PR #2327, opened 2020-05-21
Order Details: HTML in the shipping method is now decoded

### LINKED ISSUE #1873: Order Details: HTML in the shipping method is not decoded
**Describe the bug**

Similar to https://github.com/woocommerce/woocommerce-ios/issues/1808 and https://github.com/woocommerce/woocommerce-ios/issues/1871, we probably want to decode the HTML in the shipping method data from the API.

**To Reproduce**

Prerequisite:
Create an order with a shipping method that contains some special characters (e.g. "Priority Mail Flat Rate®")

- Go to the Orders tab
- Tap on the Order that was created in the prerequisite step
- Note the shipping method field under CUSTOMER > Shipping Method

**Screenshots**

**Expected behavior**

**Mobile Environment**
Please include:
- Device: any
- iOS version: iOS 12.1/13.3
- WooCommerce iOS version: 3.6.0.2

pmusolino:
Fixes #1873

## Description
When a shipping method In Order Detail and Order Fulfillment contains something commonly escaped in HTML (like an ampersand &) it's showing up as an HTML entity.

## Testing
- Open an Order Detail, with a shipping method which contains HTML -> Now should show the correct string.
- Open an Order Fulfillment, with a shipping method which contains HTML -> Now should show the correct string.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -492,7 +492,7 @@ private extension OrderDetailsDataSource {
     private func configureShippingMethod(cell: CustomerNoteTableViewCell) {
         cell.headline = NSLocalizedString("Shipping Method",
                                           comment: "Shipping method title for customer info cell")
-        cell.body = shippingMethod
+        cell.body = shippingMethod.strippedHTML
         cell.selectionStyle = .none
     }
 
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,8 @@
+4.4
+-----
+- Order Detail: the HTML shipping method is now showed correctly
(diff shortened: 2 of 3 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -492,7 +492,7 @@ private extension OrderDetailsDataSource {
     private func configureShippingMethod(cell: CustomerNoteTableViewCell) {
         cell.headline = NSLocalizedString("Shipping Method",
                                           comment: "Shipping method title for customer info cell")
-        cell.body = shippingMethod
+        cell.body = shippingMethod.strippedHTML
```
shiki, line 495: I like that we do this on the UI side. 👍

shiki (approved):
👏 :shipit:

PR #2328, opened 2020-05-21
Add unit tests for filter product list view model

### LINKED ISSUE #2037: Product List: filter products by stock status, product status, and product type
Prerequisites:
- [x] Update existing `ListSelectorViewController` and its data source (renamed to command soon) to enable usage in the filters UI/UX https://github.com/woocommerce/woocommerce-ios/pull/2195

Subtasks:
- [x] Filters screen https://github.com/woocommerce/woocommerce-ios/pull/2204
- [x] Handle tapping on each filter row and changing the filter value (UI only) https://github.com/woocommerce/woocommerce-ios/pull/2204
- [x] Update products tab predicate based on filters https://github.com/woocommerce/woocommerce-ios/pull/2209
- [x] Sync products by each filter via Yosemite https://github.com/woocommerce/woocommerce-ios/pull/2209
- [x] Show discard changes action sheet when dismissing with outstanding changes on the filters https://github.com/woocommerce/woocommerce-ios/pull/2270
- [x] Update `ListSelectorViewController`'s table view from xib to code to enable DI'ing the table view style since it's set in initialization https://github.com/woocommerce/woocommerce-ios/pull/2239
- [x] Update bottom CTA container background color so that it looks right in Light mode https://github.com/woocommerce/woocommerce-ios/pull/2271
- [x] Update the number of filtered products in the "Filter" CTA (like "Filter (2)") in https://github.com/woocommerce/woocommerce-ios/issues/2035 if a filter option is selected https://github.com/woocommerce/woocommerce-ios/pull/2246
- [ ] Unit tests for `FilterListViewController` and its view model https://github.com/woocommerce/woocommerce-ios/pull/2328
- [ ] Reset sort order & filters on default site change?

Filter options:
- stock status
- product status
- product type

Screen Shot 2020-03-25 at 10 57 29 AM

jaclync:
Fixes #2037 

## Changes

Most of the logic and nested structs are private in `FilterListViewController`, please lemme know if you think of any ways to test the view controller interactions!

- Added unit tests for `FilterTypeViewModel` creation from `FilterProductListViewModel.ProductListFilter` in `FilterProductListViewModelProductListFilterTests`
- Added unit tests for `FilterProductListViewModel`'s criteria calculation in `FilterProductListViewModelTests`

## Testing

no user-facing changes, just CI!

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
@@ -146,7 +146,7 @@ final class FilterListViewController<ViewModel: FilterListViewModel>: UIViewCont
 
     // MARK: - Navigation
     //
-    @objc func filterActionButtonTapped() {
+    @objc private func filterActionButtonTapped() {
         dismiss(animated: true) { [weak self] in
             guard let self = self else {
                 return
@@ -156,7 +156,7 @@ final class FilterListViewController<ViewModel: FilterListViewModel>: UIViewCont
         }
     }
 
-    @objc func dismissButtonTapped() {
+    @objc private func dismissButtonTapped() {
         if hasFilterChanges() {
             UIAlertController.presentDiscardChangesActionSheet(viewController: self, onDiscard: { [weak self] in
                 self?.dismiss(animated: true) {}
@@ -169,7 +169,7 @@ final class FilterListViewController<ViewModel: FilterListViewModel>: UIViewCont
         }
     }
 
-    @objc func clearAllButtonTapped() {
+    @objc private func clearAllButtonTapped() {
         viewModel.clearAll()
         listSelectorCommand.data = viewModel.filterTypeViewModels
         updateUI(numberOfActiveFilters: viewModel.filterTypeViewModels.numberOfActiveFilters)
--- a/WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
@@ -3,23 +3,38 @@ import XCTest
 
 final class FilterProductListViewModelTests: XCTestCase {
     func testCriteriaWithDefaultFilters() {
+        // Given
         let filters = FilterProductListViewModel.Filters()
+
+        // When
         let viewModel = FilterProductListViewModel(filters: filters)
+
+        // Then
         let expectedCriteria = FilterProductListViewModel.Filters(stockStatus: nil, productStatus: nil, productType: nil, numberOfActiveFilters: 0)
         XCTAssertEqual(viewModel.criteria, expectedCriteria)
     }
 
     func testCriteriaWithNonNilFilters() {
+        // Given
         let filters = FilterProductListViewModel.Filters(stockStatus: .inStock, productStatus: .draft, productType: .grouped, numberOfActiveFilters: 3)
+
+        // When
         let viewModel = FilterProductListViewModel(filters: filters)
+
+        // Then
         let expectedCriteria = filters
         XCTAssertEqual(viewModel.criteria, expectedCriteria)
     }
 
     func testCriteriaAfterClearingAllNonNilFilters() {
+        // Given
         let filters = FilterProductListViewModel.Filters(stockStatus: .inStock, productStatus: .draft, productType: .grouped, numberOfActiveFilters: 3)
+
+        // When
         let viewModel = FilterProductListViewModel(filters: filters)
         viewModel.clearAll()
+
+        // Then
         let expectedCriteria = FilterProductListViewModel.Filters(stockStatus: nil, productStatus: nil, productType: nil, numberOfActiveFilters: 0)
         XCTAssertEqual(viewModel.criteria, expectedCriteria)
     }
--- a/WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
@@ -1,0 +1,26 @@
+import XCTest
+@testable import WooCommerce
+
+final class FilterProductListViewModelTests: XCTestCase {
+    func testCriteriaWithDefaultFilters() {
+        let filters = FilterProductListViewModel.Filters()
+        let viewModel = FilterProductListViewModel(filters: filters)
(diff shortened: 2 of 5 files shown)


WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
```diff
@@ -146,7 +146,7 @@ final class FilterListViewController<ViewModel: FilterListViewModel>: UIViewCont
 
     // MARK: - Navigation
     //
-    @objc func filterActionButtonTapped() {
+    @objc private func filterActionButtonTapped() {
```
shiki, line 149: 🎉

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
```diff
@@ -0,0 +1,26 @@
+import XCTest
+@testable import WooCommerce
+
+final class FilterProductListViewModelTests: XCTestCase {
+    func testCriteriaWithDefaultFilters() {
+        let filters = FilterProductListViewModel.Filters()
+        let viewModel = FilterProductListViewModel(filters: filters)
+        let expectedCriteria = FilterProductListViewModel.Filters(stockStatus: nil, productStatus: nil, productType: nil, numberOfActiveFilters: 0)
+        XCTAssertEqual(viewModel.criteria, expectedCriteria)
+    }
```
shiki, line 10: Even if we don't use the Given-When-Then comments, I think it's still a good idea to space these out so we can easily identify which are the actions and assertions. 

```suggestion
    func testCriteriaWithDefaultFilters() {
        let filters = FilterProductListViewModel.Filters()

        let viewModel = FilterProductListViewModel(filters: filters)

        let expectedCriteria = FilterProductListViewModel.Filters(stockStatus: nil, productStatus: nil, productType: nil, numberOfActiveFilters: 0)
        XCTAssertEqual(viewModel.criteria, expectedCriteria)
    }
```

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
```diff
@@ -0,0 +1,26 @@
...
+
+    func testCriteriaWithNonNilFilters() {
+        let filters = FilterProductListViewModel.Filters(stockStatus: .inStock, productStatus: .draft, productType: .grouped, numberOfActiveFilters: 3)
```
shiki, line 13: ⚠️ So, I changed the `numberOfActiveFilters` in this line to `100` and the test still passed! 😁 

It looks like we implement our own `Equatable` here:

https://github.com/woocommerce/woocommerce-ios/blob/eec8ec29b4cfc8e215ecd1550d6f5ee434d07dcd/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductListViewModel.swift#L124-L130

So, the `numberOfActiveFilters` was **not included**. Was there a reason that we don't include it? I don't remember. 😄 

If we should include it, in case you don't know, the Equatable can now be [generated by the compiler for us](https://useyourloaf.com/blog/how-to-get-equatable-and-hashable-for-free/).

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
```diff
@@ -0,0 +1,26 @@
+import XCTest
+@testable import WooCommerce
+
+final class FilterProductListViewModelTests: XCTestCase {
```
shiki, line 4: It looks like good coverage for `FilterProductListViewModel`. We can also add an additional test for the calculated value of `numberOfActiveFilters`. The tests only use `0` and `3`. We can add a test for `2` or `1`. 🙂

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
jaclync: ooh great catch and tip! from the post, there's a default one to compare all fields and we implement it manually only if we have a different implementation:

> Now with Swift 4.1 you can remove the boilerplate extension code. You only need to tell the compiler that you want this type to be Equatable by declaring conformance to the protocol

updated in: https://github.com/woocommerce/woocommerce-ios/pull/2328/commits/31bfd56a4f54e9352993dd303d97170feececaae

now it should fail when you change the `numberOfActiveFilters` in this line :)

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
jaclync: it's better to add the Given-When-Then comments! updated here: https://github.com/woocommerce/woocommerce-ios/pull/2328/commits/8a103d5262069c107c87077d4246d17f4f63bdbd

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
jaclync: we have unit tests on `numberOfActiveFilters` before: https://github.com/woocommerce/woocommerce-ios/blob/87760dc3e2df9c09182ef117a1111cf0241526f7/WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModel+numberOfActiveFiltersTests.swift

for `FilterProductListViewModel`, it's not impossible to test modifying the value of each filter type in `filterTypeViewModels` since each `FilterTypeViewModel`'s `selectedValue` is mutable. However, because `FilterTypeViewModel`'s data type is a protocol we have to assume the order of the filter types in `FilterProductListViewModel` (stock status, product status, product type) or do some casting and I thought it's easier to test the `numberOfActiveFilters` given `[FilterTypeViewModel]` like the `numberOfActiveFiltersTests` tests above. do you think there it's worth testing modifying the selected value of `FilterProductListViewModel`'s `filterTypeViewModels`?

jaclync:
Thank you for the suggestions @shiki ! I updated the PR and responded to your comments for another look 😄

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
shiki: Oh, I must have forgotten that `FilterProductListViewModel+numberOfActiveFilterTests` exists. That's good enough for me.

PR #2318, opened 2020-05-19
Update PList storage implementation

jaclync:
Prerequisite for #1388 

## Background

Currently, we have a `FileStorage` protocol for reading/writing data from/to local storage usually for some lightweight data (similar to the usage of `UserDefaults`). So far, our `AppSettingsStore`/`AppSettingsAction` uses a `FileStorage` implementation to read/write app settings - custom shipment provider under Order Details, and the switches under Settings > Experimental Features. We use a PList implementation `PListFileStorage` to persist the app settings, and a mock in unit tests. However, before this PR, `PListFileStorage` didn't actually have any PList logic - all it used to do was to read/write data from/to a file at a URL locally without the knowledge of PList file type. Instead, the PList logic used to be in `AppSettingsStore`. This PR moves the PList logic (encode/decode) to `PListFileStorage` so that we can use this `FileStorage` implementation for persisting push notification count in #1388 as well.

## Changes

- Updated `FileStorage` protocol so that the data in read/write interface conform to `Decodable`/`Encodable`, respectively
- Moved the PList encode/decode logic from `AppSettingsStore` (private `read`/`write` functions) to `PListFileStorage`, and updated the read/write calls in `AppSettingsStore`
- Unit tests:
  - Updated `MockInMemoryStorage` from `FileStorage` protocol changes
  - Updated `AppSettingsStoreTests` to reuse `MockInMemoryStorage` instead of a very similar implementation `MockFileLoader`
  - Updated `FileStorageTests`

## Testing

The app should behave as before. The affected parts are custom carrier in shipment tracking (an extension), and feature switches under Settings > Experimental Features (stats v3/4 and Products). These are pretty well unit tested, and I'd suggest just sanity checking the following:

### Feature switches

Prerequisite: the site has WC 4+ (so that stats v4 is supported)

- Tap on the settings icon ⚙️ then "Experimental Features" row
- Toggle both switches
- Relaunch the app
- Tap on the settings icon ⚙️ then "Experimental Features" row --> the switches should have the toggled values set right before the relaunch

### Custom shipping provider

Prerequisite: the site has the shipment tracking extension

- Go to the Orders tab
- Tap on an order
- Tap "Add Tracking" row under "OPTIONAL TRACKING INFORMATION" section
- Tap the shipping carrier row
- Tap "Custom Carrier" row under "CUSTOM" section
- Type in any name and tracking number, then save by tapping "Add" to go back to the order details
[...]

Issue #1687, opened 2019-12-31
Flaky unit tests when running them locally


jaclync:
### 2022 March 2

Related: https://github.com/woocommerce/woocommerce-ios/issues/6252 / p1645776286144959-slack-C6H8C3G23
- [ ] `DefaultImageServiceTests.testDownloadingAndCachingAndRetrievingAnImageForImageView` Assertion Failure at DefaultImageServiceTests.swift:101: XCTAssertNotNil failed / Assertion Failure at DefaultImageServiceTests.swift:105: XCTAssertNotNil failed
- [ ] `DefaultImageServiceTests.testDownloadingAndCachingAndRetrievingAnImageForImageViewFromURLWithSpecialChars` Assertion Failure at DefaultImageServiceTests.swift:137: XCTAssertNotNil failed / Assertion Failure at DefaultImageServiceTests.swift:141: XCTAssertNotNil failed 
- [x] `FeeLineDetailsViewModelTests.test_view_model_creates_fee_line_with_percentage_amount` Assertion Failure at FeeLineDetailsViewModelTests.swift:151: XCTAssertEqual failed: ("Optional("0")") is not equal to ("Optional("20")") - fixed in https://github.com/woocommerce/woocommerce-ios/pull/6326
- [x] `FeeLineDetailsViewModelTests.test_view_model_disables_done_button_for_matched_percentage_value` Assertion Failure at FeeLineDetailsViewModelTests.swift:112: XCTAssertTrue failed - fixed in https://github.com/woocommerce/woocommerce-ios/pull/6326
- [x] `FeeLineDetailsViewModelTests.test_view_model_disables_done_button_for_prefilled_data_and_enables_with_changes` Assertion Failure at FeeLineDetailsViewModelTests.swift:86: XCTAssertTrue failed / Assertion Failure at FeeLineDetailsViewModelTests.swift:90: XCTAssertTrue failed - fixed in https://github.com/woocommerce/woocommerce-ios/pull/6326
- [x] `FeeLineDetailsViewModelTests.test_view_model_prefills_input_data_correctly` Assertions: Assertion Failure at FeeLineDetailsViewModelTests.swift:50: XCTAssertEqual failed: ("$0.00") is not equal to ("$15.30") - fixed in https://github.com/woocommerce/woocommerce-ios/pull/6326
[...]

PR #1672, opened 2019-12-27
Update Networking model remote ID type from `Int` to `Int64`

jaclync:
Fixes #1667

## Changes

(More context regarding why this PR is in #1667)

- Updated remote ID type from `Int` to `Int64` (thanks to AppCode's Replace in Path)
- Renamed `noteId` to `noteID` for consistent naming
- Fixed build errors and unit tests from the above changes
- Removed unnecessary `Int64` conversions
- For decoding a JSON payload into a dictionary `[String: AnyCodable]`, let's keep `Int` usage because iOS seems to default a number to `Int` type https://github.com/woocommerce/woocommerce-ios/commit/1e59c114b60ff723b8b4819f60abb7a64e5ba3b6

## Testing

- Log out from the app
- Log in to the app --> make sure the app runs normally and all the data are fetched correctly

Update release notes:

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2330, opened 2020-05-21
Improvement to FileStorageTests: managed URLs that never exists

pmusolino:
Part of #1687. Discussions started here https://github.com/woocommerce/woocommerce-ios/pull/1672#issuecomment-569953807 and here https://github.com/woocommerce/woocommerce-ios/pull/2318#discussion_r427922592

Since URLs are the preferred way to refer to local files, but also to external resources, I modified how `FileStorageTests` manage non-existing URLs, pointing to a non-existing path.
Previously, a URL that points to a nonexisting website in some cases will be rendered because locally the router redirects us to the web page of the internet provider (after the redirect, the website exist and there is a file to download, so tests fail).

## Testing
- Just CI
- Look at the code

Update release notes:

### DIFF
--- a/Storage/StorageTests/Tools/FileStorageTests.swift
+++ b/Storage/StorageTests/Tools/FileStorageTests.swift
@@ -25,23 +25,23 @@ final class FileStorageTests: XCTestCase {
     }
 
     func testErrorIsTriggeredWhenFileFailsToLoad() {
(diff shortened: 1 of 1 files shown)


jaclync (approved):
Thanks for fixing the flaky tests from the semi-non existent URL! :shipit:

PR #2350, opened 2020-05-26
Order Detail: fixed Order Status Cell wrong cell appearance

### LINKED ISSUE #2256: Order Detail: Order Status Cell wrong cell appearance
Navigating to an Order Detail -> Order Status, and selecting the last cell, the end separator is not shown, so it seems that the cell its part of the background.

## Mobile Environment
- Device: iPhone Xs simulator
- iOS version: 12.2
- WooCommerce iOS version: 4.2

pmusolino:
Fixes #2256 

## Description
Navigating to an Order Detail -> Order Status, and selecting the last cell, the end separator is not shown, so it seems that the cell it's part of the background.
The fix suggested by @Garance91540, is to not change the background color of the cell when it's selected, like in other sections of the app.

## Testing
- Navigate to an order detail
- Go to the order status screen
- Select one of the cells
- Make sure the background is still white.

## Screenshots
| Before           |  After |  After (Dark) |
81335878-9a8c1a80-90a8-11ea-9b8b-269a8f9bcfce | Simulator Screen Shot - iPhone 11 Pro - 2020-05-26 at 12 10 25 | Simulator Screen Shot - iPhone 11 Pro - 2020-05-26 at 12 18 07

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,7 @@
 - [*] The product name is now shown in the product details navigation bar so that the name is always visible.
 - [*] The images pending upload should be visible after editing product images from product details.
 - [*] The discard changes prompt does not appear when navigating from product settings detail screens with a text field (slug, purchase note, and menu order) anymore.
+- [*] Fix the wrong cell appearance in the order status list.
 
 
 4.3
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
@@ -233,6 +233,7 @@ extension OrderStatusListViewController: UITableViewDataSource {
 
         let status = statusResultsController.object(at: indexPath)
         cell.textLabel?.text = status.name
+        cell.selectionStyle = .none
 
         return cell
     }


Garance91540 (approved):
👏

jaclync (approved):
:shipit: 

Code: ✅ 
Testing: ✅ 

Unrelated to the PR, it'd be nice for the cell to have a highlighted state. before this PR, it has the dimmed background color but also the checkmark (like screenshot below), it'd be nice to have just the dimmed background color when the cell is highlighted:

pmusolino:
> Unrelated to the PR, it'd be nice for the cell to have a highlighted state. before this PR, it has the dimmed background color but also the checkmark (like screenshot below), it'd be nice to have just the dimmed background color when the cell is highlighted

I believe in this case we need to change the background color of the tableview, since the color of the selected cell has the same color currently used by the background.
cc @Garance91540

Garance91540:
> Unrelated to the PR, it'd be nice for the cell to have a highlighted state. before this PR, it has the dimmed background color but also the checkmark (like screenshot below), it'd be nice to have just the dimmed background color when the cell is highlighted

> I believe in this case we need to change the background color of the tableview, since the color of the selected cell has the same color currently used by the background.
cc @Garance91540

Hi @jaclync and @pmusolino ! 

For more context, we did have a different background color on row selection for this screen but it blended with the screen background color when the last row was selected. I noticed we were implementing it this way on [new Products screens](https://d.pr/i/eAjvQi), and it seems clear enough to me so that's why I think we're good with just the checkmark icon to convey selection.

Issue #2198, opened 2020-04-27
Products: change the product type from simple to variant, grouped or external product


pmusolino:
As part of Products M3, we want to offer the possibility to change the product type from simple to variant, grouped, or external product. 

- [ ] When editing `ProductFormActionsFactory`, let's also update the enabled features to be an array of feature flags like discussion in https://github.com/woocommerce/woocommerce-ios/pull/2357#pullrequestreview-421995127
 
Feel free to divide this issue into more detailed subtasks.

<img src="https://user-images.githubusercontent.com/495617/80390667-07cfcc80-88ad-11ea-81b2-d5b3be06305e.png" width=400 />

jaclync:
the product type row is added in https://github.com/woocommerce/woocommerce-ios/issues/2262 - this issue implements the update action that sets the product's `productType` and also updates the available fields in the product form

PR #2357, opened 2020-05-28
Edit Products: consolidate logic for visible actions in primary section, settings section, and bottom sheet

### LINKED ISSUE #2317: Edit Products: A better way to create product settings rows and bottom sheet actions
Right now, we just have a thin logic `isVisible(product:)` on `ProductFormBottomSheetAction`/`ProductFormSection.SettingsRow` to decide which settings rows / bottom sheet actions are to be displayed. It might be nice to have a factory that creates both sets of product editing actions since they are based on the same logic whenever the product is updated.

jaclync:
Fixes #2317

- [x] ⚠️ Update PR base to `develop` after https://github.com/woocommerce/woocommerce-ios/pull/2348 is merged ⚠️ 

## Background

Apology for yet another big PR 🙇🏻‍♀️ I wasn't sure how to break this down since lots of the diffs are from the updates on the unit tests from the factory changes (maybe 66% of the diffs are from unit tests, mostly just updating the interface). Please lemme know if you have any ideas!

The goal of this PR is to centralize the logic for the visible "edit actions" on the product form for editing different product properties since we have custom logic for when an action is visible and available in each UI.

For example, before this PR, the visible rows in the product form settings section (e.g. price settings, inventory settings) are [determined in `DefaultProductFormTableViewModel`](https://github.com/woocommerce/woocommerce-ios/blob/9420a214faa675cd5eb5d36de1a4f6ac509f8cfa/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/DefaultProductFormTableViewModel.swift#L52-L66), while their complementary actions in the bottom sheet are [calculated in `ProductFormBottomSheetActionsFactory`](https://github.com/woocommerce/woocommerce-ios/blob/9420a214faa675cd5eb5d36de1a4f6ac509f8cfa/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/BottomSheetListSelector/ProductFormBottomSheetActionsFactory.swift#L16-L27).

As we are supporting more product types in Products M3/M4, it would help a lot for unit testing and development if we can centralize the calculation of visible actions for different parts of product form UI. This PR converted the existing `ProductFormBottomSheetActionsFactory` to `ProductFormActionsFactory` that can be reinitialized on product changes and provides the actions for product form UI (primary section, settings section, bottom sheet for now).

## Changes

- Created a new enum `ProductFormEditAction` for each editable subset of product attributes from the product form UI. Converted `ProductFormBottomSheetActionsFactory` to `ProductFormActionsFactory` that takes in a product and M2/M3 switches and determines which actions are visible in the primary section, settings section, and bottom sheet.
- Added an `init` for `ProductFormBottomSheetAction` enum to optionally initialize from a `ProductFormEditAction` enum
- In `DefaultProductFormTableViewModel`, removed the rows calculation and created the rows based on the actions from `ProductFormActionsFactory`
- In `ProductFormViewController`, updated the usage of `DefaultProductFormTableViewModel` and `ProductFormActionsFactory`
- Lots of updates to the existing unit tests, mostly just replacing with the new interface

## Testing

The only visual change is the order in the product form bottom sheet - in the design, the shipping row goes before inventory row in the bottom sheet but it's the reverse order in the product details screen:

bottom sheet | product details
-- | --
Screen Shot 2020-05-28 at 1 50 40 PM | Screen Shot 2020-05-28 at 1 50 50 PM

@Garance91540 I have 2 questions here:
- do you know if this ordering difference is intentional? In this PR, I updated the order in the bottom sheet to follow the same as on the product details screen (shipping before inventory) but I can update if the original order is intentional.
- when do we hide the inventory row from the product details screen? right now, the row is visible whether stock management is enabled at the product level. If the stock management is enabled for the product, do we hide it when the stock quantity is 0?

The actions are pretty well tested, but we can also ensure the behavior with the following steps:

### Products M2 switch is off

- Go to Settings > Experimental Features and make sure the Products switch is off
- Go to the Products tab
- Tap on a simple physical Product with inventory settings but without shipping information (weight & dimensions) --> should see a "+ Add more details" CTA at the bottom
- Tap on the "+ Add more details" CTA at the bottom --> should see a list selector with an action to edit shipping settings
- Tap on the shipping row in the CTA to edit shipping settings
- Enter some values to the weight and dimension fields, then tap "Done" --> it should go back to the product details without the "+ Add more details" CTA at the bottom
- Feel free to test opening other simple products, and the "+ Add more details" CTA at the bottom should only be shown if there are any missing product fields 

### Products M2 switch is on

- Go to Settings > Experimental Features and make sure the Products switch is on
- Go to the Products tab
- Tap on a simple physical Product with inventory settings but without shipping information (weight & dimensions) --> should see a "+ Add more details" CTA at the bottom
- Tap on the "+ Add more details" CTA at the bottom --> should see a list selector with actions to edit shipping settings, short description (if empty), and categories (if empty)
- Feel free to test opening other simple products, and the "+ Add more details" CTA at the bottom should only be shown if there are any missing product fields

Update release notes:


loremattei:
Hey! I'm moving this to 4.5 because 4.4 has been cut. If you want this to make it to 4.4, please feel free to ping me.

pmusolino:
On it now ;)

jaclync:
thanks for the review Paolo!

> Since we are working with feature flags that appears and disappears pretty quick (in theory we need to drop the M2 feature flag from the next release), it will be good if the future our methods accept an array of feature flags. The method should just check if a specific feature flag is present and enabled. This should allow us to change methods that contain feature flags less often, and to complete all the milestones of a feature without affecting the history of a function. Do you think that makes sense?

yea having to update the feature flags is a bit cumbersome and adds to the git history. I like the idea of passing an array of feature flags that are enabled (so the check could be like `featureFlags.contains(.editProductsRelease2)` for whether M2 flag is enabled). I added this subtask to the the issue description https://github.com/woocommerce/woocommerce-ios/issues/2198 when this factory is most likely updated to support different product types, and I'll also remember to work on this if I change this class when working on other issues 🙂

jaclync:
I'm merging this PR now since there are many changes, but @Garance91540 lemme know if you have any thoughts about these 2 questions (they are pretty minor) 😄 

> 2 questions here:
> - do you know if this ordering difference is intentional? In this PR, I updated the order in the bottom sheet to follow the same as on the product details screen (shipping before inventory) but I can update if the original order is intentional.
bottom sheet | product details
-- | --
Screen Shot 2020-05-28 at 1 50 40 PM | Screen Shot 2020-05-28 at 1 50 50 PM
> - when do we hide the inventory row from the product details screen? right now, the row is visible whether stock management is enabled at the product level. If the stock management is enabled for the product, do we hide it when the stock quantity is 0?

Garance91540:
Sorry @jaclync  I missed this! As summarized in [this post](https://woomobilep2.wordpress.com/2020/06/03/products-product-types-updates-and-final-states-design/), I'm suggesting to display by default the Inventory row so users can quickly address the products out of stock.

So we would take it out from the bottom sheet, and then the order you have on the product page looks right.

PR #2358, opened 2020-05-28
Setup crash logging on Sentry before main window is setup

pmusolino:
After the discussion here p91TBi-2PM it turned out that that our crashes triggered by setupMainWindow() aren't logged on Sentry because `setupCrashLogging()` is called after 
`setupMainWindow()` and `setupComponentsAppearance()`.

Update release notes:


jaclync (approved):
LGTM! I haven't got to get the crash log showing up on Sentry yet but this looks good for the first step in any case 👍

shiki:
Related to #2371

Issue #1543, opened 2019-12-02
Invalid Number Of Rows/Sections Crash


astralbodies:
**Describe the bug**
Text from one of the crashes:

```Invalid update: invalid number of rows in section 0. The number of rows contained in an existing section after the update (8) must be equal to the number of rows contained in that section before the update (7), plus or minus the number of rows inserted or deleted from that section (1 inserted, 7 deleted) and plus or minus the number of rows moved into or out of that section (0 moved in, 0 moved out).```

Crash report: https://sentry.io/share/issue/63aa8ffcc1634d5da1c7c2c0f8155ac8/

**To Reproduce**
Unsure of how to reproduce.

**Expected behavior**
No crashes when orders list is updated.

astralbodies:
FWIW the ghost feature caused a crash in WPiOS that might be similar in nature. https://github.com/wordpress-mobile/WordPress-iOS/pull/12974

pmusolino:
I'm reopening this issue since it occurs also on release 3.5

shiki:
Moved to _Waiting_ for now. We'll observe how #2361 and #2362 will affect this.

shiki:
Looks like this is still happening in 4.5 so #2447 may not actually be enough. 🙁

pmusolino:
@shiki I'm thinking if it's possible that there is a case which we haven't considered, where the number of sections returned for a table view is 0, that is a value not allowed (there must always be a section, and the minimum number of sections allowed is 1)! For example, it seems that the number of sections returned by a `ResultsController` permit to return an empty array, and when we call the method `.count` on it, this will return 0. 💥

shiki:
That's a good idea to look into. We might be able to easily simulate that scenario and see if it crashes by making a new test case [like this one in ResultsControllerUIKitTests](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/WooCommerceTests/Tools/ResultsControllerUIKitTests.swift#L210). 

The other thing I totally forgot about is that some of our ViewControllers are still using the Ghost table `dataSource` switching which we've proven before that it can cause "invalid number of rows" crashes. (https://github.com/woocommerce/woocommerce-ios/issues/2448)

shiki:
Suggestion by Apple was implemented in Simplenote https://github.com/Automattic/simplenote-ios/pull/798. Ref p91TBi-2Yr-p2#comment-1239

shiki:
[...]

PR #2360, opened 2020-05-28
Add Additional Tests for ResultsController

shiki:
This just adds an additional test in `ResultsControllerUIKitTests`. I was using this to find the cause of #1543. I thought it would be a good permanent addition.  

## MockupTableView

I changed the `MockupTableView` so that it will always call `super` instead of just absorbing everything. This makes it closer to production behavior.

## DataSource

I also updated the `ResultsControllerUIKitTests` so it is a `UITableViewDataSource`. Again, this is to simulate how things work in production. 

## Testing

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/WooCommerceTests/Tools/ResultsControllerUIKitTests.swift
+++ b/WooCommerce/WooCommerceTests/Tools/ResultsControllerUIKitTests.swift
@@ -23,8 +23,8 @@ final class ResultsControllerUIKitTests: XCTestCase {
     // MARK: - Overridden Methods
 
     override func setUp() {
+        super.setUp()
         storageManager = MockupStorageManager()
-        tableView = MockupTableView()
 
         resultsController = {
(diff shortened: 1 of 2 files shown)


WooCommerce/WooCommerceTests/Tools/ResultsControllerUIKitTests.swift
```diff
@@ -38,13 +44,17 @@ class ResultsControllerUIKitTests: XCTestCase {
             )
         }()
 
+        tableView = MockupTableView()
+        tableView.dataSource = self
+
         resultsController.startForwardingEvents(to: tableView)
-        try? resultsController.performFetch()
+        try! resultsController.performFetch()
```
jaclync, line 51: just curious, is this change intentional? I tried the old diff and the tests passed as well

PR #2361, opened 2020-05-28
Use NSFetchedResultsSectionInfo.numberOfObjects Instead of objects.count

### LINKED ISSUE #1543: Invalid Number Of Rows/Sections Crash

shiki:
This is just an attempt in fixing #1543. 

## Findings

We currently use `objects.count` when determining the number of rows in the section. 

https://github.com/woocommerce/woocommerce-ios/blob/9420a214faa675cd5eb5d36de1a4f6ac509f8cfa/Yosemite/Yosemite/Tools/ResultsController.swift#L266-L270

I found from these resources that `objects.count` can sometimes have a different value compared to [`numberOfObjects`](https://developer.apple.com/documentation/coredata/nsfetchedresultssectioninfo/1622289-numberofobjects):

- http://mikeabdullah.net/nsfetchedresultscontroller.html
- https://stackoverflow.com/a/39232007
- https://stackoverflow.com/q/3967895

Sadly, I have not been able to prove that this is the cause of the crash. 😞 

The [official Apple doc](https://developer.apple.com/documentation/coredata/nsfetchedresultscontroller) also shows that `numberOfObjects` should be used.

## Changes

This changes `SectionInfo` to use `numberOfObjects` instead of `objects.count`. I also made it store the `NSFetchedResultsSectionInfo` internally so `SectionInfo` would act more like just a wrapper. 

While we can't really prove if this fixes #1543, this change at least eliminates some _possible_ causes that we should look into. We can now ignore this part and move on.

## Testing

Please do a regression test that this does not affect the table view loading. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/Yosemite/Yosemite/Tools/ResultsController.swift
+++ b/Yosemite/Yosemite/Tools/ResultsController.swift
@@ -257,7 +257,7 @@ public extension ResultsController {
 
     // MARK: - ResultsController.SectionInfo
     //
-    class SectionInfo {
+    final class SectionInfo {
 
         /// Name of the section
         ///
--- a/Yosemite/Yosemite/Tools/ResultsController.swift
+++ b/Yosemite/Yosemite/Tools/ResultsController.swift
@@ -259,6 +259,10 @@ public extension ResultsController {
     //
     final class SectionInfo {
 
+        /// The real SectionInfo that we're hiding.
+        ///
+        private let mutableSectionInfo: NSFetchedResultsSectionInfo
+
         /// Name of the section
         ///
         public let name: String
@@ -284,6 +288,8 @@ public extension ResultsController {
         /// Designated Initializer
         ///
         init(mutableSection: NSFetchedResultsSectionInfo) {
+            mutableSectionInfo = mutableSection
+
             name = mutableSection.name
             mutableObjects = mutableSection.objects as? [T] ?? []
         }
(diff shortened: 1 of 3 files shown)


Yosemite/Yosemite/Tools/ResultsController.swift
```diff
@@ -256,36 +256,45 @@ public extension ResultsController {
...
-        /// Array of Mutable Objects!
-        ///
-        private let mutableObjects: [T]
+            guard let objects = mutableSectionInfo.objects else {
+                return []
+            }
+            guard let castedObjects = objects as? [T] else {
+                assertionFailure("Failed to cast objects into an array of \(T.self)")
+                return []
+            }
```
jaclync, line 289: nit: would it also work by combining these 2 guard statements like:

```
private(set) public lazy var objects: [T.ReadOnlyType] = {
        guard let objects = mutableSectionInfo.objects as? [T] else {
            return []
        }
        return objects.map { $0.toReadOnly() }
}()
```

PR #2376, opened 2020-06-01
Add More ResultsController+UIKit Tests

shiki:
I added more test cases based on some of our assumptions.

### Test That `ResultsController` Will Emit a `move` `ChangeType`

https://github.com/woocommerce/woocommerce-ios/blob/5dd0dc967c7e5042dfd442496325ff4f0dae0262/WooCommerce/WooCommerceTests/Tools/ResultsControllerUIKitTests.swift#L259-L264

### Test That the `fixedType` fix is Still Necessary

https://github.com/woocommerce/woocommerce-ios/blob/5dd0dc967c7e5042dfd442496325ff4f0dae0262/WooCommerce/WooCommerceTests/Tools/ResultsControllerUIKitTests.swift#L317-L333

## Other Changes

I added `UIWindow` creations in the test cases. Without it, the `UITableView` will not actually crash if it ends up with an inconsistent state. The `UITableView` checks whether it is visible before doing any cell change. 

https://github.com/woocommerce/woocommerce-ios/blob/5dd0dc967c7e5042dfd442496325ff4f0dae0262/WooCommerce/WooCommerceTests/Tools/ResultsControllerUIKitTests.swift#L337-L338

## Testing

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/WooCommerceTests/Testing/XCTestCase+Assertions.swift
+++ b/WooCommerce/WooCommerceTests/Testing/XCTestCase+Assertions.swift
@@ -7,7 +7,7 @@ import XCTest
 extension XCTestCase {
     /// Asserts that an array is empty.
     ///
-    func assertEmpty<Element>(_ array: [Element]) {
-        XCTAssertTrue(array.isEmpty)
+    func assertEmpty<Element>(_ array: [Element], file: StaticString = #file, line: UInt = #line) {
+        XCTAssertTrue(array.isEmpty, "Expected array \(array) to be empty.", file: file, line: line)
     }
 }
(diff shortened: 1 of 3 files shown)


WooCommerce/WooCommerceTests/Testing/XCTestCase+Assertions.swift
```diff
@@ -0,0 +1,13 @@
+
+import Foundation
+import XCTest
+
+/// Additional assertions to help with readability in tests.
+///
+extension XCTestCase {
+    /// Asserts that an array is empty.
+    ///
+    func assertEmpty<Element>(_ array: [Element], file: StaticString = #file, line: UInt = #line) {
```
jaclync, line 10: this looks helpful 👍

PR #2381, opened 2020-06-03
Remove Some Unused Code

shiki:
While working on #2281, I found that these do not seem to be used anymore.

## Testing

1. Navigate to Orders 
2. Pick an order that has a refund
3. Confirm that you can still the product details
4. Navigate back to Orders
5. Pick a processing order.
6. Tap on Begin Fulfillment
7. Confirm that you can still see the product details.

## Reviewing 

Only 1 reviewer is needed but anyone can review. 

## Submitter Checklist

- [x] If it's feasible, I have added unit tests. 
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #2281, opened 2020-05-13
Order Details → Show Price on the Right Instead of Quanity


shiki:
![image](https://user-images.githubusercontent.com/198826/81869745-d0f6e900-9531-11ea-881d-c5cc9b1aa855.png)

- [x] Subtitle: Show "quantity x price" 
- [x] Right column: Price. This makes it easier to quickly match and add up the amount of each product with the payment card below, where all the amounts are right-aligned as well.
- [x] Header should use a white background
- [x] “Products” should just be “Product” if there is only one product. https://github.com/woocommerce/woocommerce-ios/pull/2383#issuecomment-638757751
- [x] Remove SKU from Products in Order Details

Ref p91TBi-2CE-p2#comment-627

PR #2383, opened 2020-06-03
Order Details → Show Item Total on the Right Instead of Quantity

shiki:
⚠️ Requires #2381 to be merged first.

Refs #2281. 

### Order Details

This swaps the quantity shown on the right side with the item total. The header also now just says “Products” instead of “Product” and “QTY” on the right. According to #2281, the header should be using a white background. I'll get to that in a follow-up PR. 

There are no changes in the layout. 

@Garance91540 Should the font sizes of the SKU and "Quantity x Price" be bigger? I couldn't find an existing style of that in the app so I wasn't sure. 

Before | After 
--------|-------
 image    |       image

After (Dark) | After (Large)
--------|-------
Simulator Screen Shot - iPhone 11 Dark - 2020-06-03 at 13 18 50        |     image

### Refunded Products

Since the Refunded Products view is using the same components as the Order Details, it is also affected by the changes too. 

Before | After 
--------|-------
image        |     image

@Garance91540 Please let me know if this is okay or not. 

## Testing

1. Navigate to a processing Order.
2. Confirm that the price is on the right side.
3. Tap on Begin Fulfillment. 
4. Confirm that the quantity is still on the right side. 
5. Navigate to an Order with a refund. 
6. Tap on the Refunded Products
7. Confirm that the price is shown on the right. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,6 @@
 4.5
 -----
+- [*] In Order Details, the item subtotal is now shown on the right side instead of the quantity. The quantity can still be viewed underneath the product name.
  
 4.4
 -----
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -567,7 +567,7 @@ extension OrderDetailsDataSource {
                     rows.append(.details)
                 }
 
-                return Section(title: Title.product, rightTitle: Title.quantity, rows: rows)
+                return Section(title: Title.products, rightTitle: nil, rows: rows)
             }
 
             var rows = [Row]()
@@ -589,7 +589,7 @@ extension OrderDetailsDataSource {
                 return nil
             }
 
-            return Section(title: Title.product, rightTitle: Title.quantity, rows: rows)
+            return Section(title: Title.products, rightTitle: nil, rows: rows)
         }()
 
         let refundedProducts: Section? = {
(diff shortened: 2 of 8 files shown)


Garance91540:
> The header also now just says “Products” instead of “Product” and “QTY” on the right

One quick note on this: actually the desired behavior is to have the header be "Product" when there's only one product in the orders and "Products" as soon as there's more than one

shiki:
> I'm not sure why we'd want to show SKU in the product rows actually, I think it's kind of useless at this level and it overcrowds the page. What do you think?

@Garance91540 It's always been there so I didn't touch it. I personally have no idea how SKUs are used. 😬 Should it be removed?

Garance91540:
>  😬 Should it be removed?
Yep let's remove! you can see them when you select the product row

shiki:
Thank you, @jaclync. Excellent catch on the comments and method names. That was embarrassing. 🙈 Fixed in 5524620bd3d1a251d2536fcdf161e12fda3930b1, e973d123b46dc0114825693dd3cdbf4de904daca, and a349acec1967315fab82a79b8e640459b5b2046b.

@Garance91540 I'll remove the SKU on a separate PR.

designsimply:
At first, I was confused because I thought that Fulfill Order screen should also show the price and not the quantity but then I re-read the instructions and found that's the expected behavior. 🙂 I wanted to drop in some screenshots from my test for reference.

IMG_3784 IMG_3785 IMG_3786

<p style="clear:both;"><sup>Tested with Woo iOS 4.5.0.2 TestFlight beta on iPhone 6S iOS 13.3.1.</sup></p>

Issue #2000, opened 2020-03-17
Edit Products: Manage Product Categories


shiki:
<img src="https://user-images.githubusercontent.com/198826/76883532-ec9b8700-6841-11ea-89d1-8ed04ad9ae55.png" width="400">

This will need to be broken down to a list of tasks/features that should be built. 

## Task Breakdown

The task breakdown is in https://github.com/woocommerce/woocommerce-ios/issues/2000#issuecomment-602852371.

Ecarrion:
Hi @shiki and @Garance91540 could any of you point me in the direction on that figma file or any  feature spec so I can start digesting and scoping this feature?

Thanks! :-)

jaclync:
Shared the design link in DM 🙂

Garance91540:
> EDIT: @jaclync gave me the figma link but any other resource is well appreciated :-)
I shared a Figma walkthrough in DM with you Ernesto hope it helps to navigate the Figma file!

Ecarrion:
I'll be taking this issue as part of my trial 
https://woomobilep2.wordpress.com/2020/03/23/manage-product-categories/

**Development tasks**
- [ ] Add manage categories feature flag
- [ ] Update Storage layer with ProductCategory new fields
- [ ] Update Network layer with [create new ProductCategory request](https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#create-a-product-category)
- [ ] Update Network layer with [syncrhonize ProductCategories](https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#list-all-product-categories)
- [ ] Update Yosemite layer to orchestrate new Network and Storage actions
- [ ] Display product category in product screen
- [ ] Build ProductCategory selector screen
- [ ] Show action sheet before saving changes on product screen
- [ ] Build Add/Create ProductCategory Screen
- [ ] Build screen to select a parent ProductCategory
- [ ] Remove manage categories feature flag

pmusolino:
I found a small issue when the Product Categories are loading in Dark Mode:

<img src="https://user-images.githubusercontent.com/495617/80382198-48761880-88a2-11ea-890b-78d3649ed016.png" width=300 />

pmusolino:
Extra tasks:
- [ ] Manage category selection
- [ ] Category Save

PR #2326, opened 2020-05-21
Add Mocks and Test Data

shiki:
Still chugging away at #2254. It seems that we could use a lot of testing utilities. 😅 

I wanted to send this as a separate PR to control the size. My solution for #2254 including the contents of this PR was already at 650 lines and I was not even done writing unit tests. 😱 

The problem is that these are all testing utilities and are useless on their own. Let me know if that's not okay! 

## Remote Mocks

I added mocks for some of the `Remote` classes that I needed to use in a test:

- [MockNotificationsRemote](https://github.com/woocommerce/woocommerce-ios/blob/issue/2254-add-remote-mocks/Yosemite/YosemiteTests/Mockups/Networking/Remote/MockNotificationsRemote.swift)
- [MockProductReviewsRemote](https://github.com/woocommerce/woocommerce-ios/blob/issue/2254-add-remote-mocks/Yosemite/YosemiteTests/Mockups/Networking/Remote/MockProductReviewsRemote.swift)
- [MockProductsRemote](https://github.com/woocommerce/woocommerce-ios/blob/issue/2254-add-remote-mocks/Yosemite/YosemiteTests/Mockups/Networking/Remote/MockProductsRemote.swift)

These mocks allow us to provide models instead of JSON files. I believe we have enough tests for the JSON parsing so we don't really need to use JSON files for `Store` test cases. Also, JSON files are so hard to debug and maintain. 

Here is an example of how these can be used: 

https://github.com/woocommerce/woocommerce-ios/blob/7127fcb04d2d8622066f07634a4e8b3ac6a99dc0/Yosemite/YosemiteTests/Stores/ProductReview/RetrieveProductReviewFromNoteUseCaseTests.swift#L50-L62

The API is based on Mockito which sort of inspires all other test frameworks. 

## Test Data

As helpers for test cases that need to use the mocks above, I also created test data factories:

- [MockNote](https://github.com/woocommerce/woocommerce-ios/blob/issue/2254-add-remote-mocks/Yosemite/YosemiteTests/Mockups/Networking/Model/MockNote.swift)
- [MockProductReview](https://github.com/woocommerce/woocommerce-ios/blob/issue/2254-add-remote-mocks/Yosemite/YosemiteTests/Mockups/Networking/Model/MockProductReview.swift)

## Testing

These files are currently unused. Making sure the build is green should be enough. 

[...]

PR #2387, opened 2020-06-04
Product Categories: updated Network/Yosemite layers with create new ProductCategory request

pmusolino:
Part of #2000

In order to build the add product category screen, I need to do some groundwork first adding the network request for adding a new product category.

## Changes
- Added the new `ProductCategoryMapper` + tests
- Added the new method “createProductCategory` in `ProductCategoriesRemote` + tests
- Added a new mock json `category.json`
- Added `addProductCategory` action in `ProductCategoryAction` and `ProductCategoryStore` + tests

## Testing
- Look at the code
- CI

Update release notes:

### DIFF
--- a/Networking/Networking/Mapper/ProductCategoryMapper.swift
+++ b/Networking/Networking/Mapper/ProductCategoryMapper.swift
@@ -1,0 +1,13 @@
+//
+//  ProductCategoryMapper.swift
+//  Networking
+//
+//  Created by Paolo Musolino on 04/06/2020.
+//  Copyright © 2020 Automattic. All rights reserved.
+//
+
+import UIKit
+
+class ProductCategoryMapper: NSObject {
+
+}
--- a/Networking/Networking/Mapper/ProductCategoryMapper.swift
+++ b/Networking/Networking/Mapper/ProductCategoryMapper.swift
@@ -1,13 +1,38 @@
-//
-//  ProductCategoryMapper.swift
-//  Networking
-//
-//  Created by Paolo Musolino on 04/06/2020.
-//  Copyright © 2020 Automattic. All rights reserved.
-//
+import Foundation
 
-import UIKit
 
-class ProductCategoryMapper: NSObject {
+/// Mapper: ProductCategory
+///
+struct ProductCategoryMapper: Mapper {
 
+    /// Site Identifier associated to the `ProductCategory`s that will be parsed.
+    ///
+    /// We're injecting this field via `JSONDecoder.userInfo` because SiteID is not returned in any of the ProductCategory Endpoints.
+    ///
+    let siteID: Int64
+
+
+    /// (Attempts) to convert a dictionary into ProductCategory.
+    ///
+    func map(response: Data) throws -> ProductCategory {
+        let decoder = JSONDecoder()
+        decoder.userInfo = [
+            .siteID: siteID
+        ]
+
+        return try decoder.decode(ProductCategoryEnvelope.self, from: response).productCategory
+    }
+}
+
+
+/// ProductCategoryEnvelope Disposable Entity:
+/// `Load Product Category` endpoint returns the updated products document in the `data` key.
+/// This entity allows us to do parse all the things with JSONDecoder.
+///
+private struct ProductCategoryEnvelope: Decodable {
+    let productCategory: ProductCategory
+
+    private enum CodingKeys: String, CodingKey {
+        case productCategory = "data"
+    }
 }
--- a/Networking/Networking/Remote/ProductCategoriesRemote.swift
+++ b/Networking/Networking/Remote/ProductCategoriesRemote.swift
@@ -30,6 +30,33 @@ public final class ProductCategoriesRemote: Remote {
 
         enqueue(request, mapper: mapper, completion: completion)
     }
+
+    /// Create a new `ProducCategory`.
+    ///
+    /// - Parameters:
+    ///     - siteID: Site for which we'll add a new product categories.
+    ///     - pageNumber: Number of page that should be retrieved.
+    ///     - pageSize: Number of categories to be retrieved per page.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func createProductCategory(for siteID: Int64,
+                                      name: String,
+                                      parentID: Int64?,
+                                      completion: @escaping ([ProductCategory]?, Error?) -> Void) {
+        var parameters = [
+            ParameterKey.name: name
+        ]
+
+        if let parentID = parentID {
+            parameters[ParameterKey.parent] = String(parentID)
+        }
+
+        let path = Path.categories
+        let request = JetpackRequest(wooApiVersion: .mark3, method: .post, siteID: siteID, path: path, parameters: parameters)
+        let mapper = ProductCategoryListMapper(siteID: siteID)
+
+        enqueue(request, mapper: mapper, completion: completion)
+    }
 }
 
 
@@ -48,5 +75,7 @@ public extension ProductCategoriesRemote {
     private enum ParameterKey {
         static let page: String = "page"
         static let perPage: String = "per_page"
+        static let name: String = "name"
+        static let parent: String = "parent"
     }
 }
--- a/Networking/Networking/Remote/ProductCategoriesRemote.swift
+++ b/Networking/Networking/Remote/ProductCategoriesRemote.swift
@@ -42,7 +42,7 @@ public final class ProductCategoriesRemote: Remote {
     public func createProductCategory(for siteID: Int64,
                                       name: String,
                                       parentID: Int64?,
-                                      completion: @escaping ([ProductCategory]?, Error?) -> Void) {
+                                      completion: @escaping (ProductCategory?, Error?) -> Void) {
         var parameters = [
             ParameterKey.name: name
         ]
@@ -53,7 +53,7 @@ public final class ProductCategoriesRemote: Remote {
 
         let path = Path.categories
         let request = JetpackRequest(wooApiVersion: .mark3, method: .post, siteID: siteID, path: path, parameters: parameters)
-        let mapper = ProductCategoryListMapper(siteID: siteID)
+        let mapper = ProductCategoryMapper(siteID: siteID)
 
         enqueue(request, mapper: mapper, completion: completion)
     }
--- a/Networking/Networking/Remote/ProductCategoriesRemote.swift
+++ b/Networking/Networking/Remote/ProductCategoriesRemote.swift
@@ -31,12 +31,12 @@ public final class ProductCategoriesRemote: Remote {
         enqueue(request, mapper: mapper, completion: completion)
     }
 
-    /// Create a new `ProducCategory`.
+    /// Create a new `ProductCategory`.
     ///
     /// - Parameters:
     ///     - siteID: Site for which we'll add a new product categories.
-    ///     - pageNumber: Number of page that should be retrieved.
-    ///     - pageSize: Number of categories to be retrieved per page.
+    ///     - name: Category name.
+    ///     - parentID: The ID for the parent of the category.
     ///     - completion: Closure to be executed upon completion.
     ///
     public func createProductCategory(for siteID: Int64,
--- a/Networking/Networking/Remote/ProductCategoriesRemote.swift
+++ b/Networking/Networking/Remote/ProductCategoriesRemote.swift
@@ -42,7 +42,7 @@ public final class ProductCategoriesRemote: Remote {
     public func createProductCategory(for siteID: Int64,
                                       name: String,
                                       parentID: Int64?,
-                                      completion: @escaping (ProductCategory?, Error?) -> Void) {
+                                      completion: @escaping (Result<ProductCategory, Error>) -> Void) {
         var parameters = [
             ParameterKey.name: name
         ]
--- a/Yosemite/Yosemite/Actions/ProductCategoryAction.swift
+++ b/Yosemite/Yosemite/Actions/ProductCategoryAction.swift
@@ -9,6 +9,12 @@ public enum ProductCategoryAction: Action {
     /// `onCompletion` will be invoked when the sync operation finishes. `error` will be nill if the operation succeed.
     ///
     case synchronizeProductCategories(siteID: Int64, fromPageNumber: Int, onCompletion: (ProductCategoryActionError?) -> Void)
+
+    /// Create a new product category associated with a given Site ID.
+    /// `onCompletion` will be invoked when the add operation finishes. `error` will be nill if the operation succeed.
+    ///
+    case addProductCategory(siteID: Int64, name: String, parentID: Int64?, onCompletion: (ProductCategory?, Error?) -> Void)
+
 }
 
 /// Defines all errors that a `ProductCategoryAction` can return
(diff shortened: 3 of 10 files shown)


Networking/Networking/Remote/ProductCategoriesRemote.swift
```diff
@@ -30,6 +30,33 @@ public final class ProductCategoriesRemote: Remote {
...
+    /// - Parameters:
+    ///     - siteID: Site for which we'll add a new product categories.
+    ///     - name: Category name.
+    ///     - parentID: The ID for the parent of the category.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func createProductCategory(for siteID: Int64,
+                                      name: String,
+                                      parentID: Int64?,
+                                      completion: @escaping (ProductCategory?, Error?) -> Void) {
```
shiki, line 45: ⚠️ We should use `Result` now that it's available. 🙂 We added it in https://github.com/woocommerce/woocommerce-ios/pull/2261. 

```suggestion
    public func createProductCategory(for siteID: Int64,
                                      name: String,
                                      parentID: Int64?,
                                      completion: @escaping (Result<ProductCategory, Error>) -> Void) {
```

Yosemite/Yosemite/Stores/ProductCategoryStore.swift
```diff
@@ -89,6 +91,23 @@ private extension ProductCategoryStore {
         }
     }
 
+    /// Create a new product category associated with a given Site ID.
+    ///
+    func addProductCategory(siteID: Int64, name: String, parentID: Int64?, onCompletion: @escaping (ProductCategory?, Error?) -> Void) {
```
shiki, line 96: ⚠️ We should use `Result` now that's available. See https://github.com/woocommerce/woocommerce-ios/pull/2261. 

```suggestion
    func addProductCategory(siteID: Int64, name: String, parentID: Int64?, onCompletion: @escaping (Result<ProductCategory, Error>) -> Void) {
```

Yosemite/Yosemite/Actions/ProductCategoryAction.swift
```diff
@@ -9,6 +9,12 @@ public enum ProductCategoryAction: Action {
     /// `onCompletion` will be invoked when the sync operation finishes. `error` will be nill if the operation succeed.
     ///
     case synchronizeProductCategories(siteID: Int64, fromPageNumber: Int, onCompletion: (ProductCategoryActionError?) -> Void)
+
+    /// Create a new product category associated with a given Site ID.
+    /// `onCompletion` will be invoked when the add operation finishes. `error` will be nill if the operation succeed.
+    ///
+    case addProductCategory(siteID: Int64, name: String, parentID: Int64?, onCompletion: (ProductCategory?, Error?) -> Void)
```
shiki, line 16: ⚠️ Like my other comments somewhere in this PR, we should use `Result`. 🙂

Yosemite/YosemiteTests/Stores/ProductCategoryStoreTests.swift
```diff
@@ -194,6 +193,70 @@ final class ProductCategoryStoreTests: XCTestCase {
         XCTAssertNotNil(errorResponse)
     }
 
+    func testAddProductCategoryAddsStoredCategorySuccessfulResponse() {
```
shiki, line 196: As I've shared in https://github.com/woocommerce/woocommerce-ios/pull/2326, we should really consider not using JSON files in `StoreTests`. In my experience, JSON files are just so difficult to debug and maintain:

- The data to test is in a separate file (the JSON) so we'll have to go back and forth to check the data that we're testing against. 
- The JSON file is already shared with the `RemoteTests`. If we change the JSON file, this `StoreTests` will most probably break too. I believe unit tests should be as encapsulated as possible. They should maintain their own "inputs" whenever possible.
- Using simple objects in `StoreTests` gives us more flexibility than JSON. It's easier to add different combinations of test data. 

So, my suggestion would be to do the same thing as https://github.com/woocommerce/woocommerce-ios/pull/2326.

### Add Protocol

```swift
public protocol ProductCategoriesRemoteProtocol {
    func loadAllProductCategories(for siteID: Int64,
                                         pageNumber: Int,
                                         pageSize: Int,
                                         completion: @escaping ([ProductCategory]?, Error?) -> Void)

    func createProductCategory(for siteID: Int64,
                                      name: String,
                                      parentID: Int64?,
                                      completion: @escaping (ProductCategory?, Error?) -> Void)
}

/// Product Categories: Remote Endpoints
///
public final class ProductCategoriesRemote: Remote, ProductCategoriesRemoteProtocol {
```

### Add Mock

```swift
final class MockProductCategoriesRemote: ProductCategoriesRemoteProtocol {
    private struct CreationResultKey: Hashable {
        let siteID: Int64
        let name: String
        let parentID: Int64?
    }

    private var creationResults = [CreationResultKey: Result<ProductCategory, Error>]()

    func loadAllProductCategories(for siteID: Int64, pageNumber: Int, pageSize: Int, completion: @escaping ([ProductCategory]?, Error?) -> Void) {
    }

    func createProductCategory(for siteID: Int64, name: String, parentID: Int64?, completion: @escaping (ProductCategory?, Error?) -> Void) {
        DispatchQueue.main.async { [weak self] in
            guard let self = self else {
                return
            }

            let key = CreationResultKey(siteID: siteID, name: name, parentID: parentID)
            if let result = self.creationResults[key] {
                switch result {
                case .success(let category):
                    completion(category, nil)
                case .failure(let error):
                    completion(nil, error)
                }
            } else {
                XCTFail("\(String(describing: self)) Could not find Result for \(key)")
            }
        }
    }

    func whenCreating(siteID: Int64, name: String, parentID: Int64?, thenReturn result: Result<ProductCategory, Error>) {
        let key = CreationResultKey(siteID: siteID, name: name, parentID: parentID)
        creationResults[key] = result
    }
}
```

We could probably make a generator for this in the future when it gets painful to create mocks. 

### Make Remote a Dependency in `ProductCategoryStore`

```swift

    private let remote: ProductCategoriesRemoteProtocol

        self.remote = ProductCategoriesRemote(network: network)
    }

    init(dispatcher: Dispatcher, storageManager: StorageManagerType, network: Network, remote: ProductCategoriesRemoteProtocol) {
        self.remote = remote
    }

    func addProductCategory(siteID: Int64, name: String, parentID: Int64?, onCompletion: @escaping (ProductCategory?, Error?) -> Void) {
        remote.createProductCategory(for: siteID, name: name, parentID: parentID) { [weak self] (productCategory, error) in

```

### Use the Mock in the Tests

```swift
    func testAddProductCategoryAddsStoredCategorySuccessfulResponse() throws {
        // Given a stubed product category network response
        let category = Yosemite.ProductCategory(categoryID: 104, siteID: sampleSiteID, parentID: 0, name: "Dress", slug: "Shirt")

        let remote = MockProductCategoriesRemote()
        remote.whenCreating(siteID: sampleSiteID, name: "Dress", parentID: 0, thenReturn: .success(category))

        let store = ProductCategoryStore(dispatcher: Dispatcher(), storageManager: storageManager, network: network, remote: remote)

        // When dispatching a `addProductCategory` action
        var errorResponse: Error?
        waitForExpectation { exp in
            let action = ProductCategoryAction.addProductCategory(siteID: sampleSiteID, name: "Dress", parentID: 0) { (productCategory, error) in
                errorResponse = error
                exp.fulfill()
            }

            store.onAction(action)
        }

        // Then the category should be added
        XCTAssertNil(errorResponse)

        let addedCategory = try XCTUnwrap(viewStorage.loadProductCategory(siteID: sampleSiteID, categoryID: 104))
        XCTAssertNotNil(addedCategory)
        XCTAssertEqual(addedCategory.categoryID, 104)
        XCTAssertEqual(addedCategory.parentID, 0)
        XCTAssertEqual(addedCategory.siteID, sampleSiteID)
        XCTAssertEqual(addedCategory.name, "Dress")
        XCTAssertEqual(addedCategory.slug, "Shirt")
    }
```

--- 

What do you think? 

Consider this just a discussion for now. It's fine if we don't have it in this PR. 🙂

Networking/Networking/Mapper/ProductCategoryMapper.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+
+/// Mapper: ProductCategory
+///
+struct ProductCategoryMapper: Mapper {
```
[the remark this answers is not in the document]
pmusolino: I would say that the ideal would be not to have a lot of similar code, which could be subject to fewer errors. In particular, all endpoints under Woo use either the mapper implemented in this way or similar to this, but returning a list, and we almost always injecting the `siteID`. 😃

Yosemite/YosemiteTests/Stores/ProductCategoryStoreTests.swift
pmusolino: Thanks for your suggestions @shiki.

I'd say that ideally, I'd prefer a situation where we don't need to write a lot of code just to maintain our mock objects and a situation where a JSON file doesn't break everything because something changed. 😃 But sometimes there is no ideal and unique solution to things.

I think we could start using a mix of the two, maintaining JSON files for our Mapper class tests, and below I explain why I wouldn't completely abandon JSON files.
In our case, I found useful to have a JSON representation of the model, since they also include some fields which we don't parse at the moment, and the data follow the same type according to the backend rules.

Ofter happened to me to discovered that I was parsing a field in the wrong way, just because the JSON was returning an Int as String. In those specific cases, it will be useful to use a real JSON for parsing the data.

Another example is the use of real JSON files used for the `Product` model. There are some strange cases where some plugins alter a field value, for example, for the `price`, where the original value can be a String, an Int, or a Decimal. How would you handle this in a mock object which wasn't parsed from a real JSON?

Use simple objects can allow us to simplify how we use mocks in some cases, but testing every class Mapper can be a real pain, but I agree with you on using mock objects in `Store` tests. Another case is how we parse the fields which represent a date. Not going to use the correct date formatter and using objects that are not parsed, we risk getting false positives, but in reality, the date format has returned differently and we will notice it only when that field is used.

I know they are borderline cases, but having a real JSON copied and pasted from the documentation, or from a real network request, I think they make unit tests more useful and less free to interpretations. Ideally, a switch from a version X to a version Y of our API endpoints can be less painful, because copying the new JSON we could easily see which cases are failing, and what we need to update.

Do you think that make sense to use both approaches?

pmusolino:
@shiki thank you so much for your review, and for suggest me the use of `Result`, which is fantastic 🤩 I had lost that PR, and I didn't know that we had started to adopt it.

Yosemite/YosemiteTests/Stores/ProductCategoryStoreTests.swift
shiki: > Do you think that make sense to use both approaches?

@pmusolino Yes and you're **absolutely right**. Thank you for sharing all that background. We'd definitely want the ability to test the cases that you mentioned. Using JSON files is probably the only way we can test that, aside from real API calls. 

I believe this is what we are both agreeing to: 

image

- Remote Tests:
  - Test simulated API responses via JSON files
  - Test API errors
  - Endpoint and corresponding expected response testing
  - Some JSON parsing tests
- Mapper Tests:
  - Highly specialized JSON parsing testing
  - Test that we can _normalize_ the various and weird API responses 
- Store Tests:
  - Assumes that the Mapper and Remote Tests have fully covered the API response testing
  - Assumes that the responses from the (mocked) Remote are correctly parsed. There should be no "JSON parsing" testing here. 
  - Focused on business logic testing
  - Tests for database operations

How does this look? 

---

> for example, for the price, where the original value can be a String, an Int, or a Decimal. How would you handle this in a mock object which wasn't parsed from a real JSON?

If we follow the specs above, I believe this should be covered by the Mapper Tests. Or did I misunderstand this? 😅

shiki: FYI @jaclync. In case you're interested in this discussion. 😄

jaclync: Great discussion! I like the plan above to test with the JSON files in Remote & Mapper tests, and a mock remote in Store tests 👍 

Testing with JSON simulates the API layer and allows us to test the handling of unexpected API response, so it makes the most sense to test with JSON in the Networking layer. I don't think we currently have many tests for unexpected JSON, but this JSON setup would be useful when we work on bug fixes. In the Yosemite layer, testing with a mock remote would be helpful to set the expected data more easily than with JSON (especially if we want to test different variations of the response).

pmusolino: > How does this look?

Your proposal sounds good to me and makes totally sense.

> If we follow the specs above, I believe this should be covered by the Mapper Tests. Or did I misunderstand this?

This is correct 😃👍 

What do you think if we create a doc page with this information, just to not miss them anymore? What to do with all our existing test cases? Do we leave them as they are, and will we migrate them as soon as they need a makeover?

pmusolino:
Thanks @shiki, I'm going to merge it now 👍

Yosemite/YosemiteTests/Stores/ProductCategoryStoreTests.swift
shiki: > What do you think if we create a doc page with this information, just to not miss them anymore?

Yup. Thanks for the nudge. I created a task for it. https://github.com/woocommerce/woocommerce-ios/issues/2394

> What to do with all our existing test cases? Do we leave them as they are, and will we migrate them as soon as they need a makeover?

Yes, that's what I think. We can change them later if we really need to. For new test cases, we can start creating the mocks.

Ecarrion: A little late to this convo but I just saw it on the iOS-handbook.
I normally don't like json on unit tests because they are a pain to maintain, specially when we fail to update them and that gives us a sense of false safety.

In my ideal world we would be able to run our remote objects and mappers tests against some kind of local/test server that can serve JSONs from web/GH where we don't need to maintain them.

But since I've never accomplished that I'm happy to follow the current rules :-)

shiki: > In my ideal world we would be able to run our remote objects and mappers tests against some kind of local/test server that can serve JSONs from web/GH where we don't need to maintain them.

I did this before. A downside to this is that the unit tests will no longer run without internet connection. Or if there are problems with the server, everyone gets blocked. 😬 

We do have some [Woo REST API unit tests](https://github.com/wordpress-mobile/WordPress-FluxC-Android/tree/develop/tests/api/src/test/java) as part of FluxC. That one runs on a real server I believe. We should get alerted if there's an API change that we did not expect. 

It's probably still useful to have a similar one fo the iOS side in the future. 

Or we could set up our mapper tests so that they can be tested against a local server, possibly using our variant of [VVV](https://github.com/Varying-Vagrant-Vagrants/VVV) that anyone can easily set up.

shiki: Created https://github.com/woocommerce/woocommerce-ios/issues/2480 so the idea will not die.

Issue #2200, opened 2020-04-27
Products: edit an existing external product


pmusolino:
As part of Products M3, we should provide a way to edit an existing external product.

- [x] schema update:
  - [x] add `buttonText: String` field to `Product` that corresponds to `button_text` field in the API response https://github.com/woocommerce/woocommerce-ios/pull/2391
  - [x] Networking + Yosemite layers https://github.com/woocommerce/woocommerce-ios/pull/2392
- [x] display an external product with readonly properties for the new rows (product link, attributes) https://github.com/woocommerce/woocommerce-ios/pull/2432:
  - always shown in the settings section: price, product type (implemented in https://github.com/woocommerce/woocommerce-ios/issues/2262), product link
  - bottom sheet actions: attributes (edit action is implemented in https://github.com/woocommerce/woocommerce-ios/issues/2385), SKU, categories, tags, short description

![external products flow ios](https://user-images.githubusercontent.com/495617/80397036-dc9dab00-88b5-11ea-8188-a609ae0b0b92.jpg)

- [x] editable product link https://github.com/woocommerce/woocommerce-ios/pull/2442
<img width="156" alt="Screen Shot 2020-06-04 at 4 03 08 PM" src="https://user-images.githubusercontent.com/1945542/83731325-e7015080-a67c-11ea-8442-4f9321ea5378.png">

- [ ] editing product link followup: local validation for external URL
- [ ] editing product link followup: auto-fill button text with "Buy product" when the external URL field becomes non-empty

- [x] editable SKU https://github.com/woocommerce/woocommerce-ios/pull/2437
<img width="231" alt="Screen Shot 2020-06-04 at 4 01 46 PM" src="https://user-images.githubusercontent.com/1945542/83731215-b9b4a280-a67c-11ea-8bad-e2cd95044488.png">

jkmassel:
We're freezing version `4.7` today, so I'm bumping this to `4.8`. If you need this to be part of `4.7`, please merge this PR into `release/4.7` and DM me – I'll be happy to cut a new beta release!

jaclync:
Moved the 2 design enhancements to a separate task https://github.com/woocommerce/woocommerce-ios/issues/2609 so that we can close the main issue now

PR #2392, opened 2020-06-05
Networking & Yosemite layers: add `buttonText` to `Product`

jaclync:
Networking & Yosemite layers for #2200 

- [x] ⚠️ depends on schema update in the storage layer https://github.com/woocommerce/woocommerce-ios/pull/2391

## Changes

- Added `buttonText: String` to the readonly `Product` in the Networking layer
  - Updated a bunch of usage in unit tests in `ProductUpdater` (wouldn't have to do this with `Copiable`!)
- Networking layer:
  - Added `product-external.json` as a sample API response from retrieving an external product
  - Added a test case for loading a product
- Yosemite layer:
  - Based on our discussion in https://github.com/woocommerce/woocommerce-ios/pull/2387/files#r435272569, I experimented with `ProductsEndpointsProviding`/`MockProductsRemote` (especially since they were already created 😄). Added an initializer with `remote: ProductsEndpointsProviding` parameter with `ProductsRemote(network: network)` as default, and used the `remote` for loading a product for a new test case in `ProductStoreTests`

## Testing

Just a sanity check on fetching products:

- Launch the app
- Go to the Orders tab
- Tap on an order with a simple product
- Tap on a simple product --> the product details should launch as before

Update release notes:


Networking/Networking/Model/Product/Product.swift
```diff
@@ -290,6 +293,7 @@ public struct Product: Codable {
         let downloadLimit = try container.decode(Int.self, forKey: .downloadLimit)
         let downloadExpiry = try container.decode(Int.self, forKey: .downloadExpiry)
 
+        let buttonText = try container.decodeIfPresent(String.self, forKey: .buttonText) ?? ""
```
pmusolino, line 296: Do we really want to save it as an empty text? 
If it is not present, it would be better to insert the value as an optional, so as not to save it if not present. What do you think?

Networking/NetworkingTests/Remote/ProductsRemoteTests.swift
```diff
@@ -270,6 +298,7 @@ private extension ProductsRemoteTests {
                        downloads: [],
                        downloadLimit: -1,
                        downloadExpiry: -1,
+                       buttonText: "",
```
pmusolino, line 301: Also here, if it's not present maybe it's better to set it as `nil`.

Networking/NetworkingTests/Remote/ProductsRemoteTests.swift
```diff
@@ -89,6 +89,34 @@ class ProductsRemoteTests: XCTestCase {
...
+        wait(for: [expectation], timeout: Constants.expectationTimeout)
+
+        // Then
+        let result = try XCTUnwrap(resultMaybe)
+        XCTAssertTrue(result.isSuccess)
+
+        let product = try result.get()
+        XCTAssertEqual(product.productID, sampleProductID)
+        XCTAssertEqual(product.buttonText, "Hit the slopes")
+        XCTAssertEqual(product.externalURL, "https://snowboarding.com/product/rentals/")
```
pmusolino, line 117: I suggest here to use the new callback `waitForExpectation:`

```suggestion
        let remote = ProductsRemote(network: network)

        network.simulateResponse(requestUrlSuffix: "products/\(sampleProductID)", filename: "product-external")

        // When
        var resultMaybe: Result<Product, Error>?
        waitForExpectation { (exp) in
            remote.loadProduct(for: sampleSiteID, productID: sampleProductID) { aResult in
                resultMaybe = aResult
                exp.fulfill()
            }
        }

        // Then
        let result = try XCTUnwrap(resultMaybe)
        XCTAssertTrue(result.isSuccess)

        let product = try result.get()
        XCTAssertEqual(product.productID, sampleProductID)
        XCTAssertEqual(product.buttonText, "Hit the slopes")
        XCTAssertEqual(product.externalURL, "https://snowboarding.com/product/rentals/")
```

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -340,6 +340,29 @@ class ProductStoreTests: XCTestCase {
...
+        remote.whenLoadingProduct(siteID: sampleSiteID, productID: sampleProductID, thenReturn: .success(expectedProduct))
+        let productStore = ProductStore(dispatcher: dispatcher, storageManager: storageManager, network: network, remote: remote)
+
+        // Action
+        let action = ProductAction.retrieveProduct(siteID: sampleSiteID, productID: sampleProductID) { (product, error) in
+            // Assert
+            XCTAssertNil(error)
+            XCTAssertEqual(product, expectedProduct)
+
+            expectation.fulfill()
```
pmusolino, line 359: Also here, I suggest the use of `waitForExpectation:`

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -340,6 +340,29 @@ class ProductStoreTests: XCTestCase {
         wait(for: [expectation], timeout: Constants.expectationTimeout)
     }
 
+    /// Verifies that `ProductAction.retrieveProduct` returns the expected `Product` of external product type.
+    ///
+    func testRetrieveSingleExternalProductReturnsExpectedFields() {
+        // Arrange
+        let expectation = self.expectation(description: "Retrieve single external product")
+        let remote = MockProductsRemote()
+        let expectedProduct = MockProduct().product(siteID: sampleSiteID, productID: sampleProductID, buttonText: "Deal", externalURL: "https://goodeals.com")
```
pmusolino, line 349: The website seems an existing website with an old SSL certificate. I suggest using `https://example.com` if possible.

Networking/Networking/Model/Product/Product.swift
jaclync: from my testing and the [WC API example response](https://woocommerce.github.io/woocommerce-rest-api-docs/#retrieve-a-product) (on the right side), `button_text` field is empty instead of `null` if not applicable (this seems to be the default API response format for the other fields as well). That was the reason I created the new `buttonText: String` attribute in https://github.com/woocommerce/woocommerce-ios/pull/2391 as non-optional with default value to an empty string to be consistent with the API response. Also, we don't have to check if the attribute is `nil` or empty in order to know whether it's set this way.

lemme know if you see any concerns with this non-optional attribute!

Networking/NetworkingTests/Remote/ProductsRemoteTests.swift
jaclync: (same as the above)

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
jaclync: I totally made this URL up and never went to that URL 😅 to avoid confusion, I updated this URL in https://github.com/woocommerce/woocommerce-ios/pull/2392/commits/6dc3ae7dfe6bddc991c9236c9009be0b7275ca58

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
jaclync: yup! updated in https://github.com/woocommerce/woocommerce-ios/pull/2392/commits/d78a26ed188f0d47b5c37d13f48e7c6278521b61

Networking/NetworkingTests/Remote/ProductsRemoteTests.swift
jaclync: updated in https://github.com/woocommerce/woocommerce-ios/pull/2392/commits/d78a26ed188f0d47b5c37d13f48e7c6278521b61 :)

jaclync:
@pmusolino thanks for the feedback! I responded to the comments if you'd like to take another look 😄

Networking/Networking/Model/Product/Product.swift
pmusolino: Sounds good to me if this is the reason, but if the field it's always present, why use `decodeIfPresent`? I think that using `decode` is enough.

pmusolino:
Thanks, @jaclync! Looks good to me. Just a nit on `decodeIfPresent`. 💯

Networking/Networking/Model/Product/Product.swift
jaclync: good point! updated in https://github.com/woocommerce/woocommerce-ios/pull/2392/commits/2d33218d2318056543aeaa24132d50e3d9db7e99

I'm always a bit concerned if some site configurations would alter the fields, but I think it's better to make it an exception rather than failing silently.

jaclync:
thanks @pmusolino! I updated `decodeIfPresent` in https://github.com/woocommerce/woocommerce-ios/pull/2392/commits/2d33218d2318056543aeaa24132d50e3d9db7e99, merging now 😄

PR #2395, opened 2020-06-05
Order Details → Remove SKU from Products

shiki:
Part of #2281. 

## Design

This removes the SKU from the Products in the Order Details. The SKU is still shown when fulfilling the order or viewing the product details. 

Before | After  
--------|-------
image | image

magic

## Testing

1. Navigate to an order
2. Confirm that the SKU is no longer visible. 

## Reviewing 

## Submitter Checklist


designsimply:
Tested with WCiOS 4.5.0.2 (TestFlight beta) and LGTM!

PR #2396, opened 2020-06-05
Order Details → Use singular or plural "Products" section title

### LINKED ISSUE #2281: Order Details → Show Price on the Right Instead of Quanity

shiki:
Requires #2389 to be merged first. Closes #2281. 

## Design

This changes the products section header title to use the plural term only if there is more than one product. 

More Than One | One Product Only
--------|-------
image        |    image

## Testing

1. Navigate to an order containing one product. 
2. Confirm that the header shows “Product”. 
1. Navigate to an order containing more than one product. 
2. Confirm that the header shows “Products”. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -577,7 +577,7 @@ extension OrderDetailsDataSource {
                     rows.append(.details)
                 }
 
-                return Section(title: Title.products, rightTitle: nil, rows: rows, headerStyle: .primary)
+                return Section(title: Localization.pluralizedProducts(count: items.count), rightTitle: nil, rows: rows, headerStyle: .primary)
             }
 
             var rows = [Row]()
@@ -599,7 +599,7 @@ extension OrderDetailsDataSource {
                 return nil
             }
 
-            return Section(title: Title.products, rightTitle: nil, rows: rows, headerStyle: .primary)
+            return Section(title: Localization.pluralizedProducts(count: items.count), rightTitle: nil, rows: rows, headerStyle: .primary)
         }()
 
         let refundedProducts: Section? = {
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -1017,3 +1018,13 @@ extension OrderDetailsDataSource {
         static let paidByCustomerCell = 1
     }
 }
+
+// MARK: - Private Utils
+
+private extension OrderDetailsDataSource {
+    enum Localization {
+        static func pluralizedProducts(count: Int) -> String {
+            count > 1 ? Title.products : Title.product
```
jaclync, line 1027: Non-blocking 💬  It'd be nice if we could use a `.stringsdict` for noun plurals ([Apple doc](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html) > "Handling Noun Plurals and Units of Measurement"), though it seems like it's not supported yet from paaHJt-c3-p2. Do you think it's worth a ping to Platform 9 since the last conversation was in 2018?

shiki: Yeah, I looked into this not too long ago when I wrote [this WPiOS documentation](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/docs/localization.md). Documentation is awesome because I wrote in there something I forgot: 

> GlotPress currently does not support pluralization using the .stringsdict file. So, right now, you have to support plurals manually by having separate localized strings.

The limitation is in GlotPress. 😐 

Based on the discussion in https://github.com/wordpress-mobile/WordPress-iOS/issues/6327, it looks like it's still not resolved on the GlotPress side. 

Once that is resolved, I think we also need to move out of `localize.py` to something that parses the `.stringsdict` too.

Issue #2371, opened 2020-05-29
Crash on startup: NSPersistentContainer loading failure


shiki:
Found by @jaclync in p91TBi-2PM-p2. This is currently not tracked by Sentry.

For the 4.1 release, 196 devices were affected. 

---

- [ ] Investigate file access
- [ ] Investigate adding breadcrumbs
- [ ] Consider not crashing and show a message to the user instead
- [ ] Consider not deleting the databases inside the `loadPersistentStore` callback
- [ ] Consider signing the user out if we cannot recover

---

Improvements for `CoreDataManager` and Core Data related:

- [x] Fix Transformer errors/warnings in iOS 13 https://github.com/woocommerce/woocommerce-ios/issues/1372
- [ ] Use NSPersistentStoreCoordinator instead of FileManager in Core Data Backups https://github.com/woocommerce/woocommerce-ios/issues/2473
- [x] Replace `fatalError` with non-fatal error logging when failing to back up the pre-existing persistent store
- [x] Separate retrieval of all model versions to a separate class/struct with unit tests

jaclync:
> Consider not crashing and show a message to the user instead

👍  If we really cannot recover from the Core Data setup, signing the user out would still be better than the app crashing consistently

shiki:
@jaclync Do you think we should close this and the linked issues in Sentry?

shiki:
As of August 17, 2020:

- 60k events
- No number of users recorded

shiki:
I marked the Sentry issue as resolved but I'm leaving this open for now.

shiki:
I'm going to close this now. We have significantly decreased the number of crashes. We do have a few startup crashes left which we'll be tracking in #2667.

PR #2413, opened 2020-06-11
Add more logging for Core Data migration for crash investigation

jaclync:
Logging for crash investigation for #2371 

## Background

The current logging for the Core Data init crash doesn't get us very far to understand how it went during the migration process before we call `loadPersistentStores` on a `NSPersistentContainer`. This PR aims to add more logging to the Core Data init crash logging for us to know more about the migration before the crash.

## Changes

- In `CoreDataIterativeMigrator`, returns a tuple with an array of debug messages in addition to the success boolean
- In `CoreDataManager`, gathered debug messages from `migrateDataModelIfNecessary` including the call to `CoreDataIterativeMigrator` and included these debug messages under `"migrationMessages"` key in the "Additional data" logging ([example Sentry event](https://sentry.io/organizations/a8c/issues/1720861484/events/84dca9b5e32a4fe9a88f6a39986fa13a/))

## Testing

Feel free to manually log a message after the line of `let migrationDebugMessages = migrateDataModelIfNecessary()` in `CoreDataManager` like:

```swift
crashLogger.logMessage("[CoreDataManager] Testing only!", properties: ["migrationMessages": migrationDebugMessages], level: .info)
```

build with "Release-Alpha" configuration, and then check Sentry after launching the app (potentially with migration from a different model version).

at some point I wasn't able to see the events on Sentry, and noticed there were processing issues on the symbols. I wasn't able to figure out after running their script so I switched to a different simulator.

Update release notes:

### DIFF
--- a/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
+++ b/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
@@ -18,28 +18,28 @@ public struct CoreDataIterativeMigrator {
     ///
     /// - Throws: A whole bunch of crap is possible to be thrown between Core Data and FileManager.
     ///
-    static func iterativeMigrate(sourceStore: URL, storeType: String, to targetModel: NSManagedObjectModel, using modelNames: [String]) throws -> Bool {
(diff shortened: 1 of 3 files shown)


jaclync:
Thanks for reviewing & testing Paolo!

PR #2428, opened 2020-06-15
Hack Week: Today Stats Widget (Step 1)

pmusolino:
This PR completes the first step of the Stats widget Project as defined here: p91TBi-2Vw

## Changes
- Added a today stats extension to the app that displays the current day stats, if v4 stats are available for the site.
- An error message is displayed in the widget if the user is not logged in to the app.
- An error message is displayed in the widget if the store does not have WC-Admin installed or if WooCommerce 4.0 is not installed.

## Testing
1. Run the Woo app, and make sure you are not logged.
2. **Open the widget ->** Make sure you see an error about your login state.
3. Login in the Woo app, and choose an app that doesn't support stats v4.
4. **Open the widget ->** Make sure you see an error on stats v4 not available.
5. Login in the Woo app, and choose an app that supports stats v4.
6. **Open the widget ->** You will be able to see the stats on the widget, and the selected website.
7. Open the app, and change the store.
8. **Open the widget ->** Verify that the widget data is refreshed when store is changed.

## Notes for testing
The widget is disabled by default now (because it's not already completed).
To enable it:
1. Click the project file in the project navigator
2. Click the containing app target (WooCommerce)
3. Click **Build Phases** tab
4. Open **Embed App Extensions**
5. Add the extension (TodayStatsWidget.appex)

Schermata 2020-06-19 alle 11 51 12

## Screenshots
| Dark Mode off            |  Dark mode on | Error state
Simulator Screen Shot - iPhone 11 Pro - 2020-06-15 at 12 42 13 | Simulator Screen Shot - iPhone 11 Pro - 2020-06-15 at 12 42 26 | Simulator Screen Shot - iPhone 11 Pro - 2020-06-15 at 12 40 38

Update release notes:


WooCommerce/TodayStatsWidget/TodayViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+                                            group.leave()
+                                            return
+                                        }
+
+                                        tempStats.totalVisitors = siteVisitStats?.totalVisitors
+                                        group.leave()
+        }
+
+        group.notify(queue: .main) { [weak self] in
+            if tempStats.isVisitStatsEmpty() || tempStats.isOrderStatsEmpty() {
```
jaclync, line 173: nit: these two conditions feel like lower level details from the same class `StatsData`, how about grouping these two to a function in `StatsData` like `hasData()`?

WooCommerce/Classes/System/WidgetExtensionService.swift
```diff
@@ -0,0 +1,86 @@
...
+            if let loadedSite = try? decoder.decode(Site.self, from: savedSite) {
+                return loadedSite
+            }
+        }
+        return nil
+    }
+
+    static func saveSite(site: Site) {
+        let encoder = JSONEncoder()
+        if let encoded = try? encoder.encode(site) {
```
jaclync, line 76: I'm guessing this is the reason for the diffs in `Site`'s `encode` implementation? I was wondering if we could go without changing the current `Site` implementation since it isn't used by the main app, and `Site` also contains data that aren't used by the widget. Maybe creating a new codable struct (e.g. `SiteData`) with all the necessary fields used in the widget would be lighter weight, could be unit tested, and it wouldn't require changes to `Site`. Then here it just creates a `SiteData` for example from the given `Site`, and it can directly encode/decode the `SiteData` within the widget.

WooCommerce/TodayStatsWidget/TodayViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+                    self?.error = .errorFetchingStats
+                }
+                group.leave()
+                return
+            }
+            tempStats.totalOrders = orderStatsV4?.totals.totalOrders
+
+            // TODO: implement currency formatter
+            // let currencyCode = CurrencySettings.shared.symbol(from: CurrencySettings.shared.currencyCode)
+            // totalRevenueText = CurrencyFormatter().formatHumanReadableAmount(String("\(orderStats.totals.grossRevenue)"), with: currencyCode) ?? String()
```
jaclync, line 148: nit: maybe these can be added at the next PR since we generally don't keep commented out / unused code in the codebase unless it's being worked on soon? 🙂

WooCommerce/TodayStatsWidget/TodayViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+
+// MARK: - Widget Updating
+extension TodayViewController: NCWidgetProviding {
+
+    func widgetPerformUpdate(completionHandler: (@escaping (NCUpdateResult) -> Void)) {
+        // Perform any setup necessary in order to update the view.
+
+        // If an error is encountered, use NCUpdateResult.Failed
+        // If there's no update required, use NCUpdateResult.NoData
+        // If there's an update, use NCUpdateResult.NewData
```
jaclync, line 52: feels like a great candidate for unit tests? 😄

WooCommerce/TodayStatsWidget/TodayViewController.swift
```diff
@@ -0,0 +1,366 @@
...
+            "Store analytics not available! Please upgrade to the latest version of WooCommerce to view your store analytics.",
+            comment: "Store analytics error in Today Stats Widget")
+        static let errorFetchingStats = NSLocalizedString("There was an error trying to fetch stats.",
+                                                          comment: "Error while fetching stats in Today Stats Widget")
+        static let missingCredentials = NSLocalizedString("Please log in to the WooCommerce app to add a widget.",
+                                                          comment: "Login error in Today Stats Widget")
+        static let genericError = NSLocalizedString("Something goes wrong.", comment: "Generic error in Today Stats Widget")
+    }
+
+    enum Errors {
```
jaclync, line 322: nit: it's a bit odd seeing the error property like `error: Errors`, maybe this could be like `enum StatsWidgetError: Error`/`enum WidgetError: Error`?

shiki:
@pmusolino When running `rake dependencies`, I get this warning:

```
[!] The `WooCommerce [Debug]` target overrides the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` build setting defined in `../Pods/Target Support Files/Pods-WooCommerce/Pods-WooCommerce.debug.xcconfig'. This can lead to problems with the CocoaPods installation
    - Use the `$(inherited)` flag, or
    - Remove the build settings from the target.

[!] The `WooCommerce [Release]` target overrides the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` build setting defined in `../Pods/Target Support Files/Pods-WooCommerce/Pods-WooCommerce.release.xcconfig'. This can lead to problems with the CocoaPods installation
    - Use the `$(inherited)` flag, or
    - Remove the build settings from the target.

[!] The `WooCommerce [Release-Alpha]` target overrides the `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES` build setting defined in `../Pods/Target Support Files/Pods-WooCommerce/Pods-WooCommerce.release-alpha.xcconfig'. This can lead to problems with the CocoaPods installation
    - Use the `$(inherited)` flag, or
    - Remove the build settings from the target.
```

I think the cause is a change in this branch. I'm just not sure which one. This doesn't happen in `develop`. 

To reproduce:

1. `rm -rf Pods`
2. `rake dependencies`

shiki:
> I would have liked the widget to be part of a separate project, but since it is closely linked to the app and since it must share some files and some classes, the only solution for the moment has been to integrate it within the project. I also believe that, even if we find a way to manage it as a separate project (like our Network, Yosemite, and Storage frameworks), the problem would be to sign it together with the app for distribution on the App Store.

I think we should document this somewhere in the code. I'm just not sure where. 🤔

WooCommerce/Classes/System/WidgetExtensionService.swift
```diff
@@ -0,0 +1,86 @@
...
+import Networking
+
+// WidgetExtensionService class
+// Save, Load and Remove credentials used by Extensions of the Woo app, like the TodayStatsWidget
+//
+final class WidgetExtensionService: NSObject {
+
+    /// Returns the Credentials, if any, under the related App Groups.
+    ///
+    static func loadCredentials() -> Credentials? {
```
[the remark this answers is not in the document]
shiki: I'm also concerned that we have `WidgetExtensionService` injected everywhere. I believe this makes it less clear how the Widgets target depends on the WooCommerce, Yosemite, and other frameworks. 

I haven't fully grasped this PR yet but I wonder if it's possible to make it look like this:

image

### WidgetCredentialsRepository

- This is shared by the WooCommerce and the Widgets Targets. 
- Encapsulates the `UserDefaults` magic.  
- Has no static methods. It has to be instantiated. It can be kept as a private instance of its consumer.
- Maybe this can handle validation of sorts that both the `Credentials` and `Site` must exist together or the "complete credentials" is invalid/unusable. 

### SessionManager

- Dispatches change events whenever the `site` and the `credentials` are updated using `PublishSubject` + `Observable`. 
- No other code should be added here so that its only concern is keeping information about the current session. 

### WidgetCredentialsUpdater

- Belongs in the WooCommerce Target only.
- Upon activation, observes the Observables provided by `SessionManager` (Credentials and Site)
- Uses an instance of `WidgetCredentialsRepository`
- Updates `WidgetCredentialsRepository` whenever `SessionManager`'s Observables emit new changes.
- This can be kept by `AppDelegate` probably so it's not accessible by other classes. But it has to be active in memory in order to work.

### TodayViewController

- Uses an instance of `WidgetCredentialsRepository` to read the credentials and site. 

---

If we can implement it this way, we reverse the dependency. Instead of `SessionManager` **depending** on `WidgetExtensionService`, it's now the other way around. This also makes it easier to take out and refactor the Widget system. For example, to remove the widgets from WooCommerce target:

1. Delete the `WidgetCredentialsUpdater`
2. That's it. Well, except the `UserDefaults` keys I guess.

pmusolino:
Thank you both @jaclync and @shiki for your reviews! I'm closing this PR after the discussion here p91TBi-2Zp. I'll try in the near future to extract some parts of this PR, like credentials, to make it works with the new widgets of iOS 14.

Issue #2262, opened 2020-05-11
Edit Products: add product type row


jaclync:
Original discussion: https://github.com/woocommerce/woocommerce-ios/pull/2217#issuecomment-624100123

Let's add a product type row to product form (simple products for now) based on the design:

<img width="259" alt="Screen Shot 2020-05-11 at 9 29 33 AM" src="https://user-images.githubusercontent.com/1945542/81516352-26809980-936a-11ea-8509-aef370f3f15b.png">

Possible product types:

Simple product
Grouped product
External product
Variable product
Custom product type name (e.g. subscriptions)

---

from the design, it seems like we currently have 4 product types:
- Physical product
- Virtual product
- Grouped product
- External product

<img width="381" alt="Screen Shot 2020-05-11 at 9 36 03 AM" src="https://user-images.githubusercontent.com/1945542/81516561-d9e98e00-936a-11ea-911a-594df7e43143.png">

@Garance91540 I have some questions for this feature:

### Question 1

Could we display the product type as in core/API instead of categorizing a product into one of the 4 options above (so that we show "Simple product"/"Grouped product"/"External/Affiliate product"/"Variable product"/other custom product from the product type field)? some reasons I can think of:
- For variable products, each variation has its own virtual/downloadable options so that a variable product might offer a mix of virtual and physical services. In these cases, we can't simply categorize a variable product as a physical or virtual product
- There could be many other custom product types from plugins. For example, subscriptions and booking are common product types from support rotation

### Question 2

Could we only display the product type on editable simple products for M2? The design looks quite different from non-simple products that are readonly now (example screenshot below). I understand that Android might have implemented in a different way so that all product types share the same updated design, while iOS only updated the design for editable product types (simple products for M2). since we're making more product types editable in the upcoming products releases and deprecating the readonly product screens like in the screenshot, I thought we could spend time on supporting more editable product types rather than changing the readonly product screens that we're deprecating in the near future.

<img src="https://user-images.githubusercontent.com/1945542/81517391-609f6a80-936d-11ea-985d-08f47b815a9e.png" width="320" />

Garance91540:
Hi @jaclync !

>Could we display the product type as in core/API instead of categorizing a product into one of the 4 options above (so that we show "Simple product"/"Grouped product"/"External/Affiliate product"/"Variable product"/other custom product from the product type field)? 

[...]

PR #2432, opened 2020-06-16
External/affiliate products are now editable when Products M3 flag is enabled

jaclync:
Editable external/affiliate product for #2200 

Note: Pending edit actions are SKU and external URL, as separate subtasks in #2200

## Changes

- Added `UIImage.linkImage` for the icon on the external URL row on the product form settings section
- Added `sku` and `externalURL` cases to `ProductFormEditAction` enum, and created the corresponding cell view models in `DefaultProductFormTableViewModel`. In `ProductFormActionsFactory`, updated the logic to return different actions for simple and affiliate products
- Added `editSKU` case to `ProductFormBottomSheetAction` enum, that is mapped from `ProductFormEditAction.sku`
- In `ProductDetailsFactory`, updated the logic to return a product form for an affiliate product if the Products M3 flag is enabled
- Updated unit tests

### Products M3 flag is on

- Go to the Products tab
- Tap on an external/affiliate Product without SKU --> should see at least price and external URL rows in the settings section
- Tap "Add more details" at the bottom --> should see at least "SKU" action in the bottom sheet (no-op for now)
- Dismiss the bottom sheet and try editing the affiliate product
- Tap "Update" --> the affiliate product should be updated remotely

### Products M3 flag is off

- Turn off the M3 feature flag in code
- Go to the Products tab
- Tap on an external/affiliate Product --> should see readonly product details

## Example screenshots

\ | product form actions | bottom sheet actions
-- | -- | --
Product has no SKU | Simulator Screen Shot - iPhone 11 - 2020-06-16 at 11 55 22 | Simulator Screen Shot - iPhone 11 - 2020-06-16 at 11 55 22
Product has a SKU | Simulator Screen Shot - iPhone 11 - 2020-06-16 at 11 55 42 | Simulator Screen Shot - iPhone 11 - 2020-06-16 at 11 55 50

Update release notes:


jaclync:
> I created a separate issue here but actually all the "empty field" should always be an action prompt not a descriptive label (so here, if there's no price yet then we should read "Add price" instead of "price")

@Garance91540 sounds good 👍 I'll address [this issue](https://github.com/woocommerce/woocommerce-ios/issues/2434) separately this sprint!

> I'm assuming the "Product type" row will be shown by default once we support product type change?

yes, the product type row & action will be implemented in https://github.com/woocommerce/woocommerce-ios/issues/2262 and https://github.com/woocommerce/woocommerce-ios/issues/2198 respectively.

> There's a small typo in the bottom sheet "Easily identify your products with unique codes"

great catch! I copied the string from the design without noticing the typo 😅 . here's an updated screenshot with the new copy (now it takes 2 lines on iPhone 11):

WooCommerce/WooCommerceTests/Mockups/MockProduct.swift
```diff
@@ -18,6 +18,7 @@ final class MockProduct {
                  fullDescription: String? = "<p>This is the party room!</p>\n",
                  productShippingClass: ProductShippingClass? = nil,
                  backordersSetting: ProductBackordersSetting = .notAllowed,
+                 externalURL: String? = "http://somewhere.com",
```
pmusolino, line 21: nit: 
```suggestion
                 externalURL: String? = "https://example.com",
```

jaclync: updated in: https://github.com/woocommerce/woocommerce-ios/pull/2432/commits/d565e71f13a7da50b9ce91d470f9f894c93a70fe

WooCommerce/Classes/ViewRelated/Products/Edit Product/BottomSheetListSelector/ProductFormActionsFactory.swift
```diff
@@ -65,14 +52,58 @@ struct ProductFormActionsFactory {
 
     /// Retruns an array of actions that are visible in the product form bottom sheet.
```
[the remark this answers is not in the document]
jaclync: oops good catch! updated in: https://github.com/woocommerce/woocommerce-ios/pull/2432/commits/d565e71f13a7da50b9ce91d470f9f894c93a70fe

jaclync:
thanks for the review Paolo, I addressed the changes you mentioned and I'm merging this PR now!

PR #2436, opened 2020-06-16
Product Categories: managed selection status

pmusolino:
Part of #2000 

## Changes
- `ProductCategory` model now conform to `GeneratedCopiable` (generated also the `copy` method)
- Small refactoring of `ProductCategoryTableViewCell`
- Added `categoryID` in `ProductCategoryCellViewModel` which is useful to know which category is.
- Restored the state of the tableview in `ProductCategoryListViewController.xib` to the original state, it didn’t allow the cell selection.
- Introduced the array `newSelectedCategories` in `ProductCategoryListViewModel`, which it’s initialized with the categories from the original product, and that handle the changes after the selection/deselection.
- `hasCategoriesChanges()` method in `ProductCategoryListViewModel` which will be useful for the discard changes prompt.

## Testing
1. Navigate to a product with Categories
2. Test that the selection/deselection of a category works like expected (no save at the moment. It will be part of another task)

## GIF

selection-deselection categories

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/ProductCategory.swift
+++ b/Networking/Networking/Model/Product/ProductCategory.swift
@@ -3,7 +3,7 @@ import Foundation
 
 /// Represents a ProductCategory entity.
 ///
-public struct ProductCategory: Decodable {
+public struct ProductCategory: Decodable, GeneratedCopiable {
     public let categoryID: Int64
     public let siteID: Int64
     public let parentID: Int64
--- a/Networking/Networking/Model/Product/ProductCategory.swift
+++ b/Networking/Networking/Model/Product/ProductCategory.swift
@@ -3,7 +3,7 @@ import Foundation
 
 /// Represents a ProductCategory entity.
 ///
-public struct ProductCategory: Decodable, GeneratedCopiable {
+public struct ProductCategory: Decodable {
     public let categoryID: Int64
     public let siteID: Int64
     public let parentID: Int64
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/Cell/ProductCategoryTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/Cell/ProductCategoryTableViewCell.swift
@@ -40,7 +40,7 @@ private extension ProductCategoryTableViewCell {
     private func styleSelection() {
         self.selectionStyle = .none
     }
-    
+
     private func styleLabels() {
         nameLabel.applyBodyStyle()
     }
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/Cell/ProductCategoryTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/Cell/ProductCategoryTableViewCell.swift
@@ -15,6 +15,7 @@ final class ProductCategoryTableViewCell: UITableViewCell {
     override func awakeFromNib() {
         super.awakeFromNib()
         applyDefaultBackgroundStyle()
+        styleSelection()
         styleLabels()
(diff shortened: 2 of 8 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/ProductCategoryListViewModel.swift
```diff
@@ -26,6 +26,10 @@ final class ProductCategoryListViewModel {
     ///
     private let product: Product
 
+    /// Product categories that will be eventually modified by the user
+    ///
+    private var newSelectedCategories: [ProductCategory]
```
jaclync, line 31: nit: I felt like `selectedCategories` is clear enough?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/ProductCategoryListViewModel.swift
```diff
@@ -80,6 +85,28 @@ final class ProductCategoryListViewModel {
...
+            newSelectedCategories.remove(at: indexCategory)
+        }
+        else if let category = resultController.fetchedObjects.first(where: { $0.categoryID == categoryViewModel.categoryID }) {
+            newSelectedCategories.append(category)
+        }
+
+        updateViewModelsArray()
+    }
+
+    func hasCategoriesChanges() -> Bool {
```
jaclync, line 107: nit: maybe `haveCategoriesChanged` or `hasUnsavedChanges`? it doesn't seem to be used yet?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/Cell/ProductCategoryTableViewCell.swift
```diff
@@ -43,6 +36,20 @@ final class ProductCategoryTableViewCell: UITableViewCell {
     }
 }
 
+private extension ProductCategoryTableViewCell {
+    private func styleSelection() {
```
jaclync, line 40: nit: `private` not needed? (and same for the two style functions below :) )

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/Cell/ProductCategoryTableViewCell.swift
```diff
@@ -43,6 +36,20 @@ final class ProductCategoryTableViewCell: UITableViewCell {
...
+        self.selectionStyle = .none
```
jaclync, line 41: nit: `self` probably isn't necessary?

Networking/Networking/Model/Product/ProductCategory.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 /// Represents a ProductCategory entity.
 ///
-public struct ProductCategory: Decodable {
+public struct ProductCategory: Decodable, GeneratedCopiable {
```
jaclync, line 6: nit: is the `ProductCategory`'s `copy` function used in this PR?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/ProductCategoryListViewModel.swift
pmusolino: Thanks for the suggestion, I used `hasUnsavedChanges`, and Yes, it will be used in the next PR.

Networking/Networking/Model/Product/ProductCategory.swift
pmusolino: I'll use it in the next PR in my plans, but since it's not used here I'm removing it for the moment

PR #2437, opened 2020-06-17
External products: edit SKU

jaclync:
Edit SKU for #2200

- [x] ⚠️ Make sure PR base is `develop` after https://github.com/woocommerce/woocommerce-ios/pull/2432 is merged ⚠️ 

## Changes

Originally, I created a new view controller for the Edit SKU screen as the SKU field design was a bit different from the inventory settings. Then I clarified with G, and we are reusing the SKU field UI for both Edit Inventory Settings (SKU plus other inventory fields) and Edit SKU screens. Since there is error handling when the SKU is already used by another product, I decided to reuse `ProductInventorySettingsViewController` for Edit SKU screen. I'm open to any other suggestions!

- In `ProductInventorySettingsViewController`, added an enum `FormType` that indicates the type of form this view controller is used for: inventory settings (SKU plus other fields), or SKU form (SKU field only). The differences between these 2 form types are:
  - Navigation bar title
  - In `viewDidAppear`, the keyboard is prompted for Edit SKU form
  - The sections for the table view: Edit SKU form only has the SKU section
- In `ProductFormViewModel`, added a function to update SKU + test case
- In `ProductFormViewController`, implemented SKU editing action

## Testing

- Go to the Products tab
- Tap on an external/affiliate Product
- If the product already has a SKU, tap on the SKU row; otherwise, tap on the "Add more details" CTA at the bottom, then tap on the SKU row --> the SKU screen should appear with keyboard up
- Enter a SKU that is used by another product --> an error should appear
- Enter a new valid SKU and tap "Done"
- Tap "Update" --> the product SKU should be updated remotely
- Go back to the Products tab
- Tap on a simple product
- Tap on the inventory row --> the inventory settings screen should appear without the keyboard up

## Example screenshots/screencast

simulator

existing SKU | empty SKU | duplicate SKU
-- | -- | --
Simulator Screen Shot - iPhone 11 - 2020-06-17 at 09 16 10 | Simulator Screen Shot - iPhone 11 - 2020-06-17 at 09 16 15 | Simulator Screen Shot - iPhone 11 - 2020-06-17 at 09 16 23

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventorySettingsViewController.swift
```diff
@@ -97,37 +116,53 @@ private extension ProductInventorySettingsViewController {
...
-        ]
+        self.sections = sections
 
         tableView.reloadData()
     }
+
+    func createSKUSection() -> Section {
+        if let error = error {
+            return Section(errorTitle: error.alertMessage,
+                           rows: [.sku])
```
pmusolino, line 148: I would write everything on the same line of code in this case.

jaclync: yup! updated in: https://github.com/woocommerce/woocommerce-ios/pull/2437/commits/c2f702290f49a21411d46a5074200ff271a74bab

jaclync:
> I noticed that the SKU cell inside the bottom sheet seems not much responsive to the touch. Sometimes I need to tap it two times 🤔

@pmusolino which type of device/simulator did you encounter this issue? I tested on iPhone 11 (iOS 13) and iPhone Xs (iOS 12) simulators, and I was able to tap the SKU row in the bottom sheet from various touch areas 🤔

PR #2441, opened 2020-06-17
Fixed tracking events opened/closed called multiple times 

### LINKED ISSUE #2440: Application opened event possibly firing too much
Looking at activity of the application_opened event, I'm seeing nearly double the events than Android. Comparing to Apple's usage statistics we should be more in line with Android's counts for active sessions in a day. Take a bit of time to test the application_opened event to make sure it's firing correctly - especially when considering push notifications.

pmusolino:
Fixes #2440 

Investigating the issue #2440 seems that the events `application_opened` and `application_closed` are called two times every time they are fired because we start to observe notifications two times if the user has opted in the analytics.
To make sure to not register the notifications two times, I remove all the observers before adding a new one.

## Testing
- Launch the app, and make sure that in the console you are able to see only one `application_opened` event. 
- Send the app in foreground, make sure that the event `application_closed` is fired only one time. (if you relaunch the app, you will notice another `application_opened` event)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Analytics/WooAnalytics.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalytics.swift
@@ -139,7 +139,8 @@ private extension WooAnalytics {
         guard userHasOptedIn == true else {
             return
         }
-
+        
+        stopObservingNotifications()
         NotificationCenter.default.addObserver(self,
                                                selector: #selector(trackApplicationOpened),
                                                name: UIApplication.didBecomeActiveNotification,
--- a/WooCommerce/Classes/Analytics/WooAnalytics.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalytics.swift
@@ -118,13 +118,11 @@ extension WooAnalytics {
         userHasOptedIn = !optedOut
 
         if optedOut {
-            stopObservingNotifications()
             analyticsProvider.clearEvents()
             analyticsProvider.clearUsers()
             DDLogInfo("🔴 Tracking opt-out complete.")
         } else {
             refreshUserData()
-            startObservingNotifications()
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/Analytics/WooAnalytics.swift
```diff
@@ -140,6 +140,7 @@ private extension WooAnalytics {
             return
         }
 
+        stopObservingNotifications()
```
astralbodies, line 143: I believe the better course of action here might be to only call `startObservingNotifications()` once. I understand why we conditionally wrap the calls to loop into the NSNotificationCenter but in the end the `track()` method will ignore those events.

astralbodies: So that means removing all of the `stopObservingNotifications()`, leave the one `startObservingNotifications()` in the `initialize()` method, and then add a `deinit` to call `stopObservingNotifications()`.

astralbodies (changes requested):

WooCommerce/Classes/Analytics/WooAnalytics.swift
pmusolino: You are right, I didn't notice that there was a check on the `track()` method. 
Since we target >= iOS 12.0 and we use `addObserver:selector:name:object:`, we don't need to unregister the observer inside the `deinit` method. The system will take care of it. It would have been different if we were using `addObserverForName:object:queue:usingBlock:`. 👍 
Can you take a second look at it?

astralbodies (approved):
I didn't know about the iOS 12 thing with observing - nice! :shipit:

PR #2442, opened 2020-06-18
External products: external link & button text

### LINKED ISSUE #2200: Products: edit an existing external product

jaclync:
"editable product link" for #2200 

## Changes

- Encoded `externalURL` and `buttonText` for updating an external product
- Added a function to update `externalURL` and `buttonText` in `ProductFormViewModel` + test case
- Created `ProductExternalLinkViewController` that enabled editing the external URL and button text fields for an external product
- Created an enum `TextFieldContentAlignment` for different ways of text field content alignment for the text and placeholder (I was debating between this name and `TextFieldTextAlignment`, lemme know if you think another name would sound more clear!)
  - Because `NSTextAlignment` only has `left`/`right` instead of `leading`/`trailing` that takes LTR/RTL layout into account, this enum is mapped to `NSTextAlignment` based on the app layout
- Added `keyboardType` and `textFieldAlignment` parameters to `TitleAndTextFieldTableViewCell` for customizations in the external link screen
- Enabled editing external link in `ProductFormViewController`

## Testing

- Go to the Products tab
- Tap on an external product
- Tap on the Product link row --> should see the screen with two text fields for external URL and button text as in the screenshots
- Edit the external URL and button text
- (If not on iOS 13.4 simulator/device) Tap "<" in the navigation bar --> a discard changes prompt should be shown
- Tap "Cancel" to keep editing
- Tap "Done" to save changes
- Tap "Update" --> the external URL and button text should be updated remotely

## Example screenshots

simulator

empty state | filled state
-- | --
Simulator Screen Shot - iPhone 11 - 2020-06-18 at 10 34 25 | Simulator Screen Shot - iPhone 11 - 2020-06-18 at 10 34 48
Simulator Screen Shot - iPhone 11 - 2020-06-18 at 10 35 33 | Simulator Screen Shot - iPhone 11 - 2020-06-18 at 10 35 28

Simulator Screen Shot - iPhone 11 - 2020-06-18 at 10 33 48

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
@@ -13,6 +13,13 @@ final class ProductExternalLinkViewController: UIViewController {
     private var externalURL: String?
     private var buttonText: String
 
+    private lazy var keyboardFrameObserver: KeyboardFrameObserver = {
+        let keyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
+            self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
+        }
+        return keyboardFrameObserver
+    }()
+
     typealias Completion = (_ externalURL: String?, _ buttonText: String) -> Void
     private let onCompletion: Completion
 
@@ -44,6 +51,7 @@ final class ProductExternalLinkViewController: UIViewController {
         configureNavigation()
         configureMainView()
         configureTableView()
+        startListeningToNotifications()
     }
 }
 
@@ -173,6 +181,22 @@ private extension ProductExternalLinkViewController {
 
     func registerTableViewCells() {
         tableView.register(TitleAndTextFieldTableViewCell.loadNib(), forCellReuseIdentifier: TitleAndTextFieldTableViewCell.reuseIdentifier)
+    }
+}
+
+// MARK: - Keyboard management
+//
+private extension ProductExternalLinkViewController {
+    /// Registers for all of the related Notifications
+    ///
+    func startListeningToNotifications() {
+        keyboardFrameObserver.startObservingKeyboardFrame()
+    }
+}
+
+extension ProductExternalLinkViewController: KeyboardScrollable {
+    var scrollable: UIScrollView {
+        return tableView
     }
 }
 
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
@@ -125,8 +125,9 @@ private extension ProductExternalLinkViewController {
     }
 
     func configureExternalURL(cell: TitleAndTextFieldTableViewCell) {
-        let title = NSLocalizedString("Product URL", comment: "")
-        let placeholder = NSLocalizedString("Enter URL", comment: "")
+        let title = NSLocalizedString("Product URL", comment: "Title of the text field for editing the external URL for an external/affiliate product")
+        let placeholder = NSLocalizedString("Enter URL",
+                                            comment: "Placeholder of the text field for editing the external URL for an external/affiliate product")
         let viewModel = TitleAndTextFieldTableViewCell.ViewModel(title: title,
                                                                  text: externalURL,
                                                                  placeholder: placeholder,
@@ -139,8 +140,9 @@ private extension ProductExternalLinkViewController {
     }
 
     func configureButtonText(cell: TitleAndTextFieldTableViewCell) {
-        let title = NSLocalizedString("Button Text", comment: "")
-        let placeholder = NSLocalizedString("Enter button text", comment: "")
+        let title = NSLocalizedString("Button Text", comment: "Title of the text field for editing the button text for an external/affiliate product")
+        let placeholder = NSLocalizedString("Buy product",
+                                            comment: "Placeholder of the text field for editing the button text for an external/affiliate product")
         let viewModel = TitleAndTextFieldTableViewCell.ViewModel(title: title,
                                                                  text: buttonText,
                                                                  placeholder: placeholder,
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
@@ -13,12 +13,9 @@ final class ProductExternalLinkViewController: UIViewController {
     private var externalURL: String?
     private var buttonText: String
 
-    private lazy var keyboardFrameObserver: KeyboardFrameObserver = {
-        let keyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
-            self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
-        }
-        return keyboardFrameObserver
-    }()
+    private lazy var keyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
+        self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
+    }
 
     typealias Completion = (_ externalURL: String?, _ buttonText: String) -> Void
     private let onCompletion: Completion
@@ -51,7 +48,7 @@ final class ProductExternalLinkViewController: UIViewController {
         configureNavigation()
         configureMainView()
         configureTableView()
-        startListeningToNotifications()
+        startObservingKeyboardNotifications()
     }
 }
 
@@ -191,7 +188,7 @@ private extension ProductExternalLinkViewController {
 private extension ProductExternalLinkViewController {
     /// Registers for all of the related Notifications
     ///
-    func startListeningToNotifications() {
+    func startObservingKeyboardNotifications() {
         keyboardFrameObserver.startObservingKeyboardFrame()
     }
 }
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
@@ -1,0 +1,208 @@
+import UIKit
+import Yosemite
+
+/// Contains text fields for editing an external product's external URL and button text.
+final class ProductExternalLinkViewController: UIViewController {
+    private lazy var tableView: UITableView = {
+        return UITableView(frame: .zero, style: .grouped)
+    }()
+
+    private let product: Product
+    private let sections: [Section]
+
+    private var externalURL: String?
+    private var buttonText: String
+
+    typealias Completion = (_ externalURL: String?, _ buttonText: String) -> Void
+    private let onCompletion: Completion
+
+    init(product: Product, onCompletion: @escaping Completion) {
+        self.product = product
+        self.externalURL = product.externalURL
+        self.buttonText = product.buttonText
+        self.onCompletion = onCompletion
+
+        let externalURLFooter = NSLocalizedString("Enter the external URL to the product.",
+                                                  comment: "Footer text for editing product external URL")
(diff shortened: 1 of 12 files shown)


jaclync:
> if users add the button url but not the button text, is the product visible? I just remembe that's why initially I had a default "Buy now" button text already filled out instead of the "Enter button text" placeholder, can we please go back to this behavior?

@Garance91540 I just tested the behavior in core, "Buy product" is shown on the shop page (at least for my site) when the button text field is empty with "Buy product" as its placeholder. I think it'd be nice to follow core, to update the button text field placeholder to "Buy product"? Automatically filling in the "Buy now" or "Buy product" button text feels a bit surprising if I were the user 🤔  Also, the site could have a different localization setting from the device.

Screen Shot 2020-06-18 at 4 29 12 PM

> Do we display an error message under the "add link" text field if users enter an invalid url?

Good point! we could add some local validation for the URL, but it wouldn't cover all possible malformed URLs or URLs that have no response or any network errors. Still, it is better than allowing any text. Do we have design for the error state? (maybe like the duplicate SKU in the inventory settings)

WooCommerce/Classes/ViewRelated/ReusableViews/TextFieldContentAlignment.swift
```diff
@@ -0,0 +1,22 @@
+import UIKit
+
+/// Horizontal alignment for content (text or placeholder) in a text field.
+enum TextFieldContentAlignment {
```
shiki, line 4: ⚠️  As you implied, the name is a bit confusing. 😬 But I think what would be more valuable is to explain the use cases of this `enum`. Why and when would we want to use this instead of just `NSTextAlignment`? What do you think about adding a comment about that?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
```diff
@@ -0,0 +1,232 @@
...
+        tableView.register(TitleAndTextFieldTableViewCell.loadNib(), forCellReuseIdentifier: TitleAndTextFieldTableViewCell.reuseIdentifier)
+    }
+}
+
+// MARK: - Keyboard management
+//
+private extension ProductExternalLinkViewController {
+    /// Registers for all of the related Notifications
+    ///
+    func startListeningToNotifications() {
```
shiki, line 192: I would not be opposed to using a longer but more specific name so when reading this inside `viewDidLoad`, it's much clearer what to expect. 

```suggestion
    func startObservingKeyboardNotifications() {
```

I'd also argue that it's probably not worth it to create a method for this since it's just one line. It probably makes the class easier to read since it's shorter? 🤷 Up to you though.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
```diff
@@ -0,0 +1,232 @@
...
+
+    private var externalURL: String?
+    private var buttonText: String
+
+    private lazy var keyboardFrameObserver: KeyboardFrameObserver = {
+        let keyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
+            self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
+        }
+        return keyboardFrameObserver
+    }()
```
shiki, line 21: You could make this shorter by:

```suggestion
    private lazy var keyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
        self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
    }

```

WooCommerce/Classes/ViewRelated/ReusableViews/TextFieldContentAlignment.swift
```diff
@@ -0,0 +1,22 @@
...
+    case leading
+    case trailing
+    case center
+}
+
+extension TextFieldContentAlignment {
+    func toTextAlignment(layoutDirection: UIUserInterfaceLayoutDirection = UIApplication.shared.userInterfaceLayoutDirection) -> NSTextAlignment {
```
shiki, line 11: Were you planning to add a test for this? Is that why you exposed the `layoutDirection` argument?

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormViewModel+UpdatesTests.swift
```diff
@@ -225,6 +225,25 @@ final class ProductFormViewModel_UpdatesTests: XCTestCase {
         // Assert
         XCTAssertEqual(viewModel.product.sku, sku)
     }
+
+    func testUpdatingExternalLink() {
+        // Arrange
+        let product = MockProduct().product(sku: "")
```
shiki, line 231: You probably just wanted: 

```suggestion
        let product = MockProduct().product()
```

😁

shiki:
By the way, good separation of commits in this PR. 👍

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
```diff
@@ -0,0 +1,208 @@
...
+        case let cell as TitleAndTextFieldTableViewCell where row == .buttonText:
+            configureButtonText(cell: cell)
+        default:
+            fatalError("Unexpected row: \(row)")
+        }
+    }
+
+    func configureExternalURL(cell: TitleAndTextFieldTableViewCell) {
+        let title = NSLocalizedString("Product URL", comment: "")
+        let placeholder = NSLocalizedString("Enter URL", comment: "")
```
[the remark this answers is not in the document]
jaclync: oops I tend to test out the UI and forgot to add a comment 🤦🏻‍♀️  thanks for catching this! updated in https://github.com/woocommerce/woocommerce-ios/pull/2442/commits/35c60bafe9089c6d403129dfff1e56ac490dd72b

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
```diff
@@ -0,0 +1,208 @@
...
+                                                                 textFieldAlignment: .trailing,
+                                                                 onTextChange: { [weak self] text in
+                                                                    self?.externalURL = text
+        })
+        cell.configure(viewModel: viewModel)
+    }
+
+    func configureButtonText(cell: TitleAndTextFieldTableViewCell) {
+        let title = NSLocalizedString("Button Text", comment: "")
+        let placeholder = NSLocalizedString("Enter button text", comment: "")
```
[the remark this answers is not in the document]
jaclync: haha yea, thanks for catching this! updated in https://github.com/woocommerce/woocommerce-ios/pull/2442/commits/35c60bafe9089c6d403129dfff1e56ac490dd72b

WooCommerce/Classes/ViewRelated/ReusableViews/TextFieldContentAlignment.swift
jaclync: I renamed this to `TextFieldTextAlignment` and updated the comment in https://github.com/woocommerce/woocommerce-ios/pull/2442/commits/1e378d324ae0088c28309c6471d46615905e6988 - lemme know if any part can still be improved!

WooCommerce/Classes/ViewRelated/ReusableViews/TextFieldContentAlignment.swift
jaclync: yes I did plan to add tests for this, but thought this mapping feels straightforward. it's easy to add tests though, so I added them in https://github.com/woocommerce/woocommerce-ios/pull/2442/commits/1e378d324ae0088c28309c6471d46615905e6988

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
jaclync: nice! updated in https://github.com/woocommerce/woocommerce-ios/pull/2442/commits/8a9971de59fadf9b837a9e21d92a724e23627d93

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
jaclync: I think it's still nice having a descriptive function even though the code is just one line. I like the name you suggested, updated in https://github.com/woocommerce/woocommerce-ios/pull/2442/commits/8a9971de59fadf9b837a9e21d92a724e23627d93

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormViewModel+UpdatesTests.swift
jaclync: cool! updated in https://github.com/woocommerce/woocommerce-ios/pull/2442/commits/bc3de89a077dd574de02c7cb19a2e9c71e17b368

jaclync:
@shiki thanks for the review! I updated the PR and responded to all your comments for another look 😄 

> Are you planning to work on the URL validation on another PR? That sounds better to make this simpler

yes, I added a subtask for URL validation to #2200 description and will handle this in a separate PR.

@Garance91540 I updated the button text placeholder following core:

Garance91540:
> @Garance91540 I updated the button text placeholder following core:

@jaclync Great! can we make the "Buy Product" color iOS System Label/ Primary Color? (right now it seems like it's disabled)

jaclync:
> @jaclync Great! can we make the "Buy Product" color iOS System Label/ Primary Color? (right now it seems like it's disabled)

@Garance91540 it's the disabled color since I set "Buy product" as the placeholder for the button text field, just like in core. did you mean we want to fill out the button text if it's currently empty, which is then saved to the product if the user taps "Done" on this screen? If so, I have some concerns about this:

- If a product doesn't have the button text field set before coming to this screen, and we automatically set this field, then it has unsaved changes immediately. since we show the discard changes prompt when they try exiting the screen without saving, it could be confusing if the user hasn't manually changed anything
- I felt leaving the default button text to core (the default text that is shown on the shop page if the button text field isn't set) would be more ideal. If our assumption of the default button text isn't the same as in core for certain languages, then it could be confusing if they have different default button text for various external products

WooCommerce/Classes/ViewRelated/ReusableViews/TextFieldContentAlignment.swift
shiki: Awesome! The description and the unit tests really help!

jaclync:
@Garance91540 since the PR has been reviewed and tested, also with WWDC happening, I'm leaving the remaining followup tasks from this PR to 2 separate subtasks for #2200 🙂 

the 2 remaining subtasks (also in the issue description):
- editing product link followup: local validation for external URL
- editing product link followup: auto-fill button text with "Buy product" when the external URL field becomes non-empty

Issue #2364, opened 2020-05-29
Log messages/crashes in Storage framework


jaclync:
Since the `CrashLogging` class that logs messages to Sentry is in the Tracks framework, and the message in `fatalError` does not show up in the crash log in Xcode/Sentry, it'd be great if we could log these fatal errors with debug message to Sentry.

shiki:
Since we now have Encrypted Logging, we can make use of `DDLogError()`. https://github.com/woocommerce/woocommerce-ios/pull/2867 shows how this can be on in the Storage framework. I _think_ it will work.

joshheald:
@jaclync it looks like your #2366 PR should have closed this issue. Is the `CrashLogger.logFatalErrorAndExit` sufficient, or should it be changed to use `DDLogError` as @shiki suggests, do you think?

jaclync:
It looks like the PR https://github.com/woocommerce/woocommerce-ios/pull/2366 meant to close this issue but because it was targeting a release branch and not `trunk`/`develop` this issue remains open 😅 closing now, thanks for following up on this old issue Josh!

PR #2447, opened 2020-06-18
Call reloadData after performFetch executions

Ecarrion:
Related: https://github.com/woocommerce/woocommerce-ios/issues/1543

# Why

For the past 7 months(or more!) users have been [experiencing a crash](https://github.com/woocommerce/woocommerce-ios/issues/1543) where a table view ended with an inconsistent state after updating/deleting rows.

After much experimentation, @shiki have been able to reproduce a [scenario](https://gist.github.com/shiki/9c5627b23eed8085c297cf6aeae45a62) where this crash happens. 

Our hypothesis is that the table has the wrong `ResultsController` state initially and then receiving new events from FRC makes it inconsistent and crash.

# How

In order to prevent this from happening we need to make sure that the table view is in sync with the `ResultsController` immediately after a `performFetch` is executed.

So after all `performFetch` executions we have made sure that a `reloadData` is being called.

# Testing Steps
- Go to orders module and verify that everything is loading correctly
- Go to reviews module and verify that everything is loading correctly
- Go to order search module and verify that everything is loading correctly
- Go to order status module and verify that everything is loading correctly
- Go to order details module and verify that everything is loading correctly
- Go to product module(initial load, filter, sort) and verify that everything is loading correctly
- Go to product shipping classes and tax classes and verify that everything is loading correctly

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ListSelector/PaginatedListSelectorViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/ListSelector/PaginatedListSelectorViewController.swift
@@ -2,6 +2,8 @@ import UIKit
 import WordPressUI
 import Yosemite
 
+import class AutomatticTracks.CrashLogging
+
 /// A generic data source for the paginated list selector UI `PaginatedListSelectorViewController`.
 ///
 protocol PaginatedListSelectorDataSource {
@@ -387,6 +389,10 @@ private extension PaginatedListSelectorViewController {
             onReload()
         }
 
-        try? resultsController.performFetch()
+        do {
+            try resultsController.performFetch()
+        } catch  {
+            CrashLogging.logError(error)
+        }
     }
 }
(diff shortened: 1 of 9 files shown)


shiki:
@loremattei Hi! We would like to get this merged into 4.5 since it can potentially fix one of our topmost crashes. Is it still okay to merge this? 

---

@jaclync @pmusolino Please review whenever you get the time. 🙂

WooCommerce/Classes/ViewRelated/ListSelector/PaginatedListSelectorViewController.swift
```diff
@@ -387,6 +389,12 @@ private extension PaginatedListSelectorViewController {
             onReload()
         }
 
-        try? resultsController.performFetch()
+        do {
+            try resultsController.performFetch()
+        } catch {
+            CrashLogging.logError(error)
```
jaclync, line 395: non-blocking thought: once we consolidate the error logging after https://github.com/woocommerce/woocommerce-ios/issues/2364, it'd be nice to use the crash logger in the app layer as well

Ecarrion:
Thanks for the review @jaclync !
And heads up to @loremattei that this has been merged :-)

loremattei:
Thanks for the ping @Ecarrion! 
A beta was released yesterday, so I think I'll release the new one on Monday.

Issue #1372, opened 2019-10-17
Coredata: Transformer errors on iOS 13


ctarda:
Discussed in p99K0U-1RH-p2
When running on iOS 13, we get a generic error in the console:

```
[error] fault: One or more models in this application are using transformable properties with transformer names that are either unset, or set to NSKeyedUnarchiveFromDataTransformerName. Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead. At some point, Core Data will default to using "NSSecureUnarchiveFromData" when nil is specified, and transformable properties containing classes that do not support NSSecureCoding will become unreadable.
```
followed by more specific errors pointing to Transformable attributes, like:
```
CoreData: warning: Property 'upsellIDs' on Entity 'Product' is using nil or an insecure NSValueTransformer.  Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead.
```

jaclync:
From Xcode 11.4, iOS 13.4, migrating from model version 26 to 28:

```
[error] fault: One or more models in this application are using transformable properties with transformer names that are either unset, or set to NSKeyedUnarchiveFromDataTransformerName. Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead. At some point, Core Data will default to using "NSSecureUnarchiveFromData" when nil is specified, and transformable properties containing classes that do not support NSSecureCoding will become unreadable.
CoreData: fault: One or more models in this application are using transformable properties with transformer names that are either unset, or set to NSKeyedUnarchiveFromDataTransformerName. Please switch to using "NSSecureUnarchiveFromData" or a subclass of NSSecureUnarchiveFromDataTransformer instead. At some point, Core Data will default to using "NSSecureUnarchiveFromData" when nil is specified, and transformable properties containing classes that do not support NSSecureCoding will become unreadable.
[...]

PR #2453, opened 2020-06-22
Fix the model name in debug messages for Core Data crash investigation

jaclync:
Logging improvement for #2371 for release 4.5

## Changes

I took a look at the [log events for the Core Data crash from release 4.5 beta build](https://sentry.io/organizations/a8c/issues/1711092978/events/?project=1458804&query=release%3A4.5.0.0&statsPeriod=14d) where we started logging the migration process, and noticed that the app was migrating first from model v0 (the first version) to model v1:
```
⚠️ [CoreDataManager] Migration required for persistent store
⚠️ Attempting migration from Model to Model 2
☠️ [CoreDataManager] Unable to migrate store.
```

 It's quite unlikely that the app was first on model v0, and I saw that we were using the index from the array of **models to migrate** (model versions between the source and destination models) for the array of all model names.

This PR fixed the model name by first finding the index of the source and destination models in the array of model names, and then got the model name from the index.

## Testing

This can be tested by just console logging, like `print`ing the `debugMessages` after this line:

https://github.com/woocommerce/woocommerce-ios/blob/b4d71241de758523b520150360c5473f440072f5/Storage/Storage/CoreData/CoreDataManager.swift#L211

- Launch the app, logged in, at an earlier Core Data model version (e.g. from [release 3.9](https://github.com/woocommerce/woocommerce-ios/releases/tag/3.9))
- Launch the app with the PR branch --> the debug messages should include a message attempting migration from the expected model versions (e.g. from my testing from 3.9.0.0 to this branch, I got `["⚠️ [CoreDataManager] Migration required for persistent store", "⚠️ Attempting migration from Model 26 to Model 27", "⚠️ Attempting migration from Model 27 to Model 28"]`)

Update release notes:

### DIFF
--- a/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
+++ b/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
@@ -92,8 +92,13 @@ public struct CoreDataIterativeMigrator {
                     return (false, debugMessages)
             }
 
+            guard let modelFromIndex = objectModels.firstIndex(of: modelFrom),
+                let modelToIndex = objectModels.firstIndex(of: modelTo) else {
+                    return (false, debugMessages)
+            }
+
             // Migrate the model to the next step
-            let migrationAttemptMessage = "⚠️ Attempting migration from \(modelNames[index]) to \(modelNames[index + 1])"
+            let migrationAttemptMessage = "⚠️ Attempting migration from \(modelNames[modelFromIndex]) to \(modelNames[modelToIndex])"
             debugMessages.append(migrationAttemptMessage)
             DDLogWarn(migrationAttemptMessage)
 


jaclync:
@woocommerce/ios-developers @loremattei it'd be great if we could include this for release 4.5, otherwise we wouldn't know the model versions that are causing issues without the correct model names since we early return whenever a migration fails. lemme know if you have other thoughts!

jaclync:
@pmusolino just checking, did you manually print a message to the console? if not, no debug messages are shown by default since they are only logged to Sentry when Core Data fails to set up. feel free to share your diffs on the PR branch so I can look into why the message isn't logged :)

> a lot of warnings about Core Data

we have an open issue for these warnings: https://github.com/woocommerce/woocommerce-ios/issues/1372. we thought these weren't harming the app with some research at the time, so we haven't addressed them yet. I didn't see the mention of crashes on the post you linked 🤔  From Sentry https://github.com/woocommerce/woocommerce-ios/issues/2371, more than 99% of the Core Data crashes were from not being able to migrate from one version to another.

loremattei:
> it'd be great if we could include this for release 4.5

@jaclync sounds good, though since we want to release earlier this week, it would be great if we manage to get this merged asap.

pmusolino:
> did you manually print a message to the console

this was the missing part of my testing process 🌊 I'll retry.

> I didn't see the mention of crashes on the post you linked

The link mention only how to solve the problem. 
Here some link to people who encountered some crashes:
https://stackoverflow.com/questions/57144796/coredata-crash-error-xcode-11-beta-ios-13-beta
https://stackoverflow.com/questions/59722771/core-data-error-switch-to-using-nssecureunarchivefromdata-or-a-subclass-of-nss
https://developer.apple.com/forums/thread/123266
 

> more than 99% of the Core Data crashes were from not being able to migrate from one version to another.

👍 👍

jaclync:
thanks for reviewing & testing this again quickly, @pmusolino ! 🙏 merging now, @loremattei thank you for wrangling another build!

PR #2456, opened 2020-06-22
Run SwiftLint on Build Phase

shiki:
Ref paaHJt-14G-p2. 

This adds a new build phase in the WooCommerce target to run SwiftLint. 

image

### SwiftLint Dependency

In the post, we discussed using CocoaPods as a dependency manager for SwiftLint. I abandoned that idea in favor of using the binary installed by running [`rake dependencies`](https://github.com/woocommerce/woocommerce-ios/blob/a60d39bca5e065db153a0f38379eb3a73e176adf/Rakefile#L94-L120). I concluded that changing how SwiftLint is installed should be a _different discussion_. and this PR should only focus on adding the Build Phase to keep things simple. 

There is also the possibility that we're not going to like the Build Phase linting so we might just end up canceling everything altogether. 😐 

### CircleCI

Based on what I understood, CircleCI doesn't run SwiftLint. Rather, it's Hound that runs it and that is configured by [`.hound.yml`](https://github.com/woocommerce/woocommerce-ios/blob/develop/.hound.yml) which uses the [`.swiftlint.yml`](https://github.com/woocommerce/woocommerce-ios/blob/develop/.swiftlint.yml) configuration. With this, I opted to still skip running SwiftLint in CircleCI by adding an environment check: 

https://github.com/woocommerce/woocommerce-ios/blob/2ac594deefb92c7b6c3414f2fe2a23a062435bbe/Scripts/build-phases/swiftlint.sh#L9-L13

In the future, we can consider setting up SwiftLint in CircleCI so that if there are lint errors, the build is immediately aborted. 

### SwiftLint Autocorrect

I played around with using `swiftlint autocorrect`. It works in the build phase. However, as @mokagio pointed out, it destroys the undo/redo stack. I believe that keeping the stack is quite important especially if you're actively writing the code so I abandoned that. 

### Linting Changed Files Only

Running SwiftLint on all the files specified in the [`include` argument](https://github.com/woocommerce/woocommerce-ios/blob/2ac594deefb92c7b6c3414f2fe2a23a062435bbe/.swiftlint.yml#L1-L6) adds `0.4` seconds to the build. I tried (timeboxed) setting it up so that only the changed files are listed. 

https://github.com/woocommerce/woocommerce-ios/blob/2ac594deefb92c7b6c3414f2fe2a23a062435bbe/Scripts/build-phases/swiftlint.sh#L31-L43

It adds some complexity to the script but the time spent is only 0.1 seconds if there are no changed files and between 0.2 and 0.3 seconds if there are changed files. I guess that's a win so I kept it. 🤷 

### Failing the Build on Lint Errors

Currently, if SwiftLint fails with a _serious error_, the Xcode build is aborted. This will probably get annoying and interrupt our focus when coding. If we want, we can consider changing our config so they all spit out warnings instead of errors. For example, this `vertical_whitespace` rule sounds like it should just be a warning instead of an error: 

https://github.com/woocommerce/woocommerce-ios/blob/a60d39bca5e065db153a0f38379eb3a73e176adf/.swiftlint.yml#L138-L140

I didn't focus on fixing this too much as I think it's best if we experiment with what we have in here first and improve accordingly. 

## Testing

1. Use this branch. 
2. Build. Confirm that there are no errors. 
3. Modify the code so that you'll get a line length violation. For example, you can add a long class name:

    ```swift
    class SomethingSomethinSomethingSomethinSomethingSomethinSomethingSomethinggggSomethingSomethinSomethingSomethinSomethingSomethinSomethingSomethinSomethingSomethinSomethingSomethinggggggSomethingSomething {

    }
    ```
4. Build. Confirm that the build is successful but you'll get a warning about the line length.
5. Add more than 3 empty lines on an existing swift file.
6. Build. Confirm that the build fails and you'll get an error about the maximum empty lines.
7. If you'd like to, play around with the other SwiftLint rules and see how they feel. 

## Reviewing 

One Platform 9 member and another Woo iOS member reviewing this would be neat. 

## Submitter Checklist


Scripts/build-phases/swiftlint.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+#
+# Runs SwiftLint on the whole workspace
```
mokagio, line 4: Nitpick, feel free to ignore. The comment below has a `.` 😳

```suggestion
# Runs SwiftLint on the whole workspace.
```

Scripts/build-phases/swiftlint.sh
```diff
@@ -0,0 +1,54 @@
...
+#
+# This does not run in Continuous Integration.
+#
+
+# Abort if we are running in CI
+# See https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables
+if [ "$CI" = true ] ; then
```
mokagio, line 11: +1 for using `$CI`, which seems to be the standard env var most CI providers use. That is, if we'll move away from Circle CI, we won't have to update this script 🙌.

mokagio: How do you feel about echoing something like the following, just to add a note about it in the CI logs?

```
echo "warning: skipping SwiftLint build phase because running on CI."
```

_I swear I saw someone use something different from warning or error to show just an info icon, but I can't find the syntax_ 😞

Scripts/build-phases/swiftlint.sh
```diff
@@ -0,0 +1,54 @@
...
+# are no matches. Xcode's build will fail if we don't do this.
+#
+MODIFIED_FILES=`git diff --name-only --diff-filter=d HEAD | grep -G "\.swift$" || true`
+echo $MODIFIED_FILES | xargs $SWIFTLINT --config $CONFIG_FILE --quiet
+MODIFIED_FILES_LINT_RESULT=$?
+
+# Run SwiftLint on the added files
+ADDED_FILES=`git ls-files --others --exclude-standard | grep -G "\.swift$" || true`
+echo $ADDED_FILES | xargs $SWIFTLINT --config $CONFIG_FILE --quiet
+ADDED_FILES_LINT_RESULT=$?
```
mokagio, line 43: This is pretty cool bash foo 👏

Scripts/build-phases/swiftlint.sh
shiki: Thanks! Added in 1896de78f06f9bf290bbe4e3d3ef6b37971d3d22.

Scripts/build-phases/swiftlint.sh
shiki: Thanks! Fixed in 1896de78f06f9bf290bbe4e3d3ef6b37971d3d22.

PR #2457, opened 2020-06-22
Fix: order_open event is now fired correctly

pmusolino:
It turned out that the event `order_open` is no longer fired, because we substituted the “segue”, pushing directly the order detail view controller. Since the `track:` method was called inside the segue method, it’s no more fired correctly. I fixed it removing the unused segue method and moving the `track:` method in the right place.

## Testing
1. From the Order List screen, tap an order. -> Check that in the console you are able to see the `order_open` event logged.
2. No app changes should be introduced, please feel free to sanity check that you are able to open an order detail from all the entry points.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift
@@ -497,7 +497,11 @@ extension OrdersViewController: UITableViewDelegate {
             assertionFailure("Expected OrderDetailsViewController to be instantiated")
             return
         }
-        orderDetailsVC.viewModel = viewModel.detailsViewModel(at: indexPath)
+        let orderDetailsViewModel = viewModel.detailsViewModel(at: indexPath)
+        orderDetailsVC.viewModel = orderDetailsViewModel
+
+        ServiceLocator.analytics.track(.orderOpen, withProperties: ["id": orderDetailsViewModel.order.orderID,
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift
```diff
@@ -497,7 +497,11 @@ extension OrdersViewController: UITableViewDelegate {
             assertionFailure("Expected OrderDetailsViewController to be instantiated")
             return
         }
-        orderDetailsVC.viewModel = viewModel.detailsViewModel(at: indexPath)
+        let orderDetailsViewModel = viewModel.detailsViewModel(at: indexPath)
+        orderDetailsVC.viewModel = orderDetailsViewModel
+
+        ServiceLocator.analytics.track(.orderOpen, withProperties: ["id": orderDetailsViewModel.order.orderID,
+        "status": orderDetailsViewModel.order.statusKey])
```
shiki, line 504: I'm not sure how I should feel that Hound didn't complain about this indentation but complains about extra spaces. 🙄

shiki (approved):
Thank you for fixing this!

WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift
pmusolino: In this case, I believe that the indentation is ok, and does not exceed 160 characters 👍

PR #2461, opened 2020-06-23
Fix: now we send the `application_upgraded` event correctly

### LINKED ISSUE #2455: The event `application_upgraded` is not correctly tracked
[Discussion here](https://a8c.slack.com/archives/C6H8C3G23/p1592573739215700).

While I was looking at our `application_installed` event, it seems that on iOS we are not tracking the event `application_upgraded` which is declared but never used. On the contrary, the `application_installed` event is used in both cases (install + update).
Since we want to update the behavior following the same did on Android, we need to use `application_upgraded` when the app is updated, and `application_installed` only when there is an update.

pmusolino:
Fixes #2455 

When the app is updated, we send erroneously the event `application_installed`. I fixed it in this PR.

## Testing
Try to upgrade from an old version to a new version of the app, and check that in the console you are able to see the `application_upgraded` event fired correctly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/AppDelegate.swift
+++ b/WooCommerce/Classes/AppDelegate.swift
@@ -331,7 +331,7 @@ private extension AppDelegate {
             ServiceLocator.analytics.track(.applicationInstalled)
         } else if versionOfLastRun != currentVersion {
             // App was upgraded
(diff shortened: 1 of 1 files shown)


shiki (approved):
✔️ Tested and LGTM!

Issue #1946, opened 2020-03-12
Stats banner update for Woo 4.0


anitaa1990:
There are a couple of changes needed in the app now that Woo 4.0 is released and includes WooCommerce Admin plugin.

1. If we know for sure the new stats endpoint is available, then default to the new stats UI instead of displaying a banner message.
2. Site does not have v4 wc-analytics namespace available OR Woo version is less than 4.0 - inform user that old stats will eventually go away, upgrade to 4.0 or install WCAdmin.

- [ ] if we want to check if the WC version, it is a new endpoint that we are going to implement: `/wc/v3/system_status?_fields=environment`

<img width="541" src="https://user-images.githubusercontent.com/22608780/76535968-b658b300-64a1-11ea-91ae-fd6bed185e9a.png">

AmandaRiu:
NOTE: Use September 1, 2020 as the cutoff date for old stats support.

shiki:
Ref: p91TBi-2tl-p2#comment-207

shiki:
Closed by #2467.

Issue #2465, opened 2020-06-23
Stats → Show Red Icon on Top of the Sync Icon in the Banner


shiki:
Continuation of #1946. 

<img src="https://user-images.githubusercontent.com/198826/85429177-98d4c400-b53b-11ea-8852-d8caa3f9ce3e.png" width="320">

Ecarrion:
Hey @Garance91540 

I tried to add the pink dot indicator by code but it turned out to be very tricky due to some dynamism between dark mode and RTL orientation.

We have decided that the best way to move forward is to get a composed image that includes the sync icon + the dot.
I can grab that from Figma for the light version, but could you do one for dark mode, please?

Thanks!

Garance91540:
Hi @Ecarrion sounds good, I added the dark mode designs in the Figma file.

Issue #2452, opened 2020-06-19
Remove Improved Stats from Experimental Features


shiki:
When #1946 is completed, the Improved Stats in the Experimental Features settings is probably no longer needed.

<img src="https://user-images.githubusercontent.com/198826/85148137-036cc380-b20d-11ea-93da-8712d2632c60.png" width="320">

PR #2467, opened 2020-06-23
Show stats v4 by default if possible

Ecarrion:
# Why
As part of https://github.com/woocommerce/woocommerce-ios/issues/1946 we will be updating the stats banner to let the user know that the v3 stats will be deprecated soon.

This PR changes the behavior to the following:
- If stats v4 are available, we default to show them and show no banner
- If stats v4 are not available, we show stats v3 dashboards and show a collapsable banner indicating the user that the v3 stats will be deprecated soon.

Changes in following PRs will include:
- Dismiss and remind me later actions
- Empty state screen if the user hasn't updated it's Woo version and September 1st has arrived.

# How

First, we have reduced the `StatsVersionState` enum to only have one case(`initial`) because we don't care anymore about transitioning between v4 or v3 states. In the future we aim to remove this enum and only store it's associated property `statsVersion`. While doing that we have also trimmed a lot of tests that weren't needed anymore. 

Second, we have added a new banner using `TopBannerView` to indicate the user that they are using and old and soon to be deprecated version of the stats. With this we have deleted the code associated with `V3ToV4` and `V4ToV3` banners.

# Screenshots

V3 | V4
--- | ---
v3|v4

# Testing steps
- Run the app using a store that doesn't have v4 stats available and see that the correct banner(indicating the user to upgrade) is being shown.
- Run the app using a store that has v4 stats available and see that the new stats dashboard is being shown and not banner is present.

# Questions
- The android designs have a black tone for the icon on the banner, other banners on the iOS app have a gray-ish tone, should we keep the gray tone for this banner too @Garance91540 ?

Update release notes:


WooCommerce/WooCommerceTests/Testing/XCTestCase+Wait.swift
```diff
@@ -0,0 +1,24 @@
...
+    /// waitForExpectation(timeout: TimeInterval(10)) { expectation in
+    ///     doSomethingInTheBackground {
+    ///         expectation.fulfill()
+    ///     }
+    /// }
+    /// ```
+    ///
+    func waitForExpectation(description: String? = nil,
+                            timeout: TimeInterval = Constants.expectationTimeout,
+                            _ block: (XCTestExpectation) -> ()) {
```
jaclync, line 19: thanks for adding this!

shiki:
> Just double checking, is the removal of v4 stats feature switch under Settings > Experimental Features a separate task?

@jaclync That's right. We're going to work on it in #2452.

Ecarrion:
> When there's no experimental feature available, instead of opening a blank screen, can we make the "Experimental Features" link disabled on the Settings page?

@Garance91540 How do you feel about hiding the "Experimental Features" row rather tan disabling it? I think that is more inline with how apple manages the settings on the phone.

Garance91540:
> @Garance91540 How do you feel about hiding the "Experimental Features" row rather tan disabling it? I think that is more inline with how apple manages the settings on the phone.

That works!

Ecarrion:
@Garance91540 I've updated the view to match the figma design. The red dot will be implemented in a different PR https://github.com/woocommerce/woocommerce-ios/issues/2465
new-aligment-text

Please review again :-)

PR #2468, opened 2020-06-23
Products (M3): Add a new Product Category

### LINKED ISSUE #2000: Edit Products: Manage Product Categories

pmusolino:
Resolves #2000 

## Description
This PR implements the last part for completing the management of Product Categories (part of Products Milestone 3). In particular, in this PR it's possible to create a new product category and to select a parent category.

## Changes
- The `addCloseNavigationBarButton:` method helper now accept also a custom title.
- Implemented the `AddProductCategoryViewController`
- Implemented the `ProductParentCategoriesViewController`
- `ProductCategoryListViewController` now contains the new “Add Button” on top of the view.
- Implemented a new method `addAndSelectedNewCategory:` in `ProductCategoryListViewModel`
- `TextFieldTableViewCell` now contains a new method `resignFirstResponder:`

## Testing
1. Open a product with or without product categories
2. Open the Product Categories screen
3. Tap "Add Category" button on top
4. Try to create a new category without parent category -> When it's saved, you will see it in category list selected. Save the product and make sure that it will be updated correctly.
5. Try to create a new category with parent category -> The new parent category should be selected in "New category" screen. Save the product and make sure that it will be updated correctly.
6. Try to create a new category where the title it's just a space ` `. -> Make sure to see an error notice which appears on the screen, and where you are able to retry.

## Add Product Category in action (GIF)
add product category

## Screenshots (dark mode enabled)
| Categories             |  Add Category | Select Parent Category |
Simulator Screen Shot - iPhone 11 Pro - 2020-06-23 at 18 31 01 | Simulator Screen Shot - iPhone 11 Pro - 2020-06-23 at 18 31 12 | Simulator Screen Shot - iPhone 11 Pro - 2020-06-23 at 18 31 17

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIViewController+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UIViewController+Helpers.swift
@@ -18,10 +18,15 @@ extension UIViewController {
         navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .plain, target: nil, action: nil)
     }
 
-    /// Show the X close button on the left bar button item position
+    /// Show the X close button or a custom close button with title on the left bar button item position
     ///
-    func addCloseNavigationBarButton(target: Any? = self, action: Selector? = #selector(dismissVC)) {
-        navigationItem.leftBarButtonItem = UIBarButtonItem(image: .closeButton, style: .plain, target: target, action: action)
+    func addCloseNavigationBarButton(title: String? = nil, target: Any? = self, action: Selector? = #selector(dismissVC)) {
+        if let title = title {
+            navigationItem.leftBarButtonItem = UIBarButtonItem(title: title, style: .plain, target: target, action: action)
+        }
+        else {
+            navigationItem.leftBarButtonItem = UIBarButtonItem(image: .closeButton, style: .plain, target: target, action: action)
+        }
     }
 
 }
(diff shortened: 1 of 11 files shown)


pmusolino:
Thank you so much @jaclync for your review. I implemented all your suggestions.

I also added the disclosure indicator.

pmusolino:
Thank you @Garance91540 for your review!

> Additionally, I think we should give more guidance to the users on the error so they can easily fix it.

I'd say this would be ideal, but our API doesn't return the specific reason why the API failed. They return a generic error, which is not really readable from a user. I added it in the description, but I'm not sure it's useful for the end-user. I'd say that in this case, we should show a generic error message until our API endpoints return better error messages for the users (but, they should also be localized).

Garance91540:
@pmusolino can we make this request to get more information on error cases from the API somewhere? 

Until this is possible, I think what you have work!

pmusolino:
Thanks @Garance91540. I opened a discussion here p7bje6-2c4. Going deeper, I noticed that I can handle the error, but I'd like to handle all the possible scenario 👍

PR #2469, opened 2020-06-23
Use main branch in .configure

shiki:
Ref paaHJt-13T-p2. 

After deleting `master` and replacing it `main`, I ran a text search for "master" in the code and this seems to be the only file that would probably be concerned about. 

Admittedly, I have no idea what this file does. 😅 So please let me know how it should be tested!

### DIFF
--- a/.configure
+++ b/.configure
@@ -1,6 +1,6 @@
 {
   "project_name": "woocommerce-ios",
-  "branch": "master",
+  "branch": "main",
   "pinned_hash": "1546d3128a60831f58226f7baae0e239068a6c78",
   "files_to_copy": [
     {


jkmassel:
oooh – this would definitely break things 😂 

We'll need to do the same steps in the mobile secrets repo as here – create `main` and delete `master`, but we'd need PRs for every project that uses mobile secrets. 

I have a task on my list that requires updating all of the projects anyway, so if you want to leave this branch open I can add my changes to it and then we can merge? (it'd just be a change to `pinned_hash` as well)

shiki:
@jkmassel Sounds good! Let me know if there's anything else I can help with.

shiki:
Awesome. Thank you for taking the time to check, @jkmassel! 🙇

PR #2474, opened 2020-06-25
Remove `ProductCategory` entry from model 26 to 27 mapping model XML with a test case

jaclync:
Migration from model version 26 to 27 and attempted improvement for #2371 

## Changes

Our Sentry crash logs indicated that most of the crashes were migrating from model version 26 to 27 (*). While inspecting the [Core Data v26 to v27 mapping model](https://github.com/woocommerce/woocommerce-ios/blob/9e605417971a9dda9333b8f63c8b00d83b2de598/Storage/Storage/Model/WooCommerceModelV26toV27.xcmappingmodel/xcmapping.xml), I noticed that `ProductCategory` is still in the XML even though it was removed in Xcode. The `ProductCategory` entry had the same `mappingnumber` value at 24 (the position in the mapping entity list) as `Account`, and I just had a slight suspicion that the duplicate entry in the XML might cause some issues in the migration process for certain devices.

- This PR removed `ProductCategory` entry from model version 26 to 27 mapping model XML
- A test case was added to migrate from model version 26 to 27 with some data (`Product`, `ProductCategory`, `Account`) populated

*: we can check the model version that the user's device was stuck on by comparing the `NSStoreModelVersionHashes` value. Lots of events shared the same `NSStoreModelVersionHashes` as that of model version 26.

## Testing

Please test migration from a model version earlier than v27 (e.g. from [release 3.9](https://github.com/woocommerce/woocommerce-ios/releases/tag/3.9)).

- Launch the app, logged in, at an earlier Core Data model version than v27
- Launch the app with the PR branch --> the app should run as expected, and feel free to try editing product categories

Update release notes:


jaclync:
I responded to the CR comments above, merging this PR now to make adding unit tests easier for https://github.com/woocommerce/woocommerce-ios/pull/2475! 😄

Issue #2473, opened 2020-06-24
Use NSPersistentStoreCoordinator instead of FileManager in Core Data Backups


shiki:
From @jaclync 

> Let’s update how we copy and remove the SQLite file to use NSPersistentStoreCoordinator API instead of FileManager.

Ref p91TBi-2YF-p2

astralbodies:
Taking this on next week.

astralbodies:
I am going to close this issue for now. Upon further investigation, the two areas we do backups from (#2816 removes one of these) require the DB to be in a migrate-able state. The operations Apple suggested do object copy from one persistent store to the other rather than copying the actual SQLite file. In our case, we’re trying to backup outside the scope of Core Data. I left the backup operation as is in the iterative migrator and used that logic in the CoreDataManager for the recovery operation.

Issue #2370, opened 2020-05-29
Untangle Core Data Setup and CrashLogging


shiki:
Found by @jaclync in p91TBi-2PM-p2#comment-860. When CrashLogging is initialized, the Core Data setup is also unexpectedly started because of this line: 

https://github.com/woocommerce/woocommerce-ios/blob/a10080aa477ba4388c7af5eb63f70ec813ddafd9/WooCommerce/Classes/Tools/Logging/CrashLogging.swift#L17-L24

shiki:
This may help with #2371

shiki:
Related discussion by @koke in paNNhX-f9-p2.

itsmeichigo:
CrashLogging initialization seems to no longer be tangled with Core Data after the change in #10049. I've added the initialization of Core Data as part of the AppDelegate setup in #14013, so the order of initialization should be as follows:

CrashLogging > CoreDataManager > DefaultStoresManager.

I'll mark #14013 as closing this issue for now.

PR #2481, opened 2020-06-27
Remove `fatalError` in case of persistent store backup failure and add more logging to store migration

jaclync:
Part of #2371 

## Changes

Since we really see a flood of crashes on Sentry and mentions on App Store reviews, I think removing the `fatalError` when the app fails to back up the store (removing and copying the pre-existing store) would be a good first step. After checking the [crash events from release 4.5](https://sentry.io/organizations/a8c/issues/1711092978/events/?project=1458804&query=release%3A4.5%2A&statsPeriod=14d), it seems like the crashes weren't only migrating from model version 26 to 27 but also from 27 to 28. Now that we know lots of users couldn't get past the backup step, I think it'd be great to know whether the [retry of persistent store loading](https://github.com/woocommerce/woocommerce-ios/blob/8a982eb73f30057aa1fa55c2c0f6f6003f6b8d97/Storage/Storage/CoreData/CoreDataManager.swift#L83-L96) would work.

- This PR removed the `fatalError` when the app couldn't remove or copy the pre-existing persistent store when it failed to migrate for the first time. The errors that occurred in the backup step are then recorded in the log event where the app either recovers from the retry or not
- Added more logging to `CoreDataIterativeMigrator` on the migration step (including a similar backup step)

It'd still be great to understand why the migration fails in the first place. We'll continue monitoring the logs in the upcoming crash events to know more about the model versions and migration process, and see if the app could recover from the retry.

## Testing

I think just CI is sufficient, since we haven't got to reproduce the crash yet. Feel free to manually add a similar logging event, and run in WooCommerce Alpha target with "Release-Alpha" configuration to verify on Sentry. [Here](https://sentry.io/organizations/a8c/issues/1737817640/events/cc7d6d5941004b32845eb2f82120ce4f/?project=1458804&statsPeriod=14d)'s an example of a similar Sentry event that I added.

Update release notes:

### DIFF
--- a/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
+++ b/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
@@ -102,7 +102,16 @@ public struct CoreDataIterativeMigrator {
             debugMessages.append(migrationAttemptMessage)
             DDLogWarn(migrationAttemptMessage)
 
-            guard migrateStore(at: sourceStore, storeType: storeType, fromModel: modelFrom, toModel: modelTo, with: migrateWithModel) == true else {
+            let (success, migrateStoreError) = migrateStore(at: sourceStore,
+                                                            storeType: storeType,
+                                                            fromModel: modelFrom,
+                                                            toModel: modelTo,
+                                                            with: migrateWithModel)
+            guard success else {
+                if let migrateStoreError = migrateStoreError {
+                    let errorInfo = (migrateStoreError as NSError?)?.userInfo ?? [:]
+                    debugMessages.append("Migration error: \(migrateStoreError) [\(errorInfo)]")
+                }
                 return (false, debugMessages)
             }
         }
@@ -197,7 +206,7 @@ private extension CoreDataIterativeMigrator {
                              storeType: String,
                              fromModel: NSManagedObjectModel,
                              toModel: NSManagedObjectModel,
-                             with mappingModel: NSMappingModel) -> Bool {
+                             with mappingModel: NSMappingModel) -> (success: Bool, error: Error?) {
         let tempDestinationURL = createTemporaryFolder(at: url)
 
         // Migrate from the source model to the target model using the mapping,
@@ -212,7 +221,7 @@ private extension CoreDataIterativeMigrator {
                                       destinationType: storeType,
                                       destinationOptions: nil)
         } catch {
-            return false
+            return (false, error)
         }
 
         do {
@@ -220,10 +229,10 @@ private extension CoreDataIterativeMigrator {
             try copyMigratedOverOriginal(from: tempDestinationURL, to: url)
             try deleteBackupCopies(at: backupURL)
         } catch {
-            return false
+            return (false, error)
         }
 
-        return true
+        return (true, nil)
     }
 
     static func metadataForPersistentStore(storeType: String, at url: URL) throws -> [String: Any]? {
--- a/Storage/Storage/CoreData/CoreDataManager.swift
+++ b/Storage/Storage/CoreData/CoreDataManager.swift
@@ -48,32 +48,22 @@ public class CoreDataManager: StorageManagerType {
 
             /// Backup the old Store
             ///
+            var persistentStoreBackupError: Error?
             do {
                 let sourceURL = self.storeURL
                 let backupURL = sourceURL.appendingPathExtension("~")
                 try FileManager.default.copyItem(at: sourceURL, to: backupURL)
             } catch {
-                let message = "☠️ [CoreDataManager] Cannot backup Store: \(error)"
-                self.crashLogger.logMessageAndWait(message,
-                                                   properties: ["persistentStoreLoadingError": persistentStoreLoadingError,
-                                                                "backupError": error,
-                                                                "appState": UIApplication.shared.applicationState.rawValue,
-                                                                "migrationMessages": migrationDebugMessages],
-                                                   level: .error)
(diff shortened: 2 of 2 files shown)


loremattei:
Hey! I'm moving this to 4.7 because 4.6 has been cut. If you need this to make it to 4.6, please feel free to ping me.

jaclync:
@woocommerce/ios-developers I was thinking about trying this for release 4.6, lemme know if you have other thoughts! @shiki I saw that you were also thinking about working on recovery from backup failure, does this align with your plan?

I looked at more crash events on Sentry from release 4.5, and it seems like the crashes were mostly from model version 27 to 28 now (v28 was released in the same release 4.5). So it looks like one of these migration steps failed when migrating from model version 27 to 28:

https://github.com/woocommerce/woocommerce-ios/blob/39e5c7854d48899b34a227cb418c10fc56b104c7/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L201-L224

I added more logging for the error that occurred in this function, and I have a strong suspicion that it might be related to the file system access.

This PR also removed `fatalError` from the database backup steps, the error(s) are logged later when trying to load the persistent stores again. Feel free to share any concerns!

shiki:
> I saw that you were also thinking about working on recovery from backup failure, does this align with your plan?

Yup. You pretty much did it. 😅 I'll still take a look at what else we can do though. Perhaps maybe adding unit tests. Or I could work on #2370. 

I'll review this today. Thank you, Jaclyn. 🙇

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -102,7 +102,16 @@ public struct CoreDataIterativeMigrator {
             debugMessages.append(migrationAttemptMessage)
             DDLogWarn(migrationAttemptMessage)
 
-            guard migrateStore(at: sourceStore, storeType: storeType, fromModel: modelFrom, toModel: modelTo, with: migrateWithModel) == true else {
+            let (success, migrateStoreError) = migrateStore(at: sourceStore,
```
shiki, line 105: Not related to this line. 

I noticed that the errors we were getting are missing reasons for the failed migrations. For example:

image

It looks like we don't return any debugging messages for cases where we return `false`. Example: 

https://github.com/woocommerce/woocommerce-ios/blob/4350b456efd71fd1f3956a58fda997c0b3573d8a/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L31-L33

Do you think we should add messages for those?

Storage/Storage/CoreData/CoreDataManager.swift
```diff
@@ -48,34 +48,22 @@ public class CoreDataManager: StorageManagerType {
...
-                                                   level: .fatal)
-                fatalError(message)
+                persistentStoreBackupError = error
             }
 
             /// Remove the old Store
             ///
+            var persistentStoreRemovalError: Error?
             do {
                 try FileManager.default.removeItem(at: self.storeURL)
```
shiki, line 64: This one might actually be wrong because it does not consider the existence of the `.sqlite-wal` and `.sqlite-shm` files. 

image

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: Looking more closely, it looks like this will be handled by the `do catch`.

jaclync: there are many points of failure in this function `CoreDataIterativeMigrator. iterativeMigrate`, and I think the most critical one is probably where it migrates the store in this call to `migrateStore`. later in this file I updated to return the error for each point of failure (actual migration error, or backup error) and adding this error to `debugMessages` below (L113)

Storage/Storage/CoreData/CoreDataManager.swift
```diff
@@ -48,34 +48,22 @@ public class CoreDataManager: StorageManagerType {
 
             /// Backup the old Store
             ///
+            var persistentStoreBackupError: Error?
```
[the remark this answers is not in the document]
jaclync: ugh yea I don't feel super good about having `persistentContainer: NSPersistentContainer` as a public lazy var, since it could be first triggered by anywhere (like the Reviews data source before). maybe this could be part of the effort to untangle Core Data setup from the app setup? https://github.com/woocommerce/woocommerce-ios/issues/2370

Storage/Storage/CoreData/CoreDataManager.swift
jaclync: yea they mentioned this at the WWDC lab. I believe https://github.com/woocommerce/woocommerce-ios/issues/2473 is going to replace this with [`NSPersistentStoreCoordinator`'s `destroyPersistentStore`](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468888-destroypersistentstore)

jaclync:
@shiki thanks a lot for the review! I responded to your comments, lemme know if I missed anything. the release branch usually gets merged back to `develop` within 1-2 days for your testability PR (can ping Platform 9 too) 😄

shiki:
Merge it! :shipit:

jaclync:
hey @loremattei I just merged this PR to release 4.6 branch! if possible, it'd be great if the changes here in release 4.6 can be merged back to `develop` for another PR that depends on these. thanks a bunch for wrangling the releases!

Issue #2199, opened 2020-04-27
Products: edit an existing grouped product


pmusolino:
As part of Products M3, we should provide a way to edit an existing grouped product.

Feel free to split this issue into more subtasks.

- [x] display a grouped product with readonly properties for the new row (linked products) https://github.com/woocommerce/woocommerce-ios/pull/2449
- [x] editing products for the grouped product
  - [x] Networking & Yosemite layer changes https://github.com/woocommerce/woocommerce-ios/pull/2459
  - [x] Product cell: selected state, accessory view https://github.com/woocommerce/woocommerce-ios/pull/2483
  - [x] existing grouped products screen with delete functionality https://github.com/woocommerce/woocommerce-ios/pull/2505
  - [x] adding grouped products screen https://github.com/woocommerce/woocommerce-ios/pull/2517
  - [x] adding grouped products screen - search https://github.com/woocommerce/woocommerce-ios/pull/2541
- [ ] more unit tests: `ProductsRemote` and `ProductStore` on `loadProducts`'s new params (`pageNumber` and `pageSize`) (Ref https://github.com/woocommerce/woocommerce-ios/pull/2505#discussion_r455203673)

<img width="421" alt="Screen Shot 2020-06-18 at 3 25 35 PM" src="https://user-images.githubusercontent.com/1945542/84990631-02d61d80-b178-11ea-8757-d923fcd99d5b.png">

This is the entire flow:
<img src="https://user-images.githubusercontent.com/495617/80395967-40bf6f80-88b4-11ea-9e5b-191185224617.jpg" width=800 />

jkmassel:
We're freezing version `4.7` today, so I'm bumping this to `4.8`. If you need this to be part of `4.7`, please merge this PR into `release/4.7` and DM me – I'll be happy to cut a new beta release!

PR #2483, opened 2020-06-29
Add selected state UI and enable accessory view for product cell

jaclync:
Prerequisite for adding grouped products in #2199 

## Changes

- Added support for accessory view to the product cell by updating the bottom border's trailing constraint to pin to the cell instead of the cell content view
  - Otherwise, the bottom border ends before the leading edge of the accessory view like the screenshot below:
Screen Shot 2020-06-29 at 9 02 52 PM

- Added selected state to the product cell:
  - In its view model `ProductsTabProductViewModel `, added `isSelected: Bool` property that is default to `false`
  - Added an optional property `selectedProductImageOverlayView: UIView?`
  - Created an icon `UIImage.checkmarkInCellImageOverlay` with the size and color for display in `ProductsTabProductTableViewCell`
- Enabled SwiftUI preview for :
  - Created an initializer with default properties for `Product` in debug builds only
  - Added SwiftUI preview with non-selected and selected states to `ProductsTabProductViewModel`

## Testing

Currently `ProductsTabProductTableViewCell` is used in 3 places, and there should no visual changes to them:
1. Products tab
2. Product search results
3. Product variation list

- Go to the Products tab --> the loading state and list view should be the same as in `develop`
- Tap on the search icon --> the search results should look the same as in `develop`
- Tap on a variable product, then tap on the variations row --> the product variation list should be the same as in `develop`

## Example screenshots

SwiftUI preview for the non-selected and selected states in Light/Dark/Large font modes (thanks to @shiki's example earlier!). The selected state will be used in a future PR for #2199. Dynamic type isn't fully supported for the product cell yet.

Screen Shot 2020-06-29 at 9 06 49 PM

Note: **no changes** are expected in the before and after screenshots.

\ | before | after
-- | -- | --
Products tab - light | Simulator Screen Shot - iPhone 11 - 2020-06-29 at 17 02 10 | Simulator Screen Shot - iPhone 11 - 2020-06-29 at 17 11 20
Search - dark | Simulator Screen Shot - iPhone 11 - 2020-06-29 at 17 02 40 | Simulator Screen Shot - iPhone 11 - 2020-06-29 at 17 11 49
Product variations - light | Simulator Screen Shot - iPhone 11 - 2020-06-29 at 17 03 11 | Simulator Screen Shot - iPhone 11 - 2020-06-29 at 17 11 34

Update release notes:


jaclync:
@woocommerce/ios-developers 👋 would anyone be able to review this one? since I have a bunch of work based on this cell change, it'd be great if it can be reviewed when any of you gets a chance 😄 the CI indicator shows pending because the PR was made when GitHub services were unstable (that led to an outage), but the [tests already passed on CI](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios?branch=issue%2F2199-update-product-cell).

pmusolino:
@jaclync on it 🙇‍♂️

WooCommerce/Classes/Extensions/Product+SwiftUIPreviewHelpers.swift
```diff
@@ -0,0 +1,72 @@
+#if canImport(SwiftUI) && DEBUG
```
pmusolino, line 1: I still not used the SwiftUI Preview, so maybe I'm wrong, but is it really needed this optional import here?

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
```diff
@@ -41,6 +42,12 @@ final class ProductsTabProductTableViewCell: UITableViewCell {
         // Border color is not automatically updated on trait collection changes and thus manually updated here.
         productImageView.layer.borderColor = Colors.imageBorderColor.cgColor
     }
+
+    override func prepareForReuse() {
+        super.prepareForReuse()
+
+        selectedProductImageOverlayView?.removeFromSuperview()
```
pmusolino, line 49: What do you think about moving this line of code under the `else` option of line 87?
Since the `prepareForReuse()` is always called, this will remove it from the superview, also if it's not necessary.

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
```diff
@@ -159,3 +189,56 @@ private extension ProductsTabProductTableViewCell {
...
     }
 }
+
+#if canImport(SwiftUI) && DEBUG
+
+import SwiftUI
+
+import Yosemite
+
+private struct ProductsTabProductTableViewCellRepresentable: UIViewRepresentable {
```
pmusolino, line 199: This SwiftUI preview is really really cool! 💘

WooCommerce/Classes/Extensions/Product+SwiftUIPreviewHelpers.swift
jaclync: this conditional check isn't necessary for this `Product` extension, but I thought it's better making it only available for SwiftUI preview in debug mode since this helper is not used in the app 🙂

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
jaclync: Good suggestion! it is easier to read when all the overlay view logic is in the same place. I updated how the overlay view is removed/configured in: https://github.com/woocommerce/woocommerce-ios/pull/2483/commits/a3420a3f52a827a85a4d0bb360e8f9c918b4bae5

jaclync:
Thanks for reviewing & testing, Paolo! I responded to your comments and updated the PR 😄 merging now!

PR #2485, opened 2020-06-29
Improve CoreDataIterativeMigrator Testability

shiki:
These are just some small improvements that will possibly help with #2371 and #2473. 

~~🛑 This should be targeting `develop` but I wanted to use @jaclync's changes in #2481 to avoid merge conflicts.~~
⚠️ This is now targeting `develop` but waiting for #2481 to be merged in `develop` before merging this.

### No Static Methods

I changed all the methods in `CoreDataIterativeMigrator` so it no longer uses static methods. This allowed mocking of the `FileManager` and possibly more dependencies in the future:

https://github.com/woocommerce/woocommerce-ios/blob/b8b416c0c1672c33ca2807857a6fa0568d872bac/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L7-L13

It's now also a `class`. 

### Mocked `FileManager`

I created a [`FileManagerProtocol`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2371-iterative-migrator-improvements/Storage/Storage/Protocols/FileManagerProtocol.swift) `protocol` which defines the methods of `Foundation.FileManager`. The protocol is then used in `CoreDataIterativeMigrator` instead of directly referencing `FileManager.default`. 

There is a companion [`MockFileManager`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2371-iterative-migrator-improvements/Storage/StorageTests/Mocks/MockFileManager.swift) which allows mocking but **not extending** `FileManager`. This mock uses the same pattern as [`MockProductsRemote`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2371-iterative-migrator-improvements/Yosemite/YosemiteTests/Mockups/Networking/Remote/MockProductsRemote.swift) and others.  

### Additional Test

I added a single test for now to make use of all the changes. 

https://github.com/woocommerce/woocommerce-ios/blob/b8b416c0c1672c33ca2807857a6fa0568d872bac/Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift#L21-L42

We probably should add more before tackling #2473.  

## Testing

Please run a migration to test that this did not cause any regressions. 

1. `git checkout 3.9`
2. Run and log in. 
3. `git checkout -t origin/issue/2371-iterative-migrator-improvements` (this PR's branch)
4. Run. This should perform a migration from Model 27 to 28. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
+++ b/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
@@ -4,7 +4,7 @@ import CoreData
 /// CoreDataIterativeMigrator: Migrates through a series of models to allow for users to skip app versions without risk.
 /// This was derived from ALIterativeMigrator originally used in the WordPress app.
 ///
-public class CoreDataIterativeMigrator {
+class CoreDataIterativeMigrator {
 
     /// Migrates a store to a particular model using the list of models to do it iteratively, if required.
     ///
--- a/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
+++ b/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
@@ -18,8 +18,10 @@ public struct CoreDataIterativeMigrator {
     ///
     /// - Throws: A whole bunch of crap is possible to be thrown between Core Data and FileManager.
     ///
-    static func iterativeMigrate(sourceStore: URL, storeType: String, to targetModel: NSManagedObjectModel, using modelNames: [String])
-        throws -> (success: Bool, debugMessages: [String]) {
+    static func iterativeMigrate(sourceStore: URL,
+                                 storeType: String,
+                                 to targetModel: NSManagedObjectModel,
+                                 using modelNames: [String]) throws -> (success: Bool, debugMessages: [String]) {
         // If the persistent store does not exist at the given URL,
         // assume that it hasn't yet been created and return success immediately.
         guard FileManager.default.fileExists(atPath: sourceStore.path) == true else {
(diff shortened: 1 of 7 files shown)


Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
```diff
@@ -205,7 +232,17 @@ private extension CoreDataIterativeMigratorTests {
...
 /// Helpers for the Core Data migration tests
-extension CoreDataIterativeMigratorTests {
+private extension CoreDataIterativeMigratorTests {
+
+    var documentsDirectory: URL {
+        let path = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).last!
+        return URL(fileURLWithPath: path)
+    }
+
+    private func managedObjectModel(for modelName: String) throws -> NSManagedObjectModel {
```
jaclync, line 242: super nit: `private` isn't needed anymore

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -4,7 +4,13 @@ import CoreData
 /// CoreDataIterativeMigrator: Migrates through a series of models to allow for users to skip app versions without risk.
 /// This was derived from ALIterativeMigrator originally used in the WordPress app.
 ///
-public struct CoreDataIterativeMigrator {
+final class CoreDataIterativeMigrator {
```
jaclync, line 7: wow nice catch that it doesn't have to be public!

jaclync (approved):
Thanks for making the Core Data setup more testable! 😍 

Code: ✅ 
Testing: ✅

Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
shiki: 🙈 Fixed in ab52c6841b5a35f14d8aac43eb4a16bc29449cd6. Thanks!

PR #2487, opened 2020-06-30
Show pink dot icon on Deprecated Stats Banner

### LINKED ISSUE #2465: Stats → Show Red Icon on Top of the Sync Icon in the Banner

Ecarrion:
closes #2465 

# Why

In order to make the deprecated stats banner stand out more, it should include a pink button on top of the sync image.

# How

I tried to draw the pink dot by code but it turns out to be quite complex when dealing with dark mode and RTL orientations. 
We have decided that a composed image would be the best for this case.

# Screenshots

Light | Dark | RTL
---- | ---- | ----
light|dark|rtl

# Testing Steps

- Test the app with a store whose WooCommerce is lower than 4.0 or has the WCAdmin plugin disabled and see the stats banner with thew new sync image.

Update release notes:


WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -399,7 +399,7 @@ extension UIImage {
 
 
     static var syncImage: UIImage {
-        return UIImage.gridicon(.sync).imageFlippedForRightToLeftLayoutDirection()
+        return UIImage(imageLiteralResourceName: "icon-sync-dot")
```
jaclync, line 402: not super sure if the design looks the same, but we used to have a gift icon with a dot on the top right corner using code. feel free to use the assets if there are design requirements that are harder to achieve with code though!

https://github.com/woocommerce/woocommerce-ios/blob/e81fdc73468c2ed43a4ec7a3cfc4b6b3df888cf6/WooCommerce/Classes/Extensions/UIImage+Woo.swift#L172-L183

light | dark
-- | --
Simulator Screen Shot - iPhone 11 - 2020-07-01 at 21 36 43 | Simulator Screen Shot - iPhone 11 - 2020-07-01 at 21 36 56

Ecarrion: Wow I didn't know about that function 🙇 

Unfortunately(if I understand things correctly), that code does not change the icon colors from light to dark, when the user changes rendering mode while the app/screen is active. :-(

PR #2493, opened 2020-07-02
Add custom paging support to `SyncingCoordinator`

jaclync:
Prerequisite for #2199 

## Changes

More details in p91TBi-30X-p2

This custom paging behavior for syncing is required for #2199 since we exclude certain products from the paginated product list.

- Added `SyncingCoordinatorCustomPagingDelegate` as an optional delegate to `SyncingCoordinator` to manually determine the timing for syncing the next page. This only changed how the last visible index is calculated, and not triggering syncing when the delegate's `hasSyncedLastPage` is `true`.
- Added test cases for `SyncingCoordinator` when `SyncingCoordinatorCustomPagingDelegate` is set

## Testing

This PR is not expected to introduce any changes on all the `SyncingCoordinator` use cases, but it'd be nice to sanity check on the more prominent screens that use `SyncingCoordinator`:

- Go to the orders tab --> the infinite scroll should work correctly for the order list on both "processing" and "all" tabs, as well as the search results
- Go to the products tab --> the infinite scroll should work correctly for the product list, as well as the search results
- Go to the reviews tab --> the infinite scroll should work correctly for the review list

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Tools/SyncCoordinator.swift
+++ b/WooCommerce/Classes/Tools/SyncCoordinator.swift
@@ -16,6 +16,19 @@ protocol SyncingCoordinatorDelegate: class {
 }
 
 
+/// In case the results from a sync and the results for UI display might differ (e.g. when the results controller predicate excludes certain sync results in a
+/// table view), this delegate can provide the necessary information for `SyncingCoordinator` to determine whether to trigger sync for the next page.
+/// To keep the states simple, only the last visible index (`totalNumberOfVisibleElements` minus 1) could trigger sync for the next page.
+protocol SyncingCoordinatorCustomPagingDelegate: class {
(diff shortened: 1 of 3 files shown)


jaclync:
@woocommerce/ios-developers I first thought that this custom paging support isn't necessary anymore because we could exclude the product IDs in a param of a Products API request. However, when I later implemented the paginated grouped product list while supporting delete action on each linked product, this custom paging support is still needed. Just wanted to clarify since I said in the p2 that this PR isn't needed anymore 🙂

jaclync:
after more discussion in p91TBi-30X-p2, I'm going to try creating a separate lighter-weight class for this case. closing this PR for now, thanks for the review @Ecarrion still!

PR #2494, opened 2020-07-02
Merge 4.6 release notes into develop

loremattei:
Merge 4.6 release notes into develop.

Update release notes:


mokagio:
What do we do with these failing UI tests? Is it okay to run them again?

They are failing on the same spot:

```
LoginTests
    ✗ testEmailPasswordLoginLogout, XCTAssertTrue failed
      - Page WooCommerceUITests.PeriodStatsTable is not loaded.
```

It'd be interesting to see how they behave.

loremattei:
Yeah! Tests were broken at code freeze, but are now fixed in develop.
I'll cherry pick the fixes in the release branch, even though they update the test and not the app code.

Issue #2464, opened 2020-06-23
Stats → Implement Remind Me Later button on banner


shiki:
<img src="https://user-images.githubusercontent.com/198826/85420857-fa436580-b530-11ea-80cc-e150c24d49d0.png" width="320">

Ref p91TBi-2tl-p2#comment-1117. 

Garance: 

> Initially I envisioned “Dismiss” as a way to permanently dismiss this banner, but I don’t think that’s wise given how important it is for users to be able to see stats. So I think we should only show “Remind me later”, and remove dismiss. Selecting “Remind me later” will remove the banner, but users will see it again next time they open the app.

PR #2502, opened 2020-07-13
Update TopBannerView to be expandable and actionable

Ecarrion:
Refers to https://github.com/woocommerce/woocommerce-ios/issues/2464

# Why

The new deprecated stats banner is supposed to be expandable and actionable which is something our current `TopBannerView` does not support.

This PR aims to add such functionality to be able to dismiss the banner.
The implementation of that feature can be found [here](https://github.com/woocommerce/woocommerce-ios/pull/2503)

# How

1. Add a new enum to `TopBannerViewModel` to indicate which type of button we want with it's respective handler.
```swift
enum TopButtonType {
  case chevron(handler: (() -> Void)?)
  case dismiss(handler: (() -> Void)?)
}
```

This allows us to remove the previous `onDismiss` and `onExpandedStateChange` parameters on `TopBannerViewModel`

2. Reorganize the internal view layout to be able to hide the action button when the user wants to collapse the banner.

*PS:*  I've decided to not go into a deeper refactor on this view as it could go out of scope.

# GIF
correct-banners

# Testing steps
Run the app and see that the current banners(stats and products) behave correctly.
- To test the stats banner you need to manage a store running an older WC version or disable the wc-admin using [a plugin](https://wordpress.org/plugins/disable-dashboard-for-woocommerce/)

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
+++ b/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
@@ -44,23 +44,22 @@ final class TopBannerView: UIView {
         return button
     }()
 
+    private lazy var actionStackView = createActionStackView()
+
     private let isActionEnabled: Bool
 
     private(set) var isExpanded: Bool
 
-    private let onDismiss: (() -> Void)?
+    private let onTopButtonTapped: (() -> Void)?
     private let onAction: (() -> Void)?
-    private let onExpandedStateChange: (() -> Void)?
 
     init(viewModel: TopBannerViewModel) {
         isActionEnabled = viewModel.actionHandler != nil
         isExpanded = viewModel.isExpanded
-        onDismiss = viewModel.dismissHandler
         onAction = viewModel.actionHandler
-        onExpandedStateChange = viewModel.expandedStateChangeHandler
+        onTopButtonTapped = viewModel.topButton.handler
         super.init(frame: .zero)
-        configureSubviews()
-        configureSubviews(viewModel: viewModel)
+        configureSubviews(with: viewModel)
     }
 
     required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
```diff
@@ -122,19 +111,63 @@ private extension TopBannerView {
...
+    func createIconInformationStackView(with viewModel: TopBannerViewModel) -> UIStackView {
+        let informationStackView = createInformationStackView(with: viewModel)
+        let iconInformationStackView = UIStackView(arrangedSubviews: [iconImageView, informationStackView])
+
+        iconInformationStackView.translatesAutoresizingMaskIntoConstraints = false
+        iconInformationStackView.axis = .horizontal
+        iconInformationStackView.spacing = 16
+        iconInformationStackView.alignment = .leading
+        iconInformationStackView.layoutMargins = .init(top: 16, left: 16, bottom: 16, right: 16)
+        iconInformationStackView.isLayoutMarginsRelativeArrangement = true
```
jaclync, line 156: nit: just curious, was `isLayoutMarginsRelativeArrangement` set to fix some UI issue?

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerViewModel.swift
```diff
@@ -3,30 +3,41 @@ import UIKit
 /// View model for `TopBannerView`.
 ///
 struct TopBannerViewModel {
+
+    enum TopButtonType {
+        case chevron(handler: (() -> Void)?)
+        case dismiss(handler: (() -> Void)?)
```
jaclync, line 9: nit: it looks like we aren't using `dismiss` style (cross icon) anymore, maybe we can remove it for now. @Garance91540 do you know if we might be using a top banner that has "X" dismiss CTA at the top that isn't expandable anymore? (for example, in the old stats banners)

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
Ecarrion: This is to make the content laid-out inside `iconInformationStackView` to respect the provided layout margins.

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerViewModel.swift
Ecarrion: I'll merge this, and will remove the `dismiss` type in a next PR if/when design confirms :-)

PR #2503, opened 2020-07-13
Allow stats banner to be dismissed

Ecarrion:
Depends on https://github.com/woocommerce/woocommerce-ios/pull/2502
Fixes https://github.com/woocommerce/woocommerce-ios/issues/2464

# Why
Users should be able to dismiss the deprecated stats banner and be reminded about it in the next app launch.

# How
- Leverage on the changes made on #2502 to be able to have an actionable and expandable stats banner

- Track the "remind me later" state in a simple in-memory variable `preventStatsBannerToBeShown` in `DashboardStatsV3ViewController.swift`. Since that VC would be removed soon (After Sept 1st) I decided that this simpler approach is good enough. But let me know if you think there is a better/robust way to handle this.

# GIF
remind-me-later

# Testing Steps
- Launch the app with an old woo commerce store or by [disabling the wc-admin plugin](https://wordpress.org/plugins/disable-dashboard-for-woocommerce/)
- See that the banner can be dismissed and it won't show again until the app is re-launched.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Factories/DashboardUIFactory.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Factories/DashboardUIFactory.swift
@@ -14,7 +14,7 @@ protocol DashboardUI: UIViewController {
     /// Called when the default account was updated
     func defaultAccountDidUpdate()
 
-    /// Called when the user has decided to not be bother with the deprecated stats banner right now.
+    /// Called when the user has decided to not be bothered with the deprecated stats banner right now.
     func remindStatsUpgradeLater()
 
     /// Reloads data in Dashboard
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Stats v3/DashboardStatsV3ViewController.swift
```diff
@@ -7,6 +7,9 @@ class DashboardStatsV3ViewController: UIViewController {
 
     var onPullToRefresh: () -> Void = {}
 
+    /// Prevent stats banner to be shown, useful when the user has opted out of the banner for this session
+    var preventStatsBannerToBeShown = false
```
jaclync, line 11: nit: this could probably be `private`?

WooCommerce/Classes/ViewRelated/Dashboard/Factories/DashboardUIFactory.swift
```diff
@@ -14,6 +14,9 @@ protocol DashboardUI: UIViewController {
     /// Called when the default account was updated
     func defaultAccountDidUpdate()
 
+    /// Called when the user has decided to not be bother with the deprecated stats banner right now.
```
jaclync, line 17: super nit: be bothered :)

PR #2504, opened 2020-07-13
Add ResultsController.safeObject(at:)

### LINKED ISSUE #2241: Crash when selecting a row in Orders → NSInvalidArgumentException: cannot access fetched objects before -performFetch:
Sentry Issue: [WOOCOMMERCE-IOS-1S](https://sentry.io/organizations/a8c/issues/1511257574/?referrer=github_integration)

```
NSInvalidArgumentException: cannot access fetched objects before -performFetch:
  File "ResultsController.swift", line 129, in ResultsController.object
  File "Classes/ViewRelated/Orders/OrdersViewModel.swift", line 180, in OrdersViewModel.detailsViewModel
  File "OrdersViewController.swift", line 479, in OrdersViewController.tableView
  File "<compiler-generated>", in OrdersViewController.tableView
  File "main.swift", line 7, in main
...
(15 additional frame(s) were not displayed)
```

This started happening in 3.4.0.3.

As of May 5, 2020: 

- 24 events
- 24 users affected

shiki:
Fixes #2241. 

My attempts at fixing #2241 didn't work. Even though [we're calling `performFetch` on `viewDidLoad`](https://github.com/woocommerce/woocommerce-ios/blob/10eb6cf847c11da1179dab35fb04f7854610271a/WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift#L118), the crash log still indicate that somehow, the user is able to _select_ something before `performFetch` is called.

I could not reproduce the crash but we can pinpoint that the crash happens when [`NSFetchedResultsController.object(at:)` is accessed](https://github.com/woocommerce/woocommerce-ios/blob/10eb6cf847c11da1179dab35fb04f7854610271a/Yosemite/Yosemite/Tools/ResultsController.swift#L129).

https://github.com/woocommerce/woocommerce-ios/blob/10eb6cf847c11da1179dab35fb04f7854610271a/Yosemite/Yosemite/Tools/ResultsController.swift#L126-L130

## Solution

I added a safer version of `object(at:)` that can return `nil` in these conditions:

- `performFetch` was not called
- the `IndexPath` is invalid

I added test cases to prove its resilience. 

https://github.com/woocommerce/woocommerce-ios/blob/2c3adfeb26793a49f64c6a33e06b6817c75eddb4/Yosemite/YosemiteTests/Tools/ResultsControllerTests.swift#L314-L324

I then used this new method in `OrdersViewModel` which is used by `OrdersViewController`, the source of the crash. 

https://github.com/woocommerce/woocommerce-ios/blob/2c3adfeb26793a49f64c6a33e06b6817c75eddb4/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L273-L279

## Testing

1. In `OrdersViewController.viewDidLoad`, add this line after  `configureGhostableTableView()`:

    ```swift
        self.tableView(tableView, didSelectRowAt: IndexPath(row: 0, section: 0))
    ```
2. Run the app
3. Navigate to Orders
4. Confirm that the app did not crash.

The same steps will make the app crash with the same exception when using `develop`.

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
@@ -38,29 +38,29 @@ final class OrderTableViewCell: UITableViewCell & SearchResultCell {
 
     /// Renders the specified Order ViewModel
     ///
+    /// If the `viewModel` is not given, then the UI will be set to empty.
+    ///
     func configureCell(viewModel: OrderDetailsViewModel?, orderStatus: OrderStatus?) {
-        if let viewModel = viewModel {
-            titleLabel.text = title(for: viewModel.order)
-            totalLabel.text = viewModel.totalFriendlyString
-            dateCreatedLabel.text = viewModel.formattedDateCreated
-        } else {
-            titleLabel.text = nil
-            totalLabel.text = nil
-            dateCreatedLabel.text = nil
+        guard let viewModel = viewModel else {
+            resetLabels()
+            return
         }
+
+        titleLabel.text = title(for: viewModel.order)
+        totalLabel.text = viewModel.totalFriendlyString
+        dateCreatedLabel.text = viewModel.formattedDateCreated
 
         if let orderStatus = orderStatus {
             paymentStatusLabel.applyStyle(for: orderStatus.status)
             paymentStatusLabel.text = orderStatus.name
-        } else if let statusKey = viewModel?.order.statusKey {
+        } else {
+            let statusKey = viewModel.order.statusKey
+
             // There are unsupported extensions with even more statuses available.
             // So let's use the order.statusKey to display those as slugs.
             let statusEnum = OrderStatusEnum(rawValue: statusKey)
             paymentStatusLabel.applyStyle(for: statusEnum)
             paymentStatusLabel.text = statusKey
-        } else {
-            paymentStatusLabel.applyStyle(for: .failed)
-            paymentStatusLabel.text = nil
         }
     }
 
@@ -103,6 +103,17 @@ final class OrderTableViewCell: UITableViewCell & SearchResultCell {
 // MARK: - Private
 //
 private extension OrderTableViewCell {
+
+    /// Reset the UI to a "no data" state.
+    ///
+    func resetLabels() {
+        titleLabel.text = nil
+        totalLabel.text = nil
+        dateCreatedLabel.text = nil
+        paymentStatusLabel.applyStyle(for: .failed)
+        paymentStatusLabel.text = nil
+    }
+
     /// Preserves the current Payment BG Color
     ///
     func preserveLabelColors(action: () -> Void) {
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
```diff
@@ -38,21 +38,29 @@ final class OrderTableViewCell: UITableViewCell & SearchResultCell {
 
     /// Renders the specified Order ViewModel
     ///
-    func configureCell(viewModel: OrderDetailsViewModel, orderStatus: OrderStatus?) {
-        titleLabel.text = title(for: viewModel.order)
-        totalLabel.text = viewModel.totalFriendlyString
-        dateCreatedLabel.text = viewModel.formattedDateCreated
+    func configureCell(viewModel: OrderDetailsViewModel?, orderStatus: OrderStatus?) {
```
jaclync, line 41: nit: I was wondering if we could keep this function with non-optional `OrderDetailsViewModel` as before, and have another function when the view model is `nil` since it's an edge case?

WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift
```diff
@@ -434,8 +434,8 @@ private extension OrdersViewController {
 //
 private extension OrdersViewController {
 
-    func lookUpOrderStatus(for order: Order) -> OrderStatus? {
-        for orderStatus in currentSiteStatuses where orderStatus.slug == order.statusKey {
+    func lookUpOrderStatus(for order: Order?) -> OrderStatus? {
+        for orderStatus in currentSiteStatuses where orderStatus.slug == order?.statusKey {
```
jaclync, line 438: nit: might it make sense to early return `nil`? otherwise it might return some order status that happens to have an empty slug 🤔

WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
shiki: I'm not sure if that would be a good idea. I find that overloading methods can be confusing. Especially for `Optionals`. 

```swift
    func configureCell(viewModel: OrderDetailsViewModel?, orderStatus: OrderStatus?) {
 
    }

    func configureCell(viewModel: OrderDetailsViewModel, orderStatus: OrderStatus?) {

    }
```

I believe it sort of makes things harder to understand. 

---

I changed this again so that there's a `guard` and a `resetLabels()` function. This is kind of close to having a state machine. Let me know if that's better for reading or not? 489ab91b1ca44e74cf98bfb5de1b5c3272eafc4f

WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift
shiki: Good catch! Fixed in 6b2b5e8626f799bdb8db4a3ce20107fe1b7d8330.

WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
jaclync: the changes look great! 👍

Issue #2369, opened 2020-05-29
App Incorrectly Shows Order Deletion Notice


shiki:
## Describe the bug

The app shows an "Order has been deleted from your store" notice even though the order was not really deleted from the store. 

<img src="https://user-images.githubusercontent.com/198826/83298887-75279380-a1b2-11ea-90c5-1fc618761422.png" width="240">

## To Reproduce

This can be reproduced in 2 scenarios. 

### Scenario A

1. Navigate to Orders → Processing.
2. Pull to refresh. 
3. Quickly tap to the All Orders tab and select an order.
4. Stay on that order's details page and it should disappear once the pull-to-refresh loading has completed. The notice will then be shown.

### Scenario B 

1. Navigate to Orders. 
2. Pick any order. 
3. Navigate to My Store → Settings. 
4. Switch to a different site. 

## Expected behavior

It's probably better to just not show the notice since the message is inaccurate. 

We don't show this message in Android. 

## Isolating the problem (mark completed items with an [x]):

- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

shiki:
Discussed in https://github.com/woocommerce/woocommerce-ios/pull/2515#issuecomment-660141275, we should also consider removing the _pop_ for the current `ViewController`. 

https://github.com/woocommerce/woocommerce-ios/blob/66d9e2403fa0126ba3c1193677e072e99a7e8d86/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/OrderDetailsViewController.swift#L124

PR #2515, opened 2020-07-16
Fix Crash When Dismissing Message Composer

### LINKED ISSUE #1361: Crash: EXC_BAD_ACCESS Attempted to dereference garbage pointer
[Public crash report](https://sentry.io/share/issue/8ca314ff9e5b4f0aaa8acef4d3af987c/). Please log into Sentry to see full details.

```
EXC_BAD_ACCESS
_forwardStackInvocation: > forwardInvocation: > respondsToSelector: > selector > te:inRange:options:usingBlock: > webView:didCommitLoadForFrame: >
Attempted to dereference garbage pointer 0x66ab90770.
```

Research possible solutions from this [Stack Overflow post](https://stackoverflow.com/questions/51940031/ios-exc-bad-access-attempted-to-dereference-garbage-pointer-twiliopoco).

- [ ] Check for zombies

shiki:
Fixes #1361 

## Findings

I got lucky with this one. I noticed that the crash had something like this:

```
messageComposeViewController:didFinishWithResult: > retainWeakReference > smsComposeControllerCancelled >
Attempted to dereference garbage pointer 0x3fe5607c.
```

I searched for `messageComposeViewController:didFinishWithResult` and found that in [`OrderMessageComposer`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewModels/Order%20Details/OrderMessageComposer.swift). I didn't understand why this would crash but luckily I remembered how things like this could happen in Objective-C. After further investigation, I found out that [`messageComposeDelegate`](https://developer.apple.com/documentation/messageui/mfmessagecomposeviewcontroller/1614070-messagecomposedelegate) is an `unowned` reference. This gave me an idea that _maybe_ the `OrderMessageComposer` was deallocated **before** the `MFMessageComposerViewController` was dismissed. 

And it looks like that might be the case here. The crash can be reproduced by these steps:

1. In the web, submit an order with a phone number (preferably yours).
1. In `OrdersViewController.sync(...)`, replace the `ServiceLocator.stores.dispatch(action)` line with this:

```swift
        DispatchQueue.main.asyncAfter(deadline: .now() + .seconds(10)) {
        }
```
2. Run the app on a real device.
3. Navigate to Orders.
4. Pull to refresh. 
5. **Very quickly** navigate to an Order → View Billing Info → Tap phone number → Message.
6. Wait until 10 seconds have passed and the order list should have been refreshed. 
7. Cancel the SMS dialog or send a message.

The app should crash after that. This is because the owner of `OrderMessageComposer`, [`BillingInformationViewController`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Customer%20Section/Billing%20Information/BillingInformationViewController.swift) was dismissed. This deallocates `OrderMessageComposer` which is the delegate of `MFMessageComposerViewController` whose UI is **still present**. When the user submits or cancels the dialog, `MFMessageComposerViewController` will try to inform the `messageComposeDelegate` but since it no longer exists → 💥 

The `BillingInformationViewController` was dismissed during the pull-to-refresh because we delete all the orders during pull-to-refresh and when an order is deleted, the `OrderDetailsViewController` dismisses the topmost VC:

 https://github.com/woocommerce/woocommerce-ios/blob/789ff4347ebb2f153982cedc578587735b71b29a/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/OrderDetailsViewController.swift#L119-L126

I'm not really sure if that code is a good idea. I don't think it's a good experience to suddenly dismiss a view while the user is doing something with it. But I'm opting not to touch that for now.

## Solution

I initially added a `deinit` in the `OrderMessageComposer` like this:

```swift
    deinit {
        // If this class is dismissed while a `ComposeViewController` is active, a crash will
        // happen because `ComposeViewController.delegate` is an `unowned` reference to `self`.
        messageComposeViewController?.dismiss(animated: false, completion: nil)
    }
```

After testing it, I found that it's not a good experience to dismiss the SMS dialog **while** the user is not finished with it. So, I went with the singleton (ew) approach by moving `OrderMessageComposer` to a reusable service as part of `ServiceLocator`: [`MessageComposerPresenter`](https://github.com/woocommerce/woocommerce-ios/blob/issue/1361-fix-crash-in-sms-compose/WooCommerce/Classes/ServiceLocator/MessageComposerPresenter.swift). This class is kept forever so it should not be deallocated. And it can dismiss `MFMessageComposerViewController` after the user is done. 

## Testing

1. Follow the steps to reproduce using `develop`. 
2. Confirm the crash happens.
3. Follow the same steps using this PR's branch. 
4. Confirm the crash no longer happens.

## Reviewing 

- Only 1 reviewer is needed but anyone can review. 
- Please review that I did not accidentally remove any critical code when I replaced `OrderMessageComposer`.

## Submitter Checklist

- [x] I have considered adding accessibility improvements for my changes.


Ecarrion:
Little note that `Coordinators` could prevent this kind of scenarios as the delegate of the `MFMessageComposerViewController` would be the coordinator and not the `OrderMessageComposer` 😇

shiki:
> have you considered just removing this line so that it doesn't pop the screen from order details when an order is deleted? I tried it and it seems fine, but I'm not sure if there might be other implications.

@jaclync I have considered it and have thought about it at some point. I think it's better to remove that popping and the notice together (https://github.com/woocommerce/woocommerce-ios/issues/2369). And yeah, I think we should do a little bit of research to make sure we're not breaking anything once we do that. 😬

shiki:
> Little note that Coordinators could prevent this kind of scenarios as the delegate of the MFMessageComposerViewController would be the coordinator and not the OrderMessageComposer 😇

@Ecarrion That'd be nice! I have a PR for a new `Coordinator` coming soon. I'd love to have your input on it. 🙂

shiki:
Thank you for reviewing, @jaclync!

PR #2517, opened 2020-07-16
Edit Grouped Products: edit linked products (Part 2/3) - show a product list selector to add linked products

jaclync:
Product selector for grouped product's linked products #2199 

This PR implements the highlighted screen in the flow to edit linked products for a grouped product:

this_pr

## Changes

- Networking layer:
  - In `ProductsRemote.loadAllProducts`, added a parameter `excludedProductIDs: [Int64]` to exclude a given list of product IDs in the API call. Also updated the completion data type from `([Product]?, Error?)` to `Result<[Product], Error>`.
  - Updated unit tests:
    - Added the same `MockupNetwork` extension helper `pathComponents` that is exactly the same from `YosemiteTests`.
    - Added a test case to verify the new param in the API call from excluded product IDs is as expected
- Yosemite layer:
  - In `ProductAction.synchronizeProducts`, added two parameters `excludedProductIDs: [Int64] = []` and `shouldDeleteStoredProductsOnFirstPage: Bool = true`. In `ProductStore`'s handling of this action, `excludedProductIDs` is passed to `ProductsRemote` and `shouldDeleteStoredProductsOnFirstPage` is used to check whether to delete all the stored data when fetching the first page.
    - `shouldDeleteStoredProductsOnFirstPage` is needed because we don't want to delete all the stored products when fetching a list of products on the list selector screen
- App layer:
  - In `GroupedProductListSelectorDataSource`, added a new function `addProducts` that is called when the user adds products from the product list selector screen. Just to prevent from any duplicates due to `ResultsController` or any other mistakes, I also added a `Sequence` helper `removingDuplicates()` to remove all the duplicates after the first instance.
  - Prior to this product list selector screen, the cell always shows a checkmark in its accessory view if selected. However, we have a different selected UI in the product list selector screen so I removed the cell's accessory type configuration from `PaginatedListSelectorViewController` and implemented it in `PaginatedProductShippingClassListSelectorDataSource` and `ProductTaxClassListSelectorDataSource`
  - Created `ProductListSelectorViewController` that contains a child view controller `PaginatedListSelectorViewController` with data source `ProductListMultiSelectorDataSource` that contains an observable list of the latest product IDs
  - In `GroupedProductsViewController`, showed `ProductListSelectorViewController` and updated its data source on product list selector completion
  - Added/updated unit tests for the changes above

## Testing

- Go to the Products tab
- Tap on a grouped product
- Tap on the "Add Products" button --> should see a list of products that **do not contain the pre-selected products plus the grouped product itself** (even though it's allowed to select itself in core), and the navigation bar has a search icon on the right
- Select a few products --> the navigation bar should indicate the number of selected products, and the search icon becomes "Done" button
- Tap "Done" --> the "Grouped products" row should show the updated total number of linked products
- Tap "Update" --> the linked products for the grouped product should be updated remotely

### Sanity check on affected screens

- Go to the Products tab
- Tap on a simple & physical product
- Tap on the price row
- Tap on the "Tax class" row --> should be able to select the tax class as before
- Go back to product details
- Tap on the shipping row
- Tap on the "Shipping class" row --> should be able to select the tax class as before

## Example screenshots

simulator

state | light | dark
-- | -- | -- 
no products selected | Simulator Screen Shot - iPhone 11 - 2020-07-16 at 20 44 00 | Simulator Screen Shot - iPhone 11 - 2020-07-16 at 20 48 24
1 product selected | Simulator Screen Shot - iPhone 11 - 2020-07-16 at 20 49 58 | Simulator Screen Shot - iPhone 11 - 2020-07-16 at 20 49 46
more than 1 products selected | Simulator Screen Shot - iPhone 11 - 2020-07-16 at 20 44 34 | Simulator Screen Shot - iPhone 11 - 2020-07-16 at 20 45 14

Update release notes:


Networking/Networking/Remote/ProductsRemote.swift
```diff
@@ -40,11 +41,16 @@ public final class ProductsRemote: Remote, ProductsRemoteProtocol {
                                 productType: ProductType? = nil,
                                 orderBy: OrderKey = .name,
                                 order: Order = .ascending,
-                                completion: @escaping ([Product]?, Error?) -> Void) {
+                                excludedProductIDs: [Int64] = [],
+                                completion: @escaping (Result<[Product], Error>) -> Void) {
```
pmusolino, line 45: A 🎖️ for having converted it in `Result<>`

WooCommerce/Classes/Extensions/Array+Helpers.swift
```diff
@@ -51,3 +51,16 @@ extension Sequence where Element: Numeric {
     /// Returns the sum of all elements in the collection.
     func sum() -> Element { return reduce(0, +) }
 }
+
+extension Sequence where Element: Equatable {
+    /// Returns the sequence with any duplicate elements after the first one removed.
+    func removingDuplicates() -> [Element] {
```
pmusolino, line 57: nit (feel free to ignore it because I'm not 100% sure that it's the correct thing to do): what do you think of `removeDuplicates()`?

jaclync: good question! for this case, [Swift API design guidelines](https://swift.org/documentation/api-design-guidelines/) have a section "Name Mutating/nonmutating method pairs consistently" that recommends naming nonmutating function as the verb's past tense "ed" or "ing" whereas mutating functions have the verb in current tense. Lemme know if you have any thoughts!

jaclync:
Thanks for the many reviews @pmusolino! responded to your feedback and merging this now :)

PR #2519, opened 2020-07-16
Replace StatsVersionState Usages with StatsVersion

### LINKED ISSUE #2466: Stats → Consider Deleting StatsVersionState Enum
After #1946 is done, the `StatsVersionState` `enum` will only have one state:

```swift
enum StatsVersionState: Equatable {
    /// if initial(v3) = then show banner “Upgrade to keep seeing your stats”
    /// if initial(v4) = then default to stats
    case initial(statsVersion: StatsVersion)
}
```

We could probably simplify this and only use the `StatsVersion` `enum` instead.

shiki:
Closes #2466.

In #2467, we removed all the cases of the `StatsVersionState` `enum` except one. This makes the `enum` less useful. 

https://github.com/woocommerce/woocommerce-ios/blob/66d9e2403fa0126ba3c1193677e072e99a7e8d86/WooCommerce/Classes/ViewRelated/Dashboard/Factories/StatsVersionStateCoordinator.swift#L4-L11

This PR replaces `StatsVersionState` usages with just the [`StatsVersion`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Storage/Storage/Model/StatsVersion.swift). I also renamed the `StatsVersionStateCoordinator` to `StatsVersionCoordinator`. The rest of the changes are property and variable renames. 

## Testing

There should be no functional changes here. Confirming that the tests are passing should be good enough. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Factories/DashboardUIFactory.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Factories/DashboardUIFactory.swift
@@ -25,23 +25,23 @@ protocol DashboardUI: UIViewController {
 
 final class DashboardUIFactory {
     private let siteID: Int64
-    private let stateCoordinator: StatsVersionCoordinator
+    private let statsVersionCoordinator: StatsVersionCoordinator
 
     private var lastStatsV3DashboardUI: (DashboardUI & TopBannerPresenter)?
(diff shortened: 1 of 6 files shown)


shiki:
Thank you, Ernesto!

PR #2523, opened 2020-07-16
Issue/2254 rename foreground notification

shiki:
Refs #2254. 

I'm planning to add an additional observable to [`PushNotesManager`](https://github.com/woocommerce/woocommerce-ios/blob/66d9e2403fa0126ba3c1193677e072e99a7e8d86/WooCommerce/Classes/ServiceLocator/PushNotesManager.swift#L5-L10). The observable will emit `ForegroundNotification` objects when the app is inactive. The observable code will look like this:

```swift
    /// An observable that emits values when the app is activated due to a Remote Notification.
    /// The Remote Notification is emitted.
    ///
    var inactiveNotifications: Observable<ForegroundNotification> { get }
```

I can use the same [`ForegroundNotification`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ServiceLocator/ForegroundNotification.swift) object for the result. But the name is not really appropriate anymore. 

This PR renames `ForegroundNotification` to `PushNotification`. 

## Testing

There are no functional changes. Checking that the build passes should be good enough.

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/Notifications/PushNotificationsManager.swift
+++ b/WooCommerce/Classes/Notifications/PushNotificationsManager.swift
@@ -14,12 +14,12 @@ final class PushNotificationsManager: PushNotesManager {
     let configuration: PushNotificationsConfiguration
 
     /// Mutable reference to `foregroundNotifications`.
-    private let foregroundNotificationsSubject = PublishSubject<ForegroundNotification>()
+    private let foregroundNotificationsSubject = PublishSubject<PushNotification>()
 
     /// An observable that emits values when the Remote Notifications are received while the app is
     /// in the foreground.
     ///
-    var foregroundNotifications: Observable<ForegroundNotification> {
+    var foregroundNotifications: Observable<PushNotification> {
         foregroundNotificationsSubject
     }
 
@@ -326,7 +326,7 @@ private extension PushNotificationsManager {
             return false
         }
 
-        if let foregroundNotification = ForegroundNotification.from(userInfo: userInfo) {
+        if let foregroundNotification = PushNotification.from(userInfo: userInfo) {
             configuration.application.presentInAppNotification(message: foregroundNotification.message)
 
             foregroundNotificationsSubject.send(foregroundNotification)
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/Notifications/PushNotificationsManager.swift
```diff
@@ -480,16 +480,16 @@ private extension PushNotificationsManager {
 
 // MARK: - ForegroundNotification Extension
```
jaclync, line 481: nit: this maybe also be updated to `PushNotification`?

WooCommerce/Classes/ServiceLocator/PushNotesManager.swift
```diff
@@ -7,7 +7,7 @@ protocol PushNotesManager {
     /// An observable that emits values when the Remote Notifications are received while the app is
     /// in the foreground.
     ///
-    var foregroundNotifications: Observable<ForegroundNotification> { get }
+    var foregroundNotifications: Observable<PushNotification> { get }
```
jaclync, line 10: I'm sure everything will make sense in a later PR, just wondering why this `foregroundNotifications` emits a `PushNotification` while `inactiveNotifications` emits a `ForegroundNotification` from the PR description? or did you mean that `inactiveNotifications` will also emit a `PushNotification`?

WooCommerce/WooCommerceTests/Mockups/MockPushNotificationsManager.swift
```diff
@@ -55,7 +55,7 @@ extension MockPushNotificationsManager {
     /// Send a ForegroundNotification that will be emitted by the `foregroundNotifications`
```
jaclync, line 55: nit: `ForegroundNotification ` -> `PushNotification`? 😄

WooCommerce/Classes/ServiceLocator/PushNotesManager.swift
shiki: > while inactiveNotifications emits a ForegroundNotification from the PR description? 

My bad. Copy-paste mistake. 🤪 Yep, the PR description should have said `PushNotification`. I fixed it.

WooCommerce/WooCommerceTests/Mockups/MockPushNotificationsManager.swift
shiki: Thank you! Fixed in a499f5856fce27fa87ce56fd2edb87af1b58a54d

WooCommerce/Classes/Notifications/PushNotificationsManager.swift
shiki: 🤦  Fixed in a499f5856fce27fa87ce56fd2edb87af1b58a54d

shiki:
Thank you for catching those issues, @jaclync! 🙏

PR #2525, opened 2020-07-16
Fix Remote Protocols Naming

shiki:
Following up on the protocol naming convention described in #2524, I'm renaming these silly `*EndpointsProviding` protocols that I created to just `*RemoteProtocol`. 

I thought I'd name these right away to prevent ourselves from copying that name in the future. "Endpoints Providing" really doesn't mean anything. I just invented that phrase. 😁 

## Testing

There are no functional changes. Confirming that the build passes should be good enough. 

## Reviewing 

## Submitter Checklist


shiki:
Thanks, @Ecarrion!

PR #2190, opened 2020-04-24
CircleCI: Update Xcode to 11.4

shiki:
In the [`issue/927-observables`](https://github.com/woocommerce/woocommerce-ios/tree/issue/927-observables) branch, I wanted to use [`XCTSkipIf`](https://developer.apple.com/documentation/xctest/3521325-xctskipif) to conditionally skip a unit test if the simulator is less than 13.0. Example:

https://github.com/woocommerce/woocommerce-ios/blob/c5d0cff51d44b957a1772bed5efc1804be0cfee1/WooCommerce/WooCommerceTests/Tools/Observables/PublishSubjectTests.swift#L234-L238

Using `XCTSkipIf()` allows better messaging when the tests are executed using an older simulator. The skipped tests are mentioned in the result:

```
Test Suite 'PublishSubjectTests' passed at 2020-04-24 09:20:55.451.
	 Executed 10 tests, with 5 tests skipped and 0 failures (0 unexpected) in 0.024 (0.068) seconds
Test Suite 'WooCommerceTests.xctest' passed at 2020-04-24 09:20:55.452.
	 Executed 10 tests, with 5 tests skipped and 0 failures (0 unexpected) in 0.024 (0.069) seconds
Test Suite 'Selected tests' passed at 2020-04-24 09:20:55.452.
	 Executed 10 tests, with 5 tests skipped and 0 failures (0 unexpected) in 0.024 (0.070) seconds
```

However, `XCTSkipIf` is only available for Xcode 11.4+. This PR updates the CircleCI config from 11.2.1 to 11.4.1 to solve this. 

## Questions

Is this okay? 😬Will this affect any of the release processes? 

## Testing

I believe checking if the build passed is good enough. 

## Reviewing 

Only 1 reviewer is needed but anyone can review. 

---

Heads up @pmusolino @jaclync. This shouldn't be a problem for us too, right?

Issue #2138, opened 2020-04-14
Color Palettes contain the wrong colors


mindgraffiti:
**Describe the bug**
gray shade0 is now drawing in as red, shade0, alpha 95%, all over the app. I do not know how many other colors are incorrectly coded.

**Context**
When dark mode was first introduced (5 months ago), Color-Studio.blog version 2.2.1 released the new color palette. Color-Studio's assets changed (3 months ago). A new version was not released when the color corrections were made, so nobody was notified of the changes.

When [Xcode 11.4 was released](https://developer.apple.com/documentation/xcode_release_notes/xcode_11_4_release_notes), Apple added `dynamic system gray colors to inspector color pickers`. One can only guess that other behind-the-scenes color changes occurred.  

Combine the incorrect color palette assets with Xcode 11.4, and now we're seeing the strange tinge.

**To Reproduce**
1. Use Xcode 11.4 to see the color palettes now getting rendered and used in the apps
2. Build and run from develop
3. Log out of the app
4. Log in. Notice that the listBackground gray is actually red shade 0.

**Screenshots**
| Xcode 11.3 | Xcode 11.4 |
| --- | --- |
| <img src="https://user-images.githubusercontent.com/1062444/79278354-28416500-7e71-11ea-87d1-2d45896fe7a1.png" /> | <img src="https://user-images.githubusercontent.com/1062444/79278357-29729200-7e71-11ea-9f9b-68c5269e86a7.png" /> |

**Expected behavior**
Re-import the [Color Palette assets](https://github.com/Automattic/color-studio/tree/2.2.1/dist/ios/ColorStudioPalette.xcassets). Problem fixed. The only downside to importing the color palette wholesale is that other color changes have occurred. I do not know which other colors were the wrong shades that were also updated.

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

**Mobile Environment**
Please include:
- Device: iPhone SE
- iOS version: ~13.4~ 12.4 (edit: Xcode 11.4.1 no longer uses the Color Palette directly for 13.4 sims) 
- WooCommerce iOS version: any version that has the app built and released using Xcode 11.4

**WordPress Environment**
n / a

jaclync:
[...]

PR #2528, opened 2020-07-17
CircleCI: Update Xcode to 11.5.0

### LINKED ISSUE #2471: Update CircleCI Xcode version to the latest version
We are still using a very old version of Xcode in CircleCI. 

https://github.com/woocommerce/woocommerce-ios/blob/9e605417971a9dda9333b8f63c8b00d83b2de598/.circleci/config.yml#L18-L22

Let's update it to the latest available [in CircleCI](https://circleci.com/docs/2.0/testing-ios/). 

---

~🛑 This is currently blocked by #2138.~

Previous attempt to work on this: https://github.com/woocommerce/woocommerce-ios/pull/2190

shiki:
Closes #2471. 

I attempted to upgrade the Xcode version in #2190 but we decided that we have to get #2138 resolved first. After some deliberation, we decided that we should just push through with the upgrade since:

- #2138 only happens on iOS 12.
- We're dropping iOS 12 very soon anyway.
- We weren't sure if it's really a problem because we're having trouble reproducing it. 

Ref: C70PAM3RA-p1594949432474000-slack

## Reason For Upgrading

Please see the description in https://github.com/woocommerce/woocommerce-ios/pull/2190 for why we would like to upgrade the Xcode version.

## Changes

This PR upgrades the Xcode version in CircleCI to 11.5.0. At this time, it's a GM version which I think should be fine? I also upgraded the iOS simulator to 13.5. 

## Testing 

I believe checking if the build passed is good enough. Unit tests and UI tests should be green. 

## Reviewing 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
   Build Tests:
     executor:
       name: ios/default
-      xcode-version: "11.2.1"
+      xcode-version: "11.5.0"
     steps:
       - git/shallow-checkout
       - fix-circleci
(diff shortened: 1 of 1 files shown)


shiki:
Thank you, @mokagio! 

I didn't think of testing the installable build. Heh. 😅

PR #2529, opened 2020-07-18
Allow Injection of Dependencies for SwitchStoreNoticePresenter

shiki:
Ref #2254. 

I needed to use `SwitchStoreUseCase` and `SwitchStoreNoticePresenter` for my [WIP branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/2254-fix-review-push-navigation?expand=1) for #2254. I wanted to write a unit test involving `SwitchStoreNoticePresenter` but I needed to be able to inject `NoticePresenter`. This can be tricky because it's using `ServiceLocator` **inline** instead of injection. 

https://github.com/woocommerce/woocommerce-ios/blob/02d00261738596044f92c91b84d6d9a18873adc6/WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift#L20

It's always better to inject it directly without relying on another third-party (ServiceLocator). 

## Changes

This PR refactors `SwitchStoreNoticePresenter` a bit so that:

- Dependencies can be injected
- There is no `static` function

I also added a small unit test.

## Testing

Please do a quick regression test:

1. Log in to the app. 
2. Confirm that there is no notice shown after the site picker. 
3. Switch to a different store.
4. Confirm that a notice is shown, informing the user that they will only receive notifications for the current store. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/WooCommerceTests/Authentication/Epilogue/SwitchStoreNoticePresenterTests.swift
+++ b/WooCommerce/WooCommerceTests/Authentication/Epilogue/SwitchStoreNoticePresenterTests.swift
@@ -1,0 +1,75 @@
+
+import Foundation
+import XCTest
+
+import Yosemite
+
+@testable import WooCommerce
+
+/// Test cases for `SwitchStoreNoticePresenter`.
+///
+final class SwitchStoreNoticePresenterTests: XCTestCase {
+
+    private var sessionManager: SessionManager!
+    private var noticePresenter: MockNoticePresenter!
+
+    override func setUp() {
+        super.setUp()
+
+        sessionManager = SessionManager.testingInstance
+        noticePresenter = MockNoticePresenter()
+    }
+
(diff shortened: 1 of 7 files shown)


WooCommerce/WooCommerceTests/Authentication/Epilogue/SwitchStoreNoticePresenterTests.swift
```diff
@@ -0,0 +1,75 @@
...
+import Foundation
+import XCTest
+
+import Yosemite
+
+@testable import WooCommerce
+
+/// Test cases for `SwitchStoreNoticePresenter`.
+///
+final class SwitchStoreNoticePresenterTests: XCTestCase {
```
pmusolino, line 11: 👏 👏

shiki:
Thank you, @pmusolino and @jaclync!

PR #2536, opened 2020-07-21
Stats: Show Contact Us Page When "Contact Us" Button is Tapped

### LINKED ISSUE #2535: Navigate to Zendesk Support after tapping on action link on deprecated stats banner
Deprecated stats banner action link should follow the same behavior as 

> show the Settings → Help & Support → Contact Support (Zendesk)

shiki:
Closes #2535. 

In https://github.com/woocommerce/woocommerce-ios/pull/2521, Ernesto implemented the empty state shown if the user is still using WC < 4.0. The empty state implemented in that PR currently opens a URL about WCiOS. This PR changes that so that the Zendesk Contact Us page is shown instead.

### EmptyStateViewController - withSupportRequest

I added a new `Config` option, [`withSupportRequest`](https://github.com/woocommerce/woocommerce-ios/blob/80f94af5719219b361f77ae94e41a733eb0875e6/WooCommerce/Classes/ViewRelated/ReusableViews/EmptyStateViewController/EmptyStateViewController.swift#L276-L280). When used, this option: 

- Shows all the UI elements
- Shows a text-style button (like the one implemented in #2521)
- Shows the Zendesk “Contact us” page when the button is tapped. 

I contemplated allowing consumers to specify a custom `block` but I believe supporting this *inside* `EmptyStateViewController` is simpler. And it keeps the original goal I had with `EmptyStateViewController` where it should provide simple options for configuration and the options are based on the design standards. I explained that more in a new comment:

https://github.com/woocommerce/woocommerce-ios/blob/80f94af5719219b361f77ae94e41a733eb0875e6/WooCommerce/Classes/ViewRelated/ReusableViews/EmptyStateViewController/EmptyStateViewController.swift#L253-L263

My bad, I should have documented this important fact beforehand. I even forgot what I was trying to do here and only remembered once I was going through the code. 🤦 

### EmptyStateViewController - LinkAction

I removed the `LinkAction` added in #2521 since it is no longer used after this change. If we ever need it again, we'll probably have to question why we have so many different empty state designs variations. 

## Testing

1. In `DashboardUIFactory`, change `september1st2020` to be `2019`.
2. In `StatsVersionCoordinator`, change the `let nextVersion: StatsVersion = isStatsV4Available ? .v4: .v3` line to:

    ```swift
    let nextVersion = StatsVersion.v3
    ```
3. Run the app. 
4. You should be seeing the empty state shown in the GIF above. 
5. Tap on the “Still need help?...” button. 
6. Confirm that the “Contact us” dialog was shown. 
7. Submit a message. 
8. Confirm that the message was submitted to Zendesk. 

## Reviewing 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/Tools/Zendesk/ZendeskManager.swift
+++ b/WooCommerce/Classes/Tools/Zendesk/ZendeskManager.swift
@@ -13,11 +13,21 @@ extension NSNotification.Name {
     static let ZDPNCleared = NSNotification.Name(rawValue: "ZDPNCleared")
 }
 
+/// Defines methods for showing Zendesk UI.
+///
+/// This is primarily used for testability. Not all methods in `ZendeskManager` are defined but
+/// feel free to add them when needed.
+///
+protocol ZendeskManagerProtocol {
+    /// Displays the Zendesk New Request view from the given controller, for users to submit new tickets.
+    ///
+    func showNewRequestIfPossible(from controller: UIViewController, with sourceTag: String?)
+}
 
 /// This class provides the functionality to communicate with Zendesk for Help Center and support ticket interaction,
 /// as well as displaying views for the Help Center, new tickets, and ticket list.
 ///
-final class ZendeskManager: NSObject {
+final class ZendeskManager: NSObject, ZendeskManagerProtocol {
 
     /// Shared Instance
     ///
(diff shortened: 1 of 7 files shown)


shiki:
Thank you!

PR #2550, opened 2020-07-23
Integrates wc-analytics/leaderboards API

### LINKED ISSUE #2422: v4 Top Performers: Add API support
Add the necessary network components to start using the v4-analytics API to fetch the top products (a.k.a Top Performers): `\wc-analytics\leaderboards`. This endpoint supports the following parameters for a `GET`:
```
"endpoints": [
	{
		"methods": [
			"GET"
		],
		"args": {
			"page": {
				"required": false,
				"default": 1,
				"description": "Current page of the collection.",
				"type": "integer"
			},
			"per_page": {
				"required": false,
				"default": 5,
				"description": "Maximum number of items to be returned in result set.",
				"type": "integer"
			},
			"after": {
				"required": false,
				"description": "Limit response to resources published after a given ISO8601 compliant date.",
				"type": "string"
			},
			"before": {
				"required": false,
				"description": "Limit response to resources published before a given ISO8601 compliant date.",
				"type": "string"
			},
			"persisted_query": {
				"required": false,
				"description": "URL query to persist across links.",
				"type": "string"
			}
		}
	}
],
```

Ecarrion:
Closes https://github.com/woocommerce/woocommerce-ios/issues/2422

# Why

This PR takes over the work on adapting the new `wc-analytics/leaderboards` API into the top performers' view of the app.
This is mainly done in order to solve some inconsistency problems between the TopEarners API and the web dashboard.
 
# How

The `wc-analytics/leaderboards` is not mobile-friendly and misses some of the data. This [issue](woocommerce/woocommerce#32274) asks wc-admin team to improve it.

While that is improved, we have decided to take the approach of converting the `leaderboards` into our current `TopEearnersStats` object, the rationale is to not rearchitect our model if/when this API improves.

 In order to do ☝️  we have to:
- Extract the productIDs from the `leaderboards` API, which come inside an HTML.
- Fetch those products from the internal storage or the remote service.
- Assemble the `TopEearnersStats` object with the leaderboard + product information

To accomplish that, this PR adds changes to 3 main components:
-  `LeaderboardsRemote`: Requests leaderboard API and parses its response them into the concrete `Leaderboard` data structure.
- `LeaderboardStatsConverter`: Provides a set of static helpers to allow us to convert a `Leaderboard` into a `TopEarnerStat` object.
- `StatsStoreV4`: Adds changes to orchestrate `LeaderboardsRemote` and `LeaderboardStatsConverter` when the app requests top performers data.

**PS: This PR has gotten way too big and I couldn't split it successfully without losing the commit history :-(**
**Let me know if you still prefer it split**

# Screenshots

Web|App
----|----
web|app

# Testing Steps
- Launch the app and navigate through the top performer's section
- Make sure the values displayed in the app are the same as in the web WC dashboard for the given time period.

Update release notes:


Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
```diff
@@ -0,0 +1,82 @@
...
+    /// EG:  `<a href='https://store.com?products=9'>Product</a>`
+    ///
+    static func infeerProductID(fromHTMLString html: String) -> Int64? {
+
+        // Parse and extract the `products` parameter out the the html using `Aztec parser` and `URLComponents`
+        let parsed = HTMLParser().parse(html)
+        guard let a = parsed.firstChild(ofType: .a),
+            let href = a.attribute(ofType: .href)?.value.toString(),
+            let queryItems = URLComponents(string: href)?.queryItems,
+            let productItemValue = queryItems.first(where: { $0.name == "products"} )?.value else {
```
jaclync, line 55: wow! 😵

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
```diff
@@ -0,0 +1,82 @@
...
+        // Try to convert the `productID` to an `Int`
+        return Int64(productItemValue)
+    }
+
+    /// Converts a `leaderboard row(top product)` into a `TopEarnerStatsItem` using an array of stored products to match
+    ///
+    static func topEearnerStatItem(from topProduct: LeaderboardRow, using storedProducts: [Product]) -> TopEarnerStatsItem? {
+
+        // Match the product that corresponds to the leaderboard row(top product)
+        guard let productID = Self.infeerProductID(fromHTMLString: topProduct.subject.display),
```
jaclync, line 68: how's the performance of `infeerProductID` extracting the product ID from HTML? if it's not immediately fast, I was wondering if it's worth calling `func topProductsIDs` once in the beginning to get the product IDs and pass the product ID to this function

Yosemite/YosemiteTests/Stores/StatsStoreV4Tests.swift
```diff
@@ -239,50 +239,63 @@ final class StatsStoreV4Tests: XCTestCase {
...
 
-        statsStore.onAction(action)
-        wait(for: [expectation], timeout: Constants.expectationTimeout)
+        // Then
+        XCTAssertNil(topEarnersError)
+        XCTAssertEqual(self.viewStorage.countObjects(ofType: Storage.TopEarnerStats.self), 1)
+        XCTAssertEqual(self.viewStorage.countObjects(ofType: Storage.TopEarnerStatsItem.self), 2)
+
+        let readOnlyTopEarnerStats = self.viewStorage.firstObject(ofType: Storage.TopEarnerStats.self)?.toReadOnly()
+        XCTAssertEqual(readOnlyTopEarnerStats, self.sampleTopEarnerStats())
```
jaclync, line 267: nit: `self` probably isn't needed for the above 5 lines?

Yosemite/YosemiteTests/Stores/StatsStoreV4Tests.swift
```diff
@@ -239,50 +239,63 @@ final class StatsStoreV4Tests: XCTestCase {
...
 
-        statsStore.onAction(action)
-        wait(for: [expectation], timeout: Constants.expectationTimeout)
+        // Then
+        XCTAssertNil(topEarnersError)
+        XCTAssertEqual(self.viewStorage.countObjects(ofType: Storage.TopEarnerStats.self), 1)
+        XCTAssertEqual(self.viewStorage.countObjects(ofType: Storage.TopEarnerStatsItem.self), 2)
+
+        let readOnlyTopEarnerStats = self.viewStorage.firstObject(ofType: Storage.TopEarnerStats.self)?.toReadOnly()
+        XCTAssertEqual(readOnlyTopEarnerStats, self.sampleTopEarnerStatsMutated())
```
jaclync, line 298: nit: `self` probably isn't needed for the above 5 lines?

Yosemite/YosemiteTests/Tools/Stats/LeaderboardStatsConverterTests.swift
```diff
@@ -0,0 +1,96 @@
...
+        // Then
+        XCTAssertTrue(statItems.isEmpty)
+    }
+}
+
+/// MARK: Test functions to generate sample data
+///
+private extension LeaderboardStatsConverterTest {
+
+    static func sampleLeaderboard(productIds: [Int64]) -> Leaderboard {
```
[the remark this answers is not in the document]
jaclync: super nit: we usually call `productIDs` 🙂

Yosemite/YosemiteTests/Tools/Stats/LeaderboardStatsConverterTests.swift
```diff
@@ -0,0 +1,96 @@
...
+        let leaderboard = Self.sampleLeaderboard(productIds: Array((1...5)))
+
+        // When
+        let missingIds = LeaderboardStatsConverter.missingProductsIDs(from: leaderboard, in: products)
+
+        // Then
+        XCTAssertTrue(missingIds.isEmpty)
+    }
+
+    func testConvertToProductsIntoStatItemsUsinfStoredProducts() {
```
jaclync, line 47: nit: "UsinfStoredProducts" -> "UsingStoredProducts"? 😄

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
```diff
@@ -0,0 +1,82 @@
...
+        let storedProductsIDs = storedProducts.map { $0.productID }
+
+        // Figure out which top products IDs are missing from store products IDs
+        let missingIDs = Set(topProductsIDs).subtracting(storedProductsIDs)
+        return Array(missingIDs)
+    }
+
+    /// Converts a the`leaderboard rows(top products)` into an array `TopEarnerStatsItem` using an array of stored products to match
+    ///
+    static func topEearnerStatItems(from topProducts: Leaderboard, using storedProducts: [Product]) -> [TopEarnerStatsItem] {
```
jaclync, line 34: super nit: maybe "*Stat**s**Items" to match the return type `[TopEarnerStatsItem]`?

Yosemite/YosemiteTests/Tools/Stats/LeaderboardStatsConverterTests.swift
```diff
@@ -0,0 +1,96 @@
...
+
+/// LeaderboardProductParser unit tests
+///
+final class LeaderboardStatsConverterTest: XCTestCase {
+
+    /// Sample site ID
+    ///
+    let siteID: Int64 = 1234
+
+    func testCorrectProductIdsFormLeaderboardOfTopProducts() {
```
jaclync, line 12: nit: maybe "testCorrectProductIDsFromLeaderboardOfTopProducts`?

jaclync (commented):
@Ecarrion Thanks so much for working through the response format from leaderboards API that isn't mobile friendly at all! It's crazy that we even have to extract the product ID from an `a` tag 🤦🏻‍♀️

Code: mostly nits and some questions!
Testing: I noticed that the API and core are sorting the top products by "items sold", shall we follow that in mobile as well? We might want to change the current "TOTAL SPEND" label to "ITEMS SOLD"? otherwise, in the example of "this year" screenshots below, the top 3 products aren't the correct top products sorted by total $.

range | iOS | core
-- | -- | --
this year | Simulator Screen Shot - iPhone 11 - 2020-07-23 at 16 26 45 | Screen Shot 2020-07-23 at 4 27 29 PM
this month | Simulator Screen Shot - iPhone 11 - 2020-07-23 at 16 26 43 | Screen Shot 2020-07-23 at 4 27 03 PM

Networking/Networking/Model/Stats/Leaderboard.swift
```diff
@@ -0,0 +1,68 @@
...
+        // [ [ { content1 }, { content2 }, { content3 }], [ { content1 }, { content2 }, { content3 } ] ]
+        //
+        // Into:
+        // [ leaderboardRow, leaderboardRow ]
+        //
+        // By re-encoding the underlying json into:
+        // [ mergeDic(content1, content2, content3), mergeDic(content1, content2, content3) ]
+        //
+        //
+        self.rows = try rawRows.map { rawRow in
```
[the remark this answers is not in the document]
Ecarrion: I know! Took my a while to figure out the best way(a way?) to model this API 🙃

Storage/Storage/Tools/StorageType+Extensions.swift
```diff
@@ -257,6 +257,11 @@ public extension StorageType {
         return allObjects(ofType: Product.self, matching: predicate, sortedBy: [descriptor])
     }
 
+    func loadProducts(siteID: Int64, productsIDs: [Int64]) -> [Product] {
```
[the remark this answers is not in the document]
Ecarrion: Totally! 👍

Networking/NetworkingTests/Remote/LeaderboardsRemoteTests.swift
```diff
@@ -0,0 +1,75 @@
...
+        XCTAssertFalse(topProducts.rows.isEmpty)
+
+        // Each prodcut should have non-empty values
+        topProducts.rows.forEach { product in
+            XCTAssertFalse(product.subject.display.isEmpty)
+            XCTAssertFalse(product.subject.value.isEmpty)
+            XCTAssertFalse(product.quantity.display.isEmpty)
+            XCTAssertTrue(product.quantity.value > 0)
+            XCTAssertFalse(product.total.display.isEmpty)
+            XCTAssertTrue(product.total.value > 0)
```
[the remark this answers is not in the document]
Ecarrion: My rationale was that this may be easier to maintain in cases where we updated the mock JSON file.
But I can add the concrete assert if you prefer them!

Networking/NetworkingTests/Remote/LeaderboardsRemoteTests.swift
```diff
@@ -0,0 +1,75 @@
+import XCTest
+@testable import Networking
+
+/// Leaderboards Unit Tests
+///
+final class leaderboardsRemoteV4Tests: XCTestCase {
```
shiki, line 6:
```suggestion
final class LeaderboardsRemoteV4Tests: XCTestCase {
```

🙂

Networking/NetworkingTests/Remote/LeaderboardsRemoteTests.swift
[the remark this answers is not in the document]
shiki: I think the actual value is preferred since we don't expect the mock JSON file to change a lot. And it will help catch numerical JSON conversion. Unlikely to fail but still. 🤷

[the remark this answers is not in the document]
shiki: Also for future reference → [XCTAssertGreaterThanOrEqual](https://developer.apple.com/documentation/xctest/xctassertgreaterthanorequal) exists. 🙂

Yosemite/Yosemite/Stores/StatsStoreV4.swift
```diff
@@ -285,6 +296,97 @@ extension StatsStoreV4 {
...
 
+
+// MARK: Convert Leaderboard into TopEarnerStats
+//
+private extension StatsStoreV4 {
+
+    /// Converts and stores a top-product `leaderboard` into a `StatsTopEarner`
+    /// Since  a `leaderboard` does not containt  the necesary product information, this method fetches the related product before starting the convertion.
+    ///
+    private func convertAndsStoreLeaderboarsIntoTopEarners(siteID: Int64,
```
shiki, line 307: The `private` is not needed since this is a `private` extension. 🙂 

```suggestion
    func convertAndsStoreLeaderboarsIntoTopEarners(siteID: Int64,
```

shiki (commented):
I just glanced at this and left nitpicks since Jaclyn's pretty much covered this already. 🙂

Yosemite/YosemiteTests/Tools/Stats/LeaderboardStatsConverterTests.swift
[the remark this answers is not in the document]
Ecarrion: 👍

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
```diff
@@ -0,0 +1,82 @@
...
+    ///
+    static func topProductsIDs(from topProducts: Leaderboard) -> [Int64] {
+        return topProducts.rows.compactMap {
+            Self.infeerProductID(fromHTMLString: $0.subject.display)
+        }
+    }
+
+    /// Indicates what products are missing from a leaderboard in an array of stored products
+    ///
+    static func missingProductsIDs(from topProducts: Leaderboard, in storedProducts: [Product]) -> [Int64] {
```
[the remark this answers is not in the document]
Ecarrion: I thought about doing that, but passing an array of `Ints` semed less readable and more error-prone than passing the concrete object.

Since the products set is small I think It's fine to prefer readability over performance here :-)

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
```diff
@@ -0,0 +1,82 @@
...
+import class Aztec.HTMLParser
+import struct Aztec.Element
+
+/// Helper functions to  convert a `Leaderboard` type into a `TopEearnerStat`
+///
+struct LeaderboardStatsConverter {
+
+    /// Returns the ids of the products associated with a top products leaderboar
+    ///
+    static func topProductsIDs(from topProducts: Leaderboard) -> [Int64] {
```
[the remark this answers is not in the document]
Ecarrion: I use static functions when `self` is not a dependency on the work that I have to do.
I think that using a static function expresses the reader a clear intention that:
- The function does not depend on any stored state
- The function does not mutate any state within the object.

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
Ecarrion: I started from the baseline that parsing that HTML using Aztec parser was quite performant.
After running a quick `measure test` in Xcode I don't think we have much to worry about, especially because the product set is so small.

Screen Shot 2020-07-23 at 3 40 05 PM

Networking/NetworkingTests/Remote/LeaderboardsRemoteTests.swift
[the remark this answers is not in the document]
Ecarrion: Done 556c4b43a6fd530d5aeb4b4e9cacc53ac8c40849

Yosemite/Yosemite/Stores/StatsStoreV4.swift
```diff
@@ -285,6 +296,91 @@ extension StatsStoreV4 {
...
+    /// Since  a `leaderboard` does not containt  the necesary product information, this method fetches the related product before starting the convertion.
+    ///
+    private func convertLeaderboarsAndStoreTopEearners(siteID: Int64,
+                                                       granularity: StatGranularity,
+                                                       date: Date,
+                                                       leaderboards: [Leaderboard],
+                                                       onCompletion: @escaping (Error?) -> Void) {
+
+        // API guarantees a fixed index for the top products leaderboard
+        let topProducts = leaderboards[Constants.topProductsIndex]
```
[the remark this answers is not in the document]
Ecarrion: Done a44bda16061c7313255b851a4657c05a43fdc2c5

Ecarrion:
Hey @jaclync! 
I've made all the necessary changes and have answered the pending questions.
Please take another look. Also, sorry for the huge PR :-(

Yosemite/YosemiteTests/Tools/Stats/LeaderboardStatsConverterTests.swift
```diff
@@ -0,0 +1,96 @@
...
+
+/// LeaderboardProductParser unit tests
+///
+final class LeaderboardStatsConverterTest: XCTestCase {
+
+    /// Sample site ID
+    ///
+    private let siteID: Int64 = 1234
+
+    func testCorrectProductIDsFormLeaderboardOfTopProducts() {
```
jaclync, line 12: nit: `testCorrectProductIDsFromLeaderboardOfTopProducts` ("Form" -> "From")?

Yosemite/Yosemite/Stores/StatsStoreV4.swift
```diff
@@ -285,6 +296,100 @@ extension StatsStoreV4 {
...
+    }
+
+    /// Returns all stored products for a given site ID
+    ///
+    func loadStoredProducts(siteID: Int64, productIDs: [Int64] ) -> [Networking.Product] {
+        let products = storageManager.viewStorage.loadProducts(siteID: siteID, productsIDs: productIDs)
+        return products.map { $0.toReadOnly() }
+    }
+
+    /// Merges and stored a top-product leaderboard with an array of stored products into  a `StatsTopEarner` object
```
jaclync, line 375: super nit: "Merges and stored" -> "Merges and stores" and "StatsTopEarner" -> "TopEarnerStats"?

Yosemite/Yosemite/Stores/StatsStoreV4.swift
```diff
@@ -285,6 +296,100 @@ extension StatsStoreV4 {
...
 
+
+// MARK: Convert Leaderboard into TopEarnerStats
+//
+private extension StatsStoreV4 {
+
+    /// Converts and stores a top-product `leaderboard` into a `StatsTopEarner`
+    /// Since  a `leaderboard` does not containt  the necesary product information, this method fetches the related product before starting the convertion.
+    ///
+    func convertAndsStoreLeaderboarsIntoTopEarners(siteID: Int64,
```
jaclync, line 307: super nit: "convertAndsStore" -> "convertAndStore" and "Leaderboars" -> "Leaderboards"?

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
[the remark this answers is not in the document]
jaclync: in my previous work, we also had the same reasons for using `static`. personally, I slightly prefer using member functions for all in case a state is added and it looks nicer without `Self` 😅  I'm fine with either way though, I remember having a conversation with @shiki earlier just thought Jayson might be interested 🙂

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
jaclync: thanks for checking this Ernesto! sounds good 👍

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
```diff
@@ -0,0 +1,82 @@
...
+        let storedProductsIDs = storedProducts.map { $0.productID }
+
+        // Figure out which top products IDs are missing from store products IDs
+        let missingIDs = Set(topProductsIDs).subtracting(storedProductsIDs)
+        return Array(missingIDs)
+    }
+
+    /// Converts a the`leaderboard rows(top products)` into an array `TopEarnerStatsItem` using an array of stored products to match
+    ///
+    static func topEearnerStatsItems(from topProducts: Leaderboard, using storedProducts: [Product]) -> [TopEarnerStatsItem] {
```
jaclync, line 34: nit: "Eearner" -> "Earner"?

jaclync (approved):
Thank you for addressing the comments 💟  I left some very small nits, the PR LGTM! :shipit: 

Regarding the "Total spend" label, feel free to update separately based on G's response!

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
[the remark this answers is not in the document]
Ecarrion: I'm happy to follow any convention that the team has previously decided tho! :-)

Yosemite/Yosemite/Stores/StatsStoreV4.swift
Ecarrion: lol I think I'm dyslexic! 🤪

Ecarrion:
Thanks again @jaclync for the review!
> Regarding the "Total spend" label, feel free to update separately based on G's response!

Will do!

Yosemite/Yosemite/Tools/Stats/LeaderboardStatsConverter.swift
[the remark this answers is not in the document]
shiki: I want to say that it's fine to have `static` methods on helpers. For the case of `LeaderboardStatsConverter`, I think it's grown beyond a simple helper. On a quick glance, the methods do not look related to me at all. First, the comment is:

```swift
/// Helper functions to  convert a `Leaderboard` type into a `TopEearnerStat`
```

But only one of the 4 methods return `TopEearnerStatItem`. The _helper_ failed in the contract that it tried to represent, which is converting something to `TopEearnerStat`. 

The separation of the 3 additional _accessible_ methods also adds friction to understanding how the methods should be used together. And it will get cumbersome to refactor since they are all accessible and you don't know where they are truly used. 

Based on how I see them being used in [`loadProduct`](https://github.com/woocommerce/woocommerce-ios/blob/9a0409d302549e9548744ef8c138f739301e284f/Yosemite/Yosemite/Stores/StatsStoreV4.swift#L337-L340), a possible improvement would be using the  `*UseCase` ([example](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Stores/ProductReview/RetrieveProductReviewFromNoteUseCase.swift)) strategy. We can gather these three lines maybe:

```swift
        // Workout if we have stored all products that relate to the given leaderboard
        let topProductIDs = LeaderboardStatsConverter.topProductsIDs(from: topProducts)
        let topStoredProducts = loadStoredProducts(siteID: siteID, productIDs: topProductIDs)
        let missingProductsIDs = LeaderboardStatsConverter.missingProductsIDs(from: topProducts, in: topStoredProducts)
```

And then let's ask ourselves. What are the inputs and outputs? For the lines above, I think the input is `Leaderboard` and the output is `[Int64]`. Notice that we don't need `LeaderboardStatsConverter.topProductsIDs` for this routine so we can **eliminate** that right away. The `*UseCase` can probably be something like this:

```swift
struct DoSomethingUseCase {
    private let leaderboard: Leaderboard

    func getMissingProductIds() -> [Int64] {
        // call topProductIds()
        // call loadStoredProducts
        return missingProductIds
    }

    private func topProductIds(from: Leaderboard) -> [Int64] {
    
    }
}
```

This `UseCase` is still testable in isolation. The big win here for me is that there is only 1 method and everything else is private. It is much easier to reason about how to use this kind of _helper_. In the end, you can think of the `UseCase` as a `static func` that contains **a lot** of `private static func`. 😄 

There's a lot that goes into these of course so I definitely skipped a lot of things to discuss here. 

----

And as I'm writing this, I realized I didn't answer Jaclyn's points. 😄 

> I slightly prefer using member functions for all in case a state is added 

I agree with Ernesto about having clearer intentions that there is no state. Though this comes with the problem of having too many arguments. I think there needs to be a balance. With the `UseCase` strategy, we've eliminated both problems. 

---

Another point! 😅 

Our Yosemite Store classes tend to get really large. Eventually, we're really going to have to split things up into logical chunks. It would be a shame if we split them using multiple `static funcs` instead of business-logic UseCases: 

https://github.com/woocommerce/woocommerce-ios/blob/9a0409d302549e9548744ef8c138f739301e284f/Yosemite/Yosemite/Stores/ProductReviewStore.swift#L104-L114

---

In summary:

- `static func`: Good for _short_ helpers but probably not as the number lines get longer. And probably not for business logic.
- private members: Good for short classes but as the class gets bigger, the number of methods grow and everything becomes harder to digest. 
- logical chunks (like UseCase): good for encapsulating a single input and output that involves a lot of different routines. 

Note: I'm not saying what's the _best_ thing here. I'm just sharing what I think when to use what. 🙂 And every day my opinions change! 😁

Issue #2085, opened 2020-04-03
Edit Variable Products: edit product variation


jaclync:
More details in p91TBi-33Q-p2

Differences between Product and Product Variation screens:

- Edit product main screen
  - A “Visible” row where the user can show/hide a product variation - will be implemented in https://github.com/woocommerce/woocommerce-ios/issues/2511
  - [x] Product variation name row is now readonly https://github.com/woocommerce/woocommerce-ios/pull/2588
  - [x] Hide the ellipsis menu from the navigation bar for a product variation https://github.com/woocommerce/woocommerce-ios/pull/2588
- Product images screen
  - [ ] Update UI to be https://github.com/woocommerce/woocommerce-ios/pull/2589
![image](https://user-images.githubusercontent.com/1945542/89001966-6c784880-d32e-11ea-9a27-4ae34bde732a.png)
  - [ ] Remove trash icon in the image details due to an API issue #2576  https://github.com/woocommerce/woocommerce-ios/pull/2589
- Price settings screen
  - [x] Remove tax section (tax class, tax status) https://github.com/woocommerce/woocommerce-ios/pull/2566
- Inventory settings screen
  - [x] Remove “Limit one per order” (`sold_individually`) field https://github.com/woocommerce/woocommerce-ios/pull/2579
  - [x] When "Manage stock" toggle is disabled, the stock status is readonly (from the parent product) https://github.com/woocommerce/woocommerce-ios/pull/2588
- Shipping settings screen
  - [x] Hide the shipping class row due to an API issue https://github.com/woocommerce/woocommerce-ios/pull/2588

jaclync:
hey @Garance91540! I'd like to clarify the product type options for a product variation:

<img width="134" alt="Screen Shot 2020-05-20 at 3 21 24 PM" src="https://user-images.githubusercontent.com/1945542/82417215-cd251280-9aad-11ea-8e00-746675221dd4.png">

for a product variation, we can't change the product type since it's part of a variable product. so I think the "product type" here is to set the "virtual" field like in core:

<img width="300" alt="Screen Shot 2020-05-20 at 3 27 06 PM" src="https://user-images.githubusercontent.com/1945542/82417643-6ce2a080-9aae-11ea-89b6-18d22bc892ce.png">

in this case, I think we can use the "Physical product" and "Virtual product" options in the original design?

[...]

PR #2552, opened 2020-07-23
Replace concrete usage with protocol: `ProductFormDataModel`, `ProductFormActionsFactoryProtocol`, and `TaxClassRequestable`

jaclync:
Prerequisite for #2085 

This PR abstracts parts of concrete usage with protocol in the product form and its related classes, so that `ProductVariation` can be supported for these classes in a future subtask. These changes should be pretty straightforward. I left the more complicated protocol that abstracts the current `ProductFormViewModel` for a future subtask. In addition, I left the `Product` protocol replacement in price and inventory settings for later since these 2 screens have UI changes between `Product` and `ProductVariation` (p91TBi-33Q-p2).

## Changes

- Created `ProductFormActionsFactoryProtocol` protocol that the existing `ProductFormActionsFactory` conforms to, and updated its existing usage
- Created `ProductFormDataModel` protocol that abstracts the `Product` usage in product form. It contains all the necessary fields for `ProductFormViewController` and its navigated view controllers. Replaced `Product` with `ProductFormDataModel` in product form related classes where replacement is straightforward
- Created `TaxClassRequestable` protocol that abstracts the `Product` usage in `TaxClassAction.requestMissingTaxClasses` within price settings. Updated existing usage in price settings classes
- Removed unused `product: Product` associated value in `ProductFormSection.PrimaryFieldRow.images` enum case
- Removed redundant extension helper `isShippingEnabled` in Yosemite layer and moved its unit tests from Yosemite to app layer.

## Testing

No changes in app behavior are expected, feel free to test for regression with the following steps:

- Go to the Products tab
- Tap on a simple product
- Tap on the description row, and make sure the editing behavior stays the same
- Go back to the product details
- Tap on the price settings row, and make sure the editing behavior stays the same especially on the tax class selector
- Go back to the product details
- Tap on the shipping settings row, and make sure the editing behavior stays the same especially on the shipping class selector
- Go back to the product details
- Tap on the "+" cell in the images header, and make sure image editing behavior stays the same

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
@@ -1,0 +1,81 @@
+import Yosemite
+
+/// Describes a data model that contains necessary properties for rendering a product form (`ProductFormViewController`).
+protocol ProductFormDataModel {
+    // General
+    var siteID: Int64 { get }
+    var productID: Int64 { get }
+    var name: String { get }
+    var description: String? { get }
+    var shortDescription: String? { get }
+
+    // Settings
+    var virtual: Bool { get }
+    var downloadable: Bool { get }
+    var permalink: String { get }
+
+    // Images
+    var images: [ProductImage] { get }
+
+    // Price
+    var regularPrice: String? { get }
+    var salePrice: String? { get }
+    var dateOnSaleStart: Date? { get }
+    var dateOnSaleEnd: Date? { get }
+    var taxStatusKey: String { get }
+    var taxClass: String? { get }
+
+    // Shipping
+    var weight: String? { get }
+    var dimensions: ProductDimensions { get }
+    var shippingClass: String? { get }
+    var shippingClassID: Int64 { get }
+
+    // Inventory
+    var sku: String? { get }
+    var manageStock: Bool { get }
+    var stockStatus: ProductStockStatus { get }
+    var stockQuantity: Int64? { get }
+    var backordersKey: String { get }
+}
+
+// MARK: Helpers that can be derived from `ProductFormDataModel`
+//
+extension ProductFormDataModel {
+    /// Returns the full description without the HTML tags and leading/trailing white spaces and new lines.
+    var trimmedFullDescription: String? {
+        guard let description = description else {
+            return nil
+        }
+        return description.removedHTMLTags.trimmingCharacters(in: .whitespacesAndNewlines)
+    }
+
+    /// Whether shipping settings are available for the product.
+    var isShippingEnabled: Bool {
+        return downloadable == false && virtual == false
+    }
+
+    /// Returns `ProductTaxStatus` given the raw value from `taxStatusKey` field.
+    var productTaxStatus: ProductTaxStatus {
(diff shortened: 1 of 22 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
```diff
@@ -0,0 +1,81 @@
...
+        fullDescription
+    }
+
+    var shortDescription: String? {
+        briefDescription
+    }
+
+    var stockStatus: ProductStockStatus {
+        productStockStatus
+    }
```
Ecarrion, line 80: Nit: It looks a bit odd that these don't use the return keyword but the ones above do. Do you think we should make it consistent?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
```diff
@@ -0,0 +1,81 @@
...
+
+    /// Returns `ProductTaxStatus` given the raw value from `taxStatusKey` field.
+    var productTaxStatus: ProductTaxStatus {
+        return ProductTaxStatus(rawValue: taxStatusKey)
+    }
+
+    /// Returns `ProductBackordersSetting` given the raw value from `backordersKey` field.
+    var backordersSetting: ProductBackordersSetting {
+        return ProductBackordersSetting(rawValue: backordersKey)
+    }
```
Ecarrion, line 66: Q: is it worth to add unit tests to these functions?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
jaclync: yeaa I'm all for consistency! update in https://github.com/woocommerce/woocommerce-ios/pull/2552/commits/3f200d10f2fe67ef154aae7ad9807cf636d2031b to remove unnecessary `return` in the ones above (one line return)

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
jaclync: yes! we have some existing unit tests and I just added some:

- `trimmedFullDescription`: [`Product_ProductFormTests.testTrimmedFullDescriptionWithLeadingNewLinesAndHTMLTags`](https://github.com/woocommerce/woocommerce-ios/blob/09573496fe95d51dbeb4125a8fb57ea2e76cfd2c/WooCommerce/WooCommerceTests/ViewRelated/Products/Product+ProductFormTests.swift#L10-L15)
- `isShippingEnabled`: [`Product_ShippingTests`](https://github.com/woocommerce/woocommerce-ios/blob/fcc5b976b691127efbbd215a5cef99577ee28242/WooCommerce/WooCommerceTests/Extensions/Product%2BShippingTests.swift)
- `productTaxStatus` and `backordersSetting`: just added in https://github.com/woocommerce/woocommerce-ios/pull/2552/commits/27ac6231e47020fa94dd7d95f22750cd757582bd

jaclync:
Thanks for the suggestions Ernesto! updated the PR based on your feedback 😄 merging this now!

PR #2555, opened 2020-07-23
Add Feature Flag for In-app Feedback Project

### LINKED ISSUE #2538: [IAFeedback] Create In-app Feedback Feature Flag
The feature flag will only be used to hide the In-app Feedback project work until we can release it.

shiki:
Closes #2538

This just adds a feature flag. I also tried to improve `DefaultFeatureFlagService` a little bit.

## Testing

There are no functional changes. Please just confirm that the build passes. 🙂 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
+++ b/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
@@ -1,20 +1,22 @@
 struct DefaultFeatureFlagService: FeatureFlagService {
     func isFeatureFlagEnabled(_ featureFlag: FeatureFlag) -> Bool {
+        let buildConfig = BuildConfiguration.current
+
         switch featureFlag {
         case .barcodeScanner:
-            return BuildConfiguration.current == .localDeveloper || BuildConfiguration.current == .alpha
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/System/DefaultFeatureFlagService.swift
```diff
@@ -1,18 +1,22 @@
...
         case .editProductsRelease2:
             return true
         case .editProductsRelease3:
-            return BuildConfiguration.current == .localDeveloper || BuildConfiguration.current == .alpha
+            return buildConfig == .localDeveloper || buildConfig == .alpha
         case .readonlyProductVariants:
             return true
         case .refunds:
             return true
+        case .inAppFeedback:
```
jaclync, line 18: nit: it looks like the cases are currently sorted by name, might be nice to keep the order? 😄

shiki: OMG. You don't miss anything. 😄 Fixed in 041c954e670ecebb88025e7e140bef8d68bca411.

shiki:
Thanks, Jaclyn!

Issue #2537, opened 2020-07-21
[IAFeedback] Display In-app Feedback Question in Stats


shiki:
<img src="https://user-images.githubusercontent.com/198826/88111962-1fe18000-cb6c-11ea-84cb-c305e27bdcdb.png" width="240">

Ref p91TBi-2Af-p2

## What 

This task only involves adding the section UI. The handling of the buttons should be handled in a separate task. 

## When To Show It

> We’d like users to have had time to interact with the app before providing feedback, and we also don’t want to annoy them too much by polling them too often.

1. Users who’ve installed the app at least 3 months ago
2. Once every 6 months

## When It Disappears

Pressing any of the buttons will hide the section. 

## Others

- [ ] Must be hidden behind feature flag

PR #2556, opened 2020-07-23
Add AppSettingsAction for In-app Feedback

shiki:
Part of #2537. 

This adds `AppSettingsActions` to persist information to support checking the conditions for #2537:

Condition | New `AppSettingAction`
--------|-------
Users who’ve installed the app at least 3 months ago | `setInstallationDateIfNecessary`
Once every 6 months | `setLastFeedbackDate`

As far as I can tell, we do not currently track the _installation date_ but please let me know if we do! 

## Changes

The actions eventually lead to persisting a new `struct`, [`GeneralAppSettings`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2537-add-app-settings/Storage/Storage/Model/GeneralAppSettings.swift). I pondered about this for a while. We save **a lot** of files in `AppSettingsStore`. And I think because of these, we now have so many boilerplate. It felt quite cumbersome to save the `installationDate` and `lastFeedbackDate` into separate structures and separate files. I wish we could have just saved all the settings in one file. But please let me know if I'm misunderstanding the purpose of why we did it this way. 

## Testing

These are currently not used. Please just review. 🙂 

## Submitter Checklist

### DIFF
--- a/Storage/Storage/Model/GeneralAppSettings.swift
+++ b/Storage/Storage/Model/GeneralAppSettings.swift
@@ -1,0 +1,26 @@
+
+import Foundation
+
+/// An encodable/decodable data structure that can be used to save files. This contains
+/// miscellaneous app settings.
+///
+/// Sometimes I wonder if `AppSettingsStore` should just use one plist file. Maybe things will
+/// be simpler?
+///
+public struct GeneralAppSettings: Codable, Equatable {
+    /// The known `Date` that the app was installed.
+    ///
+    /// Note that this is not accurate because this property/setting was created when we have
+    /// thousands of users already.
+    ///
+    public let installationDate: Date?
+
+    /// The last time that the user interacted with the in-app feedback (https://git.io/JJ8i0).
+    ///
+    public let lastFeedbackDate: Date?
+
+    public init(installationDate: Date?, lastFeedbackDate: Date?) {
+        self.installationDate = installationDate
+        self.lastFeedbackDate = lastFeedbackDate
+    }
+}
--- a/Yosemite/Yosemite/Actions/AppSettingsAction.swift
+++ b/Yosemite/Yosemite/Actions/AppSettingsAction.swift
@@ -71,4 +71,9 @@ public enum AppSettingsAction: Action {
     /// given `date`.
     ///
     case setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Void, Error>) -> Void))
+
+    /// Saves the `date` as the last known date that the user interacted with the in-app
+    /// feedback prompt (https://git.io/JJ8i0).
+    ///
+    case setLastFeedbackDate(date: Date, onCompletion: ((Result<Void, Error>) -> Void))
 }
--- a/Yosemite/Yosemite/Actions/AppSettingsAction.swift
+++ b/Yosemite/Yosemite/Actions/AppSettingsAction.swift
@@ -63,4 +63,12 @@ public enum AppSettingsAction: Action {
     /// Clears all the states related to stats version
     ///
     case resetStatsVersionStates
+
+    // MARK: - General App Settings
+
+    /// Saves the `date` as the last known date that the app was installed. This does not do
+    /// anything if there is a persisted installation date already and it is older than the
+    /// given `date`.
+    ///
+    case setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Void, Error>) -> Void))
 }
--- a/Yosemite/Yosemite/Actions/AppSettingsAction.swift
+++ b/Yosemite/Yosemite/Actions/AppSettingsAction.swift
@@ -70,7 +70,10 @@ public enum AppSettingsAction: Action {
     /// anything if there is a persisted installation date already and it is older than the
     /// given `date`.
     ///
-    case setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Void, Error>) -> Void))
+    /// - Parameter onCompletion: The `Result`'s success value will be `true` if the installation
+    ///                           date was changed and `false` if not.
+    ///
+    case setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Bool, Error>) -> Void))
 
     /// Saves the `date` as the last known date that the user interacted with the in-app
     /// feedback prompt (https://git.io/JJ8i0).
--- a/Yosemite/Yosemite/Stores/AppSettingsStore.swift
+++ b/Yosemite/Yosemite/Stores/AppSettingsStore.swift
@@ -131,8 +131,8 @@ private extension AppSettingsStore {
         do {
             let settings = try loadOrCreateGeneralAppSettings()
 
-            let installationDate = settings.installationDate ?? .distantFuture
-            guard date < installationDate else {
+            if let installationDate = settings.installationDate,
+                date > installationDate {
                 return onCompletion(.success(()))
             }
 
--- a/Yosemite/Yosemite/Stores/AppSettingsStore.swift
+++ b/Yosemite/Yosemite/Stores/AppSettingsStore.swift
@@ -127,19 +127,22 @@ private extension AppSettingsStore {
     /// Save the `date` in `GeneralAppSettings` but only if the `date` is older than the existing
     /// `GeneralAppSettings.installationDate`.
     ///
-    func setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Void, Error>) -> Void)) {
+    /// - Parameter onCompletion: The `Result`'s success value will be `true` if the installation
+    ///                           date was changed and `false` if not.
+    ///
+    func setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Bool, Error>) -> Void)) {
         do {
             let settings = try loadOrCreateGeneralAppSettings()
 
             if let installationDate = settings.installationDate,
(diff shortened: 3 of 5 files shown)


Yosemite/Yosemite/Actions/AppSettingsAction.swift
```diff
@@ -63,4 +63,17 @@ public enum AppSettingsAction: Action {
...
     ///
     case resetStatsVersionStates
+
+    // MARK: - General App Settings
+
+    /// Saves the `date` as the last known date that the app was installed. This does not do
+    /// anything if there is a persisted installation date already and it is older than the
+    /// given `date`.
+    ///
+    case setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Void, Error>) -> Void))
```
Ecarrion, line 73: [Newbie Q]: Is the reason on why these methods are asynchronous because the dispatch method don't allow for any return value?

Yosemite/Yosemite/Stores/AppSettingsStore.swift
```diff
@@ -108,11 +113,67 @@ public class AppSettingsStore: Store {
...
+    /// Save the `date` in `GeneralAppSettings` but only if the `date` is older than the existing
+    /// `GeneralAppSettings.installationDate`.
+    ///
+    func setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Void, Error>) -> Void)) {
+        do {
+            let settings = try loadOrCreateGeneralAppSettings()
+
+            let installationDate = settings.installationDate ?? .distantFuture
+            guard date < installationDate else {
+                return onCompletion(.success(()))
```
Ecarrion, line 136: I'm not sure if returning a `success` when the date wasn't expresses the correct intention.

What do you think on returning a `Boolean` to indicated if the date was actually saved or not?

Yosemite/Yosemite/Stores/AppSettingsStore.swift
```diff
@@ -108,11 +113,67 @@ public class AppSettingsStore: Store {
...
+private extension AppSettingsStore {
+    /// Save the `date` in `GeneralAppSettings` but only if the `date` is older than the existing
+    /// `GeneralAppSettings.installationDate`.
+    ///
+    func setInstallationDateIfNecessary(date: Date, onCompletion: ((Result<Void, Error>) -> Void)) {
+        do {
+            let settings = try loadOrCreateGeneralAppSettings()
+
+            let installationDate = settings.installationDate ?? .distantFuture
+            guard date < installationDate else {
```
Ecarrion, line 135: Maybe a doing something like 
```swift
guard let date = date, date < installationDate else {
``` 
would be easier to understand, it took me a while to get what/how `.distantFuture` played in here!

Yosemite/Yosemite/Stores/AppSettingsStore.swift
```diff
@@ -108,11 +113,67 @@ public class AppSettingsStore: Store {
...
+
+            onCompletion(.success(()))
+        } catch {
+            onCompletion(.failure(error))
+        }
+    }
+
+    /// Load the `GeneralAppSettings` from file or create an empty one if it doesn't exist.
+    func loadOrCreateGeneralAppSettings() throws -> GeneralAppSettings {
+        guard let settings: GeneralAppSettings = try fileStorage.data(for: generalAppSettingsFileURL) else {
```
Ecarrion, line 165: TIL: That you can cast things things way!

Yosemite/Yosemite/Actions/AppSettingsAction.swift
shiki: I believe so. Also, if I remember correctly, [Redux or Flux](https://facebook.github.io/flux/docs/in-depth-overview) does not even allow return values. The store should always be updated. And the UI should _observe_ the Store's state. Unfortunately, we don't have that architecture right now. It's something that I want to add in the future. 

So, for now, we are left with `onCompletion`. It's a bit cumbersome. 

I do like the concept that this enforces: updating the store should be asynchronous. This means we can improve things in the future so that they will happen in the background.

Yosemite/Yosemite/Stores/AppSettingsStore.swift
shiki: We could but I don't expect to use the value. This will only be called on app start. 

I also wanted this function to be just fire and forget. 😅 

I can still add it though! Maybe it can help in unit testing.

Yosemite/Yosemite/Stores/AppSettingsStore.swift
shiki: @Ecarrion I just realized that I couldn't do `guard let date = date` because it's not the `date` that is `Optional` but the `settings.installationDate`. I changed it to this instead:

```swift
            if let installationDate = settings.installationDate,
                date > installationDate {
                return onCompletion(.success(()))
            }
```

Changed in 8a4603f6bedaa5fbee68a9faa26367708bdd612f. Is that better?

Yosemite/Yosemite/Stores/AppSettingsStore.swift
shiki: Added `Bool` argument in 922904ce4528a5b2d9f78563c6136fb7cec628bd.

Yosemite/Yosemite/Stores/AppSettingsStore.swift
```diff
@@ -58,6 +58,11 @@ public class AppSettingsStore: Store {
         return documents!.appendingPathComponent(Constants.productsFeatureSwitchFileName)
     }()
 
+    private lazy var generalAppSettingsFileURL: URL = {
+        let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
+        return documents!.appendingPathComponent(Constants.generalAppSettingsFileName)
+    }()
+
```
Ecarrion, line 65: [nit] I'm not really a fan of `!` but how do you feel on indicating that this is somehow an unsafe path by marking the return type as `URL!`

```swift
private lazy var generalAppSettingsFileURL: URL!
```

Storage/Storage/Model/GeneralAppSettings.swift
```diff
@@ -0,0 +1,26 @@
+
+import Foundation
+
+/// An encodable/decodable data structure that can be used to save files. This contains
+/// miscellaneous app settings.
+///
+/// Sometimes I wonder if `AppSettingsStore` should just use one plist file. Maybe things will
+/// be simpler?
```
pmusolino, line 8: I'm not sure if this wanted to be a temporary comment, or if you want to remember it :D

Yosemite/YosemiteTests/Stores/AppSettingsStoreTests.swift
```diff
@@ -185,4 +185,117 @@ final class AppSettingsStoreTests: XCTestCase {
 
         waitForExpectations(timeout: 2, handler: nil)
     }
+
+    // MARK: - General App Settings
```
pmusolino, line 189: I don't know if it makes 100% sense, but maybe yes. What do you think about adding a case where `Date` is using a different timezone from the one stored?

pmusolino:
> Do we have any rule on when to use CD and when to use the file manager?
None that I know of. I think it would have been nice if we didn't use files. I find them a little finicky. There are pros and cons of course.
Maybe @pmusolino and @jaclync have some more info about this.

I think there are some historical reasons behind this choice, but I don't know the details 😃 maybe @astralbodies know why we are using Core Data and also some plist file to store other information (app settings)?

Yosemite/Yosemite/Stores/AppSettingsStore.swift
shiki: I didn't know you could do that! Added in be715fe08f5b6f4fd6bd6fe6a8db305405496028. 

Re: `!`. Yeah, I wasn't exactly sure how to avoid this in this class. 😞

Storage/Storage/Model/GeneralAppSettings.swift
shiki: It's intentional. Maybe somebody will respond at some point. 🤣

Yosemite/YosemiteTests/Stores/AppSettingsStoreTests.swift
shiki: So I tried this for a while and it looks like it doesn't matter because from the [documentation of `Date`](https://developer.apple.com/documentation/foundation/date):

> A Date value encapsulate a single point in time, independent of any particular calendrical system or time zone. 

There is no timezone in a `Date` so I could not even create a reliable unit test that simulates different time zones. 😞  

As for the plist file, it looks like the `Date` is saved as UTC:

```xml
<plist version="1.0">
<dict>
	<key>installationDate</key>
	<date>2020-07-29T18:18:01Z</date>
</dict>
</plist>
```

So I think we are safe here. 🙂

shiki:
> I've just one question. It's not clear what is the format of the Date stored in the plist file. Does it maintain the timezone?

@pmusolino The `Date` is saved in UTC even though I was in MDT. So I think we're good here. 🙂 

```xml
<plist version="1.0">
<dict>
	<key>installationDate</key>
	<date>2020-07-29T18:18:01Z</date>
</dict>
</plist>
```

shiki:
Thanks, Paolo and Ernesto!

PR #2557, opened 2020-07-24
Update Fastlane to version 2.152.0

mokagio:
This is a preventive update to avoid issues with Fastlane's `deliver`.

While working on Simplenote iOS 4.21, I experienced a failure updating the localized metadata via `deliver`, which updating Fastlane to the latest version solved.

I didn't spend extra time tracking down the culprit or the exact version fixing the issue because staying relatively up to date with Fastlane doesn't hurt anyways.

Also bumped our Fastlane release plugin to its latest version, 0.9.11, along the way, too.

---

Update release notes:


mokagio:
I milestoned this 4.7 despite it being so close to code freeze, because I think you might run in the same issues as I did with `deliver` so it would be great to get this in.

PR #2565, opened 2020-07-28
Abstract product form view model with `ProductFormViewModelProtocol`

jaclync:
Prerequisite for #2085 

## Changes

- Created `ProductFormViewModelProtocol` that `ProductFormViewModel` conforms to with associated type `ProductVariationFormViewModel` being `Product` (in a future PR, there will be a `ProductVariationFormViewModel` where its `ProductModel` is `ProductVariation`)
- Updated `ProductFormViewController` with generics changes - if a class has generics (`ProductFormViewModelProtocol`'s `ProductModel` in this case), all `@objc` and OjcC protocol conformance (e.g. `UITableViewDelegate`) cannot be in the extension. Also, there cannot be nested enums under a class with generics, so `ActionSheetStrings` and `Constants` were moved to private enums and `ProductFormViewController .ProductFormPresentationStyle` enum was updated to be `ProductFormPresentationStyle`.
- In `ProductFormViewController`, updated its initializer to take in a generic `viewModel: ViewModel` and `ProductImageActionHandler` because the view model depends on it.
- In `ProductDetailsFactory`, updated how `ProductFormViewController` is initialized by passing a view model and `ProductImageActionHandler`.

## Testing

No changes in app behavior are expected, feel free to test editing a product with the following steps:

- Go to the Products tab
- Tap on a simple product
- Tap on each row, potentially edit some fields, open the bottom sheet --> everything should work as before

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -48,31 +49,28 @@ final class ProductFormViewController: UIViewController {
...
-        self.tableViewModel = DefaultProductFormTableViewModel(product: product,
+        self.tableViewModel = DefaultProductFormTableViewModel(product: viewModel.productModel,
                                                                actionsFactory: viewModel.actionsFactory,
                                                                currency: currency)
         self.tableViewDataSource = ProductFormTableViewDataSource(viewModel: tableViewModel,
                                                                   productImageStatuses: productImageActionHandler.productImageStatuses,
                                                                   productUIImageLoader: productUIImageLoader,
                                                                   canEditImages: isEditProductsRelease2Enabled)
-        super.init(nibName: nil, bundle: nil)
+        super.init(nibName: "ProductFormViewController", bundle: nil)
```
Ecarrion, line 73: [Q] Can we use String(describing: self) or it won't work due to generics?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -130,8 +128,173 @@ final class ProductFormViewController: UIViewController {
 
         view.endEditing(true)
     }
+
+    // MARK: - Navigation actions handling
+
+    override func shouldPopOnBackButton() -> Bool {
+        if viewModel.hasUnsavedChanges() {
+            presentBackNavigationActionSheet()
+            return false
```
Ecarrion, line 137: What do you think about using `guard let` to force the return statement and to  be aligned with code added here https://github.com/woocommerce/woocommerce-ios/pull/2565/files#diff-2929b8aeca63d242a89ab243c9f7ddacR167

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -130,8 +128,173 @@ final class ProductFormViewController: UIViewController {
...
+                editGroupedProducts()
+                break
+            case .variations:
+                // TODO-2509 Edit Product M3 analytics
+                guard let product = product as? Product else {
+                    return
+                }
+                guard product.variations.isNotEmpty else {
+                    return
+                }
```
Ecarrion, line 265: Should we merge this into one?
```suggestion
                guard let product = product as? Product, product.variations.isNotEmpty else {
                    return
                }
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -1,14 +1,16 @@
...
+    typealias ProductModel = Product
+
     /// Emits product on change, except when the product name is the only change (`productName` is emitted for this case).
     var observableProduct: Observable<Product> {
         productSubject
     }
 
     /// Emits product name on change.
-    var productName: Observable<String> {
+    var productName: Observable<String>? {
```
Ecarrion, line 13: Are we using RX in the project? 😅

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModelProtocol.swift
```diff
@@ -0,0 +1,81 @@
...
+
+/// A view model for `ProductFormViewController` to add/edit a generic product model (e.g. `Product` or `ProductVariation`).
+///
+protocol ProductFormViewModelProtocol {
+    associatedtype ProductModel: ProductFormDataModel & TaxClassRequestable
+
+    /// Emits product on change, except when the product name is the only change (`productName` is emitted for this case).
+    var observableProduct: Observable<ProductModel> { get }
+
+    /// Emits product name on change. If product name is not editable, `nil` is returned.
```
Ecarrion, line 11: 🤔

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: I can't use `String(describing: self)` since it's not after the self init (error: `'self' used before 'super.init' call`). I did try `type(of: self).nibName` and it didn't work because the type includes the generics like `ProductFormViewController<ProductFormViewModel>` and it crashes the app because it couldn't find a nib with this name that includes generics 😞

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
jaclync: haha not the whole Rx/Combine, but we do have [`Observable` and `PublishSubject` suppport](https://github.com/woocommerce/woocommerce-ios/tree/058d800ee744f75cd6a0c06bb4f1ec98c4514f1e/WooCommerce/Classes/Tools/Observables) that @shiki added before we can use Combine in iOS 13! 💯 (though after playing with Combine for a little, I find RxSwift much easier to use... like returning an Observable from a Subject with `asObservable()`, creating a derived @Published var, and @ObservedObject is just kind of odd. I think I just haven't got the hang of Combine 🤯 )

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: the code link doesn't really show related code on my browser, but I like the idea of using `guard let` to force the return statement. updated here: https://github.com/woocommerce/woocommerce-ios/pull/2565/commits/55a1f13fa8573d4bf66a0359e920e2acaf78d97f

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: yup great suggestion! updated in: https://github.com/woocommerce/woocommerce-ios/pull/2565/commits/55a1f13fa8573d4bf66a0359e920e2acaf78d97f

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModelProtocol.swift
jaclync: sorry it sounds confusing not knowing the differences between editing a `Product` and `ProductVariation`! since `ProductVariation`'s name is derived from its attributes (the option values in the variation), it's not editable like `Product`. I updated the comment here to hopefully clarify a bit: https://github.com/woocommerce/woocommerce-ios/pull/2565/commits/cfb7811f75f987a73e60e94c8aa0636e79c5d5df

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -637,20 +692,8 @@ private extension ProductFormViewController {
...
-        return true
-    }
-
-    override func shouldPopOnSwipeBack() -> Bool {
-        return shouldPopOnBackButton()
-    }
-
-    private func presentBackNavigationActionSheet() {
+private extension ProductFormViewController {
+    func presentBackNavigationActionSheet() {
```
pmusolino, line 696: Do you think it's worth moving this method next to the others (`shouldPopOnBackButton` and `shouldPopOnSwipeBack`)?
Right now we have navigation actions handling in two separate places.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -806,7 +857,7 @@ private extension ProductFormViewController {
         defer {
             navigationController?.popViewController(animated: true)
         }
-        let hasChangedData = newBriefDescription != product.briefDescription
+        let hasChangedData = newBriefDescription != product.shortDescription
```
pmusolino, line 860: nit: rename `newBriefDescription" to `newShortDescription`

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: the functions are usually together but this is a special case, since @objc functions like `shouldPopOnBackButton` and `shouldPopOnSwipeBack` can't be in an extension of a class with generics. I think I'd prefer this staying in an extension rather than in the already crowded main class. lemme know if you have other concerns!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: yup! updated in https://github.com/woocommerce/woocommerce-ios/pull/2565/commits/0c07513322a54454c42e4bf13688164c355c1454

jaclync:
Thanks for the review Ernesto and Paolo! merging now :D

Issue #2382, opened 2020-06-03
Allow Testing of Localized Strings


shiki:
There are times when you would like to test a localized string. For example, you may want to make sure that the quantity and price are the values attached to the final string of this:

```swift
    enum Localization {
        static func subtitle(quantity: String, price: String) -> String {
            let format = NSLocalizedString("%1$@ x %2$@", comment: "")
            return String.localizedStringWithFormat(format, quantity, price)
        }
    }
```

We can technically do this now. But if the tests are running on a non-English simulator, then the test would fail because the `NSLocalizedString()` call would return the localized value. 

## Possible Solutions 

We'd probably have to create our own localization functions. These functions can probably allow overrides so that the tests can _expect_ a specific language to be used. 

We'd probably also need to modify `localize.py` so it will not just look for `NSLocalizedString()` calls. 

### Stretch Goal

It'd probably be better if we can also organize our localized strings into a single file. Kind of like how [SwiftGen does it](https://github.com/SwiftGen/SwiftGen#strings).

jaclync:
I think we don't generally assume the language of the simulator to run unit tests?

Re stretch goal: I think our current way of localizing strings using the same key and value in `Localizable.strings` makes it harder to have an automated helper like the SwiftGen examples. In the SwiftGen example and my previous projects, the key is usually a semantic string in some format like `products_title` so that the codegen helper can make it to like `L10n.productsTitle`. With our localizable strings, I'm not sure how SwiftGen is going to map like `%1$@ - %2$@ (System)` 🤔

shiki:
> I think we don't generally assume the language of the simulator to run unit tests?

@jaclync I'm not sure what you mean by this. Would you mind expanding? 

> I think our current way of localizing strings using the same key and value in Localizable.strings makes it harder to have an automated helper like the SwiftGen examples. 

Yeah, that is a problem. I wonder if that's a GlotPress requirement. 🤔 

How do you feel about having a public `enum Localization` containing **all of our strings**. We'll allow this `enum` to be accessible in the unit tests so the unit tests can use them directly. 

```swift
    enum Localization {
        static func subtitle(quantity: String, price: String) -> String {
            let format = NSLocalizedString("%1$@ x %2$@", comment: "")
            return String.localizedStringWithFormat(format, quantity, price)
        }

        static let justATitle = NSLocalizedString("Products", ...)

        // all other `NSLocalizedString()` calls here
    }
```

```swift
[...]

Issue #2229, opened 2020-05-03
Ignore NSLocalizedString Declarations in Unit Tests


shiki:
Our `localize.py` script currently parses `NSLocalizedString`  declarations in unit tests. These strings are unnecessarily submitted for translation.

joshheald:
Can likely be resolved using [this sort of strategy](https://unix.stackexchange.com/a/31269) to remove WooCommerceTests and other test directories from the find output.

I had some success with 

```
find WooCommerce Pods/WordPress* \( -type d -name WooCommerceTests -prune \) -o \( -type f -name "*.m" -o -name "*.swift" \) | grep -v Vendor
```

Which could be used [in the localize.py script](https://github.com/woocommerce/woocommerce-ios/blob/014a7e9c467c7952fd20ff5c47c00b79fe3b1ac3/Scripts/localize.py#L159)

Replacing

```
find_cmd = 'find ' + target_folders + ' -name "*.m" -o -name "*.swift" | grep -v Vendor'
```

with

```
find_cmd = 'find ' + target_folders + '\( -type d -name '+ exclude_folders ' -prune \) -o \( -type f -name "*.m" -o -name "*.swift" \) | grep -v Vendor'
```

However `-name` does not support multiple folders, so the prune would need to combine the contents of `exclude_folders` for that to work.

JorgeMucientes:
Closing this as part of a maintenance pass on the GitHub issue backlog. This issue matched the auto-close criteria for this pass: it has not been updated since 2022-05-19, is unassigned, is not labeled as a bug or high/medium priority, and does not appear to involve POS, payments, login, data loss, crash, security, vulnerability, auth, or fraud.

If this is still relevant, please comment with current details and we can reopen it or move it into Linear for triage.

PR #2566, opened 2020-07-28
`ProductVariation` now conforms to `ProductFormDataModel` and `TaxClassRequestable` for price settings customization

jaclync:
Prerequisite for #2085 

## Changes

- Updated `ProductVariation` to conform to `ProductFormDataModel` and `TaxClassRequestable` so that it can be passed to `ProductPriceSettingsViewModel`
- In `ProductPriceSettingsViewModel`, updated its sections logic so that `Product` form includes a tax section while `ProductVariation` form does not
  - I didn't really like how it checks the concrete type of `product` to differentiate the types, please let me know if you have other ideas!
- Added `MockProductVariation` to `WooCommerceTests` target for tests
- Created `ProductPriceSettingsViewModel_ProductVariationTests` that tests on the sections logic (no tax section)
- Created `ProductVariation_ProductFormTests` that tests on the derived properties of `ProductVariation`

## Testing

- Go to the Products tab
- Tap on a simple product
- Tap on the price settings row, potentially edit some fields --> everything should work as before

Update release notes:


WooCommerce/WooCommerceTests/Mockups/MockProductVariation.swift
```diff
@@ -0,0 +1,58 @@
...
+                                shippingClass: "",
+                                shippingClassID: 0,
+                                menuOrder: 1)
+
+    }
+
+    private func dateFromGMT(_ dateStringInGMT: String) -> Date {
+        let dateFormatter = DateFormatter.Defaults.dateTimeFormatter
+        return dateFormatter.date(from: dateStringInGMT)!
+    }
```
shiki, line 57: 🤣

WooCommerce/WooCommerceTests/Mockups/MockProductVariation.swift
```diff
@@ -0,0 +1,58 @@
+import Foundation
+@testable import Yosemite
+
+final class MockProductVariation {
+    func productVariation(siteID: Int64 = 2019,
+                          productID: Int64 = 2020,
+                          variationID: Int64 = 2783) -> ProductVariation {
```
shiki, line 7: I have doubts about these arguments and default values (e.g. `"chocolate.com"`). I feel like it's better to just have an empty default object:

```swift
    func productVariation() -> ProductVariation {
        siteID: 0, // numbers are zero
        permalink: "" // required strings are empty strings
```

And then let the unit test define the values using `.copy()`

```swift
let variation = MockProductVariation().productVariation().copy(siteID: 12, productID: 12, variationID: 3, something: "", anotherThing: "")
```

In this way, I think it's more explicit what data the unit test needs. And we can guarantee that no default data in `productVariation()` is being used. It's just a helper so we don't need to write lots of (unnecessary) default arguments. 

This is just a not well thought out idea. Just sharing in case you have some thoughts. 🙂

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductPriceSettingsViewModel+ProductVariationTests.swift
```diff
@@ -0,0 +1,82 @@
...
+            Section(title: nil, rows: [.scheduleSale, .scheduleSaleFrom, .scheduleSaleTo]),
+        ])
+    }
+}
+
+private extension ProductPriceSettingsViewModel_ProductVariationTests {
+    enum Strings {
+        static let priceSectionTitle = NSLocalizedString("Price", comment: "Section header title for product price")
+        static let taxSectionTitle = NSLocalizedString("Tax Settings", comment: "Section header title for product tax settings")
+    }
```
shiki, line 81: This is a bit of a problem, isn't it? 😁 What do you think about grouping all our localized strings into a single (`public`) file? I added that as a possible solution for https://github.com/woocommerce/woocommerce-ios/issues/2382. But maybe you have better ideas.

Also note that translations in unit tests are _captured_ to be translated in Glotpress. 😱 https://github.com/woocommerce/woocommerce-ios/issues/2229

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
```diff
@@ -79,3 +79,17 @@ extension Product: ProductFormDataModel {
         productStockStatus
     }
 }
+
+extension ProductVariation: ProductFormDataModel {
```
shiki, line 83: ⚠️  One concern with the `ProductFormDataModel` `protocol` approach is that it adds additional properties to `ProductVariation` itself. Considering the [multitude of properties](https://github.com/woocommerce/woocommerce-ios/blob/develop/Networking/Networking/Model/Product/ProductVariation.swift) that `ProductVariation` already has, the lines between "which ones are from the API?" and "which ones are owned by us" get blurry. I can see myself getting confused by these properties in the future. 

For example, in this test, I had thought that `isShippingEnabled` is just a regular property. 😅 

https://github.com/woocommerce/woocommerce-ios/blob/d9325a2751567caff3587dcd484fe580ab8e498c/WooCommerce/WooCommerceTests/ViewRelated/Products/ProductVariation%2BProductFormTests.swift#L34-L37

Also, consider the fact that this is an `extension`. This means that I can now access `isShippingEnabled` from **anywhere** in the WooCommerce package. For example, in here:

https://github.com/woocommerce/woocommerce-ios/blob/d9325a2751567caff3587dcd484fe580ab8e498c/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsDataSource.swift#L392-L401

I can add: 

```swift
        let isShippingEnabled = product?.isShippingEnabled
```

I don't think it even makes sense to access that property in `OrderDetailsDataSource`. But I can. 🤷 I believe this adds confusion for us down the road. 

This is why I'm always wary of extensions. And would try to shy away from them if there's another way. 

For this specific case, one idea that comes to mind is to consider encapsulating the `Product` and `ProductVariation` inside their own respective classes instead — specifically for product editing. And then those classes can be joined by a `protocol`. And it is that `protocol` that we will always interact with in the product editing module. 

Here's a rough example:

```swift
protocol EditableProductEntity {
    var description: String? { get }
    var shortDescription: String? { get }
}

final class EditableProduct: EditableProductEntity {
    private let product: Product

    init(product: Product) {
        self.product = product
    }

    var description: String? {
        product.fullDescription
    }

    var shortDescription: String? {
        product.shortDescription
    }
}

final class EditableProductVariation: EditableProductEntity {
    private let variation: ProductVariation

    init(variation: ProductVariation) {
        self.variation = variation
    }

    var description: String? {
        variation.description
    }

    var shortDescription: String? {
        nil
    }
}
```

It's a bit longer, that's true. But I believe it's sometimes worth it to have longer code in order to maintain a **clean namespace**.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
```diff
@@ -142,15 +142,25 @@ final class ProductPriceSettingsViewModel: ProductPriceSettingsViewModelOutput {
         }
         let salesSection = Section(title: nil, rows: saleScheduleRows)
 
-        // Tax section
-        let taxSection = Section(title: NSLocalizedString("Tax Settings", comment: "Section header title for product tax settings"),
+        switch product {
+        case is Product:
+            // Tax section
+            let taxSection: Section
+            taxSection = Section(title: NSLocalizedString("Tax Settings", comment: "Section header title for product tax settings"),
```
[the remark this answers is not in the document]
jaclync: Originally, a product variation also has its tax class editable but it's taken out for now due to an API issue (p91TBi-34H-p2). Now that only a product has the tax section, I moved it to the `Product` case. Lemme know if you think it's easier to read if the tax section is outside of this case!

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
```diff
@@ -142,15 +142,25 @@ final class ProductPriceSettingsViewModel: ProductPriceSettingsViewModelOutput {
         }
         let salesSection = Section(title: nil, rows: saleScheduleRows)
 
-        // Tax section
-        let taxSection = Section(title: NSLocalizedString("Tax Settings", comment: "Section header title for product tax settings"),
+        switch product {
```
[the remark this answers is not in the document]
jaclync: yeaa it sounds confusing. in the view model protocol in https://github.com/woocommerce/woocommerce-ios/pull/2565, I changed `product` in the view model to `productModel`. do you think `productModel` would sound more like a protocol? if so, I'll create a ticket to rename `product: ProductFormDataModel` everywhere after the feature work :)

WooCommerce/WooCommerceTests/Mockups/MockProductVariation.swift
jaclync: This is a good point! The code here is from the existing mock from `Yosemite`, but since I've been using `copy` for fixtures it does make sense to just return an empty object with empty default values. will update!

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductPriceSettingsViewModel+ProductVariationTests.swift
jaclync: TIL'ed the translations in unit tests are also captured in the localization script! in this case, I moved all the localizable strings in `ProductPriceSettingsViewModel` to an enum extension that this test can access: https://github.com/woocommerce/woocommerce-ios/pull/2566/commits/9d121ed4533fb6ee7d23ea56f4a0a0b816f376ae

I also commented on https://github.com/woocommerce/woocommerce-ios/issues/2382 🙂

WooCommerce/WooCommerceTests/Mockups/MockProductVariation.swift
jaclync: updated in: https://github.com/woocommerce/woocommerce-ios/pull/2566/commits/deef6afc0e41e5aba543c82ebb8083450509d2fe

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
jaclync: yea it might be nice to differentiate the helpers vs. API properties. how would you have the `isShippingEnabled` in the `EditableProductEntity` example above? I'm still trying to understand how the encapsulation helps with the situation 🤔 do you think updating the extension helper variables to functions might help? so that the properties are from the API, and functions are helpers / derived values.

personally, I think it's okay for anyone to access the extension helpers in this case - like if the order details might use `isShippingEnabled` to configure its product related UI, that's totally fine. on the other hand, I could see that we might not want certain helpers to be accessible to other classes. I could also move these extensions to where they are used and mark them private, at the cost of removing the unit tests.

leaving this PR open so that I can update based on the feedback 🙂

shiki: > I could also move these extensions to where they are used and mark them private, at the cost of removing the unit tests.

This was the first thing that I tried first. It didn't work for me. It was saying that I can't add `private` extensions to conform to `ProductFormDataModel`, which has an `internal` access level. ¯\\\_(ツ)\_/¯

A possible solution would be having the product editing as a separate module (Swift Package Manager?). But I think we're a long way from that. 

> personally, I think it's okay for anyone to access the extension helpers in this case

Yeah, this might be okay for now since the project is not that big yet. I'm just highlighting this now as I have seen how extensions can be abused if we are not vigilant. WordPress is probably a good example of this:



You'll notice that some of those extensions are not related to each other and are **tightly bound** to a specific feature in the app. The _extensions_ that they add are not even useful anywhere else. 

> how would you have the isShippingEnabled in the EditableProductEntity example above?

It will be like the `description` and `shortDescription`. 

```swift
    var isShippingEnabled: Bool {
        product.downloadable == false && product.virtual == false
    }
```

As I said, there's going to be a lot of code. And I guess I didn't mention the duplication. 😅 

We can hide the duplication by (surprise!) another `protocol`. This time, a stricter one that only defines properties that are truly defined in both `Product` and `Variation`.

image

With this, properties and methods that are only for product editing will still be hidden inside `EditableProductEntity`. Other modules will not be able to use them. 

---

Again, this is just for discussion. 🙂 This may actually be over-engineering! Let's first observe how the app evolves and be vigilant of the effects of extensions.

jaclync: Thanks for the further explanation and diagram! I understand more about the potential issues now. yea, certain extensions methods/variables just don't make sense out of the feature's scope. in this case, I think `var productTaxStatus: ProductTaxStatus` and `var backordersSetting: ProductBackordersSetting` make sense to be accessible within the app layer and can be in an extension of the data models like the common properties protocol. I created a ticket to do this after M3: https://github.com/woocommerce/woocommerce-ios/issues/2575

jaclync:
Thank you always for your insightful feedback, Jayson! It's really helpful knowing what challenges could be ahead when the app grows. I'm merging this PR now and will come back to https://github.com/woocommerce/woocommerce-ios/issues/2575 after M3!

PR #2568, opened 2020-07-28
Unit tests for ProductSettingsRows

### LINKED ISSUE #2526: Unit tests for ProductSettingsRowMediator
It would be good to have tests for `ProductSettingsRowMediator`. As suggested by @shiki and @Ecarrion in [this PR](https://github.com/woocommerce/woocommerce-ios/pull/2516), we can probably test this in isolation. For example, we can test that the `ProductSettings` is updated whenever the `SwitchTableViewCell` emits a change.

Non-compiling pseudocode:
```
    func testWhenCellIsChangedThenReviewsAllowedIsUpdated() {
        // Given
        let settings = ProductSettings()
        let enableReviews = EnableReviews(settings: settings)

        let cell = SwitchTableViewCell()
        enableReviews.configure(cell: cell)

        // When
        cell.onChange(false)

        // Then
        XCTAssertEqual(settings.reviewsAllowed, false)
    }
```

pmusolino:
Fixes #2526 

## Description
I added some unit tests for two specific cases of `ProductSettingsRows`, in particular the one related to VirtualProduct and ReviewsAllowed rows, since only in these two rows we handle the value of settings inside the cell (the value change inside it because we use a `SwitchTableViewCell` that contains a switch). 
The other rows do not seem to need testing, as the updated value always comes from the outside (other view controllers), but if you have any suggestion, feel free to suggest it 😃 

## Testing 
- Look at the code
- CI

Update release notes:

### DIFF
--- a/WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/Settings/ProductSettingsRowsTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/Settings/ProductSettingsRowsTests.swift
@@ -1,0 +1,65 @@
+import XCTest
+@testable import WooCommerce
+import Yosemite
+
+/// Test cases for `ProductSettingsRows`.
+///
+final class ProductSettingsRowsTests: XCTestCase {
+
+    var settings = ProductSettings(status: .draft,
+                                   featured: false,
+                                   password: nil,
+                                   catalogVisibility: .catalog,
+                                   virtual: false,
+                                   reviewsAllowed: false,
+                                   slug: "slug",
+                                   purchaseNote: nil,
+                                   menuOrder: 0)
+
+    override func setUp() {
+        super.setUp()
+
+        // Given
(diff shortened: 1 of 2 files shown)


WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/Settings/ProductSettingsRowsTests.swift
```diff
@@ -0,0 +1,61 @@
+import XCTest
+@testable import WooCommerce
+import Yosemite
+
+/// Test cases for `ProductSettingsRows`.
+///
+final class ProductSettingsRowsTests: XCTestCase {
+
+    var originalSettings: ProductSettings?
```
Ecarrion, line 9: [nit] Should this be private?

pmusolino: Yes, thanks @Ecarrion! Fixed

Issue #2544, opened 2020-07-22
[IAFeedback] Show Crowdsignal Survey


shiki:
The #2357 task should be completed first. 

<img src="https://user-images.githubusercontent.com/198826/88212683-e022a300-cc14-11ea-8741-2de850ebce0d.png" width="240">

Pressing on the **Could Be Better** button should show the **generic** Crowdsignal survey. 

See p91TBi-2Sy-p2 for details. There is a [WIP branch](https://github.com/woocommerce/woocommerce-ios/compare/spike/web-view-survey?expand=1) that we can use as a starting point.

We can use a temporary Crowdsignal survey URL for now. We can replace it later. 

## Reusability

The components we create for the Crowdsignal survey will be reused in #2547. 

## Others

- [ ]  Must be hidden behind the feature flag

Ecarrion:
Hey @Garance91540 
Do we need dark versions for the assets in these screens?
The lady with the tablet and the CrowdSignal icon? 

Thanks

Garance91540:
Hi @Ecarrion, if I understood correctly, the crowdsignal survey will be opened in a webview, so there's no dark mode needed for the "[Feedback form](https://d.pr/i/md1xFg)" screen. 

Once users have selected their answer in the feedback form screen, are they sent back to the app when they select "share feedback" and so the "[feedback received](https://d.pr/i/CjU4LD)" screen is native? I thought it was going to be still shown on a webview, and in this case there's no need for the dark mode view?

Ecarrion:
Hi @Garance91540 

> If I understood correctly, the crowdsignal survey will be opened in a webview, so there's no dark mode needed for the "Feedback form" screen.

That is correct.

> Once users have selected their answer in the feedback form screen, are they sent back to the app when they select "share feedback" and so the "feedback received" screen is native?

Yes, that screen is native. So we do need dark mode versions for the icons on that screen, (Lady and Crodwsignal). Unless you consider that the light mode ones work well for dark mode.

Garance91540:
Hi @Ecarrion , I added the dark mode design for the "Feedback received" screen in Figma and tagged you on it!

Issue #2543, opened 2020-07-22
[IAFeedback] Show App Store Review Dialog


shiki:
The #2357 task should be completed first. 

<img src="https://user-images.githubusercontent.com/198826/88209433-f1b57c00-cc0f-11ea-885f-443f0a7a4e4a.png" width="240">

Pressing on the **I Like It** button should show the App Store Review dialog. 

We will use `SKStoreReviewController`. No UI customizations are needed since the dialog cannot be customized. But please double-check if this is still true.

## Others

- [ ]  Must be hidden behind feature flag

Ecarrion:
It's worth noticing that while our rules to display the in-app feedback banner seem to match the ones expressed on the internet about `SKStoreReviewController`. It's possible that after tapping the "I like button" the app store review prompt doesn't appear to the user. That's why Apple doesn't recommend linking that `requestReview()` action to a particular user input. 

Are we ok with that edge case/possible scenario? cc @shiki @Garance91540 @astralbodies 

ref: https://developer.apple.com/documentation/storekit/skstorereviewcontroller/requesting_app_store_reviews

----

Alternatively, we can redirect the user to the woo's app store page with a "write-review" parameter that is guaranteed to work 100% of the time, but is not an in-app experience.

![app-store](https://user-images.githubusercontent.com/562080/89078649-685a2400-d34a-11ea-8cbe-b61dc3151ac2.gif)

Garance91540:
Hi @Ecarrion , 

>It's possible that after tapping the "I like button" the app store review prompt doesn't appear to the user. 

In that case, is it possible to simply open the WooCommerce Apple page? https://apps.apple.com/us/app/woocommerce/id1389130815

Ecarrion:
Hey @Garance91540,

We can't tell if the in-app review form was displayed(or not) in order to redirect the user to the app store page 😢 
Also, there is a chance that the user has this setting disabled, so no in-app review form will appear ever.

![IMG_E536B31F940E-1](https://user-images.githubusercontent.com/562080/89222450-b06f8580-d59a-11ea-949f-6464bcba63da.jpeg)

Garance91540:
ah that's really annoying! I didn't realize that was even possible...

Could we do:
1. user select "I like it"
2. we open modal "would you mind leaving us a review? yes/no"
3. user selects yes > open https://apps.apple.com/us/app/woocommerce/id1389130815

let me know if this is possible and I'll update the designs accordingly...

[...]

PR #2569, opened 2020-07-28
Add In-app Feedback Card

shiki:
Part of #2537. 

This adds the in-app feedback card in the Today tab for Stats V4. The card currently does not do anything. This is mostly just a UI implementation. This should unblock the other In-app Feedback tasks like #2544 and #2543.  

## Design

Large Font | Light  | Dark
--------|-------|-
Simulator Screen Shot - iPhone 8 (13 6) (Large Font) - 2020-07-28 at 17 37 59        |       Simulator Screen Shot - iPhone 8 - 2020-07-28 at 17 35 16 | Simulator Screen Shot - iPhone 8 (13 6) (Dark) - 2020-07-28 at 17 36 29

We'll ask Garance for a design review when she's back. 

## Changes

The card has to be added as a child of `StoreStatsAndTopPerformersPeriodViewController` so I created it as a `UIViewController`, [`InAppFeedbackCardViewController`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2537-add-in-app-feedback-card/WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift).

The card is only shown when:

- The `inAppFeedback` feature flag is turned on
- If the [`StoreStatsAndTopPerformersPeriodViewController .canDisplayInAppFeedback`](https://github.com/woocommerce/woocommerce-ios/blob/acbefae18c95b9a7d70f9439d159ad73837cdced/WooCommerce/Classes/ViewRelated/Dashboard/Stats%20v4/StoreStatsAndTopPerformersPeriodViewController.swift#L82-L86) is `true`. It's `true` for only the Today tab. 

I'll add unit tests in upcoming PRs since I still need to add another condition for when to display the in-app feedback card. 

## Testing

1. Log in to a store with WC > 4.0
2. Navigate to My Store → Today.
3. Confirm the card is there.
4. Confirm that the card is not visible in the other tabs. 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
@@ -75,10 +75,15 @@ final class StoreStatsAndTopPerformersPeriodViewController: UIViewController {
         return [storeStatsPeriodViewController, topPerformersPeriodViewController]
     }
 
-    init(timeRange: StatsTimeRangeV4, currentDate: Date) {
+    private let featureFlagService: FeatureFlagService
+
+    init(timeRange: StatsTimeRangeV4,
+         currentDate: Date,
+         featureFlagService: FeatureFlagService = ServiceLocator.featureFlagService) {
         self.timeRange = timeRange
         self.granularity = timeRange.intervalGranularity
         self.currentDate = currentDate
+        self.featureFlagService = featureFlagService
         super.init(nibName: nil, bundle: nil)
         configureChildViewControllers()
     }
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
@@ -79,12 +79,20 @@ final class StoreStatsAndTopPerformersPeriodViewController: UIViewController {
 
     private let featureFlagService: FeatureFlagService
 
+    /// If applicable, present the in-app feedback. The in-app feedback may still not be presented
+    /// depending on the constraints. But setting this to `false`, will ensure that it will
+    /// never be presented.
+    ///
+    private let canDisplayInAppFeedback: Bool
+
     init(timeRange: StatsTimeRangeV4,
          currentDate: Date,
+         canDisplayInAppFeedback: Bool,
          featureFlagService: FeatureFlagService = ServiceLocator.featureFlagService) {
         self.timeRange = timeRange
         self.granularity = timeRange.intervalGranularity
         self.currentDate = currentDate
+        self.canDisplayInAppFeedback = canDisplayInAppFeedback
         self.featureFlagService = featureFlagService
         super.init(nibName: nil, bundle: nil)
         configureChildViewControllers()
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
```diff
@@ -65,20 +65,35 @@ class StoreStatsAndTopPerformersPeriodViewController: UIViewController {
...
-        return [storeStatsPeriodViewController, topPerformersPeriodViewController]
+        return [storeStatsPeriodViewController, inAppFeedbackCardViewController, topPerformersPeriodViewController]
     }
 
-    init(timeRange: StatsTimeRangeV4, currentDate: Date) {
+    private let featureFlagService: FeatureFlagService
+
+    /// If applicable, present the in-app feedback. The in-app feedback may still not be presented
+    /// depending on the constraints. But setting this to `false`, will ensure that it will
+    /// never be presented.
```
jaclync, line 84: nit: maybe these comments could be in the public initializer so that it's visible to the caller? feel free to include just `canDisplayInAppFeedback`

WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.xib
```diff
@@ -0,0 +1,66 @@
...
+                    <rect key="frame" x="15" y="16" width="290.5" height="108"/>
+                    <subviews>
+                        <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" text="Enjoying the WooCommerce app?" textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="knB-Iy-ibQ">
+                            <rect key="frame" x="16" y="0.0" width="258.5" height="20.5"/>
+                            <fontDescription key="fontDescription" style="UICTFontTextStyleBody"/>
+                            <nil key="textColor"/>
+                            <nil key="highlightedColor"/>
+                        </label>
+                        <stackView opaque="NO" contentMode="scaleToFill" distribution="fillEqually" spacing="15" translatesAutoresizingMaskIntoConstraints="NO" id="nuf-l1-QC2" userLabel="ButtonsStackView">
```
jaclync, line 33: nit: I was looking at how the buttons are aligned and saw that the configuration is in the xib. do you think these stack view customizations could be in the code? so that if we ever move to SwiftUI, we can map the key parameters right away since the stack view in the xib has so many properties.

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
shiki: Thanks! Added in c006b7900164fc36a2d35cd1324cb6e78a5625aa.

WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.xib
shiki: That makes sense! Thanks for pointing it out. Fixed in c11a1f054c4314a551b25891db52c2f9a40703d7.

shiki:
Thank you, @jaclync!

PR #2570, opened 2020-07-29
Adds SurveyViewController

Ecarrion:
Relates to #2544 

# Why
For the in-app feedback project, we have decided to render the survey in a web view to speed up development times as this module is not a core part of our product.

This PR adapts [part of the work](https://github.com/woocommerce/woocommerce-ios/compare/spike/web-view-survey?expand=1) done by the great @jaclync and introduces `SurveyViewController` into the project, which is configured with a `SurveySource` internal enum.

The idea to have an internal enum rather than exposing a public link is to have total control over what kind of web page/survey we support.

# How

- Added survey URL in `WooConstants.swift`
- Configure view from `SurveySource` enum
- Copy navigation handling work from [spike branch](https://github.com/woocommerce/woocommerce-ios/compare/spike/web-view-survey?expand=1) to know when the survey has been completed.

# GIF
simple-survey

# Testing Steps
- Copy following code in `DashboardViewController`

In `ViewDidLoad`
```swift
let leftBarButton = UIBarButtonItem(barButtonSystemItem: .action, target: self, action: #selector(showSurvey))
navigationItem.leftBarButtonItem = leftBarButton
```

Below `ViewDidLoad` 
```swift
@objc private func showSurvey() {
    let surveyVC = SurveyViewController(survey: .inAppFeedback, onCompletion: { [weak self] in
        self?.dismiss(animated: true, completion: nil)
    })
    let navigationController = WooNavigationController(rootViewController: surveyVC)
    present(navigationController, animated: true, completion: nil)
}
```

- Launch the App
- Tap on the "Share" button on the left bar button item
- Fill the survey
- See that the VC is dismissed after completing the survey.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2572, opened 2020-07-29
Add tests to WooConstant

Ecarrion:
# Why

While working on #2570 I had to add a new URL to `WooConstants. I noticed that we use a function called `trustedURL` to return a non-optional URL. If the URL is malformed the app will crash. While the regular procedure is that we test that the URLs are correctly formed I think there is still a chance(human error) that some error could go unnoticed, especially during PRs with big changes.

To mitigate this I've moved these URLs out to an enum that conforms to `CaseIterable` where we can test that all URLs are correct.

**NOTE:** Feel free to tell me if you think that I may be worrying too much and this is not that necessary.

# Testing Steps
- CI Passes
- Do a quick smoke test in the app to see that the correct URLs are being launched.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -35,7 +35,7 @@ enum WooConstants {
 //
 extension WooConstants {
 
-    /// List of thrusted URLs
+    /// List of trusted URLs
     ///
     enum URLs: String, CaseIterable {
 
@@ -82,14 +82,14 @@ extension WooConstants {
         /// Returns the URL version of the receiver
         ///
         func asURL() -> URL {
-            WooConstants.trustedURL(self.rawValue)
+            Self.trustedURL(self.rawValue)
         }
     }
 }
 
 // MARK: - Utils
 
-private extension WooConstants {
+private extension WooConstants.URLs {
     /// Convert a `string` to a `URL`. Crash if it is malformed.
     ///
     private static func trustedURL(_ url: String) -> URL {
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -87,6 +87,62 @@ enum WooConstants {
     static let systemEventCount = 10
 }
 
+// MARK: URLs
+//
+extension WooConstants {
+
+    /// List of thrusted URls
+    ///
+    enum URLs: String, CaseIterable {
+
+        /// Jetpack Setup URL
+        ///
+        case jetpackSetup = "https://jetpack.com/support/getting-started-with-jetpack/"
+
+        /// Terms of Service Website. Displayed by the Authenticator (when / if needed).
+        ///
+        case termsOfService = "https://wordpress.com/tos/"
+
+        /// Cookie policy URL
+        ///
+        case cookie = "https://automattic.com/cookies/"
+
+        /// Privacy policy URL
+        ///
+        case privacy = "https://automattic.com/privacy/"
+
+        /// Privacy policy for California users URL
+        ///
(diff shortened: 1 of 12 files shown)


WooCommerce/Classes/System/WooConstants.swift
```diff
@@ -21,70 +21,70 @@ enum WooConstants {
...
+        ///
+#if DEBUG
+        case inAppFeedback = "https://wasseryi.survey.fm/woo-mobile-app-test-survey"
+#else
+        case inAppFeedback = "https://automattic.survey.fm/woo-app-general-feedback-user-survey"
+#endif
+
+        /// Returns the URL version of the receiver
+        ///
+        func asURL() -> URL {
```
shiki, line 84: This bugs me a bit because the `enum` name is `URLs` but it's actually a `String` enumeration. But I guess there's no other option, right?

WooCommerce/WooCommerceTests/System/WooConstantsTests.swift
```diff
@@ -0,0 +1,18 @@
...
+
+    func testAllThrustedURLsAreValid() {
+        // Given
+        let allTrustedURLPaths = WooConstants.URLs.allCases
+
+        // When
+        let allTrustedURLs = allTrustedURLPaths.map { $0.asURL() }
+
+        // Then
+        zip(allTrustedURLPaths, allTrustedURLs).forEach { (path, url) in
```
shiki, line 14: A problem when one of the URLs is malformed is that a `fatalError` will happen and **all the unit tests** will be canceled. It's not a good experience for us if it happens. It's quite unlikely though so I think it's fine. 

Or maybe we could change the unit test to do the `URL(string: )` call itself. But I guess that defeats the purpose. 

It's fine. I'm over-analyzing this. 😄

WooCommerce/Classes/System/WooConstants.swift
Ecarrion: I was also struggling with this name but at the end I noticed that making it something more technically close to the raw type made it feel less semantically cohesive in when using it.

I'm open for suggestions tho! :-)

WooCommerce/WooCommerceTests/System/WooConstantsTests.swift
Ecarrion: Yeah, I know this isn't ideal.
One potential solution would be to add some kind of [catchable fatal error](https://medium.com/@marcosantadev/how-to-test-fatalerror-in-swift-e1be9ff11a29) to the project. 

What do you think?

shiki: I'll read that and get back to you. 🙂

WooCommerce/Classes/System/WooConstants.swift
shiki: Yeah, I tried `enum URLs: URL`. It didn't work well. 😄 But I really like the automated unit test! I wonder if we can do something similar for the [`IconsTests`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/WooCommerceTests/Extensions/IconsTests.swift).

WooCommerce/WooCommerceTests/System/WooConstantsTests.swift
shiki: @Ecarrion I just read it and it looks like a neat idea! I'm not so sure about overriding `fatalError()` itself as that may be very surprising for new developers. Maybe we can make our own global function? We'd also have to think about how we can make it reusable across the projects. 

There's an issue that's somewhat related to this: https://github.com/woocommerce/woocommerce-ios/issues/2364

PR #2573, opened 2020-07-29
Add PushNotesManager.inactiveNotifications Observable

shiki:
Ref #2254. 

This will be used as part of #2254. I'm submitting it earlier since the [WIP branch](https://github.com/woocommerce/woocommerce-ios/compare/develop...issue/2254-fix-review-push-navigation?expand=1) is almost 500 lines and I have not even removed the old code yet. 😱 

## Changes

This adds a new `inactiveNotifications` `Observable`, similar to the existing `foregroundNotifications`.

https://github.com/woocommerce/woocommerce-ios/blob/98f4369f151c3154eb282b88c79870312491e609/WooCommerce/Classes/Notifications/PushNotificationsManager.swift#L26-L33

## Testing for Regression

Please test that you can still receive new order push notifications and tapping on them will open the order details.

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ServiceLocator/PushNotesManager.swift
+++ b/WooCommerce/Classes/ServiceLocator/PushNotesManager.swift
@@ -9,6 +9,11 @@ protocol PushNotesManager {
     ///
     var foregroundNotifications: Observable<PushNotification> { get }
 
+    /// An observable that emits values when the app is activated due to a Remote Notification.
+    /// The Remote Notification is emitted.
+    ///
+    var inactiveNotifications: Observable<PushNotification> { get }
+
     /// Resets the Badge Count.
     ///
     func resetBadgeCount(type: Note.Kind)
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ServiceLocator/PushNotesManager.swift
```diff
@@ -9,6 +9,10 @@ protocol PushNotesManager {
     ///
     var foregroundNotifications: Observable<PushNotification> { get }
 
+    /// An observable that emits values when the app is activated due to a Remote Notification.
```
jaclync, line 12: nit: shall the comment mention the "inactive" state?

shiki: Thanks for pointing that out. I changed the documentation to match `foregroundNotifications`. I hope that's clearer! 6dd59203e33a48f1ecd4b86da2691cfc3405977d

PR #2578, opened 2020-07-30
Create a product form view model for a product variation 

jaclync:
Product form view model for a variation - part of #2085

- [x] ⚠️ Make sure https://github.com/woocommerce/woocommerce-ios/pull/2577 is merged first

## Changes

Note: sorry about the PR size, this PR only contains one class (~215 diffs) and the rest of the code are on unit tests with a mock stores manager.

- Created `ProductVariationFormViewModel` that conforms to `ProductFormViewModelProtocol` with unit tests
  - Unfortunately, the difference in the editable fields between `Product` and `ProductVariation` results in a few no-op functions and variables in `ProductVariationFormViewModel`. I'm open to any other ideas to make the form work for both product models that allow customizations in the actions!

## Testing

Since the changes don't affect the app behavior yet, just CI!

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
@@ -1,0 +1,215 @@
+import Yosemite
+
+/// Provides data for product form UI on a `ProductVariation`, and handles product editing actions.
+final class ProductVariationFormViewModel: ProductFormViewModelProtocol {
+    typealias ProductModel = ProductVariation
+
+    /// Emits product variation on change.
+    var observableProduct: Observable<ProductVariation> {
+        productVariationSubject
+    }
+
+    /// The latest product variation including potential edits.
+    var productModel: ProductVariation {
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
```diff
@@ -0,0 +1,215 @@
...
+        // no-op
+        return false
+    }
+}
+
+// MARK: Action handling
+//
+extension ProductVariationFormViewModel {
+    func updateName(_ name: String) {
+        // no-op
```
pmusolino, line 109: I would add that, explicitly, we don't want to set the name because a product variation doesn't have it, and the same for all the fields that we don't need to update. 👍

jaclync: sounds good! updated here: https://github.com/woocommerce/woocommerce-ios/pull/2578/commits/7afaed7b2ad39821c4c5714e78d677058f7c0e35

Issue #1985, opened 2020-03-17
Edit Products: extract view data logic from view controller to a separate view model


jaclync:
With the introduction of discard changes action sheet, the logic becomes more complicated especially when fixing price and inventory settings in https://github.com/woocommerce/woocommerce-ios/pull/1979. I intended to add unit tests in that PR, but it wasn't suitable for a release hot fix. Let's spend a little time extracting the view data logic from the view controller to a separate view model for the following screens just to start:

- [x] `ProductPriceSettingsViewController`
- [x] `ProductFormViewController`: thread https://github.com/woocommerce/woocommerce-ios/pull/1977#discussion_r393014680
- [x] `ProductInventorySettingsViewController` https://github.com/woocommerce/woocommerce-ios/pull/2602
- [ ] `ProductShippingSettingsViewController` https://github.com/woocommerce/woocommerce-ios/pull/2714

shiki:
I also would not be opposed to incrementally extracting the complex parts of the code to **Use Cases** / **Interactors** like this [Domain Layer example](https://tech.olx.com/clean-architecture-and-mvvm-on-ios-c9d167d9f5b3#3dc5) so they can be individually tested.

jaclync:
Moving this to M2 since this doesn't block M1

PR #2579, opened 2020-07-30
Update product inventory settings to take in `ProductFormDataModel` protocol to allow `Product` and `ProductVariation`

jaclync:
Inventory settings for a variation - part of #2085

## Changes

- Updated `ProductInventorySettingsViewController` to take in `ProductFormDataModel` protocol instead of concrete `Product`, so that it works for both `Product` and `ProductVariation`
  - Since `soldIndividually` is only available/editable in `Product`, I had to switch on the product model's type for this field in some instances. If you have better ideas on handling this difference, please lemme know!
  - It'd be nice if this view controller has a view model to be unit tested, but that would involve a bunch of other changes. I reopened this issue https://github.com/woocommerce/woocommerce-ios/issues/1985 to track this since these screens support two data models now.
- Updated `soldIndividually` in `ProductInventoryEditableData` to be optional, and its initializer now takes in `ProductFormDataModel` protocol instead of concrete `Product`
- In `ProductFormViewController`, removed TODOs and `Product` checks for editing inventory settings to allow a product variation

## Testing

No app behavior changes are expected, please sanity check the affected part:

- Go to the Products tab
- Tap on a simple product
- Tap on the inventory row --> make sure the "Limit one per order" (`soldIndividually`) is shown as expected
- Edit the "Limit one per order" and optionally other inventory fields
- Tap "Done"
- Tap "Update" --> the inventory settings should be updated remotely

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventoryEditableData.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventoryEditableData.swift
@@ -1,34 +1,23 @@
 import Yosemite
 
-/// Contains editable properties of a Product in the inventory settings.
+/// Contains editable properties of a product model in the inventory settings.
 ///
-struct ProductInventoryEditableData {
+struct ProductInventoryEditableData: Equatable {
     let sku: String?
     let manageStock: Bool
-    let soldIndividually: Bool
+    let soldIndividually: Bool?
     let stockQuantity: Int64?
     let backordersSetting: ProductBackordersSetting?
     let stockStatus: ProductStockStatus?
 }
 
 extension ProductInventoryEditableData {
-    init(product: Product) {
-        self.sku = product.sku
-        self.manageStock = product.manageStock
-        self.soldIndividually = product.soldIndividually
-        self.stockQuantity = product.stockQuantity
-        self.backordersSetting = product.backordersSetting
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventoryEditableData.swift
```diff
@@ -1,34 +1,23 @@
 import Yosemite
 
-/// Contains editable properties of a Product in the inventory settings.
+/// Contains editable properties of a product model in the inventory settings.
 ///
-struct ProductInventoryEditableData {
+struct ProductInventoryEditableData: Equatable {
```
pmusolino, line 5: Is the `Equatable` conformance missing, or I missed something?

jaclync: if not providing a custom `Equatable` implementation and all the properties already conform to `Equatable`, it will generate one internally based on all the properties! since our original implementation is just to check all the properties, it'd be safer for the struct to generate its own in case we add another property and forget to update the `Equatable` implementation 🙂

jaclync:
Thanks for the review Paolo, I responded to your question! Merging this now, the optional CI tasks are blocking the check but all required tasks have completed (Hound, build tests, run tests):

Screen Shot 2020-07-31 at 9 24 24 AM

Issue #2583, opened 2020-07-30
Stats: This Week stats for top performers is wrong/doesn't update


rachelmcr:
**Describe the bug**

The "Top Performers" section in stats is correct (matching the data in wc-admin) for Today, This Month, and This Year. However, the "This Week" tab is wrong and doesn't update (even when I pull to refresh or force-close and relaunch the app).

**To Reproduce**

0. Make sure you have some orders this week on your store.
1. On the web, go to your store's WooCommerce > Dashboard (wc-admin).
2. Under "Date Range" select "Week to Date" and then "Update." (This will take you to `/wp-admin/admin.php?page=wc-admin&period=week&compare=previous_year` for your store.)
3. Notice the data displayed under "Top Products - Items Sold" on that page.
4. Open the My store tab in the app.
5. Select the "This Week" tab.
6. Scroll down to Top Performers.
7. Compare the data to the data on the web. Notice it doesn't match and doesn't update in the app.

**Screenshots**

This week's data in wc-admin:

![Screenshot 2020-07-30 13 44 54](https://user-images.githubusercontent.com/8658164/88925319-45594280-d26c-11ea-883a-d9932e79cd0f.png)

I checked the app on two devices. They both showed the wrong data (only on this tab) and were different from each other. I had been actively using the app on my iPad before noticing this issue, while I only opened the app on my iPhone after I saw it on my iPad:

WCiOS on iPad Mini 4|WCiOS on iPhone XS
-|-
![IMG_0473](https://user-images.githubusercontent.com/8658164/88925324-47230600-d26c-11ea-911a-c696dea3f2a6.PNG)|![IMG_3858](https://user-images.githubusercontent.com/8658164/88925329-48ecc980-d26c-11ea-8501-82b110573b08.PNG)

**Mobile Environment**
Please include:
- Device: iPhone XS & iPad Mini 4
- iOS version: 13.5.1 (iPhone) & 14.0 beta (iPad)
- WooCommerce iOS version: 4.4.0-beta.1

Ecarrion:
I'll take a look at this!

Ecarrion:
I haven't been able to reproduce this issue 😢 
Using the same store and setting my phone to UK's time and region I'm seeing this.

### Web 
<img width="898" alt="WebDashboard" src="https://user-images.githubusercontent.com/562080/88959893-543df600-d268-11ea-83b5-503960312c8b.png"> 

### App 
[...]

PR #2585, opened 2020-07-30
Fix wrong parameter in leaderboards remote when requesting weekly leaderboards

Ecarrion:
# Why
[An error was reported](https://github.com/woocommerce/woocommerce-ios/issues/2583) where stats were not being properly rendered on the weekly leaderboard in contrast with that is shown on the web dashboard.

I couldn't reproduce the reported bug but after comparing the URLs requested by the web-dashboard and by the WC-app I noticed that we were sending the wrong "unit" parameter. We were sending `day` instead of `week`.

# How
- Added a new `leaderboardsGranularity` method to return the correct `StatsGranularityV4` for the leaderboards API.

# Testing Steps
- Make sure you have some orders this week at your store.
- On the web, go to your store's WooCommerce > Dashboard (wc-admin).
- Under "Date Range" select "Week to Date" and then "Update." 
- Notice the data displayed under "Top Products - Items Sold" on that page.
- Open the My store tab in the app.
- Select the "This Week" tab.
- Scroll down to Top Performers.
- Compare the data to the data on the web.

Note: The order of the products may not be the same as we sort them by total spend on the app and web sorts them by item count.

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Model/Enums/StatsTimeRangeV4.swift
+++ b/Yosemite/Yosemite/Model/Enums/StatsTimeRangeV4.swift
@@ -51,6 +51,20 @@ extension StatsTimeRangeV4 {
         }
     }
 
+    /// Represents the period unit of the leaderboards v4 API  given a time range.
+    public var leaderboardsGranularity: StatsGranularityV4 {
+        switch self {
+        case .today:
+            return .daily
+        case .thisWeek:
+            return .weekly
+        case .thisMonth:
+            return .monthly
+        case .thisYear:
+            return .yearly
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


shiki:
I'll let @jaclync or @pmusolino review this since they're more knowledgeable about this. 🙂

Ecarrion:
Thanks for the insightful comment @jaclync.

Regarding the actual change, It may not fix the problem but I think it does fix the parameter value. Do you think is not worth merging?

----

Regarding the sorting situation, it was quickly discussed here C6H8C3G23-slack-p1595511065103300 and it was decided to keep the existing order, but I'm happy to make the change if @Garance91540 agrees.

PR #2587, opened 2020-07-31
Attempt to fix crash after leaving WP media picker when the second page of images just returns

jaclync:
Attempted fix for #2581 

- [x] ⚠️ Make sure the media picker pod is pointing to a production version before merging ⚠️ 

## Cause of the crash

After I was able to reproduce the crash, I looked at the memory debugger to see why `WPMediaPickerViewController`'s collection view is still trying to refresh data after the media picker has been dismissed. At the time of crash, `WordPressMediaLibraryImagePickerViewController` and `WordPressMediaLibraryPickerDataSource` have both been deallocated, but `WPMediaPickerViewController` is still lingering around because its collection view's `performBatchUpdates` completion block is holding on to it.

## Changes

- Please review the weak self changes in MediaPicker-iOS: https://github.com/wordpress-mobile/MediaPicker-iOS/pull/357 - updated to use weak self in the batch updates completion block
- In `WordPressMediaLibraryPickerDataSource`, guarded when `self` is nil (these aren't related to fixing the crash)

## Testing

Prerequisite: please test on a device type where the WP media picker has the first page of images **partially** filling the screen (e.g. iPhone 11 simulator), so that the second page of images is requested

On `develop`, try reproducing the crash first:

- Go to the Products tab
- Tap on a simple product
- Tap on the "+" cell in the images header
- Tap "Add Photos"
- Tap "WordPress Media Library" from the action sheet, and make sure to do the following step quickly!
- After the first page of photos is loaded and **before or right when the second page comes in**, quickly select an image and tap "Add 1" in the navigation bar --> the app crashes shortly with a similar stack trace as [on Sentry](https://sentry.io/organizations/a8c/issues/1723508991/?project=1458804&query=is%3Aunresolved&statsPeriod=14d)

I was able to reproduce the crash consistently. Please lemme know if you don't see the crash!

Then on the PR branch, the above steps result in no crash and the selected image(s) is added to the product. Feel free to try a few times with slightly different timings when the second page comes back.

Update release notes:


jaclync:
The PR is ready for a quick review now that the pod has the attempted fix now!

Btw @Ecarrion here's the doc on how to release a beta pod version for developers if you haven't seen this: paNNhX-33-p2 (it can also be found on Apps p2 -> Resources in the side bar -> iOS – Cocoapods)

Issue #2372, opened 2020-06-01
Enable crash logging in `Networking` layer


jaclync:
Now that we are enabling crash logging for a top crash in `Storage` layer https://github.com/woocommerce/woocommerce-ios/issues/2364, let's also add it to the `Networking` layer. Maybe we can start logging where we have `DDLogError` in `Networking` (like when it cannot map the response in `Remote`) @woocommerce/ios-developers ?

Also, since `Networking` and `Storage` are independent frameworks, I was wondering if it's worth adding a base framework where some shared protocols/helpers can be in without any dependencies?

shiki:
While reviewing #2366, I thought about using `DDLogError` instead of `SentryCrashLogger` directly. Are you proposing we **only use** `DDLogError` for both `Storage` and `Networking`? I feel like we should merge these logging APIs into one. I'm just not sure which one is better. 🤔 

What do you think? 

> Maybe we can start logging where we have DDLogError in Networking (like when it cannot map the response in Remote)

That makes sense to me. 

> I was wondering if it's worth adding a base framework where some shared protocols/helpers can be in without any dependencies?

Well, this has always been the plan. I just haven't gotten to it yet. 😄 Feel free to take it though!

jaclync:
> While reviewing #2366, I thought about using `DDLogError` instead of `SentryCrashLogger` directly. Are you proposing we **only use** `DDLogError` for both `Storage` and `Networking`? I feel like we should merge these logging APIs into one. I'm just not sure which one is better. 🤔

From my understanding, `DDLogError` is only logged to the application log that is included in a support ticket, is it correct? For logging in https://github.com/woocommerce/woocommerce-ios/pull/2366, I think logging to Sentry would be more helpful for getting error information from all the users (if they opted in) than `DDLogError`.

Maybe we could define a few levels of logging like below, feel free to edit directly or suggest edits @woocommerce/ios-developers !

severity level | where events are sent | details
-- | -- | --
fatal (since the debug message in `fatalError` isn't recorded) | Sentry | when the app cannot recover from the current state and should terminate
non-fatal error | Sentry | when the app encounters an error that deserves our attention
error | `DDLogError` | when the app encounters an error that could be helpful when they submit a support ticket
warning | `DDLogWarn` | when the app receives a warning that could be helpful when they submit a support ticket
info (higher priority) | Sentry | when the app is in a scenario that isn't a warning nor an error but deserves our attention
[...]

PR #2588, opened 2020-07-31
Product variations are now editable - description, price, shipping, and inventory

jaclync:
Variation form - part of #2085 

This PR includes various changes and fixes to make a product variation editable on the basic fields - description, price, shipping, and inventory.

## Changes

Please see the differences between a product and variation form in the description of https://github.com/woocommerce/woocommerce-ios/issues/2085, this PR includes all the changes except for the product images!

- Networking layer - a few fixes were made to `ProductVariation` to allow editing inventory and saving a product variation without an image:
  - Updated the logic on how `manageStock` is decoded from the API response: there are two states, whether a variation has its stock management enabled at variation level (`manage_stock` is `true` in the API) or it follows its parent product's stock status (`manage_stock` is `"parent"` in the API). In the past, we've been considered the latter case as the stock management is enabled at variation level but it's not the case. Therefore, the logic and 2 test cases were added on these 2 states, and `manageStock` was updated in a few unit tests that had the incorrect assumption.
  - It now only encodes its `image` field only when the variation has an image. This is due to an [API issue](https://github.com/woocommerce/woocommerce/issues/27116), and if the `image` field isn't set to a non-nil product image the update request will fail (e.g. when updating a variation without an image).
  - `stockStatusKey` is not encoded anymore. Since stock status is only available when a variation's stock management is set to follow its parent product, it's a readonly value from the parent product and shouldn't be encoded.
- App layer:
  - In `ProductVariationsViewController`, navigated to `ProductFormViewController` with a product variation form view model when the user taps on a variation in the list if M3 flag is enabled. Also, `syncingCoordinator.synchronizeFirstPage()` is now only called once in `viewDidLoad` instead of every time in `viewWillAppear`. Updated `ProductDetailsViewModel` to provide `isEditProductsRelease3Enabled` for `ProductVariationsViewController`.
  - In `ProductShippingSettingsViewController`, the shipping class row is hidden for a variation due to an [API issue](https://github.com/woocommerce/woocommerce/issues/27131) (more details in p91TBi-33Q-p2).
  - In `ProductFormViewModelProtocol`, added a new method `func canEditProductSettings() -> Bool` so that the ellipsis menu that shows the product settings in the navigation bar can be hidden for a variation. `ProductFormViewModel` returns `true` for this method, while `ProductVariationFormViewModel` returns `false`. The value of this method is used to determine whether to show the ellipsis menu in the navigation bar in `ProductFormViewController`.
  - In `ProductInventorySettingsViewController`, updated the stock status row to be readonly for a variation: disabled tap action for a variation, and no disclosure indicator is shown in the cell.
  - In `PrimaryFieldRow.name`, added another parameter `isEditable` to be `PrimaryFieldRow.name(name: String?, isEditable: Bool)` because the name field for a variation is readonly. Updated a bunch of related files, and set the value to `product is Product` in `DefaultProductFormTableViewModel`. If the name row is editable, `TextFieldTableViewCell` is used. Otherwise, `BasicTableViewCell` is used with headline style set in `ProductFormTableViewDataSource`.

## Testing

- Go to the Products tab
- Tap on a variable product with non zero variations --> sanity check that the name field is editable
- Tap on the variations row
- Tap on a variation --> a variation form should be presented, without an ellipsis menu in the navigation bar, and the name row is readonly
- Tap on the description row and edit the text
- Go back to the form and tap on the price row, then edit some fields --> no tax section should be shown on price settings
- Go back to the form and tap on the shipping row, then edit some fields --> the shipping class row should not be shown on the shipping settings
- Go back to the form and tap on the inventory row, then toggle the "Manage stock" switch and edit some fields --> the "Limit one per order" toggle row shouldn't be shown as in a product form, and the stock status row is readonly when "Manage stock" is disabled
- Go back to the form and tap "Update" --> the variation should be updated remotely with the changes above

## Example screenshots

simulator

Dark | Light
-- | --
Simulator Screen Shot - iPhone 11 - 2020-07-31 at 17 59 53 | Simulator Screen Shot - iPhone 11 - 2020-07-31 at 18 00 31

Price | Shipping | Inventory (stock management disabled)
-- | -- | --
Simulator Screen Shot - iPhone 11 - 2020-07-31 at 18 00 00 | Simulator Screen Shot - iPhone 11 - 2020-07-31 at 18 00 08 | Simulator Screen Shot - iPhone 11 - 2020-07-31 at 18 00 12

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/ProductVariation.swift
+++ b/Networking/Networking/Model/Product/ProductVariation.swift
@@ -196,7 +196,9 @@ public struct ProductVariation: Codable, GeneratedCopiable, Equatable {
                                                             alternativeTypes: [
                                                                 .string(transform: { value in
                                                                     guard value.lowercased() == Values.manageStockParent else {
-                                                                        assertionFailure("Unexpected manage stock value: \(value)")
+                                                                        let message = "Unexpected manage stock value: \(value)"
+                                                                        assertionFailure(message)
+                                                                        DDLogError(message)
                                                                         return false
                                                                     }
                                                                     return false
--- a/Networking/Networking/Model/Product/ProductVariation.swift
+++ b/Networking/Networking/Model/Product/ProductVariation.swift
@@ -252,7 +252,11 @@ public struct ProductVariation: Codable, GeneratedCopiable, Equatable {
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
 
-        try container.encode(image, forKey: .image)
+        // TODO-2576: When the image removal is feasible in the API, let's remove this condition.
+        // Ref: https://github.com/woocommerce/woocommerce/issues/27116
+        if let image = image {
+            try container.encode(image, forKey: .image)
+        }
 
         try container.encode(description, forKey: .description)
 
--- a/Networking/Networking/Model/Product/ProductVariation.swift
+++ b/Networking/Networking/Model/Product/ProductVariation.swift
@@ -188,14 +188,18 @@ public struct ProductVariation: Codable, GeneratedCopiable, Equatable {
 
         // Even though the API docs claim `manageStock` is a bool, it's possible that `"parent"`
         // could be returned as well (typically with variations) — we need to account for this.
-        // A "parent" value means that stock mgmt is turned on + managed at the parent product-level, therefore
-        // we need to set this var as `true` in this situation.
+        // A "parent" value means that stock mgmt is turned off at the product variation and it is managed at the parent product-level.
+        // Therefore, we need to set this var as `false` in this situation.
         // See: https://github.com/woocommerce/woocommerce-ios/issues/884 for more deets
         let manageStock = container.failsafeDecodeIfPresent(targetType: Bool.self,
                                                             forKey: .manageStock,
                                                             alternativeTypes: [
                                                                 .string(transform: { value in
-                                                                    value.lowercased() == Values.manageStockParent ? true : false
+                                                                    guard value.lowercased() == Values.manageStockParent else {
+                                                                        assertionFailure("Unexpected manage stock value: \(value)")
+                                                                        return false
+                                                                    }
+                                                                    return false
                                                                 })
         ]) ?? false
 
@@ -293,7 +297,6 @@ public struct ProductVariation: Codable, GeneratedCopiable, Equatable {
         try container.encode(manageStock, forKey: .manageStock)
         try container.encode(stockQuantity, forKey: .stockQuantity)
         try container.encode(backordersKey, forKey: .backordersKey)
-        try container.encode(stockStatus.rawValue, forKey: .stockStatusKey)
     }
 }
 
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
@@ -48,6 +48,8 @@ private extension DefaultProductFormTableViewModel {
         switch product {
         case let product as Product:
             return settingsRows(product: product, actions: actions)
+        case let product as ProductVariation:
+            return settingsRows(productVariation: product, actions: actions)
         default:
             fatalError("Unexpected product form data model: \(type(of: product))")
         }
@@ -76,6 +78,21 @@ private extension DefaultProductFormTableViewModel {
                 return .groupedProducts(viewModel: groupedProductsRow(product: product))
             case .variations:
                 return .variations(viewModel: variationsRow(product: product))
+            default:
+                fatalError("Unexpected action in the settings section: \(action)")
+            }
+        }
+    }
+
+    func settingsRows(productVariation: ProductVariation, actions: [ProductFormEditAction]) -> [ProductFormSection.SettingsRow] {
+        return actions.map { action in
+            switch action {
+            case .priceSettings:
+                return .price(viewModel: priceSettingsRow(product: productVariation))
+            case .shippingSettings:
+                return .shipping(viewModel: shippingSettingsRow(product: productVariation))
+            case .inventorySettings:
+                return .inventory(viewModel: inventorySettingsRow(product: productVariation))
             default:
                 fatalError("Unexpected action in the settings section: \(action)")
             }
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
@@ -35,7 +35,7 @@ private extension DefaultProductFormTableViewModel {
             case .images:
                 return .images
             case .name:
-                return .name(name: product.name)
+                return .name(name: product.name, isEditable: product is Product)
             case .description:
                 return .description(description: product.trimmedFullDescription)
             default:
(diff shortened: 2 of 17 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventorySettingsViewController.swift
```diff
@@ -379,6 +379,10 @@ extension ProductInventorySettingsViewController: UITableViewDelegate {
 
         switch rowAtIndexPath(indexPath) {
         case .stockStatus:
+            // Stock status is only editable for `Product`.
+            guard product is Product else {
+                return
+            }
```
Ecarrion, line 385: I'm a bit concern about these conditions because they seem very hard to spot and maintain as we introduce more product types(not sure if that is possible).

Do you think it's with it to group them in a testable/rehusable type?
something like `isStockEnabled(for product: ProductFormData)`

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewModel.swift
```diff
@@ -75,8 +75,8 @@ extension ProductFormSection.PrimaryFieldRow: Equatable {
         switch (lhs, rhs) {
         case (.images, .images):
             return true
-        case (let .name(name1), let .name(name2)):
-            return name1 == name2
+        case (let .name(name1, isEditable1), let .name(name2, isEditable2)):
+            return name1 == name2 && isEditable1 == isEditable2
```
Ecarrion, line 79: I think we could rely on the synthesized Equatable conformance, right?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Shipping Settings/ProductShippingSettingsViewController.swift
```diff
@@ -46,6 +43,21 @@ final class ProductShippingSettingsViewController: UIViewController {
...
+        switch product {
+        case is Product:
+            sections = [
+                Section(rows: [.weight, .length, .width, .height]),
+                Section(rows: [.shippingClass])
+            ]
+        case is ProductVariation:
+            sections = [
+                Section(rows: [.weight, .length, .width, .height])
+            ]
```
Ecarrion, line 56: Should we extract this into a testable helper/function? or do you think it's not necessary?

Networking/Networking/Model/Product/ProductVariation.swift
```diff
@@ -188,14 +188,18 @@ public struct ProductVariation: Codable, GeneratedCopiable, Equatable {
...
+        // A "parent" value means that stock mgmt is turned off at the product variation and it is managed at the parent product-level.
+        // Therefore, we need to set this var as `false` in this situation.
         // See: https://github.com/woocommerce/woocommerce-ios/issues/884 for more deets
         let manageStock = container.failsafeDecodeIfPresent(targetType: Bool.self,
                                                             forKey: .manageStock,
                                                             alternativeTypes: [
                                                                 .string(transform: { value in
-                                                                    value.lowercased() == Values.manageStockParent ? true : false
+                                                                    guard value.lowercased() == Values.manageStockParent else {
+                                                                        assertionFailure("Unexpected manage stock value: \(value)")
```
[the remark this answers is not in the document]
jaclync: Right now we don't have a way to log events on Sentry in the Networking layer yet since it depends on the Tracks dependency (it's a TODO https://github.com/woocommerce/woocommerce-ios/issues/2372), but lemme update this to a `DDLogError` so at least it can be logged when a user writes to support (it's recorded in the application log)

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -81,6 +83,21 @@ private extension DefaultProductFormTableViewModel {
...
+        return actions.map { action in
+            switch action {
+            case .priceSettings:
+                return .price(viewModel: priceSettingsRow(product: productVariation))
+            case .shippingSettings:
+                return .shipping(viewModel: shippingSettingsRow(product: productVariation))
+            case .inventorySettings:
+                return .inventory(viewModel: inventorySettingsRow(product: productVariation))
+            default:
+                fatalError("Unexpected action in the settings section: \(action)")
```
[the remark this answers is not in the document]
jaclync: good question! I thought about it more, I'll map it to `nil` with `assertionFailure` with `compactMap` here

WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventorySettingsViewController.swift
jaclync: yea this is not ideal to have this logic without tests. I originally meant to extract some view-level logic to a view model but haven't got to: https://github.com/woocommerce/woocommerce-ios/issues/1985 - I'll make sure to work on this for M3!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewModel.swift
jaclync: oh wow, I didn't realize that we don't need the custom `Equatable` implementation for an enum with associated values anymore! I couldn't find the reference where it's fixed in a Swift version, but will update to just have the form enums conform to `Equatable`

WooCommerce/Classes/ViewRelated/Products/Edit Product/Shipping Settings/ProductShippingSettingsViewController.swift
jaclync: Similar to [this comment](https://github.com/woocommerce/woocommerce-ios/pull/2588#discussion_r464170659), I'll make sure to extract the view logic to a view model to have unit tests

Networking/Networking/Model/Product/ProductVariation.swift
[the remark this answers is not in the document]
jaclync: Updated here: https://github.com/woocommerce/woocommerce-ios/pull/2588/commits/e1c0277ca3fdd3f048bfe1ae73e72652cd296a23

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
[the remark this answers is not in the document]
jaclync: Updated here: https://github.com/woocommerce/woocommerce-ios/pull/2588/commits/d43634f664750d34ae5f618161c94f924f6dbd57

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewDataSource.swift
```diff
@@ -118,21 +118,31 @@ private extension ProductFormTableViewDataSource {
...
-            self?.onNameChange?(newName)
-            }, onTextDidBeginEditing: {
-                ServiceLocator.analytics.track(.productDetailViewProductNameTapped)
-        }, inputFormatter: nil, keyboardType: .default)
-        cell.configure(viewModel: viewModel)
+            cell.accessoryType = .none
+            cell.textLabel?.text = name
+            cell.textLabel?.applyHeadlineStyle()
+            cell.textLabel?.textColor = .text
+        }
```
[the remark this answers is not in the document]
jaclync: yup! updated here: https://github.com/woocommerce/woocommerce-ios/pull/2588/commits/b5b62549f8de092634be55bcbaa44c498fe40e2f

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewModel.swift
jaclync: updated here: https://github.com/woocommerce/woocommerce-ios/pull/2588/commits/31be2f9813645ff6329bbc19d157a1451ab43de9

jaclync:
Thanks for the review/testing/suggestions Ernesto! I responded to your comments, merging now for another upcoming PR.

Yea it's a bit overwhelming having subtle differences in a product vs. variation form in several screens. writing it down and making a checklist in https://github.com/woocommerce/woocommerce-ios/issues/2085 helps 😄

PR #2593, opened 2020-08-04
Enable/disable a product variation

### LINKED ISSUE #2511: Hide a variant product in Store
As part of Milestone 3, we want to give to the user the possibility to hide a variant of a product from the store.

- Whether the product variation is visible: `status` field is `private` or `publish`

This the option on the web
Schermata 2020-07-15 alle 15 35 06

This is what we expect to see on mobile:
Schermata 2020-07-15 alle 15 34 38

jaclync:
Fixes #2511 

## Changes

- In the Networking layer, encoded `status` for `ProductVariation` 
- Added `UIImage.visibilityImage` using Gridicons' `visible` icon with a test case
- Added a case `status` to `ProductFormEditAction` for editing a variation's status
- In `ProductVariationFormActionsFactory`, always showed `status` action below the price settings
  - Updated unit tests to include this row
- In `ImageAndTitleAndTextTableViewCell`, added a different type of view model `SwitchableViewModel` to configure the new cell type with a switch toggle in the cell's accessory view
- In `DefaultProductFormTableViewModel`, configured the status row view model for a `ProductVariation`
- Added `status: ProductStatus` to `ProductFormDataModel` protocol, and return the value for a `Product`
- In `ProductFormTableViewDataSource`, added an action to handle status changes and configured the status row with `SwitchableViewModel`
- In `ProductFormTableViewModel`'s `ProductFormSection.SettingsRow`, added the corresponding `status` row that takes in a `SwitchableViewModel`
- In `ProductFormViewController`, configured the status action handling
- Added `func updateStatus(_ isEnabled: Bool)` to `ProductFormViewModelProtocol`, and implemented it for `ProductVariationFormViewModel` and no-op for `ProductFormViewModel`
  - Added unit tests for `ProductVariationFormViewModel`

## Testing

- Go to the Products tab
- Tap on a variable product with non-zero variations
- Tap on the variations row
- Tap on a variation that is disabled before in core --> the "Enable" row toggle should be off
- Turn on the toggle on the "Enabled" row
- Make sure there is a price for the variation
- Tap "Update" --> the variation status should be updated remotely ("Enabled" toggle in core)
- Navigate back to the variations list --> the variation cell should not show "Hidden" anymore (note that a price needs to be set, otherwise the variation is still hidden even though it's enabled)
- Tap on the same variation cell again --> the "Enable" row toggle should be on
- Turn off the toggle on the "Enabled" row
- Tap "Update" --> the variation status should be updated remotely ("Enabled" toggle in core)
- Navigate back to the variations list --> the variation cell should now show as "Hidden"

## Example screenshots

simulator

Mode | Disabled | Enabled
-- | -- | --
Dark | Simulator Screen Shot - iPhone 11 - 2020-08-04 at 09 53 13 | Simulator Screen Shot - iPhone 11 - 2020-08-04 at 09 53 15
Light | Simulator Screen Shot - iPhone 11 - 2020-08-04 at 09 53 27 | Simulator Screen Shot - iPhone 11 - 2020-08-04 at 09 53 28

Update release notes:


WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -427,6 +427,12 @@ extension UIImage {
         return UIImage.gridicon(.types).imageFlippedForRightToLeftLayoutDirection()
     }
 
+    /// Visibility Image
+    ///
+    static var visibilityImage: UIImage {
+        return UIImage.gridicon(.visible)
+    }
```
pmusolino, line 434: This seems a duplicated icon since we use it inside the Product Settings Password row. It's called `passwordFieldImage`. What do you think about removing `passwordFieldImage` and substitute it with `visibilityImage`?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
```diff
@@ -21,6 +21,19 @@ final class ImageAndTitleAndTextTableViewCell: UITableViewCell {
...
+        let viewModel: ViewModel
+        let isSwitchOn: Bool
+        let onSwitchChange: (_ isOn: Bool) -> Void
+
+        init(viewModel: ViewModel, isSwitchOn: Bool, onSwitchChange: @escaping (_ isOn: Bool) -> Void) {
+            self.viewModel = viewModel
+            self.isSwitchOn = isSwitchOn
+            self.onSwitchChange = onSwitchChange
+        }
+    }
```
pmusolino, line 35: I think that for this use case, it could have been used the `SwitchTableViewCell`, the cell `ImageAndTitleAndTextTableViewCell` seems to have too many roles :D

jaclync: yea it might be making `ImageAndTitleAndTextTableViewCell` a bit too versatile, I just thought the `accessoryView` is separate from the cell's content view and the content view is exactly the same. `SwitchTableViewCell` doesn't have the layout that we'd like for the product form 😞 if it gets more complicated, we can move the switch cell to a new cell in the future

jaclync:
Hey Paolo! I responded to your comments and question here for another pass 🙂 

> Seems that also after an update (on core or in the app) the status "hidden" is always shown on every variation.

From the GIF, it looks like the variations don't have a price set. A variation's visibility is determined by more than just the "enabled" switch, a price is also required. If you add a price to a variation that is enabled, the variation isn't hidden anymore. It is quite confusing though, I asked G earlier and created an issue to add a warning: https://github.com/woocommerce/woocommerce-ios/issues/2594

WooCommerce/Classes/Extensions/UIImage+Woo.swift
jaclync: good point, updated here: https://github.com/woocommerce/woocommerce-ios/pull/2593/commits/4d3453cdd435b2c49093f804084619d06d24e1a8

Garance91540:
Hi guys! 

I shared a post [here](https://woomobilep2.wordpress.com/2020/08/05/add-edit-products-variations-updates/) about the impact of price settings and enabled/disabled status on the variations list:

1. When "enabled" is turned off on the variation and price is set, let's display "disabled" 
2. When the variation doesn't have a price set, let's display "set price" 
3. In any other case, let's display the price

Is that ok with you?

jaclync:
hey @Garance91540, since the goal of this PR is only to add the switch to enable/disable a variation, we will work on the warning states separately: https://github.com/woocommerce/woocommerce-ios/issues/2594

Issue #2595, opened 2020-08-04
Update the name of a product variation to include value of all attributes (including "Any")


jaclync:
Ref p91TBi-33Q-p2#comment-1383

Depends on https://github.com/woocommerce/woocommerce-ios/issues/2575 so that we can also pass a variation's parent product's attributes for the name of a variation

Right now, we just show the non-empty attributes (where the value is not "Any") for a variation's name. When a variation has no non-empty attributes, then the name is empty. Let's update a variation's name to include "Any" for all possible attributes.

<img width="590" alt="screen-shot-2020-08-05-at-9 44 07-pm" src="https://user-images.githubusercontent.com/1945542/89484429-12b1cb80-d7d1-11ea-9f87-9838faf07ceb.png">

PR #2597, opened 2020-08-04
Create wrappers for `ProductFormDataModel` conformance based on `Product` and `ProductVariation`

### LINKED ISSUE #2575: Refactor so that `Product` and `ProductVariation` don't directly conform to `ProductFormDataModel` protocol
Original discussion: https://github.com/woocommerce/woocommerce-ios/pull/2566#discussion_r462664774

To prevent from exposing feature-specific extensions on a data model, let's move the feature-specific extension methods/variables to a feature-specific protocol.

After M3, let's break `ProductFormDataModel` protocol into two protocols:
- one protocol with readonly/editable properties in `Product`/`ProductVariation` like `ProductFormCommonProperties` or `ProductAndVariationProperties`
- another protocol like `EditableProductEntity` or `ProductFormDataModel` with product form fields/functions like `isShippingEnabled` and `trimmedFullDescription`

jaclync:
Fixes #2575 

I apologized for the large PR, most of the diffs are just from renaming, and all the diffs in unit tests (probably more than half of the total diffs) are just creating the wrapper to replace in the original tests without new logic.

## Changes

- Created `EditableProductModel` and `EditableProductVariationModel` as wrappers that conform to `ProductFormDataModel` based on `Product` and `ProductVariation` respectively.
  - The goal here is so that `Product` and `ProductVariation` don't directly implement `ProductFormDataModel` protocol, since a few of the methods are feature-specific and it doesn't make sense for a `ProductVariation` to have a `shortDescription` just because it has to be used in a generic product form (original discussion https://github.com/woocommerce/woocommerce-ios/pull/2566#discussion_r462664774). Another goal is so that if the data model has dependencies other than `Product`/`ProductVariation`, they can be injected to the wrapper. An example use case is to support the variation name update in https://github.com/woocommerce/woocommerce-ios/issues/2595 that requires its parent product's attributes.
- Replaced all the existing concrete usage of `Product` and `ProductVariation` in product form related classes with `EditableProductModel` and `EditableProductVariationModel`
  - I meant to rename `product` to `productModel` or `model` in product form and a few related view controllers, but thought this PR is already very large. After the renaming, it would be less awkward to see some `product.product` calls to access `Product` for `Product` specific operations
- In `ProductFormDataModel`, replaced `virtual` and `downloadable` properties with `func isShippingEnabled() -> Bool` since these two properties were only used for this shipping logic
- Renamed a few `Product`/`ProductVariation` checks to `EditableProductModel`/`EditableProductVariationModel` respectively.
- Updated unit tests (this was the most time consuming part!)

## Testing

No app behavior changes are expected, please feel free to sanity check the product and variation form UI:

- Go to the Products tab
- Tap on a simple product --> make sure all the rows and bottom sheet actions navigate to the expected screens, and product settings are accessible
- Feel free to edit a few properties and update --> the product should be saved remotely
- Go back to the Products tab
- Tap on a variable product
- Tap on the variations row
- Tap on a variation --> make sure all the rows and bottom sheet actions navigate to the expected screens
- Feel free to edit a few properties and update --> the variation should be saved remotely

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventoryEditableData.swift
```diff
@@ -15,7 +15,7 @@ extension ProductInventoryEditableData {
     init(productModel: ProductFormDataModel) {
         self.sku = productModel.sku
         self.manageStock = productModel.manageStock
-        self.soldIndividually = (productModel as? Product)?.soldIndividually
+        self.soldIndividually = (productModel as? EditableProductModel)?.product.soldIndividually
```
shiki, line 18: The `soldIndividually` could probably be hidden behind the `ProductFormDataModel` too, right?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormDataModel.swift
```diff
@@ -35,6 +33,8 @@ protocol ProductFormDataModel {
     var dimensions: ProductDimensions { get }
     var shippingClass: String? { get }
     var shippingClassID: Int64 { get }
+    // Whether shipping settings are available for the product.
+    func isShippingEnabled() -> Bool
```
shiki, line 37: Just curious, why are these methods instead of properties? Was it to distinguish them from the DB properties? Is my question confusing? 🤪

jaclync: Yes I was trying to distinguish the visibility logic from the DB properties, it might be better if they are in two separate protocols but trying to keep the changes not too large for now 🙂

WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventoryEditableData.swift
jaclync: yes great catch! it can be in `ProductFormDataModel` now without making `ProductVariation` support a field that doesn't apply to it 😄 updated here: https://github.com/woocommerce/woocommerce-ios/pull/2597/commits/fd7d9cf161c6d2ce7db6aefdd86594b872bdc11e

jaclync:
Thanks a lot for the review and suggesting this change earlier, Jayson! The `ProductFormDataModel` could probably be broken into two, but at least `Product` and `ProductVariation` don't have to conform to the protocol directly with a bunch of fields that don't really apply to them 😃

PR #2601, opened 2020-08-04
Use Snake Case for Some Test Method Names

shiki:
Ref p91TBi-35M-p2

This changes **some** of the test names to snake-case. It doesn't seem too bad to change all of them. I'm just sending this initial set of files for us to see how it looks like. 

Let me know if we shouldn't proceed with this style. 😄 

## Testing

No functional changes. 

## Submitter Checklist

### DIFF
--- a/WooCommerce/WooCommerceTests/Model/NoteWooTests.swift
+++ b/WooCommerce/WooCommerceTests/Model/NoteWooTests.swift
@@ -17,7 +17,7 @@ class NoteWooTests: XCTestCase {
 
     /// Verifies that `blockForSubject` returns the first block in the `.subject` collection.
     ///
-    func testBlockForSubjectReturnsTheFirstBlockInTheSubjectArray() {
+    func test_blockForSubject_returns_the_first_block_in_the_subject_array() {
         for note in sampleNotes {
             XCTAssertEqual(note.blockForSubject, note.subject.first)
         }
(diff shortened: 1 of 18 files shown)


WooCommerce/WooCommerceTests/Model/NoteWooTests.swift
```diff
@@ -17,7 +17,7 @@ class NoteWooTests: XCTestCase {
 
     /// Verifies that `blockForSubject` returns the first block in the `.subject` collection.
     ///
-    func testBlockForSubjectReturnsTheFirstBlockInTheSubjectArray() {
+    func test_blockForSubject_returns_the_first_block_in_the_subject_array() {
```
jaclync, line 20: It is nicer having the property/function subject here (`blockForSubject`) 👍

WooCommerce/WooCommerceTests/ViewRelated/ReusableViews/TextFieldTextAlignmentTests.swift
```diff
@@ -5,7 +5,7 @@ final class TextFieldTextAlignmentTests: XCTestCase {
 
     // MARK: .leading
 
-    func testTextFieldTextLeadingAlignmentMapsToLeftForLTR() {
+    func test_TextFieldText_leading_alignment_maps_to_left_for_LTR() {
```
jaclync, line 8: nit: maybe `test_TextField_text_leading_alignment_maps_to_left_for_LTR` since `TextField` refers to a text field? (similar for other tests in this file)

shiki: Thanks. I was not sure about how to do this. 😄 Fixed in c464599b9c98112a7a239f74488b19f248e1750c.

shiki:
I see that @pmusolino reacted with a 🚀 so I'll merge this now. 😄

Issue #2596, opened 2020-08-04
Product Variation > Inventory: hide the stock status when stock management is disabled for a variation


jaclync:
Ref p91TBi-33Q-p2#comment-1389

Let's hide the stock status when stock management is disabled for a product variation, since this value comes from its parent product and is not editable.

<img src="https://user-images.githubusercontent.com/1945542/89262238-f04d7000-d661-11ea-8668-bd36619dda3c.png" width="300" />

PR #2602, opened 2020-08-05
Edit Product > Inventory: extract view data logic from view controller to view model

jaclync:
Part of #1985 
Prerequisite for https://github.com/woocommerce/woocommerce-ios/issues/2596

Apology again for another big refactoring PR 🙇🏻‍♀️, this time extracting the view logic from inventory settings view controller to a view model. About 300 diffs are in unit tests (including a mock for SKU validation), and I added unit tests for more complicated functions for now. No app behavior changes are expected. This will make https://github.com/woocommerce/woocommerce-ios/issues/2596 very easy to implement in a testable way.

## Changes

- Created 2 protocols for inventory settings: `ProductInventorySettingsViewModelOutput` that returns the latest readonly values and `ProductInventorySettingsActionHandler` that handles actions, similar to `ProductPriceSettingsViewModel`. Then `ProductInventorySettingsViewModel` implements both protocols.
- All the view-level logic and action handling were moved from `ProductInventorySettingsViewController` to `ProductInventorySettingsViewModel` including the SKU validation logic.
- In `ProductFormDataModel`, added `func isStockStatusEnabled() -> Bool` so that we don't have to make hacky checks on the product model type like `if product is EditableProductModel` to make stock status readonly anymore, and this is now tested.
- Added unit tests for more complicated functions like sections from a `Product`, `ProductVariation`, and SKU form type. A mock stores manager `MockProductSKUValidationStoresManager` was also created to test a duplicate SKU.

## Testing

No user-facing changes are expected. Please sanity check on the inventory settings for a product and product variation:

### Product inventory

- Go to the Products tab
- Tap on a simple product
- Tap on the inventory row --> the inventory settings should work like before. try switching on and off "Manage stock", and updating a few fields.
- Tap "Update" --> the product should be updated remotely

### Product variation inventory

- Go to the Products tab
- Tap on a variable product
- Tap on the variations row
- Tap on a variation
- Tap on the inventory row --> the inventory settings should work like before. try switching on and off "Manage stock", and updating a few fields. Stock status should be readonly, and there is no "Limit one per order" row.
- Tap "Update" --> the variation should be updated remotely

### Product SKU

- Go to the Products tab
- Tap on a grouped or external product
- Tap on the SKU row (or from the bottom sheet) --> only the SKU row is shown, try updating its value to an existing SKU in another product (an error should be shown), then updating to a unique value
- Tap "Update" --> the product should be updated remotely

Update release notes:


jaclync:
@pmusolino I just updated the `PublishSubject` in [`ProductInventorySettingsViewModel`](https://github.com/woocommerce/woocommerce-ios/blob/9feb83060cd6d464259b5207c9247a1b72467c8b/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/Inventory%20Settings/ProductInventorySettingsViewModel.swift) to `BehaviorSubject` https://github.com/woocommerce/woocommerce-ios/pull/2602/commits/0bc105b8779b4234edfa517bb019694adc3de4dd thanks to @shiki's timely addition! Also updated the unit tests (https://github.com/woocommerce/woocommerce-ios/pull/2602/commits/9feb83060cd6d464259b5207c9247a1b72467c8b) and tested manually again to verify the sections react as expected. Merging now!

Issue #2082, opened 2020-04-03
Edit Products: show product reviews on the main screen


jaclync:
- It probably requires creating a new cell
- When tapping on the cell, it navigates to the same Reviews screen as the Reviews tab but only for the product

Tasks:
- [x] Show product reviews cell.
- [x] Networking: fetch reviews for a specific product.
- [x] Present product reviews screen.

<img width="357" alt="Screen Shot 2020-04-03 at 2 20 21 PM" src="https://user-images.githubusercontent.com/1945542/78330539-c5b9b080-75b6-11ea-8f23-4be250c8971b.png">

PR #2604, opened 2020-08-05
Product Reviews: fetch reviews for a specific list of product IDs with different statuses

pmusolino:
Part of #2082 

## Description
As part of #2082 for showing the reviews for a specific product, we need to fetch all the reviews in a specific status for the specified product IDs. Then, I added two new parameters `product` and `status` in the method `loadAllProductReviews()` in `ProductReviewsRemote` and I changed accordingly `ProductReviewAction` and `ProductReviewStore`. I also moved `ProductFormActionsFactory` out from `BottomSheetListSelector` folder (since they are no more related).

## Testing
- Just a sanity check on Reviews tab, checking that everything works like before.

Update release notes:


Networking/Networking/Remote/ProductReviewsRemote.swift
```diff
@@ -25,19 +25,27 @@ public final class ProductReviewsRemote: Remote, ProductReviewsRemoteProtocol {
     ///                determines fields present in response. Default is view.
     ///     - pageNumber: Number of page that should be retrieved.
     ///     - pageSize: Number of Orders to be retrieved per page.
+    ///     - products: Limit result set to reviews assigned to specific product IDs.
+    ///     - status: Limit result set to reviews assigned a specific status. If empty, the default is `all`.
```
Ecarrion, line 29: > If nil, the default is `all`

Could be a bit more accurate 😇

pmusolino: You are right. I improved it here 4ca37984edb45a048b01cc8fcd5f9c872ca37076 👍

pmusolino:
Thank you @Ecarrion for your quick review 🎸

PR #2607, opened 2020-08-05
Add BehaviorSubject observable

shiki:
## Premise

I'm hoping to use this when implementing #2537. I could not use [`PublishSubject`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/Tools/Observables/PublishSubject.swift) because I needed this behavior provided by Combine's [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject) or ReactiveX's [`BehaviorSubject`](http://reactivex.io/documentation/subject.html):

> maintains a buffer of the most recently published element.

## BehaviorSubject

This PR only adds [`BehaviorSubject`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2537-add-behavior-subject/WooCommerce/Classes/Tools/Observables/BehaviorSubject.swift). It is a simple version of one of the Observables/Publishers that Combine supports — [`CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject). I named it as `BehaviorSubject` to avoid conflicts and to also match the similarly featured ReactiveX Observable [with the same name](http://reactivex.io/documentation/subject.html). 

### Example Usage

In a class that you would like to emit values (or events), add the `BehaviorSubject` defining the value type:

```swift
class ViewModel {
    /// Calls observers/subscribers whenever a feature's availability changes.
    private let isFeatureAvailable = BehaviorSubject<Bool>()
}
```

Since `BehaviorSubject` exposes `send()` which makes this a **mutable** Observable, we recommend exposing only the `Observable<Bool>` interface:

```swift
class ViewModel {
    private let isFeatureAvailableSubject = BehaviorSubject<Bool>()

    /// The public Observable that the ViewController will subscribe to
    var isFeatureAvailable: Observable<Bool> {
        isFeatureAvailableSubject
    }
}
```

The `ViewController` can then subscribe to the `isFeatureAvailable` Observable:

```swift
func viewDidLoad() {
    viewModel.isFeatureAvailable.subscribe { isAvailable in
        // Present an amazing new UI if the feature is available
        self.featureSubview.isHidden = !isAvailable
    }
}
```

Whenever the conditions for the feature's availability changes, like after fetching a setting from the API, the `ViewModel` can _notify_ the `ViewController` by updating `isFeatureAvailableSubject`:

```swift
fetchFromAPI { isFeatureAvailable
    // Notify the observers (e.g. ViewController) that the feature availability has changed.
    isFeatureAvailableSubject.send(isFeatureAvailable)
}
```

### Compatibility with Combine

I made sure to implement this in a way that we can replace this with `CurrentValueSubject` later. The [`BehaviorSubjectTests`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2537-add-behavior-subject/WooCommerce/WooCommerceTests/Tools/Observables/BehaviorSubjectTests.swift) are grouped into two sections:

1. The [tests for `BehaviorSubject`](https://github.com/woocommerce/woocommerce-ios/blob/afbeddb22beb8d6e0c817548cd63720d105ba954/WooCommerce/WooCommerceTests/Tools/Observables/BehaviorSubjectTests.swift#L12-L138).
2. [Similar feature tests but using Combine's `CurrentValueSubject`](https://github.com/woocommerce/woocommerce-ios/blob/afbeddb22beb8d6e0c817548cd63720d105ba954/WooCommerce/WooCommerceTests/Tools/Observables/BehaviorSubjectTests.swift#L140-L328).

In the tests, you can also see that their usages are quite similar. 

## Testing

The changes are currently unused. Please review the unit tests instead. 🙂 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/Tools/Observables/BehaviorSubject.swift
+++ b/WooCommerce/Classes/Tools/Observables/BehaviorSubject.swift
@@ -1,0 +1,115 @@
+
+import Foundation
+
+/// Emits the current value to observers upon subscription and also when new values arrive.
+///
+/// Acts like `CurrentValueSubject` in Combine and `BehaviorSubject` in ReactiveX.
+///
+/// This observable is a bridge between the imperative and reactive programming paradigms. It
+/// allows consumers to _manually_ emit values using the `send()` method.
+///
+/// Multiple observers are allowed which makes this a possible replacement for
+/// `NSNotificationCenter` observations.
+///
+/// ## Example Usage
+///
+/// In a class that you would like to emit values (or events), add the `BehaviorSubject` defining
+/// the value type:
+///
+/// ```
+/// class ViewModel {
+///     /// Calls observers/subscribers whenever a feature's availability changes.
+///     private let isFeatureAvailable = BehaviorSubject<Bool>()
+/// }
+/// ```
+///
+/// Since `BehaviorSubject` exposes `send()` which makes this a **mutable** Observable, we recommend
+/// exposing only the `Observable<Bool>` interface:
+///
+/// ```
+/// class ViewModel {
+///     private let isFeatureAvailableSubject = BehaviorSubject<Bool>()
+///
+///     /// The public Observable that the ViewController will subscribe to
+///     var isFeatureAvailable: Observable<Bool> {
+///         isFeatureAvailableSubject
+///     }
+/// }
+/// ```
+///
+/// The `ViewController` can then subscribe to the `isFeatureAvailable` Observable:
+///
+/// ```
+/// func viewDidLoad() {
+///     viewModel.isFeatureAvailable.subscribe { isAvailable in
+///         // Present an amazing new UI if the feature is available
+///         self.featureSubview.isHidden = !isAvailable
+///     }
+/// }
+/// ```
+///
+/// Whenever the conditions for the feature's availability changes, like after fetching a setting
+/// from the API, the `ViewModel` can _notify_ the `ViewController` by updating
+/// `isFeatureAvailableSubject`:
+///
+/// ```
+/// fetchFromAPI { isFeatureAvailable
+///     // Notify the observers (e.g. ViewController) that the feature availability has changed.
+///     isFeatureAvailableSubject.send(isFeatureAvailable)
+/// }
+/// ```
+///
+/// ## References
+///
+/// See here for info about similar observables in other frameworks:
+///
+/// - https://developer.apple.com/documentation/combine/currentvaluesubject
+/// - http://reactivex.io/documentation/subject.html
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/Tools/Observables/BehaviorSubject.swift
```diff
@@ -0,0 +1,115 @@
...
+
+    private typealias OnCancel = () -> ()
+
+    /// The list of Observers that will be notified when a new value is sent.
+    ///
+    private var observers = [UUID: Observer<Element>]()
+
+    /// The last value that was emitted or the initial value passed in `init()`.
+    ///
+    private(set) var value: Element
```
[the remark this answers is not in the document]
jaclync: I see that this is to match the [`value` field in `CurrentValueSubject`](https://developer.apple.com/documentation/combine/currentvaluesubject/value), let's follow it then 👍

WooCommerce/WooCommerceTests/Tools/Observables/BehaviorSubjectTests.swift
```diff
@@ -0,0 +1,329 @@
...
+        let subject = CurrentValueSubject<String, Error>("dolorem")
+        var disposeBag = Set<AnyCancellable>()
+
+        // When
+        var emittedValues = [String]()
+        subject.sink(receiveCompletion: { _ in
+            // noop
+        }) { value in
+            emittedValues.append(value)
+        }.store(in: &disposeBag)
```
jaclync, line 185: dispose bag could be an eyesore 🙄

WooCommerce/WooCommerceTests/Tools/Observables/BehaviorSubjectTests.swift
```diff
@@ -0,0 +1,329 @@
...
+import Foundation
+@testable import WooCommerce
+
+import Combine
+
+/// Tests cases for `BehaviorSubject`
+///
+final class BehaviorSubjectTests: XCTestCase {
+
+    func test_it_keeps_the_initial_value() {
```
jleandroperez, line 12: Nipticky: The standard is CamelCase 😄 

(Passing by, "annoying" by!!!)

WooCommerce/WooCommerceTests/Tools/Observables/BehaviorSubjectTests.swift
shiki: I learned that I can use `cancellables` for the name. But if you meant adding all these variables, then I have no good answer. 😄 We can probably find something to make them seamless.

PR #2623, opened 2020-08-06
Add AppSettingsAction.loadInAppFeedbackCardVisibility

shiki:
Ref #2537

This adds `AppSettingsAction.loadInAppFeedbackCardVisibility`, which determines whether the in-app feedback card should be displayed to the user. 

The result is only `true` if these conditions are met:

- The known installation date is more than 3 months ago
- The user has not given feedback for more than 6 months ago.

These values are based on the actions created in https://github.com/woocommerce/woocommerce-ios/pull/2556. 

## Installation Date

In https://github.com/woocommerce/woocommerce-ios/pull/2556, we added actions to set the `GeneralAppSettings.installationDate`. However, if we simply just use this value, then that will mean that even old users will not be able to see the in-app feedback card **until 3 months** after we've released the in-app feedback feature. 

So we can show the in-app feedback card early to existing users, I added another "installation date" to consider by checking the creation date of the Documents directory:

https://github.com/woocommerce/woocommerce-ios/blob/2f05e04ba3d6af11aebb09f338a53005f78f6dfa/Yosemite/Yosemite/Stores/AppSettings/InAppFeedbackCardVisibilityUseCase.swift#L65-L87

The oldest of the two dates (`GeneralAppSettings.installationDate` and Documents creation date) will be used as the final installation date. 

## Small Fix

While working on this, I found that I made a mistake in `loadOrCreateGeneralAppSettings `:

https://github.com/woocommerce/woocommerce-ios/blob/8c2b080fad70acf91105c3a0bad7bf138b0259b6/Yosemite/Yosemite/Stores/AppSettingsStore.swift#L166-L173

I meant that to auto-recover and create a new `GeneralAppSettings` instance if the file does not exist. I didn't realize that the `fileStorage.data(for:)` fails if the file does not exist. 🙈 I didn't see it initially because I was using a mocked `FileStorage` which behaved a bit differently. 

I fixed this in eea3f95 and ec58890. 

## Testing

The action is currently not used. The follow-up PR should use this. Please review the unit test for now. 

## Submitter Checklist

### DIFF
--- a/Yosemite/Yosemite/Stores/AppSettings/InAppFeedbackCardVisibilityUseCase.swift
+++ b/Yosemite/Yosemite/Stores/AppSettings/InAppFeedbackCardVisibilityUseCase.swift
@@ -71,18 +71,15 @@ struct InAppFeedbackCardVisibilityUseCase {
     /// consider the users who have already installed before we started tracking that value.
     ///
     private func inferInstallationDate() -> Date? {
-        let documentDirCreationDate = creationDateOfDocumentDir()
-        let savedInstallationDate = settings.installationDate
-
-        if let documentDirCreationDate = documentDirCreationDate,
-            let savedInstallationDate = savedInstallationDate {
-            return min(documentDirCreationDate, savedInstallationDate)
-        } else if documentDirCreationDate != nil {
-            return documentDirCreationDate
-        } else if savedInstallationDate != nil {
-            return savedInstallationDate
-        } else {
-            return nil
+        switch (creationDateOfDocumentDir(), settings.installationDate) {
+            case let (documentDirCreationDate?, savedInstallationDate?):
+                return min(documentDirCreationDate, savedInstallationDate)
+            case let (documentDirCreationDate?, nil):
+                return documentDirCreationDate
+            case let (nil, savedInstallationDate?):
+                return savedInstallationDate
+            default:
+                return nil
         }
     }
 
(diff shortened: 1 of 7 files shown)


Yosemite/Yosemite/Stores/AppSettings/InAppFeedbackCardVisibilityUseCase.swift
```diff
@@ -0,0 +1,115 @@
...
+        if let documentDirCreationDate = documentDirCreationDate,
+            let savedInstallationDate = savedInstallationDate {
+            return min(documentDirCreationDate, savedInstallationDate)
+        } else if documentDirCreationDate != nil {
+            return documentDirCreationDate
+        } else if savedInstallationDate != nil {
+            return savedInstallationDate
+        } else {
+            return nil
+        }
```
Ecarrion, line 86: Do you think a switch could help a little bit with readability?

```swift
switch (creationDateOfDocumentDir(), settings.installationDate) {
    case let (creationDate?, installationDate?):
        return min(creationDate, installationDate)
    case let (creationDate?, nil):
        return creationDate
    case let (nil, instalationDate?):
        return instalationDate
    default:
        return nil
}
```

shiki: Wow! I didn't know I could do that. Thanks for suggesting it! Changed in ed51de386ed1368aecd52a0bfa21a9d5d2a4f649.

shiki:
Thanks, Ernesto!

PR #2626, opened 2020-08-07
Shows in-app review from Feedback card

### LINKED ISSUE #2543: [IAFeedback] Show App Store Review Dialog

Ecarrion:
closes #2543 

# Why
This is a simple PR that presents the system in-app store review form.

# How

- Call `SKStoreReviewController.requestReview()` method to present the in-app review form.
- Hide `SKStoreReviewController.requestReview()`behind a protocol so we can mock it and add tests to this.

# Gif
in-app-review

# Testing Steps
- Launch the app and take on the "I Like it" button from the in-app feedback card.
- See that the in-app store review form appears.
**Note that in debug builds you won't be able to submit the review**

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
@@ -74,8 +74,8 @@ private extension InAppFeedbackCardViewController {
     func configureLikeButton() {
         likeButton.applyPrimaryButtonStyle()
         likeButton.setTitle(Localization.iLikeIt, for: .normal)
-        likeButton.on(.touchUpInside) { _ in
-            self.storeReviewControllerType.requestReview()
+        likeButton.on(.touchUpInside) { [weak self] _ in
+            self?.storeReviewControllerType.requestReview()
         }
     }
 }
--- a/WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
@@ -1,5 +1,16 @@
 
 import UIKit
+import StoreKit
+
+/// Defines methods for presenting the in-app app store review form.
+///
+protocol SKStoreReviewControllerProtocol {
+    /// Displays the in app app store review alert.
+    ///
+    static func requestReview()
+}
+
+extension SKStoreReviewController: SKStoreReviewControllerProtocol { }
 
 /// Displays a small view asking the user to provide a feedback for the app.
 ///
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
```diff
@@ -51,6 +74,9 @@ private extension InAppFeedbackCardViewController {
     func configureLikeButton() {
         likeButton.applyPrimaryButtonStyle()
         likeButton.setTitle(Localization.iLikeIt, for: .normal)
+        likeButton.on(.touchUpInside) { _ in
+            self.storeReviewControllerType.requestReview()
```
jaclync, line 78: nit: maybe `weak self` just in case?

WooCommerce/WooCommerceTests/ViewRelated/inAppFeedback/InAppFeedbackCardViewControllerTests.swift
```diff
@@ -0,0 +1,66 @@
+import XCTest
+
+@testable import WooCommerce
+
+class InAppFeedbackCardViewControllerTests: XCTestCase {
```
jaclync, line 5: nit: `final` if this isn't meant to be subclassed :)

WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
```diff
@@ -12,6 +23,18 @@ final class InAppFeedbackCardViewController: UIViewController {
     /// The stackview containing the `titleLabel` and the horizontal view for the buttons.
     @IBOutlet private var verticalStackView: UIStackView!
 
+    /// SKStoreReviewController type wrapper. Needed for testing
+    private let storeReviewControllerType: SKStoreReviewControllerProtocol.Type
```
jaclync, line 27: ah DI'ing a static protocol does look a bit funny 😆

Ecarrion: Yeah. It looks a bit ugly! 
But I couldn't come up with a better way 😓

Ecarrion:
> Do we just follow the timing logic as in the Apple doc so that it most likely shows the review prompt when "I Like It" CTA is visible?

Yes, we are trying to match those rules :-)

> Do we hide the feedback card after they tap the "I Like It" CTA since there is no callback when they are done with the review prompt?

Yes, we have separate tasks to hide the card and to add analytics!

WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
shiki: That's smart and weird at the same time!

Issue #2629, opened 2020-08-07
Hide in-app feedback card when tapping on CTAs


Ecarrion:
#2537 Should be completed first

After tapping the "Could be better" or "I like it button" the in-app feedback card on the dashboard card should be hidden.

Ecarrion:
@shiki should the card be dismissed when tapping the "Could be better" button or should we wait for the survey completion? 🤔

shiki:
Thanks for creating this issue! 

I think it should be the same as the "I Like It" button so we should hide it right away. Or else, it would feel like we're *requiring* the feedback. If the user wants to give some feedback later, they can use the button that will be implemented in the settings (#2553).

Ecarrion:
[Initial implementation here](https://github.com/woocommerce/woocommerce-ios/tree/issue/2669-hide-app-feedback-card) while #2537 is completed

PR #2605, opened 2020-08-05
Adds in-app Survey loading view

Ecarrion:
fixes #2571 

# Why

Prior to this PR, the user was left staring at an empty screen while the in-app survey loads. Since the survey can take a couple of seconds to load, we have decided to add a loading indicator view.

# How

- Creates a new `SuveryLoadingView` to hold the loading indicator
- Present the loading view when the request to load the survey is fired
- Hide the loading view when the survey has finished loading

PS: I've also updated some tests to use the new method name case.

# GIF

Survey-loading-view

# Testing Steps

- Add the following code in `InAppFeedbackCardViewController`, `configureDidNotLikeButton()` method
```swift
didNotLikeButton.on(.touchUpInside) { [weak self] _  in
    let vc = SurveyViewController(survey: .inAppFeedback, onCompletion: { })
    let nav = WooNavigationController(rootViewController: vc)
    self?.present(nav, animated: true)
}
```
- Tap on the "Could be better" button in the feedback card on the dashboard
- See that the loading spinner is present while the survey loads.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2633, opened 2020-08-07
Integrate In-app Feedback Card Conditions for Displaying

### LINKED ISSUE #2537: [IAFeedback] Display In-app Feedback Question in Stats

shiki:
Closes #2537. 

This depends on #2623 which has been reviewed and approved but hasn't been merged yet. I'll merge it after the release cut.

This PR integrates the action [`AppSettingsAction.loadInAppFeedbackCardVisibility`](https://github.com/woocommerce/woocommerce-ios/blob/fd208e19dee798df19578b636ba03a4a075dd5a7/Yosemite/Yosemite/Actions/AppSettingsAction.swift#L83-L90) (#2623) to the card UI. The UI will now only show up in these conditions:

1. Users who’ve installed the app at least 3 months ago.
2. Once every 6 months. (this will have to be tested later)

ℹ️  The buttons' functionalities will be integrated later with #2626 and #2605. 

### ViewModel

I moved the `canDisplayInAppFeedbackCard` logic into a new ViewModel, [`StoreStatsAndTopPerformersPeriodViewModel`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2537-in-app-feedback-card-visibility-conditions/WooCommerce/Classes/ViewRelated/Dashboard/Stats%20v4/StoreStatsAndTopPerformersPeriodViewModel.swift). This ViewModel is responsible for dictating when the in-app feedback card should be shown. 

Later, in #2629, the buttons will also call a function inside here to call [`AppSettingsAction.setLastFeedbackDate`](https://github.com/woocommerce/woocommerce-ios/blob/f0828fe1f156c33d44b37614574f697dedbed35a/Yosemite/Yosemite/Actions/AppSettingsAction.swift#L78-L81). 

### UI

The UI uses the new `isInAppFeedbackCardVisible` `Observable` exposed by the ViewModel to hide or show the in-app feedback card: 

https://github.com/woocommerce/woocommerce-ios/blob/fd208e19dee798df19578b636ba03a4a075dd5a7/WooCommerce/Classes/ViewRelated/Dashboard/Stats%20v4/StoreStatsAndTopPerformersPeriodViewController.swift#L173-L190

## Testing

We can only easily test the condition number 1 for now: "Users who’ve installed the app at least 3 months ago.". The second condition will be tested later after the functionalities of the buttons are integrated.

2. Run the app and log in. 
3. Time travel to at least 90 days from now. 
4. Navigate to the Orders tab and navigate back to My Store.
5. Confirm that the in-app feedback card is now visible.
6. Time travel back to the present. 
7. Navigate to the Orders tab and navigate back to My Store.
8. Confirm that the in-app feedback card is no longer visible.

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
@@ -170,9 +170,16 @@ private extension StoreStatsAndTopPerformersPeriodViewController {
                 return
             }
 
-            UIView.animate(withDuration: 0.2) {
-                self.inAppFeedbackCardViewsForStackView.forEach {
-                    $0.isHidden = !isVisible
+            let isHidden = !isVisible
+
+            self.inAppFeedbackCardViewsForStackView.forEach { subView in
+                // Check if the subView will change first. It looks like if we don't do this,
+                // then StackView will not animate the change correctly.
+                if subView.isHidden != isHidden {
+                    UIView.animate(withDuration: 0.2) {
+                        subView.isHidden = isHidden
+                        self.stackView.setNeedsLayout()
+                    }
                 }
             }
         }
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
@@ -111,6 +111,11 @@ final class StoreStatsAndTopPerformersPeriodViewController: UIViewController {
 
         configureSubviews()
     }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        viewModel.onViewDidAppear()
+    }
 }
 
 // MARK: Public Interface
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
@@ -66,6 +66,15 @@ final class StoreStatsAndTopPerformersPeriodViewController: UIViewController {
     }()
 
     private lazy var inAppFeedbackCardViewController = InAppFeedbackCardViewController()
+    /// An array of UIViews for the In-app Feedback Card. This will be dynamically shown
+    /// or hidden depending on the configuration.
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewController.swift
```diff
@@ -160,6 +164,32 @@ private extension StoreStatsAndTopPerformersPeriodViewController {
         }
     }
 
+    /// Observe and react to visibility events for the in-app feedback card.
+    func configureInAppFeedbackCardViews() {
+        guard viewModel.canDisplayInAppFeedbackCard else {
+            return
+        }
+
+        _ = viewModel.isInAppFeedbackCardVisible.subscribe { [weak self] isVisible in
```
Ecarrion, line 173: I thought that the observation token had to be retained in order for the subscription to be alive, but after checking the implementation I see that the closure is saved within the subject. 👍 

My only concern would be, isn't this different on how combine works? I think I remember you have to store the `cancelable` object after `sink` in order to keep the subscription active. I may be wrong tho! Haven't really used Combine 😅

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersPeriodViewModel.swift
```diff
@@ -0,0 +1,69 @@
...
+        updateIsInAppFeedbackCardVisibleValue()
+    }
+
+    /// Calculates and updates the value of `isInAppFeedbackCardVisibleSubject`.
+    private func updateIsInAppFeedbackCardVisibleValue() {
+        // Abort right away if we don't need to calculate the real value.
+        let isEnabled = canDisplayInAppFeedbackCard && featureFlagService.isFeatureFlagEnabled(.inAppFeedback)
+        if !isEnabled {
+            return isInAppFeedbackCardVisibleSubject.send(isEnabled)
+        }
```
Ecarrion, line 55: [nit]: Maybe a `guard let` would be better as it enforces the return statement.
```swift
guard canDisplayInAppFeedbackCard && featureFlagService.isFeatureFlagEnabled(.inAppFeedback) else {
    return isInAppFeedbackCardVisibleSubject.send(false)
}
```

Also, TIL that you could return a void function with a statement when the statement returns void.

shiki:
Thanks, Ernesto and Jaclyn!

PR #2638, opened 2020-08-11
Hide inApp feedback card after CTAs

### LINKED ISSUE #2629: Hide in-app feedback card when tapping on CTAs

Ecarrion:
closes #2629

# Why
This PR hides the in-App feedback card in the dashboard after the user taps on any call to action. Which could be the "I like it" button which will trigger the app store review or the "Could be Better" button which will trigger the survey flow.

# How

- Expose closure on `InAppFeedbackCardViewController` to inform it's parent controller when the feedback has been given.
- Adds a new method `onInAppFeedbackCardAction` on `StoreStatsAndTopPerformersPeriodViewModel` which sets the `lastFeedbackDate` and recalculates the `inAppFeedbackCard` visibility.

# Gifs

dismiss-card

# Testing Steps
- Delete the app on the device.
- Run the app and log in.
- Time travel to at least 90 days from now.
- Navigate to the Orders tab and navigate back to My Store.
- Confirm that the in-app feedback card is now visible.
- Tap on any call to action.
- See that the CTA happened and the card was dismissed
Note: In-app review won't show if the system is date is altered

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
@@ -20,6 +20,9 @@ final class InAppFeedbackCardViewController: UIViewController {
     @IBOutlet private var didNotLikeButton: UIButton!
     @IBOutlet private var likeButton: UIButton!
 
+    /// Closure invoked after the user has choose what kind feedback to give.
+    var onFeedbackGiven: (() -> Void)?
+
     /// The stackview containing the `titleLabel` and the horizontal view for the buttons.
     @IBOutlet private var verticalStackView: UIStackView!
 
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/InAppFeedback/InAppFeedbackCardViewController.swift
```diff
@@ -20,6 +20,9 @@ final class InAppFeedbackCardViewController: UIViewController {
     @IBOutlet private var didNotLikeButton: UIButton!
     @IBOutlet private var likeButton: UIButton!
 
+    /// Closure invoked after the user has choose what kind feedback to give.
```
jaclync, line 23: super nit: choose --> chosen?

Ecarrion:
Thanks for the review @jaclync and nice catch about the crash! Weird that it didn't occur on the simulator 🤔

PR #2639, opened 2020-08-11
Removes Sign in with Google temporarily until we have SIWA

astralbodies:
This temporarily removes the Sign in with Google option. Users can still log in with their e-mail address using the magic link option until we're ready to deploy Sign in with Apple.

Before this change:
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-08-11 at 12 51 42

After this change:
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-08-11 at 12 50 57

The fix was to add an initialization parameter to the authenticator library, which is kind of a hack. We're using a different prologue which bypasses some of the logic that the boolean being true would trigger. It effectively removes the Google button from our initial screen used in this app.

Update release notes:


jkmassel (approved):
This looks very straightforward from a code perspective, so LGTM

:shipit:

PR #2649, opened 2020-08-13
Docs: Mention rbenv Preference

shiki:
I promised to do this more than 2 months ago during a discussion with Platform. 😅 It's pretty common to have Ruby issues and the usual solution is to use rbenv. The built-in macOS Ruby is so easy to break. This PR updates the `README.md` to mention that we prefer rbenv when installing Ruby. I also made the Ruby requirement as the second step. 

## Testing

Please view the updated `README.md` [here](https://github.com/woocommerce/woocommerce-ios/blob/update/docs-rbenv/README.md). 

## Submitter Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,9 @@ We use [SwiftLint](https://github.com/realm/SwiftLint) to enforce a common style
 
 If your code has any style violations, you can try to automatically correct them by running:
 
-`rake lint:autocorrect`
+```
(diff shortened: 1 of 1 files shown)


pmusolino:
> The built-in macOS Ruby is so easy to break

💯 % true 😢

shiki:
Thank you, both!

PR #2650, opened 2020-08-13
Updated Wormholy from v1.6.0 to v1.6.2

pmusolino:
## Description
This PR updates Wormholy from version 1.6.0 to version 1.6.2.
This last version introduces the dark mode, and the requests body is now present again. It was removed wrongly in release 1.6.0.

## Testing
- Try to open Wormholy from the Settings screen running the app in `debug`.

Update release notes:


pmusolino:
Thanks for testing it @shiki. You are right, it's not possible to call Wormholy if you are not logged. It's a good idea to have the possibility to launch it also on the login page (and not only). I filled an issue here https://github.com/woocommerce/woocommerce-ios/issues/2660

PR #2653, opened 2020-08-13
Edit Product: the product screen now shows the "Product type" row

### LINKED ISSUE #2262: Edit Products: add product type row

pmusolino:
Closes #2262 

## Description
In this PR, I handled the logic for showing the Product Type cell in the Edit Product screen.
The Product Type cell can show these Product types:
- Physical (type is simple, the virtual field is off)
- Virtual (type is simple, the virtual field is on)
- Variable (type is variable)
- Grouped (type is grouped)
- External (type is external)
- A custom product type sent through the API (supported as it comes for free, but there doesn't seem to be any way to use a custom product type).

No actions are supported at the moment, will be part of another PR (here the [related issue](https://github.com/woocommerce/woocommerce-ios/issues/2198)).

## Testing
1. Navigate to different products (simple physical, simple virtual, variable, grouped, external).
2. For each product, make sure that the cell "Product type" shows the right value.

## Screenshots
| Product Type row             |  Product Type row (dark mode enabled) |
Simulator Screen Shot - iPhone 11 Pro - 2020-08-13 at 17 19 40 | Simulator Screen Shot - iPhone 11 Pro - 2020-08-13 at 17 19 49

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
@@ -165,6 +165,10 @@ extension ProductFormViewModel {
         product = EditableProductModel(product: product.product.copy(averageRating: averageRating, ratingCount: ratingCount))
     }
 
+    func updateProductType(productType: ProductType) {
+        product = EditableProductModel(product: product.product.copy(productTypeKey: productType.rawValue))
+    }
+
     func updateInventorySettings(sku: String?,
                                  manageStock: Bool,
                                  soldIndividually: Bool?,
--- a/WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
@@ -241,8 +241,6 @@ extension ProductDetailsViewModel {
             configureProductName(cell)
         case let cell as OldProductReviewsTableViewCell:
             configureReviews(cell)
-        case let cell as TitleBodyTableViewCell where row == .productType:
-            configureProductType(cell)
         case let cell as WooBasicTableViewCell where row == .permalink:
             configurePermalink(cell)
         case let cell as WooBasicTableViewCell where row == .affiliateLink:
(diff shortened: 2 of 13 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -165,6 +165,10 @@ extension ProductFormViewModel {
         product = EditableProductModel(product: product.product.copy(averageRating: averageRating, ratingCount: ratingCount))
     }
 
+    func updateProductType(productType: ProductType) {
```
jaclync, line 168: this doesn't seem to be used in this PR, maybe it can be saved for the PR that implements the product type update?

pmusolino: Yes correct, it will be used in the next PR 😉

pmusolino:
Thank you @jaclync for the quick review! 🙌

Garance91540:
Hi @pmusolino great work! quickly glancing over this, is it possible to move the order of the product type row in the future so it's at the bottom of the list? https://d.pr/i/ZqJZQF

PR #2656, opened 2020-08-14
Fix inconsistent failure in SurveyVC tests

Ecarrion:
# Why

@pmusolino Reported the other day that some tests on about SurveyViewController were randomly failing.
Today I was able to get the failure to happen multiple times and it seems that when the `waitUntil` timeout was `1` sometimes the test failed even when the condition was met on the same run loop(This is my assumption).

# How 

The fix was to increase the timeout to use the default value and tests appear to run normally now.

# Testing Steps
- See that tests pass

Update release notes:


jaclync (approved):
I wasn't able to reproduce, but the changes look good in any case.

It's a bit confusing to me why this test case `test_it_completes_after_receiving_a_form_submitted_POST_callback_request` might have failed 🤔 `onCompletion` is called directly in the webview `decidePolicy` function, and I can't think of any reasons how this could have taken longer than 1 second. I was wondering if using XCTest `wait` + fulfilling the expectation would make any difference?

pmusolino (approved):
Now it works for me 👍 I'm no more able to reproduce the failing test on this branch

PR #2659, opened 2020-08-14
Settings: add back Products feature switch for M3 release

### LINKED ISSUE #2482: Settings: re-enable beta features > Products feature switch for M3 release
Since we removed the beta features section in settings after the stats v4 updates, let's re-enable the section and Products feature switch when we are ready to release Products M3.

jaclync:
Fixes #2482 

## Changes

- Updated `BetaFeaturesViewController` to show the products feature switch again, and removed code that involves the old feature flag `editProducts`
- Updated `SettingsViewController` to show "Experimental Features" row again
- Product settings: DI'ed `isEditProductsRelease3Enabled ` from `ProductFormViewController` to `ProductSettingsViewController` to `ProductSettingsViewModel` down to `ProductSettingsSectionMediator` protocol's init (`ProductSettingsSections.PublishSettings` and `ProductSettingsSections.MoreOptions`) to show different rows based on the flag and product type
- In `ProductDetailsFactory`, updated to enable editing for core product types when the feature switch is on (except that simple products are also editable) and pass `isEditProductsRelease2Enabled` as `true` (since we fully launched M2) and `isEditProductsRelease3Enabled` as the feature switch value
- In `ProductsTopBannerFactory`, updated to show the message based on whether the products feature switch is on
- Added `AppSettingsAction.resetFeatureSwitches` to reset all the feature switches including M2 and M3 feature switches. Since the user could have had the previous M2 feature switch on and we want M3 feature switch to be off for all users by default, I updated the file URL for M3 switch and created a reset function to delete these switch settings on logout. Please lemme know if you have any other suggestions!
- Updated/added unit tests for most of the changes

### Feature switch off

- Go to Settings > Experimental Features --> make sure the Products feature switch is off by default (before any actions on the switch; you can also delete the app to reset)
- Go to the Products tab --> the top banner should show the M2 messaging (`We've added more editing functionalities to products! You can now update images, see previews and share your products.`)
- Tap on a simple product --> it should be editable like before
- Tap on a variable product --> it should not be editable
- Tap on a grouped product --> it should not be editable
- Tap on an external product --> it should not be editable

### Feature switch on

- Go to Settings > Experimental Features
- Turn on the feature switch
- Go to the Products tab --> the top banner should show the M3 messaging (`You can now edit grouped, external and variable products, change product type and update categories and tag`)
- Tap on a simple product --> it should be editable like before, and the product settings should show the product reviews switch and virtual switch
- Tap on a variable product --> it should be editable, and the product settings should show the product reviews switch
- Tap on a grouped product --> it should be editable, and the product settings should show the product reviews switch
- Tap on an external product --> it should be editable, and the product settings should show the product reviews switch
- Log out of the app and log in again
- Go to Settings > Experimental Features --> the feature switch should be off (reset)

## Example screenshots

Products tab with M3 messaging:

light | dark
-- | --
Simulator Screen Shot - iPhone 11 - 2020-08-19 at 13 21 51 | Simulator Screen Shot - iPhone 11 - 2020-08-19 at 13 22 05

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/Beta features/BetaFeaturesViewController.swift
```diff
@@ -83,7 +83,9 @@ private extension BetaFeaturesViewController {
     ///
     func configureSections() {
         // This is empty because there aren't any ongoing experiments
```
pmusolino, line 85: Should we remove this comment?

Yosemite/YosemiteTests/Stores/AppSettingsStoreTests+ProductsFeatureSwitch.swift
```diff
@@ -21,12 +21,24 @@ final class AppSettingsStoreTests_ProductsFeatureSwitch: XCTestCase {
     ///
     private var subject: AppSettingsStore!
 
+    // Previous feature switch file URL
+    private lazy var productsFeatureSwitchURL: URL = {
+        let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
+        return documents!.appendingPathComponent("products-feature-switch.plist")
+    }()
```
pmusolino, line 28: Should we still keep this file since it is no longer used?

jaclync: it's still used when deleting feature switches - since it's possible the device had the previous version with the Products M2 feature switch PList file, I thought it'd be nice to delete all previous feature switches here. Lemme know if you think there's a better way to clean up the previous feature switch file!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Beta features/BetaFeaturesViewController.swift
jaclync: yup! removed here: https://github.com/woocommerce/woocommerce-ios/pull/2659/commits/ec1bbe522374ef9859dd9205f2585484d560091c

jaclync:
Thanks for catching the issue @pmusolino, I was able to repro this! It was because on log out, the app was trying to delete the previous feature switch and the current feature switch files while the previous feature switch file did not exist anymore (either from a fresh launch or previous logout). In https://github.com/woocommerce/woocommerce-ios/pull/2659/commits/d93f5351aec4f589d06eb67793ddfc03613ff28c, I updated the delete method in `PListFileStorage` to only delete a file if the file exists so that a series of delete calls doesn't fail on one without a file.

I also responded to other comments for another look!

Yosemite/YosemiteTests/Stores/AppSettingsStoreTests+ProductsFeatureSwitch.swift
pmusolino: Thanks for clarifying this aspect. The only thing is that probably if we remove it in the future, we will always have users with this plist file stored, maybe because they are updating from an old version to a recent version of the app. In short I think we will never be able to completely remove all the old plist files unless we keep track of them somewhere forever.

pmusolino:
> Thanks for catching the issue @pmusolino, I was able to repro this! It was because on log out, the app was trying to delete the previous feature switch and the current feature switch files while the previous feature switch file did not exist anymore (either from a fresh launch or previous logout). In d93f535, I updated the delete method in PListFileStorage to only delete a file if the file exists so that a series of delete calls doesn't fail on one without a file.

Looks good to me! Thank you 💯 Jaclyn!

jaclync:
> One quick note, can we update copy so it reads "tags"? thank you!

@Garance91540 yep! I updated the copy in https://github.com/woocommerce/woocommerce-ios/pull/2659/commits/285b2faac63baefb0bd0e3e2a876588ace87a4a6 and also the screenshots in PR description 👍

PR #2661, opened 2020-08-14
Handle General inApp Feedback Survey Errors

### LINKED ISSUE #2655: [IAFeedback] Survey Completion Does Not Consider Errors
When finishing a survey without filling in the required fields, the app incorrectly shows the "Feedback Sent" page. 

However, the survey didn't really finish yet since there are validation errors. This is what it looks like when submitting the **production survey** without filling in the fields.:

Ecarrion:
fixes #2655 

# Why
Prior to this PR the `SurveyViewController` didn't take survey errors into consideration before navigating the user to the survey completion screen.

# How
After looking at how the survey behaves I came to the conclusion the at the user is redirected to the survey URL with a special parameter(`msg=done`) when the user has successfully completed a survey. I couldn't find official documentation for this but it was confirmed in p1597419047017800-slack-crowdsignal

# GIF
survey

# Testing Steps
- Add the following code in `DashboardViewController`
```swift
let button = UIBarButtonItem(image: .productReviewsImage, style: .plain, target: self, action: #selector(showSurvey))
navigationItem.setLeftBarButton(button, animated: false)

@objc private func showSurvey() {
    let coordinator = SurveyCoordinatingController(survey: .inAppFeedback)
    present(coordinator, animated: true, completion: nil)
}
```
- Launch the app and tap on the left bar button item
- Tap on finish survey without filling it
- See that the errors are rendered on screen
- Fill the survey
- Tap on finish
- See that you are redirected to the survey completion screen

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Survey/SurveyCoordinatingController.swift
+++ b/WooCommerce/Classes/ViewRelated/Survey/SurveyCoordinatingController.swift
@@ -47,7 +47,7 @@ private extension SurveyCoordinatingController {
         }, onBackToStoreAction: { [weak self] in
             self?.finishSurveyNavigation()
         })
-        show(completionViewController, sender: self)
+        setViewControllers([completionViewController], animated: true)
     }
 
     /// Dismisses the flow modally
--- a/WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift
@@ -110,11 +110,11 @@ extension SurveyViewController: WKNavigationDelegate {
         guard case .formSubmitted = navigationAction.navigationType,
             let url = navigationAction.request.url,
             let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: false)?.queryItems,
-            let surveyMessageTag = queryItems.first(where: { $0.name == Constants.surveyMessageTagKey })?.value else {
+            let surveyMessageValue = queryItems.first(where: { $0.name == Constants.surveyMessageParameterKey })?.value else {
                 return
         }
 
-        if surveyMessageTag == Constants.surveyCompletionTagValue {
+        if surveyMessageValue == Constants.surveyCompletionParameterValue {
             onCompletion()
         }
     }
@@ -128,7 +128,7 @@ extension SurveyViewController: WKNavigationDelegate {
 //
 private extension SurveyViewController {
     enum Constants {
-        static let surveyMessageTagKey = "msg"
-        static let surveyCompletionTagValue = "done"
+        static let surveyMessageParameterKey = "msg"
+        static let surveyCompletionParameterValue = "done"
     }
 }
--- a/WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift
@@ -99,16 +99,36 @@ extension SurveyViewController {
 extension SurveyViewController: WKNavigationDelegate {
     func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
 
-        // The condition for the POST HTTP method is necessary, since the `formSubmitted` callback is triggered when showing the Crowdsignal completion screen
-        // (a GET request).
-        if case .formSubmitted = navigationAction.navigationType, navigationAction.request.httpMethod == "POST" {
-            onCompletion()
+        defer {
+            decisionHandler(.allow)
         }
 
-        decisionHandler(.allow)
+        // To consider the survey as completed, the following conditions need to occur:
+        // - Survey Form is submitted.
+        // - The request URL contains a `msg` parameter key with `done` as it's value
+        //
+        guard case .formSubmitted = navigationAction.navigationType,
+            let url = navigationAction.request.url,
+            let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: false)?.queryItems,
+            let surveyMessageTag = queryItems.first(where: { $0.name == Constants.surveyMessageTagKey })?.value else {
+                return
+        }
+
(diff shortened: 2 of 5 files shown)


WooCommerce/Classes/ViewRelated/Survey/SurveyCoordinatingController.swift
```diff
@@ -47,7 +47,7 @@ private extension SurveyCoordinatingController {
         }, onBackToStoreAction: { [weak self] in
             self?.finishSurveyNavigation()
         })
-        show(completionViewController, sender: self)
+        setViewControllers([completionViewController], animated: true)
```
shiki, line 50: Nice idea!

WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift
```diff
@@ -99,16 +99,36 @@ extension SurveyViewController {
...
-        if case .formSubmitted = navigationAction.navigationType, navigationAction.request.httpMethod == "POST" {
-            onCompletion()
+        defer {
+            decisionHandler(.allow)
+        }
+
+        // To consider the survey as completed, the following conditions need to occur:
+        // - Survey Form is submitted.
+        // - The request URL contains a `msg` parameter key with `done` as it's value
+        //
```
shiki, line 109: Thanks for adding this comment!

WooCommerce/WooCommerceTests/ViewRelated/Survey/SurveyViewControllerTests.swift
```diff
@@ -118,19 +141,36 @@ private extension SurveyViewControllerTests {
...
+
+        static var nonCompletedAction: FormSubmittedNavigationAction {
+            return FormSubmittedNavigationAction(messageTag: "other")
+        }
+
+        static var completedAction: FormSubmittedNavigationAction {
+            return FormSubmittedNavigationAction(messageTag: "done")
+        }
+
+        private let messageTag: String?
```
shiki, line 156: Not important. "Tag" may be a bit unclear since `msg` is really a [query parameter](https://branch.io/glossary/query-parameters/). Maybe we can name it as such?

WooCommerce/WooCommerceTests/ViewRelated/Survey/SurveyViewControllerTests.swift
```diff
@@ -118,19 +141,36 @@ private extension SurveyViewControllerTests {
...
-            var request = URLRequest(url: WooConstants.URLs.inAppFeedback.asURL())
-            request.httpMethod = httpMethod
-            return request
+            var urlComponents = URLComponents(url: WooConstants.URLs.inAppFeedback.asURL(), resolvingAgainstBaseURL: false) ?? URLComponents()
+            if let messageTag = messageTag {
+                let item = URLQueryItem(name: "msg", value: messageTag)
+                urlComponents.queryItems = [item]
+            }
+
+            let finalURL = (try? urlComponents.asURL()) ?? WooConstants.URLs.inAppFeedback.asURL()
```
shiki, line 172: I think it's fine to forcefully unwrap the result here. The fallback value was more confusing for me. 🙂 

```suggestion
            let finalURL = try! urlComponents.asURL()
```

WooCommerce/WooCommerceTests/ViewRelated/Survey/SurveyViewControllerTests.swift
```diff
@@ -57,7 +57,30 @@ final class SurveyViewControllerTests: XCTestCase {
...
-        let navigationAction = FormSubmittedNavigationAction(httpMethod: "GET")
+        let navigationAction = FormSubmittedNavigationAction.nonCompletedAction
+        viewController.webView(mirror.webView, decidePolicyFor: navigationAction, decisionHandler: { _ in })
+        waitForExpectations(timeout: Constants.expectationTimeout)
+
+        // Then
+        XCTAssertFalse(surveyCompleted)
+    }
+
+    func test_it_does_not_complete_after_receiving_a_form_submitted_empty_callback_request() throws {
```
shiki, line 68: ⚠️ This test and the test above takes at least 10 seconds. A second is quite expensive in CircleCI, especially when we have more developers submitting PRs in the future. 😅 

I did a quick test and it looks like we can use the `decisionHandler` for waiting. 

```swift
    func test_it_does_not_complete_after_receiving_a_form_submitted_empty_callback_request() throws {
        // Given
        var surveyCompleted = false
        let viewController = SurveyViewController(survey: .inAppFeedback, onCompletion: {
            surveyCompleted = true
        })

        // When
        _ = try XCTUnwrap(viewController.view)
        let mirror = try self.mirror(of: viewController)

        // Fakes a form submission GET navigation
        let navigationAction = FormSubmittedNavigationAction.emptyAction

        waitForExpectation { exp in
            viewController.webView(mirror.webView, decidePolicyFor: navigationAction, decisionHandler: { _ in
                exp.fulfill()
            })
        }

        // Then
        XCTAssertFalse(surveyCompleted)
    }
```

This pattern seems to work fine when I used it for `test_it_completes_after_receiving_a_form_submitted_completed_callback_request` too so I think it's testing the functionality accurately. `:nervouslaugh:`

Also, thank you very much for testing this. This is a very important logic! 🙏

WooCommerce/WooCommerceTests/ViewRelated/Survey/SurveyViewControllerTests.swift
Ecarrion: Yeah, I used `Tag` because that is how CrowdSignal docs refer to it. So I had that terminology in mind. But I'll rename it.

WooCommerce/WooCommerceTests/ViewRelated/Survey/SurveyViewControllerTests.swift
Ecarrion: Uuu nice idea! I did not like that waiting at all!

Ecarrion:
Thanks for the review @shiki 
I've made the changes, please take another look!

Issue #1549, opened 2019-12-03
"Unable to simultaneously satisfy constraints" in Product List


pmusolino:
Opening Product List on iPhone 11 Pro (simulator), the console prints a lot of errors about auto layout. They seem related to `TopBannerView`.

![Simulator Screen Shot - iPhone 11 Pro - 2019-12-03 at 17 25 54](https://user-images.githubusercontent.com/495617/70069419-0813b680-15f2-11ea-9806-f8a4e350b1d4.png)

```[LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x600002cf27b0 UIView:0x7fb8dc75d590.leading == UIStackView:0x7fb8dc75d400.leading - 16   (active)>",
    "<NSLayoutConstraint:0x600002cf1720 H:[UIStackView:0x7fb8dc75d400]-(10)-|   (active, names: '|':UIView:0x7fb8dc75d590 )>",
    "<NSLayoutConstraint:0x600002cf1270 WooCommerce.TopBannerView:0x7fb8dc75c570.leading == UIStackView:0x7fb8dc75d870.leading   (active)>",
    "<NSLayoutConstraint:0x600002cf0b90 H:[UIStackView:0x7fb8dc75d870]-(0)-|   (active, names: '|':WooCommerce.TopBannerView:0x7fb8dc75c570 )>",
    "<NSLayoutConstraint:0x600002cf1d10 UIView:0x7fb8dc7510c0.leading == WooCommerce.TopBannerView:0x7fb8dc75c570.leading   (active)>",
    "<NSLayoutConstraint:0x600002cf10e0 H:[WooCommerce.TopBannerView:0x7fb8dc75c570]-(0)-|   (active, names: '|':UIView:0x7fb8dc7510c0 )>",
    "<NSLayoutConstraint:0x600002cdaee0 'UISV-canvas-connection' UIStackView:0x7fb8dc75ca30.leading == UILabel:0x7fb8dc75bec0'Work in progress!'.leading   (active)>",
[...]

PR #2664, opened 2020-08-17
Add a warning banner on top of the variation list when any variations are missing a price

### LINKED ISSUE #2594: Update product variation form and variations list to add warning UI to better reflect when price is missing
There are some design changes on the variation for and variation list:

## Product variation form

- [x] If a variation is enabled but doesn't have a price set, let's update the following https://github.com/woocommerce/woocommerce-ios/pull/2627:
  - [x] update the price row to be warning color "Add Price" 
  - [x] add a warning banner row below
  - We can create a new view model like `NoticeViewModel` or `WarningViewModel` to `ImageAndTitleAndTextTableViewCell` for the orange background and icon. 

Screen Shot 2020-08-06 at 10 18 54 AM

## Variations list

- [x] Update variation cell UI based on the follow states https://github.com/woocommerce/woocommerce-ios/pull/2627:
  - Variation is disabled: "Disabled" is shown
  - Variation is enabled and has a price: the price is shown as before
  - Variation is enabled but doesn't have a price: "No price set" is shown in warning color
- [ ] Add a warning banner on top of the variations if there are any variations without a price set
  - It's probably easier to implement the warning view as a separate `UIView` as its header view so that it doesn't affect the table view data source.

Screen Shot 2020-08-06 at 10 22 10 AM

jaclync:
Fixes #2594 

## Changes

- Created `ProductVariationsTopBannerFactory` that generates the top banner view when some variations are missing a price, the copy mostly following the [Android PR](https://github.com/woocommerce/woocommerce-android/pull/2717)
- Updated `TopBannerView` and `TopBannerViewModel` to allow no top button and support normal/warning banner type
- In `ProductVariationsViewController`, showed/hid the warning top banner based on whether the latest variations have any missing a price.
  - Showing/hiding the top banner dynamically took me 2ish hours 😰 I've tried setting `tableView.tableHeaderView` to `nil`/`topBannerView` whenever the variations change, but it either leaves a gap after hiding the banner or covering the top cells after showing the banner. I also tried showing the banner as the section header and that also had weird UI issue after showing/hiding the banner dynamically. Eventually, I got it working by making `tableView.tableHeaderView` a frame-based container view that contains a stack view that contains the top banner then showing/hiding the banner by setting its `isHidden` property. If you think I might have missed something with the previous approaches or know of other ways to show/hide a table header view dynamically, please lemme know!
  - The callback when the variation results change in storage layer is moved to a separate function `transitionToResultsUpdatedState`, because after transitioning from ghost view to normal results we have to set up the results controller forwarding to table view again to include updating the top banner visibility (right now we don't). the top banner visibility also needs to be updated in `transitionToResultsUpdatedState` since the table view doesn't receive the results controller change events in ghost view state (animated placeholder state).

## Testing

- Go to the Products tab
- Tap on a variable product where all the variations have a price
- Tap on the variations row --> there should be no warning banner on top of the list
- Tap on a variation, remove the price, and tap "Update" to save remotely
- Go back to the variations screen --> a warning banner should show up
- Tap on the variation whose price was removed, set a price again, and tap "Update" to save remotely
- Go back to the variations screen --> a warning banner should disappear

## Example screenshots

simulator

light | dark
-- | --
Simulator Screen Shot - iPhone 11 - 2020-08-17 at 17 02 57 | Simulator Screen Shot - iPhone 11 - 2020-08-17 at 17 03 15

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,8 +1,8 @@
 4.9
 -----
+- [**] In Settings > Experimental Features, a Products switch is now available for turning Products M3 features on and off for core products (default off for beta testing). Products M3 features: edit grouped, external and variable products, enable/disable reviews, change product type and update categories and tags.
 - [*] Edit Products: the update action now shows up on the product details after updating just the sale price.
 - [*] Variable product > variation list: a warning banner is shown if any variations do not have a price, and warning text is shown on these variation rows.
-- [**] In Settings > Experimental Features, a Products switch is now available for turning Products M3 features on and off for core products (default off for beta testing). Products M3 features: edit grouped, external and variable products, enable/disable reviews, change product type and update categories and tag.
  
 4.8
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 4.9
 -----
 - [*] Edit Products: the update action now shows up on the product details after updating just the sale price.
+- [*] Variable product > variation list: a warning banner is shown if any variations do not have a price, and warning text is shown on these variation rows.
  
 4.8
 -----
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsTopBannerFactory.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsTopBannerFactory.swift
@@ -4,8 +4,8 @@ import UIKit
 ///
 final class ProductVariationsTopBannerFactory {
     static func missingPricesTopBannerView() -> TopBannerView {
-        let viewModel = TopBannerViewModel(title: Constants.title,
-                                           infoText: Constants.info,
+        let viewModel = TopBannerViewModel(title: Localization.title,
+                                           infoText: Localization.info,
                                            icon: Constants.icon,
                                            isExpanded: true,
                                            topButton: .none,
@@ -16,10 +16,13 @@ final class ProductVariationsTopBannerFactory {
 
 private extension ProductVariationsTopBannerFactory {
     enum Constants {
+        static let icon = UIImage.infoOutlineImage
+    }
+
+    enum Localization {
         static let title = NSLocalizedString("Some variations do not have prices",
                                              comment: "Banner title in product variation list top banner when some variations do not have a price")
         static let info = NSLocalizedString("Variations without price will not be shown in your store.",
                                             comment: "Banner caption in my store when the stats will be deprecated")
-        static let icon = UIImage.infoOutlineImage
     }
 }
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsTopBannerFactory.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsTopBannerFactory.swift
@@ -1,0 +1,25 @@
+import UIKit
+
+/// Generates top banner view that is shown at the top of variation list screen when at least one variation is missing a price.
+///
+final class ProductVariationsTopBannerFactory {
+    static func missingPricesTopBannerView() -> TopBannerView {
+        let viewModel = TopBannerViewModel(title: Constants.title,
+                                           infoText: Constants.info,
+                                           icon: Constants.icon,
+                                           isExpanded: true,
+                                           topButton: .none,
+                                           type: .warning)
+        return TopBannerView(viewModel: viewModel)
+    }
+}
+
+private extension ProductVariationsTopBannerFactory {
+    enum Constants {
+        static let title = NSLocalizedString("Some variations do not have prices",
+                                             comment: "Banner title in product variation list top banner when some variations do not have a price")
+        static let info = NSLocalizedString("Variations without price will not be shown in your store.",
+                                            comment: "Banner caption in my store when the stats will be deprecated")
+        static let icon = UIImage.infoOutlineImage
+    }
+}
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -185,7 +185,6 @@ private extension ProductVariationsViewController {
     func updateTopBannerView() {
         let hasVariationsMissingPrice = resultsController.fetchedObjects.contains { $0.price.isEmpty }
         topBannerView.isHidden = hasVariationsMissingPrice == false
-        tableView.layoutIfNeeded()
         tableView.updateHeaderHeight()
     }
 }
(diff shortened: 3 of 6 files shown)


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 4.9
 -----
 - [*] Edit Products: the update action now shows up on the product details after updating just the sale price.
+- [*] Variable product > variation list: a warning banner is shown if any variations do not have a price, and warning text is shown on these variation rows.
```
pmusolino, line 4: Should we remove this since this will be released as part of M3?

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsTopBannerFactory.swift
```diff
@@ -0,0 +1,25 @@
...
+                                           icon: Constants.icon,
+                                           isExpanded: true,
+                                           topButton: .none,
+                                           type: .warning)
+        return TopBannerView(viewModel: viewModel)
+    }
+}
+
+private extension ProductVariationsTopBannerFactory {
+    enum Constants {
```
pmusolino, line 18: nit: maybe it's better to call it `Localization` or `Strings` and move the icon under `enum Constants`

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -179,6 +213,22 @@ private extension ProductVariationsViewController {
...
+
+        resultsController.onDidChangeContent = { [weak tableView] in
+            tableView?.endUpdates()
+            onReload()
+        }
+
+        resultsController.onDidResetContent = {
+            onReload()
+        }
+    }
```
pmusolino, line 231: This looks good, but it got me thinking about what would happen if there were a lot of variations. In that case, the banner would not appear immediately if the first product without the price was beyond page 25, and it would probably also cause a small graphic glitch since the tableview offset would change. Is it right? Or maybe we already know the price for all the variations?

RELEASE-NOTES.txt
jaclync: The warning UI updates https://github.com/woocommerce/woocommerce-ios/issues/2594 aren't behind the M3 feature flag, and this variation list screen is available for old product details as well (tap on a variable product > variations row, just the cells aren't tappable). I think it could be helpful to include this in the release notes

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
jaclync: There's no other API for us to know if a product has variations without a price, we can only tell after fetching variations from the API in pages. Yea it is possible where the variations without a price are not in the first page of results, so the banner could appear after fetching a later page. Here's a gif of a product with 71 variations with the last row as the only variation without a price (feel free to check out the Woo BBQ product on my test site as well):

simulator

I didn't find the offset too jarring, the next page was triggered when viewing the bottom of the results so I felt rows coming in was natural. Lemme know if you think another way to implement the header would be a better UX though!

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsTopBannerFactory.swift
jaclync: updated in https://github.com/woocommerce/woocommerce-ios/pull/2664/commits/8e00aacf957c9b3397edc91b44211fe9b852f1bc

jaclync:
Thanks for review Paolo! I responded to your questions in the comments and below for another look 🙂 

> I just noticed that we continue to receive this console log about a breaking constraint on the banner. It appears in every place where we use it, but it's ok to fix it separately since it was already existing.

Thanks for raising this! There seems to be 2 different types of Auto Layout warnings: one in https://github.com/woocommerce/woocommerce-ios/issues/1549 that is caused by temporary 0px table header view width (`'UIView-Encapsulated-Layout-Width' UIView:0x7fb8dc7510c0.width == 0`) and the other one as you posted which is caused by a product/variation cell's default height as 44px (`UITableViewCellContentView:0x7f8f7bc296a0.height == 44`) whereas the constraints don't satisfy temporarily. It's quite frustrating how Auto Layout gives warnings when the views were at temporary size that isn't visible to the user. I've seen workarounds to adjust the priority of constraint(s) that cause the warning by like the original priority +- 1, but it just felt not so ideal to me. Do you know any other ways to fix warnings at temporary sizes? Maybe it's better to research more about other workarounds and fix them separately.

> Did you try to use this?
> func tableView(_ tableView: UITableView, estimatedHeightForHeaderInSection section: Int) -> CGFloat {
         return CGFloat.leastNormalMagnitude
     }
There is an old tableview bug (if I remember correctly from iOS 11): it must return a tiny value to collapse nil or empty section headers.

Yes I did make sure to return `.leastNormalMagnitude` in case of no warning banner, like other places in the code base! It still had issues though, like when it was hidden before and then set to be shown, the header view height appeared but still no view was shown. After scrolling the header view was shown immediately. Another side effect of using the section header view is that it becomes sticky and I'm not sure if that's intended UX.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
pmusolino: Sounds good, don't worry @jaclync. I think that at the moment with the API at our disposal we have only two ways of improving it:
1. fetch all the variants (iterating through all the pages), but it's not ideal.
2. show the banner always fixed on top, and it will be correct to hide it if the page doesn't contain variants without price. What do you think about this @Garance91540?

pmusolino:
> Do you know any other ways to fix warnings at temporary sizes? Maybe it's better to research more about other workarounds and fix them separately.

I definitely agree. 👍 

> Yes I did make sure to return .leastNormalMagnitude in case of no warning banner, like other places in the code base! It still had issues though, like when it was hidden before and then set to be shown, the header view height appeared but still no view was shown. After scrolling the header view was shown immediately. Another side effect of using the section header view is that it becomes sticky and I'm not sure if that's intended UX.

Oh, that's weird! Maybe in the future, we can move the banner to be part of the datasource, and manage it like a separate section/cell

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
jaclync: I generally avoid fetching all the pages if not really necessary, and I think in this case of warning banner it's fine if it doesn't show up until later pages are fetched.

If a user is looking for a variation to edit, I think they will scroll down to find the exact variation and the row also shows "No price" in warning color in addition to the warning UI in variation details. Also, I don't have the data yet but I'd think the majority of active products have fewer than 25 variations (it feels hard to manage more than 25 individually 🙂 ).

jaclync:
Thanks for the feedback Paolo! I'm merging the PR now, we can still continue the discussion on the paginated variations with G 🙂

PR #2665, opened 2020-08-17
Edit Product: now is possible to change the product type from simple to variant, grouped or external product

pmusolino:
Fixes #2198 

## Description
As part of Products Milestone 3, we want to give users the possibility to change the product type from simple to variant, to grouped or external product.
After [having introduced the Product Type row](https://github.com/woocommerce/woocommerce-ios/pull/2653), in this PR I'll show the action sheet which allows the user to change the product type, and to update it remotely accordingly.

## Changes
- Now in `Product` the `productTypeKey` is one of the fields which will be encoded.
- New image `externalProductImage` + tests.
- Implemented the `ProductTypeBottomSheetListSelectorCommand` which takes care of all the actions that we will show in the action sheet + tests.
- Adapted `ProductFormViewController` to handle the changes after the product type update.
- Removed `updateReviews` method in `ProductFormViewModel` since it wasn’t used.
- Added the `updateProductType` method in `ProductFormViewModelProtocol` and added it’s conformance where required.

#### Case 1: Simple product
1. Open a product that is NOT Simple.
2. Tap the product type cell, and choose "Simple product".
3. Fill the missing fields if necessary, and tap update. 
4. The product should be updated remotely, and every field should be populated correctly.

#### Case 2: Variable product
1. Open a product that is NOT Variable.
2. Tap the product type cell, and choose "Variable product".
3. Fill the missing fields if necessary, and tap update. 
4. The product should be updated remotely, and every field should be populated correctly.

#### Case 3: Grouped product
1. Open a product that is NOT Grouped.
2. Tap the product type cell, and choose "Grouped product".
3. Fill the missing fields if necessary, and tap update. 
4. The product should be updated remotely, and every field should be populated correctly.

#### Case 4: External product
1. Open a product that is NOT External.
2. Tap the product type cell, and choose "External product".
3. Fill the missing fields if necessary, and tap update. 
4. The product should be updated remotely, and every field should be populated correctly.

Feel free to play with other possible combinations.

## Screenshots
| Product Type action sheet             | Product Type action sheet (dark mode enabled) |
:-------------------------:|:-------------------------:
Simulator Screen Shot - iPhone 11 Pro - 2020-08-17 at 12 58 07 | Simulator Screen Shot - iPhone 11 Pro - 2020-08-17 at 12 58 00

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2666, opened 2020-08-17
Edit Product: enhancements to external product URL & link button

### LINKED ISSUE #2609: External product: external URL & link button design enhancements
There are two UX enhancements for editing an external product's external URL & link button from this PR https://github.com/woocommerce/woocommerce-ios/pull/2442:

- local validation for external URL (we have a helper `String+URLValidation` `isValidURL() -> Bool` if that helps)

Screen Shot 2020-08-06 at 4 45 15 PM

- auto-fill button text with "Buy product" if the external URL field is empty in the beginning

pmusolino:
Fixes #2609 

## Description
As part of Milestone 3, in this PR there are two enhancements on External product type. 
In particular:
- Local validation for external URL. If the URL is not valid, an error is shown.
- Auto-fill button text with "Buy product" if the external URL field is empty in the beginning. This will come in handy after that [this PR](https://github.com/woocommerce/woocommerce-ios/pull/2665) will be merged.

## Testing
1. Open an External product.
2. Tap the cell 
3. If the Button Text in the API is empty, "Buy Product" will be shown as text.
4. Try to edit the Product URL adding a non-valid URL. An error will be shown.
5. Try to re-edit the Product URL using a valid URL. The error should disappear. 
 
## Screenshots
| Valid URL             |  Valid URL (dark mode enabled) |  Non-Valid URL |  Non-Valid URL (dark mode enabled) |
:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:
Simulator Screen Shot - iPhone 11 Pro - 2020-08-17 at 16 51 02 | Simulator Screen Shot - iPhone 11 Pro - 2020-08-17 at 16 51 25 | Simulator Screen Shot - iPhone 11 Pro - 2020-08-17 at 16 51 00 | Simulator Screen Shot - iPhone 11 Pro - 2020-08-17 at 16 51 23

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
@@ -25,7 +25,7 @@ final class ProductExternalLinkViewController: UIViewController {
     init(product: Product, onCompletion: @escaping Completion) {
         self.product = product
         self.externalURL = product.externalURL
-        self.buttonText = product.buttonText
+        self.buttonText = product.buttonText.isNotEmpty ? product.buttonText : Strings.buyProductPlaceholder
         self.onCompletion = onCompletion
 
         super.init(nibName: nil, bundle: nil)
@@ -345,7 +345,7 @@ private extension ProductExternalLinkViewController {
                                                        comment: "Title of the text field for editing the button text for an external/affiliate product")
         static let buttonTextFooter = NSLocalizedString("This text will be shown on the button linking to the external product.",
                                                  comment: "Footer text for editing external product button text")
-        static let buyProductPlaceholder = NSLocalizedString("Buy product",
+        static let buyProductPlaceholder = NSLocalizedString("Buy Product",
         comment: "Placeholder of the text field for editing the button text for an external/affiliate product")
     }
 }
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
@@ -136,10 +136,9 @@ private extension ProductExternalLinkViewController {
                                                                  textFieldAlignment: .trailing,
                                                                  onTextChange: { [weak self] text in
                                                                     self?.externalURL = text
-                                                                    if text?.isValidURL() == true {
+                                                                    if text?.isValidURL() == true || text?.isEmpty == true {
                                                                         self?.hideError()
-                                                                    }
-                                                                    else {
+                                                                    } else {
                                                                         self?.displayError(error: Strings.errorMalformedURL)
                                                                     }
         })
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
```diff
@@ -122,24 +127,29 @@ private extension ProductExternalLinkViewController {
...
                                                                  text: externalURL,
                                                                  placeholder: placeholder,
                                                                  keyboardType: .URL,
                                                                  textFieldAlignment: .trailing,
                                                                  onTextChange: { [weak self] text in
                                                                     self?.externalURL = text
+                                                                    if text?.isValidURL() == true {
+                                                                        self?.hideError()
+                                                                    }
+                                                                    else {
```
jaclync, line 142: nit: we usually have `} else {` on the same line :)

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
```diff
@@ -151,11 +161,50 @@ private extension ProductExternalLinkViewController {
     }
 }
 
+// MARK: - UITableViewDelegate Conformance
+//
+extension ProductExternalLinkViewController: UITableViewDelegate {
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+
+    }
```
jaclync, line 169: nit: I think this unimplemented method isn't necessary?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
```diff
@@ -122,24 +127,29 @@ private extension ProductExternalLinkViewController {
...
+        let title = Strings.productURLTitle
+        let placeholder = Strings.enterURLPlaceholder
+        var viewModel = TitleAndTextFieldTableViewCell.ViewModel(title: title,
                                                                  text: externalURL,
                                                                  placeholder: placeholder,
                                                                  keyboardType: .URL,
                                                                  textFieldAlignment: .trailing,
                                                                  onTextChange: { [weak self] text in
                                                                     self?.externalURL = text
+                                                                    if text?.isValidURL() == true {
```
jaclync, line 139: I think we might want to consider empty string a valid case so that we don't block the user from saving an empty URL (which is allowed in core)?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
pmusolino: I forgot always the same thing 😱

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit External Link/ProductExternalLinkViewController.swift
pmusolino: Good catch! Updated here 347a76b49f073ce44476c5bd89310ec57c848b58

pmusolino:
Yes @Garance91540 but I think that we should solve this issue probably as part of another PR, since this is the component that we use all the screens where we use this type of error (the error is near the cell), for example in the SKU screen. 
If we use the same distance like on top, this is the result:

Garance91540:
> I think that we should solve this issue probably as part of another PR, since

Sounds good to me!

Garance91540 (approved):

PR #2668, opened 2020-08-17
RetrieveProductReviewFromNoteUseCase: Use Storage if Possible

shiki:
Ref #2254. 

## Findings

I'm almost done with fixing #2254. I found that handling push notifications using [`RetrieveProductReviewFromNoteUseCase`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Stores/ProductReview/RetrieveProductReviewFromNoteUseCase.swift) can be slow. It would take a while for the Review Details ViewController to be presented. The current implementation makes use of existing records in Core Data so I copied the same strategy here. 

## Solution

The `RetrieveProductReviewFromNoteUseCase` will now use any existing `Note`, `ProductReview`, or `Product` in Core Data and will only fetch them if they do not exist locally yet. 

https://github.com/woocommerce/woocommerce-ios/blob/d1e95de346c6d50287985feeccabfa00ce2795d4/Yosemite/Yosemite/Stores/ProductReview/RetrieveProductReviewFromNoteUseCase.swift#L107-L114

https://github.com/woocommerce/woocommerce-ios/blob/d1e95de346c6d50287985feeccabfa00ce2795d4/Yosemite/Yosemite/Stores/ProductReview/RetrieveProductReviewFromNoteUseCase.swift#L168-L176

https://github.com/woocommerce/woocommerce-ios/blob/105c9c946c56e6cd383c7ff87f968cec1c2110ab/Yosemite/Yosemite/Stores/ProductReview/RetrieveProductReviewFromNoteUseCase.swift#L130-L142

## Testing

This `UseCase` is not currently used. Please review the unit test instead. 🙂 

## Submitter Checklist

### DIFF
--- a/Yosemite/YosemiteTests/Stores/ProductReview/RetrieveProductReviewFromNoteUseCaseTests.swift
+++ b/Yosemite/YosemiteTests/Stores/ProductReview/RetrieveProductReviewFromNoteUseCaseTests.swift
@@ -65,6 +65,62 @@ final class RetrieveProductReviewFromNoteUseCaseTests: XCTestCase {
         XCTAssertEqual(parcel.product, product)
     }
 
+    func test_it_uses_the_existing_Note_in_Storage_if_it_is_available() throws {
+        // Given
+        let useCase = makeUseCase()
+        let note = TestData.note
+        let productReview = TestData.productReview
+        let product = TestData.product
+
+        let storageNote = viewStorage.insertNewObject(ofType: StorageNote.self)
+        storageNote.update(with: note)
+        viewStorage.saveIfNeeded()
+
+        productReviewsRemote.whenLoadingProductReview(siteID: productReview.siteID,
+                                                      reviewID: productReview.reviewID,
+                                                      thenReturn: .success(productReview))
+        productsRemote.whenLoadingProduct(siteID: product.siteID,
+                                          productID: product.productID,
+                                          thenReturn: .success(product))
+
+        // When
+        let result = try retrieveAndWait(using: useCase, noteID: note.noteID)
+
+        // Then
+        XCTAssertTrue(result.isSuccess)
+        XCTAssertEqual(notificationsRemote.invocationCountOfLoadNotes, 0)
+
+        let parcel = try XCTUnwrap(result.get())
+        XCTAssertEqual(parcel.note.noteID, note.noteID)
+    }
+
+    func test_it_uses_the_existing_Product_in_Storage_if_it_is_available() throws {
+        // Given
+        let useCase = makeUseCase()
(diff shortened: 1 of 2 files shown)


Yosemite/YosemiteTests/Stores/ProductReview/RetrieveProductReviewFromNoteUseCaseTests.swift
```diff
@@ -14,27 +14,33 @@ final class RetrieveProductReviewFromNoteUseCaseTests: XCTestCase {
...
+        storageManager = nil
         productsRemote = nil
         productReviewsRemote = nil
         notificationsRemote = nil
 
         super.tearDown()
     }
 
-    func testItFetchesAllEntitiesAndReturnsTheParcel() throws {
+    func test_it_fetches_all_entities_and_returns_the_Parcel() throws {
```
jaclync, line 41: nit: is `parcel` capitalized intentionally?

Yosemite/YosemiteTests/Stores/ProductReview/RetrieveProductReviewFromNoteUseCaseTests.swift
```diff
@@ -59,46 +65,125 @@ final class RetrieveProductReviewFromNoteUseCaseTests: XCTestCase {
...
+        // When
+        let result = try retrieveAndWait(using: useCase, noteID: note.noteID)
+
+        // Then
+        XCTAssertTrue(result.isSuccess)
+        XCTAssertEqual(notificationsRemote.invocationCountOfLoadNotes, 0)
 
-        let useCase = makeUseCase(storage: storage)
+        let parcel = try XCTUnwrap(result.get())
+        XCTAssertEqual(parcel.note.noteID, note.noteID)
```
jaclync, line 94: nit: maybe we can compare the whole `note`? `XCTAssertEqual(parcel.note, note)` if `Note` supports `Equatable`?

Yosemite/YosemiteTests/Stores/ProductReview/RetrieveProductReviewFromNoteUseCaseTests.swift
shiki: I think so. 🙈 I think I wanted it to refer to [`ProductReviewFromNoteParcel`](https://github.com/woocommerce/woocommerce-ios/blob/061c2bce126fa84064d2125c45339be65269ab7a/Yosemite/Yosemite/Stores/ProductReview/RetrieveProductReviewFromNoteUseCase.swift#L8).

PR #2671, opened 2020-08-18
Save the Installation Date 

### LINKED ISSUE #2632: [IAFeedback] Call setInstallationDateIfNecessary when app is started
The [`setInstallationDateIfNecessary`](https://github.com/woocommerce/woocommerce-ios/blob/f0828fe1f156c33d44b37614574f697dedbed35a/Yosemite/Yosemite/Actions/AppSettingsAction.swift#L69-L76) was added in #2556 but it's not integrated yet. 

It should be called when the app is started.

shiki:
Closes #2632. 

## Findings

This got complicated than expected. At first, I thought that I could just simply dispatch the [`AppSettingsAction.setInstallationDateIfNecessary`](https://github.com/woocommerce/woocommerce-ios/blob/74d9f5b92b55fc9fa51336e4d72e1d3c0831bcf7/Yosemite/Yosemite/Actions/AppSettingsAction.swift#L69-L76) in `AppDelegate`. And then I found out that calling that action has no effect if the user is not logged in because `DeauthenticatedState` doesn't handle any actions. 

https://github.com/woocommerce/woocommerce-ios/blob/74d9f5b92b55fc9fa51336e4d72e1d3c0831bcf7/WooCommerce/Classes/Yosemite/DeauthenticatedState.swift#L20-L22

So, I guess we can't really save the installation date until the user is logged in. This is fine because we currently have no need for the installation date to be very accurate. We do use some sort of inference to determine the _oldest_ date that we know:

https://github.com/woocommerce/woocommerce-ios/blob/74d9f5b92b55fc9fa51336e4d72e1d3c0831bcf7/Yosemite/Yosemite/Stores/AppSettings/InAppFeedbackCardVisibilityUseCase.swift#L65-L73

## Solution

With all that said, I opted to call the action in `MainTabBarController` instead of `AppDelegate`:

- Calling it in `AppDelegate` may make us assume that we expect it to be called even for not logged in users. 
- It was easier to have some sort of unit test within `MainTabViewModel`.

## Testing

1. Delete the app from the simulator. 
2. Run and log in.
3. Open the `Documents` directory.
4. Confirm that a file named `general-app-settings.plist` was created. The contents should have an `installationDate` key:

    ```xml
	<plist version="1.0">
	<dict>
		<key>installationDate</key>
		<date>2020-08-18T22:20:20Z</date>
	</dict>
	</plist>
    ```

5. Forcefully close the app. 
6. Delete the `general-app-settings.plist` file. 
7. Run the app again. 
8. Confirm that a `general-app-settings.plist` was created again. 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewModels/MainTabViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/MainTabViewModel.swift
@@ -1,6 +1,7 @@
 import Foundation
 import Yosemite
 
+import class AutomatticTracks.CrashLogging
 
 // MARK: - MainTabViewModel Notifications
 //
@@ -30,6 +31,13 @@ final class MainTabViewModel {
     /// passing the string to be presented in the badge as a parameter
     ///
     var onBadgeReload: ((String?) -> Void)?
+
+    /// Must be called during `MainTabBarController.viewDidAppear`. This will try and save the
+    /// app installation date.
+    ///
+    func onViewDidAppear() {
+        saveInstallationDateIfNecessary()
+    }
 
     /// Bootstrap the data pipeline for the orders badge
     /// Fetches the initial badge count and observes notifications requesting a refresh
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewModels/MainTabViewModel.swift
```diff
@@ -16,11 +17,28 @@ extension NSNotification.Name {
...
     /// passing the string to be presented in the badge as a parameter
     ///
     var onBadgeReload: ((String?) -> Void)?
 
+    /// Must be called during `MainTabBarController.viewDidAppear`. This will try and save the
+    /// app installation date.
+    ///
+    func onViewDidAppear() {
+        saveInstallationDateIfNecessary()
+    }
```
pmusolino, line 40: Since `saveInstallationDateIfNecessary()` is called only on `MainTabBarController.viewDidAppear`, maybe it's better to move the note on top of `saveInstallationDateIfNecessary` function it self, making it public, and remove the `onViewDidAppear()` function? In that way maybe it's more clear what the MainTabBarController is doing, saving the installation date.

pmusolino (approved):
Well done :shipit: 

Code ✅ just a nit but feel free to ignore it if you think it's better to leave it as is.
Testing ✅ everything works as expected.

WooCommerce/WooCommerceTests/ViewModels/MainTabViewModelTests.swift
```diff
@@ -0,0 +1,50 @@
...
+        featureFlagService = MockFeatureFlagService(isInAppFeedbackOn: true)
+    }
+
+    override func tearDown() {
+        featureFlagService = nil
+        storesManager = nil
+        super.tearDown()
+    }
+
+    func test_onViewDidAppear_will_save_the_installation_date() throws {
```
Ecarrion, line 29: I there a way to test this with the `DeauthenticatedState` state? 🤔

shiki: Added in 30830df74bb390147a9974f4f849be46e8f186f5. 🤪

shiki:
Thank you, both! 🙇

WooCommerce/Classes/ViewModels/MainTabViewModel.swift
pmusolino: Make totally sense! Thank you @shiki for explaining it 👍

PR #2680, opened 2020-08-19
Add multiple button support to TopBannerView

### LINKED ISSUE #2646: [IAFeedback] Extend BannerView to support multiple action buttons
The new Product Banner view shows two action buttons at the bottom of the banner which is not currently supported.

Screen Shot 2020-08-12 at 4 24 52 PM

Tasks:
- [x] Modify BannerView to support multiple actions #2680 
- [ ] Modify ProductBanner to include the new CTAs

Ecarrion:
closes #2646 

# Why
For the new products feedback survey, we need to modify the products top banner. Prior this PR the `TopBannerView` class only supported one action button.

# How
- Adds a new `ActionButton` struct inside `TopBannerViewModel` that will represent an action button.
- Update `TopBannerView` to build its view hierarchy based on the new `actionButtons` property from `TopBannerViewModel`.
- Adds some tests.

# Screenshot
This is not in the app, but is only to show you how it would look.
mutile-layout-button

# Testing Steps
Make sure that the current banners(Products and Stats) keep looking as they used to.
* To show the stats banner you may need to replace line 45 of `StatsVersionCoordinator` with
```swift
let nextVersion: StatsVersion = .v3
```

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
+++ b/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
@@ -44,7 +44,13 @@ final class TopBannerView: UIView {
         return button
     }()
 
-    private lazy var actionStackView = createActionStackView()
+    private lazy var actionStackView: UIStackView = {
+        let stackView = UIStackView()
+        stackView.axis = .vertical
+        return stackView
+    }()
+
+    private let actionButtons: [UIButton]
 
     private let isActionEnabled: Bool
 
@@ -58,6 +64,7 @@ final class TopBannerView: UIView {
         isExpanded = viewModel.isExpanded
         onAction = viewModel.actionHandler
         onTopButtonTapped = viewModel.topButton.handler
+        actionButtons = viewModel.actionButtons.map { _ in UIButton() }
         super.init(frame: .zero)
         configureSubviews(with: viewModel)
     }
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
```diff
@@ -108,7 +102,10 @@ private extension TopBannerView {
 
         iconImageView.image = viewModel.icon
 
-        actionButton.setTitle(viewModel.actionButtonTitle, for: .normal)
+        zip(viewModel.actionButtons, actionButtons).forEach { buttonInfo, button in
```
jaclync, line 105: cool use of `zip`! ⚡

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
```diff
@@ -179,10 +177,34 @@ private extension TopBannerView {
...
+    func configureActionStackView() {
+        // StackView to hold the action buttons
+        let buttonsStackView = UIStackView()
+        buttonsStackView.distribution = .fillEqually
+        buttonsStackView.spacing = 0.5
+
+        // Background to simulate a separator by giving the buttons some spacing
+        let separatorBackground = createButtonsBackgroundView()
+        buttonsStackView.addSubview(separatorBackground)
+        buttonsStackView.pinSubviewToAllEdges(separatorBackground)
```
jaclync, line 189: this is a cool way of showing the separators between arranged subviews :)

jaclync (approved):
LGTM! 🚢 

Code: ✅ 
Testing: ✅

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
Ecarrion: I think it's crazy that StackView don't support this by default 😅

Ecarrion:
Thanks @jaclync for the review 🙇

Issue #2548, opened 2020-07-22
[IAFeedback] Add Tracking


shiki:
Add Tracks events and track them when the following scenarios happen. 

### Prompt

This prompt is shown only after the user has passed some conditions (e.g. 3 months usage). See p91TBi-2Af-p2 for the conditions. 

<img src="https://user-images.githubusercontent.com/198826/88236893-4f12f280-cc3b-11ea-8308-ab4359e809f3.png" width="240">

Event Name | Trigger | Custom Properties
--------|-------|--
`app_feedback_prompt`  |  When the “Enjoying the WooCommerce app?” and the buttons are displayed to the user. | `action="shown"`
`app_feedback_prompt` | When the **I Like It** button is activated. | `action="liked"`
`app_feedback_prompt` | When the **Could Be Better** button is activated. | `action="didnt_like"`

### Survey Screen

This screen is shown after activating the **Could Be Better** button. 

<img src="https://user-images.githubusercontent.com/198826/88237024-9600e800-cc3b-11ea-9f7b-565cf80a98b9.png" width="240">

Event Name | Trigger | Custom Properties
--------|-------|---
`survey_screen` | When the screen is shown. | `context="general"` <br> `action="opened"`
`survey_screen` | When the user closes the screen without finishing the survey.  | `context="general"` <br> `action="canceled"`

### Survey Completion Screen

This screen is shown when the user successfully submitted the answers to the survey.

<img src="https://user-images.githubusercontent.com/198826/88237071-ae710280-cc3b-11ea-830a-542260ef3e61.png" width="240">

Event Name | Trigger | Custom Properties
--------|-------|---
`survey_screen` | When this survey completion screen is shown.  | `context="general"` <br> `action="completed"`

### Banner

This banner is shown for WIP features. The Products feature is just an example and we might have similar banners for other features in the future. 

<img src="https://user-images.githubusercontent.com/198826/88230329-75329580-cc2f-11ea-9b74-19fb4bc0381d.png" width="240">

Event Name | Trigger | Custom Properties
--------|-------|--
`feature_feedback_banner` | When the **Give Feedback** button is activated. | `context="products_m3"` <br> `action="gave_feedback"`
`feature_feedback_banner` | When the **Dismissed** button is activated. | `context="products_m3"` <br> `action="dismissed"`

### Survey

[...]

PR #2687, opened 2020-08-20
Statically-Typed Event Properties Convention and In-app Feedback Card Tracking

shiki:
Ref #2548. 

## Changes

This adds two things.  

### Statically-Typed Event Properties Convention

I thought about incorporating @diegoreymendez' analytics tracking improvements (paNNhX-98-p2) into our codebase. Today seems to be a good day to implement it. Thanks to @diegoreymendez for the advice and for allowing me to _borrow_ the code. 🤪 

I believe Diego has explained in the post what sort of problems this solves so I won't explain that here. But let me know if you have any questions. 

A small difference in our implementation is I used `WooAnalyticsStat` instead of `String`:

https://github.com/woocommerce/woocommerce-ios/blob/2b2d41459d0eaeb551daf016f35e7e8144e0b9d6/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift#L40-L43

It is implemented like this in [WordPress-iOS-Shared](https://github.com/wordpress-mobile/WordPress-iOS-Shared/blob/fcfef1e440509d92514118d91631708ee67ebdaa/WordPressShared/Core/Analytics/AnalyticsEvent.swift#L41-L49):

```swift
public final class AnalyticsEvent {
    public let name: String
    public let properties: [String: String]
}
```

I wanted to keep the usage of `WooAnalyticsStat` for now because of Aaron's concern in paNNhX-98-p2#comment-283. I documented the rationale in `WooAnalyticsStat`:

https://github.com/woocommerce/woocommerce-ios/blob/2b2d41459d0eaeb551daf016f35e7e8144e0b9d6/WooCommerce/Classes/Analytics/WooAnalyticsStat.swift#L6-L9

### In-app Feedback Card Events

To test this convention, I added tracking for the In-app Feedback Card. 

Event Name | Trigger | Custom Properties
--------|-------|--
`app_feedback_prompt`  |  When the “Enjoying the WooCommerce app?” and the buttons are displayed to the user. | `action="shown"`
`app_feedback_prompt` | When the **I Like It** button is activated. | `action="liked"`
`app_feedback_prompt` | When the **Could Be Better** button is activated. | `action="didnt_like"`

## Testing

1. Delete the app from the device to reset any existing state. 
2. Run the app.
3. Navigate to Orders tab. 
4. Time travel to more than 3 months from now. 
5. Go back to the app and navigate to My Store. The In-app Feedback Card should be visible.
6. Confirm that you can see this log in the console:

    ```
    🔵 Tracked app_feedback_prompt, properties: [AnyHashable("action"): "shown", ...
    ```
7. Tap on the Could Be Better button. 
8. Confirm that you can see this log in the console:

    ```
    🔵 Tracked app_feedback_prompt, properties: [AnyHashable("action"): "didnt_like", ...
    ```
9. Navigate to Orders tab.
10. Time travel to 6 months from the current device's date. 
11. Go back to the app and navigate to My Store. The In-app Feedback Card should be visible again.
12. Tap on the I Like It button. 
13. Confirm that you can see this log in the console:
    ```
    🔵 Tracked app_feedback_prompt, properties: [AnyHashable("action"): "liked", ...
    ```
14. Don't forget to travel back to the present. We need you here! 🤪 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
@@ -1,32 +1,33 @@
 import Foundation
 
 /// This struct represents an analytics event.
+///
 /// Declaring this class as final is a design choice to promote a simpler usage and implement events
 /// through parametrization of the `name` and `properties` properties.
 ///
 /// An example of a static event definition (in the client App or Pod):
 ///
 /// ~~~
-/// extension AnalyticsEvent {
-///     static let loginStart = AnalyticsEvent(name: "login", properties: ["step": "start"])
+/// extension WooAnalyticsEvent {
+///     static let loginStart = WooAnalyticsEvent(name: "login", properties: ["step": "start"])
 /// }
 /// ~~~
 ///
 /// An example of a dynamic / parametrized event definition (in the client App or Pod):
 ///
 /// ~~~
-/// extension AnalyticsEvent {
+/// extension WooAnalyticsEvent {
 ///     enum LoginStep: String {
 ///         case start
 ///         case success
 ///     }
 ///
-///     static func login(step: LoginStep) -> AnalyticsEvent {
+///     static func login(step: LoginStep) -> WooAnalyticsEvent {
 ///         let properties = [
 ///             "step": step.rawValue
 ///         ]
 ///
-///         return AnalyticsEvent(name: "login", properties: properties)
+///         return WooAnalyticsEvent(name: "login", properties: properties)
 ///     }
 /// }
 /// ~~~
@@ -38,7 +39,7 @@ import Foundation
 /// WPAnalytics.track(.loginStart)
 /// ~~~
 ///
-public final class AnalyticsEvent {
+public final class WooAnalyticsEvent {
     let name: String
     let properties: [String: String]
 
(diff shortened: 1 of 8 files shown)


WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
```diff
@@ -0,0 +1,88 @@
...
+        WooAnalyticsEvent(statName: .appFeedbackPrompt, properties: ["action": action.rawValue])
+    }
+
+    static func surveyScreen(context: FeedbackContext, action: SurveyScreenAction) -> WooAnalyticsEvent {
+        WooAnalyticsEvent(statName: .surveyScreen, properties: ["context": context.rawValue, "action": action.rawValue])
+    }
+
+    static func featureFeedbackBanner(context: FeedbackContext, action: FeatureFeedbackBannerAction) -> WooAnalyticsEvent {
+        WooAnalyticsEvent(statName: .featureFeedbackBanner, properties: ["context": context.rawValue, "action": action.rawValue])
+    }
```
shiki, line 87: These ones are currently not used. I just added them so we can see how this will look like. 🙂

PR #2689, opened 2020-08-21
Removed some unused code all around the app

pmusolino:
As part of p91TBi-39Q I continued to remove some unused code from our codebase.
I found some unused properties or entire extensions where the functions inside are no more used.

## Testing
- Run the app (if one of the properties/methods is used, you will encounter a compile error)
- Just CI

Update release notes:


Ecarrion (approved):
Love to see code go away! 🔪 

There is also a lot of `equatable` conformances that we can get for free with Swift. I guess we should remove those at some point too!

pmusolino:
Thanks @Ecarrion for the review. You are right, it will be good to remove all the `equatable` conformance that we get for free. 👍

PR #2692, opened 2020-08-21
[IAFeedback] Track Survey Screen Events

shiki:
Part of #2548.

This adds tracking for these events:

Event Name | Trigger | Custom Properties
--------|-------|---
`survey_screen` | When the survey screen is shown. | `context="general"` <br> `action="opened"`
`survey_screen` | When the user closes the survey screen without finishing the survey.  | `context="general"` <br> `action="canceled"`
`survey_screen` | When this survey completion screen is shown.  | `context="general"` <br> `action="completed"`

## Notes

I could not make a unit test for the `canceled` event because I could not find a way to make `isBeingDismissed` return `true` in a unit test. 

https://github.com/woocommerce/woocommerce-ios/blob/7b0ffcd8fa64a7f92c00b95d65a875aa04360ef3/WooCommerce/Classes/ViewRelated/Survey/SurveyCoordinatingController.swift#L40-L42

Please let me know if you have any ideas! 

## Testing

1. Delete the app from the device to reset any existing state. 
2. Run the app.
3. Navigate to Orders tab. 
4. Time travel to more than 3 months from now. 
5. Go back to the app and navigate to My Store. The In-app Feedback Card should be visible.
6. Tap on Could Be Better.
6. Confirm that you can see this log in the console:

    ```
    Tracked survey_screen, properties: [AnyHashable("action"): "opened", AnyHashable("context"): "general", ...]
    ```
8. Dismiss the survey window.
9. Confirm that you can see this log in the console:

    ````
    Tracked survey_screen, properties: [AnyHashable("context"): "general", AnyHashable("action"): "canceled", ...]
    ````
9. Navigate to Orders tab.
10. Time travel to 6 months from the current device's date. 
11. Go back to the app and navigate to My Store. The In-app Feedback Card should be visible again.
12. Tap on the Could Be Better button again.
13. Fill in and submit the survey.
13. Confirm that you can see this log in the console:
    ```
    Tracked survey_screen, properties: [AnyHashable("action"): "completed", AnyHashable("context"): "general", ...]
    ```

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Survey/SurveyCoordinatingController.swift
+++ b/WooCommerce/Classes/ViewRelated/Survey/SurveyCoordinatingController.swift
@@ -12,15 +12,19 @@ final class SurveyCoordinatingController: WooNavigationController {
 
     private let analytics: Analytics
 
+    /// What kind of survey to present.
+    private let survey: SurveyViewController.Source
+
     init(survey: SurveyViewController.Source,
          zendeskManager: ZendeskManagerProtocol = ZendeskManager.shared,
          viewControllersFactory: SurveyViewControllersFactoryProtocol = SurveyViewControllersFactory(),
          analytics: Analytics = ServiceLocator.analytics) {
+        self.survey = survey
         self.zendeskManager = zendeskManager
         self.viewControllersFactory = viewControllersFactory
         self.analytics = analytics
         super.init(nibName: nil, bundle: nil)
-        startSurveyNavigation(survey: survey)
+        startSurveyNavigation()
     }
 
     required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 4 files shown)


jaclync (approved):
> I could not make a unit test for the canceled event because I could not find a way to make isBeingDismissed return true in a unit test.

since `receivedSurveyFinishedEvent` is default to be `false`, maybe we can call the mock survey VC's `viewDidDisappear` if that also triggers the VC's `isBeingDismissed`? if I understand correctly, the `canceled` event is when `receivedSurveyFinishedEvent` isn't set to `true`

and thanks for adding the unit tests!

Code: ✅ 
Testing: ✅ (I wasn't able to see the in-app feedback banner again after dismissing the banner for the first time, going to the Orders tab, and updating the device time to 3+ months from the last time. I reinstalled the app again and updated the time to 3+ months from now to test the second scenario `completed`)

shiki:
> since receivedSurveyFinishedEvent is default to be false, maybe we can call the mock survey VC's viewDidDisappear if that also triggers the VC's isBeingDismissed?

Thanks for the nudge, @jaclync. I tried again and I found a way by using a `UIWindow` → 8ed59db4f7019a31282ff51b1af5aa01fd43e8a3.

PR #2695, opened 2020-08-24
Fix Build Failure and Clarify FeedbackContext Event Argument

shiki:
Ref #2548. 

This fixes the broken `develop` build. 

I also realized that the `FeedbackContext.products` was too ambiguous. It was meant to be specific for the type of feedback message (e.g. banner) we were presenting. If we have a new feedback request for products, we might name that as `products_m4` or `products_creation`. Leaving this `FeedbackContext.products` as is would just confuse us in the future. 😬 

## Testing

## Submitter Checklist


Ecarrion (approved):
👍

PR #2110, opened 2020-04-08
shouldPop: delegate method is not called on simulators with iOS 13.4

pmusolino:
Resolve #2089 

Sometimes, working on the bugs that Apple creates is really frustrating.
This time, I lost a lot of time investigating on why the `shouldPop:` delegate method is no more called on iOS 13.4. The curious thing, after [investigating it](https://a8c.slack.com/archives/C6H8C3G23/p1585927493166300), is that the similar delegate `didPop:` continue to be called without any issue. This created a lot of issues on our side because we use a method called `shouldPopOnBackButton:` to present the Discard Changes Action Sheet in Products.

After a lot of tests, I discovered that it seems an issue that is present only on the iOS simulator using iOS 13.4. In fact, testing it with the help of @rachelmcr on a real device with iOS 13.4 seems that everything works properly. I also tested the issue on my device with iOS 13.4.1, and everything continues to work as expected.

## Changes
In this PR, just to remind us that maybe something may not work as it should, I added a comment.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2705, opened 2020-08-25
Edit Product: move price settings unsaved changes logic to view model

jaclync:
Part of #1985 

## Changes

- Pass unsaved changes boolean from `ProductPriceSettingsViewModel` to `ProductPriceSettingsViewController` to `ProductFormViewController`
- Minor polishes in `ProductFormViewController`

## Testing

Please sanity check on editing price settings for a product:

- Go to the Products tab
- Tap on a simple product
- Tap on the price settings row
- Tap "Done" without editing any fields --> "Update" CTA should not appear in the navigation bar, and should see ` 🔵 Tracked product_price_settings_done_button_tapped` with `has_changed_data: false`
- Tap on the price settings row again
- Edit any one of the price fields and tap "Done" --> "Update" CTA should appear in the navigation bar, and should see ` 🔵 Tracked product_price_settings_done_button_tapped` with `has_changed_data: true`

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
@@ -793,13 +793,14 @@ private extension ProductFormViewController {
 private extension ProductFormViewController {
     func editPriceSettings() {
         let priceSettingsViewController = ProductPriceSettingsViewController(product: product) { [weak self]
-            (regularPrice, salePrice, dateOnSaleStart, dateOnSaleEnd, taxStatus, taxClass) in
+            (regularPrice, salePrice, dateOnSaleStart, dateOnSaleEnd, taxStatus, taxClass, hasUnsavedChanges) in
             self?.onEditPriceSettingsCompletion(regularPrice: regularPrice,
                                                 salePrice: salePrice,
                                                 dateOnSaleStart: dateOnSaleStart,
                                                 dateOnSaleEnd: dateOnSaleEnd,
                                                 taxStatus: taxStatus,
-                                                taxClass: taxClass)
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -717,6 +716,19 @@ extension ProductFormViewController: KeyboardScrollable {
     }
 }
 
+// MARK: - Navigation actions handling
+//
+private extension ProductFormViewController {
```
Ecarrion, line 721: I did not see this being called when navigating back from a product with unsaved changes. Is that expected?

Ecarrion (approved):
Looks good! 💪 
Left only one question!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: @Ecarrion did you test on an iOS 13.4+ simulator/device? there's an unfortunate bug from Apple that Paolo looked into in https://github.com/woocommerce/woocommerce-ios/pull/2110. if you tried on an iOS 12 simulator or a TestFlight build, the discard changes alert is expected to show up 🙂

jaclync:
Thanks for the review Ernesto, I responded to your question and am merging the PR now!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
Ecarrion: Oh yeah, I did test it in iOS 13.6.
Thanks for the heads up about that issue :-)

PR #2710, opened 2020-08-25
TopBannerView: Fix button layout on large accessibility traits

### LINKED ISSUE #2698: [IAFeedback] Products Banner buttons clipped on Large Font
Parent: #2684 

The products banner clips the "give feedback" and "dismiss" button when the user uses a large enough font

banner-max-2

Ecarrion:
fixes #2698 

# Why

Prior to this PR the `TopBannerView` class didn't support well multiple action buttons on bigger accessibility traits.

banner-max-2

# How

I tried to make the buttons to spawn over multiple lines but it didn't look good due to how letters are wrapped.
accessible

In the end, I decided to modify the button layout from horizontal to vertical when an accessibility trait is used.

accessible-buttons

# Testing Steps
- Delete the app to erase saved settings
- Set the bigger text size on your iPhone 
- Open the app and navigate to the products tab
- Expand the top banner
- See that the buttons are laid out correctly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
+++ b/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
@@ -44,6 +44,9 @@ final class TopBannerView: UIView {
         return stackView
     }()
 
+    // Needed to change the axis on larger accesibility traits
+    private let buttonsStackView = UIStackView()
+
     private let actionButtons: [UIButton]
 
     private let isActionEnabled: Bool
@@ -80,6 +83,7 @@ private extension TopBannerView {
 
         renderContent(of: viewModel)
         configureBannerType(type: viewModel.type)
+        configureStackViewsAxis()
     }
 
     func renderContent(of viewModel: TopBannerViewModel) {
@@ -175,7 +179,6 @@ private extension TopBannerView {
 
     func configureActionStackView(with viewModel: TopBannerViewModel) {
         // StackView to hold the action buttons
-        let buttonsStackView = UIStackView()
         buttonsStackView.distribution = .fillEqually
         buttonsStackView.spacing = 0.5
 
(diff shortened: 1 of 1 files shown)


Garance91540 (approved):
👏 thanks for making sure our app is accessible!!

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
```diff
@@ -44,6 +44,9 @@ final class TopBannerView: UIView {
         return stackView
     }()
 
+    // Needed to change the axis on larger accesibility traits
```
jaclync:
```suggestion
    // Needed to change the axis on larger accessibility traits
```

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
```diff
@@ -175,7 +179,6 @@ private extension TopBannerView {
 
     func configureActionStackView(with viewModel: TopBannerViewModel) {
         // StackView to hold the action buttons
```
jaclync: nit: maybe this comment could be moved to the property?

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
```diff
@@ -236,6 +239,12 @@ private extension TopBannerView {
             return .warningBackground
         }
     }
+
+    /// Changes the axis of the stack views that need speacial treatment on larger size categories
```
jaclync:
```suggestion
    /// Changes the axis of the stack views that needs special treatment on larger size categories
```

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
```diff
@@ -236,6 +239,12 @@ private extension TopBannerView {
...
+    ///
+    func configureStackViewsAxis() {
```
jaclync: super nit: I usually call `update*` if a function might be called more than once after initial setup, to differentiate from `configure*` for one-time initialization. up to you!

jaclync (approved):
This is really thoughtful, thanks for the accessibility enhancement!

Code: ✅ 
Testing: ✅ (tried landscape as well)

WooCommerce/Classes/ViewRelated/Top Banner/TopBannerView.swift
Ecarrion: You are right! :-)

Ecarrion:
Thanks for the review @jaclync :-)

PR #2717, opened 2020-08-26
Edit Product: support for multiline product title

### LINKED ISSUE #2603: Product Details > Support multiline for product Title
I realize that we don't support multiline on iOS for product titles on the product page, so the only place a user can read the full name of their product with long name is on the product list. Can we display the full product title on the product page? 

(Feel free to move this to M4 if necessary!)

pmusolino:
Fixes #2603 

## Description
From now we show the entire title on a Product Detail screen since the cell now contains a text view and no more a text field. I edited the cell used to be a `TextViewTableViewCell`, and I refactored it introducing the `ViewModel` struct for the configuration of the cell. In that way, all the properties are private now. I also adapted all the code where `TextViewTableViewCell` was used before. I also fixed the placeholder showed inside `EnhancedTextView` because it was not correctly aligned.

## Testing
1. Navigate to the product list.
2. Open a product detail with a long title. -> The title should be shown entirely.
3. Try to edit it. -> The cell will be automatically expanded based on the content of the product title.
4. Open a product without a title. -> The placeholder should appear on the cell, and it will disappear when you try to edit it.

#### Extra testing 1 (sanity check):
1. Open an order detail.
2. Try to add a new order note. -> The cell note appearance should appear like before.

#### Extra testing 2 (sanity check):
1. Open a product detail.
2. Open product settings.
3. Tap “Purchase note”. -> The cell appearance should appear like before.

## Gif

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
@@ -139,13 +139,13 @@ private extension NewNoteViewController {
                           comment: "Spoken accessibility label for an icon image that indicates it's a private note and is not seen by the customer.")
         let cellViewModel = TextViewTableViewCell.ViewModel(icon: .asideImage,
                                                             iconAccessibilityLabel: iconAccessibilityLabel,
-                                        iconTint: isCustomerNote ? .primary : .textSubtle,
-                                        onTextChange: { [weak self] (text) in
-                                            self?.navigationItem.rightBarButtonItem?.isEnabled = !text.isEmpty
-                                            self?.noteText = text
+                                                            iconTint: isCustomerNote ? .primary : .textSubtle,
+                                                            onTextChange: { [weak self] (text) in
+                                                                self?.navigationItem.rightBarButtonItem?.isEnabled = !text.isEmpty
+                                                                self?.noteText = text
         })
 
-        cell.configure(viewModel: cellViewModel)
+        cell.configure(with: cellViewModel)
     }
 
     private func setupEmailCustomerCell(_ cell: UITableViewCell) {
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
@@ -132,18 +132,20 @@ private extension NewNoteViewController {
             fatalError()
         }
 
-        cell.iconImage = .asideImage
-        cell.iconTint = isCustomerNote ? .primary : .textSubtle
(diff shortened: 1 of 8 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
```diff
@@ -132,18 +132,20 @@ private extension NewNoteViewController {
...
-            self?.noteText = text
-        }
+        let iconAccessibilityLabel = isCustomerNote ?
+        NSLocalizedString("Note to customer",
+                          comment: "Spoken accessibility label for an icon image that indicates it's a note to the customer.") :
+        NSLocalizedString("Private note",
+                          comment: "Spoken accessibility label for an icon image that indicates it's a private note and is not seen by the customer.")
+        let cellViewModel = TextViewTableViewCell.ViewModel(icon: .asideImage,
+                                                            iconAccessibilityLabel: iconAccessibilityLabel,
+                                        iconTint: isCustomerNote ? .primary : .textSubtle,
```
Ecarrion: nit: This looks like it's not properly indented!

WooCommerce/Classes/ViewRelated/ReusableViews/TextViewTableViewCell.swift
```diff
@@ -37,8 +62,60 @@ class TextViewTableViewCell: UITableViewCell {
 
         noteIconButton.accessibilityTraits = .image
     }
+
+    func configure(viewModel: ViewModel) {
```
Ecarrion: nit: `configure(with viewModel: ViewModel)` seems more aligned with swift naming guidelines!

WooCommerce/Classes/ViewRelated/ReusableViews/TextViewTableViewCell.swift
```diff
@@ -1,14 +1,39 @@
...
+
+    /// Constraints
+    @IBOutlet private weak var textViewHeightConstraint: NSLayoutConstraint!
+    @IBOutlet private weak var bottomContraint: NSLayoutConstraint!
+    @IBOutlet private weak var trailingContraint: NSLayoutConstraint!
+    @IBOutlet private weak var leadingContraint: NSLayoutConstraint!
+    @IBOutlet private weak var topContraint: NSLayoutConstraint!
 
-    @IBOutlet var noteTextView: EnhancedTextView!
+    private var viewModel: ViewModel?
```
Ecarrion: [Q] Is it needed to for us to store the view model? I don't think views should store this extra state, as it could get outdated easily. EG: internal UIKit state changes but the view model doesn't.

Ecarrion (approved):
Looks good @pmusolino 
Just left a minor concern!

pmusolino:
Thank you so much @Ecarrion for your review! I applied your suggestions here 21f0cbb1e63bfc302c44405ea83f9fced58fbcb1

🙇

Garance91540 (approved):
👏

PR #2723, opened 2020-08-27
Products: fix the variations warning banner when one variation has no price but is disabled

### LINKED ISSUE #2708: Products: Can't tell which variation is missing a price if variations are disabled
**Describe the bug**

If I have a variable product and one or more variations are missing the price, I expect to see a banner at the top of the variations list and "No price set" on the variation(s) missing a price.

However, if the variation without a price is also disabled, the "No price set" label doesn't appear. It only says "Disabled." If more than one variation is disabled, I can't tell which variation is missing the price. (Also, I'm not sure if it matters/if the banner warning is needed in that case, since I've intentionally disabled the variation with the missing price.)

**To Reproduce**

1. In wp-admin in a browser, select a variable product to edit.
2. In the Variations section, remove the price from a variation and disable it (uncheck the Enabled setting).
3. In the app, open the Products tab.
4. Select that variable product.
5. Open the Variations section to view the variations list.

Notice that the warning banner appears, but the variation without a price only says "Disabled."

**Screenshots**

In this screenshot, one of the disabled variations has a price and the other doesn't:

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 13.6
- WooCommerce iOS version: 4.9.0.0 TestFlight beta

jaclync:
Fixes #2708

## Changes

- For variations price warning banner visibility, updated to check `isEnabledAndMissingPrice` instead of just whether the price is empty. It is not a concern if a variation doesn't have a price when it's disabled.

## Testing

Prerequisite: the store has a variable product whose variations all have a price except for one, and this variation is disabled ("Enabled" switch is off)

- Go to the Products tab
- Tap on the variable product in the prerequisite
- Tap on the variations row --> there should not be a warning banner (on `develop`, a warning banner shows up in this case. if you turn on the "Enabled" switch for the disabled variation that doesn't have a price, the warning banner should show up)

## Example screenshots

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-08-27 at 13 52 37 | Simulator Screen Shot - iPhone 11 - 2020-08-27 at 13 47 46

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -183,7 +183,12 @@ private extension ProductVariationsViewController {
     }
 
     func updateTopBannerView() {
-        let hasVariationsMissingPrice = resultsController.fetchedObjects.contains { $0.price.isEmpty }
+        let hasVariationsMissingPrice = resultsController.fetchedObjects.contains {
+            EditableProductVariationModel(productVariation: $0,
+                                          allAttributes: allAttributes,
+                                          parentProductSKU: parentProductSKU)
+                .isEnabledAndMissingPrice
+        }
         topBannerView.isHidden = hasVariationsMissingPrice == false
         tableView.updateHeaderHeight()
     }


pmusolino (approved):
:shipit: 

Code ✅ 
Testing ✅

PR #2686, opened 2020-08-20
Launch product feedback survey

Ecarrion:
closes #2648

# Why
In order to complete the product feedback survey flow this PR:
- Adds product feedback survey URL
- Updates products banner to include "Give Feedback" and "Dismiss" action button
- Launch the product survey when tapping on the "Give Feedback" CTA
- Dismiss product banner when tapping on the "Dismiss" CTA
- Update `ProductsViewController` to query the product feedback visibility to know if the banner has to be shown or hidden.

# Screenshots

Dismiss | Give Feedback
---- | ----
dismiss | give-feedback

# Testing Steps

## Give Feedback
- Delete the app to reset any saved settings
- Open the app and login into a store
- Navigate to products
- Tap on the collapsed products banner
- Tap on "Give Feedback" and see that the product feedback is presented
- Complete the feedback and see that the product banner is hidden
- Relaunch the app and see that the product banner is not shown

## Dismiss
- Delete the app to reset any saved settings
- Open the app and login into a store
- Navigate to products
- Tap on the collapsed products banner
- Tap on "Dismiss" and see that the banner is dismissed
- Relaunch the app and see that the product banner is not shown

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2728, opened 2020-08-27
Fixes issue where products filter bar dissapears

### LINKED ISSUE #2720: [IAFeedback] Dismissing Product Banner also dismisses the Product filters
To reproduce:

1. Delete the app from the device.
2. Run and navigate to Products.
3. Tap on Give Feedback. 
4. Dismiss the survey URL.
5. Notice that the product filters are no longer visible. Only the list is visible.

Ecarrion:
fixes #2720 

# Why 

While trying to [dismiss the product banner](https://github.com/woocommerce/woocommerce-ios/pull/2686#issuecomment-678276742) I introduced a bug that hides the products filter bar as well 😇 

# How

- Do not `nil` the table header view when trying to dismiss the top banner
- Adjust the spacing of the outer container stack view(that serves as the header view for the table view) to be `0` when it does not have any top banner presented, to remove an unwanted separator space.

# Gif
Before | After
---- | ----
 | 

# Testing Steps
- Delete the app from the device.
- Run and navigate to Products.
- Tap on Give Feedback. 
- Dismiss the survey URL.
- Make sure that the products filter bar is still visible

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
@@ -39,7 +39,7 @@ final class ProductsViewController: UIViewController {
         let subviews = [topBannerContainerView, toolbar]
         let stackView = UIStackView(arrangedSubviews: subviews)
         stackView.axis = .vertical
-        stackView.spacing = 8
+        stackView.spacing = Constants.headerViewSpacing
         stackView.translatesAutoresizingMaskIntoConstraints = false
         return stackView
     }()
@@ -171,7 +171,7 @@ final class ProductsViewController: UIViewController {
     override func viewDidLayoutSubviews() {
         super.viewDidLayoutSubviews()
 
-        updateTableHeaderView()
+        updateTableHeaderViewHeight()
     }
 }
 
@@ -388,7 +388,7 @@ private extension ProductsViewController {
         ProductsTopBannerFactory.topBanner(isExpanded: isExpanded,
                                            isInAppFeedbackFeatureEnabled: isInAppFeedbackEnabled,
                                            expandedStateChangeHandler: { [weak self] in
-            self?.updateTableHeaderView()
+            self?.updateTableHeaderViewHeight()
         }, onGiveFeedbackButtonPressed: { [weak self] in
             self?.presentProductsFeedback()
         }, onDismissButtonPressed: { [weak self] in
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -388,22 +388,28 @@ private extension ProductsViewController {
...
         topBannerView?.removeFromSuperview()
         topBannerView = nil
-        tableView.tableHeaderView = nil
+        updateTableHeaderView()
+    }
+
+    /// Updates table header view with the correct spacing / edges depending if `topBannerContainerView` is empty or not.
+    ///
+    func updateTableHeaderView() {
+        topStackView.spacing = topBannerContainerView.subviews.isNotEmpty ? 8 : 0
```
shiki: Where do these magic numbers come from? I guess this is from this declaration right? 

https://github.com/woocommerce/woocommerce-ios/blob/1c0bb1cfbb3d99a4f42efc7aaadab1ab1d9f8d2a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift#L42

If so, should we consider moving `8` to a constant (enum)?

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -388,22 +388,28 @@ private extension ProductsViewController {
...
     func hideTopBannerView() {
         topBannerView?.removeFromSuperview()
         topBannerView = nil
-        tableView.tableHeaderView = nil
+        updateTableHeaderView()
+    }
+
+    /// Updates table header view with the correct spacing / edges depending if `topBannerContainerView` is empty or not.
+    ///
+    func updateTableHeaderView() {
```
shiki: Do you think `updateTableHeaderViewHeight` would be less ambiguous? 

```suggestion
    func updateTableHeaderViewHeight() {
```

shiki (approved):
LGTM! Thanks for fixing this. 

I just have a couple of nitpicks. 🙃

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
Ecarrion: I had it like that, but since it does a little bit more than updating just the height I left it as `updateTableHeaderView()` I also had it as `updateTableHeaderViewHeightAndSpacing()` 🤔

shiki: I guess we can consider changing the spacing as changing the "height" too. 🤣 Anyway, up to you!

PR #2743, opened 2020-08-31
Order List: added "Guest" placeholder in order cell when there is no customer name

### LINKED ISSUE #2739: [Order Details] Add "Guest" placeholder on Order List when there's no customer name
On Order List, there's currently a blank label when there's no customer name. Let's add "Guest" instead, like did in Order Detail.

pmusolino:
Fixes #2739 

## Description
On Order List, there's currently a blank label when there's no customer name. In this PR I added "Guest" instead, like did in Order Detail.

## Testing
1. Select the tab "Order"
2. All the orders without customer name and last name should show "Guest".

## Screenshots
91700140-afb81800-eb75-11ea-9515-3540d8ae1719 | Simulator Screen Shot - iPhone 11 Pro - 2020-08-31 at 12 05 27

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
@@ -128,7 +128,7 @@ private extension OrderTableViewCell {
     /// For example, #560 Pamela Nguyen
     ///
     func title(for order: Order) -> String {
-        if let billingAddress = order.billingAddress {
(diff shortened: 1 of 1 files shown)


Ecarrion (approved):
Looks good!
Should we add a unit test for that? 🤔

Garance91540 (approved):
👏

pmusolino:
@Ecarrion I don't think they are useful in this specific case, since the Localization is involved, but let me know if you think we should add them and I'll take care of it in another PR.

pmusolino:
Thank you both for your review! 😉

Issue #2553, opened 2020-07-23
[IAFeedback] Replace Feature Request with Send Feedback in Settings


shiki:
Discussed in CGPNUU63E-p1595456921037700-slack. 

<img src="https://user-images.githubusercontent.com/198826/88321981-41ae4480-ccdd-11ea-91c4-ddd7894b0e18.png" width="240">

Remove the “Feature Request” button in Settings and replace it with a “Send Feedback” button. The Send Feedback button opens the **general** survey screen created in #2544. 

---

Corresponding Android issue: https://github.com/woocommerce/woocommerce-android/issues/2647

PR #2672, opened 2020-08-18
Settings → Replace "Feature Request" With "Send Feedback"

shiki:
Closes #2553.

~⚠️ Leaving this as a draft until we have released the In-app Feedback feature (#2554). I found that this is easier to do so I don't have to hide this behind a feature flag. 😅~ 

This replaces the Feature Request button in Settings with Send Feedback. The Send Feedback button displays the survey. 

## Design

Before | After 
--------|-------
before | after

## Testing

1. Navigate to Settings.
2. Tap on Send Feedback. 

## Submitter Checklist

- [x] If it's feasible, I have added unit tests. 
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2745, opened 2020-08-31
Add RELEASE-NOTES for Settings Button Change

shiki:
Ref #2553 

I think the other features in the In-app Feedback project do not need to be announced since they're "difficult" to find.

## Testing

Please review the release notes text. 🙂 

## Submitter Checklist

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -9,7 +9,7 @@
   - A product variation's stock status is now editable in inventory settings.
   - Reviews row is now hidden if reviews are disabled.
   - Now it's possible to open the product's reviews screen also if there are no reviews.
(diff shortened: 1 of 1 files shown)


jaclync (commented):
not sure if I missed something, I didn't see "Feature Request" CTA in settings 🤔

shiki:
@jaclync Yep. The "Feature Request" button was removed. It was replaced with "Send Feedback". https://github.com/woocommerce/woocommerce-ios/pull/2672

I updated the release notes. Is that better? 😬

jaclync (approved):
Thanks for clarifying! I forgot that we had a "Feature Request" button before 😅 looks great! 👍

Issue #2522, opened 2020-07-16
Remove DashboardV3 after September 1st, 2020


Ecarrion:
After September 1st, 2020 the DashboardV3 will be deprecated and not shown in the app anymore.
We should all the code associated with it.

Ref: https://github.com/woocommerce/woocommerce-ios/pull/2521

PR #2746, opened 2020-09-01
Fix unit tests that depend on September 1st

jaclync:
## Changes

In `DashboardUIFactory`, we have logic that returns different view controllers based on whether the current date is September 1st:

https://github.com/woocommerce/woocommerce-ios/blob/5bbccdf6597e2f8df1472f773441ed0956816b54/WooCommerce/Classes/ViewRelated/Dashboard/Factories/DashboardUIFactory.swift#L73-L77

Now that today is September 1st (time flies 🍂 ), this PR replaced `DashboardStatsV3ViewController` with `DeprecatedDashboardStatsViewController` when stats v4 is not available.

Alternatively, we could update the unit tests to use a different current date in the future so that the tests don't depend on the actual current date.

## Testing

Just CI passing, no code changes outside of unit tests!

Update release notes:


pmusolino (approved):
Thanks for fixing it Jaclyn! 🎉 

Code ✅ 
Testing ✅ tests are green also locally

Ecarrion (approved):
Thanks @jaclync 
I'll work on https://github.com/woocommerce/woocommerce-ios/issues/2522 this week!

Issue #2738, opened 2020-08-31
Refactor `SyncingCoordinator` to fix infinite scroll where the response size does not match the list view size


jaclync:
Ref: p91TBi-30X-p2 - plan: p91TBi-30X#comment-1291-p2

The main work is to separate `SyncingCoordinator` into two components: scroll watcher and pagination tracker.

The first step is to fix an edge case when editing a grouped product's linked products with steps to reproduce:

Prerequisite: the store has a grouped product with more than 25 linked products
- Go to the Products tab
- Tap on the grouped product that has more than 25 linked products
- Tap on the "Grouped products" row
- Tap "X" to delete a linked product from the list
- Scroll down to the end of the list --> the infinite scroll does not work (if you didn't delete a linked product, the infinite scroll works)

After the issue above is resolved with the new infinite scroll architecture, we can start replacing other `SyncingCoordinator` usage in the code base.

jaclync:
Now that the grouped products use case is fixed in https://github.com/woocommerce/woocommerce-ios/pull/2750 with `PaginationTracker` and `PaginationTrackerDelegate`, let's replace the rest in the future after Products M3.

joshheald:
Automated triage note: this was tested and commented by Codex using device/simulator automation. I checked the current linked-products path in the WooCommerce iOS worktree using an iPhone 15 Pro simulator running iOS 17.5 as the target context. The original grouped-products linked-product selector now goes through PaginatedListSelectorViewController and PaginationTrackerDelegate rather than SyncingCoordinator. There are still other SyncingCoordinator usages in the app, but the specific infinite-scroll repro path from this issue appears to have been superseded by the pagination tracker work, so I would treat this issue as no longer relevant for the original bug.

I am leaving this GitHub issue open and applying the `triage: automation attempted` label for follow-up visibility.

dangermattic:
🚫 Please add a type label (e.g. **type: enhancement**) and a feature label (e.g. **feature: stats**) to this issue.

jaclync:
Closing, the main issue was fixed.

PR #2747, opened 2020-09-01
Infinite scroll update - part 1: scroll watcher

jaclync:
Part of #2738 
More details about the context: p91TBi-30X-p2

- [x] ⚠️ Make sure PR base is `develop` after https://github.com/woocommerce/woocommerce-ios/pull/2746 is merged ⚠️ 

The [full diffs](https://github.com/woocommerce/woocommerce-ios/compare/issue/2738-replace-SyncingCoordinator-in-paginated-list-selector?expand=1) are too big (~1000 diffs), so I'm breaking up the changes into a few parts for easier review.

## Changes

- Created `ScrollWatcher` that observes scroll view's content offset and emits scroll positions above the given threshold
  - In a future PR, the scroll watcher's `scrollTrigger: Observable<Double>` will be used to ensure the next page of results has been synced

## Testing

Just CI, since the new class isn't used yet 🙂 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
+++ b/WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
@@ -11,6 +11,7 @@ final class ScrollWatcher {
     private let triggerSubject: PublishSubject<Double> = PublishSubject<Double>()
     private let positionThreshold: Double
 
+    private var lastPosition: Double = 0.0
     private var offsetObservation: NSKeyValueObservation?
 
     /// - Parameter positionThreshold: the threshold of scroll position percentage (from 0.0 to 1.0) to trigger an event.
@@ -34,9 +35,10 @@ final class ScrollWatcher {
             // top of the table view and thus we deduct the table view's height.
             let contentHeight = tableView.contentSize.height - tableView.frame.height
             let scrollPosition = Double(1.0 * newContentOffsetY / contentHeight)
-            if scrollPosition >= self.positionThreshold {
+            if scrollPosition >= self.positionThreshold && scrollPosition > self.lastPosition {
                 self.triggerSubject.send(scrollPosition)
             }
+            self.lastPosition = scrollPosition
         }
     }
 }
--- a/WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
+++ b/WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
@@ -1,0 +1,43 @@
+import UIKit
+
+/// Watches for scroll content offset changes and notifies its subscribers if the scroll position is over a threshold.
+///
+final class ScrollWatcher {
+    /// Emits a stream of scroll position percentages from 0.0 (the beginning) to 1.0 (the end) if over a given threshold whenever the user scrolls a
+    /// `UIScrollView` or subclass.
+    var scrollTrigger: Observable<Double> {
+        scrollTriggerSubject
+    }
+    private let scrollTriggerSubject: PublishSubject<Double>
+    private let scrollPositionThreshold: Double
+
+    private var offsetObservation: NSKeyValueObservation?
+
+    /// - Parameter scrollPositionThreshold: the threshold of scroll position percentage (from 0.0 to 1.0) to trigger an event.
+    init(scrollPositionThreshold: Double = 0.7) {
+        self.scrollPositionThreshold = scrollPositionThreshold
+        self.scrollTriggerSubject = PublishSubject<Double>()
+    }
+
+    deinit {
+        offsetObservation?.invalidate()
+    }
+
+    func startObservingScrollPosition(tableView: UITableView) {
+        offsetObservation = tableView.observe(\UITableView.contentOffset, options: .new) { [weak self] tableView, change in
+            guard let self = self else {
+                return
+            }
+            guard let newContentOffsetY = change.newValue?.y else {
+                return
+            }
+            // When scrolling to the bottom of the list, the maximum content offset is at the top of the table view and thus we deduct the table view's
+            // height.
+            let contentHeight = tableView.contentSize.height - tableView.frame.height
+            let scrollPosition = Double(1.0 * newContentOffsetY / contentHeight)
+            if scrollPosition >= self.scrollPositionThreshold {
+                self.scrollTriggerSubject.send(scrollPosition)
+            }
+        }
+    }
+}
--- a/WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
+++ b/WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
@@ -5,18 +5,17 @@ import UIKit
 final class ScrollWatcher {
     /// Emits a stream of scroll position percentages from 0.0 (the beginning) to 1.0 (the end) if over a given threshold whenever the user scrolls a
     /// `UIScrollView` or subclass.
-    var scrollTrigger: Observable<Double> {
-        scrollTriggerSubject
+    var trigger: Observable<Double> {
+        triggerSubject
     }
-    private let scrollTriggerSubject: PublishSubject<Double>
-    private let scrollPositionThreshold: Double
+    private let triggerSubject: PublishSubject<Double> = PublishSubject<Double>()
+    private let positionThreshold: Double
 
     private var offsetObservation: NSKeyValueObservation?
 
-    /// - Parameter scrollPositionThreshold: the threshold of scroll position percentage (from 0.0 to 1.0) to trigger an event.
-    init(scrollPositionThreshold: Double = 0.7) {
-        self.scrollPositionThreshold = scrollPositionThreshold
-        self.scrollTriggerSubject = PublishSubject<Double>()
+    /// - Parameter positionThreshold: the threshold of scroll position percentage (from 0.0 to 1.0) to trigger an event.
+    init(positionThreshold: Double = 0.7) {
+        self.positionThreshold = positionThreshold
     }
 
     deinit {
@@ -31,12 +30,12 @@ final class ScrollWatcher {
             guard let newContentOffsetY = change.newValue?.y else {
                 return
             }
-            // When scrolling to the bottom of the list, the maximum content offset is at the top of the table view and thus we deduct the table view's
-            // height.
+            // When scrolling to the bottom of the list, the maximum content offset is at the
+            // top of the table view and thus we deduct the table view's height.
             let contentHeight = tableView.contentSize.height - tableView.frame.height
             let scrollPosition = Double(1.0 * newContentOffsetY / contentHeight)
-            if scrollPosition >= self.scrollPositionThreshold {
-                self.scrollTriggerSubject.send(scrollPosition)
+            if scrollPosition >= self.positionThreshold {
+                self.triggerSubject.send(scrollPosition)
             }
         }
     }
--- a/WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
+++ b/WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
@@ -59,7 +59,7 @@ final class ScrollWatcherTests: XCTestCase {
         tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 1000), size: frame.size), animated: false)
         // Scrolls to 99%
         tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 990), size: frame.size), animated: false)
-        // Scrolls to 99%
+        // Scrolls to 80%
         tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 800), size: frame.size), animated: false)
 
         // Assert
--- a/WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
+++ b/WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
@@ -1,0 +1,34 @@
+import XCTest
+@testable import WooCommerce
+
+final class ScrollWatcherTests: XCTestCase {
+    private var cancellable: ObservationToken?
+
+    func test_scroll_trigger_is_emitted_only_after_scrolling_to_above_threshold() {
+        // Arrange
+        let frame = CGRect(origin: .zero, size: CGSize(width: 320, height: 100))
+        let tableView = UITableView(frame: frame, style: .grouped)
+        let contentSize = CGSize(width: 320, height: 1100)
+        tableView.contentSize = contentSize
+        let threshold = 0.8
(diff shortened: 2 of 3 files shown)


WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
```diff
@@ -0,0 +1,43 @@
+import UIKit
+
+/// Watches for scroll content offset changes and notifies its subscribers if the scroll position is over a threshold.
+///
+final class ScrollWatcher {
```
pmusolino, line 5: I'm thinking if it would be better to call it `ScrollObserver`, but maybe it's confusing? I don't know, I don't have a strong opinion about this name 👍 feel free to leave it 😉

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
```diff
@@ -0,0 +1,43 @@
...
+    func startObservingScrollPosition(tableView: UITableView) {
+        offsetObservation = tableView.observe(\UITableView.contentOffset, options: .new) { [weak self] tableView, change in
+            guard let self = self else {
+                return
+            }
+            guard let newContentOffsetY = change.newValue?.y else {
+                return
+            }
+            // When scrolling to the bottom of the list, the maximum content offset is at the top of the table view and thus we deduct the table view's
+            // height.
```
pmusolino: nit:
```suggestion
            // When scrolling to the bottom of the list, the maximum content offset is at the
            // top of the table view and thus we deduct the table view's height.
```

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
```diff
@@ -0,0 +1,34 @@
...
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 500), size: frame.size), animated: false)
+        // Scrolls to 70%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 700), size: frame.size), animated: false)
+        // Scrolls to 80%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 800), size: frame.size), animated: false)
+        // Scrolls to 100%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 1000), size: frame.size), animated: false)
+
+        // Assert
+        XCTAssertEqual(triggeredScrollPositions, [0.8, 1.0])
```
pmusolino, line 38: 💯

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
```diff
@@ -0,0 +1,43 @@
...
+    var scrollTrigger: Observable<Double> {
+        scrollTriggerSubject
+    }
+    private let scrollTriggerSubject: PublishSubject<Double>
+    private let scrollPositionThreshold: Double
+
+    private var offsetObservation: NSKeyValueObservation?
+
+    /// - Parameter scrollPositionThreshold: the threshold of scroll position percentage (from 0.0 to 1.0) to trigger an event.
+    init(scrollPositionThreshold: Double = 0.7) {
```
pmusolino: Is there a reason on why we init it with `0.7` by default?
I suggest to not set a default value unless there is a reason behind it 👍 
```suggestion
    init(scrollPositionThreshold: Double) {
```

pmusolino (approved):
Looks good to me :shipit: 

I added just some nits

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
```diff
@@ -0,0 +1,43 @@
...
+/// Watches for scroll content offset changes and notifies its subscribers if the scroll position is over a threshold.
+///
+final class ScrollWatcher {
+    /// Emits a stream of scroll position percentages from 0.0 (the beginning) to 1.0 (the end) if over a given threshold whenever the user scrolls a
+    /// `UIScrollView` or subclass.
+    var scrollTrigger: Observable<Double> {
+        scrollTriggerSubject
+    }
+    private let scrollTriggerSubject: PublishSubject<Double>
+    private let scrollPositionThreshold: Double
```
shiki: I think we can leave off the `scroll*` in the names of these properties. I think they're obvious because the class has `Scroll*` too. 🤔

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    private var offsetObservation: NSKeyValueObservation?
+
+    /// - Parameter scrollPositionThreshold: the threshold of scroll position percentage (from 0.0 to 1.0) to trigger an event.
+    init(scrollPositionThreshold: Double = 0.7) {
+        self.scrollPositionThreshold = scrollPositionThreshold
+        self.scrollTriggerSubject = PublishSubject<Double>()
```
shiki: Just curious. Why is this in `init()` instead of the property? 

```swift
    private let scrollTriggerSubject = PublishSubject<Double>()
```

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
```diff
@@ -0,0 +1,34 @@
...
+        let tableView = UITableView(frame: frame, style: .grouped)
+        let contentSize = CGSize(width: 320, height: 1100)
+        tableView.contentSize = contentSize
+        let threshold = 0.8
+        let scrollWatcher = ScrollWatcher(scrollPositionThreshold: threshold)
+        scrollWatcher.startObservingScrollPosition(tableView: tableView)
+        var triggeredScrollPositions = [Double]()
+        cancellable = scrollWatcher.scrollTrigger.subscribe { scrollPosition in
+            triggeredScrollPositions.append(scrollPosition)
+        }
```
shiki: To avoid the ivar, we can also use `addTeardownBlock`

```suggestion
        let cancellable = scrollWatcher.scrollTrigger.subscribe { scrollPosition in
            triggeredScrollPositions.append(scrollPosition)
        }

        addTeardownBlock {
            cancellable.cancel()
        }
```

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
```diff
@@ -0,0 +1,43 @@
...
+        offsetObservation = tableView.observe(\UITableView.contentOffset, options: .new) { [weak self] tableView, change in
+            guard let self = self else {
+                return
+            }
+            guard let newContentOffsetY = change.newValue?.y else {
+                return
+            }
+            // When scrolling to the bottom of the list, the maximum content offset is at the top of the table view and thus we deduct the table view's
+            // height.
+            let contentHeight = tableView.contentSize.height - tableView.frame.height
```
shiki, line 36: I suppose at some point we should also consider the `contentSize` to determine the threshold. I'm thinking if there are 100 "pages", the 30% would probably be too early to _trigger_ a page load. Something we can think about in the future. 😄

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
```diff
@@ -0,0 +1,34 @@
...
+
+        // Action
+        // Scrolls to 50%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 500), size: frame.size), animated: false)
+        // Scrolls to 70%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 700), size: frame.size), animated: false)
+        // Scrolls to 80%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 800), size: frame.size), animated: false)
+        // Scrolls to 100%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 1000), size: frame.size), animated: false)
```
shiki, line 35: I wonder if we should consider scrolling down versus scrolling up. I added this after this line:

```swift
        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 900), size: frame.size), animated: false)
        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 800), size: frame.size), animated: false)
```

The value of `triggeredScrollPositions` is then:

```
[0.8, 1.0, 0.9, 0.8]
```

I think this means we're also triggering events for scrolling up. Do you think this is something to consider?

shiki (approved):
Thanks for splitting this up, Jaclyn! This is looking great. I have a couple of non-blocking questions.

shiki:
>  ⚠️ Make sure PR base is develop after #2746 is merged ⚠️

AFAIK, this is now automatic.

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
jaclync: *Observer might sound like the Observer in Rx, especially now that we have some lightweight Rx code like [`Observer`](https://github.com/woocommerce/woocommerce-ios/blob/74332bed838a5f4312586c450bcb16d2d1cc69ff/WooCommerce/Classes/Tools/Observables/Observer.swift) in the code base 🤔 keeping this name `ScrollWatcher` for now but lemme know if you have other suggestions!

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
jaclync: I thought most of the table views would have the same trigger for fetching the next page, that's why I set 70% as a guess and we can tweak this default value later too. do you think each paginated list might want to have their own threshold?

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
jaclync: makes sense! updated in https://github.com/woocommerce/woocommerce-ios/pull/2747/commits/2f42e68a3b2bbbec9a582a2341dbbcfbc15f9e6d

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
jaclync: no reason! 😆 updated in https://github.com/woocommerce/woocommerce-ios/pull/2747/commits/2f42e68a3b2bbbec9a582a2341dbbcfbc15f9e6d

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
jaclync: TIL'ed `addTeardownBlock`! in case we add more test cases, it might be nice to have the `cancellable` as a property. I added the `tearDown` override to cancel the observation token in https://github.com/woocommerce/woocommerce-ios/pull/2747/commits/2f42e68a3b2bbbec9a582a2341dbbcfbc15f9e6d

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
jaclync: yea it helps with reading the comment in two lines! updated in: https://github.com/woocommerce/woocommerce-ios/pull/2747/commits/2f42e68a3b2bbbec9a582a2341dbbcfbc15f9e6d

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
jaclync: yes the current implementation wouldn't work well with large amount of pages like you mentioned. after we integrate with the pagination tracker in a future PR, I'm thinking we could create an observable for the highest page number in the pagination tracker and `ScrollWatcher` can subscribe to get the latest highest page number `page: Int` and trigger the next page when the scroll position is higher than `(page - 1)/page` assuming each page is roughly the same size. any feedback is welcome!

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
jaclync: great suggestion! I added a check to only emit a value if it's higher than the previous one in https://github.com/woocommerce/woocommerce-ios/pull/2747/commits/ef9ace08fe4fc1fab9ad8ba04ab0dbb546ede84f

jaclync:
>> ⚠️ Make sure PR base is develop after #2746 is merged ⚠️

> AFAIK, this is now automatic.

@shiki yes, just wanted to put a note to self to make sure the PR base is `develop` before merging 🙂 

@pmusolino @shiki thanks for the feedback/suggestions, I updated the PR and responded to your comments! Since there are some open questions, I'm leaving the PR open in case you'd like to take another look 😄

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
pmusolino: It makes sense, thanks for clarifying it 👍

WooCommerce/Classes/Tools/InfiniteScroll/ScrollWatcher.swift
shiki: That sounds like a good plan. What we have here is good enough for now and we can always improve things later.

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        // Action
+        // Scrolls to 80%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 800), size: frame.size), animated: false)
+        // Scrolls to 100%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 1000), size: frame.size), animated: false)
+        // Scrolls to 99%
+        tableView.scrollRectToVisible(CGRect(origin: CGPoint(x: 0, y: 990), size: frame.size), animated: false)
+        // Scrolls to 99%
```
shiki:
```suggestion
        // Scrolls to 80%
```

🙂

shiki (approved):
:pandance:

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/ScrollWatcherTests.swift
jaclync: Oops! sharp eyes 🦅 👀 updated in https://github.com/woocommerce/woocommerce-ios/pull/2747/commits/7c3d59e185ab6b6beb7e13149a11f0af829857a7

PR #2748, opened 2020-09-01
Infinite scroll update - part 2: send `hasNextPage` on sync completion

jaclync:
Part of #2738
More details about the context: p91TBi-30X#comment-1291-p2

- [x] ⚠️ Make sure PR base is `develop` after https://github.com/woocommerce/woocommerce-ios/pull/2746 is merged ⚠️ 

The [full diffs](https://github.com/woocommerce/woocommerce-ios/compare/issue/2738-replace-SyncingCoordinator-in-paginated-list-selector?expand=1) are too big (~1000 diffs), so I'm breaking up the changes into a few parts for easier review.

## Changes

The goal of this PR is to enable the infinite support based on the updated architecture in p91TBi-30X#comment-1291-p2. Currently, the infinite scroll only works if the API response size matches the table view size for each page. To move away from this assumption, we need another way to know whether there are more pages to fetch from the API. This PR provides `hasNextPage: Bool` in Yosemite layer based on whether the API response is full - if the API response size is smaller than the requested page size, we can assume that is no next page. Note that `hasNextPage` isn't used in the app layer yet, and you can see how it's used in the [branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/2738-replace-SyncingCoordinator-in-paginated-list-selector?expand=1).

- Networking layer:
  - Updated `ProductShippingClassRemote` to return `Result<[ProductShippingClass], Error>` instead of `([ProductShippingClass]?, Error?)` in the completion block, and its unit tests
- Yosemite layer:
  - Updated completion callback to include `hasNextPage` info in `ProductShippingClassAction.synchronizeProductShippingClassModels` with unit tests
  - Updated completion callback to include `hasNextPage` info in `ProductAction.synchronizeProducts` and `ProductAction.retrieveProducts` with unit tests
- App layer:
  - Updated affected action usage in `ProductListMultiSelectorDataSource`, `PaginatedProductShippingClassListSelectorDataSource`, and `ProductsViewController` while keeping the existing behavior

## Testing

Please sanity check on the affected screens, no user-facing changes are expected:

### Product shipping classes

- Go to the Products tab
- Tap on a simple physical product
- Tap on the shipping row or from the bottom sheet, and then the shipping class row --> the paginated list should load as before, with the correct selection

### Grouped products

- Go to the Products tab
- Tap on a grouped product
- Tap on the grouped products row --> the paginated list should load as before
- Tap "Add Products" CTA --> the paginated list of products should load as before

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/ProductShippingClassRemote.swift
+++ b/Networking/Networking/Remote/ProductShippingClassRemote.swift
@@ -1,5 +1,4 @@
 import Foundation
-import Alamofire
 
 /// ProductShippingClass: Remote Endpoints
 ///
@@ -19,7 +18,7 @@ public class ProductShippingClassRemote: Remote {
                         context: String? = nil,
                         pageNumber: Int = Default.pageNumber,
                         pageSize: Int = Default.pageSize,
-                        completion: @escaping ([ProductShippingClass]?, Error?) -> Void) {
+                        completion: @escaping (Result<[ProductShippingClass], Error>) -> Void) {
         let parameters = [
             ParameterKey.page: String(pageNumber),
             ParameterKey.perPage: String(pageSize),
--- a/Yosemite/Yosemite/Stores/ProductShippingClassStore.swift
+++ b/Yosemite/Yosemite/Stores/ProductShippingClassStore.swift
@@ -40,22 +40,23 @@ private extension ProductShippingClassStore {
 
     /// Synchronizes the `ProductShippingClass`s associated with a given Site ID (if any!).
     ///
-    func synchronizeProductShippingClassModels(siteID: Int64, pageNumber: Int, pageSize: Int, onCompletion: @escaping (Error?) -> Void) {
+    func synchronizeProductShippingClassModels(siteID: Int64, pageNumber: Int, pageSize: Int, onCompletion: @escaping (Result<Bool, Error>) -> Void) {
         let remote = ProductShippingClassRemote(network: network)
 
-        remote.loadAll(for: siteID, pageNumber: pageNumber, pageSize: pageSize) { [weak self] (models, error) in
-            guard let models = models else {
-                onCompletion(error)
-                return
-            }
+        remote.loadAll(for: siteID, pageNumber: pageNumber, pageSize: pageSize) { [weak self] result in
+            switch result {
+            case .failure(let error):
+                onCompletion(.failure(error))
+            case .success(let models):
+                if pageNumber == Default.firstPageNumber {
+                    self?.deleteStoredProductShippingClassModels(siteID: siteID)
+                }
 
-            if pageNumber == Default.firstPageNumber {
-                self?.deleteStoredProductShippingClassModels(siteID: siteID)
-            }
-
-            self?.upsertStoredProductShippingClassModelsInBackground(readOnlyProductShippingClassModels: models,
-                                                                     siteID: siteID) {
-                                                                        onCompletion(nil)
+                self?.upsertStoredProductShippingClassModelsInBackground(readOnlyProductShippingClassModels: models,
+                                                                         siteID: siteID) {
+                                                                            let hasNextPage = models.count == pageSize
+                                                                            onCompletion(.success(hasNextPage))
+                }
             }
         }
     }
--- a/Yosemite/Yosemite/Stores/ProductShippingClassStore.swift
+++ b/Yosemite/Yosemite/Stores/ProductShippingClassStore.swift
@@ -48,12 +48,16 @@ private extension ProductShippingClassStore {
             case .failure(let error):
                 onCompletion(.failure(error))
             case .success(let models):
-                if pageNumber == Default.firstPageNumber {
-                    self?.deleteStoredProductShippingClassModels(siteID: siteID)
+                guard let self = self else {
+                    return
                 }
 
-                self?.upsertStoredProductShippingClassModelsInBackground(readOnlyProductShippingClassModels: models,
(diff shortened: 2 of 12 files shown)


Networking/Networking/Remote/ProductShippingClassRemote.swift
```diff
@@ -19,7 +18,7 @@ public class ProductShippingClassRemote: Remote {
                         context: String? = nil,
                         pageNumber: Int = Default.pageNumber,
                         pageSize: Int = Default.pageSize,
-                        completion: @escaping ([ProductShippingClass]?, Error?) -> Void) {
+                        completion: @escaping (Result<[ProductShippingClass], Error>) -> Void) {
```
pmusolino, line 21: 💯 thanks for converting it in `Result < >`

shiki, line 21: :success:

Yosemite/Yosemite/Stores/ProductShippingClassStore.swift
```diff
@@ -40,22 +40,23 @@ private extension ProductShippingClassStore {
...
+                onCompletion(.failure(error))
+            case .success(let models):
+                if pageNumber == Default.firstPageNumber {
+                    self?.deleteStoredProductShippingClassModels(siteID: siteID)
+                }
+
+                self?.upsertStoredProductShippingClassModelsInBackground(readOnlyProductShippingClassModels: models,
+                                                                         siteID: siteID) {
+                                                                            let hasNextPage = models.count == pageSize
+                                                                            onCompletion(.success(hasNextPage))
```
shiki: This might be a good opportunity to add `guard let self = self` so that we can be sure that both `deleteStoredProductShippingClassModels` and `upsertStoredProductShippingClassModelsInBackground` are executed as **all or nothing**.

Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -143,14 +143,15 @@ private extension ProductStore {
...
+                                    onCompletion(.failure(error))
                                 case .success(let products):
                                     if pageNumber == Default.firstPageNumber && shouldDeleteStoredProductsOnFirstPage {
                                         self?.deleteStoredProducts(siteID: siteID)
                                     }
 
                                     self?.upsertStoredProductsInBackground(readOnlyProducts: products) {
-                                        onCompletion(nil)
+                                        let hasNextPage = products.count == pageSize
+                                        onCompletion(.success(hasNextPage))
```
shiki: Yep. All these `self?` usage scares me a bit. 😄 It's probably unlikely since this is running on the same thread. I think. 🤔

Yosemite/YosemiteTests/Stores/ProductShippingClassStoreTests.swift
```diff
@@ -134,6 +134,58 @@ final class ProductShippingClassStoreTests: XCTestCase {
...
+            }
+            store.onAction(action)
+        }
+
+        // Assert
+        guard case let .success(hasNextPage) = result else {
+            XCTFail("Unexpected result: \(String(describing: result))")
+            return
+        }
+        XCTAssertTrue(hasNextPage)
```
shiki: This may be enough:

```suggestion
        XCTAssertTrue(try XCTUnwrap(result).get())
```

It's not explicit but I think it's okay. ¯\\\_(ツ)\_/¯

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -1075,6 +1133,58 @@ final class ProductStoreTests: XCTestCase {
         XCTAssertEqual(viewStorage.countObjects(ofType: Storage.Product.self), 1)
     }
 
+    func test_retrieving_products_of_the_same_page_size_has_next_page() {
+        // Arrange
+        let remote = MockProductsRemote()
+        let expectedProducts: [Yosemite.Product] = .init(repeating: MockProduct().product(), count: 25)
```
shiki, line 1131: Cool!

shiki (approved):
LGTM! 🤸

Yosemite/YosemiteTests/Stores/ProductShippingClassStoreTests.swift
jaclync: nice TIL'ed `Result.get()`! it shortens the assert code quite a bit, updated in https://github.com/woocommerce/woocommerce-ios/pull/2748/commits/397b12849fe849982b584d115c12ebba2a8a4734

Yosemite/Yosemite/Stores/ProductShippingClassStore.swift
jaclync: yup this certainly sounds more robust! updated in https://github.com/woocommerce/woocommerce-ios/pull/2748/commits/37fbfe7bce3667ee25ffda170a17351f38eb2718

Yosemite/Yosemite/Stores/ProductStore.swift
jaclync: yup thanks for bringing this up 😄 updated in https://github.com/woocommerce/woocommerce-ios/pull/2748/commits/37fbfe7bce3667ee25ffda170a17351f38eb2718

jaclync:
Thanks a lot for the review and feedback Jayson! merging this now 😄

pmusolino (approved):
Looks good to me and everything works perfectly :shipit: 

Code ✅ 
Testing ✅

PR #2750, opened 2020-09-02
Fix infinite scroll for grouped product's linked products after deletion

jaclync:
Fix infinite scroll for grouped product's linked products #2738

- [x] ⚠️ Make sure PR base is `develop` after https://github.com/woocommerce/woocommerce-ios/pull/2747 is merged ⚠️ 

## Changes

- Created `PaginationTracker` and `PaginationTrackerDelegate` that is an updated version of `SyncingCoordinator` with major differences:
  - No more caching behavior and syncing an earlier page. I don't feel like we need to sync an earlier page when scrolling to it, because if an earlier page has changes, we probably want to fetch the other earlier pages anyway. Our API doesn't work great with syncing, letting the client know of the changes. I think just keeping fetching the next highest page would be sufficient, and the user can still sync the first page by pulling down to refresh
  - The sync completion block in the delegate now takes `Result<Bool, Error>` instead of just a success boolean. This allows the pagination tracker to know if the API results have reached the end from the delegate, rather than the logic to compare the table view maximum index path with the page size.
- Updated `PaginatedListSelectorDataSource` and `PaginatedListSelectorViewController` to replace `SyncingCoordinator` with `ScrollWatcher` and `PaginationTracker`.
- Updated `PaginatedListSel` implementations to set `hasNextPage` on sync completion: `GroupedProductListSelectorDataSource`, `ProductListMultiSelectorDataSource`, `PaginatedProductShippingClassListSelectorDataSource`, and `ProductTaxClassListSelectorDataSource` (always no next page)

## Testing

Prerequisite: the store has a grouped product with more than 25 linked products
- Go to the Products tab
- Tap on the grouped product that has more than 25 linked products
- Tap "X" to delete a linked product from the list
- Scroll down toward the end of the list --> the infinite scroll should work (this wouldn't work on `develop`)

Please also sanity check on other affected screens (now infinite scroll is triggered when scrolling to ~70% of the list):

### Product shipping classes

- Go to the Products tab
- Tap on a simple physical product
- Tap on the shipping row or from the bottom sheet, and then the shipping class row --> the paginated list should load as before, with the correct selection

### Product tax classes

- Go to the Products tab
- Tap on a simple product
- Tap on the price row, and then the tax class row --> the list should load as before, with the correct selection

### Grouped products

- Go to the Products tab
- Tap on a grouped product
- Tap on the grouped products row --> the paginated list should load as before
- Tap "Add Products" CTA --> the paginated list of products should load as before

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Grouped Products/GroupedProductListSelectorDataSource.swift
```diff
@@ -75,17 +75,17 @@ final class GroupedProductListSelectorDataSource: PaginatedListSelectorDataSourc
         }
     }
 
-    func sync(pageNumber: Int, pageSize: Int, onCompletion: ((Bool) -> Void)?) {
+    func sync(pageNumber: Int, pageSize: Int, onCompletion: ((Result<Bool, Error>) -> Void)?) {
```
pmusolino, line 78: 👏

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
```diff
@@ -0,0 +1,120 @@
...
+        paginationTracker.ensureNextPageIsSynced()
+
+        // Assert
+        XCTAssertEqual(syncedPages, [
+            PageInformation(pageNumber: 0, pageSize: 30, reason: nil),
+            PageInformation(pageNumber: 1, pageSize: 30, reason: nil)
+        ])
+    }
+
+    func test_ensuring_next_page_after_syncing_first_page_without_next_page_is_noop() {
```
pmusolino:
```suggestion
    func test_ensuring_next_page_after_syncing_first_page_without_next_page_is_no_op() {
```
Maybe separating `noop` in `no_op` it's better understandable?

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
```diff
@@ -0,0 +1,120 @@
+import XCTest
+@testable import WooCommerce
+
+final class PaginationTrackerTests: XCTestCase {
```
pmusolino, line 4: 👍 great test cases!

pmusolino (approved):
Looks good to me :shipit: 

Code ✅ 
Testing ✅ everything works like expected

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
```diff
@@ -0,0 +1,152 @@
...
+    private let pageFirstIndex: Int
+
+    /// Number of elements retrieved per request.
+    private let pageSize: Int
+
+    /// Syncing delegate
+    weak var delegate: PaginationTrackerDelegate?
+
+    /// Maps from a page number to the last synced date.
+    private var syncedDatesByPage = [Int: Date]()
```
shiki: I don't see the `Date` value being used anywhere. Is this a placeholder for this?

> I don't feel like we need to sync an earlier page when scrolling to it, because if an earlier page has changes, we probably want to fetch the other earlier pages anyway.

There's also a chance I misunderstood what you meant there ☝️ 😅

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
```diff
@@ -0,0 +1,152 @@
...
+    /// - Parameter onCompletion: called on sync completion that returns either an error or a boolean that indicates
+    ///                           whether there might be the next page to sync.
+    func sync(pageNumber: Int, pageSize: Int, reason: String?, onCompletion: SyncCompletion?)
+}
+
+/// Keeps track of the pagination for API syncing to support infinite scroll.
+final class PaginationTracker {
+    /// Default Settings
+    private enum Defaults {
+        static let pageFirstIndex = Store.Default.firstPageNumber
```
shiki, line 20: Imagine all the bugs we can cause if we change the value of `Store.Default.firstPageNumber`. 🤣

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
```diff
@@ -0,0 +1,152 @@
...
+    ///
+    /// - Parameter reason: A value passed from `resync` or `syncFirstPage`. This can
+    ///                     be used to decide how to perform the sync.
+    /// - Parameter onCompletion: called on sync completion that returns either an error or a boolean that indicates
+    ///                           whether there might be the next page to sync.
+    func sync(pageNumber: Int, pageSize: Int, reason: String?, onCompletion: SyncCompletion?)
+}
+
+/// Keeps track of the pagination for API syncing to support infinite scroll.
+final class PaginationTracker {
```
shiki, line 17: Should we mark `SyncCoordinator` as `@deprecated`?

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
```diff
@@ -0,0 +1,152 @@
...
+
+    /// Returns the highest page number that is currently being synced, if any.
+    private var highestPageBeingSynced: Int? {
+        pagesBeingSynced.max()
+    }
+
+    /// Returns the highest page number that has been successfully synced.
+    private var highestPageSynced: Int? {
+        syncedDatesByPage.keys.max()
+    }
```
shiki: What do you think about making this return zero by default instead? 

```suggestion
    private var highestPageSynced: Int {
        syncedDatesByPage.keys.max() ?? pageFirstIndex - 1
    }
```

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
shiki: I'm thinking we should rename this to `pagesSynced` or something that's closer to `pagesBeingSynced` (below). I think they're closely related and it would help if they're similarly named.

WooCommerce/Classes/ViewRelated/Products/Edit Product/List Selector Data Source/ProductTaxClassListSelectorDataSource.swift
```diff
@@ -42,12 +42,15 @@ struct ProductTaxClassListSelectorDataSource: PaginatedListSelectorDataSource {
...
-    func sync(pageNumber: Int, pageSize: Int, onCompletion: ((Bool) -> Void)?) {
+    func sync(pageNumber: Int, pageSize: Int, onCompletion: ((Result<Bool, Error>) -> Void)?) {
         let action = TaxClassAction.retrieveTaxClasses(siteID: siteID) { (taxClasses, error) in
             if let error = error {
                 DDLogError("⛔️ Error synchronizing tax classes: \(error)")
+                onCompletion?(.failure(error))
+                return
             }
-            onCompletion?(error == nil)
+            // Tax classes endpoint does not support pagination, thus we assume there is no next page on every retrieval.
```
shiki, line 52: Thanks for that!

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
```diff
@@ -0,0 +1,120 @@
...
+    private let paginationTracker = PaginationTracker(pageFirstIndex: 0, pageSize: 30)
+    private let mockDelegate = MockPaginationTrackerDelegate()
+
+    override func setUp() {
+        super.setUp()
+        paginationTracker.delegate = mockDelegate
+    }
```
shiki: ⚠️  We should do a `tearDown` even though these objects are light. 🙂 

```suggestion
    private var paginationTracker: PaginationTracker!
    private var mockDelegate: MockPaginationTrackerDelegate!

    override func setUp() {
        super.setUp()
        mockDelegate = MockPaginationTrackerDelegate()
        paginationTracker = PaginationTracker(pageFirstIndex: 0, pageSize: 30)
        paginationTracker.delegate = mockDelegate
    }

    override func tearDown() {
        paginationTracker = nil
        mockDelegate = nil
        super.tearDown()
    }
```

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
```diff
@@ -0,0 +1,120 @@
...
+private struct PageInformation: Equatable {
+    let pageNumber: Int
+    let pageSize: Int
+    let reason: String?
+}
+
+/// `PaginationTrackerDelegate` Closure-based wrapper, for unit testing purposes.
+///
+private class MockPaginationTrackerDelegate: PaginationTrackerDelegate {
+    typealias OnSyncClosure = (Int, Int, String?, SyncCompletion?) -> Void
```
shiki: Based on the way we used the values in the tests, it looks like we can save some words to read by using the `PageInformation` instead:

```suggestion
    typealias OnSyncClosure = (PageInformation, SyncCompletion?) -> Void
```

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
shiki: Yeah, this is so nice!

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
```diff
@@ -0,0 +1,120 @@
...
+
+        // Action
+        paginationTracker.syncFirstPage()
+        paginationTracker.ensureNextPageIsSynced()
+
+        // Assert
+        XCTAssertEqual(syncedPages, [
+            PageInformation(pageNumber: 0, pageSize: 30, reason: nil),
+            PageInformation(pageNumber: 1, pageSize: 30, reason: nil)
+        ])
```
shiki: OMG I just tried this out of curiosity and it works!

```swift
        XCTAssertEqual(syncedPages, [
            .init(pageNumber: 0, pageSize: 30, reason: nil),
            .init(pageNumber: 1, pageSize: 30, reason: nil)
        ])
```

TIL

shiki (approved):
This looks great! 

I just have one request, marked with ⚠️ . But feel free to merge if you agree. 🙂 

:shipit:

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
jaclync: yeaa suddenly many screens wouldn't get the first page of data, and we will get so many reviews/reports about the outdated data 😅 😨

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
jaclync: sure sounds good I'll keep this in mind next time! updated in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/d3a058fd52f17405824e28fedfb66a8665eee178

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
jaclync: niice! it does shorten the code a lot, updated in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/e196384c52d2f0549c001ef79be8b9fc2f3581bc

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
jaclync: Great suggestion! looks much more concise now: https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/83045affdd260821e25cadfc58093307ee21e378

WooCommerce/WooCommerceTests/Tools/InfiniteScroll/PaginationTrackerTests.swift
jaclync: yup that's a good point! updated in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/83045affdd260821e25cadfc58093307ee21e378

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
jaclync: yup I renamed it to `pagesSynced` to be consistent with `pagesBeingSynced` and also because it has the same type as `pagesBeingSynced` now! updated in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/85d2cbc94229ee56a5487eb03b9996b1dfb943f8

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
jaclync: I felt a bit odd to return a non-nil value when no pages have been synced yet 🤔 does it look okay to you to keep it optional?

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
jaclync: > There's also a chance I misunderstood what you meant there ☝️ 😅

The type should just be `IndexSet()` similar to `pagesBeingSynced` here, thanks for pointing it out! I forgot to update after borrowing from `SyncingCoordinator`, changes made in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/85d2cbc94229ee56a5487eb03b9996b1dfb943f8. I think we're on the same page that the caching behavior (thus recording the last sync date) isn't necessary. Sorry for the long and not so clear sentence before!

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
jaclync: good idea, I did so in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/f2c848641884637de6fa106e10f92748ebb806e4 but now we have 23 issues in WooCommerce and 7 issues in WooCommerceTests from the usage of `SyncingCoordinator` - do you think it's fine we leave the warnings in the project and I can work on replacing `SyncingCoordinator` for hack week next week?

jaclync:
thanks a lot for the review & testing @shiki @pmusolino! 😄 I updated the PR and responded to all the comments. I'd like to confirm some changes (especially the [`@deprecated` one](https://github.com/woocommerce/woocommerce-ios/pull/2750#discussion_r483893923)) so leaving the PR open for another look and targeting 5.1 🙂

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
shiki: I asked for opinions in CGPNUU63E-p1599574615070200-slack

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
shiki: Yeah, I guess the optional has some value too. I think I just commented on this because it took me a while to understand. 😅 But all good!

shiki (approved):
Changes look good. We can add the `@deprecated` later if you'd like. 🙂

WooCommerce/Classes/Tools/InfiniteScroll/PaginationTracker.swift
jaclync: I reverted the deprecation in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/cd9406458dc8a13fdf4b8aa00beb027924cdc05f so that we have no build warnings!

jaclync:
I removed the deprecation of `SyncingCoordinator` in https://github.com/woocommerce/woocommerce-ios/pull/2750/commits/cd9406458dc8a13fdf4b8aa00beb027924cdc05f so that we don't get distracted from the build warnings and accidentally add a new one (CGPNUU63E-p1599574615070200-slack) 🙂 merging now!

PR #2761, opened 2020-09-03
Move WooCommerce assert* Functions to TestKit

shiki:
I wanted to use these for a test case in Yosemite but I can't cause they are still in WooCommerce. Luckily, TestKit now exists. So I moved them there. 😄 

In the original implementation of these methods, they were inside an extension:

```swift
extension XCTestCase {
    func assertEmpty(...) { } 
}
```

I opted to make them global functions this time around to closely match them with [`XCT*` assertions](https://developer.apple.com/documentation/xctest/boolean_assertions) which are also global. Or I think they're macros. Whatever. 🙃 

```swift
public func assertEmpty(...) { } 
```

Global functions have the added benefit of making them usable in other places [like Mock classes](https://github.com/woocommerce/woocommerce-ios/blob/6c58e2fa3d37a3e5099c179ef54854f9a5402586/Yosemite/YosemiteTests/Mockups/Networking/Remote/MockProductsRemote.swift#L57). 

## Testing

Please make sure that the tests still pass. 

## Submitter Checklist


jaclync (approved):
LGTM! :shipit: 

Code: ✅ 
Testing: ✅ (all tests passed locally!)

shiki:
Thanks, Jaclyn!

PR #2777, opened 2020-09-04
DiffableDataSource: Add FetchResultsSnapshotsProvider

shiki:
Part of the fix for #1543. I'm aiming to use the new `DiffableDataSource` instead of the old `ResultsController`. 

Big thanks to @jaclync for helping me get through this ordeal! 🙏 

## About

This adds a [`FetchResultsSnapshotsProvider`](https://github.com/woocommerce/woocommerce-ios/blob/issue/1543-add-snapshotsprovider/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift) which encapsulates the intricacies of working with `DiffableDataSource`:

- Hides the `NSFetchRequest` and tries to give a bit simpler API.
- Hides working with `Storage*` models. Hides the fact that we're using `NSManagedObject` underneath. 
- Encapsulates the wiring of `NSFetchedResultsController` and `NSDiffableDataSourceSnapshot`. 
- Provides the converter for immutable objects. 

This provides a [Combine framework `Publisher`](https://developer.apple.com/documentation/combine/publisher) which will be subscribed by the UI (`UITableView`). 

https://github.com/woocommerce/woocommerce-ios/blob/eb464bb63a482f5eebf4c15af9ed6ceaa0cc124a/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L89-L92

A simplified diagram of the data flow:

image

### Example Usage

```swift
// Create the snapshots provider
let query = FetchResultSnapshotsProvider<StorageOrder>.Query(
    sortDescriptor: .init(keyPath: \StorageOrder.number, ascending: false)
)
let provider = FetchResultSnapshotsProvider(storageManager: storageManager, query: query)

// Set up the TableView and DataSource
let dataSource =
   UITableViewDiffableDataSource<String, FetchResultSnapshotObjectID>(tableView: tableView,
                                                                       cellProvider: makeCellProvider())
tableView.dataSource = dataSource

// The makeCellProvider() replaces the `tableView:cellForRow` delegate method. (used above)
private func makeCellProvider() -> UITableViewDiffableDataSource<String, FetchResultSnapshotObjectID>.CellProvider {
   return { [weak self] tableView, indexPath, objectID in
       guard let self = self, let cell = tableView.dequeueReusableCell(withIdentifier: ...) else {
           fatalError("YOLO!")
       }

       let order = provider.object(withID: objectID)
       cell.configure(using: order)
       return cell
   }
}

// Observe and apply all emitted snapshots to the TableView's DataSource. 
// This _continuously_ updates the `UITableView` whenever there are data changes
provider.snapshot.sink { snapshot in
    dataSource.apply(snapshot)
}.store(in: &self.cancellables)

provider.start()
```

That's it! This makes the table-binding code a lot shorter. The sections are missing from this example but it's not really handled by `FetchResultsSnapshotsProvider`. Examples of that will be provided later. 

### Next Step

Later on, this will also encapsulate a workaround for a [bug in `NSFetchedResultsController` snapshot handling](https://openradar.appspot.com/FB7588509). 

### iOS 13 Only

This class is only available for iOS 13.0. We will be dropping iOS 12.0 support soon but for now, we'll just target iOS 13.0 users for now. 

### Consciuosly-created Technical Debt

I'm temporarily exposing `obtainPermanentIDs` in `StorageType`.

https://github.com/woocommerce/woocommerce-ios/blob/eb464bb63a482f5eebf4c15af9ed6ceaa0cc124a/Storage/Storage/Protocols/StorageType.swift#L59

When synchronizing, `NSFetchedResultsController`'s snapshot delegate method _emits_ temporary IDs and then immediately emits another snapshot with permanent IDs. This causes an undesirable confusing animation (dual refresh) when the subsequent snapshots are applied to the `UITableView`.

Here's a video of the confusing animation when using temporary IDs → https://drive.google.com/file/d/1HoS0P66TSwfmF29QzBUqw-oTKj5FXrHn/view?usp=sharing. Notice how the `UITableView` looks like it refreshed everything instead of just adding new rows to the bottom. It also automatically scrolled the user (me) to the top of the list. 

We will find a better solution for this later. Obtaining permanent IDs can cause a performance hit. I'm not sure by how much but it doesn't look like a lot since we'll be saving the objects in the background thread anyway. My current idea is to use a separate **read-only** `NSManagedObjectContext` to use for the `NSFetchedResultsController`. This will ensure that `NSFetchedResultsController` will not ever receive temporary IDs.

I had a discussion with Jaclyn regarding possible solutions for this in https://github.com/woocommerce/woocommerce-ios/commit/7e06f6d9854a756a39c03b03175076ed26676303#r41846521. 

### Recommended WWDC Videos

I based this implementation from what I understood from these videos:

- [Making Apps with Core Data - WWDC 2019 - Videos - Apple Developer](https://developer.apple.com/videos/play/wwdc2019/230/)
- [Advances in UI Data Sources - WWDC 2019 - Videos - Apple Developer](https://developer.apple.com/videos/play/wwdc2019/220)

## Testing

1. Please make sure the app still compiles on iOS 12.0.
2. Since this is not used yet, please review the unit tests for now. 

## Submitter Checklist


Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
```diff
@@ -0,0 +1,239 @@
...
+import Storage
+
+@available(iOS 13.0, *)
+final class FetchResultSnapshotsProviderTests: XCTestCase {
+
+    private var storageManager: MockupStorageManager!
+
+    private var cancellables = Set<AnyCancellable>()
+
+    var viewStorage: StorageType {
```
jaclync: nit: could this be private?
```suggestion
    private var viewStorage: StorageType {
```

Storage/Storage/Protocols/StorageType.swift
```diff
@@ -46,6 +46,18 @@ public protocol StorageType: class {
...
+    ///
+    /// This is temporarily exposed since `NSFetchedResultsController`'s `DiffableDataSource`
+    /// support exposes temporary IDs in snapshots and immediately converts them to permanent IDs.
+    /// This causes an undesirable confusing animation (dual refresh) when the subsequent
+    /// snapshots are applied to the `UITableView`.
+    ///
+    /// We will find a better solution for this later. My current idea is to use a separate
+    /// **read-only** `NSManagedObjectContext` to use for the `NSFetchedResultsController`.
+    /// This will ensure that `NSFetchedResultsController` will not ever receive temporary IDs.
+    func obtainPermanentIDs(for objects: [NSManagedObject]) throws
```
Ecarrion, line 59: Gotta love apple and their unfinished APIs! 🤪

TestKit/Sources/TestKit/XCTestCase+Wait.swift
```diff
@@ -54,4 +54,43 @@ extension XCTestCase {
...
+    /// let value: String = try waitFor { promise in
+    ///     fetchFromAPI { responseString in
+    ///         promise(responseString)
+    ///     }
+    /// }
+    ///
+    /// XCTAssertEquals("expected_value", value)
+    /// ```
+    ///
+    public func waitFor<ValueType>(file: StaticString = #file,
```
Ecarrion, line 72: Is this a replacement for `waitUntil`?

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -0,0 +1,156 @@
...
+import CoreData
+import Combine
+
+/// The mutable (`Storage` model) type that is used by `FetchResultSnapshotsProvider`.
+public typealias FetchResultSnapshotsProviderMutableType = NSManagedObject & ReadOnlyConvertible
+
+/// The type of the items emitted by `FetchResultSnapshot`.
+///
+/// This is only declared to _hide_ `NSManagedObjectID` from the UI.
+public typealias FetchResultSnapshotObjectID = NSManagedObjectID
```
Ecarrion, line 11: Maybe `StorageObjectID` or will this only be needed in the `FetchResultsSnapshop` context?

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
```diff
@@ -0,0 +1,239 @@
...
+        XCTAssertEqual(snapshot.itemIdentifiers(inSection: "Y"), expectedSecondSection.map(\.objectID))
+    }
+
+    func test_snapshot_continuously_emits_values_for_structural_changes() throws {
+        // Given
+        let zanza = insertAccount(displayName: "Z", username: "Zanza")
+        let zagato = insertAccount(displayName: "Z", username: "Zagato")
+        let yamada = insertAccount(displayName: "Y", username: "Yamada")
+
+        try viewStorage.obtainPermanentIDs(for: [zanza, zagato, yamada])
```
Ecarrion, line 200: Since `obtainPermanentIDs` is temporary, maybe we should not rely on it for tests, WDYT?

Ecarrion (approved):
Excellent Job on hiding `CoreData` implementation details from the Woo target! 👏 
I have to admit that I did not understand how `obtainPermanentIDs` fixes the animation problem, or will this be tackled in a different PR?

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -0,0 +1,156 @@
...
+
+    /// The `StorageType` to perform the fetch on.
+    ///
+    /// This is currently forced to be the `viewStorage` (main thread). In the future, we can
+    /// probably move this to a background `StorageType` since `UITableViewDiffableDataSource`
+    /// allows consuming snapshots in the background.
+    private let storage: StorageType
+    /// The conditions to use when fetching the results.
+    ///
+    /// In the future, we can allow this to be mutable if necessary.
```
jaclync, line 86: yea, I think the query can change where there is sorting/filtering UI like the products tab. I also see the invalid number of rows crash in `ProductsViewController` (by searching ProductsViewController in Sentry) so it'd be nice if this `FetchResultSnapshotsProvider` could work for the products tab as well. Feel free to add the mutable support separately or whenever we get to add DiffableDataSource to the products tab!

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -0,0 +1,156 @@
...
+    ///    return { [weak self] tableView, indexPath, objectID in
+    ///        guard let self = self, let cell = tableView.dequeueReusableCell(withIdentifier: ...) else {
+    ///            fatalError()
+    ///        }
+    ///
+    ///        let order = provider.object(withID: objectID)
+    ///        cell.configure(using: order)
+    ///        return cell
+    ///    }
+    ///}
```
jaclync: super nit:
```suggestion
    /// }
```

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -0,0 +1,156 @@
...
+            return nil
+        }
+    }
+
+    /// Converts the `NSFetchedResultsController` results to a snapshot that is then emitted
+    /// to observers.
+    ///
+    /// This will be hidden later. It should not be publicly accessible.
+    public func controller(_ controller: NSFetchedResultsController<NSFetchRequestResult>,
+                    didChangeContentWith snapshot: NSDiffableDataSourceSnapshotReference) {
```
jaclync, line 152: super cool that we don't need to implement those FRC delegate methods anymore! 🎉

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
```diff
@@ -0,0 +1,239 @@
...
+        }
+        cancellables.removeAll()
+        super.tearDown()
+    }
+
+    func test_objectWithID_returns_the_expected_immutable_object() throws {
+        // Given
+        let insertedAccount = insertAccount(displayName: "Reina Feil", username: "reinafeil")
+
+        viewStorage.saveIfNeeded()
```
jaclync: just curious, any reasons `viewStorage.saveIfNeeded()` is only called in two of the test functions in this class where some models are also inserted?

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
```diff
@@ -0,0 +1,239 @@
...
+                promise(snapshot)
+            }.store(in: &self.cancellables)
+
+            try provider.start()
+        }
+
+        // Then
+        let expectedObjectIDs = [accounts[1].objectID, accounts[2].objectID, accounts[0].objectID]
+        let actualObjectIDs = snapshot.itemIdentifiers
+        XCTAssertEqual(expectedObjectIDs, actualObjectIDs)
```
jaclync: super nit: I think we usually put the actual value before the expected value in `XCTAssertEqual`? Just thought it might be more consistent when reading the error message 🙂 (I think other frameworks/languages might use the reverse order? maybe the code base hasn't been using the more common convention? 😅 )

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
jaclync: just wondering, why only this test case needs to call `viewStorage.obtainPermanentIDs`?

jaclync (approved):
Nice work wrangling the tricky animation issue with temporary/permanent IDs! I love that so much Core Data internals are encapsulated in `FetchResultSnapshotsProvider`, and just the `Query` requires dealing with Core Data / Storage now 😄 

Code: ✅ 
Testing: ✅  (was able to run on an iOS 12 simulator)

TestKit/Sources/TestKit/XCTestCase+Wait.swift
shiki: Not a replacement but an additional helper. 😄 Most of the time, I noticed that we're using `waitUntil` to wait for some value. We can use this instead for that. There are still times that `waitUntil` can be useful though.

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
shiki: Thanks. 😄 Fixed in db5d362cacf573fda6cdd0358c0d2e32cb1beff6.

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
shiki: I don't know why I keep missing `private` in computed properties. This doesn't feel like the first time I did this. 😄 Thanks for catching it! Fixed in 0e617e4903d9194b0fd8d5e1cf3518a602d05049.

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
shiki: This is one of the problems with `FRC`. If I don't save it, then `FetchResultSnapshot` would return **temporary** `NSManagedObjectIDs`. And this would fail:

https://github.com/woocommerce/woocommerce-ios/blob/0e617e4903d9194b0fd8d5e1cf3518a602d05049/Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift#L56

The other tests don't need this because they don't use `provider.object(withID:)`. 

I added a comment to hopefully help 59eb6bb66f27c5b718353331aa28c2850864e856.

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
shiki: I always get confused by this. I want to be consistent but I can't quite _remember_ which parameter is for "actual". 🙃 The [documentation](https://developer.apple.com/documentation/xctest/xctassertequal) doesn't help. 

Do you have a trick for remembering which one goes first? 😄 

Fixed in 4d19ac453b57c45096d002d4aa59741d9de2894b.

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
shiki: @jaclync I used `obtainPermanentIDs` here because I needed the permanent IDs to correctly compare the `objectID` later in the asserts:

https://github.com/woocommerce/woocommerce-ios/blob/0e617e4903d9194b0fd8d5e1cf3518a602d05049/Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift#L218-L226

If I didn't, then the asserts will fail because they were comparing temporary IDs with the permanent IDs. I don't know why FRC would sometimes return permanent IDs or temporary IDs in the snapshot. 🙈 

I added a comment which hopefully helps: de9394f0a67236e199322cd309743ae3a2e4c47d

shiki: > Since obtainPermanentIDs is temporary, maybe we should not rely on it for tests, WDYT?

@Ecarrion As explained in my comment above, I needed it to make the assertions work. I also tried re-querying the objects and do the comparison with the results. But that also didn't work because the predicate for the fetch is using the **temporary** object IDs so the re-query failed because the temporary IDs no longer exist. 😬 

Is it okay if I keep them for now? I think the obtain will be removed later and this test is probably gonna change.

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
shiki: It looks like we could use this:

https://github.com/woocommerce/woocommerce-ios/blob/de9394f0a67236e199322cd309743ae3a2e4c47d/Storage/Storage/Protocols/Object.swift#L8-L10

But it doesn't seem to be a simple refactor. I'll take a look at migrating to that on another PR. 

Thanks for the nudge!

shiki:
> I have to admit that I did not understand how obtainPermanentIDs fixes the animation problem, or will this be tackled in a different PR?

@Ecarrion Yeah, sorry I'm not so good at explaining that. 😄 Since the `obtainPermanentIDs` isn't technically used right now (it's just exposed), I'll give testable examples of the animation problem in a follow-up PR.

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
jaclync: How about this other test case that also compares the `objectID`s:

https://github.com/woocommerce/woocommerce-ios/blob/0e617e4903d9194b0fd8d5e1cf3518a602d05049/Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift#L178-L179

just curious why `obtainPermanentIDs` was necessary in one test case but not the others that also use object IDs 😄

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
jaclync: Yea what a shame the official documentation doesn't have an example 🙄 I think it's muscle memory for me to always pass the actual value first, and I'll be really confused if I ever write in a different language that follows the opposite 😆

jaclync:
@shiki I have another non-blocking question about DiffableDataSource 🙋🏻‍♀️ just thinking ahead for the use case where we want to scroll to a newly created product in the list, do you think we'll just get the index path from the internal FRC in `FetchResultSnapshotsProvider`?

also feel free to merge the PR, the changes LGTM! ✅

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
shiki: @jaclync  Thanks for asking! And I really appreciate you for digging and understanding everything. 

On that test, the FRC returned **temporary IDs**. I didn't see the need to use `obtain` because both the IDs in the `snapshot` and `expectedFirstSection` have temporary IDs and comparing them would pass. 

However, for the `test_snapshot_continuously_emits_values_for_structural_changes`, things are a bit different. There is a call to `saveIfNeeded()`

https://github.com/woocommerce/woocommerce-ios/blob/de9394f0a67236e199322cd309743ae3a2e4c47d/Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift#L215

The `saveIfNeeded()` call was needed to force FRC to generate the second snapshot. It's really confusing why FRC works this way but I think it makes sense.
 
But then, the `saveIfNeeded()` seems to emit snapshots that have permanent IDs. And so the assertions would fail because the objects like `zanza`, `zagato`, etc. were still using temporary IDs. 🤷 

---

I changed everything to use `obtain` for now so they're all consistent and we won't confuse ourselves more in the future. 😅  0cca077f97545f80323990a795486c992e23a47f

shiki:
Thank you, @jaclync and @Ecarrion. I'm gonna merge this now since there are now 3 PRs in the chain. 😄 But let me know if you feel like something needs to be changed.

shiki:
>  just thinking ahead for the use case where we want to scroll to a newly created product in the list, do you think we'll just get the index path from the internal FRC in FetchResultSnapshotsProvider?

@jaclync The way we would do that is:

1. Grab the index path from the **UITableViewDiffableDataSource** using [`indexPath(for:)`](https://developer.apple.com/documentation/uikit/uitableviewdiffabledatasource/3255151-indexpath)
2. Scroll to the row using [`scrollToRow`](https://developer.apple.com/documentation/uikit/uitableview/1614997-scrolltorow). 

We would need the `ObjectID` in this case so I think we'll have to find a way to expose that from somewhere.

Yosemite/YosemiteTests/SnapshotsProvider/FetchResultSnapshotsProviderTests.swift
jaclync: thanks for the explanation and updates! I see that we have to call `obtainPermanentIDs` to make sure the snapshot returns permanent IDs now 👍

Issue #2740, opened 2020-08-31
Create a new product (simple/grouped/external)


jaclync:
- [x] Add a CTA to the products tab ("+") https://github.com/woocommerce/woocommerce-ios/pull/2749
- [x] Networking/Yosemite: add an action to create a product. `ProductAction.addProduct(product: Product, onCompletion: (Result<Product, ProductUpdateError>) -> Void)` https://github.com/woocommerce/woocommerce-ios/pull/2754
- [x] Refactor password/product remote update so that they can be updated differently when adding and editing a product
  - [x] Only retrieve password when editing a product https://github.com/woocommerce/woocommerce-ios/pull/2779
  - [x] Encapsulate async product & password updates into a use case https://github.com/woocommerce/woocommerce-ios/pull/2785
- [x] Create a coordinator to show a bottom sheet for selecting a product type, and navigate to a product form based on the selected product type with default values following core behavior https://github.com/woocommerce/woocommerce-ios/pull/2802
- [x] Fix bottom sheet pointing to the CTA on products tab on iPad https://github.com/woocommerce/woocommerce-ios/pull/2860 and https://github.com/wordpress-mobile/WordPressUI-iOS/pull/72
- [x] Update UI differences from editing a product (ref p91TBi-3f7-p2)
  - [x] Navigation bar "Publish" CTA is always shown (where "Update" CTA is when editing a product) and it creates a product with the status in product settings https://github.com/woocommerce/woocommerce-ios/pull/2846
  - Navigation bar > ellipsis menu:
    - [x] Add "Save as draft" action to create a product with status "draft", regardless of the status in product settings https://github.com/woocommerce/woocommerce-ios/pull/2859
    - [x] Disable "View Product in Store" and "Share" actions https://github.com/woocommerce/woocommerce-ios/pull/2846
  - [x] Unsaved changes action sheet (when the user navigates away from product details with outstanding changes):
    - Add "Save as draft" action (same behavior as “Save as draft” above) https://github.com/woocommerce/woocommerce-ios/pull/2859
  - [x] Hide product type row (it will be editable again in M5 when we support adding a variable product) https://github.com/woocommerce/woocommerce-ios/pull/2846

We're not supporting creating a variable product for M4 since there is no way to create a variation yet.

The whole flow:

[...]

PR #2779, opened 2020-09-07
Add Product > Product Settings: set password to empty without fetching password remotely

jaclync:
This PR includes a subset of backend changes for adding/editing a product #2740 

## Background

When editing a product, the product already has a remote ID and we fetch the product (post)'s password from WP.com API in the product settings screen before the user can tap into the visibility row. On the other hand, when adding a product, there is no password so we don't have to perform this fetch anymore.

This PR also updated password update's completion block to `Result`.

## Changes

- Updated `SitePostAction.updateSitePostPassword` to return `Result<String?, Error>` instead of `(_ password: String?, _ error: Error?)`
  - Networking layer: updated `SitePostsRemote.updateSitePost` similarly
  - In `ProductFormViewController`, updated to use the `Result` type
  - Updated unit tests with Arrange/Action/Assert
- Created `ProductFormType` enum that contains two cases: `.add` for adding a product and `.edit` for editing a product (existing feature)
  - Added `ProductFormType` enum to `ProductFormViewModelProtocol`
    - `ProductFormViewModel` implementation takes in a `ProductFormType` enum and is privately mutable
    - `ProductVariationFormViewModel` implementation always returns `.edit` because we don't plan to support adding a product variation with similar UI/UX
  - Updated unit tests to pass `ProductFormType.edit`
- Only fetching the password via WP.com API when editing a product
  - In `ProductSettingsViewModel`, the password is fetched when editing a product; otherwise, the password is set to empty
  - DI'ed `ProductFormType` from `ProductFormViewController` to `ProductSettingsViewController` to `ProductSettingsViewModel`

## Testing

There's no UI integration for adding a product yet, please sanity check on viewing a product's settings:

### Product without a password

- Go to the Products tab
- Tap on any product that is not password protected
- Tap on the ellipsis (...) menu and then "Product Settings"
- Tap on the "Visibility" row when it's tappable --> the expected visibility row should be selected (public/private)

### Product with a password

- Go to the Products tab
- Tap on any product that is password protected
- Tap on the ellipsis (...) menu and then "Product Settings"
- Tap on the "Visibility" row when it's tappable --> the "Password Protected" row should be selected with the correct password

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/SitePostsRemote.swift
+++ b/Networking/Networking/Remote/SitePostsRemote.swift
@@ -1,5 +1,4 @@
 import Foundation
-import Alamofire
 
 
 /// WP.com Site Posts API: Remote Endpoints
@@ -33,7 +32,7 @@ public class SitePostsRemote: Remote {
     ///     - post: Post that we will use to update the post.
     ///     - completion: Closure to be executed upon completion.
     ///
-    public func updateSitePost(for siteID: Int64, postID: Int64, post: Post, completion: @escaping (Post?, Error?) -> Void) {
+    public func updateSitePost(for siteID: Int64, postID: Int64, post: Post, completion: @escaping (Result<Post, Error>) -> Void) {
         do {
             var parameters = try post.toDictionary()
             let parametersFields = ["fields": "site_ID,password"]
@@ -44,7 +43,7 @@ public class SitePostsRemote: Remote {
 
             enqueue(request, mapper: mapper, completion: completion)
         } catch {
-            completion(nil, error)
+            completion(.failure(error))
         }
     }
 }
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
@@ -24,6 +24,9 @@ final class ProductFormViewModel: ProductFormViewModelProtocol {
         product
     }
 
+    /// The form type could change from .add to .edit after creation.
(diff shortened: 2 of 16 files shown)


Networking/Networking/Remote/SitePostsRemote.swift
```diff
@@ -33,7 +32,7 @@ public class SitePostsRemote: Remote {
     ///     - post: Post that we will use to update the post.
     ///     - completion: Closure to be executed upon completion.
     ///
-    public func updateSitePost(for siteID: Int64, postID: Int64, post: Post, completion: @escaping (Post?, Error?) -> Void) {
+    public func updateSitePost(for siteID: Int64, postID: Int64, post: Post, completion: @escaping (Result<Post, Error>) -> Void) {
```
pmusolino, line 35: Thank you for converting it! 🙇

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -24,6 +24,9 @@ final class ProductFormViewModel: ProductFormViewModelProtocol {
         product
     }
 
+    /// The form type could change from .add to .edit after creation.
+    private(set) var formType: ProductFormType
```
pmusolino, line 28: I like the idea of the Product Form Type 👍

pmusolino (approved):
Looks good to me and thanks for converting the response of SitePostsRemote in `Result < >` 💯 

Code ✅ 
Testing ✅

PR #2784, opened 2020-09-09
SnapshotsProvider: Hide the delegate method

shiki:
Part the DiffableDataSource fix for #1543 and is a continuation of #2777. 

## Problem

In #2777, the `NSFetchedResultsControllerDelegate` method used by `FetchResultSnapshotsProvider` is **`public`**:

https://github.com/woocommerce/woocommerce-ios/blob/de9394f0a67236e199322cd309743ae3a2e4c47d/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L147-L155

And it's `public` because `FetchResultSnapshotsProvider` is public and Swift requires them to match. 

Keeping the as `public` means that consumers of `FetchResultSnapshotsProvider` would be able to access it. 🙈 

```swift
let provider = FetchResultSnapshotsProvider()

// (╯°□°)╯︵ ┻━┻
provider.controller(NSFetchedResultsController(), didChangeContentWith: NSDiffableDataSourceSnapshot())
```

It's hard to call it but it's still possible.

## Solution

In this PR, I hid the method by using a hidden delegate class. 

https://github.com/woocommerce/woocommerce-ios/blob/b531afb23ffa098085ca6ef6c012e4c85e19dfc3/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L153-L174

The delegate class access the parent class (`FetchResultSnapshotsProvider`) directly and emits the snapshot. 

An alternative I could have used is to call a block like the deisgn of [`FetchedResultsControllerDelegateWrapper`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Tools/Internal/FetchedResultsControllerDelegateWrapper.swift). But I chose to avoid for now since I've seen how Sentry weirdly handles blocks in `FetchedResultsControllerDelegateWrapper` (see p91TBi-2OT-p2).

## Testing

`FetchResultSnapshotsProvider` is currently unused but is covered with a unit test. Please make sure the build is passing.

## Submitter Checklist

### DIFF
--- a/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
+++ b/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
@@ -48,7 +48,7 @@ public typealias FetchResultSnapshot = NSDiffableDataSourceSnapshot<String, Fetc
 /// That's it! The `UITableView` should be automatically updated whenever the data changes.
 ///
 @available(iOS 13.0, *)
-public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshotsProviderMutableType>: NSObject, NSFetchedResultsControllerDelegate {
+public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshotsProviderMutableType> {
 
     /// Defines the conditions for fetching the results.
     public struct Query {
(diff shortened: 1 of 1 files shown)


jaclync (approved):
This is a very thoughtful change! ✅ I wonder if the class doesn't have generics, it's probably possible to implement `NSFetchedResultsControllerDelegate` at internal level? not worth sacrificing generics for that though :)

shiki:
> I wonder if the class doesn't have generics, it's probably possible to implement NSFetchedResultsControllerDelegate at internal level? not worth sacrificing generics for that though :)

I haven't thought about generics. 😄 The only reason I used a nested class was because the class is highly coupled with `FetchedResultSnapshotsProvider`. 🙈 

https://github.com/woocommerce/woocommerce-ios/blob/b531afb23ffa098085ca6ef6c012e4c85e19dfc3/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L171-L172

Thanks for the review!

PR #2781, opened 2020-09-08
Improve unexpected product update error handling

jaclync:
Fixes #2636 

- [x] ⚠️ Potential merge conflicts with https://github.com/woocommerce/woocommerce-ios/pull/2785 ⚠️ 

## Changes

This PR implements the proposed handling by @shiki in https://github.com/woocommerce/woocommerce-ios/issues/2636#issuecomment-673151434:

- Added an associated value `error: Error` to `ProductUpdateError.unknown` case
  - Unfortunately, because `Error` is generic we have to manually implement `Equatable` for `ProductUpdateError` enum for existing asserts in unit tests
- Updated `ProductUpdateError` to implement `LocalizedError` that returns the existing user-facing error message plus the `unknown` case ("Unexpected error")
- In `ProductFormViewController`, logged the `Error` instead of just the localized description via `DDLog` and `CrashLogging` (Sentry)

## Testing

Please test for an error case, and sanity check on the existing error cases:

### API error in Booking product update

Thanks to @rachelmcr for finding this that makes it easy to test this PR!

Note that Sentry non-fatal logging is not testable yet, https://github.com/woocommerce/woocommerce-ios/pull/2775 potentially fixes it.

Please follow the steps to repro in https://github.com/woocommerce/woocommerce-ios/issues/2780:

Prerequisite: Add the WooCommerce Bookings extension to your store.

- Add at least one bookable product
- Go to Settings > Experimental Features, and turn on the Products switch
- Go to the Products tab
- Tap on the bookable product
- Try editing some of the editable fields
- Tap "Update" --> an error alert should appear with "Unexpected error" message, and the error is shown in the debug console (`⛔️ Error updating Product: unknown(error: Dotcom Error: [rest_invalid_param] Invalid parameter(s): type)`)

### Duplicate SKU

- Go to the Products tab
- Tap on a simple product
- Tap on the inventory row
- Enter a SKU that is used in another product --> there should be an error message like before ("SKU already in use by another product")

### Invalid SKU update

- Go to the Products tab
- Tap on a simple product
- Tap on the inventory row
- Enter a SKU that is used in another product **and quickly tap "Done"** before the inline error shows up
- Tap "Update" --> an error alert should appear with message "This SKU is used on another product or is invalid.", and the error is shown in the debug console (`⛔️ Error updating Product: invalidSKU`)

## Example screenshot

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2785, opened 2020-09-09
Encapsulate async product & password updates into a use case

jaclync:
Backend support for #2740 

- [x] ⚠️ Potential merge conflicts with https://github.com/woocommerce/woocommerce-ios/pull/2781 ⚠️ 

Apology for the bigger PR size, there are a lot of boilerplates setting up for each test case of `ProductFormRemoteActionUseCase` since I iterated through all the possible combinations of product & password results that I can think of. Please lemme know if the test cases can be simplified!

## Context

When adding/editing a product, we update not only the **product via Woo API** but also the **password via WP.com API** (ideally, this could be part of the Woo API). When editing a product, the order of updating a product/password doesn't matter because the product/post remote ID already exists for both API updates so that they can run in parallel. However, when adding a product, a product/post remote ID is only available after creating it remotely. Thus, we have to create the product first then update its password. To summarize the behaviors:

- Editing a product: product update API call and password update API call can run in parallel.
- Creating a product: product creation API call is made first, and if successful the password update API call is made.

## Changes

- Created `ProductFormRemoteActionUseCase` that encapsulates the different async logic when creating and editing a product based on the logic above
  - Added unit tests for possible scenarios
- Updated `ProductFormViewModel.updateProductRemotely` to call `ProductFormRemoteActionUseCase` based on the form type (`.add` or `.edit`)
- Updated `ProductFormViewController` so that the product & password API calls are both encapsulated in the view model, removed unused `func hasProductChanged() -> Bool` and `func hasPasswordChanged() -> Bool` from `ProductFormViewModelProtocol`
- Added a new enum case for password update error `ProductUpdateError.password`

## Testing

No behavior changes are intended, please test for regressions when editing a product. The product creation UI integration will be in a separate PR.

### Adding password to a product

- Go to the Products tab
- Tap on any product that is not password protected
- Tap on the ellipsis (...) menu and then "Product Settings"
- Tap on the "Visibility" row when it's tappable
- Select the "Password Protected" row, and enter a non-empty password
- Navigate back to product settings and tap "Done"
- Tap "Update" --> the product should have the password set remotely

### Removing password from a product

- Go to the Products tab
- Tap on any product that is password protected
- Tap on the ellipsis (...) menu and then "Product Settings"
- Tap on the "Visibility" row when it's tappable
- Select a different visibility
- Navigate back to product settings and tap "Done"
- Tap "Update" --> the product's visibility should be updated remotely

### Editing product but not password

- Go to the Products tab
- Tap on any product
- Update any fields that you like
- Tap "Update" --> the product be updated remotely

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormRemoteActionUseCase.swift
```diff
@@ -0,0 +1,152 @@
...
+
+        group.enter()
+        updatePasswordRemotely(product: product, password: password, originalPassword: originalPassword) { result in
+            passwordResult = result
+            group.leave()
+        }
+
+        group.notify(queue: .main) {
+            guard let productResult = productResult, let passwordResult = passwordResult else {
+                assertionFailure("Unexpected nil result after updating product and password remotely")
```
shiki, line 80: ⚠️ I'd prefer it if we propagate this failure instead. The worst case is that the async update will stall because we would never call `onCompletion`. 

```swift
                onCompletion(.failure(.unknownIThinkSmileyFace), nil)
```

Keeping the `assertionFailure` is still okay for me. It's the stalled async call that worries me. 😄

Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -565,11 +565,13 @@ extension ProductStore {
 ///
 /// - duplicatedSKU: the SKU is used by another Product.
 /// - invalidSKU: the SKU is invalid or duplicated.
+/// - password: the password of a product cannot be updated.
 /// - unknown: other error cases.
 ///
 public enum ProductUpdateError: Error {
     case duplicatedSKU
     case invalidSKU
+    case password
```
shiki: It would be nice if the `case` itself is descriptive too. 🙂 

```suggestion
    case passwordCannotBeUpdated
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -239,19 +231,57 @@ extension ProductFormViewModel {
...
+                        onCompletion(.failure(.unknown))
+                        return
+                    }
+                    switch passwordResult {
+                    case .failure(let passwordError):
+                        onCompletion(.failure(.password))
+                    case .success(let password):
+                        self?.formType = .edit
+                        self?.resetProduct(product)
+                        self?.resetPassword(password)
```
shiki: Should we add `guard let self = self`? 😁 Sorry. It's a pet-peeve now at this point. 🙃

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -239,19 +231,57 @@ extension ProductFormViewModel {
...
-                onCompletion(.failure(error))
-            case .success(let product):
-                ServiceLocator.analytics.track(.productDetailUpdateSuccess)
-                let model = EditableProductModel(product: product)
-                self?.resetProduct(model)
-                onCompletion(.success(model))
+        let remoteActionUseCase = ProductFormRemoteActionUseCase()
+        switch formType {
+        case .add:
+            remoteActionUseCase.addProduct(product: product, password: password) { [weak self] productResult, passwordResult in
```
shiki: ⚠️  For both `addProduct` and `editProduct` methods, I notice that it's a bit unclear _when_ errors should be handled.

- Should I handle errors in `productResult`?
- Should I handle errors in `passwordResult` first?

I feel like there's an opportunity to merge the errors into one and keep this `ViewModel` simpler. 

Would it be possible to have a completion block like this?

```swift
    typealias Completion = (Result<(product: EditableProductModel, password: String?), Error>) -> Void
```

And this `ViewModel` would _trust_ that the `UseCase` will correctly propagate the right error. Perhaps this block here would then be simplified to something like:

```swift
            remoteActionUseCase.addProduct(product: product, password: password) { [weak self] result in
                switch result {
                case .failure (let error):
                    onCompletion(.failure(productError))
                case .success ((let product, let password)):
                    self.formType = .edit
                    self.resetProduct(product)
                    self.resetPassword(password)
                    onCompletion(.success(product))
                }
            }
```

Just an idea for discussion. Not sure if it will really work or not. 😬

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -239,19 +231,57 @@ extension ProductFormViewModel {
...
+                case .failure(let productError):
+                    onCompletion(.failure(productError))
+                case .success(let product):
+                    guard let passwordResult = passwordResult else {
+                        assertionFailure("Password result is expected to be non-nil when product result is successful.")
+                        onCompletion(.failure(.unknown))
+                        return
+                    }
+                    switch passwordResult {
+                    case .failure(let passwordError):
```
shiki: I see a build warning here. 🤪

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormRemoteActionUseCaseTests.swift
```diff
@@ -0,0 +1,310 @@
...
+        useCase.addProduct(product: model, password: password) { aProductResult, aPasswordResult in
+            productResult = aProductResult
+            passwordResult = aPasswordResult
+        }
+
+        // Assert
+        XCTAssertEqual(storesManager.receivedActions.count, 2)
+        XCTAssertNotNil(storesManager.receivedActions[0] as? ProductAction)
+        XCTAssertNotNil(storesManager.receivedActions[1] as? SitePostAction)
+        XCTAssertEqual(productResult, .success(model))
```
shiki: Nice! I never thought we could assert `Result<>` like this. 👏

shiki (changes requested):
Thanks for making the code better, @jaclync. I have some discussion points I'd like to thoughts on (⚠️). The others are just nitpicks. 🙃

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormRemoteActionUseCase.swift
jaclync: my original thought for not passing anything to the completion block was because it's theoretically impossible to reach here (`group.notify` block) without both dispatch blocks complete where both results are set. It might be confusing to read though, I'll update to add a call to `onCompletion`

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
jaclync: I was originally going to pass the password error as an associated value, but I actually don't have a use for it yet. Updated in https://github.com/woocommerce/woocommerce-ios/pull/2785/commits/ec900ac19483a08c697256300524b4cf15b86bcf 🙂

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
jaclync: Just passing one result actually sounds cleaner keeping the two results internal, thanks for the suggestion! (like if one day Woo API supports updating the password in the product update endpoint) updated in https://github.com/woocommerce/woocommerce-ios/pull/2785/commits/9070211337b66f29a84e3bbee28c1cf8312d91d8 and https://github.com/woocommerce/woocommerce-ios/pull/2785/commits/ec900ac19483a08c697256300524b4cf15b86bcf (I created a struct `ResultData` as a wrapper for the product and password, since a tuple can't be compared in unit tests)

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormRemoteActionUseCase.swift
jaclync: updated in https://github.com/woocommerce/woocommerce-ios/pull/2785/commits/ec900ac19483a08c697256300524b4cf15b86bcf (I created another error case `.unexpected` to differentiate from `.unknown` that is for other errors that have an associated error in https://github.com/woocommerce/woocommerce-ios/pull/2781)

Yosemite/Yosemite/Stores/ProductStore.swift
jaclync: updated in: https://github.com/woocommerce/woocommerce-ios/pull/2785/commits/282dfd9948a658854eefeb749d2be10a491d5362

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
jaclync: yup! updated in https://github.com/woocommerce/woocommerce-ios/pull/2785/commits/282dfd9948a658854eefeb749d2be10a491d5362

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormRemoteActionUseCase.swift
shiki: Nice! Yeah, I agree with the technically impossible point. I'm more concerned about my future self possibly editing this code and making a mistake (highly probable). 🤪

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
shiki: Tiny structs are great!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -239,19 +231,41 @@ extension ProductFormViewModel {
...
+                case .failure(let error):
+                    onCompletion(.failure(error))
+                case .success(let data):
+                    guard let self = self else {
+                        return
+                    }
+                    self.formType = .edit
+                    self.resetProduct(data.product)
+                    self.resetPassword(data.password)
+                    onCompletion(.success(data.product))
```
shiki, line 248: So much simpler! 🤸

shiki (approved):

jaclync:
Thank you Jayson for the feedback to make the code better 💟 merging now!

PR #2790, opened 2020-09-09
CurrencySettings: it no more depends on `ResultController` and `ServiceLocator`

pmusolino:
## Description
As discussed here p91TBi-3bP since it will come in handy in the future, it would be good to not expose `ResultController` and `ServiceLocator` to the `CurrencySettings` class, that will be useful in the future for using it in a Widget extension. In that way, we don't need to expose to a potential widget these two big dependencies.
I created a `SelectedSiteSettings` class where I moved all the logic about fetching the latest Site Settings, that takes care also to update the Currency Settings used in the app, and in the future will be helpful also for other things (for example, for selecting the Site Country. I've in the plan to open a separate PR for it).

## Testing
- CI should pass
- **Sanity Check:**
1. Open a store with a specific currency, and navigate to the Order List.
2. The prices of the orders should be correct like in `develop`.
3. Switch to another store that uses a different currency.
4. Navigate to the Order List -> Prices should be updated using the new currency value.

Update release notes:


WooCommerce/Classes/Tools/Shared Site Settings/SelectedSiteSettings.swift
```diff
@@ -0,0 +1,57 @@
...
+            DDLogError("Error: no siteID found when attempting to refresh CurrencySettings results predicate.")
+            return
+        }
+
+        let sitePredicate = NSPredicate(format: "siteID == %lld", siteID)
+        let settingTypePredicate = NSPredicate(format: "settingGroupKey ==[c] %@", SiteSettingGroup.general.rawValue)
+        resultsController.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [sitePredicate, settingTypePredicate])
+        try? resultsController.performFetch()
+        resultsController.fetchedObjects.forEach {
+            CurrencySettings.shared.updateCurrencyOptions(with: $0)
```
Ecarrion: If I understand the `ServiceLocator` pattern correctly, shouldn't this access the `CurrencySettings` though the `ServiceLocator` so we can access the mocked/"injected" instance and not the singleton directly?

I wonder if a way to improve this singleton management is to make those `shared` accessors private to `ServiceLocator` and somehow forbid any `shared` usage on the codebase? 😅 

cc @woocommerce/ios-developers

Ecarrion (approved):
Code looks good! I did leave some questions about singleton handling but that may be not a topic for this PR 😇

WooCommerce/Classes/Tools/Shared Site Settings/SelectedSiteSettings.swift
pmusolino: Do you mean to remove `CurrencySettings` from being a singleton, and use its instance by adding a new static variable in `ServiceLocator`? If you mean this, I think that it's a good idea to not have an extra singleton, and use it as a global dependency inside the `ServiceLocator`

Ecarrion: Yes, something like that!

pmusolino: Sounds good to me, I'm going to edit the code for including `CurrencySettings` in `ServiceLocator` 👍 I think it makes sense

pmusolino:
Hey @Ecarrion I made all the changes suggested, updating tests and moved `CurrencySettings` under `ServiceLocator`. Can you take a second look at it, please? 🙏

Ecarrion (approved):
Looks great! 💪

pmusolino:
Thank you so much @Ecarrion for your quick review! 💯

WooCommerce/Classes/Tools/Shared Site Settings/SelectedSiteSettings.swift
shiki: Noice! I'll probably get to documenting this in HACK Week. 🤞

Issue #1962, opened 2020-03-13
Orders: Refreshing the Processing tab doesn't sync order status with remote changes


rachelmcr:
**Describe the bug**

If an order with `Processing` status has its status updated remotely (e.g. the order is refunded in WP Admin), it isn't immediately removed from the Processing tab in the app. Pulling to refresh the Processing tab also doesn't update the order status or remove that order from the list.

The only way I found to update the order to the correct status is to switch to the All Orders tab and pull to refresh, or open the order detail and pull to refresh it there.

**To Reproduce**
1. Go to the Orders tab.
2. Select the Processing tab and identify a recent order in the list.
3. In WP Admin web, open that order and fully refund it (confirm the order status is changed to `Refunded`).
4. Open the app again and notice that the order is still in the list on the Processing tab.
5. Pull to refresh the list on the Processing tab. Notice the order is still in the list.
6. Switch to the All Orders tab and pull to refresh. Notice that the order status updates and the order is now removed from the list on the Processing tab.

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 13.3.1
- WooCommerce iOS version: 3.8.0.0 TestFlight beta

PR #2791, opened 2020-09-09
Move Order List Sync Action Logic to UseCase

shiki:
Part of the DiffableDataSource fix for #1543. This is a continuation of #2784.  

😱 This is 500+ lines but it's a simple change. I just moved some logic and tests. 

## What

To implement the DiffableDataSource fix for #1543, I am planning to create **new** [OrderListViewController](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift) and ViewModel classes. These classes will be used if the device is using iOS 13.0. This allows us to use the DiffableDataSource while keeping the older (FetchedResultsController) classes for iOS 12.0 users. Eventually, we will remove the old classes. 

For now, there will be some redundancy. In this PR, I'm trying to eliminate a little bit of redundancy by moving the most important logic, the [sync action logic in the ViewModel](https://github.com/woocommerce/woocommerce-ios/blob/560430494033065825ea291668b3e36709f503f3/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L145-L240), to a `*UseCase` called [`OrderListSyncActionUseCase`](https://github.com/woocommerce/woocommerce-ios/blob/issue/1543-move-orders-vm-action/WooCommerce/Classes/ViewRelated/Orders/OrderListSyncActionUseCase.swift). This logic was used to fix this bug a long time ago: https://github.com/woocommerce/woocommerce-ios/issues/1962.

Here's a diagram of the plan:

image

Please let me know if you have any concerns. 

## Testing

The changes were previously covered by unit tests. Please take a quick look at the Order list to make sure the fetching is not broken. 

## Submitter Checklist


Ecarrion (approved):
It all looks great to me! 🚀

shiki:
Thank you, Ernesto!

PR #2792, opened 2020-09-09
Refactor of SiteCountry, that now use SelectedSiteSettings

pmusolino:
## Description
In a [previous PR](https://github.com/woocommerce/woocommerce-ios/pull/2790), I introduced the `SelectedSiteSettings` class where I moved all the logic about fetching the latest Site Settings and takes care also to update the Currency Settings used in the app.
Since `SiteCountry` was using `ResultsController` for fetching the interested site setting, I substituted it with `SelectedSiteSettings` and I added also a test case since it's easy to introduce a regression here.

## Testing (Sanity Check)
1. Navigate to an order detail.
2. Press "Add Tracking" (you need the shipping extension on your store).
3. Fill the shipping carrier, the tracking number, and date shipped. Save it. 
4. Everything should work like before.

Update release notes:


jaclync (commented):
💬 After also checking the [previous PR](https://github.com/woocommerce/woocommerce-ios/pull/2790), it looks like multiple classes depend on `[SiteSetting]`. How about making `SelectedSiteSettings` available in the `ServiceLocator` instead of the `CurrencySettings`, which can be derived from `SelectedSiteSettings.siteSettings` similar to `SiteCountry`? It seems that both `CurrencySettings` and `SiteCountry` already have a designated/convenient initializer that takes in `siteSettings: [SiteSetting]` so perhaps we can just replace `init(siteSettings: [SiteSetting] = SelectedSiteSettings.shared.siteSettings)` with `init(siteSettings: [SiteSetting] = ServiceLocator.selectedSiteSettings.siteSettings)`? This way we won't need a static singleton `shared` for `SelectedSiteSettings` 😄

loremattei:
Hey there! I'm moving this to 5.2 because 5.1 has been cut. If you want this to make it to 5.1, please feel free to ping me.

pmusolino:
Sorry for my late response Jaclyn, I'm handling it after the hack week 🙇 

> How about making `SelectedSiteSettings` available in the `ServiceLocator` instead of the `CurrencySettings`, which can be derived from `SelectedSiteSettings.siteSettings` similar to `SiteCountry`? It seems that both `CurrencySettings` and `SiteCountry` already have a designated/convenient initializer that takes in `siteSettings: [SiteSetting]` so perhaps we can just replace `init(siteSettings: [SiteSetting] = SelectedSiteSettings.shared.siteSettings)` with `init(siteSettings: [SiteSetting] = ServiceLocator.selectedSiteSettings.siteSettings)`? This way we won't need a static singleton `shared` for `SelectedSiteSettings`

This is a good idea. Done here 1444699d3d7c72f436b2681735c057c9ecaf479d
Note: I added the tests in `ServiceLocatorTests` and renamed also all the test functions according to the new naming convention.

Can you take a second look please?

WooCommerce/Classes/Tools/Shared Site Settings/SelectedSiteSettings.swift
```diff
@@ -34,8 +34,9 @@ extension SelectedSiteSettings {
     /// Setup: ResultsController
     ///
     private func configureResultsController() {
-        resultsController.onDidChangeObject = { (object, indexPath, type, newIndexPath) in
+        resultsController.onDidChangeObject = { [weak self] (object, indexPath, type, newIndexPath) in
             ServiceLocator.currencySettings.updateCurrencyOptions(with: object)
```
jaclync, line 39: nit: maybe we can add `guard let self = self else { return }` before this line so that both this line and the line below with `self?` are either both these two lines are executed or not?

WooCommerce/WooCommerceTests/Tools/ServiceLocatorTests.swift
```diff
@@ -4,78 +4,82 @@ import XCTest
...
         XCTAssertTrue(pushNotes is PushNotificationsManager)
     }
 
-    func testServiceLocatorProvidesAuthenticationManager() {
+    func test_ServiceLocator_provides_authenticationManager() {
         XCTAssertNotNil(ServiceLocator.authenticationManager)
     }
 
-    func testAutenticationManagerDefaultsAuthenticationManager() {
+    func test_autenticationManager_defaults_AuthenticationManager() {
```
jaclync: nit:
```suggestion
    func test_autenticationManager_defaults_to_AuthenticationManager() {
```

WooCommerce/WooCommerceTests/Tools/ServiceLocatorTests.swift
```diff
@@ -4,78 +4,82 @@ import XCTest
...
+    func test_ServiceLocator_provides_selectedSiteSettings() {
+        XCTAssertNotNil(ServiceLocator.selectedSiteSettings)
+    }
+
+    func test_ServiceLocator_provides_storageManager() {
         XCTAssertNotNil(ServiceLocator.storageManager)
     }
 
-    func testServiceLocatorProvidesLogger() {
+    func test_ServiceLocator_Provides_fileLogger() {
```
jaclync: nit:
```suggestion
    func test_ServiceLocator_provides_fileLogger() {
```

WooCommerce/WooCommerceTests/Tools/ServiceLocatorTests.swift
```diff
@@ -4,78 +4,82 @@ import XCTest
 
 final class ServiceLocatorTests: XCTestCase {
 
-    func testServiceLocatorProvidesAnalytics() {
+    func test_ServiceLocator_provides_analytics() {
```
jaclync, line 7: thanks for renaming the rest of the test class!

jaclync (approved):
Sorry I only got back to this PR today! The changes LGTM 👍 

Code: ✅ 
Testing: ✅

Issue #1838, opened 2020-02-07
Refactor each tab view controller to be associated with a constant site ID


jaclync:
Currently, each of our tab view controller (`DashboardViewController`, `OrdersMasterViewController`, `ProductsViewController`, `ReviewsViewController`) is created only once per app lifetime. Thus, it has to rely on the default `NotificationCenter` to update its data whenever the site changes from site switcher or logging out and in. We've had bugs like https://github.com/woocommerce/woocommerce-ios/issues/1737 related to the site change notification, and we have many instances of checking `sessionManager.defaultStoreID` for the site ID because the site could change during the class' lifetime.

I think we can start with replacing the tab view controllers whenever the site changes (e.g. on the `defaultAccountWasUpdated` notification). Then we can add a constant site ID to each tab view controller, and replace the usage of `sessionManager.defaultStoreID`.

Prep
- [x] Reviews tab: move VCs away from Storyboard https://github.com/woocommerce/woocommerce-ios/pull/1919
- [x] Dashboard tab: move VCs away from Storyboard https://github.com/woocommerce/woocommerce-ios/pull/2819

DI site ID
- [x] Reviews tab: DI site ID to VCs, and tear down the root VC when the site changes
- [x] Dashboard tab: DI site ID to VCs, and tear down the root VC when the site changes
- [x] Orders tab: DI site ID to VCs, and tear down the root VC when the site changes
- [x] Products tab: DI site ID to VCs, and tear down the root VC when the site changes

shiki:
From @astralbodies 

> Don’t forget to consider the iPad interface and what it looks like if a controller is recreated.

jaclync:
All tabs done now, closing this!

PR #2796, opened 2020-09-10
Reviews: delete stored Product Reviews when syncing Reviews from the first page

### LINKED ISSUE #2335: Reviews: Remotely trashed review remains in app
**Describe the bug**

When a review is trashed remotely (e.g. via the web) it remains in the app, even after pulling to refresh or force-closing the app.

The trashed review wasn't removed from the app until I completely refreshed the Reviews screen, e.g. by switching stores and then switching back.

I'm not sure if this is related to https://github.com/woocommerce/woocommerce-ios/issues/1908 (an issue where reviews trashed from **within** the app remain in the app).

**To Reproduce**

1. View the Reviews list in the app.
2. On the web, trash a review (from the product details screen).
3. In the app, pull to refresh the Reviews screen and notice the review remains.
4. Notice you can open the review in the app, navigate to others and back, and even force-close the app and the review remains there.

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 13.4.1
- WooCommerce iOS version: 4.2.0.1 and 4.3.0.0 TestFlight Beta

pmusolino:
Fixes #2335 

## Changes
- Moved all the Reviews classes under the folder `Reviews` (before the name of the folder was Notifications).
- Deleted all stored Product Reviews for a given site ID after successfully loading the first page of Reviews from the API and before inserting the data from the API to the storage.
- Added 3 test cases for `ProductReviewStore` for this change.
- Updated the name of all the tests in `ProductReviewStore`.

## Testing
1. View the Reviews list in the app.
2. On the web, trash a review (from the product details screen) or set it as spam.
3. In the app, pull to refresh the Reviews screen and notice the review is correctly removed from the list.

Update release notes:


Yosemite/Yosemite/Stores/ProductReviewStore.swift
```diff
@@ -81,6 +81,10 @@ private extension ProductReviewStore {
                 return
             }
 
+            if pageNumber == 1 {
```
jaclync: nit: we usually compare with `Default.firstPageNumber` like in `ProductCategoryStore`

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,134 @@ final class ProductReviewStoreTests: XCTestCase {
...
+    ///
+    func test_syncing_product_reviews_on_the_first_page_resets_stored_product_reviews() {
+
+        // Given
+        let reviewID1: Int64 = 1
+        let reviewID2: Int64 = 2
+        let productReviews = [self.sampleProductReview(reviewID: reviewID1), self.sampleProductReview(reviewID: reviewID2)]
+        self.store.upsertStoredProductReviews(readOnlyProductReviews: productReviews,
+                                              in: self.viewStorage,
+                                              siteID: self.sampleSiteID)
```
jaclync: nit: `self` probably not needed?

```suggestion
        store.upsertStoredProductReviews(readOnlyProductReviews: productReviews,
                                          in: viewStorage,
                                          siteID: sampleSiteID)
```

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,134 @@ final class ProductReviewStoreTests: XCTestCase {
...
+
+                exp.fulfill()
+            }
+            store.onAction(action)
+        }
+    }
+
+    /// Verifies that `ProductReviewAction.synchronizeProductReviews` after the first page does not delete stored Product Reviews for the given
+     /// site ID.
+     ///
```
jaclync: super nit!

```suggestion
    /// site ID.
    ///
```

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -206,7 +331,7 @@ final class ProductReviewStoreTests: XCTestCase {
 
     /// Verifies that `ProductReviewAction.retrieveProductReview` returns an error whenever there is an error response from the backend.
     ///
-    func testRetrieveSingleProductReviewReturnsErrorUponReponseError() {
+    func test_retrieve_single_product_review_returns_rrror_upon_reponse_error() {
```
jaclync:
```suggestion
    func test_retrieve_single_product_review_returns_error_upon_reponse_error() {
```

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,134 @@ final class ProductReviewStoreTests: XCTestCase {
...
+
+                exp.fulfill()
+            }
+            store.onAction(action)
+        }
+    }
+
+    /// Verifies that `ProductReviewAction.synchronizeProductReviews` for the first page does not delete stored Product Reviews if the API call fails.
+    ///
+    func test_syncing_product_reviews_on_the_first_page_doest_not_delete_stored_product_reviews_upon_response_error() {
```
jaclync:
```suggestion
    func test_syncing_product_reviews_on_the_first_page_does_not_delete_stored_product_reviews_upon_response_error() {
```

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,134 @@ final class ProductReviewStoreTests: XCTestCase {
...
+
+        // Given
+        let reviewID1: Int64 = 1
+        let reviewID2: Int64 = 2
+        let productReviews = [self.sampleProductReview(reviewID: reviewID1), self.sampleProductReview(reviewID: reviewID2)]
+        self.store.upsertStoredProductReviews(readOnlyProductReviews: productReviews,
+                                              in: self.viewStorage,
+                                              siteID: self.sampleSiteID)
```
jaclync: nit: `self` probably not needed?

```suggestion
        let productReviews = [sampleProductReview(reviewID: reviewID1), sampleProductReview(reviewID: reviewID2)]
        store.upsertStoredProductReviews(readOnlyProductReviews: productReviews,
                                              in: viewStorage,
                                              siteID: sampleSiteID)
```

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,134 @@ final class ProductReviewStoreTests: XCTestCase {
...
+                    siteID: self.sampleSiteID,
+                    reviewID: reviewID1)
+                XCTAssertNotNil(storedProductReview1)
+
+                let storedProductReview2 = self.viewStorage.loadProductReview(
+                    siteID: self.sampleSiteID,
+                    reviewID: reviewID2)
+                XCTAssertNotNil(storedProductReview2)
+
+                XCTAssertGreaterThan(self.viewStorage.countObjects(ofType: Storage.ProductReview.self), 2)
```
jaclync, line 203: nit: similar to the above comment about moving these to `// Then`!

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,134 @@ final class ProductReviewStoreTests: XCTestCase {
...
+                // The previously upserted Product Reviews should be deleted.
+                let storedProductReview1 = self.viewStorage.loadProductReview(
+                    siteID: self.sampleSiteID,
+                    reviewID: reviewID1)
+                XCTAssertNil(storedProductReview1)
+
+                let storedProductReview2 = self.viewStorage.loadProductReview(
+                    siteID: self.sampleSiteID,
+                    reviewID: reviewID2)
+                XCTAssertNil(storedProductReview2)
```
jaclync, line 159: nit: I think the wait for `ProductReviewAction.synchronizeProductReviews` could be moved to the `// When` section, and then these asserts can be under the `// Then` section?

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,134 @@ final class ProductReviewStoreTests: XCTestCase {
...
+                    siteID: self.sampleSiteID,
+                    reviewID: reviewID1)
+                XCTAssertNotNil(storedProductReview1)
+
+                let storedProductReview2 = self.viewStorage.loadProductReview(
+                    siteID: self.sampleSiteID,
+                    reviewID: reviewID2)
+                XCTAssertNotNil(storedProductReview2)
+
+                XCTAssertEqual(self.viewStorage.countObjects(ofType: Storage.ProductReview.self), 2)
```
jaclync, line 245: nit: similar to the earlier comment about When/Then sections!

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -77,7 +77,7 @@ final class ProductReviewStoreTests: XCTestCase {
 
     /// Verifies that ProductReviewAction.synchronizeProductReviews effectively persists any retrieved product reviews.
     ///
-    func testRetrieveProductReviewsEffectivelyPersistsRetrievedProductReviews() {
+    func test_retrieve_product_reviews_effectively_persists_retrieved_product_reviews() {
```
jaclync, line 80: thanks for updating all the test case names!

jaclync (commented):
Thanks for the p2 post discussion in addition to working on this bug!

Code: ✅ (mostly nits)
Testing: I noticed some issues most likely from the `viewWillAppear` sync to `viewDidLoad`:

- When I first launched the app from the PR branch (logged in), the Reviews tab first showed the empty screen (same screenshot as below) while it was actually syncing --> it might be nice to keep the placeholder state as before?
- After switching stores, the Reviews tab is always empty and I cannot pull down to refresh either



Maybe we need to sync again when the site changes, `ProductsViewController` might be a good reference if we want to move the sync from `viewWillAppear` sync to `viewDidLoad`.

pmusolino:
Thank you so much @jaclync for your review and for discovering these issues! Can you take a second look at the PR, considering also two other cases:

- Switching between two stores and going to the reviews tab
- Login into the app and navigate to the reviews tab

Everything should be in place now. I also think that the state machine and the management of data source it's a little bit over-engineered in the reviews. I'd like in the future to refactor the view controller and the data source, and simplify the methods inside `ReviewsViewController` because there are some nested methods, so you risk to call a method two or more times, and you don't know it.

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
```diff
@@ -122,7 +123,9 @@ final class ReviewsViewController: UIViewController {
         resetApplicationBadge()
         transitionToResultsUpdatedState()
 
-        syncingCoordinator.synchronizeFirstPage()
+        if isEmpty {
+            transitionToSyncingState(pageNumber: SyncingCoordinator.Defaults.pageFirstIndex)
```
jaclync, line 127: I'm a bit confused by calling the syncing UI manually here, since `transitionToSyncingState` is usually triggered by `SyncingCoordinator`'s delegate callback 🤔 I don't have a store that has empty reviews to test, but I wonder if revisiting the reviews tab without any actual reviews will get stuck at placeholder UI state?

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
```diff
@@ -291,7 +294,8 @@ private extension ReviewsViewController {
 
         tableView.setContentOffset(.zero, animated: false)
         tableView.reloadData()
-        transitionToSyncingState()
+
+        transitionToSyncingState(pageNumber: SyncingCoordinator.Defaults.pageFirstIndex)
```
jaclync: is this call necessary? since `reloadResultsController` is only called in `defaultSiteWasUpdated` and is followed by `syncingCoordinator.resynchronize()` which triggers syncing the first page, I think it's safe to let the sync callback transition to the expected state after switching sites?

I understand that this call is pre-existing, just thought it might simplify the code a bit by just letting the syncing callback triggers UI state transitions :)

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
```diff
@@ -363,7 +367,8 @@ private extension ReviewsViewController {
     ///
     @objc func defaultSiteWasUpdated() {
         reloadResultsController()
-        syncingCoordinator.resetInternalState()
+        syncingCoordinator.resynchronize()
+        tableView.reloadData()
```
jaclync: nit: is this call necessary? I commented this line and it seems to work still

Yosemite/YosemiteTests/Stores/ProductReviewStoreTests.swift
```diff
@@ -123,9 +123,136 @@ final class ProductReviewStoreTests: XCTestCase {
...
+
+        // When
+        network.simulateResponse(requestUrlSuffix: "products/reviews", filename: "generic_error")
+        XCTAssertEqual(viewStorage.countObjects(ofType: Storage.ProductReview.self), 2)
+
+        waitForExpectation { exp in
+            let action = ProductReviewAction.synchronizeProductReviews(siteID: sampleSiteID,
+                                                                       pageNumber: defaultPageNumber,
+                                                                       pageSize: defaultPageSize) { error in
+        // Then
```
jaclync: nit: indentation 😄

jaclync (approved):
Thanks for the updates Paolo!

> I also think that the state machine and the management of data source it's a little bit over-engineered in the reviews. I'd like in the future to refactor the view controller and the data source, and simplify the methods inside ReviewsViewController because there are some nested methods, so you risk to call a method two or more times, and you don't know it.

yes I agree! I'm also hoping to make each tab view controller's life cycle tied to a store, so that we don't have to handle when the site changes manually via `NSNotificationCenter` and all the setup can be in `viewDidLoad`: https://github.com/woocommerce/woocommerce-ios/issues/1838

Code: ✅ (just questions about some calls that might not be necessary)
Testing: ✅ (tested logging out, switching stores, pulling down to refresh with a remotely trashed review!)

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
pmusolino: I think there is a bug in updating the UI only from the `SyncingCoordinator` because some cases are not managed. For example, we don't want to reload all the things when the Reviews screen appears, so it's not useful to resynchronize the SyncingCoordinator in `viewWillAppear`, and without the `transitionToSyncingState` here when the screen is empty, you will see the loading/ghost cells with the data of the old store. It can be caused by the wrong reuse of the cells but should be fixed when we will drop the old `SyncingCoordinator`. I also noticed that we are using `transitionToResultsUpdatedState` here, but in theory, also it should be called only inside the `SyncingCoordinator` delegates methods.

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
pmusolino: It make sense, fixed here 1e26382ce06a7391d4cd4f088d4913837f9e5dd4 🙇

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
pmusolino: I confirm that it works properly also without this `reloadData()`. Removed here 3ddb199db0e0e2eaf0204e6729e710c3d1159409 🙇

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
jaclync: I see, yea we can remove the manual UI state transitions after replacing `SyncingCoordinator` 👍 this might also not be necessary once a `ReviewsViewController` is tied to one site ID

jaclync (approved):
The changes look good!

pmusolino:
Thank you so much @jaclync for your review! <3

PR #2797, opened 2020-09-10
Add Copied OrderListViewController and OrderListViewModel

shiki:
Part of the DiffableDataSource fix for #1543. This is a continuation of #2791.  

🇨🇦 Sorry for the 732 lines. The contents are mainly copy-pasta from the original `OrdersViewController` and `OrdersViewModel` code. 

## The Plan

The plan is still the same as what I mentioned in #2791. There will be **new** `OrderListViewController` and `OrderListViewModel` classes. These classes will be similar to the existing [`OrdersViewController`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift) and [`OrdersViewModel`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift) classes except that the new versions will be using `DiffableDataSource` instead of `FetchedResultsController`. These new classes will only be used if the device is using iOS 13.0. Eventually, we will remove the old classes when the minimum is set to iOS 13.0.

Here's a diagram of the plan:

image

Please let me know if you have any concerns. 

## What This PR Does

This PR copies the original `OrdersViewController` and `OrdersViewModel` to the new classes. 

Source | Target 
--------|-------
`OrdersViewController`        |       `OrderListViewController`
`OrdersViewModel` | `OrderListViewModel`

The only changes from the originals are:

- I exluded the references to the [`OrdersViewModel.resultsController`](https://github.com/woocommerce/woocommerce-ios/blob/c720e7415d83df7412237d304d6d8fc3c16c8a35/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L52-L76)
- I excluded the [`UITableViewDataSource` conformance](https://github.com/woocommerce/woocommerce-ios/blob/c720e7415d83df7412237d304d6d8fc3c16c8a35/WooCommerce/Classes/ViewRelated/Orders/OrdersViewController.swift#L455-L493). This will be replaced by `UITableViewDiffableDataSource`. 
- I [marked the classes](https://github.com/woocommerce/woocommerce-ios/blob/91b45a8477fbfab272a4788f1f57204beac4044c/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L27) as `iOS 13.0` only.
[...]

PR #2798, opened 2020-09-10
Use DiffableDataSource on OrderListViewModel and OrderListViewController

shiki:
Part of the DiffableDataSource fix for #1543. This is a continuation of #2797.

⚠️ Depends on #2797. Please review that first. 🙂    

## What

Continuing the plan mentioned in #2797, this implements DiffableDataSource and `FetchResultSnapshotsProvider` on the new (#2797) `OrderListViewController` and `OrderListViewModel` classes. 

image

### OrderListViewModel Key Changes

`OrderListViewModel` now declares a `snapshotsProvider` which uses the same conditions as the old [`OrdersViewModel.resultsController`](https://github.com/woocommerce/woocommerce-ios/blob/c720e7415d83df7412237d304d6d8fc3c16c8a35/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L52-L76). 

https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift#L53-L75

It's not accessible by the view controller. The view controller will only care about subscribing to new snapshots which are emitted by:

https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift#L77-L80

### OrderListViewController Key Changes

`OrderListViewController` now declares a `DiffableDataSource` which is bound to `tableView.dataSource`. 

https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L38-L46

In order to _feed_ new data to this `DiffableDataSource`, we only need to subscribe to `OrderListViewModel.snapshot`:

https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L201-L204

This subscription is how the `UITableView` will update its sections and rows.

### The DiffableDataSource Is The Source of Truth

A key difference between `DiffableDataSource` and the `FetchedResultsController` we used is that the `DiffableDataSource` is the source of truth. Once it has a snapshot, `DiffableDataSource` **does not need** the `FetchedResultsController` to determine how it will present its data. 

Question | Old FRC Implementation | New DiffableDataSource |
--------|-------|---
How many sections?        |    FRC ([sections.count](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L209-L213))   | DataSource (automatically handled)
How many rows in a section? | FRC ([sectionInfo.numberOfObjects](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L215-L219)) | DataSource (automatically handled)
What object is at this `indexPath`? | FRC ([objectIndex](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L203-L207)) | [DataSource.itemIdenfier](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L515-L518) but with support from [FetchResultSnapshotsProvider.object(withID:)](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift#L184)

As you can see, in the old FRC implementation, there are two sources of truth: `UITableView` and `FRC`. The `UITableView` keeps its own list of objects and the FRC also keeps its own list of objects. In the new implementation, there is only one: `DiffableDataSource`. In fact, if the `FetchResultSnapshotsProvider` stops working and stops emitting snapshots, the `DiffableDataSource` can still work. It will be presenting stale data but it will still work. 😄 

### Avoid Using Index Paths

We should avoid using index paths. As much as possible, we should use the `ObjectID`. If you look at the implementation in this PR, we interact with the `DiffableDataSource` first before trying to access more data. For example, in the [`didSelectRowAt` delegate method](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L508), we are only given an `indexPath`. In order to retrieve the actual object (`OrderDetailsViewModel`), we retrieve the `ObjectID` from the `DiffableDataSource` first. And then we _fetch_ the object using the `ObjectID`. 

https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L515-L518

Fetching the `ObjectID` [uses the `StorageType`](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L140-L149) directly and doesn't care whether the FRC (SnapshotsProvider) is in an incorrect state or not. Thereby avoiding the #1543 crash. 

What happens if we tried to use index paths for retrieving the object instead? That would mean that we're now back to the _old way_ of [accessing the object from the `FRC` using index paths](https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift#L190). And that means we're back to having **two sources of truth**. So if it happens that the FRC and the UITableView are not in sync, we get #1543 again 💥.

The same is true with sections, we should use section identifiers instead of the section index as much as possible. 

https://github.com/woocommerce/woocommerce-ios/blob/a13bf7699bcc76576100ebbfc67586603705852d/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L549-L553

In summary, **never trust** the data in `FetchedResultsController`. Only trust its snapshots and the `DiffableDataSource`. 😄 

## Intentional Bug

In the previous PRs (like #2777), I mentioned that I exposed `obtainPermanentIDs` to fix a bug that is caused by using the snapshots of `NSFetchedResultsController`. I intentionally did not use that in this PR yet in order for you, dear reviewer, to be aware of the bug. 😄 I will fix it in a separate PR. 

For now, expect to see empty cells like this:

## Testing

    ```swift
        if #available(iOS 13.0, *) {
            return makeViewControllers()
        } else {
        }
    ```
3. Navigate to Orders. 
4. Confirm that orders are loaded and displayed.
5. Pull to refresh and scroll down to load subsequent pages. Confirm that new pages are loaded. Expect to see empty cells at this point. Also, **notice** that when loading subsequent pages, the animation is a little funky and you seem to be thrown to a different location in the table view. That's another bug that will be fixed with `obtainPermanentIDs`. 😁 
6. Tap on an order. Confirm that the correct order details page is shown. 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
@@ -1,0 +1,8 @@
+import UIKit
+
+@available(iOS 13.0, *)
+extension UITableViewDiffableDataSource {
+    var numberOfItems: Int {
+        snapshot().numberOfItems
+    }
+}
--- a/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
@@ -13,4 +13,8 @@ extension UITableViewDiffableDataSource {
     var numberOfItems: Int {
         snapshot().numberOfItems
     }
+
+    var isEmpty: Bool {
+        numberOfItems == 0
+    }
 }
--- a/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
@@ -2,6 +2,14 @@ import UIKit
 
 @available(iOS 13.0, *)
 extension UITableViewDiffableDataSource {
+    func indexOfItem(for indexPath: IndexPath) -> Int? {
+        guard let identifier = itemIdentifier(for: indexPath) else {
+            return nil
+        }
+
+        return snapshot().indexOfItem(identifier)
+    }
+
     var numberOfItems: Int {
         snapshot().numberOfItems
     }
--- a/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
@@ -17,4 +17,8 @@ extension UITableViewDiffableDataSource {
     var isEmpty: Bool {
         numberOfItems == 0
     }
+
+    func sectionIdentifier(for sectionIndex: Int) -> SectionIdentifierType? {
+        snapshot().sectionIdentifiers[safe: sectionIndex]
+    }
 }
--- a/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
@@ -1,7 +1,11 @@
 import UIKit
 
+/// Additional utilities for `UITableViewDiffableDataSource`.
+///
 @available(iOS 13.0, *)
 extension UITableViewDiffableDataSource {
+    /// Returns the index of the item pointed to by `indexPath` relative to the number of
+    /// items in `snapshot()`.
     func indexOfItem(for indexPath: IndexPath) -> Int? {
         guard let identifier = itemIdentifier(for: indexPath) else {
             return nil
@@ -10,14 +14,20 @@ extension UITableViewDiffableDataSource {
         return snapshot().indexOfItem(identifier)
     }
 
+    /// Returns the total number of items in the current `snapshot()`.
     var numberOfItems: Int {
         snapshot().numberOfItems
     }
 
+    /// Returns `true` if there are no items in the current `snapshot()`.
     var isEmpty: Bool {
         numberOfItems == 0
     }
 
+    /// Returns the section identifier for the given index.
+    ///
+    /// Note that the identifier is _not_ the title of the section. It should probably be converted
+    /// to something else before it is presented to the user.
     func sectionIdentifier(for sectionIndex: Int) -> SectionIdentifierType? {
         snapshot().sectionIdentifiers[safe: sectionIndex]
     }
--- a/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
@@ -575,8 +575,7 @@ private extension OrderListViewController {
     /// we've got cached results, or not.
     ///
     func transitionToSyncingState() {
-        // WIP Replace with DiffableDataSource logic later
-        // state = viewModel.isEmpty ? .placeholder : .syncing
+        state = dataSource.isEmpty ? .placeholder : .syncing
     }
 
     /// Should be called whenever the results are updated: after Sync'ing (or after applying a filter).
--- a/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
@@ -105,6 +105,8 @@ final class OrderListViewController: UIViewController {
         }
     }
 
+    private var cancellables = Set<AnyCancellable>()
+
     // MARK: - View Lifecycle
 
     /// Designated initializer.
@@ -120,6 +122,12 @@ final class OrderListViewController: UIViewController {
 
     required init?(coder aDecoder: NSCoder) {
         fatalError("Not supported")
+    }
+
+    deinit {
+        cancellables.forEach {
+            $0.cancel()
+        }
     }
 
     override func viewDidLoad() {
@@ -187,6 +195,10 @@ private extension OrderListViewController {
         }
 
         viewModel.activate()
+
+        viewModel.snapshot.sink { snapshot in
+            self.dataSource.apply(snapshot)
+        }.store(in: &cancellables)
 
         // Reload table because the activate call above executes a performFetch()
         tableView.reloadData()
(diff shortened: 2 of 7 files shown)


WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
```diff
@@ -0,0 +1,34 @@
...
+@available(iOS 13.0, *)
+extension UITableViewDiffableDataSource {
+    /// Returns the index of the item pointed to by `indexPath` relative to the number of
+    /// items in `snapshot()`.
+    func indexOfItem(for indexPath: IndexPath) -> Int? {
+        guard let identifier = itemIdentifier(for: indexPath) else {
+            return nil
+        }
+
+        return snapshot().indexOfItem(identifier)
```
Ecarrion, line 14: [Q] Would this be expensive to access?

Ecarrion (approved):
I don't have many comments on the code as I don't have much(none) experience with `DiffableDataSource` 😇 

I really like how everything is being driven by the datasource but I do have some concerns about how this will scale. I'm gonna write some thoughts and let me know what you think!

- Since everything will be driven by core data, does that mean that we can't have any other visual components on the table that is not stored in our storage layer? Wouldn't that limit the design team in any way?

- Would we have any issue for cases where we need to hide an element from the snapshot based on conditions that can't be represented by the predicates?

WooCommerce/Classes/Extensions/UITableViewDiffableDataSource+Helpers.swift
shiki: I was concerned about this too. I tested it with an order list that has more than 1K orders. I used an iPhone 8, the slowest device that I have. It doesn't seem to cause any issues in scrolling. 

The [`snapshot()`](https://developer.apple.com/documentation/uikit/uitableviewdiffabledatasource/3255156-snapshot) returns an `NSDiffableDataSourceSnapshot` which hints that it is a `struct`. But then there is also an [Objective-C equivalent of `NSDiffableDataSourceSnapshot`](https://developer.apple.com/documentation/uikit/nsdiffabledatasourcesnapshot?language=objc) which is a `class`. I wonder if the `struct` is just bridged to the `class` and we're simply using a reference here. 🤔 That could mean that this is fast. 

In any case, if this becomes a problem, an option for us is to preserve a reference to the snapshot during `apply`:

https://github.com/woocommerce/woocommerce-ios/blob/5631e4e1f86a10ead37430d6d8c60f9885aea1be/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L201-L204

We'd have to do wrap that with something to reduce duplication.

shiki:
Thanks for the questions, @Ecarrion! 

> Since everything will be driven by core data, does that mean that we can't have any other visual components on the table that is not stored in our storage layer? Wouldn't that limit the design team in any way?

For the case of `FetchResultSnapshotsProvider`, yes. We can only use it for interacting with `NSManagedObjects`. And yes, it limits the design. And that was also my goal here. 🙂 I simply aimed to replace [`ResultsController`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Tools/ResultsController.swift), the source of the #1543 crash. 

That said, there are ways to modify the snapshot. Or to return a different snapshot. For example, if we didn't care about the `ObjectID` and we just wanted `Date` then we could transform the returned snapshot into something else:

```swift
    var customSnapshot: AnyPublisher<NSDiffableDataSourceSnapshot<String, Date>, Never> {
        snapshotsProvider.snapshot.map { snapshot -> NSDiffableDataSourceSnapshot<String, Date> in
            // Create a different snapshot with different signature
            var customSnapshot = NSDiffableDataSourceSnapshot<String, Date>()

            // loop over the snapshot from FRC and convert items to dates (somehow :D)
            snapshot.itemIdentifiers.forEach { identifier in
                customSnapshot.appendItems([Date()])
            }

            // Exclude some items from the result
            customSnapshot.deleteItems([Date()])

            return customSnapshot
        }.eraseToAnyPublisher()
    }
```

The `UITableViewDiffableDataSource` will now have to accept `Date` too:

```swift
    private lazy var dataSource: UITableViewDiffableDataSource<String, Date> = {
```

Linking the `UITableViewDiffableDataSource` is still the same:

```swift
        /// Update the `dataSource` whenever there is a new snapshot.
        viewModel.customSnapshot.sink { customSnapshot in
            self.dataSource.apply(customSnapshot)
        }.store(in: &cancellables)
```

We can also replace `Date` with `OrderDetailsViewModel` or a different `struct`. 

For tables that do not rely on Core Data, we have the option to build our own snapshots from scratch. There's an example of that [here](https://wwdcbysundell.com/2019/diffable-data-sources-first-look/). 

> Would we have any issue for cases where we need to hide an element from the snapshot based on conditions that can't be represented by the predicates?

We can do this by using the same strategy I showed above. 

1. Use [`Publisher.map`](https://developer.apple.com/documentation/combine/publisher/map(_:)-99evh) to change the `snapshot` **before** it is emitted to the `UITableViewDiffableDataSource`
2. Use one of the [`delete*`](https://developer.apple.com/documentation/uikit/nsdiffabledatasourcesnapshot/3375765-deleteitems) methods for removing an item from a snapshot.
3. Return the modified snapshot in the `map`. 

Let me know if you have more questions!

shiki:
Thanks for reviewing, Ernesto!

PR #2802, opened 2020-09-11
Add Product: present product type bottom sheet then navigate to product form

jaclync:
Create a coordinator for Add Product navigation #2740 

## Changes

- Created `ProductFactory` that generates a new `Product` given a site ID and product type
- Created `AddProductCoordinator` that implements `Coordinator` (I decided to go with `Coordinator` instead of `CoordinatorController: UINavigationController` because this navigation flow does not use a new navigation controller)
  - First, it presents a product type bottom sheet (note that there's a current issue with `UIBarButtonItem` anchor on iPad that I'll address in a separate subtask that involves updating a shared framework `WordPressUI`)
  - Once a product type is selected, it creates an empty product via `ProductFactory` and navigates to a product form `ProductFormViewController`
- In `ProductsViewController`, implemented `addProduct` that starts `AddProductCoordinator` navigation
- Added one test case for `AddProductCoordinator`, but very limited to the presented bottom sheet since the content of the bottom sheet is private. Please lemme know if you can think of other ways to test the navigation further!

## Testing

Note that there are still UI/UX differences that are to be implemented in a separate subtask in #2740 (ref p91TBi-3f7-p2), so that a product can only be created after making a change from the default empty product. Please also test on an iPhone simulator/device since there's another subtask to fix the bottom sheet anchoring on a `UIBarButtonItem`.

I noticed a slight delay after selecting a product type from the bottom sheet and before the product form navigation on iPhone 11 iOS 13.7 simulator. I tested on an iOS 13.7 device and iOS 12.1 simulator and didn't see the delay. Time profiler also didn't show anything suspicious during the delay, so it might be an iOS 13 simulator issue.

- Go to the Products tab
- Tap on the "+" CTA --> a bottom sheet of Simple, Grouped, and External product types should be shown (no Variable)
- Select a product type --> the bottom sheet should be dismissed and then it navigates to the product details
- Make some changes to the new product (e.g. adding an image, updating title, adding price/shipping/inventory info)
- Tap "Update" (this will be updated to "Publish" / "Save as draft" in the future) --> the product should be added remotely

## Example screenshots

Please note that there are still UI/UX differences that will be implemented separately (ref p91TBi-3f7-p2) - navigation CTA, ellipsis menu, product type row etc.

simulator

### Product type bottom sheet

light | dark
-- | --
Simulator Screen Shot - iPhone 11 - 2020-09-11 at 17 58 35 | Simulator Screen Shot - iPhone 11 - 2020-09-11 at 17 59 03

Update release notes:


Garance91540 (approved):
Real cool!!! Such a big milestone!

WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
```diff
@@ -0,0 +1,67 @@
...
+        let model = EditableProductModel(product: product)
+
+        let currencyCode = ServiceLocator.currencySettings.currencyCode
+        let currency = ServiceLocator.currencySettings.symbol(from: currencyCode)
+        let productImageActionHandler = ProductImageActionHandler(siteID: product.siteID,
+                                                                  product: model)
+        let viewModel = ProductFormViewModel(product: model,
+                                             formType: .add,
+                                             productImageActionHandler: productImageActionHandler,
+                                             isEditProductsRelease3Enabled: true)
```
Ecarrion, line 61: Is this true, because the usage of this coordinator is behind the feature flag?

WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
```diff
@@ -0,0 +1,67 @@
...
+
+// MARK: Navigation
+private extension AddProductCoordinator {
+    func presentProductTypeBottomSheet() {
+        let title = NSLocalizedString("Select a product type",
+                                      comment: "Message title of bottom sheet for selecting a product type to create a product")
+        let viewProperties = BottomSheetListSelectorViewProperties(title: title)
+        let command = ProductTypeBottomSheetListSelectorCommand(selected: nil) { selectedProductType in
+            self.navigationController.dismiss(animated: true) {
+                // Strong reference to `self` is required since `AddProductCoordinator` is not strongly referenced by any class.
```
Ecarrion: Maybe we should add a comment on the top of this class to warn the consumer that this class should not be retained, WDYT?

WooCommerce/Classes/ViewRelated/Products/Add Product/ProductFactory.swift
```diff
@@ -0,0 +1,85 @@
...
+    ///
+    /// - Parameters:
+    ///   - type: The type of the product.
+    ///   - siteID: The site ID where the product is added to.
+    func createNewProduct(type: ProductType, siteID: Int64) -> Product {
+        switch type {
+        case .simple, .grouped, .variable, .affiliate:
+            return createEmptyProduct(type: type, siteID: siteID)
+        default:
+            fatalError("Unable to create product of type: \(type)")
```
Ecarrion: I was checking that the `ProductType` enum has a `.custom` case for extensions. With that in mind, wouldn't it be better for this function to throw an error or return nil instead of crashing the app?

WooCommerce/Classes/ViewRelated/Products/Add Product/ProductFactory.swift
```diff
@@ -0,0 +1,85 @@
...
+struct ProductFactory {
+    /// Creates a new product (does not exist remotely) for a site ID of a given type.
+    ///
+    /// - Parameters:
+    ///   - type: The type of the product.
+    ///   - siteID: The site ID where the product is added to.
+    func createNewProduct(type: ProductType, siteID: Int64) -> Product {
+        switch type {
+        case .simple, .grouped, .variable, .affiliate:
+            return createEmptyProduct(type: type, siteID: siteID)
```
Ecarrion, line 14: Do you think it's worth it to create tests for this?

Ecarrion (commented):
I was testing this feature and noticed that you can turn a product into a `variable product` on the product screen, which defeats the purpose of not allowing the user to create variable products until adding a variation is supported. Should we hide that option? It would be worth to assess if there is any other unwanted option.

product with variations

WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
```diff
@@ -0,0 +1,67 @@
...
+                self.presentProductForm(productType: selectedProductType)
+            }
```
Ecarrion: Making the `presentProductForm` form outside the `dismiss` completion block seems to fix the delay problem in my simulator!
```suggestion
            self.navigationController.dismiss(animated: true)
            self.presentProductForm(productType: selectedProductType)
```

fast-transitions

WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
jaclync: yes! adding products is M4 and we always make the previous milestones publicly available 🙂

WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
jaclync: good idea! added in https://github.com/woocommerce/woocommerce-ios/pull/2802/commits/5dd038da898e1407f480e277ede5d555eababe9e

WooCommerce/Classes/ViewRelated/Products/Add Product/ProductFactory.swift
jaclync: `.custom` is a possible case for editing and not creating, but maybe some day this won't be true. Updated to return nil in: https://github.com/woocommerce/woocommerce-ios/pull/2802/commits/5dd038da898e1407f480e277ede5d555eababe9e

WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
jaclync: Nice suggestion! it isn't necessary to wait for the dismiss completion to navigate to the product form indeed. updated in: https://github.com/woocommerce/woocommerce-ios/pull/2802/commits/5dd038da898e1407f480e277ede5d555eababe9e

WooCommerce/Classes/ViewRelated/Products/Add Product/ProductFactory.swift
jaclync: yes! just asserting on very few properties now, it's still nice having unit tests ready: https://github.com/woocommerce/woocommerce-ios/pull/2802/commits/6f29da1aa44a495a880e73a9997bdbcebe905f21

jaclync:
Thanks for the feedback/suggestions @Ecarrion! I updated the PR and responded to your comments for another look 😄 

> I was testing this feature and noticed that you can turn a product into a variable product on the product screen, which defeats the purpose of not allowing the user to create variable products until adding a variation is supported. Should we hide that option? It would be worth to assess if there is any other unwanted option.

Yes this is to be updated! The product type row will be updated to be readonly in a separate subtask (Update UI differences from editing a product (ref p91TBi-3f7-p2)) for M4, since we don't support adding a variation yet.

Ecarrion (approved):
This looks great 🚀 
Thanks @jaclync

PR #2811, opened 2020-09-11
[DiffableDataSource] Fix Empty Orders Cell And Inconsistent Animation

### LINKED ISSUE #2806: [DiffableDataSource] Fix Empty Order Cells and Funky Animation
Part of the fix for #1543.

The DiffableDataSource was initially implemented in #2798 but it has a bug.

In the previous PRs (like #2777), I mentioned that I exposed `obtainPermanentIDs` to fix a bug that is caused by using the snapshots of `NSFetchedResultsController`. That method was not used yet. 

Right now, we expect to see empty cells like this:

shiki:
Closes #2806. This continues the work and fixes the bug in #2798 where cells can be empty and the scrolling animation seems inconsistent. 

Before | After 
--------|-------
        |       Simulator Screen Shot - iPhone 11 - 2020-09-11 at 13 22 31

## Findings

This main reason for this is `NSFetchedResultsController` can emit snapshots that contain **temporary `ObjectIDs`** here: 

https://github.com/woocommerce/woocommerce-ios/blob/0680c7738bf9adfe21076ae7041a740c5e4bb5ff/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L188-L194

Because of this, these undesirable effects can happen:

### Empty Cells

The table can display empty cells because even though the snapshot has temporary IDs, those temporary IDs were immediately converted to permanent IDs. And hence, [`self.object(withID:)`](https://github.com/woocommerce/woocommerce-ios/blob/0680c7738bf9adfe21076ae7041a740c5e4bb5ff/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L160) will return `nil`.

### Inconsistent Animation

The table will perform a lot of animations in sequence when showing new records. As I observed, the `FRC` can emit two snapshots in sequence in a short period of time.

Snapshot | Contents
--------|-------
1st        |       `[temporaryID, temporaryID]`
2nd | `[permanentID, permanentID]`

The first snapshot contains the temporary IDs and the second one contains the permanent IDs. These IDs point to the same orders. However, for `UITableViewDiffableDataSource`, the objects are **not the same** because it only relies on the values of `FetchResultSnapshotObjectID` to determine whether two objects are the same or not.

https://github.com/woocommerce/woocommerce-ios/blob/0680c7738bf9adfe21076ae7041a740c5e4bb5ff/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L18

This would cause the `UITableViewDiffableDataSource` (and `UITableView`) to animate the same rows in and out. Here is a sample of how this undesirable animation looks like https://tinyurl.com/y62lwzg9. 

### Alternative Considered

In https://github.com/woocommerce/woocommerce-ios/commit/7e06f6d9854a756a39c03b03175076ed26676303#r41918839, I discussed alternatives with @jaclync. 

We have the option to use [`NSManagedObjectContext.object(withID:)`](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506197-object) in `FetchResultSnapshotsProvider.object(withID:)` so that temporary IDs are considered and it will not return `nil`. While this fixes the empty cells issue, the `FRC` can still emit two snapshots in sequence in a short period of time. And we would still end up with the inconsistent animation. 

Also, `NSManagedObjectContext.object(withID:)` can throw an **uncatcheable** exception, which is concerning. 

### Implemented Solution

To avoid the empty cells and the inconsistent animation, I opted to use [`obtainPermanentIDs`](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506793-obtainpermanentids) for the inserted orders. 

This adds some overhead when inserting orders but it doesn't seem to be noticeable. 

### Possible Improvement In The Future

A possible long-term solution to avoid the overhead is to have a read-only `NSManagedObjectContext` that is only used for presenting data in the UI. We could solve this together with our long-standing concurrency issues in the way we use derived `StorageType` (p91TBi-2ND-p2). 

Note: I haven't actually tested if this will work. It's just what's been in my mind ever since I learned about p91TBi-2ND-p2. 😅 

## Changes 

I mainly moved the [`OrderStore.upsertStoredOrders`](https://github.com/woocommerce/woocommerce-ios/blob/7931091a280a01024d47d82c48058654f0e65bc4/Yosemite/Yosemite/Stores/OrderStore.swift#L386-L402) to a new `UseCase`, [`OrdersUpsertUseCase`](https://github.com/woocommerce/woocommerce-ios/blob/0680c7738bf9adfe21076ae7041a740c5e4bb5ff/Yosemite/Yosemite/Stores/Order/OrdersUpsertUseCase.swift#L25). So most of the changes here are just moved code. 

And in the `UseCase`, I added the `obtainPermanentIDs`. 

https://github.com/woocommerce/woocommerce-ios/blob/0680c7738bf9adfe21076ae7041a740c5e4bb5ff/Yosemite/Yosemite/Stores/Order/OrdersUpsertUseCase.swift#L30-L32

I opted to add a `UseCase` so that I could easily test the permanent IDs requirement. It turns out it also removed a lot of code from `OrdersStore`. 🤸 

## Testing

    ```swift
        if #available(iOS 13.0, *) {
            return makeViewControllers()
        } else {
        }
    ```
3. Navigate to Orders. 
4. Confirm that orders are loaded and displayed.
5. Pull to refresh and scroll down to load subsequent pages. Confirm that new pages are loaded and there are no empty cells. 6. Tap on an order. Confirm that the correct order details page is shown. 

## Submitter Checklist


Ecarrion (approved):
I confirm that the animation issue or the empty cells are not present anymore.
There is still a somehow funky animation when doing a pull to refresh but I assume that is because we delete and re-fetch the same orders, so `UITableViewDiffableDataSource` has no way to know that the objects are the same in practice.

----
PS: I still can't believe that this API has this problem, I wonder what other people are doing 🤔

shiki:
Thank you, Ernesto!

> There is still a somehow funky animation when doing a pull to refresh but I assume that is because we delete and re-fetch the same orders, so UITableViewDiffableDataSource has no way to know that the objects are the same in practice.

Yeah, that one is expected. But thanks for reporting it.

PR #2817, opened 2020-09-14
DiffableDataSource: Add Support for Auto-Updating Cells

### LINKED ISSUE #2807: [DiffableDataSource] Fix Cells Not Refreshing on Update
Part of the fix for #1543.

There is an open Radar for UITableViewDiffableDataSource where cells, whose underlying data were changed, are not automatically refreshed.

https://openradar.appspot.com/FB7588509

shiki:
Closes #2807. This is part of the fix for #1543. This adds an additional feature that's supported by [`ResultsController`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Tools/ResultsController.swift) but not by [`FetchResultSnapshotsProvider`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift). 

## Problem 

This scenario is possible with `ResultsController`: 

1. The user opens the Orders tab. All the listed orders are up to date.
2. The user leaves the app.
3. An order is changed on the web.
4. The user opens the app (still in the Orders tab). This triggers a synchronization in the background.
5. After the sync, the updated order is received and the cell is refreshed.

However, when using `FetchResultSnapshotsProvider`, the cell is not refreshed. 

### But Why?

To recap, we rely on this delegate method in order to emit snapshots: 

https://github.com/woocommerce/woocommerce-ios/blob/2c8d67d899fa7236e26ed41bb6bf081a109fa833/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L170-L174

Consider this object update code:

```swift
let object = derivedStorage.loadObject(...)
object.dateCreated = Date()

derivedStorage.saveIfNeeded()
```

When `saveIfNeeded()` is called, the FRC will _correctly_ call the `controller(:didChangeContentWith snapshot)` delegate with a new snapshot. The problem is that this new snapshot is pretty much the same as the old snapshot. Meaning, the `itemIdentifiers` (`NSManagedObjectIDs`) are the same since only “data updates” happened. And when this new snapshot is applied to `UITableViewDiffableDataSource`, the `UITableViewDiffableDataSource` probably thinks that nothing has changed because the `itemIdentifiers` are **the same**. 

https://github.com/woocommerce/woocommerce-ios/blob/2c8d67d899fa7236e26ed41bb6bf081a109fa833/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L201-L204

I'm not sure if this is a bug in `NSFetchedResultsController` or is just how things are supposed to be. ¯\\\_(ツ)\_/¯

There's a related discussion about this [here](https://developer.apple.com/forums/thread/120320). 

## Solution

To circumvent this, I added a `NSManagedObjectContextObjectsDidChange` observer in `FetchResultSnapshotsProvider`. 

https://github.com/woocommerce/woocommerce-ios/blob/a6aa69faa5a2981fb0294d52699931cf11c6b5b7/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L222-L232

From this `Notification`, we will grab the _updated_ objects. A new snapshot will be created and `reloadItems` will be called on the updated object IDs:

https://github.com/woocommerce/woocommerce-ios/blob/a6aa69faa5a2981fb0294d52699931cf11c6b5b7/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L305-L307

This [`reloadItems`](https://developer.apple.com/documentation/uikit/nsdiffabledatasourcesnapshot/3375783-reloaditems) method somehow informs `UITableViewDiffableDataSource` to reload the appropriate cells. And this fixes the problem.

### Alternative Considered

As discussed with @jaclync in the past, we considered using the [`didChangeContent:` delegate methods of `NSFetchedResultsControllerDelegate`](https://developer.apple.com/documentation/coredata/nsfetchedresultscontrollerdelegate). However, these methods are no longer called because declaring `controller(:didChangeConetntWith snapshot:)` [silences the other methods](https://developer.apple.com/documentation/coredata/nsfetchedresultscontrollerdelegate/3235742-controller).

> If this method is implemented, no other delegate methods are invoked. 

## Testing

    ```swift
        if #available(iOS 13.0, *) {
            return makeViewControllers()
        } else {
        }
    ```
3. Navigate to Orders. 
4. When the orders are loaded, make note of one of the orders in the visible list. 
5. Place the app in the background. 
6. In wp-admin, change the status of the order that you noted. 
7. Open the app again. This should trigger and auto-sync. 
8. Wait for a bit. Confirm that the order's status is updated.

## Submitter Checklist

### DIFF
--- a/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
+++ b/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
@@ -137,7 +137,7 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
     }
 
     deinit {
-        stopObservingObjectsDidChangeNotification()
+        stopObservingObjectsDidChangeNotifications()
     }
 
     /// Start fetching and emitting snapshots.
@@ -216,7 +216,7 @@ private extension FetchResultSnapshotsProvider {
 
     func startObservingObjectsDidChangeNotifications() {
         // Remove token in case this method was called already.
-        stopObservingObjectsDidChangeNotification()
+        stopObservingObjectsDidChangeNotifications()
 
         objectsDidChangeObservationToken =
             notificationCenter.addObserver(forName: .NSManagedObjectContextObjectsDidChange, object: storage, queue: nil) { [weak self] notification in
@@ -226,7 +226,7 @@ private extension FetchResultSnapshotsProvider {
         }
     }
 
-    func stopObservingObjectsDidChangeNotification() {
+    func stopObservingObjectsDidChangeNotifications() {
         if let token = objectsDidChangeObservationToken {
             notificationCenter.removeObserver(token)
 
--- a/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
+++ b/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
@@ -253,7 +253,7 @@ private extension FetchResultSnapshotsProvider {
     /// object.dateCreated = Date()
     ///
     /// // A new snapshot will be emitted here
-    /// derviedStorage.saveIfNeeded()
+    /// derivedStorage.saveIfNeeded()
     /// ```
     ///
     /// This new snapshot will have the same `itemIdentifiers` (`NSManagedObjectID`) as expected.
--- a/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
+++ b/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
@@ -129,9 +129,15 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
     /// The delgate for `fetchedResultsController`.
     private lazy var hiddenFetchedResultsControllerDelegate = HiddenFetchedResultsControllerDelegate(self)
 
+    private var objectsDidChangeObservationToken: Any?
+
     public init(storageManager: StorageManagerType, query: Query) {
         self.storage = storageManager.viewStorage
         self.query = query
+    }
+
+    deinit {
+        stopObservingObjectsDidChangeNotification()
     }
 
     /// Start fetching and emitting snapshots.
@@ -169,6 +175,8 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
     }
 }
 
+// MARK: - NSFetchedResultsControllerDelegate
+
 @available(iOS 13.0, *)
 private extension FetchResultSnapshotsProvider {
     /// Part of `FetchResultSnapshotsProvider` which submits new `FetchResultSnapshot` objects to
@@ -192,5 +200,39 @@ private extension FetchResultSnapshotsProvider {
             let snapshot = snapshot as FetchResultSnapshot
             snapshotsProvider?.snapshotSubject.send(snapshot)
         }
+    }
+}
+
+// MARK: - ObjectsDidChange Notification Handling
+
+@available(iOS 13.0, *)
+private extension FetchResultSnapshotsProvider {
+
+    var notificationCenter: NotificationCenter {
+        NotificationCenter.default
+    }
+
+    func startObservingObjectsDidChangeNotifications() {
+        // Remove token in case this method was called already.
+        stopObservingObjectsDidChangeNotification()
+
+        objectsDidChangeObservationToken =
+            notificationCenter.addObserver(forName: .NSManagedObjectContextObjectsDidChange, object: storage, queue: nil) { [weak self] notification in
+                if let self = self {
+                    self.maybeEmitSnapshotFromObjectsDidChangeNotification(notification)
+                }
+        }
+    }
+
+    func stopObservingObjectsDidChangeNotification() {
+        if let token = objectsDidChangeObservationToken {
+            notificationCenter.removeObserver(token)
+
+            objectsDidChangeObservationToken = nil
+        }
+    }
+
+    func maybeEmitSnapshotFromObjectsDidChangeNotification(_ notification: Notification) {
+
     }
 }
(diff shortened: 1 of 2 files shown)


Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -194,3 +204,123 @@ private extension FetchResultSnapshotsProvider {
...
+
+// MARK: - ObjectsDidChange Notification Handling
+
+@available(iOS 13.0, *)
+private extension FetchResultSnapshotsProvider {
+
+    /// The NotificationCenter to use for observing notifications.
+    var notificationCenter: NotificationCenter {
+        NotificationCenter.default
+    }
```
Ecarrion: [Q] Any reason on why not injecting this? 😁

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -194,3 +204,123 @@ private extension FetchResultSnapshotsProvider {
...
+    ///
+    /// - SeeAlso: maybeEmitSnapshotFromObjectsDidChangeNotification
+    func startObservingObjectsDidChangeNotifications() {
+        // Remove token in case this method was called already.
+        stopObservingObjectsDidChangeNotifications()
+
+        objectsDidChangeObservationToken =
+            notificationCenter.addObserver(forName: .NSManagedObjectContextObjectsDidChange, object: storage, queue: nil) { [weak self] notification in
+                if let self = self {
+                    self.maybeEmitSnapshotFromObjectsDidChangeNotification(notification)
```
Ecarrion, line 230: [Q] When we end up with multiple `FetchResultSnapshotsProvider`s do you expect this function to become something that could block the UI while all of our instances process the change information?

Ecarrion (approved):
Code looks great and it's very well documented! 👏 
Also tested and I'm able to confirm that orders are being updated after coming from the background.

----

> I'm not sure if this is a bug in NSFetchedResultsController or is just how things are supposed to be. 

This makes me really nervous 🙃. The amount of issues you are having to circumvent makes me feel that we could end with more problems than with the current implementation. I have no data to back that up tho!

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
shiki: I don't know. 😁 I probably thought I didn't need to because it was not used in the tests. Fixed in d7c214dcf188f034fe321b0376613edb70119a37.

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
shiki: That's a good point. I tried to make `maybeEmitSnapshotFromObjectsDidChangeNotification` as performant as it can be. For example, I delayed the creation of `currentObjectIDs` since I wasn't sure of its performance. 

https://github.com/woocommerce/woocommerce-ios/blob/d7c214dcf188f034fe321b0376613edb70119a37/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L297

The `maybeEmitSnapshotFromObjectsDidChangeNotification` method will get called whenever there's a `saveIfNeeded()` on a background (derived) storage. Usually, that's 25 rows (our default page size). Let's consider having 10 `SnapshotsProviders` open, which can be pretty high based on my understanding of the app. That's around `(10 * 25) + (10 * 25) + (10 * 25) + (10 * 25) = 1000` loops in the worst case. That's considering [`intersection`](https://developer.apple.com/documentation/swift/set/1779150-intersection) to be `O(n)` too (I'm not really sure what the performance of that is).  

I think 1000 loops on the main thread are negligible. 

I took a sample on an iPhone 8 Plus and here's what I got:

```
shiki: start maybeEmitSnapshotFromObjectsDidChangeNotification rows = 26
shiki: ended filter 0.0680685043334961 ms
shiki: ended objectIDsDiff 0.030994415283203125 ms
shiki: ended intersection 0.011920928955078125 ms
shiki: ended newSnapshot 0.026941299438476562 ms
shiki: ended sending new snapshot 27.009963989257812 ms
shiki: ended maybeEmitSnapshotFromObjectsDidChangeNotification 27.25505828857422 ms
```

For 26 updated rows, everything performed well except for this:

https://github.com/woocommerce/woocommerce-ios/blob/d7c214dcf188f034fe321b0376613edb70119a37/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L310

That part took 27ms and is only because that call would cause the `UITableView` to update itself. It will block the main thread no matter what we do. 

Using that sample, the loops for 1000 rows are just `(27.25505828857422 - 27.009963989257812) * 5 = 1ms`. As for the `UITableView`, the `27ms` is probably close to the upper limit cause I think the cost of update is proportional to the number of visible rows. 

---

In summary, based on my limited and not science-backed performance testing, I think we're okay here. 😅 A future enhancement would be making everything run on the background thread. I learned that `dataSource.apply(snapshot)` can be called on a background thread so we can gain something for that. But I think it's probably over-engineering at this point. For now, I think it's best if we focus on fixing the crashes first.

shiki (commented):
> This makes me really nervous 🙃. The amount of issues you are having to circumvent makes me feel that we could end with more problems than with the current implementation. I have no data to back that up tho!

It's a valid concern. It feels weird to “fix” a crash by another thing that may also cause problems. I actually don't 100% expect everything to go well. Based on what I read about Swift UI (not good), we should expect DiffableDataSource to be the same. We're going to hit bumps. My hope is it's not as worse as the current crashes. 🙈 

I'm considering this as an experiment so there will be a _switch_ of some sort to back out in case this is worse than what we currently have.

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
Ecarrion: Thanks a lot for the analysis! 🙇

shiki:
Thank you for the review, @Ecarrion!

PR #2818, opened 2020-09-15
SIWA: handle when Apple ID credential is revoked

### LINKED ISSUE #2762: SIWA: Handle Apple Credential Revoked
A user's credentials can be invalid on Apple's side, either by the user removing the connection to WordPress or by the iCloud account becoming inactive. The app should check the credential state upon launch and also listen to a notification in the app while running for the state to change.

WordPress iOS issues implementing this functionality:
https://github.com/wordpress-mobile/WordPress-iOS/pull/12549
https://github.com/wordpress-mobile/WordPress-iOS/pull/12552

jaclync:
Fixes #2762 

### SIWA token observation
- Added `var isLoggedIn: Observable<Bool>` to `StoresManager` with implementation in `DefaultStoresManager`, so that `AppDelegate` can start/stop observing when Apple ID credential is revoked
  - Update unit tests with the observable
- In `AuthenticationManager`, saved Apple ID to Keychain when the app authentication completes with SIWA
- In `AppDelegate`, if iOS 13+:
  - Start/stop observing the events when Apple ID credential is revoked on login/logout. Upon any Apple ID credential revoked event, log out the app and remove Apple ID from Keychain
  - Check if the Apple ID is still valid on launch launch: get the Apple ID from Keychain, and if it's available check with WPAuthenticator. Log out the app if the Apple ID credential has been revoked and remove Apple ID from Keychain

### Crash from resetting Cord Data context

In scenario 3 below, a crash encountered during testing before the fix in https://github.com/woocommerce/woocommerce-ios/commit/a864fe46e36dd15589229ae5b07a40925e6937dd. The cause of the crash is from logging out the app shortly after app launch, when a lot of network requests have been fired before logging out. After resetting Core Data's main context and its persistent store is reset on logout, any child context that was created before the reset doesn't have a valid persistent store anymore. When the result of a network request fired before the reset but comes back, we still delete/upsert and save to the view context's child context and the app crashes because `Object's persistent store is not reachable from this NSManagedObjectContext's coordinator`.

I've tried a few ways to fix the crash (more details in a p2 post), and I decided on this approach with all the changes within `CoreDataManager`:
- Keep a unique string internally that is set to a different UUID on every storage reset
- When creating a child context (`newDerivedStorage`), set the unique string to the child [context's `name`](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506231-name):
> The developer-provided name of the context
- When saving a child context (`saveDerivedType`), only perform saving the child context asynchronously if the child context's name matches the current unique string. Note that a child context's `name` has to be accessed in the corresponding queue (within its `perform` block)

## Testing

Notes:
- This can only be tested on a real iOS 13+ device as it seems the revoked notification does not work in simulators
- The revoked notification is only triggered when switching apps via the app switcher

### Scenario 1: SIWA is revoked when the app is active
- Sign up / log in with SIWA
- Open app switcher (i.e. double tap the home button, or swipe up on iPhone X devices)
- Go to iOS Settings and disconnect your Apple account (see steps at the bottom)
- Open app switcher (i.e. double tap the home button)
- Select the app --> the app should be logged out

### Scenario 2: SIWA is revoked when the app is backgrounded
- Sign up / log in with SIWA
- Background the app
- Go to iOS Settings and disconnect your Apple account (see steps at the bottom)
- Return to the app --> the app should be logged out with this log message in the console: `checkAppleIDCredentialState: Unauthorized Apple ID. User signed out.`

### Scenario 3: SIWA is revoked on app launch
- Terminate the app
- Go to iOS Settings and disconnect your Apple account (see steps at the bottom)
- Relaunch the app --> the app might crash (https://github.com/woocommerce/woocommerce-ios/issues/2830) but relaunching the app should bring it back to logged out state

---
Steps to disconnect in your Apple account on a device:
  - Go to iOS Settings
  - Select your iCloud account at the top
  - Go to Password & Security > Apps Using Your Apple ID
  - Select WordPress > Stop Using Apple ID (we share the same app group as WordPress)

Update release notes:


WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
```diff
@@ -50,6 +51,11 @@ class DefaultStoresManager: StoresManager {
         return state is AuthenticatedState
     }
 
+    var isLoggedIn: Observable<Bool> {
+        isLoggedInSubject
+    }
+    private let isLoggedInSubject = BehaviorSubject<Bool>(false)
+
```
shiki, line 58: This is a small but significant change. I've always thought that the StoresManager is missing the [subscribable state part of the Redux flow](https://redux.js.org/api/store#subscribelistener). This opens us up to a lot of possibilities:

- Observing if the current `siteID` changed. Effect: Delete all the controllers! 
- Observing if a specific order changes like `storesManager.orders[orderID].subscribe { ... }` 
- Create `ResultsController` objects like `storesManager.orders.resultsControllerForOrderList()`. Effect: We no longer have a reference to `Storage` in `WooCommerce`! Of course, this assumes that `DefaultStoresManager` **is part of Yosemite**, which it isn't. 🙈 Still a lot to work on here! 
- Read-only queries **without** `ResultsController` like `storesManager.orders.orderWith(id:)`. 

Exciting stuff!

Yosemite/Yosemite/Base/Dispatcher.swift
```diff
@@ -47,8 +47,6 @@ public class Dispatcher {
     /// Unregisters the specified Processor from *ALL* of the dispatcher queues.
     ///
     public func unregister(processor: ActionsProcessor) {
-        assertMainThread()
```
shiki, line 50: ⚠️ This change seems important. Would you mind explaining more? I only found the commit message in 4977438654ed4405f6b384c01263aeaca84f5b1d.

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -124,7 +129,7 @@ class AuthenticationManager: Authentication {
 //
 extension AuthenticationManager: WordPressAuthenticatorDelegate {
     func userAuthenticatedWithAppleUserID(_ appleUserID: String) {
-        // Sign in with Apple is not supported yet.
+        keychain[WooConstants.keychainAppleIDKey] = appleUserID
```
shiki: Not important: I wonder if should do something like:

```swift
// untested code :D
extension Keychain {
    var wooAppleID: String {
        get { self[WooConstants.keychainAppleIDKey] }
        set { self[WooConstants.keychainAppleIDKey] = value }
    }
}
```

Storage/Storage/CoreData/CoreDataManager.swift
```diff
@@ -14,6 +14,10 @@ public final class CoreDataManager: StorageManagerType {
 
     private let modelsInventory: ManagedObjectModelsInventory
 
+    /// Set to a unique value whenever the storage is reset.
+    /// Used to identify a child context so that we do not perform save actions on a child context whose parent has been reset.
+    private var contextName: String?
```
shiki: Not important: Should we also set an initial `UUID()` value?

Storage/Storage/CoreData/CoreDataManager.swift
```diff
@@ -132,7 +137,13 @@ public final class CoreDataManager: StorageManagerType {
...
     ///
     public func saveDerivedType(derivedStorage: StorageType, _ closure: @escaping () -> Void) {
-        derivedStorage.perform {
+        derivedStorage.perform { [weak self] in
+            guard let self = self else {
+                return
+            }
+            guard derivedStorage.name == self.contextName else {
+                return
+            }
```
shiki: I was hoping that we could check a property (like [`persistentStoreCoordinator`](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506618-persistentstorecoordinator)). I'm curious if you've tried that? There doesn't seem to be a property like `isValid`. 🤷

WooCommerce/Classes/AppDelegate.swift
```diff
@@ -401,3 +415,89 @@ extension AppDelegate {
...
+
+    func logout() {
+        ServiceLocator.stores.deauthenticate()
+    }
+}
+
+// MARK: - Apple Account Handling
+//
+@available(iOS 13.0, *)
+private extension AppDelegate {
```
shiki: ⚠️  I'm not sure but based on the code that I see here, we can encapsulate all of these into its own `class` to keep the `AppDelegate` lean-ish. We can _retain_ the object as a property of `AppDelegate` to keep it alive. 

We can use the [`didBecomeActive`](https://developer.apple.com/documentation/uikit/uiapplication/1622953-didbecomeactivenotification) `NSNotification` instead of relying on `AppDelegate`. And we could probably make that new `class` testable too. 

What do you think?

shiki (changes requested):
Tested and it's working nicely! 🤸 

I have a few things I'd like to discuss if that's okay. 🙂 

### Re: Crash from resetting Core Data context

I performed the steps in https://github.com/woocommerce/woocommerce-ios/issues/2251#issuecomment-625520675 and the crash still happens. 🙈 

The cause is not from a network request but from accessing this:

https://github.com/woocommerce/woocommerce-ios/blob/3c5d9dcc6d5b2c7e01ebd68498e158829279a91b/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsResultsControllers.swift#L64-L68

And a root cause of that is because `ResultsController` reacts on [storage reset](https://github.com/woocommerce/woocommerce-ios/blob/3c5d9dcc6d5b2c7e01ebd68498e158829279a91b/Yosemite/Yosemite/Tools/ResultsController.swift#L260-L267).

When I encountered this last time, I think a fix can either be (or all of these):

- Destroy the view controllers when logging out. View Controllers shouldn't react on session changes. They should just die. 
- Do not reset the `CoreDataManager` in the first place. I think the database coupled with the session is too much for us to handle. It would probably be simpler if we kept the database as is during session changes. Then again, this would mean keeping `siteID` on **all** entities (especially `Order`). So maybe what we have is the simpler solution. I still like having the database session-less though. 😄 
- Cancel network requests on log out. Challenging to do since this probably requires an architecture that supports it. 

❓ @jaclync Considering that the fix in here did not fully fix this crash, do you think we should remove the Core Data-related changes in this PR? Or is this a blocker that must be solved right now?

### Unlikely Scenario Bug

I found an _unlikely_ scenario that causes some weirdness. I thought you'd probably want to still know about it. 😄 

1. Login with SIWA
2. Stay on the store picker screen.
3. Unlike the Apple ID. 
4. Go back to the app. 
5. The store picker screen is still active and you can still proceed with logging in. No data loads happen. 

✅  Force-closing the app logs the user out. So all good!

jaclync:
Thanks so much @shiki for the thorough testing & review! I apologized for [the comment on #2251](https://github.com/woocommerce/woocommerce-ios/issues/2251#issuecomment-692618424) that I left at the end of my day yesterday that might sound like this PR also fixes that crash 🙇🏻‍♀️ The fix for #2251 is a separate one since the cause is similar but different: https://github.com/woocommerce/woocommerce-ios/pull/2821

Also, I wrote up some findings in p91TBi-3iz-p2 if you'd like to hear more details about these two crashes 🙂 unfortunately, after more testing today I still ran into race conditions with the fix in the PR / p2 when the reset occurs right after the child context name comparison and before the child context save. I've tried making the `contextName` property thread-safe with a property wrapper like [this gist](https://gist.github.com/jaclync/2ba5da6c5dfbf2ca873c34895ed8c55b) but still cannot completely avoid this edge case.

One thing I'd like to try before making a more significant change to **not** reset the storage layer like you suggested:

consolidate the updates to a child context (like the code in a store's [`perform` here](https://github.com/woocommerce/woocommerce-ios/blob/2fb6346361477633c210b39c265d72ed2b5a5b75/Yosemite/Yosemite/Stores/OrderStatusStore.swift#L82-L97)) and the save action so that we can early return if the context name doesn't match. It doesn't feel super promising to me though 😞 

> When I encountered this last time, I think a fix can either be (or all of these):
> - Destroy the view controllers when logging out. View Controllers shouldn't react on session changes. They should just die.

yes this would fix the crashes with FRC when the app is logged out, and potentially improve performance by not keeping many unnecessary objects around. I'm hoping to make progress on https://github.com/woocommerce/woocommerce-ios/issues/1838 this HACK week though I've spent sooo much time on this crash already!

> - Do not reset the CoreDataManager in the first place. I think the database coupled with the session is too much for us to handle. It would probably be simpler if we kept the database as is during session changes. Then again, this would mean keeping siteID on all entities (especially Order). So maybe what we have is the simpler solution. I still like having the database session-less though. 😄

Personally I'm not a fan of keeping user data after they log out, we already keep part of the site data after switching stores but I think logging out has a different intention. It's probably not a common case, but users could be using the app on a shared device and the database is readable by anyone with the device.

That said, after spending more time looking into the race condition today, I've ran out of options other than not resetting the Core Data stack. And if we ship the SIWA revoke update, the crash has to be fixed.

> - Cancel network requests on log out. Challenging to do since this probably requires an architecture that supports it.

I thought about it before, but didn't know how feasible it is. Today I looked into it more with Alamofire, and it is possible with the [SO answer](https://stackoverflow.com/questions/41478122/cancel-a-request-alamofire/43042569#43042569)! However, this only helps to a degree because some closures might already receive the network results but are still caught in the async storage update (`derivedStorage.perform`) when the reset occurs.

> @jaclync Considering that the fix in here did not fully fix this crash, do you think we should remove the Core Data-related changes in this PR? Or is this a blocker that must be solved right now?

> Unlikely Scenario Bug
> I found an unlikely scenario that causes some weirdness. I thought you'd probably want to still know about it. 😄

Wow great catch! Thank you for letting me know, I'll look into this scenario later (the day went by so quickly looking into the race condition) 😓

shiki:
> Also, I wrote up some findings in p91TBi-3iz-p2 if you'd like to hear more details about these two crashes

Responded! 

> Personally I'm not a fan of keeping user data after they log out, we already keep part of the site data after switching stores but I think logging out has a different intention

I agree. Yesterday, I didn't realize that we can have both. We can have a persistent Core Data stack **and** delete the data when logging out. 

> I've ran out of options other than not resetting the Core Data stack

In my response to the p2 post, this also doesn't look like a short term solution. 😞

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
jaclync: yea I'm excited to see more observable states in the codebase!

> Observing if the current siteID changed. Effect: Delete all the controllers!

yes! currently, the site info is in `SessionManager` and I plan to add an observable there.

> Observing if a specific order changes like storesManager.orders[orderID].subscribe { ... }
> Create ResultsController objects like storesManager.orders.resultsControllerForOrderList(). Effect: We no longer have a reference to Storage in WooCommerce! Of course, this assumes that DefaultStoresManager is part of Yosemite, which it isn't. 🙈 Still a lot to work on here!
> Read-only queries without ResultsController like storesManager.orders.orderWith(id:)

I think FRC and `EntityListener` are really the only reasons we have to expose Storage / Core Data in the app layer. It'd be really cool if we could have a way to observe the storage layer changes in Yosemite to replace FRC and `EntityListener`. `FetchResultSnapshotsProvider` feels like a great start with an observable snapshot!

One tricky part to replace FRC I can think of is to translate a query in terms of readonly models to storage layer predicate. Like from a predicate `Yosemite.Product.productType == .simple` to `Storage.Product.productTypeKey == "simple"` since some fields don't match exactly due to Core Data model constraints. We might have to create a function for each use case like `storesManager.products.productsWith(productType:)`

Yosemite/Yosemite/Base/Dispatcher.swift
jaclync: Great question! For all the scenarios in the PR testing section, I saw different stores being deinitialized in the background thread after the network results came back:

Screen Shot 2020-09-17 at 2 45 31 PM

I am actually quite confused by why this isn't an issue on `develop`, just like why I couldn't reproduce the same crash by logging out in settings right after launch 🤔 Logging out from settings is calling the same thing `ServiceLocator.stores.deauthenticate()` as the Apple ID credential revoked case, and both are called on the main thread: 

https://github.com/woocommerce/woocommerce-ios/blob/5163bf09618feca964118f6d52f9c3057f8dffbe/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift#L379-L382

I don't think this PR affects how the stores are initialized/deinitialized, lemme know if you have any insights!

WooCommerce/Classes/Authentication/AuthenticationManager.swift
jaclync: Good idea! Added an extension setter/getter here: https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/1be6823158938d21488ea14874bd37c2d2463db9

Storage/Storage/CoreData/CoreDataManager.swift
jaclync: yep that makes sense to be consistent! I updated it in https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/f171bca15eff789e4f056b0a21f32da5b5d02142 but I also reverted the commits on the Core Data changes to save the crash fix for later

Storage/Storage/CoreData/CoreDataManager.swift
jaclync: yea, I tried comparing the child context's `persistentStoreCoordinator` and its persistent stores with the main context's, but they seem to be always the same because it's a singleton (`CoreDataManager`'s `persistentContainer.persistentStoreCoordinator`) and I didn't find any Core Data API to check on if a context / persistent store has been reset 😞 

It's quite frustrating how little official guidance there is for Core Data best practices for different scenarios (app reset, parent/child context etc.).

WooCommerce/Classes/AppDelegate.swift
jaclync: I like the idea of encapsulating the Apple ID checks to its own class, rather than expanding the global `AppDelegate`! I created a class `AppleIDCredentialChecker` in https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/55d07b2ad68c438944edd9b51bb31b68cf1ad363 and then updated `WordPressAuthenticator` to a lazy var in https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/ad1786ee0ad4422bfc3887511af0715944cdff5c since the shared instance is only initialized in app delegate's `willFinishLaunchingWithOptions` after its first call in `applicationDidBecomeActive` 😅  (otherwise it crashes the app on launch)

For the check in `applicationDidBecomeActive`, I still let the app delegate call `AppleIDCredentialChecker`'s `checkAppleIDCredentialState`. But lemme know if you think observing the `NSNotification` within the class would be better!

jaclync:
@shiki thank you for looking into the crash, and the comments on the p2 post! I had an update to the crash fix with the context name that seems to work (by saving the child context in the main context's queue), but I agree it would be better leaving the crash fix to a separate PR. This PR now only has the Apple ID credential checks and is ready for another pass! I retested the scenarios and updated the PR testing steps with the expected crash in scenario 3.

I also created 2 issues for the case you found and the crash:
- Crash in scenario 3: https://github.com/woocommerce/woocommerce-ios/issues/2830
- App getting into a weird state: https://github.com/woocommerce/woocommerce-ios/issues/2829
  - I looked into it a bit, and it's because `StoresManager` does not consider the app is "authenticated" until a site is selected with its credentials. Thus, the Apple ID check early returns because it doesn't consider the app "logged in". I think this logic might help (where we show the store picker on launch):
https://github.com/woocommerce/woocommerce-ios/blob/5163bf09618feca964118f6d52f9c3057f8dffbe/WooCommerce/Classes/AppDelegate.swift#L370

WooCommerce/Classes/AppDelegate.swift
shiki: > since the shared instance is only initialized in app delegate's willFinishLaunchingWithOptions after its first call in applicationDidBecomeActive 😅

Ugh. Singletons. ʕノ•ᴥ•ʔノ ︵ ┻━┻

This is fine for now. 🙂 

> For the check in applicationDidBecomeActive, I still let the app delegate call AppleIDCredentialChecker's checkAppleIDCredentialState. But lemme know if you think observing the NSNotification within the class would be better!

I would think so. If this is handled by `AppleIDCredentialChecker` then `AppDelegate` doesn't really need to care about it aside from calling `observeLoggedInStateForAppleIDObservations`. And _using_ `AppleIDCredentialChecker` would be easier. There's only one entry point. 🙂 

But up to you if you still feel like making the change. It's not too much trouble, especially since we've moved most of the noise to `AppleIDCredentialChecker`. 🙂

Yosemite/Yosemite/Base/Dispatcher.swift
shiki: I see. Thank you for explaining more! I didn't realize that this is called in `deinit`

https://github.com/woocommerce/woocommerce-ios/blob/085a3db48a8858e1ad72343cbc0d3747996e83cd/Yosemite/Yosemite/Base/Store.swift#L40-L44

I think we can't really control when the object is deinitialized. In the example that you shared, it looks like it was deinitialized in the background thread because it just stopped being used and got automatically deallocated during `storageManager.saveDerivedType`. It's just a guess. I don't know much about how Swift's memory management works. 😅 

But yeah, thanks for sharing. I guess this is fine then! 

Also, IIRC, this assert is one of the causes of our weird unit test failures. I'm glad that it's gone!

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
shiki: > One tricky part to replace FRC I can think of is to translate a query in terms of readonly models to storage layer predicate. Like from a predicate Yosemite.Product.productType == .simple to Storage.Product.productTypeKey == "simple" since some fields don't match exactly due to Core Data model constraints. We might have to create a function for each use case like storesManager.products.productsWith(productType:)

I was hoping to have more exposed classes in `Yosemite` that hide these. Maybe `UseCase`? 

```swift
let useCase = Yosemite.ProductQueriesUseCase()
// Translates Yosemite/Networking productType to Storage productType
let query: FetchResultSnapshotsProvider.Query = useCase.queryFor(productType: productType) 
// Apply new query
snapshotsProvider.query = query
```

I see a lot to be gained in moving a lot of business logic to Yosemite and keeping WooCommerce mostly UI stuff. This is still valid based on [the description](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/YOSEMITE.md):

> This module encapsulates the application business logic, and interacts with the Networking and Storage layers.

This would probably help us when we get around to having different apps in the future.

WooCommerce/Classes/Authentication/AppleIDCredentialChecker.swift
```diff
@@ -0,0 +1,104 @@
...
+/// Checks and listens for observations when the Apple ID credential is revoked when the user previously signed in with Apple.
+///
+@available(iOS 13.0, *)
+final class AppleIDCredentialChecker {
+    /// Keychain access for SIWA auth token
+    private lazy var keychain = Keychain(service: WooConstants.keychainServiceName)
+
+    private var cancellable: ObservationToken?
+
+    private lazy var authenticator: WordPressAuthenticator = WordPressAuthenticator.shared
```
shiki: ℹ️  Maybe we should explain why we didn't inject this. 🙂

shiki (approved):
Re-reviewed and re-tested. Thank you, Jaclyn! :shipit:

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
jaclync: Creating `UseCase`s to encapsulate the query sounds cool, I'm all for it for Yosemite to have most of the business logic and App layer to just handle UI! 👍

Yosemite/Yosemite/Base/Dispatcher.swift
jaclync: yea, I'm still wondering why `deinit` hasn't happened in the background thread in production (logging out right after app launch) while this PR does 🤔

WooCommerce/Classes/Authentication/AppleIDCredentialChecker.swift
jaclync: actually, after observing `didBecomeActive` inside the checker class https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/01bce49859e57d5592a8027e82bac0d46be49b87, this lazy var change can be reverted (https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/e3c73b13968fc096cbc2eb9bed91264bad8c3efc) since it's always after the class is initialized and after `WordPressAuthenticator` singleton init 😄

WooCommerce/Classes/AppDelegate.swift
jaclync: yea it makes sense to observe `applicationDidBecomeActive` within the class itself! updated in https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/01bce49859e57d5592a8027e82bac0d46be49b87 and there's no need to make the authenticator a lazy var anymore (https://github.com/woocommerce/woocommerce-ios/pull/2818#discussion_r490672218)!

jaclync:
Thanks for the feedback @shiki, I just updated the PR and responded to your comments! I'm going to move this to 5.2 milestone and merge it after code freeze next week so that I have some time to fix the crash https://github.com/woocommerce/woocommerce-ios/issues/2830 and bug https://github.com/woocommerce/woocommerce-ios/issues/2829 🙂

jaclync:
Just [updated the release notes from 5.1 to 5.2](https://github.com/woocommerce/woocommerce-ios/pull/2818/commits/40be037cf8498645f802fdc284d9e6a715600e82), merging now!

Issue #2250, opened 2020-05-07
Orders → Search: Status totals aren't updated with remote changes until order list is refreshed


rachelmcr:
**Describe the bug**

In https://github.com/woocommerce/woocommerce-ios/pull/2208 we added the total number of orders of a given status in Orders → Search.

However, if a new order is received or an order status is changed remotely, the order list updates immediately to show the new order/status but the status totals in Orders → Search don't change until the order list is refreshed (e.g. with a pull to refresh).

cc @shiki 

**To Reproduce**

1. Open the Orders tab and check the Processing badge count on the tab.
2. Open Orders → Search and confirm the Processing status count matches the badge count.
3. Receive a new order with Processing status on your store.
4. Notice the order appears in the orders list and the Processing badge count on the Orders tab updates.
5. Open Orders → Search and notice the Processing status count hasn't changed.
6. Go back to the orders list and pull to refresh.
7. Open Orders → Search and notice the Processing status count is now correct.

**Screenshots**

Orders badge (accurate count)|Status totals (outdated count)
-|-
<img src="https://user-images.githubusercontent.com/8658164/81302954-fb006500-9072-11ea-898b-328a59eb1a40.PNG" width="300px">|<img src="https://user-images.githubusercontent.com/8658164/81302943-f6d44780-9072-11ea-9ae2-23c7612d69ba.PNG" width="300px">

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 13.4.1
- WooCommerce iOS version: 4.2.0.0 TestFlight beta

shiki:
Also experienced by Jaclyn when testing #2820 → https://github.com/woocommerce/woocommerce-ios/pull/2820#pullrequestreview-489447474

itsmeichigo:
Closing this issue because this seems like an outdated design and we're no longer showing the order status list when searching order now. Let me know if I'm missing something 😅.

PR #2820, opened 2020-09-15
Enable DiffableDataSource for Orders Tab and Orders → Search → Filter

### LINKED ISSUE #1543: Invalid Number Of Rows/Sections Crash

### LINKED ISSUE #2809: [DiffableDataSource] Enable the new Orders View Controllers for iOS 13 users
The new `OrderListViewController` and `OrderListViewModel` classes are currently unused. They need to be enabled in `OrdersMasterViewController` for iOS 13 users.

shiki:
Closes #2809. 

This enables all the work that was done for implementing DiffableDataSource for Orders to circumvent #1543. This does not fully fix #1543 because that crash happens (sometimes) on our other view controllers too. But if this is successful, then we can then convert everything else to use DiffableDataSource. 

All the PRs that led to this moment are in [this Github project](https://github.com/woocommerce/woocommerce-ios/projects/22). There are still some remaining debts in there that can be tackled later. 

## Reversal

If it turns out that this DiffableDataSource implementation is worse than the original, we can reverse everything by reverting the changes in this PR. 

## Testing

1. Use an iOS 13.0+ device.
1. Navigate to Orders → Processing. Confirm that the orders are filtered by Processing. 
2.. Navigate to Orders → All Orders. Confirm that all orders are shown. 
3. Confirm that the infinite load still works. 
4. Navigate to Orders → Search. 
5. Tap on a filter. Confirm that an order list is shown with orders having the selected filter. 

### Push Notifications 

1. Enable push notifications in your sandbox.
6. Place the app in the background.
7. Submit a new order on the web.
8. Tap on the push notification. Confirm that the order details page is opened.

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Search/Order/OrderSearchStarterViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Search/Order/OrderSearchStarterViewController.swift
@@ -90,29 +90,9 @@ extension OrderSearchStarterViewController: UITableViewDelegate {
 
         analytics.trackSelectionOf(orderStatusSlug: cellViewModel.slug)
 
-        let emptyStateMessage: NSAttributedString = {
-            guard let statusName = cellViewModel.name else {
-                return NSAttributedString(string: NSLocalizedString("We're sorry, we couldn't find any orders.",
-                                                                    comment: "Default message to show if a filtered Orders list is empty."))
-            }
+        let listViewController = makeOrderListViewController(for: cellViewModel)
 
-            let boldStatusName = NSAttributedString(string: statusName,
-                                                    attributes: [.font: EmptyStateViewController.Config.messageFont.bold])
-
-            let format = NSLocalizedString("We're sorry, we couldn't find any “%@” orders",
-                                           comment: "Message shown if a filtered Orders list is empty. The %@ is a placeholder for the order status.")
-            let message = NSMutableAttributedString(string: format)
-            message.replaceFirstOccurrence(of: "%@", with: boldStatusName)
-
-            return message
-        }()
-        let ordersViewController = OrdersViewController(
-            title: cellViewModel.name ?? Localization.defaultOrderListTitle,
-            viewModel: OrdersViewModel(statusFilter: cellViewModel.orderStatus),
-            emptyStateConfig: .simple(message: emptyStateMessage, image: .emptySearchResultsImage)
-        )
-
-        navigationController?.pushViewController(ordersViewController, animated: true)
+        navigationController?.pushViewController(listViewController, animated: true)
 
         tableView.deselectSelectedRowWithAnimation(true)
     }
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Search/Order/OrderSearchStarterViewController.swift
```diff
@@ -106,23 +128,27 @@ extension OrderSearchStarterViewController: UITableViewDelegate {
...
+        if #available(iOS 13, *) {
+            return OrderListViewController(
+                title: title,
+                viewModel: .init(statusFilter: cellViewModel.orderStatus),
+                emptyStateConfig: emptyStateConfig
+            )
+        } else {
+            return OrdersViewController(
+                title: title,
+                viewModel: OrdersViewModel(statusFilter: cellViewModel.orderStatus),
```
jaclync: nit: this could probably be `.init` like the iOS 13+ case?

```suggestion
                viewModel: .init(statusFilter: cellViewModel.orderStatus),
```

jaclync (approved):
Great work, @shiki! I'm feeling very hopeful about fixing the infamous invalid number of rows crash for Orders 😄 💯  

Code: ✅ 
Testing: ✅ 

I noticed 2 non-blocking minor issues during testing, since both happen in production as well:

- In Orders Search, the count of order status doesn't seem updated after changing the order status of an order
- If I background the app while it's on Orders Search and then tap on a push notification, it stays on Orders Search and I see the new order after dismissing the search UI

WooCommerce/Classes/ViewRelated/Search/Order/OrderSearchStarterViewController.swift
shiki: You got me! Fixed in b43c3dcbb2ba0f20c1ba263de38036bea88ece54.

shiki (commented):
Thanks, Jaclyn!

> In Orders Search, the count of order status doesn't seem updated after changing the order status of an order

Thanks. I believe this is the same as https://github.com/woocommerce/woocommerce-ios/issues/2250. 

> If I background the app while it's on Orders Search and then tap on a push notification, it stays on Orders Search and I see the new order after dismissing the search UI

Good find! Created https://github.com/woocommerce/woocommerce-ios/issues/2823. I wonder if we should present modals in this case instead. 🤔

PR #2821, opened 2020-09-16
Order Details: refetch results for all results controllers after storage is reset

jaclync:
Attempted fix for #2251

While looking into another crash of the same reason (`Object's persistent store is not reachable from this NSManagedObjectContext's coordinator`), I came to understand a bit more when this type of crash might occur after the Core Data reset (resetting the main context & persistent store):
- Loading results from a context that has been reset without another fetch
- Saving objects to a child context created before the reset

For #2251, it's the first case where the `ResultsController` (FRC underneath) is reloading UI that access the objects from another FRC whose objects haven't been refreshed. After the storage reset, each FRC has to call `performFetch` again so that it fetches the objects from the updated context. The objects from a FRC without `performFetch` after the reset have un unknown persistent store since [the previous store was destroyed](https://github.com/woocommerce/woocommerce-ios/blob/2fb6346361477633c210b39c265d72ed2b5a5b75/Storage/Storage/CoreData/CoreDataManager.swift#L153-L166).

The crash only happens for UI that relies on more than one FRC and whose UI involves another FRC, because `ResultsController` [already calls `performFetch` before the `onDidResetContent` callback](https://github.com/woocommerce/woocommerce-ios/blob/2fb6346361477633c210b39c265d72ed2b5a5b75/Yosemite/Yosemite/Tools/ResultsController.swift#L256-L267) after storage reset. As the [stack trace](https://sentry.io/organizations/a8c/issues/1703307167/?project=1458804&query=is%3Aunresolved&sort=user&statsPeriod=14d) indicates, it's the **refund FRC** reloading UI that reads `[Product]` from **product FRC** that hasn't been refreshed. From my testing, the refund FRC's `onDidResetContent` is always called first before other FRCs.

[FRC: Fetched results controller ([`NSFetchedResultsController`](https://developer.apple.com/documentation/coredata/nsfetchedresultscontroller))]

## Changes

- Since Order Details UI has 4 FRCs, and each result controller's `onDidResetContent` is called upon NSNotification observation, I thought it'd be safer to call `performFetch` for all FRCs before any reload UI action in each result controller's `onDidResetContent`.

## Testing

- Go to the Orders tab
- Tap on an order with the products section shown
- Go to the My Store tab > Settings, and log out --> the app should be logged out without crashing (it crashes consistently before the PR)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
@@ -100,7 +100,11 @@ private extension OrderDetailsResultsControllers {
             onReload()
         }
 
-        trackingResultsController.onDidResetContent = {
+        trackingResultsController.onDidResetContent = { [weak self] in
+            guard let self = self else {
+                return
+            }
+            self.refetchAllResultsControllers()
             onReload()
         }
 
@@ -112,7 +116,11 @@ private extension OrderDetailsResultsControllers {
             onReload()
         }
 
-        productResultsController.onDidResetContent = {
+        productResultsController.onDidResetContent = { [weak self] in
+            guard let self = self else {
+                return
+            }
+            self.refetchAllResultsControllers()
             onReload()
         }
 
@@ -124,10 +132,21 @@ private extension OrderDetailsResultsControllers {
             onReload()
         }
 
-        refundResultsController.onDidResetContent = {
+        refundResultsController.onDidResetContent = { [weak self] in
+            guard let self = self else {
+                return
+            }
+            self.refetchAllResultsControllers()
             onReload()
         }
 
         try? refundResultsController.performFetch()
+    }
+
+    func refetchAllResultsControllers() {
+        try? productResultsController.performFetch()
+        try? refundResultsController.performFetch()
+        try? trackingResultsController.performFetch()
+        try? statusResultsController.performFetch()
     }
 }
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
@@ -143,6 +143,8 @@ private extension OrderDetailsResultsControllers {
         try? refundResultsController.performFetch()
     }
 
+    /// Refetching all the results controllers is necessary after a storage reset in `onDidResetContent` callback and before reloading UI that
+    /// involves more than one results controller.
     func refetchAllResultsControllers() {
         try? productResultsController.performFetch()
         try? refundResultsController.performFetch()


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
```diff
@@ -124,10 +132,21 @@ private extension OrderDetailsResultsControllers {
...
+            guard let self = self else {
+                return
+            }
+            self.refetchAllResultsControllers()
             onReload()
         }
 
         try? refundResultsController.performFetch()
     }
+
```
pmusolino, line 145: I'd say to add a small comment about why we are doing these refetch. 
To the careless eye, it may seem superfluous and may be accidentally removed in the future.

pmusolino (approved):
Thanks Jaclyn for figuring it and explaining the issue really well!

The fix is simple and effective. I'm thinking if we can have other similar issues around the app because they are really hard to find. It would be ideal if we found a way for refreshing automatically the content of an instantiated `ResultsController` when we reset the context, without firing an `NSNotification`. It should fix the issue without performing a fetch manually, right? 💭 

Code ✅ 
Testing ✅ I'm able to replicate the crash on `develop`, but not there.

shiki:
> The objects from a FRC without performFetch after the reset have un unknown persistent store since the previous store was destroyed.

This is an interesting fact. If the FRC can recover after `CoreDataManager.reset()`, that means the `viewContext` also recovered. 🤔 But the `derivedStorage` could not recover and cause the crash 2 described in p91TBi-3iz-p2. 

Multi-threading == 💥

jaclync:
> This is an interesting fact. If the FRC can recover after CoreDataManager.reset(), that means the viewContext also recovered. 🤔 But the derivedStorage could not recover and cause the crash 2 described in p91TBi-3iz-p2.
> Multi-threading == 💥

Ugh yea multi-threading is hard! 🤯 Crash 2 in p91TBi-3iz-p2 is extra tricky because the updates/save happen on the child context's queue in the background, and we can't stop it if the main context is reset on the main queue in the middle of child context's async action 😞

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
jaclync: yes good idea! it does help with anyone reading the code in the future, added in https://github.com/woocommerce/woocommerce-ios/pull/2821/commits/f75f00336f8133fe704b16b0d33acee749387168

jaclync:
> I'm thinking if we can have other similar issues around the app because they are really hard to find. It would be ideal if we found a way for refreshing automatically the content of an instantiated ResultsController when we reset the context, without firing an NSNotification. It should fix the issue without performing a fetch manually, right? 💭

Yea the current setup with the `NSNotification` doesn't feel so safe. I checked other places where we implemented `onDidResetContent`, and it looks like they either only have one `ResultsController` or the UI is always presented as a modal so the user cannot log out (this could change when the app is logged out from Apple ID credential being revoked though - it's the [shipping providers screen](https://github.com/woocommerce/woocommerce-ios/blob/5163bf09618feca964118f6d52f9c3057f8dffbe/WooCommerce/Classes/ViewModels/ShippingProvidersViewModel.swift#L126-L138) which has very low usage now).

Ideally, after logging out (when the storage reset happens), all the view controllers/views should be torn down and gone and none of these `ResultsController`s should be around. It's a lot of work to change from our current tabs hanging around in logged out state, but I think it's worth it 🙂

jaclync:
thank you for the review & testing, Paolo & Jayson! 💟 I responded to the comments above, and merging this now!

shiki:
I am somewhat confident that this fixes #2251 so I will close that issue and mark the Sentry issue as resolved too. 

I'm just hoping to decrease the noise in Github and Sentry even if it's just for a few weeks. 🤪 

Thank you, Jaclyn!

PR #2828, opened 2020-09-16
Enables selection guide in ProductTableViewCell

Ecarrion:
# Why 

I noticed a strange lag when tapping on a top performer's cell. To debug that issue @astralbodies figured it would be nice to see the selection state of the cell to identify any touch problem.

# How

After enabling the selection style on the cell xib, Xcode changed the IB version, and the lag disappeared. 🤷 

# Testing Steps
- Launch the app
- Tap on a top performers product row
- Make sure that there is no lag and that the cell shows a selection guide

# Gif 
Lag + no Selection | No lag + Selection
---- | ----
original-lag |  new-no-lag
 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.xib
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.xib
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="15702" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="16097.3" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
     <device id="retina4_7" orientation="portrait" appearance="light"/>
     <dependencies>
         <deployment identifier="iOS"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="15704"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="16087"/>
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.xib
```diff
@@ -1,16 +1,16 @@
...
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="15704"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="16087"/>
         <capability name="Safe area layout guides" minToolsVersion="9.0"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
         <placeholder placeholderIdentifier="IBFilesOwner" id="-1" userLabel="File's Owner"/>
         <placeholder placeholderIdentifier="IBFirstResponder" id="-2" customClass="UIResponder"/>
-        <tableViewCell contentMode="scaleToFill" selectionStyle="none" indentationWidth="10" rowHeight="104" id="KGk-i7-Jjw" customClass="ProductTableViewCell" customModule="WooCommerce" customModuleProvider="target">
+        <tableViewCell contentMode="scaleToFill" selectionStyle="default" indentationWidth="10" rowHeight="104" id="KGk-i7-Jjw" customClass="ProductTableViewCell" customModule="WooCommerce" customModuleProvider="target">
```
jaclync, line 13: nit: do you think we can update this `selectionStyle` in code in `ProductTableViewCell` so that it's clear that this is something we intentionally changed? It's a bit harder to tell what's default vs. edited by us in xib 🙂

jaclync (approved):
Woo this works like magic! Though I kind of think it's a bug in the iOS 13+ simulator as I couldn't reproduce on a device (I tried checking time profiler for the bottom sheet case and nothing is going on in our code during the lag) 😅 The selected state is nice in any case!

Code: ✅ 
Testing: ✅ (I was able to repro the lag on iPhone 11 iOS 13.7 simulator, and now it's presented instantly with the highlighted state!)

WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.xib
Ecarrion: Sure https://github.com/woocommerce/woocommerce-ios/pull/2828/commits/2570a1f77f6de5deb3cb2b7008bac8a9125a59e9

PR #2832, opened 2020-09-17
Update StatsVersionCoordinator to return V4 as the default stats version

Ecarrion:
# Why

@astralbodies noticed that the experience for a new user is not the best since the "Deprecated Stats" screen is shown for a bit while the app fetches the stats version availability.

# How

To mitigate this, I'm making `.V4` to be the default version. This implies that a new user will see the stats loading screen immediately and will change to the deprecated stats screen only after validating that `.V4` is not available.

# Gifs

Before - V4 YES | After - V4 YES | After - V4 NO
---- | ---- | ----
before-v4-yes | after-v4-yes | after-no-v4

# Testing steps
- Make sure v4 is enabled by changing the line `45` in `StatsVersionCoordinator.swift` to
```swift
let nextVersion: StatsVersion = .v4
```
- Logout from the app.
- Login to your store.
- Make sure there is no glitch(screen swapping) when logging in.

Update release notes:

### DIFF
--- a/WooCommerce/WooCommerceTests/ViewRelated/Dashboard/DashboardUIFactoryTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Dashboard/DashboardUIFactoryTests.swift
@@ -61,7 +61,7 @@ final class DashboardUIFactoryTests: XCTestCase {
     }
 
     /// Stats v4 --> v4
-    func testWhenV4IsAvailableWhileStatsV4IsLastShown() {
+    func test_when_V4_is_available_while_stats_V4_is_last_shown() {
         mockStoresManager = MockupStatsVersionStoresManager(initialStatsVersionLastShown: .v4,
                                                             sessionManager: SessionManager.testingInstance)
         mockStoresManager.isStatsV4Available = true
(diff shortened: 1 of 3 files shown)


WooCommerce/WooCommerceTests/ViewRelated/Dashboard/DashboardUIFactoryTests.swift
```diff
@@ -15,47 +15,49 @@ final class DashboardUIFactoryTests: XCTestCase {
         super.tearDown()
     }
 
-    func testWhenV4IsAvailableWhileNoStatsVersionIsSaved() {
+    func test_it_loads_only_statsV4_when_V4_is_available_while_no_stats_version_is_saved() {
```
pmusolino, line 18: 💯 I'd say to refactor also the other method names in this class 🗡️

pmusolino (approved):
Looks good to me and it works like a charm @Ecarrion!

Code ✅ just a nit
Testing ✅

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/DashboardUIFactoryTests.swift
Ecarrion: Sure! https://github.com/woocommerce/woocommerce-ios/pull/2832/commits/7c3f2343619a42dfcdce56aff5c1e47b257e0afb

Issue #2830, opened 2020-09-17
SIWA: app crashes shortly after app launch when the Apple ID credential is revoked


jaclync:
**Describe the bug**

If the app relaunches after the Apple ID credential is revoked, it crashes shortly after launch. Relaunching the app brings it back to the logged out state. More details about the crash in p91TBi-3iz-p2 (crash 2).

**To Reproduce**

- Sign up / log in with SIWA, and wait for some data to load
- Terminate the app
- Go to iOS Settings and disconnect your Apple account (Settings > Apple ID > Password & Security > Apps using Apple ID > Remove WordPress app)
- Relaunch the app –> the app crashes shortly

**Expected behavior**

The app should be logged out after the Apple ID credential is revoked without crashing.

**Isolating the problem (mark completed items with an [x]):**
- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

**Mobile Environment**
Please include:
- Device: iPhone 6s Plus
- iOS version: iOS 13.7
- WooCommerce iOS version: any

**WordPress Environment**
<details>
```
Copy and paste the system status report from **WooCommerce > System Status** in WordPress admin.
```
</details>

PR #2833, opened 2020-09-17
Reset SnapshotsProvider When The StorageManager Is Reset

### LINKED ISSUE #2831: [DiffableDataSource] Crash When Logging Out And Back In
Found by @jaclync. See C6H8C3G23-p1600333375278800-slack.

To reproduce:

1. Launch the app in logged in state
2. Navigate to the Orders tab and wait for some data to load
3. Log out from the app
4. Log in to a store
5. Navigate to the Orders tab --> the empty cells appear (screenshot below) and the app crashes

shiki:
Fixes #2831. Thanks to @jaclync for finding this crash before production! 😌 

## Findings

The crash that happens after performing the steps in #2831 is something like this:

```
[error] error: Serious application error.  Exception was caught during Core Data change processing.  This is usually a bug within an observer of NSManagedObjectContextObjectsDidChangeNotification.  Object's persistent store is not reachable from this NSManagedObjectContext's coordinator with userInfo (null)
```

The stack trace indicates that the root cause is the same as what we experienced in https://github.com/woocommerce/woocommerce-ios/pull/2821 and https://github.com/woocommerce/woocommerce-ios/issues/2830. It's because we destroy the persistent store when the user logs out:

https://github.com/woocommerce/woocommerce-ios/blob/feb952d722b96decdac07fd42e9e75a678155432/Storage/Storage/CoreData/CoreDataManager.swift#L153-L158

As for why the `FetchResultSnapshotsProvider` reacts to this, I had thought that this was because we have an observer for `NSManagedObjectContextObjectsDidChange`:

https://github.com/woocommerce/woocommerce-ios/blob/feb952d722b96decdac07fd42e9e75a678155432/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L276-L281

But when I removed that code, the crash still happens. I'm positive that the crash is coming from [`NSFetchedResultsController`  itself](https://github.com/woocommerce/woocommerce-ios/blob/feb952d722b96decdac07fd42e9e75a678155432/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L122-L134), which also uses that `Notification`. 

## Solution

Luckily, this problem was already solved in our [`ResultsController`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2831-diffabledatasource-react-on-reset/Yosemite/Yosemite/Tools/ResultsController.swift). `ResultsController` simply executes `performFetch()` again if the `StorageManager` is reset. 

https://github.com/woocommerce/woocommerce-ios/blob/feb952d722b96decdac07fd42e9e75a678155432/Storage/Storage/CoreData/CoreDataManager.swift#L153-L158

And that's also what I came up with for `FetchResultSnapshotsProvider`:

https://github.com/woocommerce/woocommerce-ios/blob/feb952d722b96decdac07fd42e9e75a678155432/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift#L405-L408

## Testing

1. Launch the app in a logged-in state
2. Navigate to the Orders tab and wait for some data to load.
3. Log out from the app
4. Log in to a store.
5. Navigate to the Orders tab. Confirm that the app did not crash and orders are loaded and displayed on the table.

## Submitter Checklist

### DIFF
--- a/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
+++ b/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
@@ -167,8 +167,9 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
         try fetchedResultsController.performFetch()
     }
 
-    /// Returns `true` if the `start()` method was previously called.
-    private var isStarted: Bool {
+    /// Returns `true` if the `performFetch()` method was previously called, which activates
+    /// the `fetchedResultsController`.
+    private var fetchedResultsControllerIsActive: Bool {
         fetchedResultsController.fetchedObjects != nil
     }
 
@@ -388,7 +389,7 @@ private extension FetchResultSnapshotsProvider {
     ///    “tested” that it works because it was already called in `start()`.
     ///
     func restart() {
-        guard isStarted else {
+        guard fetchedResultsControllerIsActive else {
             return
         }
 
--- a/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
+++ b/Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
@@ -155,6 +155,11 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
         startObservingObjectsDidChangeNotifications()
     }
 
+    /// Returns `true` if the `start()` method was previously called.
+    private var isStarted: Bool {
+        fetchedResultsController.fetchedObjects != nil
+    }
+
     /// Retrieve the immutable type pointed to by `objectID`.
     ///
     /// This is typically used in the `UITableViewDiffableDataSource`'s `CellProvider` in order to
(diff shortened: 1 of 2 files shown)


Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -325,3 +374,48 @@ private struct ObjectsDidChangeNotification {
...
+    ///
+    /// This crash comes from `NSFetchedResultsController` itself and not from the observation
+    /// in `startObservingObjectsDidChangeNotifications()`.
+    ///
+    func startObservingStorageManagerDidResetNotifications() {
+        // Remove just in case this method was called already.
+        stopObservingStorageManagerDidResetNotifications()
+
+        storageManagerDidResetObservationToken =
+            notificationCenter.addObserver(forName: .StorageManagerDidResetStorage, object: storageManager, queue: nil) { _ in
```
jaclync: nit non-blocking suggestion: since this class is iOS 13+, we can take advantage of the [Combine way of observing a notification](https://developer.apple.com/documentation/combine/routing-notifications-to-combine-subscribers) 😄 then `stopObservingStorageManagerDidResetNotifications` could be just cancelling the `AnyCancellable` from the observation

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -325,3 +374,48 @@ private struct ObjectsDidChangeNotification {
...
+    ///
+    /// ```
+    /// [error] error: Serious application error.  Exception was caught during Core Data change
+    /// processing.  This is usually a bug within an observer of
+    /// NSManagedObjectContextObjectsDidChangeNotification.  Object's persistent store is not
+    /// reachable from this NSManagedObjectContext's coordinator with userInfo (null)
+    /// ```
+    ///
+    /// This crash comes from `NSFetchedResultsController` itself and not from the observation
+    /// in `startObservingObjectsDidChangeNotifications()`.
```
jaclync, line 387: thank you for the detailed comment 😍

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -183,6 +191,47 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
...
+@available(iOS 13.0, *)
+private extension FetchResultSnapshotsProvider {
+    /// Start `fetchedResultsController` fetching and dispatching of snapshots.
+    func activateFetchedResultsController() throws {
+        try fetchedResultsController.performFetch()
+    }
+
+    /// Returns `true` if the `activateFetchedResultsController()` method was previously called.
+    var fetchedResultsControllerIsActive: Bool {
+        fetchedResultsController.fetchedObjects != nil
```
jaclync, line 205: it's interesting that this doesn't crash the app when accessing `fetchedObjects` after the reset and before `performFetch()` again 🤔 like in the [stack trace](https://sentry.io/organizations/a8c/issues/1703307167/?project=1458804&query=Object%27s+persistent+store+is+not+reachable&statsPeriod=14d), it's crashing from getting the `fetchedObjects` from product FRC. I might just misunderstand the root cause of this crash 😅 

update: I tested adding `print(fetchedResultsController.fetchedObjects)` to above this nil check, and the app crashes on this print. It's possible that doing a nil check is fine while accessing the individual fetched objects is not (nil check might not involve accessing the objects from the store).

jaclync (approved):
Thanks for fixing this quickly! 💯 

Code: ✅ 
Testing: ✅ (used to crash on `develop` and not anymore on PR branch!)

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
shiki: Good find! Unfortunately, I don't know what else to use to determine if `performFetch` has been called. But if it doesn't crash on a `nil` check, I guess that's fine. 😄 Though I suppose we can use a `Bool` ivar in the class if it becomes a problem. 

I'm guessing that accessing `fetchedObjects` in a `print` causes the array objects to be _accessed_ from `NSManagedObjectContext` and that causes the crash. And maybe a `nil` check does not need to do that. 🤷

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
shiki: Great idea! Changed in 96000ead32790a98e49aadd85c8558ecf467ea9c and 711366647a6f4209789789dfdfec729da94ae611. I removed the `stop*` methods because I read that we don't need to also call `AnyCancellable.cancel()` in `deinit()`. Though I kept the `deinit` calls for now until I verify it myself. 😄

shiki (commented):
Thank you, @jaclync!

PR #2838, opened 2020-09-19
Issue/2642 persists downloadable files in original order for a product into storage

### LINKED ISSUE #2642: Support product downloads
From a unit test failure after updating `Product` to use the default `Equatable` implementation that compares all the properties (https://github.com/woocommerce/woocommerce-ios/pull/2641), an issue was discovered that we don't currently import `downloads` when upserting a `Product` to storage.

https://github.com/woocommerce/woocommerce-ios/blob/415ef2c77e378b14d68c05acdd313cad2d085edb/Yosemite/Yosemite/Stores/ProductStore.swift#L328-L335

When we're ready to support product downloads, let's add back the `downloads` to the test cases https://github.com/woocommerce/woocommerce-ios/pull/2641/commits/ad0db045026a2c6f75d2e2610a52c326d14e5859.

### LINKED ISSUE #2755: Add/Edit Products M5: Downloadable files Step 1 - display downloadable files for a product
As part of the M5 release, we need to provide the option to:
- display the downloadable files count for a product, if it exists.
- When clicking on the downloadable files count, we need to display the list of downloadable files for the product.

| Downloadable files cell            |  File List screen |
Frame 194 | iOS-add-price

**Note that this issue only takes care of displaying the downloadable files. The option to update or add files will be taken care of in another issue and is not part of the scope of this issue.**

partho-maple:
Fixes #2642 
Partially Fixes #2755 

## Description
This PR persists the downloadable files in the original order with their details on the product into the CoreData Storage. No visible changes on the UI. Verifiable only by unit test cases.

## Changes
- Added CoreData model version 31
- Changed `downloads` relationship of a `Product` entity, from unsorted to sorted and from `Nullify` to `Cascade`
- Regenerated `Product+CoreDataProperties.swift` file to reflect the changes made on CoreData
- Added a new method `handleProductDownloadableFiles` on `ProductStore.swift` to persist the downloadable files info
- Added a few extensions method
- Modified/added test cases

## Screenshots
No visible changes on the UI layer yet. Only verifiable via Test Cases

## Test Cases
Test cases have been added and modified accordingly.

### Testing
- Run the test cases on `ProductStoreTests` and `ProductMapperTests`

Update release notes:


Networking/NetworkingTests/Mapper/ProductMapperTests.swift
```diff
@@ -189,6 +190,23 @@ final class ProductMapperTests: XCTestCase {
...
+    func test_that_product_downloadable_files_are_properly_mapped() {
+        let product = mapLoadProductResponse()
+
+        guard let files = product?.downloads else {
+            XCTFail("Failed to parse product category")
+            return
+        }
+        XCTAssertEqual(files.count, 2)
+
+        let productDownloadableFile = files[0]
```
pmusolino: I suggest you to do something like that:
```
guard let productDownloadableFile = files?[0] else {
            XCTFail("Missing product downlodable file")
            return
        }
```

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
```diff
@@ -189,6 +190,23 @@ final class ProductMapperTests: XCTestCase {
...
+        XCTAssertEqual(productDownloadableFile.downloadID, "b6843219-775a-409d-b428-dd2ffe3af2c2")
+        XCTAssertEqual(productDownloadableFile.name, "T-shirt")
+        XCTAssertEqual(productDownloadableFile.fileURL, "https://mywootesting.mystagingwebsite.com/wp-content/uploads/2020/09/рок-метал-24-scaled.jpg")
```
pmusolino: Can you edit it, pointing it to a nonexisting URL? E.g: `https://example.com`

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
```diff
@@ -253,7 +271,7 @@ private extension ProductMapperTests {
     /// Returns the ProductMapper output upon receiving `product`
     ///
     func mapLoadProductResponse() -> Product? {
-        return mapProduct(from: "product")
+        return mapProduct(from: "product-downloadable")
```
pmusolino: This is not clear to me. Why are you pointing to a specific JSON file now? The JSON file `product`  it should continue to exist. If you need to use a modified version, since the only changes are to downloadable files, you can edit the `product.json` file, changing downloadable files according to your needs.

Storage/Storage/Tools/StorageType+Extensions.swift
```diff
@@ -319,6 +319,13 @@ public extension StorageType {
         return firstObject(ofType: ProductTag.self, matching: predicate)
     }
 
+    /// Retrieves the Stored Product Downloadable File.
+    ///
+    func loadProductDownloadableFile(downloadableFileID: String) -> ProductDownload? {
+        let predicate = NSPredicate(format: "downloadID = %@", downloadableFileID)
+        return firstObject(ofType: ProductDownload.self, matching: predicate)
+    }
```
pmusolino: Should we have in plan to use this method? I mean... do we want to get a `ProductDownload` from a `downloadID`? Perhaps a method for getting all the `ProductDownload`s of a specific product would be more useful. Or maybe just adding it maintaining `loadProductDownloadableFile`. What do you think?

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 30.xcdatamodel/contents
```diff
@@ -255,7 +255,7 @@
         <relationship name="categories" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="ProductCategory" inverseName="products" inverseEntity="ProductCategory"/>
         <relationship name="defaultAttributes" optional="YES" toMany="YES" deletionRule="Cascade" destinationEntity="ProductDefaultAttribute" inverseName="product" inverseEntity="ProductDefaultAttribute"/>
         <relationship name="dimensions" optional="YES" maxCount="1" deletionRule="Cascade" destinationEntity="ProductDimensions" inverseName="product" inverseEntity="ProductDimensions"/>
-        <relationship name="downloads" optional="YES" toMany="YES" deletionRule="Nullify" destinationEntity="ProductDownload" inverseName="product" inverseEntity="ProductDownload"/>
+        <relationship name="downloads" optional="YES" toMany="YES" deletionRule="Nullify" ordered="YES" destinationEntity="ProductDownload" inverseName="product" inverseEntity="ProductDownload"/>
```
pmusolino: This change should be always applied to a new model.
You should also update the `MIGRATIONS.md` file 👍

Networking/NetworkingTests/Responses/product-downloadable.json
```diff
@@ -0,0 +1,238 @@
...
+                }
+            ],
+            "variations": [
+                192,
+                194,
+                193
+            ],
+            "grouped_products": [],
+            "menu_order": 0,
+            "meta_data": [
```
pmusolino: This field can be removed 👍

Networking/NetworkingTests/Responses/product-downloadable.json
```diff
@@ -0,0 +1,238 @@
...
+                        "to": "thuy.copeland@automattic.com",
+                        "subject": "Booking Event",
+                        "show_subject": "no",
+                        "widget": 0,
+                        "id": 282,
+                        "submit_button_text": "Submit"
+                    }
+                }
+            ],
+            "jetpack_publicize_connections": [],
```
pmusolino: also this one 👍

Yosemite/Yosemite/Model/Storage/Product+ReadOnlyConvertible.swift
```diff
@@ -8,6 +8,10 @@ extension Storage.Product {
     var tagsArray: [Storage.ProductTag] {
         return tags?.toArray() ?? []
     }
+    var downloadableFilesArray: [Storage.ProductDownload] {
+        let downloadableFiles: [Storage.ProductDownload]? =  downloads?.toArray()
+        return downloadableFiles ?? []
```
pmusolino:
```suggestion
        return downloads?.toArray() ?? []
```

Yosemite/Yosemite/Model/Storage/Product+ReadOnlyConvertible.swift
```diff
@@ -119,7 +123,7 @@ extension Storage.Product: ReadOnlyConvertible {
                        totalSales: Int(totalSales),
                        virtual: virtual,
                        downloadable: downloadable,
-                       downloads: productDownloads.sorted(),
+                       downloads: productDownloads,
```
pmusolino, line 125: nice catch 👍

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
```diff
@@ -189,6 +190,23 @@ final class ProductMapperTests: XCTestCase {
         XCTAssert(productImage.alt?.isEmpty == true)
     }
 
+    /// Test that product downloadable files are properly mapped.
+    ///
+    func test_that_product_downloadable_files_are_properly_mapped() {
+        let product = mapLoadProductResponse()
+
+        guard let files = product?.downloads else {
+            XCTFail("Failed to parse product category")
```
pmusolino:
```suggestion
            XCTFail("Failed to parse product downloads")
```

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
```diff
@@ -189,6 +190,23 @@ final class ProductMapperTests: XCTestCase {
         XCTAssert(productImage.alt?.isEmpty == true)
     }
 
+    /// Test that product downloadable files are properly mapped.
+    ///
+    func test_that_product_downloadable_files_are_properly_mapped() {
```
pmusolino, line 195: Can you use our practice Given/When/Then or Arrange/Action/Assert? Look at [this](https://github.com/woocommerce/woocommerce-ios/blob/2b32dcd38531c3bb320fd793a77bec5f85684847/WooCommerce/WooCommerceTests/ViewRelated/inAppFeedback/InAppFeedbackCardViewControllerTests.swift) for an example.

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -1578,6 +1592,16 @@ private extension ProductStoreTests {
         return [defaultAttribute1]
     }
 
+    func sampleDownloadsMutated() -> [Networking.ProductDownload] {
+        let download1 = ProductDownload(downloadID: "1f9c11f99ceba63d4403c03bd5391b11",
+                                        name: "Song #1",
+                                        fileURL: "https://woocommerce.files.wordpress.com/2017/06/woo-single-1.ogg")
```
pmusolino: Can you change it to `https://example.com`? 🙇

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -1578,6 +1592,16 @@ private extension ProductStoreTests {
...
+        let download2 = ProductDownload(downloadID: "ec87d8b5-1361-4562-b4b8-18980b5a2cae",
+                                        name: "Artwork",
+                                        fileURL: "https://thuy-test.mystagingwebsite.com/wp-content/uploads/2018/01/cd_4_angle.jpg")
```
pmusolino: Same here 👍

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -961,9 +969,11 @@ final class ProductStoreTests: XCTestCase {
         let expectedProductSaleEnd = date(with: "2019-10-27T21:29:50")
         let expectedProductTaxStatus = "taxable"
         let expectedProductTaxClass = "reduced-rate"
+        let expectedDownloadableFileCount = 0
+        let expectedDownloadable = false
 
         network.simulateResponse(requestUrlSuffix: "products/\(expectedProductID)", filename: "product-update")
-        let product = sampleProduct(productID: expectedProductID)
+        let product = sampleProduct(productID: expectedProductID, downloadable: true)
```
pmusolino: I'd say that in this specific case you can maintain `downloadable` to `false`.

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -737,7 +743,7 @@ final class ProductStoreTests: XCTestCase {
     ///
     func testUpdateStoredProductEffectivelyPersistsNewProduct() {
         let productStore = ProductStore(dispatcher: dispatcher, storageManager: storageManager, network: network)
-        let remoteProduct = sampleProduct()
+        let remoteProduct = sampleProduct(downloadable: true)
```
pmusolino, line 746: I'm curious about this change: why `true`? 💭 It seems that it's not used.

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -1578,6 +1592,16 @@ private extension ProductStoreTests {
         return [defaultAttribute1]
     }
 
+    func sampleDownloadsMutated() -> [Networking.ProductDownload] {
```
pmusolino: I suggest you to move it after `sampleDownloads()` 👍

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
pmusolino: Might be worth also comparing the full `ProductDownload` after making it Equatable in case we add more properties to the model.

pmusolino (changes requested):
Thanks for opening this PR @partho-maple! 
In general, the approach is right, but I encountered a crash on the model migration trying to run the tests, after having installed an app version coming from the `develop` branch. This happens because probably, you did the changes on an existing model, instead of creating a new storage model (the storage model n° 30 was already existing, but you should have created model 31), and it's really dangerous because this means crashing all app updates.

I also added some extra suggestions 👍

Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -512,6 +513,31 @@ private extension ProductStore {
...
+
+    /// Updates, inserts, or prunes the provided StorageProduct's downloadable files using the provided read-only Product's downloadable files
+    ///
+    func handleProductDownloadableFiles(_ readOnlyProduct: Networking.Product, _ storageProduct: Storage.Product, _ storage: StorageType) {
+
+        // Removes all the downloadable files first.
+        storageProduct.downloadableFilesArray.forEach { existingStorageDownloadableFile in
+            storage.deleteObject(existingStorageDownloadableFile)
+            storageProduct.removeFromDownloads(existingStorageDownloadableFile)
+        }
```
Ecarrion, line 537: [Q] If we are deleting all files for a product here, wouldn't the `loadProductDownloadableFile` in line 530 always be nil? 🤔

Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -512,6 +513,31 @@ private extension ProductStore {
...
+
+        var storageDownloads = [StorageProductDownload]()
```
Ecarrion: I think this could be an `ordered set` and avoid the conversion in line 539!

Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -512,6 +513,31 @@ private extension ProductStore {
...
+            if let existingStorageDownloadableFile = storage.loadProductDownloadableFile(downloadableFileID: readOnlyDownloadableFile.downloadID) {
+                existingStorageDownloadableFile.update(with: readOnlyDownloadableFile)
+                storageDownloads.append(existingStorageDownloadableFile)
+            } else {
+                let newStorageDownloadableFile = storage.insertNewObject(ofType: Storage.ProductDownload.self)
+                newStorageDownloadableFile.update(with: readOnlyDownloadableFile)
+                storageDownloads.append(newStorageDownloadableFile)
+            }
+        }
+        storageProduct.addToDownloads(NSOrderedSet(array: storageDownloads))
```
Ecarrion: What do you think about extracting from line 522 to line 525 and from line 528 to line 538 into their own private methods? That would increase the readability of this chunk!

Ecarrion (commented):
Good job on the PR @partho-maple!
I left a few questions and suggestions for you to look at!

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
pmusolino: In another PR I saw @jaclync using `https://photo.jpg` I think it would be nice to use it also here 👍

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
partho-maple: @pmusolino , thanks a lot for the review and pointers regarding CD model versioning. 😃 

Regarding this one, I am not really sure about the suggestion. But what I have done here is,

- Unwrapped the `product?.downloads` into `files`. So `files` is already unwrapped.
- `XCTAssertEqual(files.count, 2)` this line confirm is we have 2 files or not. 
- And `product?.downloads` is also non-optional

So it won't allow to unwrap a non-optional again like this `guard let productDownloadableFile = files?[0]` or this `if let productDownloadableFile = files?[0]`

But I can check the count manually. like so,
`
if files.count < 1 {
     XCTFail("Missing product downloadable file")
     return
}
`
Although it was a nice catch. A possible crash point. 👏

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
partho-maple: Got it. Edited accordingly

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
partho-maple: Yes, you are right. Initially, I tried with changing the existing `product .json` file but it broke a few existing test cases, so I decided not to touch the existing ones and add a new file. But since you questioned this choice, now I think it would be better to do what you suggested and fix the failing test cases.

Storage/Storage/Tools/StorageType+Extensions.swift
partho-maple: Currently, a `product` can have multiple downloadable files. And a downloadable file would belong to only one `product`.

But can a single downloadable file(with the same `downloadID ` ) belong to multiple `product`? If yes, we might need this method.

Also, I was actually using that method on `ProductStore.swift` file at line `530`. But based on @Ecarrion 's suggestion, I removed that chunk of code, hence that method is no longer necessary. SO removed that also.

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 30.xcdatamodel/contents
partho-maple: Thanks for the pointer. Changed accordingly.

Networking/NetworkingTests/Responses/product-downloadable.json
partho-maple: Removed the whole `product-downloadable.json` file

Networking/NetworkingTests/Responses/product-downloadable.json
partho-maple: Removed the whole `product-downloadable.json` file

Yosemite/Yosemite/Model/Storage/Product+ReadOnlyConvertible.swift
partho-maple: Yeap. changed. 
Went a bit verbose 😛

Yosemite/Yosemite/Model/Storage/Product+ReadOnlyConvertible.swift
partho-maple: 😄

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
partho-maple: Changed

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
partho-maple: Great suggestion. Changed accordingly

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
partho-maple: Changed accordingly

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
partho-maple: Changed accordingly

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
partho-maple: It has not been used there. I did some tests and later removed it. Somehow I missed this one. Thanks for pointing out.

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
partho-maple: Actually it has been used here.
That `remoteProduct ` will have 3 downloadable files since it's `sampleProduct(downloadable: true)`
So at line 755, we are checking that there are not downloadable files before `upsertStoredProduct`

And after `upsertStoredProduct` the `remoteProduct` we should have 3 `ProductDownload` which has been checked at line 767

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
partho-maple: Moved

Yosemite/Yosemite/Stores/ProductStore.swift
partho-maple: Yes, you are right. But for safety, I added that block.
Though, now it seems redundant to have that. Removed that block.
Thanks for the great insight and for pointing that out. 🙂

Yosemite/Yosemite/Stores/ProductStore.swift
partho-maple: In the end, we need an `NSOrderedSet` which is not mutable. The only way we can create it is from an Array or `NSMutableOrderedSet`
So we need atleast one conversion. Either `Array` >> `NSOrderedSet` or `NSMutableOrderedSet` >> `Array` >> `NSOrderedSet`

Yosemite/Yosemite/Stores/ProductStore.swift
partho-maple: I think that would be a great idea and also readability would improve a lot. 
But I am not sure on one point. All other methods on this class that are performing similar types of tasks (as an example `handleProductTags` , `handleProductImages` , `handleProductDefaultAttributes` , `handleProductAttributes` etc) are designed in the similar fashion as it is currently/above. Taking that into consideration what do you suggest?

@pmusolino , would you please share your thoughts on this?

Yosemite/Yosemite/Stores/ProductStore.swift
Ecarrion: I think you can return a `NSMutableOrderedSet` since it inherits from an `NSOrderedSet`. But I guess it's not a big deal.

Yosemite/Yosemite/Stores/ProductStore.swift
Ecarrion: I bring that up because I had to read the method several times in order to understand what it was doing. Having smaller helper methods with meaningful names would have turned that into a single read. I think we can always improve what we have been doing so far, but I'll let @pmusolino to decide! :-)

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
pmusolino: It's totally fine to update the tests cases 👍

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
pmusolino: I wasn't noticed that you was unwrapping it, so it's totally fine to remove this check 
```
if files.count < 1 {
            XCTFail("Missing product downloadable file")
            return
        }
```
since you are also checking `XCTAssertEqual(files.count, 3)`. 👍

Yosemite/Yosemite/Stores/ProductStore.swift
pmusolino: I'd say that, just for not having discrepancies inside the same class, we can split it, adding some comments that better describe every pass of the method. In the future, we can think about some generic ways for better split things. What do you think?

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
partho-maple: OK.

Networking/NetworkingTests/Mapper/ProductMapperTests.swift
partho-maple: Already updated.  🙂

Yosemite/Yosemite/Stores/ProductStore.swift
partho-maple: I agree with both of you. I have splited that into 2 smaller methods.

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 31.xcdatamodel/contents
```diff
@@ -0,0 +1,519 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
pmusolino, line 1: I noticed that the "Delete Rule" of `downloads` is set to "nullify", but should it be set to "Cascade"? 
What do you think we should use in this case @Ecarrion @jaclync @shiki?

pmusolino (commented):
Looks good to me @partho-maple!

I've just a concern about the Delete Rule of `downloads`, I'm not sure we should use `Nullify` for them.

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 31.xcdatamodel/contents
partho-maple: **So far what I have tested:**
The `downloadID` of each `ProductDownload` entity is unique, even if the `name` and `fileURL` are same. If you add a downloadable file with the same name and same URL on a single product or multiple products, the `downloadID` is unique everywhere. This proves that a single downloadable file belongs to only one product.

In my opinion, when we are deleting a `Product` from CoreData, all of its downloadable files need to be deleted. So, to me, `Cascade ` makes more sense.

Please rectify me if I am wrong or miss anything.

pmusolino: I agree with you, `Cascade` makes more sense because we don't want to continue to store `ProductDownload` entities when the `Product` in relation to one of these entities was deleted. I'd say to change it to `Cascade`, unless there is a valid reason to keep it set to `nullify` (and as of today, it doesn't seem like there is).

Yosemite/Yosemite/Stores/ProductStore.swift
pmusolino: Any change on this @partho-maple?

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 31.xcdatamodel/contents
partho-maple: Updated to `Cascade `

Yosemite/Yosemite/Stores/ProductStore.swift
partho-maple: Yeap. pushed.  😛

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 31.xcdatamodel/contents
jaclync: yea "cascade" makes sense to me, since each download is unique to a product (it can't be shared with other products) so we can just delete them when a product is deleted 👍

pmusolino: Thank you both! 💯

pmusolino (approved):
👏 congrats @partho-maple! I'm going to merge your first (important) PR after that the CI finish to run 💯

PR #2845, opened 2020-09-21
SIWA: fix crash shortly after app launch when the Apple ID credential is revoked

### LINKED ISSUE #2830: SIWA: app crashes shortly after app launch when the Apple ID credential is revoked

jaclync:
Fixes #2830 

## Context

The crash is most likely from a race condition from saving on a child's background queue while the main context is reset on the main context's queue (main thread):

Core Data child context save crash

Originally, I thought it's enough to create a thread-safe identifier before saving a child context but it only reduces the chance and it still crashes once in a while because this identifier could change in the middle of a child context's `perform` in a background queue. After thinking more about the crash, the only way I can think of is to compare this identifier **on the main context's queue** and perform the save there. I haven't got to repro the crash after testing 10 times. One thing I was a bit worried about is saving a child context on the main thread, but I didn't notice frozen UI that's different from `develop`. Feel free to share any concerns and any lag you might notice during testing though!

## Changes

In `CoreDataManager`:
- Keep a unique string internally that is set to a different UUID on every storage reset
- When creating a child context (`newDerivedStorage`), set the unique string to the child [context's `name`](https://developer.apple.com/documentation/coredata/nsmanagedobjectcontext/1506231-name):
> The developer-provided name of the context
- When saving a child context (`saveDerivedType`):
  - Get the child context's `name` (this can only be done on child context's queue via `perform` block)
  - Switch to the main context's queue 
    - Compare the child context's name against `CoreDataManager`'s context name, early return if different
    - Perform saving the child context
    - Perform saving the main context like before

## Testing

- Terminate the app
- Go to iOS Settings and disconnect your Apple account (see steps at the bottom)
- Relaunch the app --> the app is not supposed to crash (on `issue/2762-siwa-revoked-logout` branch, it crashes pretty often)

Update release notes:


shiki:
> One thing I was a bit worried about is saving a child context on the main thread, but I didn't notice frozen UI that's different from develop

The UI is supposed to freeze all the time as far as I understand. Saving from child to parent is usually fast because no disk access happens. But saving from the parent context that is directly attached to the persistent store will be slow since that requires disk access. And because our “parent” context is the main context (main thread), the app will always wait in the main thread when Core Data is saving to the persistent store (Sqlite). 😬 

I'll get back to this PR tomorrow! I'm kinda scared about calling `save` inside the wrong thread of the `NSManagedObjectContext`. I can't remember how I crashed the app last time. 🤪

shiki (changes requested):
Hi @jaclync! Thanks for the diagram. ❤️  This does fix the crash! The problem is, as I feared, we shouldn't be calling `NSManagedObjectContext.save()` on the wrong thread. 😞 

### ConcurrencyDebug

Core Data has this launch argument that we can turn on: `-com.apple.CoreData.ConcurrencyDebug 1`. When that is turned on, the app will crash if it sees a Core Data multithreading violation. I added this to the launch arguments:



And then I ran the app normally. I immediately got a crash, pointing to the `derivedStorage.saveIfNeeded()`. 



I wanted to have this flag turned on for our development scheme for some time. I just forget every time. This flag will save us from Core Data headaches in the future. It's already turned on in WPiOS. 

So, I believe we shouldn't proceed with this change. It will prevent us from turning on this concurrency flag. And it's a violation of Core Data best practices. 😞 

### The Error Logs

Even if this did fix the crash, there are still the underlying errors logged like this: 

```
2020-09-22 14:25:41.093805-0600 WooCommerce[2318:1942253] [error] error:  API Misuse: Attempt to serialize store access on non-owning coordinator (PSC = 0x282a9a040, store PSC = 0x0)
CoreData: error:  API Misuse: Attempt to serialize store access on non-owning coordinator (PSC = 0x282a9a040, store PSC = 0x0)
2020-09-22 14:25:41.094872-0600 WooCommerce[2318:1942253] [error] error:  API Misuse: Attempt to serialize store access on non-owning coordinator (PSC = 0x282a9a040, store PSC = 0x0)
CoreData: error:  API Misuse: Attempt to serialize store access on non-owning coordinator (PSC = 0x282a9a040, store PSC = 0x0)
2020-09-22 14:25:41.096324-0600 WooCommerce[2318:1942253] [error] error:  API Misuse: Attempt to serialize store access on non-owning coordinator (PSC = 0x282a9a040, store PSC = 0x0)
```

This log happens both in `develop` (when it crashes) and for this branch (when it doesn't crash). I believe this means we still didn't fix the concurrency issues. 😞 

--- 

I'm not quite sure what do at this point. I think an update to the architecture is due. Do you have some short-term ideas we could try? I'd be happy to work with you on this.

shiki:
@jaclync An idea that I'm not sure will work:

1. Do not destroy the persistent store
2. Lock all `derivedStorage.perform` inside a thread that **we** manage. 
3. Change `CoreDataManager.reset()` to [delete all entities from the database](https://stackoverflow.com/a/33931528) instead of destroying the persistent store. But **before doing that**, wait for all `derivedStorage` calls to finish. And then prevent them from running until the deletion is done. 

---

On second thought, there will be some network calls lying around so we might still end up saving some data in the background. I guess that's fine? 🤷 

Just thinking out loud for now. 😄

jaclync:
@shiki Thanks for sharing the Core Data concurrency debug flag, I had a feeling saving on the main thread wouldn't be a great idea in any case. For the error messages on `CoreData: error:  API Misuse: Attempt to serialize store access on non-owning coordinator`, I believe they are from the child context performing some storage operations before saving after the main context is reset.

Yea, it doesn't seem like it's possible to solve the race condition when resetting the main context & saving the child context in two different threads asynchronously. Not resetting the main context sounds like the best option.

It's not great for `CoreDataManager` to depend on other singletons like `StoresManager`, but maybe we can DI the `isLoggedIn` observable to `CoreDataManager` so that it knows when to discard the `saveDerivedType` calls from a child context. So the flow is like:

- The app changes from logged in to logged out state:
  - Update the `isLoggedIn: Bool` internal property
  - Delete all objects in the main context and save (we might have to go through each entity to delete all)
  - When any `saveDerivedType` comes in, early return when `isLoggedIn` is `false`
    - There could still be some save calls that already reach the `derivedStorage.perform { save }` queue before the `isLoggedIn` change, we can delete all objects again in [`self.viewStorage.perform { ... }`](https://github.com/woocommerce/woocommerce-ios/blob/3030e66b8433a6965030931d6d2e4b02cd79b564/Storage/Storage/CoreData/CoreDataManager.swift#L133) when it gets back to the main context's queue

> Lock all derivedStorage.perform inside a thread that we manage.

Right now we have all different child contexts for each Yosemite store action, and each context has its own queue/thread. Do you know how to run perform on a child context in a thread we own, and why?

jaclync:
closing this PR since saving a child context on parent context's queue isn't a good practice, we'll figure out another way to fix the crash later!

PR #2846, opened 2020-09-21
UI differences between adding and editing a product (except for "saving as draft")

jaclync:
Part of #2740 

This PR implements the following UI differences between adding and editing a product:

- Navigation bar "Publish" CTA is always shown (where "Update" CTA is when editing a product) and it creates a product with the status in product settings
- Navigation bar > ellipsis menu:
  - **[NEXT PR]** Add "Save as draft" action to create a product with status "draft", regardless of the status in product settings
  - Disable "View Product in Store" and "Share" actions
- **[NEXT PR]** Unsaved changes action sheet (when the user navigates away from product details with outstanding changes):
  - Add "Save as draft" action (same behavior as “Save as draft” above)
- Hide product type row (it will be editable again in M5 when we support adding a variable product)

## Changes

- In `ProductFormActionsFactory`, DI'ed `formType: ProductFormType` and only showed the product type row when it is editing a simple/external/grouped product
 - Updated unit tests and added test cases for adding a product in `ProductFormActionsFactory_AddProductTests`
- In `ProductFormViewController `'s `updateNavigationBar(isUpdateEnabled: Bool)`, always showed a Publish CTA (same functionality as Update CTA for edit product) when adding a product
- Added `func canShareProduct() -> Bool` to `ProductFormViewModelProtocol`, and updated its implementation `ProductFormViewModel`'s `canShareProduct` and `canViewProductInStore` to be enabled only when editing a product
  - Added test cases in `ProductFormViewModelTests`

### Creating a simple product

- Go to the Products tab
- Tap on "+" CTA in the navigation bar
- Tap "Simple product" --> there should be no product type row in the form, and "Publish" CTA is always visible
- Tap the ellipsis menu in the navigation bar --> there should be only one action "Product Settings"
- Tap "Product Settings" and feel free to edit any fields you like
- Go back to the form and fill in any fields you like
- Tap "Publish" --> the simple product should be created remotely with the fields you entered
- Tap the ellipsis menu in the navigation bar --> there should be "View Product in Store" if the product has published status, "Share", and "Product Settings" in the action sheet

### Creating an external product

Repeat the same steps as creating a simple product, but select "External product" in step 4 and make sure the expected external product is created remotely.

### Creating a grouped product

Repeat the same steps as creating a grouped product, but select "Grouped product" in step 4 and make sure the expected grouped product is created remotely.

## Example screenshots

Please note the navigation bar, ellipsis menu action sheet, and the lack of product type row when adding a product. "Save as draft" will be implemented in the next PR.

adding a simple product | editing a simple product
-- | --
Simulator Screen Shot - iPhone 11 - 2020-09-21 at 14 30 48 | Simulator Screen Shot - iPhone 11 - 2020-09-21 at 14 32 17

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -663,6 +670,11 @@ private extension ProductFormViewController {
         }
     }
 
+    func createPublishBarButtonItem() -> UIBarButtonItem {
+        let updateTitle = NSLocalizedString("Publish", comment: "Action for creating a new Product remotely")
+        return UIBarButtonItem(title: updateTitle, style: .done, target: self, action: #selector(updateProduct))
+    }
```
pmusolino:
```suggestion
    func createPublishBarButtonItem() -> UIBarButtonItem {
        let publishTitle = NSLocalizedString("Publish", comment: "Action for creating a new Product remotely")
        return UIBarButtonItem(title: publishTitle, style: .done, target: self, action: #selector(updateProduct))
    }
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -663,6 +670,11 @@ private extension ProductFormViewController {
         }
     }
 
+    func createPublishBarButtonItem() -> UIBarButtonItem {
+        let updateTitle = NSLocalizedString("Publish", comment: "Action for creating a new Product remotely")
+        return UIBarButtonItem(title: updateTitle, style: .done, target: self, action: #selector(updateProduct))
```
pmusolino: Do we really want to call the `updateProduct` method here? Probably the behavior it's the same, but inside `updateProduct` we call also the analytics event `eventLogger.logUpdateButtonTapped()`. This would affect our stats.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -120,7 +122,11 @@ extension ProductFormViewModel {
     }
 
     func canViewProductInStore() -> Bool {
-        return originalProduct.product.productStatus == .publish
+        originalProduct.product.productStatus == .publish && formType == .edit
```
pmusolino, line 125: Probably in the future should be better to rename `publish` to `published`. The key `publish` sounds like something that has yet to be done, and not like a status. 🌚

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
```diff
@@ -93,12 +93,16 @@ final class ProductVariationFormViewModel: ProductFormViewModelProtocol {
 //
 extension ProductVariationFormViewModel {
     func canEditProductSettings() -> Bool {
-        return false
+        false
     }
 
     func canViewProductInStore() -> Bool {
         // no-op
```
pmusolino: Should we remove this comment?

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormActionsFactory+AddProductTests.swift
```diff
@@ -0,0 +1,109 @@
+import XCTest
+
+@testable import WooCommerce
+import Yosemite
+
+final class ProductFormActionsFactory_AddProductTests: XCTestCase {
```
pmusolino, line 6: Nice 💯

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormViewModelTests.swift
```diff
@@ -0,0 +1,118 @@
+import Photos
```
pmusolino: nit: this import seems not needed 👍

pmusolino (approved):
Looks good to me 💯 great work @jaclync 

Code ✅ just some nits, one in particular about using the method `updateProduct()` when a navigation bar button is pressed

Testing:
- Tap "Simple product" --> there should be no product type row in the form, and "Publish" CTA is always visible ✅ 
- Tap the ellipsis menu in the navigation bar --> there should be only one action "Product Settings" ✅ 
- Tap "Publish" --> the simple product should be created remotely with the fields you entered ✅ 
- Tap the ellipsis menu in the navigation bar --> there should be "View Product in Store" if the product has published status, "Share", and "Product Settings" in the action sheet ✅ 

I've just doubt about adding a new "empty" simple Product (without the title). It will be saved, but the name of the product will be `Product`. Should we leave it empty, like when we edit a product with an empty title?

Garance91540 (approved):
Looking great! Can't wait to see more 🤩

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
jaclync: yes, it is confusing to see product status is `.publish` while it means "published"! I created this issue to track this change https://github.com/woocommerce/woocommerce-ios/issues/2851 🙂

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: great catch on the analytics event! I created a different function `publishProduct` for this case, and renamed the functions for saving a product remotely (either from adding/editing) here: https://github.com/woocommerce/woocommerce-ios/pull/2846/commits/7d03e6f5f713f2bce57f4d1127f4f50307cf0c96

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: oops! updated in https://github.com/woocommerce/woocommerce-ios/pull/2846/commits/19396ff0470831b2f357f600d2619d1f8373dfe5 😄

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
jaclync: yea makes sense! updated in https://github.com/woocommerce/woocommerce-ios/pull/2846/commits/19396ff0470831b2f357f600d2619d1f8373dfe5

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormViewModelTests.swift
jaclync: indeed! removed in https://github.com/woocommerce/woocommerce-ios/pull/2846/commits/19396ff0470831b2f357f600d2619d1f8373dfe5

jaclync:
Thanks for the review, Paolo & G and for catching the analytics event! @pmusolino I updated the PR and responded to your comments for another look 🙂

jaclync:
> I've just doubt about adding a new "empty" simple Product (without the title). It will be saved, but the name of the product will be Product. Should we leave it empty, like when we edit a product with an empty title?

@pmusolino I forgot to reply to this question: the "Product" name is actually populated by the server when the submitted product has an empty name 😐 we could set "Product" to be the default name, but I'm not sure how we localize this string to be the same as the backend

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
pmusolino: This is great! 💯

pmusolino:
> @pmusolino I forgot to reply to this question: the "Product" name is actually populated by the server when the submitted product has an empty name 😐 we could set "Product" to be the default name, but I'm not sure how we localize this string to be the same as the backend

Ohh I didn't know that. I'd say then to leave it populated by the backend :nod:

pmusolino:
BTW, changes look great! 💯

PR #2850, opened 2020-09-22
Reviews tab: refactor to have one instance per site ID

jaclync:
Part of #1838 

## Background

In this PR, a new observable for the currently logged in site ID is introduced to `SessionManager`/`StorageManager` so that the main tab (`MainTabBarController`) can subscribe to site ID changes.

- When the app is logged out (`nil` site ID), the main tab's view controllers become empty
- If the user is logged in to a site (non-nil site ID), the main tab's view controllers are set up given the site ID (only Reviews tab is taking in the site ID in this PR; other tabs will follow in future PRs)

Any feedback on the observable implementation is welcome!

## Changes

- Added an observable site ID to `SessionManager` and `StoresManager` (just forwarding `SessionManager`'s observable)
  - Since the site ID is stored in `UserDefaults`, I created `UserDefaults.defaultStoreID` so that we can observe this dynamic var. The return type has to be Objective-C compatible for KVO, so I converted `Int` to `Int64` before emitting the value.
  - Updated the struct to a class, so that we can invalidate the KVO (even though it's a singleton in the app). This also allowed us to update some `var` to `let` in unit tests.
- In `MainTabBarController`'s `viewDidLoad`, replaced the current `viewControllers` setup with subscription for site ID changes so that `viewControllers` follow the behavior:
  - Nil site ID: `viewControllers` are empty, and the site ID specific Reviews coordinator is set to nil
  - Non-nil site ID: `viewControllers` are set up with the existing Dashboard, Orders, and Products tabs plus a new instance of Reviews coordinator that is tied to a site ID. In future PRs, Dashboard, Orders, and Products tabs are also going to be torn down on each site ID change.
- DI'ed site ID to Reviews tab classes - `DefaultReviewsDataSource`, `ReviewDetailsViewController` `ReviewsCoordinator`, `ReviewsViewController`, `ReviewsViewModel`. In `ReviewsViewController`, removed subscription for `StoresManagerDidUpdateDefaultSite` notifications to reload results controller and resync data.
- Added unit tests for `DefaultStoresManager` and `MainTabBarController`

### Logging in / switching stores

- Log out from the app
- Log in to an account with multiple sites, select a site to continue --> all the tabs should load as before
- Navigate to the Reviews tab --> the reviews should be loaded for the selected site
- Navigate to the My Store tab, tap Settings, then switch to the same store --> nothing should happen
- Tap Settings, then switch to a different store --> all the tabs should be updated to load data for the selected site

### Review push notifications

- Enable dev push notifications on sandbox p99K0U-1qz-p2
- Log in and navigate to a tab that is not Reviews
- Place the app in the background
- Submit a product review on the web
- Wait for the push notification
- Tap on the push notification
- Wait for a second --> the app should navigate to the Reviews tab and show the review details

Update release notes:


WooCommerce/Classes/ViewRelated/MainTabBarController.swift
```diff
@@ -351,6 +334,58 @@ extension MainTabBarController {
...
+            self.updateViewControllers(siteID: siteID)
+        }
+    }
+
+    func updateViewControllers(siteID: Int64?) {
+        guard let siteID = siteID else {
+            viewControllers = []
+            reviewsCoordinator = nil
+            return
+        }
```
Ecarrion: One thing I don't quite understand well is why we set up `viewControllers` to an empty array but still hold a reference to all root view controllers. Is this something that will change once you refactor the other VC's to accept a site ID?

WooCommerce/Classes/System/SessionManager.swift
```diff
@@ -21,10 +21,15 @@ extension NSNotification.Name {
     public static let applicationTerminating = Foundation.Notification.Name(rawValue: "ApplicationTerminating")
 }
 
+private extension UserDefaults {
+    @objc dynamic var defaultStoreID: Int {
+        integer(forKey: Key.defaultStoreID.rawValue)
```
Ecarrion, line 26: Not sure if possible, but can't we expose this as an observable so we don't have to deal with KVO observations? 😅

Ecarrion (commented):
Code looks good! I just have some questions to understand it correctly.
Testing also looks, good but I think I found a minor glitch(VC starts popping but then it changes suddenly) on the animation while logging out.
I think we didn't have any animation before but the glitch does look a bit weird 😅 

logout

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
jaclync: This PR only has the Reviews tab associated with a site ID, and the other tabs still rely on listening to the site change notification so I keep them around. After we finalize the approach for the Reviews tab, I'll do the same for other tab view controllers so that all of them will be set to `nil` here.

WooCommerce/Classes/System/SessionManager.swift
jaclync: Did you mean opening this up directly for whoever interested in the default store ID to subscribe? `UserDefaults` seems an implementation detail of `SessionManager`, and that's why I created a public observable `siteID: Observable<Int64?>` from `UserDefaults` KVO. Also, `UserDefaults` KVO has some limitation since it's ObjC API and doesn't work with `Int64` directly.

Not sure if I misunderstood anything, feel free to suggest another way!

jaclync:
> Testing also looks, good but I think I found a minor glitch(VC starts popping but then it changes suddenly) on the animation while logging out.
I think we didn't have any animation before but the glitch does look a bit weird 😅

Nice catch @Ecarrion! When the user logs out, we deauthenticate the app, then we pop the dashboard view controller's navigation controller to root animated:

https://github.com/woocommerce/woocommerce-ios/blob/ed8be1f7e79e975545fa384b64d832719d83495b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift#L379-L382

`ServiceLocator.stores.deauthenticate()` triggers the `state` to change from `AuthenticatedState` to `DeauthenticatedState`, and `DeauthenticatedState`'s `didEnter` event triggers the app delegate to display authentication UI:

https://github.com/woocommerce/woocommerce-ios/blob/ed8be1f7e79e975545fa384b64d832719d83495b/WooCommerce/Classes/Yosemite/DeauthenticatedState.swift#L10-L14

In the meantime, this PR also empties the main tab bar's `viewControllers`. I'm guessing this interferes with the `navigationController?.popToRootViewController(animated: true)` in `SettingsViewController`'s log out above and causes the weird animation.

In https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/3436ef3b61c44984c4cf77c6cd1feaa4877f8bd2, I removed the line to pop to root view controller animated and then turn on animation for displaying the authentication UI:

simulator

@Garance91540 does the animation after logging out look good to you? ☝️ 

This is what it looks like without animating the authentication UI:

simulator

Garance91540:
> @Garance91540 does the animation after logging out look good to you? ☝️

Yes! definitily add some smooth in there :)

WooCommerce/Classes/System/SessionManager.swift
Ecarrion: NVM, I think I misunderstood this 😇

Ecarrion:
@jackiejade I just test this and while the animation looks great I think we introduced another bug, but I'm not sure if it is by [not clearing the navigation controller](https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/3436ef3b61c44984c4cf77c6cd1feaa4877f8bd2#diff-56863e4b60e22cc16f786c5393dec589L381).

- Log in to a store in the app
- Navigate to orders and tap on one
- Log out
- Log in to a different store
- Navigate to orders and see that the order from the previous store is still visible

jaclync:
> I just test this and while the animation looks great I think we introduced another bug, but I'm not sure if it is by not clearing the navigation controller.
Log in to a store in the app
Navigate to orders and tap on one
Log out
Log in to a different store
Navigate to orders and see that the order from the previous store is still visible

@Ecarrion I think you mentioned another Jackie 😄 Thanks for testing and noticing this one, I was able to reproduce this issue in production as well. Also, this issue occurs after switching to another store without logging out as well. The code to [not pop the navigation controller to root](https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/3436ef3b61c44984c4cf77c6cd1feaa4877f8bd2#diff-56863e4b60e22cc16f786c5393dec589L381) just affects the dashboard tab (settings VC's navigation stack).

I just updated the PR with this commit https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/690c59e08fdca0b38dd4440cbb81ae83012ae852 to recreate all tab view controllers, without DI'ing the site ID to Dashboard/Orders/Products tabs. This should fix the issue above, and all tab view controllers now have life cycle tied to a logged in site. In the future, we can update the Dashboard/Orders/Products view controllers to DI site ID and remove their default site change notification observation. Lemme know what you think!

Ecarrion (approved):
👏 
Thanks, @jaclync for the changes. I tested the logout and switch store flow and it's working correctly ✅ 

The only thing I noticed was a strange notice message when changing store "700 products have been deleted from your store" which I assume it's because we are clearing the DB from the old store.  

Should I file a bug for that scenario or is that expected?

Ecarrion:
Also sorry for mentioning the wrong person 🤦

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
```diff
@@ -80,61 +80,26 @@ final class MainTabBarController: UITabBarController {
...
+    /// Tab view controllers
+    ///
+    private var dashboardViewController: UIViewController?
+    private var ordersTabViewController: UIViewController?
+    private var productsTabViewController: UIViewController?
+    private var reviewsTabCoordinator: Coordinator?
 
-        return navController
-    }()
+    private var cancellableSiteID: ObservationToken?
```
shiki, line 87: It's probably best to still add `cancellableSiteID.cancel()` on `deinit` even though `MainTabBarController` lives forever.

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
```diff
@@ -80,61 +80,26 @@ final class MainTabBarController: UITabBarController {
...
-                                                comment: "Title of the Products tab — plural form of Product")
-        navController.tabBarItem = UITabBarItem(title: navigationTitle,
-                                                image: UIImage.productImage,
-                                                tag: 0)
-        navController.tabBarItem.accessibilityIdentifier = "tab-bar-products-item"
+    /// Tab view controllers
+    ///
+    private var dashboardViewController: UIViewController?
+    private var ordersTabViewController: UIViewController?
+    private var productsTabViewController: UIViewController?
```
shiki: These properties are not used anywhere outside of `updateViewControllers`. It's probably best to remove them unless we're planning to use these later. 

```suggestion

```

The `reviewsTabCoordinator` is needed though. It needs to be kept in memory because it uses an observer too. Bummer.

WooCommerce/Classes/ViewRelated/Reviews/ReviewDetailsViewController.swift
```diff
@@ -45,23 +47,18 @@ final class ReviewDetailsViewController: UIViewController {
 
     /// Designated Initializer
     ///
-    init(productReview: ProductReview, product: Product?, notification: Note?) {
+    init(productReview: ProductReview, siteID: Int64, product: Product?, notification: Note?) {
```
shiki: I _think_ we can infer the `siteID` from `ProductReview.siteID`. We can keep the `siteID` property but we don't need to have the `init `argument. What do you think? 🙂

WooCommerce/Classes/ViewRelated/Reviews/ReviewDetailsViewController.swift
```diff
@@ -164,10 +161,6 @@ private extension ReviewDetailsViewController {
     /// Synchronizes the Notifications associated to the active WordPress.com account.
     ///
     func synchronizeReview(reviewID: Int64, onCompletion: @escaping () -> Void) {
-        guard let siteID = ServiceLocator.stores.sessionManager.defaultStoreID else {
```
shiki, line 167: Fantastic! 🤸

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
```diff
@@ -357,17 +346,9 @@ private extension ReviewsViewController {
     ///
     func startListeningToNotifications() {
         let nc = NotificationCenter.default
-        nc.addObserver(self, selector: #selector(defaultSiteWasUpdated), name: .StoresManagerDidUpdateDefaultSite, object: nil)
         nc.addObserver(self, selector: #selector(applicationDidBecomeActive), name: UIApplication.didBecomeActiveNotification, object: nil)
     }
 
-    /// Default Site Updated Handler
```
shiki, line 364: 👏

WooCommerce/WooCommerceTests/Internal/SessionManager+Internal.swift
```diff
@@ -16,7 +16,7 @@ extension SessionManager {
     /// Create an instance of unit testing.
     ///
     static func makeForTesting(authenticated: Bool = false) -> SessionManager {
-        var manager = SessionManager.testingInstance
+        let manager = SessionManager.testingInstance
```
shiki: ⚠️ We should create an instance now instead of using `testingInstance`. The effect of changing `SessionManager` from `struct` to `class` is that `SessionManager.testingInstance` is now **a singleton**. 🙈 That can become problematic since its `defaultCredentials` property is mutable. 

```suggestion
        let manager = SessionManager(defaults: SessionSettings.defaults, keychainServiceName: SessionSettings.keychainServiceName)
```

We should get rid of `testingInstance` later and test cases should use `makeForTesting()`.

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
```diff
@@ -0,0 +1,111 @@
...
+final class MainTabBarControllerTests: XCTestCase {
+    private var stores: StoresManager!
+    private var tabBarController: MainTabBarController!
+
+    override func setUp() {
+        super.setUp()
+        let mockAuthenticationManager = MockAuthenticationManager()
+        ServiceLocator.setAuthenticationManager(mockAuthenticationManager)
+        stores = DefaultStoresManager.testingInstance
+        ServiceLocator.setStores(stores)
```
shiki, line 14: The disadvantage of `ServiceLocator`. I can't tell when or where these are used. 🤪

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
```diff
@@ -0,0 +1,111 @@
...
+        // Action
+        let siteID: Int64 = 134
+        stores.updateDefaultStore(storeID: siteID)
+        stores.deauthenticate()
+
+        // Assert
+        XCTAssertEqual(tabBarController.viewControllers?.count, 0)
+    }
+
+    func test_tab_view_controllers_contain_a_different_reviews_tab_after_updating_to_a_different_site() throws {
```
shiki: Maybe: 

```suggestion
    func test_viewControllers_are_replaced_after_updating_to_a_different_site() throws {
```

Or something along those lines. Just wanted to highlight that it's not just “reviews” tab anymore. 😅

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    func test_tab_view_controllers_are_empty_after_deauthentication() {
+        // Arrange
+        // Trigger `viewDidLoad`
+        XCTAssertNotNil(tabBarController.view)
+
+        // Action
+        let siteID: Int64 = 134
+        stores.updateDefaultStore(storeID: siteID)
```
shiki: What do you think about adding an assertion for the previous state? 

```suggestion
        let siteID: Int64 = 134
        stores.updateDefaultStore(storeID: siteID)

        XCTAssertEqual(tabBarController.viewControllers?.count, 4)

        // Action
```

Maybe it will help with understanding more what changed.

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
```diff
@@ -0,0 +1,111 @@
+import TestKit
+import XCTest
+@testable import WooCommerce
+
+final class MainTabBarControllerTests: XCTestCase {
```
shiki, line 5: Thanks for adding tests! This is awesome! 👏

shiki (commented):
This is great! Thank you for working on this! 

✔️ Tested store switching
✔️ Tested logging out
✔️ Checked memory after logging out
✔️ Tested order and review push notifications

I just have a few suggestions and some celebratory comments. 😁 

## Objects Not Deallocated

These are separate problems and I'm glad that this PR helps us with catching these.

Looks like `DashboardViewController` is not deallocated after logging out. 

image

Also, and this is terrible. And obviously my mistake. I forgot to add a `weak self` to the `Publisher.sink` calls. So `OrderListViewController` instances are also not deallocated. 😅 

image

Would you mind creating issues for this after this PR is merged? I can tackle the OrderList one. 😄

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
jaclync: yea sounds good! updated in https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/5a9d65197d94f1343205132c6b5f2d20adbf6c17

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
jaclync: yup `reviewsTabCoordinator` is needed for push notification presentation! removed the rest in https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/5a9d65197d94f1343205132c6b5f2d20adbf6c17

WooCommerce/Classes/ViewRelated/Reviews/ReviewDetailsViewController.swift
jaclync: makes sense! updated in https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/d8f33ecce5c04540fc3f7cb16f6be691e1013920

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
jaclync: oops I forgot to update the test case name after replacing all tab view controllers! updated in https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/d8f33ecce5c04540fc3f7cb16f6be691e1013920

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
jaclync: yea good suggestion! updated in https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/d8f33ecce5c04540fc3f7cb16f6be691e1013920

WooCommerce/WooCommerceTests/Internal/SessionManager+Internal.swift
jaclync: ooh I actually thought it generates a new instance of in `.testingInstance` until I saw the `static`! I updated `makeForTesting()` to generate a new instance in https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/d8f33ecce5c04540fc3f7cb16f6be691e1013920, and created an issue for the replacement in https://github.com/woocommerce/woocommerce-ios/issues/2885

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
jaclync: yea, I'm not a fan of this 😞 however, `MainTabBarController` is initialized from `Main.storyboard` right now. It'd be nice to remove `Main.storyboard` at some point so we can DI to `MainTabBarController` more easily.

jaclync:
@shiki thanks for the review and noticing these retain issues after logout! I updated the PR and responded to your comments. Regarding the retain issues after logout:

> Looks like DashboardViewController is not deallocated after logging out.

This is very interesting! I took a look and I believe it's `MainTabBarController`'s `selectedViewController` holding onto `DashboardViewController` and I couldn't easily reset `selectedViewController`. I tried a few things and will fix this in a separate issue: https://github.com/woocommerce/woocommerce-ios/issues/2884

> Also, and this is terrible. And obviously my mistake. I forgot to add a weak self to the Publisher.sink calls. So OrderListViewController instances are also not deallocated. 😅

This is an easy fix, exactly what you said about the sink call! I fixed this in https://github.com/woocommerce/woocommerce-ios/pull/2850/commits/6d1ea95434499db18cc2ac7ea56efe0423944a18 if you don't mind 🙂

jaclync:
> The only thing I noticed was a strange notice message when changing store "700 products have been deleted from your store" which I assume it's because we are clearing the DB from the old store.
Should I file a bug for that scenario or is that expected?

@Ecarrion somehow I haven't seen the notice message after switching stores 🤔 I see the notice about site switch:



Do you happen to have a screenshot or steps to reproduce this "deleted" notice message?

jaclync:
I created another PR https://github.com/woocommerce/woocommerce-ios/pull/2886 to fix the `DashboardViewController` being retained after logging out. The PR updates the implementation in `MainTabBarController` a bit with authentication animation improvement (only animates when transitioning from logged in to logged out state). Since this PR has been reviewed and contains other changes like the site ID in Reviews tab classes, I'm planning to merge both PRs in the same release when they are ready. Lemme know if you have other suggestions!

Ecarrion:
> Do you happen to have a screenshot or steps to reproduce this "deleted" notice message?

I think I misread the message it says: "order xxx has been deleted from your store" 😛 

And it happens if you do the following
- Open the app
- Navigate to orders tab
- Open an order
- Tap on "my store" tab
- Change store
- See notice message

Screen Shot 2020-09-29 at 7 34 50 AM

shiki:
@Ecarrion That deletion message looks like https://github.com/woocommerce/woocommerce-ios/issues/2369. It's quite misleading and we should remove it soon. 🙈

shiki (approved):
👏

jaclync:
> the message it says: "order xxx has been deleted from your store" 😛

thanks for the steps to repro Ernesto and pointing to the existing issue #2369 Jayson! yea I agree it makes sense to remove the notice message, since we delete all the stored orders when the user pulls down to refresh on the orders tab as well and we currently don't have UI to manually delete an order.

merging now to unblock other open PRs!

PR #2848, opened 2020-09-21
[issue Refund] Add issue refund button

Ecarrion:
fixes #1487 

# Why

This would be the entry point for the refunds flow.

# How
- Added a new `IssueRefundTableViewCell` to represent the "issue refund button "
- Hide the button if the `.issueRefunds` feature flag is disabled or if the order is refunded.

# Screenshots

Screen Shot 2020-09-21 at 3 08 51 PM

# Testing Steps
- Create an order on the web-store
- Navigate to the order in the app and see that the "issue refund" button is visible.
- Do a full refund on web admin
- Refresh the order in the app and see that the "issue refund" button is not visible.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2857, opened 2020-09-22
Update Order to use OrderStatusEnum

Ecarrion:
# Why

While working on #2848 I noticed that the `Order` model was using a `String` to represent the order status. This PR updates the `Order` model to use the existing `OrderStatusEnum`

# Testing Steps
- Make sure tests pass.
- Do a smoke test around orders and updating orders to and verify that everything is working

Update release notes:


Networking/Networking/Remote/OrdersRemote.swift
```diff
@@ -116,9 +116,9 @@ public class OrdersRemote: Remote {
     ///     - status: New Status to be set.
     ///     - completion: Closure to be executed upon completion.
     ///
-    public func updateOrder(from siteID: Int64, orderID: Int64, statusKey: String, completion: @escaping (Order?, Error?) -> Void) {
+    public func updateOrder(from siteID: Int64, orderID: Int64, statusKey: OrderStatusEnum, completion: @escaping (Order?, Error?) -> Void) {
```
jaclync, line 119: nit: similar to the comment above, maybe `statusKey` could be renamed to `status` now that it is an enum?

Networking/Networking/Model/Order.swift
```diff
@@ -14,7 +14,7 @@ public struct Order: Decodable, GeneratedCopiable {
     ///
     /// Maps to `OrderStatus.slug`.
     ///
-    public let statusKey: String
+    public let statusKey: OrderStatusEnum
```
jaclync: nit: maybe `statusKey` could be renamed to `status` now that it is an enum? no worries if it involves lots of changes though!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -539,7 +539,7 @@ private extension OrderDetailsDataSource {
 // MARK: - Lookup orders and statuses
 extension OrderDetailsDataSource {
     func lookUpOrderStatus(for order: Order) -> OrderStatus? {
-        return currentSiteStatuses.filter({$0.slug == order.statusKey}).first
+        return currentSiteStatuses.filter({$0.status == order.statusKey}).first
```
jaclync: nice!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -348,9 +348,9 @@ extension OrderDetailsViewModel {
         let providerName = tracking.trackingProvider ?? ""
 
         ServiceLocator.analytics.track(.orderTrackingDelete, withProperties: ["id": orderID,
-                                                                         "status": statusKey,
-                                                                         "carrier": providerName,
-                                                                         "source": "order_detail"])
+                                                                              "status": statusKey,
```
jaclync: this might require the raw value from the enum? otherwise, I get a crash in Tracks:

```
[error] error: SQLCore dispatchRequest: exception handling request: <NSSQLSaveChangesRequestContext: 0x600002304840> , The data couldn’t be written because it isn’t in the correct format. with userInfo of {
    NSUnderlyingError = "Error Domain=NSCocoaErrorDomain Code=4866 \"The data couldn\U2019t be written because it isn\U2019t in the correct format.\" UserInfo={NSUnderlyingError=0x600001b48330 {Error Domain=NSCocoaErrorDomain Code=4864 \"This decoder will only decode classes that adopt NSSecureCoding. Class '__SwiftValue' does not adopt it.\" UserInfo={NSDebugDescription=This decoder will only decode classes that adopt NSSecureCoding. Class '__SwiftValue' does not adopt it.}}}";
}
```

```suggestion
                                                                              "status": statusKey.rawValue,
```

WooCommerce/Classes/ViewModels/Order Details/OrderEmailComposer.swift
```diff
@@ -11,8 +11,8 @@ final class OrderEmailComposer: NSObject, MFMailComposeViewControllerDelegate {
 
         displayEmailComposer(for: email, from: from)
         ServiceLocator.analytics.track(.orderContactAction, withProperties: ["id": order.orderID,
-                                                                        "status": order.statusKey,
-                                                                        "type": "email"])
+                                                                             "status": order.statusKey,
```
jaclync:
```suggestion
                                                                             "status": order.statusKey.rawValue,
```

jaclync (approved):
Thanks for refactoring the order status field! It definitely feels nicer using the enum in the app 👍 

Code: ✅ mostly the comments on the Tracks events
Testing: ✅ updating order status or viewing custom order statuses still working great!

Ecarrion:
Thanks, @jaclync for spotting those crashes! Also I've renamed all the possible `statusKey` to `status`!

PR #2859, opened 2020-09-23
Add Product: save as draft

jaclync:
Part of #2740 

## Background

This PR implements the "Save as draft" feature as in UI differences from editing a product (ref p91TBi-3f7-p2):
- Navigation bar > ellipsis menu:
  - Add "Save as draft" action to create a product with status "draft", regardless of the status in product settings
- Unsaved changes action sheet (when the user navigates away from product details with outstanding changes):
  - Add "Save as draft" action (same behavior as “Save as draft” above)

## Changes

- Added `func canSaveAsDraft() -> Bool` to `ProductFormViewModelProtocol` so that Add Product form's ellipsis menu shows "Save as draft" action
- In `ProductFormViewModelProtocol`, updated its `func updateProductRemotely(onCompletion: @escaping (Result<ProductModel, ProductUpdateError>) -> Void)` to include an optional status `status: ProductStatus?` to override in case the user chooses to save the product as draft
  - In `ProductFormViewModel` implementation, if the given `status` is non-nil the product is copied with the given status. Otherwise, it keeps using the original product.
- In `UIAlertController+Helpers`, added a different action sheet helper `presentDiscardNewProductActionSheet` for the discard changes action sheet plus "Save as draft" action
- In `ProductFormViewController`, created another save action `func saveProductAsDraft()` in addition to the existing `publishProduct` and `updateProduct`. this function passes the optional status down to view model's `saveProductRemotely`
- In `ProductFormViewController`, showed "Save as draft" action in the ellipsis menu if view model's `canSaveAsDraft` is true
- In `ProductFormViewController`'s `presentBackNavigationActionSheet`, showed the action sheet with "Save as draft" action
- Added unit tests for `ProductFormViewModel` to make sure it's saving the product with the expected status

## Testing

Please test this on an iOS 12 simulator because the back navigation event isn't triggered on iOS 13 simulators in debug build 😞

### Discard changes action sheet

- Go to the Products tab, tap on "+" CTA in the navigation bar, and tap "Simple product"
- Tap "<" in the navigation bar to leave the product form --> it should leave the screen without an action sheet
- Repeat the first step to create a product
- Fill in some details for the new product
- Tap "<" in the navigation bar to leave the product form --> it should show an action sheet with "Save as draft" action
- Tap on "Save as draft" --> the product should be published with "Draft" status remotely

### Discard changes action sheet

- Go to the Products tab, tap on "+" CTA in the navigation bar, and tap "Simple product"
- Fill in some details for the new product
- Tap "..." in the navigation bar --> there should be an "Save as draft" action
- Tap on "Save as draft" --> the product should be published with "Draft" status remotely

## Example screenshots

ellipsis menu | discard changes action sheet
-- | --
Simulator Screen Shot - iPhone Xs - 2020-09-23 at 12 00 22 | Simulator Screen Shot - iPhone Xs - 2020-09-23 at 12 00 18

Example screencast for saving as draft from the ellipsis menu:

simulator

Update release notes:


pmusolino:
Am I wrong, or draft products are part of milestone 5?
As far I know, inside the M4 I see only these stories:
- I want to create a new physical product.
- I want to create a new virtual product.
- I want to add a new grouped product.
- I want to add linked products to a product.

Garance91540 (approved):
👏

WooCommerce/Classes/Extensions/UIAlertController+Helpers.swift
```diff
@@ -29,6 +29,35 @@ extension UIAlertController {
...
+        }
+
+        actionSheet.addCancelActionWithTitle(ActionSheetStrings.cancel) { _ in
+            onCancel?()
+        }
+
+        if let popoverController = actionSheet.popoverPresentationController {
+            popoverController.sourceView = viewController.view
+            popoverController.sourceRect = viewController.view.bounds
+            popoverController.permittedArrowDirections = []
```
Ecarrion, line 55: [nit] Should we pass the source button so this looks better positioned on iPad?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -236,11 +240,18 @@ extension ProductFormViewModel {
...
 extension ProductFormViewModel {
-    func updateProductRemotely(onCompletion: @escaping (Result<EditableProductModel, ProductUpdateError>) -> Void) {
+    func saveProductRemotely(status: ProductStatus?, onCompletion: @escaping (Result<EditableProductModel, ProductUpdateError>) -> Void) {
+        let productModelToSave: EditableProductModel
+        if let status = status, status != product.status {
+            let productWithStatusUpdated = product.product.copy(statusKey: status.rawValue)
+            productModelToSave = EditableProductModel(product: productWithStatusUpdated)
+        } else {
+            productModelToSave = product
+        }
```
Ecarrion: nit: I'd like to write this kind of code as an inline function
```swift
let productModelToSave: EditableProductModel = {
    guard let status = status, status != product.status else {
        return product
    }
    let productWithStatusUpdated = product.product.copy(statusKey: status.rawValue)
    return EditableProductModel(product: productWithStatusUpdated)
}
```
So it's clear that the variables and code created are limited to the  `productModelToSave` scope

Ecarrion (commented):
Code ✅ 
Testing ✅ 

One thing I noticed is that when "saving as draft" the loading overlay says "Publishing your product"
To me that makes me feel that the product is being published live! Do you think we should change that copy when saving a draft or as a draft? 

cc @Garance91540

Ecarrion (approved):

jaclync:
> Am I wrong, or draft products are part of milestone 5?

@pmusolino from my understanding, "Save as draft" is part of product creation in M4 (p91TBi-3f7-p2). This way, the user might be more inclined to create a product without the fear of having to publish it immediately. We might not have "Save as draft" in design earlier, it was suggested by G in p91TBi-3f7-p2#comment-1650 🙂

WooCommerce/Classes/Extensions/UIAlertController+Helpers.swift
jaclync: Since the discard changes action sheet is triggered by tapping on the back navigation bar button or swipe-back gesture, we currently don't anchor the bottom sheet at a particular button and just show the popover in the middle of the view:

Simulator Screen Shot - iPad Air 2 - 2020-09-24 at 10 50 02

I tried passing `navigationItem.backBarButtonItem` to see if we can anchor the popover at the back button if it's triggered by the button tap action. However, I think `navigationItem.backBarButtonItem` is to be displayed on the next view controller and not the current one (the app crashes because it's not anchoring at a valid view).

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
jaclync: Nice it looks much better! Updated in https://github.com/woocommerce/woocommerce-ios/pull/2859/commits/22603684e69a1dd30c302619734d5608358e710a

jaclync:
> One thing I noticed is that when "saving as draft" the loading overlay says "Publishing your product"
To me that makes me feel that the product is being published live! Do you think we should change that copy when saving a draft or as a draft?

Good point! I asked in the async meeting channel CGPNUU63E/p1600923509001700-slack to continue the discussion 😄 

Also thanks for the suggestions, Ernesto! I responded to the comments and updated the PR. Merging this now!

PR #2862, opened 2020-09-23
Switches to an in-memory SQLite store which works better for testing

astralbodies:
The traditional in-memory store doesn't have all of the same functionality that SQLite does. In particular, it can't enforce certain constraints from the model like uniqueness and a few others. In 2018, [Apple mentioned](https://developer.apple.com/videos/play/wwdc2018/224/?time=1838) that the SQLite in-memory store option was available by specifying the file location of `/dev/null`.

I've updated the unit testing Core Data stack to use this type of store. All of the unit tests continue to pass!

Update release notes:

### DIFF
--- a/Yosemite/YosemiteTests/Mockups/MockupStorage.swift
+++ b/Yosemite/YosemiteTests/Mockups/MockupStorage.swift
@@ -115,7 +115,8 @@ extension MockupStorageManager {
     ///
     var storeDescription: NSPersistentStoreDescription {
         let description = NSPersistentStoreDescription()
-        description.type = NSInMemoryStoreType
+        description.type = NSSQLiteStoreType
+        description.url = URL(fileURLWithPath: "/dev/null")
         return description
     }
 }


shiki (approved):
TIL! 

A downside to this is that it is noticeably slower. 

### Using `InMemoryStoreType`

```
Test Suite 'All tests' passed at 2020-09-23 13:14:06.107.
	 Executed 455 tests, with 0 failures (0 unexpected) in 8.920 (9.111) seconds
```

### Using `SQLiteStoreType`

```
Test Suite 'All tests' passed at 2020-09-23 13:15:07.003.
	 Executed 455 tests, with 0 failures (0 unexpected) in 11.230 (11.461) seconds
```

But the benefits outweigh that downside. 👍 Thanks, Aaron!

astralbodies:
> A downside to this is that it is noticeably slower.

I hadn't considered that @shiki - thanks for looking at the timing!

I'll wait until the others get a chance to look at it. While the code change is minor, I wanted to make sure the fact I was changing it was apparent to everyone. 😄

shiki:
@astralbodies Looks like you got **1** 🎉 and **3** 🚀 . 

I guess you can now :shipit: 😄

jaclync:
Merging on behalf of @astralbodies 🚀

astralbodies:
Thanks again for merging this for me @jaclync!

PR #2865, opened 2020-09-23
Deletion of an existing Product

### LINKED ISSUE #2853: Ability to delete an existing Product
As part of Product Milestone 5, we should allow users to delete an existing product.

#### TODO:
- [x] Implement the network API request.
- [x] Implement the Yosemite layer for deleting a product (in reality it goes in your Trashed items, as [discussed here](https://a8c.slack.com/archives/C6H8C3G23/p1600696854038900))
- [x] Implement the action in the action sheet (to be showed only when the product is in the `editing` status)
- [x] Show the alert like on Android, asking the user for confirmation.

#### Screen:
Schermata 2020-09-22 alle 16 30 04

pmusolino:
Closes #2853 

## Description
This is the last PR, part of #2853, for enabling under the feature flag `editProductsRelease5` the deletion of a product.

## Changes
- Added `isEditProductsRelease5Enabled` param in `ProductFormViewModel`.
- Added `deleteProduct()` function in `ProductFormRemoteActionUseCase`.
- Implemented `canDeleteProduct()` protocol.
- Implemented the confirmation delete alert in `ProductFormViewController`.
- Updated and implemented new tests.  

### Case 1
1. Navigate to a product detail
2. Tap the more button "•••" item in the navigation bar. -> The new delete button should be one of the options.
3. Tap "Delete". -> An alert asking for confirmation would appear.
4. Tap "Move to Trash" -> The product should be deleted locally, and you will be popped to the product list.
5. Navigate to the wp-admin of your store. Open products, and click "Trash" between the options on top.
6. The product you just deleted should appear in the list.

### Case 2
1. Navigate to an order detail.
2. Tap on one of the products.
3. Tap the more button "•••" item in the navigation bar. -> The new delete button should be one of the options.
4. Tap "Delete". -> An alert asking for confirmation would appear.
5. Tap "Move to Trash" -> The product should be deleted locally, and you will be popped to the order detail.
6. Navigate to the wp-admin of your store. Open products, and click "Trash" between the options on top.
7. The product you just deleted should appear in the list.

## Screenshots
| Action sheet             |  Delete alert confirmation |
Simulator Screen Shot - iPhone 11 Pro - 2020-09-23 at 15 44 11 | Simulator Screen Shot - iPhone 11 Pro - 2020-09-23 at 15 52 34

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormRemoteActionUseCase.swift
```diff
@@ -102,6 +102,24 @@ final class ProductFormRemoteActionUseCase {
...
+    ///   - product: The product to be deleted remotely.
+    ///   - onCompletion: Called when the remote process finishes.
+    func deleteProduct(product: EditableProductModel, onCompletion: @escaping EditProductCompletion) {
+        deleteProductRemotely(product: product) { productResult in
+            switch productResult {
+            case .failure(let error):
+                // TODO: M5 analytics
+                onCompletion(.failure(error))
+            case .success(let product):
+                // `self` is retained because the use case is not usually strongly held.
```
jaclync: nit: it looks like `self` isn't used in this block 😄 so maybe this comment isn't needed anymore?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -236,6 +243,33 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
         present(alertController, animated: true)
     }
 
+    /// Product Confirmation Delete alert
+    ///
+    func presentProductConfirmationDeleteAlert(completion: @escaping (Bool) -> ()) {
```
jaclync: nit: just a suggestion to move this alert setup to like `UIAlertController+Helpers` or some other helper class since `ProductFormViewController` is already pretty huge?

jaclync: nit: maybe the `Bool` parameter can have an internal description like:
```suggestion
    func presentProductConfirmationDeleteAlert(completion: @escaping (_ isConfirmed: Bool) -> ()) {
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -620,6 +654,35 @@ private extension ProductFormViewController {
         SharingHelper.shareURL(url: url, title: product.name, from: view, in: self)
     }
 
+    func displayDeleteProduct() {
```
jaclync: nit suggestion:

```suggestion
    func displayDeleteProductAlert() {
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -620,6 +654,35 @@ private extension ProductFormViewController {
...
+                    // Presents the error alert.
+                    self.displayError(error: error)
+                case .success:
+                    // Pop to the previous view controller
+                    switch self.presentationStyle {
+                    case .navigationStack:
+                        self.navigationController?.popViewController(animated: true)
+                    default:
+                        self.dismiss(animated: true, completion: nil)
+                    }
```
jaclync: nit: maybe these could be moved to a separate function for potential reuse in the future whenever we want to dismiss the product form? then the comment can be replaced by the function name :)

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -128,6 +131,10 @@ extension ProductFormViewModel {
     func canShareProduct() -> Bool {
         formType == .edit
     }
+
+    func canDeleteProduct() -> Bool {
+        isEditProductsRelease5Enabled && originalProduct.product.productStatus == .publish && formType == .edit
```
jaclync: any reasons we don't allow product deletion for other product statuses like draft/pending/private/custom? I've got so many draft products that I personally want to delete (from testing M4 😆 )

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormRemoteActionUseCaseTests.swift
```diff
@@ -250,6 +250,46 @@ final class ProductFormRemoteActionUseCaseTests: XCTestCase {
...
+            result = aResult
+        }
+
+        // Assert
+        XCTAssertEqual(storesManager.receivedActions.count, 1)
+        XCTAssertNotNil(storesManager.receivedActions[0] as? ProductAction)
+        XCTAssertEqual(result, .success(ResultData(product: model, password: nil)))
+    }
+
+    func test_delete_product_returns_failure_result_with_product_error() {
```
jaclync: nit: just to be consistent with the previous test case name

```suggestion
    func test_deleting_product_returns_failure_result_with_product_error() {
```

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormViewModelTests.swift
```diff
@@ -103,6 +103,56 @@ final class ProductFormViewModelTests: XCTestCase {
...
+        let viewModel = createViewModel(product: product, formType: .add)
+
+        // Action
+        let canDeleteProduct = viewModel.canDeleteProduct()
+
+        // Assert
+        XCTAssertFalse(canDeleteProduct)
+    }
+
+    func test_edit_product_form_with_non_published_status_can_delete_product() {
```
jaclync, line 133: I think the test case name conflicts with the assert 😄 (related to a previous comment about deleting product for other product statuses)

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -195,6 +195,13 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
             }
         }
 
+        if viewModel.canDeleteProduct() {
+            actionSheet.addDefaultActionWithTitle(ActionSheetStrings.delete) { [weak self] _ in
```
jaclync: nit: maybe the "Delete" action could be destructive?
```suggestion
            actionSheet.addDestructiveActionWithTitle(ActionSheetStrings.delete) { [weak self] _ in
```

jaclync (commented):
The code looks good! I just have some questions about the deletion UX in the testing section -

Code: ✅ (mainly a question about supporting deletion for statuses that are not "published" and "trash")
Testing: ✅ ❓ 

- After trashing a product, I can still see the product in the Products tab. Maybe the predicate of the list could be updated to include `statusKey != 'trash'` here:

https://github.com/woocommerce/woocommerce-ios/blob/d42b886b808ce821261bf98cafda9928ac398e43/Yosemite/Yosemite/Tools/Products/ResultsController+FilterProducts.swift#L9

- This might involve design changes (cc @Garance91540): After trashing a product, I can still open the product from order details and update the product. However, updating the product remotely always returns an error (expected). Maybe we want to show an alert when tapping on a trashed product (status is 'trash')?

- This might also involve design changes (cc @Garance91540): If a trashed product is deleted permanently from core, we currently show a "Cannot load product" UI with a retry CTA when opening the product from order details. Since there isn't actually a way to retry/recover this product, the retry CTA feels quite misleading. Maybe we can show a different message or an alert in this case?




On the dev side, it looks like we get this API response when a product is deleted permanently and we can tell from the error code:

```
{
    "error": "woocommerce_rest_product_invalid_id",
    "message": "Invalid ID."
}
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
pmusolino: Make sense 👍 updated here c0119edeea4afab72a994635c392358766599597 (together with other nits)

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
pmusolino: Omg, You are right! Fixed

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormViewModelTests.swift
pmusolino: Fixed 👍

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
pmusolino: I'm not totally 100% sure about it, since it will differ from the original design proposed by @Garance91540, visible [here](https://github.com/woocommerce/woocommerce-ios/issues/2853).

This is the result of using the `Destructive` option.

pmusolino:
Thank you Jaclyn for your review!

> After trashing a product, I can still see the product in the Products tab. Maybe the predicate of the list could be updated to include statusKey != 'trash' here:

This is super strange because I'm not able to replicate this behavior. 🤔 When I try to delete a product, it will be immediately trashed, and also refreshing the list doesn't show to me the product. What are the steps you follow?

> This might involve design changes (cc @Garance91540): After trashing a product, I can still open the product from order details and update the product. However, updating the product remotely always returns an error (expected). Maybe we want to show an alert when tapping on a trashed product (status is 'trash')?

Nice catch. I didn't expect that I can't update a product inside the trash, since it's just a product in another status.
@jaclync your idea is to show an alert when a user opens a trashed product, or you want to show a notice (which disappear after a few seconds)?

> This might also involve design changes (cc @Garance91540): If a trashed product is deleted permanently from core, we currently show a "Cannot load product" UI with a retry CTA when opening the product from order details. Since there isn't actually a way to retry/recover this product, the retry CTA feels quite misleading. Maybe we can show a different message or an alert in this case?

I agree with showing a different message in this case. What do you think @Garance91540?

pmusolino:
> This might also involve design changes (cc @Garance91540): If a trashed product is deleted permanently from core, we currently show a "Cannot load product" UI with a retry CTA when opening the product from order details. Since there isn't actually a way to retry/recover this product, the retry CTA feels quite misleading. Maybe we can show a different message or an alert in this case?

This will be no more true because of this new task https://github.com/woocommerce/woocommerce-ios/issues/2869 👍

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: We usually use the destructive for deleting/removing something from the site, like deleting a tracking number from order details or "Discard changes" in the product form discard changes action sheet. Maybe "Product Settings" can be moved up so that "Delete" is at the bottom @Garance91540 ?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
jaclync: not sure if it makes sense to move the alert logic to a helper since `ProductFormViewController` is already 1200 LOC? 🙂

jaclync:
> This is super strange because I'm not able to replicate this behavior. 🤔 When I try to delete a product, it will be immediately trashed, and also refreshing the list doesn't show to me the product. What are the steps you follow?

Interesting, I can't reproduce anymore either, sorry for the confusion 😅 since we delete the trashed product in Yosemite, it should just disappear from the list 👍

jaclync:
I was testing the PR at a coffee shop with slower internet, and it took some time for the product details to go away after trashing the product:

simulator

If it takes a bit longer or I'm fast enough, I can probably even edit the product and update it again. @Garance91540 do you think we might want to show some waiting UI while we wait for the API request to go through?

jaclync (approved):
Code looks good 👍 just a suggestion to move the alert setup code away from `ProductFormViewController`. We can see what G thinks about the design questions!

Garance91540:
Thanks @pmusolino for all these changes and @jaclync for the thorough review!

> We usually use the destructive for deleting/removing something from the site, like deleting a tracking number from order details or "Discard changes" in the product form discard changes action sheet. Maybe "Product Settings" can be moved up so that "Delete" is at the bottom @Garance91540 ?

@jaclync this is a good point, I agree with your suggestion and updated the designs in Figma.

> This might also involve design changes (cc @Garance91540): If a trashed product is deleted permanently from core, we currently show a "Cannot load product" UI with a retry CTA when opening the product from order details. Since there isn't actually a way to retry/recover this product, the retry CTA feels quite misleading. Maybe we can show a different message or an alert in this case?

mmm this one is annoying. is there no way to open the product as a read only? As a merchant I can see how I could delete a product from my store because I'll stock making it for example, but still want to see the product details attached to an existing order. 

> If it takes a bit longer or I'm fast enough, I can probably even edit the product and update it again. @Garance91540 do you think we might want to show some waiting UI while we wait for the API request to go through?

Absolutely: added a screen here: https://d.pr/i/kfCIJQ

Did I answer all design questions?

Garance91540 (approved):

pmusolino:
> @jaclync this is a good point, I agree with your suggestion and updated the designs in Figma.

Perfect, I'll move the delete action to be the last one of the list (like in other action sheets).

> mmm this one is annoying. is there no way to open the product as a read only? As a merchant I can see how I could delete a product from my store because I'll stock making it for example, but still want to see the product details attached to an existing order.

Since the product is deleted permanently, the only way to do this is by opening the read only product screen from the Order model. I'll work on that in another PR. Opened an issue here https://github.com/woocommerce/woocommerce-ios/issues/2898

> Absolutely: added a screen here: https://d.pr/i/kfCIJQ

💯 beautiful


> Did I answer all design questions?

Yes, thanks Garance 😄 👍

Issue #2864, opened 2020-09-23
Rake Generate Fails


Ecarrion:
I'm getting the following error when running `rake generate` to update copiable models.

> while reading .yml 'CodeGeneration/Networking-Copiable.sourcery.yaml'. 'The element XCSwiftPackageProductDependency is not supported'

Update:
- Tried updating Sourcery to v1.0 with no luck.

shiki:
@woocommerce/ios-developers This is failing because Sourcery is using a very old Xcode project parser. This started happening when I added the TestKit Swift Package. 

For now, what we can do is:

1. Temporarily remove the TestKit references from the WooCommerceTests, YosemiteTests, NetworkingTests, and StorageTests targets.

    ![image](https://user-images.githubusercontent.com/198826/94032868-a3426c00-fd7d-11ea-824e-b3a18f2fa84b.png)

2. Run` rake generate` 

That should work. And then revert the Xcode project changes. Sorry for the trouble! 

Hopefully, Sourcery will fix this on their end soon.

Ecarrion:
@jaclync Ha! yes. I just saw this issue: https://github.com/krzysztofzablocki/Sourcery/issues/816

PR #2866, opened 2020-09-23
Updates copiable docs to include SPM temporary fix

Ecarrion:
Relates to: https://github.com/woocommerce/woocommerce-ios/issues/2864

# Why

This PR updates our docs to point to the temporary when generating copiable methods.


shiki (approved):
Thank you!

PR #2867, opened 2020-09-23
Yosemite: Log Fatal Errors to Sentry

shiki:
This is supposed to help with diagnosing #2858. We should also do the same for Storage (#2364) and Networking (#2372).   

## Changes

I added a new function, `logErrorAndExit()` which uses `DDLogError` and `fatalError` internally. The `DDLogError` will be sent to the Encrypted Logs. The `fatalError()` will be sent to Sentry with just “Fatal error” like before. But now, the underlying message will be accessible in the Encrypted Logs. Read more about Encrypted Logging at paaHJt-1ms-p2. 

https://github.com/woocommerce/woocommerce-ios/blob/5732be3f36caa0358edc33d265bc31110dd7a773/Yosemite/Yosemite/Internal/LogErrorAndExit.swift#L9-L12

## Testing

1. Use the WooCommerce Alpha scheme. 
2. Turn the `force-crash-logging` launch argument on

    

3. Add a breakpoint on [this line](https://github.com/Automattic/Automattic-Tracks-iOS/blob/5f96537a62a8b7d3eb6635cb4dd8489dd39fc62a/Automattic-Tracks-iOS/Event%20Logging/EventLogging%2BSentry.swift#L33) of the Tracks `EventLogging+Sentry.swift` file. 
4. Change [this line](https://github.com/woocommerce/woocommerce-ios/blob/5732be3f36caa0358edc33d265bc31110dd7a773/Yosemite/Yosemite/Model/Extensions/OrderStatsV4Interval%2BDate.swift#L6) of the `OrderStatsV4Internal+Date` file to this so that it will crash later:

    ```swift
        guard let date = createDateFormatter(timeZone: timeZone).date(from: "abc") else {
    ```
5. Run and install the app on the simulator. 
6. Log in. The app will crash.
7. Stop debugging. 
8. Open the app on the simulator again. Not from Xcode. Open it a few times. This is to make sure that the crash will be queued. 
9. Revert the change in Step 4. 
10. Run the app from Xcode again. 
11. The app should stop in the breakpoint in Step 3. 
12. Capture and save the value of `logFile.uuid`. 
13. Let the app finish running so it will update the crash to Sentry. 
14. Head to the Encrypted Logs Console. 
15. Enter the `UUID` you captured in step 12 and click on View.
16. Confirm that you see the error message like this:

    

## Submitter Checklist

### DIFF
--- a/WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce Alpha.xcscheme
+++ b/WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce Alpha.xcscheme
@@ -70,6 +70,12 @@
             ReferencedContainer = "container:WooCommerce.xcodeproj">
          </BuildableReference>
       </BuildableProductRunnable>
+      <CommandLineArguments>
+         <CommandLineArgument
+            argument = "-force-crash-logging 1"
+            isEnabled = "NO">
+         </CommandLineArgument>
+      </CommandLineArguments>
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Release"
--- a/Yosemite/Yosemite.xcodeproj/project.pbxproj
+++ b/Yosemite/Yosemite.xcodeproj/project.pbxproj
@@ -101,6 +101,7 @@
 		45ED4F16239E939A004F1BE3 /* TaxClassStoreTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45ED4F15239E939A004F1BE3 /* TaxClassStoreTests.swift */; };
 		570B05CF246B6AAD00C186AE /* RetrieveProductReviewFromNoteUseCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 570B05CE246B6AAD00C186AE /* RetrieveProductReviewFromNoteUseCase.swift */; };
 		57150E1124F462D900E81611 /* TestKit in Frameworks */ = {isa = PBXBuildFile; productRef = 57150E1024F462D900E81611 /* TestKit */; };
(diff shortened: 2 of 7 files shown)


jaclync (commented):
Just an initial question: I think it'd be nice for the message to show in Sentry directly, without having to find it from Encrypted Logs Console for each crash event. Not sure if there's a way to log the crash on Sentry without having to call `fatalError` to exit the app which creates a duplicate crash on Sentry?

shiki:
@jaclync My concern with that is we would end up with different error messages like the [CoreDataManager Cannot Backup Store errors](https://sentry.io/share/issue/cec3ecfc89bc4cf8abbfc07901b87bb4/) which causes Sentry to consider them as separate crashes. 

But it looks like this `fatalError` issue is going to be fixed in Sentry soon: https://github.com/getsentry/sentry-cocoa/issues/662. 

This `logErrorAndExit` may just be a temporary fix that we'll delete later. What do you think about keeping this for now until it's resolved in Sentry? Or would you still prefer to have duplicate issues, one with no message, and the other with the message? Looks like that what this person did: https://github.com/getsentry/sentry-cocoa/issues/662#issuecomment-675016062.

jaclync:
> What do you think about keeping this for now until it's resolved in Sentry? Or would you still prefer to have duplicate issues, one with no message, and the other with the message? Looks like that what this person did: getsentry/sentry-cocoa#662 (comment).

Keeping this `logErrorAndExit` sounds great! It doesn't look like there's a timeline for the Sentry fix, and I didn't really like seeing two entries of the same crash on Sentry (it's confusing for anyone not familiar with the crash like for CTrO from the Android team). Testing now!

jaclync (approved):
Nice work! :shipit:

Code: ✅ (I was wondering where `-force-crash-logging` comes from, and found it by searching in the code base)
Testing: ✅ (I also generated an installable build on a testing branch and got to see the [event on Sentry](https://sentry.io/organizations/a8c/issues/1912794968/?project=1458804&query=is%3Aunresolved) and log in Encrypted logs console as well!)

shiki:
Thanks, Jaclyn!

PR #2873, opened 2020-09-25
Show stock status in variation list

### LINKED ISSUE #2863: Products: Show stock status in variations list for variable products
### Problem

Right now, if I manage stock at the variations level for a variable product, there are only two places where I see the stock status:

* Products list
* Variation details

If a variation is out of stock the product will show that status in the Products list, but I can't tell which variation(s) is out of stock without opening each one to check its individual stock status.

### Enhancement

As @Garance91540 suggested:

> I think surfacing the “out of stock”/”low stock”/”1 left in stock” labels in the Variation list in a similar way as we do for the Product list would be the best way!

### Screenshots (current behavior)

Products list|Product details|Variations list|Variation details
-|-|-|-
image|image|image|image

Internal ref: p5T066-1xg-p2#comment-5892

jaclync:
Fixes #2863 

## Changes

- Moved `func createStockText() -> String` from a private `Product` extension to `ProductFormDataModel` protocol extension (which `EditableProductModel` and `EditableProductVariationModel` implement)
  - Updated the return value from an optional string to non-optional, since it's already returning non-nil value for all cases
- Updated `ProductsTabProductViewModel`'s initializer for `Product` to use `EditableProductModel` and `ProductFormDataModel`'s extensions
- Updated `ProductsTabProductViewModel`'s initializer for `EditableProductVariationModel` from just a status/price to `"\(stock status) • \(the original status or price)"`

## Testing

- Go to the Products tab --> each product cell should look the same as before
- Tap on a variable product
- Tap on the variations row --> each variation cell should show the stock status first

## Example screenshots

\ | before | after
-- | -- | --
variation list | Simulator Screen Shot - iPhone 11 - 2020-09-25 at 11 12 00 | Simulator Screen Shot - iPhone 11 - 2020-09-25 at 11 14 15
product list (regression testing) | Simulator Screen Shot - iPhone 11 - 2020-09-25 at 11 11 51 | Simulator Screen Shot - iPhone 11 - 2020-09-25 at 11 14 08

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductsTabProductViewModel+ProductVariation.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductsTabProductViewModel+ProductVariation.swift
@@ -15,6 +15,28 @@ extension ProductsTabProductViewModel {
 
 private extension EditableProductVariationModel {
     func createDetailsAttributedString(currencySettings: CurrencySettings) -> NSAttributedString {
+        let stockStatusAttributedString = createStockStatusAttributedString()
+        let variationStatusOrPriceAttributedString = createVariationStatusOrPriceAttributedString(currencySettings: currencySettings)
+
+        let detailsAttributedString = NSMutableAttributedString(attributedString: stockStatusAttributedString)
+        detailsAttributedString.append(NSAttributedString(string: " • ", attributes: [
+            .foregroundColor: UIColor.textSubtle,
+            .font: Style.detailsFont
+        ]))
+        detailsAttributedString.append(variationStatusOrPriceAttributedString)
+        return NSAttributedString(attributedString: detailsAttributedString)
+    }
+
+    func createStockStatusAttributedString() -> NSAttributedString {
+        let stockText = createStockText()
+        return NSAttributedString(string: stockText,
+                                  attributes: [
+                                    .foregroundColor: UIColor.textSubtle,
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/ProductsTabProductViewModel+ProductVariation.swift
```diff
@@ -46,6 +68,12 @@ private extension EditableProductVariationModel {
     }
 }
 
+private extension EditableProductVariationModel {
+    enum Style {
+        static let detailsFont = StyleManager.footerLabelFont
```
Ecarrion, line 73: Newbie question: Why did you choose to create an `enum` instead of using `StyleManager.footerLabelFont` directly?

Ecarrion (approved):
Looks good! 🚀

Garance91540 (approved):
Thanks for the quick fix!

WooCommerce/Classes/ViewRelated/Products/Variations/ProductsTabProductViewModel+ProductVariation.swift
jaclync: this is just to make sure all the parts in the details attributed strings use the same font except for the parts with custom color! in case one day we decide to change the font for the stock status text but forget to also update other parts of the same font (e.g. the " • " separator)

PR #2876, opened 2020-09-25
Add Unit Test To Help Prevent Core Data Migration Mistakes

shiki:
## Problem

I'm about to add another migration to fix the Xcode 12 build. And from my experience, I always felt uneasy because I'm never sure if I'm doing things right or I forgot something. 

As @astralbodies also mentioned, when we create a new model but don't make any structural changes, we should always make sure to change the Hash Modifier value. If we don't, then this will pass and `CoreDataManager` will just _skip_ the migration:

https://github.com/woocommerce/woocommerce-ios/blob/365f2c7c64c1cdf74097e5913c2b934ffdf9606e/Storage/Storage/CoreData/CoreDataManager.swift#L188-L191

That can lead to unknown consequences. 

This fact is also mentioned in this [Core Data documentation](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreDataVersioning/Articles/vmUnderstandingVersions.html). 

> There may, however, be some situations in which you have two versions of a model that Core Data would normally treat as equivalent that you want to be recognized as being different. For example, you might change the name of the class used to represent an entity, or more subtly you might keep the model the same but change the internal format of an attribute such as a BLOB—this is irrelevant to Core Data, but it is crucial for the integrity of your data. To support this, Core Data allows you to set a hash modifier for an entity or property see versionHashModifier (NSEntityDescription) and versionHashModifier (NSPropertyDescription).

## Solution

This PR adds an automated test that ensures that all the Core Data model versions are **not compatible with each other**. This helps prevent us from creating new Core Data model versions that `CoreDataManager.migrateDataModelIfNecessary()` will just ignore. 

### Other Changes

I organized some of the `NSManagedObjectModel` creation to use [`ManagedObjectModelsInventory`](https://github.com/woocommerce/woocommerce-ios/blob/add/unit-test-for-core-data-migrations/Storage/Storage/CoreData/ManagedObjectModelsInventory.swift) so there is less repetition in how we read these objects. 

https://github.com/woocommerce/woocommerce-ios/blob/be33b0678990493cc4b9f8d2a7ec39d2b58443aa/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L291-L294

## Testing

1. Run the `test_all_model_versions_are_not_compatible_with_each_other` unit test. We expect this to pass.
2. Add a new Model Version to `WooCommerce.xcdatamodeld`. Use the last model version as the base. This is probably named “Model 32”.
3. In “Model 32”, change the class name of the `Attribute` entity to `GenericAttribute`. 
4. Set “Model 32” as the current model version. 
5. Run the `test_all_model_versions_are_not_compatible_with_each_other` again. It should fail now. This is the “feature” of this unit test. 
6. Back to “Model 32”, add a value for the Hash Modifier.

    image

7. Run the `test_all_model_versions_are_not_compatible_with_each_other` again. It should pass. 

Run `git reset --hard HEAD` to discard all the changes above. 

## Submitter Checklist

### DIFF
--- a/Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
+++ b/Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
@@ -81,6 +81,11 @@ struct ManagedObjectModelsInventory {
                                             versions: modelVersions)
     }
 
+    /// Load the corresponding `NSManagedObjectModel` for the given `version`.
+    ///
+    /// This is intentionally not part of `ModelVersion` itself because this involves a file
+    /// access and we usually would not need all of the `NSManagedObjectModel` instances.
+    ///
     func model(for version: ModelVersion) -> NSManagedObjectModel? {
         let expectedMomURL = packageURL.appendingPathComponent(version.name).appendingPathExtension("mom")
         return NSManagedObjectModel(contentsOf: expectedMomURL)
--- a/Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
+++ b/Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
@@ -80,6 +80,11 @@ struct ManagedObjectModelsInventory {
                                             currentModel: currentModel,
                                             versions: modelVersions)
     }
+
+    func model(for version: ModelVersion) -> NSManagedObjectModel? {
+        let expectedMomURL = packageURL.appendingPathComponent(version.name).appendingPathExtension("mom")
+        return NSManagedObjectModel(contentsOf: expectedMomURL)
+    }
 }
 
 // MARK: - Utils
--- a/Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
+++ b/Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
@@ -26,7 +26,7 @@ struct ManagedObjectModelsInventory {
     /// This is intentionally a `struct` with a single property instead of a `String` because I
     /// foresee that this will be used to contain the `NSManagedObjectModel` in the near future.
     ///
-    struct ModelVersion {
+    struct ModelVersion: Equatable {
         /// The name excluding the extension.
         ///
         /// For example, if the model file name is "Model 10.mom", then this would be "Model 10".
--- a/Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
+++ b/Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
@@ -27,7 +27,7 @@ final class CoreDataIterativeMigratorTests: XCTestCase {
             .appendingPathComponent(UUID().uuidString)?
             .appendingPathExtension("sqlite"))
 
-        try modelsInventory.versions.forEach { currentVersion in
+        for (currentVersionIndex, currentVersion) in modelsInventory.versions.enumerated() {
             // Given
             let currentModel = try XCTUnwrap(modelsInventory.model(for: currentVersion))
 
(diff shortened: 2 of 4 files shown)


Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
```diff
@@ -80,6 +80,16 @@ struct ManagedObjectModelsInventory {
...
                                             versions: modelVersions)
     }
+
+    /// Load the corresponding `NSManagedObjectModel` for the given `version`.
+    ///
+    /// This is intentionally not part of `ModelVersion` itself because this involves a file
+    /// access and we usually would not need all of the `NSManagedObjectModel` instances.
+    ///
+    func model(for version: ModelVersion) -> NSManagedObjectModel? {
+        let expectedMomURL = packageURL.appendingPathComponent(version.name).appendingPathExtension("mom")
```
jaclync, line 90: just curious, when do we use a URL with extension `mom` vs. `momd` like in the deleted `func urlForModel(name: String, in directory: String?) -> URL?`?

jaclync (approved):
Thanks for taking the time to ensure the migration runs smoothly! Somehow I can't repro the test failure, but feel free to check out [my testing branch](https://github.com/woocommerce/woocommerce-ios/compare/add/unit-test-for-core-data-migrations...jaclync/unit-test-for-core-data-migrations-testing?expand=1) and merge it! (One commit adds and sets the current version to model version 32 without any changes, and the second commit renames `Attribute` to `GenericAttribute`)

TIL'ed that we have to make sure the version hash is different, and a simple renaming wouldn't change the hash 😮

Storage/Storage/CoreData/ManagedObjectModelsInventory.swift
astralbodies: I'm pulling this from my brain, not a verified source so I might be a little wrong. Using the `momd` URL will look at the package of models and will return the current model version only. There's a plist that gets compiled in with those files pointing at the current model version. Therefore, `mom` files are referring to a specific model .

Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
```diff
@@ -7,10 +7,10 @@ import CoreData
 final class CoreDataIterativeMigratorTests: XCTestCase {
     private var modelsInventory: ManagedObjectModelsInventory!
 
-    override func setUp() {
-        super.setUp()
+    override func setUpWithError() throws {
```
astralbodies, line 10: I really am happy with this new pattern allowing setUp and tearDown to have failures and they're presented properly. 😄

Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
```diff
@@ -19,6 +19,85 @@ final class CoreDataIterativeMigratorTests: XCTestCase {
...
+    /// 1. Creating a new model version for `WooCommerce.xcdatamodeld`, copying the latest version.
+    /// 2. Running this test.
+    ///
+    /// And then make this pass again by setting a Hash Modifier value for the new model.
+    ///
+    func test_all_model_versions_are_not_compatible_with_each_other() throws {
+        // Given
+        // Use in-memory type or else this would be too slow.
+        let storeType = NSInMemoryStoreType
+        let storeURL = try XCTUnwrap(NSURL(fileURLWithPath: NSTemporaryDirectory())
```
astralbodies, line 50: ❤️ `XCTUnwrap`

shiki:
Thank you for the reviews!

---

> Somehow I can't repro the test failure

@jaclync I'm so sorry. The testing steps were not clear. I meant renaming just the "Class Name". In your branch, both the Entity Name and Class Name were renamed. When I reverted the Entity Name change, the test failed. 

image

I'll merge this now but please do try it again if you get some time. 🙂

jaclync:
> I meant renaming just the "Class Name". In your branch, both the Entity Name and Class Name were renamed. When I reverted the Entity Name change, the test failed.

Ooh yea reverting the entity name makes the test case fail! I renamed from the Entities list on the left panel, which updates both the entity and class name 😆

Issue #2757, opened 2020-09-03
Add/Edit Products M5: Downloadable files Step 2.1 - Mark/unmark a product as downloadable


pmusolino:
As part of the M5 release, we need to provide the option to change the `is_downloadable` flag of a product by adding a checkbox "Product Downloadable" to the product settings:

<img width="300" alt="iOS-product-settings" src="https://user-images.githubusercontent.com/495617/92255555-cd1d1700-eed2-11ea-8f56-4e88abcaec0a.png">

PR #2872, opened 2020-09-24
Issue/2756 update downloadable file info for a product - part 1

partho-maple:
Fixes: #2756 
Part 2: https://github.com/woocommerce/woocommerce-ios/pull/2877

## Description
This PR allows updating the downloadable files info with their details. This PR implements the following capabilities.
- Edit individual downloadable files info in a new scene, when tapped from the downloadable file list table view cell

## Changes
- Add `ProductDownloadFileViewController` and related files to show individual downloadable files' details to edit when tappen from `ProductDownloadListViewController`s' `tableView` cell 
- Encodes downloadable files related properties in Product.swift file
- dropSessionDidUpdate method updated to handle datasource modification
- Handle validation on product update

## Test Cases
Test cases have been included with this PR

## Screenshots

Screen Shot 2020-09-25 at 1 51 12 AM
Screen Shot 2020-09-25 at 1 51 38 AM

#### Prerequisites
- Create a downloadable product from wp-admin
- Add a few downloadable files from there
- Then navigate to the downloadable files' list scene on the app

#### Case 1 
- Tap on any downloadable file's cell from the list which will navigate to the details scene.
- Check the scene navigation bar title is properly updated with the filename
- Write an invalid URL
- Notice the `Done` button is disabled
- Add a different valid URL
- Notice the `Done` button is enabled
- Now add the previous original URL, and notice the `Done` button is disabled again

#### Case 2
- Do the same changes as `Case 1`, but this time for Filename.
- `Done` button should be disabled when the file name is empty or the same as the original name.
- Notice the downloadable files count to reflect the changes

#### Case 3
- Go to the same screen like the above 2 cases and make some valid changes, press done
- Navigate back to the product details screen and press update.
- Verify that the downloadable file information had been properly updated.
- Also verify from the wp-admin portal

#### Case 4
- Navigate to the "Downloadable Files List" scene and reorder files, press `Done`
- Navigate back to the product details screen and press update.
- Verify that the downloadable file information had been properly updated.
- Also verify from the wp-admin portal

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2878, opened 2020-09-26
Issue/2757 mark unmark a product as downloadable

partho-maple:
Fixes: #2757 
Depends on: https://github.com/woocommerce/woocommerce-ios/pull/2872 for its actually remote API call implementation.

## Description
This PR allows the capability for a product to mark/unmark it as downloadable, from `Product Settings` on product details

## Changes
- Add `downloadable ` property to `ProductSettings` model
- Add a new row for downloadable product settings on `ProductSettingsRows `
- Add `isEditProductsRelease5Enabled` feature flag of `Product Settings` related places

## Test Cases
Test cases have been included with this PR

## Screenshots
Screen Shot 2020-09-26 at 12 47 36 PM

#### Case 1 
- From the `Products` tab, select any `Simple` product, and show details.
- Tap on the top right navigation bar button >> "Product Settings"
- You should be able to see a new switchable row, titled `Downloadable Product`
- Make any changes on the downloadable switch and update the product.
- Come back again to this page to verify that your changes are reflected properly (Currently the remote API depends on https://github.com/woocommerce/woocommerce-ios/pull/2872 . So that PR needs to be merged for the API to work properly]

#### Case 2 
- From the same `Products` tab, select a `non-simple` product
- Go to the product settings and the Downloadable Product row shouldn't be visible

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
@@ -55,18 +55,19 @@ private extension AddProductCoordinator {
         let currency = ServiceLocator.currencySettings.symbol(from: currencyCode)
         let productImageActionHandler = ProductImageActionHandler(siteID: product.siteID,
                                                                   product: model)
+        let isEditProductsRelease5Enabled = ServiceLocator.featureFlagService.isFeatureFlagEnabled(.editProductsRelease5)
         let viewModel = ProductFormViewModel(product: model,
                                              formType: .add,
                                              productImageActionHandler: productImageActionHandler,
                                              isEditProductsRelease3Enabled: true,
-                                             isEditProductsRelease5Enabled: ServiceLocator.featureFlagService.isFeatureFlagEnabled(.editProductsRelease5))
+                                             isEditProductsRelease5Enabled: isEditProductsRelease5Enabled)
         let viewController = ProductFormViewController(viewModel: viewModel,
                                                        eventLogger: ProductFormEventLogger(),
                                                        productImageActionHandler: productImageActionHandler,
                                                        currency: currency,
                                                        presentationStyle: .navigationStack,
                                                        isEditProductsRelease3Enabled: true,
-                                                       isEditProductsRelease5Enabled: true)
+                                                       isEditProductsRelease5Enabled: isEditProductsRelease5Enabled)
         // Since the Add Product UI could hold local changes, disables the bottom bar (tab bar) to simplify app states.
         viewController.hidesBottomBarWhenPushed = true
         navigationController.pushViewController(viewController, animated: true)
--- a/WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
@@ -65,7 +65,8 @@ private extension AddProductCoordinator {
                                                        productImageActionHandler: productImageActionHandler,
                                                        currency: currency,
                                                        presentationStyle: .navigationStack,
-                                                       isEditProductsRelease3Enabled: true)
+                                                       isEditProductsRelease3Enabled: true,
+                                                       isEditProductsRelease5Enabled: true)
         // Since the Add Product UI could hold local changes, disables the bottom bar (tab bar) to simplify app states.
         viewController.hidesBottomBarWhenPushed = true
         navigationController.pushViewController(viewController, animated: true)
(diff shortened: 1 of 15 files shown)


WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
```diff
@@ -65,7 +65,8 @@ private extension AddProductCoordinator {
                                                        productImageActionHandler: productImageActionHandler,
                                                        currency: currency,
                                                        presentationStyle: .navigationStack,
-                                                       isEditProductsRelease3Enabled: true)
+                                                       isEditProductsRelease3Enabled: true,
+                                                       isEditProductsRelease5Enabled: true)
```
pmusolino: Please, pay attention to these details, or we risk to enable a feature that is still a work in progress to our users. This should be set to `ServiceLocator.featureFlagService.isFeatureFlagEnabled(.editProductsRelease5)` 👍

WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/ProductSettingsSections.swift
```diff
@@ -20,12 +20,13 @@ enum ProductSettingsSections {
 
         let rows: [ProductSettingsRowMediator]
 
-        init(_ settings: ProductSettings, productType: ProductType, isEditProductsRelease3Enabled: Bool) {
-            if isEditProductsRelease3Enabled && productType == .simple {
+        init(_ settings: ProductSettings, productType: ProductType, isEditProductsRelease3Enabled: Bool, isEditProductsRelease5Enabled: Bool) {
+            if (isEditProductsRelease3Enabled || isEditProductsRelease5Enabled) && productType == .simple {
```
pmusolino: I'd say that this condition is wrong. We don't want to show the downloadable product row if only the product release 3 flag is enabled.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -305,7 +305,8 @@ extension ProductVariationsViewController: UITableViewDelegate {
                                                            productImageActionHandler: productImageActionHandler,
                                                            currency: currency,
                                                            presentationStyle: .navigationStack,
-                                                           isEditProductsRelease3Enabled: isEditProductsRelease3Enabled)
+                                                           isEditProductsRelease3Enabled: isEditProductsRelease3Enabled,
+                                                           isEditProductsRelease5Enabled: false)
```
pmusolino: Why is `false` here? 🤔  The edit product release 5 will not include only downloadable files.

pmusolino (changes requested):
The case 1 is not testable, since it depends on another PR 😅 

In that case, what I do for implementing a PR like this is: 
1) create a branch from the main PR (in this case, #2872)
2) Open the PR, but the PR target will be the #2872 branch. In this way, we can see only the changes applied to this PR but we can still use the features implemented in the main branch.
3) When this PR is reviewed and approved, we have to attend that #2872 is merged into our working branch, then we can change the target branch for this PR to the working branch and merge it.

So I'm going to review only the code and to test the second use case since the first it's not testable until it will be merged in `feat/products-downloadable-files`.

WooCommerce/Classes/ViewRelated/Products/Add Product/AddProductCoordinator.swift
partho-maple: I am so sorry. I didn't know that. I thought `true ` was OK since the other one was `true `
Updated accordingly

WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/ProductSettingsSections.swift
partho-maple: Hi @pmusolino , I think I didn't quite get your point. Can you please elaborate more on when do we want to show the downloadable product row and when we don't

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
partho-maple: Product variation doesn't have any downloadable file that's why I thought it would be false. But yes, you are right.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/ProductSettingsSections.swift
partho-maple: Updated the logic. Is it OK now?

Issue #279, opened 2018-09-04
StorageType+Woo: siteID is part of the pK!


jleandroperez:
### Details:
`StorageType+Woo` implements a set of methods used by  the OrderStore. We need to extend those methods, so that the **siteID** is also considered as part of the Order's primary key.

PR #2879, opened 2020-09-28
SIWA: fix crash shortly after app launch when the Apple ID credential is revoked

### LINKED ISSUE #2830: SIWA: app crashes shortly after app launch when the Apple ID credential is revoked

jaclync:
Fixes #2830

## Changes

There are 2 major changes in this PR:

- Crash fix (https://github.com/woocommerce/woocommerce-ios/commit/256d04cbc648f44d8008f676e32004e0116e77f9) in `CoreDataManager`: instead of destroying and adding another persistent store, it now just resets and **deletes all stored objects** in the main context.
  - Unit tests in https://github.com/woocommerce/woocommerce-ios/commit/275e4a1d5e7a1a818c87eeb7b590132bc6446fff. Unfortunately, I wasn't able to reproduce the crash in the last test case. In real life, the main context's `reset` and child context's `save` most likely happen at the same time when the crash occurs (it's hard to make two operations happening exactly the same time since code is sequential).
- To minimize the stored objects **after** the logout, this commit https://github.com/woocommerce/woocommerce-ios/commit/1b9829840a6cadf342862db2d833286ce47797c7 updates `Remote`'s networking completion blocks to use `weak self` so that the network requests are not retained after a Yosemite store is deinitialized on logout. Now a `Remote` and a `Store` have the same life cycle (unless it's retained by one of the closures). While updating the `Remote` usage, I also added some `weak self` to some completion blocks.
  - Note that this still does not eliminate the stored objects being saved after logging out. I noticed that with the test scenario 1, `OrderCount` and `OrderCountItem` are often the only tables with leftover data. It's possible that because `OrderCount`/`OrderCountItems` are requested on app launch in `MainTabViewModel` which is a singleton throughout the app lifetime.

## Testing

Note on inspecting the database:

- In Xcode, go to Window > Devices and Simulators
- In the selected device that is used for testing, scroll to "Installed apps" and click on Woo with the identifier `com.automattic.woocommerce` (to differentiate from the alpha build)
- At the bottom of the "Installed apps" section, click on the ⚙️ icon and then "Download Container..."
- After the download is complete, right click on the downloaded file and select "Show Package Contents"
- In AppData > Documents, you'd find the database `WooCommerce.sqlite`
- Open the database file in your desired inspector (I use DataGrip, @shiki mentioned https://sqlitebrowser.org/ and https://betamagic.nl/products/coredatalab.html)

### Scenario 1: app launch after Apple ID credential is revoked

Please test on an iOS 13+ device for SIWA.

- Terminate the app
- Go to iOS Settings and disconnect your Apple account (Settings > Apple ID > Password & Security > Apps using Apple ID > Remove WordPress app)
- Relaunch the app –> the app should not crash, and the database should contain minimal data (I've been seeing `OrderCount` and `OrderCountItem` in the database consistently, while all the other tables are empty). On `develop`, this results in a crash very frequently

### Scenario 2: logging out right after refreshing Orders tab

Thanks to @shiki for discovering this scenario where lots of Order data are still around after just the first commit!

- Set device networking link conditioner to “Very bad network”
- Navigate to the Orders tab
- Pull to refresh
- Quickly log out –> the app should not crash, and the database should contain minimal data (I've been seeing zero data in all the tables after https://github.com/woocommerce/woocommerce-ios/commit/1b9829840a6cadf342862db2d833286ce47797c7)

Update release notes:


Yosemite/Yosemite/Stores/SettingStore.swift
```diff
@@ -6,11 +6,19 @@ import Storage
 // MARK: - SettingStore
 //
 public class SettingStore: Store {
+    private let remote: SiteSettingsRemote
```
shiki: It might be better to call this `siteSettingsRemote` since there are 2 `Remote` properties in here now.

shiki (approved):
This looks good to me! 

As discussed, we should also prioritize implementing `siteID` keys (required) for Orders so that even if we make a mistake of forgetting `weak self` in the future, users would still not see data from different users and/or sites. https://github.com/woocommerce/woocommerce-ios/issues/279

I'm not sure if there are other entities that don't have `siteID`?

shiki:
Oh, we should probably add an `[internal]` note in `RELEASE-NOTES.txt` to mention the change in database behavior.

Yosemite/Yosemite/Stores/SettingStore.swift
jaclync: yup! updated in https://github.com/woocommerce/woocommerce-ios/pull/2879/commits/3fd77b03869f2adc5b19f448d2355362c00ccce5

jaclync:
> As discussed, we should also prioritize implementing siteID keys (required) for Orders so that even if we make a mistake of forgetting weak self in the future, users would still not see data from different users and/or sites. #279
I'm not sure if there are other entities that don't have siteID?

I believe for entities that we don't delete after switching stores already require a siteID in the predicate, `Order`s are deleted after switching stores (`OrderAction.resetStoredOrders`). I'll do a pass to make sure all `ResultsController`s are setting `siteID` in the predicate now that the database might have little left over data, or like you said in case of us forgetting to weak self in remote completion blocks.

Also thanks for pointing out #279 - I'll work on this and have this PR depend on that!

Issue #2880, opened 2020-09-28
Fix Xcode 12 build failure


shiki:

asbajaj001:
I am also getting error on 11.5 which is a recommended version on readme file. I kept getting issue in xcode 12 and kept trying for two days and just stumbled upon this issue.

The build (feels like) is randomly failing. e.g., latest issue it is giving me is: Swift Compiler Error "Type 'ApiCredentials' has no member 'loggingEncryptionKey'" in EventLogging.swift AND "Command CompileSwift failed with a nonzero exit code" in two files viz. Ghost+Woo.swift and WooConstants.swift.

I am not trying to update the code because I want the code to first run as-is as downloaded from official github.

Any pointers or ideas would be appreciated.

shiki:
Hi @asbajaj001! Sorry about that. It looks like the template used for generating `ApiCredentials.swift` was missing a property. I have submitted a PR to fix this https://github.com/woocommerce/woocommerce-ios/pull/2926. 

But you should be able to fix it by changing your `ApiCredentials.swift` file to the contents of this [`ApiCredentials-Template.swift`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Credentials/Templates/ApiCredentials-Template.swift) file. You'll have to re-enter the Client ID and Client Secret after that. 

Please let me know how it goes!

asbajaj001:
Hi @shiki thanks for your quick help. I replaced content of ApiCredentials.swift in woocommerce/derivedsources folder and then I am trying to view canvas but nothing appears for swift files. Attaching a screenshot. What am I doing wrong?

Thanks!
![Uploading Screen Shot 2020-10-05 at 7.10.11 PM.png…]()

shiki:
Hello again @asbajaj001! I got an email that you responded to this issue but it's not on Github anymore. I guess it was deleted. 🙂 Let me know if you still need help!

asbajaj001:
@shiki thanks for following up. I deleted my previous response because I confused it with something else. As for this issue, I went with 5.0 release and it worked fine. Thanks for your awesomeness.

asbajaj001:
@shiki I downloaded 5.1 release version and trying to rebuild but getting the same error again. I followed the steps but it is failing continuously with message: Type 'ApiCredentials' has no member 'loggingEncryptionKey'

shiki:
[...]

PR #2881, opened 2020-09-28
Upgrade WPAuth from 1.24.0 to 1.26.0-beta.12

shiki:
Ref #2880. 

This is part of fixing the build errors when compiling with Xcode 12. This PR fixes this error in WPKit:

WordPressKit error
 
That error was fixed in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/280. And WPKit was subsequently updated in WPAuth in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/475. 

## Changes in WPAuth

Here is a [diff between 1.24.0 and 1.26.0-beta.12](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/compare/1.24.0...1.26.0-beta.12). I don't see anything that might cause a regression. But please double-check. 

### Xcode 12

1. Compile with Xcode 12.
2. Confirm that the build error above is no longer shown. 

There are still some more build errors that will be taken care of in upcoming PRs.

### Test for Regression

Using Xcode 11, please test different methods of logging in like:

- Log in with Google
- SIWA
- Log in with site address
- Log in with WP

## Submitter Checklist

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -32,7 +32,7 @@ target 'WooCommerce' do
   pod 'Gridicons', '~> 1.0'
 
   # To allow pod to pick up beta versions use -beta. E.g., 1.1.7-beta.1
-  pod 'WordPressAuthenticator', '~> 1.24.0'
+  pod 'WordPressAuthenticator', '~> 1.26.0-beta'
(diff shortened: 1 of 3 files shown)


Ecarrion (approved):
🚀

shiki:
Thank you very much, @Ecarrion!

PR #2883, opened 2020-09-28
Rename Storage.Attribute to GenericAttribute to fix Xcode 12 build errors

### LINKED ISSUE #2880: Fix Xcode 12 build failure

shiki:
Closes #2880. 

🙇  Sorry for the 500+ lines. This adds a new Core Data model version and a mapping model, both totaling 2600+ lines. 

This is the final PR to fix the build errors when compiling with Xcode 12. There are warnings that remain but they can be fixed later. 

## Problem

The last remaining error is the duplicate `Attribute` definitions: 

## Solutions Considered

I thought I could just easily fix this by renaming this `@objc` declaration and hardcoding the `Attribute.entityName`:

https://github.com/woocommerce/woocommerce-ios/blob/e10545c0e516bc460d620825889535548578ddea/Storage/Storage/Model/GenericAttribute%2BCoreDataClass.swift#L14

But that turned out to be more complex because of [the way `entityName` is declared](https://github.com/woocommerce/woocommerce-ios/blob/e10545c0e516bc460d620825889535548578ddea/Storage/Storage/Extensions/NSManagedObject%2BObject.swift#L14-L20). It would look like I'd introduce overrides and some architectural changes. Also, having a different class name (`Attribute`) and a different Core Data entity name (`GenericAttribute`) would probably cost us a lot of confusion in the future. 

## Solution

So, to fix the error, I followed what @pmusolino suggested and renamed the `Attribute` class to `GenericAttribute`, both the Swift class name and the Core Data entity name. 

### Model 32

image

The renaming in the new model, “Model 32”, was straightforward. I just did a normal rename. 😄 

### Mapping Model

It turns out that renaming an entity would not make Core Data automatically keep the data during migration. 🙈 To fix this, I added a new mapping model. And then I added a new mapping between `Attribute` and `GenericAttribute`. This was not generated by Xcode. 

image

But that wasn't enough. Because we also have to fix all the relationships using `Attribute`. I questioned my life choices at this point. 😃 It turned out that I just had to declare the mapping name for the relationship to keep the data:

image

### Migration Unit Test 

And then I added a [unit test for this migration](https://github.com/woocommerce/woocommerce-ios/blob/e10545c0e516bc460d620825889535548578ddea/Storage/StorageTests/CoreData/MigrationTests.swift#L51) that proves that:

- The entity was renamed.
- The existing data was kept. 

### Migration

1. Log in using `develop` and load some data.
2. Log in using this branch. Confirm that there are no failed migrations logged in the console. 
3. Confirm that you can load products in the Products tab. 

### Xcode 12

1. Build the project using Xcode 12.
2. Confirm that the build is successful. The warnings are expected. 

## Submitter Checklist


Storage/Storage/Model/ProductVariation+CoreDataProperties.swift
```diff
@@ -69,28 +69,28 @@ extension ProductVariation {
...
-    @NSManaged public func replaceAttributes(at indexes: NSIndexSet, with values: [Attribute])
+    @NSManaged public func replaceAttributes(at indexes: NSIndexSet, with values: [GenericAttribute])
 
     @objc(addAttributesObject:)
-    @NSManaged public func addToAttributes(_ value: Attribute)
+    @NSManaged public func addToAttributes(_ value: GenericAttribute)
 
     @objc(removeAttributesObject:)
-    @NSManaged public func removeFromAttributes(_ value: Attribute)
+    @NSManaged public func removeFromAttributes(_ value: GenericAttribute)
```
pmusolino, line 93: I think that these methods needs to be re-regenerated since they should be called:
- `insertIntoGenericAttributes`
- `replaceGenericAttributes`
- `addToGenericAttributes`
and so on.

Storage/Storage/Model/GenericAttribute+CoreDataClass.swift
```diff
@@ -0,0 +1,17 @@
...
+/// A generic Attribute class with 3 core properties: `id`, `key`, and `value`.
+///
+/// This is currently only used as the corresponding storage for `ProductVariationAttribute` but
+/// we might be able to use it for others later.
+///
+/// ## Naming
+///
+/// This would have been named as just “Attribute” but Xcode 12 build issues prevents us from
+/// doing so because there's an existing unrelated Attribute class in Aztec. ¯\_(ツ)_/¯
+///
```
pmusolino, line 13: Thanks for documenting it 💯

Storage/StorageTests/CoreData/MigrationTests.swift
```diff
@@ -0,0 +1,191 @@
...
+/// ## Notes
+///
+/// In general, we should avoid using the entity classes like `Product` or `Order`. These classes
+/// may **change** in the future. And if they do, the migration tests would have to be changed.
+/// There's a risk that the migration tests would no longer be correct if this happens.
+///
+/// That said, it is understandable that we are sometimes under pressure to finish features that
+/// this may not be economical.
+///
+final class MigrationTests: XCTestCase {
```
pmusolino, line 20: What is the main difference between `MigrationTests` and `CoreDataIterativeMigratorTests`? They seem to accomplish the same purpose

pmusolino (commented):
Thank you for handling this @shiki! 

There is a thing that is not clear to me: is there a reason why you used a new model (model 32) instead of model 31, which is still a work in progress for release 5.2.0.0? Since the model 31 was still a work in progress, I think that the changes can be handled inside this storage model. What do you think?

Storage/Storage/Model/ProductVariation+CoreDataProperties.swift
shiki: @pmusolino I generated the code and it didn't change this part at all. 😅 It looks like this is actually correct because the relationship name is still `attributes`. 

image

It's only the type that changed. 

Anyway, I added a simple test that one of the methods here still works. 62b7e2b778b95c44ecccb2485a453108c90d2a32

Storage/StorageTests/CoreData/MigrationTests.swift
shiki: I realized that some of the `CoreDataIterativeMigratorTests` are not related to each other. Some test the functionality of the `CoreDataIterativeMigrator` itself while some are testing the actual model changes. This `MigrationTests` is for testing the model changes. In #2903, I moved the other tests to `MigrationTests` and simplified them.

shiki (commented):
@pmusolino I responded to your inline comments. This should be ready for review again. 🙂

Storage/Storage/Model/ProductVariation+CoreDataProperties.swift
pmusolino: Oh great! 👍

pmusolino (approved):
Xcode 12 is coming 🤩 thanks for handling this @shiki!

shiki:
Thank you, Paolo!

PR #2886, opened 2020-09-29
Fix `DashboardViewController` retained after logging out

### LINKED ISSUE #2884: `DashboardViewController` is retained after logging out
This is a followup task from #1838 where we replaced all the tab view controllers after logging out or site ID changes. However, after logging out, all the tab view controllers were deallocated except for `DashboardViewController` via Xcode memory debugger.

After some debugging, I believe it is `MainTabBarController`'s `selectedViewController` still pointing to `DashboardViewController` and thus retaining it. I've tried a few things:
- Setting `MainTabBarController`'s `selectedViewController` to `nil`: this crashes the app
- Setting `MainTabBarController`'s `selectedIndex` to `NSNotFound` or some value that is not 0: this has no effect
- Setting `MainTabBarController`'s `viewControllers` to contain a single blank navigation controller: this successfully deallocates `DashboardViewController`, however there's a visual side effect where a blank screen appears for a split second because `DashboardViewController` is gone before the authenticator UI appears.

I'm planning to go with the last option and also fix the blank screen by only resetting `MainTabBarController`'s `viewControllers` to a single blank navigation controller **only when the authentication presentation completes** instead of when site ID becomes `nil`.

jaclync:
Fixes #2884 
Based on https://github.com/woocommerce/woocommerce-ios/pull/2850

- [x] ⚠️ Update `Podfile` to point `WordPressAuthenticator` to production version after https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/488 is merged and deployed ⚠️ 

## The challenge of resetting `MainTabBarController`'s `viewControllers`

The last tab (`DashboardViewController`) before logging out is retained by `MainTabBarController` because once a `UITabBarController`'s `selectedViewController` has been set, it cannot be set to nil or it crashes the app. I tried replacing `MainTabBarController`'s `viewControllers` to a single placeholder navigation controller and set it to its `selectedViewController`, but `DashboardViewController` is still retained because of an internal strong reference `_selectedViewControllerDuringWillAppear` 😨 My guess is because `DashboardViewController` is still present when `MainTabBarController` presents the authentication UI, and I could not find a way to fix this internal strong reference nor find much information about it online.

Screen Shot 2020-09-29 at 11 41 50 AM

In order to fix this, I think it'd better to keep `MainTabBarController`'s `viewControllers` the same throughout its lifetime. In this PR, I updated the original implementation (mutating `viewControllers` after site ID changes) to replace each tab navigation controller's `viewControllers` instead. This deallocates all the tab root view controllers after logging out.

In `MainTabBarController`, its `viewControllers` are now fixed to the same `UINavigationController`s and we update each navigation controller's `viewControllers` when site ID changes.

Note: I noticed that `StoreStatsV4PeriodViewController`s are still lingering after `DashboardViewController` is deallocated. It looks like it's from `ChartYAxis` from the charting library. I created a separate issue https://github.com/woocommerce/woocommerce-ios/issues/2887 to track this.

Screen Shot 2020-09-29 at 4 32 27 PM

### Tear down the tab root view controllers after authentication UI is presented

When the user logs out, it triggers `AppDelegate`'s `displayAuthenticator` to show the authentication UI. However, now that the dashboard tab is torn down and deallocated on `nil` site ID (when `displayAuthenticator` is triggered), the blank dashboard tab is shown for a split second before the authentication is presented. Before this PR, this issue doesn't occur because `DashboardViewController` was still retained. To fix this UI issue, I made a PR https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/488 to add a completion block for authentication UI presentation completion and tear down the tab root view controllers there.

### Authentication UI presentation animation

Another UI issue I noticed from https://github.com/woocommerce/woocommerce-ios/pull/2850 is when launching the app in logged out state where the authentication UI is presented from a black screen (since the tab root view controllers aren't set up when logged out). The desired behavior is to animate the authentication UI presentation only after logging out from the app (logged in --> logged out). To fix this, I added an `animated: Bool` parameter to allow animation only when transitioning from logged in to logged out state (`DeauthenticatedState`'s `didEnter`).

## Future tasks

(I created an issue for these two https://github.com/woocommerce/woocommerce-ios/issues/2888)

- I'd like to refactor the app navigation to a coordinator (maybe not a subclass of `Coordinator` because the main window is using a `UITabBarController` instead of a `UINavigationController`). I have some WIP code for this but thought it's better leaving them for a separate task.
- I noticed a glitch for store picker flow when the app is previously logged in but the user hasn't selected a site, and we present a store picker on launch. Because the main tab bar isn't set up yet since the app is still logged out, the store picker is presented from a tab bar with a black screen. I tried updating its presentation to try presenting from the tab bar controller, but `StorePickerCoordinator` currently takes in a `UINavigationController` only and it uses the tab bar's `selecteViewController` right now 🤨 (feels a bit hacky to me to grab a `UINavigationController` from the tab bar's selected tab) I think this is not a major flow, but I'd like to fix it sometime soon.

### Logging in / switching stores

- Log out from the app
- Log in to an account with multiple sites, select a site to continue and navigate to each tab --> all the tabs should load as before
- Navigate to the My Store tab, tap Settings, then switch to the same store --> nothing should happen
- Tap Settings, then switch to a different store --> all the tabs should be updated to load data for the selected site
- Log out from the app --> in Xcode memory debugger, all tab root view controllers should be deallocated (before this PR, `DashboardViewController` stays around after logging out)

### Confidence check for Review push notifications

- Enable dev push notifications on sandbox p99K0U-1qz-p2
- Log in and navigate to a tab that is not Reviews
- Place the app in the background
- Submit a product review on the web
- Wait for the push notification
- Tap on the push notification
- Wait for a second --> the app should navigate to the Reviews tab and show the review details

Update release notes:

### DIFF
--- a/WooCommerce/Classes/AppDelegate.swift
+++ b/WooCommerce/Classes/AppDelegate.swift
@@ -354,22 +354,24 @@ extension AppDelegate {
 
     /// Whenever there is no default WordPress.com Account, let's display the Authentication UI.
     ///
-    func displayAuthenticatorIfNeeded() {
+    private func displayAuthenticatorIfNeeded() {
         guard ServiceLocator.stores.isAuthenticated == false else {
             return
         }
 
-        displayAuthenticator()
+        displayAuthenticator(animated: false)
     }
 
     /// Displays the WordPress.com Authentication UI.
     ///
-    func displayAuthenticator() {
+    func displayAuthenticator(animated: Bool) {
         guard let rootViewController = window?.rootViewController else {
             fatalError()
         }
 
-        ServiceLocator.authenticationManager.displayAuthentication(from: rootViewController)
+        ServiceLocator.authenticationManager.displayAuthentication(from: rootViewController, animated: animated) { [weak self] in
+            self?.tabBarController?.removeViewControllers()
+        }
     }
 
     /// Whenever the app is authenticated but there is no Default StoreID: Let's display the Store Picker.
@@ -388,13 +390,13 @@ extension AppDelegate {
         storePickerCoordinator = StorePickerCoordinator(navigationController, config: .standard)
         storePickerCoordinator?.start()
         storePickerCoordinator?.onDismiss = { [weak self] in
-            self?.displayAuthenticator()
+            self?.displayAuthenticator(animated: false)
         }
     }
 
     /// Whenever we're in an Authenticated state, let's Sync all of the WC-Y entities.
     ///
-    func synchronizeEntitiesIfPossible() {
+    private func synchronizeEntitiesIfPossible() {
         guard ServiceLocator.stores.isAuthenticated else {
             return
         }
--- a/Podfile
+++ b/Podfile
@@ -32,9 +32,9 @@ target 'WooCommerce' do
   pod 'Gridicons', '~> 1.0'
 
   # To allow pod to pick up beta versions use -beta. E.g., 1.1.7-beta.1
-  pod 'WordPressAuthenticator', '~> 1.24.0'
+  # pod 'WordPressAuthenticator', '~> 1.24.0'
   # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :commit => ''
-  # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => ''
+  pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => 'wcios-2884/authenticator-presentation-completion'
   # pod 'WordPressAuthenticator', :path => '../WordPressAuthenticator-iOS'
 
   pod 'WordPressShared', '~> 1.11'
--- a/Podfile
+++ b/Podfile
@@ -32,9 +32,9 @@ target 'WooCommerce' do
   pod 'Gridicons', '~> 1.0'
 
   # To allow pod to pick up beta versions use -beta. E.g., 1.1.7-beta.1
-  # pod 'WordPressAuthenticator', '~> 1.24.0'
+  pod 'WordPressAuthenticator', '~> 1.26.0-beta'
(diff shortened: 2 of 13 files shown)


WooCommerce/Classes/AppDelegate.swift
```diff
@@ -354,22 +354,24 @@ extension AppDelegate {
...
     ///
-    func displayAuthenticator() {
+    func displayAuthenticator(animated: Bool) {
         guard let rootViewController = window?.rootViewController else {
             fatalError()
         }
 
-        ServiceLocator.authenticationManager.displayAuthentication(from: rootViewController)
+        ServiceLocator.authenticationManager.displayAuthentication(from: rootViewController, animated: animated) { [weak self] in
+            self?.tabBarController?.removeViewControllers()
```
shiki, line 374: It looks like we can achieve the same effect by calling `removeViewControllers` inside `MainTabBarController` instead. 

```swift
    func observeSiteIDForViewControllers() {
        cancellableSiteID = stores.siteID.subscribe { [weak self] siteID in
            guard let self = self else {
                return
            }
            self.updateViewControllers(siteID: siteID)

            if siteID == nil {
                self.removeViewControllers()
            }
        }
    }
```

But we end up with the black screen that you mentioned. It's not nice. 😅

shiki (approved):
Looks good to me! Thank you again for being so thorough with this! 

I did not get to test the push notification because Push V2 changed the spec. I did follow the code using breakpoints and it looks like push notifications will be handled fine. 🤞 

I'm loving this view controller destruction. So much easier to understand! 😄

WooCommerce/Classes/AppDelegate.swift
jaclync: yeaa I'd much prefer just responding to site ID changes, but the authentication UI presentation doesn't happen fast enough even without animation 😞

jaclync:
Thank you for the review & testing with breakpoints, Jayson! I tested the review push notification when I created the PR on a device. Merging https://github.com/woocommerce/woocommerce-ios/pull/2850 and this PR now to make the site ID PR easier!

PR #831, opened 2019-04-02
Product Details: Update ProductStore to handle multiple sites better

bummytime:
This is a quick fix for a situation I overlooked in my last Product Details PR — we need to take into account the `siteID` when upserting/pruning `ProductAttribute`, `ProductDefaultAttribute`, `ProductImage`, `ProductCategory`, and `ProductTag` entities. To do that, this PR updates the `StorageType+Extensions.swift` loaders and `ProductStore`'s `handleProduct*()` funcs accordingly.

Ref: #279 

## Testing
* Review the updated code
* Verify the unit tests are ✅ 

@mindgraffiti or @ctarda 1st :shipit: wins! 😄

PR #2889, opened 2020-09-29
Add site ID to `Order`, `OrderCoupon`, `OrderRefundCondensed`, and `ShippingLine` in storage lookup

### LINKED ISSUE #279: StorageType+Woo: siteID is part of the pK!

jaclync:
Fixes #279

There are not many details in the issue description #279, so I referred to a previous PR that linked to the issue https://github.com/woocommerce/woocommerce-ios/pull/831.

## Changes

- In `StorageType` extensions, added site ID to the lookup functions for `Order`, `OrderCoupon`, `OrderRefundCondensed`, and `ShippingLine`
- Updated its usage in `OrdersUpsertUseCase`
- Updated unit tests, and added a test case `test_it_persists_order_relationships_in_storage`
- Fixed a build warning from `develop`: https://github.com/woocommerce/woocommerce-ios/commit/104bd005387854592534b50980e7da75d38760ed

## Testing

Please do a confidence check on the Orders tab - especially refunded orders, order's shipping info, and orders with at least a coupon!

Here are screenshots of an order with coupon, and an order with refund info and shipping info:

order with a coupon | order with refund and shipping info
-- | --
Simulator Screen Shot - iPhone 11 - 2020-09-29 at 22 18 56 | Simulator Screen Shot - iPhone 11 - 2020-09-29 at 22 19 22

Update release notes:

### DIFF
--- a/Storage/Storage/Tools/StorageType+Extensions.swift
+++ b/Storage/Storage/Tools/StorageType+Extensions.swift
@@ -30,8 +30,8 @@ public extension StorageType {
 
     /// Retrieves the Stored Order.
     ///
-    func loadOrder(orderID: Int64) -> Order? {
-        let predicate = NSPredicate(format: "orderID = %ld", orderID)
+    func loadOrder(siteID: Int64, orderID: Int64) -> Order? {
+        let predicate = NSPredicate(format: "orderID = %ld AND siteID = %ld", orderID, siteID)
         return firstObject(ofType: Order.self, matching: predicate)
     }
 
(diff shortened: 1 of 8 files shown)


Ecarrion (approved):
I'm not sure if there was a problem but changes and tests look good! 🚀

jaclync:
Thanks for the review Ernesto!

We don't currently have a (noticeable) problem because we delete all stored orders on site change. Since another PR https://github.com/woocommerce/woocommerce-ios/pull/2879 stops resetting the persistent store to fix a crash from race condition saving a child context in a background thread, there might be leftover data that still get saved after logging out. This PR is to make sure we are loading an order and handling its relationships with a site ID, in case there could be order objects around from another site.

PR #2891, opened 2020-09-29
Enabled Products Milestone 3 by default 

### LINKED ISSUE #2815: Remove the M3 feature flag
The M3 features have been released as a beta in a couple of releases now and we're now ready to make them available for all users. We need to:

- [ ] Disable all M3 feature flag checks
- [ ] Hide the beta switch from the settings

pmusolino:
Fixes #2815 

## Description
In the next release 5.2 we will enable Products M3 for everyone. So I disabled the M3 feature flag checks and I hide the beta switch from the settings.

## Changes
- In `DefaultFeatureFlagService`, editProductsRelease3 now returns always true.
- Removed the products section in `BetaFeaturesViewController`.
- Updated `SettingsViewController` for hiding the experimental features.
- In `ProductDetailsFactory` now we always returns `true` also for non-simple products.
- Updated `ProductsTopBannerFactory` for not considering edit products release 3.
- Updated tests in `ProductDetailsFactoryTests`.

## Testing
**Prerequisites:** the site has non-simple products (e.g. variable), and orders with non-simple products.

Try these steps from your current installation, but also deleting and reinstalling the app.

#### Case 1
1. Go to Settings. Make sure that the Experimental Features cell is not shown.
2. Go to the Orders tab.
3. Tap on an order with a non-simple product.
4. Tap on the product under the Products section --> you should be able to open and edit that product.

#### Case 2
1. Go to the Products tab.
2. Make sure that the top banner is shown. Press the "dismiss" button.
3. Close and re-open the app.
4. Go to the Products tab. -> Make sure that the top banner is now hidden.
5. Tap a non-simple product. -> you should be able to open and edit that product.

Update release notes:


Ecarrion (commented):
Code: 
I just have one question, is there a particular reason why you prefer leaving all the products_M3 conditionals to return true instead of deleting that code? 

Testing:
I tried to edit a variable product from an existing order and got the following error.
error-on updating product
Console said: 
> ⛔️ Error updating Product: unknown(error: Yosemite.AnyError(error: Dotcom Error: [woocommerce_rest_invalid_product_id] To manipulate product variations you should use the /products/&lt;product_id&gt;/variations/&lt;id&gt; endpoint.))

Additionally, loading that product made it appear on the product list as a separate product, which doesn't seem correct, right?
Screen Shot 2020-09-29 at 1 56 24 PM

pmusolino:
Thanks for the review @Ecarrion! 

> I just have one question, is there a particular reason why you prefer leaving all the products_M3 conditionals to return true instead of deleting that code?

We usually leave all the things about the feature flag for at least one or two releases before deleting it definitely. I think that it is the best option since if something goes wrong, we can request an expedit review re-enabling it quickly. This time there is another reason: there are some branches that need to be merged (related to Products M5) and removing all the things about the M3 flags means managing big conflicts in a lot of different files, so personally I prefer to leave it until these branches will be merged. I opened an issue as a reminder here https://github.com/woocommerce/woocommerce-ios/issues/2893

> I tried to edit a variable product from an existing order and got the following error.

Good catch Ernesto and 100 points to Gryffindor 😄😭 you discovered a new bug, not specifically related to this PR (also if you pull `develop` you will be able to replicate it). I opened a new issue https://github.com/woocommerce/woocommerce-ios/issues/2894

> Additionally, loading that product made it appear on the product list as a separate product, which doesn't seem correct, right?

This bug is really really weird and I'm not able to replicate it. 🤔 Are you able to replicate it constantly? If yes, can you give me the access to your store, please?

Ecarrion:
@pmusolino 
> We usually leave all the things about the feature flag for at least one or two releases before deleting it definitely. I think that it is the best option since if something goes wrong, we can request an expedit review re-enabling it quickly. This time there is another reason: there are some branches that need to be merged (related to Products M5) and removing all the things about the M3 flags means managing big conflicts in a lot of different files, so personally I prefer to leave it until these branches will be merged. I opened an issue as a reminder here https://github.com/woocommerce/woocommerce-ios/issues/2893

👍 

> Good catch Ernesto and 100 points to Gryffindor 😄😭 you discovered a new bug, not specifically related to this PR (also if you pull `develop` you will be able to replicate it). I opened a new issue https://github.com/woocommerce/woocommerce-ios/issues/2894

Yeah, I also tested in develop to see if it was related to this PR(which I was sure it was not by looking at the code changes) but since it failed on the testing steps I thought it was worth it to flag it here.

> This bug is really really weird and I'm not able to replicate it. 🤔 Are you able to replicate it constantly? If yes, can you give me the access to your store, please?

Yes I can replicate it easily, Just added your A8C email as a user on my store.

pmusolino:
> Yeah, I also tested in develop to see if it was related to this PR(which I was sure it was not by looking at the code changes) but since it failed on the testing steps I thought it was worth it to flag it here.

You did well 👍 

> Yes I can replicate it easily, Just added your A8C email as a user on my store.

Thank you so much! I tried to replicate it in your store, and I noticed that it happens only after trying to replicate this bug https://github.com/woocommerce/woocommerce-ios/issues/2894. It's like if we try to add a product to the list. I'm going to investigate this issue today.

pmusolino:
@Ecarrion I handled these two cases in this PR https://github.com/woocommerce/woocommerce-ios/pull/2900

Ecarrion (approved):
> @Ecarrion I handled these two cases in this PR #2900

Great @pmusolino 🚀 

I still was able to reproduce the [issue where the variation appears on the product list](https://github.com/woocommerce/woocommerce-ios/issues/2894#issuecomment-701271869) tho.
Should we track that in a separate issue?

pmusolino:
> I still was able to reproduce the issue where the variation appears on the product list tho.
Should we track that in a separate issue? 

I merged the work done in #2900, but I confirm that it's still present and I think it's a thing related to the way we store products in core data. I'm going to open a separate issue for that 😿

PR #2896, opened 2020-09-29
Configure the same user-agent header for all the network requests made through the app

### LINKED ISSUE #2895: Use the same user agent header for all the network requests made through the app
As discussed here **pau2Xa-1OO** and here **C6H8C3G23/p1601048110035500**, we should align the user agent headers sent in our network requests, because they are not aligned, and in our API requests we send "WooCommerce iOS" instead of "wc-ios".

We should align all the user agent headers to be a combination of `webkitUserAgent + "wc-ios" + bundle short version` similar to Android that sends a user agent like this:
```
Mozilla/5.0 (Linux; Android 9; AOSP on IA Emulator Build/PSR1.180720.117; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/69.0.3497.100 Mobile Safari/537.36 wc-android/5.1-rc-1
```
We should set also the user agent sent through our `WKWebView`s.

pmusolino:
Fixes #2895 

## Description
As discussed here pau2Xa-1OO and here C6H8C3G23/p1601048110035500 I aligned all the user agents sent through the app to be a combination of `webkitUserAgent + "wc-ios" + bundle short version`.

## Changes
- Moved the class `UserAgent` under Networking.
- Moved the class `UserAgentTests` under Networking.
- The `userAgent` variable was removed from `Settings`.
- In all the `WKWebView` (the one in Licenses and the one in Surveys) now we pass the new user agent.

## Testing
Just CI and a confidence check: login, API requests and `WKWebView's should continue to work as before.

Update release notes:

### DIFF
--- a/Networking/Networking/Settings/UserAgent.swift
+++ b/Networking/Networking/Settings/UserAgent.swift
@@ -1,28 +1,25 @@
-import Foundation
-import UIKit
 import WebKit
-
 
 /// WooCommerce User Agent!
 ///
-class UserAgent {
+public class UserAgent {
 
     /// Private: NO-OP
     ///
     private init() { }
 
 
-    /// Returns the WooCommerce User Agent
+    /// Returns the default WooCommerce iOS User Agent
     ///
-    static var defaultUserAgent: String = {
+    public static var defaultUserAgent: String = {
         return webkitUserAgent + " " + Constants.woocommerceIdentifier + "/" + bundleShortVersion
     }()
 
     /// Returns the WebKit User Agent
     ///
-    static var webkitUserAgent: String {
+    public static var webkitUserAgent: String {
         guard let userAgent = WKWebView().value(forKey: Constants.userAgentKey) as? String,
-            userAgent.isNotEmpty else {
+            !userAgent.isEmpty else {
                 return ""
         }
         return userAgent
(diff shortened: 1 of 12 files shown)


Networking/Networking/Settings/UserAgent.swift
```diff
@@ -1,36 +1,28 @@
...
     ///
-    static var defaultUserAgent: String = {
+    public static var defaultUserAgent: String = {
         return webkitUserAgent + " " + Constants.woocommerceIdentifier + "/" + bundleShortVersion
     }()
 
     /// Returns the WebKit User Agent
     ///
-    static var webkitUserAgent: String {
+    public static var webkitUserAgent: String {
```
jaclync: nit: this property could probably be internal to Networking? it seems to be used only in unit tests

jaclync (approved):
LGTM thanks for taking on this one!

Code: ✅ 
Testing: ✅ 

Login works as before!
For API requests and survey webview, I see user agent: `Mozilla/5.0 (iPhone; CPU iPhone OS 13_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 wc-ios/5.1` on my device

I didn't see a request from opening the license page so didn't verify that 🤔

Networking/Networking/Settings/UserAgent.swift
pmusolino: 👍 fixed here f94baa4368d619cff3379dbc13b2d1cbf1f288e3

pmusolino:
> I didn't see a request from opening the license page so didn't verify that 

Yes, it's correct because it loads a local file at the moment 😃

Issue #2885, opened 2020-09-29
Unit tests: replace `SessionManager.testingInstance` static singleton with `makeForTesting()`


jaclync:
Original discussion: https://github.com/woocommerce/woocommerce-ios/pull/2850#discussion_r495245234

We have a bunch of usage of `SessionManager.testingInstance` in unit tests. However, `.testingInstance` is actually a static singleton and let's instantiate a new instance every time in the updated `makeForTesting()` so that each test case is using a brand new `SessionManager`.

https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/WooCommerceTests/Internal/SessionManager+Internal.swift#L10-L14

joshheald:
I've checked this: `testingInstance` isn't actually a singleton, and it makes a brand new `SessionManager` whenever it's called.

As a static computed var, the closure is run every time it's called, it's not a `lazy` instantiation which gets stored.

I've confirmed this with a breakpoint in `testingInstance`, running a suite of tests, and printing the `sessionManager` each time we hit the breakpoint:

```
(lldb) po sessionManager
<SessionManager: 0x600001302f00>
(lldb) po sessionManager
<SessionManager: 0x600001337a00>
(lldb) po sessionManager
<SessionManager: 0x600001308e00>
(lldb) po sessionManager
<SessionManager: 0x60000130c800>
```

PR #2897, opened 2020-09-30
Orders tab (iOS 12 & iOS 13+): add site ID predicate for Orders

jaclync:
Precaution for #2830 

Before this PR, there is no predicate on site ID in Orders tab FRC. After #2830, we're not tearing down the persistent store and there could be remaining data that are saved after logging out. Having site ID as part of the FRC predicate is a good way to make sure any leftover data from another site do not get used for the currently logged in site. I did a pass of FRCs in the code base and just found the Orders tab, plus two models `SiteVisitStats` and `TopEarnerStats` don't have any site ID info. This PR focuses on the FRC for the Orders tab (2 implementations).

This change also allows us to support keeping orders for multiple sites in the future.

## Changes

- In https://github.com/woocommerce/woocommerce-ios/commit/b599ff9ed0cd43f1fbb5238bae9d9381a8ad6f97 (iOS 12 implementation, soon to be deprecated): refactored how a predicate is created in `OrdersViewModel`, and created a function `refreshOrdersPredicate` that is called when site changes in `OrdersViewController`'s observation. Note that after https://github.com/woocommerce/woocommerce-ios/issues/1838, this `refreshOrdersPredicate` won't be necessary anymore. From my testing, `.defaultAccountWasUpdated` does not get triggered after switching sites and `.StoresManagerDidUpdateDefaultSite` does.
- In https://github.com/woocommerce/woocommerce-ios/commit/5e92137089f8d700f20e460df77ce42d704e50a1: similar changes for iOS 13+ implementation.

## Testing

No user-facing changes are expected, please test for confidence check:

- Go to the Orders tab --> make sure both tabs show the same orders as before
- Go to My store tab and > Settings, and log out from the app
- Go to the Orders tab --> make sure both tabs show the orders from the new site
- Log out from the app
- Log in to a store
- Go to the Orders tab --> make sure both tabs show the orders from the new site

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
```diff
@@ -298,14 +298,21 @@ extension OrderListViewController {
...
     /// Runs whenever the default Account is updated.
     ///
     @objc func defaultAccountWasUpdated() {
-        refreshStatusPredicate()
         syncingCoordinator.resetInternalState()
     }
+
+    /// Default Site Updated Handler
+    ///
+    @objc func defaultSiteWasUpdated() {
```
Ecarrion: nit: Can this be private?

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
```diff
@@ -149,6 +132,40 @@ final class OrderListViewModel {
...
+            if !includesFutureOrders, let nextMidnight = Date().nextMidnight() {
+                // Exclude orders on and after midnight of today's date
+                let dateSubPredicate = NSPredicate(format: "dateCreated < %@", nextMidnight as NSDate)
+                predicates.append(dateSubPredicate)
+            }
+
+            return NSCompoundPredicate(andPredicateWithSubpredicates: predicates)
+        }()
+
+        guard let siteID = ServiceLocator.stores.sessionManager.defaultStoreID else {
```
Ecarrion, line 146: nit: Should we inject the `siteID`?

WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift
```diff
@@ -124,6 +118,30 @@ final class OrdersViewModel {
...
+            return NSCompoundPredicate(andPredicateWithSubpredicates: predicates)
+        }()
+
+        guard let siteID = stores.sessionManager.defaultStoreID else {
+            assertionFailure("Trying to create results controller without a site ID")
+            return predicate
+        }
+
+        let siteIDPredicate = NSPredicate(format: "siteID = %lld", siteID)
+        return NSCompoundPredicate(andPredicateWithSubpredicates: [siteIDPredicate, predicate])
```
Ecarrion, line 137: After seeing this code duplicated, I wonder if we should extract the predicates creation into a helper to avoid duplication? 🤔

Ecarrion (approved):
Everything looks ok! Just left some minor suggestions for your consideration

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
```diff
@@ -298,14 +298,21 @@ extension OrderListViewController {
     func startListeningToNotifications() {
         let nc = NotificationCenter.default
         nc.addObserver(self, selector: #selector(defaultAccountWasUpdated), name: .defaultAccountWasUpdated, object: nil)
+        nc.addObserver(self, selector: #selector(defaultSiteWasUpdated), name: .StoresManagerDidUpdateDefaultSite, object: nil)
```
shiki: ⚠️ When https://github.com/woocommerce/woocommerce-ios/pull/2850 is merged, it doesn't look like we will need this at all since the Orders tab and other VCs will be recreated anyway.

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -189,6 +177,19 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
...
+        guard query != self.query else {
+            return
+        }
+        self.query = query
+        fetchedResultsController = createFetchedResultsController(query: query)
+        do {
+            try activateFetchedResultsController()
+        } catch {
+            DDLogError("⛔️ FetchResultSnapshotsProvider: Failed to activate after updating query with error \(error)")
+        }
```
shiki: ⚠️  I'd recommend that we use `restart` instead so that the FRC will not be incorrectly activated. We have to ensure that `start()` was already called beforehand. 

```suggestion
        restartFetchedResultsController()
```

If `start` is not called, we don't want FRC to be active because we will not be observing the `NSManagedObjectContextObjectsDidChange` notifications.

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
```diff
@@ -189,6 +177,19 @@ public final class FetchResultSnapshotsProvider<MutableType: FetchResultSnapshot
             return nil
         }
     }
+
+    public func updateQuery(_ query: Query) {
```
shiki: We're probably not gonna need this (right now) since the OrderListViewController will be recreated anyway but would you mind adding a comment on what happens if this is called? 🙂 Also, good choice about using a function instead of exposing the `query` property!

WooCommerce/WooCommerceTests/ViewRelated/Orders/OrdersViewModelTests.swift
```diff
@@ -28,9 +29,14 @@ final class OrdersViewModelTests: XCTestCase {
     override func setUp() {
         super.setUp()
         storageManager = MockupStorageManager()
+        stores = MockupStoresManager(sessionManager: .makeForTesting(authenticated: true))
+        stores.sessionManager.setStoreId(siteID)
     }
 
     override func tearDown() {
+        // If not resetting store ID back to `nil`, it could cause other test failures since `setStoreId` changes UserDefaults.
+        stores.sessionManager.setStoreId(nil)
```
shiki, line 38: Yeah. Even though we use a different instance, they're still using the same `UserDefaults`. 🙈 So it's still technically a singleton! I wonder how we can improve this. 🤔

shiki (changes requested):
I didn't know that `Order` already has `siteID`! We should make that required in the future.

Thanks for working on this, @jaclync. In addition to Ernesto's comments, I have some requests. 🙂

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
jaclync: yes in a future PR! now that #2850 was merged, there is only one site ID for its lifetime. DI requires more levels of changes, so I plan to save it for a separate PR soon.

WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift
jaclync: since we are most likely deprecating iOS 12 support in the near future now that iOS 14 is gaining adoption, I think it's okay to keep the code in two places like the rest of Order list implementation. the iOS 12 usage is already low enough that we're ready to drop it, but we wanted to wait for Products M3 full release (release from this sprint!) so that users on iOS 12 get this feature.

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
jaclync: yes! originally I wasn't sure if the tab bar controller recreation PRs (https://github.com/woocommerce/woocommerce-ios/pull/2850 and https://github.com/woocommerce/woocommerce-ios/pull/2886) would be merged this sprint, and wanted to make the crash fix https://github.com/woocommerce/woocommerce-ios/pull/2879 along with this PR to make sure orders tab are only showing orders with the current site. now that the tab bar controller PRs were merged, no need for these changes anymore! removed in https://github.com/woocommerce/woocommerce-ios/pull/2897/commits/d6f8daf8ff384469d850a857ee718cfc090877ad

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
jaclync: Even though the changes for `updateQuery` were reverted in https://github.com/woocommerce/woocommerce-ios/pull/2897/commits/d6f8daf8ff384469d850a857ee718cfc090877ad, we might still need it in the future - e.g. in products tab, we might want to update the query after the user changes the filtering / sorting options when we update it to use diffable table view data source.

Originally, I wanted to call `restartFetchedResultsController` but hesitated on the guard statement `fetchedResultsControllerIsActive` (`fetchedResultsController.fetchedObjects != nil`). What if `activateFetchedResultsController` hasn't been called for the original query for some reason (probably a developer mistake)? What do you think if we call `start()` in case `fetchedResultsControllerIsActive` is `false`? Or there might some undesirable side effects?

```swift
guard fetchedResultsControllerIsActive else {
    start()
    return
}
```

Yosemite/Yosemite/SnapshotsProvider/FetchResultSnapshotsProvider.swift
jaclync: yup this was removed in https://github.com/woocommerce/woocommerce-ios/pull/2897/commits/d6f8daf8ff384469d850a857ee718cfc090877ad 🙂

WooCommerce/WooCommerceTests/ViewRelated/Orders/OrdersViewModelTests.swift
jaclync: yea it's a tricky singleton in unit tests that could affect the rest of the tests 😞 I think a straightforward improvement is to call `setStoreId(nil)` or reset `SessionManager.SessionSettings.defaults: UserDefaults` in `SessionManager.makeForTesting` (and we should always use this instead of `SessionManager.testingInstance` https://github.com/woocommerce/woocommerce-ios/issues/2885),  though I haven't tested. Another way is to move away from using the site ID from `SessionManager` and DI the site ID as much as possible.

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
jaclync: this function was removed in https://github.com/woocommerce/woocommerce-ios/pull/2897/commits/d6f8daf8ff384469d850a857ee718cfc090877ad 🙂

jaclync:
Thanks a lot for the prompt review @Ecarrion and @shiki! I updated and responded to the PR comments for another look - when it's ready I'll merge https://github.com/woocommerce/woocommerce-ios/pull/2879 along with this PR.

> I didn't know that Order already has siteID! We should make that required in the future.

@shiki did you mean make `StorageOrder`'s `siteID` required (not optional) in the schema? makes sense to me 👍 
Currently, there are still two entities that don't have a site ID yet: `SiteVisitStats` and `TopEarnerStats`. It'd be nice to add site ID to them as well.

WooCommerce/WooCommerceTests/ViewRelated/Orders/OrdersViewModelTests.swift
shiki: We could probably add a `SessionManagerProtocol` and provide a `MockSessionManager` in unit tests. The mock will not use `UserDefaults`, just in-memory storage. 

And then we'll inject the `SessionManagerProtocol` directly instead of the `StoresManager`. I think it's better if the session and the store are separate entities. 

And yeah, injecting `siteID` is probably better since most views don't need to _react_ to session changes. Injecting `siteID` is how it's usually done in WPAndroid: https://github.com/wordpress-mobile/WordPress-Android/search?q=siteid

shiki (approved):
> did you mean make StorageOrder's siteID required (not optional) in the schema? 

That's correct. It's another migration. 🤣 Not important at the moment. 

I created an issue for the `SiteVisitStats` and `TopEarnerStats`. https://github.com/woocommerce/woocommerce-ios/issues/2916

WooCommerce/WooCommerceTests/ViewRelated/Orders/OrdersViewModelTests.swift
jaclync: ah yea, refactoring `SessionManager` to a protocol with a mock implementation using in-memory storage would be the ultimate fix. I created an issue for this: https://github.com/woocommerce/woocommerce-ios/issues/2917

jaclync:
Thank you for creating the issue for the site ID and the review again @shiki! Merging this now along with the crash fix to make release 5.2

PR #2903, opened 2020-10-01
Move Data-Tests from IterativeMigratorTests to MigrationTests

### LINKED ISSUE #2398: Conventions or Framework for Creating Unit Tests for Core Data Migrations
Lightweight Core Data migrations like https://github.com/woocommerce/woocommerce-ios/pull/2391 require us to create a new `xcdatamodel`. It is very easy to screw these up and they are very difficult to review. 

We should make some sort of framework to protect us from mistakes. We should make it easy to create unit tests for migrations, including skipping versions.

### LINKED ISSUE #2478: Explore persistence store mocks for Core Data migration unit tests
Original discussion: https://github.com/woocommerce/woocommerce-ios/pull/2474#pullrequestreview-437981177

When adding unit tests for Core Data heavier weight migration that requires a mapping model, it's currently difficult to populate the database with the source model version if any property of an entity at the source model version has been renamed/deleted. Let's explore exporting a database with populated data (`WooCommerce.sqlite*`) at the expected source model version, and adding a migration unit test from this mock database to migrate to the destination model version.

shiki:
Refs #2883. Closes #2398. 

The [current tests in `CoreDataIterativeMigrator`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2880-fix-attribute-xcode-12/Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift) are quite mixed. Some of the tests are for testing the functionality of `CoreDataIterativeMigrator`. The rest is for testing the effects of migrating between different versions. In #2883, I created a new test class, [`MigrationTests`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2880-fix-attribute-xcode-12/Storage/StorageTests/CoreData/MigrationTests.swift), which is to be used for adding tests for migrating between different model versions.  

## Changes

In this PR, I moved and significantly simplified the unit tests for migrating between 2 versions. The `CoreDataIterativeMigratorTests` now contain only the tests for validating the functionality of `CoreDataIterativeMigrator`. 

When new model versions are created, unit tests should now be added to the `MigrationTests` file. 

## Testing

The build passing should be good enough. 🙂 

## Submitter Checklist


pmusolino (approved):
LGTM :shipit:  and thanks for changing the model 0 to the correct model 1 🗡️

Issue #2890, opened 2020-09-29
[Issue Refund] Add Issue Refund ViewController


Ecarrion:
- Create View controller and set up the necessary infrastructure to render the table view with all the necessary cells 
- Add next button

<img width="351" alt="Screen Shot 2020-09-29 at 11 37 58 AM" src="https://user-images.githubusercontent.com/562080/94587460-409f1380-0248-11eb-9e06-700e923e6f1d.png">

PR #2908, opened 2020-10-01
Adds type-safe TableView cell dequeue

Ecarrion:
# Why

_We deserve nice things_

----

While working on the refund screen #2890 I end up writing a lot of repeated code that looked like this:
```swift
guard let cell = tableView.dequeueReusableCell(withIdentifier: CellType.reuseIdentifier, for: indexPath) as? CellType else {
    fatalError("Cell could not be dequeed")
}
```

To mitigate this I've added a typesafe dequeue method on `TableView` that will shorten the code to:
```swift
let cell = tableView.dequeueReusableCell(CellType.self, for: indexPath)
```

In this PR I've also refactored all possible TableView dequeues to use this new method which gives us a bit of logging since most of the code only called fatal error without any message.

# Testing Steps.
- Making sure that the code compiles should be enough but doing a sanity check trough the app and see that all cells are working correctly doesn't hurt!

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
+++ b/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
@@ -551,19 +551,14 @@ extension StorePickerViewController: UITableViewDataSource {
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         guard let site = state.site(at: indexPath) else {
             hideActionButton()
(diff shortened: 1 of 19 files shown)


shiki (approved):

shiki:
We could probably do the same thing for `tableView.register(, forCellReuseIdentifier)`!!!

Ecarrion:
@shiki 
> We could probably do the same thing for `tableView.register(, forCellReuseIdentifier)`!!!
> 

Submitted https://github.com/woocommerce/woocommerce-ios/pull/2910 !

PR #2910, opened 2020-10-01
Shorter TableView register nib method

Ecarrion:
# Why 

@shiki while reviewing #2908 suggested that we should apply the same strategy to the `tableView.register(...` method

I think this is good because:
- It's less code to write
- We guarantee that we always use the class `reuseIdentifier` and avoid any possible typo.

# How
Added
```swift
func register(_ type: UITableViewCell.Type)
```
and 

```swift
 func registerNib(for type: UITableViewCell.Type) 
```
methods, additionally I've refactored the existing code to use these new methods.

# Testing steps
- Making sure that the code compiles should be enough but doing a sanity check trough the app and see that all cells are working correctly doesn't hurt!

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UITableView+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UITableView+Helpers.swift
@@ -27,7 +27,7 @@ extension UITableView {
         register(type, forCellReuseIdentifier: type.reuseIdentifier)
     }
 
-    /// Registers a `UITableViewCell`nib  using its `reuseIdentifier` property as the reuse identifier.
+    /// Registers a `UITableViewCell` nib  using its `reuseIdentifier` property as the reuse identifier.
     ///
     func registerNib(for type: UITableViewCell.Type) {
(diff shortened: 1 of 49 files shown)


WooCommerce/Classes/Extensions/UITableView+Helpers.swift
```diff
@@ -18,9 +18,21 @@ extension UITableView {
...
+// MARK: Typesafe Register & Dequeue
 extension UITableView {
 
+    /// Registers a `UITableViewCell` using its `reuseIdentifier` property as the reuse identifier.
+    ///
+    func register(_ type: UITableViewCell.Type) {
+        register(type, forCellReuseIdentifier: type.reuseIdentifier)
+    }
+
+    /// Registers a `UITableViewCell`nib  using its `reuseIdentifier` property as the reuse identifier.
```
pmusolino:
```suggestion
    /// Registers a `UITableViewCell` nib  using its `reuseIdentifier` property as the reuse identifier.
```

pmusolino (approved):
Looks good to me @Ecarrion :shipit:

Issue #2768, opened 2020-09-04
[Issue Refunds] Refund Summary View


Ecarrion:
After tapping the `next` button on the [refund by item view](https://github.com/woocommerce/woocommerce-ios/issues/1094), a summary of the amount to be refunded should be shown.

<img width="357" alt="summary" src="https://user-images.githubusercontent.com/562080/92197114-7eba3a80-ee36-11ea-9398-4a622aee3d04.png">

On non-automatic refunds, a disclaimer will be shown
<img width="272" alt="Screen Shot 2020-09-03 at 10 47 43 PM" src="https://user-images.githubusercontent.com/562080/92197523-91813f00-ee37-11ea-9355-5d12e515bccd.png">

## Known Goals

- Show refund method
- Include refunded shipping costs in the total amount to be refunded
- Allow entry of Reason for Refund

## Unknowns

Is the refund method hardcoded depending on the payment method or is it something the store owner can change?

## Split Tasks 

This has been split into multiple tasks so it's clearer what's left. 

- https://github.com/woocommerce/woocommerce-ios/issues/2980
- https://github.com/woocommerce/woocommerce-ios/issues/2981
- https://github.com/woocommerce/woocommerce-ios/issues/2982

shiki:
I have split the remaining into separate tasks and marking this to be closed by the PR #2979.

Issue #2211, opened 2020-04-29
Rename `SettingTitleAndValueTableViewCell` so it doesn't sound like it is meant for settings


shiki:
As discussed in https://github.com/woocommerce/woocommerce-ios/pull/2208#pullrequestreview-402433382, that cell is now used not just for "settings". 

Jaclyn's suggestion:

> SettingTitleAndValueTableViewCell looks like a good fit for this use case, maybe we could rename it to a more generic name like TitleAndValueTableViewCell in the future so that it doesn't sound like this cell is meant for settings?

Issue #2222, opened 2020-05-03
Document how to Create UI


shiki:
- Describe that we use xibs for layout only 
- Mention that styling is in the code
- Give examples

shiki:
I'll cancel this for now because things will change as we write more SwiftUI.

PR #2911, opened 2020-10-01
Add Initial UI for Refund Confirmation

shiki:
Part of #2768. This adds the initial UI for the Refunds Summary (Confirmation) page. This is currently just using hardcoded data.  

## Design

@Garance91540 Please let me know what you think of this design implementation. 

Before | After 
--------|-------
Simulator Screen Shot - iPhone 11 - 2020-10-01 at 16 00 09        |       Simulator Screen Shot - iPhone 11 (Dark) (13 7) - 2020-10-01 at 16 02 55

## Changes

I initially planned to make the [ProductSettings Mediator Pattern](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/Product%20Settings/ProductSettingsRows.swift) reusable but bailed out. I realized that it would be better to approach that when we have access to `UITableViewDiffableDataSource`. That should make things easier. 

In here, I'm following the design pattern used in [Product Editing](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/ProductFormTableViewModel.swift) but a bit lightweight and keeping things a bit reusable. I think some of these can be reused somewhere else like the rows structs:

https://github.com/woocommerce/woocommerce-ios/blob/b0e3ea981a34f6e76fde694fd0773e2e1deeaf90/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Issue%20Refunds/RefundConfirmationViewModel.swift#L42-L67

But I'm still not sure how this is gonna pan out when the UI becomes editable. 😬 

### SettingTitleAndValueTableViewCell

Note that this cell is supposed to be reusable. It just has an outdated name. https://github.com/woocommerce/woocommerce-ios/issues/2211

### HeadlineLabelTableViewCell

I also modified this to support the “Manual Refund via Stripe” styling shown in the screenshot. 

## Testing

1. Change the contents of [this function](https://github.com/woocommerce/woocommerce-ios/blob/b0e3ea981a34f6e76fde694fd0773e2e1deeaf90/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift#L506-L530) to:

     ```swift
        #warning("do not commit")

        let vc = RefundConfirmationViewController()
        navigationController?.pushViewController(vc, animated: true)
        return

    ```
2. Run the app using an iOS 13.0+ device.
3. Navigate to Orders. 
4. Pick an order. 
5. The UI above should be shown. These are currently static data. 

## Submitter Checklist


Garance91540:
Hi @shiki , I made some light color/font updates: https://d.pr/i/TyBCtA

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundConfirmationViewController.swift
```diff
@@ -0,0 +1,95 @@
+import Foundation
+import UIKit
+
+/// Presents a screen to confirm the refund with the user.
+///
+/// Shows the total amount to be refunded and allows the user to enter the reason for the refund.
+///
+final class RefundConfirmationViewController: UIViewController {
+
+    private lazy var tableView = UITableView(frame: .zero, style: .grouped)
```
Ecarrion, line 10: I was under the impression that the guideline was to create `xibs` for every VC or View 🤔

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundConfirmationViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+private extension RefundConfirmationViewController {
+    func configureTableView() {
+        // Register cells
+        let supportedCellTypes = [
+            SettingTitleAndValueTableViewCell.self,
+            TitleAndEditableValueTableViewCell.self,
+            HeadlineLabelTableViewCell.self
+        ]
+        supportedCellTypes.forEach {
+            tableView.register($0.loadNib(), forCellReuseIdentifier: $0.reuseIdentifier)
```
Ecarrion: You can now use the new `registerNib` method :-)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundConfirmationViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+                cell.applyDefaultLabelsStyle()
+            }
+            cell.selectionStyle = .none
+            return cell
+        case let row as RefundConfirmationViewModel.TitleAndEditableValueRow:
+            let cell = tableView.dequeueReusableCell(TitleAndEditableValueTableViewCell.self, for: indexPath)
+            cell.update(title: row.title, placeholder: row.placeholder)
+            return cell
+        case let row as RefundConfirmationViewModel.TitleAndBodyRow:
+            let cell = tableView.dequeueReusableCell(HeadlineLabelTableViewCell.self, for: indexPath)
```
Ecarrion, line 79: Much cleaner!

Ecarrion (approved):
Looks good to me! Just left a minor question and suggestion.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundConfirmationViewController.swift
shiki: 😱 Sorry about that. I can't find a reference to it. IIRC, in one of the PRs, @jaclync and I talked about not creating a xib file if the xib is going to be empty anyway. We have a few files that are just like this. They are just noise. If we do have UI elements like buttons, labels, and stack views, then that's when we create xibs. I had this logged in https://github.com/woocommerce/woocommerce-ios/issues/2222 but I'm really falling behind in documentation. 😞

Ecarrion:
@shiki one thing I forgot to mention 
> But I'm still not sure how this is gonna pan out when the UI becomes editable. 

I think this screen is not meant to be editable, or am I missing something?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundConfirmationViewController.swift
shiki: 🤸 

Changed in 807bb7a08025b0df542bda28a1e594985b88baa2.

shiki (commented):
> I think this screen is not meant to be editable, or am I missing something?

@Ecarrion I believe it's just the _reason for refund_ that is editable. It's editable on Android. 🙂

shiki:
@Garance91540 Here is the updated Ui based on the design you shared in https://github.com/woocommerce/woocommerce-ios/pull/2911#issuecomment-702701827. Please let me know if I missed anything. 

Before | After 
--------|-------
 |   

It doesn't seem possible to change the text color of the “Reason for refunding order” so I left it as is. This is how it looks like if the text field is filled in:

Garance91540 (approved):
👏

shiki:
Thank you, both!

Issue #2904, opened 2020-10-01
After opening a Product Variation from an Order Detail, the product variation appears in the Product List


pmusolino:
@Ecarrion while was reviewing this PR #2891, discovered a new bug where opening a product variation starting from an order detail, display the product variation in the product list. 😨 

After opening a Product Variation from an Order Detail, the product variation appears in the Product List.

**Steps to reproduce the bug:**
1. Navigate to the Order List.
2. Open an order that contains Variable Products.
3. Tap one of the variable products.
4. Close it, and navigate to the Products tab.
5. You will see that specific Product Variation in the list.

<img src="https://user-images.githubusercontent.com/495617/94667245-db622580-030e-11eb-87dd-3a87a3b916fe.png" width=300 />

PR #2915, opened 2020-10-02
Updated Networking and Yosemite layers for fetching a single ProductVariation

pmusolino:
Part of #2904 

## Description
To start fixing the issue encountered in #2904 we could have the possibility to fetch a single `ProductVariation` instead of the `Product` itself. In this PR I updated the Networking and the Yosemite layers for fetching and store a `ProductVariation` locally.

## Changes
- Added `loadProductVariation` method in `ProductVariationsRemote`.
- Added `ProductVariationsRemote` tests.
- Added a new mock json response `product-variation.json`.
- Implemented `deleteProductVariation` metjod in `StorageType+Deletions` extension.
- Added `retrieveProductVariation` action.
- Implemented the `retrieveProductVariation` action in `ProductVariationStore`.
- Added `ProductVariationStore` tests.

## Testing
- Just CI

Update release notes:


Yosemite/Yosemite/Stores/ProductVariationStore.swift
```diff
@@ -76,6 +78,34 @@ private extension ProductVariationStore {
         }
     }
 
+    /// Retrieves the product variation associated with a given siteID + productID + variationID.
+    ///
+    func retrieveProductVariation(siteID: Int64, productID: Int64, variationID: Int64,
+                                  onCompletion: @escaping (Result<ProductVariation, ProductUpdateError>) -> Void) {
+        let remote = ProductVariationsRemote(network: network)
```
jaclync: after https://github.com/woocommerce/woocommerce-ios/pull/2879, let's use its `remote` instance directly instead of creating a new one (which will be deallocated now that the remote completion block is not retained)

Yosemite/Yosemite/Stores/ProductVariationStore.swift
```diff
@@ -125,6 +155,14 @@ private extension ProductVariationStore {
         }
     }
 
+    /// Delete a specific Storage.ProductVariation with the specified `siteID` and `variationID`
+    ///
+    func deleteStoredProductVariation(siteID: Int64, variationID: Int64) {
```
jaclync: this doesn't seem used?

Yosemite/YosemiteTests/Stores/ProductVariationStoreTests.swift
```diff
@@ -290,6 +290,153 @@ final class ProductVariationStoreTests: XCTestCase {
...
+    // MARK: - ProductVariationAction.retrieveProductVariation
+
+    /// Verifies that `ProductVariationAction.retrieveProductVariation` effectively persists the retrieved product variation.
+    ///
+    func test_retrieveProductVariation_effectively_persisted() throws {
+        // Given
+        let store = ProductVariationStore(dispatcher: dispatcher, storageManager: storageManager, network: network)
+        let sampleProductVariationID: Int64 = 1275
+
+        network.simulateResponse(requestUrlSuffix: "products/\(sampleProductID)/variations/\(sampleProductVariationID)", filename: "product-variation")
```
jaclync: nit: it'd be nice to use `MockProductVariationsRemote` instead of the network response so that we can return any kind of result (same for the other instances in this test class)

jaclync (approved):
Just some comments after the merge of https://github.com/woocommerce/woocommerce-ios/pull/2879! I'll take over the PR and update it tomorrow for another round of review from the rest of the iOS team

Yosemite/Yosemite/Actions/ProductVariationAction.swift
```diff
@@ -10,6 +10,10 @@ public enum ProductVariationAction: Action {
     ///
     case synchronizeProductVariations(siteID: Int64, productID: Int64, pageNumber: Int, pageSize: Int, onCompletion: (Error?) -> Void)
 
+    /// Retrieves the specified ProductVariation.
+    ///
+    case retrieveProductVariation(siteID: Int64, productID: Int64, variationID: Int64, onCompletion: (Result<ProductVariation, ProductUpdateError>) -> Void)
```
jaclync: I think it'd be better for the completion block to return a generic `Error` instead of `ProductUpdateError`, since this action isn't related to saving/updating a product

jaclync:
I updated the PR based on my own review comments 😅 could anyone else from @woocommerce/ios-developers take a look at the PR when you get a chance?

Networking/NetworkingTests/Remote/ProductVariationsRemoteTests.swift
```diff
@@ -117,6 +117,55 @@ final class ProductVariationsRemoteTests: XCTestCase {
         wait(for: [expectation], timeout: Constants.expectationTimeout)
     }
 
+    // MARK: - Load single product variation tests
+
+    /// Verifies that loadProductVariation properly parses the `product-variation` sample response.
+    ///
+    func test_load_single_ProductVariation_properly_returns_parsed_ProductVariation() throws {
```
shiki: I think we should stop using "properly". I feel like it's implied. 😄

Networking/NetworkingTests/Remote/ProductVariationsRemoteTests.swift
```diff
@@ -117,6 +117,55 @@ final class ProductVariationsRemoteTests: XCTestCase {
...
+        waitForExpectation { exp in
+            remote.loadProductVariation(for: sampleSiteID, productID: sampleProductID, variationID: sampleProductVariationID) { aResult in
+                resultMaybe = aResult
+                exp.fulfill()
+            }
+        }
+
+        // Then
+        let result = try XCTUnwrap(resultMaybe)
+        XCTAssertTrue(result.isFailure)
```
shiki: We can now also use [`waitFor`](https://github.com/woocommerce/woocommerce-ios/blob/fe221b6ce87d5018a8ce21d5547859ddd6e14639/TestKit/Sources/TestKit/XCTestCase%2BWait.swift#L58-L95) so we can have an immutable variable. 

```suggestion
        // When
        let result = try waitFor { promise in
            remote.loadProductVariation(for: self.sampleSiteID, productID: self.sampleProductID, variationID: sampleProductVariationID) { result in
                promise(result)
            }
        }

        // Then
        XCTAssertTrue(result.isFailure)
```

shiki: Do you think we should also add an assertion for the type or message of the error?

Yosemite/YosemiteTests/Stores/ProductVariationStoreTests.swift
```diff
@@ -290,6 +290,138 @@ final class ProductVariationStoreTests: XCTestCase {
         wait(for: [expectation], timeout: Constants.expectationTimeout)
     }
 
+    // MARK: - ProductVariationAction.retrieveProductVariation
+
+    /// Verifies that `ProductVariationAction.retrieveProductVariation` effectively persists the retrieved product variation.
+    ///
+    func test_retrieveProductVariation_effectively_persisted() throws {
```
shiki: Do you think this describes the scenario better? 🤔 

```suggestion
    func test_retrieveProductVariation_persists_the_ProductVariation() throws {
```

shiki (approved):
Thanks for taking over, @jaclync. I just have minor suggestions that you can freely ignore. 😄

Networking/NetworkingTests/Remote/ProductVariationsRemoteTests.swift
jaclync: yup! I added another assertion `XCTAssertEqual(result.failure as? NetworkError, .notFound)` in https://github.com/woocommerce/woocommerce-ios/pull/2915/commits/26e5adeae232736e45eb0e54e0409e259a190c16

Networking/NetworkingTests/Remote/ProductVariationsRemoteTests.swift
jaclync: hah yea, removed in https://github.com/woocommerce/woocommerce-ios/pull/2915/commits/26e5adeae232736e45eb0e54e0409e259a190c16

Networking/NetworkingTests/Remote/ProductVariationsRemoteTests.swift
jaclync: thanks for the suggestion! this `waitFor` helper certainly simplifies a bunch of things, no need to try unwrapping to the non-optional value anymore. I updated all the usage in this PR in https://github.com/woocommerce/woocommerce-ios/pull/2915/commits/26e5adeae232736e45eb0e54e0409e259a190c16

Yosemite/YosemiteTests/Stores/ProductVariationStoreTests.swift
jaclync: it does read nicer! updated in https://github.com/woocommerce/woocommerce-ios/pull/2915/commits/26e5adeae232736e45eb0e54e0409e259a190c16

jaclync:
thanks for the suggestions @shiki! I updated the PR and responded to your comments, merging now 😃

PR #2918, opened 2020-10-05
Remove `SettingsViewController` from storyboard

jaclync:
Part of #1838

Before this PR, `SettingsViewController` was instantiated from `Dashboard.storyboard` and prevented us from DI'ing anything to it (still possible, but will the dependencies have to be optional properties). I started this refactoring so that we can DI the site ID to `SettingsViewController` then `BetaFeaturesViewController`. Later I realized `BetaFeaturesViewController` isn't actually using the site ID (it was used for the stats feature switch), but I thought it'd still be nice to pull this from the storyboard for easier DI in the future.

## Changes

- Removed `SettingsViewController` from `Dashboard.storyboard`, and created `SettingsViewController.xib` with the same UI as in the storyboard (a grouped `UITableView`)
- Replaced segues with navigation code in `SettingsViewController`
- Removed unused `siteID` from `BetaFeaturesViewController`

## Testing

- Launch the app
- Tap on the ⚙️ icon on the dashboard tab --> the settings screen should be shown as before
- Tap on each row just to make sure the navigation works as before

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Dashboard.storyboard
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Dashboard.storyboard
@@ -8,53 +8,16 @@
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <scenes>
-        <!--Settings View Controller-->
-        <scene sceneID="s9m-Yn-qYL">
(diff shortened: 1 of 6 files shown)


Ecarrion (approved):
👏 
So happy to see segues going away!

Issue #2920, opened 2020-10-05
Sometimes switching stores does not work


jaclync:
Could be similar to the issue reported by @pmusolino, or a different issue. When this bug happens, I see the following warning in the console:

```
Warning: Attempt to present <UINavigationController: 0x107a73200>  on <WooCommerce.WooNavigationController: 0x10785b600> which is already presenting (null)
```

After more testing/debugging, it seems like it was cause by `SettingsViewController` receiving tap events on "Switch Store" more than once at the same time, and `StorePickerCoordinator` calls its `lazy var StorePickerViewController` more than once so that the previous `StorePickerViewController`'s reference to `StorePickerCoordinator` is gone while being visible. I was also able to reproduce in release 5.1 (5.1.0.0) TestFlight build.

## Example timeline:

- `2020-10-05 17:06:49.108383+0800` presenting store: <WooCommerce.StorePickerViewController: 0x1222bba00>
- `2020-10-05 17:06:49.119110+0800` presenting store: <WooCommerce.StorePickerViewController: 0x12231ca00>

In Xcode memory debugger, both instances exist in memory while the first store picker instance `0x1222bba00` is presented in the UI via view debugger.

## Somewhat reliable steps to reproduce:

- Switch stores at least once
- Every time after switching store, quickly tap on ⚙️ icon to go to the settings screen, then quickly tap on "Switch Store" multiple times --> if you see `settings_selected_site_tapped` event being fired more than once within 0.1 second in the console, this bug is most likely to occur

jaclync:
What I've tried:

## Check `navigationController.presentedViewController` before presenting
I tried replacing L86-L87 of 

https://github.com/woocommerce/woocommerce-ios/blob/22a4a2fbf4f32181cb5ccf1b31e8d840dd068b3d/WooCommerce/Classes/Authentication/Epilogue/StorePickerCoordinator.swift#L85-L87

with:

```swift
if navigationController.presentedViewController == nil {
    let wrapper = UINavigationController(rootViewController: storePicker)
    navigationController.present(wrapper, animated: true)
}
```

this removes the warning `Attempt to present <UINavigationController: *>  on <WooCommerce.WooNavigationController: *> which is already presenting (null)` in the console, but still does not fix the bug.

## Replace delegate with closure

[...]

PR #2922, opened 2020-10-05
Store picker: fix a retain issue and `didSelectStore`'s completion block

jaclync:
Related to #2920 

While testing switching stores for #2920, I encountered some issues in the store picker:

- After switching stores `n` times, there are `n` `StorePickerViewController` instances strongly retained by `AccountHeaderView` in the memory
- In `StorePickerViewController.actionWasPressed` > `delegate.didSelectStore(with: site.siteID)`'s completion block, it's calling `cleanupAndDismiss` which could call `delegate?.didSelectStore` again whereas it can just dismiss itself

## Changes

In `StorePickerViewController`:
- Updated to use weak self in `AccountHeaderView.onHelpRequested` completion block
- Separated the dismiss part of `func cleanupAndDismiss` to its own function `func dismiss` and called this in `actionWasPressed` > `delegate.didSelectStore(with: site.siteID)`'s completion block

## Testing

- Go to "My store" tab
- Tap on the ⚙️ icon and switch to another store --> the store switch should work as before. after store picker is dismissed and the app is back to the dashboard, there should be no `StorePickerViewController` instance in Xcode memory debugger

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
+++ b/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
@@ -333,6 +333,10 @@ private extension StorePickerViewController {
             })
         }
 
+        dismiss()
+    }
+
+    func dismiss() {
         switch configuration {
         case .switchingStores:
             dismiss(animated: true)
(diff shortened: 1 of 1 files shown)


Ecarrion (commented):
Confirmed that the instance is no longer in memory!

Ecarrion (approved):

PR #2892, opened 2020-09-29
[Issue Refunds] Products refunds cell

Ecarrion:
closes #2767 

# Why 

End goal is to build the refunds UI, this PR focus on building the products refund costs detail cell(the blue rectangle).
Screen Shot 2020-09-29 at 3 36 53 PM

# How
- Created `RefundProductsTotalTableViewCell` which it's backed up by it's xib
- Created `RefundProductsTotalViewModel` to help to decouple the rendering content duties.

# Screenshots
 Regular |  Large
---- | ----
regular | large

# Testing Steps
Nothing as this is not yet visible on any screen.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2882, opened 2020-09-28
Adds RefundShippingDetailsTableViewCell

Ecarrion:
closes #2834 

# Why
The end goal is to build the refund UI, but this PR focuses on building the shipping details section (enclosed in the blue rectangle).

end-goal

# How
- Created `RefundShippingDetailsTableViewCell` backed up by its xib.
- Created simple `RefundShippingDetailsViewModel` to aid rendering content on `RefundShippingDetailsTableViewCell`. This just declares properties for now. More logic will be added as the development progresses.

# Screenshots

Light | Dark | XXL | RTL
---- | ---- | ---- | ---- 
normal |  dark |  big-font |  rtl

# Testing Steps
- Nothing, as the cell is not yet integrated into any screen

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2871, opened 2020-09-24
[Issue Refund] Refund Item Cell

Ecarrion:
closes #2839

# Why

The end goal is to build the refund UI, but this PR focuses on building a single refund item cell(enclosed in the blue rectangle). 
end-goal

# How

- Created `RefundItemTableViewCell` backed up by it's `xib`
- Created simple `RefundItemViewModel` to aid rendering content on `RefundItemTableViewCell`. This just declares properties for now. More logic will be added as the development progresses.
- I'm changing a bit the alignment of components as the UI grows in order to make content more readable.

One particularly weird thing that I did was to have a separate image view for the product placeholder which is inseted from the main product image view. This is to avoid the contentMode change we do in most of our cells. I think this looks a bit cleaner!

https://github.com/woocommerce/woocommerce-ios/blob/d42b886b808ce821261bf98cafda9928ac398e43/WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.swift#L83-L96

# Screenshots

screenshots

# Testing Steps
- Nothing much to test as the cell is not integrated yet 

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2925, opened 2020-10-05
[Issue Refund] Add Issue Refund View Controller

### LINKED ISSUE #2890: [Issue Refund] Add Issue Refund ViewController

Ecarrion:
#closes #2890 

# Why

This PR implements the full(and static) UI of the funds screen.
old-style

PS: That design has legacy app styles.

# How

- Created `IssueRefundsViewController` that renders all previously created cells #2871 #2882 #2892 
- Created `IssueRefundsViewModel` to provide what information to render, this info is static at the moment
- Launch `IssueRefundsViewController` from "issue refund" button on the Order Detail screen
- Updated some cell fonts from `body` to `semibold`.

# Screenshots

Light(Up) | Light(Down)
---- | ----
light-upZ | light-down

Dark(Up) | Dark(Down)
---- | ----
dark-up | dark-down

# Testing Steps

- Launch the app and navigate to a non-refunded order
- Tap on the issue refund button
- See static content being rendered correctly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
@@ -46,10 +46,8 @@ private extension RefundProductsTotalTableViewCell {
         subtotalPriceLabel.applyBodyStyle()
         taxTitleLabel.applyBodyStyle()
         taxPriceLabel.applyBodyStyle()
-        productsRefundTitleLabel.applyBodyStyle()
-        productsRefundTitleLabel.font = .font(forStyle: .body, weight: .semibold)
-        productsRefundPriceLabel.applyBodyStyle()
-        productsRefundPriceLabel.font = .font(forStyle: .body, weight: .semibold)
+        productsRefundTitleLabel.applyHeadlineStyle()
+        productsRefundPriceLabel.applyHeadlineStyle()
 
         taxTitleLabel.text = Localization.taxTitle
         subtotalTitleLabel.text = Localization.subtotalTitle
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
@@ -47,9 +47,9 @@ private extension RefundProductsTotalTableViewCell {
         taxTitleLabel.applyBodyStyle()
         taxPriceLabel.applyBodyStyle()
         productsRefundTitleLabel.applyBodyStyle()
-        productsRefundTitleLabel.font = .font(forStyle: .body, weight: .bold)
+        productsRefundTitleLabel.font = .font(forStyle: .body, weight: .semibold)
         productsRefundPriceLabel.applyBodyStyle()
-        productsRefundPriceLabel.font = .font(forStyle: .body, weight: .bold)
+        productsRefundPriceLabel.font = .font(forStyle: .body, weight: .semibold)
 
         taxTitleLabel.text = Localization.taxTitle
         subtotalTitleLabel.text = Localization.subtotalTitle
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
@@ -16,15 +16,65 @@ final class IssueRefundViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        configureTableView()
     }
 }
 
 // MARK: View Configuration
 private extension IssueRefundViewController {
+    func configureTableView() {
+        registerCells()
+        tableView.delegate = self
+        tableView.dataSource = self
+    }
+
     func registerCells() {
-        tableView.register(RefundItemTableViewCell.loadNib(), forCellReuseIdentifier: RefundItemTableViewCell.reuseIdentifier)
-        tableView.register(RefundProductsTotalTableViewCell.loadNib(), forCellReuseIdentifier: RefundProductsTotalTableViewCell.reuseIdentifier)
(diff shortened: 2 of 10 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
```diff
@@ -47,9 +47,9 @@ private extension RefundProductsTotalTableViewCell {
         taxTitleLabel.applyBodyStyle()
         taxPriceLabel.applyBodyStyle()
         productsRefundTitleLabel.applyBodyStyle()
-        productsRefundTitleLabel.font = .font(forStyle: .body, weight: .bold)
+        productsRefundTitleLabel.font = .font(forStyle: .body, weight: .semibold)
```
shiki: ⚠️ It's probably best if we stick to what's available in the styles. What about `applyHeadlineStyle()`? If that's not possible, we can make a different style in that file. Or maybe it's the headline style that should be changed to semi-bold. It looks like the design standard mentions that it should be semi-bold: https://woomobiledesignhandbook.wordpress.com/typography/

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+protocol IssueRefundRow {}
+
+extension IssueRefundViewModel {
+
+    struct Section {
+        let rows: [IssueRefundRow]
+    }
+
+    /// ViewModel that represents the shipping switch row.
+    struct ShippingSwitchViewModel: IssueRefundRow {
```
shiki, line 51: Do you think we can name it as `ShippingSwitchRowViewModel`? It's a bit longer but maybe it's clearer? Let me know what you think. Cause I'm not sure. 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -0,0 +1,61 @@
...
+        let title: String
+        let isOn: Bool
+    }
+}
+
+extension RefundItemViewModel: IssueRefundRow {}
+
+extension RefundProductsTotalViewModel: IssueRefundRow {}
+
+extension RefundShippingDetailsViewModel: IssueRefundRow {}
```
shiki, line 61: Nothing to change here. Just wanted to mention that I was initially confused about why these were not nested under `IssueRefundViewModel`. 😄 

I'm always concerned about implementing protocols outside of the declaration as they can be difficult to track and pollutes the namespace. I wonder if having a separate nested `struct`, which would clearly state that the `struct` can only be used by `IssueRefundViewModel` and the ViewController, would be better. 

```swift
extension IssueRefundViewModel {
    struct ItemRowViewModel: IssueRefundRow {
        let refundItem: RefundItemViewModel
    }
}
```

It does become repetitive though. 

And again, just thinking out loud for now. I'm open to seeing how this would evolve. 🙂

shiki (approved):
Looking good, @Ecarrion! I have one tiny concern (marked ⚠️ ) that I would love to hear your thoughts on.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
```diff
@@ -0,0 +1,141 @@
+import UIKit
+
+/// Screen that allows the user to refund items (products and shipping) of an order
+///
+final class IssueRefundViewController: UIViewController {
```
shiki, line 5: When the project is close to completion, we should probably think about renaming them so the names are closely related. 😅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
Ecarrion: Thanks for bringing this up! I totally missed that the `headline` style already defaults to semibold 🤦 
Screen Shot 2020-10-06 at 3 18 12 PM

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
Ecarrion: I didn't feel it was necessary but I'm happy to change it if you think it will lead to confusion.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
Ecarrion: Yeah, I hear you on that the protocol conformance can be difficult to track, but I also didn't want to overcomplicate it too much. As you say, let's see how this evolves.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
shiki: Oh, cool! Where is this reference? 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
shiki: It's fine!

shiki (approved):
:shipit:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundProductsTotalTableViewCell.swift
Ecarrion: https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/typography/

Garance91540 (approved):
👏

PR #2933, opened 2020-10-07
Update coding-style-guide to allow one liner guard self

Ecarrion:
# Why

Update the style guide to allow one-liner `guard let self = self else { return }`

Ref: CGPNUU63E-p1601997825048400-slack

Update release notes:


shiki (approved):
🤸

PR #2935, opened 2020-10-07
Document Core Data Model Version Guidelines

shiki:
Documents the discussion about Core Data model versions (p91TBi-3pE-p2).

## Testing

Please read and review the documentation [here](https://github.com/woocommerce/woocommerce-ios/blob/24769a1c918df2656776970b972172459ec7f2b4/docs/creating-core-data-model-versions.md). 

## Submitter Checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -10,6 +10,7 @@
         - [String Constants in Nested Enums](#string-constants-in-nested-enums)
         - [Test Methods](#test-methods)
     - [Choosing Between Structures and Classes](#choosing-between-structures-and-classes)
+    - [Creating Core Data Model Versions](creating-core-data-model-versions.md)
 - [Design Patterns](#design-patterns)
     - [Copiable](#copiable)
         - [Generating Copiable Methods](#generating-copiable-methods)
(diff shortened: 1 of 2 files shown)


jaclync (approved):
Thanks a lot for the detailed documentation 💟 Nice link to @astralbodies's Core Data migration post!

I like the section about adding unit tests, enforcing unit tests for each model version and explaining why `NSManagedObject` is preferred. If we ever have a model version that renames an existing required attribute (pretty unlikely and it requires manual migration), we might have to version its `insert*(to:) -> NSManagedObject` helper.

Ecarrion (approved):
Thank you for this! 🚀

shiki:
> If we ever have a model version that renames an existing required attribute (pretty unlikely and it requires manual migration), we might have to version its insert*(to:) -> NSManagedObject helper.

I think it can be done using a mapping model. And yeah, we'll have to version the `insert()`. I did not think that far ahead yet. 😄 

Thank you, both!

astralbodies:
@shiki Flattery will get you nowhere!

Who am I kidding, of course, it helps hahahahaha

PR #2936, opened 2020-10-07
Make FulfillButtonTableViewCell Reusable

shiki:
Ref #2768. 

I'm planning to reuse this cell for #2768 so I made it reusable.

## Testing

Please test that fulfillment still works.

1. Navigate to an order that is in “Processing”. 
2. Tap on “Begin Fulfillment”. 
3. Confirm that the fulfillment screen is shown. 

## Submitter Checklist


WooCommerce/Classes/ViewRelated/ReusableViews/ButtonTableViewCell.swift
```diff
@@ -0,0 +1,55 @@
+import UIKit
+
+/// Displays a button inside a `UITableViewCell`.
+///
+final class ButtonTableViewCell: UITableViewCell {
+    @IBOutlet private var button: UIButton!
```
jaclync, line 6: thanks for updating the button property from public to private!

jaclync (approved):
LGTM! :shipit: 

Code: ✅ (the new cell is also cleaner!)
Testing: ✅

PR #2938, opened 2020-10-08
Make FulfillmentVC Footer View Reusable

shiki:
Ref #2768. 

I made a mistake in #2936, it turns out I could not use that for the button in #2768. But since it's done and that button is now reusable anyway, I'm leaving the PR open. 😅 

## Changes 

This PR transfers this button in `FulfillViewController.xib` to a reusable class, [`ButtonTableFooterView`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2768-reuse-button-footerview/WooCommerce/Classes/ViewRelated/ReusableViews/ButtonTableFooterView.swift). 

I then added a `ButtonTableFooterView` to the refund confirmation page. 

## Testing

1. Navigate to an unfulfilled order.
2. Tap “Begin Fulfillment”. 
3. Tap “Mark Order Complete”. Confirm that the order status was changed to “Completed”. 
4. Navigate to another order.
5. Tap “Issue Refund”. 
6. Tap “Next”. 
7. Confirm there is a “Refund” button. This page is currently read-only so there's nothing else to do here. 🤪 

## Submitter Checklist


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
```diff
@@ -38,7 +38,7 @@ final class IssueRefundViewController: UIViewController {
 // MARK: Actions
 private extension IssueRefundViewController {
     @IBAction func nextButtonWasPressed(_ sender: Any) {
-        print("Next button pressed")
+        navigationController?.pushViewController(RefundConfirmationViewController(), animated: true)
```
Ecarrion:
```suggestion
        show(RefundConfirmationViewController(), animated true)
```
We can use `show` to decouple ourselves a bit from the optional navigation controller.

Ecarrion (approved):
Neat! I guess I should use this on `IssueRefundViewController`

----

While I was testing that this button rotated correctly I noticed that the table view on `RefundConfirmationViewController` does not expand all the way to the screen edges which makes it look a bit odd when having the safe area margins. I think we should pin it to the view edges and let the table view handle the safe area insets.

Screen Shot 2020-10-08 at 7 58 28 AM

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
shiki: Thanks! TIL. 

Fixed in bd759a67aecf50d3d7e05da86c9375353db1c65f.

shiki (commented):
> I think we should pin it to the view edges and let the table view handle the safe area insets.

@Ecarrion It looks like this has to be applied to iPhones with notches and on landscape only. I don't know of a good way to approach this at the moment. 😅 I created an issue so we can tackle it later. https://github.com/woocommerce/woocommerce-ios/issues/2943

Thanks for the review!

Ecarrion:
> @Ecarrion It looks like this has to be applied to iPhones with notches and on landscape only. 

@shiki I think you just need to pin the view to the edges and the table view automatically adjusts the offsets based on the safe area!

PR #2939, opened 2020-10-08
Refactor `ProductAction.retrieveProduct` to return `Result<Product, Error>` instead of `(Product?, Error?)`

jaclync:
Prep for #2904 

This PR is not meant to introduce user-facing changes, and is to simplify the logic for the main PR for #2904.

## Changes

- Updated `ProductAction.retrieveProduct` to return `Result<Product, Error>` instead of `(Product?, Error?)`
- Refactored affected unit tests to follow Arrange/Action/Assert style with non-nil result

## Testing

Confidence check on affected screens:

- Go to the Orders tab
- Tap on an order
- Tap on a product on order details --> it should open the readonly product details successfully
- Pull down to refresh the product details --> it should refresh the product successfully

(I tested the case when the device has no network connection, and the error in the console shows the error details)

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/ProductStore.swift
+++ b/Yosemite/Yosemite/Stores/ProductStore.swift
@@ -233,8 +233,7 @@ private extension ProductStore {
             case .success(let product):
                 self.upsertStoredProductsInBackground(readOnlyProducts: [product]) { [weak self] in
                     guard let storageProduct = self?.storageManager.viewStorage.loadProduct(siteID: siteID, productID: productID) else {
-                        onCompletion(.failure(ProductLoadError.notFoundInStorage))
-                        return
+                        return onCompletion(.failure(ProductLoadError.notFoundInStorage))
                     }
                     onCompletion(.success(storageProduct.toReadOnly()))
                 }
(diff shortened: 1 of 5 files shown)


Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -229,11 +229,14 @@ private extension ProductStore {
...
                 }
-                onCompletion(nil, error)
+                onCompletion(.failure(error))
             case .success(let product):
                 self.upsertStoredProductsInBackground(readOnlyProducts: [product]) { [weak self] in
-                    let storageProduct = self?.storageManager.viewStorage.loadProduct(siteID: siteID, productID: productID)
-                    onCompletion(storageProduct?.toReadOnly(), nil)
+                    guard let storageProduct = self?.storageManager.viewStorage.loadProduct(siteID: siteID, productID: productID) else {
+                        onCompletion(.failure(ProductLoadError.notFoundInStorage))
+                        return
```
Ecarrion: nit: I think you merge this two into one line!
```suggestion
                        return onCompletion(.failure(ProductLoadError.notFoundInStorage))
```

Ecarrion (approved):
LGTM! Thanks for improving this! 🚀

Yosemite/Yosemite/Stores/ProductStore.swift
jaclync: TIL'ed! updated in https://github.com/woocommerce/woocommerce-ios/pull/2939/commits/9fd4e2606316f8aeb751caab43ab81a79c314d0c

Issue #2931, opened 2020-10-07
Product/variation details: add support for readonly state


jaclync:
We're ready to deprecate the old product details UI (`ProductDetailsViewController`), but we recently changed the product requirement from editable to readonly product details from outside of the Products tab. Since `ProductFormViewController` does not have a readonly mode, we still use the old UI.

Let's add readonly support to `ProductFormViewController` so that we can fully deprecate `ProductDetailsViewController` to have consistent UI. Also, https://github.com/woocommerce/woocommerce-ios/issues/2904 occurs because `ProductDetailsViewController` only supports `Product` and not a `ProductVariation`. It'd be easier to support both readonly product & product variation details with `ProductFormViewController`.

- Ellipsis menu: view product in store & share product if applicable
- Product description row: hidden when readonly & empty description
- Product images row: hidden when there are no images

### External/affiliate products
- External URL row: hidden when external URL is empty
- SKU row: hidden when SKU is empty

### Grouped products
- SKU row: hidden when SKU is empty

PR #2940, opened 2020-10-08
Opening a Product Variation from order details now shows editable variation details and the variation does not appear in Products tab

### LINKED ISSUE #2904: After opening a Product Variation from an Order Detail, the product variation appears in the Product List

jaclync:
Fixes #2904

Depends on https://github.com/woocommerce/woocommerce-ios/pull/2939

Note that we don't have support for readonly variation details yet https://github.com/woocommerce/woocommerce-ios/issues/2931, so we show editable variation details for now. Readonly `ProductDetailsViewController` is deprecated soon and I don't think it's worth adding variation support to it.

Apology for the ~600 diff PR 🙇🏻‍♀️ I tried covering all cases in the unit tests and they turned out to be ~289 lines.

## Changes

- Created `ProductLoaderViewController.Model` enum that describes the type of product model to load - it could be a product or product variation with its associated info to load data remotely
  - Added convenience initializers for `ProductLoaderViewController.Model` from `OrderItem`, `OrderItemRefund`, `AggregateOrderItem`, and `TopEarnerStatsItem` with the existing logic based on its `variationID == 0`
  - Updated current usage of `ProductLoaderViewController`
- Created `ProductVariationLoadUseCase` that encapsulates the async logic for fetching a `ProductVariation` and its parent `Product` remotely
- Created `ProductVariationDetailsFactory` that makes a view controller that displays product variation details
- Updated `ProductLoaderViewController` to take in a `ProductLoaderViewController.Model`, load product variation data using `ProductVariationLoadUseCase`, and present variation details using `ProductVariationDetailsFactory`
- Added unit tests for the new classes

### Top Performer > variation details

- On the store shop page, generate an order that contains a variable product's variation
- Launch the app
- Go to the Products tab and wait for the product list to load (so that we can verify the variation does not appear later)
- Go to the Dashboard tab, tap on the top performer product row that is a product variation (should be in the "Today" tab) --> the editable variation details should be shown
- Edit some variation fields and tap "Update" --> the variation details should be saved remotely
- Dismiss the variation details and go to the Products tab --> the variation should not appear in the Products tab

### Order details > variation details

- Go to the Orders tab, tap on the order that contains a variation, and tap on on the variation row (the name is its parent product's name) --> the editable variation details should be shown
- Dismiss the variation details and go to the Products tab --> the variation should not appear in the Products tab

### Order details > begin fulfillment > variation details

- Go to the Orders tab, tap on the order that contains a variation
- Tap on "Begin Fulfillment"
- Tap on the variation row --> the editable variation details should be shown
- Dismiss the variation details and go to the Products tab --> the variation should not appear in the Products tab

### Order details > Refunded row > variation details

- In wp-admin, refund the variation line item
- Go to the Orders tab, tap on the order that contains a variation and whose variation is refunded
- Tap on the "Refunded" row under payment section that contains the variation
- Tap on on the variation row (the name is its parent product's name) --> the editable variation details should be shown
- Dismiss the variation details and go to the Products tab --> the variation should not appear in the Products tab

## Example screencast

simulator

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Product Loader/ProductLoaderViewController.swift
```diff
@@ -3,18 +3,22 @@ import UIKit
...
-/// ProductLoaderViewController: Loads asynchronously a Product (given it's ProductID + SiteID).
-/// On Success the ProductDetailsViewController will be rendered "in place".
+/// ProductLoaderViewController: Loads product and/or product variation async and render its details.
 ///
 final class ProductLoaderViewController: UIViewController {
+    /// Indicates the type of data model with the associated information to load data remotely.
+    enum Model: Equatable {
+        case product(productID: Int64)
+        case productVariation(productID: Int64, variationID: Int64)
+    }
```
pmusolino, line 13: I like the idea of passing a `Model` in the `init` method 💯

WooCommerce/Classes/ViewRelated/Products/Product Loader/ProductVariationLoadUseCase.swift
```diff
@@ -0,0 +1,72 @@
+import Yosemite
+
+/// Encapsulates the async logic for loading product variation and related data (its parent product) for UI display.
+final class ProductVariationLoadUseCase {
```
pmusolino, line 4: Do you think that it worth moving this class under Yosemite? It seems related, also if here we don't do storage operations.

WooCommerce/Classes/ViewRelated/Products/ProductVariationDetailsFactory.swift
```diff
@@ -0,0 +1,59 @@
...
+                                        presentationStyle: ProductFormPresentationStyle,
+                                        currencySettings: CurrencySettings = ServiceLocator.currencySettings,
+                                        forceReadOnly: Bool,
+                                        onCompletion: @escaping (UIViewController) -> Void) {
+        let vc = productVariationDetails(productVariation: productVariation,
+                                         parentProduct: parentProduct,
+                                         presentationStyle: presentationStyle,
+                                         currencySettings: currencySettings,
+                                         isEditProductsEnabled: forceReadOnly ? false: true,
+                                         isEditProductsRelease3Enabled: true,
```
pmusolino: Do you think it's worth to pass `ServiceLocator.featureFlagService.isFeatureFlagEnabled(.editProductsRelease3)` here?
Just because we can remember easily when we will drop the `editProductsRelease3` feature flag to remove it.

pmusolino (approved):
Thanks @jaclync for taking care of this! ❤️ 


Code ✅ Looks good to me! 
Testing ✅ Everything works as expected

WooCommerce/Classes/ViewRelated/Products/Product Loader/ProductVariationLoadUseCase.swift
jaclync: In Yosemite, we can't easily dispatch actions to different stores via `StoresManager` though - in this use case, it dispatches actions to both `ProductStore` and `ProductVariationStore`. If the actions are all in the same store, it certainly makes sense!

WooCommerce/Classes/ViewRelated/Products/ProductVariationDetailsFactory.swift
jaclync: good point! updated in https://github.com/woocommerce/woocommerce-ios/pull/2940/commits/159456efb8123558650cd18a3e0719a5e27defd3

jaclync:
Thanks for the review & testing Paolo, I responded to your comments and I'm merging the PR now!

Issue #2842, opened 2020-09-20
[Issue refund] Render order content


Ecarrion:
- Render items that can be refunded
- Take into consideration previously refunded items
- Make sure taxs are calculated correctly

PR #2946, opened 2020-10-08
[Issue Refunds] Shipping Switch Handling

Ecarrion:
part of #2842

# Why
This PR takes care of two things:
1. Render the shipping details to refund.
2. Show/Hide the shipping detail when the user enables/disables the "Refund Shipping" switch.

# How
1. Update `RefundShippingDetailsViewModel` to include a new initializer that will format its fields appropriately.
2. Added a `State` struct in `IssueRefundViewController` to hold on to the state generated by the users. 
   Modify that state when the user toggles the shipping switch. 
   Recreate the table view sections everything that states changes.

# Screenshots
No Shipping | Shipping Disabled | Shipping Enabled
--- | --- | ---
no-shipping | shipping-disabled | shipping-enabled

# Testing steps

## No shipping
- Launch app and navigate to an underfunded order that does not have shipping
- Tap on the "Issue refund" button
- See that the issue refund screen does not show a shipping section

## Shipping
- Launch app and navigate to an underfunded order that does have shipping
- Tap on the "Issue refund" button
- See that the screen shows a disabled switch to refund shipping
- Enable the switch
- See that the screen shows the shipping details to refund

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundShippingDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundShippingDetailsViewModel.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Yosemite
 
 /// Represents shipping details for an order to be refunded. Meant to be rendered by `RefundShippingDetailsTableViewCell`
 ///
@@ -8,4 +9,31 @@ struct RefundShippingDetailsViewModel {
     let shippingTax: String
     let shippingSubtotal: String
     let shippingTotal: String
+}
+
+// MARK: Convenience Initializers
+extension RefundShippingDetailsViewModel {
+    /// Creates a `RefundShippingDetailsViewModel` based on a `ShippingLine`
+    ///
+    init(shippingLine: ShippingLine, currency: String, currencySettings: CurrencySettings) {
+        let currencyFormatter = CurrencyFormatter(currencySettings: currencySettings)
+        self.carrierRate = shippingLine.methodTitle
+        self.carrierCost = currencyFormatter.formatAmount(shippingLine.total, with: currency) ?? ""
+        self.shippingTax = currencyFormatter.formatAmount(shippingLine.totalTax, with: currency) ?? ""
+        self.shippingSubtotal = currencyFormatter.formatAmount(shippingLine.total, with: currency) ?? ""
+        self.shippingTotal = {
+            let calculatedTotal = Self.calculateShippingTotal(of: shippingLine, using: currencyFormatter)
+            return currencyFormatter.formatAmount(calculatedTotal, with: currency) ?? ""
+        }()
+    }
+
+    /// Calculates the shipping total by adding the shipping cost + the shipping tax
+    ///
+    private static func calculateShippingTotal(of shippingLine: ShippingLine, using currencyFormatter: CurrencyFormatter) -> String {
+        guard let cost = currencyFormatter.convertToDecimal(from: shippingLine.total),
+            let tax = currencyFormatter.convertToDecimal(from: shippingLine.totalTax) else {
+                return ""
+        }
+        return cost.adding(tax).stringValue
+    }
 }
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundShippingDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundShippingDetailsViewModel.swift
@@ -29,11 +29,11 @@ extension RefundShippingDetailsViewModel {
 
     /// Calculates the shipping total by adding the shipping cost + the shipping tax
     ///
-    private static func calculateShippingTotal(of shippingLine: ShippingLine, using currencyFormatter: CurrencyFormatter) -> String {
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundShippingDetailsViewModel.swift
```diff
@@ -9,3 +10,30 @@ struct RefundShippingDetailsViewModel {
...
+    }
+
+    /// Calculates the shipping total by adding the shipping cost + the shipping tax
+    ///
+    private static func calculateShippingTotal(of shippingLine: ShippingLine, using currencyFormatter: CurrencyFormatter) -> String {
+        guard let cost = currencyFormatter.convertToDecimal(from: shippingLine.total),
+            let tax = currencyFormatter.convertToDecimal(from: shippingLine.totalTax) else {
+                return ""
+        }
+        return cost.adding(tax).stringValue
```
jaclync: how about returning an `NSDecimalNumber` directly and pass this value to `CurrencyFormatter.formatAmount(_ amount: NSDecimalNumber, ...)`? Otherwise, `CurrencyFormatter.formatAmount(_ amount: String, ...)` calls `convertToDecimal` again while we already have the `NSDecimalNumber` value

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
```diff
@@ -42,6 +44,21 @@ final class IssueRefundViewController: UIViewController {
...
 
+// MARK: ViewModel observation
+private extension IssueRefundViewController {
+    func observeViewModel() {
+        viewModel.onChange = { [weak self] in
+            self?.updateVithViewModelContent()
+        }
+    }
+
+    func updateVithViewModelContent() {
```
jaclync:
```suggestion
    func updateWithViewModelContent() {
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -5,9 +5,30 @@ import Yosemite
 ///
 final class IssueRefundViewModel {
 
-    /// Order to be refunded
+    /// Struct to hold the necesary state to perform the refund and update the view model
```
jaclync:
```suggestion
    /// Struct to hold the necessary state to perform the refund and update the view model
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -5,9 +5,30 @@ import Yosemite
...
     ///
-    private let order: Order
+    private struct State {
+        /// Order to be refunded
+        ///
+        let order: Order
+
+        /// Current currency settings
+        ///
+        let currencySettings: CurrencySettings
```
jaclync, line 17: I was wondering why `order` and `currencySettings` are part of the `State` while `shouldRefundShipping` is the only property that could change by the user? Would it make sense if these two properties belong to `IssueRefundViewModel` instead?

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -0,0 +1,54 @@
+import XCTest
+import Yosemite
+
+@testable import WooCommerce
+
+/// Test cases for `IssueRefundViewModel`
+///
+final class IssueRefundViewModelTests: XCTestCase {
+
+    func test_viewModel_does_not_have_shippping_section_on_order_without_shipping() {
```
jaclync:
```suggestion
    func test_viewModel_does_not_have_shipping_section_on_order_without_shipping() {
```

jaclync (approved):
Exciting to see UX now!

Code: ✅ some suggestions for your consideration
Testing: ✅ 

order with virtual product | refund shipping off | refund shipping on
-- | -- | --
Simulator Screen Shot - iPhone 11 - 2020-10-09 at 18 17 52 | Simulator Screen Shot - iPhone 11 - 2020-10-09 at 18 18 11 | Simulator Screen Shot - iPhone 11 - 2020-10-09 at 18 18 17

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
Ecarrion: I added them in the `state` struct because they are the immutable part of the state, and they are needed together with `shouldRefundShipping` to compute the properties that the view model uses to render the screen. 

I'm trying to have those two concerns separated. In the state struct, the underlying data. In the VM the data that is going to be presented to the user.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/Cells/RefundShippingDetailsViewModel.swift
Ecarrion: Nice call!

PR #2948, opened 2020-10-09
Orders tab: DI site ID to `OrdersRootViewController ` and its downstream classes

jaclync:
Part of #1838 

## Changes

- DI'ed `siteID: Int64` to `OrdersRootViewController` and its related classes, including refund, search, and shipping. Also removed unnecessary observation for `defaultAccountWasUpdated` notification
- Updated unit tests

## Testing

For confidence check on the Orders tab:

(I tested tapping a new order push notification)

Prerequisite: the site has at least one refunded order

- Log out from the app
- Log in to a site
- Go to the Orders tab --> the orders should load on both sub-tabs as before
- Tap on the search icon --> it should display the order statuses available in the selected site
- Tap on the Refunded order status, and tap on a refunded order --> the refund UI should be shown as before
- Tap on a "Refunded" row that contains at least a product --> the product should be shown in the refund details as before
- Switch to another site and repeat the steps above, making sure the Orders tab only shows data from the selected site

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
@@ -8,6 +8,7 @@ final class OrderDetailsResultsControllers {
     private let storageManager: StorageManagerType
 
     private let order: Order
+    private let siteID: Int64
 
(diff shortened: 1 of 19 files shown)


Ecarrion (approved):
Thanks for pushing DI forward! 🙇

PR #2953, opened 2020-10-12
Products tab: DI site ID to `ProductsViewController` and its downstream classes

jaclync:
Part of #1838 

## Changes

- DI'ed `siteID: Int64` to `OrdersRootViewController` and its related classes, including refund, search, and shipping. Also removed unnecessary observation for `defaultAccountWasUpdated` notification
- Updated unit tests

## Testing

For confidence check on the Products tab:

- Log out from the app
- Log in to a site
- Go to the Products tab --> the products should load as before
- Go to the Dashboard tab, and tap settings to switch to a different store
- Go to the Products tab --> the products of the selected site should load
- Tap on the search icon and search for some products --> the search results should load as before
- Tap on a product with reviews
- Tap on the reviews row --> the product reviews should load as before
- Add a category to the product, then tap "Update" --> the product should be saved remotely with the new category
- Go back to the Products tab, and tap "+" in the navigation bar
- Fill in some details, and tap "Publish" --> the product should be published remotely

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
+++ b/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
@@ -375,7 +375,7 @@ private extension MainTabBarController {
         let ordersViewController = createOrdersViewController(siteID: siteID)
         ordersNavigationController.viewControllers = [ordersViewController]
 
-        let productsViewController = createProductsViewController()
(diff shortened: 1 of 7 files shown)


Ecarrion (approved):
Looks great! 🚀

PR #2955, opened 2020-10-12
Remove `ProductFormActionsFactory` unit tests when Products M3 feature flag is off

jaclync:
Part of #2954

## Changes

This PR just removed the outdated unit tests on `ProductFormActionsFactory` that have Products M3 feature flag off, to make #2931 easier (since it touches a bunch of `ProductFormEditAction` cases). The rest of Products M3 feature flag removal will be in a separate PR.
- Replaced `ProductFormActionsFactoryTests` with the original `ProductFormActionsFactory+EditProductsM3Tests` where Products M3 feature flag is enabled

## Testing

Just CI! All changes are in the WooCommerce tests target

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #2956, opened 2020-10-12
Add `ProductFormType.readonly` and `editable` flag to product form rows

jaclync:
Part of #2931 
Based on https://github.com/woocommerce/woocommerce-ios/pull/2955

This PR is not intended to introduce user-facing changes. The end goal is add readonly mode to `ProductFormViewController` (a few rows already support readonly mode), and I'd like to make sure nothing breaks with the introduction of `editable` flag and the implementation looks good. More customizations for full readonly support will be in the next PR.

## Changes

- Added `ProductFormType.readonly` case: this case will be used in the next PR (it's actually used in the soon to be deprecated `ProductDetailsViewModel` when navigating to the variation list. however, this scenario isn't possible because an order can only have a "variation" not a variable product)
- Added `editable: Bool` associated value to `ProductFormEditAction` cases that can be editable or readonly
- Updated `ProductFormActionsFactory` and `ProductVariationFormActionsFactory` to determine whether the form is editable by its form type (not `ProductFormType.readonly`)
- Updated how each row is rendered with the new `editable` flag:
  - Added `isEditable: Bool` associated value to `ProductFormSection.SettingsRow` cases that can be editable or readonly
  - Added `isEditable: Bool` associated value to `ProductFormSection.PrimaryFieldRow` cases that can be editable or readonly
  - Updated `DefaultProductFormTableViewModel` and `ProductFormTableViewDataSource` given the `isEditable: Bool` changes
- In `ProductFormViewController`, implemented `ProductFormType.readonly` case: no right navigation bar button items, and no-op when the user navigates away from the form
- Updated unit tests: mostly just adding `(editable: true)` to the existing `ProductFormEditAction` asserts except for test cases on a non-core product (`test_actions_for_non_core_product_with_price` and `test_actions_for_non_core_product_without_price` where some rows are readonly). Unit tests for readonly mode will be in a separate PR.

## Testing

Please confidence check on product editing:

### Simple product
- Go to the Products tab
- Tap on a simple product --> make sure each field is editable, and the ellipsis menu includes product settings

### Variable product
- Go to the Products tab
- Tap on a variable product
- Tap on the variations row
- Tap on a variation --> make sure the variation is editable like before

### Add product
- Go to the Products tab
- Tap on "+" CTA in the navigation bar
- Tap "Simple product" --> make sure each field is editable, and the ellipsis menu includes product settings

Update release notes:


pmusolino (approved):
I like that we will have the read-only version of the new product screen 👏 

Everything works like before and the code looks good to me 💯 

I just noticed one thing that probably shouldn't be in the "Add Simple Product" screen. I don't think that we should show the `Reviews` row since it will be always empty when a product will be added.

Let me know if I should create a new issue for that 😉

jaclync:
> I just noticed one thing that probably shouldn't be in the "Add Simple Product" screen. I don't think that we should show the Reviews row since it will be always empty when a product will be added.

Good question! I originally listed the reviews row to be removed for product creation in the differences post (p91TBi-3f7-p2), then removed it based on G's comment (p91TBi-3f7#comment-1650-p2) - I think it's still nice showing the "no reviews" row if product reviews are enabled in the settings, so that they know they can start seeing reviews later on. Feel free to share more of your thoughts on the p2 post with G!

PR #2958, opened 2020-10-12
Refund Summary → Calculate Total Refunded

shiki:
Ref #2768. 

This adds calculation for the “Previously Refunded” row in the `RefundConfirmationViewModel`. 

 
It's not a fake number anymore. 🤸 

## Changes

I added a new `UseCase` for the calculation, [`TotalRefundedCalculationUseCase`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2768-add-refund-total/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Issue%20Refunds/TotalRefundedCalculationUseCase.swift), and used that in the `ViewModel`. 

Side note: I sometimes wonder if this should be in Yosemite since it's business logic. 🤷 But that requires us to transfer `CurrencySettings` and `CurrencyFormatter` to Yosemite too. 

### Net Calculation

The Net calculation also has the total refunded logic so I replaced that with the `UseCase` as well. 

## Testing

1. Navigate to an order that has more than one refund items. 
2. In the Order Details, confirm that the Net value is still the same as what you would see in `develop`. 
3. Tap on Issue Refund. 
4. Tap Next. 
5. Confirm that the Previously Refunded field shows the sum of all the refund totals. 

## Submitter Checklist


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/TotalRefundedCalculationUseCase.swift
```diff
@@ -0,0 +1,26 @@
...
+    private let locale: Locale
+
+    init(order: Order, currencyFormatter: CurrencyFormatter, locale: Locale = .current) {
+        self.order = order
+        self.currencyFormatter = currencyFormatter
+        self.locale = locale
+    }
+
+    /// Sums up the `OrderRefundCondensed.total` inside `self.order`. This is returned as a
+    /// **negative** number because that's how the API returns them as.
```
jaclync, line 19: the comment on the negative number is helpful!

WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
```diff
@@ -175,17 +175,8 @@ final class OrderPaymentDetailsViewModel {
...
-
-        // Can't use .reduce(0,+) here because we're working with NSDecimalNumber.
-        var refundTotal: NSDecimalNumber = .zero
-        for r in refunds {
-            if let refund = r {
-                refundTotal = refundTotal.adding(refund)
-            }
-        }
+        let totalRefundedUseCase = TotalRefundedCalculationUseCase(order: order, currencyFormatter: currencyFormatter)
+        let refundTotal = totalRefundedUseCase.totalRefunded()
```
jaclync, line 179: nice that this logic is in a use case that is unit tested, with `reduce` :)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundConfirmationViewModel.swift
```diff
@@ -7,32 +8,47 @@ final class RefundConfirmationViewModel {
...
-            ),
-            Section(
-                title: Localization.refundVia,
-                rows: [
-                    TitleAndBodyRow(title: Localization.manualRefund(via: "Stripe"),
-                                    body: Localization.refundWillNotBeIssued(paymentMethod: "Stripe"))
-                ]
-            )
-        ]
+    lazy private(set) var sections: [Section] = [
```
jaclync, line 15: it's a variable just because of the usage of `makePreviouslyRefundedRow` right? I wish there's a way to call a function before self.init

jaclync (approved):
Looks good! :shipit: 

> Side note: I sometimes wonder if this should be in Yosemite since it's business logic. 🤷 But that requires us to transfer `CurrencySettings` and `CurrencyFormatter` to Yosemite too.

yea, I can see `CurrencySettings` and `CurrencyFormatter` in Yosemite as they are business logic as well!


Code: ✅ 
Testing: ✅ 

payment details | previously refunded
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-13 at 18 11 26 | Simulator Screen Shot - iPhone 11 - 2020-10-13 at 18 11 29

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundConfirmationViewModel.swift
shiki: That's correct. I _had_ to do it. (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

Issue #2888, opened 2020-09-29
Improvements for app navigation based on logged out / logged in state


jaclync:
Followup tasks from https://github.com/woocommerce/woocommerce-ios/pull/2886

- [x] Refactor the app navigation to a coordinator (maybe not a subclass of `Coordinator` because the main window is using a `UITabBarController` instead of a `UINavigationController`). I have some WIP code for this, would like to add some unit tests. https://github.com/woocommerce/woocommerce-ios/pull/2963
- [ ] There is a glitch when the app is previously logged in but the user hasn't selected a site, and we present a store picker on launch. Because the main tab bar isn't set up yet since the app is still logged out, the store picker is presented from a tab bar with a black screen for a split second. I tried updating its presentation to try presenting from the tab bar controller, but `StorePickerCoordinator` currently takes in a `UINavigationController` only and it uses the tab bar's `selecteViewController` right now 🤨 (feels a bit hacky to me to grab a `UINavigationController` from the tab bar's selected tab) I think this is not a major flow, but I'd like to fix it sometime soon.
- [x] An issue with store picker: https://github.com/woocommerce/woocommerce-ios/issues/2888#issuecomment-701602846 - fixed in https://github.com/woocommerce/woocommerce-ios/pull/3498
- [ ] Log in with credentials, force close the app, then restart login flow --> there is a glitch when it animates to the login UI, it requires changes in the store picker

shiki:
I'm thinking the `Coordinator` that we also have is also a bit too rigid because it requires a `UINavigationController`. Perhaps that can be taken out and be something more generic. 

```swift
protocol Coordinator {
    var rootViewController: UIViewController { get }
}
```

shiki:
In relation to the Store Picker, I also found another bug. 

1. Log in and stop when you reach the store picker. 
2. Force close the app. 
3. Open the app again. The store picker will be shown. 
4. Pick a store and continue with the login process. 

The app will show the dashboard but immediately show the login page again. This happens in `develop` (3b29761b4).

jaclync:
Closing - the issue is more than 5 years old, and high priority issues should already have been reported separately.

PR #2963, opened 2020-10-13
Consolidate auth-based navigation to `AppCoordinator`

jaclync:
Part of #2888

## Why

Before this PR, the authentication navigation code is scattered in different places:
- When logging out, the login UI is triggered by `DeauthenticatedState`'s `didEnter`
- When the app is launched in logged out state, or the app is logged in but requires a selected site, the logic to show the UI in these cases are in `AppDelegate`. We also want to keep `AppDelegate` lean and only have high-level app setup logic there.

This PR aims to consolidate all the authentication based navigation to one place (`AppCoordinator`). 

## Changes

- Created `AppCoordinator` that takes in `MainTabBarController` and dependencies from the `ServiceLocator`.
- In `AppDelegate`, replaced existing auth navigation code with `AppCoordinator`.
- Added/updated unit tests (there were quite some mysterious tests failing that took me a while to fix 😰 ). I had to [remove some asserts](https://github.com/woocommerce/woocommerce-ios/commit/65dc48d9b267a42fb9a0fd7e6b250a11c43c7af4) because I couldn't reproduce the [test failures on CI](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/8101/workflows/32a8139d-ba51-4a9c-8510-889ba3c16817/jobs/13647) (I tried reordering the tests as in CI).

### Known pre-existing issue https://github.com/woocommerce/woocommerce-ios/issues/2888#issuecomment-701602846

This PR preserves the same logic as before, just moving the logic around. I'll fix this in a separate PR.
- Log in and stop when you reach the store picker.
- Force close the app (stopping or relaunching the app from Xcode, since [closing the app triggers deauthentication](https://github.com/woocommerce/woocommerce-ios/blob/20db5f84fbb75e3bda6e08179c3aff5b0860d321/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift#L273-L277))
- Open the app again. The store picker will be shown.
- Pick a store and continue with the login process. --> the app will show the dashboard but immediately show the login page again. The app should stay in logged in state.

Please confidence check on the authentication flow:

- Log out the app
- Log in to the app --> the auth flow should work as before
- Relaunch the app --> the app should stay logged in

Update release notes:

### DIFF
--- a/WooCommerce/Classes/AppDelegate.swift
+++ b/WooCommerce/Classes/AppDelegate.swift
@@ -29,15 +29,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     ///
     var window: UIWindow?
 
+    /// Coordinates app navigation based on authentication state.
+    ///
+    private var appCoordinator: AppCoordinator?
+
     /// Tab Bar Controller
     ///
     var tabBarController: MainTabBarController? {
         return window?.rootViewController as? MainTabBarController
     }
-
-    /// Store Picker Coordinator
-    ///
-    private var storePickerCoordinator: StorePickerCoordinator?
 
     /// Checks on whether the Apple ID credential is valid when the app is logged in and becomes active.
     ///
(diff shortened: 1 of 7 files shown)


Ecarrion (approved):
Looks good to me!
It's nice to see some tests around this area.

👏

Issue #2967, opened 2020-10-13
[Issue Refunds] Handle previous refunds


Ecarrion:
When refunding an order we need to make sure that.

- Do not show the "refund shipping" switch if shipping was already refunded.
- Take into account already refunded items to display the proper max quantities to refund.

shiki:
This is the REST API issue we're waiting for to identify if shipping costs were already refunded. https://github.com/woocommerce/woocommerce/issues/27566

Ecarrion:
PR https://github.com/woocommerce/woocommerce/pull/28241

PR #2968, opened 2020-10-13
[Issue Refund] Selected Items count label

### LINKED ISSUE #2840: [Issue Refunds] Navigation and header handling
- Update Navigation
- Update item selection 
- Select All button

Screen Shot 2020-09-20 at 11 45 04 AM

Ecarrion:
closes #2840 

# Why

This PR takes care of two things:
- Updating the selected item label text to reflect the current selected state
- Enabling the "select all" button.

# How
- Added a `selectAllOrderItems` method on `IssueRefundViewModel` to select all orders items.
- Added a `createSelectedItemsCount` method on `IssueRefundViewModel` to update the selected items label text. 
- Added a `count` method to `RefundQuantityStore` to know how many items are marked for refund.

# Demo
demo

# Testing steps
- Navigate to a non-refunded order
- Tap the **issue refund** button
- See that the label reads: "0 items selected"
- Select one item to refund
- See that the label reads: "1 item selected"
- Tap on the "select all" button
- See that the label reads correctly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
@@ -229,7 +229,7 @@ extension IssueRefundViewModel {
         return productsTotalUseCase.calculateRefundValues().total + shippingTotalUseCase.calculateRefundValue()
     }
 
-    /// Returns a string with the count ofhow many items are selected for refund.
+    /// Returns a string with the count of how many items are selected for refund.
     ///
     private func createSelectedItemsCount() -> String {
         let count = state.refundQuantityStore.count()
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
@@ -31,6 +31,7 @@ final class IssueRefundViewModel {
         didSet {
             sections = createSections()
             title = calculateTitle()
+            selectedItemsTitle = createSelectedItemsCount()
             onChange?()
         }
     }
@@ -44,9 +45,8 @@ final class IssueRefundViewModel {
     private(set) var title: String = ""
 
     /// String indicating how many items the user has selected to refund
-    /// This is temporary data, will be removed after implementing https://github.com/woocommerce/woocommerce-ios/issues/2842
     ///
-    let selectedItemsTitle: String = "0 items selected"
+    private(set) var selectedItemsTitle: String = ""
 
     /// The sections and rows to display in the `UITableView`.
     ///
@@ -64,6 +64,7 @@ final class IssueRefundViewModel {
         state = State(order: order, currencySettings: currencySettings)
         sections = createSections()
         title = calculateTitle()
+        selectedItemsTitle = createSelectedItemsCount()
     }
 
     /// Creates the `ViewModel` to be used when navigating to the page where the user can
@@ -136,6 +137,8 @@ private extension IssueRefundViewModel {
 private extension IssueRefundViewModel {
     enum Localization {
         static let refundShippingTitle = NSLocalizedString("Refund Shipping", comment: "Title of the switch in the IssueRefund screen to refund shipping")
+        static let itemSingular = NSLocalizedString("1 item selected", comment: "Title of the label indicating that there is 1 item to refund.")
+        static let itemsPlural = NSLocalizedString("%d items selected", comment: "Title of the label indicating that there are multiple items to refund.")
     }
 }
 
@@ -225,6 +228,13 @@ extension IssueRefundViewModel {
         let shippingTotalUseCase = RefundShippingCalculationUseCase(shippingLine: shippingLine, currencyFormatter: formatter)
         return productsTotalUseCase.calculateRefundValues().total + shippingTotalUseCase.calculateRefundValue()
     }
+
+    /// Returns the a string with the count of many items are selected for refund.
+    ///
+    private func createSelectedItemsCount() -> String {
+        let count = state.refundQuantityStore.count()
+        return String.pluralize(count, singular: Localization.itemSingular, plural: Localization.itemsPlural)
+    }
 }
 
 extension RefundItemViewModel: IssueRefundRow {}
@@ -261,6 +271,13 @@ private extension IssueRefundViewModel {
         ///
         func map<T>(transform: (_ item: OrderItem, _ quantity: Quantity) -> (T)) -> [T] {
             store.map(transform)
+        }
+
+        /// Returns the number of items referenced for refund.
+        /// Calculated by aggregating all stored quantities.
+        ///
+        func count() -> Int {
+            store.values.reduce(0) { $0 + $1 }
         }
     }
 }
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -109,6 +110,15 @@ extension IssueRefundViewModel {
         }
         state.refundQuantityStore.update(quantity: quantity, for: item)
     }
+
+    /// Marks all items as to be refunded
+    ///
+    func selectAllOrderItems() {
+        state.order.items.forEach { item in
+            let quantity = Int(truncating: item.quantity as NSDecimalNumber)
```
jaclync, line 118: Non-blocking thought: this reminds me of an edge case where the store supports decimal quantity for orders (I tested this from a support issue recently). For example, I just tested refunding an item with quantity 0.5 out of 2 in an order:

Screen Shot 2020-10-15 at 4 47 24 PM

Then I checked the app, and I was able to select up to 2 of this product to refund (1.5 left). If I submit this refund, the server will probably return an error (which is fine as well). I was wondering if using the `floor` of `item.quantity` would be safer? Also, as an improvement we can update "Select All" to include the decimal quantity so that the user can at least refund all of the remaining.

This requires changing `RefundQuantityStore.Quantity` to double and that might complicate things though, feel free to keep the quantity as integer for the refund project! The support for decimal quantity is certainly out of scope :)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -127,6 +137,8 @@ private extension IssueRefundViewModel {
 private extension IssueRefundViewModel {
     enum Localization {
         static let refundShippingTitle = NSLocalizedString("Refund Shipping", comment: "Title of the switch in the IssueRefund screen to refund shipping")
+        static let itemSingular = NSLocalizedString("1 item selected", comment: "Title of the label indicating that there is 1 item to refund.")
+        static let itemsPlural = NSLocalizedString("%d items selected", comment: "Title of the label indicating that there are multiple items to refund.")
```
jaclync, line 141: wish we had strings dict support for singular/plural soon!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -216,6 +228,13 @@ extension IssueRefundViewModel {
         let shippingTotalUseCase = RefundShippingCalculationUseCase(shippingLine: shippingLine, currencyFormatter: formatter)
         return productsTotalUseCase.calculateRefundValues().total + shippingTotalUseCase.calculateRefundValue()
     }
+
+    /// Returns a string with the count ofhow many items are selected for refund.
```

```suggestion
    /// Returns a string with the count of how many items are selected for refund.
```

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -158,4 +179,60 @@ final class IssueRefundViewModelTests: XCTestCase {
...
+            MockOrderItem.sampleItem(itemID: 2, quantity: 2, price: 12.50),
+            MockOrderItem.sampleItem(itemID: 3, quantity: 1, price: 13.50),
+        ]
+        let order = MockOrders().makeOrder(items: items)
+
+        // When
+        let viewModel = IssueRefundViewModel(order: order, currencySettings: currencySettings)
+
+        // Then
+        let selectedItemsTitle = String(format: NSLocalizedString("%d items selected", comment: ""), 0)
```
jaclync, line 197: nit: I think it's reasonable to make `IssueRefundViewModel.Localization` internal instead of private so that we can refer to the Localization constants in unit tests in case they change in the future? (Same comment for other `NSLocalizedString`s in this test class)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -253,5 +272,12 @@ private extension IssueRefundViewModel {
         func map<T>(transform: (_ item: OrderItem, _ quantity: Quantity) -> (T)) -> [T] {
             store.map(transform)
         }
+
+        /// Returns the number of items referenced for refund.
+        /// Calculated by aggregating all stored quantities.
+        ///
+        func count() -> Int {
+            store.values.reduce(0) { $0 + $1 }
```
jaclync, line 280: nice and concise!

jaclync (approved):
LGTM! :shipit: 

Code: ✅ 
Testing: ✅ 

Just wanted to mention an issue on a partially refunded order that you probably have planned to fix later:

If an order has been partially refunded, the quantity on Issue Refund screen is still showing the original quantity instead of the remaining refundable quantity:

(The order originally has quantity 2 for both products)

order details with remaining quantity | issue refund with original quantity
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-15 at 17 06 29 | Simulator Screen Shot - iPhone 11 - 2020-10-15 at 17 06 57

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
Ecarrion: Hmm, I normally don't like to expose a type's implementation details just for testing but I also don't like to write twice a localized string 🤔 

Maybe we should make a small vote within the team and decide what's preferred?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
Ecarrion: Ohhh thanks a lot for bringing that case into my eyes. I have added a [small task](https://github.com/woocommerce/woocommerce-ios/projects/17#card-47428684 ) to corroborate work consistent with android and don't break anything. 

🙇

Ecarrion:
Thanks @jaclync for the review!

> Just wanted to mention an issue on a partially refunded order that you probably have planned to fix later:

> If an order has been partially refunded, the quantity on Issue Refund screen is still showing the original quantity instead of the remaining refundable quantity:

Yes, I have [this issue](https://github.com/woocommerce/woocommerce-ios/issues/2967) to track that and this functionality will come in the next PR!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -216,6 +228,13 @@ extension IssueRefundViewModel {
         let shippingTotalUseCase = RefundShippingCalculationUseCase(shippingLine: shippingLine, currencyFormatter: formatter)
         return productsTotalUseCase.calculateRefundValues().total + shippingTotalUseCase.calculateRefundValue()
     }
+
+    /// Returns a string with the count of how many items are selected for refund.
+    ///
+    private func createSelectedItemsCount() -> String {
+        let count = state.refundQuantityStore.count()
+        return String.pluralize(count, singular: Localization.itemSingular, plural: Localization.itemsPlural)
+    }
```
shiki, line 237: Another way is to move this to `Localization` instead since it's a logic that we shouldn't really need to care about if only we supported Strings dict. 

```swift
enum Localization {
    static func selectedItemsCountLabel(_ count: Int) -> String {
        String.pluralize(count, singular: Localization.itemSingular, plural: Localization.itemsPlural)
    }
}
```

PR #2976, opened 2020-10-14
Show unsaved changes alert when the user drags to dismiss the product selector search modal

### LINKED ISSUE #2561: Show unsaved changes alert when the user drags to dismiss the product selector search modal
Original discussion: https://github.com/woocommerce/woocommerce-ios/pull/2541#pullrequestreview-455135957

Since a modal in iOS 13 can be dismissed by dragging down the modal, it'd be nice if we show the unsaved changes prompt when the user dismisses the product selector search modal with `UIAdaptivePresentationControllerDelegate`:

adaptive-dismiss

jaclync:
Fixes #2561 

## Changes

- Updated `ProductListMultiSelectorSearchUICommand` to implement `UIAdaptivePresentationControllerDelegate` so that we can show the discard changes action sheet when the user has selected at least one product from the search modal

### Dismiss search modal without any selected products

- Launch the app on an iOS 13+ simulator/device
- Go to the Products tab
- Tap on a grouped product
- Tap "Add Products"
- Tap on the search icon
- Swipe to dismiss the modal --> the search modal should be dismissed right away

### Dismiss search modal with selected products

- Continue from the last step above
- Tap on the search icon
- Tap to select at least one product
- Swipe to dismiss the modal --> a discard changes action sheet should be presented
- Tap "Cancel" or outside of the action sheet --> the action sheet should be dismissed and it stays on the search modal
- Swipe to dismiss the modal --> a discard changes action sheet should be presented
- Tap "Discard changes" --> the search modal should be dismissed

## Example screenshots

iPhone:

simulator

iPad:

simulator

Update release notes:


Garance91540 (approved):
👏

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Grouped Products/ProductListMultiSelectorSearchUICommand.swift
```diff
@@ -92,13 +95,43 @@ final class ProductListMultiSelectorSearchUICommand: SearchUICommand {
...
+        }
+        guard hasUnsavedChanges() else {
+            presentingViewController.dismiss(animated: true)
+            return
+        }
+        UIAlertController.presentDiscardChangesActionSheet(viewController: presentingViewController,
+                                                           onDiscard: { [weak self] in
+                                                            self?.presentingViewController?.dismiss(animated: true)
+        })
+    }
```
Ecarrion: While I was testing this something in the drag gesture felt off when dismissing a view controller that does not have any pending changes. And it looks like this is because we are prohibiting all gestures from the start and then deciding what to do.

In order to have the same experience when dismissing a `non-dirty` view we could do this

```suggestion
    func presentationControllerShouldDismiss(_ presentationController: UIPresentationController) -> Bool {
        return !hasUnsavedChanges()
    }

    func presentationControllerDidAttemptToDismiss(_ presentationController: UIPresentationController) {
        UIAlertController.presentDiscardChangesActionSheet(viewController: presentationController.presentedViewController, onDiscard: {
            presentationController.presentedViewController.dismiss(animated: true)
        })
    }
```

Here the `presentationControllerShouldDismiss` method controls if the controller should be dismissed or not.
and `presentationControllerDidAttemptToDismiss` is only called when the user attempted to dismiss it when we didn't allow it.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Grouped Products/ProductListSelectorViewController.swift
```diff
@@ -71,7 +71,12 @@ private extension ProductListSelectorViewController {
         let searchViewController = SearchViewController(storeID: siteID,
                                                         command: searchProductsCommand,
                                                         cellType: ProductsTabProductTableViewCell.self)
+        // Disables interactive dismiss action so that we can prompt the discard changes alert.
+        if #available(iOS 13.0, *) {
```
Ecarrion: With the previous change, I think this won't be necessary.

Ecarrion (commented):
Thanks for tackling this @jaclync 🙇 

I left you a small suggestion that could make the experience better, let me know what you think.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Grouped Products/ProductListSelectorViewController.swift
jaclync: yup! removed in https://github.com/woocommerce/woocommerce-ios/pull/2976/commits/2ad8c7d871dabfe25364030026e3159f250473de

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Grouped Products/ProductListMultiSelectorSearchUICommand.swift
jaclync: oooh thanks for the suggestions! this feels like a much better experience, updated in https://github.com/woocommerce/woocommerce-ios/pull/2976/commits/2ad8c7d871dabfe25364030026e3159f250473de and https://github.com/woocommerce/woocommerce-ios/pull/2976/commits/ce115266d1acd8327ced3274576a8247b1658424

jaclync:
thank you for the suggestion @Ecarrion! I updated the PR for another look 😃

Ecarrion (approved):
Works great! 🚀

PR #2978, opened 2020-10-14
Update Download File Settings for a specific Product

### LINKED ISSUE #2756: Add/Edit Products M5: Downloadable files Step 2 - update downloadable file info for a product
As part of the M5 release, we need to provide the option to:
- Tap on a downloadable file from the list.
- Update the downloadable file details such as file url, download limit and download expiry. We would also need to add the option to update the downloadable file details in Networking/Storage/Yosemite.

| Select a file             |  Edit the file screen |
iOS-add-price | iOS-add-price copia

- Edit the downloadable file settings.

| Select more button in downloadable file list             |  Download Settings |
iOS-add-price | iOS-product-settings

**Note that this issue only takes care of updating existing downloadable files for a product. The option to add new files will be taken care of in another issue and is not part of the scope of this issue.**

pmusolino:
Fixes: #2756

## Description
This PR allows updating the download file settings for a specific product.
In particular, it's possible to add or edit downloadable file limit and expiry in a separate scene when tapped from the downloadable file list navigation right bar button item.
The work was started by @partho-maple but I made some changes to some logics and cleaned the code to conform to our code style.

## Changes
- Changed the location of some files since they were added under the wrong path initially.
- Updated `ProductDownloadListViewController` that now can present an action sheet when tapping on the More Bar Button item. It also can show the new screen for configuring the Download Settings for a product.
- New `Product+DownloadSettingsViewModels` used for creating the `TitleAndTextFieldTableViewCell` view models.
- Added the new `ProductDownloadSettingsViewController` that shows the download file settings.
- Implemented the `ProductDownloadSettingsViewModel` that provides data needed for download settings + added unit tests.

#### Prerequisites

- Create a downloadable product from wp-admin.
- Then navigate to the downloadable files' list scene on the app.

#### Cases

1. Tap on any downloadable file's settings button on the right navigation bar button item.
2. Change the Limit and/or Expiry, and tap done. It should update the settings.
3. Done button should only be enabled if and only if there is any change/update on the Limit and/or Expiry value.
4. Removing any of the values should reset Limit and/or Expiry value to default.

## Screenshots
| Dark mode OFF            |  Dark mode ON |
Simulator Screen Shot - iPhone 11 Pro - 2020-10-14 at 17 51 54 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-14 at 17 52 16

Update release notes:


Garance91540 (approved):
👏

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/File List/ProductDownloadListViewController.swift
```diff
@@ -190,14 +190,45 @@ extension ProductDownloadListViewController {
...
+                                                      comment: "Button title Download Settings in Downloadable Files More Options Action Sheet")
+        let downloadSettingsAction = UIAlertAction(title: downloadSettingsTitle, style: .default) { [weak self] (action) in
+            self?.showDownloadableFilesSettings()
+        }
+        menuAlert.addAction(downloadSettingsAction)
+
+        let cancelTitle = NSLocalizedString("Cancel",
+                                            comment: "Button title Cancel in Downloadable Files More Options Action Sheet")
+        let cancelAction = UIAlertAction(title: cancelTitle, style: .cancel) { (action) in
+        }
```
jaclync: nit:

```suggestion
        let cancelAction = UIAlertAction(title: cancelTitle, style: .cancel)
```

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
```diff
@@ -3439,12 +3461,13 @@
 			isa = PBXGroup;
 			children = (
 				773077E7251E943600178696 /* Add or Edit File */,
-				77E53EC62510C3F4003D385F /* FIle List */,
+				77E53EC62510C3F4003D385F /* File List */,
```
jaclync, line 3464: nice catch!

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewController.swift
```diff
@@ -0,0 +1,260 @@
...
+
+    func configureNavigationBar() {
+        title = NSLocalizedString("Download Settings",
+                                  comment: "Download file settings navigation title")
+
+        navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .done,
+                                                            target: self,
+                                                            action: #selector(completeUpdating))
+        removeNavigationBackBarButtonText()
+        self.enableDoneButton(false)
```
jaclync: nit:
```suggestion
        enableDoneButton(false)
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+        // We consider valid values: -1, positive numbers and if at least one new value is different from the previous one
+        let downloadLimitIsValid = downloadLimit == -1 || downloadLimit >= 0
+        let downloadExpiryIsValid = downloadExpiry == -1 || downloadExpiry >= 0
+        let isDownloadLimitOrDownloadExpiryChanged = downloadLimit != product.downloadLimit || downloadExpiry != product.downloadExpiry
+        return downloadLimitIsValid && downloadExpiryIsValid && isDownloadLimitOrDownloadExpiryChanged
+    }
+}
+
+extension ProductDownloadSettingsViewModel {
+    enum Strings {
```
jaclync: nit: we started using `Localization` as in our [style guide](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/naming-conventions.md#string-constants-in-nested-enums)

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+
+    func hasUnsavedChanges() -> Bool {
+        return downloadLimit != product.downloadLimit || downloadExpiry != product.downloadExpiry
+    }
+}
+
+// MARK: - Convenience Methods
+//
+private extension ProductDownloadSettingsViewModel {
+    func isChangesValid() -> Bool {
```
jaclync: nit:
```suggestion
    func areChangesValid() -> Bool {
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+extension ProductDownloadSettingsViewModel: ProductDownloadSettingsActionHandler {
+    func handleDownloadLimitChange(_ downloadLimit: String?, onValidation: @escaping (_ isValid: Bool, _ shouldBringUpKeyboard: Bool) -> Void) {
+
+        if downloadLimit?.isEmpty == true {
+            self.downloadLimit = -1
+            onValidation(isChangesValid(), true)
+            return
+        }
+
+        guard let downloadLimit = downloadLimit, let downloadLimit_unwrapped = Int64(downloadLimit), downloadLimit_unwrapped >= 0 else {
```
jaclync: nit: the use of underscore in `downloadLimit_unwrapped` is quite uncommon in our codebase - how about `downloadLimitValue` or `downloadLimitUnwrapped`? or `guard let downloadLimitString = downloadLimit, let downloadLimit = Int64(downloadLimitString), downloadLimit >= 0`?

(similar comment in `handleDownloadExpiryChange`)

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+        default:
+            fatalError("Unsupported product type: \(product)")
+        }
+    }
+}
+
+// MARK: - UI changes
+//
+extension ProductDownloadSettingsViewModel: ProductDownloadSettingsActionHandler {
+    func handleDownloadLimitChange(_ downloadLimit: String?, onValidation: @escaping (_ isValid: Bool, _ shouldBringUpKeyboard: Bool) -> Void) {
```
jaclync: I'm a bit confused by the use case of `shouldBringUpKeyboard` - we had this in inventory settings because SKU validation is an async operation, but it seems like local validation is all fine here? I tested without handling `shouldBringUpKeyboard` in the callback, and the UX seems fine to me. Maybe we can remove this?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+
+        if downloadLimit?.isEmpty == true {
+            self.downloadLimit = -1
```
jaclync: nit: I was confused by the value of -1 first, then saw that -1 is the default value in `Product.swift`. Maybe this default value -1 could be in a constant so that it's clear that this is the default value? (similar comment in `handleDownloadExpiryChange`)

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+            onValidation(isChangesValid(), true)
+            return
+        }
+
+        guard let downloadLimit = downloadLimit, let downloadLimit_unwrapped = Int64(downloadLimit), downloadLimit_unwrapped >= 0 else {
+            self.downloadLimit = -2
```
jaclync: nit: seeing -2 here is quite confusing to me, later I realized this is to set it to an invalid value. How about making a constant for -2 like `invalidDownloadLimitValue`?  (similar comment in `handleDownloadExpiryChange`)

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
```diff
@@ -0,0 +1,137 @@
...
+        onValidation(isChangesValid(), false)
+    }
+
+    // MARK: - Navigation actions
+
+    func completeUpdating(onCompletion: ProductDownloadSettingsViewController.Completion) {
+        if isChangesValid() {
+            onCompletion(downloadLimit, downloadExpiry, hasUnsavedChanges())
+        }
+        return
```
jaclync: nit: this return doesn't seem necessary?

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/Downloadable Files/ProductDownloadSettingsViewModelTests.swift
```diff
@@ -0,0 +1,259 @@
...
+        // Arrange
+        let product = MockProduct().product(downloadable: true)
+        let model = EditableProductModel(product: product)
+
+        // Act
+        let viewModel = ProductDownloadSettingsViewModel(product: model)
+
+        // Assert
+        XCTAssertEqual(viewModel.downloadLimit, 1)
+        XCTAssertEqual(viewModel.downloadExpiry, 1)
```
jaclync, line 19: nit: I was wondering where these values come from, maybe we can make them more explicit by setting them via `copy(downloadLimit:downloadExpiry:)` in L11 above?

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/Downloadable Files/ProductDownloadSettingsViewModelTests.swift
```diff
@@ -0,0 +1,259 @@
...
+        // Act
+        var isValidResult: Bool?
+        var shouldBringUpKeyboard: Bool?
+        waitForExpectation { exp in
+            viewModel.handleDownloadLimitChange("") { (isValid, shouldBringUpKeyboardValue) in
+                isValidResult = isValid
+                shouldBringUpKeyboard = shouldBringUpKeyboardValue
+                exp.fulfill()
+            }
+        }
```
jaclync: nit: I'd suggest using `let result = try waitFor { promise in ... }` which I find it simpler (no need to unwrap it!) - there are some examples in the codebase, Jayson first shared this great test helper to me :)

jaclync (commented):
Thanks for working on this, Paolo and Partho! The settings work nicely, just shared some comments (mostly the -1/-2 values and `shouldBringUpKeyboard`, the rest are just nit suggestions)

Code: 💬 
Testing: ✅ working great, I was able to see the values in core

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
pmusolino: Updated 👍

WooCommerce/Classes/ViewRelated/Products/Edit Product/Downloadable Files/Settings/ProductDownloadSettingsViewModel.swift
pmusolino: You are right @jaclync, in that specific case we handle an async operation and hide the keyboard if we should show an error (reloading the tableview). I'm going to remove it now

pmusolino:
Thank you @jaclync and @Garance91540 for your review! I made all the changes and I'm going to merge it 😉

PR #2979, opened 2020-10-14
Accept Reason for Refund

### LINKED ISSUE #2768: [Issue Refunds] Refund Summary View

shiki:
Closes #2768. 

This makes the Reason for Refund text field value to be accessible inside the [`RefunndCofirmationViewModel`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Issue%20Refunds/RefundConfirmationViewModel.swift). 

To support this, I made the existing `TitleAndEditableValueTableViewCell` more reusable by requiring an observable `ViewModel` for it, [`TitleAndEditableValueTableViewCellViewModel`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2768-accept-reason-for-refund/WooCommerce/Classes/ViewRelated/ReusableViews/TitleAndEditableValueTableViewCell/TitleAndEditableValueTableViewCell.swift). It is this `CellViewModel` that is accessed by `RefundConfirmationViewModel`. And there are **no wiring necessary** in the `ViewController`. 

The data flow looks like this:

image

### Submission

I made a dummy submission, for now, to prove that the value is accessible. 

https://github.com/woocommerce/woocommerce-ios/blob/0394ee2738e6f0d1ed735dd823b668fb4de155cb/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Issue%20Refunds/RefundConfirmationViewModel.swift#L47-L50

1. Navigate to an Order. 
2. Tap Issue Refund. 
3. Tap Next. 
4. Enter something in the Reason for Refund text field. 
4. Tap Refund.
5. Confirm that the value of the text field is printed out in the console. 

### Manual Tracking.

Please do a quick test that you can still add manual tracking. You would need the [Shipment Tracking](https://woocommerce.com/products/shipment-tracking/) plugin for this. 

1. Navigate to an Order. 
2. Tap on Add Tracking. 
3. Tap the right arrow o the Shipping carrier. 
4. On the list that shows up, tap on Custom Carrier. 
5. Add a carrier name, tracking number, and tracking link. 
6. Tap on the Add button located at the top right. 
7. You should be navigated back to the Order Details. Confirm that you see the custom carrier details that you added in the Order List. 

## Submitter Checklist


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipment Tracking Section/Add Tracking/ManualTrackingViewController.swift
```diff
@@ -230,58 +230,66 @@ extension ManualTrackingViewController: UITableViewDataSource {
...
-        cell.value.addTarget(self, action: #selector(didChangeProviderName), for: .editingChanged)
+        let cellViewModel = TitleAndEditableValueTableViewCellViewModel(
+            title: NSLocalizedString("Carrier name", comment: "Add Custom shipping carrier. Title of cell presenting the carrier name"),
+            placeholder: NSLocalizedString("Enter carrier name", comment: "Add custom shipping carrier. Placeholder of cell presenting carrier name"),
+            initialValue: viewModel.providerName
+        )
+        cell.update(viewModel: cellViewModel)
         cell.accessoryType = .none
+
+        _ = cellViewModel.value.subscribe { [weak self] in
```
Ecarrion, line 252: Don't these subscriptions have to be stored somewhere? 🤔

WooCommerce/Classes/ViewRelated/ReusableViews/TitleAndEditableValueTableViewCell/TitleAndEditableValueTableViewCell.swift
```diff
@@ -2,10 +2,10 @@ import UIKit
...
 final class TitleAndEditableValueTableViewCell: UITableViewCell {
     @IBOutlet private weak var stackView: UIStackView!
-    // TODO Make this private and use `update()` instead.
-    @IBOutlet weak var title: UILabel!
-    // TODO Make this private and use `update()` instead.
-    @IBOutlet weak var value: UITextField!
+    @IBOutlet private weak var title: UILabel!
+    @IBOutlet private weak var value: UITextField!
+
+    private var viewModel: TitleAndEditableValueTableViewCellViewModel?
```
Ecarrion, line 8: Interesting... 🤔  
I normally don't wire "models" to a view because we end up with two sources of truth. The model state and the UIKit state. Additionally, this also increases complexity in views which I considered to be dumb. 

I understand how it's more convenient to get the value every time the field is updated rather than fetch it through the view hierarchy when the action button is pressed. But still, It feels riskier to me. 

Am I worrying about nothing? 😛

Ecarrion (approved):
Looks and works good! 👍 
I did leave a concern in case you want to discuss it.

WooCommerce/Classes/ViewRelated/ReusableViews/TitleAndEditableValueTableViewCell/TitleAndEditableValueTableViewCell.swift
shiki: > I normally don't wire "models" to a view because we end up with two sources of truth. The model state and the UIKit state.

In this case, I was actually aiming for the `ViewModel` to be the only source of truth. If you look at `self.update()`, once a `ViewModel` is set, the View is reset to whatever the `ViewModel`'s _state_ is. So I believe this View is still dumb. 

I think the complexity here is because we have to use `UITableView` and `UITableView` requires us to make the View (Cells) **reusable**. If there was a way to bind 1 `Cell` to 1 `ViewModel` **forever** then I would opt for that. But I don't know if that's possible right now. 

> But still, It feels riskier to me.

I'm curious about the risks! 

I was aiming for this to be reusable everywhere else. What I noticed in our app is that we use a lot of UI components and they are mostly very similar (especially in Products). My vision is for these UI components to be reusable and not tightly bound to the ViewModel. And also that we don't need to write a lot of code in order to bind them together. There's also an opportunity for us to have one `ViewModel` but different views for iPhone and iPad. 

I don't really know how it's gonna look like yet. I started proposing this for Product Settings (diagrams and discussions here https://github.com/woocommerce/woocommerce-ios/pull/1854#pullrequestreview-357680040) but I'm waiting for iOS 13 to start looking at a design proposal.

Ecarrion: > In this case, I was actually aiming for the ViewModel to be the only source of truth. If you look at self.update(), once a ViewModel is set, the View is reset to whatever the ViewModel's state is. So I believe this View is still dumb.

Hmm, thinking a bit more about it, I guess this is a fancy(or more complex) way of expressing a delegate.

> I'm curious about the risks!

The risk that I fear is that somehow the UI gets updated but the view fails to transmit the value to the view model, so we end up with two sources of truth.

shiki: > The risk that I fear is that somehow the UI gets updated but the view fails to transmit the value to the view model, so we end up with two sources of truth.

That makes sense! I added a test for that but it's certainly possible.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipment Tracking Section/Add Tracking/ManualTrackingViewController.swift
shiki: I could have made a mistake but I looked at the Memory Graph after this VC is dismissed and these objects are all deallocated. So I believe it's fine. 

The block is also not retaining `self` so it's only owned by the `Observable` (`BehaviorSubject`).

shiki:
Thanks, Ernesto!

Issue #2983, opened 2020-10-15
Grouped product > grouped products: support readonly mode


jaclync:
Right now, the grouped products screen is editable with "Add Products" CTA and "X" on each product row to delete a product. Let's support readonly mode so that we can navigate to `GroupedProductsViewController` from a readonly grouped product form.

<img src="https://user-images.githubusercontent.com/1945542/96085289-31c88a00-0ef3-11eb-9b44-b56d9242d723.png" width="300" />

JorgeMucientes:
Closing this as part of a maintenance pass on the GitHub issue backlog. This issue matched the auto-close criteria for this pass: it has not been updated since 2020-10-15, is unassigned, is not labeled as a bug or high/medium priority, and does not appear to involve POS, payments, login, data loss, crash, security, vulnerability, auth, or fraud.

If this is still relevant, please comment with current details and we can reopen it or move it into Linear for triage.

PR #2984, opened 2020-10-15
Product/variation details: support readonly mode

### LINKED ISSUE #2931: Product/variation details: add support for readonly state

jaclync:
Fixes #2931

Apology for the PR size slightly larger than 500 diffs, 300ish diffs are on unit tests for each product type & variation scenario. There are also a few different parts of UI that require changes to support readonly mode.

## Changes

- `ProductDetailsFactory` now always returns `ProductFormViewController` with `readonly` or `edit` form type depending on whether product details are editable
- `ProductVariationDetailsFactory ` now returns `ProductFormViewController` with `readonly` or `edit` form type depending on whether variation details are editable

### Product & variation rows in readonly mode

- Primary rows in both `ProductFormActionsFactory` and `ProductVariationFormActionsFactory`:
  - Description row: hidden when readonly & empty description
  - Images row: hidden when there are no images
- Setting rows in product details (`ProductFormActionsFactory`):
  - Product type row is always shown and readonly for all product types
  - Affiliate/external product:
    - External URL row: hidden when external URL is empty
    - SKU row: hidden when there is no SKU
  - Grouped product:
    - SKU row: hidden when there is no SKU
- Setting rows in variation details (`ProductVariationFormActionsFactory`):
  - Price row: hidden when regular price is empty
- Bottom sheet actions: always empty in both `ProductFormActionsFactory` and `ProductVariationFormActionsFactory`

### More options menu in readonly mode

- More options menu is now always shown:
  - "View Product in Store" is shown when the product is published
  - "Share" is always shown
- Implementation:
  - Added an extension helper to `ProductFormViewModelProtocol.shouldShowMoreOptionsMenu` which returns true when any of the more options menu item is visible. This value is used to show the more options menu icon ("..." in the navigation bar)
  - In `ProductFormViewModel`, updated `canViewProductInStore` and `canShareProduct` based on the logic above
  - Updated `ProductFormViewController` based on the logic above

### UI support for readonly mode

- Added `isActionable: Bool` to `ImageAndTitleAndTextTableViewCell.SwitchableViewModel` so that a switch cell can be readonly
- In `DefaultProductFormTableViewModel`, updated to pass the editable flag to the view model for rows that support readonly mode
- Removed `ProductFormEditAction.variations`'s `editable: Bool` associated value, now that variation details support readonly mode. The variations row is always actionable, and navigates to the variations list (`ProductVariationsViewController`) which now shows each row with a disclosure indicator
- In `ProductFormViewController`, no-op when tapping a readonly row

### Unit tests

- Added unit tests for `ProductFormActionsFactory` and `ProductVariationFormActionsFactory` in readonly mode
- Reorganized the test files so that all tests for `ProductFormActionsFactory` and `ProductVariationFormActionsFactory` are in the same folder "Actions Factory"

## Testing

It's way too time consuming to test all the combinations of states as mentioned above, so I just listed the key scenarios to test for the PR (feel free to remove some fields from the product/variation to make sure they aren't shown in readonly mode). I also tested adding/editing a product from the products tab as confidence check.

### Readonly variation details

Prerequisite: an order includes a variation

- Go to the orders tab
- Tap on the order that includes a variation
- Tap on the variation (its parent product's name is shown) --> a variation details modal should be shown, only rows with non-empty data are shown
- Tap on each row --> nothing should happen as all UI are readonly except for the reviews row

### Readonly product details

- Go to the orders tab
- Tap on the order that includes a product of any type
- Tap on the product --> a product details modal should be shown, only rows with non-empty data are shown (except for the price warning)
- Tap on each row --> nothing should happen as all UI are readonly except for the reviews row

## Example screenshots

product type | readonly (from order details) | editable
-- | -- | --
grouped product | Simulator Screen Shot - iPhone 11 - 2020-10-15 at 14 33 10 | Simulator Screen Shot - iPhone 11 - 2020-10-15 at 14 33 20
external/affiliate product without images, description, and external URL | Simulator Screen Shot - iPhone 11 - 2020-10-15 at 14 36 55 | Simulator Screen Shot - iPhone 11 - 2020-10-15 at 14 37 04
product variation | Simulator Screen Shot - iPhone 11 - 2020-10-15 at 14 35 50 | Simulator Screen Shot - iPhone 11 - 2020-10-15 at 14 36 01

Update release notes:


Ecarrion (approved):
Looks and works good! 🚀 
I see some repeated code but I don't have any clear suggestion to improve it 😅

Garance91540 (commented):
Hi @jaclync !

This is looking good, I have a few questions:

1. Can you plz share with me a screenshot of a variable product? (Not just the variations)

2. On the variation screenshot, I noticed the "Enabled" toggle is still actionable, should we disable it so the page is all read only?

3. I also think every data that's not displayed on the Product Details page  (ie: products composing grouped products, product variations, multi-line product description) should still be accessible for merchants to check, so we should enable users to tap these rows on the Product Details page, and we'll need to signal this by adding chevrons on the rows users can expand.

jaclync:
Hey @Garance91540 great questions!

> Can you plz share with me a screenshot of a variable product? (Not just the variations)

Yes I did test on a variable product! It's quite an edge case, because an order can only have product variations and not a variable product itself and we only show readonly product/variation details when opening a product from order details right now. However, it's still possible if the user changes an existing non-variable product in an order to a variable product (this is how I tested on variable products).

(Sorry my internet was a bit sluggish when recording the screencast)

simulator


> On the variation screenshot, I noticed the "Enabled" toggle is still actionable, should we disable it so the page is all read only?

The toggle is actually not actionable, it's hard to tell from the screenshot. We can update the color to look like in disabled state, here's an example trying with `systemGray3` - lemme know any other colors you'd like to use!

dark | light
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-16 at 09 39 56 | Simulator Screen Shot - iPhone 11 - 2020-10-16 at 09 40 17

> I also think every data that's not displayed on the Product Details page (ie: products composing grouped products, product variations, multi-line product description) should still be accessible for merchants to check, so we should enable users to tap these rows on the Product Details page, and we'll need to signal this by adding chevrons on the rows users can expand.

- For products of grouped products, it's not tappable because we don't have readonly mode for the grouped products screen which requires non-trivial work and I created an issue for it: https://github.com/woocommerce/woocommerce-ios/issues/2983
- For product variations, we do support readonly mode as in the screencast above
- For product description: this is a good point, it'd be nice if we have readonly mode for the Aztec editor. I created a separate issue since it also requires non-trivial work: https://github.com/woocommerce/woocommerce-ios/issues/2994

Garance91540:
Everything looks good! Thank you!

jaclync:
Thanks for the review Ernesto & G!

@Ecarrion I added a small commit https://github.com/woocommerce/woocommerce-ios/pull/2984/commits/98591a1866df4e00e8a2432448e19ae301899ec6 for updating the disabled on color of the switch (https://github.com/woocommerce/woocommerce-ios/pull/2984#issuecomment-709808582) - lemme know if you have any concerns from this commit and I'll fix it separately! Merging now for release 5.3 code freeze today

PR #2987, opened 2020-10-15
Mark or unmark a Product as Downloadable from Product Settings

### LINKED ISSUE #2757: Add/Edit Products M5: Downloadable files Step 2.1 - Mark/unmark a product as downloadable

pmusolino:
Fixes #2757  

## Description
Now it's possible to mark/unmark every simple product as downloadable, from the Product Settings. This is part of Products MIlestone 5.

## Changes
- Introduced `isEditProductsRelease5Enabled` when needed for hiding the downloadable row in settings if the feature flag is not enabled.
- Added `DownloadableProduct` in `ProductSettingsRows`.
- Managed the appearance of the downloadable row in `ProductSettingsSections`.
- Added `downloadable` in `ProductSettings`.
- Added and updated unit tests.

#### Case 1
1. From the Products tab, select any Simple product.
2. Tap on the top right navigation bar button, then open the "Product Settings".
3. You should be able to see a new switchable row, titled `Downloadable Product`.
4. Make any changes on/off to the downloadable switch and update the product.
5. Come back again to this page to verify that your changes are reflected properly.

#### Case 2
1. From the Products tab, select a non-simple product.
2. Go to the product settings -> the Downloadable Product row shouldn't be visible.

## Screenshots
| Dark mode OFF             |  Dark mode ON |
Simulator Screen Shot - iPhone 11 Pro - 2020-10-15 at 15 23 06 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-15 at 15 24 58

## GIF

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
@@ -56,7 +56,7 @@ final class ProductDetailsViewModel {
     var productID: Int64 {
         return product.productID
     }
-    
+
     let isEditProductsRelease5Enabled: Bool
 
     // MARK: - private variables
@@ -125,7 +125,7 @@ final class ProductDetailsViewModel {
     init(product: Product, isEditProductsRelease5Enabled: Bool) {
         self.product = product
         self.isEditProductsRelease5Enabled = isEditProductsRelease5Enabled
-        
+
         refreshResultsController()
     }
 
--- a/WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
@@ -56,6 +56,8 @@ final class ProductDetailsViewModel {
     var productID: Int64 {
         return product.productID
     }
+    
+    let isEditProductsRelease5Enabled: Bool
 
     // MARK: - private variables
 
@@ -120,9 +122,10 @@ final class ProductDetailsViewModel {
 
     /// Designated initializer.
     ///
-    init(product: Product) {
+    init(product: Product, isEditProductsRelease5Enabled: Bool) {
         self.product = product
-
+        self.isEditProductsRelease5Enabled = isEditProductsRelease5Enabled
+        
         refreshResultsController()
     }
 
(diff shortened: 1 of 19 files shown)


WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
```diff
@@ -709,8 +709,10 @@ extension ProductDetailsViewModel {
             WebviewHelper.launch(product.externalURL, with: sender)
         case .productVariants:
             ServiceLocator.analytics.track(.productDetailViewVariationsTapped)
+            let isEditProductsRelease5Enabled = ServiceLocator.featureFlagService.isFeatureFlagEnabled(.editProductsRelease5)
```
Ecarrion: Do you think it's worth it to inject `ServiceLocator.featureFlagService` in the init of this view model(and set it as a default parameter), so it's clear that this view model has that dependency?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/ProductSettingsRows.swift
```diff
@@ -185,6 +185,37 @@ enum ProductSettingsRows {
...
+          init(_ settings: ProductSettings) {
+             self.settings = settings
+         }
+
+          func configure(cell: UITableViewCell) {
+             guard let cell = cell as? SwitchTableViewCell else {
+                 return
+             }
+
+              let title = NSLocalizedString("Downloadable Product", comment: "Downloadable Product label in Product Settings")
```
Ecarrion: Should this go into a `Localization` enum as with other classes? 
I thought that was the standard on the codebase but correct me if I'm wrong!

WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/ProductSettingsRows.swift
```diff
@@ -185,6 +185,37 @@ enum ProductSettingsRows {
...
+
+              cell.title = title
```
Ecarrion: nit: extra space here

Ecarrion (approved):
Looks good! 
Just left a couple of suggestions for your consideration

WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/ProductSettingsRows.swift
pmusolino: Good idea. I moved all the `NSLocalizationString` of this class under `Localization` enum b71e516738260c44b7d775f455c3f5f7438bafbe

WooCommerce/Classes/ViewModels/ProductDetailsViewModel.swift
pmusolino: Sounds good 👍 done here 027119293fe8f2fc5f4d26f1a0adb66e9bdd34eb

pmusolino:
Thank you @Ecarrion for your review. 🙇 I'll wait for the @Garance91540 review before merging it.

Garance91540 (approved):
👏

pmusolino:
Thank you both! Merging it now. <3

PR #2989, opened 2020-10-15
[Issue Refunds] Handle previous refunds

Ecarrion:
# Why

This PR takes care of handling previous refunds on the issue refund screen. Particularly handles two cases:
1. Reduce the number of items available to refund 
2. Hide items that were previously refunded

# How
To achieve this I'm running a filter method `filterItems(from order: Order, with refunds: [Refund])` on `IssueRefundViewModel`. This method updates the item quantity based on previous refunds and filter items when the updated quantity is 0.

In order to modify the item quantity, I've(sourcery) added methods to update an `OrderItem`

Additionally, I've modified the `IssueRefundViewModel` init to receive the the previous refunds from `OrderDetailViewModel`

# Screenshots
Initial | After Refunds
--- | --- 
initial | later

# Testing Steps
- Navigate to an order with multiple items and quantities
- Tap Issue Refund button
- See all items and quantities being available for refund
- On Wp-Admin issue a refund by quantity on any item
- Refresh the order and tap on issue refund again
- See that the item quantities have been reduced or that the item disappears if there is no quantity left to refund.

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -235,6 +240,36 @@ extension IssueRefundViewModel {
...
         return String.pluralize(count, singular: Localization.itemSingular, plural: Localization.itemsPlural)
     }
+
+    /// Return an array of `orderItems` by taking out all previously refunded items
+    ///
+    private static func filterItems(from order: Order, with refunds: [Refund]) -> [OrderItem] {
+        // Flattened array with all items refunded
+        let allRefundedItems = refunds.flatMap { $0.items }
+
+        // Transform `order.items` by substracting the quantity left to refund and evicting those who were fully refunded.
```
```suggestion
        // Transform `order.items` by subtracting the quantity left to refund and evicting those who were fully refunded.
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -235,6 +240,36 @@ extension IssueRefundViewModel {
...
+        return order.items.compactMap { item -> OrderItem? in
+
+            // Calculate how many times an item has been refunded. This number is negative.
+            let timesRefunded = allRefundedItems.reduce(0) { timesRefunded, refundedItem -> Decimal in
```
jaclync, line 253: nit: just a thought of optimization - maybe `allRefundedItems` can be reduced to a dictionary that maps from `refundedItem.productOrVariationID` to the number of times this product/variation has been refunded before looping through `order.items`. This way, it saves some time looping through `allRefundedItems` for each order item.

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -228,11 +228,56 @@ final class IssueRefundViewModelTests: XCTestCase {
...
-        let viewModel = IssueRefundViewModel(order: order, currencySettings: currencySettings)
+        let viewModel = IssueRefundViewModel(order: order, refunds: [], currencySettings: currencySettings)
         viewModel.selectAllOrderItems()
 
         // Then
         let selectedItemsTitle = String(format: NSLocalizedString("%d items selected", comment: ""), 6)
         XCTAssertEqual(viewModel.selectedItemsTitle, selectedItemsTitle)
     }
+
+    func test_viewModel_correcly_reduces_refunded_items() {
```
jaclync: nit:

```suggestion
    func test_viewModel_correctly_reduces_refunded_items() {
```

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -228,11 +228,56 @@ final class IssueRefundViewModelTests: XCTestCase {
...
+        // When
+        let viewModel = IssueRefundViewModel(order: order, refunds: [refund], currencySettings: currencySettings)
+
+        // Then
+        XCTAssertEqual(viewModel.quantityAvailableForRefundForItemAtIndex(0), 1)
+        XCTAssertEqual(viewModel.quantityAvailableForRefundForItemAtIndex(1), 1)
+        XCTAssertEqual(viewModel.quantityAvailableForRefundForItemAtIndex(2), 1)
+    }
+
+    func test_viewModel_correcly_filters_items_already_refunded() {
```
jaclync: nit:
```suggestion
    func test_viewModel_correctly_filters_items_already_refunded() {
```

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -228,11 +228,56 @@ final class IssueRefundViewModelTests: XCTestCase {
...
+            MockRefunds.sampleRefundItem(productID: 1, quantity: -3),
+            MockRefunds.sampleRefundItem(productID: 2, quantity: -2),
+        ])
+
+        // When
+        let viewModel = IssueRefundViewModel(order: order, refunds: [refund], currencySettings: currencySettings)
+
+        // Then
+        let itemRows = viewModel.sections[0].rows
+        XCTAssertEqual(itemRows.count, 2) // One item left + summary row
```
jaclync, line 304: just wondering, does it make sense to assert on the quantity like the previous test case `XCTAssertEqual(viewModel.quantityAvailableForRefundForItemAtIndex(0), 1)`?

jaclync (commented):
Thanks for handling this somewhat complicated scenario!

Code: ✅ 
Testing: ❓ 

I noticed that "Select All" seems a bit odd - it's still showing the total quantity from the original order (2+2) and the quantity label for each item still shows 0:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
Ecarrion: I tried that at the beginning but felt the algorithm was a little bit harder to read and understand. Then I weighed that against lopping through the array(which should not have a lot of items) and I decided that it was fast enough 😅

Ecarrion:
Hey @jaclync, Thank you a lot for the review! Thanks to it I caught two bugs.

1. Selecting all bug, this happened because I was using `state.order.items` and not `state.itemsToRefund`. The fix was just using the correct array.

2. Tax calculations for an item were wrong after the item had any refunds, this happened because the calculation to get the tax is `item.totalTax / item.quantity` but my new filter method was updating the quantity hence giving me wrong tax info. The fix was to create a new struct that contains the original item + the real quantity available to refund in a separate field.

```swift
/// Groups an order item and its quantity available for refund
///
struct RefundableOrderItem {
    /// Original purchased item
    ///
    let item: OrderItem

    /// Current quantity available for refund
    ///
    let quantity: Int
}
```

----

Additionally, I've 
- Added tests cases for these bugs
- Remove the Orderitem+copiable methods since they are no longer needed.

Please take another look ♻️

loremattei:
Hey there! I'm moving this to 5.4 as 5.3 has been cut. If you still want this to make it to 5.3, please feel free to ping me.

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -134,6 +134,29 @@ final class IssueRefundViewModelTests: XCTestCase {
         XCTAssertEqual(viewModel.currentQuantityForItemAtIndex(3), nil)
     }
 
+    func test_viewModel_updates_refund_quantities_after_selecting_all_while_having_previous_refunds() {
```
jaclync, line 137: nice :)

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -134,6 +134,29 @@ final class IssueRefundViewModelTests: XCTestCase {
...
+            MockRefunds.sampleRefundItem(productID: 2, quantity: -2),
+        ])
+
+        // When
+        let viewModel = IssueRefundViewModel(order: order, refunds: [refund], currencySettings: currencySettings)
+        viewModel.selectAllOrderItems()
+
+        // Then
+        XCTAssertEqual(viewModel.currentQuantityForItemAtIndex(0), 1) // Product 1
+        XCTAssertEqual(viewModel.currentQuantityForItemAtIndex(1), 1) // Product 3 is at index 2 because Product 2 was refunded
```
jaclync: nit: Product 3 is at index 1 (0-indexed)?

jaclync (approved):
LGTM! :shipit: 

Nice catch on the tax calculation with previous refunds, it gets really complicated! I retested on the order and "select all" works great and the tax looks correct 👍

PR #3007, opened 2020-10-19
Rename `briefDescription` to `shortDescription` to be consistent with naming in core

### LINKED ISSUE #1899: Rename `briefDescription` to `shortDescription`
In core, the field is called "short description" while we call it `briefDescription` in the codebase. Since `briefDescription` is already in the core data schema, we could probably rename every `briefDescription` to `shortDescription` except for in Storage

jaclync:
Fixes #1899

## Changes

This PR just contains renaming from `briefDescription` to `shortDescription` so that it is more consistent with the naming in core. An exception is in the storage layer, where the name `briefDescription` is still kept in the schema and `StorageProduct.briefDescription` since it doesn't feel worth a migration.

[API doc:](https://woocommerce.github.io/woocommerce-rest-api-docs/#product-properties)
> short_description | string | Product short description.

Core:
Screen Shot 2020-10-19 at 9 52 55 AM

## Testing

All the diffs are 1-1 renaming in the Networking/App layers, feel free to confidence check on the app by launching it and tapping on products with/without a short description.

Update release notes:


pmusolino (approved):
LGTM, thanks for handling this @jaclync! :shipit: 

I did also a confidence check, everything seems working like before. 
Just a thing: do you think it's worth to rename also the `briefDescription` property in CoreData or to leave it as is without doing a migration?

jaclync:
> Just a thing: do you think it's worth to rename also the `briefDescription` property in CoreData or to leave it as is without doing a migration?

I believe renaming in Core Data requires changing the schema though - if we just update the property of `Product` in storage, it will be overwritten the next time someone regenerates the attributes for this entity (like after adding a new attribute). I think it's okay leaving `briefDescription` in the storage layer, since we always use the readonly version in the app layer 🙂

Issue #2843, opened 2020-09-20
[Issue Refunds] Refund Network Calls


Ecarrion:
- Perform refund network call 
- Make sure that the new refund is reflected on the order screen

PR #3031, opened 2020-10-19
[Issue Refund] Refund API encoding

Ecarrion:
part of #2843 

# Why

As part of the refund shipping investigation, we noticed that the way to refund items on the API is not very mobile-friendly because its structure does not match with proper JSON objects.
```
{
  "amount": "97,99",
  "reason": "refund-product-wp-console",
  "api_refund": false,
  "line_items": {
    "63": {
      "refund_total": "90.0",
      "refund_tax": {
        "1": "7,99"
      }
    }
  }
}
```
ref: p91TBi-3hR-p2

This PR updates the `Refund` and `OrderItemRefund` entities to match the encoding required by the API.

# How

- On `Refund.swift` I added a method `itemsDictionary` that reduces the items array into a dictionary in the correct format.
- On `OrderItemRefund.swift` I added a method `taxesDictionary` that reduces the taxes array into a dictionary in the correct format
- On `OrderItemRefund.swift` I added an `EncodingKeys` enum as those keys are different from the `Decoding` ones

# Testing steps
Make sure the tests pass.

Update release notes:

### DIFF
--- a/Networking/NetworkingTests/Mapper/RefundMapperTests.swift
+++ b/Networking/NetworkingTests/Mapper/RefundMapperTests.swift
@@ -61,6 +61,64 @@ final class RefundMapperTests: XCTestCase {
         XCTAssertEqual(item.sku, "HOODIE-SHIP-YOUR-IDEA-BLUE-XL")
         XCTAssertEqual(item.price, NSDecimalNumber(integerLiteral: 27))
     }
+
+    func test_refund_is_encoded_correctly_with_items_and_taxes() throws {
+        // Given
+        let refund = sampleRefund(includeTaxes: true)
+        let mapper = RefundMapper(siteID: dummySiteID, orderID: 123)
+
+        // Ref: https://github.com/woocommerce/woocommerce/blob/0a81188e42fd3797ea2dd09f35c777e787801808/includes/wc-order-functions.php#L535-L573
+        let expectedDictionary: [String: Any] = [
+            "amount": refund.amount,
+            "api_refund": refund.createAutomated ?? false,
+            "reason": refund.reason,
+            "line_items": [
+                "\(refund.items[0].itemID)": [
+                    "qty": refund.items[0].quantity,
+                    "refund_total": refund.items[0].total,
+                    "refund_tax": [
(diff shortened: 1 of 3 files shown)


Networking/NetworkingTests/Mapper/RefundMapperTests.swift
```diff
@@ -61,6 +61,64 @@ final class RefundMapperTests: XCTestCase {
         XCTAssertEqual(item.sku, "HOODIE-SHIP-YOUR-IDEA-BLUE-XL")
         XCTAssertEqual(item.price, NSDecimalNumber(integerLiteral: 27))
     }
+
+    func test_refund_is_encoded_correctly_with_items_and_taxes() throws {
+        // Given
+        let refund = sampleRefund(includeTaxes: true)
+        let mapper = RefundMapper(siteID: dummySiteID, orderID: 123)
+
+        // Ref: https://github.com/woocommerce/woocommerce/blob/0a81188e42fd3797ea2dd09f35c777e787801808/includes/wc-order-functions.php#L535-L573
```
pmusolino: I suggest to use a short link here 👍 
```suggestion
        // Ref: https://git.io/JTRsF
```

Networking/NetworkingTests/Mapper/RefundMapperTests.swift
```diff
@@ -61,6 +61,64 @@ final class RefundMapperTests: XCTestCase {
...
+        let dictionaryEquals = NSDictionary(dictionary: refundDictionary ?? [:]).isEqual(to: expectedDictionary)
+        XCTAssertTrue(dictionaryEquals)
+    }
+
+    func test_refund_is_encoded_correctly_with_items_and_no_taxes() throws {
+        // Given
+        let refund = sampleRefund(includeTaxes: false)
+        let mapper = RefundMapper(siteID: dummySiteID, orderID: 123)
+
+        // Ref: https://github.com/woocommerce/woocommerce/blob/0a81188e42fd3797ea2dd09f35c777e787801808/includes/wc-order-functions.php#L535-L573
```
pmusolino: 👍 same here

pmusolino (approved):
LGTM :shipit:

Issue #3033, opened 2020-10-20
[Issue Refund] Support api_refund property


Ecarrion:
API exposes an `api_refund` property that lets the store to automatically refund payments using the order payment gateway.

We need to:
- See what Android is doing with that flag
- Research if we can fetch that value that from the order payment gateway
- Mark that flag if the payment gateway supports it

PR #3035, opened 2020-10-20
[Issue Refunds] Create order item refunds

Ecarrion:
Part of #2843 

# Why

This PR takes care of providing the necessary interfaces to create remote refunds for order items(Shipping refunds will be supported in a next PR).

# How 

- Extracted `RefundableOrderItem` from `IssueRefundViewModel` as it widely used and needed across the refund module.
- Created a `RefundCreationUseCase` type to help transform `OrderItems` into a `Refund` object with `OrderItemRefund` items.
- Added a `localize` method on `CurrencyFormatter` that works with `Decimal` types

# Testing Steps

This code is not yet integrated into the app, but you can try it in the following way.

1. Replace the code in `createRefundConfirmationViewModel` method on `IssueRefundViewController.swift` line 77 with
```swift
func createRefundConfirmationViewModel() -> RefundConfirmationViewModel {

    let items = state.refundQuantityStore.map { RefundableOrderItem(item: $0, quantity: Decimal($1)) }
    let useCase = RefundCreationUseCase(amount: "\(calculateRefundTotal())",
                                        reason: "PR Review",
                                        automaticallyRefundsPayment: false,
                                        items: items,
                                        currencyFormatter: CurrencyFormatter(currencySettings: state.currencySettings))
    let refund = useCase.createRefundObject()

    let action = RefundAction.createRefund(siteID: state.order.siteID,
                                           orderID: state.order.orderID,
                                           refund: refund) { refund, error  in
        print("completed")
        if let error = error {
            print("Error: \(error)")
        }
    }

    return RefundConfirmationViewModel(order: state.order, currencySettings: state.currencySettings)
}
```
This will fire a create refund action after tapping the next button in the issue refund screen.

2. Go to an un-refunded order

3. Select some items to refund (Please do not enable refund shipping)

4. Tap Next

5. Go to the sore admin and see that the correct items were refunded for the order

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/RefundableOrderItem.swift
```diff
@@ -0,0 +1,28 @@
...
+    ///
+    let item: OrderItem
+
+    /// Current quantity available for refund
+    ///
+    let quantity: Int
+
+    /// Quantity expressed as decimal
+    ///
+    var decimalQuantity: Decimal {
```
jaclync, line 20: nit: maybe this derived property could be in an extension so that it's easy to see the main properties?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/UseCases/RefundCreationUseCase.swift
```diff
@@ -0,0 +1,89 @@
...
+    ///
+    let items: [RefundableOrderItem]
+
+    /// Currency formatted needed for decimal calculations
+    ///
+    let currencyFormatter: CurrencyFormatter
+
+    /// Creates a `Refund` object ready to be used  on `RefundStore.createRefund` action
+    ///
+    func createRefundObject() -> Refund {
```
jaclync: nit: how about `createRefund`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/UseCases/RefundCreationUseCase.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+
+    /// Calculates the refundable tax from a tax line by diving its total tax value by the purchased quantity and mutiplying it by the refunded quantity.
+    ///
+    private func calculateTax(of taxLine: OrderItemTax, purchasedQuantity: Decimal, refundQuantity: Decimal) -> String {
+        let totalTax = currencyFormatter.convertToDecimal(from: taxLine.total) ?? 0
+        let itemTax = (totalTax as Decimal) / purchasedQuantity
+        let refundableTax = itemTax * refundQuantity
+        return currencyFormatter.localize(refundableTax) ?? "\(refundableTax)"
```
jaclync, line 79: just curious, does the API expect formatted amount for a refund with the site currency settings? I just thought that the formatting is usually for UI display. Also, since we only fetch site settings on app launch, there's a slight chance that currency settings are outdated

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/UseCases/RefundCreationUseCase.swift
```diff
@@ -0,0 +1,89 @@
+import Foundation
+import Yosemite
+
+/// Creates a `Refund` object ready to be used  on `RefundStore.createRefund` action
+///
+struct RefundCreationUseCase {
+
+    /// Total amount to be refunded.
+    ///
+    let amount: String
```
jaclync, line 10: just wondering, can `amount` be derived from `items`? or this total amount could include other information that this use case doesn't know?

jaclync (approved):
Woo it's really close! It's easy to review the changes in the use case with tests.

Just a note that I replaced the code in `IssueRefundViewModel` L77 instead of `IssueRefundViewController` as in the PR description 🙂 

Code: ✅ just some minor questions!
Testing: ✅ working nicely! I was able to see the refund in core 🎉 Do you know if automatic/manual refund is supported? Right now the refunds are manual

app | core
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-21 at 16 48 38 | Screen Shot 2020-10-21 at 4 48 35 PM

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/UseCases/RefundCreationUseCase.swift
Ecarrion: From what I could see this method formats the amount without the currency and uses a `.` separator. 
It felt more secure than using only `"\(amount)"`

In any case I'll add a task to test this with different currencies and variations!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/UseCases/RefundCreationUseCase.swift
Ecarrion: `amount` could be derived from `items` + `shipping` but the screen that will call this method already has the amount calculated and displayed to the user, that's why I decided to just pass it on. But will consider it again once I add the shipping support!

Ecarrion:
@jaclync 

> Just a note that I replaced the code in IssueRefundViewModel L77 instead of IssueRefundViewController as in the PR description 🙂

Sorry, I meant `IssueRefundViewModel` 😇 

>  Do you know if automatic/manual refund is supported? Right now the refunds are manual

Yes, right now refunds are manual but I have [a task](https://github.com/woocommerce/woocommerce-ios/issues/3033) to analyze automatic refund support!

Issue #2280, opened 2020-05-13
Order Details → Separate Product Name from Variation Name


shiki:
⚠️ This is not as easy as it looks as it may involve a few API fetches in order to get the variation names. See p91TBi-2Lt-p2.

![image](https://user-images.githubusercontent.com/198826/81869558-71004280-9531-11ea-99d8-8e3de92d5ea6.png)

- Title: product name only
- Subtitle: Variant selected (if any), quantity and price. This will make it easier to visually separate the product from the variations selected, especially if several have been selected for the same product (ie “White” and “Medium”)

Ref p91TBi-2CE-p2#comment-627

shiki:
Related to https://github.com/woocommerce/woocommerce-ios/issues/1127 and blocked by https://github.com/woocommerce/woocommerce-rest-api/issues/9.

jaclync:
@woocommerce/ios-developers would love to get your thoughts on implementing this, since it involves schema change:

The variation attributes are in each `OrderItem`'s `meta_data` field in the API response `GET /orders/{{order_id}}`:

```
"line_items": [
  {
    "id": 720,
    "name": "(Variable) Chocolate bars",
    "product_id": 846,
    "variation_id": 1275,
    ...
    "meta_data": [
      {
        "id": 6041,
        "key": "darkness",
        "value": "99%",
        "display_key": "Darkness",
        "display_value": "99%"
      },
       ...
     ]
  }
]
```

We have a few ways to store the variation attributes from `meta_data` in `Storage.OrderItem` if its `variationID` is non-zero:

- Add a new relationship `attributes: [GenericAttribute]` or `variationAttributes: [GenericAttribute]`. Its `id`/`key`/`value` are from `id`/`display_key`/`display_value` in the API response `meta_data` field respectively.
- Create a new model `OrderItemAttribute` that contains all the attributes from `meta_data` (`id`/`key`/`value`/`display_key`/`display_value`) even though we're only using `display_key` and `display_value` for this issue. Then add a new relationship `attributes: [OrderItemAttribute]`.
- Add a new attribute `details: String`: this freeform string can potentially be parsed from other fields in the future, but is less flexible in how we display the variation attributes (like comma separated)

In `Networking.OrderItem`, I'll add the corresponding field and parsing the meta data with a decodable struct (similar to `ProductVariation.attributes: [ProductVariationAttribute]`).

I personally prefer adding a relationship of attributes, what do you think?

[...]

PR #3036, opened 2020-10-21
Products tab: fix earlier draft products appearing at the top when sorting by "Newest to Oldest"

### LINKED ISSUE #2727: [Product List] When sorting by "Newest to Oldest", draft products appear on top of the list even if I created them before other products
When sorting by "Newest to Oldest" my products in my product list, the draft products I have created appear on top of the list even if I created them before other products. 

Example: on this screenshot, the "test" product was created months ago and the "[simple] Grand Ficus Elastica" was created an hour ago

IMG_3517CEFBB341-1

jaclync:
Fixes #2727 

The PR has ~2700 diffs 😱 However, the actual changes of this PR are pretty small (under 200 diffs), just a lot of diffs in the Core Data mapping model and model version.

- The changes in `Model 33.xcdatamodel/contents` are in https://github.com/woocommerce/woocommerce-ios/commit/d9c5aa0512525c9fc82e5a90ed16679b811ff699#diff-c2640637c23ff4b62d2a9dc19c9290438213cab88f068214fb49c9d647b5d616
- The only change I made in `WooCommerceModelV32toV33.xcmappingmodel/xcmapping.xml`:
Screen Shot 2020-10-21 at 11 14 19 AM

## Changes

- Created a new model version 33 with a mapping model that sets the new non-optional `Product.date` to the previous `Product.dateCreated`, plus a test case in `MigrationTests`.
- Added `date: Date` to `Product` model in Networking and Storage layers, updated unit tests. In `Networking.Product`, the new `date` property is set to `dateModified` for a draft product and `dateCreated` otherwise.
- In `ResultsController+SortProducts`, updated `sortDescriptors` for `dateAscending` and `dateDescending` cases to use the new `date` attribute instead of `dateCreated`. Added a test case in `ResultsController_StorageProductTests` that would fail without changes in this PR.

## Testing

Note: this bug is only reproducible on certain sites where a draft product has `null` creation date and non-nil modified date. On my atomic site (WP.com business plan), a draft product has non-nil creation/modified date. However, in G and Bananza sites, the bug occurs. Feel free to test on any site you prefer, regardless of the draft product API behavior!

Prerequisite: the site has a draft product created/modified earlier than a published product.

- Go to the Products tab
- Tap "Sort by"
- Tap "Date: Newest to Oldest" --> the products should be shown in the same order as in core, especially for the draft product(s)

## Example screenshots

product list in core:

Screen Shot 2020-10-21 at 11 07 11 AM

before | after
-- | -- 
Simulator Screen Shot - iPhone 11 - 2020-10-21 at 10 48 07 | Simulator Screen Shot - iPhone 11 - 2020-10-21 at 11 06 36
Update release notes:


Ecarrion (approved):
Looks good to me! 🚀 
I think the only thing that is missing is to update the migrations readme.

Ecarrion:
Now that I take another look I see some inconsistencies with core, but I think it's a core problem.


Screen Shot 2020-10-21 at 7 42 21 PM

Both draft products were updated today, but core is not picking up the last modified date correctly hence sorting the products wrong. I think the apps is good as per the PR description tho.

Am I missing something? 😅

jaclync:
Thank you for reviewing & testing this @Ecarrion! Merging now to unblock https://github.com/woocommerce/woocommerce-ios/issues/2280 which requires schema update as well, response to your comments below!

> I think the only thing that is missing is to update the migrations readme.

Oops I did forget to update the migrations README 😳 Thanks a lot for noticing this! Updated this in https://github.com/woocommerce/woocommerce-ios/pull/3036/commits/6661f1eddff94c95e12ad98b2b48d0f70b583ab0

> Both draft products were updated today, but core is not picking up the last modified date correctly hence sorting the products wrong. I think the apps is good as per the PR description tho.
Am I missing something? 😅

This does look like a core issue. I was able to reproduce this in my test store with WC 4.7.0-beta, but not in Bananza site with WC 4.1.1. However, when I switched my site to WC 4.1.1 it still doesn't fix the issue. @Ecarrion what's the WC version in your store? I'll create an issue in core repo after testing more in different sites.

Ecarrion:
>  @Ecarrion what's the WC version in your store? I'll create an issue in core repo after testing more in different sites.

I'm on `4.6.0`, let me know if you want me to add you to my store!

Ecarrion:
> Merging now to unblock #2280 which requires schema update as well, response to your comments below!

@jaclync FYI I also submitted https://github.com/woocommerce/woocommerce-ios/pull/3045 which involves a new CD model 🙈

jaclync:
> I'm on 4.6.0, let me know if you want me to add you to my store!

@Ecarrion actually it would be helpful if you can add me to your store! I'd like to check if it's an issue because of some common plugins since the issue still persists with much earlier WC version on my site.

> FYI I also submitted #3045 which involves a new CD model 🙈

No problem at all! I actually didn't get to start on the work that requires schema changes 😄 I'll ping Woo iOS devs in mobile-woo when I start!

Ecarrion:
@jaclync Added to your A8C email!

PR #3042, opened 2020-10-22
Add/Edit Product > Price: sale date enhancements

jaclync:
Fixes #3040 

## Changes

- In `DatePickerTableViewCell`, configured the date picker to have `preferredDatePickerStyle = .wheels` in iOS 14+. I've tried using the new iOS 14 picker style [`preferredDatePickerStyle = .inline`](https://developer.apple.com/documentation/uikit/uidatepickerstyle/inline), but I couldn't get the height to work properly. With the current date picker cell height 216px (I also tested 250px), the top part of the iOS 14 style picker is truncated (weekday labels and year/month selector).
  - I also tried setting the cell height to automatic and setting the `UIDatePicker` to have `contentCompressionResistancePriority` as `required` vertically, but this didn't work either. Feel free to share any other ideas to show the new iOS 14 date picker inline!
- During testing, I also noticed that the sale end date editing is automatically ended whenever I changed just the year/month/day once. It felt weird since I might want to change other date fields. I made an update to fix this in `ProductPriceSettingsViewModel` with two test cases

## Testing

- Launch the app in iOS 14 simulator/device from Xcode 12
- Go to the products tab
- Tap on a product whose price is editable
- Tap on the price row
- Turn on "Schedule sale" toggle
- Tap on either "From" or "To" row --> the date picker should behave like in iOS 13
- Change the sale date to a different date --> the date picker should stay editable on any year/month/day changes
- Tap on "Remove end date" row --> the "Remove end date" row and date picker row should disappear and the sale end date is reset

## Example screenshots

iOS version | before | after
-- | -- | --
iOS 14.0 | Simulator Screen Shot - iPhone 11 - 2020-10-22 at 17 06 24 | Simulator Screen Shot - iPhone 11 - 2020-10-22 at 16 41 56
iOS 13.1 | Simulator Screen Shot - iPhone 11 - 2020-10-22 at 17 09 09 | Simulator Screen Shot - iPhone 11 - 2020-10-22 at 17 10 09

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #3040, opened 2020-10-22
Add/Edit Product > Price: sale date picker UI looks odd in iOS 14


jaclync:
**Describe the bug**

When building from Xcode 12 to an iOS 14 simulator/device, the sale date UI looks odd as the current date takes up a lot of vertical space. The UX still works, the user can still tap the current date to bring up the new date picker and pick another date. Let's update the UI to make it look nicer with the new date picker in iOS 14, and test the changes in both iOS 13 and 14.

**To Reproduce**

- Launch the app in iOS 14 simulator/device from Xcode 12
- Go to the products tab
- Tap on a product whose price is editable
- Tap on the price row
- Turn on "Schedule sale" toggle
- Tap on either "From" or "To" row --> notice the current date takes a lot of vertical space

**Screenshots**

iOS 14 | iOS 13
-- | --
![Simulator Screen Shot - iPhone 11 - 2020-10-22 at 14 40 39](https://user-images.githubusercontent.com/1945542/96835449-2ab2f600-1476-11eb-8a1b-d721c06a9fdf.png) | ![Simulator Screen Shot - iPhone 11 - 2020-10-22 at 14 41 44](https://user-images.githubusercontent.com/1945542/96835466-2f77aa00-1476-11eb-9d9b-a808260658e0.png)

**Expected behavior**

The date picker should be editable inline or doesn't take so much vertical space.

**Isolating the problem (mark completed items with an [x]):**
- [ ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [ ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

**Mobile Environment**
Please include:
- Device: iPhone 11 simulator
- iOS version: iOS 14.0
- WooCommerce iOS version: 5.4

PR #3044, opened 2020-10-22
Date pickers now have the look introduced in iOS 14

pmusolino:
Part of #3040. 

## Description
This is an enhancement to #3042 originally started by @jaclync. I made some changes for showing correctly the new inline date picker in Product Price and in Add Tracking (under Order detail).
Now when a user is running iOS 14+ we will show the new inline date picker for choosing a date.

## How I solved the original issue
It has always been impossible to self-define the correct height of a UIDatePicker, because of a limitation on the Apple side. What we should do in this case, is to set a fixed height based on the style that we are showing.
Before, we always set the fixed height to be 216px, but the new date picker needs 345px. So I implemented a static method inside the `DatePickerTableViewCell` for getting the default cell height based on the iOS on which we are running.
I also updated the xib of the cell, since there was a little bug in the constraints that don't allow us to show the bottom border of the cell (it was hidden under the date picker).
In the end, I removed the `estimatedHeightRowAt` delegate method in `ManualTrackingViewController` because isn't needed.

#### Case 1
- Launch the app in iOS 14 simulator/device from Xcode 12
- Go to the products tab
- Tap on a product whose price is editable
- Tap on the price row
- Turn on "Schedule sale" toggle
- Tap on either "From" or "To" row --> the date picker should have the new appearance introduced in iOS 14

#### Case 2
- Launch the app in iOS 14 simulator/device from Xcode 12 and make sure to have the Shipping Tracking extension on your website
- Go to the Orders tab
- Tap an Order detail
- Press "Add tracking"
- Press "Date shipped" --> the date picker should have the new appearance introduced in iOS 14

## Screenshots
| Produce Price pickers             |  Add Tracking picker |
Simulator Screen Shot - iPhone 11 Pro - 2020-10-22 at 13 34 48 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-22 at 13 35 40

Update release notes:


Garance91540 (approved):
cool stuff! 👏

jaclync:
The code looks great! I'm not sure why, the date picker is still truncated in iOS 14.0 simulator and device 🤔 I checked in view debugger that the date picker cell height is 345px (including a 1px bottom border). I also tried adjusting the font size in accessibility but it didn't fix it. Which device type/iOS version did you test on?

iOS 14.0 iPhone 11 sim | iOS 14.0 iPhone XR device
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-23 at 10 00 55 | IMG_6345

pmusolino:
Thank you @jaclync for testing this!

I did some tests with different devices with different languages and everything seems ok on my side. The only difference that I noticed is that you are using iOS 14.0. What happens on your side if you use iOS 14.1? 🤔 

|  iPhone XR iOS 14.1 | iPhone XR iOS 14.1 (Chinese simplified)    |  iPhone 11 Pro iOS 14.1 | iPhone 8 Pro iOS 12.2 | iPhone 5s iOS 12.2 |
:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:|:-------------------------:
iphone xr ios 14 1 | iphone xr ios 14 1 chinese | iphone 11 pro | iphone 8 | iphone 5s 12 2

jaclync:
> The only difference that I noticed is that you are using iOS 14.0. What happens on your side if you use iOS 14.1? 🤔

Ah yea, it's iOS 14.1 / 14.0 that makes the difference! It seems like the spacing between calendar rows is slightly larger in iOS 14.0 and thus there's no space for the text on top. This is quite tricky, since most of our users are still on iOS 14.0 (based on App Store Analytics) but people are gradually upgrading to iOS 14.1 😓 I wish the date picker worked with Auto Layout in `UITableViewCell`, or they maintain the same height between iOS versions.

We can probably set different height for iOS 14.0 / 14.1 (but there's a chance we might have to do this for each iOS 14 version), or keep the wheel style for now?

iOS 14.1 | iOS 14.0
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-27 at 14 07 00 | Simulator Screen Shot - iPhone 11 - 2020-10-27 at 14 09 50

pmusolino:
Thank you @jaclync for confirming it. 
Then, I believe that increasing the size to 380px would solve all problems on both iOS 14.0 and iOS 14.1. 

The appearance is still good on iOS 14.1, and on iOS 14.0 it would be OK, without any issue. Can you confirm that it runs properly on your device, setting the height to 380px?

jaclync:
380px works for both iOS 14.0 and 14.1 (a bit tall but looks good), let's go for this! I also tested with the largest font size and the calendar just stays the same size 👍 Let's hope they don't increase the content size height in the later iOS versions 😅 

iOS 14.0 | iOS 14.1
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-29 at 11 57 33 | Simulator Screen Shot - iPhone 11 - 2020-10-29 at 11 57 34

pmusolino:
Thank you for testing it @jaclync! 💯 

> Let's hope they don't increase the content size height in the later iOS versions

I really hope they didn't change, but as far I see yesterday, there are other people in GitHub that are using our same solution because it seems that the date picker is not working properly with Autolayout. 
Maybe in the future, we can try to implement this view using SwiftUI, and manage the constraints using it. Maybe in SwiftUI the issue is not existing at all. 🤷‍♂️

PR #3046, opened 2020-10-22
Merge Release/5.3 into develop

loremattei:
Merge release notes for release/5.3 and a fix into develop.

Update release notes:


AliSoftware:
Side question: why do Peril always comment that many times (3 times, including one which has the same message duplicated) instead of replacing the existing message every time it's run like Danger usually does?

AliSoftware (approved):

loremattei:
It shouldn't 🤔 It used to work the same way. 
Let's keep an eye on this to see if it was a hiccup or something it does on every PR.

AliSoftware:
Oh wait now that I'm back on this page I only see one single comment while I saw 3 before. So it fixed itself? (still received 3 distinct email notifications tho)

Maybe it was a GitHub UI refresh bug actually (I've seen those before with labels not refreshing properly for example, I assume because everything is XHR / async refresh… except a couple of corner cases)

PR #3053, opened 2020-10-22
[Refunds] Update deleted refunds on web

### LINKED ISSUE #3052: Deleting refunds on web does not reflect in the app
It seems that when you delete a refund on wen, the app is not able to pick up those changes and displays wrong refund information.

Ecarrion:
fix https://github.com/woocommerce/woocommerce-ios/issues/3052

# Why

While testing the refunds creation feature I noticed that deleting a refund on web does not properly reflect in the app.
The issue happens because the app is only upserting new refunds as they come from web but does not handle the scenario of cleaning refunds deleted on web.

# How

To fix that I've added a method on `RefundStore` that deletes stale(web removed) refunds when indicated on `RefundAction.retrieveRefunds - deleteSlateRefunds flag` 

# Screenshots
**WEB**
core

Before | After
---- | ----
before | after

# Testing Steps
- Navigate to a refunded order
- Tap on refunded items and see items being refunded
- On web, go to that order and delete one or more refunds
- On the app, refresh the order detail and see that the previous refund disappears.

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Actions/RefundAction.swift
+++ b/Yosemite/Yosemite/Actions/RefundAction.swift
@@ -7,7 +7,7 @@ import Networking
 public enum RefundAction: Action {
     case createRefund(siteID: Int64, orderID: Int64, refund: Refund, onCompletion: (Refund?, Error?) -> Void)
     case retrieveRefund(siteID: Int64, orderID: Int64, refundID: Int64, onCompletion: (Refund?, Error?) -> Void)
-    case retrieveRefunds(siteID: Int64, orderID: Int64, refundIDs: [Int64], onCompletion: (Error?) -> Void)
+    case retrieveRefunds(siteID: Int64, orderID: Int64, refundIDs: [Int64], deleteStaleRefunds: Bool, onCompletion: (Error?) -> Void)
     case synchronizeRefunds(siteID: Int64, orderID: Int64, pageNumber: Int, pageSize: Int, onCompletion: (Error?) -> Void)
     case resetStoredRefunds(onCompletion: () -> Void)
 }
--- a/Yosemite/Yosemite/Stores/RefundStore.swift
+++ b/Yosemite/Yosemite/Stores/RefundStore.swift
@@ -36,8 +36,8 @@ public class RefundStore: Store {
             createRefund(siteID: siteID, orderID: orderID, refund: refund, onCompletion: onCompletion)
         case .retrieveRefund(let siteID, let orderID, let refundID, let onCompletion):
             retrieveRefund(siteID: siteID, orderID: orderID, refundID: refundID, onCompletion: onCompletion)
-        case .retrieveRefunds(let siteID, let orderID, let refundIDs, let onCompletion):
-            retrieveRefunds(siteID: siteID, orderID: orderID, refundIDs: refundIDs, onCompletion: onCompletion)
+        case .retrieveRefunds(let siteID, let orderID, let refundIDs, let deleteStaleRefunds, let onCompletion):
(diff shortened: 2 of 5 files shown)


Yosemite/Yosemite/Actions/RefundAction.swift
```diff
@@ -7,7 +7,7 @@ import Networking
 public enum RefundAction: Action {
     case createRefund(siteID: Int64, orderID: Int64, refund: Refund, onCompletion: (Refund?, Error?) -> Void)
     case retrieveRefund(siteID: Int64, orderID: Int64, refundID: Int64, onCompletion: (Refund?, Error?) -> Void)
-    case retrieveRefunds(siteID: Int64, orderID: Int64, refundIDs: [Int64], onCompletion: (Error?) -> Void)
+    case retrieveRefunds(siteID: Int64, orderID: Int64, refundIDs: [Int64], deleteStaleRefunds: Bool, onCompletion: (Error?) -> Void)
```
jaclync, line 10: I'd think we always want to delete stale refunds, do you know if there are any scenarios when we want to set `deleteStaleRefunds: false`?

Yosemite/Yosemite/Stores/RefundStore.swift
```diff
@@ -36,8 +36,8 @@ public class RefundStore: Store {
             createRefund(siteID: siteID, orderID: orderID, refund: refund, onCompletion: onCompletion)
         case .retrieveRefund(let siteID, let orderID, let refundID, let onCompletion):
             retrieveRefund(siteID: siteID, orderID: orderID, refundID: refundID, onCompletion: onCompletion)
-        case .retrieveRefunds(let siteID, let orderID, let refundIDs, let onCompletion):
-            retrieveRefunds(siteID: siteID, orderID: orderID, refundIDs: refundIDs, onCompletion: onCompletion)
+        case .retrieveRefunds(let siteID, let orderID, let refundIDs, let deleteStaleRefunds, let onCompletion):
+            retrieveRefunds(siteID: siteID, orderID: orderID, refundIDs: refundIDs, deleteStaleRefunds: deleteStaleRefunds, onCompletion: onCompletion)
         case .synchronizeRefunds(let siteID, let orderID, let pageNumber, let pageSize, let onCompletion):
```
jaclync, line 41: deleting stale refunds made me wonder how we handle deleted refunds in `synchronizeRefunds`, and it looks like it's not even used 🤔

jaclync (approved):
Good catch on this scenario! Just checking, it makes sense to still show an empty refunded row in the payment section right?

before deleting refund | after deleting refund
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-23 at 18 08 35 | Simulator Screen Shot - iPhone 11 - 2020-10-23 at 18 09 34

In core, the deleted refund row is also shown as $0:

Screen Shot 2020-10-23 at 6 14 39 PM

Also, an unrelated observation: the refunds under payment don't seem to follow the same order as in the screenshots 🤔

Yosemite/Yosemite/Actions/RefundAction.swift
Ecarrion: I added the flag because in the action `retrieveRefunds` you can specify which refunds to retrieve. So If a consumer chooses just to refresh a few they may not want to delete the ones they didn't fetch.

Yosemite/Yosemite/Stores/RefundStore.swift
Ecarrion: Yeah, the only action used `retrieveRefunds` do you think we should delete the rest of them?

Ecarrion:
Thanks for the review @jaclync, regarding your findings I've added a separate issue to continue working on that screen! https://github.com/woocommerce/woocommerce-ios/issues/3055

PR #3054, opened 2020-10-23
Fix: added Core Data model 32 and 33 inside the test for loading the expected model version

pmusolino:
Super minor fix: the method `test_it_can_load_the_expected_model_versions` was not considering the last Core Data models, n°32 and n°33.

## Testing
- Just CI

Update release notes:


Ecarrion (approved):
🚀 
I wonder if we could write a rule in hound or something that alert us when this file is not modified after adding a new CD model 🤔

pmusolino:
@Ecarrion this would be useful, but I think it's easy to add this to our doc :nod:

shiki:
I don't believe it's important to update this as I just wanted the unit test to test that `ManagedObjectModelsInventory` works as originally intended. But it's not a bad thing to keep it up to date also. 🙂

PR #3057, opened 2020-10-23
Add a RELEASE-NOTES format example at the top

shiki:
Asking @loremattei if this is possible. p91TBi-3Ah-p2

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,6 @@
 5.4
 -----
+- [*] PLEASE FOLLOW THIS FORMAT: [<priority indicator>] <description> [<PR URL>]
 - [*] fix: the footer in app Settings is now correctly centered.
 - [*] fix: Products tab: earlier draft products now show up in the same order as in core when sorting by "Newest to Oldest".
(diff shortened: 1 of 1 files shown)


RELEASE-NOTES.txt
```diff
@@ -1,5 +1,6 @@
 5.4
 -----
+- [*] PLEASE FOLLOW THIS FORMAT: [<priority indicator, more stars = higher priority>] <description> [<PR URL>]
```
loremattei: This is great @shiki!
But could we put this on top of the doc, before the first version number (I mean, above `5.4` in this example)?
And could we use an easy to identify sequence at the beginning of the string (like `*** PLEASE FOLLOW ...`)?
This would make processing the file so much easier 😅

shiki: @loremattei Thanks for the feedback! I updated it. How does it look? Will it be a problem for the deployment workflow/automation?

loremattei (approved):
Thank you @shiki!
This looks good to me! We'll update our release tooling to handle it if problems arise.

shiki:
Thanks, Lorenzo!

PR #3048, opened 2020-10-22
[Issue Refunds] Create shipping refunds

Ecarrion:
closes #2843 

# Why

This PR updates the `RefundCreationUseCase` to be able to create a refund object that will refund a shipping line remotely.

# How 

- Added `shippingLine` property on `RefundCreationUseCase` that will create a specific `OrderItemRefund` for refunding shipping values.

# Testing Steps

This code is not yet integrated into the app, but you can try it in the following way.

1. Replace the code in `createRefundConfirmationViewModel` method on `IssueRefundViewModel` line 77 with
```swift
func createRefundConfirmationViewModel() -> RefundConfirmationViewModel {
        let items = state.refundQuantityStore.map { RefundableOrderItem(item: $0, quantity: Decimal($1)) }
        let shippingLine = state.shouldRefundShipping ? state.order.shippingLines.first : nil
        let useCase = RefundCreationUseCase(amount: "\(calculateRefundTotal())",
                                            reason: "iOS Shipping refund",
                                            automaticallyRefundsPayment: false,
                                            items: items,
                                            shippingLine: shippingLine,
                                            currencyFormatter: CurrencyFormatter(currencySettings: state.currencySettings))
        let refund = useCase.createRefund()

        let action = RefundAction.createRefund(siteID: state.order.siteID,
                                               orderID: state.order.orderID,
                                               refund: refund) { _, error  in
            if let error = error {
                print("Error: \(error)")
            }
        }
        ServiceLocator.stores.dispatch(action)

        return RefundConfirmationViewModel(order: state.order, currencySettings: state.currencySettings)
    }
```
This will fire a create refund action after tapping the next button on the issue refund screen.

2. Go to an un-refunded order that has shipping

3. Toggle the refund shipping switch. You can select some order items too.

4. Tap Next

5. Go to the sore admin and see that shipping was refunded correctly.

refund

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3060, opened 2020-10-26
[Issue Refund] Handle free shipping

### LINKED ISSUE #3058: [Issue Refund] Handle Free Shipping refund
We should hide the shipping refund switch if the shipping was free but first, we need to corroborate how core handles this and replicate the behavior.

Ecarrion:
fixes #3058

# Why

While reviewing #3048, @jaclync noticed that enabling the "refund shipping" switch for free shipping orders made not much sense so this PR takes care of hiding that section for that scenario

# How

Use `RefundShippingCalculationUseCase` to know if there is no amount to refund before returning the "refund shipping" section

# Gif
free-shipping-demo

# Testing Steps

- Navigate to an order with free shipping
- Tap on issue refund button
- Corroborate that the "refund shipping" section is not visible 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
@@ -199,6 +199,13 @@ extension IssueRefundViewModel {
             return nil
         }
 
+        // If there is no amount to refund (EG: free shipping), hide the refund shipping section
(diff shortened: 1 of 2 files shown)


jaclync (approved):
LGTM! thanks for handling this edge case 😄 

Code: ✅ 
Testing: ✅ (tested on orders with free shipping and paid shipping!)

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-10-27 at 16 16 44 | Simulator Screen Shot - iPhone 11 - 2020-10-27 at 16 20 26

PR #3064, opened 2020-10-27
Products: add downloadable file from Device or from WordPress Media library

### LINKED ISSUE #2759: Add/Edit Products M5: Downloadable files Step 4 - add a downloadable file for a product
As part of the M5 release, we need to provide the option to add a new downloadable product by:
- choosing a file from the mobile device and uploading the file.
- choosing a file from the media picker library.
- manually entering a file url, the download expiry and download limit.

This issue should also take care to display the downloadable files as part of the bottom sheet, if there are no existing files for the product.
Frame 207

#### Choose to add a downloadable file
iOS-add-price

Clicking on the Add file option should display a bottom sheet with the following options:
Frame 208

#### Choose from device
iOS-product-add-photos

#### Choose from media picker library
iOS-product-add-photos-from-media-picker

#### Manually add a file
iOS-product-add-photos-from-media-picker

Feel free to split this issue into smaller parts!

pmusolino:
Fixes #2759 

## Description
This is the last part for the implementation of the core functionalities of Downloadable Files, part of Products M5. In this PR, I added the possibility to upload a new downloadable file from the device or from the WordPress Media library. I also refactored the `SurveyLoadingView` to be a generic `LoadingView`, since I use it while a new image is uploading.

## Changes
- Now the done button in the Product Download File detail is enabled or disabled based on unsaved changes.
- Implemented all the methods needed for handling the images from the device or from the WordPress media picker in `ProductDownloadListViewController`.
- Refactoring of the `SurveyLoadingView`. I made it more generic and reusable, and the loading view now has the size of the superview. In that way, all the elements in the background are not tappable while the loading view is shown.
- Updated `SurveyViewController` for using the new loading view.
- Updated the `SurveyViewControllerTests` for using the new loading view.

#### Case 1
1. Tap the Products tab.
2. Open a product with downloadable files or enable them for a product (from product settings).
3. Tap the cell "Downloadable Files"
4. Tap "Add File"
5. Press the option in the action sheet "From device".
6. Choose a photo. After tapping it, the uploading will start, showing a loader.
7. When the upload finished, you will see the screen for adding the downloadable file.
8. Tap add and update the product.

#### Case 2
1. Tap the Products tab.
2. Open a product with downloadable files or enable them for a product (from product settings).
3. Tap the cell "Downloadable Files"
4. Tap "Add File"
5. Press the option in the action sheet "From WordPress Media Library".
6. Choose a photo. After tapping it, you will see the screen for adding the downloadable file.
7. Tap add and update the product.

## GIF
| New downloadable file uploading it from device             |  New downloadable file choosing it from WordPress media library |
gif selection from device | selection from wordpress media library

## Screenshot of the loader
I used the same loader used initially for the Survey screen. I've only a concern about it when there are no downloadable files in the list since the loader background is the same as the table view. It's ok, or we should update it @Garance91540? 

| Loader with cell background             |  Loader with empty background |
Simulator Screen Shot - iPhone 11 Pro - 2020-10-26 at 18 26 43 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-26 at 18 41 04

Update release notes:


Ecarrion (commented):
This looks great @pmusolino!

I only have one concern regarding the `LoadingView`. I would prefer if this view was not generalized as I think the big blocking view is a pattern we would not like to extend across our views. 

I'm not sure if we have thought about alternatives like progress bars or any other kind of loading indicator for uploading a file.

What do you think @Garance91540?

pmusolino:
Thank you @Ecarrion for the review! 🙇 

> I only have one concern regarding the LoadingView. I would prefer if this view was not generalized as I think the big blocking view is a pattern we would not like to extend across our views.

I believe that when we show a loader, we always need to block the interaction with the view below. In Survey, it's an exception because it will be shown while the content it's still loading in the web view and nothing will be shown in the meanwhile, but If we allow a user to tap on an area while there is something loading, we can create unexpected behaviors.

> I'm not sure if we have thought about alternatives like progress bars or any other kind of loading indicator for uploading a file.

In products images, while an image is uploading, we don't show a progress bar indicator, but a perpetual loader indicator.
I believe that in the future will be good to have a specific design for this type of loader, with a better appearance, that will be easily recognized and substituted inside the `LoadingView`.

Garance91540 (commented):
@pmusolino is it possible to add a darker layer between the screens and the loading dialog anywhere we use a loading modal?

Screen Shot 2020-10-28 at 5 11 10 PM

pmusolino:
For sure @Garance91540! Is it `black` with an alpha layer? How much `alpha` do you used?

Ecarrion (approved):
@pmusolino Thanks for the answer, let monitor and see if this case happens again to think about a more holistic solution!

PR #3068, opened 2020-10-27
Display "(No Title)" as Product title in the products list when a product doesn't have the title set

### LINKED ISSUE #3049: Display "(No Title)" in Product List when a product doesn't have the title
As discussed here p5T066-1zC#comment-6119 we should display `(No Title)` when a product in the Product List doesn't have the title set. Currently, on iOS, we always hide the title label.

pmusolino:
Fixes #3049 

As discussed here p5T066-1zC#comment-6119, we should display `(No Title)` when a product in the Product List doesn't have the title set.

## Testing
1. Open the product list
2. Make sure that a product without the title in the list shows `(No Title)` 

Update release notes:


Ecarrion (approved):
🚀

PR #3069, opened 2020-10-27
Core Data: Reduce Chances of Recovery Failed Crashes

shiki:
Ref #2667. 

This is an attempt to fix the Recovery Failed crash which seems to be ramping up. 

### Migration

From what I can see, the pattern of all crashes is that they all point to `WooCommerce.sqlite.~` being inaccessible. For example, in this [event](https://sentry.io/share/issue/d1497b3ad58e4cba8249599fd9efc0eb/), when migrating, we have this error:

```
[
⚠️ [CoreDataManager] Migration required for persistent store, 
⚠️ Attempting migration from Model 30 to Model 31, 
Migration error: Error Domain=NSCocoaErrorDomain Code=4 "“WooCommerce.sqlite.~” couldn’t be moved to “backup” because either the former doesn’t exist, or the folder containing the latter doesn’t exist." UserInfo={NSSourceFilePathErrorKey=/var/mobile/Containers/Data/Application/A7E09E5A-41FF-4081-89D7-B1C2384DAC56/Documents/WooCommerce.sqlite.~, NSUserStringVariant=(
    Move
), NSDestinationFilePath=/var/mobile/Containers/Data/Application/A7E09E5A-41FF-4081-89D7-B1C2384DAC56/Documents/backup/WooCommerce.sqlite.~, NSFilePath=/var/mobile/Containers/Data/Application/A7E09E5A-41FF-4081-89D7-B1C2384DAC56/Documents/WooCommerce.sqlite.~, NSUnderlyingError=0x28282c7e0 {Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"}} [["NSSourceFilePathErrorKey": /var/mobile/Containers/Data/Application/A7E09E5A-41FF-4081-89D7-B1C2384DAC56/Documents/WooCommerce.sqlite.~, "NSDestinationFilePath": /var/mobile/Containers/Data/Application/A7E09E5A-41FF-4081-89D7-B1C2384DAC56/Documents/backup/WooCommerce.sqlite.~, "NSUnderlyingError": Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory", "NSUserStringVariant": <__NSSingleObjectArrayI 0x282508170>(
Move
)
, "NSFilePath": /var/mobile/Containers/Data/Application/A7E09E5A-41FF-4081-89D7-B1C2384DAC56/Documents/WooCommerce.sqlite.~]]
, 
☠️ [CoreDataManager] Unable to migrate store.
]
```

Even though we got that file description from `FileManager`, `FileManager` eventually said that the file does not exist. This happens during the [**backup phase**](https://github.com/woocommerce/woocommerce-ios/blob/ebe9ff4cd6425291dc34b02afc9b22f46317f211/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L189-L202). That means that the app never actually got the chance to do the **actual migration** which is supposed to happen right after the backup. 

The app should not also try to back up `WooCommerce.sqlite.~` since that file itself is a backup. We stopped creating that backup file in https://github.com/woocommerce/woocommerce-ios/pull/2816. 

### Recovery 

So, since the migration failed, the app will try to recover, but the recovery fails again when we try to [delete the existing files](https://github.com/woocommerce/woocommerce-ios/blob/ebe9ff4cd6425291dc34b02afc9b22f46317f211/Storage/Storage/CoreData/CoreDataManager.swift#L61-L76):

```
{
Code: 4, 
Description: “WooCommerce.sqlite.~” couldn’t be removed., 
Domain: NSCocoaErrorDomain, 
User Info: ["NSUserStringVariant": <__NSSingleObjectArrayI 0x282509070>(
Remove
)
, "NSFilePath": /var/mobile/Containers/Data/Application/A7E09E5A-41FF-4081-89D7-B1C2384DAC56/Documents/WooCommerce.sqlite.~, "NSUnderlyingError": Error Domain=NSPOSIXErrorDomain Code=2 "No such file or directory"]
}
```

Again, the failure is because of `WooCommerce.sqlite.~`.

## Changes

I still have no idea why these happen. There's a chance that these are because of the `Documents` directory being inaccessible. We haven't really been able to prove that though. 

For now, I'm basing these changes on the guess that there's inherently something wrong with the `WooCommerce.sqlite.~` file and `FileManager` fails to copy it. @jaclync also did a test and found that the sequence of deletion is:

```
WooCommerce.sqlite-wal
WooCommerce.sqlite
WooCommerce.sqlite-shm
WooCommerce.sqlite.~
```

So it could be that the migration backup phase succeeded except for `WooCommerce.sqlite.~`. Though this theory is not consistent for the [recovery phase](https://github.com/woocommerce/woocommerce-ios/blob/develop/Storage/Storage/CoreData/CoreDataManager.swift#L61-L76). If the app finished deleting everything, then it should have fully recovered. 

### Migration Changes

I changed the backup phase of the migration so that the `WooCommerce.sqlite.~` file is not included. Only `WooCommerce.sqlite`, `WooCommerce.sqlite-shm`, and `WooCommerce.sqlite-wal` will be copied to the `backup` folder. 

https://github.com/woocommerce/woocommerce-ios/blob/432bc1f80a6404dd65e3ee88499beb01fcdc9876/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L167-L175

I also made the `backup` directory recreation to fail loudly instead of silently. 

https://github.com/woocommerce/woocommerce-ios/blob/432bc1f80a6404dd65e3ee88499beb01fcdc9876/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L161-L164

This should allow us to catch errors where the `backup` folder was not created, which could lead us to prove the theory that the `Documents` directory is not writeable when these crashes happen. 

I'm not sure if we should be creating the `backup` folder in the `Documents` directory. I think we should do that in the temporary folder so the OS can clean it up for us. I decided not to make this change for now. 

It looks like we can actually use [`replacePersistentStore`](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468917-replacepersistentstore) instead of file operations. But that requires some architectural changes that I'm not comfortable with doing since I'm targeting the release branch. 

### Recovery Changes

For the recovery phase, instead of using file operations, I was able to use [`destroyPersistentStore`](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468888-destroypersistentstore). 

https://github.com/woocommerce/woocommerce-ios/blob/432bc1f80a6404dd65e3ee88499beb01fcdc9876/Storage/Storage/CoreData/CoreDataManager.swift#L82-L84

Based on my observation, the `destroyPersistentStore` method: 

- Does **not** delete the `.sqlite*` files. 
- Instead, it restructures the database so it matches the current `NSManagedObjectModel`. 
- It deletes all the data.

I was able to prove these in [a unit test](https://github.com/woocommerce/woocommerce-ios/blob/432bc1f80a6404dd65e3ee88499beb01fcdc9876/Storage/StorageTests/CoreData/CoreDataManagerTests.swift#L108). 

### Upgrade Regression Test

1. Install 5.2 and run.
2. Run this branch and confirm that the database is still loaded.

### Migration Test

1. Delete the app. 
2. Using this branch, delete Models 31 and 32. 
3. Set the Current Model Version to Model 30.
3. Run the app using a simulator, login but don't load Orders and Products because Model 30 is incompatible and the app will crash. 🤪 
5. Stop the app. 
6. Restore the deleted models and set the Current Model Version to Model 32. 
6. Add a breakpoint in this line:

    https://github.com/woocommerce/woocommerce-ios/blob/32a1305ef210d26b1198fd1fb56b00c07867167f/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L262

7. Open the `Documents` directory and copy `WooCommerce.sqlite` to a new file named `WooCommerce.sqlite.~` (note the two periods).
7. Run the app. This should perform a migration.  
8. When the app stops at the breakpoint, open the simulator's `Documents` directory. Confirm that `WooCommerce.sqlite.~` is still there. Confirm that these files are no longer present:
    - `WooCommerce.sqlite`
    - `WooCommerce.sqlite-wal`
    - `WooCommerce.sqlite-shm`

    However, these files should be present in `Documents/migration`. 

9. Step over. 
10. Check the `Documents` directory again. 
    - Confirm that the `Documents/migration` directory is empty. 
    - Confirm that the `WooCommerce.sqlite`, `*.sqlite-wal`, and `*.sqlite-shm` files are in the `Documents` directory. 
11. Continue running the app. The breakpoint should be hit twice because two migrations will happen. If you repeat the checks in steps 10-12, note that the `*.sqlite-wal` and `*.sqlite-shm` files may no longer be present. Core Data may have merged them into `*.sqlite` during the migration. I'm not quite familiar with _when_ that should happen. 
12. Confirm that you can load the Orders and Products tab now. 

### Recovery Test

1. Delete the app. 
2. Using this branch, run and log in using a simulator.
3. Stop the app. 
4. Add a breakpoint in this line:

    https://github.com/woocommerce/woocommerce-ios/blob/972f62bb45e80f14d8f7cc851d60ac048751a216/Storage/Storage/CoreData/CoreDataManager.swift#L82-L84

5. Open the simulator's Documents directory. Edit the `WooCommerce.sqlite` file with a text editor. Add some gibberish and save it. This will _corrupt_ the file. 😄 
6. Run the app again.
7. The breakpoint should be hit, which means that the app is trying to recover. 
8. Step over until you reach this line **but before it is executed**:

    https://github.com/woocommerce/woocommerce-ios/blob/972f62bb45e80f14d8f7cc851d60ac048751a216/Storage/Storage/CoreData/CoreDataManager.swift#L91

    You will notice logs like this in the console too:

    ```
        CoreData: annotation: <NSPersistentStoreCoordinator: 0x600000b3ccb0>: Attempting recovery from error encountered during addPersistentStore: Error Domain=NSCocoaErrorDomain Code=259 "The file “WooCommerce.sqlite” couldn’t be opened because it isn’t in the correct format." UserInfo={NSFilePath=/Users/shiki/Library/Developer/CoreSimulator/Devices/5B01E811-4206-4D5B-915C-EAD55AAA88C7/data/Containers/Data/Application/0399B895-E830-44D1-8696-A77FC44D844D/Documents/WooCommerce.sqlite, NSSQLiteErrorDomain=26}
    2020-10-27 12:19:30.413268-0600 WooCommerce[18033:3110659] [logging] database corruption page 95 of /Users/shiki/Library/Developer/CoreSimulator/Devices/5B01E811-4206-4D5B-915C-EAD55AAA88C7/data/Containers/Data/Application/0399B895-E830-44D1-8696-A77FC44D844D/Documents/WooCommerce.sqlite at line 72984 of [02c344acea]
    2020-10-27 12:19:30.413412-0600 WooCommerce[18033:3110659] [logging] database corruption at line 71240 of [02c344acea]
    2020-10-27 12:19:30.413531-0600 WooCommerce[18033:3110659] [logging] database corruption page 95 of /Users/shiki/Library/Developer/CoreSimulator/Devices/5B01E811-4206-4D5B-915C-EAD55AAA88C7/data/Containers/Data/Application/0399B895-E830-44D1-8696-A77FC44D844D/Documents/WooCommerce.sqlite at line 73134 of [02c344acea]
    2020-10-27 12:19:30.413619-0600 WooCommerce[18033:3110659] [logging] database corruption at line 71240 of [02c344acea]
    2020-10-27 12:19:30.413746-0600 WooCommerce[18033:3110659] [logging] database disk image is malformed in "PRAGMA main.journal_mode"
    ```

9. Check the Documents directory. Notice that the `WooCommerce.sqlite*` files are still there but their file sizes have been reduced to something like this:

    image

    The `WooCommerce.sqlite` file is typically more than 100KB before the recovery.

10. Continue running the app. Confirm that you can load the Orders and Products tabs. 

## Submitter Checklist


shiki:
@loremattei Is it okay if we include this for 5.3?

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -144,20 +144,34 @@ private extension CoreDataIterativeMigrator {
...
+    /// - {store_filename}.sqlite-shm
+    ///
+    /// Where {store_filename} is most probably "WooCommerce".
+    ///
+    func moveStoreFilesToBackupFolder(storeURL: URL) throws -> URL {
+        let storeFolderURL = storeURL.deletingLastPathComponent()
+        let backupURL = storeFolderURL.appendingPathComponent("backup")
+
+        if fileManager.fileExists(atPath: backupURL.path) {
+            try fileManager.removeItem(at: backupURL)
```
astralbodies: I think this needs to be a `try?` so that if it fails it continues with the next statement.

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -144,20 +144,34 @@ private extension CoreDataIterativeMigrator {
...
+
         do {
-            let files = try fileManager.contentsOfDirectory(atPath: storeURL.deletingLastPathComponent().path)
-            try files.forEach { (file) in
-                if file.hasPrefix(storeURL.lastPathComponent) {
-                    let fullPath = storeURL.deletingLastPathComponent().appendingPathComponent(file).path
-                    let toPath = URL(fileURLWithPath: backupURL.path).appendingPathComponent(file).path
-                    try fileManager.moveItem(atPath: fullPath, toPath: toPath)
-                }
+            try fileManager.contentsOfDirectory(atPath: storeFolderURL.path).map { fileName in
```
astralbodies, line 166: This is pretty darn elegant 😄

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -144,20 +144,34 @@ private extension CoreDataIterativeMigrator {
...
+    /// Where {store_filename} is most probably "WooCommerce".
+    ///
+    func moveStoreFilesToBackupFolder(storeURL: URL) throws -> URL {
+        let storeFolderURL = storeURL.deletingLastPathComponent()
+        let backupURL = storeFolderURL.appendingPathComponent("backup")
+
+        if fileManager.fileExists(atPath: backupURL.path) {
+            try fileManager.removeItem(at: backupURL)
+        }
+        try fileManager.createDirectory(at: backupURL, withIntermediateDirectories: false, attributes: nil)
```
astralbodies: Same here, I think it needs to be `try?` to prevent stopping of execution.

astralbodies (changes requested):
Just a little tweak; I will test more tomorrow as well after that change!

Storage/Storage/CoreData/CoreDataManager.swift
```diff
@@ -62,15 +79,9 @@ public final class CoreDataManager: StorageManagerType {
...
-                let files = try fileManager.contentsOfDirectory(atPath: pathToStore)
-                try files.forEach { (file) in
-                    if file.hasPrefix(self.storeURL.lastPathComponent) {
-                        let fullPath = URL(fileURLWithPath: pathToStore).appendingPathComponent(file).path
-                        try fileManager.removeItem(atPath: fullPath)
-                    }
-                }
+                try container.persistentStoreCoordinator.destroyPersistentStore(at: self.storeURL,
+                                                                                ofType: storeDescription.type,
+                                                                                options: nil)
```
jaclync, line 87: Nice replacing `FileManager` operations with PSC call 👍

Storage/Storage/CoreData/CoreDataManager.swift
```diff
@@ -14,19 +14,25 @@ public final class CoreDataManager: StorageManagerType {
 
     private let modelsInventory: ManagedObjectModelsInventory
 
-    /// Designated Initializer.
+    /// Module-private designated Initializer.
     ///
     /// - Parameter name: Identifier to be used for: [database, data model, container].
     /// - Parameter crashLogger: allows logging a message of any severity level
```
jaclync, line 20: nit: documentation for the new parameter `modelsInventory` could be helpful

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
jaclync: I believe this is intentional! In the PR description:

> I also made the `backup` directory recreation to fail loudly instead of silently. CoreDataIterativeMigrator.swift L161-L164
> This should allow us to catch errors where the `backup` folder was not created, which could lead us to prove the theory that the `Documents` directory is not writeable when these crashes happen.

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -144,20 +144,34 @@ private extension CoreDataIterativeMigrator {
...
-                    try fileManager.moveItem(atPath: fullPath, toPath: toPath)
-                }
+            try fileManager.contentsOfDirectory(atPath: storeFolderURL.path).map { fileName in
+                storeFolderURL.appendingPathComponent(fileName)
+            }.filter { fileURL in
+                // Only include files that have the same filename as the store (sqlite) filename.
+                fileURL.deletingPathExtension() == storeURL.deletingPathExtension()
+            }.forEach { fileURL in
+                let toPath = backupURL.appendingPathComponent(fileURL.lastPathComponent).path
+                try fileManager.moveItem(atPath: fileURL.path, toPath: toPath)
```
jaclync: nit: could we also use `URL` instead of path string here? In the [Apple documentation on `FileManager`](https://developer.apple.com/documentation/foundation/filemanager):

> The use of the NSURL class is generally preferred for specifying file-system items because URLs can convert path information to a more efficient representation internally. 

```suggestion
                let toURL = backupURL.appendingPathComponent(fileURL.lastPathComponent)
                try fileManager.moveItem(at: fileURL, to: toURL)
```

and adding this to `FileManagerProtocol`:

```swift
func moveItem(at srcURL: URL, to dstURL: URL) throws
```

Storage/StorageTests/Mocks/SpyFileManager.swift
```diff
@@ -0,0 +1,59 @@
+import Foundation
+
+import XCTest
+
+@testable import Storage
+
+/// Unlike `MockFileManager`, this performs the actual `FileManager` operations but can also
+/// record or stub method calls if needed.
```
jaclync, line 8: this is cool! I was wondering if it's possible to reset all the file operations after this class is deinitialized?

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
jaclync: +1!

Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
```diff
@@ -182,6 +182,50 @@ final class CoreDataIterativeMigratorTests: XCTestCase {
         XCTAssertNotNil(ps)
     }
 
+    func test_iterativeMigrate_creates_backups_of_the_SQLite_files() throws {
```
jaclync: this test case is 💯 !!

Storage/StorageTests/CoreData/CoreDataManagerTests.swift
```diff
@@ -104,4 +104,73 @@ class CoreDataManagerTests: XCTestCase {
...
+    func deleteStoreFiles(at storeURL: URL) throws {
+        let fileManager = FileManager.default
+        let expectedExtensions = ["sqlite", "sqlite-wal", "sqlite-shm"]
+
+        try expectedExtensions.forEach { ext in
+            let fileURL = storeURL.deletingPathExtension().appendingPathExtension(ext)
+            if fileManager.fileExists(atPath: fileURL.path) {
+                try fileManager.removeItem(at: fileURL)
+            }
+        }
```
jaclync, line 174: nit: would `CoreDataManager.persistentContainer.destroyPersistentStore(...)` also work here?

jaclync:
Thanks a lot for taking on this mysterious crash surge @shiki! 🙏 The unit tests are really nice, and the testing steps are thorough 💯 

Upgrade Regression Test: ✅ (installed 5.2.0.2 from TestFlight, then launched the app from the PR branch)
Migration Test: ✅ (I think the breakpoint is to be placed at `moveStoreFilesToBackupFolder` in step 7 since the crash happened when copying files from store folder to `backup`?)
Recovery Test: ✅ (I performed step 5 when [loading persistent store](https://github.com/woocommerce/woocommerce-ios/blob/972f62bb45e80f14d8f7cc851d60ac048751a216/Storage/Storage/CoreData/CoreDataManager.swift#L71) for the first time and was able to go through the following steps without relaunching the app)

> I'm not sure if we should be creating the backup folder in the Documents directory. I think we should do that in the temporary folder so the OS can clean it up for us. I decided not to make this change for now.

yea a temporary folder for backup files also makes more sense to me, but not necessary to include in a hotfix.

> It looks like we can actually use replacePersistentStore instead of file operations. But that requires some architectural changes that I'm not comfortable with doing since I'm targeting the release branch.

it'd be awesome if we can replace as much as `FileManager` operations in `CoreDataIterativeMigrator` and potentially simplifies the backup process during migration. I agree that it's too risky to make significant architectural changes in a hotfix.

jaclync (approved):

loremattei:
@shiki yes, please ping me as soon as this is merged so that I'll release a new beta as quickly as possible.

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: @astralbodies What Jaclyn said is correct. I forgot to highlight that the `FileManager` error is not very clear:

```
Error Domain=NSCocoaErrorDomain Code=4 "“WooCommerce.sqlite.~” 
couldn’t be moved to “backup” because either the former doesn’t exist, or the 
folder containing the latter doesn’t exist.
```

The problem can either be:

- The `backup` folder does not exist
- The `WooCommerce.sqlite.~` does not exist

Here, I'm intentionally making things fail if the `backup` folder fails to be recreated so that we'll have more info if that is indeed the problem. 

Another thing we could do is to actually just not create a backup at all since we don't do anything with it. 😬 What do you think?

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: Responded in [the thread](https://github.com/woocommerce/woocommerce-ios/pull/3069/files#r513120209) above . 🙂

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: Thank you! I definitely missed that we're using `FileManagerProtocol` and to think I was the one who added it in `IterativeMigrator`. 😄 

Fixed in b6d2b2121290b76bf6c8b1c633fbfbce04972524 and 9d7b883702bc216c0ced62a7b3bcc727881a848f.

Storage/Storage/CoreData/CoreDataManager.swift
shiki: Thanks! Added in 8eab54ec772647c5fda5c4432043bfd4e5c19de1.

Storage/StorageTests/CoreData/CoreDataManagerTests.swift
shiki: Probably not since `destroyPersistentStore` does not delete the files so it's not fit for the use case. I really wanted the files to be killed before the unit test is run. 😄

Storage/StorageTests/Mocks/SpyFileManager.swift
shiki: Sorry, I'm not sure what you mean by _reset all the file operations_. 😅 Please let me know! There's probably an opportunity to improve this documentation. 😬

shiki (commented):
Thank you for the reviews, @astralbodies and @jaclync. I responded to your comments and this is ready for another look. 🙂

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
astralbodies: @shiki Skipping the backup makes the most sense. The only reason we backed things up was because the original ALIterativeMigrator did it. We *never* wrote code to use that file. If migration fails, the db should just get whacked.

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
astralbodies: @shiki What happens if the backup folder exists from a previous migration? Will this fail?

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: As you wish! I shall happily delete that backup routine.

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: The backup folder is deleted right before this line is executed so this should pass. 

```swift
        if fileManager.fileExists(atPath: backupURL.path) {
            try fileManager.removeItem(at: backupURL)
        }
```

But this is no longer relevant since I removed the backup logic. 🙂

shiki (commented):
@astralbodies @jaclync Based on Aaron's suggestion, I removed the backup logic so there are lesser file operations now! 🎉 

Here is the diff of the changes: https://github.com/woocommerce/woocommerce-ios/pull/3069/files/972f62bb45e80f14d8f7cc851d60ac048751a216..32a1305ef210d26b1198fd1fb56b00c07867167f

I have also updated the Migration and Recovery Testing Steps above. ☝️

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
jaclync: note for future: maybe we can leave the more significant changes (like changing the migration flow) for later next time, since it's a hotfix and already tested :)

Storage/StorageTests/Mocks/SpyFileManager.swift
jaclync: Ohh I was just thinking if the file system operations in unit tests might be altering the file system and might have some side effects on other unit tests that depend on the file system. I don't think this is a problem now!

Storage/StorageTests/CoreData/CoreDataManagerTests.swift
jaclync: interesting, looks like they used to have an API to delete all the store files but it's deprecated and I couldn't find much documentation on the alternatives: [removeUbiquitousContentAndPersistentStore](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468923-removeubiquitouscontentandpersis). Deleting the files with `FileManager` sounds good here 👍

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -144,40 +144,55 @@ private extension CoreDataIterativeMigrator {
...
-                if file.hasPrefix(storeURL.lastPathComponent) {
-                    let fullPath = storeURL.deletingLastPathComponent().appendingPathComponent(file).path
-                    let toPath = URL(fileURLWithPath: backupURL.path).appendingPathComponent(file).path
-                    try fileManager.moveItem(atPath: fullPath, toPath: toPath)
-                }
+            try fileManager.contentsOfDirectory(atPath: storeFolderURL.path).map { fileName in
+                storeFolderURL.appendingPathComponent(fileName)
+            }.filter { fileURL in
+                // Only include files that have the same filename as the store (sqlite) filename.
+                fileURL.deletingPathExtension() == storeURL.deletingPathExtension()
```
jaclync, line 170: nit: would it make sense to also delete the `*.~` backup file altogether? though, not worth the risk for the hotfix since the goal is to not touch the `*.~` backup file 😐

jaclync (approved):
Changes look good! :shipit: it'd be nice to leave more significant changes for outside of the hotfix after testing next time 😄 

Upgrade Regression Test: ✅ 
Migration test: ✅  with the latest changes to just delete the store files in `Documents`, there are no store files in step 8 anymore. I created the `*.sqlite.~` by copying the db file in `Documents/migration` folder and then moved to `Documents` folder.
Recovery Test: ✅ like last round of testing, because the simulator store URL changes every launch I performed step 5 when [loading persistent store](https://github.com/woocommerce/woocommerce-ios/blob/972f62bb45e80f14d8f7cc851d60ac048751a216/Storage/Storage/CoreData/CoreDataManager.swift#L71) for the first time and was able to go through the following steps without relaunching the app

astralbodies (approved):
Looks good!!

astralbodies:
Thank you for the effort here @shiki and @jaclync! 🥳

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: Yeah. I think I wanna leave that file alone. It's probably cursed. It's the file that always shows up in the crash logs. 😟

PR #3070, opened 2020-10-27
Fix StorateType_loadOrderItemTax method

Ecarrion:
closes https://github.com/woocommerce/woocommerce-ios/projects/17#card-47838232

# Why 

While working on refunds I noticed that the `loadOrderItemTax` was not properly passing the `itemID` parameter to the search predicate. This was used in `OrdersItemUpsertUseCase` to know if an `OrderItemTax` already existed, and since the method always failed to match an `OrderItemTax` the app was inserting new copies on every order synchronization.

# How
The fix is to pass the `itemID` value to the predicate

# Testing
- Make sure unit tests pass

Update release notes:

### DIFF
--- a/Yosemite/YosemiteTests/Stores/Order/OrdersUpsertUseCaseTests.swift
+++ b/Yosemite/YosemiteTests/Stores/Order/OrdersUpsertUseCaseTests.swift
@@ -85,7 +85,7 @@ final class OrdersUpsertUseCaseTests: XCTestCase {
             Networking.OrderItemTax(taxID: 2, subtotal: "", total: "0.2"),
             Networking.OrderItemTax(taxID: 3, subtotal: "", total: "0.6")
         ]
-        let item = sampleOrderItem(itemID: 22, taxes: taxes)
+        let item = makeOrderItem(itemID: 22, taxes: taxes)
         let order = makeOrder().copy(orderID: 98).copy(items: [item])
         let useCase = OrdersUpsertUseCase(storage: viewStorage)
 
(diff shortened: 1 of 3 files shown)


Yosemite/YosemiteTests/Stores/Order/OrdersUpsertUseCaseTests.swift
```diff
@@ -102,6 +123,23 @@ final class OrdersUpsertUseCaseTests: XCTestCase {
 }
 
 private extension OrdersUpsertUseCaseTests {
+
+    func sampleOrderItem(itemID: Int64, taxes: [Networking.OrderItemTax]) -> Networking.OrderItem {
```
jaclync: nit: maybe a name that is consistent with `makeOrder` below?
```suggestion
    func makeOrderItem(itemID: Int64, taxes: [Networking.OrderItemTax]) -> Networking.OrderItem {
```

Yosemite/YosemiteTests/Stores/OrderStoreTests.swift
```diff
@@ -360,7 +360,7 @@ final class OrderStoreTests: XCTestCase {
         XCTAssertEqual(storageOrder2?.toReadOnly(), sampleOrderMutated2())
         XCTAssertEqual(viewStorage.countObjects(ofType: Storage.Order.self), 1)
         XCTAssertEqual(viewStorage.countObjects(ofType: Storage.OrderItem.self), 1)
-        XCTAssertEqual(viewStorage.countObjects(ofType: Storage.OrderItemTax.self), 5)
+        XCTAssertEqual(viewStorage.countObjects(ofType: Storage.OrderItemTax.self), 4)
```
jaclync, line 363: nice that the issue is caught in a test case!

unrelated to the PR: would be nice for the sample orders in this test case using `copy` so it's more clear what fields change :) (it's a bit hard to follow what's being upserted and its expected changes right now)

jaclync (approved):
Great catch! and nice post coming from this fix p91TBi-3BJ-p2 😃

PR #3077, opened 2020-10-28
[Issue Refunds] Fix refund values format

Ecarrion:
# Why

I noticed two "edge?" cases when refunding items:

1. Format of values that were bigger than `999` had a thousand separator that the API can't read well.
2. Some values were being rounded down after when the value had more than two decimals EG: 13.456 was 13.45

# How

To fix this I'm migrating away from `CurrencyFormatter.localize()` when creating the refund object, since we want the raw number without any thousand separations and rounding.

Also I'm adding a `round up` rule on `CurrencyFormatter.localize()` so values like `13.456` are displayed as `13.46` and not `13.45` like core does it.

# Screenshots
**Before:**
App | Web
---- | ----
before-app | before-refund

**Tax was rendered as `121.14` and value refunded was `121.14`**

**After**
App | Web
---- | ----
Screen Shot 2020-10-28 at 9 12 53 AM | Screen Shot 2020-10-28 at 10 32 27 AM

**Tax is rendered as `121.15` and value refunded is `121.145`**
**Refunding the other item makes the refund complete**

Screen Shot 2020-10-28 at 11 01 20 AM

# Testing steps

This code is not yet integrated into the app, but you can try it in the following way.

1. Replace the code in `createRefundConfirmationViewModel` method on `IssueRefundViewModel` line 77 with
```swift
func createRefundConfirmationViewModel() -> RefundConfirmationViewModel {
        let items = state.refundQuantityStore.map { RefundableOrderItem(item: $0, quantity: Decimal($1)) }
        let shippingLine = state.shouldRefundShipping ? state.order.shippingLines.first : nil
        let useCase = RefundCreationUseCase(amount: "\(calculateRefundTotal())",
                                            reason: "iOS Shipping refund",
                                            automaticallyRefundsPayment: false,
                                            items: items,
                                            shippingLine: shippingLine,
                                            currencyFormatter: CurrencyFormatter(currencySettings: state.currencySettings))
        let refund = useCase.createRefund()

        let action = RefundAction.createRefund(siteID: state.order.siteID,
                                               orderID: state.order.orderID,
                                               refund: refund) { _, error  in
            if let error = error {
                print("Error: \(error)")
            }
        }

        return RefundConfirmationViewModel(order: state.order, currencySettings: state.currencySettings)
    }
```
This will fire a create refund action after tapping the next button on the issue refund screen.

2. Create an order where the final tax value(tax * quantity) has an odd number.
3. Refund items until the order gets totally refunded
3. Corroborate in the website that everything makes sense!

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
+++ b/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
@@ -81,9 +81,9 @@ public class CurrencyFormatter {
     ///       Note this assumption will be wrong for India.
     ///
     func localize(_ decimalAmount: Decimal,
-                  with decimalSeparator: String? = ".",
-                  in decimalPosition: Int = 2,
-                  including thousandSeparator: String? = ",") -> String? {
+                  decimalSeparator: String? = ".",
+                  decimalPosition: Int = 2,
+                  thousandSeparator: String? = ",") -> String? {
         localize(decimalAmount as NSDecimalNumber, with: decimalSeparator, in: decimalPosition, including: thousandSeparator)
     }
 
--- a/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
+++ b/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
@@ -67,6 +67,7 @@ public class CurrencyFormatter {
         numberFormatter.generatesDecimalNumbers = true
         numberFormatter.minimumFractionDigits = decimalPosition
         numberFormatter.maximumFractionDigits = decimalPosition
+        numberFormatter.roundingMode = .halfUp
 
         return numberFormatter.string(from: absoluteAmount)
     }
--- a/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
+++ b/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
@@ -82,9 +82,9 @@ public class CurrencyFormatter {
     ///       Note this assumption will be wrong for India.
     ///
     func localize(_ decimalAmount: Decimal,
-                  decimalSeparator: String? = ".",
-                  decimalPosition: Int = 2,
-                  thousandSeparator: String? = ",") -> String? {
+                  with decimalSeparator: String? = ".",
+                  in decimalPosition: Int = 2,
+                  including thousandSeparator: String? = ",") -> String? {
         localize(decimalAmount as NSDecimalNumber, with: decimalSeparator, in: decimalPosition, including: thousandSeparator)
     }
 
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
```diff
@@ -67,6 +67,7 @@ public class CurrencyFormatter {
         numberFormatter.generatesDecimalNumbers = true
         numberFormatter.minimumFractionDigits = decimalPosition
         numberFormatter.maximumFractionDigits = decimalPosition
+        numberFormatter.roundingMode = .halfUp
```
pmusolino, line 70: Are we sure we want to follow the same behavior as the web? I'm not sure this is the right way to handle this. I've some concerns about the payments and the "bureaucracy" around it.
Eg:
If the price is `13.456$`, probably in your bank account you will receive a payment of `13.45$`. If we show `13.46$`, we are showing the wrong info to the user. Maybe it is the web that should adapt to show the correct information? 🤔

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/UseCases/RefundCreationUseCase.swift
```diff
@@ -106,22 +106,20 @@ struct RefundCreationUseCase {
...
-
-
     /// Calculates the refundable tax from a tax line by diving its total tax value by the purchased quantity and mutiplying it by the refunded quantity.
     ///
     private func calculateTax(of taxLine: OrderItemTax, purchasedQuantity: Decimal, refundQuantity: Decimal) -> String {
         let totalTax = currencyFormatter.convertToDecimal(from: taxLine.total) ?? 0
         let itemTax = (totalTax as Decimal) / purchasedQuantity
         let refundableTax = itemTax * refundQuantity
-        return currencyFormatter.localize(refundableTax) ?? "\(refundableTax)"
+        return "\(refundableTax)"
```
pmusolino, line 115: I think that here is correct to not localize the final value 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/UseCases/RefundCreationUseCase.swift
```diff
@@ -106,22 +106,20 @@ struct RefundCreationUseCase {
...
     }
 
     /// Calculates the refundable total price from a `RefundableOrderItem` by diving the item price value by the purchased quantity
     /// and mutiplying it by the refunded quantity.
     ///
     private func calculateTotal(of refundable: RefundableOrderItem) -> String {
         let total = (refundable.item.price as Decimal) * refundable.decimalQuantity
-        return currencyFormatter.localize(total) ?? "\(total)"
+        return "\(total)"
```
pmusolino, line 123: 👍

pmusolino (approved):
Thanks, @Ecarrion for working on this!

Everything looks good to me. I've just a question about the `halfUp` rounding mode.

WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift
Ecarrion: Good point, I'll make sure to ask core why and who are they rounding up numbers.

PR #3078, opened 2020-10-29
Order Details > variation details: storage layer changes

jaclync:
Storage layer changes for #2280

This PR only contains changes in the storage layer, since the Networking/Yosemite changes are going to be huge as well 🙂 

## Changes

- Schema updates (please review the diffs on model 35 in https://github.com/woocommerce/woocommerce-ios/commit/2c63d3fcef2155e2fdda64909343faa6506c275c#diff-2f9c7ea9bf9793125e3c162b4fcb6dbd5ff9c50054f30aacfc2b996b9c11de71)
  - Added `OrderItemAttribute` entity
  - Added  `attributes: [OrderItemAttribute]` relationship to `OrderItem`
- Added a unit test for migrating from mode 34 to 35 in `MigrationTests`

## Testing

Please run a confidence check on migration!

- Launch the app from `develop` and log in or stay logged in
- Stop the app
- Launch the app from PR branch --> the app should launch in logged in state with the same data

Update release notes:

### DIFF
--- a/Storage/Storage/Model/MIGRATIONS.md
+++ b/Storage/Storage/Model/MIGRATIONS.md
@@ -2,7 +2,7 @@
 
 This file documents changes in the WCiOS Storage data model. Please explain any changes to the data model as well as any custom migrations.
 
-## Model 35 (Release 5.4.0.0)
+## Model 35 (Release 5.5.0.0)
 - @jaclync 2020-10-29
 - Added `OrderItemAttribute` entity.
 - Added  `attributes: [OrderItemAttribute]` relationship to `OrderItem`.
--- a/Storage/Storage/Model/MIGRATIONS.md
+++ b/Storage/Storage/Model/MIGRATIONS.md
@@ -2,6 +2,11 @@
 
 This file documents changes in the WCiOS Storage data model. Please explain any changes to the data model as well as any custom migrations.
 
+## Model 35 (Release 5.4.0.0)
+- @jaclync 2020-10-29
+- Added `OrderItemAttribute` entity.
+- Added  `attributes: [OrderItemAttribute]` relationship to `OrderItem`.
+
 ## Model 34 (Release 5.4.0.0)
 - @ecarrion 2020-10-21
 - Added `ShippingLineTax` entity.
--- a/Storage/Storage/Model/OrderItemAttribute+CoreDataProperties.swift
+++ b/Storage/Storage/Model/OrderItemAttribute+CoreDataProperties.swift
@@ -1,0 +1,16 @@
+import Foundation
+import CoreData
+
+
+extension OrderItemAttribute {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<OrderItemAttribute> {
+        return NSFetchRequest<OrderItemAttribute>(entityName: "OrderItemAttribute")
+    }
+
+    @NSManaged public var metaID: Int64
+    @NSManaged public var name: String
+    @NSManaged public var value: String
+    @NSManaged public var orderItem: OrderItem?
+
+}
--- a/Storage/Storage.xcodeproj/project.pbxproj
+++ b/Storage/Storage.xcodeproj/project.pbxproj
@@ -160,6 +160,7 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
+		021EAA42254938AA00AA8CCD /* Model 35.xcdatamodel */ = {isa = PBXFileReference; lastKnownFileType = wrapper.xcdatamodel; path = "Model 35.xcdatamodel"; sourceTree = "<group>"; };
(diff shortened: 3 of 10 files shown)


Storage/Storage/Model/MIGRATIONS.md
```diff
@@ -2,6 +2,11 @@
 
 This file documents changes in the WCiOS Storage data model. Please explain any changes to the data model as well as any custom migrations.
 
+## Model 35 (Release 5.4.0.0)
```
pmusolino: Probably it's better to merge this PR in release `5.5.0.0`?

Storage/Storage/Model/OrderItemAttribute+CoreDataProperties.swift
```diff
@@ -0,0 +1,16 @@
...
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<OrderItemAttribute> {
+        return NSFetchRequest<OrderItemAttribute>(entityName: "OrderItemAttribute")
+    }
+
+    @NSManaged public var metaID: Int64
+    @NSManaged public var name: String
+    @NSManaged public var value: String
+    @NSManaged public var orderItem: OrderItem?
+
+}
```
pmusolino, line 16: Do you think that it worth adding the optional string field `type` that [will be introduced in the future](https://a8c.slack.com/archives/CGPNUU63E/p1603926767142700?thread_ts=1602796453.088400&cid=CGPNUU63E)?

pmusolino (approved):
Thanks, @jaclync for introducing the new entity.

I've left some comments and I've two questions:
- Can I ask you to add the `Model 35` also inside the test method `test_it_can_load_the_expected_model_versions` under `ManagedObjectModelsInventoryTests` please?
- An `OrderItemAttribute` can be used by other Order Items or they are unique? If they can be reused by other Order Items, probably it makes sense to change the Delete Rule of `attributes` inside `OrderItem` from `Cascade` to `Nullify`.

Storage/Storage/Model/OrderItemAttribute+CoreDataProperties.swift
jaclync: Hmm I was thinking we can add this when the value is available in the API (the migration doesn't require a mapping model, we can set its default to empty string) - @shiki what do you think, since you're thinking about adding `type`?

Storage/Storage/Model/MIGRATIONS.md
jaclync: Hah yea, let's start following the no-merge-big-changes-on-Thursday/Friday-before-code-freeze guideline! I'll merge it after code freeze on my Tuesday next week

jaclync:
> - Can I ask you to add the `Model 35` also inside the test method `test_it_can_load_the_expected_model_versions` under `ManagedObjectModelsInventoryTests` please?

interesting, I didn't see this item in our [Core Data migration documentation](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/creating-core-data-model-versions.md) and the test case `test_it_can_load_the_expected_model_versions` intentionally not fails on missing latest model versions. @shiki do you think it makes sense to just enforce the complete model version names in the test case? like removing the truncation here:

https://github.com/woocommerce/woocommerce-ios/blob/bb6e0a7defc4ba99e1955f2060e2c35e63e9aeec/Storage/StorageTests/CoreData/ManagedObjectModelsInventoryTests.swift#L74-L77

just wondering if you have other concerns since you created the test case this way :) @pmusolino I also added "Model 35" in https://github.com/woocommerce/woocommerce-ios/pull/3078/commits/4a3692151dbbb5d04553ae8712030170f0b7f0d1

> - An `OrderItemAttribute` can be used by other Order Items or they are unique? If they can be reused by other Order Items, probably it makes sense to change the Delete Rule of `attributes` inside `OrderItem` from `Cascade` to `Nullify`.

I don't think an `OrderItemAttribute` is reused by more than one `OrderItem`, since an attribute is specific to an order item. I think it's easier to maintain this way, and will be more obvious in the next PR with Yosemite/Networking changes

Storage/Storage/Model/MIGRATIONS.md
jaclync: updated in https://github.com/woocommerce/woocommerce-ios/pull/3078/commits/26050612ded63832d6dd6d1824a3869a8d34e5c9

Storage/Storage/Model/OrderItemAttribute+CoreDataProperties.swift
shiki: Let's probably wait because that will have to through a code review. Maybe the Core team will not like it and suggest something different. 🙂

jaclync:
Merging the Core Data schema changes now that `develop` is 5.5!

shiki:
> @shiki do you think it makes sense to just enforce the complete model version names in the test case? like removing the truncation here:

@jaclync Sorry I missed this. 

I never intended for that test to fully capture all the models. That test is really just proving that it loads model versions in sequence. It's a specifically scoped unit test for the `ManagedObjectModelsInventory` struct itself and not for the model versions. 

It doesn't need to be updated but it doesn't hurt to update it. 🙂  I don't believe we should enforce it though. It's an unnecessary addition to the process that doesn't provide a lot of benefits in my opinion.

PR #3079, opened 2020-10-29
Remove `OrderItem`'s custom `Equatable` implementation

jaclync:
Prerequisite for #2280

## Changes

While I was working the Yosemite & Networking layer changes for #2280, I removed the custom implementation of `OrderItem`'s `Equatable` since I was adding `attributes` property and then I noticed that it actually already ignores `taxes` on `develop`. It's dangerous to have a custom `Equatable` implementation because we don't have a good way to enforce comparisons on every property. I think it's a good practice that we use the default `Equatable` implementation whenever possible.

This PR:
- Removed the custom implementation of `OrderItem`'s `Equatable`
- Updated `Storage.OrderItem`'s conversion to readonly type by specifying an order for `taxes` relationship by its `taxID` tentatively (it's currently a `Set` in storage layer) so that we can expect the same result in unit tests
- Fixed unit tests

Update release notes:

### DIFF
--- a/Networking/Networking/Model/OrderItem.swift
+++ b/Networking/Networking/Model/OrderItem.swift
@@ -117,22 +117,7 @@ private extension OrderItem {
 
 // MARK: - Comparable Conformance
 //
-extension OrderItem: Comparable {
-    public static func == (lhs: OrderItem, rhs: OrderItem) -> Bool {
-        return lhs.itemID == rhs.itemID &&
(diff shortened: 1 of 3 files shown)


pmusolino (approved):
LGTM :shipit: 

> I think it's a good practice that we use the default Equatable implementation whenever possible.

I agree with you. We have a lot of custom `Equatable` implementation that can be removed in favor of the default implementation.

Issue #3081, opened 2020-10-29
Disabled state on buttons not working


Ecarrion:
**Describe the bug**
Our `applyPrimaryButtonStyle` method on `UIButton` seems to be applying the wrong style for the disabled state.

**To Reproduce**
1. Apply `applyPrimaryButtonStyle` to any button
2. Set it as disabled `button.isEnabled = false`
3. See wrong style applied

**Screenshots**
Current | Expected
---- | ----
<img width="439" alt="bad" src="https://user-images.githubusercontent.com/562080/97575197-60963400-19ba-11eb-9675-5960cbb1c39b.png"> | <img width="131" alt="good" src="https://user-images.githubusercontent.com/562080/97575201-612eca80-19ba-11eb-8e85-06c38fea8847.png">

**Mobile Environment**
Happening on iOS 13 and 14, both simulator and device

br3akzero:
I would like to do this if this task is still active ?

Ecarrion:
Sure @Denpex, let us know if you need any assistance. 👍

jaclync:
I think the design for disabled primary button is the following (in "System - iOS Components" design board):

@Garance91540 feel free to correct!

\ | Light mode | Dark mode
-- | -- | --
Design | <img width="381" alt="Screen Shot 2020-10-30 at 3 12 49 PM" src="https://user-images.githubusercontent.com/1945542/97671438-3c724a00-1ac3-11eb-85b5-b0c77b48b3c7.png"> | <img width="357" alt="Screen Shot 2020-10-30 at 3 12 58 PM" src="https://user-images.githubusercontent.com/1945542/97671471-4eec8380-1ac3-11eb-89df-981962d2c36d.png">
Text color | iOS System Quaternary Label Color | iOS System Tertiary Label Color
Border color | Same as text color | Same as text color

br3akzero:
@jaclync thanks for the heads up, I will make sure to follow these guidelines.

Garance91540:
That's correct @jaclync , thank you!

Ecarrion:
Hi @Denpex do you still plan to work on this?

PR #3083, opened 2020-10-29
[Issue Refunds] Disable next button on Issue Refund Screen

### LINKED ISSUE #3066: [Issue Refunds] Disable Next button when there is no item selected for refund
We should allow the user to navigate further when;
- Has selected at least one item for refund
or
- Has enabled the shipping refund switch

Ecarrion:
closes #3066 

# Why

There is no point in letting a user refund on items, so this PR disables the next button of the "Issue Refund" screen if there is no item selected to refund.

PS: While doing this I noticed that there is a problem with the disabled state of the buttons and since it's a broader problem I'll tackle it in a separate PR https://github.com/woocommerce/woocommerce-ios/issues/3081

# How

- Added a `isNextButtonEnabled` property on `IssueRefundViewModel` which is calculated everytime the `state` struct mutates.
- Updates `IssueRefundViewController` to change the button enabled state.

# Screenshot
button gif

# Testing Steps

- Go to an no-refunded order
- Tap issue refund button
- See next button disabled
- Select an item or the shipping refund switch
- See the next button enabled

Update release notes:

### DIFF
--- a/WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
@@ -342,4 +342,52 @@ final class IssueRefundViewModelTests: XCTestCase {
         // Price is 11.50 and tax is 0.99 (2.97 / 3(quantity))
         XCTAssertEqual(viewModel.title, "$12.49")
     }
+
+    func test_viewModel_starts_with_next_button_disabled() {
+        // Given
+        let currencySettings = CurrencySettings()
+        let items = [
+            MockOrderItem.sampleItem(itemID: 1, quantity: 3, price: 11.50),
+        ]
+        let order = MockOrders().makeOrder(items: items)
+
+        // When
+        let viewModel = IssueRefundViewModel(order: order, refunds: [], currencySettings: currencySettings)
+
+        // Then
+        XCTAssertFalse(viewModel.isNextButtonEnabled)
+    }
+
+    func test_viewModel_next_button_gets_enabled_after_selecting_items() {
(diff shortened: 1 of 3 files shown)


WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -342,4 +342,52 @@ final class IssueRefundViewModelTests: XCTestCase {
         // Price is 11.50 and tax is 0.99 (2.97 / 3(quantity))
         XCTAssertEqual(viewModel.title, "$12.49")
     }
+
+    func test_viewModel_starts_with_next_button_disabled() {
+        // Given
+        let currencySettings = CurrencySettings()
```
ctarda, line 348: I noticed that the three tests start with the same arrange section. 

I am curious here, are we trying to avoid extracting shared code to private methods in the tests, or do we consider that a bit of repetition in the individual tests actually makes them more readable, and therefore it's a good thing?

ctarda (approved):
I tested by:
- Tapping the Select All button
- Selecting each item in a 2 item order separately
- All the combinations of selecting one item, then tapping Select All, then activating and deactivating single items that I could think of

And it works perfectly 🥳 

:shipit:

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
Ecarrion: I like to have unit tests as isolated as possible so they make sense and exists on their own. I guess this is something we have not yet discussed but I'm happy to hear alternatives to improve them!

PR #3085, opened 2020-10-30
Order Details > variation details: Yosemite/Networking layer changes

jaclync:
Yosemite/Networking layer changes for #2280

Based on https://github.com/woocommerce/woocommerce-ios/pull/3078

## Changes

Networking layer:
- Created `OrderItemAttribute` struct
- Updated `OrderItem` to add `attributes: [OrderItemAttribute]` property, and decoded `name` and `attributes` differently based on whether the order item is a variation or not. Also, it filters the known attribute generated by core (`_reduced_stock`).
- Added 2 JSONs for testing in `OrderMapperTests`

Yosemite layer:
- Updated `Storage.OrderItem`'s `ReadOnlyConvertible` to include `attributes` from `NSOrderedSet` to array
- Added `Storage.OrderItemAttribute`'s `ReadOnlyConvertible` implementation
- Updated `OrdersUpsertUseCase` to handle order item attributes by replacing the `NSOrderedSet`, with unit tests in `OrdersUpsertUseCaseTests`

## Testing

The user-facing changes will be in the next PR, just CI for now!

Update release notes:


Networking/Networking/Model/OrderItemAttribute.swift
```diff
@@ -0,0 +1,33 @@
+public struct OrderItemAttribute: Decodable, Hashable, Equatable {
+    public let metaID: Int64
+    public let name: String
+    public let value: String
```
pmusolino, line 6: nit: Is it maybe better to rename it as `displayValue` since there is another field called `value` that maybe in the future we want to parse?

Networking/Networking/Model/OrderItemAttribute.swift
```diff
@@ -0,0 +1,33 @@
+public struct OrderItemAttribute: Decodable, Hashable, Equatable {
```
pmusolino, line 3: nit: Maybe a comment on the top of the struct could help identify what type of entity is this. 👍

Networking/Networking/Model/OrderItem.swift
```diff
@@ -52,17 +55,36 @@ public struct OrderItem: Decodable, Hashable {
...
         let variationID = try container.decode(Int64.self, forKey: .variationID)
 
+        let isVariation = variationID > 0
+        let name: String
+        let attributes: [OrderItemAttribute]
+        if isVariation {
+            name = try ((container.decodeIfPresent(String.self, forKey: .variationParentName))
+                        ?? container.decode(String.self, forKey: .name)).strippedHTML
+            let allAttributes = (try? container.decodeIfPresent([OrderItemAttribute].self, forKey: .attributes)) ?? []
+            attributes = allAttributes.filter { $0.name != "_reduced_stock" }
```
pmusolino, line 77: nit: consider to add a comment for remembering why we are not considering `_reduced_stock`.

Networking/Networking/Model/OrderItem.swift
```diff
@@ -52,17 +55,36 @@ public struct OrderItem: Decodable, Hashable {
...
+
+            // Logs error for unexpected attributes.
+            let unexpectedAttributes = attributes.filter { $0.name.hasPrefix("_") }
+            if unexpectedAttributes.isEmpty == false {
+                DDLogError("⚠️ Unexpected order item attributes: \(unexpectedAttributes)")
+            }
```
pmusolino, line 83: nit: I think that we can remove the check on `_reduced_stock` and filter by removing just the attributes with an underscore `_`prefix, right?

Networking/NetworkingTests/Responses/order-with-line-item-attributes-old-version.json
```diff
@@ -0,0 +1,93 @@
+{
```
pmusolino: nit: add a field `note` at the same level of `data`, that explain why this is an old version

pmusolino (approved):
Looks good to me 😸 I just left minor nits.

Networking/Networking/Model/OrderItemAttribute.swift
jaclync: I'd prefer adding just the fields we need now and add new ones if we need them in the future 🙂 also, there's only `name` and `value` and in the storage layer so it'd be nice keeping them consistent

Networking/Networking/Model/OrderItem.swift
jaclync: the reason I only filtered `_reduced_stock` is that this is the only attribute added by core that we know of. It's not impossible some users might actually name their attributes with an underscore prefix, so I didn't want to assume for the cases we don't know. It's probably rare though

Networking/Networking/Model/OrderItem.swift
jaclync: yup! added in https://github.com/woocommerce/woocommerce-ios/pull/3085/commits/98fa53c122601a936293eca646fca77fd2da2790

Networking/Networking/Model/OrderItemAttribute.swift
jaclync: added here: https://github.com/woocommerce/woocommerce-ios/pull/3085/commits/98fa53c122601a936293eca646fca77fd2da2790

Networking/NetworkingTests/Responses/order-with-line-item-attributes-old-version.json
jaclync: Hmm actually I cannot add a comment to JSON ([not supported](https://stackoverflow.com/a/244858/9185596)), so I renamed the file to `order-with-line-item-attributes-before-API-support.json` and test case to `test_OrderLineItem_attributes_are_empty_before_API_support` with a more clear comment. I thought a comment in the code would be more visible :)

jaclync (commented):
Thanks for the suggestions @pmusolino! I updated the PR and responded to your comments 😃

PR #3093, opened 2020-11-02
merge Release 5.4 back into develop

AliSoftware:
Updated `Podfile` after the release of `WPAuthenticator` pod v1.28 [EDIT: and WPKit 4.20]

Update release notes:


Podfile.lock
```diff
@@ -66,7 +66,7 @@ PODS:
     - WordPressKit (~> 4.18-beta)
     - WordPressShared (~> 1.12-beta)
     - WordPressUI (~> 1.7.0)
-  - WordPressKit (4.19.0):
+  - WordPressKit (4.20.0-beta.1):
```
AliSoftware: Is this normal?
This pod wasn't listed in the list of beta pods when I ran the `code_freeze` lane so not sure why it appears as beta now. (I see WordPressShared is in 1.12-beta above too 🤔 )

Should I end up releasing WPKit as a 4.20.0 too and re-update this here? or is this ok because those seem to be transitive dependencies from WPAuthenticator itself and not direct deps after all?

AliSoftware: Ok I just say 4.20.0 being released to trunk like 5mn ago, I'm guessing by @jkmassel when you were doing WPiOS 16.0? Meaning I was supposed to bump it too (in that case not sure why our `code_freeze` lane didn't list it for me to update earlier today?)

jkmassel (approved):

Issue #2407, opened 2020-06-09
Inventory barcode scanner


jaclync:
Refs:
p91TBi-2Si-p2 (design)
p77Llu-dwV-p2 (first hack week post)

Subtasks:

- [ ] create a feature flag and add a CTA
- [ ] TBD

toupper:
Replaced by https://github.com/woocommerce/woocommerce-ios/issues/11289

PR #3097, opened 2020-11-04
Product inventory/SKU settings: barcode scanner for SKU

jaclync:
Part of #2407 

Sorry about the PR size going slightly over 500 diffs! I originally thought I'd PR the barcode scanner separately, but then it's hard to test the integration. About 100 diffs are in two xibs. When I started the barcode scanner back in June, I referred to some example projects in [Vision documentation](https://developer.apple.com/documentation/vision).

## Changes

- Created `BarcodeScannerViewController` that is meant to be embedded as a child view controller, for future reuse in inventory scanner (more details in https://github.com/woocommerce/woocommerce-ios/issues/2407). This starts up camera for live video stream, and makes Vision requests for barcode detection from video output frames at a throttled rate.
  - Added a `CGRect` helper to map the scan area (see screenshots below) to the video frame (which is often bigger than the device screen). This mapping took me quite some time, I tried adding comments but let me know if I can explain the calculation further. I tested this by adding a debug image view that shows the cropped image and made sure that it is the same as the scan area.
- Created `ProductSKUInputScannerViewController` that is just a container of `BarcodeScannerViewController` with a navigation bar title.
- Integration with product inventory settings:
  - In `ProductInventorySettingsActionHandler`, added `handleSKUFromBarcodeScanner` similar to the existing `handleSKUChange` to display SKU before validation (since the SKU comes from the barcode scanner and we need to update the UI manually).
  - In `ProductInventorySettingsViewController`, navigated to `ProductSKUInputScannerViewController` when the user taps on the scan CTA.

## Testing

Camera is only available on a device, please try on a device - iPad would be super helpful since I mostly tested on iPhone!

- Go to the products tab
- Tap on a product whose SKU is editable
- Tap on the inventory/SKU row
- Tap on the scan CTA in the accessory view of the SKU row --> should be prompted to use camera if you haven't used camera for other use cases in the app, tap yes in this case. No camera permission scenario will be handled separately as part of #2407.
- Scan any barcode that is available to you, make sure it's only scanned when the barcode is in the scan area --> after a barcode is detected, the barcode should be populated on the SKU row

## Example screencast/screenshots

Screencast: https://d.pr/free/v/C0DvGY

Light mode | Dark mode
-- | --
IMG_7722 | IMG_7723

Landscape:

Update release notes:


Garance91540 (approved):
👏

WooCommerce/Classes/Extensions/CGRect+Scaling.swift
```diff
@@ -0,0 +1,38 @@
...
+    /// Returns a CGRect that maps the current frame to a target frame that has been scaled to aspect fill in the reference frame.
+    /// This assumes the reference and target frame have zero origin, and the target frame is resized to fill the shorter dimension of reference frame and
+    /// centered on the other dimension (the behavior of `resizeAspectFill`).
+    ///
+    /// - Parameters:
+    ///   - referenceFrame: the reference frame for self.
+    ///   - targetFrame: the frame that has been scaled to aspect fill in the reference frame.
+    func scaling(in referenceFrame: CGRect, to targetFrame: CGRect) -> CGRect {
+        // Aspect fill indicates that it is scaled by the maximum factor of the two dimensions.
+        let scaleFactor = min(targetFrame.width/referenceFrame.width, targetFrame.height/referenceFrame.height)
```
Ecarrion: Maximum or minimum?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventorySettingsViewController.swift
```diff
@@ -335,7 +335,17 @@ private extension ProductInventorySettingsViewController {
 //
 private extension ProductInventorySettingsViewController {
     @objc func scanSKUButtonTapped() {
-        // TODO-2704: scan barcode for product SKU
+        let scannerViewController = ProductSKUInputScannerViewController(onBarcodeScanned: { [weak self] barcode in
+            self?.onSKUBarcodeScanned(barcode: barcode)
+            self?.navigationController?.popViewController(animated: true)
+        })
+        navigationController?.pushViewController(scannerViewController, animated: true)
```
Ecarrion: nit 
```suggestion
        show(scannerViewController, sender: self)
```

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
```diff
@@ -0,0 +1,247 @@
...
+                let scaledScanAreaRect = scanAreaRect.scaling(in: self.videoOutputImageView.frame, to: imageExtent)
+
+                // Crops scan area from the original video output image.
+                let croppedCIImage = ciImage.cropped(to: scaledScanAreaRect)
+
+                // Configures options for `VNImageRequestHandler`.
+                var requestOptions: [VNImageOption: Any] = [:]
+                if let cameraData = CMGetAttachment(sampleBuffer, key: kCMSampleBufferAttachmentKey_CameraIntrinsicMatrix, attachmentModeOut: nil) {
+                    requestOptions = [.cameraIntrinsics: cameraData]
+                }
```
Ecarrion: nit: I think this could be extracted to a method, it will help a lot with readability :-)

WooCommerce/Classes/Extensions/CGRect+Scaling.swift
```diff
@@ -0,0 +1,38 @@
...
+
+extension CGRect {
+    /// Returns a CGRect that maps the current frame to a target frame that has been scaled to aspect fill in the reference frame.
+    /// This assumes the reference and target frame have zero origin, and the target frame is resized to fill the shorter dimension of reference frame and
+    /// centered on the other dimension (the behavior of `resizeAspectFill`).
+    ///
+    /// - Parameters:
+    ///   - referenceFrame: the reference frame for self.
+    ///   - targetFrame: the frame that has been scaled to aspect fill in the reference frame.
+    func scaling(in referenceFrame: CGRect, to targetFrame: CGRect) -> CGRect {
```
Ecarrion: QQ: is this function meant to be used by other components or only by the barcode scanner?
If it's only by the barcode scanner I think it may be best to have a helper class like `BarcodeFrameScaler` than to have a general `CGRect` extension.

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
```diff
@@ -0,0 +1,247 @@
...
+    func configureDimmingView(_ view: UIView) {
+        view.backgroundColor = Constants.dimmingColor
+    }
+
+    func configureInstructionLabel() {
+        instructionLabel.backgroundColor = Constants.dimmingColor
+        instructionLabel.textAlignment = .center
+        instructionLabel.applyHeadlineStyle()
+        instructionLabel.textColor = .white
+        instructionLabel.textInsets = .init(top: 11, left: 0, bottom: 11, right: 0)
```
Ecarrion: Nit: Should we move this to constants?

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
```diff
@@ -0,0 +1,247 @@
+import AVFoundation
+import UIKit
+import Vision
+
+/// Starts live stream video for scanning barcodes.
+/// This view controller is meant to be embedded as a child view controller for navigation customization.
+final class BarcodeScannerViewController: UIViewController {
+    /// UI properties that could differ in each use case.
+    struct ViewProperties {
+        let instructionText: String
```
Ecarrion: QQ: Do you plan to have more view properties in the future?

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/ProductSKUInputScannerViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+                return
+            }
+            self.hasDetectedBarcode = true
+            self.onBarcodeScanned(barcode)
+        }
+    }()
+
+    private let onBarcodeScanned: (String) -> Void
+
+    /// Makes sure the barcode detection callback is only handled once.
```
Ecarrion: Did you meant to remove this comment?

Ecarrion (approved):
Wow, this is great! Such a nice addition! 🚀 

I left you some minor comments but looks very good in general.

PS: I noticed that the keyboard comes up when you press the barcode scanner button, is that intentional?
PS2: Sorry for taking so long to review 😇 
 Scanning | Result
---- | ----
IMG_8913 | IMG_8914

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/ProductSKUInputScannerViewController.swift
jaclync: ooh I meant for this comment to indicate that this line is to make sure the callback is handled once, but it does sound more like a TODO comment 😅 I'll rephrase this comment

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
jaclync: Not at the moment, maybe in the future. do you think it's better to just pass this one property to the view controller instead of a struct now that there's only one?

WooCommerce/Classes/Extensions/CGRect+Scaling.swift
jaclync: yea this scaling logic is specific to barcode scanner, updated to a separate struct in: https://github.com/woocommerce/woocommerce-ios/pull/3097/commits/27ea4d0fa81affcd61f5e21a8bd60265075618ea

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
jaclync: yup! updated in https://github.com/woocommerce/woocommerce-ios/pull/3097/commits/b33b3434f9edbfc2341d700065cff6e740cb9aa4

WooCommerce/Classes/ViewRelated/Products/Edit Product/Inventory Settings/ProductInventorySettingsViewController.swift
jaclync: updated in https://github.com/woocommerce/woocommerce-ios/pull/3097/commits/b33b3434f9edbfc2341d700065cff6e740cb9aa4

WooCommerce/Classes/Extensions/CGRect+Scaling.swift
jaclync: great catch! updated in https://github.com/woocommerce/woocommerce-ios/pull/3097/commits/b33b3434f9edbfc2341d700065cff6e740cb9aa4

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
jaclync: yup makes sense! extracted to a separate function in https://github.com/woocommerce/woocommerce-ios/pull/3097/commits/ef648ba5c3ceabe45443710e62d0147c59ad5c14

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/ProductSKUInputScannerViewController.swift
jaclync: Updated the comment in https://github.com/woocommerce/woocommerce-ios/pull/3097/commits/b33b3434f9edbfc2341d700065cff6e740cb9aa4

jaclync:
Thanks for testing & reviewing @Ecarrion! I responded to your comments and updated the PR for another look 😃 

> PS: I noticed that the keyboard comes up when you press the barcode scanner button, is that intentional?

I couldn't seem to reproduce this keyboard issue 🤔 I tried tapping at different areas around the scan CTA and it triggered either the keyboard for editing the SKU or showing the scanner for me. Did you see the keyboard come up while the scanner is shown at the same time?

> PS2: Sorry for taking so long to review 😇

No worries at all! Last week was HACK Week and hacking was the focus 💻

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
Ecarrion: Not a big deal, but if it's only one property I would just pass one property. Up to you.

Ecarrion:
> I couldn't seem to reproduce this keyboard issue 🤔 I tried tapping at different areas around the scan CTA and it triggered either the keyboard for editing the SKU or showing the scanner for me. Did you see the keyboard come up while the scanner is shown at the same time?

I noticed the issue when opening the scanner for the first time, the app asked for permissions, and while I was tapping the "accept" button I saw the keyboard go up.

Also, if you go to the scanner and then tap back, you will the the keyboard is present.

Ecarrion (approved):
🚀

WooCommerce/Classes/ViewRelated/Products/SKU Scanner/BarcodeScannerViewController.swift
jaclync: makes sense! updated in https://github.com/woocommerce/woocommerce-ios/pull/3097/commits/a08089b4de36dba1ac05347d1ef67f9374322682

jaclync:
> I noticed the issue when opening the scanner for the first time, the app asked for permissions, and while I was tapping the "accept" button I saw the keyboard go up.
Also, if you go to the scanner and then tap back, you will the the keyboard is present.

Interesting, I wasn't able to reproduce the keyboard issues on my test device iPhone 6s Plus iOS 14.1 (either by reinstalling the app to accept the camera permission or going back from the scanner, if I never tapped on the SKU field to bring up the keyboard in the first place) 🤔 I added your comment to "Other issues" in https://github.com/woocommerce/woocommerce-ios/issues/2407 to look out for in the future 🙂 

Merging now!

PR #3101, opened 2020-11-05
Extract all Build Settings into xcconfig files

mokagio:
_WIP_


mokagio:
Closing this because it's become stale. We'd still like to get there at some point. When the time will come, it'll be easier to start from scratch.

Issue #1810, opened 2020-02-03
Review push notification doesn't open product review details when tapped


astralbodies:
**Describe the bug**
When I received a push notification for a new product review, tapping it did not bring up the details of the review. I had product reviews up as the last screen I was using in the app before backgrounding it. The product review didn't appear in the list either until I pulled to refresh.

**To Reproduce**
1. Go to the Reviews tab.
2. Background the app.
3. Leave a product review to receive a push notification.
4. Tap the push notification to open the app.

**Screenshots**
Didn't take a screenshot as the text really covers it all.

**Expected behavior**
Tapping the push notification should force the review to be loaded and the details show up.

**Mobile Environment**
Please include:
- Device: iPhone 6s
- iOS version: 13.1.1
- WooCommerce iOS version: 3.5.0.0

astralbodies:
Also the product name wasn't listed - only "Unknown".

jaclync:
I noticed this issue in WCiOS 3.8.0.0 sometimes, this seemed to happen while the app has navigated to an Order or Review from a push notification before

jaclync:
Re the comment above: I tried reproducing the issue by leaving the app in the background when it received the new order push notif, then receiving a new review push notif. Wasn't able to repro it this time though, it's probably caused by other reasons

shiki:
I shared a similar scenario about order notifications in https://github.com/woocommerce/woocommerce-ios/issues/1867#issuecomment-639660324. I can reproduce the same for reviews. 

- It consistently opens the last screen only if the device was locked. 
- If the device was not locked, then the app opens the review details correctly. 

My device was also in low power mode.

rachelmcr:
I saw this issue repeatedly while testing 5.4.0.0, exactly as described in the original issue (app was backgrounded but device wasn't locked, had to pull refresh the review list to see the new review, and product name wasn't listed).

I backgrounded the app once on an order detail screen and twice on the My store tab, and each time I got the same results.

shiki:
We changed the way Product Review push notifications are handled in https://github.com/woocommerce/woocommerce-ios/pull/2669 so that the product will always be loaded. However, it looks like there was a regression in https://github.com/woocommerce/woocommerce-ios/pull/2879. If the product was not previously downloaded on the device, then the Review details page is never shown. 

In summary: There's a regression. 

---

New Order push notifications seem to be working though.

shiki:
[...]

PR #3103, opened 2020-11-05
Fix Product Review Push Notification Handling Regressions

shiki:
Refs #1810. 

This fixes two regressions for handling Product Review push notifications. 

### To Reproduce

1. Delete the app from the device to reset all the data.
2. Run and log in. Do not open the Products and Reviews tab. 
3. Place the app in the background. 
4. Send a push notification. 
5. Tap on the push notification. 

The review details page will not be shown. 

### Why It Happens

Currently, we load all the information for presenting a Review Details page via this action handler:

https://github.com/woocommerce/woocommerce-ios/blob/ec6f85b6eeb6d06fb03fe9670a0e5343d3a1b67e/Yosemite/Yosemite/Stores/ProductReviewStore.swift#L125-L130

This relied on the fact that API completion blocks are always called regardless of the `Remote` instances being deallocated. However, we changed that behavior in #2879. The `Remote` completion blocks are no longer called if the `Remote` instance has been deallocated. We did that by adding `weak self` in 1b98298:

https://github.com/woocommerce/woocommerce-ios/blob/ec6f85b6eeb6d06fb03fe9670a0e5343d3a1b67e/Networking/Networking/Remote/Remote.swift#L30-L34

This caused a problem in the `RetrieveProductReviewFromNoteUseCase` shown above. Because the `UseCase` is not retained, it gets deallocated right away, and its network calls are essentially _cancelled_. The result is that tapping the push notification will still cause the API calls to be run but [the final completion block](https://github.com/woocommerce/woocommerce-ios/blob/ec6f85b6eeb6d06fb03fe9670a0e5343d3a1b67e/Yosemite/Yosemite/Stores/ProductReview/RetrieveProductReviewFromNoteUseCase.swift#L188-L195) will not be called, and so the review details page is never shown by [ReviewsCoordinator](https://github.com/woocommerce/woocommerce-ios/blob/ec6f85b6eeb6d06fb03fe9670a0e5343d3a1b67e/WooCommerce/Classes/ViewRelated/Reviews/ReviewsCoordinator.swift#L64-L94).

### Solution

I fixed this by storing the `UseCase` as an ivar (https://github.com/woocommerce/woocommerce-ios/commit/7896fcb509ad8f76443f7df7f5c787c8a813cde8). I tried to create a unit test to reproduce the scenario but failed. I could not find a way to make the `UseCase` be forcefully deallocated immediately during a unit test scenario. 

### To Reproduce

1. Load all the products of the site.
1. Terminate the app.
2. Send a Product Review push notification. 
3. Tap on the push notification. 
4. The Review Detail page will be shown. But then it will be dismissed right away and the user will be navigated to the Reviews page.

### Why It Happens

I found this after fixing the first regression. This happens because in https://github.com/woocommerce/woocommerce-ios/pull/2796, we started deleting all the reviews when loading the Reviews list page for the first time. 

https://github.com/woocommerce/woocommerce-ios/blob/ec6f85b6eeb6d06fb03fe9670a0e5343d3a1b67e/Yosemite/Yosemite/Stores/ProductReviewStore.swift#L88-L90

That seemed fine but is causing an issue if the `ReviewDetailsViewController` is presented. The `ReviewDetailsViewController` has special handling for when the underlying `ProductReview` (`NSManagedObject`) has been deleted from the `NSManagedObjectContext`. It dismisses itself when that happens. 

https://github.com/woocommerce/woocommerce-ios/blob/ec6f85b6eeb6d06fb03fe9670a0e5343d3a1b67e/WooCommerce/Classes/ViewRelated/Reviews/ReviewDetailsViewController.swift#L118-L121

### Solution

I _fixed_ this by removing the `onDelete` handling (e59a526). As far as I can tell, the `onDelete` handling does not seem to be necessary. The `ReviewDetailsViewController` and its buttons like _Spam_ and _Delete_ still seems to work well even if the underlying `NSManagedObject` no longer exists. 

It also shows a misleading notice to the user because the Review could not have been deleted from the site, it was just deleted in the local database. We have a similar problem with Orders https://github.com/woocommerce/woocommerce-ios/issues/2369. 

## Testing

Please follow the **To Reproduce** steps in the sections above. 🙂 

## Submitter Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Reviews/ReviewDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Reviews/ReviewDetailsViewController.swift
@@ -114,11 +114,6 @@ private extension ReviewDetailsViewController {
         entityListener.onUpsert = { [weak self] productReview in
             self?.productReview = productReview
         }
-
-        entityListener.onDelete = { [weak self] in
-            self?.navigationController?.popToRootViewController(animated: true)
-            self?.displayNoteDeletedNotice()
-        }
     }
 
     /// Reports a significant event to the App Rating Manager
@@ -175,15 +170,6 @@ private extension ReviewDetailsViewController {
 // MARK: - Display Notices
 //
 private extension ReviewDetailsViewController {
-
-    /// Displays a Notice onScreen, indicating that the current Review has been deleted from the Store.
-    ///
-    func displayNoteDeletedNotice() {
-        let title = NSLocalizedString("The review has been removed", comment: "Displayed whenever a review that was onscreen got deleted.")
-        let notice = Notice(title: title, feedbackType: .error)
-
-        ServiceLocator.noticePresenter.enqueue(notice: notice)
-    }
 
     /// Displays the Error Notice.
     ///
--- a/Yosemite/Yosemite/Stores/ProductReviewStore.swift
+++ b/Yosemite/Yosemite/Stores/ProductReviewStore.swift
@@ -8,6 +8,9 @@ import Storage
 public final class ProductReviewStore: Store {
     private let remote: ProductReviewsRemote
 
+    private lazy var productReviewFromNoteUseCase =
+        RetrieveProductReviewFromNoteUseCase(network: network, derivedStorage: sharedDerivedStorage)
+
     private lazy var sharedDerivedStorage: StorageType = {
         return storageManager.newDerivedStorage()
     }()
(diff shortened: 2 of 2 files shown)


Ecarrion (approved):
Thank you for fixing these 🙇!

shiki:
Thanks for reviewing so quickly, Ernesto!

AliSoftware:
New beta 5.4.0.1 has been deployed including those changes 🎉

PR #3114, opened 2020-11-10
Removed error logs on Sentry from ProductFormViewController

pmusolino:
It seems that [we are logging on Sentry](https://sentry.io/share/issue/dc6647450c4d45b896c03b6dd7fb4e88/) some errors when a product is saved or deleted, but they are not useful for us since we are logging just the error messages, which are generic and localized in the store language.

In this PR I removed the `CrashLogging` from `ProductFormViewController`. We discussed it [here](https://a8c.slack.com/archives/C70PAM3RA/p1605022224162900).

I added just you @jaclync as a reviewer since you probably know if there are other reasons to log these messages, that maybe we are ignoring. 

## Testing
- Just CI

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
@@ -3,8 +3,6 @@ import UIKit
 import WordPressUI
 import Yosemite
 
-import class AutomatticTracks.CrashLogging
-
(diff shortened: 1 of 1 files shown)


pmusolino:
Also related to this: https://sentry.io/share/issue/ba9aac138dc04f039cc79f69e6b80889/

jaclync (approved):
LGTM! thanks for removing the Sentry error logs that aren't as useful as Tracks events 😄

PR #3136, opened 2020-11-11
Pass boolean flag to enable unified login and signup

### LINKED ISSUE #3124: [UL&S] Activate WPAuthenticator flag for Unified Login
Pass `enableUnifiedAuth` and `enableUnifiedCarousel` to `WordPressAuthenticatorConfiguration` as true.

This will make WPAuthenticator display the UL&S flow.

This issue is not to be merged into `develop`, but into a long lived feature branch.

There is some indetermination here about the actual scope of the task. I think an acceptable scope here is that the app builds, and it is possible to login, even if the flow does not look like it should

ctarda:
Closes #3124 

⚠️  Notice how the PR is not against develop, but against the long-lived feature branch were we will implement Unified Login and signup.

This PR does not fully implement the new Unified Login flow, it only activates the configuration flag necessary to start working on customising the flow and the UI. The actual implementation, the customisation of the UI and the navigation flows begins after this PR.

# Changes
* Pass a boolean flag to WPAuthenticator (via WordPressAuthenticatorConfiguration), so that the authentication library starts displaying the unified login flow.

# Testing
* Check the branch, build and run. 
* It should be possible to log in with wordpress.com and with a site address. 
* Please avoid using SIWA for now. 

# Example screenshots
| | Before | After |
| ---- | ---- | ---- |
| Prologue |  |  |
| Login with WP.com |  |  |
| Login with Site Address |  |  |

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/AuthenticationManager.swift
+++ b/WooCommerce/Classes/Authentication/AuthenticationManager.swift
@@ -42,7 +42,8 @@ class AuthenticationManager: Authentication {
                                                                 userAgent: UserAgent.defaultUserAgent,
                                                                 showLoginOptions: true,
                                                                 enableSignUp: false,
-                                                                enableSignInWithApple: isSignInWithAppleEnabled)
+                                                                enableSignInWithApple: isSignInWithAppleEnabled,
+                                                                enableUnifiedAuth: true)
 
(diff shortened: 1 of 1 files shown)


pmusolino (approved):
LGTM :shipit:

ctarda:
Thanks for the review @pmusolino 🙇

Issue #3110, opened 2020-11-09
[Issue Refund] Download Payment Gateways


Ecarrion:
In order to support automatic refunds, we need to download all payment gateways to see which one of them supports automatic refunds.

Tasks:
- [x] Update Storage layer with payment entities.
- [x] Update Networking layer with payment entities.
- [x] Create Yosemite action to sync Payment gateways
- [x] Update Woo target to sync Payment gateways on startup

PR #3140, opened 2020-11-11
[Issue Refunds] Add GatewayPayment storage structure

### LINKED ISSUE #3019: [Xcode 12] Fix Core Data insecure value transformer warnings
This is shown when compiling with Xcode 12. It seems to happen in Model 32. We did something similar for older model versions in https://github.com/woocommerce/woocommerce-ios/pull/2475. If fixing this warning in Xcode will cause more warnings to show up in older versions, @astralbodies suggested that we should look into _safely_ deleting the older **unused** model versions.

Ecarrion:
part of #3110 
fix https://github.com/woocommerce/woocommerce-ios/issues/3019

# Why

This PR does two things(sorry)

1. As a continuation of adding `PaymentGateway` networking support this PR takes care of adding the necessary `Storage` classes so we can use it(`PaymentGateway`) in our [future Yosemite actions and store](https://github.com/woocommerce/woocommerce-ios/pull/3141)

2. While adding a `Transformable` property on `Storage.PaymentGateway` I noticed that it generated a new warning because now, by default, those types of properties use an insecure NSCoding value transformer. This PR fixes that warning for all other transformable properties in the project.

# How

1. 
- Added Model 36 that includes `PaymentGateway` Core Data Entity
- Added auto-gen `+CoreDataProperties` and `+CoreDataClass` extensions
- Updated `StorageType+Extensions` to provide functions to load payment gateways from storage.
- Added `ReadOnlyConvertible` methods to `PaymentGateway`

2.
- Update `transformable` properties on Product, to use `NSSecureUnarchiveFromData` as the property value transformer instead of the default(unsecure) one.

# Testing Steps

1. Not much to test here, but you can see it working on [this PR](https://github.com/woocommerce/woocommerce-ios/pull/3141)

2. Make sure that after compiling(cleaning it first) the project, no more core data warnings are present

Update release notes:


Storage/Storage/Model/MIGRATIONS.md
```diff
@@ -2,6 +2,11 @@
 
 This file documents changes in the WCiOS Storage data model. Please explain any changes to the data model as well as any custom migrations.
 
+## Model 36 (Release 5.5.0.0)
+- @ecarrion 2020-11-10
+- Added `PaymentGateway` entity.
+- Fixed warning from `transformable` properties byt setting an `NSSecureCoding` value transformer.
```
jaclync: nit:

```suggestion
- Fixed warning from `transformable` properties by setting an `NSSecureCoding` value transformer.
```

jaclync (approved):
Thanks for fixing the Core Data warnings! I don't see the warnings anymore 🎉 

Screen Shot 2020-11-12 at 4 18 09 PM

Do you think it's worth adding a test case for migrating to the new model, based on [this doc](https://github.com/woocommerce/woocommerce-ios/blob/e98c5099860fc740ee8871470831a4a2b4feb6ce/docs/creating-core-data-model-versions.md#always-add-unit-tests)? It's probably not very useful for a new entity without relationships to other existing entities (like `PaymentGateway`). Would it be worth making sure `Product`'s attributes with the value transformer change have the same value after the migration?

Ecarrion:
@jaclync 
> Do you think it's worth adding a test case for migrating to the new model, based on this doc? 

Totaly worth it, thanks for suggesting that! I've added it in https://github.com/woocommerce/woocommerce-ios/pull/3140/commits/fc04369f0222e6683b7da3f5583627697b0de0e9

PR #3149, opened 2020-11-12
[Issue Refund] Decode and store Order.paymentMethod

### LINKED ISSUE #3111: [Issue Refund] Fetch payment gateway from order.
In order to display info needed in #2982 we need to create a `Yosemite` action to fetch a payment gateway from the payment ID.

We also need update the order entity to store the payment information code

Ecarrion:
closes #3111

# Why
As a continuation of the "Refund ~ PaymentGateway" work, this PR takes care of decoding and storing the payment gateway id when synchronizing orders.

This `ID` will be used later to fetch the full payment gateway to know if it supports automatic money refunds or not. 😇 

# How
- Added `paymentMethodID` to `Networking.Order`
- Added `paymentMethodID` to `Storage.Order` and updated CoreData model to `v37`

# Testing Steps
- Smoke test around orders and check that everything is working as before.

Update release notes:


Networking/Networking/Model/Order.swift
```diff
@@ -189,6 +194,7 @@ private extension Order {
         case shippingTax        = "shipping_tax"
         case total              = "total"
         case totalTax           = "total_tax"
+        case paymentMethodID    = "payment_method"
```
shiki, line 197: At first, I was confused about why we're diverging from the REST API name but after looking [at the docs](https://woocommerce.github.io/woocommerce-rest-api-docs/#order-properties), I believe using `ID` makes more sense!

Storage/Storage/Model/MIGRATIONS.md
```diff
@@ -2,6 +2,10 @@
 
 This file documents changes in the WCiOS Storage data model. Please explain any changes to the data model as well as any custom migrations.
 
+## Model 37 (Release 5.5.0.0)
```
shiki, line 5: ⚠️  We should add a simple test in `MigrationTests`. 😇

Networking/Networking/Model/Order.swift
```diff
@@ -28,6 +28,7 @@ public struct Order: Decodable, GeneratedCopiable {
     public let shippingTax: String
     public let total: String
     public let totalTax: String
+    public let paymentMethodID: String
```
shiki, line 31: As far as I can tell, it looks like `payment_method` is not optional. 👍  

It's just scary that it's saved in an optional meta database in WordPress. 😬

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 37.xcdatamodel/contents
```diff
@@ -0,0 +1,546 @@
...
+        <attribute name="dateCreated" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="dateModified" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="datePaid" optional="YES" attributeType="Date" usesScalarValueType="NO"/>
+        <attribute name="discountTax" optional="YES" attributeType="String"/>
+        <attribute name="discountTotal" optional="YES" attributeType="String"/>
+        <attribute name="exclusiveForSearch" attributeType="Boolean" defaultValueString="NO" usesScalarValueType="YES"/>
+        <attribute name="number" optional="YES" attributeType="String"/>
+        <attribute name="orderID" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
+        <attribute name="parentID" optional="YES" attributeType="Integer 64" defaultValueString="0" usesScalarValueType="YES"/>
+        <attribute name="paymentMethodID" optional="YES" attributeType="String"/>
```
shiki, line 62: I wonder if we should also consider this to not be optional but have a default empty string value instead. 🤔

Storage/Storage.xcodeproj/project.pbxproj
```diff
@@ -1467,7 +1469,7 @@
 				746A9D14214071F90013F6FF /* Model 2.xcdatamodel */,
 				B59E11D920A9D00C004121A4 /* Model.xcdatamodel */,
 			);
-			currentVersion = 261CF1C1255AFDC40090D8D3 /* Model 36.xcdatamodel */;
+			currentVersion = 261CF32E255D988F0090D8D3 /* Model 37.xcdatamodel */;
```
shiki, line 1472: I wonder why this is included in the `project.pbxproj`. 😑

Storage/Storage/Model/WooCommerce.xcdatamodeld/Model 37.xcdatamodel/contents
```diff
@@ -0,0 +1,546 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
```
shiki, line 1: Compared with Model 36. MIGRATIONS.md advertisement seems accurate. 👍

WooCommerce/WooCommerceUITests/Mocks/mappings/jetpack-blogs/wc/orders/orders.json
```diff
@@ -54,6 +54,7 @@
                     "postcode": "94119",
                     "country": "US"
                 },
+                "payment_method": "",
```
shiki, line 57: Adding all these stops being fun after a while. 🤣

shiki (approved):
✔️ Tested. Looks like the payment method ID is correctly captured from the API.
⚠️ I only have one request, which is adding a simple test in `MigrationTests`. 🙂

Ecarrion:
Thanks for the quick review @shiki! I've added the unit test here https://github.com/woocommerce/woocommerce-ios/pull/3149/commits/a283d468fc73071371f36865e860c8b92e14ea41

shiki:
Nice migration test! 🎉

PR #3155, opened 2020-11-16
Print a shipping label: networking layer

jaclync:
Part of #2166 
Part of #3117

## Changes

- Added a new API version `WooAPIVersion.wcConnectV1` ("wc/v1/connect")
- Added a new model `ShippingLabelPrintData` and its mapper (this model is not persisted in storage and thus does not have a corresponding storage model)
- Created `ShippingLabelRemote` with an endpoint to print a shipping label (`printShippingLabel`)
- Added unit tests and a json file

## Testing

Just CI for now, the endpoint isn't used in the app yet!

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #2166, opened 2020-04-21
Shipping Labels M1 release: add backend support to print a shipping label


anitaa1990:
As part of the shipping labels project, we need to add an endpoint to print a shipping label.

- [x] Networking layer https://github.com/woocommerce/woocommerce-ios/pull/3155
- [x] Yosemite layer https://github.com/woocommerce/woocommerce-ios/pull/3156

PR #3156, opened 2020-11-16
Print a shipping label: Yosemite layer

jaclync:
Part of #2166 

Based on Networking layer changes: https://github.com/woocommerce/woocommerce-ios/pull/3155

## Changes

Networking layer changes:
- Created `ShippingLabelRemoteProtocol`, and updated `ShippingLabelRemote` to implement this protocol 
- `ShippingLabelPrintData`: implemented `Equatable` for comparing two instances in unit tests

Yosemite layer changes:
- Created `ShippingLabelAction` and `ShippingLabelStore` to handle the action
- Created `MockShippingLabelRemote: ShippingLabelRemoteProtocol` and added unit tests for success and failure cases

## Testing
Just CI for now, the action isn't used in the app yet!

Update release notes:


Yosemite/Yosemite/Stores/ShippingLabelStore.swift
```diff
@@ -0,0 +1,72 @@
...
+        switch self {
+        case .label:
+            return "label"
+        case .legal:
+            return "legal"
+        case .letter:
+            return "letter"
+        }
+    }
+}
```
pmusolino: What do you think about removing this extension, and add `String` to the `ShippingLabelPaperSize` enum?

Like this:
```
public enum ShippingLabelPaperSize: String {
     case label
     case legal
     case letter
 }
```

Yosemite/YosemiteTests/Mockups/Networking/Remote/MockShippingLabelRemote.swift
```diff
@@ -0,0 +1,43 @@
...
+/// Mock for `ShippingLabelRemote`.
+///
+final class MockShippingLabelRemote {
+    private struct PrintResultKey: Hashable {
+        let siteID: Int64
+        let shippingLabelID: Int64
+        let paperSize: String
+    }
+
+    /// The results to return based on the given arguments in `loadProduct`
```
pmusolino: nit: I think this is a typo -> `loadProduct`

pmusolino (approved):
LGTM :shipit: I just added two minor nits

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
jaclync: I replaced the enum here with the one from the networking layer in: https://github.com/woocommerce/woocommerce-ios/pull/3155/commits/7fd4a2596885cd00b700ab2465af51f05102663d

There I conformed the enum to `RawRepresentable` following other enums in the networking layer, and the string raw value is an internal detail.

Yosemite/YosemiteTests/Mockups/Networking/Remote/MockShippingLabelRemote.swift
jaclync: fixed in: https://github.com/woocommerce/woocommerce-ios/pull/3156/commits/88bc08df311eeb0c06fa17fc8f9e9c08599f2b9a

PR #3159, opened 2020-11-16
Support removing an Image from a Product Variation

### LINKED ISSUE #2576: Support removing a product image from a product variation after it's fixed in the API
Ref: p91TBi-34W-p2

As of writing, there is an issue on removing a product image from a product variation in the API. Let's re-enable this feature in the app after it's fixed in the API.

TBD: whether to support this with a WC version check or wait until the WC version with the fix has been widely adopted.

---

Corresponding Android issue: https://github.com/woocommerce/woocommerce-android/issues/3066

pmusolino:
Fixes #2576

## Description
Originally, we disabled the possibility for a user to remove an Image from a Product Variation, since [there was a bug in WooCommerce <=4.6](https://github.com/woocommerce/woocommerce/issues/27116).
Since this bug was fixed in WooCommerce 4.7 and greater, I enabled the possibility for a user to remove an image from a Product Variation, re-enabling it in the code. I also handled the case when the API returns the error `woocommerce_variation_invalid_image_id` which means we are sending `image = null` to an old version of WooCommerce. In that case, a custom error will appear.

## Testing
For testing purposes, I suggest installing on your testing website the plugin [WooCommerce Beta Tester](https://wordpress.org/plugins/woocommerce-beta-tester/) that allow you to switch easily between different versions of WooCommerce.

#### Case 1
1. Configure your website using WooCommerce <=4.6.
2. Run the app, and open a Product Variation with an image.
3. Open that image, and remove it, then update the product variation -> You should get an alert that suggests you to update your website to WooCommerce 4.7 or greater.

#### Case 2
1. Configure your website using WooCommerce 4.7 or greater.
2. Run the app, and open a Product Variation with an image.
3. Open that image, and remove it, then update the product variation -> the image should be removed correctly, with no errors. **Note**: if the parent product has an image, the product variation image will be set to the one used by the parent product. This is due to another bug (?), [discussed here](https://a8c.slack.com/archives/CGPNUU63E/p1605526543184500).

## Screenshots and GIF
| Before (under WooCommerce 4.6)            |  After (under WooCommerce 4.6) | After (under WooCommerce 4.7) |
WooCommerce  4 7 | Simulator Screen Shot - iPhone 11 Pro - 2020-11-16 at 15 57 06 | 

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/EditableProductVariationModel.swift
```diff
@@ -178,7 +178,7 @@ extension EditableProductVariationModel: ProductFormDataModel, TaxClassRequestab
     }
 
     func isImageDeletionEnabled() -> Bool {
```
ctarda, line 180: I'm just curious. Since this is always going to return true, would it make sense to declare it just as a let constant?

ctarda (approved):
Tested on both 4.6 and 4.7 

The error string on 4.6 is updated, and images can be deleted.

I understand this might be not part of the scope of this issue, but one thing I am a bit confused about (I don't know if it's an issue or not) is that if there is already one image, and then I tap the "add image" button (see screenshot) I can not add an extra image, only edit the existing one.

pmusolino:
> I understand this might be not part of the scope of this issue, but one thing I am a bit confused about (I don't know if it's an issue or not) is that if there is already one image, and then I tap the "add image" button (see screenshot) I can not add an extra image, only edit the existing one. 

This is correct because for every Product Variation is possible to add a maximum of one image. So, if you already have uploaded an image, when you tap the button for adding an image, the button on top will suggest you replace the existing image.

WooCommerce/Classes/ViewRelated/Products/Edit Product/EditableProductVariationModel.swift
pmusolino: Good question. One of the reasons is that in `EditableProductModel` and `EditableProductVariationModel` we are using the properties for returning the properties related to a product or a variation, and methods for managing the visibility logics. In reality, we can transform everything into a property in these classes, but maybe using properties and methods makes the roles more clear.

pmusolino:
@ctarda thank you for the review and testing, I'm going to merge this PR right now 🙇

PR #3160, opened 2020-11-16
release/5.5 -> develop

AliSoftware:
Code freeze for 5.5

NB: This PR already includes the Release Notes validated by Editorial [p2JRYi-57q-p2]


RELEASE-NOTES.txt
```diff
@@ -1,7 +1,11 @@
+*** PLEASE FOLLOW THIS FORMAT: [<priority indicator, more stars = higher priority>] <description> [<PR URL>]
```
jkmassel, line 1: Will this change break the "automatically add the version number to the top of the file" functionality?

jkmassel (approved):
LGTM – just one question around the release notes file change.

RELEASE-NOTES.txt
AliSoftware: Yes, it actually already does (as you can see from commit history 😓 )

This new note has been added like just 2 releases ago, and we haven't got around updating our scripts to take it into consideration just yet, so for now I update its position manually. Hopefully we'll get around updating the script to something like "ignore leading lines that start with `*** ` when adding the new version number" or something like that (so we can add even more instructions/notes here in the future if we need to)

AliSoftware:
Closing in favor of https://github.com/woocommerce/woocommerce-ios/pull/3162 to fix conflict resolution

PR #3163, opened 2020-11-17
Edit Products: show the Linked products cell in Product Detail settings and inside the bottom action sheet

### LINKED ISSUE #3071: Linked Products: show the Linked products cell (part 1)
Ref p91TBi-2XV-p2.

The first task to be implemented for Linked Products, will be to insert the Linked Products cell inside the "Add more details" action sheet.

Linked Products are shown for every type of product (simple, grouped products, etc...).

## Goals
- [ ]  Display the menu item in Add More Details if there are no linked products. 
- [ ]  Show the number of _linked_ upsell products and cross-sell products in product details.

## Design

| Add more details             |  Linked Products cell |
Schermata 2020-10-28 alle 15 10 14 | Schermata 2020-10-28 alle 15 05 02

pmusolino:
Fixes #3071 

## Description
As part of Milestone 5, in this PR I introduced the logic for showing/hiding the Linked Products cell inside the Product Detail screen or in the action sheet under the button "Add more details".

## Changes
- Added a new icon + test for Linked Products.
- Added `editLinkedProducts` case in `ProductFormBottomSheetAction` for managing title and description of the cell inside the bottom sheet.
- Added the new linked products row in `DefaultProductFormTableViewModel`.
- Added the properties `upsellIDs` and `crossSellIDs` used for Linked Products inside `EditableProductModel`, `EditableProductVariationModel` and `ProductFormDataModel`.
- Managed the appearance of the Linked Products cell in `ProductFormActionsFactory`.
- Added linked products empty (for the moment) actions in `ProductFormViewController`.
- Updated the old tests, and added two new test cases about the Linked Products appearance in `ProductFormActionsFactoryTests`.

#### Case 1
1. Open any type of product (simple, variable, grouped, etc...) with some Linked Products. You can some linked products from wp-admin.
2. If everything is in place, you will see the new `Linked Products` cell at the end of the list, showing upsell and cross-sell products info. The action will be implemented in the next PR.
3. If for that product the "Add more details" button is enabled, tapping it you shouldn't see the Linked Products cell in the bottom action sheet.

#### Case 2
1. Open any type of product (simple, variable, grouped, etc...) without Linked Products.
2. Tap the "Add more details" button.
3. You should see the `Linked Products` cell in the bottom action sheet. The action will be implemented in the next PR.

## Screenshots
| Product with Linked products   |  If the Product has no Linked products, you will see the cell inside the bottom action sheet |
Simulator Screen Shot - iPhone 11 Pro - 2020-11-17 at 20 01 24 | Simulator Screen Shot - iPhone 11 Pro - 2020-11-17 at 20 01 17

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
@@ -578,7 +578,7 @@ private extension DefaultProductFormTableViewModel {
                                                            comment: "Format of the number of Downloadable Files row in the plural form. Reads, `5 files`")
 
         // Linked Products
-        static let linkedProductsTitle = NSLocalizedString("Linked Products",
+        static let linkedProductsTitle = NSLocalizedString("Linked products",
                                                            comment: "Title of the Linked Products row on Product main screen")
         static let singularUpsellProductFormat =
             NSLocalizedString("%ld upsell product",
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
@@ -432,12 +432,10 @@ private extension DefaultProductFormTableViewModel {
         let icon = UIImage.linkedProductsImage
         let title = Localization.linkedProductsTitle
 
-        var linkedProductsDetails = [String]()
-
-        linkedProductsDetails.append(Localization.upsellProducts(count: product.upsellIDs.count))
-        linkedProductsDetails.append(Localization.crossSellProducts(count: product.crossSellIDs.count))
-
-        let details = linkedProductsDetails.isEmpty ? nil : linkedProductsDetails.joined(separator: "\n")
+        let details = [
+            Localization.upsellProducts(count: product.upsellIDs.count),
+            Localization.crossSellProducts(count: product.crossSellIDs.count),
+        ].joined(separator: "\n")
 
         return ProductFormSection.SettingsRow.ViewModel(icon: icon,
                                                         title: title,
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
@@ -434,19 +434,9 @@ private extension DefaultProductFormTableViewModel {
 
         var linkedProductsDetails = [String]()
 
-        if product.upsellIDs.count <= 1 {
-            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.singularUpsellProductFormat, product.upsellIDs.count))
-        }
-        else {
-            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.pluralUpsellProductsFormat, product.upsellIDs.count))
-        }
+        linkedProductsDetails.append(Localization.upsellProducts(count: product.upsellIDs.count))
+        linkedProductsDetails.append(Localization.crossSellProducts(count: product.crossSellIDs.count))
 
-        if product.crossSellIDs.count <= 1 {
-            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.singularCrossSellProductFormat, product.crossSellIDs.count))
-        }
-        else {
-            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.pluralCrossSellProductsFormat, product.crossSellIDs.count))
-        }
         let details = linkedProductsDetails.isEmpty ? nil : linkedProductsDetails.joined(separator: "\n")
 
         return ProductFormSection.SettingsRow.ViewModel(icon: icon,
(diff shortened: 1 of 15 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -545,5 +576,21 @@ private extension DefaultProductFormTableViewModel {
                                                                comment: "Format of the number of Downloadable Files row in the singular form. Reads, `1 file`")
         static let pluralDownloadsFormat = NSLocalizedString("%ld files",
                                                            comment: "Format of the number of Downloadable Files row in the plural form. Reads, `5 files`")
+
+        // Linked Products
+        static let linkedProductsTitle = NSLocalizedString("Linked Products",
```
shiki: ⚠️  The casing seems inconsistent with the other titles. For example, for “Downloadable files”, the *f* is not capitalized. But for this “Linked Products”, the *p* is. I'm not sure which one we should be using. 🤷

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -423,6 +425,35 @@ private extension DefaultProductFormTableViewModel {
...
+        let title = Localization.linkedProductsTitle
+
+        var linkedProductsDetails = [String]()
+
+        if product.upsellIDs.count <= 1 {
+            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.singularUpsellProductFormat, product.upsellIDs.count))
+        }
+        else {
+            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.pluralUpsellProductsFormat, product.upsellIDs.count))
+        }
```
shiki: ℹ️  It might be better to move this logic to `Localization` instead. 

```swift
enum Localization {
    static func upsellProducts(count: Int) -> String {
            let format: String = {
                if count <= 1 {
                    return NSLocalizedString("%ld upsell product",
                                             comment: "Format of upsell linked products row in the singular form. Reads, `1 upsell product`")
                } else {
                    return NSLocalizedString("%ld upsell products",
                                             comment: "Format of upsell linked products row in the plural form. Reads, `5 upsell products`")
                }
            }()

            return String.localizedStringWithFormat(format, count)
        }
}
```

And then this line would just be:

```swift
        var linkedProductsDetails = [String]()

        linkedProductsDetails.append(Localization.upsellProducts(count: product.upsellIDs.count))
```

In this way, I believe we're giving more emphasis to what is actually happening in the `linkedProductsRow` function and leaving the tiny language details inside `Localization`. This will probably also make it easier to find pluralization logic if we ever migrate to SwiftGen.

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -423,6 +425,35 @@ private extension DefaultProductFormTableViewModel {
...
+
+        if product.crossSellIDs.count <= 1 {
+            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.singularCrossSellProductFormat, product.crossSellIDs.count))
+        }
+        else {
+            linkedProductsDetails.append(String.localizedStringWithFormat(Localization.pluralCrossSellProductsFormat, product.crossSellIDs.count))
+        }
+        let details = linkedProductsDetails.isEmpty ? nil : linkedProductsDetails.joined(separator: "\n")
```
shiki: It doesn't look like `linkedProductsDetails.isEmpty` will ever be `true`, right? In any case, I tried this in a playground:

```swift
let names = [String]()
let x = names.joined(separator: "_SEPARATOR_")

print(x)
```

And `x` is just an empty string. So I believe we can simplify this as:

```suggestion
        let details = linkedProductsDetails.joined(separator: "\n")
```

With my other suggestion above, this block could look like:

```swift
let details = [
    Localization.upsellProducts(count: product.upsellIDs.count),
    Localization.crossSellProducts(count: product.crossSellIDs.count),
].joined(separator: "\n")
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormActionsFactory.swift
```diff
@@ -208,6 +219,8 @@ private extension ProductFormActionsFactory {
             return product.product.categories.isNotEmpty
         case .tags:
             return product.product.tags.isNotEmpty
+        case .linkedProducts:
+            return isEditProductsRelease5Enabled && (product.upsellIDs.count > 0 || product.crossSellIDs.count > 0)
```
shiki, line 223: I think we have `isNotEmpty`. 🙂 

```suggestion
            return isEditProductsRelease5Enabled && (product.upsellIDs.isNotEmpty || product.crossSellIDs.isNotEmpty)
```

WooCommerce/WooCommerceTests/Mockups/MockProduct.swift
```diff
@@ -42,7 +42,9 @@ final class MockProduct {
                  menuOrder: Int = 0,
                  categories: [ProductCategory] = [],
                  tags: [ProductTag] = [],
-                 images: [ProductImage] = []) -> Product {
+                 images: [ProductImage] = [],
+                 upsellIDs: [Int64] = [99, 1234566],
+                 crossSellIDs: [Int64] = [1234, 234234, 3]) -> Product {
```
shiki, line 47: I would suggest leaving these as empty arrays and let the unit tests define their values instead. But since these are the original values, we'd probably break other unit tests. 🙈

WooCommerce/WooCommerceTests/ViewRelated/Products/Actions Factory/ProductFormActionsFactoryTests.swift
```diff
@@ -87,6 +87,62 @@ final class ProductFormActionsFactoryTests: XCTestCase {
         XCTAssertEqual(factory.bottomSheetActions(), expectedBottomSheetActions)
     }
 
+    func test_viewModel_for_physical_simple_product_without_linked_products() {
```
shiki: What do you think about describing the expectations too? 

```suggestion
    func test_viewModel_for_product_without_linked_products_shows_editLinkedProducts_in_bottom_sheet() {
```

I don't believe we need to mention “physical and simple” since the constraints seem to be only for the `upsellIDs` and `crossSellIDs`.

WooCommerce/WooCommerceTests/ViewRelated/Products/Actions Factory/ProductFormActionsFactoryTests.swift
```diff
@@ -87,6 +87,62 @@ final class ProductFormActionsFactoryTests: XCTestCase {
...
+                                                                       .tags(editable: true),
+                                                                       .shortDescription(editable: true),
+                                                                       .productType(editable: true)]
+        XCTAssertEqual(factory.settingsSectionActions(), expectedSettingsSectionActions)
+
+        let expectedBottomSheetActions: [ProductFormBottomSheetAction] = [.editLinkedProducts]
+        XCTAssertEqual(factory.bottomSheetActions(), expectedBottomSheetActions)
+    }
+
+    func test_viewModel_for_physical_simple_product_with_linked_products() {
```
shiki: Same as above, what about mentioning the expectation? 🙂 

```suggestion
    func test_viewModel_for_product_with_linked_products_shows_linkedProducts_action_in_settings_section() {
```

shiki (approved):
Looking good, @pmusolino! I have one concern that I'm sure will be fixed (or ignored) easily —marked with ⚠️. The others are suggestions. 🙂

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
pmusolino: Good idea, done ✅

WooCommerce/WooCommerceTests/Mockups/MockProduct.swift
pmusolino: Yes, unfortunately, we followed this logic from the start also with other properties, so I followed the same convention.

pmusolino:
Thank you so much for the review @shiki! I applied all the suggested changes :) I'll wait for the review of @Garance91540 before merging it

Garance91540 (commented):
Hi @pmusolino ,

one quick note about the order of the "Linked product" row on the Product Details page: it should appear before the "Product Type" row, which should always be last. https://d.pr/i/uHwh0c

pmusolino:
Nice catch @Garance91540, I fixed it here 23db87f23344388fba0af13db1ed918d128fef14

PR #3168, opened 2020-11-18
Shipping Labels M1: create new entities in storage layer

### LINKED ISSUE #3113: Shipping Labels M1: create new entities in storage layer
We have four new entities for shipping labels M1: `ShippingLabel`, `ShippingLabelRefund`, `ShippingLabelSettings`, and `ShippingLabelAddress`.

Also, we're adding relationships to `Order`:

## Order

relationship name | type | description | notes
-- | -- | -- | --
shippingLabels | ShippingLabel | A collection of shipping labels. | Optional, to-many
shippingLabelSettings | ShippingLabelSettings | Shipping label settings for this order. | Optional, to-one

## ShippingLabel

attribute name | type | corresponding API field | description | notes
-- | -- | -- | -- | --
shippingLabelID | Int64 | label_id | The remote ID of the shipping label. |
orderID | Int64 | passed externally - `orderId` in the top level API response | The remote ID of the order that owns the shipping label. | The value is not at the same level of the shipping label API response and has to be passed to the mapper.
siteID | Int64 | passed externally | The remote ID of the site that owns this shipping label. | Just like how we pass `siteID` to other model mappers.
carrierID | String | carrier_id | The remote ID of the shipping carrier. | 
dateCreated | Date | created_date (requires type transformation) | The date the shipping label was created. | The API returns an integer UNIX timestamp in ms.
packageName | String | package_name | The name of the package for the shipping label. | 
rate | Double | rate | The cost of the shipping label. | 
currency | String | currency | The currency of the shipping label. |
trackingNumber | String | tracking | The tracking number of the shipping label. | 
serviceName | String | service_name | The name of service for the shipping label (e.g. "USPS - Media Mail").
refundableAmount | Double | refundable_amount | The amount that is refundable.
status | String | status | The status of the shipping label (e.g. "PURCHASED").
productIDs | [Int64] | product_ids | A list of product IDs included in the shipping label. | Optional, since this is only available in the new API (p91TBi-2G0-p2#comment-1952)
productNames | [String] | product_names | A list of product names included in the shipping label. | Non-optional, this is used to look up a product in case `productIDs` is not available.

relationship name | type | corresponding API field | description | notes
-- | -- | -- | -- | --
refund | ShippingLabelRefund | refund | Optional refund details if the user has requested refund for the shipping label. | Optional, to-one
originAddress | ShippingLabelAddress | formData.origin | The sender (store owner)'s address. | Non-optional, to-one
destinationAddress | ShippingLabelAddress | formData.destination | The recipient (customer)'s address. | Non-optional, to-one

## ShippingLabelRefund

attribute name | type | corresponding API field | description | notes
-- | -- | -- | -- | --
dateRequested | Date | request_date | The date of refund request. | The API returns an integer UNIX timestamp in ms.
status | String | status |  The status of the refund (e.g. "pending").

## ShippingLabelSettings

attribute name | type | corresponding API field | description | notes
-- | -- | -- | -- | --
paperSize | String | paperSize | The default paper size for reprinting a shipping label (current options are "label", "legal", and "letter"). | TODO: clarify if this is needed
orderID | Int64 | passed externally - `orderId` in the top level API response | The remote ID of the order that owns the shipping label settings. | 
siteID | Int64 | passed externally | The remote ID of the site that owns the order. | Just like how we pass `siteID` to other model mappers.

## ShippingLabelAddress

attribute name | type | corresponding API field | description | notes
-- | -- | -- | -- | --
company | String | company | The name of the company at the address. | 
name | String | name | The name of the sender/receiver at the address. | 
phone | String | phone | The contact phone number at the address. |
country | String | country | The country the address is in (ISO code). | 
state | String | state | The state the address is in (ISO code). | 
address1 | String | address | The first line of address (street, number, floor, etc.). | 
address2 | String | address_2 | The second line of address, empty if the address is only one line. | 
city | String | city | The city the address is in. | 
postcode | String | postcode | Postal code of the address. |

jaclync:
Fixes #3113 

The PR size is ~830 diffs, but ~592 diffs are on the new model version 38 and bunch of auto-generated code for Core Data subclasses!

## Changes

- Added model version 38 with changes in the following commits: https://github.com/woocommerce/woocommerce-ios/commit/14047aa272263c4bb2538cd8a5e10a9c6c056b93, https://github.com/woocommerce/woocommerce-ios/commit/d7a0d27f85fb5be4d047c96ff2f154d3255ca806, https://github.com/woocommerce/woocommerce-ios/commit/a9341bfca00b52f8e830ce2d994ef969f3df07cd, and https://github.com/woocommerce/woocommerce-ios/pull/3168/commits/759695fe4f2cbc52edd1bace0a32b46f97f57d1d - the last three commits were made after initial UI integration and PR feedback.
- Model version 38 has 4 new entities: `ShippingLabel`, `ShippingLabelAddress`, `ShippingLabelRefund`, and `ShippingLabelSettings`. Generated subclasses for these 4 entities.
- Added a test case to `MigrationTests` which verifies that these 4 new entities can be created in storage layer in model version 38.

## Testing

Please sanity check on migration:

- Launch the app from `develop` or any other branch without changes in this PR, and log in if necessary
- Launch the app from PR branch --> the app should run normally without crashing

Update release notes:


astralbodies (commented):
✅ Model 37 is unchanged from this PR and release/5.5.
✅ Only new entities between 37 & 38
✅ Relationship between ShippingLabel & other entities has cascade delete and an inverse relationship present.
❓ Should there be a relationship between Order and ShippingLabel & ShippingLabelSettings to prevent stale data from sticking around when the Order gets deleted?

Storage/StorageTests/CoreData/MigrationTests.swift
```diff
@@ -305,6 +305,43 @@ final class MigrationTests: XCTestCase {
...
+        let shippingLabelRefund = insertShippingLabelRefund(to: targetContext)
+        let shippingLabel = insertShippingLabel(to: targetContext)
+        shippingLabel.setValue(originAddress, forKey: "originAddress")
+        shippingLabel.setValue(destinationAddress, forKey: "destinationAddress")
+        shippingLabel.setValue(shippingLabelRefund, forKey: "refund")
+
+        // Creates an `ShippingLabelSettings`.
+        _ = insertShippingLabelSettings(to: targetContext)
+
+        try targetContext.save()
```
astralbodies: You may wish to wrap this in `XCTAssertNoThrow` and remove the throws from the test method. I know the rest of the class doesn't follow this, but it's nice to make sure there aren't any unexpected throws and it looks nicer on execution of the tests if there is a failure.

jaclync: good suggestion, updated in https://github.com/woocommerce/woocommerce-ios/pull/3168/commits/5520ca5e34f23076f981fbfec044e97c8ba6b1f1! however, I wasn't able to remove the `throws` from this test case because there are other `try` statements that assign to a constant value and I couldn't find a good way to use `XCTAssertNoThrow` for these two:

- `let sourceContainer = try startPersistentContainer("Model 37")`
- `let targetContainer = try migrate(sourceContainer, to: "Model 38")`

do you have any ideas?

jaclync:
> ❓ Should there be a relationship between Order and ShippingLabel & ShippingLabelSettings to prevent stale data from sticking around when the Order gets deleted?

Really great point @astralbodies! I added the relationships in https://github.com/woocommerce/woocommerce-ios/pull/3168/commits/759695fe4f2cbc52edd1bace0a32b46f97f57d1d and verified that the 4 shipping label tables became empty after switching to another store using a database inspector (DataGrip) 🧹 

Ready for another look!

Storage/StorageTests/CoreData/MigrationTests.swift
Ecarrion: What about?
```swift
let sourceContainer = XCTUnwrap(try? startPersistentContainer("Model 37"))
```

Not sure if that is better or worse 😅 
On one hand, the failures are easy to see but on the other, the code looks a bit odd.

jaclync: Ooh nice! I agree on the tradeoff, I think I slightly prefer code readability to how the test fails. What do you think @astralbodies?

shiki: I don't believe we need to change the things like `let sourceContainer = try startPersistentContainer("Model 37")`. 

Using `XCTAssertNoThrow` _can_ make sense for `try targetContext.save()` if we are _proving_ that it will not fail. I think for this case, we're not. From what I can tell, the proofs that we truly want are the `XCTAssertEqual()` below.

So, in this case, I would prefer to have a less awkward and more readable option.

Storage/Storage/Model/ShippingLabel+CoreDataClass.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
```
shiki, line 1: LOC-conservation trick for the future: We don't actually need to import `Foundation` if there are other imports that import it already, like `CoreData`. 

There! I saved you one line of change! 🤣

This is hinted in: https://github.com/raywenderlich/swift-style-guide/#minimal-imports

Storage/Storage/Model/ShippingLabel+CoreDataProperties.swift
```diff
@@ -0,0 +1,30 @@
...
+    @NSManaged public var carrierID: String
+    @NSManaged public var dateCreated: Date
+    @NSManaged public var packageName: String
+    @NSManaged public var rate: Double
+    @NSManaged public var currency: String
+    @NSManaged public var trackingNumber: String
+    @NSManaged public var serviceName: String
+    @NSManaged public var refundableAmount: Double
+    @NSManaged public var status: String
+    @NSManaged public var productIDs: [Int64]?
```
shiki: I wish this was not optional. I somehow find it unnecessarily tricky to deal with optional arrays. But no worries.

Storage/Storage/Model/ShippingLabelAddress+CoreDataProperties.swift
```diff
@@ -0,0 +1,21 @@
...
+
+    @NSManaged public var company: String
+    @NSManaged public var name: String
+    @NSManaged public var phone: String
+    @NSManaged public var country: String
+    @NSManaged public var state: String
+    @NSManaged public var address1: String
+    @NSManaged public var address2: String
+    @NSManaged public var city: String
+    @NSManaged public var postcode: String
```
shiki, line 19: ⚠️ This is missing `originShippingLabel` and `destinationShippingLabel`. I guess that's intentional, right? We probably don't need them.

shiki (approved):
✔️ Compared before and after. Looks legit. 
✔️ Tested migration. It didn't cry. 
✔️ Checked that the `Order` relationships are there. I wish that those are included in the unit tests though. 😉 

I just have one clarifying question but I don't think it's that important to be done. 

:shipit:

Storage/Storage/Model/ShippingLabel+CoreDataClass.swift
jaclync: haha, these files were automatically generated by Xcode I think I'll leave them as they are for now 🙂 guess Apple doesn't follow this style guide 🙄

Storage/Storage/Model/ShippingLabel+CoreDataProperties.swift
jaclync: Hmm actually, I'll make this non-optional. I agree that optional arrays are a bit unnecessary since we can often check whether it's empty (unless an empty array has other meaning). The readonly model in the networking layer actually has this property non-optional as well 👍

Storage/Storage/Model/ShippingLabelAddress+CoreDataProperties.swift
jaclync: That must have been my oversight not regenerating the subclass for `ShippingLabelAddress` when I made other changes, thanks for spotting this 🦅 👁️

jaclync: Added here: https://github.com/woocommerce/woocommerce-ios/pull/3168/commits/3cf9c86e0d00a6c19283ccc3c3c2492516d444bc

Storage/Storage/Model/ShippingLabel+CoreDataProperties.swift
jaclync: Updated here: https://github.com/woocommerce/woocommerce-ios/pull/3168/commits/9784a6e40078766f2801cbc3754ca1ecbdc7e0fc

jaclync:
Thanks a lot @shiki and @astralbodies for the thorough review 💟 I responded to your comments and did integration testing, merging now!

> Checked that the Order relationships are there. I wish that those are included in the unit tests though. 😉

Great point! added here: https://github.com/woocommerce/woocommerce-ios/pull/3168/commits/a2dfbd18f0fba3710fd35865e80382d1f50d1eca

PR #3179, opened 2020-11-19
Linked Products: added Upsells and Cross-sells screen

### LINKED ISSUE #3072: Linked Products: add Upsells and Cross-sells screen (part 2)
Ref p91TBi-2XV-p2

Tapping the Linked Product cell, will present the option to see/add Upsells and Cross-sells products options. 

## Goals 
- [x] Show the page from the cell mentioned above. 
- [x] Show the number of linked products (upsells and cross-sells). Show “Add Products” if there are no linked Upsells or Cross-sells products.
- [x] Show “Edit Products” if there are already linked upsells or cross-sells. 
- [ ] Show “<NUMBER> products“ if there are already linked products for that category. The screenshot below is **incorrect**. We don't need to show “added”. 

## Design
| Empty state            |  Populated state |
Schermata 2020-10-28 alle 15 20 21 | Schermata 2020-10-28 alle 15 20 34

pmusolino:
Fixes #3072 

## Description
In this PR I worked on implementing the first screen of the Linked Products flow that shows the number of Upsells and Cross-sell products already selected, and show the two buttons for opening the details (the actions will be handled in another PR). 
I'm sorry for the 600+ lines of code, but around 100+ lines are related to the new Xibs files.

## Changes
- Added two new icons, used in the screen for Linked Products + tests.
- Created a new cell `NumberOfLinkedProductsTableViewCell` used for showing the number of upsells and cross-sells products.
- Implemented the new `LinkedProductsViewController` that handles the UI for the Linked Products screen.
- implemented the `LinkedProductsViewModel` that provides the view data for Linked Products and handles the init/UI/navigation actions needed + tests.
- Edited `ProductFormViewController` that now is handling the opening of the Linked Products screen.

## Testing
1. Open a product.
2. If that product already had some Linked products, tap the cell showed; otherwise, tap "Add more details" and tap the "Linked products" cell.
3. The new screen "Linked Products", will be presented, showing Upsells and Cross-sells products.
4. If one of the two sections has some products already added, you will see another cell showing the number of products under upsells or/and cross-sells, and the button will show the new title "Edit Products" instead of "Add Products".

## Screenshots
| Linked Products screen             |  Linked Products screen (dark mode enabled) |
Simulator Screen Shot - iPhone 11 Pro - 2020-11-19 at 16 46 42 | Simulator Screen Shot - iPhone 11 Pro - 2020-11-19 at 16 46 57

Update release notes:


WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -13,6 +13,12 @@ extension UIImage {
         return UIImage.gridicon(.addOutline)
     }
 
+    /// Arrow Up Icon
```
ctarda, line 16: I would not consider this a blocker, but as far as I recall, we used to add a test for these icons in `IconsTests` (a very simple test that ensures that the image is not nil). A bit overkill, but prevents a runtime crash if an image is missing from the final binary.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/Cells/NumberOfLinkedProductsTableViewCell.swift
```diff
@@ -0,0 +1,23 @@
+import UIKit
+
+class NumberOfLinkedProductsTableViewCell: UITableViewCell {
```
ctarda: Would it make sense to declare this class as final? Or are we actually planning on subclassing it in the future?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewModel.swift
```diff
@@ -0,0 +1,76 @@
+import UIKit
+import Yosemite
+
+/// Provides data needed for Linked Products.
+///
+protocol LinkedProductsViewModelOutput {
```
ctarda: I'm just curious here: what is the reason behind the two protocols? I understand this provides a kind of [command-query separation](https://martinfowler.com/bliki/CommandQuerySeparation.html) but I don't see the practical benefit of formalizing that separation. Wha am I missing?

Leaving that aside, there is one more thing I am curious about: conformance to LinkedProductsViewModelActionHandler is implemented in an extension, while conformance to LinkedProductsViewModelOutput is implemented in the class itself. That reads a bit un-symmetric, would it make sense to implement conformance to each protocol in a separate extension?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewModel.swift
```diff
@@ -0,0 +1,76 @@
...
+        self.product = product
+        upsellIDs = product.upsellIDs
+        crossSellIDs = product.crossSellIDs
+    }
+
+    var sections: [Section] {
+        var rows = [Row]()
+
+        rows.append(.upsells)
+        if upsellIDs.count > 0 {
```
ctarda: I think Array implements an `isEmpty` property. It seems like that property would clarify the intent more than checking the count? What do you think?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewModel.swift
```diff
@@ -0,0 +1,76 @@
...
+    func handleUpsellIDsChange(_ upsellIDs: [Int64]) {
+        self.upsellIDs = upsellIDs
+    }
+
+    func handleCrossSellIDsChange(_ crossSellIDs: [Int64]) {
+        self.crossSellIDs = crossSellIDs
+    }
+
+    func hasUnsavedChanges() -> Bool {
+        guard upsellIDs != product.upsellIDs || crossSellIDs != product.crossSellIDs else {
```
ctarda, line 53: Would it make sense to encapsulate this logic in a private method, with a name that conveys what this logic check actually does? The reason I ask is because, in my opinion, that would make this guard easier to understand in the long term.

WooCommerce/WooCommerceTests/Extensions/IconsTests.swift
```diff
@@ -7,6 +7,10 @@ final class IconsTests: XCTestCase {
         XCTAssertNotNil(UIImage.addOutlineImage)
     }
 
+    func test_arrowUp_icon_is_not_nil() {
```
ctarda, line 10: Oops! Now I see the tests I commented about above. Sorry!

ctarda (approved):
This is great @pmusolino !

I made a few comments (mostly questions 😊), but I nothing is really a blocker.

This is what I see when a Product does not have any previously linked product:

Product without linked products. Step 1:


Product without linked products. Step 2: (after tapping Add more details > Linked Products)



Product with a linked Product: Step 1:


Product with linked Products: Step 2 (the linked product count shows correctly in the correct section):


I won't include Dark mode screenshots, but the results are the same.

One thing I am not sure is worth pointing out: none of the buttons in the Linked Products Screen (including "Done") seem to work.

WooCommerce/Classes/Extensions/UIImage+Woo.swift
ctarda: Aaaaand I saw the tests down below. Sorry!

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/Cells/NumberOfLinkedProductsTableViewCell.swift
pmusolino: It makes sense, done ✅

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewModel.swift
pmusolino: Probably you are right, declaring two protocols for formalizing a separation is a little bit overkill here (also because in the future this view model will not grow), so I removed the two protocols 👍 5cfc7159a499e53bb55cceff22401fb6e1e8cbd4

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewModel.swift
pmusolino: Agree ;) done ✅

pmusolino:
Thank you so much @ctarda for the review! ;) I updated the PR following your suggestions. 💯

Garance91540 (commented):
Hi @pmusolino ! 

I've some styling notes:
Screen Shot 2020-11-20 at 5 35 02 PM

1. Vertical padding between "Upsells" card and screen header. 
Can we update the vertical padding between these 2 items to 16px?

2. Icons in the "Upsells" and "Cross-sells" cards
Can we update the left and right padding of these icons to be 16px, and vertical padding to 10px? 

3. "Add Products" CTA
Can we update the vertical and horizontal padding for these buttons to 16px? 

Thank  you!

pmusolino:
I'm going to merge this PR since the changes needed on the UI was partially handled in this other PR https://github.com/woocommerce/woocommerce-ios/pull/3199

PR #3184, opened 2020-11-20
[Issue Refund] Handle previous shippings refunds

### LINKED ISSUE #2967: [Issue Refunds] Handle previous refunds

Ecarrion:
closes #2967 

_Disclaimer the PR is not that big, 598 from the 835 lines are from the new `xcdatamodel` file_

# Why 

This PR takes care of handling previous shipping refunds on the issue refund screen. The logic goes as:

**Prerequisite:** Order must have a non-free shipping line
- If there are no shipping refunds, allow the user to refund shipping.
- If there is any shipping refund**¹**, don't allow the user to refund shipping.
- If there is no information about shipping refunds**²**, don't allow the user to refund shipping.

**¹** Since we don't allow partial shipping refund on the app, I'm treating any shipping refund as a complete shipping refund
**²** Since the shipping refund information only comes on stores `v.4.8.0` or greater. I'm disabling shipping refunds if the order has a refund but previous shipping refund information is not available.

# How

**On Networking**
- Added `shippingLines: [ShippingLine]?` property to `Refund` entity.

**On Storage**
- Added `shippingLines: Set<ShippingLine>?` relationship to `Refund` CD class with `cascade` deletion rule.
- Added `refund: Refund?` inverse relationship to `ShippingLine` CD class with `nulify` deletion rule.
- Rename `loadShippingLine` method on `StorageType+Extension` to `loadOrderShippingLine`.
- Added a `loadRefundShippingLine` method on `StorageType+Extension`.

**On Yosemite**
- Update `upsertStoredRefunds` method on `RefundStore` to handle the refund's shipping lines which internally also handles the refund's shipping lines taxes.

**On WooCommerce**
- Added `hasShippingBeenRefunded() -> Bool?` method on `IssueRefundViewModel` to be used when deciding if the refund shipping section should be shown or not.

# Demo 

handling-shipping-refunds

# Testing Steps

**No Previous Shipping Refunds**
- Go to a non-refunded order with shipping costs
- Tap on "Issue Refund" button
- Confirm that the "Refund Shipping" section is visible

**Unknown Previous Shipping Refunds**
- Use a store with a version older than `4.8.0`
- Go to an order that has at least one refund but is not totally refunded.
- Tap on "Issue Refund" button
- Confirm that the "Refund Shipping" section is not visible.

**Previous Shipping Refunds**
- Update your store to `4.8.0 beta 1` using the [beta tester plugin](https://wordpress.org/plugins/woocommerce-beta-tester/)
- Go to an order that has a previous shipping refund but is not totally refunded.
- Tap on "Issue Refund" button
- Confirm that the "Refund Shipping" section is not visible.

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
+++ b/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
@@ -16,6 +16,7 @@ extension Storage.Refund: ReadOnlyConvertible {
         amount = fullRefund.amount
         reason = fullRefund.reason
         byUserID = fullRefund.refundedByUserID
+        supportShippingRefunds = fullRefund.shippingLines != nil
 
         if let automated = fullRefund.isAutomated {
             isAutomated = automated
@@ -30,7 +31,14 @@ extension Storage.Refund: ReadOnlyConvertible {
     ///
     public func toReadOnly() -> Yosemite.Refund {
         let orderItems = items?.map { $0.toReadOnly() } ?? [Yosemite.OrderItemRefund]()
-        let readOnlyShippingLines = shippingLines?.map { $0.toReadOnly() }
+
+        // Assign nil if the refund does not support shipping information
+        let readOnlyShippingLines: [ShippingLine]? = {
+            guard supportShippingRefunds else {
+                return nil
+            }
+            return shippingLines?.map { $0.toReadOnly() }
+        }()
 
         return Refund(refundID: refundID,
                       orderID: orderID,
--- a/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
+++ b/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
@@ -30,7 +30,7 @@ extension Storage.Refund: ReadOnlyConvertible {
     ///
     public func toReadOnly() -> Yosemite.Refund {
         let orderItems = items?.map { $0.toReadOnly() } ?? [Yosemite.OrderItemRefund]()
-        let readOnlyShippingLines = shippingLines.map { $0.toReadOnly() }
+        let readOnlyShippingLines = shippingLines?.map { $0.toReadOnly() }
 
         return Refund(refundID: refundID,
                       orderID: orderID,
--- a/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
+++ b/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
@@ -30,6 +30,7 @@ extension Storage.Refund: ReadOnlyConvertible {
     ///
     public func toReadOnly() -> Yosemite.Refund {
         let orderItems = items?.map { $0.toReadOnly() } ?? [Yosemite.OrderItemRefund]()
+        let readOnlyShippingLines = shippingLines.map { $0.toReadOnly() }
 
         return Refund(refundID: refundID,
                       orderID: orderID,
@@ -41,6 +42,6 @@ extension Storage.Refund: ReadOnlyConvertible {
                       isAutomated: isAutomated,
                       createAutomated: createAutomated,
                       items: orderItems,
-                      shippingLines: [])
+                      shippingLines: readOnlyShippingLines)
     }
 }
--- a/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
+++ b/Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
@@ -40,6 +40,7 @@ extension Storage.Refund: ReadOnlyConvertible {
                       refundedByUserID: byUserID,
                       isAutomated: isAutomated,
                       createAutomated: createAutomated,
-                      items: orderItems)
+                      items: orderItems,
+                      shippingLines: [])
     }
 }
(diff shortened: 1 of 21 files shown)


jaclync:
Nice work! Supporting multiple WC versions is hard 😅 

Core Data data model version checklist: ✅ ❓ maybe a test case migrating from model version 38 to 39 based on [this doc](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/creating-core-data-model-versions.md#always-add-unit-tests)?
No Previous Shipping Refunds: ✅ 
Previous Shipping Refunds: ✅ 
Unknown Previous Shipping Refunds: ❓ - on a test store with WC 4.7.0, I refunded one of two products (I tried one refund without shipping and another refund on partial shipping), and the "Refund Shipping" section was visible:

2 refunds | issue refund (shipping still visible)
-- | --
Simulator Screen Shot - iPhone 11 - 2020-11-23 at 21 02 09 | Simulator Screen Shot - iPhone 11 - 2020-11-23 at 21 02 13

Also, I encountered a weird crash every time I tried issuing refund on a product in the app (Xcode 12.1, iPhone 11 iOS 14.1 simulator):


stack trace (partial main thread):

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x0000000114ca5638 in swift_retain ()
#1	0x0000000114cf572c in swift_bridgeObjectRetain ()
#2	0x000000011434e1ff in outlined init with copy of Refund ()
#3	0x0000000114397f65 in RefundMapper.map(response:) at /Users/jaclynchen/Developer/testing/woocommerce-ios/Networking/Networking/Mapper/RefundMapper.swift:30
#4	0x00000001143984c7 in protocol witness for Mapper.map(response:) in conformance RefundMapper ()
#5	0x00000001142dfe1c in closure #1 in Remote.enqueue<A>(_:mapper:completion:) at /Users/jaclynchen/Developer/testing/woocommerce-ios/Networking/Networking/Remote/Remote.swift:85
#6	0x00000001142e03c0 in partial apply for closure #1 in Remote.enqueue<A>(_:mapper:completion:) ()
#7	0x000000011431177f in closure #1 in AlamofireNetwork.responseData(for:completion:) at /Users/jaclynchen/Developer/testing/woocommerce-ios/Networking/Networking/Network/AlamofireNetwork.swift:50
#8	0x000000011158bdf1 in closure #1 in closure #1 in DataRequest.response<A>(queue:responseSerializer:completionHandler:) at /Users/jaclynchen/Developer/testing/woocommerce-ios/Pods/Alamofire/Source/ResponseSerialization.swift:167
```

```
warning: Swift error in fallback scratch context: error: /Users/jaclynchen/Developer/woocommerce-ios/Pods/Sodium-Fork/Sodium/libsodium/module.modulemap:1:8: error: redefinition of module 'Clibsodium'
module Clibsodium {
       ^

/Users/jaclynchen/Developer/testing/woocommerce-ios/Pods/Sodium-Fork/Sodium/libsodium/module.modulemap:1:8: note: previously defined here
module Clibsodium {
       ^

error: could not build Objective-C module 'Alamofire'


note: This error message is displayed only once. If the error displayed above is due to conflicting search paths to Clang modules in different images of the debugged executable, this can slow down debugging of Swift code significantly, since a fresh Swift context has to be created every time a conflict is encountered.
```

Ecarrion:

> Core Data data model version checklist: ✅ ❓ maybe a test case migrating from model version 38 to 39 based on this doc?

Sure, added in https://github.com/woocommerce/woocommerce-ios/pull/3184/commits/e8dc6d441e54517c223e733d1f1f2a4ede108196


> Unknown Previous Shipping Refunds: ❓ - on a test store with WC 4.7.0, I refunded one of two products (I tried one refun> d without shipping and another refund on partial shipping), and the "Refund Shipping" section was visible:

Apparently, I made a mistake when testing this because it turns out that you can't have a `nil` `one-to-many` relationship in core data entities 🤦‍♂️  and I was using that to know if shipping refunds were unknown.

Now, I've added a new property called `supportShippingRefunds` which is set to false when the `Networking.Refund` has `nil` shipping lines https://github.com/woocommerce/woocommerce-ios/pull/3184/commits/5432059747c4308e06273229694a0b1d89a88f13

unknown-refund

> Also, I encountered a weird crash every time I tried issuing refund on a product in the app (Xcode 12.1, iPhone 11 iOS 14.1 simulator):

I've been issuing refunds on many stores and haven't been able to reproduce this 🤔

Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
```diff
@@ -16,6 +16,7 @@ extension Storage.Refund: ReadOnlyConvertible {
         amount = fullRefund.amount
         reason = fullRefund.reason
         byUserID = fullRefund.refundedByUserID
+        supportShippingRefunds = fullRefund.shippingLines != nil
```
jaclync, line 19: nit: what about also having this property in the Networking layer and logic there so that we can test this property along with other properties? it might be hard to find this property here in the future 🙂

jaclync (approved):
Code and testing are both ✅ now! Thanks for adding the migration test case 😄

Yosemite/Yosemite/Model/Storage/Refund+ReadOnlyConvertible.swift
Ecarrion: I didn't add this to the networking entity because of two reasons:
1. I thought networking entities should be a 1-1 match with API objects.
2. I plan to delete this once we add support to something like this https://github.com/woocommerce/woocommerce-ios/issues/3200

Issue #3151, opened 2020-11-12
Linked Products: Add Tracking


shiki:
These events are supposed to answer the question “Are people using the new feature?”. If they don't, please let me know. 🙂 I'm also avoiding proposing too many event triggers. If you feel like these are still too many then you're probably right! 

## The Button

<img src="https://user-images.githubusercontent.com/198826/99010707-bf55c480-2507-11eb-9977-36f8b280b944.png" width="300">

Event Name | Trigger
--------|-------
`product_detail_view_linked_products_tapped`        |  When the Linked products button is tapped as shown above.     

This is already implemented in Android.

## Linked Products

<img src="https://user-images.githubusercontent.com/198826/99008473-78fe6680-2503-11eb-949b-b907e9dd4dab.png" width="300">

Event Name | Trigger | Custom Properties
--------|-------|----
`linked_products`        |  When this screen is shown.      |  `action="shown"`
`linked_products` | When the Done button is tapped. This indicates that the user has changed something and is definitely using the linked products feature. | `action="done"`

## Linked Products Selector 

<img src="https://user-images.githubusercontent.com/198826/99009362-2faf1680-2505-11eb-8947-cd74aa5b4b35.png" width="300">

Let's rename the existing grouped products selector events so they are more generic. We'll use `context` to differentiate between grouped products, cross-sells, and upsells.

Old Event Name | New Event Name | Custom Properties
--------|------|--
   `grouped_product_linked_products_add_tapped`    |  `connected_products_selector` | `action: "add_tapped", context: "grouped_products OR upsells OR cross_sells"`
   `grouped_product_linked_products_added`    |  `connected_products_selector` | `action: "added", context: "grouped_products OR upsells OR cross_sells"`
   `grouped_product_linked_products_done_button_tapped `    |  `connected_products_selector` | `action: "done_tapped", context: "grouped_products OR upsells OR cross_sells"`
   `grouped_product_linked_products_delete_tapped `    |  `connected_products_selector` | `action: "delete_tapped", context: "grouped_products OR upsells OR cross_sells"`

---

Corresponding Android issue: https://github.com/woocommerce/woocommerce-android/issues/3159

pmusolino:
@shiki Since there are some existing events in the product list selector and we are reusing all these components, should they match the events used before in grouped products? 
Consider that I make this flow generic, so we can pass new events for "products add tapped", "products added", "products done button tapped", and "delete tapped". 

These are the existing events for grouped products:
[...]

PR #3185, opened 2020-11-20
Linked Products: added Upsell and Cross-sell screen selectors + Products Selector Searching

### LINKED ISSUE #3073: Linked Products: add Upsell and Cross-sell screen selectors (part 3)
Ref p91TBi-2XV-p2

## Goals 

- [ ] Show a products list when tapping on “Add products” (see #3072) 
- [ ] Show the "Upsell/Cross-sell already added" screen when tapping on “Edit Products” (see #3072)
- [ ]  Allow multiple selections. This used to be “tap and hold” and Jayson confirmed with Garance that **multiple selections happen right away**. Just tapping will enable multiple selections. 
- [ ]  Tapping on Done will show the Linked Products screen.
- [ ] The list of selectable products should not include the products that are already linked. 

## Notes

- We could probably reuse the Grouped Products screens for this.
- Search will be handled in a different task. 

## Screenshots

| Add Upsell/Cross-sell screen             |  Upsell/Cross-sell already added | Selection style (similar to Grouped Products) |
Schermata 2020-10-28 alle 15 31 46 | Schermata 2020-10-28 alle 15 34 24 | Simulator Screen Shot - iPhone 11 Pro - 2020-10-28 at 15 33 35

### LINKED ISSUE #3074: Linked Products: Add Products Selector Searching (part 4)
Ref p91TBi-2XV-p2

This is shown after tapping on the search icon from the Products Selector. The search behaves similarly to the Grouped Products search.

## Goals
- [ ] Allow searching XD. We should probably not show the products that are already linked. 
- [ ] Allow multiple selection.

## Screenshots
Schermata 2020-10-28 alle 15 41 41

pmusolino:
Fixes #3073 
Fixes #3074 

## Description
In this PR, which should be merge after #3179, I modified our existing view controllers for selecting the products that should be linked to grouped products type, making them generic, and allowing us to open it also from the Linked Products screen.

## Changes
Unfortunately, it seems that Github isn't able to recognize that a file was renamed if I moved it under another folder, so some files that have been renamed like `LinkedProductsListSelectorViewController` are showed as a new file, also if I changed only some lines of code.

- `GroupedProductListSelectorDataSource` has been renamed in `LinkedProductListSelectorDataSource` + tests. I made it generic, removing all the references to grouped products, and I modified the init method that now accepts a list of product IDs and the event that should be fired when a user presses the delete button.
- `GroupedProductsListSelectorViewController` has been renamed in `LinkedProductsListSelectorViewController`. I made it generic, removing also here all the references to grouped products, and implementing a `ViewConfiguration` struct that will be useful for configuring externally the view, for example with a custom title view and with all the generic events that will be fired after an action.
- `GroupedProductsViewController.xib` has been renamed in `LinkedProductsListSelectorViewController`. Also, I fixed some constraints issues that were added earlier.
- `LinkedProductsViewController`: I added the way for presenting the `LinkedProductsListSelectorViewController`, handle the data and refresh the table view. The analytics events are temporary and will be part of the final issue related to Linked Products.
- I updated the `ProductFormViewController` with the new data requests by `LinkedProductsListSelectorViewController` when we try to present it related to grouped products.

#### Case 1 (Selection)
1. Open a product.
2. If that product already had some Linked products, tap the cell showed; otherwise, tap "Add more details" and tap the "Linked products" cell.
3. The new screen "Linked Products", will be presented, showing Upsells and Cross-sells products.
4. Tap the button "Add/Edit Products" under upsell or cross-sell products.
5. Tap "Add Products" on the new screen.
6. Select multiple products, and press "done". -> The products selected should be shown on the previous screen. 
7. Try to delete some of the added products.
8. Tap the button "Done" in the navigation bar.
9. The relative section under the Linked Products screen should be updated with the new number of products.

#### Case 2 (Search)
1. Open a product.
2. If that product already had some Linked products, tap the cell showed; otherwise, tap "Add more details" and tap the "Linked products" cell.
3. The new screen "Linked Products", will be presented, showing Upsells and Cross-sells products.
4. Tap the button "Add/Edit Products" under upsell or cross-sell products.
5. Tap "Add Products" on the new screen.
6. Press the "search" icon in the navigation bar.
7. Search and select the products that you want to add, and press "Done".
8. The selected products should appear in the list, selected.
9. Press done, and make sure that the relative section under the Linked Products screen are updated with the new number of products.

#### Case 3 (Grouped Products)
Since I made these Products List Selector screen generic, this had a small impact also on Grouped Products.
Please, open a grouped product, and try to select/remove/search/update grouped products, making sure no regressions have been introduced. The gif below better explain an update of grouped products.

## Screenshots
| Upsell Products selection             |  Upsell Products Search and selection | Grouped products update (no regressions introduced) |
:-------------------------:|:-------------------------: |:-------------------------:
Product Selection | product search | grouped products

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/LinkedProductListSelectorDataSource.swift
```diff
@@ -12,51 +12,58 @@ final class GroupedProductListSelectorDataSource: PaginatedListSelectorDataSourc
...
+    // Not used: a product's linked product list is not selectable in this use case.
     var selected: Product?
 
     private let siteID: Int64
     private let product: Product
     private let imageService: ImageService
+    private let deleteButtonTappedEvent: WooAnalyticsStat
 
-    init(product: Product, imageService: ImageService = ServiceLocator.imageService) {
+    init(product: Product,
```
ctarda, line 45: 🤔 I see one big drawback of DI via the constructor: it is easy to end up passing too many arguments, which ends up, eventually, making the code more difficult to understand. 

I am not advocating or suggesting making any change here, I am genuinely curious about your opinion: would it make sense to [pack these parameters into one struct/object](https://refactoring.com/catalog/introduceParameterObject.html), in order to pass only one parameter? 

On the one side that would make this constructor easier to read, on the other hand it introduces one layer of indirection that might not be necessary. What do you think?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/LinkedProductListSelectorDataSource.swift
```diff
@@ -70,21 +77,22 @@ final class GroupedProductListSelectorDataSource: PaginatedListSelectorDataSourc
         cell.update(viewModel: viewModel, imageService: imageService)
 
         cell.configureAccessoryDeleteButton { [weak self] in
-            ServiceLocator.analytics.track(.groupedProductLinkedProductsDeleteButtonTapped)
-            self?.deleteProduct(model)
+            guard let self = self else { return }
+            ServiceLocator.analytics.track(self.deleteButtonTappedEvent)
```
ctarda, line 81: This is also a genuine question: I am not sure why we want to inject this event. As far as I can tell from looking at the tests, (it might be I am missing something), we do not seem to test that the event is actually sent.

Are we reusing this data source with different view controllers, and therefore we want to be able to send different events to tracks depending on the context?

Because if that's the case, and building on the previous comment, maybe we could think about the parameters we pass to this class in a way that we provide an abstraction that models/describes the context?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/LinkedProductsListSelectorViewController.swift
```diff
@@ -0,0 +1,206 @@
+import UIKit
```
ctarda: I checked out the branch and checked the history of this file, and the history seems to be there, it has not been lost. I guess that means it's okay if it shows as a new file here?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/ProductsListSelectorViewController.swift
```diff
@@ -166,3 +168,19 @@ private extension GroupedProductsViewController {
         }
     }
 }
+
+extension LinkedProductsListSelectorViewController {
```
ctarda: I am curious about this. Being as it is a wrapper on a string, I was wondering if this s really necessary. I understand it would be consistent with other VCs though.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewController.swift
```diff
@@ -142,16 +146,57 @@ private extension LinkedProductsViewController {
...
+        let buttonTitle = Localization.buttonTitle(count: viewModel.upsellIDs.count)
+        cell.configure(style: .secondary, title: buttonTitle) { [weak self] in
+            self?.show(viewController, sender: self)
         }
     }
 
     func configureCrossSellsButton(cell: ButtonTableViewCell) {
-        let buttonTitle = Localization.buttonTitle(count: viewModel.crossSellIDs.count)
-        cell.configure(style: .secondary, title: buttonTitle) {
+        guard let product = product as? EditableProductModel else {
```
ctarda, line 177: It seems like the implementation of this method (configureCrossSellsButton) and the implementation of configureUpsellsButton are pretty similar. Would it make sense to attempt to extract and delegate as much as possible of that shared implementation to a separate method?

ctarda (commented):
The code looks great!. I asked a few questions here and there, I am curious about your take 😊

I didn't see any regressions with Grouped Products, but there seems to be a couple of things that might not be working as expected.

1. The "Done" button in the Linked Products screen does not seem to work all the time.


A way to break that button consistently seems to be:
* Navigate to a product with linked products.
* Tap the cell that navigates to the Linked Products screen
* Tap "Edit Products" or "Add Products"
* Add or remove a Linked Product. 
* When reaching the Linked Products screen again, the Done button does not seem to work.

2. When a Product does not have a linked product (see gif:)

* Tap "Add more details"
* Tap "Linked Products"
* Tap Upsells > Add Product
* Tap Add Products
* Tap the loupe icon 
* Search (notice also how the search results are a bit unexpected, I did not expect "Cup" to match a search for "Me")
* Select all items
* Tap Done
* Tap Done again
* Tap Done one more time to navigate back tot he Linked Products list
* The Done button does not work, so the only way out is tapping Back
* And therefore the linked products do not seem to be linked.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/LinkedProductListSelectorDataSource.swift
pmusolino: Good question @ctarda. In general, I prefer to introduce a `struct` if there are a lot of different parameters like I did in `LinkedProductsListSelectorViewController`, where I introduced the `struct ViewConfiguration`. Probably for this case is still good to not pack these parameters, just to not introduce another layer that is probably not necessary here.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/LinkedProductListSelectorDataSource.swift
pmusolino: > Are we reusing this data source with different view controllers, and therefore we want to be able to send different events to tracks depending on the context?

Correct. This data source will be used by `LinkedProductsListSelectorViewController`, which is now generic and used in two points of the app: Grouped Products and Linked Products. Since there are two points where the list selector will be presented, we need to track different events, with different names.

> Because if that's the case, and building on the previous comment, maybe we could think about the parameters we pass to this class in a way that we provide an abstraction that models/describes the context?

Do you mean to add a `struct` to be passed to the `init` method which contains only the events that we want to track (in this case, just one)?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/LinkedProductsListSelectorViewController.swift
pmusolino: It's showed as a new file, but as you can see, there are just a small bunch of differences from the previous version. Github shows it as a new file just because it was moved under a new folder 🤷‍♂️

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/ProductsListSelectorViewController.swift
pmusolino: It seems that this file called `ProductsListSelectorViewController` is in reality an old version of `LinkedProductsListSelectorViewController` which is not visible inside Xcode. I removed it 🙇

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewController.swift
pmusolino: This is true, for the moment they are similar, but after this task https://github.com/woocommerce/woocommerce-ios/issues/3151, and after adding all the missing events, they will be different since we need to pass different events inside the `ViewConfiguration`, different button titles, different ViewModel handling for updating the different values.

pmusolino:
Thank you @ctarda for your quick review! 🙇‍♂️ 💯 

> I didn't see any regressions with Grouped Products, but there seems to be a couple of things that might not be working as expected. The "Done" button in the Linked Products screen does not seem to work all the time.

This is right since the last task of the testing process is to make sure that the Linked Products screen shows the new info. 
The `Done` button is a thing that will be handled in a new PR 😄 I apologize for not making this clear right away.

Garance91540 (approved):
looking good!

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products List Selector/LinkedProductListSelectorDataSource.swift
ctarda: > Do you mean to add a struct to be passed to the init method which contains only the events that we want to track (in this case, just one)?

I was wondering if there is anything else (logic, or other configuration) that is shared between the view controllers that share this list selector. But it seems like it's only one VC (LinkedProductsListSelectorViewController) so it might not be worth complicating things more.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Linked Products/LinkedProductsViewController.swift
ctarda: Makes sense!

ctarda:
@pmusolino Sorry for holding the review. 😊 Knowing that the behaviour of the Done button is going to be implemented in a separate PR... :shipit:

ctarda (approved):
:shipit:

pmusolino:
Thank you both for your review 🙇‍♂️

PR #3187, opened 2020-11-23
Removed the Product Title in product screen navigation bar

### LINKED ISSUE #3178: Remove Product Title in product screen header
It's currently a bit hard to see the "Update" and "Publish" CTAs right now on the Product Screen. https://d.pr/i/f8xfcu

I think we need to remove the product title in the header so we can better see the CTAs

pmusolino:
Fixes #3178 

## Description
Since it's currently a bit hard to see the "Update" and "Publish" CTAs right now on the Product Screen, as suggested by @Garance91540 I removed the product title from the navigation bar.

#### Case 1
1. Try to create a new product.
2. The title of the screen should be empty, and it should be shown only the "Publish" CTA.

#### Case 2
1. Try to edit an existing product.
2. Try to edit the title of the product. The title of the screen should be empty, showing only the "Update" CTA.

## Screenshots
Screen Shot 2020-11-19 at 10 33 24 | update product name

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
@@ -724,15 +724,6 @@ private extension ProductFormViewController {
 // MARK: Navigation Bar Items
 //
 private extension ProductFormViewController {
-    func updateNavigationBarTitle(productName: String) {
-        navigationItem.title = productName
-        switch presentationStyle {
-        case .contained(let containerViewController):
-            containerViewController.navigationItem.title = productName
-        default:
-            break
-        }
-    }
 
     func updateNavigationBar(isUpdateEnabled: Bool) {
         var rightBarButtonItems = [UIBarButtonItem]()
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -383,7 +382,6 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
 private extension ProductFormViewController {
     func configureNavigationBar() {
         updateNavigationBar(isUpdateEnabled: false)
-        updateNavigationBarTitle(productName: product.name)
```
jaclync, line 386: nit: `updateNavigationBarTitle` function can also be deleted now that it's not used anymore?

jaclync (approved):
Thanks for taking this one Paolo! :shipit: 

Code: ✅ 
Testing: ✅

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
pmusolino: Good catch, removed here 540c20b893c98ab5a16643dceab1a6a1f95d1471 💯

Garance91540 (approved):
👏

pmusolino:
Thank you both for your quick review, I appreciate it! 🙇

Issue #2167, opened 2020-04-21
Shipping Labels M1: UI to display shipping labels purchased for an order


anitaa1990:
Landing on the Order Detail screen will now give you the option to view shipping label details for an order.

Prerequisite: https://github.com/woocommerce/woocommerce-ios/issues/2164
Dev plan: p91TBi-3Lo-p2

### Order details > shipping label card

- [x] Display shipping label cards for an order with partial UI https://github.com/woocommerce/woocommerce-ios/pull/3188
- [x] Show aggregated order items for a shipping label (the logic is quite complicated) https://github.com/woocommerce/woocommerce-ios/pull/3197
- [x] Sync missing variations for an order https://github.com/woocommerce/woocommerce-ios/pull/3278
- [x] Polish order item rows without a price/quantity https://github.com/woocommerce/woocommerce-ios/pull/3254
- [x] Show ellipsis menu in the card header if the shipping label hasn't been refunded https://github.com/woocommerce/woocommerce-ios/pull/3209
- [x] Add “Don’t know how to print from your phone?” CTA row (the screen is implemented in #2174) https://github.com/woocommerce/woocommerce-ios/pull/3230
- [x] Update design for tracking number row with two actions: https://github.com/woocommerce/woocommerce-ios/pull/3226
  - “Copy tracking number”: same as existing copy tracking number action
  - “Track shipment”: opens a webview with a URL based on the carrier and tracking number (this is implemented in https://github.com/woocommerce/woocommerce-ios/issues/2564) ([Android code](https://href.li/?https://github.com/woocommerce/woocommerce-android/blob/26590149ffa14773b49758e13a215e1e00168d75/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/orders/shippinglabels/ShipmentTrackingUrls.kt))
- [x] Hide Shipment tracking sections if an order has non-empty shipping labels https://github.com/woocommerce/woocommerce-ios/pull/3268
- [x] The Products section only shows products/variations that aren't in a non-refunded shipping label https://github.com/woocommerce/woocommerce-ios/pull/3268

### Polishes

- (Moved to https://github.com/woocommerce/woocommerce-ios/issues/3334) Hide separators of "Reprint Shipping Label" CTA and remove top spacing in the printing info row (requires refactoring of `TopLeftImageTableViewCell` to use Auto Layout)
[...]

PR #3188, opened 2020-11-23
Order Details: display shipping label cards with partial UI

jaclync:
Part of #2167 

## Changes

- In `OrderDetailsResultsControllers`, added a FRC of `StorageShippingLabel` for an order's shipping labels
- In `OrderDetailsViewController` and `OrderDetailsViewModel`, triggered syncing for shipping labels
- In `OrderDetailsDataSource`, added new sections for all shipping labels when reloading sections:
  - Added an enum `Section.Category` so that we can look up the first shipping label section for `IndexPath` mapping to a specific `ShippingLabel` because the number of shipping label sections is variable
  - If a shipping label is refunded, only the tracking number row and details row are shown
  - Otherwise, these rows are shown:
    - `shippingLabel.productNames.count` product rows (without any setup so you might see some reused cells),
    - "Reprint Shipping Label" button row
    - Tracking number row
    - Details row

## Testing

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/) and there is an order with at least one non-refunded shipping label and an order with at least one refunded shipping label

**Notes**: the product rows and tracking number row are to be implemented separately, you might see reused cells during your testing. All the UI subtasks in #2167.

### Refunded shipping label card

- Go to the orders tab
- Tap on an order with at least one refunded shipping label --> after syncing, a shipping label package card should be shown "Package #" with the refunded rows in the "Changes" section above (tracking number row and details row)

### Non-refunded shipping label card

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> after syncing, a shipping label package card should be shown "Package #" with the non-refunded rows in the "Changes" section above (`shippingLabel.productNames.count` product rows, "Reprint Shipping Label" button row, tracking number row, details row)

## Example screenshots

**Notes**: the **product rows** and **tracking number row** are to be implemented separately, please ignore these rows

Refunded shipping label card | Non-refunded shipping label card
-- | --
Simulator Screen Shot - iPhone 12 - 2020-11-23 at 18 29 43 | Simulator Screen Shot - iPhone 12 - 2020-11-23 at 18 15 13

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -79,6 +79,8 @@ final class OrderDetailsDataSource: NSObject {
         return resultsControllers.refunds
     }
 
+    private var shippingLabels: [ShippingLabel] = []
+
     /// Shipping Lines from an Order
     ///
     private var shippingLines: [ShippingLine] {
@@ -227,6 +229,8 @@ private extension OrderDetailsDataSource {
             configureShippingMethod(cell: cell)
         case let cell as WooBasicTableViewCell where row == .billingDetail:
             configureBillingDetail(cell: cell)
+        case let cell as WooBasicTableViewCell where row == .shippingLabelDetail:
+            configureShippingLabelDetail(cell: cell)
         case let cell as TopLeftImageTableViewCell where row == .shippingNotice:
             configureShippingNotice(cell: cell)
         case let cell as LeftImageTableViewCell where row == .addOrderNote:
@@ -245,12 +249,18 @@ private extension OrderDetailsDataSource {
             configureNetAmount(cell: cell)
         case let cell as ProductDetailsTableViewCell where row == .orderItem:
             configureOrderItem(cell: cell, at: indexPath)
+        case let cell as ProductDetailsTableViewCell where row == .shippingLabelProduct:
+            configureShippingLabelProduct(cell: cell, at: indexPath)
         case let cell as ProductDetailsTableViewCell where row == .aggregateOrderItem:
             configureAggregateOrderItem(cell: cell, at: indexPath)
-        case let cell as ButtonTableViewCell:
+        case let cell as ButtonTableViewCell where row == .fulfillButton:
             configureFulfillmentButton(cell: cell)
-        case let cell as OrderTrackingTableViewCell:
+        case let cell as ButtonTableViewCell where row == .shippingLabelReprintButton:
+            configureReprintShippingLabelButton(cell: cell)
+        case let cell as OrderTrackingTableViewCell where row == .tracking:
             configureTracking(cell: cell, at: indexPath)
+        case let cell as OrderTrackingTableViewCell where row == .shippingLabelTrackingNumber:
+            configureShippingLabelTrackingNumber(cell: cell, at: indexPath)
         case let cell as LeftImageTableViewCell where row == .trackingAdd:
             configureNewTracking(cell: cell)
         case let cell as SummaryTableViewCell:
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -609,7 +667,31 @@ extension OrderDetailsDataSource {
...
+            return Section(category: .refundedProducts, title: Title.refundedProducts, row: row)
+        }()
+
+        let shippingLabels: [Section] = {
+            let shippingLabels = resultsControllers.shippingLabels
+            guard shippingLabels.isNotEmpty else {
+                return []
+            }
+
+            self.shippingLabels = shippingLabels
```
Ecarrion: This assignment looks like an unwanted side-effect.
Is it not possible to use `resultsControllers.shippingLabels` directly?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -346,6 +346,19 @@ extension OrderDetailsViewModel {
...
     }
 
+    func syncShippingLabels(onCompletion: ((Error?) -> ())? = nil) {
+        let action = ShippingLabelAction.synchronizeShippingLabels(siteID: order.siteID, orderID: order.orderID) { result in
+            switch result {
+            case .failure(let error):
+                DDLogError("⛔️ Error synchronizing shipping labels: \(error)")
+                onCompletion?(error)
+            case .success:
+                onCompletion?(nil)
```
Ecarrion: Super nit: I normally see the `.success` case first and the `.failure` case later!

Ecarrion (approved):
Code: Everything looks good! I just left a small comment about an unwanted side-effect
Testing: I couldn't set shipping labels on my test-site without adding a credit card, is there a way to circumvent this?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: the reason I set this is so that the `shippingLabels` value is fixed until the next time `reloadSections` is called. because the number of shipping label cards and its product rows depend on the value, I personally felt it's safer to have a "fixed" value rather than a variable from FRC (even though it should trigger `reloadSections` on any change, but FRC is also not fully within our control). lemme know what you think!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
jaclync: I actually haven't noticed the common order! sounds good, updated in: https://github.com/woocommerce/woocommerce-ios/pull/3188/commits/f5fff94e8ed20edc38f2f6e4df4f7bb32f85c1f0

jaclync:
Thanks for the review Ernesto! I responded to your comments for another look 🙂 for creating a shipping label without using a real credit card, you can check out Anitaa's testing post p91TBi-3xD-p2 to set up your account and store (it's always good to double check the credit card under `WooCommerce > Settings > Shipping > WooCommerce Shipping & Tax` in core before purchasing a shipping label, I was charged on a personal card last time 😜 )

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
Ecarrion: I think that it's good but what I find strange is for the mutation to happen on `let shippingLabels: [Section] = {` creation. I wouldn't expect that variable initialization to mutate some bigger state of `self`.

Maybe something like:
```swift
self.shippingLabels = shippingLabels
let shippingLabels: [Section] = {
    // code
}
```
could express the intention better!

Ecarrion (approved):
Tested with the test creadit card and all seems to work as expected! 🚀

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: yea great point! I update this in https://github.com/woocommerce/woocommerce-ios/pull/3188/commits/f41907f2b667a0dcb1f476908cfa772d4a68a170

PR #3196, opened 2020-11-24
Add CoreDataManager Unit Tests

shiki:
Part of #2667. 

I have a [WIP branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/2667-reduce-file-operations-in-core-data) which reduces the `FileManager` operations in our Core Data stack. As we've seen observed with the effect of https://github.com/woocommerce/woocommerce-ios/pull/3069, removing the `FileManager` is effective in reducing the crashes. This is the from the [Sentry event](https://sentry.io/share/issue/d1497b3ad58e4cba8249599fd9efc0eb/):

But for now, I'm only submitting additional unit tests. AKA “Cover My Ass”. 😄 

## Changes

I added unit tests that prove that `CoreDataManager` will (or will not) automatically migrate the database. These are simple tests as more migration-specific tests will be added to [`CoreDataIterativeMigratotTests`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2667-add-core-data-unit-tests/Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift) later. 

## Testing

Please review the code and make sure the build is green. 🙂 

## Submitter Checklist

### DIFF
--- a/Storage/StorageTests/CoreData/CoreDataManagerTests.swift
+++ b/Storage/StorageTests/CoreData/CoreDataManagerTests.swift
@@ -119,15 +119,16 @@ final class CoreDataManagerTests: XCTestCase {
                                                    in: manager.viewStorage as! NSManagedObjectContext))
 
         // When
-        // Use a models inventory with an old model. this will cause a loading error and make the
-        // `CoreDataManager` recover and recreate the database.
-        let olderModelsInventory = ManagedObjectModelsInventory(
+        // Use a models inventory with an old and only one model. This will cause a loading error
+        // because it is not compatible. This will then make `CoreDataManager` recover and
+        // recreate the database.
+        let invalidModelsInventory = ManagedObjectModelsInventory(
             packageURL: modelsInventory.packageURL,
             currentModel: try XCTUnwrap(modelsInventory.model(for: .init(name: "Model 2"))),
             versions: [.init(name: "Model 2")]
         )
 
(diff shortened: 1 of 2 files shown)


Storage/StorageTests/CoreData/CoreDataManagerTests.swift
```diff
@@ -35,14 +36,14 @@ final class CoreDataManagerTests: XCTestCase {
     /// Verifies that the ContextManager's viewContext matches the PersistenContainer.viewContext
     ///
     func test_viewContext_property_returns_persistentContainer_main_context() {
-        let manager = CoreDataManager(name: "WooCommerce", crashLogger: MockCrashLogger())
+        let manager = CoreDataManager(name: storageIdentifier, crashLogger: MockCrashLogger())
```
ctarda, line 39: I was going to comment "would you consider extracting this to a private method" but, to be honest, it's just a line, I am not sure it adds any value, other than the semantic value of calling it `createDefaultManager()` or similar, so maybe it's not worth it.

Sorry, I' just thinking out loud here...

ctarda (approved):
Run the tests, and got a ✅ 

:shipit:

Storage/StorageTests/CoreData/CoreDataManagerTests.swift
shiki: I thought that too and we both ended up with the same conclusion. 😄

shiki (commented):
Thanks, Cesar!

PR #3197, opened 2020-11-24
Order Details: show aggregated order items for a shipping label

jaclync:
Part of #2167 

Depends on https://github.com/woocommerce/woocommerce-ios/pull/3188

This PR aggregates order items from an array of shipping labels, order items, products, and product variations from an order. The logic is a bit complicated due to the interesting API behavior, I tried adding comments along the way but feel free to ask any questions or request any explanations.

## Changes

- Created `ShippingLabelOrderItemsAggregator` that aggregates order items from an array of shipping labels, order items, products, and product variations from an order. The availability of products and product variations depends on whether they have been synced before. Note that syncing for product variations associated with an order is saved for a separate subtask of #2167.
- In `OrderDetailsResultsControllers`, added a FRC for `ProductVariation`s associated with an order.
- In `OrderDetailsDataSource`, set `ShippingLabelOrderItemsAggregator` when shipping labels are set in `reloadSections`. Updated the number of order item rows from `ShippingLabelOrderItemsAggregator`'s `orderItems` size.
- In `ProductDetailsCellViewModel`, added `productImageURL: URL?` so that the image URL is available in `AggregateOrderItem` without a concrete `Product` (it could also be `ProductVariation`).
- In `OrderDetailsViewModel`, handled tapping on a shipping label order item (product).
- Updated `AggregateOrderItem` from a `class` to `struct` so that we can use the auto-generated `Equatable` implementation.
- Added unit tests for `ShippingLabelOrderItemsAggregator`.

## Testing

**Notes**: if the store has not fetched product variations before order details, you might see the variation cell showing price 0.0. There are separate subtasks to fix this.

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> after syncing, a shipping label package card should be shown "Package #" with order items (products/variations) that are included in this package with the correct quantity/price (feel free to compare with core or Android)
  - Note that if the product/variation hasn't been synced, only the product name is shown.

## Example screenshots

Notes for @Garance91540:
- Please only focus on the order item rows (the other rows might not have the latest design yet)
- I plan to hide the 0 quantity/price labels in the product/variation cell if the information isn't available yet in a separate subtask. There is also another subtask to sync product variations associated with an order.

before variations are synced | after variations are synced (from opening the variations from a variable product in the Products tab)
-- | --
Simulator Screen Shot - iPhone 11 - 2020-11-24 at 14 30 42 | Simulator Screen Shot - iPhone 11 - 2020-11-24 at 14 30 57

Update release notes:


Garance91540 (approved):
👏

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
+import Yosemite
+
+/// Aggregates the order items from a shipping label.
+struct ShippingLabelOrderItemsAggregator {
```
shiki: ℹ️ Considering this is a complex logic, it'd be nice if we can explain what “aggregate” means. What are the inputs? What is the output? What are some examples? And why do we have to do this? I'm thinking we'd probably forget in a few months so it'd be nice if we can document it. 🙂

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -677,15 +705,19 @@ extension OrderDetailsDataSource {
             }
 
             self.shippingLabels = shippingLabels
+            self.shippingLabelOrderItemsAggregator = ShippingLabelOrderItemsAggregator(shippingLabels: shippingLabels,
```
shiki: ⚠️ I would like to respectfully disagree with the location of this. I also disagree with the existing `self.shippingLabels = shippingLabels` above. 😅 

This block is inside a variable creation function (`let shippingLabels: [Section] = {`) and I believe it's not a good practice to have a side-effect that **mutates self**. It's almost always a surprise for developers. 🙈 

--- 

Looking at how we're using `OrderDetailsDataSource` and `ResultsController`, I see that we don't really care much about the reactiveness of `ResultsController`. We're only using it as an overblown `SELECT * FROM` query. So, I think it's perfectly fine to move this outside of the variable declaration and possibly even at the top so we can separate the data-creation from the section-creation. 

```swift
    func reloadSections() {
        let summary = Section(category: .summary, row: .summary)

        self.shippingLabels = shippingLabels
        self.shippingLabelOrderItemsAggregator = ShippingLabelOrderItemsAggregator(shippingLabels: shippingLabels,
                                                                                   orderItems: items,
                                                                                   products: self.products,
                                                                                   productVariations: productVariations)

```

I'd be fine with this. `reloadSections()` is still very long but that's a different problem. 🤪

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -80,6 +86,7 @@ final class OrderDetailsDataSource: NSObject {
     }
 
     private var shippingLabels: [ShippingLabel] = []
+    private var shippingLabelOrderItemsAggregator: ShippingLabelOrderItemsAggregator?
```
shiki: ⚠️  I believe there's an opportunity for us to clarify that this is a data structure and not an object that does something. When I read the word "Aggregator", I assumed that there's a function that I can use. But there isn't, [they're all private](https://github.com/woocommerce/woocommerce-ios/blob/4b2532290798db2d5db06b446c2656934f03eec1/WooCommerce/Classes/ViewModels/Order%20Details/Shipping%20Labels/ShippingLabelOrderItemsAggregator.swift#L32).

What do you think about renaming this to `AggregatedShippingLabelOrderItems` instead?

```suggestion
    private var aggregatedShippingLabelOrderItems: AggregatedShippingLabelOrderItems?
```

I'd also like to suggest that we avoid nullability. I feel like a non-null aggregate would further signify that it is recomputed later on. 

```swift
struct AggregatedShippingLabelOrderItems {

    static let empty: AggregatedShippingLabelOrderItems = .init(shippingLabels: [], orderItems: [], products: [], productVariations: [])

```

```swift
    private var aggregatedShippingLabelOrderItems = AggregatedShippingLabelOrderItems.empty
```

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -67,6 +67,12 @@ final class OrderDetailsDataSource: NSObject {
         return resultsControllers.products
     }
 
+    /// ProductVariations from an Order
+    ///
+    var productVariations: [ProductVariation] {
+        return resultsControllers.productVariations
+    }
```
shiki: ℹ️  I feel like we can get away with not having this as publicly accessible `var`. Since it's only used once, I think it's fine to use it directly.

```swift
        self.aggregatedShippingLabelOrderItems = AggregatedShippingLabelOrderItems(shippingLabels: shippingLabels,
                                                                                   orderItems: items,
                                                                                   products: self.products,
                                                                                   productVariations: resultsControllers.productVariations)

```

That saves us a few lines in `OrderDetailsDataSource`. A few more lines that we don't need to understand. 🤪 This class is getting way too big. 😢

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -164,6 +171,18 @@ final class OrderDetailsDataSource: NSObject {
...
+    func shippingLabelOrderItem(at indexPath: IndexPath) -> AggregateOrderItem? {
+        guard let shippingLabel = shippingLabel(at: indexPath) else {
+            return nil
+        }
+
+        guard let orderItem = shippingLabelOrderItemsAggregator?.orderItems(of: shippingLabel)[safe: indexPath.row] else {
+            return nil
+        }
+
+        return orderItem
```
shiki: Kudos for using `safe`! 🎉 

ℹ️ It looks like we can shorten this. 🙂 

```suggestion
        return shippingLabelOrderItemsAggregator?.orderItems(of: shippingLabel)[safe: indexPath.row]
```

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -439,7 +458,16 @@ private extension OrderDetailsDataSource {
     private func configureShippingLabelProduct(cell: ProductDetailsTableViewCell, at indexPath: IndexPath) {
         cell.selectionStyle = .default
 
-        // TODO-2167: show aggregated order item (product) for a shipping label
+        guard let shippingLabel = shippingLabel(at: indexPath) else {
+            return
+        }
+
+        guard let orderItem = shippingLabelOrderItemsAggregator?.orderItems(of: shippingLabel)[safe: indexPath.row] else {
```
shiki: ℹ️ We can also have a helper that encapsulates `safe` inside it. 

```swift
    func orderItem(of shippingLabel: ShippingLabel, at index: Int) -> AggregateOrderItem? {
        guard let orderItems = orderItemsByShippingLabelID[shippingLabel.shippingLabelID] else {
            return nil
        }

        return orderItems[safe: index]
    }
```

```suggestion
        guard let orderItem = shippingLabelOrderItemsAggregator?.orderItems(of: shippingLabel at: indexPath.row) else {
```

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -439,7 +458,16 @@ private extension OrderDetailsDataSource {
...
+            return
+        }
+
+        let itemViewModel = ProductDetailsCellViewModel(aggregateItem: orderItem, currency: order.currency)
+        cell.configure(item: itemViewModel, imageService: imageService)
```
shiki, line 457: ❓ What happens if the cell is _reused_? Should we somehow clear the cell data in case the `guard` calls above fail?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
```diff
@@ -30,6 +30,15 @@ final class OrderDetailsResultsControllers {
         return ResultsController<StorageProduct>(storageManager: storageManager, matching: predicate, sortedBy: [descriptor])
     }()
 
+    /// ProductVariation ResultsController.
+    ///
+    private lazy var productVariationResultsController: ResultsController<StorageProductVariation> = {
+        let variationIDs = order.items.map { $0.variationID }.filter { $0 != 0 }
```
shiki: I was worried about this. I questioned myself “what if `order.items` changes?“. But I suppose it's fine because `self.order` is readonly. 🤷 

Nitpick! 😄 

```suggestion
        let variationIDs = order.items.map(\.variationID).filter { $0 != 0 }
```

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
```diff
@@ -147,6 +163,22 @@ private extension OrderDetailsResultsControllers {
...
+
+        productVariationResultsController.onDidResetContent = { [weak self] in
+            guard let self = self else {
+                return
+            }
+            self.refetchAllResultsControllers()
+            onReload()
+        }
+
+        try? productVariationResultsController.performFetch()
```
shiki: We should log the failure. 🙂 But I understand if you don't want to at the moment. It's weird if it's the only one in this class.

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
...
+        // Here we map a ShippingLabel's `productNames` to `ProductInformation` with an optional product ID at the corresponding index.
+        let productInfoArray = shippingLabel.productNames.enumerated().map { index, productName in
+            ProductInformation(id: shippingLabel.productIDs[safe: index], name: productName)
+        }
+        // In the API, if a product has quantity higher than 1, it is repeated in the `productNames`/`productIDs` array.
+        // Note that decimal quantity might have unexpected behavior given the current API design.
+        let uniqueProductInfoArray = productInfoArray.removingDuplicates()
+
+        // Generates a dictionary that maps a `ProductInformation` to the number of times (quantity) in a shipping label.
+        let countsByProductInformation = productInfoArray.reduce([:]) { (result, productInfo) -> [ProductInformation: Int] in
```
shiki: ℹ️  I suppose the reason why we don't use `uniqueProductInfoArray` in here is that we want to count the repeated products. What do you think about moving the `uniqueProductInfoArray` declaration after this block?. 🙂

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
...
+                                                products: [Product],
+                                                productVariations: [ProductVariation]) {
+        // ShippingLabel's `productNames` is always available, but `productIDs` is only available in WooCommerce Shipping & Tax v1.24.1+.
+        // Here we map a ShippingLabel's `productNames` to `ProductInformation` with an optional product ID at the corresponding index.
+        let productInfoArray = shippingLabel.productNames.enumerated().map { index, productName in
+            ProductInformation(id: shippingLabel.productIDs[safe: index], name: productName)
+        }
+        // In the API, if a product has quantity higher than 1, it is repeated in the `productNames`/`productIDs` array.
+        // Note that decimal quantity might have unexpected behavior given the current API design.
+        let uniqueProductInfoArray = productInfoArray.removingDuplicates()
```
shiki: ℹ️ Maybe using `Set` is better so we don't rely on our own implementation. 😄 Hopefully, the Swift developers used a better algorithm. 

```suggestion
        let uniqueProductInfoArray = Array(Set(productInfoArray))
```

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
...
+
+private extension ShippingLabelOrderItemsAggregator {
+    mutating func aggregateProductsToOrderItems(shippingLabels: [ShippingLabel],
+                                                orderItems: [OrderItem],
+                                                products: [Product],
+                                                productVariations: [ProductVariation]) {
+        shippingLabels.forEach { aggregateProductsToOrderItems(shippingLabel: $0,
+                                                               orderItems: orderItems,
+                                                               products: products,
+                                                               productVariations: productVariations) }
```
shiki: ⚠️ How do you feel about reducing the number of `mutating` functions by:

Changing line 43 to return an array of `AggregateOrderItem` instead of mutating a variable.

```swift
    func aggregateProductsToOrderItems(shippingLabel: ShippingLabel,
                                       orderItems: [OrderItem],
                                       products: [Product],
                                       productVariations: [ProductVariation]) -> [AggregateOrderItem] {
        /// LOGIC HERE
        return aggregateOrderItems
    }
```

And then changing this to:

```suggestion
        orderItemsByShippingLabelID = shippingLabels.reduce(into: [Int64: [AggregateOrderItem]]()) { result, shippingLabel in
            result[shippingLabel.shippingLabelID] = aggregateProductsToOrderItems(shippingLabel: shippingLabel,
                                                                       orderItems: orderItems,
                                                                       products: products,
                                                                       productVariations: productVariations)
        }
```

With this, we will only have one `mutating` func and it's probably easier to understand. 🙂

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
...
+        // In the API, if a product has quantity higher than 1, it is repeated in the `productNames`/`productIDs` array.
+        // Note that decimal quantity might have unexpected behavior given the current API design.
+        let uniqueProductInfoArray = productInfoArray.removingDuplicates()
+
+        // Generates a dictionary that maps a `ProductInformation` to the number of times (quantity) in a shipping label.
+        let countsByProductInformation = productInfoArray.reduce([:]) { (result, productInfo) -> [ProductInformation: Int] in
+            var countsByProductInfo = result
+            countsByProductInfo[productInfo] = (result[productInfo] ?? 0) + 1
+            return countsByProductInfo
+        }
```
shiki: ℹ️ `reduce(into:)` is more succinct and less awkward. 🙂 

```suggestion
        let countsByProductInformation = productInfoArray.reduce(into: [ProductInformation: Int]()) { result, productInfo in
            result[productInfo] = (result[productInfo] ?? 0) + 1
        }
```

But up to you!

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
...
+                                                orderItems: [OrderItem],
+                                                products: [Product],
+                                                productVariations: [ProductVariation]) {
+        shippingLabels.forEach { aggregateProductsToOrderItems(shippingLabel: $0,
+                                                               orderItems: orderItems,
+                                                               products: products,
+                                                               productVariations: productVariations) }
+    }
+
+    mutating func aggregateProductsToOrderItems(shippingLabel: ShippingLabel,
```
shiki: Great separation here! I found it easy to read, especially with the comments. 🙂

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
...
+                         quantity: Decimal(quantity),
+                         sku: orderItem?.sku ?? variation.sku,
+                         total: totalPrice,
+                         imageURL: URL(string: variation.image?.src ?? ""),
+                         attributes: orderItem?.attributes ?? [])
+        }
+    }
+
+    func lookUpProduct(by productID: Int64, products: [Product]) -> Product? {
+        products.filter({ $0.productID == productID }).first
```
shiki: ℹ️ There is also `first()`. 🙂 

```suggestion
        products.first(where: { $0.productID == productID })
```

This makes one question whether we need the function. 🤪

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
```diff
@@ -0,0 +1,153 @@
...
+            return .product(product: product, orderItem: orderItem, name: productInfo.name)
+        } else if let productVariation = lookUpProductVariation(by: productID, productVariations: productVariations) {
+            let orderItem = orderItems.first(where: { $0.variationID == productID })
+            return .productVariation(productVariation: productVariation, orderItem: orderItem, name: productInfo.name)
+        } else {
+            return .productName(name: productInfo.name)
+        }
+    }
+
+    func orderItem(from model: OrderItemModel, quantity: Int, orderItems: [OrderItem]) -> AggregateOrderItem {
```
shiki: ⚠️  The `orderItems` is not used. 🙂

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
```diff
@@ -21,6 +21,8 @@ struct ProductDetailsCellViewModel {
     ///
     private let product: Product?
 
+    private let productImageURL: URL?
+
```
shiki: ⚠️ I feel like there should be separate view model types at this point. But that's quite a refactoring. 

Would you consider renaming this to `preferredImageURL`? And maybe a comment about how it is used? 

Or I wonder if we should make `var imageURL` be intentionally initialized during `init` instead of it being a computed property. In this way, we only have one image URL property.

WooCommerce/WooCommerceTests/Tools/MockShippingLabel.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+import Yosemite
+
+/// Generates mock `ShippingLabel`
+///
+public struct MockShippingLabel {
+    /// Consider setting a subset of properties using `.copy`.
```
shiki, line 7: ❤️

WooCommerce/WooCommerceTests/Tools/MockShippingLabelAddress.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+import Yosemite
+
+/// Generates mock `ShippingLabelAddress`
+///
+public struct MockShippingLabelAddress {
+    public static func sampleAddress(company: String = "",
```
shiki, line 7: ℹ️  Perhaps this can be `empty()` too and we'll just use `copy()`.  🙂

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregatorTests.swift
```diff
@@ -0,0 +1,135 @@
+import XCTest
```
shiki: I see there's no more `Foundation` bwahahahahahahahaha

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregatorTests.swift
```diff
@@ -0,0 +1,135 @@
...
+@testable import WooCommerce
+import Yosemite
+
+final class ShippingLabelOrderItemsAggregatorTests: XCTestCase {
+    private let sampleSiteID: Int64 = 124
+    private let sampleOrderID: Int64 = 630
```
shiki: ⚠️ These are unused. 🙂 

```suggestion

```

shiki (changes requested):
Heya, @jaclync! I have a mix of strong opinions (⚠️ ) and suggestions (ℹ️ ). 😄 

---

Do you know why my shipping label is showing up as "Label"?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: Totally true! I didn't notice that this was inside a variable creation block in the first place in https://github.com/woocommerce/woocommerce-ios/pull/3188, Ernesto pointed this out and I updated it there. I'll move these two toward the top of the function.

Ideally the data assignment can be separate from `reloadSections`, and that we can be sure the data don't change between two `reloadSections` calls. I could also just get the shipping labels directly from FRC like how other order details sections, but I just didn't feel comfortable getting the data from the FRC every time in case the value changes without triggering `reloadSections`. Lemme know if you think this manual assignment of shipping labels from FRC doesn't seem necessary! (I tend to over-worry about things 😅)

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
jaclync: There's actually a potential difference in the output using `Array(Set(array))` vs. `array.removingDuplicates()` - the former could output the unique elements in any order from the conversion to `Set`, while the latter keeps the original order and removes duplicates after the first instance. For a shipping label's order items, I think we'd like to keep the order as in the API response (product IDs/names). Feel free to correct anything!

WooCommerce/WooCommerceTests/Tools/MockShippingLabelAddress.swift
jaclync: That'd be nice, though I haven't implemented `GeneratedCopiable` for `ShippingLabelAddress` yet - will do that the next time I update the tests since this PR is already big!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: Moved these two lines in: https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/67190d5cb2a9810c894071a14bfce354699dc5c5

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: makes sense! updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/67190d5cb2a9810c894071a14bfce354699dc5c5

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
jaclync: yup! added more comments in: https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/73b681404e089516b55cdd97ac1ab77d719f284d

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: `AggregatedShippingLabelOrderItems` sounds much more accurate for what this struct is! I'm still working on naming 😅 and the change to non-optional sounds good. updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/73b681404e089516b55cdd97ac1ab77d719f284d

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: It's nice optimizing the LOC in an already big class 😄 updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/7dbe1f3ebf33f3e571285748bc76007225b1618f

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
jaclync: nice! updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/7dbe1f3ebf33f3e571285748bc76007225b1618f

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsResultsControllers.swift
jaclync: good idea to log the error in case of failure even though other functions don't do this, added in: https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/7dbe1f3ebf33f3e571285748bc76007225b1618f

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
jaclync: yup makes sense, updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/7dbe1f3ebf33f3e571285748bc76007225b1618f

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
jaclync: hah that's true, I'll still keep them in a function for now. updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/7dbe1f3ebf33f3e571285748bc76007225b1618f

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
jaclync: great catch! removed in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/7dbe1f3ebf33f3e571285748bc76007225b1618f

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregatorTests.swift
jaclync: good catch! removed in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/b4cb7059e6f9f969691b3b8794c3229ae8ad32e8

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: yea sounds nice to encapsulate `safe` in a function, updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/c85b4c7e7b360c08c6038bad1dadacf1201f8f03

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: The guard statement's failure case is not expected, since it means that there is a miscalculation in the shipping label and/or order item index. I added an `assertionFailure` here: https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/71b2fe4f2d3bce38f6df320a84cd649156abf9c4

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
jaclync: oooh nice one! updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/497b99f22a680ee517ce1968364fa35301bff15d

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
jaclync: yea, initializing `imageURL` in each initializer makes the most sense to me. `Product?` parameter in its initializers is actually only used for `imageURL`, I removed the private `Product?` here as well: https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/b4315a2b690dbdf38dd9f3232484e3a2c86ce361

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
jaclync: wow TIL'ed `reduce(into:)`, this looks much more concise! updated in https://github.com/woocommerce/woocommerce-ios/pull/3197/commits/7dbe1f3ebf33f3e571285748bc76007225b1618f

jaclync:
@shiki thanks a lot for the thorough review! really helpful feedback and suggestions 🙇🏻‍♀️ I responded to all of your comments with a couple of updates to the PR for another look! also updated the PR milestone 5.7 and I'll address any comments next week after AFKs.

> Do you know why my shipping label is showing up as "Label"?

Yes! this is because I haven't implemented the latest design for the tracking number row (a subtask of https://github.com/woocommerce/woocommerce-ios/issues/2167) so it looks quite broken when it's reused 😆 working on this next!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
shiki: > but I just didn't feel comfortable getting the data from the FRC every time in case the value changes without triggering reloadSections.  

I agree. I think this actually means we need to rethink the `DataSource` and `ViewModel`. We can make that reactive in the future. 🙂

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelOrderItemsAggregator.swift
shiki: You're right! I stand corrected. 🤸

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
shiki: Love it!

shiki (approved):
:shipit: 🤸

jaclync:
Thank you for the final review 💟 merging now!

PR #3202, opened 2020-11-24
Fix disabled state for buttons

### LINKED ISSUE #3081: Disabled state on buttons not working

Ecarrion:
fixes #3081 

# Why 

While working on the issue refunds project I noticed that the disabled states for our app buttons were wrong.
previous

This happens because the background-image for the disabled state was a clear color hence showing the default view' background color(pink)

# How

The issue is fixed by not assigning a general background color anymore, but only assigning default images to the different button states (normal, disabled, highlighted)

# Screenshots

Light Disabled | Light Enabled
---- | ----
disabled-light | enabled-light

Dark Disabled | Dark Enabled
---- | ----
disabled-dark | enabled-dark

# Testing Steps
- Go to an un-refunded order
- Tap on "Issue Refund" button
- See the "next" button with the correct disabled state
- Select some items
- See the "next" button with the correct enabled state

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIButton+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UIButton+Helpers.swift
@@ -8,7 +8,6 @@ extension UIButton {
     /// Applies the Primary Button Style: Solid BG!
     ///
     func applyPrimaryButtonStyle() {
-        backgroundColor = .primaryButtonBackground
         contentEdgeInsets = Style.defaultEdgeInsets
         layer.borderColor = UIColor.primaryButtonBorder.cgColor
         layer.borderWidth = Style.defaultBorderWidth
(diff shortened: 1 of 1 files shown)


jaclync (approved):
Nice update! 🎨

Code: ✅ 
Testing: ✅ (I felt like the disabled state in dark mode could have more contrast, up to @Garance91540!)

Garance91540 (approved):
Looking good! 👏👏👏

PR #3206, opened 2020-11-24
Remove Exclusion of Future Orders

### LINKED ISSUE #2202: Orders: Predicate is not refreshed
In All Tabs, the `resultsController` `predicate` should be refreshed on `viewWillAppear`

This bug happens in All Tabs.

1. Run the app with a one-month-old date.
2. Navigate to All Orders
3. Set the device date back to today
4. Pull-to-refresh to All Orders. 

All Orders will be empty or no **recent orders** will be shown. The list will still show orders that were created more than a month ago.

## Possible Solution.

Maybe the predicate can be refreshed on the next query so that the date in the query condition is updated too.

### LINKED ISSUE #2805: [DiffableDataSource] Copy Legacy OrdersVM Tests to OrderListVM
Part of the fix for #1543.

### LINKED ISSUE #3203: All Orders tab should show future-dated orders
As mentioned in a comment on [this ticket](https://github.com/woocommerce/woocommerce-android/issues/2252#issuecomment-627536180):

> Garance mentioned in p5T066-1i2-p2#comment-4879 that she agrees that we should show future orders in the All Orders tab. So this issue might be affected by that.

The all orders tab should be updated to now show future-dated orders.

--- 

Andoid issue: https://github.com/woocommerce/woocommerce-android/issues/2412

shiki:
Closes #3203.  
Fixes #2202. 
Closes #2805. 

Before | After 
--------|-------
     |       

## Findings

We currently don't show orders that are dated in the future in the All Orders. A few months ago, we decided to show all the orders, regardless of what their `dateCreated` value is. This was done on Android (https://github.com/woocommerce/woocommerce-android/pull/2413) but I totally forgot to do the same on iOS. This fixes my mistake. 

Removing the exclusion of future orders also fixes #2202. This actually makes me wonder whether the `predicates` with the date constraint is why we're getting reports from users saying that their orders are not loading. 

https://github.com/woocommerce/woocommerce-ios/blob/1c94af3a701bd16bcbadd142f08acd50cd421666/WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift#L140-L145

We never update the date argument in this constraint so if the app is never terminated, it's possible that the constraint will stay on the same date. For example:

Scenario | Date | Predicate | Result
--------|-------|---|---
User installs and opens the app.        |    2020-11-20  | `dateCreated < 2020-11-21` | All good and all orders (except future orders) should be loaded
User comes back to the app 2 days later. The app was never terminated by iOS. | 2020-11-22 | `dateCreated < 2020-11-21` | Orders submitted on Nov 21 and 22nd are not shown. 

🤦 

## Changes

This removes the exclusion of future orders, which also fixes the bug I mentioned above. 

In addition to that, I copied the [`OrdersViewModelTests`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2202-fix-orders-predicate/WooCommerce/Classes/ViewRelated/Orders/OrdersViewModel.swift) to its newer iOS-13 only version, [`OrderListViewModelTests`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2202-fix-orders-predicate/WooCommerce/WooCommerceTests/ViewRelated/Orders/OrderListViewModelTests.swift). This should allow us to safely remove the deprecated `OrdersViewModel` when we drop iOS 12.0 support. 

## Testing

1. Submit a new order on the web.
2. Set the device's date back one month.
3. Run the app on the device.
4. Navigate to All Orders. 
5. Confirm that you still see the order you just submitted. It should be under the “Upcoming” section.
6. Set the device date back to today.
7. Pull-to-refresh to All Orders. 
8. Confirm that all the orders are still shown. But the order you submitted should be in the “Today” section. 
9. Navigate away from the All Orders tab. Maybe go to Processing.
10. Submit another order on the web. 
11. Navigate back to the All Orders tab. 
12. Don't do a pull-to-refresh. Wait a while and then confirm that the last order you submitted is shown. 

## Submitter Checklist

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,7 @@
 
 5.6
 -----
+- [**] Fixed order list sometimes not showing newly submitted orders. 
 - [*] now the date pickers on iOS 14 are opened as modal view. [https://github.com/woocommerce/woocommerce-ios/pull/3148]
 - [*] now it's possible to remove an image from a Product Variation if the WC version 4.7+. [https://github.com/woocommerce/woocommerce-ios/pull/3159]
 - [*] removed the Product Title in product screen navigation bar. [https://github.com/woocommerce/woocommerce-ios/pull/3187]
--- a/WooCommerce/Classes/ViewRelated/Orders/OrderListSyncActionUseCase.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrderListSyncActionUseCase.swift
@@ -83,10 +83,6 @@ struct OrderListSyncActionUseCase {
     let siteID: Int64
     /// The current filter mode of the Order List.
     let statusFilter: OrderStatus?
-    /// If true, orders created after today's day will be included in the result.
-    ///
-    /// This will generally only be false for the All Orders tab. All other screens should show orders in the future.
-    let includesFutureOrders: Bool
 
     /// Returns the action to use when synchronizing.
     func actionFor(pageNumber: Int,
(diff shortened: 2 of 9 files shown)


ctarda (approved):
Following the testing instructions: 
* Submit an order on web
* Change the device's date to the past
* Display orders:



Update the device date to today:



And submit a new order:



:shipit:

Issue #2168, opened 2020-04-21
Shipping Labels M1: UI to request refund a shipping label


anitaa1990:
If shipping labels are available for an order, we should provide an option to refund an already purchased shipping label. Note that [this networking layer issue](https://github.com/woocommerce/woocommerce-ios/issues/2165) should be addressed first.

<img width="645" alt="Screenshot 2020-04-21 at 9 00 30 PM" src="https://user-images.githubusercontent.com/22608780/79885716-0ae23d00-8415-11ea-9734-c2dc8606ea2b.png">

PR #3209, opened 2020-11-25
Order Details > Shipping Label: add CTA to section header that opens an action sheet for refund

jaclync:
Part of #2167 

## Changes

- Updated `PrimarySectionHeaderView`'s xib from one label to a `UIStackView` with a label and button. In `PrimarySectionHeaderView`, added a struct `ActionConfiguration` that is optional in the view's configuration.
- In `OrderDetailsDataSource`, added a new case `Section.HeaderStyle.actionablePrimary` with a `PrimarySectionHeaderView.ActionConfiguration` associated value for configuring the action. In each shipping label section setup, set the header style to `actionablePrimary` if the shipping label hasn't been refunded. The action handler is propagated from `OrderDetailsViewController` to `OrderDetailsViewModel` following the existing pattern. I first tried reusing `onCellAction` but it does not provide a source view and the action is not from a cell either.

## Testing

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> after syncing, a shipping label package card should be shown "Package #" with an ellipsis CTA at the right side of the header
- Tap on the ellipsis CTA --> an action sheet should be opened with one action to request a refund (to be implemented in #2168)

## Example screenshots

Note for reviewers: please focus on the non-refunded label's section header where an ellipsis more menu is added with "Request a Refund" action

\ | order details | refund request action sheet
-- | -- | --
dark | Simulator Screen Shot - iPhone 11 - 2020-11-25 at 12 03 01 | Simulator Screen Shot - iPhone 11 - 2020-11-25 at 12 03 03
light | Simulator Screen Shot - iPhone 11 - 2020-11-25 at 12 15 02 | Simulator Screen Shot - iPhone 11 - 2020-11-25 at 12 15 05
iPad | Simulator Screen Shot - iPad Air (3rd generation) - 2020-11-25 at 12 18 49 | Simulator Screen Shot - iPad Air (3rd generation) - 2020-11-25 at 12 00 24

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -45,6 +45,10 @@ final class OrderDetailsDataSource: NSObject {
     ///
     var onCellAction: ((CellActionType, IndexPath?) -> Void)?
 
+    /// Closure to be executed when the shipping label more menu is tapped.
+    ///
+    var onShippingLabelMoreMenuTapped: ((_ shippingLabel: ShippingLabel, _ sourceView: UIView) -> Void)?
+
     /// Closure to be executed when the UI needs to be reloaded.
     ///
     var onUIReloadRequired: (() -> Void)?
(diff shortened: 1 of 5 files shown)


Garance91540 (approved):
👏

Ecarrion (approved):
Code and test look
 good! 🚀

Ecarrion:
Screen Shot 2020-11-25 at 10 20 24 AM

I was confused about why that "more" button was visible if the code was hiding it but then I realized it's the reused cell within.

jaclync:
> I was confused about why that "more" button was visible if the code was hiding it but then I realized it's the reused cell within.

Sorry about the confusion! The reuse behavior in the cells that haven't been worked in has been quite confusing in previous code reviews as well - will update them soon!

PR #3216, opened 2020-11-26
Dropped iOS 12 support

pmusolino:
As proposed here p91TBi-3d8 and as discussed [here](https://a8c.slack.com/archives/CGPNUU63E/p1606069557212300), we decided to drop iOS 12 since the adoption is less than 3% right now.
This PR will be merged in the next release 5.7, after the Monday freeze of release 5.6.

**From now we will support iOS 13 and later.**
I have also solved the warnings that appeared after dropping iOS 12 support.

## Changes
- updated all the targets to iOS 13.
- since `keyWindow` was deprecated and we can’t use anymore `UIApplication.shared.keyWindow`, I implemented an extension that contains the `currentKeyWindow` property to get the `keyWindow` in iOS 13.
- updated `UIActivityIndicator` using the new conventions for iOS 13 (eg. `.gray` became `.medium`).
- updated `scanQuotesRanges` method in the `Scanner` extension which was using deprecated methods.
- updated `statusBarOrientation`, deprecated from iOS 13.
- the `scrollIndicatorInsets` getter is deprecated. Now we use the `verticalScrollIndicatorInsets` getters instead.

## Testing
Before starting the testing process, run `bundle exec pod install`.
It would be good to do a general test of the app, just to make sure we didn't introduce any regression. In particular, looking at the things that have been updated, it would be good to test in detail:
- push notifications: the screens are presented correctly?
- activity indicators appearance.
- barcode scanner in Products (plus try to rotate the screen here).
- the keyboard on all the screens should work like before.
- CI should pass.

Update release notes:


WooCommerce/Classes/Extensions/UIApplication+Woo.swift
```diff
@@ -1,12 +1,22 @@
...
 import UIKit
 
+// MARK: - UIApplication Utils
+//
+extension UIApplication {
+
+    /// Returns the keyWindow. Accessing `UIApplication.shared.keyWindow` is deprecated from iOS 13.
+    ///
+    var currentKeyWindow: UIWindow? {
+        UIApplication.shared.windows.filter {$0.isKeyWindow}.first
```
ctarda, line 11: Sorry if this is obvious, but I don't have enough context. 

My understanding is that the reason for deprecation was that the concept of `keyWindow` does not make sense for an iOS application anymore. 

I know it might be a bit overkill now, as it seems like we do not support multiple window, but would it make sense to take the potential scenes into account, filtering the active scene by its activationState and then grabbing its first key window?

ctarda (approved):
Great work @pmusolino . These migrations are never simple...

I did some smoke testing, and nothing seems to be broken. I left one question about the keyWindow, I'm curious about your opinion.

:shipit:

ctarda:
Oh, and this means SwiftUI, doesn't it!? 🥳

WooCommerce/Classes/Extensions/UIApplication+Woo.swift
pmusolino: Exactly. The deprecation message says: `Should not be used for applications that support multiple scenes as it returns a key window across all connected scenes`.


> I know it might be a bit overkill now, as it seems like we do not support multiple window, but would it make sense to take the potential scenes into account, filtering the active scene by its activationState and then grabbing its first key window?

If I understand correct, do you mean something like this?
```
let keyWindow = UIApplication.shared.connectedScenes
        .filter({$0.activationState == .foregroundActive})
        .map({$0 as? UIWindowScene})
        .compactMap({$0})
        .first?.windows
        .filter({$0.isKeyWindow}).first
```

pmusolino:
Thank you for the review @ctarda :D and yes, this means that we can start to introduce small SwiftUI components 🕺

WooCommerce/Classes/Extensions/UIApplication+Woo.swift
ctarda: Yes? I mean, if we take the scene into account, that looks like the way to go, but I am not sure if we need to complicate things that much...

pmusolino: So, I'd say to maintain the current behavior since we are sure it's still working like before. I'm not sure if in the future, if we will ship a mac OS app, we should handle this in a different way. :)

Issue #2844, opened 2020-09-20
[Issue Refunds] Create New Refund coordinator


Ecarrion:
- Add `CoordinatingController` to handle a "issue refund" navigation flow
- Make sure its tested

PR #3217, opened 2020-11-26
[Issue Refund] Adds Coordinating Controller

Ecarrion:
part of #2844 

# Why

This PR does not add any new functionality. It pays some tech debt by adding a coordinating controller to handle all presentation/navigation related to the issue refund feature. The main goal of this change is:

- Abstract the navigation/presentation duties from the view controllers.
- Make that navigation/presentation testable.

* Tests will be added in the next PR to maintain this one in a manageable size.

# How
- Added `IssueRefundCoordinatingController` that orquestrates `IssueRefundViewController`, `RefundConfirmattionViewController`, `ListSelectorViewController`, `UIAlertViewController`, and  `InProgressViewController` navigation and presentation.

- Refactor `IssueRefundViewController` and `RefundConfirmattionViewController` to expose actions that will be handled by `IssueRefundCoordinatingController`.

- Modify `RefundItemQuantityListSelectorCommand` to include the item index to be used when the view controller returns from the quantity selection.

# GIFS

## Success
sucess refund

## Failure
refund-error

# Testing steps

- Smoke test creating refunds, with and without internet connection(to see the error notice)

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
```diff
@@ -17,6 +17,14 @@ final class IssueRefundViewController: UIViewController {
 
     private let viewModel: IssueRefundViewModel
 
+    /// Closure invoked when the next button is tapped
+    ///
+    var onNextAction: ((RefundConfirmationViewModel) -> Void)?
```
ctarda, line 22: I am aware we are already using this pattern is another few view controllers, but I am not sure having view controllers expose a closure directly is the best way to decouple UI from navigation. 

I would argue that it couples things (view controller, view model and coordinator) even more. In this case in particular, the coordinator would need to be aware of the existence of RefundConfirmationViewController. Which might be necessary to instantiate the next step in the navigation sequence, but still, I would argue should not be knowledge of a navigation controller.

Would it make sense to encapsulate these closures in an object, and pass an instance of that object instead? Sort of a lightweight implementation of the command pattern, where what we pass to the view controller is the implementation of an interface with just one method (say, `execute()`). 

Or probably better, we could have a different "NavigationCommands" object per view controller, that would pack all navigation out of a given view controller, and notify clients via a delegate.

That way we would completely decouple view controllers from coordinators. 

On the other hand, that would mean that we would be adding a significant amount of very granular objects.

But at the same time, each of everyone of those objects would be testable as well, so we would be able to test:
* that the view controller meets expectations when navigation buttons are tapped
* that the entities that communicate with the navigation controller meet expectations.

On the other hand, I would argue that a simplest solution, that does not involve coordinators would be making viewmodels responsible for triggering navigation. In a way, a view controller should be already calling into a view model when there s user interaction in the UI...

What do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
```diff
@@ -65,7 +82,7 @@ private extension IssueRefundViewController {
 private extension IssueRefundViewController {
     @IBAction func nextButtonWasPressed(_ sender: Any) {
         let confirmationViewModel = viewModel.createRefundConfirmationViewModel()
```
ctarda, line 84: This is, in my opinion, a good example of a place where navigation could (and in my opinion probably should) be triggered by the view model. 

At this point, we request a view model for the destination view controller from the current view model, and then pass that view model to a different entity that is expecting it. So, in a way, we are leaking the current view model's abstraction.

I am curious about your opinion. Would it make sense to let IssueRefundViewModel be responsible from creating the confirmation view model (as it is now), and since it already knows that the refund conformationviewmodel is necessary, take also responsibility for navigating to the correct UI? After all, it could be argued that navigation logic is part of the logic that belongs in a view model

ctarda (commented):
Thanks for attacking this @Ecarrion 

The code looks great! I left a couple of comments, I am very curious about you opinion. Maybe there is potential for a wider discussion here?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
Ecarrion: > I would argue that it couples things (view controller, view model and coordinator) even more. In this case in particular, the coordinator would need to be aware of the existence of RefundConfirmationViewController. Which might be necessary to instantiate the next step in the navigation sequence, but still, I would argue should not be knowledge of a navigation controller.

Maybe we understand the coordinator pattern differently but I would say that the `Coordinator` does need to know about what view controller is coming next, since the coordinator's job is to define what VC comes next in the flow.

> Would it make sense to encapsulate these closures in an object, and pass an instance of that object instead? Sort of a lightweight implementation of the command pattern, where what we pass to the view controller is the implementation of an interface with just one method (say, execute()).

I'm not too sure, that object would have to be created by the `ViewController` or the `ViewModel` right? If that is the case then the `ViewController` couldn't really be reused in a different flow without adding conditionals all over the place. I think that defeats the main purpose of the `Coordinator`.

Ecarrion: FYI you can see here(p91TBi-36A-p2) the discussion and rationale when this pattern was adopted.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
Ecarrion: > This is, in my opinion, a good example of a place where navigation could (and in my opinion probably should) be triggered by the view model.

> At this point, we request a view model for the destination view controller from the current view model, and then pass that view model to a different entity that is expecting it. So, in a way, we are leaking the current view model's abstraction.

I agree with you that this isn't ideal, it is a compromise with how `RefundConfirmationViewModel` and `RefundConfirmationViewController` were created. 

> I am curious about your opinion. Would it make sense to let IssueRefundViewModel be responsible from creating the confirmation view model (as it is now), and since it already knows that the refund conformationviewmodel is necessary, take also responsibility for navigating to the correct UI? After all, it could be argued that navigation logic is part of the logic that belongs in a view model

I don't think that the `ViewController` or the `ViewModel` should be in charge of handling navigating in this specific case, it will couple them with a specific flow which is what we try to counteract with the `Coordinator` pattern.

My ideal scenario would be:

- `ViewControllerB` exposes dependencies through the `init` and uses those dependencies to create it's `ViewModelB`.

- `ViewControllerA` exposes delegate with the information that it can provide
```swift
protocol ViewControllerANavigationDelegate {
    func viewController(_ vc: ViewControllerA, didTapNextButon button: UIButton)
}
```

- The coordinator then creates `ViewControllerB` with the info that it can query from `ViewControllerA`
```swift
    func viewController(_ vc: ViewControllerA, didTapNextButon button: UIButton) {
         let dependencies = createDependencies(from: vc.state)
         let newVC = ViewControllerB(dependencies: dependencies)
         show(newVC, animated: true)
    } 
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
ctarda: The way I see it, the coordinator pattern is just the [Application Controller](https://martinfowler.com/eaaCatalog/applicationController.html) pattern, as defined in [Patterns of Enterprise Application Architecture](https://www.amazon.com/Patterns-Enterprise-Application-Architecture-Martin/dp/0321127420)

The spirit of the pattern is `A centralized point for handling screen navigation and the flow of an application.`, which is what this PR implements. That's a 💯

Yes, the Application Controller is responsible for instantiating and navigating to the next step in the sequence, encapsulating the logic to execute those actions.

What I don't understand is:
> If that is the case then the ViewController couldn't really be reused in a different flow without adding conditionals all over the place. I think that defeats the main purpose of the Coordinator.

Why would we want to reuse the view controller in a different flow?  I am genuinely curious. Assuming we wanted to reuse it in a different flow, decoupling the view controller from the application controller as much as possible seems like the safest way to do it.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
ctarda: As you mentioned, it's all about the tradeoffs. At this point, it seems to me like the benefit of encapsulating the navigation logic is offset a little bit by the fact that there is logic in ViewModelA that is leaking, and is spread in between ViewModelA (who knows how to instantiate ViewModelB) VIewControllerA (who knows that VIewModelA needs to provide a ViewModelB) and the Application Controller, which knows that needs to set a closure in ViewControllerA that requires a parameter of type ViewModelB.

I think that if we really want to achieve what the pattern aims for (decoupling view controllers from navigation) we should try to find a way to avoid that leakage.  

What you mentioned in your ideal scenario seems to reduce that leakage, at least partially. 

I am not completely sure about this though: `ViewControllerB exposes dependencies through the init and uses those dependencies to create it's ViewModelB.` I don't think a view controller should be responsible for creating its own view model. 

However, back to your ideal scenario, I think that solves the issue of the Application Controller knowing too much about the internals of the View Controller. 

If we also assume that view models are injected into view controllers, it is safe to assume that the first view controller in any navigation sequence would be aware of how the view model for that first few controller is instantiated. But even if it didn't, what matters, to me, is that if the Application Controller is responsible for instantiating view controllers, it is okay if it is also responsible for instantiating the view models necessary for any of those view controllers.

So, what do you think, should we give a try to:
- ViewControllerA declares a navigation delegate.
- The Application Controller conforms to that delegate
- The Application Controller does whatever it knows it has to do to instantiate ViewControllerB, providing it with what it needs.

I think that would encapsulate the responsibility of moving between view controllers better.

Downsides? At the more granular level, there would be one navigation delegate per view controller, which can be a bit noisy. On the other hand, it is what it is, it reflects reality: each view controller might have different navigation flows out of it, there is no point in trying to deal with those VCs as if they were the same, because they are not.

At a higher lever of abstraction, I think we would need to ask ourselves if the complexity (or lack of thereof) of the solution is really worth it, compared to the problem it solves. Would things really be better after this PR were merged?

What do you think?

ctarda: That being said... (and I want to extract this to a separate comment to separate it from the other conversation)

ViewModelA is (or should be, or could be) aware of the user interactions in ViewControllerA. In particular, if a user taps a button that should trigger a navigation action, it is logical to expect that ViewControllerA could call a method into ViewModelA. 

Is there anything stopping us from making ViewModelA be the one that actually calls the methods declared in ViewControllerANavigationDelegate? 

That would:
*  solve the problem of creating ViewModelB when there is a relation between ViewModelA and ViewModelB. ViewModelA can create it if needed. 
* make ViewControllerA completely unaware of anything that is not its view model.
* remove the need for the Application Controller to reach into ViewControllerA or ViewModelA to get anything that it needs to instantiate ViewControllerB

So, the sequence would be:
* user taps button in ViewControllerA
* ViewControllerA calls method in ViewModelA
* ViewModelA does whatever it needs to do (tracks events, call a remote service, whatever...) and then calls the corresponding method declared in ViewControllerANavigationDelegate. Provides the next view model if needed?
* which reaches the Application Controller, who navigates to ViewControllerB

My point is, one we have clear separation of concerns, and navigation is fully decoupled form the view controller, we can pretty much do whatever make sense to us. What matters is the spirit, not the letter of the pattern. 

What do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
Ecarrion: > Why would we want to reuse the view controller in a different flow? 
Some use cases I can think of would be:

- A/B Testing: Having different coordinators served to a different set of people.
- iPad Coordinators: IPad apps may choose to display or handle view controllers in a different way. Having VC's reusable will help us to achieve that. 
- Tutorials: We can have coordinators with mocked up dependencies so we can show "real-live" tutorials on how to use our apps.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
Ecarrion: > As you mentioned, it's all about the tradeoffs. At this point, it seems to me like the benefit of encapsulating the navigation logic is offset a little bit by the fact that there is logic in ViewModelA that is leaking, and is spread in between ViewModelA (who knows how to instantiate ViewModelB) VIewControllerA (who knows that VIewModelA needs to provide a ViewModelB) and the Application Controller, which knows that needs to set a closure in ViewControllerA that requires a parameter of type ViewModelB.

I would say that the main benefits I'm trying to achieve now is:
- Making View controllers as dumb as possible.
- Stop assumptions on VC's that I'm in a modally - navigation context.
- Enabling me to unit test this navigation flow.

It's true that the VM creation is leaking and we can certainly work on improving that, but it's already leaking on the current(develop) implementation so I don't think it matters much for this PR 😅 

> I am not completely sure about this though: ViewControllerB exposes dependencies through the init and uses those dependencies to create it's ViewModelB. I don't think a view controller should be responsible for creating its own view model.

Hmmm, I guess this is preferences / styles of the project. I like VC's to create their VMs because it simulates more closely apple APIs and their lifecycle where the VC is the entry point for any new screen. To me, coordinators know about view controllers not about view models, but I guess in practice there isn't much difference.

> At a higher lever of abstraction, I think we would need to ask ourselves if the complexity (or lack of thereof) of the solution is really worth it, compared to the problem it solves. Would things really be better after this PR were merged?

I think it does, unit-testing this flow is good enough for me for going through this trouble. Regarding the ideal solution, I came from a project where we had a very strict and complex coordinator implementation. I would say that with these light coordinators I'm still experimenting with what feels right and with what is worth it or not!

Ecarrion: > Is there anything stopping us from making ViewModelA be the one that actually calls the methods declared in ViewControllerANavigationDelegate?

> That would:

> solve the problem of creating ViewModelB when there is a relation between ViewModelA and ViewModelB. ViewModelA can create it if needed.
> make ViewControllerA completely unaware of anything that is not its view model.
> remove the need for the Application Controller to reach into ViewControllerA or ViewModelA to get anything that it needs to instantiate ViewControllerB

I think that may be possible(without validating it first) but to me, that just keeps adding more and more complexity to the `ViewModel` which, in my mind, is what we are trying to avoid on the `ViewController`.

> So, the sequence would be:

> user taps button in ViewControllerA
> ViewControllerA calls method in ViewModelA
> ViewModelA does whatever it needs to do (tracks events, call a remote service, whatever...) and then calls the corresponding method declared in ViewControllerANavigationDelegate. Provides the next view model if needed?
> which reaches the Application Controller, who navigates to ViewControllerB

Wouldn't this also leak the next ViewModelB abstraction that you mentioned before?

ctarda: So it seems like our approaches are pretty much at the opposite ends of the spectrum. 

Time to disagree and commit 😊? :shipit:

ctarda (approved):
:shipit:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
Ecarrion: Ha! I won't say we are at different sides, just that I don't think the `ViewModel` should know about the Navigation and flow logic I think its responsibilities should be contained to what a view controller needs. 

Speaking of that, That's one reason why I don't like the name view model for these objects, that name does not clearly define a single responsibility and quickly become bloated, like "managers" used to be in early iOS days. But I guess that's another discussion 😅

PR #3219, opened 2020-11-27
Insert custom error VC into WPAuthenticator, overriding inline error message

### LINKED ISSUE #3218: [UL&S] UI. Not a WordPress site
Implement UI for error screen when user enters a site address that does not correspond with a WordPress site

ctarda:
Closes #3218 
Ref: p91TBi-3Rn-p2

⚠️⚠️ Please note also that the coloring of the CTAs (Back button, Help button and others) is implemented in #3253 

## Design

⚠️⚠️⚠️ Please note there is a limitation that, as far as I can tell, we can not work around: at the moment of presenting the error there is no way to know the site address the user entered, so I removed it from the error message.

## Changes

First a gif of the new behaviour:

Screenshots:
| Light mode | Dark mode |
| ---- | ---- |
|  |  |

The changes to the code are:
* Added a new view model to represent this error
* Pointed WPAuthenticator to the corresponding branch
* Implemented the new methods declared in WPAuthenticatorDelegate, and passed a properly customized error UI to WPAuthenticator.

## How to test
* Check out the branch. Run `bundle exec pod install`
* Log out if necessary
* Attempt to log in with a store address. Use an address that does not correspond to a WP site (i.e. GitHub.com)
* Note: The "Log In With Another Account" button in the error screen should pop to the Login Prologue. The "Enter Another Site" button is deactivated in order to submit the PR, as the functionality to support t is part of a different PR in WPAuthenticator wordpress-mobile/WordPressAuthenticator-iOS#532 . Making progress in two different repos with the desired level of granularity is hard!

Update release notes:


WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -183,6 +183,19 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
     ///
     func createdWordPressComAccount(username: String, authToken: String) { }
 
+    func shouldDisplayError(_ error: Error) -> Bool {
```
ctarda: This is the implementation of the two new methods in WPAuthenticatorDelegate. First we signal that we want to handle errors (this should be more granular, and return true only for certain error codes)

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -183,6 +183,19 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
...
+        return true
+    }
+
+    func handleError(_ error: Error, onCompletion: @escaping (WordPressAuthenticatorResult) -> Void) {
+        let viewModel = NotWPErrorViewModel(siteURL: "http://woocommerce.com")
```
ctarda: And here we instantiate custom UI and pass it back to WPAuthenticator. WPAuthenticator then will short-circuit its own error handling flow and just present the UI we provide

ctarda: It's probably fine if we just pass a view controller, instead of an instance of WordPressAuthenticatorResult to the completion closure. After all, host apps would only call this method *if* they want to insert custom UI to present the error

Garance91540 (approved):
Looking good Cesar! Is it possible to change "Site" with "Store" on the "Try another site" CTA?

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -183,6 +183,25 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
     ///
     func createdWordPressComAccount(username: String, authToken: String) { }
 
+    func shouldHandleError(_ error: Error) -> Bool {
+        let err = error as NSError
+        /// This is the error code that signals that a site is not a WP site.
+        /// It corresponds to WordPressKit.WordPressOrgXMLRPCValidatorError
+        /// However, we do not pull WordPressKit as a dependency
+        guard err.code == 406 else {
```
Ecarrion: Do you think it's worth it for us to have those error codes mapped to an internal enum?

That way the could encapsulate that logic to something like:
```swift
func shouldHandleError(_ error: Error) -> Bool {
   return OurNewError(error) == .notWPSite
}
```

WooCommerce/Classes/Authentication/Navigation Exceptions/NotWPErrorViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+/// Configuration and actions for an ULErrorViewController, modelling
+/// an error when the site is not a WordPress site
+struct NotWPErrorViewModel: ULErrorViewModel {
+    // MARK: - Data and configuration
+    let image: UIImage = .loginNoWordPressError
+
+    var text: NSAttributedString {
+        let message = NSMutableAttributedString(string: Localization.errorMessage)
+        return message
+    }
```
Ecarrion: nit:
```suggestion
    var text: NSAttributedString {
        NSMutableAttributedString(string: Localization.errorMessage)
    }
```

WooCommerce/Classes/Authentication/Navigation Exceptions/NotWPErrorViewModel.swift
```diff
@@ -0,0 +1,65 @@
...
+                                                          + "Presented when logging in with a site address that is not a WordPress site")
+
+        static let secondaryButtonTitle = NSLocalizedString("Log In With Another Account",
+                                                            comment: "Action button that will restart the login flow."
+                                                            + "Presented when logging in with a site address that does not have a valid Jetpack installation")
+
+    }
+
+    enum Strings {
+        static let instructionsURLString = "https://docs.woocommerce.com/document/jetpack-setup-instructions-for-the-woocommerce-mobile-app/"
```
Ecarrion: I'm not seeing this being used, will it be needed later?

WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -309,12 +309,18 @@ extension UIImage {
...
+    /// Login error: no Jetpack
     ///
     static var loginNoJetpackError: UIImage {
         return UIImage(named: "woo-no-jetpack-error")!
     }
 
+    /// Login error: no WordPress
+    ///
+    static var loginNoWordPressError: UIImage {
+        return UIImage(named: "woo-wp-no-site")!
```
Ecarrion: I think this has to be rotated for RTL layouts right?
```suggestion
        return UIImage(named: "woo-wp-no-site")!.imageFlippedForRightToLeftLayoutDirection()
```

WooCommerce/Classes/Authentication/Navigation Exceptions/NotWPErrorViewModel.swift
```diff
@@ -0,0 +1,65 @@
+import UIKit
+import SafariServices
+import WordPressAuthenticator
+import WordPressUI
+
+
+/// Configuration and actions for an ULErrorViewController, modelling
```
Ecarrion: typo: `modelling` -> `modeling`
```suggestion
/// Configuration and actions for an ULErrorViewController, modeling
```

Ecarrion (commented):
Left some minor comments/suggestions. Also, I'm not sure if this is till up for debate but it seems an unnecessary transition to display an error on another screen only to be forced to go back again. Do you think it will be better to show that error on the current screen?

WooCommerce/Classes/Authentication/AuthenticationManager.swift
ctarda: it makes sense!

WooCommerce/Classes/Authentication/Navigation Exceptions/NotWPErrorViewModel.swift
ctarda: Good catch, I'll remove it!

ctarda:
@Garance91540 Do you have any input about @Ecarrion 's suggestion here? : https://github.com/woocommerce/woocommerce-ios/pull/3219#pullrequestreview-542078189

From a purely technical point of view, I don't think WPAuthenticator will let us insert custom UI without a transition to a separate screen. The view where users enter the store address is served by WPAuthenticator. The view where we present the error is served by WooCommerce.

ctarda:
I think I've covered everything:
* Added a private enumeration to map to error codes emitted by WPAuthenticator
* Various nits and bits (including making the image for the Jetpack error aware or RTL)
* Pinned WPAuthenticator to the beta release that implements support for all of this.
* Updated the copy of the primary action button:



This is ready for another round!

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -344,3 +340,25 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
...
 }
+
+
+// MARK: - Error handling
+private extension AuthenticationManager {
+
+    /// Maps error codes emitted by WPAuthenticator to a domain error object
+    enum AuthenticationError: Int, Error {
+        case notWPSite = 406
+        case unkown
```
Ecarrion:
```suggestion
        case unknown
```

Ecarrion (approved):
Looks good! Just a minor typo.
🚀

ctarda:
Alrighty, typo in the enumeration fixed! Merging! 🚀

PR #3220, opened 2020-11-27
Add keyboard dismiss in Order -> Add Tracking screen

### LINKED ISSUE #3102: The keyboard is not correctly dismissed in Order -> Add Tracking screen
Ref. p5T066-1Bt#comment-6192

@rachelmcr while was testing the beta release 5.4 discovered a bug to the keyboard that is not hidden in the Add Tracking screen.

**I report her message here:**
_When I’m adding shipment tracking to an order, once I tap into the Tracking Number field the keyboard comes up and there’s no way to dismiss it. (I usually go through that screen top to bottom, so I choose a shipping provider, enter the tracking number, and then select the date.) I can’t find any way to change the focus away from that tracking number field so the keyboard goes away.
Although I can still select the date field and use the date picker, the keyboard remains on screen. Is there a way to auto-dismiss the keyboard once the date field is selected, or otherwise make it possible to dismiss the keyboard there?_

**It would be good in my opinion to dismiss the keyboard by pressing the date picker and also the return button on the keyboard.**

ealeksandrov:
Fixes https://github.com/woocommerce/woocommerce-ios/issues/3102.

##  Description

This PR adds keyboard dismiss in Order -> Add Tracking screen:
1. on "return" when tracking number field is active
2. when carrier field tapped
3. when calendar field tapped

### Known limitations

Can be seen in the end of "after" video.
Keyboard won't dismiss if you tap inside date picker expanded cell - this is managed by date picker and I didn't want to increase complexity and affect all usage cases of this date picker. Also with regular flow user will first tap on "date" cell to expand datepicker - and keyboard will be dismissed already for next step.

## Testing

1. Open order screen
2. Tap "Add tracking" option
3. Tap "Tracking number" field to activate keyboard
4. Try tapping "return" on keyboard
5. Try tapping other fields (carrier and date) and observe keyboard dismiss

## Videos

 Before | After 
--------|------
Screen Recording 2020-11-26 at 18 36 26|Screen Recording 2020-11-27 at 12 24 21

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -8,6 +8,7 @@
 - [*] removed the Product Title in product screen navigation bar. [https://github.com/woocommerce/woocommerce-ios/pull/3187]
 - [*] the icon of the cells inside the Product Detail are now aligned at 10px from the top margin. [https://github.com/woocommerce/woocommerce-ios/pull/3199]
 - [**] Added the ability to issue refunds from the order screen. Refunds can be done towards products or towards shipping. [https://github.com/woocommerce/woocommerce-ios/pull/3204]
+- [*] Add keyboard dismiss in Add Tracking screen [https://github.com/woocommerce/woocommerce-ios/pull/3220]
 
 
 5.5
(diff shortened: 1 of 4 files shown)


pmusolino (approved):
Looks good to me :) great! 👍

PR #3221, opened 2020-11-27
Prevent banner dismiss when tapping "give feedback" on products screen

### LINKED ISSUE #2945: [IAFeedback] Do Not Dismiss the Banner When Tapping on Give Feedback
Currently, tapping on “Give Feedback” would dismiss the banner. As suggested by @rachelmcr, we should only dismiss the banner if the user tapped on “Dismiss” in case they want to provide more than one feedback. 

> Would it make sense to not permanently dismiss the banner unless it has been explicitly dismissed? (So don’t permanently dismiss it even if the user submits the survey?) Especially if I’m a user who enjoys tinkering with beta features, I might like to have that always available as a route to submit feedback e.g. if I notice something else later that I want to share feedback on, after submitting the survey once.

ealeksandrov:
Fixes https://github.com/woocommerce/woocommerce-ios/issues/2945.

##  Description

This PR adds removes banner dismiss logic when tapping "give feedback" on products screen.

## Testing

1. Open products screen
2. Tap "Give feedback"
3. Dismiss feedback screen
4. Banner should be still available

## Screenshot

Update release notes:


pmusolino (approved):
LGTM :shipit:

PR #3186, opened 2020-11-23
[UL&S] Login with site address: Jetpack not installed or not connected

ctarda:
Closes #3130 
Closes #3131 

Ref: p91TBi-3ON-p2

⚠️ Please note this PR is not against develop, but against the feature branch for UL&S 🙇

##Design

## Changes

In action:

Screenshots:

| | Light Appearance | Dark Appearance |
| ---- | ---- | ---- |
| iPhone SE |  |  |
| iPhone 12 |  |  |
| iPhone 12 Pro Max |  |  |

⚠️  Adding the help button requires another update to WPAuthenticator, so I have deferred it to a separate issue: #3214

## How it works
The heavy lifting is done by WordPressAuthenticator. What we do is provide a View Controller to the library, and let it inset it in whatever way it makes sense. See p91TBi-3ON-p2 for more context.

This PR:
* Points the Podfile to a specific beta release of WPAuthenticator.
* Adds a new view controller (ULErrorViewController). All error messages for UL&S seem to be modeled using the same UI pattern, so I decided to go "generic", and attempt to implement all these special screens (that contain an image, text, a text button and two larger buttons) using the same view controller
* The way to customize this ULErrorViewController is by injecting a view model that provides the content (i.e. text, button titles) and that handles button actions.
* Adds unit tests

## Testing
* Checkout the branch. Run `bundle exec pod install`
* Log out if necessary
* Attempt to Log In with Store Address. Enter a store address of a site that does not have Jetpack installed or does not have Jetpack connected.
* Notice if the error screen in shown. 
* The "What is Jetpack" button should load a web view pointing to this article: https://jetpack.com/about/ * The "See instructions" button should load a web view pointing to: https://docs.woocommerce.com/document/jetpack-setup-instructions-for-the-woocommerce-mobile-app/
* The "Refresh after install" button should navigate to the screen where users enter the site address. It is not actually possible to refresh this screen, so there isn't much more we can do, as far as I can tell. Better solutions are welcome!

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3225, opened 2020-11-30
Insert a Help Button into custom UI provided to WordPressAuthenticator

### LINKED ISSUE #3124: [UL&S] Activate WPAuthenticator flag for Unified Login

ctarda:
Closes #3124 

The actual work of inserting the Help button is done by WPAuthenticator, and implemented in this PR: wordpress-mobile/WordPressAuthenticator-iOS#527

What this PR does is point WPAuthenticator to the relevant branch and comment out a bit of code that will have to be un-commented when PR #3186 is merged.

| Before | After |
| ---- | ---- |
|  |  |

Notice the Help button in the navigation bar

## Changes
* Point the reference to WPAuthenticator to the branch that implements the solution
* Comment out (temporarily, only until #3186 is merged) two lines of code that would break the build. The reason for that is that PRs in WPAuthenticator that break the API are not in sync with PRs in WooCommerce.

## Testing
* Checkout the branch
* Try to log in with a Site Address, using a site that does not have Jetpack installed, or does not have Jetpack connected (as this is the only custom error we present)

Update release notes:


WooCommerce/Classes/Authentication/Navigation Exceptions/JetpackErrorViewModel.swift
```diff
@@ -49,8 +49,10 @@ struct JetpackErrorViewModel: ULErrorViewModel {
     }
 
     func didTapSecondaryButton(in viewController: UIViewController?) {
-        let refreshCommand = NavigateToRoot()
-        refreshCommand.execute(from: viewController)
+        // To be removed when PR #3186 is merged.
```
ctarda: This should not be part of the diff, it is temporary, and I will remove it before merging. 

But, the API of WPAuthenticator is a bit ahead

jaclync: Sounds tricky with different changes going on in both repos!

jaclync (approved):
Code changes look good! I tried testing with a new Pressable site with Jetpack deactivated but I still see the login by credentials screen 🤔 (I couldn't delete Jetpack plugin on my site) not sure if I missed some setup for the error screen?


plugins | after logging in with site address
-- | --
Screen Shot 2020-11-30 at 6 12 15 PM | Simulator Screen Shot - iPhone 11 Pro - 2020-11-30 at 18 13 42

Garance91540 (approved):
Hi Cesar, this is looking good! You probably will address this separately, but I wanted to make sure that all the CTAs ("Back", "Help", "Find your store address") will be displayed in Pink 50 https://d.pr/i/RsNXke

ctarda:
Thanks for the review!

I logged #3251 for setting the color of the CTAs properly, and #3252 for reviewing the flow for testable sites. 

I'll merge as soon as CI gives the ✅

PR #3227, opened 2020-11-30
Fix secondary button background color

jaclync:
Part of #2167 

## Changes

After the change to set a normal background image for primary buttons, the secondary button in a cell still keeps the background image when it's reused from a cell with primary button. This PR set the normal background image for secondary buttons similar to primary buttons in https://github.com/woocommerce/woocommerce-ios/commit/74c81971cf357ec190a70f8712969663112fd4d3#diff-cfbdf6fc7e3bd4bf03be587e71fd7e9dc505fad690ad8529962015081d2009fc so that it has the correct background color

## Testing

Sometimes I can reproduce with "Issue Refund" CTA in order details, but not all the time. I can always reproduce in order details with at least one shipping label.

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> after syncing, a shipping label package card should be shown with a "Reprint Shipping Label" CTA that has the expected background color (see screenshots below)

## Example screenshots

(Please notice the background color of the "Reprint Shipping Label" CTA)

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-11-30 at 13 44 29 | Simulator Screen Shot - iPhone 11 - 2020-11-30 at 13 48 21
Simulator Screen Shot - iPhone 11 - 2020-11-30 at 13 45 49 | Simulator Screen Shot - iPhone 11 - 2020-11-30 at 13 48 07

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIButton+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UIButton+Helpers.swift
@@ -52,6 +52,11 @@ extension UIButton {
         setTitleColor(.secondaryButtonTitle, for: .highlighted)
         setTitleColor(.buttonDisabledTitle, for: .disabled)
 
+        let normalBackgroundImage = UIImage.renderBackgroundImage(fill: .secondaryButtonBackground,
+                                                                  border: .secondaryButtonBorder)
+            .applyTintColorToiOS13(.secondaryButtonBackground)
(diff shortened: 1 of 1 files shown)


Garance91540 (approved):
👏👏👏

pmusolino (approved):
LGTM :shipit:

PR #3229, opened 2020-11-30
Create a feature flag for shipping labels M1

### LINKED ISSUE #3228: Create a feature flag for shipping labels M1
As we add shipping label M1 features to order details, let's hide them behind a feature flag until it's ready for launch.

jaclync:
Fixes #3228 

## Changes

- Added feature flag for shipping labels M1: `FeatureFlag.shippingLabelsRelease1` which is only enabled for debug or alpha builds
- In order details VC & data source, only sync and show shipping labels in order details if the feature flag is enabled

## Testing

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/) and there is an order with at least one shipping label.

- In code, manually returned `false` in `DefaultFeatureFlagService`'s `shippingLabelsRelease1` case
- Launch the app
- Go to the orders tab
- Tap on an order with at least one shipping label --> there should **not** be any shipping label package card in order details UI after a few seconds (they are visible in `develop` or when the feature flag is enabled for the same order)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
+++ b/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
@@ -7,6 +7,8 @@ struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .editProductsRelease5:
             return buildConfig == .localDeveloper || buildConfig == .alpha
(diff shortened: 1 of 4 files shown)


ctarda (approved):
:shipit:

PR #3249, opened 2020-11-30
Update FooterSpinnerView to have single appearance in all tables

### LINKED ISSUE #1922: Order List: the padding of the bottom loader indicator is wrong
It seems to me that the padding of the bottom loader in the Orders table view is wrong, it seems too close to the above cell. It should look like in the product list.

## Mobile Environment
Please include:
- Device: iPhone 11 Pro simulator
- iOS version: 13.2.2
- WooCommerce iOS version: 3.8

ealeksandrov:
Fixes https://github.com/woocommerce/woocommerce-ios/issues/1922.

##  Description

Initial issue - padding of the bottom loader in the Orders table view is wrong.

After some investigation in https://github.com/woocommerce/woocommerce-ios/issues/1922#issuecomment-734796730 I decided to simplify `FooterSpinnerView` layout to have single appearance independent of tableview style.

Here is old layout logic as I understand:
- since grouped tableview has excessive padding before footer view - pin activity indicator to the top of the footer, so it will appear centered (vertically)
- plain tableview displays footer view just fine, without extra padding - pin activity indicator to the center of the footer

I suggest to deal with grouped tableview padding by setting `tableView.sectionFooterHeight = .leastNonzeroMagnitude` where needed. And with that every footer spinner will appear in same way and same height in all tables.

While testing I observe that plain tableviews look just as before and grouped have slightly decreased footer size (because it had additional padding previously). For all grouped (and therefore changed) cases comparison screenshots provided below.

## Screens affected

Plain (no change):
- ProductsViewController
- SearchViewController
- ProductVariationsViewController

Grouped (footer height decreased to 45):
- OrderListViewController and OrdersViewController (already have `.leastNonzeroMagnitude` footer height and that's why indicator appeared wrong)
- ReviewsViewController
- ProductReviewsViewController
- ShipmentProvidersViewController

PaginatedListSelectorViewController plain implementations:
- LinkedProductsListSelectorViewController
- ProductListSelectorViewController

PaginatedListSelectorViewController grouped implementations:
- ProductShippingSettingsViewController
- ProductPriceSettingsViewController

## Testing

1. Open any of screens above and scroll to the bottom.
2. Observe that spinner is centered vertically.

### OrderListViewController
 Before | After 
--------|------
|

### ReviewsViewController
 Before | After 
--------|------
|

### ProductReviewsViewController
 Before | After 
--------|------
|

### ShipmentProvidersViewController
 Before | After 
--------|------
|

### ProductShippingSettingsViewController
 Before | After 
--------|------
Shipping1|Shipping2

### ProductPriceSettingsViewController
 Before | After 
--------|------
Price1|Price2

Update release notes:


WooCommerce/Classes/ViewRelated/ListSelector/PaginatedListSelectorViewController.swift
```diff
@@ -66,7 +66,7 @@ where DataSource.StorageModel == StorageModel, Model == DataSource.StorageModel.
     /// Footer "Loading More" Spinner.
     ///
     private lazy var footerSpinnerView = {
-        return FooterSpinnerView(tableViewStyle: tableView.style)
+        return FooterSpinnerView()
     }()
```
jaclync: nit: can further simplify to:

```swift
private lazy var footerSpinnerView = FooterSpinnerView()
```

(and similar for other instances in the PR, but not blocking!)

jaclync (approved):
Thanks for digging into the root cause of mismatching `sectionFooterHeight` (especially harder to notice in the xib) and consolidating the footer spinner view!

I created [a branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/1922-fix-spinner-padding) to run CI, and the tests look ✅  just tagging @pmusolino to double check!

Code: ✅ 
Testing: ✅ (I went through the affected screens and all of them have the expected height! Your list of affected screens is helpful for testing)

WooCommerce/Classes/ViewRelated/ListSelector/PaginatedListSelectorViewController.swift
ealeksandrov: Agree 🙂
I even did that first, but spotted that there are some examples around with similar 1 line closure syntax. So there are no style conventions about it and I can convert them to one liners?

I would like to also convert to 1 line related footer placeholder views, like this one:
https://github.com/woocommerce/woocommerce-ios/blob/19af5d0a7d1217c788d98d026a1e9dc8ffbc4aee/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift#L32-L34

jaclync: I don't think it's documented in our [style guide](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/coding-style-guide.md), but generally we've been going for the more concise way if it does not impact readability. There are a few recent instances where we use line for a lazy var like:

https://github.com/woocommerce/woocommerce-ios/blob/19af5d0a7d1217c788d98d026a1e9dc8ffbc4aee/WooCommerce/Classes/Authentication/AppleIDCredentialChecker.swift#L10

And yes, the `footerEmptyView` would be a good place to be one liner too 🙂

ealeksandrov: Sounds good, pushed styling update 👍

ealeksandrov:
Thank you for review, @jaclync! 👍

WooCommerce/Classes/ViewRelated/ListSelector/PaginatedListSelectorViewController.swift
```diff
@@ -65,13 +65,11 @@ where DataSource.StorageModel == StorageModel, Model == DataSource.StorageModel.
...
     ///
-    private lazy var footerSpinnerView = {
-        return FooterSpinnerView(tableViewStyle: tableView.style)
-    }()
+    private lazy var footerSpinnerView = FooterSpinnerView()
 
-    private lazy var footerEmptyView = {
-        return UIView(frame: .zero)
-    }()
+    /// Empty Footer Placeholder. Replaces spinner view and allows footer to collapse and be completely hidden.
```
jaclync, line 70: thanks for adding the comment!

jaclync (approved):
Thanks for the update! LGTM :shipit: 

(also reran CI in [this branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/1922-fix-spinner-padding) ✅ )

ealeksandrov:
Merged in `develop`, resolved conflict (changed file was deleted in develop, no changes added).

PR #3254, opened 2020-12-01
Order Details > Shipping Label: hide order item total & subtitle when the price is unknown

jaclync:
Part of #2167

## Why

In a shipping label's API response, we only know the product/variation IDs or names depending on the WCS plugin version. Before the products/variations are fetched given the IDs, we do not know the price/total of the product/variation for an order item in a shipping label. For example, right now we don't fetch the missing variations for an Order in order details (it's a subtask in #2167), so that the order items with missing variations don't have a valid price/total. We currently set the missing price/total values for such order items to zero as in the before/after screenshots below, and this PR updates `AggregateOrderItem` to allow its price and total values to be `nil` so that we can hide the total & subtitle labels in this scenario.

## Changes

I took the chance to refactor `ProductDetailsCellViewModel` a bit so that it's easier to customize the logic when the price/total value is `nil`.

- Updated `AggregateOrderItem`'s `price` and `total` to be nullable
- In `AggregatedShippingLabelOrderItems`, updated to set `AggregateOrderItem`'s `price` and `total` to be `nil` if only the name is known
- In `ProductDetailsCellViewModel`, refactored to remove the private properties that were only used for public calculated properties and consolidate the logic that sets the public properties in a private initializer
  - The only behavior change is so that we set the `total` and `subtitle` strings to be empty if an item's `price`/`total` is `nil`
  - Added unit tests for the existing `ProductDetailsCellViewModel` usage for `OrderItem`/`AggregateOrderItem`/`OrderItemRefund`
- Added an Auto Layout constraint to `ProductDetailsTableViewCell`'s image view: `superview.bottom >= imageView.bottom + 8px` so that the image has some margin at the bottom when the subtitle isn't shown

## Testing

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/).

- In core, create an order with a variation. Then create a shipping label for this order
- Launch the app, log out and in to the store without going to the Products tab
- Go to the orders tab
- Tap on the order with a variation & shipping label --> after syncing, the variation order item should not show any total or subtitle like in the example screenshot below

## Example screenshots

before | after
-- | --
 | 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
@@ -11,39 +11,6 @@ struct ProductDetailsCellViewModel {
         let value: String
     }
 
-    // MARK: - Private properties
-
-    /// Yosemite.Order.currency
-    ///
-    private let currency: String
-
-    /// Currency Formatter
-    ///
-    private let currencyFormatter: CurrencyFormatter
-
-    /// Item Quantity
-    /// represented as a positive value
-    ///
-    private let positiveQuantity: Decimal
-
-    /// Item Total
-    /// represented as a positive value
-    ///
-    private let positiveTotal: NSDecimalNumber
-
-    /// Item Price
-    /// represented as a positive value
-    ///
-    private let positivePrice: NSDecimalNumber
-
-    /// Item SKU
-    ///
-    private let skuText: String?
-
-    /// Attributes of an order item.
-    ///
-    private let attributes: [OrderAttributeViewModel]
-
     // MARK: - Public properties
 
     /// The first available image for a product/variation.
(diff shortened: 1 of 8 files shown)


Garance91540 (approved):
👏

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
```diff
@@ -56,54 +23,73 @@ struct ProductDetailsCellViewModel {
...
+            }
+            return currencyFormatter.formatAmount(positiveTotal, with: currency) ?? String()
+        }()
+
+        self.subtitle = {
+            guard let positivePrice = positivePrice else {
+                return ""
+            }
+            let itemPrice = currencyFormatter.formatAmount(positivePrice, with: currency) ?? String()
+            return Localization.subtitle(quantity: quantity, price: itemPrice, attributes: attributes)
```
Ecarrion: I really like how these inline closures make the code more readable while maintaining everything in the same context!

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
```diff
@@ -56,54 +23,73 @@ struct ProductDetailsCellViewModel {
...
+            }
+            return String.localizedStringWithFormat(Localization.skuFormat, sku)
+        }()
 
-    // MARK: - Initializers
+        self.total = {
+            guard let positiveTotal = positiveTotal else {
+                return ""
+            }
+            return currencyFormatter.formatAmount(positiveTotal, with: currency) ?? String()
```
Ecarrion, line 66: Newbie question: What is the difference between `String()` and `""`?

Ecarrion (approved):
Looks good! 🚀 

First Launch | After Loading product
---- | ----
Screen Shot 2020-12-01 at 2 00 18 PM | Screen Shot 2020-12-01 at 2 01 16 PM

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
jaclync: Great question! `String()` and `""` are the same, according to the [Apple doc](https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html) under "Initializing an Empty String":

> var emptyString = ""               // empty string literal
var anotherEmptyString = String()  // initializer syntax
// these two strings are both empty, and are equivalent to each other

I'll update the empty string literal to `String()` to avoid confusion 😄

jaclync: Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3254/commits/ad204b4390d4ae2165acb43535e4fb2f5865bc8a

PR #3263, opened 2020-12-01
Cleaning: removed conditions and classes no more needed on iOS 13

pmusolino:
As part of the [migration to iOS 13](https://github.com/woocommerce/woocommerce-ios/pull/3216), there was still some code and classes with old conditions for making the things working on iOS 12.

In this PR, I removed all these conditions, removed the part that references to iOS 12 and prior versions, and I also removed some classes like `OrdersViewController` and `OrdersViewModel` that are now replaced with `OrderListViewController` (this was a planned move, that now we can accomplish).
 
## Testing
I don't expect any changes to the app. It would be good to do a general test of the app, just to make sure we didn't introduce any regression. In particular, looking at the things that have been updated, it would be good to focus on:
- Order List
- Order Detail
- Light/Dark mode

Update release notes:


shiki:
Paolo! You're awesome! Thanks for doing this!

Ecarrion (approved):
It all looks good in my tests! Thanks for this @pmusolino!

pmusolino:
Thank you Ernesto for the review 💯

PR #3265, opened 2020-12-01
Merge release/5.6 into develop after new beta + Release Notes

AliSoftware:
New beta 5.6.0.1 + Release Notes update with Editorial copy [ref: p2JRYi-58I-p2]

### DIFF
--- a/WooCommerce/Resources/AppStoreStrings.pot
+++ b/WooCommerce/Resources/AppStoreStrings.pot
@@ -57,11 +57,11 @@ msgctxt "app_store_promo_text"
 msgid "Run your store from anywhere"
 msgstr ""
 
-msgctxt "v5.5-whats-new"
+msgctxt "v5.6-whats-new"
 msgid ""
-"Products: You can now add simple, grouped, and external products and save products as drafts. We’ll have more new features for you to try out soon!\n"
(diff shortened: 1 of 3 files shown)


loremattei (approved):
I don't see any change related to a new beta here, other than the version number, but it looks like this is because of https://github.com/woocommerce/woocommerce-ios/pull/3262. 

Am I right @AliSoftware?
If so, feel free to merge it.

AliSoftware:
> I don't see any change related to a new beta here, other than the version number, but it looks like this is because of #3262.
> 
> Am I right @AliSoftware?
> If so, feel free to merge it.

Yep, basically if you look at the ascii-art in #3262 you'll see that what happened is that I needed to do a new beta after the 3 fixes… but since the merge back into `develop` had not been merged post-code-freeze at that time (those hotfixes happened on the release branch faster than the approval for post-❄️-merge PR), I wasn't able to do a new beta right in #3262 as `develop` didn't yet have the new version number (and thus `new_beta_release` suggested an incorrect bump).

This is why I had to merge into `develop` once first (via #3262), to get not only the 3 fixes but also the beta-1 version bump in `develop` that never had the time to get merged post-code-freeze at that point, then immediately do a `new_beta_release` (which this time found the right version in develop to do the proper bump) then this PR. Just a matter of unfortunate timing with the fixes coming faster than the post-code-freeze-remerge-to-develop approval 😅

PR #3266, opened 2020-12-01
Add support for injectable mock object graphs

jkmassel:
This PR adds support for type-safe and translatable mocks by injecting an object graph (paired with mocked remotes that emit these objects) into the app. This will be useful for screenshots and UI tests because the app will fail to compile if the mocks are broken, which should ensure that things remain in sync. Additionally, this removes a dependency on WireMock, which should simplify the process of building and running UI Tests.

All the mocking code lives in the `Network` layer, though I could see an argument for moving `ScreenshotObjects` (and a future `UITestObjects` object) up into the `Yosemite` layer too?


shiki:
> All the mocking code lives in the Network layer, though I could see an argument for moving ScreenshotObjects (and a future UITestObjects object) up into the Yosemite layer too?

I was thinking, if possible, we could have them in a separate Swift Package. 😁

Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -17,7 +17,7 @@ public class ProductStore: Store {
         self.init(dispatcher: dispatcher, storageManager: storageManager, network: network, remote: remote)
     }
 
-    init(dispatcher: Dispatcher, storageManager: StorageManagerType, network: Network, remote: ProductsRemoteProtocol) {
+    public init(dispatcher: Dispatcher, storageManager: StorageManagerType, network: Network, remote: ProductsRemoteProtocol) {
```
jkmassel, line 20: These constructors could be combined in `...Store` classes by making the remote optional and providing a default in the method body, but that might be an over-optimization (at least until the other `...Store` classes have the second constructor?)

Networking/Networking/Model/Mocks/MockObjectGraph.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+public protocol MockObjectGraph {
```
AliSoftware, line 3: Was just wondering: why did you make this a protocol, instead of directly using the `struct`?

iinm, in the current state of that PR the only object conforming is `struct ScreenshotObjects`, right?
I'm assuming the idea is that in the future we would also have another conforming type but used for UI Tests and that's why you started it as a `protocol`? But even so, was pondering if it would not make it easier to read (debatable) and more performant (static vs dynamic dispatch) to make `MockObjectGraph` a `struct`, and have a `static let screenshotObjects = MockObjectGraph(…)` to provide the data for Screenshots (and in the future a similar one for UI Tests)

It's debatable and subjective, so 🤷  and that pondering might be triggered by the fact that the naming for `ScreenshotObjects` doesn't itself contains the term `Mock` (see below) but figured I'd start the discussion anyway.

Networking/Networking/Model/Mocks/ScreenshotsObjectGraph.swift
```diff
@@ -0,0 +1,231 @@
+import Foundation
+
+public struct ScreenshotObjects: MockObjectGraph {
```
AliSoftware, line 3: bikeshedding:
 - The name of the `struct` here doesn't match the filename
 - I think it would be nice and more explicit to have that type contain the term "Mock", e.g. "ScreenshotMocks" or "ScreenshotMocksGraph" or similar to make it more clear? (that is, if we don't decide to go with the alternative approach of making `static let screenshotsObjectGraph: MockObjectGraph = .init(…)`, see above)

WooCommerce/Classes/System/BuildConfiguration.swift
```diff
@@ -32,4 +34,8 @@ enum BuildConfiguration: String {
         BuildConfiguration.testingOverride = nil
     }
     #endif
+
+    static var shouldUseScreenshotsNetworkLayer: Bool {
+        return ProcessInfo.processInfo.arguments.contains("mocked-network-layer")
```
AliSoftware, line 39: nit: a common practice is to use the ability for Xcode and the OS to auto-inject flags passed to the command line as `UserDefaults`, see [here in Apple's doc](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/UserDefaults/AboutPreferenceDomains/AboutPreferenceDomains.html#//apple_ref/doc/uid/10000059i-CH2-96930)

TL;DR: this suggests to use `-mocked-network-layer` as part of the arguments when invoking the app (so, with an additional leading dash), and then to check it in the code here with `UserDefaults.standard.bool(forKey: "mocked-network-layer")` (without the leading dash that would have already been interpreted/consumed by the built-in argument parser) instead of using `ProcessInfo` like here.

This is not a big deal, but feels more idiomatic to how Apple suggests to do it 😉

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
```diff
@@ -70,7 +70,14 @@ class DefaultStoresManager: StoresManager {
...
         _sessionManager = sessionManager
-        self.state = AuthenticatedState(sessionManager: sessionManager) ?? DeauthenticatedState()
+
+        if BuildConfiguration.shouldUseScreenshotsNetworkLayer {
+            let screenshotObjectGraph = ScreenshotObjects()
+            self.state = MockAuthenticatedState(objectGraph: screenshotObjectGraph)
+        }
+        else {
+            self.state = AuthenticatedState(sessionManager: sessionManager) ?? DeauthenticatedState()
+        }
```
AliSoftware, line 80: This feels like the opposite of DI here, where you depend on the state of a singleton instead of using dependency injection, so wondering if we could find a better solution 🤔 

tbh not sure if there is one that doesn't involve a (too) deep refactoring, as this would mean changing the API from Yosemite anyway 🧐 . So maybe this is fine at least for the first iteration?

I mean the idea in the end if we do DI would be to pass the `StoresManagerState` in the `init` but that would also rely on the call site to be aware of which state it is supposed to create (I imagine not _any_ `StoresManagerState` might work as expected in that context?). So not sure that would be the best solution.

Another idea would be to pass some generic factory object, which provides an `authenticatedState(from: SessionManager)` method, and let the `DefaultStoresManager.init` call that method on the factory passed as parameter to get the right one. That kinds of kicks the can down the road though.

Except maybe if we consider that this factory object I'm talking about… ends up being your `MockObjectGraph` that we rename into something more generic like `ObjectFactory` – and that would be used both for mocks and for real data, with just different instances (or conforming types) provided depending on if we're running Screenshots, UITests or real app?
This solution would be more in line with the Dependency Injection principle, but also means a deeper refactoring, so I'm curious about your thoughts: do you think it would be worth it to try this approach, or is it pushing it too far?

AliSoftware (commented):
Just dropping some random ideas to get the discussion started. Mostly nits / bikesheding to propose alternative approaches, up for debate

jkmassel:
> I was thinking, if possible, we could have them in a separate Swift Package. 😁

Yep, I don't mind that at all – it could keep things nice and clean. Then we'd just `import $package` wherever we need it, so it'd live outside the `Networking` / `Storage` / `Yosemite` hierarchy?

Networking/Networking/Model/Mocks/MockObjectGraph.swift
jkmassel: Yeah! Good question. 

The reason I started with a protocol is [these lines](https://github.com/woocommerce/woocommerce-ios/pull/3266/files#diff-d6d5fd010165d9f25d5722e4615c23f775184439120e0b8bea9c99f14cd735b9R18-R26) – in the next PR, there are far more methods like this. Adding them as a protocol extension means that the `ScreenshotObjectGraph` inherits them automatically and so would a `UITestsObjectGraph`.

jkmassel: Regarding performance, this code is only ever used in a test environment, so I'm not _terribly_ worried about dispatch speed in this context, though I think this is a rare case where we can afford to deprioritize it :)

Networking/Networking/Model/Mocks/ScreenshotsObjectGraph.swift
jkmassel: >The name of the struct here doesn't match the filename

Yep you're right 😅

`ScreenshotsMockObjectGraph` might be nice and clear?

WooCommerce/Classes/System/BuildConfiguration.swift
jkmassel: Right – I always forget about that 😁 

I can implement this into the built-in `UserDefaults` wrapper code, assuming it still works from `app.launchArguments` in XCUITest

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
jkmassel: >This solution would be more in line with the Dependency Injection principle, but also means a deeper refactoring, so I'm curious about your thoughts: do you think it would be worth it to try this approach, or is it pushing it too far?

I was nervous to refactor things too much with this, so I went for the simplest possible starting approach.

I agree that it'd be nicer to clean this up – I'm personally in favour of the factory object approach (because this object _determines_ the global logged-in/out state already, so we can't inject any state into it), but we can see what @shiki suggests?

jkmassel: I was working on this some more today, and tried a couple of things (had a few dead-ends where I realized I was changing too much).

[One approach](https://github.com/woocommerce/woocommerce-ios/blob/e91456b21c5449be8bec80f0f8e353b1c65d3ba4/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift#L106-L112) allows injecting the object graph (inside the mocked state) directly from the App Delegate's [handleLaunchArguments](https://github.com/woocommerce/woocommerce-ios/blob/e91456b21c5449be8bec80f0f8e353b1c65d3ba4/WooCommerce/Classes/AppDelegate.swift#L324-L341) method.

IMHO this is a bit cleaner, but WDYT @AliSoftware / @shiki ?

shiki (commented):
@jkmassel I've only analyzed this at the top-level for now and have not dived into details yet. 

This is a nice idea! I like that it would be possible for us to get rid of WireMock in the future. 😁 

## What I Understand

This is how I understand what we're proposing here. 

image


Mainly, we're replacing the `DefaultStoresManager.state` with `MockAuthenticatedState`. And then the `MockAuthenticatedState` replaces Stores and Networking objects so that they would use the `MockObjectGraph`. 

Did I get it right? Let me know if I misunderstood. 🙂 

## Question

I think this can work. I did think of another way of doing this. Have we considered **replacing the `StoresManager`** instead? 

As far as I know, the [architecture of the app](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/YOSEMITE.md) relies heavily on using [`ServiceLocator.stores`](https://github.com/woocommerce/woocommerce-ios/blob/c7ef4dee10c40d4036c7e87760b68f80fae1561e/WooCommerce/Classes/ServiceLocator/ServiceLocator.swift#L80-L84) in communicating with Yosemite. So, we might be able to take this one level higher:

image

With this, it's possible for us to bypass `Networking` entirely and handle the dispatched `Action` ourselves. This might result in less code. 

### Prototype

I made a prototype of how this would look like [in this branch](https://github.com/woocommerce/woocommerce-ios/compare/add/injectable-mock-objects...add/injectable-mock-objects-playground-shiki). On that branch, you'll notice I replaced the `ServiceLocator.stores`:

https://github.com/woocommerce/woocommerce-ios/blob/e47ae4a2c275985b972ee5f331ddcc1df047ea74/WooCommerce/Classes/AppDelegate.swift#L230-L232

This new [`ScreenshotsStoresManager`](https://github.com/woocommerce/woocommerce-ios/blob/add/injectable-mock-objects-playground-shiki/WooCommerce/Classes/Yosemite/ScreenshotsStoresManager.swift) implements the `StoresManager` protocol. And that class will now be what the UI classes will be communicating with. It's now essentially the **Yosemite for screenshots**. 😁 

Here is an example of how we can _catch_ what the UI wanted to load: 

https://github.com/woocommerce/woocommerce-ios/blob/e47ae4a2c275985b972ee5f331ddcc1df047ea74/WooCommerce/Classes/Yosemite/ScreenshotsStoresManager.swift#L33-L46

Normally, the action showed there, `OrderAction.fetchFilteredAndAllOrders` would require an API call. But we bypassed that right then and there and simply used [our own fake data (`MockObjectGraph`)](https://github.com/woocommerce/woocommerce-ios/blob/e47ae4a2c275985b972ee5f331ddcc1df047ea74/WooCommerce/Classes/Yosemite/ScreenshotsStoresManager.swift#L87-L104). 

Running that branch, you'll see that the Orders tab will load the fake data:



And this was possible without even overriding `Networking` protocols like `ProductsRemoteProtocol`. We just went straight to saving in Core Data. That's why I concluded that this might be simpler.

### Caveat

However, this has one big hurdle. The `StoresManager` `protocol` is defined in the `WooCommerce` module and not in `Yosemite`. 🙈 I think it should be in `Yosemite` since it mostly uses Yosemite classes that we could hide. For example, I feel like the Stores classes like [`ProductStore`](https://github.com/woocommerce/woocommerce-ios/blob/c7ef4dee10c40d4036c7e87760b68f80fae1561e/Yosemite/Yosemite/Stores/ProductStore.swift#L8) do not need to be `public` at all. They should be hidden from the `WooCommerce` module. 

So, if we like this solution, we should tackle the move first. 

If we can get `StoresManager` moved to Yosemite, then we can move `ScreenshotsStoresManager` to Yosemite too. It can then use the internal utilities like [`OrdersUpsertUseCase`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Stores/Order/OrdersUpsertUseCase.swift) to interact with Core Data. 

## Moving Forward

What do you think of this option? If you agree, the next step would be asking the original author (I think Cesar) about why `StoresManager` is in `WooCommerce` and whether it's possible for us to move it to Yosemite.

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
AliSoftware: tbh I don't really like in this approach (the one you linked in the comment above) that the `DefaultStoresManager` – which is a class belonging to the app logic/model layer) has knowledge of the Mocks (and the `MockObjectGraph` type), so that `func authenticate(withObjectGraph objectGraph: MockObjectGraph) -> StoresManager` method feels a bit off to me.

Replacing both `authenticate(credentials:)` and `authenticate(withObjectGraph:)` with a unified `func authenticate(objectGraph: ObjectGraph)` – which would pull the data from that `ObjectGraph` to init both `state` and `sessionManager.defaultCredentials` no matter what that ObjectGraph comes from, as it's not the DSM's business to know the implementation details of the OG instance it received, only what data it can pull from it – would make it more SOLID and a better separation of concern imho. Then the call site can decide to pass that method either an `ObjectGraph` that contains the real credential data (when called from the app), or an `ObjectGraph` containing mocked data (when called in the context of UITests or Screenshots). The construction of the proper `ObjectGraph` the the context could still be decided at the `handleLaunchArguments` site, completing the proper usage of the ServiceLocator pattern.

What would you think about such approach @shiki ?

jkmassel:
Hey @shiki – I like this option best for sure – it's actually close to where I started, but I was trying to avoid the `Storage` layer entirely. Once I realized that all of the views relied on `FetchedResultsController` directly (AFAICT there's no way to abstract that dependency away), I abandoned it in favour of just providing the objects at the `Network` layer.

I like this better because it touches less things, and the initialization is far cleaner! Basically it sounds like exactly what I was after initially 😄

> What do you think of this option? If you agree, the next step would be asking the original author (I think Cesar) about why StoresManager is in WooCommerce and whether it's possible for us to move it to Yosemite.

I'm good with it for sure! @ctarda – WDYT? 😃

ctarda:
@jkmassel @shiki I'd (very slightly) lean towards keeping that protocol as part of the app layer, as it is supposed to abstract the global dependency the app has with Yosemite, but it looks like pushing it down to Yosemite makes lots of things easier, and it is also easy to justify that, since it abstracts out the available stores, it belongs in Yosemite. 

That's all to say, go for it! Push it down to Yosemite! 😄

shiki:
Thank you very much for responding, @ctarda! 🙏 

@jkmassel Let me know if you need some support with the move to Yosemite. 🙂

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
shiki: Heya! Since it looks like we'll be using a different approach (https://github.com/woocommerce/woocommerce-ios/pull/3266#issuecomment-738279736), let's probably discuss this later. 🙂 

Using that other approach seems to fix the issue with `DefaultStoresManager` having knowledge of the `MockObjectGraph`. And we could potentially even isolate everything related to Screenshots. 🤸

AliSoftware:
I'm gonna submit 5.6 to Apple for review later today, and this PR targeting 5.6 milestone still has conflicts and will thus likely require @jkmassel to fix them.
Since it's mainly touching the tests part, it doesn't seem to me that it's a hard requirement for it to land in 5.6 – and especially since despite being about tests/mocks it still touches the app codebase in some parts, I think it's reasonable to move its milestone to a later version.

I moved it to 5.7 for now but still 5.7 will be code-frozen on Wednesday, it's up to you @jkmassel to estimate if you think you'll have time to land this beforehand in time for 5.7, or if it's safer to postpone it further, i.e. move it to 5.8.

AliSoftware:
Postponing milestone to 5.8 then 😅

shiki:
@jkmassel I guess this PR should be closed since #3337 is the new thing?

jkmassel:
@shiki – for sure – closing in favour of #3337

PR #3267, opened 2020-12-01
Remove Duplicate Core Data Migration Logic

shiki:
Part of fixing #2667. Continuation of #3196.

Like #3196, this is another digestible chunk of the [WIP branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/2667-reduce-file-operations-in-core-data) which reduces the `FileManager` operations in our Core Data stack. The next PR will remove the `FileManager` operations and can be bloody. 

## Changes

This PR simplifies the stack by:

- Removing a duplicate logic.
- Encapsulating long routines into digestible methods.
- Throwing instead of hiding errors.

### Removing a Duplicate Logic

I found out that **both** `CoreDataManager` and `CoreDataIterativeMigrator` check if the database should be migrated or not. Here is the `CoreDataManager` logic:

https://github.com/woocommerce/woocommerce-ios/blob/5afb3bf083cd0e966b9edbb9554358a15f6dc87c/Storage/Storage/CoreData/CoreDataManager.swift#L194-L212

And here's the `CoreDataIterativeMigrator` logic:

https://github.com/woocommerce/woocommerce-ios/blob/5afb3bf083cd0e966b9edbb9554358a15f6dc87c/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L33-L49

As far as I can tell, they're both doing the same thing. I removed the `CoreDataManager` logic since I believe this logic is more apt for the `Migrator`. `CoreDataManager` will now just [trust](https://github.com/woocommerce/woocommerce-ios/blob/9a27fae9e083be16c28810277b50bd45bead8dfc/Storage/Storage/CoreData/CoreDataManager.swift#L194-L202) that the `Migrator` will do the right thing. 

### Encapsulating Long Routines into Digestible Methods

To help me be less scared of touching `CoreDataIterativeMigrator`, I moved some of the long routines like this to a method (1d964f814f03a09b9dc90707997c4ea6affa65ad):

https://github.com/woocommerce/woocommerce-ios/blob/5afb3bf083cd0e966b9edbb9554358a15f6dc87c/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L59-L88

Now, I don't need to read it all the time to figure out the whole `iterativeMigrate()` logic. 😄 

Another routine that was moved is the loading or inferring of the `NSMappingModel` (115694cf1bd392873a2d80f5e579a323d73a8a16). 

### Throwing Instead of Hiding Errors

I also noticed that we were unnecessarily hiding some of the errors in [`iterativeMigrate()`](https://github.com/woocommerce/woocommerce-ios/blob/5afb3bf083cd0e966b9edbb9554358a15f6dc87c/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L30). 

First, [loading the current database's metadata](https://github.com/woocommerce/woocommerce-ios/blob/5afb3bf083cd0e966b9edbb9554358a15f6dc87c/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L41-L43) was unnecessarily [building its own error message](https://github.com/woocommerce/woocommerce-ios/blob/5afb3bf083cd0e966b9edbb9554358a15f6dc87c/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L270-L278). And we don't even use the error message. 😬 I changed it so that it throws (6f9d993545ab3dd97bbbef528c32e3b9fbb09753):

https://github.com/woocommerce/woocommerce-ios/blob/9a27fae9e083be16c28810277b50bd45bead8dfc/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L39-L42

If `iterativeMigrate()` throws, it [will be caught](https://github.com/woocommerce/woocommerce-ios/blob/9a27fae9e083be16c28810277b50bd45bead8dfc/Storage/Storage/CoreData/CoreDataManager.swift#L211-L215) by `CoreDataManager` and it will ultimately be [uploaded to Sentry](https://github.com/woocommerce/woocommerce-ios/blob/9a27fae9e083be16c28810277b50bd45bead8dfc/Storage/Storage/CoreData/CoreDataManager.swift#L101-L109). 

Second, the routine that I mentioned above for retrieving the `NSMappingModel` also now throws if there's a problem (115694c). Throwing should be similar to the previous functionality which just [aborts everything](https://github.com/woocommerce/woocommerce-ios/blob/5afb3bf083cd0e966b9edbb9554358a15f6dc87c/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L106). 

## Testing

2. Using this branch, delete models 39 and 38. 
3. Set the current model version to 37. 
5. Stop the app. 
6. Restore the deleted model versions and reset the current model version to 39. 

    https://github.com/woocommerce/woocommerce-ios/blob/59d0f0ae842e118e8bf046800eeb7b9b5c3f7166/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L74

8. Run the app again. Migrations should happen. Confirm that the breakpoint above was hit twice.
9. Confirm that you can load the Orders and Products. 

## Submitter Checklist


Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -288,15 +242,84 @@ private extension CoreDataIterativeMigrator {
...
+    /// - Model 15
+    /// - Model 14
+    /// - Model 13
+    ///
+    /// We don't really use the descending list at the moment. I'm just keeping this logic
+    /// as is for now so I don't accidentally introduce regressions. Someday, one brave soul
+    /// will refactor this and remove the descending logic.
+    ///
+    /// - Returns: The list of models to be used for migration, including the `sourceModel` and
+    ///            the `targetModel`.
```
jaclync, line 279: nice comment! 💯

Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
```diff
@@ -260,6 +261,34 @@ final class CoreDataIterativeMigratorTests: XCTestCase {
...
+
+        // Then
+        XCTAssertTrue(result)
+
+        XCTAssertEqual(debugMessages.count, 1)
+        assertThat(try XCTUnwrap(debugMessages.first), contains: "No migration necessary.")
+
+        // There should be no file operations.
+        assertEmpty(spyFileManager.deletedItems)
+        XCTAssertEqual(spyFileManager.movedItems.count, 0)
```
jaclync: nit: any reasons this assert is different from the previous line (`XCTAssertEqual` / `assertEmpty`)?

jaclync (approved):
Nice refactoring and improvements to the migration steps with unit tests! 🧹  

Code: ✅ 
Testing: ✅

Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
shiki: Mainly because `assertEmpty` accepts an `Array` and `movedItems` is a Dictionary. So I gave up right away. 😄 

But thanks for the nudge, I changed `assertEmpty` so it accepts a `Collection`. 93ea48502c40f176c6cc40e791db340dd8d718c1 and modified this line to use `assertEmpty()` ccb007dd078012a9a4e5ef65735c36939e989f32.

shiki (commented):
Thanks for the review, @jaclync! Would you mind quickly checking the last 2 commits I added (ref: https://github.com/woocommerce/woocommerce-ios/pull/3267#discussion_r534845770) if they look good to you? 

- 93ea485
- ccb007d

Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift
jaclync: ooh I forgot that these two are different types! awesome update 💯

jaclync (approved):
Thanks for the update 💟 Still LGTM!

PR #3268, opened 2020-12-02
Order Details: update Products, Tracking, and Add Tracking sections based on shipping labels

jaclync:
Part of #2167 

## Why

Now that the shipping label sections also show order items (products/variations) in an order, we only want to show the order items & quantity that are not yet included in a non-refunded shipping label so that the store owner can still create a shipping label for the remaining items in M2. Before this PR, the Products section shows the order items combined with refunded products. This PR updates the Products section to show the **order items combined with refunded products and non-refunded shipping labels**. It also hides the tracking sections if there are non-empty non-refunded shipping labels.

## Changes

- Added a helper function to `AggregateDataHelper` to combine aggregate order items with those from non-refunded shipping labels with a test case
- Added a helper function to `AggregatedShippingLabelOrderItems` that returns an array of order items from all of the given non-refunded shipping labels
- Added an extension helper to a collection of `ShippingLabel` to return non-refunded shipping labels
- In `OrderDetailsDataSource`:
  - Updates the Products section to show the combined aggregate order items from refunded products **plus non-refunded shipping labels**
  - Hides the tracking sections if there are non-empty non-refunded shipping labels
- Implemented `GeneratedCopiable` for `AggregateOrderItem` and ran `rake generate` after removing `TestKit` to generate `Models+Copiable.generated.swift` in WooCommerce target. This is used in `AggregateDataHelper` and the test case

## Testing

There are many edge cases like with refunded products and refunded shipping labels in multi-package scenario. Feel free to test what makes sense to you!

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> after syncing, the Products section should only show the products that aren't covered by the non-refunded shipping labels (the section is hidden if there are no products to show). Also, the Tracking and Add Tracking sections should be hidden

## Example screenshots

\ | before | after
-- | -- | --
there are remaining order items | Simulator Screen Shot - iPhone 11 - 2020-12-02 at 11 56 53 | Simulator Screen Shot - iPhone 11 - 2020-12-02 at 11 54 12
no remaining order items | Simulator Screen Shot - iPhone 11 - 2020-12-02 at 11 57 04 | Simulator Screen Shot - iPhone 11 - 2020-12-02 at 11 54 36

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Copiable/Models+Copiable.generated.swift
+++ b/WooCommerce/Classes/Copiable/Models+Copiable.generated.swift
@@ -1,0 +1,41 @@
+// Generated using Sourcery 0.18.0 — https://github.com/krzysztofzablocki/Sourcery
+// DO NOT EDIT
+
+import Networking
+
+
+extension AggregateOrderItem {
+    func copy(
+        productID: CopiableProp<Int64> = .copy,
+        variationID: CopiableProp<Int64> = .copy,
+        name: CopiableProp<String> = .copy,
+        price: NullableCopiableProp<NSDecimalNumber> = .copy,
+        quantity: CopiableProp<Decimal> = .copy,
+        sku: NullableCopiableProp<String> = .copy,
+        total: NullableCopiableProp<NSDecimalNumber> = .copy,
+        imageURL: NullableCopiableProp<URL> = .copy,
+        attributes: CopiableProp<[OrderItemAttribute]> = .copy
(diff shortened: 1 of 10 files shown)


Garance91540 (approved):
Looking great!!! 👏👏👏

pmusolino (approved):
LGTM :shipit: 

Code ✅ 
Testing ✅

PR #3277, opened 2020-12-03
Order Details > Shipping Label: implement shipping label details UI

jaclync:
Part of #2167 

## Why

The user can now tap on the "View Shipment Details" row in a shipping label card to view the details about a shipping label.

## Changes

Ref: p91TBi-3Lo-p2

- Added two new icons from Gridicons: `.house` and `.creditCard` for the details screen
- In `OrderDetailsDataSource`, updated `shippingLabel(at:)` to be public for tap action handling in `OrderDetailsViewModel`
- In `OrderDetailsViewModel`, showed `ShippingLabelDetailsViewController` when tapping on the shipping label details row
- Created `ShippingLabelDetailsViewController` and `ShippingLabelDetailsViewModel` that displays a table view of shipping label details

## Testing

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/) and there is an order with at least one shipping label (refunded or not).

- Go to the orders tab
- Tap on an order with at least one shipping label
- After a shipping label card is shown, tap on "View Shipment Details" --> it should navigate to a table view of 5 rows: origin address, destination address, package details (package name), shipping carrier and rates (the rate is formatted based on site settings and currency from the label), payment method (hard-coded to "Credit Card")

## Example screenshots

Notes for design reviewers - there are two places that deviate from the iOS design:
- The destination address icon uses a solid "house" icon from [Gridicons](http://automattic.github.io/gridicons/) as I couldn't find an outlined version
- For the origin/destination address content, I followed Android to show the name and multiline formatted address (the iOS design shows a comma-separated address). Using the multiline address formatted by iOS Contacts framework ensures the order follows the country convention (you can see that the origin address from the US has a different order from the destination address to Taiwan)

dark | light
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-03 at 10 31 37 | Simulator Screen Shot - iPhone 11 - 2020-12-03 at 10 31 54

Update release notes:


ctarda (approved):
Code is 💯 

This is how it looks for me:

Garance91540 (approved):
Hi @jaclync , thank you for this and highlighting what deviates from the designs! Regarding the House icon, we'll need to add the outlined icon to our Gridicon library so we can use it. Do you know how we can do this?

jaclync:
> Regarding the House icon, we'll need to add the outlined icon to our Gridicon library so we can use it. Do you know how we can do this?

@Garance91540 I'll ask in the async meeting channel, and I added this to the pending design items! from the [README in Gridicons](https://github.com/Automattic/gridicons#add-a-proposed-icon-to-gridicons-admins-only) it looks like it might require working with Calypso designers to have a new icon if they think it's useful? In any case, if that doesn't work out we can export the asset to a PDF and add it to the project.

Issue #2113, opened 2020-04-09
Order Details > Products: the image isn't shown for product variations


jaclync:
When an order includes a product variation purchase, the image isn't shown in Order Details > Product section for the item even though the variation does have an image (example screenshots below).

I looked into the issue for a bit, and it seems like there are 2 issues:

## Issue 1

The first issue is where we request the missing products for an order - currently, we only request missing products but not product variations.

https://github.com/woocommerce/woocommerce-ios/blob/9971547b8c2c7e76ce09ff5035c008bc7282ba8d/Yosemite/Yosemite/Stores/ProductStore.swift#L114-L127

## Issue 2

When fetching the missing products (`ProductAction.requestMissingProducts`), we make an API request `wc/v3/products?include={{list_of_product_or_variation_ids}}`.  However, the `wc/v3/products`'s `include` parameter with a list of remote IDs only includes posts with post type (`post_type` in the database) `product` while the variations are of `product_variation` type. Therefore, we cannot use the same `wc/v3/products` endpoint to request missing product variations.

I think we can fix this by:
- Add a `ProductVariationAction` to request missing product variations similar to the `ProductAction` used to fetch missing products (that aren't in the storage yet): `wc/v3/products/{{product_id}}/variations?include={{list_of_variation_ids}}` - it seems like we need both the product & product variation ID(s) to make a request per product, so there could be multiple API requests for the missing variations.
- Similar to https://github.com/woocommerce/woocommerce-ios/blob/9971547b8c2c7e76ce09ff5035c008bc7282ba8d/Yosemite/Yosemite/Stores/ProductStore.swift#L114, with the difference that we need to know both the product and product variation ID.

All product variations | Order details with 2 product variations
-- | --
![Simulator Screen Shot - iPhone 11 - 2020-04-09 at 12 52 05](https://user-images.githubusercontent.com/1945542/78859218-4b2edc00-7a61-11ea-9d8d-755d8ec0a4cf.png) | ![Simulator Screen Shot - iPhone 11 - 2020-04-09 at 12 52 08](https://user-images.githubusercontent.com/1945542/78859228-508c2680-7a61-11ea-8c13-ea21065fadae.png)

pmusolino:
Similar issue here: https://github.com/woocommerce/woocommerce-ios/issues/1945

shiki:
I can confirm that this still happens on version 5.6.

PR #3278, opened 2020-12-03
Order Details: sync missing variations in order details

jaclync:
Part of #2167

## Why

Before this PR, we only sync products in order details. Since some order items could be variations instead of products, these order items would never be loaded until the user goes to another screen that syncs the variations (e.g. visiting a variable product's variations page). This PR syncs product variations in order details so that variation order items in shipping label cards are synced for UI display. 

## Changes

- Added `ProductVariationAction.requestMissingVariations` for requesting missing variations in an order in `ProductVariationStore` with test cases
  - Unfortunately, I didn't find an API endpoint to request a list of variations so the store makes an API request for each order item that misses its variation and signals completion when all requests finish. Normally, an order wouldn't have a lot of order items
- Synced missing variations in `OrderDetailsViewModel` and `OrderDetailsViewController` following when `syncProducts` is called

## Testing

- In core, create an order with a variation. Then create a shipping label for this order
- Launch the app, log out and in to the store without going to the Products tab
- Go to the orders tab
- Tap on the order with a variation & shipping label --> after syncing, the variation order item should show its price, image (if any), and total

## Example screenshots 

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-03 at 10 51 15 | Simulator Screen Shot - iPhone 11 - 2020-12-03 at 13 55 27

 
Update release notes:


Yosemite/YosemiteTests/Stores/ProductVariationStoreTests.swift
```diff
@@ -455,7 +455,7 @@ final class ProductVariationStoreTests: XCTestCase {
 
     /// Verifies that `ProductVariationAction.updateProductVariation` returns an error whenever there is an error response from the backend.
     ///
-    func testUpdatingProductVariationReturnsErrorUponReponseError() {
+    func testUpdatingProductVariationReturnsErrorUponResponseError() {
```
pmusolino: Do it worth renaming the method using the underscore?

pmusolino (approved):
Looks good to me 💯 



**I've only a question**: it seems that the sync of product variations inside an order detail, affect only the products inside the Package, but not the ones inside the `Products` section, as you can see from the screen below. Shouldn't these be updated too?

Garance91540 (approved):
👏👏👏

shiki:
@jaclync, will this also fix https://github.com/woocommerce/woocommerce-ios/issues/2113?

jaclync:
> Whether this PR fixes the missing variation order item images in the Products section

@pmusolino @shiki Great question! No, this PR does not fix the missing image in the variation order items in the Products section but now I have a better idea to fix #2113. For the order items in the Products section, it currently only tries finding the matching `Product` but not the `ProductVariation` in and the `imageURL` is from the optional `Product`.

https://github.com/woocommerce/woocommerce-ios/blob/97269f4cbdba00015524da51eff5b9dccac81b0c/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsDataSource.swift#L552-L554

I'm planning to fix this by finding the matching `ProductVariation` if the order item is associated with a variation. will work on #2113 this sprint!

Yosemite/YosemiteTests/Stores/ProductVariationStoreTests.swift
jaclync: yup updated in: https://github.com/woocommerce/woocommerce-ios/pull/3278/commits/f436d3e9aefd1ff68f23e1cc89639c70b39fd44f

PR #3282, opened 2020-12-03
Set correct, relative path to UnitTests test plan

rachelmcr:
The Xcode project file accidentally had the UnitTests test plan set to an absolute file path (on my computer) instead of the test plan file in the project. This change corrects that to the file `WooCommerceTests/UnitTests.xctestplan`, with the file path relative to the project.

The scheme already [pointed to the correct test plan file](https://github.com/woocommerce/woocommerce-ios/blob/8b229f9fdb7b538f9802ac51934cc85e66ad4e21/WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce.xcscheme#L41) so I don't think this was causing any functional issues. We should just confirm that all tests continue to run and pass after selecting and running the unit test plan, including tests in:

* WooCommerceTests
* NetworkingTests
* StorageTests
* YosemiteTests

h/t @eshurakov for noticing the absolute file path and bringing it to my attention!

Update release notes:


eshurakov (approved):
Looks good! Tried to run tests locally, everything works 👍 

:shipit:

PR #3285, opened 2020-12-04
Order Details > Shipping Label: present shipping label printing instructions screen

### LINKED ISSUE #2174: Shipping Labels M1: UI to display instructions on how to print shipping labels
As part of the M1 release, we need to add a information dialog to the order detail screen on how to print a shipping label from a device.

. Screenshot 2020-04-22 at 11 28 16 AM

jaclync:
Fixes #2174 

## Why

For users who haven't printed from an iOS device, we show a screen with instructions on how to do so with 4 steps.

## Changes

Note: please check p91TBi-3V3-p2 for us to discuss on SwiftUI adoption.

- Added two assets for `woo-shipping-label-printing-instructions` in light and dark mode
- Created 3 SwiftUI views:
  - `BoldableTextView`: this view shows a boldable string (as shown in screenshots), and has the most logic of all views in this PR. I looked into how to show text with different weights, and [this blog post](https://www.avanderlee.com/swiftui/text-weight-combinations/) best summarizes my findings and my implementation is based on the solution in the blog post. The solution in the blog post has some bugs handling space, and I changed and cleaned up quite a bit. I also added unit tests for different string content scenarios
  - `ShippingLabelPrintingStepView`: displays one step of printing instructions given its `ViewModel`
  - `ShippingLabelPrintingInstructionsView`: the top-level view that consists of an illustration at the top and four steps (`ShippingLabelPrintingStepView`). the previews show the view in light and dark mode
- In `OrderDetailsViewModel`, presented `ShippingLabelPrintingInstructionsViewController` when tapping on the shipping label printing info row

## Testing

- Go to the orders tab
- After a shipping label card shows up, tap on the info row "Don't know how to print from your phone?" --> a screen should be presented that displays 4 steps on how to print a shipping label from an iOS device
- Tap on the close button ("X") --> the instructions screen should be dismissed

## Example screenshots

\ | dark | light
-- | -- | --
iPhone 11 - portrait | Simulator Screen Shot - iPhone 11 - 2020-12-04 at 11 39 00 | Simulator Screen Shot - iPhone 11 - 2020-12-04 at 11 39 14
iPhone 11 - landscape | Simulator Screen Shot - iPhone 11 - 2020-12-04 at 11 39 04 | Simulator Screen Shot - iPhone 11 - 2020-12-04 at 11 39 19
iPad Air gen 3 | Simulator Screen Shot - iPad Air (3rd generation) - 2020-12-04 at 11 41 30 | Simulator Screen Shot - iPad Air (3rd generation) - 2020-12-04 at 11 41 58

Update release notes:


WooCommerce/WooCommerceTests/ViewRelated/SwiftUI/BoldableTextViewTests.swift
```diff
@@ -0,0 +1,69 @@
...
+@testable import WooCommerce
+
+final class BoldableTextViewTests: XCTestCase {
+    func test_content_without_matching_asterisks() throws {
+        // When
+        let view = BoldableTextView("I will *never stop learning.")
+
+        // Then
+        let text = try XCTUnwrap(view.body as? Text)
+        let expectedText = Text("") + Text(verbatim: "I will *never stop learning.")
```
jaclync: This got me stuck for a while: in SwiftUI, `Text(content)` where `let content = "string"` is not equal to `Text("string")`. To compare to a static string, I had to use `Text(verbatim: "string")` 🤯 

Also, because `BoldableTextView` reduces a list of boldable elements with an initial `Text("")`, `Text("")` is needed here.

Garance91540 (approved):
👏

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingInstructionsView.swift
```diff
@@ -0,0 +1,51 @@
...
+            "*Ensure AirPrint is enabled* in your printer settings. You may need to configure this setting on the printer itself.\n\n"
+            + "See the documentation that came with your printer for details.",
+            comment: "Step 2 of shipping label printing instructions screen. The content inside two asterisks *...* denote bolded text.")
+        static let step3 = NSLocalizedString(
+            "Ensure that your *printer firmware is up to date*. See your printer documentation for instructions on updating.",
+            comment: "Step 2 of shipping label printing instructions screen. The content inside two asterisks *...* denote bolded text.")
+        static let step4 = NSLocalizedString(
+            "If you are still experiencing issues printing from your phone, you can save your label as PDF and *send it by email* "
+            + "to print it from another device.",
+            comment: "Step 2 of shipping label printing instructions screen. The content inside two asterisks *...* denote bolded text.")
```
eshurakov: Nitpicking: comment for step 3 and 4 mentions step 2 :-)

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,113 @@
...
+            if isBold {
+                self.content = content.replacingOccurrences(of: boldRegionMark, with: "")
+            } else {
+                self.content = content
+            }
+            self.isBold = isBold
+        }
+    }
+
+    fileprivate static let boldRegionMark = "*"
```
eshurakov: Just thinking if it's better to use `**` to mark bold text. It would be similar to markdown and won't have problems if someone decides to use `*` inside the localisation string for other purposes.

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,113 @@
...
+}
+
+private extension String {
+    /// Parses the input text and returns a collection of boldable elements.
+    ///
+    /// - Returns: A collection of boldable elements.
+    func parseBoldableElements() -> [BoldableTextView.Element] {
+        let boldRegionMark = BoldableTextView.boldRegionMark
+        let regex = try! NSRegularExpression(pattern: "\\\(boldRegionMark){1}(.*?)\\\(boldRegionMark){1}")
+        let range = NSRange(location: 0, length: count)
```
eshurakov: `String` can use different representations of the text that it stores (utf8 / utf16) and `count` may differ from `NSString` length. 

`NSRange` has an initialiser that takes range and swift string. I think it's safer to use :-)
```
extension _NSRange {
    public init<R, S>(_ region: R, in target: S) where R : RangeExpression, S : StringProtocol, R.Bound == String.Index
}
```

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,113 @@
...
+
+        /// Finds all the ranges that match the regex for content within a bold region.
+        let matches: [NSTextCheckingResult] = regex.matches(in: self, options: [], range: range)
+        let matchingRanges = matches.compactMap { Range<Int>($0.range) }
```
eshurakov: `Range` has a constructor that can convert from `NSRange` to the native range of `String`

```
Range(range: NSRange, in: String)
```

I believe if we operate with this range we don't need a `subscript(range: Range<Int>) -> String` method and can use `substring` method of a string.

WooCommerce/WooCommerceTests/ViewRelated/SwiftUI/BoldableTextViewTests.swift
```diff
@@ -0,0 +1,69 @@
+import SwiftUI
+import XCTest
+@testable import WooCommerce
+
+final class BoldableTextViewTests: XCTestCase {
```
eshurakov: Just a thought: testing views are usually quite tricky. I guess that the main intention of the test here is to see if correct areas of provided text are styled as bold and for that we don't necessarily need to have a view. What do you think about extracting the code from `parseBoldableElements` to a separate object and test it?
And the role of the view would be the same as now, to combine all the Elements( or text fragments ) together applying provided style.

eshurakov (commented):
Looks nice! Sending a few comments your way!

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,113 @@
...
+/// For example, initializing with a string "*Never* will I *stop learning*." will bold the substrings "Never" and "stop learning".
+///
+struct BoldableTextView: View {
+    fileprivate struct Element {
+        let content: String
+        let isBold: Bool
+
+        init(content: String, isBold: Bool) {
+            if isBold {
+                self.content = content.replacingOccurrences(of: boldRegionMark, with: "")
```
eshurakov: One more improvement can be to pass text already without `boldRegionMark`. Our regular expression already has a matching group just for content `(.*?)`. We can leverage on that.

`NSTextCheckingResult` has an api to get a range at a certain index. Each of our matching results should have 2 ranges available: one for the whole match (that we use now) and one just for the content without the region mark. If we use that range to get the actual content, we don't need to replace occurrences later in the code.

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -1,4 +1,5 @@
 import Foundation
+import SwiftUI
```
ctarda: Is this import really necessary here?

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    /// - Returns: A string subscript based on the given range.
+    subscript(range: Range<Int>) -> String {
+        let startIndex = index(self.startIndex, offsetBy: range.lowerBound)
+        let endIndex = index(self.startIndex, offsetBy: range.upperBound)
+        return String(self[startIndex..<endIndex])
+    }
+}
+
+struct RichTextView_Previews: PreviewProvider {
```
ctarda: I would suggest considering separating previews to a different file.

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,113 @@
+import SwiftUI
+
+/// A text view that takes in a string whose substrings within two asterisks are bolded.
+///
+/// For example, initializing with a string "*Never* will I *stop learning*." will bold the substrings "Never" and "stop learning".
+///
+struct BoldableTextView: View {
+    fileprivate struct Element {
```
ctarda: I am curious. What would be the benefit of declaring a sort of model object in the namespace of a view?

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,113 @@
...
+
+    var body: some View {
+        elements.map { $0.toTextView }
+            .reduce(into: Text(""), { result, text in
+                result = result + text
+            })
+    }
+}
+
+private extension BoldableTextView.Element {
```
ctarda: Maybe Element could / should live in a view model/presenter? That way this logic, plus the little bit of logic above that maps elements and then reduces plus the string manipulation could be part of the view model as well? 

That way, in theory, the view could be completely free of logic. What do you think?

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingInstructionsView.swift
```diff
@@ -0,0 +1,51 @@
...
+/// Displays an illustration at the top and four steps on how to print a shipping label on an iOS device.
+struct ShippingLabelPrintingInstructionsView: View {
+    var body: some View {
+        ScrollView {
+            VStack {
+                Spacer()
+                    .frame(height: 20)
+                Image("woo-shipping-label-printing-instructions")
+                Spacer()
+                    .frame(height: 30)
```
ctarda, line 12: Given that creating views is supposed to be almost free (according to Apple), would it make sense to extract all the step views to a separate view? That way we would also remove the knowledge of the view model creation from here

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingInstructionsView.swift
```diff
@@ -0,0 +1,51 @@
...
+                ShippingLabelPrintingStepView(viewModel: .init(stepIndex: 1, text: Localization.step1, isLastStep: false))
```
ctarda: I understand using .init here makes the code very compact, but, to me, unless you check the code of ShippingLabelPrintingStepView to see what type of parameter it accepts, it makes difficult to understand what is happening, and what type those view models represent. What do you think?

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingStepView.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    init(viewModel: ViewModel) {
+        self.viewModel = viewModel
+    }
+
+    var body: some View {
+        VStack(alignment: .leading) {
+            Spacer()
+                .frame(height: 14)
+            HStack(alignment: .top, spacing: 11) {
```
ctarda: Would it make sense to extract this Stack to its own view?

ctarda (commented):
This is a great step forward @jaclync ! Thanks for pioneering the SwiftUI work!

It looks great to me, but I left a few questions, I am curious about your opinion.

In general, I would say that we could benefit from making some of the views a bit more granular, breaking them down in smaller parts. After all Apple suggest breaking views down in composable views as much as possible, and, at least in theory, more views with semantic names, should make the code of the views containing them easier to read

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingInstructionsView.swift
jaclync: Nice catch! updated in https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/625b8ce9ec46ae331593a3db20a799ace33ac6bc 😄

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: I agree, `**` matches the markdown syntax more and allows usage of single matching asterisk like to highlight some substring. Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/78d1ea47625e0becf31e7242d76036c35516120d

WooCommerce/WooCommerceTests/ViewRelated/SwiftUI/BoldableTextViewTests.swift
jaclync: Yea I agree on testing the logic rather than views directly (I felt the tests on `BoldableTextView` quite hacky). I moved the text parsing logic to a separate struct `BoldableTextParser` in https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/78d1ea47625e0becf31e7242d76036c35516120d

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: Great point! TIL'ed the 2 ranges in a `NSTextCheckingResult` - updated in https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/78d1ea47625e0becf31e7242d76036c35516120d 😃

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: It looks like `substring` is deprecated in Swift 5, and the range has to be in the index form like what the `String.subscript` helper currently does (also suggested in [this SO response](https://stackoverflow.com/a/46634511/9185596)) - do you have other suggestions on the replacement of `String` extension `subscript(range: Range<Int>) -> String`?

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: I didn't think of the potential difference in `count` between `String` and `NSString`! updated here: https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/b6ada23fe642cb118d1d635c280ec10d9ce79850

WooCommerce/Classes/Tools/BoldableTextParser.swift
```diff
@@ -0,0 +1,77 @@
...
+}
+
+private extension String {
+    /// - Returns: A string subscript based on the given range.
+    subscript(range: Range<Int>) -> String {
+        let startIndex = index(self.startIndex, offsetBy: range.lowerBound)
+        let endIndex = index(self.startIndex, offsetBy: range.upperBound)
+        return String(self[startIndex..<endIndex])
+    }
+}
```
eshurakov: > It looks like substring is deprecated in Swift 5, and the range has to be in the index form like what the String.subscript helper currently does (also suggested in this SO response) - do you have other suggestions on the replacement of String extension subscript(range: Range<Int>) -> String?

Ah, true, `substring` is deprecated. Because `NSTextCheckingResult` returns ranges as `NSRange` we can't directly use it with `String`, but we can convert it to swift `Range`. This will enable us to do all range calculations and later get a substring without index conversion.

Below is a suggestion that works without String extension.

```suggestion
        var startingIndex = string.startIndex
        matches.enumerated().forEach { index, match in
            // Each `NSTextCheckingResult` has two ranges: the first one (index 0) is the range that includes the bold marks (**..**).
            // The second one (index 1) is the range that excludes the bold marks, which corresponds to the bolded substring.
            guard match.numberOfRanges == 2 else {
                return
            }

            guard let matchingRange = Range(match.range(at: 0), in: string),
                  let boldedContentRange = Range(match.range(at: 1), in: string) else {
                return
            }

            // Adds an element for the range from the last index and before the matching range, if non-empty.
            let rangeBefore = startingIndex..<matchingRange.lowerBound
            if rangeBefore.isEmpty == false {
                let content = String(string[rangeBefore])
                elements.append(.init(content: content, isBold: false))
            }

            // Adds an element for the matching range which should be bold.
            let matchContent = String(string[boldedContentRange.lowerBound..<boldedContentRange.upperBound])
            elements.append(.init(content: matchContent, isBold: true))

            // Updates the last starting index so that the next range starts after the end of current matching range.
            startingIndex = matchingRange.upperBound

            // If the matching range is the last one, adds an element for the range after the last matching range, if non-empty.
            let isLastMatchingRange = match == matches.last
            if isLastMatchingRange {
                let lastRange = matchingRange.upperBound..<string.endIndex
                if lastRange.isEmpty == false {
                    let content = String(string[lastRange])
                    elements.append(.init(content: content, isBold: false))
                }
            }
        }
        return elements
    }
}
```

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: what do you think are the benefits of having the previews in a different file? before this PR, we already have a few custom `UIView`s that we add SwiftUI previews in the same file (like `InAppFeedbackCardViewController`, `PrimarySectionHeaderView`, etc.)

while developing the SwiftUI views, I also found it convenient to have the preview code in the same file to edit preview data & preview without leaving the window. curious to hear other thoughts 🙂

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: for private/fileprivate models, it probably doesn't matter much and it could be a separate model. I'd say it depends on the use case and scope, like for `ViewModel` it sometimes makes sense to have it inside a small struct/class like for a `UITableViewCell` while it might be better having its own struct/class when it is for a view controller.

I refactored the code here to a separate struct `BoldableTextParser` with a separate model struct `BoldableElement` that is public in https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/78d1ea47625e0becf31e7242d76036c35516120d 😄

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: The element model is now a separate struct `BoldableElement` in the same file as `BoldableTextParser`! feel free to comment on anything on the latest code 🙂

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingInstructionsView.swift
jaclync: hmm interesting point, the main reason I used `.init` here is because the parameter name already suggests a "view model" so it's most likely a view model for `ShippingLabelPrintingStepView` (and usually called `*.ViewModel` or `*ViewModel`). in code, we can click into the `.init` to see the view model as well and it could save a few lines of code if the view model name is long (and save time for reformatting if we ever refactor the name of the view model). I generally find it easier to read with `.init` rather than the full class/struct name to focus on the parameters for the view model, but I could see the point of not knowing the type.

@woocommerce/ios-developers any thoughts on this?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
jaclync: nope! removed in https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/d08546c510b5e062965d3fe49429b2a171ab238f

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingInstructionsView.swift
jaclync: Yea it makes sense to keep the top-level view cleaner, updated in: https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/b2df64348e583313ce738431b2e01ebbaf30dc2d

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingStepView.swift
jaclync: Hmm for this one, I'm still debating 🤔 My current logic for whether to create a separate SwiftUI view is its reusability - if it's not reusable by other views (not really like a "component"), it might be easier to read when its UI configuration in one place. Complexity of a view might also be a factor, but this could also be subjective. Curious to hear what the team thinks about when to make a separate SwiftUI view @woocommerce/ios-developers !

WooCommerce/Classes/Tools/BoldableTextParser.swift
jaclync: thanks so much @eshurakov for your suggestions! I didn't have extensive experience with handling `String` and ranges and I really appreciate your insights 🙏 I applied your suggestions here https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/da95e62728f7c19b2f138986110369a0d084b338!

jaclync:
@ctarda @eshurakov thank you both for the feedback, I updated/responded to the PR for another look! 😄

WooCommerce/Classes/Tools/BoldableTextParser.swift
```diff
@@ -0,0 +1,69 @@
...
+
+            // Adds an element for the range from the last index and before the matching range, if non-empty.
+            let rangeBefore = startingIndex..<matchingRange.lowerBound
+            if rangeBefore.isEmpty == false {
+                let content = String(string[rangeBefore])
+                elements.append(.init(content: content, isBold: false))
+            }
+
+            // Adds an element for the matching range which should be bold.
+            let matchContent = String(string[boldedContentRange.lowerBound..<boldedContentRange.upperBound])
```
eshurakov: Nitpick: I believe can just use `String(string[boldedContentRange])` here :-)

WooCommerce/Classes/Tools/BoldableTextParser.swift
```diff
@@ -0,0 +1,69 @@
...
+    private let boldTextPattern = "\\*{2}(.*?)\\*{2}"
+
+    /// Parses the input text and returns a collection of `BoldableElement`.
+    ///
+    /// - Returns: A collection of boldable elements.
+    func parseBoldableElements(string: String) -> [BoldableElement] {
+        guard let regex = try? NSRegularExpression(pattern: boldTextPattern) else {
+            return []
+        }
+        let range = NSRange(string.startIndex..., in: string)
```
eshurakov, line 20: Nice!
TIL that you can create a range like this. I always used both upper and lower bounds before.

WooCommerce/Classes/Tools/BoldableTextParser.swift
```diff
@@ -0,0 +1,69 @@
...
+/// Parses boldable text for further text concatenation with bold attributes.
+struct BoldableTextParser {
+    private let boldTextPattern = "\\*{2}(.*?)\\*{2}"
+
+    /// Parses the input text and returns a collection of `BoldableElement`.
+    ///
+    /// - Returns: A collection of boldable elements.
+    func parseBoldableElements(string: String) -> [BoldableElement] {
+        guard let regex = try? NSRegularExpression(pattern: boldTextPattern) else {
+            return []
```
eshurakov: Nitpick: maybe can return the whole string here.

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
```diff
@@ -0,0 +1,47 @@
+import SwiftUI
+
+/// A text view that takes in a string whose substrings within two asterisks are bolded.
+///
+/// For example, initializing with a string "*Never* will I *stop learning*." will bold the substrings "Never" and "stop learning".
```
eshurakov: Nitpick: can change * to ** in the comment

eshurakov (approved):
Looks great!
Sent a few nitpick comments :-)

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingStepView.swift
ctarda: I don't think reusability should be a concern here. Apple explicitly suggest breaking views down as much as possible, as it is pretty much free, and it helps with readability, which is a concern when it comes to SwiftUI

WooCommerce/Classes/Tools/BoldableTextParser.swift
jaclync: nice! updated here: https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/83595186048b3262514e558b9ab386af72af1a61

WooCommerce/Classes/Tools/BoldableTextParser.swift
jaclync: makes sense! updated in https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/83595186048b3262514e558b9ab386af72af1a61

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/BoldableTextView.swift
jaclync: great catch 😬 updated in: https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/83595186048b3262514e558b9ab386af72af1a61

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPrintingStepView.swift
jaclync: Sure! It was hard for me to tell which views are "complicated" and should be broken down, it's nice you pointed it out for readability 😄 

In https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/37152c4e00de749bf6d97f3d505d4dec90100268, I separate the `HStack` to `ShippingLabelPrintingStepContentView` (lemme know if you have other naming suggestions), and also removed the `ViewModel` per discussion in https://github.com/woocommerce/woocommerce-ios/pull/3322#discussion_r540281521. Also, in https://github.com/woocommerce/woocommerce-ios/pull/3285/commits/7fdcd60e813f1224d04cbc3a8fb686f0b0ac0a53 I renamed the plural steps view to `*StepListView` based on [your feedback](https://github.com/woocommerce/woocommerce-ios/pull/3322#discussion_r542012968) in another PR 🙂

jaclync:
@eshurakov @ctarda I updated the PR and responded to your comments! The more open one is probably the response to @ctarda's comment on SwiftUI: https://github.com/woocommerce/woocommerce-ios/pull/3285#discussion_r542246905 - lemme know how it looks!

jaclync:
Merging the PR now after resolving merge conflicts from `develop`! Feel free to comment on anything after merge 😄

PR #3286, opened 2020-12-04
Block sign ups to WordPress.com accounts.

### LINKED ISSUE #3146: [UL&S] Prevent signups to wp.com
1. Tap "Continue with wordpress.com"
2. Enter an email that does not match a wordpress.com account

At this moment, we present this screen:

However, we should present this one:

ctarda:
Closes #3146 

|  | Light | Dark |
| ---- | ---- | --- |
| Error screen |  |  |
| Error screen + alert |  |  |

And a gif showing the interactions:

I am not sure if the actual link that can be opened in a web view from the alert is correct. Changing that takes 30 seconds.

## Changes
* Pointed WPAuthenticator to the branch that provides support for short-circuiting the navigation flow provided by WPAuthenticator, and allows us to block the account creation flow.
* A new implementation of the ULErrorViewModel that provides support for this new error screen.
* A couple of minor touches to the constraints in the actual UI, so that the button that is styled as a link can stretch horizontally to occupy all the available space.

## Testing
* Checkout the branch. Run `bundle exec pod install`
* Log out if necessary
* Select Continue With WordPress.com
* Enter an email address that does not match a WP.com account. 
* You should see the custom error UI instead of the flow that allows creating a new WP.com account

Update release notes:


Garance91540 (approved):
👏👏👏  The only thing I mentioned on slack is that the link "need help finding your email" should open the Help center

ctarda:
Thanks for the review @Garance91540 

I just pushed an update to:
* Point WPAuthenticator to the latest beta
* Present the Help when tapping "need more help" in the alert:



I guess this is ready for a close look at the code 😊

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -342,18 +341,39 @@ private extension AuthenticationManager {
 
     /// Maps error codes emitted by WPAuthenticator to a domain error object
     enum AuthenticationError: Int, Error {
+        case emailDoesNotMatchWPAcount = 7
```
jaclync: nit:
```suggestion
        case emailDoesNotMatchWPAccount = 7
```

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -342,18 +341,39 @@ private extension AuthenticationManager {
...
                 return .notWPSite
             default:
                 return .unknown
             }
         }
     }
+
+    func isSupportedError(_ error: Error) -> Bool {
+        let wooAuthError = AuthenticationError.make(with: error)
+        return wooAuthError == .emailDoesNotMatchWPAcount || wooAuthError == .notWPSite
```
jaclync: nit: how about just checking if the error is not `unknown`, since we handle all the known cases?
```suggestion
        return wooAuthError != .unknown
```

WooCommerce/Classes/Authentication/Navigation Exceptions/NotWPAccountViewModel.swift
```diff
@@ -0,0 +1,63 @@
...
+import WordPressUI
+
+
+/// Configuration and actions for an ULErrorViewController, modeling
+/// an error when user attempts to log in with an invalid WordPressAccount
+struct NotWPAccountViewModel: ULErrorViewModel {
+    // MARK: - Data and configuration
+    let image: UIImage = .loginNoWordPressError
+
+    let text: NSAttributedString = NSMutableAttributedString(string: Localization.errorMessage)
```
jaclync: nit: any reasons for using the mutable version?

```suggestion
    let text: NSAttributedString = .init(string: Localization.errorMessage)
```

jaclync (approved):
Looks good!

Code: ✅ 
Testing: ✅ (all CTAs work as expected on the error screen!)

I noticed a non-blocking UI issue in landscape:

WooCommerce/Classes/Authentication/AuthenticationManager.swift
ctarda: I went with the long form because it seemed more explicit to me, but will update, thanks!

ctarda:

I pushed an update to incorporate your feedback. 

This is blocking progress a little bit, so I will defer the layout bug on landscape to a separate issue that I will attack later this week: #3300

WooCommerce/Classes/Authentication/Navigation Exceptions/NotWPAccountViewModel.swift
ctarda: Nope, good point, thanks!

PR #3287, opened 2020-12-04
Remove duplicated line in TopBannerView

ealeksandrov:
Related to https://github.com/woocommerce/woocommerce-ios/issues/1549.

## Description

No functional changes, just duplicate line removed. This line was adding array of buttons to stack view before iterating and adding same buttons one by one.

Update release notes:


pmusolino (approved):
LGTM, thanks for fixing this! :shipit:

Issue #3181, opened 2020-11-20
Variations: List Existing Attributes


shiki:
Partially depends on #3180 and #3194. 

Refs p91TBi-3Oc-p2, p91TBi-2Ur-p2.

<img src="https://user-images.githubusercontent.com/198826/99741223-d78c8d00-2a8d-11eb-9579-3647e387f8dd.png" width="500">

When tapping on “Add variations”, display an “Add attributes” screen. 

## Goals

- [ ] Display existing attributes of the store. This includes store-level (global) and product-level attributes. 
- [ ] Synchronize the attributes if we don't do that somewhere already. 

The flow of adding a new attribute can be done later. 

---

Android: https://github.com/woocommerce/woocommerce-android/issues/3207

PR #3290, opened 2020-12-04
Networking layer for List/Add/Update/Delete a Product Attribute

pmusolino:
Part of #3181 

It seems that the **Add Variable Products** functionality needs a lot of API requests that should be implemented on our side.
In this PR, I implemented the first four API requests inside the networking layer for Listing, Adding, Updating and Deleting the global [Product Attributes](https://woocommerce.github.io/woocommerce-rest-api-docs/#product-attributes) using the endpoint `/wp-json/wc/v3/products/attributes`. The listing of the Product Attributes will be shown inside the screen that will be implemented for #3181.

The Yosemite and Storage layers changes will be submitted in another PR.

## Changes
- Implemented `ProductAttributeMapper` + tests for mapping a single `ProductAttribute` entity.
- Implemented `ProductAttributeListMapper` + tests for mapping a list of `ProductAttribute` entities.
- Implemented `ProductAttributesRemote` + tests for the networking requests.
- Added some JSON files for mock purposes.

Note: this PR surpassed the 500 LOC because of the JSON files that I added.

## Testing
- just CI, since we don't use them inside the code at the moment.

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/ProductAttributesRemote.swift
+++ b/Networking/Networking/Remote/ProductAttributesRemote.swift
@@ -13,7 +13,7 @@ public final class ProductAttributesRemote: Remote {
     ///     - completion: Closure to be executed upon completion.
     ///
     public func loadAllProductAttributes(for siteID: Int64,
-                                         completion: @escaping ([ProductAttribute]?, Error?) -> Void) {
+                                         completion: @escaping (Result<[ProductAttribute], Error>) -> Void) {
 
         let path = Path.attributes
         let request = JetpackRequest(wooApiVersion: .mark3, method: .get, siteID: siteID, path: path)
--- a/Networking/Networking/Remote/ProductAttributesRemote.swift
+++ b/Networking/Networking/Remote/ProductAttributesRemote.swift
@@ -1,0 +1,121 @@
+import Foundation
+
+/// Product Attributes: Remote Endpoints for fetching global product attributes.
+///
+public final class ProductAttributesRemote: Remote {
+
+    // MARK: - Product Attributes
+
+    /// Retrieves all of the global`ProductAttribute` available.
+    ///
+    /// - Parameters:
+    ///     - siteID: Site for which we'll fetch remote products attributes.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func loadAllProductAttributes(for siteID: Int64,
+                                         completion: @escaping ([ProductAttribute]?, Error?) -> Void) {
+
+        let path = Path.attributes
+        let request = JetpackRequest(wooApiVersion: .mark3, method: .get, siteID: siteID, path: path)
+
+        let mapper = ProductAttributeListMapper(siteID: siteID)
+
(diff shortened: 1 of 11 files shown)


Networking/Networking/Remote/ProductAttributesRemote.swift
```diff
@@ -0,0 +1,121 @@
+import Foundation
+
+/// Product Attributes: Remote Endpoints for fetching global product attributes.
+///
+public final class ProductAttributesRemote: Remote {
+
+    // MARK: - Product Attributes
+
+    /// Retrieves all of the global`ProductAttribute` available.
```
ctarda, line 9: I am not sure if this actually more accurate, but maybe "Retrieves all of the `ProductAttribute`s available for a given site" sounds better than global, which might be a bit too broad?

Networking/Networking/Remote/ProductAttributesRemote.swift
```diff
@@ -0,0 +1,121 @@
...
+
+    /// Update a `ProductAttribute`.
+    ///
+    /// - Parameters:
+    ///     - siteID: Site for which we'll update the product attribute.
+    ///     - productAttributeID: ID of the Product Attribute that will be updated.
+    ///     - name: Attribute name.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func updateProductAttribute(for siteID: Int64,
```
ctarda, line 59: It seems like this method updates the attribute name only? Would it make sense to make that fact explicit in the method name?

ctarda (approved):
Looks great to me!

I asked a couple of questions, out of curiosity. Anyway, :shipit:

Networking/Networking/Remote/ProductAttributesRemote.swift
pmusolino: Ops sorry for not being totally clear here. There are two types of `ProductAttribute`. Global Attributes and Custom attributes. The global are the `ProductAttribute`s fetched using these endpoints, the custom attributes are the one added inside the specific product variation. I added the word `global` because this `*Remote` class is useful only for fetching global attributes :) more detail here p99K0U-1Mb

Networking/Networking/Remote/ProductAttributesRemote.swift
pmusolino: I left it explicitly generic because in reality the user, also on wp-admin, doesn't take care of the details under the Product Attributes, since they know only the name of the Product Attributes. All the other pieces of information are internal, and not useful.

pmusolino:
Thank you @ctarda for the review :) I responded to your questions. I'm going to merge it now, but let me know if you want to do some changes, and I can do them in another PR.

Networking/Networking/Remote/ProductAttributesRemote.swift
ctarda: Oh, I see, thanks for the clarification! 🙇

Issue #3237, opened 2020-11-30
Order Creation: add the Order Creation screen (step 3)


pmusolino:
Discussion here: p91TBi-2pw

As part of the Order Creation, we should implement this new screen which is the skeleton for creating a new order from mobile.
<img src="https://user-images.githubusercontent.com/495617/100633614-99fdee80-332e-11eb-8f45-321ac41ce02b.png" width=350 />

shiki:
I'm going to close this for now since we're not actively working on Order creation. Feel free to reopen if I made a mistake. 😅

PR #3291, opened 2020-12-04
Add order creation button on orders screen navbar

### LINKED ISSUE #3236: Order Creation: add navigation bar button for adding a new order (step 2)
As part of the Order Creation process, we should add the bar button item in the Order List. The button should be hidden if the `orderCreation` flag is disabled.

ealeksandrov:
Resolves https://github.com/woocommerce/woocommerce-ios/issues/3236.

### Description

Adds plus button in orders screen navbar, hidden behind feature flag. Action is useless for now, navigation will be added in https://github.com/woocommerce/woocommerce-ios/issues/3237.

I've decided to implement it on `OrdersTabbedViewController` level, since search action also handled there. Quick test with new view controller to be pushed there works fine - navigation bar works, whole screen area is used.

### Screenshot

Update release notes:
- This is potentially (after enabling feature flag) user-facing change, but I decided to skip release-notes for now until whole branch is ready.

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
@@ -71,6 +71,12 @@ final class OrdersTabbedViewController: ButtonBarPagerTabStripViewController {
         present(navigationController, animated: true, completion: nil)
     }
 
+    /// Shows Order Creation flow.
+    ///
+    @objc private func createOrder() {
+        // TODO-3237: show order creation screen
+    }
+
     // MARK: - ButtonBarPagerTabStripViewController Conformance
 
     /// Return the ViewControllers for "Processing" and "All Orders".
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
```diff
@@ -163,6 +169,18 @@ extension OrdersTabbedViewController {
         return button
     }
 
+    /// Create a `UIBarButtonItem` to be used as order creation trigger on the top-right.
+    ///
+    func createOrderCreationBarButtonItem() -> UIBarButtonItem {
```
pmusolino: The name of the method seems a little bit redundant, since you are using `create` and `creation`. I'd say to rename it like 
```suggestion
    func createNewOrderBarButtonItem() -> UIBarButtonItem {
```

pmusolino (commented):
Looks good to me, just a new name suggestion 👍

WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
ealeksandrov: Totally agree 😄
Fixed in f660b22

pmusolino (approved):
:shipit: thank you @ealeksandrov!

ealeksandrov:
Thank you for review, @pmusolino! 👍

PR #3293, opened 2020-12-07
[UL&S] Present error when Jetpack is either not connected or not activated

### LINKED ISSUE #3252: [UL&S] Test login with a pressable site with jetpack deactivated.
Pointed out during the review of PR #3225

ctarda:
Closes #3252 

It turns out that, when implementing the first iteration of the Jetpack error, I missed the case when Jetpack might be installed but not activated.

## Changes
* Add an extension to WordPressComSiteInfo (a struct originally declared in WPAuthenticator) to add a var that encapsulates the logic that decides if a site is accepted or not according to the state of its Jetpack installation.
* Add a couple of tests to cover that logic
* A minor change to the way the view model that backs the error deals with the case of an unknown url.
* Also, I uncommented two lines in the view model that should have been uncommented already and that support navigating back to the root of the UL&S stack.

This is the end result, for a site that has jetpack installed, but not active and not connected:

## How to test
* Check out the branch, run `bundle exec pod install`
* Attempt to log in with a site address. edgetest.mystagingwebsite.com is a Pressable site with Jetpack installed but not active and not connected, but feel free to use another site.
* Notice that we do not navigate to the "login with username and password" screen anymore, but display the error in the gif above.

cc @jaclync  who reported the original bug 🙇 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/AuthenticationManager.swift
+++ b/WooCommerce/Classes/Authentication/AuthenticationManager.swift
@@ -199,8 +199,7 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
     func shouldPresentUsernamePasswordController(for siteInfo: WordPressComSiteInfo?, onCompletion: @escaping (WordPressAuthenticatorResult) -> Void) {
 
         guard let site = siteInfo, site.hasValidJetpack == true else {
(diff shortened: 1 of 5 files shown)


pmusolino (approved):
LGTM :shipit: 

Code ✅ 
Testing ✅

PR #3294, opened 2020-12-07
Shipping Labels M1: Yosemite/Networking changes to refund a shipping label

### LINKED ISSUE #2165: Shipping Labels M1: add backend support to refund a shipping label
As part of the shipping labels M1 release, we need to add an endpoint to refund a shipping label for an order.

jaclync:
Fixes #2165 

## Why

This PR contains the backend support for requesting a refund for a shipping label.

## Changes

Networking layer:
- Added `ShippingLabelRemote.refundShippingLabel` that makes an API request to refund a shipping label, using a mapper `ShippingLabelRefundMapper`
- Added test cases for failure/success scenarios in the API request in `ShippingLabelRemoteTests`
Yosemite layer:
- Added a new action `ShippingLabelAction.refundShippingLabel` with implementation in `ShippingLabelStore` - on success, it upserts the refund to the corresponding shipping label if available
- Added support for the new action in `MockShippingLabelRemote`
- Test cases in `ShippingLabelStoreTests`

## Testing

No user-facing changes yet, just CI for now!

Update release notes:


Yosemite/Yosemite/Actions/ShippingLabelAction.swift
```diff
@@ -11,4 +11,11 @@ public enum ShippingLabelAction: Action {
                             shippingLabelID: Int64,
                             paperSize: ShippingLabelPaperSize,
                             completion: (Result<ShippingLabelPrintData, Error>) -> Void)
+
+    /// Requests a refund for a shipping label.
+    ///
+    case refundShippingLabel(siteID: Int64,
+                             orderID: Int64,
+                             shippingLabelID: Int64,
```
shiki: I wonder if we should also consider submitting the `ShippingLabel` object. As I was testing this manually, I ended up just using the object anyway. 🤷 

```swift
            guard let shippingLabel = dataSource.shippingLabel(at: indexPath) else {
                return
            }

            // This could just pass the `ShippingLabel`
            let action = ShippingLabelAction.refundShippingLabel(siteID: shippingLabel.siteID,
                                                                 orderID: shippingLabel.orderID,
                                                                 shippingLabelID: shippingLabel.shippingLabelID) { result in
                print(result)
                print("done!")
            }
            print("refunding \(shippingLabel)")
```

Probably something to consider once everything is connected.

Networking/NetworkingTests/Remote/ShippingLabelRemoteTests.swift
```diff
@@ -55,4 +55,45 @@ final class ShippingLabelRemoteTests: XCTestCase {
         XCTAssertEqual(printData.mimeType, "application/pdf")
         XCTAssertFalse(printData.base64Content.isEmpty)
     }
+
+    func test_refundShippingLabel_returns_refund_on_success() throws {
```
shiki, line 59: 👍

Networking/NetworkingTests/Remote/ShippingLabelRemoteTests.swift
```diff
@@ -55,4 +55,45 @@ final class ShippingLabelRemoteTests: XCTestCase {
...
+        let result: Result<ShippingLabelRefund, Error> = try waitFor { promise in
+            remote.refundShippingLabel(siteID: self.sampleSiteID,
+                                       orderID: orderID,
+                                       shippingLabelID: shippingLabelID) { result in
+                promise(result)
+            }
+        }
+
+        // Then
+        XCTAssertNotNil(result.failure)
```
shiki: ⚠️ Do you think it's worth it to test that we're also parsing the error? 

```
(lldb) po result.failure
▿ Optional<Error>
  ▿ some : Dotcom Error: [wcc_server_error_response] Error: The WooCommerce Shipping & Tax server returned: Bad Request Unable to request refund. The parcel has been shipped. ( 400 )
    ▿ unknown : 2 elements
      - code : "wcc_server_error_response"
      ▿ message : Optional<String>
        - some : "Error: The WooCommerce Shipping & Tax server returned: Bad Request Unable to request refund. The parcel has been shipped. ( 400 )"
```

I think this error is even useful to display to the user. I get this in WC Admin: 

image

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
```diff
@@ -135,6 +175,17 @@ private extension ShippingLabelStore {
         }
     }
 
+    func handleShippingLabelRefund(refund: ShippingLabelRefund?, storageShippingLabel: StorageShippingLabel) {
```
shiki: ⚠️ It'd be nice if we can rename this to something that is not “handle“. 🙂 If I understand it correctly, the goal is to _update the `refund`_ of `Storage.ShippingLabel`. So maybe something like 

```swift
func update(shippinglabel: ShippingLabel, withRefund refund: ShippingLabelRefund) { }
```

It's not the best but I think it's better than `handle` so future developers will have a better idea of the intent and not blow this method up. 😬

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
```diff
@@ -97,6 +122,27 @@ private extension ShippingLabelStore {
...
+                                               orderID: Int64,
+                                               shippingLabelID: Int64,
+                                               refund: ShippingLabelRefund,
+                                               onCompletion: @escaping () -> Void) {
+        let derivedStorage = sharedDerivedStorage
+        derivedStorage.perform { [weak self] in
+            guard let self = self else { return }
+            guard let shippingLabel = derivedStorage.loadShippingLabel(siteID: siteID, orderID: orderID, shippingLabelID: shippingLabelID) else {
+                return
+            }
```
shiki: Should we mention why we're not doing anything if the `ShippingLabel` does not exist?

shiki (approved):
I tested it manually and it works! 

```
▿ Result<ShippingLabelRefund, Error>
  ▿ success : ShippingLabelRefund
    ▿ dateRequested : 2020-12-07 22:22:28 +0000
      - timeIntervalSinceReferenceDate : 629072548.4000001
    - status : Networking.ShippingLabelRefundStatus.pending
```

I just have a couple of concerns (⚠️ ) that can be resolved easily. Or we can discuss if you disagree. 🙂

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
jaclync: Good idea! I was following the naming like how we [update relationships of a `Product` in `ProductStore`](https://github.com/woocommerce/woocommerce-ios/blob/fb2b7595935cd1e38f70630dd8ad0ef5145f3aa9/Yosemite/Yosemite/Stores/ProductStore.swift#L393), but this `update` function sounds much more clear. Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3294/commits/bc7ab8b7b185ef6ba7b3ec0e0563479fecdb4669

Yosemite/Yosemite/Actions/ShippingLabelAction.swift
jaclync: Makes sense, it does save a bunch of code! Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3294/commits/1686982f6e315cf92c3fca7c0d1c58671dc6dde2

Networking/NetworkingTests/Remote/ShippingLabelRemoteTests.swift
jaclync: Yes thanks for catching that! Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3294/commits/a4ea4c4b5021a6027c6837ff61f00501544ce8cf

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
jaclync: Yup, aded a comment in: https://github.com/woocommerce/woocommerce-ios/pull/3294/commits/a272a54b8772fafda0eb7f5c2ea2353db035fbe7

Networking/NetworkingTests/Remote/ShippingLabelRemoteTests.swift
jaclync: > I think this error is even useful to display to the user. I get this in WC Admin:

Hmm, I'm surprised to see this message in wp-admin. I personally find the message not very user-friendly (the 400 error code and the "Error: The WooCommerce Shipping & Tax server returned" sound very technical and out of place) 🙄 I'll respond to your p2 comment and ask G about this!

jaclync:
Thanks @shiki for the suggestions and testing the changes manually! (I tested with WIP UI integration) I updated the PR for another look 😄

Networking/NetworkingTests/Remote/ShippingLabelRemoteTests.swift
shiki: Yup. It's not user-friendly at all. Thanks for starting the discussion. 🙂

shiki (approved):
🤸

PR #3295, opened 2020-12-07
[Issue Refund] Hide "select all" CTA when there are no products left to refund

### LINKED ISSUE #3283: Refunds: Remove "Select All" CTA when there are no products left to refund
**Describe the bug**

In this scenario:

- I have already refunded all the products in the order.
- I start to issue a new refund for the shipping on the order.

I noticed that there’s still a section at the top of the refund screen where the products would appear if I hadn’t refunded them already, with e.g. a "Select All" CTA that would be used to select products to refund.

I’m still able to select “Refund Shipping” and continue the refund as expected, so functionally it works fine, but that "Select All" CTA is out of place when there are no products left to refund.

**To Reproduce**
1. Go to the Orders tab.
2. Select an order with products + a shipping cost.
3. Scroll down to the payments section and select "Issue Refund."
4. Refund all of the products (but not the shipping).
5. On the order detail, select "Issue Refund" again.
6. Notice the "Select All" CTA still appears on the refund screen.

**Screenshots**

Refund screen (no products left to refund)|Refund screen (no products, shipping selected)
-|-
image|image

**Expected behavior**

In the case where there are no products left to refund, we should hide the “Select All” CTA. (internal ref: p5T066-1Ew-p2#comment-6451)

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 14.2
- WooCommerce iOS version: 5.6.0.1 TestFlight beta

Ecarrion:
fix #3283 

# Why

It was reported that after refunding all products(but not shipping) and then going back to the "issue refund" screen the "Select All" button was still visible but it did nothing since there are no products left to refund.

This PR fixes that situation by hiding that button when there are no more products to refund.

# How

- Added a new property `isSelectAllButtonVisible` in `IssueRefundViewModel` that is calculated at init.
- Update `IssueRefundViewController` to set the button visibility based on the viewmodel.  

# Screenshot
SelectAllButton-demo

# Testing Steps
- Go to the Orders tab.
- Select an order with products + a shipping cost.
- Scroll down to the payments section and select "Issue Refund."
- Refund all of the products (but not the shipping).
- On the order detail, select "Issue Refund" again.
- Notice the "Select All" CTA don't appear anymore.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewController.swift
@@ -75,6 +75,7 @@ private extension IssueRefundViewController {
         title = viewModel.title
         itemsSelectedLabel.text = viewModel.selectedItemsTitle
         nextButton.isEnabled = viewModel.isNextButtonEnabled
+        selectAllButton.isHidden = !viewModel.isSelectAllButtonVisible
         tableView.reloadData()
     }
 }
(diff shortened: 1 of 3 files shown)


ctarda (approved):
:shipit:

PR #3296, opened 2020-12-07
[Issue Refund] Control Interactive dismiss gesture

### LINKED ISSUE #3264: Refunds: Disable swipe to dismiss
Right now it is possible to swipe down the refunds screen to dismiss it. I think it’s actually quite easy to do this by accident and all the changes will be lost. My suggestion is to disable “swipe to dismiss” and only dismiss the screen when the flow is done or when the user explicitly closes it.

**To Reproduce**
1. Go to an order details
2. Tap on Issue Refund
3. Swipe down to dismiss

**Expected behavior**
Only pressing `cancel` or `refund` dismisses the screen.

Ecarrion:
fix #3264

# Why
It was reported that the new(IOS 13+) interactive dismiss behavior for the issue refund flow was too easy to activate causing changes to the flow to be lost.

This PR fixes that by showing a confirmation action sheet when the user tries to dismiss the issue refund flow with a drag gesture but it has some unsaved changes.

# How

**On `IssueRefundCordinatingController`**
- Created a new protocol `IssueRefundInteractiveDismissDelegate` to allow its view controllers to decide if they should be dismissed by drag gesture or not.
- Set itself to be the presentation delegate and forward the `presentationControllerShouldDismiss` message.
- Present a confirmation action sheet when the user tries to dismiss the flow by a drag gesture while having unsaved changes.
- Update the `dismissIssueRefundFlow` method to be able to dismiss the flow without presenting the success notice.

**On `IssueRefundViewModel`**
- Added a new property `hasUnsavedChanges` that is computed on every state change

**On `IssueRefundViewController`**
- Conform to `IssueRefundInteractiveDismissDelegate` to allow dismissal when the view model does not have any pending changes.

**on `RefundConfirmationViewController`**
- Always return false for the `presentationControllerShouldDismiss` method as in that screen there are always changes pending to save.

# Demo
interactive-dismiss

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
@@ -41,6 +41,7 @@ final class IssueRefundViewModel {
             title = calculateTitle()
             selectedItemsTitle = createSelectedItemsCount()
             isNextButtonEnabled = calculateNextButtonEnableState()
+            hasUnsavedChanges = calculatePendingChangesState()
             onChange?()
         }
     }
@@ -60,6 +61,10 @@ final class IssueRefundViewModel {
     /// Boolean indicating if the next button is enabled
     ///
     private(set) var isNextButtonEnabled: Bool = false
+
+    /// Boolean indicating if there are refunds pending to commit
+    ///
+    private(set) var hasUnsavedChanges: Bool = false
 
     /// The sections and rows to display in the `UITableView`.
     ///
@@ -91,6 +96,7 @@ final class IssueRefundViewModel {
         title = calculateTitle()
         isNextButtonEnabled = calculateNextButtonEnableState()
         selectedItemsTitle = createSelectedItemsCount()
+        hasUnsavedChanges = calculatePendingChangesState()
     }
 
     /// Creates the `ViewModel` to be used when navigating to the page where the user can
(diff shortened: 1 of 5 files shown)


Ecarrion:
@Garance91540 this was suggested by @eshurakov and I added a confirmation alert as I found it more useful than just disabling the dismiss gesture. Let me know if you think it's ok.

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/IssueRefundViewModelTests.swift
```diff
@@ -442,6 +442,37 @@ final class IssueRefundViewModelTests: XCTestCase {
         XCTAssertFalse(viewModel.isNextButtonEnabled)
     }
 
+    func test_viewModel_starts_with_no_saved_changes() {
```
jaclync: nit: would `no_unsaved_changes` be more consistent with the `hasUnsavedChanges` name?

```suggestion
    func test_viewModel_starts_with_no_unsaved_changes() {
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Issue Refunds/IssueRefundViewModel.swift
```diff
@@ -316,7 +322,13 @@ extension IssueRefundViewModel {
     /// Calculates wether the next button should be enabled or not
     ///
     private func calculateNextButtonEnableState() -> Bool {
-        return state.refundQuantityStore.count() > 0 || state.shouldRefundShipping
+        calculatePendingChangesState()
+    }
+
+    /// Calculates wether the are pending changes to commit
```

```suggestion
    /// Calculates whether there are pending changes to commit
```

Garance91540 (approved):
Sounds good to me!

jaclync (approved):
Nice enhancement! :shipit: 

Code: ✅ 
Testing: ✅ (tested without changes, shipping enabled, and some items selected)

PR #3297, opened 2020-12-07
Add OrderFormViewController for Create order flow

### LINKED ISSUE #3237: Order Creation: add the Order Creation screen (step 3)

ealeksandrov:
Resolves https://github.com/woocommerce/woocommerce-ios/issues/3237.

### Description

This is implementation of OrderFormViewController to host root view for Orders creation/editing. Now it'll just have empty summary + 3 action buttons.

OrderFormDataSource is decoupled to handle UITableViewDataSource. Now it's the only data abstraction but as a next step I'll add ViewModel that will be passed in OrderFormDataSource on its init. Let me know if it's a good idea 🙂

### Known limitations

Summary header cell is left without changes for now. I think "draft" status support should be added on all orders later and then correct data and status will be displayed there.

### Questions

1. Question about localization - if we have identical strings ("keys") on different screens - they will be merged into single line with multiple descriptions, right? Do we have some considerations about it?
2. I18n - do we need to add `accessibilityLabel` and `accessibilityHint` on actions? I see it in some places but haven't found documentation on it. Added simple labels and hints on action cells for now.

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
@@ -1,7 +1,7 @@
 import UIKit
 
 final class OrderCreationFormDataSource: NSObject {
-    let viewModel: OrderCreationFormViewModel
+    private let viewModel: OrderCreationFormViewModel
 
     init(viewModel: OrderCreationFormViewModel) {
         self.viewModel = viewModel
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
@@ -1,6 +1,6 @@
 import UIKit
 
-final class OrderFormDataSource: NSObject {
+final class OrderCreationFormDataSource: NSObject {
     private(set) var sections: [Section] = []
 
     override init() {
@@ -23,7 +23,7 @@ final class OrderFormDataSource: NSObject {
     }
 }
 
-extension OrderFormDataSource: UITableViewDataSource {
+extension OrderCreationFormDataSource: UITableViewDataSource {
     func numberOfSections(in tableView: UITableView) -> Int {
         return sections.count
     }
@@ -62,7 +62,7 @@ extension OrderFormDataSource: UITableViewDataSource {
     }
 }
 
-extension OrderFormDataSource {
+extension OrderCreationFormDataSource {
 
     struct Section {
         enum Category {
@@ -100,7 +100,7 @@ extension OrderFormDataSource {
     }
 }
 
-private extension OrderFormDataSource {
+private extension OrderCreationFormDataSource {
     func configure(_ cell: UITableViewCell, section: Section, indexPath: IndexPath) {
         let row = section.rows[indexPath.row]
         switch cell {
@@ -156,7 +156,7 @@ private extension OrderFormDataSource {
     }
 }
 
-private extension OrderFormDataSource {
+private extension OrderCreationFormDataSource {
     enum Icons {
         static let addItemIcon = UIImage.plusImage
     }
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
@@ -1,20 +1,15 @@
 import UIKit
 
 final class OrderCreationFormDataSource: NSObject {
-    private(set) var sections: [Section] = []
+    let viewModel: OrderCreationFormViewModel
 
-    override init() {
+    init(viewModel: OrderCreationFormViewModel) {
+        self.viewModel = viewModel
         super.init()
-        sections = [
-            Section(category: .summary, title: nil, rows: [.summary]),
-            Section(category: .items, title: Localization.itemsHeader, rows: [.addOrderItem]),
-            Section(category: .customerInformation, title: Localization.customerHeader, rows: [.addCustomer]),
-            Section(category: .notes, title: Localization.orderNotesHeader, rows: [.addOrderNote])
-        ]
     }
 
     func registerTableViewCells(_ tableView: UITableView) {
-        for row in Row.allCases {
+        for row in OrderCreationFormViewModel.Row.allCases {
             tableView.registerNib(for: row.type)
         }
 
@@ -25,15 +20,15 @@ final class OrderCreationFormDataSource: NSObject {
 
 extension OrderCreationFormDataSource: UITableViewDataSource {
     func numberOfSections(in tableView: UITableView) -> Int {
-        return sections.count
+        return viewModel.sections.count
     }
 
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return sections[section].rows.count
+        return viewModel.sections[section].rows.count
     }
 
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let section = sections[indexPath.section]
+        let section = viewModel.sections[indexPath.section]
         let reuseIdentifier = section.rows[indexPath.row].reuseIdentifier
         let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath)
         configure(cell, section: section, indexPath: indexPath)
@@ -42,7 +37,7 @@ extension OrderCreationFormDataSource: UITableViewDataSource {
 
     func heightForHeaderInSection(_ section: Int, tableView: UITableView) -> CGFloat {
         // Hide header for summary
-        if sections[section].category == .summary {
+        if viewModel.sections[section].category == .summary {
             return CGFloat.leastNormalMagnitude
         }
 
@@ -55,53 +50,15 @@ extension OrderCreationFormDataSource: UITableViewDataSource {
             return nil
         }
 
-        header.leftText = sections[section].title
+        header.leftText = viewModel.sections[section].title
         header.rightText = nil
 
         return header
     }
 }
 
-extension OrderCreationFormDataSource {
-
-    struct Section {
-        enum Category {
-            case summary
-            case items
-            case customerInformation
-            case notes
-        }
-
-        let category: Category
-        let title: String?
-        let rows: [Row]
-    }
-
-    /// Rows listed in the order they appear on screen
-    ///
-    enum Row: CaseIterable {
-        case summary
-        case addOrderItem
-        case addCustomer
-        case addOrderNote
-
-        var type: UITableViewCell.Type {
-            switch self {
-            case .summary:
-                return SummaryTableViewCell.self
-            case .addOrderItem, .addCustomer, .addOrderNote:
-                return LeftImageTableViewCell.self
-            }
-        }
-
-        var reuseIdentifier: String {
-            return type.reuseIdentifier
-        }
-    }
-}
-
 private extension OrderCreationFormDataSource {
-    func configure(_ cell: UITableViewCell, section: Section, indexPath: IndexPath) {
+    func configure(_ cell: UITableViewCell, section: OrderCreationFormViewModel.Section, indexPath: IndexPath) {
         let row = section.rows[indexPath.row]
         switch cell {
         case let cell as SummaryTableViewCell:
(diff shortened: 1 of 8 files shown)


WooCommerce/WooCommerce.xcodeproj/project.pbxproj
```diff
@@ -1096,11 +1098,11 @@
...
 		0211253F25778BDF0075AD2A /* ShippingLabelDetailsViewController.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ShippingLabelDetailsViewController.swift; sourceTree = "<group>"; };
 		0211254025778BDF0075AD2A /* ShippingLabelDetailsViewController.xib */ = {isa = PBXFileReference; lastKnownFileType = file.xib; path = ShippingLabelDetailsViewController.xib; sourceTree = "<group>"; };
 		021125472577CC650075AD2A /* ShippingLabelDetailsViewModel.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ShippingLabelDetailsViewModel.swift; sourceTree = "<group>"; };
-		0211252725773F220075AD2A /* Models+Copiable.generated.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "Models+Copiable.generated.swift"; sourceTree = "<group>"; };
-		0211252D25773FB00075AD2A /* MockAggregateOrderItem.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockAggregateOrderItem.swift; sourceTree = "<group>"; };
```
ealeksandrov: Xcode doing its XML lines dance... Missed it. If needed can do a force push with removal of this changes.

WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
```diff
@@ -74,7 +74,9 @@ final class OrdersTabbedViewController: ButtonBarPagerTabStripViewController {
     /// Shows Order Creation flow.
     ///
     @objc private func createOrder() {
```
pmusolino: Following the name convention of the other methods, I'd say to rename it on 
```suggestion
    @objc private func displayCreateOrder() {
```

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormDataSource.swift
```diff
@@ -0,0 +1,187 @@
...
+                return SummaryTableViewCell.self
+            case .addOrderItem, .addCustomer, .addOrderNote:
+                return LeftImageTableViewCell.self
+            }
+        }
+
+        var reuseIdentifier: String {
+            return type.reuseIdentifier
+        }
+    }
```
pmusolino: It would be better to declare an extension with `Section` and `Row` inside, at the end of the file, since these are not crucial info. We prefer to declare properties and init methods on top of the file ;)

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormDataSource.swift
```diff
@@ -0,0 +1,187 @@
...
+        case let cell as SummaryTableViewCell:
+            configureSummary(cell: cell)
+        case let cell as LeftImageTableViewCell where row == .addOrderItem:
+            configureAddItem(cell: cell)
+        case let cell as LeftImageTableViewCell where row == .addCustomer:
+            configureAddCustomer(cell: cell)
+        case let cell as LeftImageTableViewCell where row == .addOrderNote:
+            configureAddNote(cell: cell)
+        default:
+            fatalError("Unknown cell in Order Form")
```
pmusolino:
```suggestion
            fatalError("Unknown cell in Order Creation Form")
```

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormDataSource.swift
```diff
@@ -0,0 +1,187 @@
+import UIKit
+
+final class OrderFormDataSource: NSObject {
```
pmusolino:
```suggestion
final class OrderCreationFormDataSource: NSObject {
```

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormViewController.swift
```diff
@@ -0,0 +1,85 @@
+import UIKit
+
+final class OrderFormViewController: UIViewController {
```
pmusolino:
```suggestion
final class OrderCreationFormViewController: UIViewController {
```

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormViewController.swift
```diff
@@ -0,0 +1,85 @@
...
+    func configureTableView() {
+        dataSource.registerTableViewCells(tableView)
+
+        tableView.delegate = self
+        tableView.dataSource = dataSource
+
+        view.backgroundColor = .listBackground
+        tableView.backgroundColor = .listBackground
+        tableView.sectionHeaderHeight = UITableView.automaticDimension
+        tableView.sectionFooterHeight = .leastNonzeroMagnitude
```
pmusolino: This seems redundant since you already declared the delegate method, or I'm wrong?

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormDataSource.swift
```diff
@@ -0,0 +1,187 @@
...
+            configureAddCustomer(cell: cell)
+        case let cell as LeftImageTableViewCell where row == .addOrderNote:
+            configureAddNote(cell: cell)
+        default:
+            fatalError("Unknown cell in Order Form")
+        }
+    }
+
+
+    private func configureSummary(cell: SummaryTableViewCell) {
```
pmusolino: nit: not needed to declare it as `private`, since the extension is already `private`.

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormDataSource.swift
```diff
@@ -0,0 +1,187 @@
...
+        // TODO: add 'new draft' state
+    }
+
+    private func configureAddItem(cell: LeftImageTableViewCell) {
```
pmusolino: nit: not needed to declare it as `private`, since the extension is already `private`.

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormDataSource.swift
```diff
@@ -0,0 +1,187 @@
...
+        cell.imageView?.tintColor = .accent
+        cell.textLabel?.textColor = .accent
+        cell.labelText = Localization.addItemsTitle
+
+        cell.accessibilityTraits = .button
+        cell.accessibilityLabel = Localization.addItemsAccessibilityLabel
+        cell.accessibilityHint = Localization.addItemsAccessibilityHint
+    }
+
+    private func configureAddCustomer(cell: LeftImageTableViewCell) {
```
pmusolino: nit: not needed to declare it as `private`, since the extension is already `private`.

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormDataSource.swift
```diff
@@ -0,0 +1,187 @@
...
+        cell.imageView?.tintColor = .accent
+        cell.textLabel?.textColor = .accent
+        cell.labelText = Localization.addCustomerTitle
+
+        cell.accessibilityTraits = .button
+        cell.accessibilityLabel = Localization.addCustomerAccessibilityLabel
+        cell.accessibilityHint = Localization.addCustomerAccessibilityHint
+    }
+
+    private func configureAddNote(cell: LeftImageTableViewCell) {
```
pmusolino: nit: not needed to declare it as `private`, since the extension is already `private`.

pmusolino (changes requested):
Thank you @ealeksandrov for opening this PR!

I'm curious about the logic flow that you have in plan to implement. 
What is your idea about managing the data here? Where do you have in plan to handle the changes?
It seems a little bit hard to be tested without a View Model. For example, you generate the sections inside the DataSource, but it can be hard to test all the various use cases (expected status after initializing a specific data, expected status with empty data, and so on). What do you think?

pmusolino:
> Question about localization - if we have identical strings ("keys") on different screens - they will be merged into single line with multiple descriptions, right? Do we have some considerations about it?

This is correct and expected. It's ok to keep it as is ;)  


> I18n - do we need to add accessibilityLabel and accessibilityHint on actions? I see it in some places but haven't found documentation on it. Added simple labels and hints on action cells for now.

Since at this moment this is a static table view with some TableViewCell as Button, it's ok (and great) that you considered the accessibility. :) Here you can find more info p77Llu-djJ

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormViewController.swift
ealeksandrov: I've preferred to leave `Creation` part out of VC name because we may later reuse it for editing of drafts, what do you think?
Of course we can name it `OrderCreationForm` now and change to generic `Form` later when it fits 😄

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormViewController.swift
ealeksandrov: Good catch for header - I've added this one before implementing delegate later.
For footer it's still needed, I don't have its delegate method handled.

ealeksandrov:
Thanks for feedback, @pmusolino!

> What is your idea about managing the data here? Where do you have in plan to handle the changes?

I want to have ViewModel which stores order initial state (empty or previous draft) and separately - any pending changes on top of it (until you hit save). And that ViewModel will dispatch the action to perform actual save.

> For example, you generate the sections inside the DataSource, but it can be hard to test all the various use cases

I think ViewModel should be be injected into DataSource, and DataSource will return something like `viewModel.sectionsCount`. Static sections init is a placeholder/demo for now.

I'm adding it because I think that `UITableViewDataSource` should be decoupled from general ViewModel, so ViewModel doesn't depend on UI implementation at all.

Products and Customers lists (pickers) will be separated with own ViewModels to sync data, it's out of scope for Root VC+VM.

ealeksandrov:
Actually ViewModel can be added easily on this step. It doesn't really depend on other layers for initial state, let me add it and push update.
I'll think about some related tests too.

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderFormViewController.swift
pmusolino: Sounds good

pmusolino:
Looks good and thanks for clarifying some aspect @ealeksandrov! :)

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
```diff
@@ -0,0 +1,158 @@
+import UIKit
+
+final class OrderCreationFormDataSource: NSObject {
+    let viewModel: OrderCreationFormViewModel
```
pmusolino: This should be declared as` private` since is not accessible outside the class.

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormViewController.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    init() {
+        self.viewModel = OrderCreationFormViewModel()
+        self.dataSource = OrderCreationFormDataSource(viewModel: viewModel)
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        self.viewModel = OrderCreationFormViewModel()
+        self.dataSource = OrderCreationFormDataSource(viewModel: viewModel)
```
pmusolino, line 19: Is really needed, in this case, to initialize `viewModel` and `dataSource` here?

pmusolino (approved):
Looks good to me, just a minor nit that should be fixed before merging it 👍

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormViewController.swift
ealeksandrov: `viewModel` and `dataSource` are non-optional without default value. If we won't init them here in any way - we should block this required initializer and throw `fatalError`.
Any more ideas how to handle 2 inits properly?

WooCommerce/Classes/ViewRelated/Orders/Order Create/OrderCreationFormDataSource.swift
ealeksandrov: Fixed in bc52132

PR #3299, opened 2020-12-07
List Existing Attributes in Variations: Yosemite and Storage layers changes

pmusolino:
Part of https://github.com/woocommerce/woocommerce-ios/issues/3181

After having implemented the [networking layer](https://github.com/woocommerce/woocommerce-ios/pull/3290) for fetching Product Attributes, in this PR I implemented the Yosemite and Storage layers for introducing the changes necessary to manage the Product Attributes.

**Do not be scared:** The majority of LOC of this PR is because of the new model 40 in Storage and because of the new xcmappingmodel. The tests for the `ProductAttributeStore` will be part of another PR (I didn't want to burden this PR too much).

## Changes
- Added the `siteID` parameter inside the `ProductAttribute` model. Since it’s now possible to fetch `ProductAttribute`s from a dedicated endpoint, we want to store them by siteID, and not associated with a product. Updated tests according.
- Updated the storage layer, introducing the new model 40, that adds the `siteID` field inside the `ProductAttribute` entity. 
- Updated `Product`'s `attributes` relationship with `nullify` as delete rule.
- Used mapping model: `WooCommerceModelV39toV40.xcmappingmodel` to remove product attributes without `siteID`. Implemented also a migration test from model 39 to model 40.
- Updated `ProductAttribute+CoreDataProperties` and `ProductAttribute+ReadOnlyConvertible` adding the `siteID` field.
- In `StorageType+Extensions` added a new method for fetching a stored Product Attribute by `siteID` + `attributeID`.
- Implemented `ProductAttributeAction` and `ProductAttributeStore` with methods for synchronizing, adding, updating, and deleting product attributes.

## Testing
- Just CI is enough, there are no changes in the UI. Make sure the app runs like before.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/ProductAttribute.swift
+++ b/Networking/Networking/Model/Product/ProductAttribute.swift
@@ -4,6 +4,7 @@ import Foundation
 /// Represents a ProductAttribute entity.
 ///
 public struct ProductAttribute: Decodable {
+    public let siteID: Int64
     public let attributeID: Int64
     public let name: String
     public let position: Int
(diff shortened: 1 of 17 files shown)


ctarda (approved):
I run develop, and then this branch right after and I didn't see any migration errors or warnings. The app works as expected too.

:shipit:

pmusolino:
Thank you so much @ctarda for the review 😸

PR #3301, opened 2020-12-08
Hide WordPress.com TOS 

### LINKED ISSUE #3145: [UL&S] Continue with WP.com: remove link to WP terms of service displayed right below email textfield
The link to WP.com links of service displayed below the textfield to enter email address should be removed.

This is how it looks now:

And this is how it should look:

ctarda:
Closes #3145 
Ref: wordpress-mobile/WordPressAuthenticator-iOS#536

## Changes

| Before | After |
| ---- | ---- |
|  |  |

* Point the podfile to the branch in WPAuthenticator that actually implements the solution
* Update AuthenticationManager to implement the new delegate method that allows removing the link to the TOS

## Testing
* Log out if necessary
* Attempt to log in with a WordPress.com account. Notice if the link to the Terms Of Service worth above the "Continue" button is visible.

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -38,9 +38,9 @@ target 'WooCommerce' do
   pod 'Gridicons', '~> 1.0'
 
   # To allow pod to pick up beta versions use -beta. E.g., 1.1.7-beta.1
-  pod 'WordPressAuthenticator', '~> 1.31.0-beta.3'
+  # pod 'WordPressAuthenticator', '~> 1.31.0-beta.3'
   # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :commit => ''
-  # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => ''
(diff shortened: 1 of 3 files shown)


jaclync (approved):
LGTM! 🚢 

before | after
-- | --
Simulator Screen Shot - iPhone 8 Plus - 2020-12-08 at 16 31 26 | Simulator Screen Shot - iPhone 8 Plus - 2020-12-08 at 16 32 29

Garance91540 (changes requested):
Hi Cesar, can we make sure there's 16px vertical padding between the email text field and the "Continue" button? https://d.pr/i/smfi7B

ctarda:
Hi @Garance91540 , thanks for the review!

I am afraid we can't, at least not without significant effort, as that UI is provided by WPAuthenticator and we don't have much control over it, other than hiding the TOS. 

How critical would you say that is? If it is, is this something we could defer to a second milestone?

Garance91540:
> How critical would you say that is? If it is, is this something we could defer to a second milestone?

This is purely visual, so it's fine to defer for later!

ctarda:
> This is purely visual, so it's fine to defer for later!

Thanks! I've logged #3309 for that. Merging now!

PR #3305, opened 2020-12-08
Move stores manager to Yosemite

jkmassel:
This PR moves the `StoresManager` protocol down to the `Yosemite` layer, per our discussion [here](https://github.com/woocommerce/woocommerce-ios/pull/3266#issuecomment-738815306). 

It was a little more complex that just relocating the files, so I'd like to check in here since I haven't done much work in the Woo app.

**Step 1 - Observables**
The `StoresManager` protocol relies on `Observable` in a few places. Because this class is used extensively in `WooCommerce` but nowhere in `Yosemite`, I needed to make it available in `Yosemite. It seems like a hard-and-fast rule that lower layers never import from higher layers, so I figured there were two options:

1. Move all the `Observable` stuff into `Yosemite`, and import it as needed.
2. Move all the `Observable` stuff into its own package and import that package as needed.

I went with Option 2 – all of the `Observable` objects have been moved over, and have had access modifiers added to handle their use outside of the new module.

Also tests were moved over.

**Step 2 – `SessionManager`**
The `StoresManager` object contains a reference to the current `SessionManager`, which lives in the `WooCommerce` target. There's no reasonable way to move `SessionManager` over to `Yosemite` without a more significant refactor (it depends on custom `UserDefaults` extensions in several places, as well as the `KeychainAccess` pod, and `WooConstants`. 

I can see a good argument for moving it into `Yosemite`, but it would add some complexity – and this is something I'd like feedback on – in your opinion, is it worth moving over as well? I think we'd need to refactor it to be more storage-agnostic, but I don't think that's an impossibility. 

In the interim, I've protocol-ized it, so there's a `SessionManagerProtocol` that lives in `Yosemite` and replaces every instance of `SessionManager` previously used in the `WooCommerce` layer. `SessionManager` just conforms to it and can be injected. Additionally, I refactored the tests slightly to work properly with this change.

**Step 3 – `StoresManager`**
WIth these items out of the way, we can now move `StoresManager` into `Yosemite`, and everything just works.  One thing I would like feedback/confirmation on for this – does putting `StoresManager` into the `Base` directory make sense?

--

Like I mentioned, this is a first step, so if you're ok with this as-is, we could merge it and tackle next steps after some additional discussion? It seems like your goal is to move the whole `classes/Yosemite` into `Yosemite` in order to make the `Store` objects internal? If that's the case, I can carry on with that if you'd like - LMK!! 

Update release notes:


shiki:
I'll get to this tomorrow! Sorry for the delay!

Observables/README.md
```diff
@@ -0,0 +1,3 @@
+# Observables
+
+A description of this package.
```
shiki: ⚠️  Would you mind updating this to:

```md
These classes are our own implementation of [ReactiveX observables](http://reactivex.io/) that we used when we were targeting iOS 12+. But now that our minimum iOS version is iOS 13, we should stop using them and use [Combine](https://developer.apple.com/documentation/combine) instead. 

We should also migrate the existing usages of these classes to use Combine. The classes were intentionally created to be compatible with both ReactiveX and Combine so the migration should be smooth-ish.
```

🙂

Observables/Package.swift
```diff
@@ -0,0 +1,31 @@
...
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
+        .target(
+            name: "Observables",
+            dependencies: []),
+        .testTarget(
+            name: "ObservablesTests",
+            dependencies: ["Observables"]),
+    ]
```
shiki, line 30: 🛑  When I ran all the unit tests, the tests in this package were not included. I think we'd need to add it to the `UnitTests.xctestplan`. 

image

Yosemite/Yosemite/Base/SessionManagerProtocol.swift
```diff
@@ -0,0 +1,18 @@
...
+    var defaultStoreID: Int64? { get set }
+
+    var anonymousUserID: String? { get }
+
+    var siteID: Observable<Int64?> { get }
+
+    var defaultCredentials: Credentials? { get set}
+
+    func reset()
+}
```
shiki: ⚠️ Would you mind copying over the comments from the `SessionManager`? 🙂 Since this is referred to by `StoresManager`, I think it makes more sense to have some comments here.

WooCommerce/WooCommerceTests/Yosemite/StoresManagerTests.swift
```diff
@@ -173,7 +174,13 @@ extension DefaultStoresManager {
...
     static var testingInstance: DefaultStoresManager {
-        return DefaultStoresManager(sessionManager: .testingInstance)
+        return DefaultStoresManager(concreteSessionManager: .testingInstance)
+    }
+
+    /// Allow initializing a test `DefaultStoresManager` using a concrete implementation of `SessionManager`
+    ///
+    convenience init(concreteSessionManager: SessionManager) {
+        self.init(sessionManager: concreteSessionManager)
     }
```
shiki: It doesn't look like we need the `init(concreteSessionManager:)`. Looks like this works just fine:

```suggestion
    static var testingInstance: DefaultStoresManager {
        DefaultStoresManager(sessionManager: SessionManager.testingInstance)
    }
```

Out of scope: I'm worried about this `static` instance. This means that we're keeping this object until all the tests have finished running. 😬

shiki (changes requested):
Thanks for working on this, @jkmassel! This looks great to me! I'm only requesting changes because of the unit tests that are not executed. See inline comment marked with 🛑 . I also have minor requests (⚠️). 🙂 

I like how you separated commits. It made it so much easier for me to track what happened. 🙏 

> Observables: I went with Option 2 – all of the Observable objects have been moved over, and have had access modifiers added to handle their use outside of the new module.

This is fine. We're supposed to remove it after we set the minimum to iOS 13 and the SPM makes it easier to not forget about this _debt_. 🤪 

> SessionManager 
> I can see a good argument for moving it into Yosemite, but it would add some complexity – and this is something I'd like feedback on – in your opinion, is it worth moving over as well? I think we'd need to refactor it to be more storage-agnostic, but I don't think that's an impossibility.

Gah! I forgot about that one. I'm still not sure if `SessionManager` should really belong inside `StoresManager`. It feels like they're separate objects with different responsibilities and don't need to know about each other. That's the extent of my thoughts about it. 😄 

I do think you made a good decision here of using a `protocol`. We can leave this as-is for now and make a decision later if it becomes a problem. 

> One thing I would like feedback/confirmation on for this – does putting StoresManager into the Base directory make sense?

I think so. 🤷 

> It seems like your goal is to move the whole classes/Yosemite into Yosemite in order to make the Store objects internal? If that's the case, I can carry on with that if you'd like - LMK!!

That's the goal. But that means we'll also need to move `DefaultStoresManager` and `AuthenticatedState`. But now I see that to achieve our goals in #3266, it doesn't look like we need to move those to `Yosemite` for now. So I'm fine with focusing on the goal and dealing with the `Store` objects later.

Observables/README.md
jkmassel: Well that answers that question – I was wondering why we had this given iOS 13  😁

jkmassel: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3305/commits/b11a4683399736a08f5333cae7447d4af101fe4b

Observables/Package.swift
jkmassel: Ooh yeah – done in https://github.com/woocommerce/woocommerce-ios/pull/3305/commits/cc35693657a7a9579346d0d9547453f16ecd76be

Yosemite/Yosemite/Base/SessionManagerProtocol.swift
jkmassel: For sure – happy to do so! The reason I didn't previously was it seemed like most `Yosemite` protocols didn't have this, and I assumed it was to avoid maintaining both 😅 

I think I like it better this way 🙂

Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3305/commits/b4f525d7d6b65d0aa56856747603d1cc85c6514f

WooCommerce/WooCommerceTests/Yosemite/StoresManagerTests.swift
jkmassel: >It doesn't look like we need the init(concreteSessionManager:). Looks like this works just fine:

Yeah, I just got too excited trying to preserve the `.testingInstance` syntactic sugar 😜 

Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3305/commits/5d4e4e962ce69fdb4fa1133d9b72e90bc5e57374

> I'm worried about this static instance

Is this a static instance? I thought it would be functionally equivalent to `static func getTestingInstance() -> DefaultStoresManager` – the value only lives the lifetime of the test, then each test gets a new version because it's a `getter` rather than an actual variable?

Observables/README.md
shiki: Thanks! 😄

WooCommerce/WooCommerceTests/Yosemite/StoresManagerTests.swift
shiki: You're right. Yeah, I can't read Swift. 😅

shiki (approved):
Thanks, Jeremy! 👏

PR #3307, opened 2020-12-08
Show specific message when opening a deleted product

### LINKED ISSUE #2901: Detect deleted products
If a product is deleted outside of the app, you can still open it in product detail and there's nothing to let the user know the product no longer exists. We should instead show a dialog stating that the product no longer exists.

To do this, we need to detect when the request for a product returns a 404 with this response:

```
error: "woocommerce_rest_product_invalid_id"
message: "Invalid ID."
```

slack-imgs

eshurakov:
## What
This PR adds a specific message to the user in case they try to open a product with invalid id, which most likely means that the product is deleted.

We already had a separate screen to fetch a product (`ProductLoaderViewController`) which shows a failure overlay with retry button in case fetching wasn't successful. 

In case we know that the id we're trying to fetch is invalid it doesn't make much sense to retry. In this case we're reusing existing failure overlay, but with a different message and without retry button.

Closes #2901 

## Test
1. Create an order
2. Delete one of the products that's in the order (also from trash)
3. Open the order in the app
4. Tap on the product deleted in 2.

- [ ]  Check that the error message is shown saying that the product is deleted

## Screenshots

|||
|-|-|
|Simulator Screen Shot - iPhone 11 - 2020-12-08 at 14 58 17|Simulator Screen Shot - iPhone 11 - 2020-12-08 at 14 58 28|
|Simulator Screen Shot - iPhone 11 - 2020-12-08 at 14 58 34|Simulator Screen Shot - iPad Pro (9 7-inch) - 2020-12-08 at 15 01 41|

Update release notes:


Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -661,5 +664,27 @@ public enum ProductUpdateError: Error, Equatable {
...
+        guard case let DotcomError.unknown(code, _) = error else {
+            self = .unknown(error: error.toAnyError)
+            return
+        }
+
+        self = ErrorCode(rawValue: code)?.error ?? .unknown(error: error.toAnyError)
+    }
+
+    enum ErrorCode: String {
+        case invalidId = "woocommerce_rest_product_invalid_id"
```
jaclync: super nit: we generally say `ID` in the codebase 😄 

```suggestion
        case invalidID = "woocommerce_rest_product_invalid_id"
```

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
```diff
@@ -698,8 +698,10 @@ final class ProductStoreTests: XCTestCase {
         productStore.upsertStoredProduct(readOnlyProduct: sampleProduct(), in: viewStorage)
         XCTAssertEqual(viewStorage.countObjects(ofType: Storage.Product.self), 1)
 
+        let dotComError = DotcomError.unknown(code: ProductLoadError.ErrorCode.invalidId.rawValue, message: nil)
+
         // Action
-        network.simulateError(requestUrlSuffix: "products/282", error: NetworkError.notFound)
+        network.simulateError(requestUrlSuffix: "products/282", error: dotComError)
```
jaclync, line 704: is it possible to assert `result.failure` to `ProductLoadError.notFound` later in the test case?

Yosemite/Yosemite/Stores/ProductStore.swift
```diff
@@ -661,5 +664,27 @@ public enum ProductUpdateError: Error, Equatable {
 }
 
 public enum ProductLoadError: Error, Equatable {
+    case notFound
     case notFoundInStorage
+    case unknown(error: AnyError)
+
+    init(underlyingError error: Error) {
```
jaclync, line 671: nice!

jaclync (approved):
Nice enhancement! :shipit: 

Code: ✅ 
Testing: ✅

Garance91540 (changes requested):
Hi @eshurakov ! The illustration on this screen is more used for success states. For errors messages, can we use the one here attached? 
[illustrations 3.zip](https://github.com/woocommerce/woocommerce-ios/files/5664712/illustrations.3.zip)

eshurakov:
> Hi @eshurakov ! The illustration on this screen is more used for success states. For errors messages, can we use the one here attached?
> [illustrations 3.zip](https://github.com/woocommerce/woocommerce-ios/files/5664712/illustrations.3.zip)

Hey @Garance91540, of course we can update an image!
Would it be possible to get `pdf` instead of `svg`? iOS uses `pdf` for vector images.

Yosemite/YosemiteTests/Stores/ProductStoreTests.swift
eshurakov: Good point! Updated in https://github.com/woocommerce/woocommerce-ios/pull/3307/commits/756040231175160f70b239962c1117f2e2f62e36
Thanks!

jaclync:
@Garance91540 the assets look like the ones we use when the store does not support the new stats (`woo-no-store`): [light mode](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Resources/Images.xcassets/woo-no-store.imageset/woo-no-store-light.pdf) and [dark mode](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Resources/Images.xcassets/woo-no-store.imageset/woo-no-store-dark.pdf)

Screen Shot 2020-12-10 at 4 43 53 PM


do you think we can just use the same image for this scenario?

Garance91540:
> do you think we can just use the same image for this scenario?

Hi @jaclync , this is our "general failure" illustration, so we are using it in different places (error fetching orders state, login errors, analytics, etc) at the moment. It would be nice to have more illustrations in the future though, and I'm working on our illustration library this week so I'll keep that in mind!

Garance91540:
> Hey @Garance91540, of course we can update an image!
Would it be possible to get pdf instead of svg? iOS uses pdf for vector images.

Ah that's why I never share assets directly in GH 👯 , there's always something wrong :) 

I added those assets in the file "Orders - V1" in Figma (page "Order Details 2") https://d.pr/i/V6ocNR
You can export them directly with the Figma export tool

eshurakov:
Thank you @Garance91540 and @jaclync! We have 2 screens to show product loading errors. I've updated an image for both of them.
Could you take another look please? Thanks!

||||
|-|-|-|
|Network error|Simulator Screen Shot - iPhone 11 - 2020-12-10 at 12 47 58|Simulator Screen Shot - iPhone 11 - 2020-12-10 at 12 48 32|
|Deleted / Not Found|Simulator Screen Shot - iPhone 11 - 2020-12-10 at 12 50 05|Simulator Screen Shot - iPhone 11 - 2020-12-10 at 12 49 58|

WooCommerce/WooCommerceTests/Extensions/IconsTests.swift
```diff
@@ -364,4 +364,8 @@ final class IconsTests: XCTestCase {
     func test_wordPressLogoImage_is_not_nil() {
         XCTAssertNotNil(UIImage.wordPressLogoImage)
     }
+
+    func testErrorImageIsNotNil() {
```
jaclync: nit: we use snake case for test cases now based on the [doc](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/naming-conventions.md#test-methods) 😄 
```suggestion
    func test_errorImage_is_not_nil() {
```

Garance91540 (approved):
👏👏👏

jaclync (approved):
Looks great!

Non-blocking observation: there's a very minor difference between the light and dark mode (cc @Garance91540)

Screen Shot 2020-12-11 at 4 53 31 PM

WooCommerce/WooCommerceTests/Extensions/IconsTests.swift
eshurakov: Thank you! Resolved in https://github.com/woocommerce/woocommerce-ios/pull/3307/commits/e0ff816f2e45f294cc7b6ebd063078852c826f24

Garance91540:
> Non-blocking observation: there's a very minor difference between the light and dark mode (cc @Garance91540)

Hi @jaclync , I'm working on updating all the illustrations in our current library, so we'll have to replace all of these later on. With that in mind it's fine if we have a small discrepancy between light and dark mode for now

eshurakov:
Thank you @jaclync and @Garance91540!

PR #3310, opened 2020-12-09
Order Details > Shipping Label: track a shipping label

### LINKED ISSUE #2564: Shipping Labels M1: provide option to track a shipment in WCS
**Update**: we only support tracking a shipping label with the latest product requirements (same in Android). The user cannot delete a shipping label but refund one.

---

As part of the M1 release, we need to provide an option to:
- Track a shipment from the shipping label, with or without the Shipment Tracking plugin.
- Delete a shipment from the shipping label **only if the Shipment Tracking plugin is available**.

#### Tracking shipment
- **With the Shipment Tracking plugin available:** Calling the GET shipment tracking list API endpoint will provide the list of shipment trackings for the order. We can then match the shipping label tracking number with the shipment tracking number to display the corresponding tracking link.
- **Without the Shipment Tracking plugin:** If the shipment tracking plugin is not available, we won't be able to use the GET shipmen tracking API endpoint. But WCS still displays the tracking link in `wp-admin`.  We need to [store the tracking link for each carrier in the app and we just need append the tracking Id for that shipping label to that link](https://github.com/Automattic/woocommerce-services/blob/develop/client/extensions/woocommerce/woocommerce-services/views/shipping-label/tracking-link.js), if the tracking link icon is clicked from the app.

#### Deleting shipment
- **With the Shipment Tracking plugin available:** Currently, if the ST plugin is available, we provide the option to delete the shipment id from the app. We just need to call the same DELETE shipment tracking API to delete the shipment.
- **Without the Shipment tracking plugin:** Deleting the shipment option is only visible if the ST plugin is available. If not, this option should not be displayed in the app.

Screenshot 2020-07-28 at 11 45 21 AM

jaclync:
Fixes #2564 

## Why

After the store owner ships a package with a shipping label, we provide a tracking CTA for them to track the shipment.

## Changes

- Created `ShippingLabelTrackingURLGenerator` that generates a URL given a shipping label's `trackingNumber` and `carrierID` with unit tests
- In `OrderDetailsViewController`, generated URL for a shipping label and showed the tracking action that presents a webview if the URL is non-nil

## Testing

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> a shipping label package card should be shown after syncing
- Tap on the ellipsis menu in the tracking number row
- Tap "Track shipment" --> a webview should be shown with tracking information for the label

## Example screenshots

DHL Express | USPS
-- | --
simulator | simulator

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
@@ -395,8 +395,14 @@ private extension OrderDetailsViewController {
             self?.viewModel.dataSource.sendToPasteboard(shippingLabel.trackingNumber, includeTrailingNewline: false)
         }
 
-        actionSheet.addDefaultActionWithTitle(Localization.ShippingLabelTrackingMoreMenu.trackShipmentAction) { _ in
-            // TODO-2564: track shipment of a shipping label
+        // Only shows the tracking action when there is a tracking URL.
+        if let url = ShippingLabelTrackingURLGenerator.url(for: shippingLabel) {
+            actionSheet.addDefaultActionWithTitle(Localization.ShippingLabelTrackingMoreMenu.trackShipmentAction) { [weak self] _ in
+                guard let self = self else { return }
+                let safariViewController = SFSafariViewController(url: url)
+                safariViewController.modalPresentationStyle = .pageSheet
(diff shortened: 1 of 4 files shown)


ctarda (approved):
Code is 💯  (extra points for the tests).

Functionality is 💯 as well:


:shipit:

Garance91540 (approved):
👏

Issue #3224, opened 2020-11-30
[UL&S] Fix margins of custom UI in tablets


ctarda:
The custom UI provided to WPAuthenticator spreads horizontally to cover all the available space. 

It should respect the same margins WPAuthenticator inserts in the UL UI it provides by default.

Look at LoginViewController.setTableViewMargins

shiki:
I saw that this is already in the Done column in the Unified Login board. I'll close this now but feel free to reopen if I made a mistake. 😅

PR #3311, opened 2020-12-09
Make UL error view support rotation only for iPad, in order to blend in with the rest of WPAuthenticator

ctarda:
Closes #3286 

The Unified login flow supports device rotation only in tablets. However, the custom UI WooCommerce injects into WPAuthenticator for error handling is supporting rotation also in phones. 

## Changes

| Before | After |
| ---- | ---- |
|  |  |

## Testing
* Log out if necessary
* Attempt to log in again. Trigger an error (either by login in with a Store Address with a url that does not have Jetpack installed, or by logging in with a WordPress.com account, providing a non valid email address)
* Rotate the device. On an iPhone, the screen should not rotate. On a tablet, it should rotate, but the margins will be incorrect, as reported in #3224 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Navigation Exceptions/ULErrorViewController.swift
+++ b/WooCommerce/Classes/Authentication/Navigation Exceptions/ULErrorViewController.swift
@@ -17,6 +17,10 @@ final class ULErrorViewController: UIViewController {
     @IBOutlet private weak var errorMessage: UILabel!
     @IBOutlet private weak var extraInfoButton: UIButton!
 
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        return UIDevice.isPad() ? .all : .portrait
+    }
+
     init(viewModel: ULErrorViewModel) {
         self.viewModel = viewModel
         super.init(nibName: Self.nibName, bundle: nil)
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/Authentication/Navigation Exceptions/ULErrorViewController.swift
```diff
@@ -17,6 +17,10 @@ final class ULErrorViewController: UIViewController {
     @IBOutlet private weak var errorMessage: UILabel!
     @IBOutlet private weak var extraInfoButton: UIButton!
 
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        return UIDevice.isPad() ? .all : .portrait
```
jaclync: nit simplification:
```suggestion
        UIDevice.isPad() ? .all : .portrait
```

jaclync (approved):
LGTM! There's a super edge case when the device can still go to landscape mode in iPhone after tapping on "Need help finding the connected email?" CTA then rotate but I wouldn't worry about it 😆 



While testing on iPad Air 4gen iOS 14.1, the button container in the prologue screen seems to be not taking the full width. Just double checking if it's a known issue 🙂

Garance91540 (approved):
👏

ctarda:
Thanks for the review! I just pushed a commit to remove that return, and logged #3318 and #3319 . Will merge when CI gives me the ✅

PR #3314, opened 2020-12-09
Order Details > Products: show images for product variations

### LINKED ISSUE #2113: Order Details > Products: the image isn't shown for product variations

### LINKED ISSUE #3269: Product Images for variable products missing from Order details screen
**Describe the bug**
Some product images are missing from the Order Details screen. The images are displayed in the Product list though. It seems like the Products that do not display any image in the Order Details screen are Variable Products.

Order Details:

Product List:

Tapping on any Product displays a modal view, where the images are visible, no matter if it is a variable product or not.

**To Reproduce**
0. In your store dashboard, check the existing Products, make sure all have a Product Image set.
1. In your store dashboard (WooCommerce > Orders > Add order), create a new order.
2. Add a mix of Simple Products and Variable Products to the order
3. Create the order.
4. Navigate to the Orders tab in the iOS app. Select the new order. Notice images missing for Products with variations
5. Navigate to the Products tab. Notice how the Products that were missing images in the Order Details screen now display their associated image in the Products tab.

**Screenshots**

Order Details screen:

Products tab:

**Expected behavior**
I would expect the Product Images to be displayed in the Order Details screen

- [X] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).

**Mobile Environment**
Please include:
- Device: iPhone 12 Pro, iPhone 8
- iOS version: iOS 14.2.1
- WooCommerce iOS version: 5.6 (5.6.0.1)

jaclync:
Fixes #2113 
Fixes #3269

## Why

When configuring `ProductDetailsCellViewModel` for an `AggregateOrderItem` to display an order line item, we currently only look up the corresponding `Product` but not `ProductVariation`. This PR tries looking up for the corresponding `ProductVariation` when `AggregateOrderItem` is a variation (`variationID != 0`) and setting the `imageURL` explicitly.

## Changes

- In `OrderDetailsDataSource`, added a private helper to look up a `ProductVariation` given the product and variation ID. When configuring `ProductDetailsCellViewModel` for an `AggregateOrderItem` in `configureAggregateOrderItem`, try looking up for a `ProductVariation` or `Product` for the image URL
- Removed unused code (`.orderItem` row isn't used anymore, since we've consolidated to use `AggregateOrderItem`)

## Testing

Prerequisite: the store has an order that contains a variation, and the variation has an image

- Switch to the store if its products & variations haven't been loaded before, or log out and in to the store with (since we don't delete the products/variations locally)
- Go to the orders tab
- Tap on the order that contains a variable product --> the variation image should be loaded shortly without any actions
- Confidence check: go back to the orders tab, and tap on an order with a product (not variation) --> the product image should be loaded as before

## Example screenshots

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-09 at 17 29 45 | Simulator Screen Shot - iPhone 11 - 2020-12-09 at 17 44 19

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 -----
 - [***] Dropped iOS 12 support. From now we support iOS 13 and later. [https://github.com/woocommerce/woocommerce-ios/pull/3216]
 - [*] Fixed spinner appearance in the footer of orders list. [https://github.com/woocommerce/woocommerce-ios/pull/3249]
+- [*] In order details, the image for a line item associated with a variation is shown now after the variation has been synced. [https://github.com/woocommerce/woocommerce-ios/pull/3314]
 - [internal] Refactored Core Data stack so more errors will be propagated. [https://github.com/woocommerce/woocommerce-ios/pull/3267]
 
 
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -548,10 +548,16 @@ private extension OrderDetailsDataSource {
         cell.selectionStyle = .default
 
         let aggregateItem = aggregateOrderItems[indexPath.row]
-        let product = lookUpProduct(by: aggregateItem.productOrVariationID)
-        let itemViewModel = ProductDetailsCellViewModel(aggregateItem: aggregateItem,
-                                                        currency: order.currency,
-                                                        product: product)
+        let imageURL: URL? = {
+            guard let imageURLString = aggregateItem.variationID != 0 ?
+                    lookUpProductVariation(productID: aggregateItem.productID, variationID: aggregateItem.variationID)?.image?.src:
+                    lookUpProduct(by: aggregateItem.productID)?.images.first?.src else {
+                return nil
+            }
+            return URL(string: imageURLString)
+        }()
+        let itemViewModel = ProductDetailsCellViewModel(aggregateItem: aggregateItem.copy(imageURL: imageURL),
+                                                        currency: order.currency)
 
         cell.configure(item: itemViewModel, imageService: imageService)
     }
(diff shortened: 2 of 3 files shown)


Ecarrion (approved):
Tested and it's working ok!
Good job!

PR #3316, opened 2020-12-09
List Existing Attributes in Variations: ProductAttributeStore tests

pmusolino:
Part of #3181

After having implemented the Yosemite and the Storage layers for managing the Product Attributes, in this PR I added the unit tests for the `ProductAttributeStore` class, which had not been implemented due to the magnitude of the previous PR.

## Testing
- Just CI

Update release notes:

### DIFF
--- a/Yosemite/YosemiteTests/Stores/ProductAttributeStoreTests.swift
+++ b/Yosemite/YosemiteTests/Stores/ProductAttributeStoreTests.swift
@@ -119,30 +119,51 @@ final class ProductAttributeStoreTests: XCTestCase {
         XCTAssertEqual(storedProductAttributesCount, 0)
         XCTAssertTrue(result.isFailure)
     }
-//
-//    func testAddProductTagAddsStoredTagSuccessfulResponse() {
-//        // Given a stubed product tag network response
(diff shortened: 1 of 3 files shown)


Yosemite/YosemiteTests/Stores/ProductAttributeStoreTests.swift
```diff
@@ -0,0 +1,330 @@
...
+        super.tearDown()
+    }
+
+    func test_synchronize_product_attributes_returns_attributes_upon_successful_response() throws {
+        // Given
+        network.simulateResponse(requestUrlSuffix: "products/attributes", filename: "product-attributes-all")
+        XCTAssertEqual(storedProductAttributesCount, 0)
+
+        // When
+        let result: Result<[Networking.ProductAttribute], Error> = try waitFor { promise in
```
Ecarrion, line 67: FYI: I don't think is necessary to explicitly set the result's type. But it may help ups with the compile times 🤔

Ecarrion (approved):
Thank you a lot for adding unit tests 🙇

pmusolino:
Thank you so much for the review @Ecarrion !

PR #3317, opened 2020-12-09
Order Details > Products: update header background to match cell

### LINKED ISSUE #2662: Order Details Product Header in Dark Mode Has No Bottom Margin
**Describe the bug**
There is no space between the Product header and the product list. 

**To Reproduce**
1. Use Dark Mode.
2. Navigate to an Order.

**Screenshots**
image

**Expected behavior**
There should be a space like in light mode. Or perhaps it should be like in light mode where the background of the Product header is the same as the product list. 

**Mobile Environment**
Please include:
- Device: iPad Pro 11 inch
- iOS version: iPad OS 14 beta 4
- WooCommerce iOS version: 4.8.0.0 RC

eshurakov:
Closes #2662 

### Why
When looking at the order details in dark mode, there is no space between the Product header and the product list. It doesn't happen in light mode because background is the same for the Product header and the product list.
This PR updates background of the Product header to match the background of the product list, which makes it look like there is a padding in between :-)

The changes are only in dark mode, but I took screenshots in light mode just in case.

|Before|After|
|-|-|
|Simulator Screen Shot - iPhone 11 - 2020-12-09 at 15 46 12|Simulator Screen Shot - iPhone 11 - 2020-12-09 at 15 44 05|
|Simulator Screen Shot - iPhone 11 - 2020-12-09 at 15 46 14|Simulator Screen Shot - iPhone 11 - 2020-12-09 at 15 44 08|

### Test
1. Open an order details

- [ ] Check that the Product header and the product list have the same background

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/Section Headers/PrimarySectionHeaderView.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/Section Headers/PrimarySectionHeaderView.swift
@@ -74,7 +74,7 @@ private extension PrimarySectionHeaderView {
 
 private extension PrimarySectionHeaderView {
     enum Colors {
-        static let containerViewBackgroundColor = UIColor.basicBackground
+        static let containerViewBackgroundColor = UIColor.listForeground
     }
 }
 


Ecarrion (approved):
Nice catch! 🚀

ghost (approved):

eshurakov:
Thank you @Ecarrion and @adamzelinski!

Merging this :-)

PR #3320, opened 2020-12-10
Fix background color of Prologue buttons in landscape on iPad

ctarda:
Closes #3319 
Ref: wordpress-mobile/WordPressAuthenticator-iOS#539

| Before | After |
| ---- | ---- |
|  |  |
|  |  |

Also, in dark mode:

| Portrait | Landscape |
| ---- | ---- |
|  |  |

## Changes
* Point WPAuthenticator to the relevant branch
* Pass a background color for the offending view

## How to test
* Checkout the branch, run `bundle exec pod install`
* Log out if necessary. 
* Attempt to log in both on a phone and a tablet. Check that the background color extends all the way to the margins.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3321, opened 2020-12-10
Fix margins of custom UI in tablets

### LINKED ISSUE #3224: [UL&S] Fix margins of custom UI in tablets

ctarda:
Closes #3224 

| | Before | After |
| ---- | ---- | ---- |
| iPhone |  |  |
| iPad |  |  |

This is not a great solution, but it is probably the best we can do without making large changes to WPAuthenticator.

The difficult solution is:
* We provide our very own View Controller (`ULErrorViewController`) to WordPressAuthenticator so that it can use it to render custom error messages.
* WordPressAuthenticator is applying a custom margin to all its unified views. However, the view controller we provide will not be applied those margins.

The ideal solution would be, in my opinion:
1. `ULErrorViewController` exposes a public method (i.e. setMargins)
2. WordPressAuthenticator, passes a float to that public method when inserting ULErrorViewController into its navigation stack

However, being able to accommodate that will require an intense rewriting of parts of WPAuthenticator. 

All that being said...

## Changes
* Set the margins of the two main views of ULErrorViewController, mimicking the logic in WPAuthenticator's LoginPrologueViewController

## How to test
* Log out if necessary
* Attempt to log in on an iPad. Trigger an error (by, for example, attempting to login with a Store Address and entering GitHub.com). Notice the margins.
* Attempt to log in on an iPhone. Trigger an error, and notice the margins (which should extend to use all the available width)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Navigation Exceptions/ULErrorViewController.swift
+++ b/WooCommerce/Classes/Authentication/Navigation Exceptions/ULErrorViewController.swift
@@ -17,6 +17,9 @@ final class ULErrorViewController: UIViewController {
     @IBOutlet private weak var errorMessage: UILabel!
     @IBOutlet private weak var extraInfoButton: UIButton!
 
+    /// Constraints on the view containing the action buttons
+    /// and the stack view containing the image and error text
+    /// Used to adjust the button width in unified views provided by WPAuthenticator
     @IBOutlet private weak var buttonViewLeadingConstraint: NSLayoutConstraint!
     @IBOutlet private weak var buttonViewTrailingConstraint: NSLayoutConstraint!
     @IBOutlet private weak var stackViewLeadingConstraint: NSLayoutConstraint!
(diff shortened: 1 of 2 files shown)


Garance91540 (approved):
Thank you for caring about tablets users ❤️

pmusolino (approved):
Looks good to me @ctarda!

I just noticed an issue not related to this PR, on iPad:

Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2020-12-10 at 11 42 15

pmusolino:
I'm sorry, I saw that there is another PR open for this https://github.com/woocommerce/woocommerce-ios/pull/3320

Issue #2169, opened 2020-04-21
Shipping Labels M1 release: UI to print a shipping label


anitaa1990:
Dev plan: p91TBi-3QL-p2

- [x] Reprint shipping label screen with 4 CTAs: https://github.com/woocommerce/woocommerce-ios/pull/3341 and https://github.com/woocommerce/woocommerce-ios/pull/3342
  - Paper size options
  - Printing instructions (implemented in https://github.com/woocommerce/woocommerce-ios/issues/2174)
  - Paper size selector
  - Reprint shipping label
- [ ] Present printing instructions https://github.com/woocommerce/woocommerce-ios/pull/3397
- [ ] Present paper size options screen https://github.com/woocommerce/woocommerce-ios/pull/3397
  - [x] Build paper size options screen (static) https://github.com/woocommerce/woocommerce-ios/pull/3322
- [x] Paper size selector https://github.com/woocommerce/woocommerce-ios/pull/3355
- [x] Reprint shipping label action: makes an API call and shows the in-progress modal, then present the AirPrint dialog https://github.com/woocommerce/woocommerce-ios/pull/3374

If shipping labels are available for an order, we should provide an option to print the shipping label. Note that this[ Networking issue](https://github.com/woocommerce/woocommerce-ios/issues/2166) should be addressed first.

- We should provide an option for users to select the paper size once the Print Shipping label button is clicked.
- Once paper size is selected, an API request needs to be initiated to fetch PDF of the shipping label.
- Provide option to print this PDF from the app.

order details | reprint screen | paper size selector | paper sizes info screen | in-progress screen (API call) | airprint
-- | -- | --| -- | -- | --
[...]

PR #3322, opened 2020-12-10
Reprint Shipping Label: paper size options screen

jaclync:
Part of #2169 

## Why

We provide a screen that shows all supported paper sizes with an image so that it is easier to visualize how the label is laid out on each paper size. For Shipping Labels M1, the paper sizes are hard coded to 3 options for US store owners: legal, letter, and label.

## Changes

- Added 3 sets of image assets (light & dark version - 6 PDF assets in total): 10 files changed
- Implemented the static paper size options view with SwiftUI:
  - `GridStackView`: a generic grid view that is inspired by [this blog post](https://www.hackingwithswift.com/quick-start/swiftui/how-to-position-views-in-a-grid-using-lazyvgrid-and-lazyhgrid)'s iOS 13 fallback solution
  - `ShippingLabelPaperSizeOptionView`: a reusable view for a paper size option given a `ShippingLabelPaperSize`
  - `ShippingLabelPaperSizeOptionsView`: the top-level view of the paper size options given an array of `ShippingLabelPaperSize` (in the future, ideally we pass different paper sizes based on the store's geographical region. for example, A4 is more commonly used in some countries outside of the US)
- Created a hosting controller for the top-level SwiftUI view `ShippingLabelPaperSizeOptionsViewController` that configures the navigation bar

## Testing

You can play with the previews for `ShippingLabelPaperSizeOptionsView`, or add the following testing code to `OrderDetailsViewModel` where we handle the tap action on printing instructions:

https://github.com/woocommerce/woocommerce-ios/blob/616da875295af3131a64b8a85ed9137ba587e3b9/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsViewModel.swift#L241

testing code:

```swift
let paperSizeOptionsViewController = ShippingLabelPaperSizeOptionsViewController()
let navigationController = WooNavigationController(rootViewController: paperSizeOptionsViewController)
viewController.present(navigationController, animated: true, completion: nil)
```

## Example screenshots

Please note that the entry point in the screencast (printing instructions row) is just for testing purposes. After the reprint screen is implemented in https://github.com/woocommerce/woocommerce-ios/issues/2169, the entry point would be from the reprint screen!

simulator

dark | light
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-10 at 14 30 34 | Simulator Screen Shot - iPhone 11 - 2020-12-10 at 14 30 52
Simulator Screen Shot - iPhone 11 - 2020-12-10 at 14 31 26 | Simulator Screen Shot - iPhone 11 - 2020-12-10 at 14 31 51

Update release notes:


Garance91540 (approved):
Looking neat! 👏👏

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionView.swift
```diff
@@ -0,0 +1,65 @@
...
+                title = Localization.labelSizeTitle
+                image = PaperSizeImage.label
+            case .legal:
+                title = Localization.legalSizeTitle
+                image = PaperSizeImage.legal
+            case .letter:
+                title = Localization.letterSizeTitle
+                image = PaperSizeImage.letter
+            }
+            return .init(title: title, image: image)
```
Ecarrion: I have almost 0 experience in `SwiftUI` but it does feel a tad bit weird to create a VM in a View, is that the correct way to go? 😅

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionsView.swift
```diff
@@ -0,0 +1,51 @@
...
+
+/// Displays a grid view of all available paper size options for printing a shipping label.
+struct ShippingLabelPaperSizeOptionsView: View {
+    private let paperSizeOptions: [ShippingLabelPaperSize]
+    private let numberOfColumnsPerRow = 2
+    private let numberOfRows: Int
+
+    init(paperSizeOptions: [ShippingLabelPaperSize]) {
+        self.paperSizeOptions = paperSizeOptions
+        self.numberOfRows = Int(ceil(Double(paperSizeOptions.count) * 1.0 / Double(numberOfColumnsPerRow)))
```
Ecarrion: Maybe this can be extracted to a function? Also, why are we multiplying by 1?

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionsViewController.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    required dynamic init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+private extension ShippingLabelPaperSizeOptionsViewController {
+    func configureNavigationBar() {
+        navigationItem.title = Localization.navigationBarTitle
+        navigationItem.leftBarButtonItem = UIBarButtonItem(image: .closeButton, style: .plain, target: self, action: #selector(closeButtonTapped))
```
Ecarrion: Should we use `addCloseNavigationBarButton` helper ?

Ecarrion (dismissed):
I'm not approving this because I feel I don't have enough SwiftUI knowledge to know if everything is correct 😅 

I did found some issues while using big fonts, feel free to address it in a different PR.

Screen Shot 2020-12-10 at 10 49 59 AM

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionView.swift
jaclync: Do you think it'd be better just to pass the configurable UI properties to the initializer, or a different name like the equivalent concept "Props" in React? Open to suggestions @woocommerce/ios-developers  😄

Ecarrion: I guess it's weird because in a fully `SwiftUI` world there are not VCs(right?) so this kind of code needs to go somewhere, but since we are mixing `UIKit` and `SwiftUI` I'm not totally sure how the responsibilities should be split 🤔 

In this specific case, I guess it does not hurt much since the `VM` is simple and does not contain any business logic, but I would also want to hear what the more experienced members think!

ctarda: I think that, as Ernesto mentioned, a view should not try to instantiate its own view model. 

ShippingLabelPaperSizeOptionView is instantiated by ShippingLabelPaperSizeOptionsView (by the way, it took me a while to figure out where those two names differ, maybe there is a better way to name one of the two, so that name are more clearly differentiable). 

So, if we wanted to pass a view model to ShippingLabelPaperSizeOptionView, the parent view would have to either a) create that view model, which would put us in the same position, or b) get injected a view model that can spin a view model for a single paper size.

Now, ShippingLabelPaperSizeOptionView creates the view model to encapsulate the logic that decides which title and which image it should render. 

If we encapsulated that logic in an extension to ShippingLabelPaperSize that was declared in the UI layer (something like):

```
protocol ShippingLabelPaperSIzePresentable {
var title: String { get }
var image: Image { get }
}
extension ShippingLabelPaperSize:  ShippingLabelPaperSIzePresentable {
var title: String { 
switch self.label
.....
}
}
```

We would be able to remove that logic from the UI.

Now, the next step would be that we could still initialize ShippingLabelPaperSizeOptionView passing it a ShippingLabelPaperSize or we could choose to declare the reference to  ShippingLabelPaperSize as a state variable in the parent's view. But that might be too much.

So, my suggestion would be:
1. Encapsulate the logic to decide which title and which image to use in an extension to ShippingLabelPaperSize declared in the UI layer, and that acts as kind of a presenter for ShippingLabelPaperSize
2. Pass only an instance of ShippingLabelPaperSize to ShippingLabelPaperSizeOptionView
3. Let ShippingLabelPaperSizeOptionView consume the methods/vars in that extension 

Wha do you think @jaclync ?

ctarda (commented):
@jaclync 

It looks good to me in general. I think there is a way out of declaring a view model within the details view, I would love to hear your opinion 😊

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionView.swift
jaclync: Thanks both for the feedback! I agree that `ViewModel` is quite odd in SwiftUI, since we generally use it in the "view controller - view model" context in UIKit. If we ever encounter a use case when the number of configurable properties is large, we can come up with some other name specific to the use case.

What do you think of just removing the `ViewModel` and instantiating the `title` and `image` in the initializer like in https://github.com/woocommerce/woocommerce-ios/pull/3322/commits/2456de1fe0a325d8845837a0f85ef79c39f4a188?

jaclync: > ShippingLabelPaperSizeOptionView is instantiated by ShippingLabelPaperSizeOptionsView (by the way, it took me a while to figure out where those two names differ, maybe there is a better way to name one of the two, so that name are more clearly differentiable).

I renamed `ShippingLabelPaperSizeOptionsView` to `ShippingLabelPaperSizeOptionListView` in https://github.com/woocommerce/woocommerce-ios/pull/3322/commits/b2215b500e55f8ee6a3148d7502106ca5ee4d8ee, hopefully they are more distinguishable now 😄

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionsView.swift
jaclync: Ah it was a habit to multiple by 1.0 to make sure it's a floating number, but it isn't necessary here since the `Int` values are already cast to `Double`. Removed in: https://github.com/woocommerce/woocommerce-ios/pull/3322/commits/520d99d926516915236b10b0429c77b8de632fd6

> Maybe this can be extracted to a function?

it's a bit tricky to call a function in the initializer, and it's mostly casting / handling two numbers. but lemme know if you think any way of writing would make it easier to read!

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionsViewController.swift
jaclync: Makes sense! Updated in https://github.com/woocommerce/woocommerce-ios/pull/3322/commits/7e67eccacf8ef02259874996d0886ed08259cec4 (also updated the testing code in PR description)

jaclync:
@woocommerce/ios-developers would be great having any of your review (again)! Since it's one of the first use cases we have for SwiftUI, feel free to suggest anything regardless of SwiftUI experience.

> I did found some issues while using big fonts, feel free to address it in a different PR.

@Ecarrion thanks for bringing up the issues with larger font sizes! I added a commit to make the text any number of lines in https://github.com/woocommerce/woocommerce-ios/pull/3322/commits/0fcc222c50700a32d8e4641ebb540cb1aa009bc7, and will work on the rest of the design issues (some cells losing leading margin) separately in https://github.com/woocommerce/woocommerce-ios/issues/3334 since the PR also had design review approved.

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionListView.swift
```diff
@@ -0,0 +1,51 @@
...
+    private let numberOfColumnsPerRow = 2
+    private let numberOfRows: Int
+
+    init(paperSizeOptions: [ShippingLabelPaperSize]) {
+        self.paperSizeOptions = paperSizeOptions
+        self.numberOfRows = Int(ceil(Double(paperSizeOptions.count) / Double(numberOfColumnsPerRow)))
+    }
+
+    var body: some View {
+        ScrollView {
```
shiki, line 16: Not today: I wonder if we could also make a separate `WooScrollView` component which is already configured with the standard background that we use.

shiki (commented):
I have nothing good to add here since I don't have experience with Swift UI. The code looks simple enough for me that I don't worry about it. 🙂

ctarda (approved):
Thanks for iterating on the PR @jaclync 

I think it looks great. I agree with @shiki  that it might be a good idea to extract a ScrollView configured with Woo's visual defaults, as that seems like it's something that we are going to need later, but it's definitely not a blocker.

:shipit:

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ShippingLabelPaperSizeOptionListView.swift
jaclync: Great idea! I created a separate ticket to document SwiftUI adoption and create custom components in https://github.com/woocommerce/woocommerce-ios/issues/3379 for future improvements 😃

jaclync:
Thanks for the review César and Jayson! Merging the PR now, will document and create our custom scroll view component separately in https://github.com/woocommerce/woocommerce-ios/issues/3379!

PR #3326, opened 2020-12-10
Customers fetching networking

### LINKED ISSUE #3241: Order Creation: networking, yosemite and storage layers for fetching and store Customers (step 4)
As part of the Order Creation, we should implement the networking, Yosemite, and storage layers for fetching every customer.

Here is possible to find the [API documentation](https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-customers).

ealeksandrov:
Part of #3241.

### Changes

This PR adds
- `Customer` entity
- `Customer.Role` nested enum case (with custom values handling)
- `CustomersListMapper` to decode array of customers (+ tests)
- `CustomerRemote` to request network (+ tests)

Sorry for >500 lines of changes, but it's mostly lightweight - including test cases and test JSON.

### Testing

Just check the code. No actions exposed to UI yet.

### Questions

1. `siteID` added because I see it injected in many other entities. Is it correct decision?

### DIFF
--- a/Networking/Networking/Model/Customer.swift
+++ b/Networking/Networking/Model/Customer.swift
@@ -1,0 +1,143 @@
+import Foundation
+
+/// Represent a Customer Entity.
+///
+public struct Customer: Decodable {
+
+    public let siteID: Int64
+    public let userID: Int64
+
+    public let dateCreated: Date    // gmt
+    public let dateModified: Date?  // gmt
+
+    public let email: String
+    public let username: String?
+    public let firstName: String?
+    public let lastName: String?
+    public let gravatarUrl: String?
+
+    public let isPaying: Bool
+
+    public let billingAddress: Address?
+    public let shippingAddress: Address?
+
+
+    /// Struct initializer for Customer.
+    ///
+    public init(siteID: Int64,
+                userID: Int64,
+                dateCreated: Date,
+                dateModified: Date?,
+                email: String,
+                username: String?,
+                firstName: String?,
+                lastName: String?,
+                gravatarUrl: String?,
+                isPaying: Bool,
+                billingAddress: Address?,
+                shippingAddress: Address?) {
+        self.siteID = siteID
+        self.userID = userID
+
+        self.dateCreated = dateCreated
+        self.dateModified = dateModified
+
+        self.email = email
+        self.username = username
+        self.firstName = firstName
+        self.lastName = lastName
+        self.gravatarUrl = gravatarUrl
+
+        self.isPaying = isPaying
+
+        self.billingAddress = billingAddress
+        self.shippingAddress = shippingAddress
+    }
+
+    /// The public initializer for Customer.
+    ///
+    public init(from decoder: Decoder) throws {
+        guard let siteID = decoder.userInfo[.siteID] as? Int64 else {
+            throw CustomerDecodingError.missingSiteID
+        }
+
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        let userID = try container.decode(Int64.self, forKey: .userID)
+
+        let dateCreated = try container.decodeIfPresent(Date.self, forKey: .dateCreated) ?? Date()
+        let dateModified = try container.decodeIfPresent(Date.self, forKey: .dateModified)
+
+        let email = try container.decode(String.self, forKey: .email)
+        let username = try container.decodeIfPresent(String.self, forKey: .username)
+        let firstName = try container.decodeIfPresent(String.self, forKey: .firstName)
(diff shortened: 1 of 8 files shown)


Networking/Networking/Model/Customer.swift
```diff
@@ -0,0 +1,151 @@
...
+                  billingAddress: billingAddress,
+                  shippingAddress: shippingAddress)
+    }
+}
+
+// MARK: - Equatable Conformance
+//
+extension Customer: Equatable {
+
+    public static func == (lhs: Customer, rhs: Customer) -> Bool {
```
pmusolino: We should be able to conform to Equatable without declaring a custom equatable method.

Networking/Networking/Model/Customer.swift
```diff
@@ -0,0 +1,151 @@
...
+    public let userID: Int64
+
+    public let dateCreated: Date    // gmt
+    public let dateModified: Date?  // gmt
+
+    public let email: String
+    public let username: String?
+    public let firstName: String?
+    public let lastName: String?
+    public let gravatarUrl: String?
```
pmusolino: Is the avatar always coming from gravatar? I'd say to mantaing a generic `avatarUrl` name.

Networking/Networking/Remote/CustomerRemote.swift
```diff
@@ -0,0 +1,30 @@
...
+    ///     - siteID: Site for which we'll fetch remote customers.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func getAllCustomers(for siteID: Int64,
+                                completion: @escaping (Result<[Customer], Error>) -> Void) {
+        let path = Path.customers
+        let request = JetpackRequest(wooApiVersion: .mark3, method: .get, siteID: siteID, path: path)
+        let mapper = CustomersListMapper(siteID: siteID)
+        enqueue(request, mapper: mapper, completion: completion)
+    }
```
pmusolino: This seems partially implemented, since there are some parameters missing (for example, is not possible to do any sort of pagination. Will be impossible for us to requests all the customers (imagine what can happens with a store that has thousands of customers).

Networking/NetworkingTests/Remote/CustomerRemoteTests.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    /// Dummy Site ID
+    let sampleSiteID: Int64 = 1234
+
+    override func setUp() {
+        super.setUp()
+        network.removeAllSimulatedResponses()
+    }
+
+    func test_getAllCustomers_returns_parsed_customers() throws {
```
pmusolino, line 20: Looks good to me. What do you think about adding another test when we obtain a network error, just to see if everything continue to work as expected?
Something like this: `test_getAllCustomers_relays_networking_error()`

pmusolino (commented):
In general it looks good to me 😃 I added just some nits, but the most important is that you should handle the pagination inside the API request for getting all the customers.

Networking/Networking/Model/Customer.swift
ealeksandrov: I'm not sure and will agree with you 🙂
Inspired by same mapping in `Account`:

https://github.com/woocommerce/woocommerce-ios/blob/073e714c884a5e0de3cb4515901ca6d9a2ec0d76/Networking/Networking/Model/Account.swift#L50

Networking/Networking/Remote/CustomerRemote.swift
ealeksandrov: Ah, totally right! Delayed pagination part after verifying basic stuff in Storage+Yosemite and completely forgot about it 🙂

Networking/Networking/Model/Customer.swift
ealeksandrov: Fixed in 8b391ff

Networking/Networking/Model/Customer.swift
ealeksandrov: Fixed in 9aea6a5

Networking/NetworkingTests/Remote/CustomerRemoteTests.swift
ealeksandrov: Added in cbc74ab

Networking/Networking/Remote/CustomerRemote.swift
ealeksandrov: Added optional parameters for pagination (`page` and `per_page`) and `context` in 1437120.
Other parameters available from API doc seem unneeded to me for now.

pmusolino (approved):
Thank you @ealeksandrov :shipit: looks good to me

PR #3335, opened 2020-12-11
Add Product Variations: List Existing Attributes

### LINKED ISSUE #3181: Variations: List Existing Attributes

pmusolino:
Fixes #3181 

## Description
After having implemented the [Networking layer](https://github.com/woocommerce/woocommerce-ios/pull/3290), and after the changes to the [Storage and Yosemite layers](https://github.com/woocommerce/woocommerce-ios/pull/3299) to handle global Product Attributes, in this PR we display the existing local and global Attributes in the new "Add Attribute" screen, part of the Add Product Variation flow (M5).

## Changes
- Now in Storage we delete unused Product Attributes when we synchronize all the global Product Attributes when they are not connected to a product.
- Now it’s possible to load all the `ProductAttribute`s for a specific `siteID`.
- Implemented the `AddAttributeViewController` that display the product attributes plus a text field on top where the user can add a new attribute.
- Implemented `AddAttributeViewModel` that provides view data for the Add Attribute screen, and handles init/UI/navigation actions needed. + unit tests
- Added `ProductAttributeStore` inside the list of `AuthenticatedState`.

## Testing
1. Open a Variable Product without variations.
2. Tap the cell "Add variations".
3. The new screen "Add Attribute" should be presented.
4. In the "Add Attribute" screen you should see the list of global attributes + local product attributes. No other actions are allowed at the moment.

## Screenshot

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
```diff
@@ -0,0 +1,298 @@
...
+
+    /// Keyboard management
+    ///
+    private lazy var keyboardFrameObserver: KeyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
+        self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
+    }
+
+    /// Init
+    ///
+    init(product: Product) {
```
ctarda: Would it make sense to pass the view model directly? That way this view controller wouldn't have to instantiate its own view model, which seems a bit off.

Anyway, it doesn't look like the view controller uses the product directly, so maybe it wouldn't need to retain it?

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
```diff
@@ -0,0 +1,298 @@
...
+    }
+
+    func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        guard let leftText = viewModel.sections[section].header else {
+            return nil
+        }
+
+        let headerID = TwoColumnSectionHeaderView.reuseIdentifier
+        guard let headerView = tableView.dequeueReusableHeaderFooterView(withIdentifier: headerID) as? TwoColumnSectionHeaderView else {
+            fatalError()
```
ctarda: If I understand correctly, this would stop execution also in production. Would it make sense to throw an assertion, if the reason it's here is to warn developers that a view is not properly registered?

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
```diff
@@ -0,0 +1,298 @@
...
+    /// Cells currently configured in the order they appear on screen
+    ///
+    func configure(_ cell: UITableViewCell, for row: Row, at indexPath: IndexPath) {
+        switch cell {
+        case let cell as TextFieldTableViewCell where row == .attributeTextField:
+            configureTextField(cell: cell)
+        case let cell as BasicTableViewCell where row == .existingAttribute:
+            configureAttribute(cell: cell, attribute: viewModel.localAndGlobalAttributes[safe: indexPath.row])
+        default:
+            fatalError()
```
ctarda, line 213: I was going to comment about this fatalError as well, but I searched the codebase first and I noticed that's how we do it 😊 Sorry for the previous comment ...

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewModel.swift
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /// Updates  data from the remote fetched objects.
+    ///
+    func updateSections(attributes: [ProductAttribute]) {
+
+        /// Sum fetched attributes + attributes inside the product (global + local), and remove duplicated product attributes, then sort it
+        localAndGlobalAttributes = attributes + product.attributes.filter { element in
+            return !attributes.contains { $0.attributeID == element.attributeID }
+        }.sorted(by: { (attribute1, attribute2) -> Bool in
```
ctarda: As far as I know, in theory this could be written as `sorted(by: > )`, I am not sure that makes it more readable though

ctarda (approved):
The code looks great, as usual.

There seems to be a bit of an edge case, that I don't think is a blocker but that looks like could be a simple fix before merging as well, if so you decided. 

When presenting variations for a product that does not have any variations assigned, this is what I see (notice the ghosting in the top right corner):



👆This corresponds with a product setup like this:



If the product has some "active" variations, like this:


Then the ghosting works perfect, and this would be the end result:

Garance91540 (approved):
yeees!👏

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
pmusolino: Thank you @ctarda.

You are right, it's better that the viewmodel is instantiated outside. I followed your suggestions and made the changes here df5ccf25a76277e0d2ac3a831b0953f3b96d4327

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewModel.swift
pmusolino: wow good suggestion, I implemented it 💯 8cf059f555614b9c1b47b8d207517780d4184d36

pmusolino:
Thank you both @ctarda and @Garance91540 for the review.

@ctarda I fixed the two small issues, the one about the case where there are no existing attributes 25c3718dde561c76e23b7d2ceb3a74cf63471211 and the one about the ghost cell 18660bebd6160a5b0e1659d64fe34a989dc38a3f 👍

Issue #1765, opened 2020-01-23
Enable mocking `StorageManagerType` in unit tests


jaclync:
From https://github.com/woocommerce/woocommerce-ios/pull/1762:

It looks like there isn't a way to mock the storage manager in unit tests, even though we do have an interface `StorageManagerType`. Let's make the `ServiceLocator` take a `StorageManagerType` instead of the implementation `CoreDataManager` so that we can pass a [`MockupStorageManager`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/YosemiteTests/Mockups/MockupStorage.swift) in unit tests.

https://github.com/woocommerce/woocommerce-ios/blob/21c3ebdd7a914a4f70b4cb0e0217e390d18b37ea/WooCommerce/Classes/ServiceLocator/ServiceLocator.swift#L186-L192

JorgeMucientes:
Closing this as part of a maintenance pass on the GitHub issue backlog. This issue matched the auto-close criteria for this pass: it has not been updated since 2021-02-25, is unassigned, is not labeled as a bug or high/medium priority, and does not appear to involve POS, payments, login, data loss, crash, security, vulnerability, auth, or fraud.

If this is still relevant, please comment with current details and we can reopen it or move it into Linear for triage.

PR #3337, opened 2020-12-11
Add screenshot mocks

jkmassel:
Add support for screenshot mocks.

**To Test:**
- Ensure that all screenshots CI tasks pass

There's a _lot_ of code here – the majority of it is the mocks themselves, which aren't particularly interesting.

@shiki – this follows the pattern that you laid out previously, so I'm not sure if you want to take another look at it from an architectural standpoint? I think we'd talked about punting this all into a separate Swift package, which I'm ok to do if you'd like to, but I wasn't sure if it'd still apply given the move to Yosemite?

@rachelmcr and @JavonDavis  – you've previously worked on screenshots, so tagging you though you're both optional :) 

@woocommerce/release-managers – hoping at least one of you can take a detailed look through this (sorry!!)


.github/workflows/screenshots.yml
```diff
@@ -73,24 +63,13 @@ jobs:
...
-        key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
-    
-    - name: Install Ruby Dependencies
-      run: |
-        bundle config path vendor/bundle
-        bundle install --jobs 4 --retry 3
+        bundler-cache: true
+
+    - name: Select Xcode Version
+      run: sudo xcode-select --switch /Applications/Xcode_12.app
```
AliSoftware: I wonder if we shouldn't instead use [fastlane's `xcversion` action](https://docs.fastlane.tools/actions/xcversion/) – which automatically ensures the right version of Xcode is available on the machine and used by subsequent calls to `xcodebuild`?

That would make the Xcode version also enforced locally (not just on CI), and prevent the need of this (and of `sudo`) in CI.

.github/workflows/screenshots.yml
```diff
@@ -137,42 +108,65 @@ jobs:
...
-      uses: actions/upload-artifact@v1
-      with:
-        name: raw-screenshots
-        path: fastlane/screenshots
+      - uses: actions/checkout@v1
+
+      - name: Install Native Dependencies
+        run: |
+          brew install imagemagick@6
+          brew link imagemagick@6 --force
```
AliSoftware: nit: any reason why we need to use v6 here? (not that this PR should update to v7, as this can totally be done in a separate PR later if it's worth it rather than here; but was just wondering if you already looked at what currently prevents us from updating easily)

.github/workflows/screenshots.yml
```diff
@@ -137,42 +108,65 @@ jobs:
...
+      - name: Install Screenshot Gems
+        run: bundle install --with screenshots
+
+      - name: Install Fastlane Dependencies
+        run: bundle exec fastlane run configure_apply
+
+      - name: Download Generated Screenshots
+        run: |
+          cd fastlane
+          aws s3 cp s3://$S3_BUCKET/2020-12-14-final/screenshots screenshots/  --recursive --exclude "*.html"
```
AliSoftware: Will the run ID always be the same/unchanged? 🤔  (or is this a leftover of one of your debug session?)

.github/workflows/screenshots.yml
```diff
@@ -137,42 +108,65 @@ jobs:
...
+          bundle exec fastlane create_screenshot_summary
+          aws s3 cp fastlane/screenshots/screenshots.html s3://$S3_BUCKET/$GITHUB_RUN_ID/screenshots/screenshots.html
+
+      - name: Archive Raw Screenshots
+        uses: actions/upload-artifact@v1
+        with:
+          name: raw-screenshots
+          path: fastlane/screenshots
+
+      - name: Install Promo Screenshot Fonts
```
AliSoftware, line 138: YES! The risk with missing fonts for screenshots always bothered me in the past… Nice addition 👍

WooCommerce/Classes/AppDelegate.swift
```diff
@@ -10,6 +10,8 @@ import WordPressKit
 import WordPressAuthenticator
 import AutomatticTracks
 
+import class Yosemite.ScreenshotStoresManager
```
AliSoftware, line 13: I like the targeted import there 👍

WooCommerce/Classes/System/BuildConfiguration.swift
```diff
@@ -32,4 +34,16 @@ enum BuildConfiguration: String {
...
+        return ProcessInfo.processInfo.arguments.contains("mocked-network-layer")
+    }
+
+    static var shouldLogoutAtLaunch: Bool {
+        ProcessInfo.processInfo.arguments.contains("logout-at-launch")
+    }
+
+    static var shouldDisableAnimations: Bool {
+        ProcessInfo.processInfo.arguments.contains("disable-animations")
+    }
```
AliSoftware, line 48: Nit: At some point we should use Apple's recommended way to pass launch options instead – i.e. use `-mocked-network-layer` with a dash prefix then use `UserDefaults` to read them from the Argument Domain.
Not critical, but would be good practice.

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerceScreenshots.xcscheme
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
    LastUpgradeVersion = "1210"
-   version = "1.7">
+   version = "1.3">
```
AliSoftware, line 4: Is this expected? 🤔

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerceScreenshots.xcscheme
```diff
@@ -11,24 +11,6 @@
...
-                     BuildableIdentifier = "primary"
-                     BlueprintIdentifier = "F997170123DBB97500592D8E"
-                     BuildableName = "WooCommerceScreenshots.xctest"
-                     BlueprintName = "WooCommerceScreenshots"
-                     ReferencedContainer = "container:WooCommerce.xcodeproj">
-                  </BuildableReference>
-               </EnvironmentBuildable>
-            </ActionContent>
-         </ExecutionAction>
-      </PreActions>
```
AliSoftware, line 31: Doesn't this stop from using a mocked statusbar content (9:41, cellular icon, battery, etc) that we used to do for consistent screenshots?

WooCommerce/WooCommerceScreenshots/Utils/SnapshotHelper.swift
```diff
@@ -38,22 +38,13 @@ func snapshot(_ name: String, timeWaitingForIdle timeout: TimeInterval = 20) {
 }
 
 enum SnapshotError: Error, CustomDebugStringConvertible {
-    case cannotDetectUser
-    case cannotFindHomeDirectory
     case cannotFindSimulatorHomeDirectory
-    case cannotAccessSimulatorHomeDirectory(String)
     case cannotRunOnPhysicalDevice
```
AliSoftware, line 42: ℹ️ [Note for reviewers]: this whole file is not ours, it's provided by fastlane.

_(I lost some time reviewing it and trying to understand why some code unrelated to our use cases where changed in this PR, before realizing this)_

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
```diff
@@ -54,6 +60,67 @@ class WooCommerceScreenshots: XCTestCase {
             .selectProduct(atIndex: 1)
             .thenTakeScreenshot(named: "product-details")
     }
+
+    private let port = Int.random(in: 2048...65535)
```
AliSoftware: I think we should put a mechanism in place to try the next port if the randomly-selected one fails.

For example, if it randomly selects port 12345 but that port is already used by some other app/process running on the machine, `server.start()` will fail with "port unavailable"; in those cases we should at least try the like the 5 next ports 12346...12350, until we find one working or bail.

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
```diff
@@ -54,6 +60,67 @@ class WooCommerceScreenshots: XCTestCase {
...
+            /// Run event loop
+            self.loop.runForever()
+
+            debugPrint("Web Server running on port \(self.port)")
+        }
+    }
+
+    func handleWebRequest(env: [String: Any], startResponse: ((String, [(String, String)]) -> ()), sendBody: ((Data) -> ())) {
+        /// Extract the path, so http://localhost:9285/foo-bar would be `/foo-bar` here.
+        guard let path = env["PATH_INFO"] as? String else {
```
AliSoftware, line 89: Always found it odd that Embassy used an untyped dictionary to access those values, as opposed to a type-checked structure, to expose those values; feels quite unnatural in Swift…

Maybe worth a comment to at least point to doc:
```suggestion
        // See https://github.com/envoy/Embassy#whats-swsgi-swift-web-server-gateway-interface for more about PATH_INFO
        guard let path = env["PATH_INFO"] as? String else {
```

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
```diff
@@ -54,6 +60,67 @@ class WooCommerceScreenshots: XCTestCase {
...
+            startResponse("404 Not Found", [])
+            sendBody(Data())
+            return
+        }
+
+        /// Remove the leading `/`, so we're left with `foo-bar`
+        let assetName = path.replacingOccurrences(of: "/", with: "")
```
AliSoftware: This feels a bit dangerous, especially if we start using namespaced assets at some point, e.g.
namespaced-xcasset
(which then are referenced in the code by `UIImage(named: "icons/icon_add")` using their fully qualified name)

Maybe this could be safer instead:
```suggestion
        let assetName = path.replacingOccurrences(of: "^/", with: "", options: .regularExpression)
```

WooCommerce/WooCommerceUITests/Screens/BaseScreen.swift
```diff
@@ -20,6 +20,10 @@ class BaseScreen {
     @discardableResult
     func waitForPage() throws -> BaseScreen {
         XCTContext.runActivity(named: "Confirm page \(self) is loaded") { (activity) in
```
AliSoftware, line 22: ❤️  at the use of `XCTContext.runActivity(…)` 🙌

Yosemite/Yosemite/Internal/String+Extensions.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+extension String {
+
+    private static let slugSafeCharacters = CharacterSet(charactersIn: "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-")
+
+    public var slugified: String? {
+        guard let latin = self.applyingTransform(StringTransform("Any-Latin; Latin-ASCII; Lower;"), reverse: false) else {
```
AliSoftware, line 8: ❤️  ICU transforms FTW 👌

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockActionHandler.swift
```diff
@@ -0,0 +1,56 @@
...
+}
+
+extension MockActionHandler {
+
+    /// A helper method to keep ActionHandler switch statements clean
+    func unimplementedAction<T>(action: T) where T: Action {
+        fatalError("Unable to handle action: \(action.identifier) \(String(describing: action))")
+    }
+
+    /// A helper for that immediately returns a success for `Result<Void>` closure patterns
```
AliSoftware: nit:
```suggestion
    /// A helper that immediately returns a success for `Result<Void>` closure patterns
```

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockActionHandler.swift
```diff
@@ -0,0 +1,56 @@
...
+    func success(_ onCompletion: (Result<(), Error>) -> Void) {
+        onCompletion(.success(()))
+    }
+
+    /// A helper for that immediately returns a success for `Error?` closure patterns
```
AliSoftware: nit:
```suggestion
    /// A helper that immediately returns a success for `Error?` closure patterns
```

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockAppSettingsActionHandler.swift
```diff
@@ -0,0 +1,40 @@
...
+            default: unimplementedAction(action: action)
+        }
+    }
+
+    func loadInitialStatsVersionToShow(siteId: Int64, onCompletion: (StatsVersion?) -> Void) {
+        onCompletion(.v4)
+    }
+
+    func setStatsVersionLastShown(siteId: Int64, statsVersion: StatsVersion) {
+        // TODO – we should persist this somewhere
```
AliSoftware: nit: 💡 wdyt about using a warning to not forget about this later?
```suggestion
        #warning("TODO – we should persist this somewhere")
```

(And same for other `// TODO – xxx` comments I've seen in other mocks ofc)

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockOrderActionHandler.swift
```diff
@@ -0,0 +1,48 @@
...
+        let storage = storageManager.newDerivedStorage()
+
+        storage.perform {
+            let updater = OrdersUpsertUseCase(storage: storage)
+            updater.upsert(orders)
+        }
+
+        storageManager.saveDerivedType(derivedStorage: storage) {
+            DispatchQueue.main.async(execute: onCompletion)
+        }
```
AliSoftware, line 46: nit: felt off-balance that we needed to dispatch the completion to main here but not in the other cases. Wondering if it wouldn't be better to make `saveDerivedType` dispatch its own completion on main itself 🤔 

Probably not a big deal; but this surprised me when reading the code during review and I can see this dispatch easily be missed in other call sites when people don't realize this… 🤷

Yosemite/Yosemite/Model/Mocks/Helpers/ProductStockStatus+Mocks.swift
```diff
@@ -0,0 +1,16 @@
...
+    public static func from(quantity: Int64) -> ProductStockStatus {
+        switch quantity {
+            case Int64.min ..< 0:
+                return .onBackOrder
+            case 0:
+                return .outOfStock
+            case 1...Int64.max:
+                return .inStock
+            default:
+                preconditionFailure("Unable to parse quantity")
```
AliSoftware: nit: you can leave out the extreme bounds. I find it clearer that way ;)

```suggestion
            case ..<0:
                return .onBackOrder
            case 0:
                return .outOfStock
            case 1...:
                return .inStock
            default:
                preconditionFailure("Unable to parse quantity")
```

Yosemite/Yosemite/Model/Mocks/MockObjectGraph.swift
```diff
@@ -0,0 +1,481 @@
...
+}
+
+let mockResourceUrlHost = "http://localhost:\(UserDefaults.standard.integer(forKey: "mocks-port"))/"
+
+// MARK: Product Accessors
+extension MockObjectGraph {
+
+    func product(forSiteId siteId: Int64, productId: Int64) -> Product {
+        return products(forSiteId: siteId).first { $0.productID == productId }!
+    }
```
AliSoftware, line 37: Thinking out loud: seeing this force-unwrap I'm wondering: what's the real world app behavior when such productID doesn't exist, and wouldn't we want to be able to mock/test that "product not found in siteID" case in our tests? (iow return an optional here), like you do for `order(…)` and `review(…)` below? 🤔

fastlane/Fastfile
```diff
@@ -537,6 +540,7 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
       "app_store_screenshot-3"    => {desc: "app_store_screenshot_3.txt"},
       "app_store_screenshot-4"    => {desc: "app_store_screenshot_4.txt"},
       "app_store_screenshot-5"    => {desc: "app_store_screenshot_5.txt"},
+      "app_store_screenshot-6"    => {desc: "app_store_screenshot_6.txt"},
```
AliSoftware: If we do have a new `app_store_screenshot_6.txt`, shouldn't `fastlane/screenshots.json` also have changed in that PR accordingly? I don't see the related change there in this branch though?

AliSoftware (commented):
I only looked at the code itself in GitHub so far – didn't yet test the feature itself on my Mac – but figured I'd still post my initial review notes before going deeper and coming back to it to test it for real later.

Mostly nits, though a couple of questions too.

shiki:
I can take a look, possibly later today. 🙂

.github/workflows/screenshots.yml
jkmassel: Haha I actually just learned that `xcversion` exists yesterday and didn't realize that it could be used here.

Changed in cfd0a24258937238af808732c8244279771aea31

.github/workflows/screenshots.yml
jkmassel: Yeah, good question – I was in a rush and this worked fine. IIRC we've done some work on the release toolkit to support v7, but pinning the dependency in this way prevents it from breaking in a future v8 version some day too.

I agree this might be good to come back to, but yeah probably in a future PR!

.github/workflows/screenshots.yml
jkmassel: > is this a leftover of one of your debug session

Nice catch, thank you!!

Fixed in d6bbbefca7affdc570060b76c77cfcf4ab503420

.github/workflows/screenshots.yml
AliSoftware: Yeah I actually [updated release-toolkit to imagemagick 7 during my trial when I worked on Android Screenshots](https://github.com/wordpress-mobile/release-toolkit/pull/154) back in the day. Which means that the release toolkit itself does depend on imagemagick 7, so it might not be great to depend on 2 different versions for different projects sharing the toolkit…

I think we ultimately need to unify and migrate this to v7 to be aligned with release-toolkit, at least. But that could be done later if that's too much work to include in that PR and if we want to merge this sooner than later.

Yosemite/Yosemite/Model/Mocks/MockObjectGraph.swift
jkmassel: Yeah, good question – the product should be guaranteed to exist on the object graph here – we're just finding it in the list. Because `first` returns an optional, we force-unwrap it. If this ever fails there are bigger issues at play (ie – maybe we've loaded a production database in a test context?)

.github/workflows/screenshots.yml
```diff
@@ -137,42 +105,65 @@ jobs:
...
+        run: |
+          aws s3 cp s3://$S3_BUCKET/fonts.zip fonts.zip
```
shiki, line 140: Does this mean that we're downloading and installing the font every time the workflow is executed? Just curious, I really have no idea what's happening here. 🤪

fastlane/Fastfile
jkmassel: Good call – this is no longer necessary :) 

Removed in 5c32829e2945b3654c49e0e5d23471270cf69698

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerceScreenshots.xcscheme
jkmassel: Fortunately no – this was moved into `fastlane`, so we just use `override_status_bar: true` in the `capture_ios_screenshots` and it does it all for us! 🎉

.bundle/config
```diff
@@ -0,0 +1,5 @@
+---
```
shiki, line 1: Somehow, I have this file in my local repo. Does this mean that we used to generate this?

WooCommerce/Classes/AppDelegate.swift
```diff
@@ -318,11 +320,26 @@ private extension AppDelegate {
     }
 
     func handleLaunchArguments() {
-        if ProcessInfo.processInfo.arguments.contains("logout-at-launch") {
+        if BuildConfiguration.shouldLogoutAtLaunch {
           ServiceLocator.stores.deauthenticate()
```
shiki: The indent is a bit off. 😉 Yay for nitpicking!

WooCommerce/Classes/AppDelegate.swift
```diff
@@ -318,11 +320,26 @@ private extension AppDelegate {
...
         }
 
-        if ProcessInfo.processInfo.arguments.contains("disable-animations") {
+        if BuildConfiguration.shouldUseScreenshotsNetworkLayer {
+            /// Print the location of the core data DB for debugging
+            if let storeUrl = ServiceLocator
+                .storageManager
+                .persistentContainer
```
shiki: ⚠️ I'd rather we don't use `persistentContainer` outside of `StorageManager`. I've been meaning to make this `private` because we couldn't mock `StorageManagerType` if this is still exposed. We made some progress towards that in https://github.com/woocommerce/woocommerce-ios/pull/2049 but we still have one more thing to do before we can make this private: https://github.com/woocommerce/woocommerce-ios/issues/1765.

Maybe we can expose the `storeURL` as a property of `StorageManagerType`?

WooCommerce/Classes/AppDelegate.swift
```diff
@@ -318,11 +320,26 @@ private extension AppDelegate {
...
+                .persistentStoreCoordinator
+                .persistentStores
+                .first?
+                .url {
+                debugPrint("💿 \(storeUrl.path)")
+            }
+
+            ServiceLocator.setStores(ScreenshotStoresManager(storageManager: ServiceLocator.storageManager))
```
shiki, line 328: 🎉 🎉 🎉

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerceScreenshots.xcscheme
jkmassel: Apparently - Xcode will change it back to `1.7` if you re-add a script phase, but only if it takes build arguments from a target.

Weird 🤷

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
jkmassel: Ah interesting ok!

Sounds good to me!

WooCommerce/WooCommerceScreenshots/Utils/SnapshotHelper.swift
shiki: Thanks!

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
```diff
@@ -54,6 +60,67 @@ class WooCommerceScreenshots: XCTestCase {
...
+
+            /// Run event loop
+            self.loop.runForever()
+
+            debugPrint("Web Server running on port \(self.port)")
+        }
+    }
+
+    func handleWebRequest(env: [String: Any], startResponse: ((String, [(String, String)]) -> ()), sendBody: ((Data) -> ())) {
+        /// Extract the path, so http://localhost:9285/foo-bar would be `/foo-bar` here.
```
shiki, line 87: It'd be great if there's a summary of what we're doing here and why we need to do it. I admittedly don't understand what's going on. 😅

Yosemite/Yosemite/Internal/String+Extensions.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+extension String {
+
+    private static let slugSafeCharacters = CharacterSet(charactersIn: "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-")
+
+    public var slugified: String? {
```
shiki: I believe this can be module-level only. 

```suggestion
    var slugified: String? {
```

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockOrderActionHandler.swift
jkmassel: Yeah, unfortunately that's code outside that scope of this PR. 

IMHO we should never emit changes on the main queue from a library function – the app layer should handle when it wants to bring the changes into the main queue to avoid random UI hitches. 

But it's [the pattern used elsewhere in this app](https://github.com/woocommerce/woocommerce-ios/blob/a1d8473bbaff62a6baa03e4f3a848dd0df4efb57/Yosemite/Yosemite/Stores/OrderStore.swift#L344-L346), which is why I copied it for this 🤷‍♂️

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockActionHandler.swift
```diff
@@ -0,0 +1,56 @@
...
+    func success(_ onCompletion: (Error?) -> ()) {
+        onCompletion(nil)
+    }
+
+    /// A helper for saving mock objects into Core Data
+    func save<T, U>(mocks: [T], as dataType: U.Type, onCompletion: @escaping (Error?) -> ()) where U: ReadOnlyConvertible & NSManagedObject {
```
shiki, line 39: Noice!

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockAppSettingsActionHandler.swift
jkmassel: I've seen this pattern before, but we don't allow new warnings in PRs, so I can't use it 😅

If we were looking to fully build out the backend we'd persist it to a `plist` but it's not _required_. However, leaving the message and empty function body there hints that it's a future improvement we could make without requiring it or making a lot of noise 🤷‍♂️

WooCommerce/Classes/System/BuildConfiguration.swift
jkmassel: I would like this as well. One thing I wasn't clear on is whether that works inside of other packages (or bundles?), or just from the main one?

For now isolating this over here at least makes the call sites clean and we can improve it at some point

.github/workflows/screenshots.yml
jkmassel: Ah ok awesome – in that case I've bumped it to v7 in 3a58fd0161419487ae9f969fca94d7f0fb6bd11f – let's see how it works!

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockAppSettingsActionHandler.swift
AliSoftware: Would then maybe `// TODO: xxx` (with the specific syntax that Xcode recognizes, i.e. exact `// TODO:` prefix) be better? Given that this:
1. is recognized by Xcode IDE
2. can be turned into a warning on demand [by a SwiftLint rule](https://realm.github.io/SwiftLint/todo.html) (there's even [another one that allows to put an expiration date on a TODO](https://realm.github.io/SwiftLint/expiring_todo.html) so the warning is only triggered after said date…)
 
That way we can keep this not being a warning by default… but be able to surface them as warning only on demand when handling maintenance 🤔

shiki: I think we were also supposed to remove this `setStatsVersionLastShown` action. We used this to allow the user to change between v3 or v4 of the stats feature but we removed v3 a while ago. Only v4 is available. Maybe we just forgot to remove this. 🤷

shiki: In any case, we only save these settings on files so if we need to save something, perhaps an `NSDictionary` is good enough since the screenshots flow don't really have a permanent state, right? It's just run and die?

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockNotificationCountActionHandler.swift
```diff
@@ -0,0 +1,28 @@
...
+        }
+    }
+
+    func loadNotificationCount(siteId: Int64, type: SiteNotificationCountType, onCompletion: (Int) -> Void) {
+        onCompletion(objectGraph.currentNotificationCount)
+    }
+
+    func reset(siteId: Int64, type: Note.Kind, onCompletion: () -> Void) {
+        // Don't do anything when resetting
+    }
```
shiki, line 27: I guess this is a disadvantage of `unimplementedAction(action: action)`. We're forced to add `case` for everything instead of just excluding them and letting `default` do nothing. 🙂

WooCommerce/Classes/System/BuildConfiguration.swift
AliSoftware: Yeah I like that you isolated them here 👌  will make it easier if we want to improve this later 👍 
Good question about cross-bundle 🤔  I don't have the answer but definitively something to double-check when we'll get to this!

Yosemite/Yosemite/Model/Mocks/Graphs/ScreenshotsObjectGraph.swift
```diff
@@ -0,0 +1,366 @@
...
+            country: country,
+            phone: phone,
+            email: email
+        )
+    }
+
+    static var shippingAddress: Address {
+        return billingAddress
+    }
+}
```
shiki: ℹ️ This does not seem to be used. It's either that or Xcode is lying to me. 😄 

```suggestion

```

Yosemite/Yosemite/Model/Mocks/Graphs/ScreenshotsObjectGraph.swift
```diff
@@ -0,0 +1,366 @@
...
+
+    struct DefaultSite {
+        static let name = NSLocalizedString("Your WooCommerce Store", comment: "Store Name for the screenshot demo account")
+        static let url = NSLocalizedString("example.com", comment: "")
+    }
+}
+
+struct ScreenshotObjectGraph: MockObjectGraph {
+
+     public let userCredentials = Credentials(
```
shiki: ℹ️ This and all the other `public` properties and methods below do not need to be `public`. 🙂 

```suggestion
    let userCredentials = Credentials(
```

Yosemite/Yosemite/Model/Mocks/Graphs/ScreenshotsObjectGraph.swift
```diff
@@ -0,0 +1,366 @@
...
+        createTopEarningItem(product: Products.coloradoShades, quantity: 5),
+    ])
+
+    /// The probability of a sale for each visit when generating random stats
+    ///
+    let orderProbabilityRange = 0.1...0.5
+
+    /// The possible value of an order when generating random stats
+    ///
+    let orderValueRange = 100 ..< 500
```
shiki: ℹ️  These can be `private`. 🙂 

```suggestion
    /// The probability of a sale for each visit when generating random stats
    ///
    private let orderProbabilityRange = 0.1...0.5

    /// The possible value of an order when generating random stats
    ///
    private let orderValueRange = 100 ..< 500
```

Yosemite/Yosemite/Model/Mocks/MockObjectGraph.swift
```diff
@@ -0,0 +1,481 @@
...
+    let end: Date
+
+    static func from(start: String, end: String) -> DateRange {
+
+        return DateRange(
+            start: Date.from(dateString: start),
+            end: Date.from(dateString: end)
+        )
+    }
+}
```
shiki: ℹ️ This seems unused. 🙂 

```suggestion

```

Yosemite/Yosemite/Model/Mocks/MockObjectGraph.swift
```diff
@@ -0,0 +1,481 @@
...
+        formatter.dateFormat = "yyyy-MM-dd HH:MM:ss"
+        return formatter.string(from: self)
+    }
+
+    static func from(dateString: String) -> Date {
+        let formatter = DateFormatter()
+        formatter.dateFormat = "yyyy-MM-dd"
+        return formatter.date(from: dateString)!
+    }
+}
```
shiki: Do you think you'll be able to use some of the formatters in [`DateFormatter+Woo`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Networking/Networking/Extensions/DateFormatter+Woo.swift)? I'm not sure. 🤔 I'm just sharing in case you didn't know about that.

shiki (changes requested):
Hi, @jkmassel! I like how this turned out! 🎉 

I'm requesting changes because of one little tiny thing: the usage of `persistentContainer`. Please see the inline comment marked with ⚠️ . The others are just suggestions or nitpicks. 😄 

> I think we'd talked about punting this all into a separate Swift package, which I'm ok to do if you'd like to, but I wasn't sure if it'd still apply given the move to Yosemite?

Yeah, I'm not sure either. The only concern here is that there are lots of classes/structs in Yosemite now so maybe it makes sense to have a separate Swift package. That should make it easier for product developers to grok Yosemite. So maybe it's worth a try? 

I'm not sure how much effort is involved in that. We might have some surprising things that need to be done. 😬

.bundle/config
AliSoftware: It's not surprising: it's just because it used to be `.gitignore`'d (but now we want to commit it to have the same config for everyone).

I think the right move is to use the one from the PR replace the one you had locally. (This might make people have `git pull` error about the conflicting files once the PR is merged tho, might need to warn them in Slack)

.bundle/config
```diff
@@ -0,0 +1,5 @@
...
+BUNDLE_RETRY: "3"
+BUNDLE_PATH: "vendor/bundle"
+BUNDLE_JOBS: "3"
+BUNDLE_WITHOUT: "screenshots"
```
AliSoftware, line 5: This made me realize a side effect of this: when we want to generate screenshots, we will run `bundle install --with screenshots` – or, as it's nowadays suggested by bundler instead, `bundle config --local --delete without` (see [this in WPAndroid for example](https://github.com/wordpress-mobile/WordPress-Android/blob/develop/.circleci/config.yml#L460-L462)).

In both cases this will **have the effect of modifying this `bundle/config` file** to remove this `BUNDLE_WITHOUT`, **a change which should NOT be commited** (as we don't want everyone to install gems that are useless for most people if they never have to create screenshots in most cases).

This is probably unlikely to happen as most devs will never have to do that if screenshots are generated by an MIE or on CI, but something to keep in mind still…

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
JavonDavis: +1

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockProductActionHandler.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    func upsert(products: [Product], onCompletion: @escaping () -> ()) {
+        let storage = storageManager.newDerivedStorage()
+
+        storage.perform {
+            productStore.upsertStoredProducts(readOnlyProducts: products, in: storage)
+        }
+
+        storageManager.saveDerivedType(derivedStorage: storage) {
```
JavonDavis, line 68: @jkmassel Just a quick question, it's probably not super important here either but could we end up a race condition with the perform block here? I'm wondering if _technically_ we could end up saving before the perform block completes

Yosemite/Yosemite/Model/Mocks/Graphs/ScreenshotsObjectGraph.swift
jkmassel: Ah yep – I ended up just using a `struct` for this but forgot to remove it.

Thanks! Addressed in 23a0b02ebc2f8d5b7b017f8960db6f3681f9d35a

(I'm too spoiled by Rust, which complains about this kind of thing!!)

Yosemite/Yosemite/Model/Mocks/Graphs/ScreenshotsObjectGraph.swift
jkmassel: Yeah – I had these in place in case we were moving to to a package, but I'll go back and audit visibility when we do that in a future PR :) 

Addressed in c99bc531f92a9b1b4fdbb3ba6e5b9eb5efc108b0

Yosemite/Yosemite/Model/Mocks/Graphs/ScreenshotsObjectGraph.swift
jkmassel: Sure can – also addressed in c99bc531f92a9b1b4fdbb3ba6e5b9eb5efc108b0

Yosemite/Yosemite/Model/Mocks/MockObjectGraph.swift
jkmassel: Removed in 2a5fa077afbed46e6d31b1ec318fb523d1b27c03

Yosemite/Yosemite/Model/Mocks/MockObjectGraph.swift
jkmassel: Nice – I wasn't aware those existed. For creating these mock objects I still prefer the fluent approach, but calling out to those formatters avoids reinventing this particular wheel and will also beneficially cause the code to break if we move to a different versioning scheme (say in a future statsv5?)

Addressed in 2a5fa077afbed46e6d31b1ec318fb523d1b27c03

.bundle/config
jkmassel: Yeah, this is true. It's kinda suboptimal no matter which way you go – we don't want people having to remember to run `bundle install --without screenshots` (or whatever the new syntax is 😅) 

But we also don't want to dirty the file every time. 

IMHO this is the _best_ option because as you note it'll only affect MIEs and not everybody else 😄

Once we drop ruby for all the things and only MIEs are using it, IMHO we could just remove the `without` part and make it a requirement if you want to use Ruby?

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockProductActionHandler.swift
jkmassel: Ya this is an excellent point – I can't believe I missed this 🙈 

It's the same code as is used here: 

https://github.com/woocommerce/woocommerce-ios/blob/bdbabd20ca3ee0a581c7c1b1e5c8bafd7a319eeb/Yosemite/Yosemite/Stores/ProductStore.swift#L345-L357


@shiki – WDYT? IMHO this makes sense to fix in both places right?

shiki: I have been confused by this in the past. This is fine as far as I can tell. The `saveDerivedType` has a `perform` call too. 🙂 So they should happen in sequence. 

https://github.com/woocommerce/woocommerce-ios/blob/bdbabd20ca3ee0a581c7c1b1e5c8bafd7a319eeb/Storage/Storage/CoreData/CoreDataManager.swift#L143-L152

AliSoftware:
Moving this to milestone 5.9 as I'm code-freezing 5.8 today

WooCommerce/WooCommerceUITests/Screens/BaseScreen.swift
jkmassel: That's all @rachelmcr! 😃 

I just copied it from WPiOS 😝

shiki:
@jkmassel Let me know when this is ready for another round!

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockAppSettingsActionHandler.swift
jkmassel: Addressed in 42f65e17c9df539736946c65e3c0a4f2155fe56f since yeah, there's no state. Just replaced with a no-op until it's removed!

.github/workflows/screenshots.yml
jkmassel: Yep that's exactly right, because we can't guarantee the Woo marketing fonts are on our CI machines, we need to grab them every time 🙃

WooCommerce/Classes/AppDelegate.swift
jkmassel: I love nitpicks 😉

Addressed in f49e08d46f610cc4da670bd8a7ea6404194dd716

WooCommerce/Classes/AppDelegate.swift
jkmassel: Works for me – definitely understand about isolating this!

I personally like printing the path the app's documents directory or something on startup, because when debugging I like to open up the core data DB a bunch to see what it's doing. 

But for now, I'm fine to just drop this code to avoid further changes – I'll replace it some other future tooling thing probably 😜

Removed in 046c07d7a91a0dcb5acdd438a8f2909a31421c31

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockNotificationCountActionHandler.swift
jkmassel: Ya there's definitely a trade-off there – I'd prefer the screenshots code crash rather than silently ignore signals that the app emits and create screenshots with missing data 😅

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
jkmassel: Yeah good call – went for a lazier solution in 2f483bc9cfb32938f8bfd2642539534193ef9f8f – passing `0` to the port number lets the kernel assign an open one 😜

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
jkmassel: Yeah good call – addressed in 896a0404b8efdcf96458c9fb7bd7b8f42dee8f09 – I didn't use the suggestion so I could add a formatted comment which is the woo default

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
jkmassel: Added a bit more info in 896a0404b8efdcf96458c9fb7bd7b8f42dee8f09 – hope it's helpful – happy to add more if needed!!!

jkmassel:
@shiki, @AliSoftware, and @JavonDavis – thanks for the excellent feedback. I'm _pretty_ sure I've addressed all of it in this PR (apologies if I missed one, there was a lot!! 😁)

Ready for another look at your convenience! 🙇

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
AliSoftware: TIL. 👍

WooCommerce/Classes/AppDelegate.swift
shiki: Removal was a better decision than anything I could have imagined. 🤣

Yosemite/Yosemite/Model/Mocks/Graphs/ScreenshotsObjectGraph.swift
shiki: Xcode should complain about these things too. Just saying. 🤷

shiki (approved):
Thanks, @jkmassel! 

The tests didn't run for me on iPhone 11. It just stayed at “Running tests...”. It works on iPhone 12 though. Not sure if this is the expected behavior so I thought I'd share.

WooCommerce/WooCommerceScreenshots/ScreenshotImages.xcassets/akoya-pearl-shades.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
+    {
+      "filename" : "Mask-Group-2.png",
```
AliSoftware: [super-nitpick] The OCD in me is kicking in when I see the name of the PNG file that is completely different / not based on the basename of the `.imageset` 😅  (ditto for other imagesets below)

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockStatsActionV4Handler.swift
```diff
@@ -0,0 +1,61 @@
...
+    func retrieveStats(siteID: Int64, timeRange: StatsTimeRangeV4, onCompletion: @escaping (Error?) -> ()) {
+        let store = StatsStoreV4(dispatcher: Dispatcher(), storageManager: storageManager, network: NullNetwork())
+
+        switch timeRange {
+            case .today: success(onCompletion)
+            case .thisWeek: success(onCompletion)
+            case .thisMonth: success(onCompletion)
+            case .thisYear:
+                store.upsertStoredOrderStats(readOnlyStats: objectGraph.thisYearOrderStats, timeRange: timeRange)
+                onCompletion(nil)
```
AliSoftware: My OCD speaking again: feels a bit unbalanced and harder to read to have some `case` having their action on the same line but others (just the last one here) being on separate lines (because there's more than one expression).

At first skim my mind read it as if that was a C-style switch (i.e. with implicit `fallthrough`) and thought that all 4 cases were executing the last 2 lines (`store.upsert…` + `onCompletion`), before I clicked back into realizing there were some `success(onCompleteion)` calls hiding there.

It might only be a nit but even if it would make the code spread a bit more vertically, I think it would be worth always putting the instructions after each case on their own lines. In this specific example, given the instructions are the same for the 3 first cases, we could DRY the 3 cases on one line though:

```suggestion
            case .today, .thisWeek, .thisMonth:
                success(onCompletion)
            case .thisYear:
                store.upsertStoredOrderStats(readOnlyStats: objectGraph.thisYearOrderStats, timeRange: timeRange)
                onCompletion(nil)
```

Ditto for all other similar `switch` in the same file below ofc.

fastlane/Fastfile
```diff
@@ -699,7 +708,7 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
   end
 
   def ios_locales
-    "ar de-DE es-ES fr-FR he id it ja ko nl-NL pt-BR ru sv tr zh-Hans zh-Hant en-US".split()
+    "ar de-DE en-US es-ES fr-FR he id it ja ko nl-NL pt-BR ru sv tr zh-Hans zh-Hant".split()
```
AliSoftware, line 707: Ah, one thing my OCD likes very much! 😄

fastlane/Fastfile
```diff
@@ -471,7 +477,10 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
       stop_after_first_error: true,
 
       # Allow the caller to invoke dark mode
-      dark_mode: options[:mode].to_s.downcase == "dark"
+      dark_mode: options[:mode].to_s.downcase == "dark",
+
+      # Make the status bar pretty
```
AliSoftware, line 478: 💄 😊 ✨

fastlane/Fastfile
```diff
@@ -415,6 +418,9 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
   desc "Take Screenshots"
   lane :take_screenshots do | options |
 
+    # Ensure we're using the right version of Xcode
+    xcversion(version: "~> 12.0")
```
AliSoftware, line 418: ❤️

AliSoftware (approved):
LGTM!

Only left a couple of nits that could be nice to address before merge, but feel free to ignore them.

jkmassel:
> The tests didn't run for me on iPhone 11

@shiki – thanks for pointing this out. Did you happen to run it again after shutting off the simulator? I've tested on iPhone 11 across a few iOS versions and they all seem to work. My guess would be that some kind of Xcode => Simulator XPC thing was having a bad day and that's why it didn't work? We don't use iPhone 11 for testing, so if we run into this in the future, I think it'd be ok to address it then?

Yosemite/Yosemite/Model/Mocks/ActionHandlers/MockStatsActionV4Handler.swift
jkmassel: Yeah fair enough – fixed all of the cases in this file in 11e4afa0a6e8ae757a51e5a501f8f4b2b50a8886 and verified that there aren't others in the other files.

WooCommerce/WooCommerceScreenshots/ScreenshotImages.xcassets/akoya-pearl-shades.imageset/Contents.json
jkmassel: Hah nice! Fixed in c52879a4b911a43c2b645309b9f5f893c38ad6d2 😂

jkmassel:
Force-pushed to fix a failed merge with `develop` – merging once it turns 🟢 

Thanks for the reviews y'all!! 🙌

shiki:
> Did you happen to run it again after shutting off the simulator?

Yeah, I tried everything that I know including clearing derived data and stuff. I think it just hates me. 

But yeah, I wouldn't worry about it.

Issue #3340, opened 2020-12-14
Shipping Labels: handle A4 paper size


jaclync:
h/t to @Ecarrion for discovering the A4 paper size from testing!

For now, only stores in the US can create a shipping label with 3 paper size options: legal, letter, and label. However, after a shipping label is created, if the store country is changed to outside of US, CA, MX, DO (CKNC7KACA/p1607010932150400?thread_ts=1606980019.149700&cid=CKNC7KACA-slack) A4 paper size becomes available.

This is still an edge case as most stores don't change the store country and reprint a shipping label after creation while switching store country. In the future, we might want to support A4 paper size if it is the shipping label preference (that we fetch in `ShippingLabelSettings`).

Let's add an enum case for A4 paper size, and fall back to one of the supported paper sizes in the reprint screen for now.

Steps to repro:

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/) and there is an order with at least one non-refunded shipping label and an order with at least one refunded shipping label.

You can set up your store to use a test credit card with instructions in p91TBi-3xD-p2.

- In wp-admin, create a shipping label for an order
- Change store country to outside of US, CA, MX, DO (e.g. UK) in WooCommerce settings
- Go back to the order, and try reprinting a shipping label --> A4 paper size appears in the paper size menu
- Reprint with A4 paper size
- Launch the app
- Go to the orders tab, and tap on the same order --> the app currently crashes in debug mode because A4 paper size is unexpected

PR #3341, opened 2020-12-14
Reprint Shipping Label: load shipping label settings for default selected paper size in view model

jaclync:
Part of #2169 

## Why

Because there are quite some UI changes for the reprint view controller, I separated the code for its view model and Yosemite layer changes for loading a `ShippingLabelSettings` for default selected paper size to this PR.

## Changes

Yosemite layer:
- Added `ShippingLabelAction.loadShippingLabelSettings` to load the corresponding `ShippingLabelSettings` for a `ShippingLabel` + test cases

App layer:
- Created `ReprintShippingLabelViewModel` that provides 2 data for future usage:
  - `paperSizeOptions: [ShippingLabelPaperSize]`: all supported paper sizes for paper size selector
  - `selectedPaperSize: AnyPublisher<ShippingLabelPaperSize?, Never>`: currently selected paper size, the default is set to the corresponding `ShippingLabelSettings`'s `paperSize` (if the settings paper size is not supported, it defaults to the first supported paper size - `legal`. it's possible in an edge case after changing store country https://github.com/woocommerce/woocommerce-ios/issues/3340)
- Unit tests for `ReprintShippingLabelViewModel`

## Testing

The view model and new action aren't used yet, just CI for now!

Update release notes:


WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewModel.swift
```diff
@@ -0,0 +1,43 @@
...
+    /// Paper size options that we support for reprinting a shipping label.
+    /// In the future, the options could be different per geographical region.
+    let paperSizeOptions: [ShippingLabelPaperSize] = [.legal, .letter, .label]
+
+    /// Observable selected paper size.
+    var selectedPaperSize: AnyPublisher<ShippingLabelPaperSize?, Never> {
+        selectedPaperSizeSubject.eraseToAnyPublisher()
+    }
+
+    private let selectedPaperSizeSubject = CurrentValueSubject<ShippingLabelPaperSize?, Never>(nil)
```
ctarda: I am curious about your approach here. 

My understanding of CurrentValueSubject is that it wraps a single value and published a new element whenever the value changes. 

It seems like the value is expected to change only once? I was wondering if instead of a publisher (selectedPaperSize) publishing a value emitted by a Subject (CurrentValueSubject), it wouldn't be simpler if the view model extended ObservableObject, exposing selectedPaperSize as a read-only property. and the view observed the view model to redraw when the observed object is updated. 

What do you think?

ctarda (approved):
It looks good to me. 

I might lean towards a slightly different approach to make the view model emit changes, but I am curious about your opinion. 

In the meantime, :shipit:

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewModel.swift
jaclync: The paper size could be changed later by the user in the [paper size selector screen](https://user-images.githubusercontent.com/1945542/101569513-9560c680-3a0f-11eb-8622-46a3a0b1d22d.png) as a separate subtask of https://github.com/woocommerce/woocommerce-ios/issues/2169! In the subtask, there will be a new function added to the view model like `func onPaperSizeSelected(_ paperSize: ShippingLabelPaperSize)`.

Thanks for suggesting `@Published` + `ObservableObject`, I didn't think of using that! It seems like `@Published` is actually implemented by `CurrentValueSubject` underneath but easier to read (just like a property). Since there's no error handling in this use case, I'll update it to `@Published` + `ObservableObject`!

jaclync: Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3341/commits/9a19560b433100de6cfb236856cef2f2f23134a3 - it looks much simpler 😃

jaclync:
Thank you for the suggestion @ctarda, I updated the PR for another check (it's my first time using `@Published` in production)! 😄

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewModel.swift
shiki: This is pretty cool! It turns out that you also **don't need** to conform to `ObservableObject`. The `@Published` itself is just fine in my test:

```swift
final class ReprintShippingLabelViewModel {
    /// Observable selected paper size.
    @Published private(set) var selectedPaperSize: ShippingLabelPaperSize?
```

Conforming to `ObservableObject` exposes a `objectWillChange` observable which can be misleading. 



Unless there was something else we were using `ObservableObject` for. 😬

jaclync: @shiki good point! From the [Apple documentation on `@Published`](https://developer.apple.com/documentation/combine/published) and [`ObservableObject`](https://developer.apple.com/documentation/combine/observableobject), it looks like we don't need `ObservableObject` since we are not listening for `objectWillChange`. Updated here: https://github.com/woocommerce/woocommerce-ios/pull/3341/commits/5c5c96a31c1c99e4c8cd43f74b8ecfc21a1571a0

jaclync:
Thanks for the review @ctarda @shiki, merging now!

Issue #3334, opened 2020-12-11
Shipping Labels M1: design & accessibility enhancements


jaclync:
### Design Polishes

- [x] Hide separator and remove bottom spacing in "Reprint Shipping Label" CTA row https://github.com/woocommerce/woocommerce-ios/pull/3430
- [x] Hide separator in "Don't know how to print from your phone" row and potentially add 8px spacing below ([reference](https://github.com/woocommerce/woocommerce-ios/pull/3430#issuecomment-755831580)) https://github.com/woocommerce/woocommerce-ios/pull/3435
- [x] Reprint screen (from PR https://github.com/woocommerce/woocommerce-ios/pull/3342):
  - [x] Images in "See layout and paper sizes options" and "Don't know how to print from your device?" rows are not the footnote size 20x20 as in design (same refactoring requirement for `TopLeftImageTableViewCell` as above) https://github.com/woocommerce/woocommerce-ios/pull/3396
  - [x] Info text row's image and text are not top aligned in larger font size ([screenshot](https://user-images.githubusercontent.com/198826/102104000-e9422580-3dea-11eb-9563-a0f89abbf596.png)) (same refactoring requirement for `TopLeftImageTableViewCell` as above) https://github.com/woocommerce/woocommerce-ios/pull/3438
  - [x] Paper size selector row (`SettingTitleAndValueTableViewCell`)'s value is truncated in large font size ([screenshot](https://user-images.githubusercontent.com/198826/102104138-0f67c580-3deb-11eb-9255-ffafae80d9e3.png)) https://github.com/woocommerce/woocommerce-ios/pull/3411
- [x] Update "House" icon to outlined house icon (either by adding a new one to Gridicons or exporting to a PDF) https://github.com/woocommerce/woocommerce-ios/pull/3398

### Accessibility Enhancements

- [x] ([Reference](https://github.com/woocommerce/woocommerce-ios/pull/3306#pullrequestreview-549270386)) With large font size, the header text in "Refund Shipping Label" screen dominates the top space and leaves very little room for the scrollable content. It might be worth converting the header view from section header view to table's `tableHeaderView` or a cell so that it scrolls with the content below https://github.com/woocommerce/woocommerce-ios/pull/3440
- [x] ([Reference](https://github.com/woocommerce/woocommerce-ios/pull/3322#pullrequestreview-549340020)) Paper size options screen: polish grid layout in large font sizes https://github.com/woocommerce/woocommerce-ios/pull/3462

[...]

PR #3342, opened 2020-12-14
Order Details > Shipping Label: reprint shipping label main screen

jaclync:
Part of #2169 

For code reviewers: please review https://github.com/woocommerce/woocommerce-ios/pull/3341 first. Feel free to hold off reviewing this PR if the base PR has changes requested.

## Why

This PR implements the main screen for reprinting a shipping label. The view model is implemented in the base PR https://github.com/woocommerce/woocommerce-ios/pull/3341 since the UI changes are large enough.

## Changes

- Added a new icon `pagesFootnoteImage` for the footnote size for `pages` Gridicon
- Added an extension `UIButton.applyFootnoteLinkButtonStyle` tot apply footnote link button styles (footnote has smaller font size)
- Added an extension `UITableViewCell.hideSeparator` to hide the separator for a cell - feel free to suggest other solutions!
- Created `ReprintShippingLabelViewController` with a table view and a CTA to reprint a shipping label. I decided to implement the UI above the CTA with table view cells, after struggling with dynamic height using `tableView.tableHeaderView`.)
- Created `SpacerTableViewCell` for a cell just for vertical padding
- Integration with order details in `OrderDetailsViewController` and `OrderDetailsDataSource`

## Testing

- Go to the orders tab
- After syncing, tap on "Reprint Shipping Label" CTA --> it should navigate to the reprint UI as in screenshots

## Example screenshots

Notes for reviewers: due to the current implementation of `TopLeftImageTableViewCell` still using frames and bounds, the images in "See layout and paper sizes options" and "Don't know how to print from your device?" rows are not the footnote size 20x20 as in design. I added this to the design enhancements issue https://github.com/woocommerce/woocommerce-ios/issues/3334 to work on them separately.

light | dark
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-14 at 13 58 18 | Simulator Screen Shot - iPhone 11 - 2020-12-14 at 14 24 57

Update release notes:


Garance91540 (approved):
Hi @jaclync it looks good but slightly too dense, could we update the padding under the info text "if you already used..." from 16px to 24px? Thank you!

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    deinit {
+        cancellables.forEach {
+            $0.cancel()
+        }
```
shiki, line 33: I still have to validate this myself but we probably won't need to manually call `cancel()` on `deinit`. From [the docs](https://developer.apple.com/documentation/combine/anycancellable):

> An AnyCancellable instance automatically calls cancel() when deinitialized.

This is as long as we're not capturing `self` in the `sink()` calls. 

---

Nothing to change here though. I just thought of mentioning it for the sake of sharing ideas. 🙂

WooCommerce/Classes/Extensions/UIButton+Helpers.swift
```diff
@@ -80,6 +80,17 @@ extension UIButton {
...
+    ///
+    func applyFootnoteLinkButtonStyle() {
+        backgroundColor = .clear
+        tintColor = .systemColor(.secondaryLabel)
+        titleLabel?.applyFootnoteStyle()
+        titleLabel?.textAlignment = .natural
+        setTitleColor(.systemColor(.secondaryLabel), for: .normal)
+        setTitleColor(.systemColor(.tertiaryLabel), for: .highlighted)
+    }
+
```
shiki: ⚠️  This does not seem to be used anywhere. 🙂 

```suggestion

```

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
```diff
@@ -0,0 +1,257 @@
...
+            self.reprintButton.isEnabled = paperSize != nil
+        }.store(in: &cancellables)
+    }
+}
+
+// MARK: UITableViewDataSource
+extension ReprintShippingLabelViewController: UITableViewDataSource {
+    func numberOfSections(in tableView: UITableView) -> Int {
+        1
+    }
```
shiki: ⚠️  I believe the default is 1. 

```suggestion

```

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
```diff
@@ -0,0 +1,257 @@
...
+
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        rows.count
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let row = rows[indexPath.row]
```
shiki: If there's one thing I hate, it's “invalid number of rows” exceptions. 😄 I think we should still consider using `safe` whenever possible. 

```suggestion
        guard let row = rows[safe: indexPath.row] else {
            return UITableViewCell()
        }
```

But up to you. It's unlikely. We can always fix it if we ever get a crash later on the road.

WooCommerce/Classes/ViewRelated/ReusableViews/SpacerTableViewCell.swift
```diff
@@ -0,0 +1,24 @@
...
+
+/// Used for variable vertical spacing between two table view cells.
+/// The cell currently has default cell background color (`.listForeground`).
+final class SpacerTableViewCell: UITableViewCell {
+    /// The height of the spacer cell.
+    var height: CGFloat = 0 {
+        didSet {
+            heightConstraint.constant = height
+        }
+    }
```
shiki: Do you think we should consider `configure(height:)` instead so we're consistent (kind of) with the other cells?

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
```diff
@@ -0,0 +1,257 @@
...
+        configureCommonStylesForInfoLinkCell(cell)
+    }
+
+    func configureCommonStylesForInfoLinkCell(_ cell: TopLeftImageTableViewCell) {
+        cell.imageView?.tintColor = .systemColor(.secondaryLabel)
+        cell.textLabel?.applyFootnoteStyle()
+        cell.textLabel?.textColor = .systemColor(.secondaryLabel)
+        cell.hideSeparator()
+        cell.selectionStyle = .default
+    }
```
shiki: ℹ️  I think we should consider this as a legitimate configuration style for `TopLeftImageTableViewCell`. Something like:

```swift
cell.configure(style: .secondaryOrWhatever)
```

This is similar to what we did with [`HeadlineLabelTableViewCell`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/ReusableViews/HeadlineLabelTableViewCell.swift):

https://github.com/woocommerce/woocommerce-ios/blob/b8e1d3b3b4b06b8a13b988ef98c64c0a37481f65/WooCommerce/Classes/ViewRelated/ReusableViews/HeadlineLabelTableViewCell.swift#L53-L66

The theory is that:

- We have standard “styles” to use all throughout the app.
- ViewControllers would only be using styles and not defining their own standards.
- Probably makes migration to SwiftUI easier. Probably. 

Something to think about in the future. 🙂

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
```diff
@@ -0,0 +1,257 @@
...
+}
+
+private extension ReprintShippingLabelViewController {
+    enum Constants {
+        static let cellValueTextColor = UIColor.systemColor(.secondaryLabel)
```
shiki: ⚠️ This seems to be unused.

```suggestion

```

WooCommerce/Classes/ViewRelated/ReusableViews/SpacerTableViewCell.xib
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
shiki, line 1: I'm curious if you've tried not having a XIB. Will it work? It seems unnecessary. 😄 This is nitpicky so feel free to ignore.

shiki (approved):
Looking good! I just have a few suggestions for consideration (⚠️ ). 


---

> Added a new icon pagesFootnoteImage for the footnote size for pages Gridicon

GridIcon please don't crash. 🙈 

---

### Accessibility

It looks like there some issues with larger font sizes. 

The icon is misaligned with the text:



And the Paper Size does not scale very well. 



We can probably tackle these as separate issues. 🙂

WooCommerce/Classes/Extensions/UIButton+Helpers.swift
jaclync: Wow nice catch 🦅 👁️ ! (This is from my iteration of using a custom view for `table.tableFooterView` but the dynamic height is tricky after rotation.) Removed in https://github.com/woocommerce/woocommerce-ios/pull/3342/commits/341def7b4e3b1c75abc921c16ac423e0dd4945a6

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
jaclync: Ah right! Removed in https://github.com/woocommerce/woocommerce-ios/pull/3342/commits/341def7b4e3b1c75abc921c16ac423e0dd4945a6

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
jaclync: thanks for catching this! I wish Xcode has an unused code highlighter. Removed in https://github.com/woocommerce/woocommerce-ios/pull/3342/commits/341def7b4e3b1c75abc921c16ac423e0dd4945a6

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
jaclync: Great suggestion! It is nice to have a set of predefined styles. So far, I just see the consistent usage of font size for `TopLeftImageTableViewCell ` while the image/text colors differ in use cases. Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3342/commits/830f6aede4fbbd41467f01c0b8bbe579831d0871

WooCommerce/Classes/ViewRelated/ReusableViews/SpacerTableViewCell.swift
jaclync: yup makes sense! updated in: https://github.com/woocommerce/woocommerce-ios/pull/3342/commits/79499744b164335f3b80cad620152502c4d1a96b

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
jaclync: Hah right, it doesn't hurt to be safer. Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3342/commits/79499744b164335f3b80cad620152502c4d1a96b

WooCommerce/Classes/ViewRelated/ReusableViews/SpacerTableViewCell.xib
jaclync: I thought about not using a xib for 1 subview, though I generally still create a xib whenever possible in Xcode (`UITableViewHeaderFooterView` does not have a xib option in Xcode new file window) following our practices in the codebase. Also, our table view cell registration usually expects the use of a xib (`tableView.registerNib(for:)`) so I thought it's easier not to create exception for table view configuration 😅 .

Xib does create tens of XML diffs even it just has one subview 🙄 looking forward to gradually moving to SwiftUI!

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
jaclync: Interesting! I checked the memory debugger before & after removing these lines, and saw 2 objects with `reprint` keyword in the memory after leaving the reprint screen in both cases:

Screen Shot 2020-12-15 at 11 55 01 AM

These two objects weren't marked with memory warning, and there were other `Swift.ReferenceWritableKeyPath` as well. I didn't find much information online other than the [use of `assign(to:)` in the Swift forum](https://forums.swift.org/t/does-assign-to-produce-memory-leaks/29546). We also weak self in the `sink` closure, not sure if there's any way to get rid of the key paths.

I see that we still cancel `AnyCancellable` manually in other places in the codebase, I'll leave them for now and we can remove them if we feel like in the future.

jaclync:
Thank you for the review & feedback @shiki and @Garance91540! I responded to the comments and updated the PR for another look 😃 

> Hi @jaclync it looks good but slightly too dense, could we update the padding under the info text "if you already used..." from 16px to 24px? Thank you!

Hi @Garance91540 yep! I added another 8px-tall spacer row under the info text in https://github.com/woocommerce/woocommerce-ios/pull/3342/commits/f696b7f6b743a85c4cf35178f867fd45a2c08f11 (cc @shiki) and here's the comparison:

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-15 at 11 26 08 | Simulator Screen Shot - iPhone 11 - 2020-12-15 at 11 27 23

WooCommerce/Classes/Extensions/UIButton+Helpers.swift
shiki: 🤸

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
shiki: OMG the documentation lies. Lies!!! 

Thanks for looking into it!

shiki (approved):
Still LGTM! I reviewed the added `ShippingLabelPaperSizeListSelectorCommand` too. 

I forgot to mention yesterday, thank you for adding lots of comments! 🙏

jaclync:
Sorry @shiki about the 3 commits on `ShippingLabelPaperSizeListSelectorCommand` by mistake! I decided to revert them instead of force pushing the branch because I remember from Platform9 that it is best not to alter the git history. Lemme know if you think force pushing would be cleaner 🙂

shiki (approved):
I have no opinion on force push or revert. 🤷 

:shipit:

WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/ReprintShippingLabelViewController.swift
jaclync: These `Swift.ReferenceWritableKeyPath`s were also alive after manually cancelling the `AnyCancellable` so I'm not sure what to do there 😓 at least there are no memory warnings on these 🙄

jaclync:
Thank you for the review Jayson 💟 Merging this now for other dependent subtasks!

PR #3349, opened 2020-12-15
Order Details: filter deleted refunds and update refund list order

### LINKED ISSUE #3055: Polish Refund List
While reviewing #3048 @jaclync noticed that there are some inconsistencies on how refunds are displayed between the app and web.

Specifically:

- Deleted refunds are shown as an empty refund, should we filter them? Core is not filtering them.
- Refund list order does not match Core.

Screen Shot 2020-10-23 at 9 51 55 AM

eshurakov:
Closes #3055

### Filter out deleted refunds
In a some scenarios when fetching an order or an order list Core would send refunds with `refundID` and `total` being zero. Those refunds are deleted and we don't want to show them to the user.

My first attempt to fix this was to add a code to filter refunds in `OrdersUpsertUseCase` when we update the data in the storage. This approach didn't work in all cases. I discovered that when I pull-to-refresh the order, the updated order still has refunds that I thought should be filtered out. This happens because when we sync an order, the callback passes the model that was created from the network response and not the one that is fetched from the storage.

Because of that, in my second attempt, I moved the code to filter out deleted refunds to `Order`. And it works :-)

### Refund list order

Because there is no `dateCreated` in `OrderRefundCondensed` I thought to keep it simple and order refunds by `refundID` as it seems to be an auto incremental field in Core. The order should be the same as if it is ordered by `dateCreated`.
This way refund list is in the same order as it is in Core.

## Test
1. Create an order
2. Make several refunds in Core
3. Open this order in the app
4. Check that the order of refunds is the same as in Core
5. Delete one of refunds in Core
6. Pull to refresh the order in the app. Refund should be gone.

Update release notes:

### DIFF
--- a/Networking/NetworkingTests/Mapper/OrderMapperTests.swift
+++ b/Networking/NetworkingTests/Mapper/OrderMapperTests.swift
@@ -199,16 +199,14 @@ final class OrderMapperTests: XCTestCase {
 
     /// Verifies that an Order ignores deleted refunds.
     ///
-    func test_Order_deleted_refund_fields_are_ignored() {
-        guard let order = mapLoadOrderWithDeletedRefundsResponse() else {
-            XCTFail()
-            return
-        }
+    func test_Order_deleted_refund_fields_are_ignored() throws {
+        // When
+        let order = try XCTUnwrap(mapLoadOrderWithDeletedRefundsResponse())
 
-        let refunds = order.refunds
-        XCTAssertEqual(refunds.count, 1)
+        // Then
+        XCTAssertEqual(order.refunds.count, 1)
 
-        let refund = refunds[0]
+        let refund = try XCTUnwrap(order.refunds.first)
         XCTAssertEqual(refund.refundID, 73)
         XCTAssertEqual(refund.reason, "Cap!")
         XCTAssertEqual(refund.total, "-16.00")
--- a/Networking/NetworkingTests/Mapper/OrderMapperTests.swift
+++ b/Networking/NetworkingTests/Mapper/OrderMapperTests.swift
@@ -197,6 +197,23 @@ final class OrderMapperTests: XCTestCase {
         XCTAssertEqual(partialRefund2.total, "-8.10")
     }
 
+    /// Verifies that an Order ignores deleted refunds.
+    ///
+    func test_Order_deleted_refund_fields_are_ignored() {
+        guard let order = mapLoadOrderWithDeletedRefundsResponse() else {
+            XCTFail()
+            return
+        }
+
+        let refunds = order.refunds
+        XCTAssertEqual(refunds.count, 1)
+
+        let refund = refunds[0]
+        XCTAssertEqual(refund.refundID, 73)
+        XCTAssertEqual(refund.reason, "Cap!")
+        XCTAssertEqual(refund.total, "-16.00")
+    }
+
     func test_taxes_are_parsed_correctly() throws {
         // When
         let order = try XCTUnwrap(mapLoadOrderResponse())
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -807,8 +807,8 @@ extension OrderDetailsDataSource {
 
         let payment: Section = {
             var rows: [Row] = [.payment, .customerPaid]
-            if order.refunds.count > 0 {
-                let refunds = Array<Row>(repeating: .refund, count: order.refunds.count)
+            if !condensedRefunds.isEmpty {
```
jaclync: nit: I believe we also have a helper `isNotEmpty` that might be easier to read 😄

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -873,7 +873,7 @@ extension OrderDetailsDataSource {
 
     func refund(at indexPath: IndexPath) -> Refund? {
         let index = indexPath.row - Constants.paymentCell - Constants.paidByCustomerCell
-        let condensedRefund = order.refunds[index]
+        let condensedRefund = condensedRefunds[index]
```
jaclync, line 876: nice consolidating the usage of `condensedRefunds` instead of mixed `condensedRefunds` and `order.refunds` in this class 👍

Networking/NetworkingTests/Mapper/OrderMapperTests.swift
```diff
@@ -197,6 +197,23 @@ final class OrderMapperTests: XCTestCase {
         XCTAssertEqual(partialRefund2.total, "-8.10")
     }
 
+    /// Verifies that an Order ignores deleted refunds.
+    ///
+    func test_Order_deleted_refund_fields_are_ignored() {
```
jaclync: nit: would be nice to add the [Given/When/Then](https://en.wikipedia.org/wiki/Given-When-Then) comments (or a subset of them) in the test case, similar to the test case below (`test_taxes_are_parsed_correctly`) 🙂

jaclync (approved):
Thanks for fixing these issues! :shipit: 

Code: ✅ 
Testing: ✅ 

> My first attempt to fix this was to add a code to filter refunds in OrdersUpsertUseCase when we update the data in the storage. This approach didn't work in all cases. I discovered that when I pull-to-refresh the order, the updated order still has refunds that I thought should be filtered out. This happens because when we sync an order, the callback passes the model that was created from the network response and not the one that is fetched from the storage.

This sounds tricky - I see that sometimes we pass the object from the network response and sometimes the latest from storage in the completion block like:

https://github.com/woocommerce/woocommerce-ios/blob/da61682bbe5bb1da6f75e0ababbb7c61653372b5/Yosemite/Yosemite/Stores/ProductStore.swift#L236-L243

Filtering in the networking decoding step also works 👍 

order is the same as core (latest to oldest) | one refund deleted
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-16 at 17 25 05 | Simulator Screen Shot - iPhone 11 - 2020-12-16 at 17 25 47

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
eshurakov: Indeed! Thank you, looks better 👍

Networking/NetworkingTests/Mapper/OrderMapperTests.swift
eshurakov: Sounds good. Added in https://github.com/woocommerce/woocommerce-ios/pull/3349/commits/055eb6e19065116d2ae4f4eed2c8428a62d26ba5
Thank you!

eshurakov:
@jaclync, thank you for the review and suggestions!

PR #3350, opened 2020-12-15
Add Customer entity

### LINKED ISSUE #3241: Order Creation: networking, yosemite and storage layers for fetching and store Customers (step 4)

ealeksandrov:
Part of https://github.com/woocommerce/woocommerce-ios/issues/3241.

### Changes

This PR adds `Customer` entity to new CoreData model version (41).

I've bundled address properties into Customer object, like in Order:
https://github.com/woocommerce/woocommerce-ios/blob/78010115bcbcc9ea499eac75459c0c3f1287e07b/Storage/Storage/Model/Order%2BCoreDataProperties.swift#L11-L21

I made this choice because proper one-to-one relationships without inverse would lead to constantly displayed warnings:

image

It _may_ work with separate BillingAddress and ShippingAddress entities, where we can setup inverse `customer`. But I decided to go forward with simplest implementation, already proven itself in `Order`.

### Testing

Just check the code. No actions exposed to UI yet.

### DIFF
--- a/Storage/Storage/Model/Customer+CoreDataProperties.swift
+++ b/Storage/Storage/Model/Customer+CoreDataProperties.swift
@@ -1,0 +1,45 @@
+import Foundation
+import CoreData
+
+
+extension Customer {
+
+    @nonobjc public class func fetchRequest() -> NSFetchRequest<Customer> {
+        return NSFetchRequest<Customer>(entityName: "Customer")
+    }
+
+    @NSManaged public var siteID: Int64
+    @NSManaged public var userID: Int64
+    @NSManaged public var dateCreated: Date
+    @NSManaged public var dateModified: Date?
+    @NSManaged public var email: String
+    @NSManaged public var username: String?
+    @NSManaged public var firstName: String?
+    @NSManaged public var lastName: String?
+    @NSManaged public var avatarUrl: String?
+    @NSManaged public var role: String
+    @NSManaged public var isPaying: Bool
+    @NSManaged public var billingFirstName: String?
+    @NSManaged public var billingLastName: String?
+    @NSManaged public var billingCompany: String?
+    @NSManaged public var billingAddress2: String?
+    @NSManaged public var billingCity: String?
+    @NSManaged public var billingState: String?
+    @NSManaged public var billingPostcode: String?
+    @NSManaged public var billingCountry: String?
+    @NSManaged public var billingPhone: String?
+    @NSManaged public var billingEmail: String?
+    @NSManaged public var billingAddress1: String?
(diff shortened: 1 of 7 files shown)


Storage/Storage/Model/Customer+CoreDataProperties.swift
```diff
@@ -0,0 +1,45 @@
...
+    @NSManaged public var shippingAddress1: String?
+    @NSManaged public var shippingAddress2: String?
+    @NSManaged public var shippingCity: String?
+    @NSManaged public var shippingState: String?
+    @NSManaged public var shippingPostcode: String?
+    @NSManaged public var shippingCountry: String?
+    @NSManaged public var shippingPhone: String?
+    @NSManaged public var shippingEmail: String?
+
+}
```
pmusolino, line 45: It's so strange for me to see that the storage model is not specular to the networking model. Since there are a lot of duplicated fields here and inside the `Order` model, it doesn't make sense to create a new `Address` entity in the storage layer, with a relationship 1:1 inside `Customer` and `Order`?
Anyway I know that you followed the same idea of the `Order` entity and as described, we don't want to see these warnings, so it's ok for the moment to maintain these addresses like that.

Storage/StorageTests/CoreData/ManagedObjectModelsInventoryTests.swift
```diff
@@ -66,7 +66,10 @@ final class ManagedObjectModelsInventoryTests: XCTestCase {
             "Model 35",
             "Model 36",
             "Model 37",
-            "Model 38"
+            "Model 38",
+            "Model 39",
+            "Model 40",
+            "Model 41"
```
pmusolino, line 72: 💯 great

pmusolino (approved):
LGTM :ship:

Storage/Storage/Model/Customer+CoreDataProperties.swift
ealeksandrov: I completely agree with you and it's very annoying. We can't use single inverse relationship for multiple forward relationships (this is understandable for most cases when it's possible to have 2 relationships instead of always 1 in our case):

Screen Recording 2020-12-16 at 13 12 46

We can create `customerBilling` and `customerShipping` (and `orderBilling` + `orderShipping`) but that will look very strange as well (only one of relationships is expected to be non-nil at all times):

image

(but no warnings here)

PR #3352, opened 2020-12-15
Update App Store Metadata

jkmassel:
This PR is part of merging https://github.com/woocommerce/woocommerce-ios/pull/3337.

It looks like we haven't previously synced metadata from App Store Connect into the app repo. Additionally, some of the locale codes were incorrect.

This PR fixes those inconsistencies by:

- Ensuring that we're using the correct locale codes for App Store Connect
- Ensuring that when data is downloaded from GlotPress, it's moved into the correct directory
- Ensuring that all the keys we need for screenshots are always downloaded

**To Test:**
- Check out this PR
- Delete everything in the `fastlane/metadata` directory (but not the directory itself)
- Run `bundle exec fastlane deliver download_metadata`, then run `bundle exec fastlane download_app_store_translations`. The repo should now be clean.


.gitignore
```diff
@@ -74,6 +74,7 @@ fastlane/report.xml
 fastlane/Preview.html
 fastlane/screenshots
 fastlane/promo_screenshots
+/fastlane/metadata/review_information
```
AliSoftware: You probably didn't mean to have that leading `/` in here…

```suggestion
fastlane/metadata/review_information
```

Never seen this directory personally, what is it supposed to contain?

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
-      write_file(privacy_path, "https://automattic.com/privacy/")
+    # Ensure that localizations are present in the right places (some locale codes aren't correct)
+    {
+      "ar" => "ar-SA",
+      "de" => "de-DE",
+      "es" => "es-ES",
+      "fr" => "fr-FR",
+      "nl" => "nl-NL",
+      "zh-cn" => "zh-Hans",
+      "zh-tw" => "zh-Hant"
```
AliSoftware, line 559: We already have those locale maps in `Scripts/update-translations.rb` (and `fastlane/download_metadata.swift`).
_(In WordPress, a similar map is even centralized in `ScreenshotFastfile` but we don't have such dedicated file in Woo)_

It would be nice to DRY those to avoid having to update them in multiple places when locale code changes or we add new ones (maybe a `locales.yml` file just containing that mapping, and this line would then just be a matter of `YAML.load_file('locales.yml')` instead of the inline map?

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+    }.each { |source, destination|
```
AliSoftware: nit: ruby guidelines suggest to use `do…end` instead of `{ … }` for blocks (~ closures) that are not one-liners.

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+      source_dir = Dir.pwd + "/metadata/" + source
```
AliSoftware: Prefer using the more portable `File.join`:
```suggestion
      source_dir = File.join(Dir.pwd, 'metadata', source)
```

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+      Dir.children(source_dir).select { |file|
```
AliSoftware: Avoid using `select` for side-effects. `select` is for filtering, not for acting.

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+      "es" => "es-ES",
+      "fr" => "fr-FR",
+      "nl" => "nl-NL",
+      "zh-cn" => "zh-Hans",
+      "zh-tw" => "zh-Hant"
+    }.each { |source, destination|
+      source_dir = Dir.pwd + "/metadata/" + source
+      Dir.children(source_dir).select { |file| 
+        FileUtils.mv source_dir + "/" + file, Dir.pwd + "/metadata/" + destination + "/" + file
+      }
```
AliSoftware: Also, why move each file in `source_dir` individually into `destination` manually?

Mhhh… Ok maybe I better do a more global suggestion then 😅 

```suggestion
    # Ensure that localizations are present in the right places (some locale codes aren't correct)
    YAML.load_file('locales.yml').each do |source, destination|
      source_files = Dir.glob(File.join('metadata', source, '*'))
      dest_dir = File.join('metadata', destination)
      FileUtils.mkdir_p dest_dir # In case it doesn't exist yet locally
      FileUtils.mv source_files, dest_dir
    end
```

This is assuming you also add the `locales.yml` YAML file
```yaml
ar: ar-SA
de: de-DE
es: es-ES
fr: fr-FR
nl: nl-NL
zh-cn: zh-Hans
zh-tw: zh-Hant
```

That last part is totally optional (it's ok if you prefer to keep the array hardcoded inline in the `Fastfile`) but if you go through with it it could be worth also updating `Scripts/update-translations.rb` to use it too (I think it's the same list right?)

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+      FileUtils.remove_dir(source_dir)
+    }
+
+    # Copy source files into `en-US`
+    enUS_path = Dir.pwd + "/metadata/en-US/"
```
AliSoftware:
```suggestion
    enUS_path = File.join(Dir.pwd, 'metadata', 'en-US')
```

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+
+    FileUtils.mkdir_p(enUS_path) unless File.directory?(enUS_path)
```
AliSoftware: I'm not entirely sure the `unless it exists` test is necessary with `mkdir_p` – I don't think the `-p` option fails if destination exists already?

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+    [
+      "promo_screenshot_1.txt",
+      "promo_screenshot_2.txt",
+      "promo_screenshot_3.txt",
+      "promo_screenshot_4.txt",
+      "promo_screenshot_5.txt",
+    ].each { |source|
+      FileUtils.cp( Dir.pwd + "/appstoreres/metadata/source/" + source, enUS_path + source)
     }
+
```
AliSoftware:
```suggestion
    ].each do |source|
      FileUtils.cp( File.join(Dir.pwd, 'appstoreres', 'metadata', 'source', source), File.join(enUS_path, source) )
    end

```

Note that we could also consider using the fact that `FileUtils.cp` accepts an array as its first argument (source) and will then copy all the files listed in that array into the destination (that has to be a folder in that case). So `FileUtils.cp([list,of,files], enUS_path)` would be enough, except that we would still need to `map` the list of file basenames `promo_screenshot_N.txt` to their full path to produce the array, so probably not worth going that far in refactoring.

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+    # Replace the `default` directory with the contents of `en-US`
+    default_path = Dir.pwd + "/metadata/default/"
```
AliSoftware:
```suggestion
    default_path = File.join(Dir.pwd, 'metadata', 'default')
```

fastlane/Fastfile
```diff
@@ -549,17 +548,46 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+    FileUtils.remove_dir(default_path) if File.directory?(default_path)
+    FileUtils.copy_entry(enUS_path, default_path)
+
+    # Never store review information (it contains passwords and PII)
+    review_info_path = Dir.pwd + "/metadata/review_information/"
```
AliSoftware: …Will you guess my suggestion for this one?… 😛 

```suggestion
    review_info_path = File.join(Dir.pwd, 'metadata', 'review_information')
```

.gitignore
jkmassel: > You probably didn't mean to have that leading / in here…

True, fixed in 80f9b7dd6f11c660da64917ff3788c10feb148a8 because I didn't think to just hit the button in the dropdown 🤦

jkmassel: > Never seen this directory personally, what is it supposed to contain?

It contains the name and phone number of the App Store Connect contact, as well as the demo credentials used for evaluating the app. 

We don't want those to be public, which is why they're excluded.

fastlane/Fastfile
jkmassel: TIL! Fixed in 5b26e040050cb8b57f55264aa307dd5a5fd7c1da, thanks!!

fastlane/Fastfile
jkmassel: Yes indeed – this is the way.

Fixed in 49ee666b68863303a8b29c33f6a4aba413e27109

fastlane/Fastfile
jkmassel: Also fixed in 49ee666b68863303a8b29c33f6a4aba413e27109. Not sure how this one snuck in, but I promise I know better 😜

fastlane/Fastfile
jkmassel: Fixed in 49ee666b68863303a8b29c33f6a4aba413e27109

fastlane/Fastfile
jkmassel: You're right – turns out it's not necessary. Fixed in 49ee666b68863303a8b29c33f6a4aba413e27109

fastlane/Fastfile
jkmassel: Fixed in 49ee666b68863303a8b29c33f6a4aba413e27109

fastlane/Fastfile
jkmassel: Fixed in 49ee666b68863303a8b29c33f6a4aba413e27109

fastlane/Fastfile
jkmassel: Yeah, I very much agree that this is suboptimal 😅

I could refactor this, but IMHO this is something that our "fix up the localization pipeline" project should solve for us, and we should be able to just call a `download_translated_metadata` action or something and have it do the right thing. 

I'll be happy to burn this code at that time, but I'm not sure if it's worth interim work if we're gonna replace it anyway? 🤔

fastlane/Fastfile
jkmassel: Yeah, see my note above – we could certainly do this, but IMHO it'd be nice to have a solution to this that's not on a per-project basis.

fastlane/Fastfile
jkmassel: As part of the oft-aforementioned localization tooling, it'd be nice to also look into where/how to store these resources initially. IMHO it might make sense to have the `default` directory be the source for everything that goes off for translation, which would avoid the need for this entirely.

jkmassel:
Thanks for the awesome feedback @AliSoftware!!

Right now this is a bit of a basket of hacks – I don't mind coming back to do the `.yml` file thing, but IMHO it'd be nice to solve it more permanently with something like https://github.com/wordpress-mobile/fastlane-locale-map once we're able to spend some time on this particular tooling?

Anyway, ready for another look when you've a moment, thanks!

fastlane/Fastfile
AliSoftware: TIL about our locale map helper project. GTK for when I'll be tackling the localization cleanup task!

Yeah as I said in my comment this YAML idea is very optional so given this project this can wait indeed

fastlane/Fastfile
AliSoftware: image

fastlane/Fastfile
AliSoftware: Agreed you can skip over the YAML idea but the other refactorings (`File.join`, better use of `FileUtils`, …) in that suggestion are still worth applying even if you keep the list of locales hardcoded inline for now and that we do the rest later

fastlane/Fastfile
```diff
@@ -549,17 +548,50 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+      "nl" => "nl-NL",
+      "zh-cn" => "zh-Hans",
+      "zh-tw" => "zh-Hant"
+    }.each do |source, destination|
+      source_dir = File.join(Dir.pwd, "metadata", source)
+      Dir.children(source_dir).each { |file| 
+        source_file = File.join(source_dir, file)
+        destination_file = File.join(Dir.pwd, "metadata", destination, file)
+        FileUtils.mv(source_file, destination_file)
+      }
```
AliSoftware: Still worth using the `glob` suggestion that I mentioned in preview review, to avoid looping (esp given I think `children` includes `.` and `..` iirc?), and using `do…end` instead of `{…}`etc

fastlane/Fastfile
```diff
@@ -549,17 +548,50 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+      "promo_screenshot_1.txt",
+      "promo_screenshot_2.txt",
+      "promo_screenshot_3.txt",
+      "promo_screenshot_4.txt",
+      "promo_screenshot_5.txt",
+    ].each { |filename|
+      source = File.join(Dir.pwd, 'appstoreres', 'metadata', 'source', filename)
+      destination = File.join(enUS_path, filename)
+      FileUtils.cp(source, destination)
     }
```
AliSoftware: Ditto about `do…end`

fastlane/Fastfile
jkmassel: I actually ran into this issue which was why I used `Dir.children` 😅 

From https://ruby-doc.org/core-2.5.0/Dir.html#method-c-children:
> Returns an array containing all of the filenames except for “.” and “..” in the given directory. [...]

Also, it might just be that I'm very tired, but I'm not 100% sure what the `glob` suggestion was 😅 

One tricky thing – the `fastlane deliver download_metadata` action will put everything in the correct directory, but the `download_app_store_translations` lane still doesn't have everything that App Store Connect does. For that reason, I want to merge the directory contents instead of overwriting the directory. Not sure if that helps at all?

Sorry for the headache!!

fastlane/Fastfile
jkmassel: Ah yep – fixed in 41e1543f718e8f25c1f8291538c48cdcdaa48644. 

I wonder if a linter could catch this for us? 😁

fastlane/Fastfile
AliSoftware: The glob suggestion is here: https://github.com/woocommerce/woocommerce-ios/pull/3352#discussion_r544471384

It still copy the dir content not the dir itself (thanks to the `*`), that's indeed a thing I kept in mind when I tried that code in `pry` to validate it before submitting the suggestion

AliSoftware:
I won't have time to review it further before my EOY AFK, but the only thing that I think was left nit-fixing was the use of `glob` on the `mv` (to get an array of all the files to move and then move them in one single command) instead of looping one file at a time. It's not critical and must not prevent the merge of this PR, it was merely a code style improvement suggestion.

So apart from that, LGTM (but because it's Friday and I've had a hard day with lots of mistakes and brain farts I'll avoid approve it officially, might come back a bit later once rested to give the official green thumb after a final reading tho)

AliSoftware (approved):
Finally took another look at this today.

Code LGTM, and I also tested to clean up the `fastlane/metadata` folder and repopulating it with the appropriate fastlane lanes as suggested and can confirm that I end up with a clean repo (no change) and with the metadata in the appropriate locale folders.

:shipit:

jkmassel:
@AliSoftware thanks for the approval – I'll land this even if it's not 100% perfect in terms of reading files since it'll probably just be replaced anyway – this work is actually secondary to the changes I _actually_ want to be landing 😅

Thanks for the feedback (and I can't wait until my code here is all replaced) 😄

fastlane/Fastfile
```diff
@@ -549,17 +548,50 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+    [
+      "promo_screenshot_1.txt",
+      "promo_screenshot_2.txt",
+      "promo_screenshot_3.txt",
+      "promo_screenshot_4.txt",
+      "promo_screenshot_5.txt",
+    ].each do |filename|
+      source = File.join(Dir.pwd, 'appstoreres', 'metadata', 'source', filename)
+      destination = File.join(enUS_path, filename)
+      FileUtils.cp(source, destination)
```
AliSoftware, line 584: @jkmassel Actually now that I'm trying to apply a similar PR to other repos, I'm wondering what was the point of this, what's the need of copying the txt from `fastlane/appstoreres/metadata/source` to `fastlane/metadata/en_US`? (Or was the original intention to copy them to `fastlane/appstoreres/metadata/en_US` instead?)

fastlane/Fastfile
```diff
@@ -549,17 +548,50 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+    end
+
+    # Replace the `default` directory with the contents of `en-US`
+    default_path = File.join(Dir.pwd, 'metadata', 'default')
+    FileUtils.remove_dir(default_path) if File.directory?(default_path)
+    FileUtils.copy_entry(enUS_path, default_path)
```
AliSoftware, line 590: I'm starting to wonder if that part is worth it too, vs just not having the `default` directory at all. Do we have any script that depend on this `default` directory to exist? (And if so, wouldn't it be simpler to make it a `ln -s en_US default` symlink?)

PR #3353, opened 2020-12-16
Pin WPAuthenticator to a tag, not a branch

ctarda:
I just noticed that the Podfile references a branch as a source for the dependency with WPAuthenticator

## Changes
* Pin WPAuthenticator to a tag, instead of a branch

## How to test
* CI Should be enough. The tag points to the tip of the branch the Podfile was pointing to

Update release notes:


jaclync (approved):
LGTM! 🚢

PR #3354, opened 2020-12-16
[UL&S] Better error message when entering a non-valid url address

### LINKED ISSUE #3325: [UL&S] Error message when entering a non-valid url address
This can be delayed to phase 2, if it is too complicated to implement.

The following error seems to originate in WordPressKit. In theory, we should already be able to intercept it.

ctarda:
Closes #3325 

Before:

After:

The original error was fired by WPAuthenticator. This PR builds upon the mechanism implemented to take over errors emitted by WPAuthenticator: wordpress-mobile/WordPressAuthenticator-iOS#525

## Changes
* Add a new case to the Authenticator error (the domain error we use in the host app to map errors thrown by WPAuthenticator)
* Present the same error we present when we don't find a valid WP site (as discussed in the [issue comments](https://github.com/woocommerce/woocommerce-ios/issues/3325#issuecomment-744287061))

## How to test
* Check out the branch. Run `bundle exec pod install`
* Attempt to login with a Store Address, enter a not-valid site (i.e. GitHub.cox)

Update release notes:


Garance91540 (approved):
👏

Ecarrion (approved):
🚀

PR #3359, opened 2020-12-16
Don't dismiss order details and show error message if order is deleted

### LINKED ISSUE #2369: App Incorrectly Shows Order Deletion Notice

eshurakov:
Closes #2369 

The issue is that the app shows an "Order has been deleted from your store" notice and pops the view controller even though the order was not really deleted from the store. It happens for example when we sync orders and entries are temporarily removed from the database.

In this PR we're fixing this by removing `onDelete` handling.

The motivation behind it is as mentioned by @shiki: It's confusing and scary to merchants if they come up. If a merchant has deleted an order, they would know anyway and they don't need to be reminded of it.

## Tests

Follow a steps to reproduce from #2369 or update `fetchFilteredAndAllOrders` method in `OrderStore` and change 
`serialQueue.async { [weak self] in` 
to 
`serialQueue.asyncAfter(deadline: .now() + 3.0) { [weak self] in`

### Order Details

1. Navigate to Orders
2. Pull to refresh
3. Pick any Order
4. Verify that Order is not dismissed and "Order has been deleted from your store" notice is not shown (wait for a few seconds)

### Refunded Products

1. Make sure there is an order with refunded products
2. Navigate to Orders
2. Pull to refresh
3. Pick an Order from Step 1
4. Open Refunded Products
4. Verify that Order is not dismissed and "Order has been deleted from your store" notice is not shown (wait for a few seconds)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsNotices.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsNotices.swift
@@ -2,21 +2,6 @@ import Foundation
 import Yosemite
 
 final class OrderDetailsNotices {
-    /// Displays a Notice onscreen, indicating that the current Order has been deleted from the Store.
-    ///
-    func displayOrderDeletedNotice(order: Order) {
-        let message = String.localizedStringWithFormat(
-            NSLocalizedString(
-                "Order %@ has been deleted from your store",
-                comment: "Displayed whenever an Order gets deleted. It reads: Order {order number} has been deleted from your store."
-            ),
-            order.number
-        )
-
-        let notice = Notice(title: message, feedbackType: .error)
-        ServiceLocator.noticePresenter.enqueue(notice: notice)
-    }
-
     /// Displays the `Unable to delete tracking` Notice.
     ///
     func displayDeleteErrorNotice(order: Order, tracking: ShipmentTracking, onAction: @escaping () -> Void) {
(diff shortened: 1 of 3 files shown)


shiki (changes requested):
Hi @eshurakov! Thanks for looking into this. 

I'm sorry I didn't clarify the issue more. I think we should just remove the `onDelete` handling altogether. It causes more problems than it's worth. It's confusing and scary to merchants if they come up. If a merchant has deleted an order, **they would know** anyway and they don't need to be reminded of it.

Android also doesn't have a mechanism like this and just lets the Order page remain visible. The merchant will not be able to do anything anyway if it's permanently deleted. If it's just Trashed, then they can still interact with the UI, and that behavior still makes sense. 

We also removed this similar behavior in Reviews because it was also causing issues, not just in sync but even in push notifications. Please see this commit https://github.com/woocommerce/woocommerce-ios/commit/e59a5265e5468b673df997b884e8942359a5fc57.

eshurakov:
Thank you @shiki! 
I have updated my PR with the suggested approach. I found that we also had a similar `onDelete` handling in `RefundedProductsViewController` and removed it from there as well.

shiki (approved):
Tested and works great! Thank you for being thorough! 🙏 

:shipit:

eshurakov:
Thank you for the review @shiki, merging it 👍

PR #3371, opened 2020-12-17
Set background color of StorePickerViewController to match the background color provided by WPAuthenticator

### LINKED ISSUE #3364: [UL&S] Mixed styles on epilogue/error screen when WordPress.com account has no connected WooCommerce stores
**Describe the bug**

There are odd, mixed styles on the epilogue/error screen when WordPress.com account has no connected WooCommerce stores.

**To Reproduce**
3. Enter the email address for a WordPress.com account that is *not* connected to any WooCommerce stores.
4. Finish logging in with the account password or magic link.
5. Notice that the epilogue/error screen shows the error message, illustration, and primary CTA in the middle of the screen with a white background, surrounded by the old epilogue styling (grey background).

**Screenshots**

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 14.2

ctarda:
Closes #3364 

| Before | After |
| ---- | ---- |
|  |  |

## Changes
* Set the background color of StoresPickerView controller to the background color provided by WPAuthenticator

## How to test
* Follow the repro instructions in the original issue: login with magic link to an account without linked stores

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
+++ b/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
@@ -185,13 +185,13 @@ class StorePickerViewController: UIViewController {
 private extension StorePickerViewController {
 
     func setupMainView() {
-        view.backgroundColor = .listBackground
+        view.backgroundColor = backgroundColor()
     }
 
     func setupTableView() {
         tableView.registerNib(for: EmptyStoresTableViewCell.self)
         tableView.registerNib(for: StoreTableViewCell.self)
-        tableView.backgroundColor = .listBackground
+        tableView.backgroundColor = backgroundColor()
     }
 
     func setupAccountHeader() {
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
```diff
@@ -243,6 +243,10 @@ private extension StorePickerViewController {
         preselectStoreIfPossible()
         reloadInterface()
     }
+
+    func backgroundColor() -> UIColor {
+        return WordPressAuthenticator.shared.unifiedStyle?.viewControllerBackgroundColor ?? .listBackground
```
Ecarrion, line 248: Isn't this style defined in our styles? 
I see a big optional change and also a singleton usage that makes me a bit nervous 😓

Ecarrion (approved):
Left a small comment 😇

WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
ctarda: It is, but... WordPresAuthenticator is a singleton, so there isn't much we can do there, and the style needs to match the style exposed by WPAuthenticator. 

That being said, in my opinion, exposing authentication as we do with any other global service would make sense in the long run.

PR #3374, opened 2020-12-17
Reprint Shipping Label: reprint action with a coordinator

jaclync:
Part of #2169 

## Why

This PR implements the reprint action to **request printable document for a shipping label**, and then **present the AirPrint dialog** on success. I wanted to make the navigation testable, and created a `ReprintShippingLabelCoordinatorProtocol` that has a set of navigation functions during the reprint flow. Before writing `ReprintShippingLabelCoordinatorProtocol`, I tried using `Coordinator` protocol and callback-based actions for various navigation actions in `ReprintShippingLabelViewController`. But I felt like I didn't need a `UINavigationController` (it's also an implementation detail) and the callbacks with a callback within itself (e.g. `func showPaperSizeSelector`) are quite hard to read (and we can't specify the parameter name). If you have any suggestions on how to make the coordinator/navigation code better, I'm all open to changes!

## Changes

- Moved 5 files without any changes (the ones with 0 diffs)
- Created `ReprintShippingLabelCoordinatorProtocol` with all navigation actions for reprinting a shipping label. `ReprintShippingLabelCoordinator` implements the protocol. Added unit tests for `ReprintShippingLabelCoordinator`
- In `OrderDetailsViewController`, added a lazy var `ReprintShippingLabelCoordinator` that shows reprint UI when tapping on the reprint CTA
- Updated `ReprintShippingLabelViewController` to call `ReprintShippingLabelCoordinatorProtocol` for navigation actions (paper size selector and reprint action so far. there are two more actions in the followup subtasks to present informational modals)
- Added a function to `ReprintShippingLabelViewModel` to request printable data for a shipping label with test cases

## Testing

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> after syncing, a shipping label package card should be shown
- Tap on "Reprint Shipping Label"
- Choose a different paper size or leave it as the default
- Tap "Print Shipping Label" --> an in-progress modal is shown, then an AirPrint modal is shown with a preview of the shipping label in the selected paper size
- (Optional) If you have a printer that supports AirPrint, feel free to try connecting to the printer in the AirPrint modal 🖨️ 

## Example screenshots

simulator

When there is an error generating the printable document, an error alert is shown (ref p91TBi-3QL-p2#comment-2418):

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Reprint Shipping Label/ReprintShippingLabelCoordinator.swift
```diff
@@ -0,0 +1,112 @@
...
+    /// Dismisses the in-progress modal when the shipping label document result is ready.
+    /// When the result is successful, we present an AirPrint modal.
+    /// If the result fails with an error, we present an error alert.
+    /// - Parameter result: The result of shipping label document request, could be `ShippingLabelPrintData` on success or an error.
+    func dismissReprintInProgressUIAndPresentPrintingResult(_ result: Result<ShippingLabelPrintData, ReprintShippingLabelError>)
+}
+
+/// Implements `ReprintShippingLabelCoordinatorProtocol` that handles all navigation actions.
+final class ReprintShippingLabelCoordinator {
+    private let sourceViewController: UIViewController
```
eshurakov, line 6: This looks like a retain cycle because view controller also retains coordinator. 🤔

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -27,6 +27,10 @@ final class OrderDetailsViewController: UIViewController {
         return EntityListener(storageManager: ServiceLocator.storageManager, readOnlyEntity: viewModel.order)
     }()
 
+    /// Coordinator for shipping label reprint UX navigation.
+    ///
+    private lazy var reprintShippingLabelCoordinator = ReprintShippingLabelCoordinator(sourceViewController: self)
```
eshurakov: Wondering if it is possible to have the same lifecycle for Coordinator and view controllers that it coordinates. So we would create a coordinator when we need to start a flow and it would deallocate once all the screens from the flow are gone.

eshurakov (commented):
I wanted to share my thoughts on the Coordinator implementation.

I believe that Coordinator should be the one responsible for all the transitions between the screens in the flow. For example instead of passing Coordinator as a collaborator to the view controller, we can pass a closure / delegate. After view controller is done with its part of the process it would notify Coordinator via this closure. This way view controller doesn't need to be even aware that it is presented from a certain Coordinator. There is a Russian proverb saying "The less you know, the more soundly you sleep" and its English equivalent "Ignorance is bliss", which I think can be applied to view controllers - less dependencies and less knowledge about the environment around is better :-)

Writing some pseudo code I think that Coordinator can be something like this. I didn't think much about names, it's just a high level code :-)

In this case Coordinator is responsible for transitions and it decides based on the output from view controllers what is the next step. Coordinator can also keep a data object with information that view controllers need and inject it to view controllers when they are created. 

```
class ReprintShippingLabelCoordinator {
    let sourceViewController: UIViewController
    let shippingLabel: ShippingLabel

    weak var reprintViewController: ReprintShippingLabelViewController?
    weak var inProgressViewController: InProgressViewController?

    init(...) { }

    func showReprintUI() {
        let reprintViewController = ReprintShippingLabelViewController(shippingLabel: shippingLabel) { result
            switch result {
            case .selectPaperSize:
                self?.showPaperSizeSelector()
            case .reprint(let paperSize):
                self?.reprint(with: paperSize)
            }
        }
        self.reprintViewController = reprintViewController

        // Since the reprint UI could make an API request for printing data, disables the bottom bar (tab bar) to simplify app states.
        reprintViewController.hidesBottomBarWhenPushed = true
        sourceViewController.show(reprintViewController, sender: sourceViewController)
    }

    func showPaperSizeSelector() {
        let command = ShippingLabelPaperSizeListSelectorCommand(paperSizeOptions: paperSizeOptions, selected: selectedPaperSize)
        let listSelector = ListSelectorViewController(command: command) { paperSize in
            self?.reprintViewController.updatePaperSize(paperSize)
        }
        sourceViewController.show(listSelector, sender: sourceViewController)
    }

    func reprint(with: paperSize) {
        presentReprintInProgressUI()

        requestDocumentForPrinting(shippingLabel, paperSize) {
            self?.inProgressViewController.dismiss()
            self?.presentReprint(printData)
        }
    }

    func presentReprintInProgressUI() {
        let viewProperties = InProgressViewProperties(title: Localization.inProgressTitle, message: Localization.inProgressMessage)
        let inProgressViewController = InProgressViewController(viewProperties: viewProperties)
        inProgressViewController.modalPresentationStyle = .overCurrentContext
        self.inProgressViewController = inProgressViewController
        sourceViewController.present(inProgressViewController, animated: true, completion: nil)
    }

    func presentReprint(printData) {
        let data = Data(base64Encoded: printData.base64Content)
        let printController = UIPrintInteractionController()
        printController.printingItem = data
        printController.present(animated: true, completionHandler: nil)
    }
}
```

If the code is structured this way testing Coordinator can be a bit hard as we will have only one public method `showReprintUI` with the rest of implementation being hidden. This can be solved by injecting an object that will be responsible for creating view controllers. We can then mock it in tests and provide different data to coordinator and check that the flow goes in the right direction.

```
class ReprintViewControllerProvider {
    func shippingLabelViewController(with shippingLabel: ShippingLabel, completion: @escaping (Result) -> Void) -> UIViewController
    func paperSizeSelectorViewController(completion: @escaping (PaperSize) -> Void) -> UIViewController
    ...
}
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Reprint Shipping Label/ReprintShippingLabelCoordinator.swift
jaclync: oops great catch, my bad! I was hoping to make the coordinator have the same life cycle as reprint UX to avoid the retain cycle but hadn't figured out a way. Also forgot to double check with memory debugger 😅

jaclync:
Thanks for the feedback @eshurakov! I updated the PR for another pass!

> I believe that Coordinator should be the one responsible for all the transitions between the screens in the flow. For example instead of passing Coordinator as a collaborator to the view controller, we can pass a closure / delegate. After view controller is done with its part of the process it would notify Coordinator via this closure. This way view controller doesn't need to be even aware that it is presented from a certain Coordinator.

I like it 👍 the reprint view controller was taking more responsibility than it needed, and it makes sense to have the coordinator handle the UI actions. In https://github.com/woocommerce/woocommerce-ios/pull/3374/commits/0dd260b466ced02e55989d06d71c803d2b6c9ded, I updated `ReprintShippingLabelViewController` to have a `onAction: ((ActionType) -> Void)?` closure for the coordinator to handle all actions from the reprint UI. I was able to unit test the actions getting `ReprintShippingLabelViewController` from the source view controller, see if it makes sense 😄. Also, the coordinator now has the same life cycle as `ReprintShippingLabelViewController` with strong `self` in action handling and `OrderDetailsViewController` just instantiates the coordinator and call its only public method `showReprintUI()`. I checked memory debugger and don't see the retained objects after leaving order details as before anymore.

> There is a Russian proverb saying "The less you know, the more soundly you sleep" and its English equivalent "Ignorance is bliss", which I think can be applied to view controllers - less dependencies and less knowledge about the environment around is better :-)

Haha nice, very true! Since view controllers are hard to test, it's worth making sure it's as simple/lean as possible 🙂

Networking/Networking/Model/ShippingLabel/ShippingLabelPrintData.swift
```diff
@@ -8,6 +8,11 @@ public struct ShippingLabelPrintData: Decodable, Equatable {
 
     /// The content of shipping label document for printing, with Base64 encoding.
     public let base64Content: String
+
+    public init(mimeType: String, base64Content: String) {
+        self.mimeType = mimeType
+        self.base64Content = base64Content
+    }
```
eshurakov, line 20: Idea: how about adding `Data(base64Encoded: base64Content)` here as a computed property?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Reprint Shipping Label/ReprintShippingLabelCoordinator.swift
```diff
@@ -0,0 +1,118 @@
...
+                self.requestDocumentForPrinting(paperSize: paperSize) { result in
+                    self.dismissReprintInProgressUI()
+                    switch result {
+                    case .success(let printData):
+                        self.presentAirPrint(printData: printData)
+                    case .failure(let error):
+                        DDLogError("Error generating shipping label document for printing: \(error)")
+                        self.presentErrorAlert(title: Localization.reprintErrorAlertTitle)
+                    }
+                }
```
eshurakov: Maybe extract this to a separate method to reduce indentation and improve readability a bit?

eshurakov (commented):
Looks good! Tested it on iPhone and iPad ✅ 

While testing I saw some warnings in the console that I think worth addressing:

1. Reprint View Controller has some issues with constraints. Happens every time you open this view controller.

```
[LayoutConstraints] Unable to simultaneously satisfy constraints.
    Probably at least one of the constraints in the following list is one you don't want. 
    Try this: 
        (1) look at each constraint and try to figure out which you don't expect; 
        (2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x6000023100f0 UIView:0x7fb02a0122d0.height == 8   (active)>",
    "<NSLayoutConstraint:0x600002310780 V:[UIView:0x7fb02a0122d0]-(0)-|   (active, names: '|':UITableViewCellContentView:0x7fb089cd3320 )>",
    "<NSLayoutConstraint:0x600002311bd0 V:|-(0)-[UIView:0x7fb02a0122d0]   (active, names: '|':UITableViewCellContentView:0x7fb089cd3320 )>",
    "<NSLayoutConstraint:0x600002310a00 'UIView-Encapsulated-Layout-Height' UITableViewCellContentView:0x7fb089cd3320.height == 8.5   (active)>"
)

Will attempt to recover by breaking constraint 
<NSLayoutConstraint:0x6000023100f0 UIView:0x7fb02a0122d0.height == 8   (active)>
```

2. Presenting progress view controller also results in a warning. We try to present from Order details, but it is no longer on screen, because Reprint VC was pushed on the stack.

`[Presentation] Presenting view controller <WooCommerce.InProgressViewController: 0x7fb099cbf680> from detached view controller <WooCommerce.OrderDetailsViewController: 0x7fb099c5e620> is discouraged.`

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Reprint Shipping Label/ReprintShippingLabelCoordinator.swift
jaclync: sounds good! updated in https://github.com/woocommerce/woocommerce-ios/pull/3374/commits/65812bd2ea8c065292ced25bf96fa011dc8ee16b

Networking/Networking/Model/ShippingLabel/ShippingLabelPrintData.swift
jaclync: good idea! added/updated in https://github.com/woocommerce/woocommerce-ios/pull/3374/commits/0991498889a2c811e263fcc2ce1e5f402f6a1577

jaclync:
@eshurakov thanks for testing & bringing up the warnings in the console (I'll pay more attention to them next time)! I updated the PR and responded to the console issues here:

> 1. Reprint View Controller has some issues with constraints. Happens every time you open this view controller.

Auto Layout warnings from `UITableView` reusable views always get me stuck 😓 In this case, it's `SpacerTableViewCell` with one subview of height constraint 8px and pinned to cell content view's top and bottom. However, when the cell is actually rendered, the content view's height is increased by 0.5px from the separator that we have little control over. To work around the warning, I updated the spacer view's constraints in https://github.com/woocommerce/woocommerce-ios/pull/3374/commits/1821056028864024b5348aedcf5e81baa7667ebd to remove the bottom edge constraint and add a constraint for equal height to the content view @ 750 priority (not required):

before | after
-- | --
Screen Shot 2020-12-21 at 11 18 05 AM | Screen Shot 2020-12-21 at 11 17 46 AM

I'm not sure what's the best way to account for the 0.5px separator that we hide in this case. We could assume the separator being 0.5px tall and deduct that from the 8px height constraint (so configuring `SpacerTableViewCell`'s height to 7.5px instead of 8px), but it sounds quite hacky since the height could change in future UIKit updates so I leave the extra 0.5px as it is for now.

> 2. Presenting progress view controller also results in a warning. We try to present from Order details, but it is no longer on screen, because Reprint VC was pushed on the stack.
> `[Presentation] Presenting view controller <WooCommerce.InProgressViewController: 0x7fb099cbf680> from detached view controller <WooCommerce.OrderDetailsViewController: 0x7fb099c5e620> is discouraged.`

This is interesting! I'm guessing UIKit considers `OrderDetailsViewController` as "detached" because it isn't the top view controller. I updated to pass the navigation controller instead of `OrderDetailsViewController` as the source view controller in https://github.com/woocommerce/woocommerce-ios/pull/3374/commits/4e6fe9ac180a39f42c6f3528254d8231230663fb

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Reprint Shipping Label/ReprintShippingLabelCoordinator.swift
```diff
@@ -0,0 +1,121 @@
+import UIKit
+import Yosemite
+
+/// Coordinates navigation actions for reprinting a shipping label.
+final class ReprintShippingLabelCoordinator {
```
eshurakov, line 5: Idea: I think it would be nice to mention in the comments to the `ReprintShippingLabelCoordinator` that we're intentionally retaining `self` in callbacks to avoid confusion.

eshurakov (approved):
Looks great! :shipit: 

1. Auto Layout warnings

> To work around the warning, I updated the spacer view's constraints in 1821056 to remove the bottom edge constraint and add a constraint for equal height to the content view @ 750 priority (not required)

My go to approach in a situation like this is to just lower the priority of the bottom constraint. I usually set it to 999, so it's high enough to be considered important, but not required :-) This way cell can calculate its own height using auto layout, but in case someone outside is forcing a different height for the cell, constraints will not break. It also works well for cell animations because tableview sometimes temporarily changes cells height.

Your approach definitely works too 👍 

2. Detached view controller

> I'm guessing UIKit considers OrderDetailsViewController as "detached" because it isn't the top view controller.

After navigation transition is finished and `OrderDetailsViewController` goes off screen, its view is removed from the view hierarchy (but still kept in the viewController array of nav controller) and its `window` property is nil. It can't access any other views or window, that's why it is considered to be detached. That's how I understand it :-)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Reprint Shipping Label/ReprintShippingLabelCoordinator.swift
jaclync: Makes sense! added to `showReprintUI` in https://github.com/woocommerce/woocommerce-ios/pull/3374/commits/267dda34b2cdf7a122618e2ec90b1dffb6a6f9a7

Garance91540 (approved):
👏👏

jaclync:
Merging now to add two more actions for the info links!

Issue #3377, opened 2020-12-18
[UL&S] Provide custom UI for error presented when logging in with Google account not linked to WordPress.com


ctarda:
This would be the ideal solution to #3365 .

We found that, if we provide WPAuthenticator with a flag to blog the creation of WP.com accounts with Google accounts, the error message presented contains two buttons whose:
1. Titles are not really that consistent with the rest of the language WooCommerce uses (i.e. site address vs Store Address)
2. Actions are not exactly what we would like.

So the solution to that would be to propagate up the error captured in GoogleAuthViewController in a way that it reaches AuthenticationManager so that we can provide custom UI back.

PR #3378, opened 2020-12-18
Prevent signups to WordPress.com with google accounts

### LINKED ISSUE #3365: [UL&S] Logging in with Google account not linked to WordPress.com is a dead end
**Describe the bug**
If you try to log in to WordPress.com using a Google account that isn't already linked to WordPress.com, you hit a dead end:

* The "Log in" screen says it will create a new WordPress.com account, but it doesn't.
* The screen you end up on has options to try with another email or site address, but tapping on them doesn't go anywhere.

**To Reproduce**
3. Select "Continue with Google."
4. Select or log in to a Google account that is *not* linked to a WordPress.com account.
5. On the resulting error screen, tap the options to try logging in a different way and notice they don't do anything.

**Screenshots**

"Log in" screen|Error screen
-|-
|

**Mobile Environment**
Please include:

- Device: iPhone XS
- iOS version: 14.2

ctarda:
Closes #3365 

This is not the best possible solution to the issue, but I think it strikes a reasonable balance at this moment.

What I have done is block signing up to wp.com accounts with a google account. This goes in line with the spirit of some of the other changes we have implemented to the login flow on WooCommerce. 

We could (and probably should in phase 2) present a custom UI instead of the UI presented by WPAuthenticator (that is the screen with the "Try with another email" and "try with site address" buttons, but that would require reaching deep into WPAuthenticator, which seems a bit risky at this point. 

Anyway, I have passed a flag to WPAuthenticator to block account creation with a google account and that seems to work reasonably well. 

A couple of things to have in mind:
* Customizing the text in the two buttons ("Try with another email" and "Try with site address") would require reaching into WPAuthenticator deep as well. 
* I have logged #3377 to implement a proper, custom UI for this error screen, so we don't loose track of this for phase 2.

## Changes
* Pass a flag to WordPressAuthenticator to prevent signing up to a WordPress.com account with a google account. 

## How to test
* Check out the branch. Run `bundle exec pod install` just in case.
* Log out if necessary
* Tap "Continue with WordPress.com". Tap "Continue with Google"
* Log into a Google account not linked to a WordPress.com account.
* Notice the error screen. 
* On the error screen, tap "Try with another email" That should navigate to the root of the login process
* Tap "Try with the site address". That should lead to the screen to enter the Store Address.

Update release notes:


jaclync (approved):
Looks good! :shipit: 

Some non-blocking observations:

- Rachel mentioned the copy in https://github.com/woocommerce/woocommerce-ios/issues/3365 in the screenshot below, maybe we can pass a different string to remove Google from this disclaimer?



- The CTAs in the no account screen has inconsistent color scheme from the rest of the Woo app:



> Customizing the text in the two buttons ("Try with another email" and "Try with site address") would require reaching into WPAuthenticator deep as well.

During testing, I first expected "Try with another email" to go back to the previous screen (the screen with CTA to log in with Google). Maybe "Try with another account" would be more accurate for this navigation?

ctarda:
Thanks for the review @jaclync !

I definitely think we should provide a custom UI here. That would solve all the issues you mentioned. 

In the meantime, I'll merge as soon as possible, so we can cut a new beta.

ctarda:
Bumping this @woocommerce/mobile-designers Can we merge?

ghost:
Hey, @ctarda @jaclync yes we will need to remove 'Google' from the disclaimer string as we don't actually create an account with the 'Continue with Google' flow.

Screen Shot 2020-12-22 at 09 46 55

It also looks like we're missing the *Email not associated with WP.com account* screen.

ctarda:
Bumping this up to the top. I have pushed an update to the Terms of Service that removes "google":



@adamzelinski the screen you mentioned (Email not associated with WP.com account) would be part of what we would implement in a second phase, as part of #3377 

A couple of weeks ago, when we decided to roll back UL&S from 5.7 we discussed that, this error message, or the lack of a better error screen, is not blocking any core login flow and we could defer to a second phase. What do you think @adamzelinski , would it be possible to defer and merge as it is, with the updated copy?

ghost:
> A couple of weeks ago, when we decided to roll back UL&S from 5.7 we discussed that, this error message, or the lack of a better error screen, is not blocking any core login flow and we could defer to a second phase. What do you think @adamzelinski , would it be possible to defer and merge as it is, with the updated copy?

Hey @ctarda yes, happy to merge as is with the updated copy and do the rest in the second phase.

ghost (approved):

ctarda:
Thanks @adamzelinski !

PR #3380, opened 2020-12-18
Create and use a cell view model for cells on order list screen

### LINKED ISSUE #2308: Use a tiny CellViewModel for OrdersViewController
As shown in https://github.com/woocommerce/woocommerce-ios/pull/2298, the `OrderDetailsViewModel` is coupled in 2 ViewControllers. This might affect the cost of change in the future. 

It is probably better to have a tiny `OrdersCellViewModel` that provides only the data required to present the cell in the Order list.

eshurakov:
Closes #2308 

In this PR we're introducing a `OrderListCellViewModel` - a tiny view model for cells on order list screen. Previously cells were populated using `OrderDetailsViewModel`, which is quite big and has a lot of things not needed to render cells.

## Changes 
* new `OrderListCellViewModel` - it has everything that order list cell needs
* `statusResultsController` is moved from `OrderListViewController` to `OrderListViewModel`, which allowed to pass `OrderStatus` into view model along with an order
* `OrderSearchCellViewModel` is deleted because it can be fully replaces with `OrderListCellViewModel`

## Testing
Introduced changes should not affect any UI or how the app works, but to double check please do a quick smoke test of order list and order search screens.

Update release notes:


WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import Yosemite
+
+struct OrderListCellViewModel {
```
shiki: Would you mind adding a simple comment saying what this is for and where it is used? 🙂

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
```diff
@@ -43,6 +43,23 @@ final class OrderListViewModel {
     ///
     private var isAppActive: Bool = true
 
+    /// Used for looking up the `OrderStatus` to show in the `OrderTableViewCell`.
+    ///
+    /// The `OrderStatus` data is fetched from the API by `OrdersTabbedViewModel`.
+    ///
+    private lazy var statusResultsController: ResultsController<StorageOrderStatus> = {
+        let storageManager = ServiceLocator.storageManager
```
shiki: ⚠️ This class already has a reference to `storageManager` so we should use that. 🙂 We're also trying to avoid surprising `ServiceLocator` usage and preferring to inject them so there's more clarity in when these objects change and where they come from. 

https://github.com/woocommerce/woocommerce-ios/blob/fff92231e12c54cccd4e815975f8bda57a33057f/WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift#L53-L57

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+        return formatter.string(from: order.dateCreated)
+    }
+
+    var status: OrderStatusEnum {
+        return orderStatus?.status ?? order.status
+    }
+
+    var statusString: String {
+        return orderStatus?.name ?? order.status.rawValue
+    }
```
shiki: ⚠️  I think we should copy over the important comment from the original code. It'd be great if we can add a comment for this property. 

```swift
            // There are unsupported extensions with even more statuses available.	
            // So let's use the order.status to display those as slugs.
```

WooCommerce/WooCommerceTests/ViewModels/OrderListCellViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    func test_OrderStatus_is_used_when_passed_to_view_model() {
+        // Given
+        let order = MockOrders().sampleOrder()
+        let orderStatus = OrderStatus(name: UUID().uuidString,
+                                      siteID: 0,
+                                      slug: UUID().uuidString,
+                                      total: 0)
+        let viewModel = OrderListCellViewModel(order: order, status: orderStatus)
```
shiki: Thanks for using Given-When-Then! 

I would say this is the When part. 🙂 

```suggestion

        // When
        let viewModel = OrderListCellViewModel(order: order, status: orderStatus)
```

WooCommerce/WooCommerceTests/ViewModels/OrderListCellViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
+import XCTest
+import TestKit
+
+@testable import Yosemite
```
shiki: ⚠️ This doesn't need to be `@testable`. 🙂 

```suggestion
import Yosemite
```

It seems to compile without it. 🤪

WooCommerce/WooCommerceTests/ViewModels/OrderListCellViewModelTests.swift
```diff
@@ -0,0 +1,56 @@
...
+@testable import WooCommerce
+
+final class OrderListCellViewModelTests: XCTestCase {
```
shiki: Thanks for this!

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
shiki: ⚠️ Would you mind moving this to the same folder as `OrderTableViewCell`? We're trying to organize files by feature now instead of grouping all ViewModels together. Same for the `OrderListCellViewModelTests.swift` too please. 🙂 




P.S. `ViewRelated` is a weird folder name. I think we should get rid of that soon. 😬

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import Yosemite
+
+struct OrderListCellViewModel {
+    private let order: Order
+    private let orderStatus: OrderStatus?
+
+    private let currencyFormatter = CurrencyFormatter(currencySettings: ServiceLocator.currencySettings)
```
shiki: ❓  For memory conservation purposes, it might be worth it to store this in `OrderListViewModel` instead and we will just inject the instance here. In that way, there's only one `currencyFormatter` lying around. What do you think?

WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
```diff
@@ -123,18 +114,6 @@ private extension OrderTableViewCell {
         paymentStatusLabel.backgroundColor = paymentColor
         paymentStatusLabel.layer.borderColor = borderColor
     }
-
```
shiki, line 126: Really great work being thorough in the refactoring!

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
```diff
@@ -43,6 +43,23 @@ final class OrderListViewModel {
...
+        let descriptor = NSSortDescriptor(key: "slug", ascending: true)
```
shiki, line 51: ℹ️ Would you mind adding this comment?

```swift
// TODO We should also include a siteID predicate here. 
```

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
```diff
@@ -163,9 +181,36 @@ private extension OrderListViewModel {
...
 }
 
+// MARK: - Order Status
+
+private extension OrderListViewModel {
+    /// Setup: Status Results Controller
+    ///
+    func configureStatusResultsController() {
+        statusResultsController.predicate = NSPredicate(format: "siteID == %lld", siteID)
+        try? statusResultsController.performFetch()
```
shiki: This is a low-hanging fruit and only if you have time. We can add error logging here. 

```suggestion
        do {
            try statusResultsController.performFetch()
        } catch {
            CrashLogging.logError(error)
        }
```

shiki (changes requested):
Thanks for working on this, @eshurakov! Another really great work. 

I have a couple of requests, marked with ⚠️. And something I would like your opinion on, marked with ❓. The rest are just nitpicks. 😄

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
eshurakov: Good point, done!

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
eshurakov: Totally agree, it's always good to know where objects came from and who and when calls them 👍

WooCommerce/WooCommerceTests/ViewModels/OrderListCellViewModelTests.swift
eshurakov: Thank you! Updated :-)

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
eshurakov: Sorry I wasn't aware of this. Moved it! 👍

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
eshurakov: I saw that we were adding a predicate later in `configureStatusResultsController` and moved it here as we already know `siteID`.

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
eshurakov: Done! Thanks for the suggestion 👍

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
eshurakov: That's a good point and it would for sure improve memory allocation. I have an alternative suggestion, that may help in some other places as well.

#### Suggestion 1
* `CurrencyFormatter` is now a `class`, but it doesn't need to be. It doesn't have methods that mutate internal state and has only one member, which is a `let` member (CurrencySettings). If we convert it to `struct` we won't need to store it as a member variable in `OrderListCellViewModel`, but instead create and use the formatter "inline". I believe it will be basically free to create a currency formatter like this.

```
var total: String? {
      return CurrencyFormatter().formatAmount(order.total, with: order.currency)
}
```

* Maybe set a default value for `currencySettings` in the init method of `CurrencyFormatter`, because I see that it is always created this way: `CurrencyFormatter(currencySettings: ServiceLocator.currencySettings)`

#### Suggestion 2
I think it would be nice to have it stored somewhere, similar to how date formatters are created only once in `DateFormatter+Woo`. Of course the difference is that currency formatter depends on the site settings and needs to be updated the site is updated / changed. But this way there will be only one instance and everyone can use it without creating it over and over again :-)

eshurakov: Let me know what you think, @shiki.
I think "Suggestion 2" may be a bit more beneficial in a long term run, but would require significant changes to multiple classes.
On the other hand other suggestions are also beneficial and require less code :-)

WooCommerce/Classes/ViewRelated/Orders/OrderListViewModel.swift
shiki: Ah, gotcha. I didn't notice that. Thanks for moving it!

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
shiki: Thanks!

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
shiki: Thanks for the detailed suggestion. Suggestion 1 sounds better for me. It looks like we should be passing `CurrencySettings` around instead of `CurrencyFormatter`. 

But I believe this is something that we should discuss more as a team since there was a previous discussion related to `CurrencyFormatter` (p91TBi-3bP-p2) and I might be missing something important. 

So, let's keep what we have in here as-is for now. And maybe create an issue for your suggestion? 🙂

shiki (approved):
Thanks, @eshurakov! :shipit:

WooCommerce/Classes/ViewModels/Order List/OrderListCellViewModel.swift
eshurakov: Thank you for sharing the link to a previous discussion. I will create an issue so that everyone can chip in there :-)

eshurakov:
Thank you for the review @shiki

PR #3389, opened 2020-12-21
Shipping Labels M1: handle A4 paper size (not supported in mobile yet)

### LINKED ISSUE #3340: Shipping Labels: handle A4 paper size

jaclync:
Fixes #3340 

## Why

Right now, only stores in the US can create a shipping label with 3 paper size options: legal, letter, and label. However, after a shipping label is created, if the store country is changed to outside of US, CA, MX, DO (ref CKNC7KACA/p1607010932150400?thread_ts=1606980019.149700&cid=CKNC7KACA-slack) A4 paper size becomes available.

This is still an edge case as most stores don't change the store country and reprint an existing shipping label. In the future, we might want to support A4 paper size if it is the shipping label preference (that we fetch in `ShippingLabelSettings`).

## Changes

- Added a new case `ShippingLabelPaperSize.a4`
- In the UI side, `ReprintShippingLabelViewModel` already detects an unsupported paper size and defaults to the first paper size option (`legal`) in the following code. This PR added a test case for this scenario
https://github.com/woocommerce/woocommerce-ios/blob/f09cd9ca3777cd0583452c208b9001dc8c253fa0/WooCommerce/Classes/ViewModels/Order%20Details/Shipping%20Labels/ReprintShippingLabelViewModel.swift#L33-L35

## Testing

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/). You can set up your store to use a test credit card with instructions in p91TBi-3xD-p2.

- In wp-admin, create a shipping label for an order
- Change store country to outside of US, CA, MX, DO (e.g. UK) in WooCommerce settings
- Go back to the order, and try reprinting a shipping label --> A4 paper size appears in the paper size menu
- Reprint with A4 paper size or set the default paper size to A4 in WooCommerce settings > Shipping > WooCommerce Shipping
- Launch the app
- Go to the orders tab, and tap on the same order
- Tap "Reprint Shipping Label" --> the app used to crash in `develop` in debug mode, and now it should not crash and the paper size is default to "Legal"

Update release notes:

### DIFF
--- a/Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaperSize.swift
+++ b/Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaperSize.swift
@@ -2,6 +2,7 @@ import Foundation
 
 /// Paper size options for printing a shipping label.
 public enum ShippingLabelPaperSize {
+    case a4
     case label
     case legal
     case letter
@@ -13,6 +14,8 @@ extension ShippingLabelPaperSize: RawRepresentable {
     ///
     public init(rawValue: String) {
         switch rawValue {
+        case Keys.a4:
+            self = .a4
         case Keys.label:
             self = .label
         case Keys.legal:
@@ -29,6 +32,8 @@ extension ShippingLabelPaperSize: RawRepresentable {
     ///
     public var rawValue: String {
         switch self {
+        case .a4:
+            return Keys.a4
         case .label:
             return Keys.label
         case .legal:
@@ -41,6 +46,7 @@ extension ShippingLabelPaperSize: RawRepresentable {
 
 /// Contains the supported ShippingLabelPaperSize values.
 private enum Keys {
+    static let a4 = "a4"
     static let label = "label"
     static let legal = "legal"
     static let letter = "letter"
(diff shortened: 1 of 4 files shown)


Ecarrion (approved):
Code and tests look good! 🚀

PR #3390, opened 2020-12-21
Edge to edge cells in landscape on Settings / Refunded Products / Fulfill screens

### LINKED ISSUE #2943: ButtonTableFooterView Has Unnecessary Margin in iPhones with Notches on Landscape
The `ButtonTableFooterView` was created from an existing (Fulfill) view in #2938. @Ecarrion found that on iPhones with notches in landscape mode, it has an unnecessary margin.

Portrait ✔️ | Landscape ❌
--------|-------
        |       

## iPad

It seems to work fairly well on iPad.

Portrait ✔️ | Landscape ✔️
--------|-------
        |       

## iPhone 8

On iPhone 8, which does not have a notch, it also works fine.  So we probably only need to fix it for iPhones with notches. 

Portrait ✔️ | Landscape ✔️
--------|-------
        |

eshurakov:
Closes #2943 

### What
In this PR we're making cells go edge to edge in landscape on iPhones with a notch. 
Original issue talks only about Fulfill screen, but I updated a few other simple screens as well.
More screens can be updated to support edge to edge cells, but many requires some input from a designer.

|Before|After|
|-|-|
|Screenshot 2020-12-21 at 11 50 28|Screenshot 2020-12-21 at 11 47 10|
|Screenshot 2020-12-21 at 11 50 00|Screenshot 2020-12-21 at 11 46 55|

### How

Connecting leading and trailing constraints directly to the superview and not a safe area will make cells go edge to edge. Content view of the cells will still follow safe area guidelines in this case.

### Testing (iPhone with a notch and iPhone without a notch)

Check that the cells are edge to edge and content is following the safe area guidelines on the following screens:
* Order -> Order Details -> Refunded Products
* Order -> Order Details -> Fulfill Order
* Settings

Update release notes:


pmusolino (approved):
It looks good to me :shipit: 
Now we just need to apply the same change everywhere (like in the Order detail) 😄

ghost (approved):
This looks great! Thanks @eshurakov

eshurakov:
Thank you for the review @pmusolino and @adamzelinski!

Issue #123, opened 2018-06-27
Order Fulfillment: Undo


jleandroperez:
### Mockup i5:
![undo](https://user-images.githubusercontent.com/1195260/41995337-b9bff550-7a27-11e8-8e77-2d65df5a17e9.jpg)

PR #3391, opened 2020-12-21
Localize fullName in Address

### LINKED ISSUE #2314: Localize Address.fullName and related methods
As mentioned in https://github.com/woocommerce/woocommerce-ios/pull/2298#discussion_r426330490, the strings in [`Address.fullName`](https://github.com/woocommerce/woocommerce-ios/blob/86dcbe87419b0aa2ddffb32d8c1ec91bb8b9fe83/WooCommerce/Classes/Model/Address+Woo.swift#L12-L14) and `fullNameWithCompany` should be localized.

It might be worth it to merge the localization methods that were created in https://github.com/woocommerce/woocommerce-ios/pull/2298.

eshurakov:
Closes #2314 

### What

Display localized name of the customer based on the language rules, phone locale and settings.
I'm switching the code to using `PersonNameComponentsFormatter` instead of `NSLocalizedString` because it takes into account:
* Characters used in the name. For Chinese, Japanese and Korean names there are strict ordering rules that can't be overridden by a selected locale
* Current locale (that's the only similar part to `NSLocalizedString`)
* User specified preferences. Users can override the default sort and display order of given and family names for their current locale.

The Chinese name on the screenshots is `张三` and should always be written in this order. (https://en.wikipedia.org/wiki/List_of_placeholder_names_by_language)

|Before|After|
|-|-|
|Simulator Screen Shot - iPhone 11 - 2020-12-21 at 16 20 46|Simulator Screen Shot - iPhone 11 - 2020-12-21 at 16 19 41|

### Testing

1. Open order list
2. Check that first/last names are in the same order as in contacts app on the phone
3. Open one of the orders
4. Check shipping and billing addresses

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
@@ -127,13 +127,14 @@ private extension OrderTableViewCell {
     /// For example, #560 Pamela Nguyen
     ///
     func title(for order: Order) -> String {
-        if let billingAddress = order.billingAddress, billingAddress.firstName.isNotEmpty || billingAddress.lastName.isNotEmpty {
-            return Localization.title(orderNumber: order.number,
-                                      firstName: billingAddress.firstName,
-                                      lastName: billingAddress.lastName)
-        }
+        let customerName: String = {
+            if let fullName = order.billingAddress?.fullName, fullName.isNotEmpty {
+                return fullName
+            }
+            return Localization.guestName
+        }()
 
-        return Localization.title(orderNumber: order.number)
+        return Localization.title(orderNumber: order.number, customerName: customerName)
     }
 }
 
@@ -163,22 +164,14 @@ private extension OrderTableViewCell {
 
 private extension OrderTableViewCell {
     enum Localization {
-        static func title(orderNumber: String, firstName: String, lastName: String) -> String {
-            let format = NSLocalizedString("#%1$@ %2$@ %3$@", comment: "In Order List,"
-                + " the pattern to show the order number and the full name. For example, “#123 John Doe”."
-                + " The %1$@ is the order number. The %2$@ is the first name. The %3$@ is the last name.")
-
-            return String.localizedStringWithFormat(format, orderNumber, firstName, lastName)
-        }
-
-        static func title(orderNumber: String) -> String {
+        static func title(orderNumber: String, customerName: String) -> String {
             let format = NSLocalizedString("#%@ %@", comment: "In Order List,"
                 + " the pattern to show the order number. For example, “#123456”."
                 + " The %@ placeholder is the order number.")
 
-            let guestName: String = NSLocalizedString("Guest", comment: "In Order List, the name of the billed person when there are no name and last name.")
-
-            return String.localizedStringWithFormat(format, orderNumber, guestName)
+            return String.localizedStringWithFormat(format, orderNumber, customerName)
         }
+
+        static let guestName = NSLocalizedString("Guest", comment: "In Order List, the name of the billed person when there are no name and last name.")
     }
 }
--- a/WooCommerce/Classes/Model/Address+Woo.swift
+++ b/WooCommerce/Classes/Model/Address+Woo.swift
@@ -7,10 +7,14 @@ import Yosemite
 //
 extension Address {
 
-    /// Returns the First + LastName combined.
+    /// Returns the First + LastName combined according to users Locale.
     ///
     var fullName: String {
-        return firstName + " " + lastName
+        var components = PersonNameComponents()
+        components.givenName = firstName
+        components.familyName = lastName
+
+        return PersonNameComponentsFormatter.localizedString(from: components, style: .medium, options: [])
     }
 
     /// Returns the `fullName` and `company` (on a new line). If either the `fullname` or `company` is empty,
--- a/WooCommerce/Classes/Model/Address+Woo.swift
+++ b/WooCommerce/Classes/Model/Address+Woo.swift
@@ -7,7 +7,7 @@ import Yosemite
 //
 extension Address {
 
-    /// Returns the First + LastName combined according to users Locale.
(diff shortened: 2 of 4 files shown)


eshurakov:
While I was looking into this issue, I saw that we also use `NSLocalizedString` to combine order number with a customers name. 
```
let format = NSLocalizedString("#%@ %@", comment: "In Order List,"
                + " the pattern to show the order number. For example, “#123456”."
                + " The %@ placeholder is the order number.")
```

I'm questioning if it is really useful. I think for this specific example the order of the elements is more of a UX decision -  how to structure elements relative to each other, rather than something locale dependent.
Using localised string that mostly consists of placeholders can also result in conflicts if different parts of the app would use the same string for different purposes.

What do you think, @woocommerce/ios-developers?

WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
```diff
@@ -127,13 +127,14 @@ private extension OrderTableViewCell {
...
-                                      lastName: billingAddress.lastName)
-        }
-
-        return Localization.title(orderNumber: order.number)
+        let customerName: String = {
+            if let fullName = order.billingAddress?.fullName, fullName.isNotEmpty {
+                return fullName
+            }
+            return Localization.guestName
+        }()
```
jaclync: nice rewrite 👍

WooCommerce/Classes/ViewRelated/Orders/Cells/OrderTableViewCell.swift
```diff
@@ -163,22 +164,14 @@ private extension OrderTableViewCell {
...
                 + " the pattern to show the order number. For example, “#123456”."
                 + " The %@ placeholder is the order number.")
 
-            let guestName: String = NSLocalizedString("Guest", comment: "In Order List, the name of the billed person when there are no name and last name.")
-
-            return String.localizedStringWithFormat(format, orderNumber, guestName)
+            return String.localizedStringWithFormat(format, orderNumber, customerName)
         }
+
+        static let guestName = NSLocalizedString("Guest", comment: "In Order List, the name of the billed person when there are no name and last name.")
```
jaclync: super nit suggestion:

```suggestion
        static let guestName = NSLocalizedString("Guest", comment: "In Order List, the name of the billed person when there are no first and last name.")
```

jaclync (approved):
Nice work :shipit: TIL'ed `PersonNameComponentsFormatter`, that's awesome it detects the script and supports names of different languages at the same time!

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 10 53 23 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 12 34
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 10 53 26 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 12 40



> While I was looking into this issue, I saw that we also use `NSLocalizedString` to combine order number with a customers name. 
> ```
> let format = NSLocalizedString("#%@ %@", comment: "In Order List,"
>                + " the pattern to show the order number. For example, “#123456”."
>                + " The %@ placeholder is the order number.")
> ```
> I'm questioning if it is really useful. I think for this specific example the order of the elements is more of a UX decision -  how to structure elements relative to each other, rather than something locale dependent.
Using localised string that mostly consists of placeholders can also result in conflicts if different parts of the app would use the same string for different purposes.

I looked into the translations for this string in all `Localizable.strings` files, and noticed the translation is different in Korean from others:

https://github.com/woocommerce/woocommerce-ios/blob/fc7725464bfcbae5c8b7ce7ab3190f4afabf513e/WooCommerce/Resources/ko.lproj/Localizable.strings#L9-L10

I'm not familiar with the language to explain the translation difference (would love to know more if anyone knows Korean 😄), maybe it's worth keeping it as a localizable string just to maintain the existing behavior?

eshurakov:
> I looked into the translations for this string in all Localizable.strings files, and noticed the translation is different in Korean from others. 

Good catch @jaclync. Interesting enough it was only shown when the customer is a guest. For customers with first/last names only a space was used. 
https://github.com/woocommerce/woocommerce-ios/blob/fc7725464bfcbae5c8b7ce7ab3190f4afabf513e/WooCommerce/Resources/ko.lproj/Localizable.strings#L6-L10

Also for Italian "N." is used in one case and "#" in another case:
https://github.com/woocommerce/woocommerce-ios/blob/fc7725464bfcbae5c8b7ce7ab3190f4afabf513e/WooCommerce/Resources/it.lproj/Localizable.strings#L6-L10

That's also why I'm questioning running strings like this through localisation. Translators usually don't have the whole overview and it is a very hard task to know which string is used where, which can result in UX inconsistencies.

I don't think there is any need to do anything in this PR, it is more of an open question :-)

jaclync:
Ohh great catch that the translations in Italian and Korean are different for these two cases, and it's nice that this PR consolidates the two strings to just one! The previous localizable string `"#%1$@ %2$@ %3$@"` could use some comment as well to provide some context for translators 😅 
 
Just curious about the Italian translation, @pmusolino do you know if `N. 123 Customer Name` reads more naturally than `#123 Customer Name` in Italian? (the number is the order ID)

shiki:
> I'm questioning if it is really useful. I think for this specific example the order of the elements is more of a UX decision - how to structure elements relative to each other, rather than something locale dependent.

IIRC, the reason why I did it this way was I wasn't sure if RTL languages would want to change the order so that the order number is shown on the right. 

```
"#%@ %@" → "%@ %@#"
```

But now I see that that's not even possible since it's not using positional placeholders. 🤷 

I'm also not sure if this is really the best way to do that. It's the only way I know how. 🙂

eshurakov:
> IIRC, the reason why I did it this way was I wasn't sure if RTL languages would want to change the order so that the order number is shown on the right.

Indeed, interesting case. I didn't think about that. 
Actually don't really know what is the expected output for "#123 Name" in RTL 🤔

eshurakov:
Thank you for the review. Merging it 🚀

PR #3396, opened 2020-12-22
Update `TopLeftImageTableViewCell` to use Auto Layout

jaclync:
Part of #3334 (Images in "See layout and paper sizes options" and "Don't know how to print from your device?" rows are not the footnote size 20x20 as in design)

## Why

Before this PR, `TopLeftImageTableViewCell` has no subviews and its image and text are configured for the native `imageView` and `textLabel` properties with frame-based layout. This makes it hard to add customizations like changing the top spacing and spacing between image and text. Also, the image size was hard coded to 20x20 before. This PR updated `TopLeftImageTableViewCell` to use a horizontal stack view of an `UIImageView` and `UILabel` with Auto Layout for a few customizations in #3334.

I also looked a bit into dynamic type when the UI has image and text (like [this alignment issue](https://user-images.githubusercontent.com/198826/102104000-e9422580-3dea-11eb-9563-a0f89abbf596.png) with larger font size), and it's not very easy with fixed size images like how we use Gridicons. Ideally, we can start using [symbols](https://developer.apple.com/design/human-interface-guidelines/sf-symbols/overview/) for images that are shown alongside text (there are some [system symbols](https://developer.apple.com/sf-symbols/) that we could probably start using).

## Changes

- Updated `TopLeftImageTableViewCell` to use Auto Layout
- Updated existing usage to use the `configure` method to configure its image and text (it used to just set the cell's `imageView` and `textLabel`)
- Updated `PrivacySettingsViewController`'s `TopLeftImageTableViewCell` usage to `BasicTableViewCell` since it just uses the cell's `imageView` to create spacing

## Testing

Feel free to focus on just a subset of affected screens, I shared the before/after screenshots for all screens below:

- My store tab -> settings > Privacy Settings --> the multiline info text cells should look as before
- Orders tab -> order with multiple shipping methods --> the [shipping notice](https://github.com/woocommerce/woocommerce-ios/pull/1434) cell should look good
- Orders tab -> order with a refund that has a refund reason -> refund details --> the refund reason cell should look good
- Orders tab -> order with a non-refunded shipping label --> the "Don't know how to print from your phone?" cell should look good
- Orders tab -> order with a non-refunded shipping label -> "Reprint Shipping Label" --> "See layout and paper sizes options" and "Don't know how to print from your device?" cells now should have footnote font size and 20x20px icon, while the info cell above has body font size and 24x24px icon
-  Orders tab -> order with a customer note -> "Begin Fulfillment" --> the customer note cell should look good

## Example screenshots

Notes:
- The cells with an image at the top left now have 16px spacing on all 4 edges and between the image and text. The top spacing was 11px before. Using Auto Layout allows us to customize the spacing in the future
- The comparison screenshots below might show slight different in font size, because the text is now using `body` font that adapts to dynamic type

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 54 59 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 15 13 50
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 55 40 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 15 14 20
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 55 50 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 15 14 42
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 56 40 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 15 15 15
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 56 48 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 15 15 47
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 11 58 03 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 15 16 08 

Update release notes:


Garance91540 (approved):
👏👏👏

Ecarrion (approved):
Looks good to me! 👏 

FYI I've handled dynamic size of images like this.

https://github.com/woocommerce/woocommerce-ios/blob/bfb41b2a01398a8758b9592ec5ef1e0315969e6b/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Issue%20Refunds/Cells/RefundItemTableViewCell.swift#L85-L102

jaclync:
> FYI I've handled dynamic size of images like this.

Nice thanks for the reference! I'll do the same for the accessibility improvement for #3334 👍 I also plan to look into how we can use the symbols starting iOS 13

PR #3398, opened 2020-12-22
Shipping Label Details > Ship To: update "House" icon to outlined house icon 

jaclync:
Part of #3334 (Update "House" icon to outlined house icon)

## Why

Adding a new icon to [Gridicons](https://github.com/Automattic/gridicons) is a big undertaking (requires review from Gridicons maintainers) (ref CGPNUU63E/p1607084416034700-slack), so this PR just added the house PDF from design.

## Changes

- Exported `icon-house-outlined` PDF from design (I updated the stroke color to black in a copied design file, otherwise the color is slightly off)
- Added `UIImage.houseOutlinedImage` with a test case
- In `ShippingLabelDetailsViewController`, updated "Ship To" row to use `houseOutlinedImage` image

## Testing

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/) and there is an order with at least one shipping label.

- Go to the orders tab
- Tap on an order with at least one shipping label
- After a shipping label package card appears, tap "View Shipment Details" --> the "Ship to" row should show the house outlined icon instead of solid icon

## Example screenshots

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 18 37 39 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 18 53 16
Simulator Screen Shot - iPhone 11 - 2020-12-22 at 18 37 51 | Simulator Screen Shot - iPhone 11 - 2020-12-22 at 18 53 09

Update release notes:


WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -237,6 +237,12 @@ extension UIImage {
         UIImage.gridicon(.house)
     }
 
+    /// House Outlined Image
+    ///
+    static var houseOutlinedImage: UIImage {
+        UIImage(named: "icon-house-outlined")!
```
Ecarrion: nit: you can use `UIImage(imageLiteralResourceName: "woo-no-store")` which doesn't force you to use the `!` although it will probably crash in the same way if the resource doesn't exist! 🤷

Ecarrion (approved):
👏 
One question tho, do we plan to add this icon on `GridIcons` eventually?

ghost (approved):
👏👏

WooCommerce/Classes/Extensions/UIImage+Woo.swift
jaclync: ohh I do prefer not having `!` in the code! updated in https://github.com/woocommerce/woocommerce-ios/pull/3398/commits/7196589609c05d515ae512d5ea827f836890e2f1

jaclync:
> One question tho, do we plan to add this icon on GridIcons eventually?

From my understanding in the [GitHub README](https://github.com/Automattic/gridicons#propose-a-new-icon), it's quite a lengthy process and involves review from other designers. After it's merged to Gridicons, there is another step to add the new icon to Gridicons-iOS repo, which is another non-trivial process (I did it once) 😅.

Whether this effort is worthwhile probably depends on if other teams might benefit from this icon - @woocommerce/mobile-designers do you know if other projects/apps might be using this house outlined icon?

PR #3403, opened 2020-12-22
Rename SettingTitleAndValueTableViewCell -> TitleAndValueTableViewCell

### LINKED ISSUE #2211: Rename `SettingTitleAndValueTableViewCell` so it doesn't sound like it is meant for settings

eshurakov:
Closes #2211

#### What

As suggested in the issue renaming `SettingTitleAndValueTableViewCell` to `TitleAndValueTableViewCell`

#### Testing 

`TitleAndValueTableViewCell` is used on the following screens, check some of them to see if the cell works:
* Filter list on Search
* Refund confirmation
* Reprint shipping label
* Product price settings
* Add product category

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Editor/FormatBar/LinkSettings/LinkSettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Editor/FormatBar/LinkSettings/LinkSettingsViewController.swift
@@ -216,9 +216,9 @@ private extension LinkSettingsViewController {
     ///
     func configure(_ cell: UITableViewCell, for row: Row, at indexPath: IndexPath) {
         switch cell {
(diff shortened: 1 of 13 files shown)


Ecarrion (approved):
Thanks for this!

eshurakov:
Thank you for the review @Ecarrion!

PR #3404, opened 2020-12-22
Remove interpolated strings from localization

### LINKED ISSUE #1732: Interpolated strings found in localized strings
**Describe the bug**
There are approximately 13 interpolated strings in our `Localizable.strings` files. Interpolated strings (`This sentence has a \(variableName)` ) are harder to understand by translators and they may end up translating/changing the variable name, which can cause a crash. Use `%@`, `%d`, and other acceptable placeholders for localized strings.

Example, NOK:
```
let year = 2019
let str = NSLocalizedString("© \(year) Automattic, Inc.", comment: "Copyright Notice")
```

**To Reproduce**
1. Go to the `Localizable.strings` file (in English)
2. Search for `\(` to find the beginning of an interpolated string
3. Copy the first part of the localized string to track it back to the location where the localization is happening in the code.

**Screenshots**
n/a

**Expected behavior**
We should create a template instead.

Example:
```
let template = NSLocalizedString("© %d Automattic, Inc.", comment: "Copyright Notice")
let str = String.localizeStringWithFormat(template, year)
```

**Mobile Environment**
Please include:
- Device: any
- iOS version: any 
- WooCommerce iOS version: since v1.0

eshurakov:
Closes #1732 

### What

This PR replaces interpolated strings found in localization files with placeholders.
For example, before:
```
NSLocalizedString("Dotcom Error: [\(code)] \(theMessage)", comment: "WordPress.com (unmapped!) error")
```

After:
```
let messageFormat = NSLocalizedString(
    "Dotcom Error: [%1$@] %2$@",
    comment: "WordPress.com (unmapped!) error. Parameters: %1$@ - code, %2$@ - message"
)
String.localizedStringWithFormat(messageFormat, code, theMessage)
```

### Testing

Not sure what is the best way to test this as it touches different scenarios that can be hard to simulate. Any ideas?

Update release notes:

### DIFF
--- a/Networking/Networking/Model/DotcomError.swift
+++ b/Networking/Networking/Model/DotcomError.swift
@@ -113,7 +113,11 @@ extension DotcomError: CustomStringConvertible {
             return NSLocalizedString("Dotcom Stats Module Disabled", comment: "WordPress.com error thrown when the Jetpack site stats module is disabled.")
         case .unknown(let code, let message):
             let theMessage = message ?? String()
-            return NSLocalizedString("Dotcom Error: [\(code)] \(theMessage)", comment: "WordPress.com (unmapped!) error")
+            let messageFormat = NSLocalizedString(
+                "Dotcom Error: [%1$@] %2$@",
+                comment: "WordPress.com (unmapped!) error. Parameters: %1$@ - code, %2$@ - message"
+            )
+            return String.localizedStringWithFormat(messageFormat, code, theMessage)
         }
     }
 }
--- a/WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
+++ b/WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
@@ -24,9 +24,11 @@ final class SwitchStoreNoticePresenter {
             return
         }
 
-        let message = NSLocalizedString("Switched to \(newStoreName). You will only receive notifications from this store.",
-            comment: "Message presented after users switch to a new store. "
-                + "Reads like: Switched to {store name}. You will only receive notifications from this store.")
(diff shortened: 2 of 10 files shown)


ctarda (approved):
I run this on the Spanish simulator, and:
* The strings are displayed correctly
* The strings are not localized

It makes sense that those strings are not localized, as those are not in the Localizable.strings file. 







As far as I understand, the extraction of those strings is part of the release process, but I am wondering if that's something that, in this case, we should update manually as well.

@woocommerce/ios-developers does anyone one if the Localizable string files should be updated manually in this case?

Anyway, pending that, it's a :shipit: from me.

jaclync:
> As far as I understand, the extraction of those strings is part of the release process, but I am wondering if that's something that, in this case, we should update manually as well.

I believe the Localizable.strings files are updated by platform9 team via a script (possibly [this `localize.py`](https://github.com/woocommerce/woocommerce-ios/blob/95f48633ed3cb1d0d15b30e14b89742e4e9ac18b/Scripts/localize.py)) during the release process, like how [most of the commits in the English strings file](https://github.com/woocommerce/woocommerce-ios/commits/0af9af5e38b39763bae66df25d829c46c76c0025/WooCommerce/Resources/en.lproj/Localizable.strings) are by platform9. I'd think these changes in the code are picked up by the script, then all the strings files are updated accordingly

eshurakov:
> As far as I understand, the extraction of those strings is part of the release process, but I am wondering if that's something that, in this case, we should update manually as well.

I believe as @jaclync said strings from the code will be automatically picked up during code freeze / release. 

On the other hand I thought that it would be nice if we could keep existing translations for renamed localization keys without the need to localize them again. Not sure if there is an automated process for that. Doing it manually for all languages would be crazy :-)

PR #3407, opened 2020-12-23
Use system margins for cell content

### LINKED ISSUE #1933: Edit Product: update left alignment for table view cells
Some table view cells and headings aren't perfectly aligned with each other, and is noticeable in a few Edit Product screens:

screen | screenshot | issues
-- | -- | --
Edit Product | Simulator Screen Shot - iPhone 11 - 2020-03-11 at 07 31 28 | the empty state cell's text isn't left aligned
inventory settings | Simulator Screen Shot - iPhone 11 - 2020-03-11 at 07 31 38 | the cells with a toggle ("Manage stock" and "Limit one per order") aren't left aligned with the rest of the cells
price settings | Simulator Screen Shot - iPhone 11 - 2020-03-11 at 07 31 55 | the cells aren't left aligned with the section heading label (e.g. "TAX SETTINGS"), and the "Schedule date" cell isn't left aligned with the rest

It seems like the left alignment in the toggle cells is correct, since it aligns with the default section header and cell left inset.

eshurakov:
Closes #1933 

### Why

Quite some cells in the app are using using system provided margins, which are usually 16 or 20 depending on the device. Other cells are using fixed margins defined in the code and they are not always 16 or 20.

Using system provided margins would allow us to provide better user experience and have components that align nicely with each other.

Should I split it in smaller pieces, what do you think? 🤔 

### Changes
* I walked through most of the cells and updated leading and trailing margins to use `layoutMarginsGuide`
* Some cells had constraints from subviews of stack views to a superview. Because stack view fully manages its subviews, I think it is a mistake to have additional constraints for subviews. Those are removed.
* I didn't see that @jaclync updated `TopLeftImageTableViewCell` to auto layout. I thought about doing that too, but also noticed that it is very similar to `ImageAndTitleAndTextTableViewCell` and that it would be nice to use less different types of cells. I replaced `TopLeftImageTableViewCell` usage on Order Details screen with `ImageAndTitleAndTextTableViewCell`. 
* Placeholder in `EnhancedTextView` is now using auto layout and positioned using text container insets. It fixes an issue with placeholder not showing at initial cursor position.

### Testing

Please check some of the following screens:
* order tab -> order details
* order tab -> order details -> billing information
* order tab -> order details -> add note 
* order tab -> order details -> issue refund
* products tab -> product details
* products tab -> product details -> price settings
* products tab -> product details -> inventory
* products tab -> product details -> settings -> visibility
* products tab -> new product
* reviews

### Examples

It's like a game, spot a few differences :-)

|Before|After|
|-|-|
|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 14 15|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 19 55 16|
|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 14 22|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 29 00|
|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 16 24|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 31 03|
|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 17 27|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 31 51|
|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 16 52|Simulator Screen Shot - iPhone 12 Pro Max - 2020-12-23 at 14 31 28|

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
```diff
@@ -53,13 +53,24 @@ final class ImageAndTitleAndTextTableViewCell: UITableViewCell {
         let title: String?
     }
 
+    /// View model to replace TopLeftImageTableViewCell
+    struct TopLeftImageViewModel {
+        let icon: UIImage
+        let iconColor: UIColor?
+        let title: String
+        let isFootnoteStyle: Bool
```
jaclync, line 61: nit: how about an enum like `Style` in `TopLeftImageTableViewCell` so that we can add more styles than the current usage `body` and `footnote` in the future?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
```diff
@@ -129,6 +142,24 @@ extension ImageAndTitleAndTextTableViewCell {
...
+                                  isActionable: false)
+        updateUI(viewModel: viewModel)
+
+        if topLeftImageViewModel.isFootnoteStyle {
+            titleLabel.applyFootnoteStyle()
+        } else {
+            titleLabel.applyBodyStyle()
+        }
+
+        contentImageViewWidthConstraint.isActive = true
```
jaclync, line 161: Just wondering, why not leave the image size as its `UIImage`'s content size? For example, when we are replacing `TopLeftImageTableViewCell` usage in `ReprintShippingLabelViewController`, the images of the info links (with footnote style) are smaller than the image of the info row at the top 🙂

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -330,13 +330,17 @@ private extension OrderDetailsDataSource {
         )
     }
 
-    private func configureShippingNotice(cell: TopLeftImageTableViewCell) {
+    private func configureShippingNotice(cell: ImageAndTitleAndTextTableViewCell) {
```
jaclync, line 333: Good idea replacing `TopLeftImageTableViewCell` with `ImageAndTitleAndTextTableViewCell`! I created an issue https://github.com/woocommerce/woocommerce-ios/issues/3408 to replace `TopLeftImageTableViewCell` with `ImageAndTitleAndTextTableViewCell` 😄

WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.xib
```diff
@@ -63,19 +64,19 @@
...
+                        <color key="backgroundColor" systemColor="separatorColor"/>
                         <constraints>
                             <constraint firstAttribute="height" constant="0.5" id="XW4-X6-mV7"/>
                         </constraints>
-                        <viewLayoutGuide key="safeArea" id="dER-lG-WQT"/>
                     </view>
                 </subviews>
                 <constraints>
                     <constraint firstAttribute="trailing" secondItem="lOL-Ke-vhW" secondAttribute="trailing" id="5MW-g7-JV0"/>
                     <constraint firstAttribute="trailing" secondItem="VJh-YT-pnW" secondAttribute="trailing" constant="16" id="EuK-Sc-fLf"/>
```
jaclync, line 76: do you think the leading/trailing constraint here could pin to superview's layout margin like other cells as well?

WooCommerce/Classes/ViewRelated/ReusableViews/TopLeftImageTableViewCell.xib
```diff
@@ -33,10 +33,10 @@
                     </stackView>
                 </subviews>
                 <constraints>
-                    <constraint firstAttribute="trailing" secondItem="8cW-Vq-hwR" secondAttribute="trailing" constant="16" id="JUM-av-6A1"/>
+                    <constraint firstAttribute="trailingMargin" secondItem="8cW-Vq-hwR" secondAttribute="trailing" id="JUM-av-6A1"/>
                     <constraint firstAttribute="bottom" secondItem="8cW-Vq-hwR" secondAttribute="bottom" constant="16" id="aq3-TT-4SM"/>
                     <constraint firstItem="8cW-Vq-hwR" firstAttribute="top" secondItem="H2p-sc-9uM" secondAttribute="top" constant="16" id="fwR-dS-jWv"/>
-                    <constraint firstItem="8cW-Vq-hwR" firstAttribute="leading" secondItem="H2p-sc-9uM" secondAttribute="leading" constant="16" id="xLN-q2-SMf"/>
+                    <constraint firstItem="8cW-Vq-hwR" firstAttribute="leading" secondItem="H2p-sc-9uM" secondAttribute="leadingMargin" id="xLN-q2-SMf"/>
```
jaclync, line 39: thanks for updating this one too!

WooCommerce/Classes/ViewRelated/ReusableViews/EnhancedTextView.swift
```diff
@@ -57,14 +57,24 @@ final class EnhancedTextView: UITextView {
...
+        placeholderLabel = {
+            let label = UILabel(frame: bounds)
+            label.translatesAutoresizingMaskIntoConstraints = false
+            addSubview(label)
+
+            // Make placeholder left/right margins same as for the text container
+            NSLayoutConstraint.activate([
+                label.leadingAnchor.constraint(equalTo: leadingAnchor, constant: textContainer.lineFragmentPadding + textContainerInset.left),
+                label.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -(textContainer.lineFragmentPadding + textContainerInset.right)),
+                label.topAnchor.constraint(equalTo: topAnchor, constant: Constants.margin)
```
jaclync: just curious why bottom constraint isn't needed here?

WooCommerce/Classes/ViewRelated/ReusableViews/ButtonTableFooterView.swift
```diff
@@ -37,24 +37,36 @@ final class ButtonTableFooterView: UIView {
...
-                                                   left: Dimensions.buttonDefaultInset,
-                                                   bottom: Dimensions.buttonBottomInset,
-                                                   right: Dimensions.buttonDefaultInset))
+
+        configureButtonConstraints()
+    }
+
+    private func configureButtonConstraints() {
+        let trailingConstraint = layoutMarginsGuide.trailingAnchor.constraint(equalTo: button.trailingAnchor)
+        trailingConstraint.priority = .defaultHigh
```
jaclync, line 54: just curious, why does the trailing constraint have a different setup on its priority?

jaclync (approved):
Nice work! it's so pleasant seeing the leading/trailing edges aligned beautifully 😍 just out of curiosity, do you know when the leading/trailing margin becomes 20px? I tried on iPad simulator (iPad Air 3rd gen iOS 14.1) but it's still 16px

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
eshurakov: I thought to keep it simple for now and add more styles as we need them.

Another idea that just came to me is to define styles for this cell in a way similar to the standard UITableViewCell. We can have styles that are global for the cell vs styles that are attached to a particular view model.
For example, for this cell we can have the following styles: `default`, `switch`, `warning`, `footnote`, 'topLeftImage' (can do a better work on naming those :-) )

What do you think?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
eshurakov: My motivation here was to align the text that is following after the image. That's why I fixed the size of the image view. The content mode for the image view was already set to `center`, so if the image is smaller, it will be the right size and centered within the image view.
For example, disabling that constraint produces the following result on the order screen:
Screenshot 2020-12-29 at 09 21 21

I couldn't think of a generic way to fix this without fixing the image size (or maybe text left margin).

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
eshurakov: Awesome, thank you! 👍

WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.xib
eshurakov: I tried to use system margins on a dashboard screen and learned that it is a more complex task and decided left it out of this PR. Because dashboard is created from a different components added to a scroll view and another tab / segment bar it is needed to do more work to propagate margins to all the views.
I've created and issue to fix it later: #3413

WooCommerce/Classes/ViewRelated/ReusableViews/EnhancedTextView.swift
eshurakov: If the text view height is bigger than one line and placeholder label is pinned to all edges, placeholder label would vertically centre the text and we want it to be top aligned.
I updated constraints to use width / heigh of the text view in https://github.com/woocommerce/woocommerce-ios/pull/3407/commits/8d021637ff4dff2e6a92e224f1ce0826404e9840 as I figured that using trailing constraint doesn't work well because text view is basically a scroll view.

WooCommerce/Classes/ViewRelated/ReusableViews/ButtonTableFooterView.swift
eshurakov: I saw that constraints were breaking while going from screen to screen in Issue Refund flow and that fixed it.

eshurakov:
> just out of curiosity, do you know when the leading/trailing margin becomes 20px? I tried on iPad simulator (iPad Air 3rd gen iOS 14.1) but it's still 16px

That's what I thought, but apparently it's not the case anymore. TIL :-)

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
jaclync: Got it, thanks for the explanation! Yea, it's tricky to align images with different width. Let's go with fixed width here, and I'll see how it looks after replacing `TopLeftImageTableViewCell` with `ImageAndTitleAndTextTableViewCell` later 👍

WooCommerce/Classes/ViewRelated/ReusableViews/ButtonTableFooterView.swift
jaclync: ooh nice, thanks for noticing the warnings! I need to pay attention to them more, now the console is flooded with messages from aztec 😅

WooCommerce/Classes/ViewRelated/Dashboard/MyStore/Cells/ProductTableViewCell.xib
jaclync: ah interesting, Dashboard is a bit different from other use cases. we'll handle it separately in the issue then 👍

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
jaclync: sounds like a good idea! the current approach with different view models is a bit odd. if I understand correctly, the extra parameters in each view model could become associated values in each enum case like:

```swift
enum ImageAndTitleAndTextTableViewCell.Style {
  case default
  case switch(isSwitchOn: Bool, onSwitchChange: @escaping (_ isOn: Bool) -> Void)
  case warning
  case imageAndTitleOnly
}

extension ImageAndTitleAndTextTableViewCell {
  func updateUI(style: Style, title: String?, text: String?, textTintColor: UIColor? = nil, ...)
}
```

to replace these different view models: `ViewModel`, `SwitchableViewModel`, `WarningViewModel`, `TopLeftImageViewModel`

lemme know if this is different from what you thought!

jaclync (approved):
Still LGTM, feel free to merge after design review!

> That's what I thought, but apparently it's not the case anymore

It's possible the margin is different from 16px in certain tablet models 😄

Garance91540 (approved):
👏

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
eshurakov: This is very similar to what I thought! I created an issue to keep track of it and so that it can be discussed there (https://github.com/woocommerce/woocommerce-ios/issues/3419)

Thank you!

eshurakov:
Thank you @jaclync and @Garance91540!

PR #3409, opened 2020-12-24
Shipping Labels M1: add feature announcement banner to order details with non-refunded shipping labels

### LINKED ISSUE #2995: Shipping Labels M1: add feature announcement banner in Order Detail
As part of the Shipping Labels M1 release, we need to add the following dismissible feature banner in the Order detail screen, when we are ready to launch the Shipping Labels feature.

**This will only be displayed in there are shipping labels available for the order.**

jaclync:
Fixes #2995 

## Why

If an order has at least one non-refunded shipping label, we show a top banner that informs the user about the shipping labels M1 features.

## Changes

- Added a new `FeedbackType` enum case for shipping labels M1 (I kept the name `*Release1` to be consistent with `FeatureFlag.shippingLabelsRelease1` even though we call it M1). Also added related changes for feedback survey in https://github.com/woocommerce/woocommerce-ios/commit/61c15c323ae4f2731dc6e0599185f22f7c8a0f20 and https://github.com/woocommerce/woocommerce-ios/commit/61e646e6d72312c4bd8c3b5ecd6d86e9f508b4ec
- Created `ShippingLabelsTopBannerFactory` to determine whether to return a top banner based on shipping labels and app settings (whether the feedback card has been dismissed similar to the products tab) + unit tests https://github.com/woocommerce/woocommerce-ios/commit/ce5acad714bc46b64addaea001a1ec308a961dd4
- Integration with order details in https://github.com/woocommerce/woocommerce-ios/commit/9585ae23227187d61ce8dc1b57cf39f414916e95

## Testing

Prerequisite: the shipping labels top banner has not been dismissed in the app. If it has, you can reinstall the app to clear the app settings.

### Order without shipping labels

- Go to the orders tab
- Tap on an order without any shipping labels --> there should **not** be a shipping labels top banner

### Order with refunded shipping label(s) only

- Go to the orders tab
- Tap on an order where all the shipping label(s) have been refunded --> there should **not** be a shipping labels top banner

### Order with non-refunded shipping labels

- Go to the orders tab
- Tap on an order with at least one non-refunded shipping label --> there should be a shipping labels top banner after syncing (the shipping label card should appear as well)
- Tap "Give feedback" --> a feedback survey should be presented
- Type something in the survey and submit the survey if you like 😄 
- Tap "Back to store" on the submission confirmation screen
- Tap "Dismiss" --> the top banner should be hidden
- Go back to the orders tab and tap on the same order again --> the top banner should not be shown anymore

## Example screenshots

simulator

light | dark
-- | --
Simulator Screen Shot - iPhone 11 - 2020-12-24 at 14 52 42 | Simulator Screen Shot - iPhone 11 - 2020-12-24 at 14 53 25
Simulator Screen Shot - iPhone 11 - 2020-12-24 at 14 52 44 | Simulator Screen Shot - iPhone 11 - 2020-12-24 at 14 53 27

landscape (expanded) | landscape (collapsed) | large font size | iPad
-- | -- | -- | --
Simulator Screen Shot - iPhone 11 - 2020-12-24 at 14 52 59 | Simulator Screen Shot - iPhone 11 - 2020-12-24 at 14 53 04 | Simulator Screen Shot - iPhone 11 - 2020-12-24 at 14 53 55 | Simulator Screen Shot - iPad Air (3rd generation) - 2020-12-24 at 14 56 17

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -83,6 +83,10 @@ extension WooConstants {
         ///
         case productsM4Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-products"
 
+        /// URL for the shipping labels M1 feedback survey
+        ///
+        case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-shipping-labels"
+
         /// Returns the URL version of the receiver
         ///
         func asURL() -> URL {
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -85,7 +85,11 @@ extension WooConstants {
 
         /// URL for the shipping labels M1 feedback survey
         ///
+#if DEBUG
+        case shippingLabelsRelease1Feedback = "http://automattic.survey.fm/woo-app-testing-feature-feedback-shipping-labels"
+#else
         case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-shipping-labels"
+#endif
 
         /// Returns the URL version of the receiver
         ///
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -86,7 +86,7 @@ extension WooConstants {
         /// URL for the shipping labels M1 feedback survey
         ///
 #if DEBUG
-        case shippingLabelsRelease1Feedback = "http://automattic.survey.fm/woo-app-testing-feature-feedback-shipping-labels"
+        case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-testing-feature-feedback-shipping-labels"
 #else
         case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-shipping-labels"
 #endif
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
@@ -234,6 +234,8 @@ private extension OrderDetailsViewController {
         }
 
         self.topBannerView = topBannerView
+        // A frame-based container view is needed for table view's `tableHeaderView` and its height is recalculated in `viewDidLayoutSubviews`, so that the
+        // top banner view can be Auto Layout based with dynamic height.
         let headerContainer = UIView(frame: CGRect(x: 0, y: 0, width: Int(tableView.frame.width), height: Int(Constants.headerDefaultHeight)))
         headerContainer.addSubview(topBannerView)
         headerContainer.pinSubviewToSafeArea(topBannerView, insets: Constants.headerContainerInsets)
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
@@ -11,7 +11,7 @@ final class OrderDetailsViewController: UIViewController {
 
     /// Main TableView.
     ///
-    @IBOutlet weak var tableView: UITableView!
+    @IBOutlet private weak var tableView: UITableView!
 
     /// Pull To Refresh Support.
     ///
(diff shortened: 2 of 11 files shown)


WooCommerce/Classes/System/WooConstants.swift
```diff
@@ -83,6 +83,10 @@ extension WooConstants {
         ///
         case productsM4Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-products"
 
+        /// URL for the shipping labels M1 feedback survey
+        ///
+        case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-shipping-labels"
```
Ecarrion, line 91: [Q]: Should we have a testing feedback so we don't pollute our feedback results or is it not needed?

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -194,6 +206,62 @@ private extension OrderDetailsViewController {
...
+    func showTopBannerView(_ topBannerView: TopBannerView) {
+        guard tableView.tableHeaderView == nil else {
+            return
+        }
+
+        self.topBannerView = topBannerView
+        let headerContainer = UIView(frame: CGRect(x: 0, y: 0, width: Int(tableView.frame.width), height: Int(Constants.headerDefaultHeight)))
+        headerContainer.addSubview(topBannerView)
+        headerContainer.pinSubviewToSafeArea(topBannerView, insets: Constants.headerContainerInsets)
+        tableView.tableHeaderView = headerContainer
```
Ecarrion, line 242: Is there a reason on why we can't set the top banner directly as the `tableHeaderView`? If so, it would be good to add some comments explaining it!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelsTopBannerFactory.swift
```diff
@@ -0,0 +1,113 @@
...
+}
+
+private extension ShippingLabelsTopBannerFactory {
+    func determineIfTopBannerShouldBeShown(onCompletion: @escaping (_ shouldShow: Bool) -> Void) {
+        guard shippingLabels.nonRefunded.isNotEmpty else {
+            onCompletion(false)
+            return
+        }
+
+        let action = AppSettingsAction.loadFeedbackVisibility(type: .shippingLabelsRelease1) { result in
```
Ecarrion, line 54: Not related to this PR, but having this kind of actions synchronous would make our lives much easier, don't you think? 😅

Ecarrion (approved):
Testing 🚀 
Code: 🚀 

Thanks for this @jaclync!

Garance91540 (approved):
👏

WooCommerce/Classes/System/WooConstants.swift
jaclync: that's a good idea! I deleted my responses from testing, and it was somewhat scary deleting responses from a survey in production. updated in https://github.com/woocommerce/woocommerce-ios/pull/3409/commits/e0de3bebdad95e39d7c9d17f56dbdf5772f804bc and https://github.com/woocommerce/woocommerce-ios/pull/3409/commits/9f83249f296308095ab3c29e0214317a19797ca8

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelsTopBannerFactory.swift
jaclync: Yes I agree 😓 if the operation has to be async, it'd be nice to have something like `async/await` like in Javascript.

after some quick search, it might actually be available in Swift in the future 😮 https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
jaclync: If we don't use a frame-based container, the Auto Layout based top banner can't be adjusted to fit width with dynamic height. For example, if I replaced this line with `tableView.tableHeaderView = topBannerView` it looks like:



Lemme know if you know of other ways to have an Auto Layout based table header view!

I added a comment about it in: https://github.com/woocommerce/woocommerce-ios/pull/3409/commits/7801ca4f65458cc08ec93fbc76eaa241414e0092

jaclync:
Thank you for the suggestions & testing @Ecarrion, responded to your comments!

Ecarrion (approved):
👏

Issue #3112, opened 2020-11-09
Products: Loading skeleton is incomplete on first load


rachelmcr:
**Describe the bug**

As @ThomazFB reported (internal ref: p5T066-1Bt-p2#comment-6253):

When we first open the app and go directly to the Products section the loading skeleton shows up incomplete, only the image part is glowing, meanwhile, the rest is blank.

![image](https://user-images.githubusercontent.com/8658164/98566154-591e3700-22a6-11eb-895e-65651575fff9.png)

If we call the reloading pulling the screen down then we can see the correct loading skeleton showing up

![image](https://user-images.githubusercontent.com/8658164/98566183-60dddb80-22a6-11eb-908a-cdc0e35df41e.png)

Reported in WCiOS 5.4 beta testing.

pmusolino:
I investigated this issue. it seems that this is caused by the implementation of `ProductsTabProductTableViewCell`, which compact the labels inside the cell if there is no content. So, one solution, but not ideal, would be to use a fake table view cell with the appearance of `ProductsTabProductTableViewCell`. At the same time, if we update this cell, we should remember to update also the fake one.

joshheald:
I can no longer reproduce this, it was likely fixed in the PRs relating to #2448

PR #3417, opened 2020-12-30
[Variations] Adds ProductAttributeTerms Entity

Ecarrion:
part of #3112 

# Why 

In order to support the screen where the store manager can add a new term to a product attribute, we need first to introduce the necessary API entity that will allow us to fetch and create new terms.

Screen Shot 2020-12-30 at 11 03 37 AM

# How

- Added `ProductAttributeTerm` entity and it's mappers.
- Added `product-attribute-terms.json` to support tests.

# Testing
- Nothing yet, just make sure the unit tests pass.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/ProductAttributeTerm.swift
+++ b/Networking/Networking/Model/Product/ProductAttributeTerm.swift
@@ -1,0 +1,51 @@
+import Foundation
+
+/// Represents a `ProductAttributeTerm` entity.
+///
+public struct ProductAttributeTerm: Equatable {
+    public let siteID: Int64
+    public let termID: Int64
+    public let name: String
+    public let slug: String
+    public let count: Int
+
+    /// Member wise initializer.
+    ///
+    public init(siteID: Int64, termID: Int64, name: String, slug: String, count: Int) {
+        self.siteID = siteID
+        self.termID = termID
+        self.name = name
+        self.slug = slug
+        self.count = count
+    }
+}
+
+// MARK: Codable Conformance
+extension ProductAttributeTerm: Decodable {
+    enum CodingKeys: String, CodingKey {
+        case termID = "id"
+        case name
+        case slug
+        case count
+    }
+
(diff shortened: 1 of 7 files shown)


Networking/NetworkingTests/Mapper/ProductAttributeTermListMapperTests.swift
```diff
@@ -0,0 +1,35 @@
...
+    func test_productAttributeTerm_fields_are_correctly_mapped() throws {
+        let terms = try mapLoadAllProductAttributeTermsResponse()
+        XCTAssertEqual(terms.count, 3)
+
+        let secondTerm = terms[1]
+        XCTAssertEqual(secondTerm.siteID, dummySiteID)
+        XCTAssertEqual(secondTerm.termID, 27)
+        XCTAssertEqual(secondTerm.name, "Medium")
+        XCTAssertEqual(secondTerm.slug, "medium")
+        XCTAssertEqual(secondTerm.count, 1)
```
jaclync: nit: does it make sense to assert on the whole term using the default `Equatable` implementation, since this entity doesn't seem overly large (some entities like `Product` or `Order` has tens of attributes)?

Networking/Networking/Model/Product/ProductAttributeTerm.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    /// Member wise initializer.
+    ///
+    public init(siteID: Int64, termID: Int64, name: String, slug: String, count: Int) {
+        self.siteID = siteID
+        self.termID = termID
+        self.name = name
+        self.slug = slug
+        self.count = count
+    }
```
jaclync, line 20: nit: I think this initializer isn't needed? (there is a default one that is the same as this one)

jaclync (approved):
LGTM! :shipit:

Ecarrion:
Thanks a lot @jaclync for the review!

PR #3418, opened 2020-12-30
[Variations] Adds ProductAttributeTerms Storage class

Ecarrion:
part of #3112
depends on #3417 

# Why

In order to support the screen where the store manager can add a new term to a product attribute, we need first to introduce the necessary storage class that will allow us to store and link product attributes and terms.

variations-flow

# How
- Added `ProductAttributeTerm` core data classes
- Added relationship between `ProductAttribute` and `ProductAttributeTerm`

# Testing steps
- Make sure unit tests passes

Update release notes:


ctarda (approved):
Tests pass and I don't see any coredata warnings / errors in the console. 

:shipit:

Ecarrion:
Thanks for checking @ctarda

PR #3427, opened 2021-01-05
 [UL&S] When logging with site address, redirect users to enter the wordpress.com email address

### LINKED ISSUE #3426: [UL&S] When logging with site address, redirect users to enter the wordpress.com email address
For context: p91TBi-44z-p2

Currently, there’s a bug for self-hosted sites: after entering your site address, you are sent to the site credentials screen, but you should be send to the Login with Email Screen. (pbMoDN-1KW#comment-3541-p2)

ctarda:
Closes #3426 

Before:

https://user-images.githubusercontent.com/2722505/103692828-59fddd00-4f66-11eb-9322-d6a23c29f415.mov

After:

https://user-images.githubusercontent.com/2722505/103692843-61bd8180-4f66-11eb-8443-73baa1740695.mov

As usual, most of the solution is implemented in WPAuthenticator. The corresponding PR is wordpress-mobile/WordPressAuthenticator-iOS#555

## Changes
* Point WPAuthenticator to the branch that implements the logic to navigate to the expected view
* Update the logic in AuthenticationManager

## How to test
* Attempt to login with a self-hosted site (i.e. a site created with jurassic ninja)
* Make sure the new site is connected to a valid Jetpack account
* Log out if necessary. 
* Select "Enter your Store Address". Enter the address of the new site.
* Notice if the authentication flow requests an email account. 
* Continue the login process with the email linked to the Jetpack account connected to the site.

Update release notes:


ghost (approved):

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -217,8 +219,31 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
...
+            let authenticationResult: WordPressAuthenticatorResult = .presentEmailController
+
+            onCompletion(authenticationResult)
+
+            return
+        }
+
+        /// We should never reach this point, as WPAuthenticator won't call its delegate for this case.
+        ///
+        let authenticationResult: WordPressAuthenticatorResult = .presentPasswordController(value: false)
```
jaclync, line 247: since this path is not expected to happen, does it make sense to return an error result and log an error like with `DDLogError`?

jaclync (approved):
Code & testing LGTM! :shipit: (tested with non-WP site, self-hosted site, and WP.com site)

WooCommerce/Classes/Authentication/AuthenticationManager.swift
ctarda: Good point. WPAuthenticator should implement the case and not call into the delegate, but if for some reason it does, I think it would be better to implement it just in case. I am very scared of making assumptions about what libraries do, in particular WPAuthenticator.

Logging this case makes a lot of sense. I'll add a line here

ctarda:
Thanks for the review @jaclync . I'll incorporate your feedback, point WPAuthenticator to a tag and then merge as soon as the WPAuthenticator release is propagated.

PR #3431, opened 2021-01-06
Use DeleteEntityObjectsMigrationPolicy to Delete ProductAttribute in 39 to 40 Migration

shiki:
This fixes the issue found by @Ecarrion (p91TBi-463-p2). When migrating from 40 to 41, this method returns the **incorrect mapping**:

https://github.com/woocommerce/woocommerce-ios/blob/a8ef64d3e14f64d94398333a7c5c68799c307fa7/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L316-L324

It returns `WooCommerceModelV39toV40.xcmappingmodel` which is supposed to be used only when migrating from 39 to 40. 

## Findings

Based on the team's efforts, we found that the possible cause is that the original `WooCommerceModelV39toV40.xcmappingmodel` excludes the `ProductAttributeToProductAttribute` mapping. We did this in order to delete the `ProductAttribute` rows. 

But for some reason, this causes the `NSMappingModel()` initializer to incorrectly return the `WooCommerceModelV39toV40.xcmappingmodel` when migrating from 40 to 41. The `NSMappingModel` should be _inferred_ in this situation. 

## Solution

To fix this, instead of excluding `ProductAttributeToProductAttribute`, we are keeping it but we'll use a custom `NSEntityMigrationPolicy`, [`DeleteEntityObjectsMigrationPolicy`](https://github.com/woocommerce/woocommerce-ios/blob/fix/mapping-model-39-to-40-because-core-data/Storage/Storage/Model/Migration%20Policies/DeleteEntityObjectsMigrationPolicy.swift), which deletes the entity objects. This has the same effect as the previous mapping. 

image

I also updated the tests to add more assertions and give me more confidence that this works as expected:

- bb88316 Used pre-inserted `ProductAttribute` to validate that it did not get deleted when migrating from 40 to 41.
- 4430479 Added validation that the `Product.attributes` relationship was updated and can still be used.

### Confirm

@pmusolino Can you confirm that the only change in the **original** `WooCommerceModelV39toV40.xcmappingmodel` is the exclusion of `ProductAttributeToProductAttribute`? Are there other changes? 

### Migration

1. Delete the `Model 40` and `Model 41` models. 
2. Set the `Model 39` as the current version. 
3. Run the app and log in if you haven't yet. 
4. Force close the app but don't rerun it yet.
5. Revert the model changes. 
6. Add breakpoint (A) on this line:
    https://github.com/woocommerce/woocommerce-ios/blob/5ec1fbc5847a13b8cd159253abb2c05665afd407/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L76
7. Add breakpoint (B) on this line:
    https://github.com/woocommerce/woocommerce-ios/blob/5ec1fbc5847a13b8cd159253abb2c05665afd407/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L86
7. Add breakpoint (C) on this line:
    https://github.com/woocommerce/woocommerce-ios/blob/5ec1fbc5847a13b8cd159253abb2c05665afd407/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L320
8. Add breakpoint (D) on this line:
    https://github.com/woocommerce/woocommerce-ios/blob/5ec1fbc5847a13b8cd159253abb2c05665afd407/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L323
8. Run the app. 
9. The database should be migrated from 39 to 41. 
10. Confirm that:
    * Breakpoint A was hit twice (migration steps 39 to 40 and 40 to 41)
    * Breakpoint B was never hit. This means the migration was successful.
    * Breakpoint C was hit once, before D. This means that 39 to 40 used  `WooCommerceModelV39toV40.xcmappingmodel`.
    * Breakpoint D was hit once. This means that 40 to 41 used an inferred mapping model. 
11. Confirm that you can load Orders and Products. Note that if you're using WC 4.9.0-rc.1, there's an outstanding regression that causes Products not to load. 

## Submitter Checklist


pmusolino:
> Can you confirm that the only change in the original WooCommerceModelV39toV40.xcmappingmodel is the exclusion of ProductAttributeToProductAttribute? Are there other changes?

@shiki I confirm that this was the only change. Also from the core data migrations document:  **Used mapping model: `WooCommerceModelV39toV40.xcmappingmodel` to remove product attributes without `siteID`.**

Ecarrion (approved):
Tested it and it seems to be working perfect!
Thanks for help @shiki, also TIL about `NSEntityMigrationPolicy`

shiki:
Thanks, everyone! YOLO!

PR #3432, opened 2021-01-06
[Platform] Update L10n Linter strategy to allow retry

AliSoftware:
This PR takes advantage of the new mode of the `ios_lint_localizations` action [recently added in release-toolkit 0.12.0](https://github.com/wordpress-mobile/release-toolkit/pull/182).

[Internal ref: paaHJt-1Fn-p2#comment-3524]

## What changed?

After this PR, if a localization violation is found, we will now be given the possibility to fix the violation locally and retry until we get it right – as opposed to waiting for it to be fixed in GlotPress.

This new strategy applies to **both** the `new_beta_release` and `finalize_release`.

_Note: Ideally this is still better to solve the violations at their source in GlotPress first then re-run the lanes from scratch. But it's also not always possible to wait for someone to be available to solve it in GlotPress for us; so this new behavior gives us an escape hatch when we can't wait for a fix of the copy in GP. We should still ping someone to fix the issue in GP in all cases._

## Target branch

This PR intentionally targets the ~`release/5.7`~ `release/5.8` branch and not `develop` because we're hoping to already use this new behavior for the upcoming release of 5.7, instead of waiting for the next code freeze to include it.
[EDIT: We apparently failed to get it reviewed in time 😅; now rebased and targeting `release/5.8`, hopefully we'll get it reviewed this time around 😄 ].

## To Test

The test steps are similar to the ones from https://github.com/wordpress-mobile/release-toolkit/pull/182:

 - Introduce a violation (incorrect `%d`/`%@`/… placeholders) in one of the `.strings` file
 - Make a temporary edit to the `finalize_release` lane to test the PR:
   - Comment the `ios_update_metadata()` call (to avoid your temporary violation from step above to be overwritten)
   - Replace the `ios_bump_version_beta()` and `ios_tag_build()` calls with something like `UI.message("And now new version is tagged")` (to be able to see if/when the next steps after `ios_lint_localizations` will be called as expected)
 - Run the modified lane: `bundle exec fastlane finalize_release` and ensure it finds the violations you introduced on purpose earlier.
 - Fix the violation in the local `.strings` file, then choose "Retry".
 - Check that the linter pass on this second try after violation has been fixed.

## Other: Security Fix

This PR will also incidentally close [this Dependabot security alert](https://github.com/woocommerce/woocommerce-ios/security/dependabot/Gemfile.lock/nokogiri/open) from our repo – which was a dependency of `release-toolkit`, which has been fixed in its version 0.12.0.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,16 +1,16 @@
 GIT
   remote: https://github.com/wordpress-mobile/release-toolkit
-  revision: 2f7ef851ed51fee6b5b8075056d5359c9e6277ea
-  tag: 0.11.0
+  revision: a932bd25320f2b82be0774b66e550c6862cd8765
+  tag: 0.12.0
   specs:
-    fastlane-plugin-wpmreleasetoolkit (0.11.0)
+    fastlane-plugin-wpmreleasetoolkit (0.12.0)
       activesupport (~> 5)
       bigdecimal (~> 1.4)
       chroma (= 0.2.0)
       diffy (~> 3.3)
       git (~> 1.3)
-      jsonlint
-      nokogiri (>= 1.10.4)
+      jsonlint (~> 0.3)
+      nokogiri (~> 1.11)
       octokit (~> 4.18)
       parallel (~> 1.14)
       progress_bar (~> 1.3)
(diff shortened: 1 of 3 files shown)


AliSoftware:
Rebased on top of `release/5.8` and milestone updated.

jkmassel (approved):
✅ Ensure it finds the violations you introduced on purpose earlier.
✅ Check that the linter pass on this second try after violation has been fixed.

AliSoftware:
Thanks for the review @jkmassel !

Issue #3434, opened 2021-01-07
Shipping Labels M1: remove feature flag `shippingLabelsRelease1` when it is stable in production


jaclync:
When the feature is stable in production (no crashes, reported issues, similar usage to Android in Tracks), let's remove the feature flag from the codebase.

This is also the last issue for the M1 Project!

PR #3436, opened 2021-01-07
Shipping Labels M1: turn on feature flag

### LINKED ISSUE #3433: Shipping Labels M1: turn on feature flag
Let's enable the feature flag for release 5.8 for us to beta test and potentially launch!

jaclync:
Fixes #3433

## Why

As the main features are complete for M1, let's turn on the feature flag for beta testing and public if no major showstoppers!

## Changes

Enabled the shipping labels M1 feature flag for all.

## Testing

Full testing guide: p91TBi-3xD-p2

- In Edit Scheme > Run > Build Configuration, you can update the configuration from Debug to Release
- Launch the app
- Go to the Orders tab
- Tap on an order with a shipping label --> shipping label card(s) should appear

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
+++ b/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
@@ -8,7 +8,7 @@ struct DefaultFeatureFlagService: FeatureFlagService {
         case .addProductVariations:
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .shippingLabelsRelease1:
-            return buildConfig == .localDeveloper || buildConfig == .alpha
+            return true
         default:
             return true
         }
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/System/DefaultFeatureFlagService.swift
```diff
@@ -8,7 +8,7 @@ struct DefaultFeatureFlagService: FeatureFlagService {
         case .addProductVariations:
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .shippingLabelsRelease1:
-            return buildConfig == .localDeveloper || buildConfig == .alpha
+            return true
```
ctarda, line 11: I'm just curious. The reason for setting the flag to true without removing it is so that it's easier to roll back the feature, if needed?

ctarda (approved):
Congratulations on shipping! 🎉 

This is what I see for an order with a shipping label:


:shipit:

WooCommerce/Classes/System/DefaultFeatureFlagService.swift
jaclync: yup exactly! I created another issue https://github.com/woocommerce/woocommerce-ios/issues/3434 as the last issue of the project to remove the feature flag when the feature has been beta tested, launched, and the analytics looking similar to Android 🙂

jaclync:
Thanks for the prompt review César, merging now!

PR #3438, opened 2021-01-08
Dynamic type support for `ImageAndTitleAndTextTableViewCell` image and title style

jaclync:
Part of #3334 

## Why

Before this PR, only the label font size adjusts with dynamic type in `ImageAndTitleAndTextTableViewCell` while the image stays the original size. To keep the changes small and easier to test, I limited the changes to usage of the new style `ImageAndTitleAndTextTableViewCell.Style` (you can search `imageAndTitleOnly` in the codebase).

One tradeoff is that the image quality could go down at larger font sizes (as shown in the **after** screenshots below) - I haven't found a good way to keep up the image quality, since the Gridicons generates an icon at a given size (usually 24px). One idea is to specify the maximum size (like 60px) but I'm not sure if it's worth keeping a larger image than needed for the majority of cases. I tested dynamic type on a device and the images didn't look as fuzzy as in the simulator, so I thought the benefit of dynamic support could outweigh the potential pixelated image. Lemme know if you have any ideas!

## Changes

- In `ImageAndTitleAndTextTableViewCell`, added `baseContentImageDimension: CGFloat` as the base image dimension for dynamic type scaling on the image width constraint. The value is set based on its `FontStyle` (24px for `body` and 20px for `footnote`). Please note that the image width constraint is deactivated for `updateUI(*ViewModel:)` usage so that only the `update(with:data:)` usage is affected
- Now that the image size is adjusted with content size category, passed the default image instead of footnote image from Gridicons

## Testing

There are 4 use cases for the deprecated `TopLeftImageTableViewCell`:

- Orders tab > refunded order with a refund reason > refund details --> the refund reason cell
- Orders tab > order with a customer note and can be fulfilled > Begin Fulfillment --> the customer provided note
- Orders tab > order with at least one non-refunded shipping label --> the printing info row
- Orders tab > order with at least one non-refunded shipping label > Reprint Shipping Label --> the info text, and two link rows

## Example screenshots

For design reviewer: please note that the image size is now adjusted based on the system scaling factor from dynamic type. We can change the base image size, and it will change the size at other system font sizes.

screen | before | after
-- | -- | --
Shipping label printing info row | Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 29 01 Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 29 15 Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 29 18 | Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 45 52 Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 46 28 Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 47 07
Reprint shipping label | Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 29 28 Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 29 35 Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 29 40 | Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 47 21 Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 47 29 Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 47 35
Refund details > refund reason | Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 53 17 Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 53 23 Simulator Screen Shot - iPhone 11 - 2021-01-08 at 15 53 26 | Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 33 39 Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 33 46 Simulator Screen Shot - iPhone 11 - 2021-01-12 at 11 33 56

Update release notes:


shiki:
I had some questions about this: CGPNUU63E-p1610131258011600-slack.

Garance91540 (approved):
👏

jaclync:
Based on the discussion CGPNUU63E-p1610131258011600-slack, I updated the PR to have a vertical stack view (image on top of title) when the font size is >= `.accessibilityMedium` (this is the threshold that Apple Reminders app changes its stack view axis from horizontal to vertical).

@shiki @Garance91540 the code and screenshots in the PR description are ready for review again! Also feel free to suggest any other changes like the spacing or threshold.

Garance91540 (approved):
Still looks good to me!

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
```diff
@@ -128,6 +131,11 @@ final class ImageAndTitleAndTextTableViewCell: UITableViewCell {
         configureTitleAndTextStackView()
         applyDefaultBackgroundStyle()
     }
+
+    override func prepareForReuse() {
```
shiki, line 135: Great foresight!

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
```diff
@@ -270,11 +292,27 @@ private extension ImageAndTitleAndTextTableViewCell {
...
+    func applyAccessibilityChanges(contentSizeCategory: UIContentSizeCategory) {
+        adjustContentStackViewAxis(contentSizeCategory: contentSizeCategory)
+    }
+
+    /// Changes the image view width according to the base image dimension.
+    func adjustContentStackViewAxis(contentSizeCategory: UIContentSizeCategory) {
+        let isVerticalStack = contentSizeCategory >= .accessibilityMedium
+        contentStackView.axis = isVerticalStack ? .vertical: .horizontal
+        contentStackView.alignment = isVerticalStack ? .leading: .firstBaseline
+        contentStackView.spacing = isVerticalStack ? 5: 16
```
shiki, line 316: Is there a reason why we're not using a multiple of `4`? Just curious. It can be really tricky to adjust pixels/points so I understand why we have to do this sometimes.

shiki (approved):
I love it! Thanks, @jaclync, for not skimping out on accessibility improvements.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
jaclync: Is a multiple of 4 a design preference documented somewhere? I tried 16px (too apart), 0px (too close), then 5px looks pretty good to me - didn't really think of the base multitude 😅 

Since G already did design review, let's go with this and we can always update the spacing later 🙂

shiki: I was wrong. It's not `4` but `8`. 😄 From https://woomobiledesignhandbook.wordpress.com/layout/:

> All layouts are based on a 8pt grid, with a preference for 16pt as the default reference unit (i.e. spacing, gaps, padding).

Yup. I was just curious here and it was a non-blocking question. 🙂

PR #3440, opened 2021-01-08
Refund Shipping Label: update header implementation so that it is scrollable in large font size

jaclync:
Part of #3334 

## Why

The header info text in Refund Shipping Label screen was originally implemented by a section header, which is sticky when scrolling through cells of the same section. Due to the length of the header text, it could grow very tall and prevent the user from scrolling to see the rest of the content with large font size.

This PR updated the header implementation from section header to a cell (`BasicTableViewCell`) so that it scrolls with the rest of the cells.

## Testing

- Go to the orders tab
- Tap the ellipsis menu ("...") at the top of the non-refunded shipping label card
- Tap "Request a Refund" in the action sheet
- Try various system font sizes --> the whole table view should be scrollable

## Example screenshots

(The header text is also left aligned with the cells below now!)

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2021-01-08 at 17 56 39 | Simulator Screen Shot - iPhone 11 - 2021-01-08 at 17 56 43
Simulator Screen Shot - iPhone 11 - 2021-01-08 at 17 57 02 | Simulator Screen Shot - iPhone 11 - 2021-01-08 at 17 57 15

Update release notes:


ctarda (approved):
Nice! 

This is the default font size:


And a couple of larger sizes:



:shipit:

Issue #3412, opened 2020-12-29
Variations: Networking, Storage and Yosemite for product attribute terms


Ecarrion:
- [x] Create product attribute term entity
- [x] Download product attribute terms
- [x] Create product attribute term storage entity 
- [x] Synchronize all attribute storage entities
- [x] Add action to create product attribute term
- [ ] Add action to delete product attribute term?

PR #3448, opened 2021-01-09
[Variations] Add ProductAttributeTermRemote class

Ecarrion:
part of #3412

# Why

In order to synchronize all terms for a product attribute, this PR adds the ability to download a single page of product attribute terms. The next PR will use these changes in order to sync and store all product attribute terms.

# How
- Added `ProductAttributeTermRemote` class and unit tests

# Extra
- Added a few helpers to load previously stored terms(none at the moment) and to conform to `ReadOnlyConvertible`. This was squashed here because the next PR was getting a bit too big.

# Testing Steps
- Make sure unit tests pass.

Update release notes:


Networking/Networking/Remote/ProductAttributeTermRemote.swift
```diff
@@ -0,0 +1,51 @@
...
+    ///     - attributeID: Identifier for the parent product attribute.
+    ///     - pageNumber: Number of page that should be retrieved.
+    ///     - pageSize: Number of terms to be retrieved per page.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func loadProductAttributeTerms(for siteID: Int64,
+                                          attributeID: Int64,
+                                          pageNumber: Int = Default.pageNumber,
+                                          pageSize: Int = Default.pageSize,
+                                          completion: @escaping ([ProductAttributeTerm]?, Error?) -> Void) {
```
jaclync: nit: how about returning `Result<[ProductAttributeTerm], Error>` on completion?

jaclync (approved):
LGTM! just a suggestion on using `Result` for the result 🙂

Networking/Networking/Remote/ProductAttributeTermRemote.swift
Ecarrion: Great suggestion, added here https://github.com/woocommerce/woocommerce-ios/pull/3448/commits/7d0bcdba0478558b0d398edc77e4606d752f562a

PR #3449, opened 2021-01-09
[Variations] Synchronize ProductAttributeTerm objects

Ecarrion:
Part of #3412

# Why

In order to allow merchants to create new product variations, we need to prove them with a list of product attribute terms that they can use to generate new variations.

This PR gives our codebase the `Action` and `Store` to synchronize all product variations terms given for the specified attribute.

Note that since `product attribute term` API is paginated, this synchronization method requests all pages recursively before completing the synchronization.

# How
- Add `ProductAttributeTermAction` that provides the `synchronizeProductAttributeTerms` case.
- Add `ProductAttributeTermStore` to handle the `synchronizeProductAttributeTerms` which recursively synchronize all pages.
- Delete all stale terms after a successful synchronization in the `deleteStaleTerms` method
- Added some mock JSON files to properly unit test the paginated synchronization.

# Testing steps
- Make sure unit tests pass.

Update release notes:


Networking/NetworkingTests/Responses/product-attribute-terms-empty.json
```diff
@@ -0,0 +1,4 @@
+{
+  "data": []
```
jaclync, line 2: jus

Networking/NetworkingTests/Responses/product-attribute-terms-empty.json
```diff
@@ -0,0 +1,4 @@
...
+}
+
```
jaclync, line 4: just wondering, are these two json files expected to generate diffs in the Networking layer project file? 🤔

Yosemite/YosemiteTests/Mocks/MockStorageManager+Sample.swift
```diff
@@ -80,6 +80,20 @@ extension MockStorageManager {
         return newProductAttribute
     }
 
+    /// Inserts a new (Sample) ProductAttributeTerm. and links it to a parent `ProductAttribute` if required.
```
jaclync: nit: does it make sense to update "if required" to "if available"?

Yosemite/YosemiteTests/Stores/ProductAttributeTermStoreTests.swift
```diff
@@ -0,0 +1,179 @@
...
+        let storedTerm = viewStorage.loadProductAttributeTerm(siteID: sampleSiteID, termID: sampleTermID, attributeID: sampleAttributeID)
+        let readOnlyTerm = try XCTUnwrap(storedTerm?.toReadOnly())
+        XCTAssertNotEqual(initialTerm, readOnlyTerm)
+        XCTAssertFalse(result.isFailure)
+    }
+
+    func test_synchronizeProductAttributeTerms_deletes_stale_terms() throws {
+        // Given
+        network.simulateResponse(requestUrlSuffix: sampleTermsPath, filename: "product-attribute-terms")
+        network.simulateResponse(requestUrlSuffix: sampleTermsPath, filename: "product-attribute-terms-empty")
```
jaclync, line 138: nit: we also started testing Yosemite store actions by creating a separate `*RemoteProtocol` (like `ProductsRemoteProtocol`) with a mock implementation so that we can simulate any networking layer response in code. Using the json files also works, up to you!

Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
```diff
@@ -0,0 +1,175 @@
...
+
+    override public func onAction(_ action: Action) {
+        guard let action = action as? ProductAttributeTermAction else {
+            assertionFailure("ProductAttributeTermStore received an unsupported action")
+            return
+        }
+
+        switch action {
+        case let .synchronizeProductAttributeTerms(siteID, attributeID, onCompletion):
+            synchronizeAllProductAttributeTerms(siteID: siteID, attributeID: attributeID, fromPageNumber: 0) { error in
```
jaclync: nit: how about starting from page number 1 (WC API has [1-based page number in pagination](https://woocommerce.github.io/woocommerce-rest-api-docs/#pagination)) so that `synchronizeAllProductAttributeTerms` doesn't have to use `fromPageNumber + 1`? if there are other reasons for 0-indexed page number here, a comment on the function's `fromPageNumber` parameter would be helpful 🙂

jaclync: Alternatively, since our use case in the app layer is to sync **all** attribute terms, we don't need the `fromPageNumber` param and can rename this action to be like `synchronizeAllProductAttributeTerms`? Unless we are supporting syncing a subset of pages?

Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
```diff
@@ -0,0 +1,175 @@
...
+            guard let self = self else { return }
+            switch result {
+
+            // If terms is empty, end the recursion and call `onCompletion`
+            case let .success(terms) where terms.isEmpty:
+                self.deleteStaleTerms(siteID: siteID, attributeID: attributeID, activeTerms: synchronizedTerms)
+                onCompletion(nil)
+
+            // If there could be more(non-empty) terms, request the next page recursively.
+            case let .success(terms):
```
jaclync, line 71: nit: a potential minor optimization here is to call completion if the term count is smaller than the page size (the last page is most often partially filled) so we don't have to make another API request for an empty response. If `terms.count < pageSize` is true, we can call `self.deleteStaleTerms(siteID: siteID, attributeID: attributeID, activeTerms: synchronizedTerms)` and `onCompletion` like in the case above

jaclync (approved):
Code looks good, just some non-blocking suggestions!

Yosemite/YosemiteTests/Stores/ProductAttributeTermStoreTests.swift
Ecarrion: Oh TIL! will take it into account for the next store!
I'll probably continue with the same approach for another method I have to add to this store to keep consistency tho!

Networking/NetworkingTests/Responses/product-attribute-terms-empty.json
Ecarrion: I think so, these files are shared between Yosemite and Network targets.

Yosemite/YosemiteTests/Mocks/MockStorageManager+Sample.swift
Ecarrion: Totally!

Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
Ecarrion: Great Suggestion!

Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
Ecarrion: Oh nice catch, I've updated to use page 1 as the first page 🤦.
We do need the `fromPageNumber` and the `fromPageNumber + 1` because that is the parameter that changes on the recursive call in order to fetch the next page when needed!

Ecarrion:
Thanks @jaclync for the review! I have applied your suggestions and will merge this now to not stall this feature anymore.

Networking/NetworkingTests/Responses/product-attribute-terms-empty.json
jaclync: somehow I didn't see diffs on these two json files in the project file in the PR 🤔 it seems like the tests worked without the project file diffs though

Issue #3084, opened 2020-10-29
Order Details: Rename Labels and Add Fees


shiki:
This is for the Orders → Order Details page.

## Order Total Updates

- “Paid by Customer” is becoming “Paid” for simplicity (for all orders, refunded or not)
- When a refund is processed, “Net Total” is becoming “Net Payment” to better reflect that it’s showing the total amount paid after the refund.

<img src="https://user-images.githubusercontent.com/198826/97629389-a8857d00-19f3-11eb-8b73-077259eb257b.png" width="300">

## Fees

You can add a fee item when creating a new product in wp-admin. The fee currently does not show up in Order Details. 

Add a Fee row in Order Details. 

<img src="https://user-images.githubusercontent.com/198826/97630284-fe0e5980-19f4-11eb-96c8-bccba12d22e1.png" width="300">

Ref p91TBi-36J-p2.

raziqfarid:
@shiki Would love to work on this. Can you please assign this to me?

shiki:
@raziqfarid Awesome! Thanks for contributing! I assigned this to you. 🙂

PR #3452, opened 2021-01-11
Networking: Add Fees to Orders

ctarda:
This is the first step to add support for Fees to the Order Details page. (#3084)

This PR implements support for Fees in Networking. Support for Fees in Yosemite and Storage will be added in subsequent PRs.

## Changes
- Add new model objects to Networking: `OrderFeeLine`, `OrderFeeTaxStatus` 
- Update `Order` to add a new property called `fees`
- Update mocks and json payloads in tests target
- Add a test to OrderMapperTests to test the parsing of an individual Order

## How to test
There are no user facing changes. 
- 👀 at the code 
- Check for a ✅ on CI.

Update release notes:


Networking/Networking/Model/OrderFeeLine.swift
```diff
@@ -0,0 +1,69 @@
...
+
+// MARK: - Comparable Conformance
+//
+extension OrderFeeLine: Comparable {
+    public static func == (lhs: OrderFeeLine, rhs: OrderFeeLine) -> Bool {
+        return lhs.feeID == rhs.feeID &&
+            lhs.name == rhs.name &&
+            lhs.total == rhs.total &&
+            lhs.totalTax == rhs.totalTax
+    }
```
jaclync: nit: do we intentionally choose this custom `Equatable` implementation, or does it make sense to conform to `Equatable` with the default implementation which compares all attributes? In other models, we generally use the default `Equatable` implementation so I was just curious when we choose to compare just a subset of attributes

Networking/Networking/Model/OrderFeeLine.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    public static func < (lhs: OrderFeeLine, rhs: OrderFeeLine) -> Bool {
+        return lhs.feeID < rhs.feeID ||
+            (lhs.feeID == rhs.feeID && lhs.name < rhs.name) ||
+            (lhs.feeID == rhs.feeID && lhs.name == rhs.name && lhs.total < rhs.total)
+    }
```
jaclync: nit: do we need this for the use case?

Networking/NetworkingTests/Mapper/OrderMapperTests.swift
```diff
@@ -254,6 +254,27 @@ final class OrderMapperTests: XCTestCase {
...
+        guard let fee = order.fees.first else {
+            XCTFail()
+            return
+        }
+
+        XCTAssertEqual(fee.feeID, 60)
+        XCTAssertEqual(fee.name, "$125.50 fee")
+        XCTAssertEqual(fee.taxStatus, .taxable)
+        XCTAssertEqual(fee.total, "125.50")
+        XCTAssertEqual(fee.totalTax, "0.00")
```
jaclync, line 277: nit: do we also want to compare other fields like `taxClass` and `taxes`? or compare two objects if we use the default `Equatable` implementation?

jaclync (commented):
Thanks for taking on this task adding fees!

Code: just some minor questions about the custom `Equatable` implementation and comparison!
Testing: `OrderStoreTests_FetchFilteredAndAllOrders.testWhenGivenAFilterItFetchesBothTheFilteredListAndTheAllOrdersList` didn't pass in CI, probably from the orders in `orders-load-all-2` not all having `fee_lines` field and thus these 4 orders can’t be parsed?

Networking/NetworkingTests/Mapper/OrderMapperTests.swift
ctarda: 🤔 good point. It probably makes more sense to just rely on the default implementations of Equatable and compare two objects directly. 

Thanks @jaclync ! I'll update and fix the breaking test

ctarda:
This is ready for another round @jaclync 

I removed the custom implementation of Equatable and Comparable, updated the mock so all tests should pass (sorry about that, 😬 ) and in the end went with touching all the object properties in the test, as it seemed to make that particular test a bit more explicit than testing equality, and it blended a bit better with the tests around it. Please let me know what you think

jaclync (approved):
LGTM! :shipit:

PR #3454, opened 2021-01-12
Persistence of sort order and filter options in the Product List

### LINKED ISSUE #2732: Products tab: persist sort order and filter options
On the products tab, the user can pick a sort order and filter option. Let's persist these two settings so that users see the same list of products between launches!

pmusolino:
Fixes #2732

## Description
In the Product List, on Android, users are able to see the filters that they previously applied because on that platform we store the sort order and the filters information for products.
I implemented the same behavior on iOS. From now, if a user applies custom sort orders and filters in the Product List, now when they reopen the app will be able to see the previous settings applied.
For implementing this solution, after having received some [feedback from the team](https://github.com/woocommerce/woocommerce-ios/issues/2732#issuecomment-756820161), I implemented new methods in [AppSettingsStore](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Stores/AppSettingsStore.swift) for fetching/upserting/deleting the product settings.
Why Product Settings (plural)? Since in the future probably we'll store other info about products, I decided to adopt a more general solution that can be reused for other purposes.
The Product Settings file is a `Plist` file stored locally, called `products-settings.plist`, that contains an array of `StoredProductSettings` entities, one for each `siteID`.

Schermata 2021-01-11 alle 18 22 45

## Changes
- Created a new model that conforms to `Codable` called `StoredProductSettings` that presents all the information that we want to store into the Product Settings file.
- Since we encode the product status, the stock status, and the product type, I updated the models `ProductType`, `ProductStockStatus` and `ProductStatus` to conform to `Codable`.
- Updated the `ProductsViewController` that now is able to update and sync the local product settings.
- Implemented new methods `loadProductsSettings`, `upsertProductsSettings` and `resetProductsSettings` in `AppSettingsStore` + Unit Tests.
- Since we need to store the Products sort order, now the enum `ProductsStortOrder` is of type `String`.

## Testing
1. Open the product list, then apply sorting and filters to your liking.
2. Close the app.
3. Re-open the app, and make sure that the product list is still filtered/sorted with the filters that you previously applied.
4. Change store, then apply new sorting and filters.
5. Return to your initial store. The filters should be still set.
6. Logout from the app.
7. Do a new login to the same account -> The filters/sorting options should be reset for all the stores.

Update release notes:


shiki:
> The Product Settings file is a Plist file stored locally, called products-settings.plist, that contains an array of StoredProductSettings entities, one for each siteID.

@woocommerce/ios-developers 

I would like to bring this up as a discussion with the team. Right now, the [`AppSettingsStore`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Stores/AppSettingsStore.swift) has quite [a lot of plist files](https://github.com/woocommerce/woocommerce-ios/blob/741bdea8bd119b2a7c35c1220146c5e16426343d/Yosemite/Yosemite/Stores/AppSettingsStore.swift#L22-L78). Some of those files are even deprecated. It's getting really cumbersome to read that class. And there's also a lot of boilerplate. 

With the additional `products-settings.plist` file proposed by this PR, we're adding more code in `AppSettingsStore`. The class is now at 500 lines. I think this is quite a lot of code for a supposedly simple goal: saving local settings. 

What I failed to emphasize in https://github.com/woocommerce/woocommerce-ios/issues/2732#issuecomment-756825639 is that I would have preferred to keep just **one plist file** with a **single overarching `struct`**. Something like:

```swift
public struct GeneralAppSettings: Codable, Equatable {
    public let installationDate: Date?
    public let feedbacks: [FeedbackType: FeedbackSettings]

    struct Products {
        public let siteID: Int64
        public let sort: String?
        public let stockStatusFilter: ProductStockStatus?
        public let productStatusFilter: ProductStatus?
        public let productTypeFilter: ProductType?
    }

    // Product-only settings
    public let products: Products 
}
```

I have hoped that having a single file and a single `struct` would help _simplify_ things and we wouldn't have to write a lot of code in order to add a local settings property. But I'm probably wrong. 😅 

📢  So, I'd like to hear what you all think. What should be the future of saving local settings? Perhaps we can start moving towards that. 

P.S. I miss the times when I could just use `NSUserDefaults`. 😄

jaclync:
One drawback with PList file implementation I noticed: whenever we deprecate an app setting (e.g. when Products M3 was fully launched and we removed the feature switch from settings), we add code to delete the PList file if it exists, but we don't know when we can remove this code to delete the file. It is a bit easier to reset `UserDefaults` on logout or just one setting by setting the value to `nil`. If there are no major drawbacks using `UserDefaults` (I can't think of any so far), it might be a simpler implementation?

pmusolino:
@shiki It makes sense to me to have one unique plist for managing settings. 

In theory, the process for upserting some settings remains the same. I do an example:
1. You have Products Settings and Feedback Settings in your plist file.
2. You want to update only the Product Settings, which is an array of settings. Every setting corresponds to a `siteID`.
3. Since the `plist` file should be re-generated every time you change also a single field, you should first load the current plist, search for the `siteID`, edit only that element of Product Settings, then delete and re-add the entire plist file.

With the current complexity of settings, it's not a pain, but what happens if in the future we access that file from multiple points in the app at the same time? I think that we risk experiencing concurrencies problems, for example, if you are updating the plist file while another process is reading that file.

> One drawback with PList file implementation I noticed: whenever we deprecate an app setting (e.g. when Products M3 was fully launched and we removed the feature switch from settings), we add code to delete the PList file if it exists, but we don't know when we can remove this code to delete the file. It is a bit easier to reset UserDefaults on logout or just one setting by setting the value to nil. If there are no major drawbacks using UserDefaults (I can't think of any so far), it might be a simpler implementation?

I think that for this specific case there are no big differences between plist and UserDefaults. At the same, if we use a single plist/UserDefaults, we can also forget about that property, without adding more complexity over there, since will be no more used in the future.
One pro that I see in favor of using a plist file, is that the debugging is easier with complex structs since we can access directly to the plist file and see the behavior. We can also update the plist directly in the file system to check the correct behavior.
One pro that I see in favor of UserDefaults is that it's easier to add/update/delete the preferences.

Networking/Networking/Model/Product/StoredProductSettings.swift
```diff
@@ -0,0 +1,38 @@
+import Foundation
+
+/// Models a pair of `siteID` and Product Settings
+/// These entities will be serialised to a plist file
+///
+public struct StoredProductSettings: Codable, Equatable {
```
shiki, line 6: ⚠️  I think it may be more appropriate if this is inside `Storage → Model → AppSettings`. I tried it and after fixing some compilation issues, it seems to work. 😄 My rationale is that `Networking` models would probably be best left for API-related models. But this one isn't since it's just for local settings. 

I think this should even be in Yosemite but I think there was a reason why `GeneralAppSettings` is in Storage too. 😄 

What do you think?

Networking/Networking/Model/Product/StoredProductSettings.swift
```diff
@@ -0,0 +1,38 @@
...
+                productStatusFilter: ProductStatus?,
+                productTypeFilter: ProductType?) {
+        self.siteID = siteID
+        self.sort = sort
+        self.stockStatusFilter = stockStatusFilter
+        self.productStatusFilter = productStatusFilter
+        self.productTypeFilter = productTypeFilter
+    }
+
+    public func numberOfFilters() -> Int {
```
shiki: ℹ️ This might be better named as `numberOfActiveFilters` to match the area where this is used. 🙂 

```suggestion
    public func numberOfActiveFilters() -> Int {
```

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -669,6 +685,42 @@ extension ProductsViewController: SyncingCoordinatorDelegate {
 
         ServiceLocator.stores.dispatch(action)
     }
+
+    /// Update local Products Settings (eg. sort order or filters stored in Products settings)
+    ///
+    func updateLocalProductSettings(sort: ProductsSortOrder? = nil,
+                                    stockStatusFilter: ProductStockStatus? = nil,
+                                    productStatusFilter: ProductStatus? = nil,
+                                    productTypeFilter: ProductType? = nil) {
```
shiki: ℹ️ This can be `private`. 🙂 

```suggestion
    private func updateLocalProductSettings(sort: ProductsSortOrder? = nil,
                                            stockStatusFilter: ProductStockStatus? = nil,
                                            productStatusFilter: ProductStatus? = nil,
                                            productTypeFilter: ProductType? = nil) {
```

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -669,6 +685,42 @@ extension ProductsViewController: SyncingCoordinatorDelegate {
...
+        let action = AppSettingsAction.upsertProductsSettings(siteID: siteID,
+                                                              sort: sort?.rawValue,
+                                                              stockStatusFilter: stockStatusFilter,
+                                                              productStatusFilter: productStatusFilter,
+                                                              productTypeFilter: productTypeFilter) { (error) in
+        }
+        ServiceLocator.stores.dispatch(action)
+    }
```
shiki: ⚠️  What do you think about passing the `filters` object instead? In this way, the method calls would be a bit shorter. Maybe it'll be easier to read? 🤷 

```suggestion
    private func updateLocalProductSettings(sort: ProductsSortOrder? = nil,
                                            filters: FilterProductListViewModel.Filters) {
        let action = AppSettingsAction.upsertProductsSettings(siteID: siteID,
                                                              sort: sort?.rawValue,
                                                              stockStatusFilter: filters.stockStatus,
                                                              productStatusFilter: filters.productStatus,
                                                              productTypeFilter: filters.productType) { (error) in
        }
        ServiceLocator.stores.dispatch(action)
    }

```

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -669,6 +685,42 @@ extension ProductsViewController: SyncingCoordinatorDelegate {
...
+
+    /// Fetch local Products Settings (eg.  sort order or filters stored in Products settings)
+    ///
+    func syncLocalProductsSettings(onCompletion: @escaping (Result<StoredProductSettings, Error>) -> Void) {
```
shiki: ℹ️ This can also be `private`. 🙂 

```suggestion
    private func syncLocalProductsSettings(onCompletion: @escaping (Result<StoredProductSettings, Error>) -> Void) {
```

WooCommerce/Classes/Yosemite/AuthenticatedState.swift
```diff
@@ -120,6 +120,7 @@ private extension AuthenticatedState {
         let resetStoredProviders = AppSettingsAction.resetStoredProviders(onCompletion: nil)
         let resetStoredStatsVersionStates = AppSettingsAction.resetStatsVersionStates
         let resetFeatureSwitchStates = AppSettingsAction.resetFeatureSwitches
-        ServiceLocator.stores.dispatch([resetStoredProviders, resetStoredStatsVersionStates, resetFeatureSwitchStates])
+        let resetProductsSettings = AppSettingsAction.resetProductsSettings
+        ServiceLocator.stores.dispatch([resetStoredProviders, resetStoredStatsVersionStates, resetFeatureSwitchStates, resetProductsSettings])
```
shiki, line 124: I didn't realize we have this existing dependency cycle. Gah! 🙈 



I wonder if we can use `self.dispatcher.dispatch()` but I'm too scared to try that at the moment. 😄

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -72,6 +72,10 @@ final class ProductsViewController: UIViewController {
     private var sortOrder: ProductsSortOrder = .nameAscending {
```
shiki: I'm kinda worried that we don't have unit tests for this. But I understand it's a bit challenging now since this class does not have a `ViewModel`. Something to think about in the future.

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -669,6 +685,42 @@ extension ProductsViewController: SyncingCoordinatorDelegate {
...
+    }
+
+    /// Fetch local Products Settings (eg.  sort order or filters stored in Products settings)
+    ///
+    func syncLocalProductsSettings(onCompletion: @escaping (Result<StoredProductSettings, Error>) -> Void) {
+        let action = AppSettingsAction.loadProductsSettings(siteID: siteID) { [weak self] (result) in
+            switch result {
+            case .success(let settings):
+                if let sort = settings.sort {
+                    self?.sortOrder = ProductsSortOrder(rawValue: sort) ?? .nameAscending
```
shiki: ⚠️ I suggest that we make the default `. nameAscending` as a part of `ProductsSortOrder` since `.nameAscending` is also used above to define the “default”. Maybe something like:

```swift
public enum ProductsSortOrder: String {
    static let default = .nameAscending
}
```

And then this can be:

```suggestion
                    self?.sortOrder = ProductsSortOrder(rawValue: sort) ?? .default
```

Yosemite/Yosemite/Actions/AppSettingsAction.swift
```diff
@@ -68,6 +68,25 @@ public enum AppSettingsAction: Action {
...
+    // MARK: - Products Settings
+
+    /// Loads the products settings
+    ///
+    case loadProductsSettings(siteID: Int64, onCompletion: (Result<StoredProductSettings, Error>) -> Void)
+
+    /// Add or Update products settings
+    ///
+    case upsertProductsSettings(siteID: Int64,
+                                sort: String? = nil,
```
shiki, line 80: I think it's fair to move `ProductsSortOrder` to Yosemite since Yosemite is business logic anyway. In this way, we don't need to use `String`. But up to you. I don't know if we have any guidelines on this.

Yosemite/Yosemite/Stores/AppSettingsStore.swift
```diff
@@ -467,6 +485,62 @@ private extension AppSettingsStore {
...
+            return
+        }
+
+        let settings = allSavedSettings.filter { $0.siteID == siteID }
+
+        guard let settingsUnwrapped = settings.first else {
+            let error = AppSettingsStoreErrors.noProductsSettings
+            onCompletion(.failure(error))
+            return
+        }
```
shiki: ⚠️ Have we considered encapsulating the uniqueness of `siteID` inside the `struct` instead? Something like:

```swift
/// Dummy name
public struct ProductsModuleSettings: Codable, Equatable {

    struct Filter {
        public let siteID: Int64
        public let sort: String?
        public let stockStatusFilter: ProductStockStatus?
        public let productStatusFilter: ProductStatus?
        public let productTypeFilter: ProductType?
    }

    /// SiteID: Filter
    let filters: [Int64: Filter]
}
```

In this way, we can hide this complexity as methods inside the `ProductsModuleSettings`. An example are these methods inside `GeneralAppSettings`:

https://github.com/woocommerce/woocommerce-ios/blob/42402d6b6799e24eb5b5d832476cb712fe2f7f2d/Storage/Storage/Model/GeneralAppSettings.swift#L28-L46

If we do this, I hope this class will be a bit shorter and will mostly only contain file saving operations.

shiki (commented):
I guess I'm really just worried about the boilerplate. I'm not particularly concerned with the `NSUserDefaults` and Plist file differences. 

> With the current complexity of settings, it's not a pain, but what happens if in the future we access that file from multiple points in the app at the same time?

I think this is a problem even with separate files. We'd probably still end up constraining everything in a single thread. 

> Why Product Settings (plural)? Since in the future probably we'll store other info about products, I decided to adopt a more general solution that can be reused for other purposes.

I'm still not convinced by the argument. I feel like it's premature to think about it at the moment. Who knows, maybe we'll end up not adding any more local settings in the future. 🤷 

Anyway, I'm happy to disagree and commit for now. We can start thinking about how we can improve `AppSettingsStore` in the meantime. 

---

@pmusolino I have reviewed this and have some things I would like to discuss to you about (marked ⚠️ ). And I have some tiny suggestions (ℹ️ ). The rest are random comments. 😄

Networking/Networking/Model/Product/StoredProductSettings.swift
pmusolino: Good question @shiki. 
In reality, there is a reason why I put it under Networking. Networking and Storage targets don't know each other, because there is the Yosemite layer that puts them in communication. 
If I move the `StoredProductSettings` model, Storage should import Networking because of the other models used inside it like `ProductStockStatus`, `ProductStatus` and `ProductType`. Did you solve the compilation issues importing the Networking framework?

About `GeneralAppSettings`, it's correct to keep in Storage because they don't need to import models declared in Networking. :nod:

WooCommerce/Classes/Yosemite/AuthenticatedState.swift
pmusolino: I'm scared too 🙈 I also noticed that the dispatch method you mentioned doesn't support an array of actions at the moment. :D Maybe should be part of another PR...

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
pmusolino: Absolutely agree. Initially, the `ProductsViewController` had to be just a container, but in the end, it is also fulfilling another role, which is to join the pieces between the various view controllers, sync coordinators, etc... do you think that it worth to create an issue just for that?

Yosemite/Yosemite/Stores/AppSettingsStore.swift
pmusolino: The class will be just 2 or 3 lines shorter, but at the same time we should introduce a new Codable model 😄 however if you think we can really get a benefit from it, it doesn't cost me anything to carry out this migration (also because once the model is defined, it would be difficult to modify it later)

pmusolino:
Thank you @shiki for your review, I really appreciated 🙇 

I updated the code based on your suggestions and left some responses. 

> I think this is a problem even with separate files. We'd probably still end up constraining everything in a single thread.

This can be for sure a problem also with separate files, but it would occur less frequently. Anyway, we should handle this case in the future, maybe using a read/write queue.

> Anyway, I'm happy to disagree and commit for now. We can start thinking about how we can improve AppSettingsStore in the meantime.

Do we have any setting that we should maintain and that is important for the users as far as you know? I think no since if a user lost the product filters settings or the app feedback is not a problem. If the answer is "Yes, it's not important", we can do that migration at any time, moving all the settings under a unique plist file or under UserDefaults. What do you think?

Networking/Networking/Model/Product/StoredProductSettings.swift
shiki: Doh! You're right. I had to import `Networking` inside `Storage`. 🤦 

I guess it's not really that important right now. I think we need to consolidate where these models are located at some point.

WooCommerce/Classes/Yosemite/AuthenticatedState.swift
shiki: We can take care of it later. It's just good to be aware of it. 🙂

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
shiki: I think the sync coordinator and any `Action` handling should be in a `ViewModel`. But it's not that important at the moment. Perhaps we can just add a `TODO` in the class documentation? Maybe when we modify this class next time, we can spend a bit of time to create a `ViewModel` for it.

Yosemite/Yosemite/Stores/AppSettingsStore.swift
```diff
@@ -467,6 +485,62 @@ private extension AppSettingsStore {
...
+
+        let settings = allSavedSettings.filter { $0.siteID == siteID }
+
+        guard let settingsUnwrapped = settings.first else {
+            let error = AppSettingsStoreErrors.noProductsSettings
+            onCompletion(.failure(error))
```
shiki, line 500: I just noticed this. I'm not sure how I feel about returning with an error because I believe it's not really an error but just a “there is no data“ state (`null`). And that is expected. But based on how we use this `Action`, I guess the difference is not a big deal. 😄

Yosemite/Yosemite/Stores/AppSettingsStore.swift
shiki: Yeah, if you're fine with that, let's do it. I think these parts could all be encapsulated in the `struct`: 

https://github.com/woocommerce/woocommerce-ios/blob/46b7fd986c225f0d05912fede77dca5f5a6b4c58/Yosemite/Yosemite/Stores/AppSettingsStore.swift#L498-L500

https://github.com/woocommerce/woocommerce-ios/blob/46b7fd986c225f0d05912fede77dca5f5a6b4c58/Yosemite/Yosemite/Stores/AppSettingsStore.swift#L517

https://github.com/woocommerce/woocommerce-ios/blob/46b7fd986c225f0d05912fede77dca5f5a6b4c58/Yosemite/Yosemite/Stores/AppSettingsStore.swift#L520-L525

The benefits that I see are:

- `ProductsModuleSettings` encapsulates loading `Filter` settings by `siteID` or updating by the `Filter` settings by `siteID`. It's the only object that knows how to update itself. 
- `AppSettingsStore` is only concerned about saving and loading the `plist` that contains the `ProductsModuleSettings`. 

These are significant to me even though we'd only be saving a few lines.

shiki (commented):
> Anyway, we should handle this case in the future, maybe using a read/write queue.

👍 

> If the answer is "Yes, it's not important", we can do that migration at any time, moving all the settings under a unique plist file or under UserDefaults. What do you think?

Yup, this. I agree. I just meant that it's good to discuss things like this, introspect, and question the norms. In this way, we can continue to innovate and improve in the future. 🙂 

---

I responded to your comments. Just a couple of things left. 🙂 Let me know if I can help in any way!

Yosemite/Yosemite/Stores/AppSettingsStore.swift
pmusolino: Thank you @shiki for the suggestions. I made the changes and you can find them in this commit 95be4c7048ea72f7b7f0051652a5f113886a9ad2
Consider that I decided to not use the word "filter` since we want to use them in a generic way, so the struct inside is called `Setting` :D

pmusolino:
@shiki ready for a second check 🙇

shiki (approved):
🤸

pmusolino:
Thank you for the review @shiki 🥇 I'm going to merge it right now

PR #3455, opened 2021-01-12
[Variations] Create product attribute term

### LINKED ISSUE #3412: Variations: Networking, Storage and Yosemite for product attribute terms

Ecarrion:
closes #3412 

# Why

This PR adds the ability to create a new product attribute term that will be used later to create a new product variation.
Ref:

variations-flow

# How
- Updated `ProductAttributeTermRemote` to include a `createProductAttributeTerm` function that creates the term remotely.
- Updated `ProductAttributeTermStore` to include a `createProductAttributeTerm` function that uses the remote function and stores the term internally upon completion.
- Updated `ProductAttributeTermAction` to be able to dispatch the function created on the store.

# Testing Steps
- Make sure unit tests run

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
+++ b/Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
@@ -111,9 +111,10 @@ private extension ProductAttributeTermStore {
                                     name: String,
                                     onCompletion: @escaping (Result<ProductAttributeTerm, Error>) -> Void) {
         remote.createProductAttributeTerm(for: siteID, attributeID: attributeID, name: name) { [weak self] result in
+            guard let self = self else { return }
             switch result {
             case let .success(term):
-                self?.upsertStoredProductAttributeTermsInBackground([term], siteID: siteID, attributeID: attributeID) {
+                self.upsertStoredProductAttributeTermsInBackground([term], siteID: siteID, attributeID: attributeID) {
                     onCompletion(.success(term))
                 }
             case .failure:
--- a/Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
+++ b/Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
@@ -39,6 +39,8 @@ public final class ProductAttributeTermStore: Store {
                 }()
                 onCompletion(result)
             }
+        case let .createProductAttributeTerm(siteID, attributeID, name, onCompletion):
+            createProductAttributeTerm(siteID: siteID, attributeID: attributeID, name: name, onCompletion: onCompletion)
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
```diff
@@ -101,6 +103,24 @@ private extension ProductAttributeTermStore {
...
+    /// Creates and stores a new product attribute term and links it to it's parent product attribute.
+    ///
+    func createProductAttributeTerm(siteID: Int64,
+                                    attributeID: Int64,
+                                    name: String,
+                                    onCompletion: @escaping (Result<ProductAttributeTerm, Error>) -> Void) {
+        remote.createProductAttributeTerm(for: siteID, attributeID: attributeID, name: name) { [weak self] result in
+            switch result {
+            case let .success(term):
+                self?.upsertStoredProductAttributeTermsInBackground([term], siteID: siteID, attributeID: attributeID) {
```
jaclync: nit: maybe we can guard on `self` at the beginning of the closure so that when it's `nil`, it doesn't only go through the `failure` case?

Yosemite/YosemiteTests/Stores/ProductAttributeTermStoreTests.swift
```diff
@@ -154,6 +154,29 @@ final class ProductAttributeTermStoreTests: XCTestCase {
...
+            let action = ProductAttributeTermAction.createProductAttributeTerm(siteID: self.sampleSiteID,
+                                                                               attributeID: self.sampleAttributeID,
+                                                                               name: "XXS") { result in
+                promise(result)
+            }
+            self.store.onAction(action)
+        }
+
+
+        // Then the category should be added
```
jaclync: nit:
```suggestion
        // Then the term should be added
```

jaclync (approved):
Looks great! 🚀

Yosemite/Yosemite/Stores/ProductAttributeTermStore.swift
Ecarrion: Nice catch!

Yosemite/YosemiteTests/Stores/ProductAttributeTermStoreTests.swift
Ecarrion: 🤦

Ecarrion:
Thanks for the review @jaclync, I've added your suggestions and will proceed to merge it!

PR #3459, opened 2021-01-13
Delete Unused Fulfillment Code

shiki:
Part of #3034.

The PR https://github.com/woocommerce/woocommerce-ios/pull/3453 removes the Fulfillment screen and adds the Mark Order Complete button to the Order Details screen. 

This removes the code related to the old Fulfillment screen. 

## Testing

Please test that you can still fulfill an order using the new Mark Order Complete button. 

1. Navigate to a Processing order.
2. Tap on Mark Order Complete.
3. Confirm in WC Admin that the order was changed to Completed.

## Submitter Checklist

- [x] If it's feasible, I have added unit tests. 
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #3034, opened 2020-10-20
Remove Order Fulfillment Screen


shiki:
When navigating to an order, a button named “Begin Fulfillment” is shown. 

<img src="https://user-images.githubusercontent.com/198826/96624978-c7d12b80-12ca-11eb-9f68-13f7f60e4317.png" width="300">

This leads to a fulfillment screen. This screen has since been removed in Android. Android just shows “Mark Order Complete”. 

<img src="https://user-images.githubusercontent.com/198826/96625063-e7685400-12ca-11eb-8500-0d91b847180a.png" width="300">

We should do the same in iOS and remove the fulfillment screen.

## Goals

- [ ] Remove Fulfillment Screen
- [ ] Move order completion logic to Order Details

---

Ref p91TBi-3mT-p2#comment-1857

raziqfarid:
@shiki Would love to work on this. Can you please assign this to me?

shiki:
This was done in #3453. There are a few things I need to supplement like unit tests but I'll create separate issues for those.

PR #3460, opened 2021-01-13
Remove EditableOrderTrackingTableViewCell

shiki:
Part of #3034. 

I found out after #3459 that there's yet another class that is no longer used, `EditableOrderTrackingTableViewCell`. This removes that. 🤸 

## Testing

The build passing should be enough I think. 🙂 

## Submitter Checklist


jaclync (approved):
Haha nice! 🧹 (We already have so many custom cells)

PR #3461, opened 2021-01-13
Storage: add fees to orders

ctarda:
This is the second step to add support for Fees to the Order Details page. (#3084). The first step is here: #3452 

This PR implements support for Fees in Storage. Support for Fees in Yosemite will be added in a subsequent PR.

## Changes
* Added a new version to coredata model
* Added a new entity: `OrderFeeLine` and new relationships to `Order`
* Added method to retrieve Fees to StorageType+Extensions

## How to test
* 👀 on the code. I haven't done this in a while, so I presume I have missed something.
* Coming from develop, check out the branch, build and run. There shouldn't be any coredata errors or warnings.
* CI should be ✅ 

Update release notes:


Ecarrion (commented):
Hey @ctarda, this looks solid to me!

A couple of things to confirm would be.
- Are we sure that we need a `nullify` rule on the Order -> FeeLines relationship? And what about on the FeeLine -> Taxes relationship? I asked this because I'm not tally aware of the problem domain but I think it's better to confirm! 

Also, it would be good to add a unit test on the migration, especially testing that a previously added order can be migrated successfully to have a `feeLines` set. Adding tests like that flagged out the problem we had with mapping models!

https://github.com/woocommerce/woocommerce-ios/blob/474baaf8dad9b2bac4e3c210fc3ede2283413be3/Storage/StorageTests/CoreData/MigrationTests.swift#L415-L439

ctarda:
Thanks for the review @Ecarrion , excellent points!

I just pushed an update to those relationships, and added a migration test (which surfaced an error in the way I had named the key associated to the relationship between Order and fees. Testing FTW!

Storage/StorageTests/CoreData/MigrationTests.swift
```diff
@@ -437,6 +437,42 @@ final class MigrationTests: XCTestCase {
...
+
+        // Confidence-check
+        XCTAssertEqual(try targetContext.count(entityName: "OrderFeeLine"), 1)
+
+        // The relationship between Order and OrderFeeLine should be updated.
+        XCTAssertEqual(migratedOrder.value(forKey: "fees") as? Set<NSManagedObject>, [fee])
+
+        // The OrderFeeLine.order inverse relationship should be updated.
+        XCTAssertEqual(fee.value(forKey: "order") as? NSManagedObject, migratedOrder)
+
```
Ecarrion: You may want to remove these extra empty lines!

Ecarrion (approved):
🚀

PR #3462, opened 2021-01-13
Shipping Labels > Paper Size Options: dynamic type layout polishes

jaclync:
Part of #3334 

## Why

The paper size options screen (a static screen) looks not so polished in larger font size. To better support dynamic type, this PR updated the SwiftUI layout so that the screen looks nicer with different font sizes.

## Changes

I've tried a few different layout approaches, and settled on the one in the PR (since we don't really have layout rules in the design regarding how each element grows to any optional max size). I capped the content width to 200px, because otherwise the images look unnecessarily huge in landscape.

This PR also added SwiftUI previews when the font size is XXXL and in portrait and landscape mode.

If you have other ideas for the layout, please feel free to suggest!

## Testing

Prerequisites: the site has [WooCommerce Shipping & Tax plugin](https://docs.woocommerce.com/document/woocommerce-shipping-and-tax/) and there is an order with at least one non-refunded shipping label. You can set up your store to use a test credit card with instructions in p91TBi-3xD-p2.

- Go to the orders tab
- After shipping labels are synced, tap "Reprint Shipping Label" in the non-refunded shipping label card
- Tap "See layout and paper size options" row
- Try adjusting system font size from small to large, rotating the device --> the layout should look reasonably nice (I noticed some minor leading misalignment in the first column in one system font size in iPhone 11 iOS 14.1 sim, might be an edge case)

## Example screenshots

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2021-01-13 at 17 50 05 | Simulator Screen Shot - iPhone 11 - 2021-01-13 at 17 52 26
Simulator Screen Shot - iPhone 11 - 2021-01-13 at 17 50 12 | Simulator Screen Shot - iPhone 11 - 2021-01-13 at 17 52 31
Simulator Screen Shot - iPhone 11 - 2021-01-13 at 17 50 15 | Simulator Screen Shot - iPhone 11 - 2021-01-13 at 17 52 34

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/GridStackView.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/Shipping Labels/GridStackView.swift
@@ -4,12 +4,13 @@ import SwiftUI
 struct GridStackView<Content: View>: View {
     private let rows: Int
     private let columns: Int
+    private let spacingBetweenColumns: CGFloat
     private let content: (Int, Int) -> Content
 
     var body: some View {
         VStack(alignment: .leading) {
             ForEach(0 ..< rows, id: \.self) { row in
-                HStack(alignment: .top) {
+                HStack(alignment: .top, spacing: spacingBetweenColumns) {
                     ForEach(0 ..< columns, id: \.self) { column in
                         content(row, column)
                     }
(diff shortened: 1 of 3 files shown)


ctarda (approved):
Code looks good, and the Ui looks even better to me 😄 




:shipit:

Garance91540 (approved):
👏

PR #3469, opened 2021-01-14
Order Details: fix shipping address order in different countries

### LINKED ISSUE #1872: Order Details: shipping address is flipped for some regions
**Describe the bug**

The address in the "Shipping details" and "Billing information" is flipped on both of my devices, but not on the simulators 🤔 

**To Reproduce**

- Go to the Orders tab
- Tap on an Order that contains a physical product and valid shipping address
- Note the address under CUSTOMER > Shipping details and also inside "View Billing Information"

**Screenshots**

**Expected behavior**

**Mobile Environment**
Please include:
- Device: iPhone XR & iPhone 6s Plus
- iOS version: 13.3.1 & 12.1
- WooCommerce iOS version: 3.6.0.2 (and also in 3.5 and 3.4 TestFlight beta builds)

jaclync:
Fixes #1872 

## Why

For certain countries (like some eastern Asian countries), the address starts from the largest unit to the smallest (e.g. from country to the apartment number). Currently, the address in order details just follows the "Region" device setting so all the addresses are formatted the same way (either from the largest to smallest unit or the other way). If the device is set to like Taiwan or Korea, it shows **all** the addresses from the largest to smallest unit. For stores with customers and shipping addresses in different countries, the address might look odd and cause confusion.

This PR sets the country code to `CNPostalAddress` which is used to display the address, so that each address is displayed in the order of the country.

In WC API, an [address's `country` field](https://woocommerce.github.io/woocommerce-rest-api-docs/#order-shipping-properties) follows the ISO 3166-1 alpha-2 format which is the expected value in [`CNPostalAddress`'s `isoCountryCode`](https://developer.apple.com/documentation/contacts/cnpostaladdress/1403396-isocountrycode).

## Testing

- In device settings > General > Language & Region, set the region to Korea or Taiwan
- Launch the app
- Go to the orders tab
- Tap on an order with a physical product (has a shipping address) --> the address should be displayed in the expected order (I think most countries have it from the smallest to largest unit)

## Example screenshots

address country | before | after
-- | -- | --
US | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 13 54 26 | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 14 03 54
Taiwan  (technically, the address should only be flipped when it is in Mandarin but I guess iOS doesn't consider the language it is in) | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 13 54 44 | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 14 04 19

Update release notes:


WooCommerce/Classes/Model/Address+Woo.swift
```diff
@@ -106,6 +106,7 @@ private extension Address {
         address.state = state
         address.postalCode = postcode
         address.country = country
+        address.isoCountryCode = country
```
ctarda, line 109: TIL

ctarda (approved):
:shipit:

Issue #3419, opened 2021-01-04
Setup ImageAndTitleAndTextTableViewCell with style/type


eshurakov:
Discussed briefly here: https://github.com/woocommerce/woocommerce-ios/pull/3407#discussion_r549619572

### Idea

At the moment the way the cell look depends on the view model that is used to configure it. We can break this dependency and provide a way to configure a cell with finite styles (or maybe "type" is a better word here?) and data independently. This would be somewhat similar to how UITableViewCell works.

Additional information that is specific for a style can be passed in associated values of the enum. 

Code example:

```
enum ImageAndTitleAndTextTableViewCell.Style {
  case default
  case switch(isOn: Bool, onChange: @escaping (_ isOn: Bool) -> Void)
  case warning
  case imageAndTitleOnly
}

extension ImageAndTitleAndTextTableViewCell {
  func update(with style: Style, title: String?, text: String?, image: UIImage?)

  // or combine title, text and image into data configuration object and pass along with the style
  func update(with style: Style, data: DataConfiguration)
}
```

### Subtasks

- [x] `imageAndTitleOnly` style https://github.com/woocommerce/woocommerce-ios/pull/3424 https://github.com/woocommerce/woocommerce-ios/pull/3471
- [x] `warning` style https://github.com/woocommerce/woocommerce-ios/pull/3470
- [ ] `switch` style
- [ ] default style

JorgeMucientes:
Closing as part of the Woo Mobile GitHub stale-issue audit cleanup.

Rationale for closing:
- Listed in audit section "Potentially Stale / Good Candidates To Close Still Open, Outside Grouped Commerce Rows (40)". Last GitHub activity in the audit: 2021-01-15 1994d. Audit recommendation: Keep / Manual review. Audit rationale: Open in the 2026-07-02 live snapshot; last GitHub update 2021-01-15 (1994d ago). Current-code signal from rg: `ImageAndTitleAndTextTableViewCell` appears in `WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift`. Current-code signal from rg: `UITableViewCell` appears in `WooCommerce/WordPressAuthenticator/Signin/LoginSocialErrorViewController.swift`.. No longer in a protected grouped area in this report; close as stale backlog cleanup and reopen if still reproducible or product-owned.

Reopen if this is still reproducible, product-owned, or should be tracked in a current Linear issue.

PR #3470, opened 2021-01-14
Setup `ImageAndTitleAndTextTableViewCell` with warning style

jaclync:
Part of #3419 

## Why

This is part of #3419 to refactor different `ImageAndTitleAndTextTableViewCell` view models with `Style` enum and default data configurations. Each style sets a subset of styling properties with optional parameters. This implementation simplifies different `*ViewModel` structs with enum cases, and allows other data configurations to be set to any value like `showSeparator` and `numberOfLines*` which could differ in each use case.

## Changes

This PR added a new style `warning` to replace `WarningViewModel` and updated its existing usage (missing price info row in Product Variation screen).

## Testing

Please do a confidence check on the only usage of warning style:

- Go to the Products tab
- Tap on a variable product with a variation that does not have a price
- Tap on the Variations row
- Tap on the variation without a price --> the warning row should look as before (no changes are expected)

## Example screenshots

The price warning row should look the same as `develop`.

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2021-01-14 at 15 22 32 | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 15 34 16
Simulator Screen Shot - iPhone 11 - 2021-01-14 at 15 34 53 | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 15 35 01

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Cells/ImageAndTitleAndTextTableViewCell.swift
@@ -15,6 +15,8 @@ final class ImageAndTitleAndTextTableViewCell: UITableViewCell {
     enum Style {
         /// Only the image and title label are displayed with a given font style for the title.
         case imageAndTitleOnly(fontStyle: FontStyle)
+        /// The cell's title, image, and background color are set to warning style.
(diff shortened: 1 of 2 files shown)


ctarda (approved):
| Light appearance | Dark appearance |
| ---- | ---- |
|  |  |
|  |  |

:shipit:

PR #3471, opened 2021-01-14
Replace `ImageAndTitleAndTextTableViewCell.TopLeftImageViewModel` with `imageAndTitleOnly` style

jaclync:
Part of #3419 

## Why

Similar to https://github.com/woocommerce/woocommerce-ios/pull/3470, we are replacing `ImageAndTitleAndTextTableViewCell`'s view models with `Style` enum. `imageAndTitleOnly` style case has been used in the app, and this PR replaced the last usage of `ImageAndTitleAndTextTableViewCell.TopLeftImageViewModel` with `imageAndTitleOnly` style in order details > shipping notice.

## Changes

- Removed manual tint color for `UIImage.noticeImage`, otherwise the icon cannot change tint color after switching light/dark mode
- In `OrderDetailsDataSource.configureShippingNotice`, replaced `ImageAndTitleAndTextTableViewCell.TopLeftImageViewModel` with `imageAndTitleOnly` style configuration
- In `ImageAndTitleAndTextTableViewCell`, removed `TopLeftImageViewModel` code

## Testing

The conditions when the shipping notice is shown require some work (the order needs to have multiple shipping methods), and I recommend commenting out these lines so that the shipping notice section is always shown in order details:

https://github.com/woocommerce/woocommerce-ios/blob/42402d6b6799e24eb5b5d832476cb712fe2f7f2d/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsDataSource.swift#L727-L731

- Go to the Orders tab
- Tap on an order where the shipping notice section is shown --> the shipping notice section should look as before

## Example screenshots

(No changes are expected)

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2021-01-14 at 16 02 01 | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 16 10 37
Simulator Screen Shot - iPhone 11 - 2021-01-14 at 16 04 32 | Simulator Screen Shot - iPhone 11 - 2021-01-14 at 16 10 59

Update release notes:


pmusolino (approved):
LGTM :shipit:

PR #3474, opened 2021-01-14
Improvements to Notice Present for preventing to add similar Notices in the queue

### LINKED ISSUE #1752: Improvements to NoticePresenter for cancelling a notice in the queue
We have some cases in the app, where we use the NoticePresenter to show some local errors.
The ideal would be to give the possibility to delete some Notices in the queue or prevent to add a notice more than one time if it is already in the queue.

**One case where can be useful:**
In _Products > Edit Product > Edit Price_ if you try to insert a sale price higher than the regular price, and you press save, we show an error notice. If you press saves more than one time, you will see a lot of similar Notices.

pmusolino:
Fixes #1752 

## Description
This is a small improvement, that has a big impact on our user experience. In fact, there are some cases where we use the `NoticePresenter` for presenting some alerts to our users, but sometimes it happens that we present the same error notice multiple times because we don't have any check on the queue before adding the same notice. One of these cases happens for example in **_Products > Edit Product > Edit Price_**: if you try to insert a sale price higher than the regular price, and you press save, we show an error notice. If you press the save button multiple times, you will see a lot of similar Notices. 
Now the `Notice` struct conforms to `Equatable`, and we check if a similar `Notice` is already in the queue.

## Testing
1. Open a product
2. Click on 'Price'
3. Set 'sale' price higher than 'regular' price
4. Press 'Done' button multiple times
5. Notice the error message
6. Leave the screen
7. Notice that the message disappears, and no other similar messages are shown.

## Screenshots

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 - [*] Removed fulfillment screen and moved fulfillment to the order details screen. [https://github.com/woocommerce/woocommerce-ios/pull/3453]
 - [*] Fix: billing information action sheets now are presented correctly on iPad. [https://github.com/woocommerce/woocommerce-ios/pull/3457]
 - [*] fix: the rows in the product search list now don't have double separators. [https://github.com/woocommerce/woocommerce-ios/pull/3456]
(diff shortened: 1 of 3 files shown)


Ecarrion (approved):
Makes sense, thanks for tackling this!

pmusolino:
Thank you Ernesto for the quick review! 💯

Issue #3465, opened 2021-01-13
Implement typed NSPredicates


Ecarrion:
Ref: p91TBi-3BJ-p2

Tasks:
- [x] Unit tests existing predicates.
- [x] Add [typed predicates](https://medium.com/@redhotbits/building-nspredicate-with-swift-keypaths-and-operators-5ff569db304f) support.
- [x] Replace existing predicates with typed predicates

PR #3480, opened 2021-01-15
Unit test StorageType+Extension methods

Ecarrion:
part of #3465 

# Why 
In order to safely replace our stringly type predicates( p91TBi-3BJ-p2) with a type-safe version, I think it's more than required to unit test our current implementation to minimize the chance of regressions during the migration.

# How
This PR adds unit tests for all the methods on `StorageType+Extensions.swift` which are the methods where we will be replacing the old predicates.

# Testing Steps
- Take a look at the code and make sure the tests pass!

Update release notes:


ctarda (approved):
Thanks for doing this, @Ecarrion It must have been very painful to go through all the methods one by one.

Great work. :shipit:

Ecarrion:
Thanks @ctarda for the review!

Issue #3312, opened 2020-12-09
[UL&S] Accessibility review


ctarda:
* Run the new flow using VoiceOver
* Run the new flow using larger fonts

PR #3483, opened 2021-01-17
[UL&S] Match store address with wordpress.com account after login is completed and present error if necessary

### LINKED ISSUE #3482: [UL&S] Catch the case when users initiate a log in with a store address, then use a wp account different to the store address entered
For context:
pbMoDN-1KW#comment-3684-p2
pbMoDN-1KW#comment-3687-p2

When logging in with store address, we are missing presenting an error when a user logs in with a wordpress.com or jetpack account that do not match the store address entered.

ctarda:
Closes #3482 

This is a PR against the 5.8 release branch, to fix an omission picked up during testing:

pbMoDN-1KW#comment-3684-p2
pbMoDN-1KW#comment-3687-p2

In action:

https://user-images.githubusercontent.com/2722505/104858404-cb397a80-58ec-11eb-9d8c-d3c1ad65c3bb.mov

Screenshots:

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |
|  |  |
|  |  |

## Rationale

When logging with Store Address, the flow is:
1. enter the store address
2. enter the email associated to the wp.com account
3. enter password
4. pick store

We missed checking if the email/password entered in steps 2 and 3 and the store address entered in step 1 correspond to the same account.

There is a serious limitation to do so. At the end of the login process, WooCommerce is notified by WordPressAuthenticator of the userId corresponding to the email/password entered in steps 2+3. However, there is no way for neither WPAuthenticator, nor WooCommerce, to know at that point the userID that corresponds to the store address entered in step 1.

## Changes
* After the login process is completed, and before presenting the store picker, we will check if the store address entered in step 1 is one of the stores associated to the wordpress.com account entered in steps 2+3. To do so, we use `ULAccountMatcher`
* Add a new error screen specific to this use case: `ULAccountMismatchViewController`. Notice there is an issue with the spacing of the "Wrong Account?" label and "LogOut" button. I can not figure out for the life of me why how to solve that gap. Any help or suggestions are more than welcome.
* Add a new view model specific for this case.
* Add tests
* Add a new configuration to StorePickerViewController to support displaying only the list of sites. StorePickerViewController had already three responsibilities at this point, and I am adding a fourth now, but I am afraid to break it down this close to a release. We should probably think of ways to simplifying it, by breaking it down into its four responsibilities.

## How to test
Prepare two different accounts with at least one store each. Let's call them Account A and Account B

First testing flow:
1. In the login prologue, tap Login With Store Address. 
2. Enter the address of a store associated to account A
3. In the next step, enter the email and password associated to account B
4. Notice if there is an error screen. 
5. In the error screen, tap the Log Out button.
6. The app should navigate to the login prologue again

Second testing flow:
1. In the login prologue, tap Login With Store Address. 
2. Enter the address of a store associated to account A
3. In the next step, enter the email and password associated to account B
4. Notice if there is an error screen. 
5. In the error screen, tap See Connected Stores.
6. The app should navigate to a list of stores associated with account B
7. Tap back, tap Log Out

Third testing flow:
1. In the login prologue, tap Login With Store Address. 
2. Enter the address of a store associated to account A
3. In the next step, enter the email and password associated to account A
4. The app should navigate to the store picker
5. Select a store, everything should work as expected.

cc: @designsimply 

Update release notes:

### DIFF
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -54,7 +54,7 @@ PODS:
   - WordPress-Aztec-iOS (1.11.0)
   - WordPress-Editor-iOS (1.11.0):
     - WordPress-Aztec-iOS (= 1.11.0)
-  - WordPressAuthenticator (1.34.0-beta.1):
+  - WordPressAuthenticator (1.34.0-beta.2):
     - 1PasswordExtension (~> 1.8.6)
     - Alamofire (~> 4.8)
     - CocoaLumberjack (~> 3.5)
@@ -107,7 +107,7 @@ DEPENDENCIES:
   - Kingfisher (~> 5.11.0)
   - Sourcery (~> 0.18)
   - WordPress-Editor-iOS (~> 1.11.0)
-  - WordPressAuthenticator (from `../WordPressAuthenticator-iOS`)
+  - WordPressAuthenticator (from `https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git`, branch `issue/563-allow-shortcircuit-sync`)
   - WordPressShared (~> 1.12)
   - WordPressUI (~> 1.7.2)
   - Wormholy (~> 1.6.4)
@@ -158,7 +158,13 @@ SPEC REPOS:
 
 EXTERNAL SOURCES:
   WordPressAuthenticator:
-    :path: "../WordPressAuthenticator-iOS"
+    :branch: issue/563-allow-shortcircuit-sync
+    :git: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git
+
+CHECKOUT OPTIONS:
+  WordPressAuthenticator:
+    :commit: cf8af893adeef7fd58e42b0dbb7db8febadacecc
+    :git: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git
 
 SPEC CHECKSUMS:
   1PasswordExtension: f97cc80ae58053c331b2b6dc8843ba7103b33794
@@ -185,7 +191,7 @@ SPEC CHECKSUMS:
   UIDeviceIdentifier: a79ccdfc940373835a7d8e9fc7541e6bf61b6319
   WordPress-Aztec-iOS: 050b34d4c3adfb7c60363849049b13d60683b348
   WordPress-Editor-iOS: 304098424f1051cb271546c99f906aac296b1b81
-  WordPressAuthenticator: 32b3fc31ff91c0d9e616d5f91c40f7a2294deac7
+  WordPressAuthenticator: e369a5a6ba8d92400c4672e005f6de054d3169c2
   WordPressKit: 1b61217cea9e7045ff74a9c8df391d7788776daf
   WordPressShared: 38cb62e9cb998d4dc3c1611f17934c6875a6b3e8
   WordPressUI: cb5d0c58f92778b6dffcdcb8234ed8d7a930ce90
@@ -201,6 +207,6 @@ SPEC CHECKSUMS:
   ZendeskSupportProvidersSDK: e183d32abac888c448469e2005c4a5a8c3ed73f0
   ZendeskSupportSDK: e52f37fa8bcba91f024b81025869fe5a2860f741
 
-PODFILE CHECKSUM: 0c10a2cb70f0c10b869dc5c9193f259186397a72
+PODFILE CHECKSUM: d042a1c1520cc733a33dc8011c97dbdccab5cde1
 
 COCOAPODS: 1.10.0
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -54,7 +54,7 @@ PODS:
   - WordPress-Aztec-iOS (1.11.0)
   - WordPress-Editor-iOS (1.11.0):
     - WordPress-Aztec-iOS (= 1.11.0)
-  - WordPressAuthenticator (1.34.0-beta.2):
+  - WordPressAuthenticator (1.34.0-beta.1):
     - 1PasswordExtension (~> 1.8.6)
     - Alamofire (~> 4.8)
     - CocoaLumberjack (~> 3.5)
@@ -107,7 +107,7 @@ DEPENDENCIES:
   - Kingfisher (~> 5.11.0)
   - Sourcery (~> 0.18)
   - WordPress-Editor-iOS (~> 1.11.0)
-  - WordPressAuthenticator (from `https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git`, branch `issue/563-allow-shortcircuit-sync`)
+  - WordPressAuthenticator (~> 1.34.0-beta)
   - WordPressShared (~> 1.12)
   - WordPressUI (~> 1.7.2)
   - Wormholy (~> 1.6.4)
@@ -141,6 +141,7 @@ SPEC REPOS:
     - UIDeviceIdentifier
     - WordPress-Aztec-iOS
     - WordPress-Editor-iOS
+    - WordPressAuthenticator
     - WordPressKit
     - WordPressShared
     - WordPressUI
@@ -155,16 +156,6 @@ SPEC REPOS:
     - ZendeskSDKConfigurationsSDK
     - ZendeskSupportProvidersSDK
     - ZendeskSupportSDK
-
-EXTERNAL SOURCES:
-  WordPressAuthenticator:
-    :branch: issue/563-allow-shortcircuit-sync
-    :git: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git
-
-CHECKOUT OPTIONS:
-  WordPressAuthenticator:
-    :commit: cf8af893adeef7fd58e42b0dbb7db8febadacecc
-    :git: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git
 
 SPEC CHECKSUMS:
   1PasswordExtension: f97cc80ae58053c331b2b6dc8843ba7103b33794
@@ -191,7 +182,7 @@ SPEC CHECKSUMS:
   UIDeviceIdentifier: a79ccdfc940373835a7d8e9fc7541e6bf61b6319
   WordPress-Aztec-iOS: 050b34d4c3adfb7c60363849049b13d60683b348
   WordPress-Editor-iOS: 304098424f1051cb271546c99f906aac296b1b81
-  WordPressAuthenticator: e369a5a6ba8d92400c4672e005f6de054d3169c2
+  WordPressAuthenticator: 32b3fc31ff91c0d9e616d5f91c40f7a2294deac7
   WordPressKit: 1b61217cea9e7045ff74a9c8df391d7788776daf
   WordPressShared: 38cb62e9cb998d4dc3c1611f17934c6875a6b3e8
   WordPressUI: cb5d0c58f92778b6dffcdcb8234ed8d7a930ce90
@@ -207,6 +198,6 @@ SPEC CHECKSUMS:
   ZendeskSupportProvidersSDK: e183d32abac888c448469e2005c4a5a8c3ed73f0
   ZendeskSupportSDK: e52f37fa8bcba91f024b81025869fe5a2860f741
 
-PODFILE CHECKSUM: d042a1c1520cc733a33dc8011c97dbdccab5cde1
+PODFILE CHECKSUM: 68967742aa99a2585d61cc4c38dcef68ca7cb36e
 
 COCOAPODS: 1.10.0
--- a/Podfile.lock
+++ b/Podfile.lock
@@ -107,7 +107,7 @@ DEPENDENCIES:
   - Kingfisher (~> 5.11.0)
   - Sourcery (~> 0.18)
   - WordPress-Editor-iOS (~> 1.11.0)
-  - WordPressAuthenticator (~> 1.34.0-beta)
+  - WordPressAuthenticator (from `../WordPressAuthenticator-iOS`)
   - WordPressShared (~> 1.12)
   - WordPressUI (~> 1.7.2)
   - Wormholy (~> 1.6.4)
@@ -141,7 +141,6 @@ SPEC REPOS:
     - UIDeviceIdentifier
     - WordPress-Aztec-iOS
     - WordPress-Editor-iOS
-    - WordPressAuthenticator
     - WordPressKit
     - WordPressShared
     - WordPressUI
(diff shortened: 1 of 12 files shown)


ctarda:
Sorry for the 500 lines, but there is one new xib file and several tests

Garance91540 (approved):
Hi @ctarda , just a small UI fix: is it possible to remove the spacing between "wrong account" and "log out"? https://d.pr/i/lqyA5Q

jaclync:
Hi @ctarda, sorry I didn't get to review this PR today and I can take a look tomorrow if no one has. Do you think we can target `develop` since the changes are non-trivial and this also seems to be a pre-existing issue? (I was able to reproduce with release 5.7.0.2, pre UL&S)

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -253,6 +253,18 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
     /// Presents the Login Epilogue, in the specified NavigationController.
     ///
     func presentLoginEpilogue(in navigationController: UINavigationController, for credentials: AuthenticatorCredentials, onDismiss: @escaping () -> Void) {
+        let matcher = ULAccountMatcher()
+
+        guard let siteURL = credentials.wpcom?.siteURL, matcher.match(originalURL: siteURL) else {
+            //fatalError("Self Hosted sites are not supported. Please review the Authenticator settings!")
```
Ecarrion: Should we remove this comment?

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -253,6 +253,18 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
...
+            DDLogWarn("⚠️ Present account mismatch error for site: \(String(describing: credentials.wpcom?.siteURL))")
+            let viewModel = WrongAccountErrorViewModel(siteURL: credentials.wpcom?.siteURL)
+            let mismatchAccountUI = ULAccountMismatchViewController(viewModel: viewModel)
+            navigationController.pushViewController(mismatchAccountUI, animated: true)
+
+            return
```
Ecarrion: Nit: This could be reduced into:
```suggestion
            return navigationController.show(mismatchAccountUI, sender: nil)
```

WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
```diff
@@ -220,6 +226,15 @@ private extension StorePickerViewController {
                                                            action: #selector(cleanupAndDismiss))
     }
 
+    func setupNavigationForListOfConnectedStores() {
+        title = NSLocalizedString("Connected Stores", comment: "Page title for the list of connected stores")
```
Ecarrion, line 230: Should this be inside a `Localizations` enum?

WooCommerce/Classes/Authentication/Navigation Exceptions/ULAccountMismatchViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    /// Constraints on the view containing the action buttons
+    /// and the stack view containing the image and error text
+    /// Used to adjust the button width in unified views provided by WPAuthenticator
+    @IBOutlet private weak var buttonViewLeadingConstraint: NSLayoutConstraint!
+    @IBOutlet private weak var buttonViewTrailingConstraint: NSLayoutConstraint!
+    @IBOutlet private weak var stackViewLeadingConstraint: NSLayoutConstraint!
+    @IBOutlet private weak var stackViewTrailingConstraint: NSLayoutConstraint!
+
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        UIDevice.isPad() ? .all : .portrait
```
Ecarrion, line 39: TIL! I thought we allowed all orientations all the time!

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    // MARK: - Actions
+    func didTapPrimaryButton(in viewController: UIViewController?) {
+        guard let navigationController = viewController?.navigationController else {
+            return
+        }
+
+        let storePicker = StorePickerViewController()
+        storePicker.configuration = .listStores
+        //storePickerCoordinator?.onDismiss = onDismiss
```
Ecarrion: nit: Delete?

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+        navigationController.pushViewController(storePicker, animated: true)
```
Ecarrion, line 87: nit: prefer `show` over `pushViewController`

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    func didTapAuxiliaryButton(in viewController: UIViewController?) {
+        let fancyAlert = FancyAlertViewController.makeNeedHelpFindingEmailAlertController()
+        fancyAlert.modalPresentationStyle = .custom
+        fancyAlert.transitioningDelegate = AppDelegate.shared.tabBarController
```
Ecarrion, line 93: This feels a bit drastic, but I'm not sure why is it needed 🤔 
Could you explain it a little bit more?

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
```diff
@@ -0,0 +1,138 @@
...
+        viewController?.present(fancyAlert, animated: true)
+    }
+
+    func didTapLogOutButton(in viewController: UIViewController?) {
+        // Log out and pop
+        ServiceLocator.stores.deauthenticate()
```
Ecarrion: Do you think it's worth it to inject these dependencies, would it help on unit tests?

Ecarrion (commented):
Testing was good! But I did found some issues when working with larger accessibility fonts, however, feel free to tackle them in another PR if you prefer.

Screen Shot 2021-01-18 at 8 45 28 AM

Also, here are some thoughts that may have been resolved before 😅 
- Wouldn't this be resolved/not needed if we defaulted to the correct email address once the user has entered the store address?
- Is that even possible? or will it be considered a security breach?
- I'm not finding many benefits of entering the store address if then I have to enter my account/password which will then present store picker.

Ecarrion:
@ctarda regarding the spacing on the logout button I think it's because the `logOutButton.applyLinkButtonStyle()` line applies a custom edge inset.

Removing that inset removes the extra space, it's just a matter of you to find the correct value.

Removing the inset
```swift
func configureLogOutButton() {
    logOutButton.applyLinkButtonStyle()
    logOutButton.setTitle(viewModel.logOutButtonTitle, for: .normal)
    logOutButton.contentEdgeInsets = .zero
    logOutButton.on(.touchUpInside) { [weak self] _ in
        self?.didTapLogOutButton()
    }
}
```

Removes the space
Screen Shot 2021-01-18 at 9 06 10 AM

WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
ctarda: Right. I didn't do it because this class does not declare a Localizations enum, and I didn't want to change anything that wasn't strictly necessary. I would prefer to leave this as it is for now. IMO, we really need to rethink StorePickerViewController and break it down.

WooCommerce/Classes/Authentication/Navigation Exceptions/ULAccountMismatchViewController.swift
ctarda: Not for login. WordPressAuthenticator doesn't, so we can't allow rotation in the custom UI we insert into it

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
ctarda: We need to set the transitioningDelegate otherwise the fancy alert does not grab out the background. I presume what's called you attention is pulling the tabBarController from the AppDelegate directly? 

If that's the case, I guess we could pull the tabBarController from the UIViewController itself, but then we'd need to downcast to MainTabBarController, which I am not sure makes things better. Also, it's the pattern used elsewhere.

I am not sure if that answers you question though.

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
ctarda: Sure. Just for the record, DI is not the only way to unit test services get hit, ServiceLocator also exposes setters precisely for tests

ctarda:
Thanks a lot for the review @Ecarrion , 10_000 extra points for figuring out what was going on with the spacing of the Log Out button. I spent an embarrassing amount of time trying to fix that yesterday 🙇 

@Garance91540 this is how the line with the Log Out button looks now:



I just pushed an update that implements all the feedback (with one exception pointed up in the comments) 

To answer some of your questions:

> Wouldn't this be resolved/not needed if we defaulted to the correct email address once the user has entered the store address?
> Is that even possible? or will it be considered a security breach?

WordPressAuthenticator does not provide that information just with the store address.

> I'm not finding many benefits of entering the store address if then I have to enter my account/password which will then present store picker.

I will defer to @Garance91540 to answer that 😊

ctarda:
Oh, and one more thing @Ecarrion Thanks for pointing out the issues with larger fonts. We have an ongoing issue for that: #3312

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
Ecarrion: Thanks for the explanation, yes it was due to pulling the "tabBarController from the AppDelegate directly"
I don't have a better suggestion at the moment so let's leave it like that if you prefer.

WooCommerce/Classes/Authentication/Navigation Exceptions/WrongAccountErrorViewModel.swift
Ecarrion: True, but we have faced scenarios where a test sets something in `ServiceLocator` but then does not clean it correctly, which led to inconsistencies in other tests where is not clear that `ServiceLocator` is used behind the scenes.

This is easily solved by exposing all possible dependencies in the `init` :-)

Ecarrion (approved):
🚀

ctarda:
Alright, merging now.

Podfile.lock
```diff
@@ -66,7 +66,7 @@ PODS:
     - WordPressKit (~> 4.18-beta)
     - WordPressShared (~> 1.12-beta)
     - WordPressUI (~> 1.7-beta)
-  - WordPressKit (4.24.0):
+  - WordPressKit (4.25.0-beta.3):
```
AliSoftware, line 69: Hey @ctarda. Just wanted to triple-check with you (and @Ecarrion) that this was intentional.
This bump means that this PR that targets the code-frozen 5.8 branch will now include code of `WordPressKit` from _after_ the code freeze of 5.8, potentially some that was added to provide features and changes aimed for / used only… in the current `develop` branches WordPress or Woo.

Strictly speaking, if a fix on the current release branch needs one of our dependency to be updated to add a specific fix on it, it's usually more appropriate to cut a hotfix branch from 4.24.0 (the version of the dependency we used when `release/5.8` was code-frozen) just to add the specific fix you need on top, and then make a 4.24.1 and point to it.

As you updated WordPressKit to `4.25.0-beta.3` here, that might mean that your fix on release/5.8 will not only include whatever you changed and added to `WordPressKit` specifically to fix this issue, but also any other change that was added between `4.24.0` and `4.25.0-beta.2`.

That being said, while it would be a bit dangerous to include all that if that was a hotfix on an already-published production build, that might not be that big of a deal in the case of a fix on a `release/5.8` like here that was not out yet and will still have time to be beta-tested before your fix is released to prod, but that's still something to be wary of. I think it's ok to keep that **if you're entirely sure that whatever unrelated changes between `4.24.0` and `4.25.0-beta.2` you brought (in addition to whatever was in `4.25.0-beta.3` you needed here) didn't include neither a regression, nor a new feature that was meant for `develop` or for the future `5.9` only**.

---

For reference, changes brought by this dependency bump seems to include [WordPressKit-iOS#325](https://github.com/wordpress-mobile/WordPressKit-iOS/issues/325) (4.25.0-beta.1), [WordPressKit-iOS#327](https://github.com/wordpress-mobile/WordPressKit-iOS/issues/327) (4.25.0-beta.2), and [WordPressKit-iOS#326](https://github.com/wordpress-mobile/WordPressKit-iOS/issues/326) (4.25.0-beta.3). I'm actually wondering if that bump was necessary to address your fix after all, vs if it was done entirely by mistake (and if so, we should have kept 4.24.0 in the first place).

WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
AliSoftware: ⚠️ Does this means we now have an entirely new string in the codebase after this PR? This means that this string won't be translated in time for 5.8 – as the GlotPress bundle containing all the copies to be translated has already been sent to our translations vendor on Sunday

Podfile.lock
ctarda: Good catch @AliSoftware 

It was not on purpose. I think what might have happened is:
* At one moment in time, while working on the PR, I [bumped WPAuthenticator](https://github.com/woocommerce/woocommerce-ios/pull/3483/commits/7741e305b1313096b7e9775350f6d3b78d8c1c68)
* I reverted that bump back later, but I guess that didn't revert the bump to WPKit

Anyway, I agree with this:
> while it would be a bit dangerous to include all that if that was a hotfix on an already-published production build, that might not be that big of a deal in the case of a fix on a release/5.8 l

And the three PRs in WPKit you linked to do not seem to affect WooCommerce. 

So I think it's okay to keep it

AliSoftware: Thanks for the answer. Actually keeping that change here will require me to stabilize a 4.25.0 version at the time of actual release on Friday – because we never ship dependencies in beta state to the AppStore – which doesn't seem right (will include even more new `-beta.N` changes, untested with WC 5.8). I'd prefer that we revert to 4.24 if your PR didn't mean any new fix or feature from WPKit in the first place.

ctarda: Sure. Reverted in #3489

AliSoftware:
Thanks for the fix! This PR and its patch #3489 have now been shipped to beta testers as part of new beta 5.8.0.2 🎉

PR #3484, opened 2021-01-17
Refunded Amount Show Up as Positive Instead of Negative

### LINKED ISSUE #3467: Refunded Amount Show Up as Positive Instead of Negative
**Describe the bug**

After submitting a refund, the Refund item shows up on the Order Details page as a **positive value** instead of negative. The positive value gets incorrectly added to the Net amount. This is only temporary because after pulling-to-refresh on the Order Details page, the positive value will be shown as negative (correct). 

**To Reproduce**

1. Open an order.
2. Issue a refund (any amount). 
3. Notice the issued refund amount on the order details page. 

**Screenshots**

https://user-images.githubusercontent.com/198826/104516552-c3f33380-55b1-11eb-9da1-7b479742242d.mov

**Expected behavior**

The Refund item should be shown as negative. Android presents it as negative right away. 

**Mobile Environment**
Please include:
- Device: iPhone 8
- iOS version: 14
- WooCommerce iOS version: 

**WordPress Environment**
<details>
```
```
</details>

ctarda:
Closes #3467 

Fix discrepancies in order details view in between the moment a refund is issued and the moment the order is refreshed with data from the API
  
The issue is that we were always expecting refunds to have a negative value, which is not always the case. The API always seems to return funds with a negative value, but there is a case when a refund can have a positive value: in between the refund is issued in the app and the moment we update the order from the API.
  
 The fix touches the way refunds are calculated and the way they are presented. I don't love this solution, but I tried to submit a refund with a negative value and it didn't work (the API creates the refnd but sets it to zero) and forcing all refunds to be negative at the Store seemed like could be more dangerous.

https://user-images.githubusercontent.com/2722505/104858965-cb3b7980-58f0-11eb-862e-cc8ed9a423c8.mov

  
## Changes
* Make OrderPaymentDetailsViewModel and TotalRefundedCalculationUseCase check that the values of individual refunds are negative, and if they are not, make them negative.

## How to test:

1. Open an order.
2. Issue a refund (any amount).
3. Notice the issued refund amount on the order details page. It should be negative, and the Net line should be reduced by the amount refunded.
4. Repeat with more than one item, if possible.
5. Pull to refresh at any time to pull values from the API. Nothing should change, neither in the individually refunded items, nor the Net amount.
6. Repeat issuing partial refunds and pulling to refresh to see if anything breaks.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
@@ -155,8 +155,8 @@ final class OrderPaymentDetailsViewModel {
             return nil
         }
 
-        // We want the condensed refund total because it's reported as a negative value
-        return currencyFormatter.formatAmount(condensedRefund.total, with: order.currency)
+        // We can not assume the total is negative.
+        return currencyFormatter.formatAmount(condensedRefund.normalizedTotalAsNegative, with: order.currency)
     }
 
     /// Format the net amount with the correct currency
@@ -213,5 +213,18 @@ final class OrderPaymentDetailsViewModel {
 private extension OrderPaymentDetailsViewModel {
     enum Constants {
         static let decimalZero = Decimal(0)
+    }
+}
+
+
+private extension OrderRefundCondensed {
+    /// Present the refund total as a negative number,
+    /// by prefixing it with a minus symbol.
+    var normalizedTotalAsNegative: String {
+        guard total.hasPrefix("-") else {
+            return "-" + total
+        }
+
+        return total
     }
 }
--- a/WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/TotalRefundedCalculationUseCaseTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/TotalRefundedCalculationUseCaseTests.swift
@@ -38,8 +38,7 @@ final class TotalRefundedCalculationUseCaseTests: XCTestCase {
             // Values with currency are probably not possible in the API but we're
             // considering it here for resilience.
             OrderRefundCondensed(refundID: 0, reason: nil, total: "$-8.71972"),
-            // The API returns negative numbers. We're including a positive number to indicate
-            // that we don't do special handling of the sign.
+            // The API returns negative numbers. However, refunds issued but not fetched from the API are positive.
             OrderRefundCondensed(refundID: 0, reason: nil, total: "3719.8850971"),
         ]
         let order = MockOrders().empty().copy(refunds: refundItems)
@@ -50,6 +49,6 @@ final class TotalRefundedCalculationUseCaseTests: XCTestCase {
         let totalRefunded = useCase.totalRefunded()
 
         // Then
-        XCTAssertEqual(totalRefunded, NSDecimalNumber(string: "-43427.9375229"))
+        XCTAssertEqual(totalRefunded, NSDecimalNumber(string: "-50867.7077171"))
     }
 }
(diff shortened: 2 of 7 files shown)


jaclync:
Tricky issue! I didn't know much about the refund models, so spent some time looking into the API response. It looks like the issue came from creating `OrderRefundCondensed` with the response from refund creation API request (`POST /refunds`) - the `amount` has a positive value while the response from `GET orders/\(orderID)` has a negative `amount`. There's probably some reason why the POST response has a positive value, like maybe core uses it to display on a banner. Is it worth asking core about it? Maybe there's no particular reason, and the `amount` field can be updated to be consistent with the [order response](https://woocommerce.github.io/woocommerce-rest-api-docs/#order-refunds-properties).

Given it's a pre-existing API behavior, I was wondering if it makes sense to only upsert `OrderRefundCondensed` when retrieving an order and skip creating the first instance from the refund creation. Like removing this line:

https://github.com/woocommerce/woocommerce-ios/blob/2ac34dd3fb92773e913534fc502865ed1a426cc9/Yosemite/Yosemite/Stores/RefundStore.swift#L178

I tested with this change and it seemed to work. Since we already [retrieve the latest order after creating a refund](https://github.com/woocommerce/woocommerce-ios/blob/2ac34dd3fb92773e913534fc502865ed1a426cc9/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Issue%20Refunds/RefundConfirmationViewModel.swift#L86), the `order.refunds` should be updated quickly following the refund creation under normal network conditions.

ctarda:
Thanks for taking a look @jaclync The thing is, I considered at one point doing exactly as you suggest, but I discarded it for some reason I don't remember. But now that you brought it up, it makes all sense. 

I'll update the PR later today, thanks!

ctarda:
I just pushed an update to remove the upsert operation @jaclync I think this is ready for another round.

WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
```diff
@@ -155,8 +155,8 @@ final class OrderPaymentDetailsViewModel {
             return nil
         }
 
-        // We want the condensed refund total because it's reported as a negative value
-        return currencyFormatter.formatAmount(condensedRefund.total, with: order.currency)
+        // We can not assume the total is negative.
+        return currencyFormatter.formatAmount(condensedRefund.normalizedTotalAsNegative, with: order.currency)
```
jaclync, line 159: now that we don't create `OrderRefundCondensed` with a positive amount, do you think we still want to normalize the amount to negative? I tested with just removing `handleCondensedRefundsIfNeeded` and that seems enough 🙂

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/TotalRefundedCalculationUseCaseTests.swift
```diff
@@ -50,6 +49,6 @@ final class TotalRefundedCalculationUseCaseTests: XCTestCase {
         let totalRefunded = useCase.totalRefunded()
 
         // Then
-        XCTAssertEqual(totalRefunded, NSDecimalNumber(string: "-43427.9375229"))
+        XCTAssertEqual(totalRefunded, NSDecimalNumber(string: "-50867.7077171"))
```
jaclync, line 52: just curious, how did this number change from the changes?

jaclync (approved):
Thanks for the update! Just had a question about whether to keep the changes to normalize the refund amount to negative, up to you!

WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
ctarda: I guess it doesn't hurt to be extra safe?

WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue Refund/TotalRefundedCalculationUseCaseTests.swift
ctarda: Previously, the test was not completely accurate, and it was actually passing for the case where the bug happens: if a refund was positive, the test would substract the amount refunded from the refund total, instead of adding it to it

PR #3485, opened 2021-01-18
Adds support to Typed Predicates

Ecarrion:
part of #3465 

# Why
After [adding unit tests on our predicate methods](https://github.com/woocommerce/woocommerce-ios/pull/3480) we can proceed to add the necessary tools that will allow us to create type-safe predicates. 

The next PR will replace our current predicates with safe-predicates using the tools added here.

# How
This PR adds a set of protocols, classes, and operator overloads in order to provide a type-safe API to create `NSPredicates`. 

This is achieved by overloading logical operators between a `Swift.KeyPath` which is type-safe and a `Value`. 
`KeyPaths` and `Values` can be transformed into `NSExpression` via a native `APIs` which then can be converted to an `NSComparisonPredicate` using the `init(leftExpression, rightExpression, operator)` method. This will allow us to support the `EQUAL`, `NOT EQUAL`, and `IN` predicates.

Additionally, we can group multiple `NSComparisonPredicate` into `NSCompoundPredicates` that will add support to `AND` and `OR` predicates.

# Testing Steps
- 👀 to the code and tests

Update release notes:

### DIFF
--- a/Storage/Storage/Tools/TypedPredicates.swift
+++ b/Storage/Storage/Tools/TypedPredicates.swift
@@ -33,12 +33,6 @@ public func || (p1: TypedPredicate, p2: TypedPredicate) -> CompoundPredicate {
     CompoundPredicate(type: .or, subpredicates: [p1, p2])
 }
 
-/// Overloads the `NOT` operator for a single predicate, returns a `CompoundPredicate` that evaluates predicates with the `!` rule.
-///
-public prefix func ! (p: TypedPredicate) -> CompoundPredicate {
-    CompoundPredicate(type: .not, subpredicates: [p])
-}
-
 // MARK: - Comparison Operators Overloads
 
 /// Overloads for comparison  operators.  Each operator takes a `KeyPath` and a `Value`
--- a/Storage/Storage/Tools/TypedPredicates.swift
+++ b/Storage/Storage/Tools/TypedPredicates.swift
@@ -54,6 +54,13 @@ public func == <RootType, ResultingType: Equatable>(keyPath: KeyPath<RootType, R
     ComparisonPredicate(keyPath, .equalTo, value)
 }
 
+/// Defines a new  operator `equal insensitive`  between a `KeyPath` and a `Value`. Returns a `ComparisonPredicate` that evaluates the parameters using the `==[c]` rule.
+///
+infix operator =~ : ComparisonPrecedence
+public func =~ <RootType, ResultingType: Equatable>(keyPath: KeyPath<RootType, ResultingType>, value: ResultingType) -> ComparisonPredicate {
+    ComparisonPredicate(keyPath, .equalTo, value, .caseInsensitive)
+}
+
 /// Overloads the `not equal` operator between a `KeyPath` and a `Value`. Returns a `ComparisonPredicate` that evaluates the parameters using the `!=` rule.
 ///
 public func != <RootType, ResultingType: Equatable>(keyPath: KeyPath<RootType, ResultingType>, value: ResultingType) -> ComparisonPredicate {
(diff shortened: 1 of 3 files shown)


Storage/Storage/Tools/TypedPredicates.swift
```diff
@@ -0,0 +1,80 @@
...
+/// Each operator takes `TypedPredicate` instances and returns a `CompoundPredicate` that evaluate the predicates using their respective logical rule.
+///
+
+/// Overloads the `AND` operator between two predicates, returns a `CompoundPredicate` that evaluates the two predicates with the `&&` rule.
+///
+public func && (p1: TypedPredicate, p2: TypedPredicate) -> CompoundPredicate {
+    CompoundPredicate(type: .and, subpredicates: [p1, p2])
+}
+
+/// Overloads the `OR` operator between two predicates, returns a `CompoundPredicate` that evaluates the two predicates with the `!!` rule.
```
jaclync: nit:

```suggestion
/// Overloads the `OR` operator between two predicates, returns a `CompoundPredicate` that evaluates the two predicates with the `||` rule.
```

Storage/Storage/Tools/TypedPredicates.swift
```diff
@@ -0,0 +1,80 @@
...
+
+/// Overloads the `not equal` operator between a `KeyPath` and a `Value`. Returns a `ComparisonPredicate` that evaluates the parameters using the `!=` rule.
+///
+public func != <RootType, ResultingType: Equatable>(keyPath: KeyPath<RootType, ResultingType>, value: ResultingType) -> ComparisonPredicate {
+    ComparisonPredicate(keyPath, .notEqualTo, value)
+}
+
+/// Overloads the  `IN` operator between a `KeyPath` and a `[Value]`. Returns a `ComparisonPredicate` that evaluates the parameters using the `===` rule.
+///
+public func === <RootType, ResultingType: Equatable>(keyPath: KeyPath<RootType, ResultingType>, value: Array<ResultingType>) -> ComparisonPredicate {
```
jaclync, line 72: TIL'ed this IN operator!

Storage/StorageTests/Tools/TypedPredicateTests.swift
```diff
@@ -0,0 +1,43 @@
+import XCTest
+@testable import Storage
+
+final class TypedPredicateTests: XCTestCase {
```
jaclync, line 4: nice tests!! 💯

jaclync (approved):
Nice work, this is really cool having type safe predicates! 😮 🚀

Storage/Storage/Tools/TypedPredicates.swift
```diff
@@ -0,0 +1,80 @@
...
+
+// MARK: - Convenience Initializers
+
+internal extension ComparisonPredicate {
+    /// Returns a `ComparisonPredicate` by converting the parameters into `NSExpression` instances using the provided `Operator` as a modifier.
+    ///
+    convenience init<RootType, ResultingType>(_ keyPath: KeyPath<RootType, ResultingType>, _ operator: NSComparisonPredicate.Operator, _ value: Any?) {
+        let keyPathExpression = NSExpression(forKeyPath: keyPath)
+        let valueExpression = NSExpression(forConstantValue: value)
+        let options = ResultingType.self == String.self ? NSComparisonPredicate.Options.caseInsensitive : []
```
jaclync: While reviewing https://github.com/woocommerce/woocommerce-ios/pull/3487, I noticed that some string comparisons aren't case insensitive (like `loadTaxClass(slug:)`) - is it possible to override `NSComparisonPredicate.Options` in some way?

Ecarrion: Wow, great catch I failed to notice that case! To solve it I have added a new operator `=~` that performs a `==` comparison with the `case insensitive` option. https://github.com/woocommerce/woocommerce-ios/pull/3485/commits/de1a07cfc6033c807ec233197bf89c7a23fb32e4

PR #3486, opened 2021-01-18
Core Data: Add MigrationStep to Encapsulate Index-Based Loops

shiki:
Part of fixing #2667. Continuation of #3267.

Like #3267, this is another digestible chunk of the [WIP branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/2667-reduce-file-operations-in-core-data) which reduces the `FileManager` operations in our Core Data stack. I've been trying to simplify the stack and add more unit tests to reduce the chances of regressions. 

I think the next PR will be the last one, which removes the `FileManager` operations. 

## Changes

This PR removes this index-based operation from the `CoreDataIterativeMigrator`: 

https://github.com/woocommerce/woocommerce-ios/blob/c3d5ad8d0ee956652cc57002d8ba4a27b17d0688/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L63-L68

These variables and their functions have also been removed from the `CoreDataIterativeMigrator`:

https://github.com/woocommerce/woocommerce-ios/blob/c3d5ad8d0ee956652cc57002d8ba4a27b17d0688/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L53-L56

These are now encapsulated inside a new `struct`, [`MigrationStep`](https://github.com/woocommerce/woocommerce-ios/blob/8fa4773ee1dca0019084efa1b0d177e289126533/Storage/Storage/CoreData/CoreDataIterativeMigrator%2BMigrationStep.swift#L6-L27). 

As a result, there is now a logical split: 

- `MigrationStep`: The **what**. Answers what `NSManagedObjectModels` are involved to upgrade a store to the latest version. 
- `CoreDataIterativeMigrator`:  The **how**. Blindly follows the steps given by [`MigrationStep.steps()`](https://github.com/woocommerce/woocommerce-ios/blob/8fa4773ee1dca0019084efa1b0d177e289126533/Storage/Storage/CoreData/CoreDataIterativeMigrator%2BMigrationStep.swift#L21-L27) and focuses on File and `NSMigrationManager` operations. 

## Testing

2. Using this branch, delete models 41 and 42. 
3. Set the current model version to 40. 
5. Stop the app. 
6. Restore the deleted model versions and reset the current model version to 42. 

    https://github.com/woocommerce/woocommerce-ios/blob/8fa4773ee1dca0019084efa1b0d177e289126533/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L68

8. Run the app again. Migrations should happen. Confirm that the breakpoint above was hit twice. And the `migrationAttemptMessage` variable contain these values:
    - `⚠️ Attempting migration from Model 40 to Model 41`
    - `⚠️ Attempting migration from Model 41 to Model 42`
9. Confirm that you can load the Orders and Products. 

## Submitter Checklist

### DIFF
--- a/Storage/Storage.xcodeproj/project.pbxproj
+++ b/Storage/Storage.xcodeproj/project.pbxproj
@@ -49,6 +49,7 @@
 		454005AA24AE4D350087FDD1 /* WooCommerceModelV28toV29.xcmappingmodel in Sources */ = {isa = PBXBuildFile; fileRef = 454005A924AE4D350087FDD1 /* WooCommerceModelV28toV29.xcmappingmodel */; };
 		45E1862E2370450C009241F3 /* ShippingLine+CoreDataClass.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45E1862D2370450C009241F3 /* ShippingLine+CoreDataClass.swift */; };
 		45E1863023704519009241F3 /* ShippingLine+CoreDataProperties.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45E1862F23704519009241F3 /* ShippingLine+CoreDataProperties.swift */; };
+		5707FBC425B5FE2200B7C1A6 /* CoreDataIterativeMigrator+MigrationStep.swift in Sources */ = {isa = PBXBuildFile; fileRef = 5707FBC325B5FE2200B7C1A6 /* CoreDataIterativeMigrator+MigrationStep.swift */; };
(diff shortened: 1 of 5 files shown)


Ecarrion (approved):
Nice to see this code converted into easier to read chunks. I remember I had a hard time following `CoreDataIterativeMigrator` the first time I read it. 💪 

Also, testing steps work as expected! 🚀

shiki:
Thanks, Ernesto!

PR #3488, opened 2021-01-19
Product Settings > Visibility: hide the separator between password visibility and password text field row

### LINKED ISSUE #2274: Remove the separator between Password protected and Password in Product Visibility
- [ ] Remove the separator between Password protected and Password rows in Product Visibility

Screen Shot 2020-05-11 at 09 53 44

jaclync:
Fixes #2274 

## Why

The goal is to hide the password visibility cell separator when it is selected since the password text field is shown below. @rachelmcr noticed a challenging issue if a cell is first configured with `hideSeparator()` and then later changed to `showSeparator()` (ref https://github.com/woocommerce/woocommerce-ios/issues/2274#issuecomment-760948136) - a few of us looking into this and the usage of [`CGFloat.greatestFiniteMagnitude`](https://developer.apple.com/documentation/coregraphics/cgfloat/1845207-greatestfinitemagnitude) in the separator inset is causing this issue. This PR addresses this unexpected issue by replacing `.greatestFiniteMagnitude` with a large number. Feel free to suggest any other ideas!

## Changes

- In `ProductVisibilityViewController`, separated cell configuration for `Row.passwordVisibility` so that we show/hide the cell separator when it is deselected/selected, respectively.
- In `UITableViewCell.hideSeparator()`, replaced `.greatestFiniteMagnitude` right inset with a large number - I've tried `.greatestFiniteMagnitude - 1` but it did not work 😐 

### Scenario 1: product is originally not password protected

- Go to the products tab
- Tap on a product with public or private visibility
- Tap the ellipsis menu, then tap "Product Settings"
- Tap "Visibility" row when it becomes tappable --> the "Password Protected" row should have a separator
- Tap "Password Protected" row --> the password text field should appear, and "Password Protected" row should not have a separator
- Feel free to tap around to verify that the separator is hidden when "Password Protected" is selected

### Scenario 2: product is originally password protected

- Go to the products tab
- Tap on a product that is password protected
- Tap the ellipsis menu, then tap "Product Settings"
- Tap "Visibility" row when it becomes tappable --> the "Password Protected" row should not have a separator and the password text field is shown below
- Tap "Public" or "Private" row --> the password text field should be hidden, and "Password Protected" row should now have a separator
- Feel free to tap around to verify that the separator is hidden when "Password Protected" is selected

## Example screenshots

product is not password protected | product is password protected
-- | --
Simulator Screen Shot - iPhone 11 - 2021-01-20 at 11 16 43 | Simulator Screen Shot - iPhone 11 - 2021-01-20 at 11 16 52
Simulator Screen Shot - iPhone 11 - 2021-01-20 at 11 15 22 | Simulator Screen Shot - iPhone 11 - 2021-01-20 at 11 15 27
 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UITableViewCell+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UITableViewCell+Helpers.swift
@@ -22,7 +22,9 @@ extension UITableViewCell {
     /// Be careful applying this to a reusable cell where the separator is expected to be shown in some cases.
     ///
     func hideSeparator() {
-        separatorInset = UIEdgeInsets(top: 0, left: separatorInset.left, bottom: 0, right: .greatestFiniteMagnitude)
+        // Using `CGFloat.greatestFiniteMagnitude` for the right separator inset would not work if the cell is initially configured with `hideSeparator()` then
+        // updated with `showSeparator()` later after the table view is rendered - the separator would not be shown again.
+        separatorInset = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 999999)
     }
 
     /// Shows the separator for a cell.
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/Visibility/ProductVisibilityViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Product Settings/Visibility/ProductVisibilityViewController.swift
@@ -57,6 +57,11 @@ final class ProductVisibilityViewController: UIViewController {
         }
     }
 
+    override func viewLayoutMarginsDidChange() {
+        super.viewLayoutMarginsDidChange()
+        reloadSections()
+    }
+
     private func reloadSections() {
         if visibility == .passwordProtected {
             sections = [Section(rows: [.publicVisibility, .passwordVisibility, .passwordField, .privateVisibility])]
@@ -212,7 +217,7 @@ private extension ProductVisibilityViewController {
         cell.selectionStyle = .default
         cell.textLabel?.text = row.description
         cell.accessoryType = row.visibility == visibility ? .checkmark : .none
-        cell.showSeparator()
+        cell.showSeparator(inset: .init(top: 0, left: tableView.layoutMargins.left, bottom: 0, right: 0))
     }
 
     func configurePasswordVisibilityCell(cell: BasicTableViewCell, indexPath: IndexPath) {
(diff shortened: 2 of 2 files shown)


pmusolino (commented):
I think my hawkeye is more alert than usual today 🤣 

While I was testing the app, I noticed that the first separator is more faded if compared to the second one. So I inspected them, and this is how the separators appear:

First separator:
Schermata 2021-01-19 alle 13 04 02

Second separator:
Schermata 2021-01-19 alle 13 04 13

I honestly don't know what is causing this issue. 

I also noticed using the View Hierarchy Inspector that it seems like if there is another separator between each cell with background color equal to `nil`.

In this screenshot, they are highlighted in blue, but I don't know what they are due to ...
Schermata 2021-01-19 alle 13 12 06

Any idea? 🤔

jaclync:
Hey @pmusolino thanks for testing! I don't really trust the Xcode view debugger for the cell separator anymore from the Slack discussion (the separator is shown in view debugger after being hidden in the first place, while it isn't visible in the UI). I don't have a good explanation for the view debugger discrepancies 😞 

Re different separator colors: I checked on my simulator (I zoomed in a bit so that each 0.5px separator is shown as a full pixel in the computer screen) and the color of the two separators is the same on my end. The `UITableViewCell.showSeparator` and `UITableViewCell.hideSeparator` functions don't change the separator color, I can't think of a reason how these calls result in a different color 🤔 . Sometimes these thin 0.5px separators are shown with different edges depending on the zoom scale, maybe you can check the other screens with separators (like Product Settings > Status or Catalog Visibility) and see if you only see this color difference in the Visibility screen?

jaclync:
After testing again, I noticed a small misalignment between the password text field row and the other visibility rows (screenshots below). Since the password text field row is a different cell type and does not call `showSeparator` (since it's never hidden), its separator uses the cell's layout margins left inset. The cell content also has leading alignment based on the layout margins. In https://github.com/woocommerce/woocommerce-ios/pull/3488/commits/3b90333a03f3da5912e3357e1b74d43daa19497f, I updated the cells to use the table view's layout margins left inset and reload the table view on layout margins change. From testing, the table view's layout margins are updated after `viewDidLoad` - it starts with 8px on all edges and then changes to 20px on left/right edges.

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2021-01-20 at 10 17 48 | Simulator Screen Shot - iPhone 11 - 2021-01-20 at 11 06 25

pmusolino:
Nice catch @jaclync!

pmusolino (approved):
Also if there are some obscure things, the PR LGTM :shipit:

jaclync:
Thanks for the review, Paolo! Merging now

PR #3491, opened 2021-01-19
Fix: product is now read-only when opened from the order details

### LINKED ISSUE #3274: Order: product is read-only when opened from the order and read-write when opened from order details
**Describe the bug**
When the order is in [failed, on hold, pending] state, there is a "details" row below the list of products on the order screen.
If you tap on the product on the order screen, it opens read only view.
If you tap on the product from details screen, it opens read-write view

**To Reproduce**
1. Go to an order
2. Tap on a product, it opens read-only view
3. Close product details
4. Tap on "Details" row below the list of products
5. Tap on a product, it opens read-write view

**Expected behavior**
I believe the behavior should be consistent and both actions should open read-only view of the product.

**Mobile Environment**
Please include:
- Device: iPhone 12
- iOS version: iOS 14.2
- WooCommerce iOS version: 5.6

pmusolino:
Fixes #3274 

## Description
Previously, when the order is in [failed, on hold, pending] state, there is a "details" row below the list of products on the order screen.
If you tap on the product on the order screen, it opens read only view.
If you tap on the product from details screen, it opens read-write view.
For creating a consistent behavior, both actions should open read-only view of the product.

## Testing
1. Go to an order.
2. Tap on a product, it opens read-only view.
3. Close product details.
4. Tap on "Details" row below the list of products.
5. Tap on a product, it opens read-only view.

Update release notes:


Ecarrion (approved):
Working like a charm!

PR #3496, opened 2021-01-19
Fix 5.8 metadata

jkmassel:
This PR fixes the 5.8 metadata issues cause by the `download_metadata.swift` script deleting the whole `metadata` directory each time.

Note: The Apple TV privacy policy text is all missing, but we don't have an app for that platform, so I don't mind just leaving it blank for now to avoid noise in future diffs

**To Test:**
- Run `cd fastlane && ./download_metadata.swift`. Ensure that the repo is clean.


fastlane/download_metadata.swift
```diff
@@ -105,22 +105,20 @@ func downloadTranslation(languageCode: String, folderName: String) {
         let fileManager = FileManager.default
         try? fileManager.createDirectory(atPath: languageFolder, withIntermediateDirectories: true, attributes: nil)
 
-        /// This sort of hack-ey – we're not actually downloading anything here, but the idea is
```
jkmassel, line 108: This is removed since we've downloaded it from ASC and it won't be deleted anymore now.

fastlane/download_metadata.swift
```diff
@@ -68,7 +68,7 @@ func downloadTranslation(languageCode: String, folderName: String) {
 
         jsonDict.forEach({ (key: String, value: Any) in
 
-            guard let index = key.index(of: Character(UnicodeScalar(0004))) else {
+            guard let index = key.firstIndex(of: Character(UnicodeScalar(0004))) else {
```
AliSoftware, line 71: Finally, bye bye old swift warning!

fastlane/metadata/en-US/marketing_url.txt
```diff
@@ -1 +1 @@
-https://woocommerce.com/mobile/
\ No newline at end of file
+https://woocommerce.com/mobile/
```
AliSoftware: Must be a sneaky difference like CRLF at end of file… though if it were GH would put a ⛔️ icon to indicate that… so not sure what diff it's seeing…🤷‍♂️   Oh well at least that gives me an excuse to post some meme 😄 

image

AliSoftware (approved):
LGTM!

Tested by patching `new_beta_release` to only call `ios_update_metadata` then running the patched lane. That way I ensured that not only that script but also the rest of what this action does doesn't break the metadata again 👍 

Thanks for the fix! :shipit:

PR #3495, opened 2021-01-19
Merge/release 5.8 to develop after beta (fix conflict)

AliSoftware:
Replaces #3490 to fix pbxproj merge conflict. It contains:

 - The release-toolkit update from #3432
 - The fix #3483 + its patch #3489
 - The update of all the Localized strings pulled from GlotPress
 - [UPDATE] Also includes the fixes #3496 + #3497 to fix the metadata oddities which were present in this PR before (see below for details).

```
                                          #3496  #3497
                     #3483   #3489          ▼      ▼
                        ▼       ▼         ○──○   ○──○
╔═════════════╗          ╲       ╲       ╱    ╲ ╱    ╲
║ release/5.8 ║─ ─○─ ─ ─ ─●───────●─────○──────○──────○─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ▶
╚═════════════╝        5.8.0.1  5.8.0.2  ╲             ╲
                ╱                   ╲     ╲             ╲
                                     ╲     ╲             ╲  ┌────────────────────────────┐
              ╱                     ○━◎━━━○━○━━━━━━━━━━━━━○ │merge/release-5.8-to-develop│
                                   ╱  fix                   └────────────────────────────┘
            ╱                     ╱ conflict                ╲
╔═══════╗                        ╱
║develop║─○───────────────○──○──○─────────────────────────────◌─────────────▶
╚═══════╝
```

----
### Note about metadata changes

<details><summary>Old notes about metadata oddity in this PR – fixed since</summary>

⚠️ Note: that PR seems to have also removed a lot of `fastlane/metadata/*` files in [during `new_beta_release` when updating metadata translation](https://github.com/woocommerce/woocommerce-ios/pull/3495/commits/d59a50a2bb595c412f2a2d4a52c5296919c73ecb) which seems odd 🤔  _(Tried to [re-run `download_metadata.swift` separately](https://github.com/woocommerce/woocommerce-ios/pull/3495/commits/24c075711bfbc3644ca10e4dee35dafdb5a95bab) to no avail)_

@jkmassel I think that might be due to your recent PR updating the scripts and metadata? Is that expected?
For the `fastlane/metadata/*/release_notes.txt` files I can guess that it's probably because the Release Notes are not yet translated in GlotPress. But for the other metadata files, I'm a bit more confused.

[...]

PR #3497, opened 2021-01-19
Put the App Store Screenshots Labels back

jkmassel:
Re-adds the screenshot text assets to reduce the diff noise between the `release/5.8` branch and `develop`.


AliSoftware:
@jkmassel Not sure why you kept this PR as Draft 🤔  Is it because of the slight diff on line endings that it introduces that you want to solve?

Given that I think the release -> develop resync PR has waited a bit much by now – I prefer to resync often and resync again if needed, rather than let it get stale – I'll merge this PR into `release/5.8` then update #3495 with it so it can be merged into `develop` sooner rather than later. But feel free to make any additional PR to finish what you wanted to fix if there was something you still wanted to address that made you keep this as Draft, we still have time to revisit.

AliSoftware (approved):
LGTM.

I ran the `download_app_store_translations` lane on top of this branch and it gave me almost the same content as what you have except without the line endings diffs you had when you ran it on your Mac (as we discussed in Slack). I figured that's the reason why you kept this PR as Draft.

Given that the line endings I got when I ran the script match what we have always used for those files – making those not appear as a diff in #3495 anymore, I figured it was better to push those changes and the line endings I got – see 729865cab62c2020879deb25490dae96f63f95cc, and I'll merge with those changes so we can then get #3495 merged and resync with `develop` sooner than later. We can always consider investigating this line ending mismatch (depending on the Mac running the scripts?) in a separate PR.

PR #3498, opened 2021-01-20
Fix uninitialized tab bar UI briefly shown when launching in logged out state or restarting authentication

### LINKED ISSUE #3107: Launching the app in logged out state shows uninitialized tab bar UI first in iOS 14

### LINKED ISSUE #3429: After tapping "Try another account", uninitialized tab bar UI is shown briefly before login prologue screen
**Describe the bug**

Similar to https://github.com/woocommerce/woocommerce-ios/issues/3107, presenting login UI from WPAuthenticator library tends to have some noticeable delay. In this case, it is when the user decides to restart the authentication flow by tapping "Try another account" on the store picker and the app dismisses the login UI then presents login UI again. The reason why we can't just reuse the login UI and go back to the first screen is because the prologue screen might not exist anymore when logging in with magic link (https://github.com/woocommerce/woocommerce-ios/pull/3423).

**To Reproduce**

1. Start logged out of the app
2. Select the WordPress.com login option
3. Log in with email and password or magic link
4. On the store picker screen, select "Try another account" to log in to a different account --> the uninitialized tab bar UI is shown for a split second before the prologue screen is shown

**Screenshots**

https://user-images.githubusercontent.com/1945542/103722202-676ca300-500a-11eb-82f3-fd35f11872e4.mov

**Expected behavior**

The uninitialized tab bar should not be shown when presenting the login UI.

**Mobile Environment**
Please include:
- Device: iPhone 11 sim, iPhone 6s Plus
- iOS version: iOS 14.1
- WooCommerce iOS version: 5.7

jaclync:
Fixes #3107 
Fixes #3429 

- [x] ⚠️ Update WPAuthenticator pod to release version after https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/564 is merged ⚠️ 
- [x] ⚠️ Update release notes if the PR is targeting 6.0 ⚠️ 

## Why

Because the login UI is initialized from WPAuthenticator pod, the time it takes to load the login view controller to show on launch or from store picker is significant so that a glitch is noticeable in these scenarios:

- Launching the app in logged out state ([example screencast in .gif](https://user-images.githubusercontent.com/1945542/98314874-de3cee00-2011-11eb-984c-9401939688fb.gif))
- Tapping "Try another account" on the store picker ([example screencast in .mov](https://user-images.githubusercontent.com/1945542/103722202-676ca300-500a-11eb-82f3-fd35f11872e4.mov))

If the view controller isn't from WPAuthenticator pod (like if we just show a blank `UIViewController` where we show the login UI), such glitch does not happen. I've suspected the slowness partially comes from instantiating the login view controller from a giant Storyboard (Login.storyboard), I did give it a try to move the initial login view controller off of the storyboard but it only helped save ~12% of the time. There are other factors that contribute to the loading time, but I wasn't able to identify all of them. WordPress iOS app also has the same glitch https://github.com/wordpress-mobile/WordPress-iOS/issues/15399.

In order to remove the glitch, I settled on the solution to set the `UIWindow`'s `rootViewController` based on logged in state instead of always having `MainTabBarController` as the root view controller. This is a significant change to our app, I've tested a bunch and I listed possible scenarios for testing in this PR.

## Changes

[Interactive graph](https://viewer.diagrams.net/?highlight=B266FF&edit=_blank&layers=1&nav=1&title=App%20coordinator#Uhttps%3A%2F%2Fdrive.google.com%2Fuc%3Fid%3D1IOllkEUeZSKbkK8-tpry8Y5zlAEs79n9%26export%3Ddownload)

App coordinator

There are two signals that we now subscribe: "whether the user has entered credentials" and "whether a store needs to be selected" as in the graph above. In this PR, `AppCoordinator` now sets the UI state (`UIWindow.rootViewController`) based on the graph - login flow, store picker (if login flow isn't present), and tab bar UI.

Major changes include:
- Updated `StoresManager` protocol's `isLoggedIn` observable to Combine, and added a new publisher `needsDefaultStorePublisher` that is derived from `SessionManager`'s `defaultStoreIDPublisher` (added in this PR).
- Updated `Authentication` protocol's `displayAuthentication` to return the login view controller instead of presenting the login view controller from WPAuthenticator, so that `AppCoordinator` can set the login view controller to window's root.
  - WPAuthenticator PR: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/564
- In `AppCoordinator`, it now takes in a `UIWindow` as the app's main window and sets the window's root view controller based on `StoresManager`'s `isLoggedInPublisher` and `needsDefaultStorePublisher` using `CombineLatest` operator.

### Why does `AppCoordinator` have to subscribe to two signals while it only needs the `isLoggedIn` signal before?

Before this PR, the tab bar UI is always the window's root view controller. Therefore, `AppCoordinator` only had to handle UI changes when the app is not logged in or missing the selected store. Now that `AppCoordinator` is only showing the tab bar UI when the app is **logged in with a selected store**, it needs to also know when a store has been selected by the user (`needsDefaultStorePublisher`) in addition to having the credentials (`isLoggedInPublisher`).

### Log in to a store, then log out

- Launch the app logged out --> there should not be a glitch (uninitialized tab bar UI) as in `develop` before login UI is shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Select a store (if more than one) and tap "Continue" --> the tab bar UI should be shown
- Relaunch the app --> the tab bar UI should be shown
- Log out from the app --> the login UI should be shown with cross fade animation (I didn't look into how to implement the previous modal presentation animation, but we can try that if we prefer the modal animation to cross fade)

### Log in with credentials, then restart login flow

- Launch the app logged out --> there should not be a glitch as in `develop` before login UI is shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Tap "Try another account" --> there should not be a glitch before login UI is shown

### Log in with credentials, force close the app, then continue with a store

- Launch the app logged out --> there should not be a glitch as in `develop` before login UI is shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Force close the app and relaunch (relaunch the app from Xcode while the app is still on the store picker screen) --> the store picker should be shown without a glitch
- Select a store (if more than one) and tap "Continue" --> the tab bar UI should be shown

### Log in with credentials, force close the app, then restart login flow

- Launch the app logged out --> there should not be a glitch as in `develop` before login UI is shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Force close the app and relaunch (relaunch the app from Xcode while the app is still on the store picker screen) --> the store picker should be shown without a glitch
- Tap "Try another account" --> the login UI is shown (there is a glitch when it animates to the login UI, but it requires changes in the store picker so I decided to leave it as a subtask in https://github.com/woocommerce/woocommerce-ios/issues/2888 for now)

## Example screencasts

For design reviewers: please note that the animation when logging out from the app (the first screencast) is now changed to cross fade. If we want to bring back the modal presentation animation, we can look into that separately.

Log in to a store, then log out (#3107) | Log in with credentials, then restart login flow (#3429)
-- | --
https://user-images.githubusercontent.com/1945542/105166123-1b654680-5b52-11eb-99ed-88d1681e76ff.mov | https://user-images.githubusercontent.com/1945542/105165992-f7096a00-5b51-11eb-8bb8-cd2fe7287b40.mov

Update release notes:


Podfile
```diff
@@ -38,9 +38,9 @@ target 'WooCommerce' do
   pod 'Gridicons', '~> 1.0'
 
   # To allow pod to pick up beta versions use -beta. E.g., 1.1.7-beta.1
-  pod 'WordPressAuthenticator', '~> 1.34.0-beta'
+  # pod 'WordPressAuthenticator', '~> 1.34.0-beta'
   # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :commit => ''
-  # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => ''
+  pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => 'wcios-3107/login-ui'
```
Ecarrion: [Q]: Is it ok to point this pod to any branch? is there a plan to merge and release it soon? What would happen if those changes don't get approval and our code gets released?

WooCommerce/Classes/AppDelegate.swift
```diff
@@ -36,7 +36,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     /// Tab Bar Controller
     ///
     var tabBarController: MainTabBarController? {
-        return window?.rootViewController as? MainTabBarController
+        appCoordinator?.tabBarController
```
Ecarrion, line 41: 👏

WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -110,16 +110,19 @@ class AuthenticationManager: Authentication {
...
             guard let self = self else { return }
             // Resets Apple ID at the beginning of the authentication.
             self.appleUserID = nil
 
             ServiceLocator.analytics.track(.loginPrologueContinueTapped)
-        }, onCompletion: onCompletion)
+        }) else {
+            fatalError("Cannot instantiate login UI from WordPressAuthenticator")
+        }
+        return loginViewController
```
Ecarrion: nit: I found this a little hard to read. I believe it's because the `else` condition from the the `guard` statement looks like another completion block.

What do you think about something like this:

```swift
let loginVC = WordPressAuthenticator.loginUI(onLoginButtonTapped: { [weak self] in
    // Resets Apple ID at the beginning of the authentication.
       self.appleUserID = nil

        ServiceLocator.analytics.track(.loginPrologueContinueTapped)
    })

    guard let loginViewController = loginVC else {
       fatalError("Cannot instantiate login UI from WordPressAuthenticator")
    }

    return loginViewController
}    
```

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
```diff
@@ -1,66 +1,97 @@
...
+                        self.displayAuthenticator()
+                    }
+                } else {
+                    if needsDefaultStore {
+                        self.displayStorePicker()
+                    } else {
+                        self.window.rootViewController = self.tabBarController
+                    }
+                }
+                self.isLoggedIn = isLoggedIn
```
Ecarrion: I found myself backtracking several conditions when reading these rules, do you think converting it to a switch will make it easier to read?

```swift
                switch (isLoggedIn, needsDefaultStore) {
                case (false, true):
                    self.displayAuthenticator()
                case (true, true):
                    self.displayStorePicker()
                case (true, false):
                    self.window.rootViewController = self.tabBarController
                default:
                    break
                }
```

Ecarrion (approved):
Thanks for handling this @jaclync!

I got a few suggestions so let me know if you think are valid!
Also, while testing I noticed that we don't have any animation when going from the store picker to the main tab bar, do you think this should be something that we need to add right now? I'm fine if you prefer to tackle it in a different PR.

Podfile
jaclync: Great question! No, we generally only point a pod to a release version (available via CocoaPods CDN) when merging a PR. During code review, it is okay to point a pod to a branch, make a PR for the library separately, and point the pod to release version before merging. More details on shipping a pod release: paNNhX-33-p2

Sorry for causing confusion - I didn't get time to create a PR in WPAuthenticator yesterday, also wanted to see how CR feedback is. Just opened a PR https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/564 and tested in WPiOS https://github.com/wordpress-mobile/WordPress-iOS/pull/15682 🙂

WooCommerce/Classes/Authentication/AuthenticationManager.swift
jaclync: Good point! I originally wanted to make `WordPressAuthenticator.loginUI` non-nil but that'd affect how the existing `WordPressAuthenticator.showLogin` which WPiOS still uses. Updated in: https://github.com/woocommerce/woocommerce-ios/pull/3498/commits/b77707e2bcdb0616b88b98c6e808beebd4d5b738

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
jaclync: This looks better to read! Also added a comment to refer to this PR (hopefully the diagram helps) in: https://github.com/woocommerce/woocommerce-ios/pull/3498/commits/b77707e2bcdb0616b88b98c6e808beebd4d5b738

WooCommerce/Classes/AppDelegate.swift
jaclync: I didn't realize this `tabBarController` is actually used in a bunch of places 😅

jaclync:
Thank you @Ecarrion for the prompt review & testing!! I updated the PR for another look 👀 

> Also, while testing I noticed that we don't have any animation when going from the store picker to the main tab bar, do you think this should be something that we need to add right now? I'm fine if you prefer to tackle it in a different PR.

Great suggestion! I added a function `AppCoordinator.setWindowRootViewControllerAndAnimateIfNeeded` to animate window transition if the root VC isn't nil before in https://github.com/woocommerce/woocommerce-ios/pull/3498/commits/b77707e2bcdb0616b88b98c6e808beebd4d5b738 - now it should also cross fade to the main tab bar 😄

Ecarrion (approved):
Tested again and it's working like a charm with the new transition! 🚀

Garance91540 (approved):
👏

jaclync:
I updated to point the WPAuthenticator pod to release version https://github.com/woocommerce/woocommerce-ios/pull/3498/commits/58bb21180eed2bb7ccd3e6ef11b7151b301ac740, merging now for 5.9!

PR #3500, opened 2021-01-20
Removed all the use cases of "segue" from the app

### LINKED ISSUE #2349: Remove the use of "segue" in the points of the app where they still exist
I just noticed that in some points of the app, like in the Order Detail, we use some `segue`, like `ShowProductListViewController` and `ShowOrderStatusListViewController`.

It would be a good thing to remove these `segue` and replace them by manually presenting the view controller like in the rest of the app.

pmusolino:
Fixes #2349 

## Description
In this PR, I removed the use of "segue" from the app, by removing the 4 segue still present within the code.
In particular, I removed these one:
- [x] ShowProductListViewController
- [x] ShowOrderStatusListViewController
- [x] ShowApplicationLogViewController
- [x] ShowApplicationLogDetailViewController

Now all the view controllers are presented via code.

#### Case 1
1. Open an order with some products inside. 
2. Tap "Details".
3. Make sure the list of products still appears like before.

#### Case 2
1. Open Settings.
2. Tap the cell "Help & Support"
3. Select "View Application Log" -> The Application Logs list should appear.
4. Select one of the application logs. -> The application log detail VC should appear.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
@@ -416,10 +416,3 @@ extension OrderDetailsViewModel {
         stores.dispatch(deleteTrackingAction)
     }
 }
-
-private extension OrderDetailsViewModel {
-
-    enum Constants {
-        static let orderStatusListSegue = "ShowOrderStatusListViewController"
-    }
-}
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
@@ -236,7 +236,12 @@ extension OrderDetailsViewModel {
             viewController.navigationController?.pushViewController(billingInformationViewController, animated: true)
         case .details:
             ServiceLocator.analytics.track(.orderDetailProductDetailTapped)
-            viewController.performSegue(withIdentifier: Constants.productDetailsSegue, sender: nil)
+            let identifier = ProductListViewController.classNameWithoutNamespaces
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -236,7 +236,12 @@ extension OrderDetailsViewModel {
             viewController.navigationController?.pushViewController(billingInformationViewController, animated: true)
         case .details:
             ServiceLocator.analytics.track(.orderDetailProductDetailTapped)
-            viewController.performSegue(withIdentifier: Constants.productDetailsSegue, sender: nil)
+            let identifier = ProductListViewController.classNameWithoutNamespaces
+            guard let productListVC = UIStoryboard.orders.instantiateViewController(identifier: identifier) as? ProductListViewController else {
+                return
```
allendav, line 242: To aid debugging, would it be good to DDLogError or DDLogDebug in this unlikely event?

allendav: Same comment for the other calls to `instantiateViewController`

allendav (approved):
Tests well. Thank you for cleaning up `orderStatusListSegue` too. One optional request around logging something in case we spell a `storyboardIdentifier` wrong at some point in the future to help save time.

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
pmusolino: Thank you for the suggestion @allendav  👍  e55c39e3d1caccf371ca506f0c2b33932f4da4a0

PR #3501, opened 2021-01-20
Update Products Banner for M5 Features

### LINKED ISSUE #3481: Products: Update copy on the feature announcement section
Update the copy on the new features announcement area in the Products list to the following:

**Title:** New features available!
**Body:** You can now add downloadable files to a product and link upsell & cross-sell products. No longer want a product? Trash it!

Old messaging:
screenshot_20210114-080631

Android issue: https://github.com/woocommerce/woocommerce-android/issues/3444

shiki:
Closes #3481. 

## Design

This change should be the same as Android: https://github.com/woocommerce/woocommerce-android/pull/3448. 

Before | After 
--------|-------
 Simulator Screen Shot - iPhone 11 - 2021-01-20 at 10 01 28   |       Simulator Screen Shot - iPhone 11 - 2021-01-20 at 11 06 35

## Changes

Aside from the content changes, I also fixed some typos. 

## Testing

1. Delete the app. 
2. Run using `develop` first. 
3. Log in and navigate to Products. Confirm that the old message of the banner is there. 
4. Tap Dismiss. The banner should disappear. 
5. Run using this branch. 
6. Navigate to Products. Confirm that the banner is visible again but with the updated content. 
7. Confirm that the app didn't crash. 😅 That means our plist file saving is resilient. 

## Submitter Checklist

### DIFF
--- a/Storage/Storage/Model/GeneralAppSettings.swift
+++ b/Storage/Storage/Model/GeneralAppSettings.swift
@@ -28,11 +28,11 @@ public struct GeneralAppSettings: Codable, Equatable {
     /// Returns the status of a given feedback type. If the feedback is not stored in the feedback array. it is assumed that it has a pending status.
     ///
     public func feedbackStatus(of type: FeedbackType) -> FeedbackSettings.Status {
-        guard let feedbackSeeting = feedbacks[type] else {
+        guard let feedbackSetting = feedbacks[type] else {
             return .pending
         }
 
-        return feedbackSeeting.status
+        return feedbackSetting.status
     }
 
     /// Returns a new instance of `GeneralAppSettings` with the provided feedback seetings updated.
--- a/Storage/Storage/Model/FeedbackType.swift
+++ b/Storage/Storage/Model/FeedbackType.swift
@@ -5,9 +5,9 @@ public enum FeedbackType: String, Codable {
     ///
     case general
 
-    /// identifier for the products m4 beta feedback survey
+    /// Identifier for the Products M5: Linked Products, Downloadable Files, Trashing.
     ///
-    case productsM4
+    case productsM5
 
     /// identifier for the shipping labels m1 feedback survey
     ///
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsTopBannerFactory.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsTopBannerFactory.swift
@@ -43,12 +43,12 @@ struct ProductsTopBannerFactory {
 private extension ProductsTopBannerFactory {
     enum Localization {
         static let title =
-            NSLocalizedString("Create products from the app!",
-                              comment: "The title of the Work In Progress top banner on the Products tab when Products feature switch is enabled.")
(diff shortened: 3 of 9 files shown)


Storage/Storage/Model/GeneralAppSettings.swift
```diff
@@ -28,11 +28,11 @@ public struct GeneralAppSettings: Codable, Equatable {
     /// Returns the status of a given feedback type. If the feedback is not stored in the feedback array. it is assumed that it has a pending status.
     ///
     public func feedbackStatus(of type: FeedbackType) -> FeedbackSettings.Status {
-        guard let feedbackSeeting = feedbacks[type] else {
+        guard let feedbackSetting = feedbacks[type] else {
```
jaclync, line 31: nice catch!

jaclync (approved):
LGTM! :shipit: 

Just a minor request to update the milestone tag in the products in-app survey so that we can tell which Products banner they come from (though people probably comment on whatever issues they encountered):

https://github.com/woocommerce/woocommerce-ios/blob/797c5a53b90666d98d7a3a462a492b69739ddedb/WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift#L80

and the survey enum case:

https://github.com/woocommerce/woocommerce-ios/blob/797c5a53b90666d98d7a3a462a492b69739ddedb/WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift#L66

I haven't thought of a good way to catch these when we launch a new Products milestone, other than keeping these in mind 😐

Storage/Storage/Model/GeneralAppSettings.swift
shiki: I wouldn't have caught this if you didn't teach me about the spell checker in Xcode. 😉

shiki (commented):
Thanks, Jaclyn! I fixed those in 757dc1d and 19adb0a. I didn't know about those. Or I might have but I just forgot. 🤷

Garance91540 (approved):
Looking good, but there's a widow at the end of the banner. Let's bring "Trash it" to the next line, so the text reads:
...No longer want a product?
Trash it!

Instead of:
...No longer want a product? Trash
it!

shiki:
@Garance91540 This is tricky because this is laid out by the OS. If our customers use different font sizes, they look fine. See these examples of different font sizes:

   

If we try to fix this by changing the layout or forcing a new line, there's a chance that the text flow will still look odd in other font sizes. 🤷 Perhaps a better solution would be to change the content so it flows better. Maybe you have a suggestion of how to rephrase the content? I can try it and show you how it looks like.

Garance91540:
> If we try to fix this by changing the layout or forcing a new line, there's a chance that the text flow will still look odd in other font sizes. 🤷 Perhaps a better solution would be to change the content so it flows better. Maybe you have a suggestion of how to rephrase the content? I can try it and show you how it looks like.

I see! With variable fonts, I think rephrasing the content will lead to the same issues... So I'd say let's leave it like now, it's just a small detail :)

shiki:
Thanks, Garance!

PR #3504, opened 2021-01-22
Fix inconsistent date calculation for top performers data in "This Week" tab

jaclync:
Attempted fix for #2583 

## Why

A few of us have noticed that the app does not show the correct top performers in "This Week" tab. This often occurs when the site time zone has a different start of week (Monday / Sunday) so that the `date` string calculation is off by a week. The cause is from inconsistent `date` calculation when we save a `StorageTopEarnerStats` vs. the FetchedResultsController predicate in the top performers UI.

- How `date` is stored from the API data: in `StatsStoreV4`, `date` is calculated from `latestDateToInclude: Date` in the site time zone passed to `StatsActionV4.retrieveTopEarnerStats`.

https://github.com/woocommerce/woocommerce-ios/blob/1cae0e700babcf539ed1ba4ea141e4716927f1e5/Yosemite/Yosemite/Stores/StatsStoreV4.swift#L389

where the `Date` is calculated as following:

https://github.com/woocommerce/woocommerce-ios/blob/1cae0e700babcf539ed1ba4ea141e4716927f1e5/WooCommerce/Classes/ViewRelated/Dashboard/Stats%20v4/StoreStatsAndTopPerformersViewController.swift#L142

- How `date` is calculated for the predicate in Top Performers UI: in `TopPerformerDataViewController`, `date` is calculated from the current date `Date()`.

https://github.com/woocommerce/woocommerce-ios/blob/1cae0e700babcf539ed1ba4ea141e4716927f1e5/WooCommerce/Classes/ViewRelated/Dashboard/MyStore/TopPerformerDataViewController.swift#L30

## Changes

This PR fixed the inconsistent `date` calculation by DI'ing `currentDate`, `siteTimeZone`, and `timeRange` to `TopPerformerDataViewController` so that the top performs FRC can calculate the `date` value in the same as it was stored during API sync.

## Testing

Prerequisite: the store has some orders this week.

1. On the web, go to your store's WooCommerce > Dashboard (wc-admin).
2. Under "Date Range" select "Week to Date" and then "Update." (This will take you to `/wp-admin/admin.php?page=wc-admin&period=week&compare=previous_year` for your store.)
3. Notice the data displayed under "Top Products - Items Sold" on that page.
4. Open the My store tab in the app.
5. Select the "This Week" tab.
6. Scroll down to Top Performers --> the data should match the data on the web
7. Feel free to check other tabs as well!

Update release notes:


jaclync:
@designsimply would you like to check out the installable build #17286 to see this PR fixes the issue for you https://github.com/woocommerce/woocommerce-ios/issues/2583#issuecomment-762527426?

Ecarrion (approved):
Thank you for this! 👏

designsimply:
@jaclync apologies because I didn't see your ask to test with the installable build in time! 😱 

I will for sure test it in the 6.0 beta when it comes up for beta testing though. 👍

designsimply:
Tested and confirmed I'm no longer seeing timezone mismatch problems with the Top Performers in the This Week screen in WCiOS 6.0.0.0. 🎉

Issue #2916, opened 2020-10-02
Add siteID to SiteVisitStats and TopEarnerStats


shiki:
From @jaclync in https://github.com/woocommerce/woocommerce-ios/pull/2897#issuecomment-701757804. Continuation of #2897. 

We should also add `siteID` to the `SiteVisitStats` and `TopEarnerStats` entities. Both in Core Data and in Networking models. 

Doing this would help us eliminate the need to delete entities when switching stores: 

https://github.com/woocommerce/woocommerce-ios/blob/6c58e2fa3d37a3e5099c179ef54854f9a5402586/WooCommerce/Classes/Authentication/Epilogue/SwitchStoreUseCase.swift#L39-L73

And also helps us with ensuring that we're not showing data from different sites.

jaclync:
Working on this! Since we don't delete all `TopEarnerStats` when switching stores, I noticed that the app shows top performers from another store when the current store has no orders for this week 😨 

store 1 | store 2
-- | --
![Simulator Screen Shot - iPhone 11 - 2021-01-22 at 15 24 45](https://user-images.githubusercontent.com/1945542/105460230-6f913780-5cc6-11eb-8caf-4d63c177b76a.png) | ![Simulator Screen Shot - iPhone 11 - 2021-01-22 at 15 24 55](https://user-images.githubusercontent.com/1945542/105460240-728c2800-5cc6-11eb-850f-e6be46d46d47.png)

PR #3510, opened 2021-01-25
Add site ID to `SiteVisitStats` and `TopEarnerStats`: storage layer changes

jaclync:
Storage layer changes for #2916 

## Why

Since we don't delete all data after switching to another store, adding a site ID to data models associated with a store would prevent issues from incorrect data from another store ([example](https://github.com/woocommerce/woocommerce-ios/issues/2916#issuecomment-765296344)).

## Changes

📝 in case the diff count looks scary in this PR: ~2500 diffs are on the mapping model, and ~630 diffs are on the new model!

This PR contains schema changes for the task, the rest of the changes (Yosemite/Networking/app layers) will be in a followup PR.

- Added `siteID` attribute to `SiteVisitStats` and `TopEarnerStats` entity
- Used mapping model: `WooCommerceModelV42toV43.xcmappingmodel` to remove `SiteVisitStats` and `TopEarnerStats` entities in model version 42
- Added a test case for migrating from model version 42 to 43

## Testing

Please confidence check on app migration:

- Launch the app from `develop` or any other branch not based off of PR branch
- Log in to the app if needed
- Launch the app from PR branch --> the app should run as before, the dashboard stats should look the same as before

Update release notes:


ealeksandrov (approved):
Code looks good, migration from `develop` branch works as expected 👍

PR #3512, opened 2021-01-25
Add site ID to `SiteVisitStats` and `TopEarnerStats`: Networking/Yosemite/app layer changes

### LINKED ISSUE #2916: Add siteID to SiteVisitStats and TopEarnerStats

jaclync:
Fixes #2916 

Depends on https://github.com/woocommerce/woocommerce-ios/pull/3510

## Why

This PR follows storage layer changes in https://github.com/woocommerce/woocommerce-ios/pull/3510. Since we don't delete all data after switching to another store, adding a site ID to data models associated with a store would prevent issues from incorrect data from another store ([example](https://github.com/woocommerce/woocommerce-ios/issues/2916#issuecomment-765296344)).

## Changes

- Added site ID to Networking models `SiteVisitStats` and `TopEarnerStats`, with corresponding changes in Networking and Yosemite layers
- Added site ID predicate in `TopPerformerDataViewController` in the app layer

## Testing

- Launch the app --> the visitor stats and top performers should look the same as before
- Pull down to refresh dashboard --> the visitor stats and top performers should look the same as before

Update release notes:


ealeksandrov (commented):
Code is good, stats work well :+1:

Small question - I've found

https://github.com/woocommerce/woocommerce-ios/blob/fb513f683560358dde85eaef7679fe714b881138/Yosemite/Yosemite/Model/ReadOnly/SiteVisitStats%2BReadOnlyType.swift#L7-L18

and

https://github.com/woocommerce/woocommerce-ios/blob/fb513f683560358dde85eaef7679fe714b881138/Yosemite/Yosemite/Model/ReadOnly/TopEarnerStats%2BReadOnlyType.swift#L7-L18

I'm not sure if it's used at all, but maybe we should add `siteID` comparison checks here too?

jaclync:
Wow great catch on the missing checks in `ReadOnlyType` conformance @ealeksandrov! The protocol seems to be used for `EntityListener`, though I don't see its usage for `SiteVisitStats` and `TopEarnerStats`. Just updated the PR for another look!

ealeksandrov (approved):
Looks great, thanks! 🎉

PR #3513, opened 2021-01-25
Replace all Automattic copyrights, where appropriate, with WooCommerce.

### LINKED ISSUE #3514: Replace all Automattic copyrights, where appropriate, with WooCommerce.
Replace all Automattic copyrights, where appropriate, with WooCommerce. We can retain the "work with Automattic" notice.

For more context: p6q6po-b3P-p2#comment-91948

rachelmcr:
Fixes: https://github.com/woocommerce/woocommerce-ios/issues/3514

## Description

We are updating copyright notices to say WooCommerce instead of Automattic. The guidelines:

* If they just generically say, “Automattic,” then we can update them to generically say, “WooCommerce.”
* If they list the full entity name of “Automattic Inc” we should update the new notice to say “WooCommerce, Inc and WooCommerce Ireland Ltd.”
* The hiring prompt ("Made with love by Automattic. We're hiring!") is an exception and should be left as-is.

## Changes

I only found two copyright references:

* I removed one copyright reference in the code, since it seemed unnecessary.
* I updated the About screen footer from "Automattic, Inc" to "WooCommerce, Inc and WooCommerce Ireland Ltd."

## Testing

1. Go to Settings > WooCommerce and confirm the About screen shows the updated copyright notice.
2. Confirm there are no other copyright notices (user-facing or in the code) that still refer to Automattic.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/About/AboutViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/About/AboutViewController.swift
@@ -22,7 +22,7 @@ class AboutViewController: UIViewController {
             Bundle.main.detailedVersionNumber()
         )
         let localizedTitleTextLine2 = String.localizedStringWithFormat(
-            NSLocalizedString("© %@ Automattic, Inc.", comment: "About View's Footer Text. The variable is the current year"),
+            NSLocalizedString("© %@ WooCommerce, Inc and WooCommerce Ireland Ltd.", comment: "About View's Footer Text. The variable is the current year"),
             year
         )
         return String(format: localizedTitleTextLine1, year) + "\n" + localizedTitleTextLine2
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/About/AboutViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/About/AboutViewController.swift
@@ -22,7 +22,7 @@ class AboutViewController: UIViewController {
             Bundle.main.detailedVersionNumber()
         )
         let localizedTitleTextLine2 = String.localizedStringWithFormat(
-            NSLocalizedString("© %@ WooCommerce, Inc and WooCommerce Ireland Ltd.", comment: "About View's Footer Text. The variable is the current year"),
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/About/AboutViewController.swift
```diff
@@ -22,7 +22,7 @@ class AboutViewController: UIViewController {
             Bundle.main.detailedVersionNumber()
         )
         let localizedTitleTextLine2 = String.localizedStringWithFormat(
-            NSLocalizedString("© %@ Automattic, Inc.", comment: "About View's Footer Text. The variable is the current year"),
+            NSLocalizedString("© %@ WooCommerce, Inc and WooCommerce Ireland Ltd.", comment: "About View's Footer Text. The variable is the current year"),
```
shiki: I think there should be a period after `"Inc"`. 

```suggestion
            NSLocalizedString("© %@ WooCommerce, Inc. and WooCommerce Ireland Ltd.", comment: "About View's Footer Text. The variable is the current year"),
```

shiki (approved):
This looks good to me. I also ran a full-text search and I believe you've covered everything. I just have one inline suggestion. 

And I also asked in CGPNUU63E-p1611577960000600-slack where we should keep the “Automattic” here:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/About/AboutViewController.swift
rachelmcr: Thanks! I originally copied this text directly from the team requesting the change, and it didn't include a period, so I left it as it was written. But I just double checked the Chicago Manual of Style (which our house style is based on) and confirmed that they always include the period in the abbreviations `Inc.` and `Ltd.` so I agree we should add it here.

designsimply:
I wondered why the link had been left out in the update and wanted to add a quick note to say it was done to align with the iOS app. 👍

<sup>(internal reference: CGPNUU63E/p1611577960000600-slack-mobile-woo-async-mtg)</sup>

PR #3517, opened 2021-01-25
Fix: notice presenter presented by the tab bar controller inside the AppCoordinator

pmusolino:
This is a hotfix for resolving a regression introduced in this PR https://github.com/woocommerce/woocommerce-ios/pull/3498, where the Notice Presenter was no longer working because the `presentingViewController` was not correctly set to the tab bar controller under `AppCoordinator`.
This PR will be merged inside the `release/5.9` branch.

[Slack discussion](https://a8c.slack.com/archives/C6H8C3G23/p1611336604025200).

## Testing
Make sure that the notice presenter inside the app works like before everywhere.
One of the tests can be this one:
1. Open a product
2. Click on 'Price'
3. Set 'sale' price higher than 'regular' price
4. Press 'Done' button multiple times
5. Notice the error message
6. Leave the screen
7. Notice that the message disappears, and no other similar messages are shown.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/AppDelegate.swift
+++ b/WooCommerce/Classes/AppDelegate.swift
@@ -263,7 +263,7 @@ private extension AppDelegate {
     ///
     func setupNoticePresenter() {
         var noticePresenter = ServiceLocator.noticePresenter
-        noticePresenter.presentingViewController = window?.rootViewController
+        noticePresenter.presentingViewController = appCoordinator?.tabBarController
     }
 
     /// Push Notifications: Authorization + Registration!


WooCommerce/Classes/AppDelegate.swift
```diff
@@ -263,7 +263,7 @@ private extension AppDelegate {
     ///
     func setupNoticePresenter() {
         var noticePresenter = ServiceLocator.noticePresenter
-        noticePresenter.presentingViewController = window?.rootViewController
+        noticePresenter.presentingViewController = appCoordinator?.tabBarController
```
Ecarrion, line 266: This means that we can't show any notice while in the login flow, is that ok?

Ecarrion (approved):
🎉 
Just left a minor question to be sure!

WooCommerce/Classes/AppDelegate.swift
pmusolino: It seems to work also during the login flow. This is the site picker, immediately showed after the login in. I think it's the first screen where we use the notice presenter.

Ecarrion: Maybe the login module is not using the global notice presenter but a one of it's own.

pmusolino: Currently it seems that these view controllers are using their own notice presenter:
- `StorePickerViewController`
- `ManualTrackingViewController`
- `ShipmentProvidersViewController`
- `RefundConfirmationViewController`
- `NewNoteViewController`

PR #3522, opened 2021-01-25
Merge release/5.9 to develop (code freeze)

AliSoftware:
Code freeze 5.9 happened today.


Podfile
```diff
@@ -38,11 +38,14 @@ target 'WooCommerce' do
...
 
   # To allow pod to pick up beta versions use -beta. E.g., 1.1.7-beta.1
-  pod 'WordPressAuthenticator', '~> 1.34.0-beta'
+  pod 'WordPressAuthenticator', '~> 1.34.0'
   # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :commit => ''
   # pod 'WordPressAuthenticator', :git => 'https://github.com/wordpress-mobile/WordPressAuthenticator-iOS.git', :branch => ''
   # pod 'WordPressAuthenticator', :path => '../WordPressAuthenticator-iOS'
 
+  # To allow pod to pick up beta versions use -beta. E.g., 1.1.7-beta.1
+  pod 'WordPressKit', '~> 4.25.0'
```
AliSoftware, line 47: Added this explicitly to ensure that the client app picks up the non-beta version of WordPressKit. Otherwise `WordPressKit` is only pulled as a transitive dependency from `WordPressAuthenticator`, which declares that dependency in its `WordPressAuthenticator.podspec` loosely, allowing beta versions, while in the client apps we want to ensure that we use a stable one.

This is the same trick that we've been using for a while in WordPress' `Podfile` – explicit declaration of `WordPressKit` to ensure non-beta dependency.

It never seem to have been necessary in Woo before because we were lucky enough to always do the `pod install` during release cut… right after a stable version of our dependencies (incl WordPressKit) were pushed to trunk and to CDN, but before any new beta was pushed. But this time around, and because the CDN took >1h30 to resync with master, `WordPressKit 4.26.0-beta.1` was already available on CDN by the time `WordPressAuthenticator 1.34.0` was finally synced in the CDN, i.e by the time our `pod install --repo-update` finally found it, which means that this time around, it picked the next beta of WPKit that was already available instead of the latest stable version pushed only minutes before it.

WooCommerce/Classes/AppDelegate.swift
```diff
@@ -263,7 +263,7 @@ private extension AppDelegate {
     ///
     func setupNoticePresenter() {
         var noticePresenter = ServiceLocator.noticePresenter
-        noticePresenter.presentingViewController = window?.rootViewController
+        noticePresenter.presentingViewController = appCoordinator?.tabBarController
```
AliSoftware, line 266: Comes from #3517 – fix included in release branch as we wanted to include it in 5.9 but only got ready after I did the cut of the release branch, but before had time to finished the release cut process (i.e. while I was waiting for the CocoaPods CDN to sync)

jkmassel (approved):

jkmassel:
@AliSoftware – this ended up having merge conflicts by the time I got to it – we might need to do a merge branch for it 😞

AliSoftware:
Closed in favor of #3524 to fix conflict.

PR #3523, opened 2021-01-26
Use NSPersistentStoreCoordinator Instead of FileManager to Migrate Stores

### LINKED ISSUE #2667: Find The Cause of Core Data Start-up Crashes
[CoreDataManager] Recovered from persistent store loading error

Sentry Issue: [WOOCOMMERCE-IOS-1B1C](https://sentry.io/organizations/a8c/issues/1813159058/?referrer=github_integration)

```
[CoreDataManager] Recovered from persistent store loading error
```

---

In #2481, @jaclync made changes to the migrator so that the app will try to recover if:

- The store cannot be opened
- The store cannot be migrated

This tremendously brought down the number of start-up crashes we were getting (#2371). For some users, the app has [**successfully recovered**](https://sentry.io/organizations/a8c/issues/1813159058/?referrer=github_integration). For some, the app [**failed to recover**](https://sentry.io/organizations/a8c/issues/1794501606/?referrer=github_integration).

As the next step, we should find why the Core Data stores could not be opened in the first place.

shiki:
Closes #2667. This is a continuation of #3486. 

This is the last step in fully removing the usage of `FileManager` to migrate stores. As we've seen observed with the effect of https://github.com/woocommerce/woocommerce-ios/pull/3069, removing the `FileManager` is effective in reducing the crashes. This is the from the [Sentry event](https://sentry.io/share/issue/d1497b3ad58e4cba8249599fd9efc0eb/):

Hopefully, this change fully eliminates all of them. 🤞 

### Destroying and Migrating

This primarily removes the `FileManager` operations like this: 

https://github.com/woocommerce/woocommerce-ios/blob/5cc71164342b0838c353fe9add9e5e5504bf3c8d/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L186-L189

And replaces them with `NSPersistentStoreCoordinator` functions like [`destroyPersistentStore`](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468888-destroypersistentstore) and [`replacePersistentStore`](https://developer.apple.com/documentation/coredata/nspersistentstorecoordinator/1468917-replacepersistentstore). 

### Temporary Files

Previously, we use the `Documents` directory for temporary store files. 

https://github.com/woocommerce/woocommerce-ios/blob/5cc71164342b0838c353fe9add9e5e5504bf3c8d/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L93-L99

This changes that so that the temporary directory will be used instead:

https://github.com/woocommerce/woocommerce-ios/blob/b1f3fe11610f34913ab554491946cdaaa5a15726/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L171-L175

This helps in avoiding a lot of file operations. And we also don't need to fully clean up everything if the app crashes in between. 

### Fewer Store Replacements

Currently, every incremental migration will replace the store URL (`~/Documents/WooCommerce.sqlite`). 

https://github.com/woocommerce/woocommerce-ios/blob/5cc71164342b0838c353fe9add9e5e5504bf3c8d/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L188-L189

For example, when migrating from Model 39 to Model 41, the store URL will be replaced twice. 

This PR changes that so that all incremental migrations will have to be finished first before the store URL [gets replaced](https://github.com/woocommerce/woocommerce-ios/blob/b1f3fe11610f34913ab554491946cdaaa5a15726/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L94-L103). Using the same example above, migrating from Model 39 to Model 41 will only cause the store URL to be replaced **once**. If an error happens in between, there are fewer chances that the store URL will be corrupted. 

### Flow 

Hopefully, these flow diagrams will better explain what I mentioned above. 😅 

## Testing

2. Using this branch, delete models 41 and 42. 
3. Set the current model version to 40. 
5. Stop the app. 
6. Restore the deleted model versions and reset the current model version to 42. 

    https://github.com/woocommerce/woocommerce-ios/blob/11ed114d61b80b9332d97cf598e9a1a52e8cfcfe/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L76
8. Add a breakpoint on this line:

    https://github.com/woocommerce/woocommerce-ios/blob/11ed114d61b80b9332d97cf598e9a1a52e8cfcfe/Storage/Storage/CoreData/CoreDataManager.swift#L79

9. Run the app again. Migrations should happen. Confirm that the breakpoint on Step 7 was hit twice. And the `migrationAttemptMessage` variable contain these values:
    - `⚠️ Attempting migration from Model 40 to Model 41`
    - `⚠️ Attempting migration from Model 41 to Model 42`
10. Confirm the breakpoint on Step 8 was never hit. 
11. Confirm that you can load the Orders and Products. 

## Author Checklist


Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -59,21 +68,46 @@ final class CoreDataIterativeMigrator {
...
+                                                     sourceStoreURL: currentSourceStoreURL,
+                                                     storeType: storeType)
+
+                // Destroy the `currentSourceStoreURL` if it is a temporary URL since we
+                // will no longer need it.
+                if currentSourceStoreURL != sourceStoreURL {
+                    try persistentStoreCoordinator.destroyPersistentStore(at: currentSourceStoreURL,
+                                                                          ofType: storeType,
+                                                                          options: nil)
+                }
```
jaclync, line 92: since each `migrate(step:sourceStoreURL: storeType:)` returns a different temporary URL using a UUID, this condition is most likely met for each migration step? If the PSC never calls `addPersistentStore` for the current source store URL (I don't see in the PR), do we still need to call `destroyPersistentStore`?

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -59,21 +68,46 @@ final class CoreDataIterativeMigrator {
...
+            // in `sourceStoreURL` should be fully migrated and useable.
+            try persistentStoreCoordinator.replacePersistentStore(at: sourceStoreURL,
+                                                                  destinationOptions: nil,
+                                                                  withPersistentStoreFrom: lastTempDestinationURL,
+                                                                  sourceOptions: nil,
+                                                                  ofType: storeType)
+
+            // Final clean-up. Destroy the store at `lastTempDestinationURL` since it should have
+            // been copied over to `sourceStoreURL` during `replacePersistentStore` (above).
+            try persistentStoreCoordinator.destroyPersistentStore(at: lastTempDestinationURL,
```
jaclync, line 110: is it possible when `lastTempDestinationURL` is the same as `sourceStoreURL`?

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
```diff
@@ -27,19 +36,19 @@ final class CoreDataIterativeMigrator {
...
     ///
-    func iterativeMigrate(sourceStore: URL,
+    func iterativeMigrate(sourceStore sourceStoreURL: URL,
                           storeType: String,
                           to targetModel: NSManagedObjectModel) throws -> (success: Bool, debugMessages: [String]) {
         // If the persistent store does not exist at the given URL,
         // assume that it hasn't yet been created and return success immediately.
-        guard fileManager.fileExists(atPath: sourceStore.path) == true else {
-            return (true, ["No store exists at URL \(sourceStore).  Skipping migration."])
+        guard fileManager.fileExists(atPath: sourceStoreURL.path) == true else {
```
jaclync, line 44: hah nice, this is the only place we need the file manager now

jaclync (approved):
Thank you @shiki for always including the detailed diagrams (before/after) with unit tests! I just had some questions out of curiosity, and the testing steps worked perfectly! :shipit:

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: > this condition is most likely met for each migration step?

That's correct. This is to delete previous temporary store URLs so we can keep the file usage to a minimum. 

The `if` statement is necessary because the first time that the `reduce()` block is called, `currentSourceStoreURL` will be equal to `sourceStoreURL`. And we don't want to delete `sourceStoreURL` because that's the `/Documents/WooCommerce.sqlite`. 🙈 

>  If the PSC never calls addPersistentStore for the current source store URL (I don't see in the PR), do we still need to call destroyPersistentStore?

Yup, the `destroyPersistentStore` will clean up the SQLite database files so they are **almost empty**. Using the Testing Steps in this PR, I added a breakpoint in this line to investigate the file system. This is what the SQLite file looks like before `destroyPersistentStore` is called:

before

Notice that it's at 561 KB. And this is what it looks like after `destroyPersistentStore` is called:

after

Now it's just 4 KB. This means that for every step, we're clearing the temporary databases and saving a bit of disk space. 

I added more comments in 47bb9ed7243f0c261cd7d61b1d81ea81f81f6ec0. I hope this helps. 😬 

---

Do you have a suggestion on how to make this routine clearer?

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
shiki: Good thought-provoking question! 

It's possible if the [steps](https://github.com/woocommerce/woocommerce-ios/blob/47bb9ed7243f0c261cd7d61b1d81ea81f81f6ec0/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L63) turn out to be empty. But we have the `guard` statement:

https://github.com/woocommerce/woocommerce-ios/blob/47bb9ed7243f0c261cd7d61b1d81ea81f81f6ec0/Storage/Storage/CoreData/CoreDataIterativeMigrator.swift#L64

The following tests also fail if the `steps` array returns empty because of an incorrect logic:

https://github.com/woocommerce/woocommerce-ios/blob/47bb9ed7243f0c261cd7d61b1d81ea81f81f6ec0/Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift#L165

https://github.com/woocommerce/woocommerce-ios/blob/47bb9ed7243f0c261cd7d61b1d81ea81f81f6ec0/Storage/StorageTests/CoreData/CoreDataIterativeMigratorTests.swift#L206

So I think we're fairly protected. I'll just leave it as is to keep the code shorter and hopefully easier to understand.

shiki (commented):
Thank you for the review, Jaclyn! I responded to your comments. If you have further suggestions, we can continue discussing them and making changes if needed. 🙂

Storage/Storage/CoreData/CoreDataIterativeMigrator.swift
jaclync: Thanks for checking the file system! It's interesting it doesn't fully delete the files 🙄 but they are in temporary folder so that sounds fine. I can't think of any way to make the process better, the additional comments look great!

PR #3530, opened 2021-01-26
Refactor OrderStatusListViewController to move VM code into OrderDetailsViewModel to support VM refreshing Notes on Status changes

### LINKED ISSUE #3447: Order Notes Do Not Get Automatically Refreshed When Changing Status
## Describe the Bug

When changing an order's status to cancelled, the stock levels of the products will be restored. And this is noted in the order notes. However, right now, the app does not refresh the order notes when cancelling an order. The merchant has to pull-to-refresh in order to view the updated notes. 

## To Reproduce

1. Navigate to a processing order that has a product with a managed inventory.
2. Change the order's status to Cancelled. The stock levels should be updated.
3. Scroll down to the order notes, it will not show the notes about stock level changes. 

Pulling to refresh will show the updated notes. 

## Screenshots

https://user-images.githubusercontent.com/198826/104078198-0d223c80-51da-11eb-808d-504bf5e9f6f0.mov

## Expected Behavior

The notes should be auto-refreshed too.

## Isolating the Problem 

## Mobile Environment

- Device: iPhone 11 sim
- iOS version: iOS 14
- WooCommerce iOS version: 2c0c86c

## WordPress Environment

<details>
```
```
</details>

allendav:
Fixes #3447 

To test:

- open any order
- change its status (all status changes result in an order note)
- after applying the status change, within a second or two, you should also see the order notes update with the status change note
- repeat the above steps - this time watch for a "undo" notice and tap on it to undo the change (hurry - it only appears briefly)
- ensure the change is undone and both changes appear in the order notes

Rationale for changes:

- I moved the data source/view model out of `OrderStatusListViewController` and into two new objects: `OrderStatusListViewModel` and `OrderStatusListDataSource`. I did this to reduce the concerns (and size) of OSLVC and to mirror what is done elsewhere in the app (e.g. product reviews)
- I changed `OrderStatusListViewController` to return, via closure, any requested change in status to `OrderDetailsViewController` instead of allowing it to initiate the action to change the status (and update notes). This way `OrderDetailsViewController` could decide what needs to happen in addition to the new status (e.g. re-fetching notes or showing or hiding other UI bits like shipping labels) and since the undo actions and notices that result from a status change are owned by ODVC
- I moved the action creation from `OrderStatusListViewController` to `OrderDetailsViewController` for this same reason
- I observed that, even without these changes, the pull-to-refresh behavior on `develop` was not working: #3608 #3609 - I removed the pull-to-refresh and will fix that in a follow on PR to help keep this already large PR from getting larger 

Update release notes:


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -3,11 +3,33 @@ import UIKit
 import Gridicons
 import Yosemite
 import MessageUI
+import class AutomatticTracks.CrashLogging
+
+protocol OrderDetailsViewModelStatusSubModel {
```
Ecarrion: It's a guideline to add the postfix "Protocol" when naming protocols that will be constrained to one type.
Eg: `OrderDetailsViewModelProtocol`

Also since this is just meant to be used in `OrderDetailsViewModel` I think it's worth it to add some docs on why this is needed!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -417,3 +439,83 @@ extension OrderDetailsViewModel {
...
+
+    func refreshStatuses() {
+        do {
+            try statusResultsController.performFetch()
+        } catch {
+            CrashLogging.logError(error)
+        }
+    }
+
+    func getNumStatusSections() -> Int {
```
Ecarrion: nit: probaly just `statusSectionsCount` sounds more swifty!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -417,3 +439,83 @@ extension OrderDetailsViewModel {
...
+        return statusResultsController.sections.count
+    }
+
+    func getNumStatuses(inSection section: Int) -> Int {
```
Ecarrion: nit: Same here `statusesCount()`

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -417,3 +439,83 @@ extension OrderDetailsViewModel {
...
+        return statusResultsController.sections[section].numberOfObjects
+    }
+
+    func getStatusName(at indexPath: IndexPath) -> String? {
+        let status = statusResultsController.object(at: indexPath)
```
Ecarrion: I think this could be a one-liner, also when you have a one-liner, you can omit the `return` keyword
```swift
func getStatusName(at indexPath: IndexPath) -> String? {
    statusResultsController.object(at: indexPath)
}
```

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -417,3 +439,83 @@ extension OrderDetailsViewModel {
...
+    }
+
+    func getIndexOfSelectedStatus() -> IndexPath? {
+        guard let orderStatus = self.orderStatus else {
+            return nil
+        }
+        guard let row = statusResultsController.fetchedObjects.firstIndex(of: orderStatus) else {
+            return nil
+        }
+        return IndexPath(row: row, section: 0)
```
Ecarrion: nit: I think it's a little bit misleading that you have a method that computes the section count(`getNumStatusSections()`) but here you hardcode to the first section. Maybe we should remove the `getNumStatusSections()` method if it's not really needed?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -417,3 +439,83 @@ extension OrderDetailsViewModel {
...
+                ServiceLocator.analytics.track(.orderStatusChangeSuccess)
+                return
+            }
+
+            ServiceLocator.analytics.track(.orderStatusChangeFailed, withError: error)
+            DDLogError("⛔️ Order Update Failure: [\(orderID).status = \(status)]. Error: \(error)")
+
+            //X TODO self.displayErrorNotice(orderID: orderID)
+        })
+    }
```
Ecarrion: In this method, you need to capture a weak version of `self` to avoid any memory leak if this screen goes out of memory while the request is happening. [ref](https://medium.com/flawless-app-stories/you-dont-always-need-weak-self-a778bec505ef)

```swift
return asyncAction() { [weak self] in
    self.doSomething()
}
```

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -417,3 +439,83 @@ extension OrderDetailsViewModel {
...
+        //X                                                "from": newStatus.rawValue,
+        //X                                                "to": undoStatus.rawValue])
+        //X }
+    }
+
+    /// Returns an Order Update Action that will result in the specified Order Status updated accordingly.
+    ///
+    private func updateOrderAction(siteID: Int64, orderID: Int64, status: OrderStatusEnum) -> Action {
+        return OrderAction.updateOrder(siteID: siteID, orderID: orderID, status: status, onCompletion: { error in
+            guard let error = error else {
```
Ecarrion: nit: Following the [golden path rule](https://cutt.ly/vkw47w7), I would suggest putting your error handling code inside the `if` branch and do your success path outside it.

```swift
if let error = error {
    // handle error
    return 
}

// handle success
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
```diff
@@ -57,32 +40,22 @@ final class OrderStatusListViewController: UIViewController {
...
-        } catch {
-            CrashLogging.logError(error)
-        }
-
+    private func reloadTable(completion: (() -> Void)? = nil) {
+        viewModel.refreshStatuses() //X Is this necessary everytime? Seems it could be slow.
+        //X TODO - needs a completion?
         tableView.reloadData()
         preselectStatusIfPossible()
+        completion?()
```
Ecarrion: [Q] What is the reason for this `completion` block? since all the methods are synchronous I'm not sure what the use case for this!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
```diff
@@ -57,32 +40,22 @@ final class OrderStatusListViewController: UIViewController {
...
-        configureResultsController()
-
-        preselectStatusIfPossible()
+        reloadTable()
     }
 
-    /// Setup: Results Controller
-    ///
-    private func configureResultsController() {
-        statusResultsController.startForwardingEvents(to: tableView)
```
Ecarrion, line 69: [Q]: Are you planning to replicate this line later or was it a miss?

Ecarrion (approved):
This is looking good, left you some comments! 
Let me know what you think!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
shiki: Swift method names, like Obj-C, also typically don't have `get` in them. Something like this can be:

```swift
func statusName(at indexPath: IndexPath) -> String? {
```

It's weird, I know. 😄 But following that convention makes it easier for everyone else to read I think.

allendav:
@Ecarrion @ctarda - I was about to wrap this up when I realized that I may not need/want the VM to have a ResultsController<StorageOrderStatus> like here: https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-fd43a78202fcd3f89a78e22d1c6dc8819ef103625f5f0b44847296528b124cf3R23 since the VM already has access through the OrderDetailsDataSource to a ResultsController<StorageOrderStatus> - is there any reason NOT to use the status results controller the VM already has access to?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
allendav: Captured in https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-fd43a78202fcd3f89a78e22d1c6dc8819ef103625f5f0b44847296528b124cf3R447 but I am reconsidering whether we really need another status results controller since the VM already could access one through the data source

Ecarrion:
@allendav 
> is there any reason NOT to use the status results controller the VM already has access to?

That should be fine!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3530/commits/2edfb1f91bff6b2810ca165966b53f4996e6ac35 - https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-4d59e113421bc07b78bfdc220bcb864922c8762e9ca9f8218f02ab571f8660d6R162

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: Fixed "get" in https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-931cf23b4f72cbc59934f4caaffdc26bb12776a303513ead86d06d0e8ff69914R59

Fixed one-liner in https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-931cf23b4f72cbc59934f4caaffdc26bb12776a303513ead86d06d0e8ff69914R26

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-931cf23b4f72cbc59934f4caaffdc26bb12776a303513ead86d06d0e8ff69914R25

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: Ended up removing this entirely since # of sections is always 1

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: I ended up removing the protocol and instead created a viewmodel just for the status list view, which is loaded from the order details view model. This way the child view doesn't have access to the parent view's view model directly, and getting refreshes of the parent view's view model into the child view's view model can be reasoned about more clearly.

allendav:
@ctarda @Ecarrion - with the changes I have made, I am not seeing notices present. I am a little stuck on why, i.e. in

https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-7a86646239312544a7262ed5500021395d7be2581294f37eaee62c5b4ccd308cR676

the presentingViewController is null. This is causing our success and error notices to never display.

I tried changing the statusList.dismiss (https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-7a86646239312544a7262ed5500021395d7be2581294f37eaee62c5b4ccd308cR616) to statusList.presentingViewController.dismiss to no avail

Any insights?

Ecarrion:
> @ctarda @Ecarrion - with the changes I have made, I am not seeing notices present. I am a little stuck on why, i.e. in
> 
> https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-7a86646239312544a7262ed5500021395d7be2581294f37eaee62c5b4ccd308cR676
> 
> the presentingViewController is null. This is causing our success and error notices to never display.
> 
> I tried changing the statusList.dismiss (https://github.com/woocommerce/woocommerce-ios/pull/3530/files#diff-7a86646239312544a7262ed5500021395d7be2581294f37eaee62c5b4ccd308cR616) to statusList.presentingViewController.dismiss to no avail
> 
> Any insights?

Hey @allendav, there was a [fix around the notice configuration](https://github.com/woocommerce/woocommerce-ios/pull/3517) some days ago. If your merge/rebase `develop` into your branch, you should be good to go!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3530/commits/7cc9c5638575c014e0ca397fd191cf46b31fc2b1

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
allendav: Not needed. Removed. We'll need a completion handler when pull-to-refresh is implemented in a follow on PR ( for #3608 #3609 )

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: I was surprised by this too. It was in the code that way already, and is present elsewhere https://github.com/woocommerce/woocommerce-ios/search?q=%22guard+let+error%22

If you'd like me to change it nonetheless, I am happy to do so

allendav:
@Ecarrion @shiki - good to go?

WooCommerce/Classes/ViewModels/Order Details/OrderStatusListViewModel.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// Return the number of statuses for the view.
+    ///
+    func statusCount() -> Int {
+        dataSource.statusCount()
+    }
+
+    /// Return the index of the current order status.
+    ///
+    func indexOfCurrentOrderStatus() -> IndexPath? {
+        guard let foundStatus = dataSource.statuses().filter({ $0.status == status }).first else {
```
Ecarrion, line 33: super nit: You don't need the `(` and `)` in the filter closure
```suggestion
        guard let foundStatus = dataSource.statuses().filter{ $0.status == status }.first else {
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
```diff
@@ -162,89 +119,44 @@ extension OrderStatusListViewController {
...
     func numberOfSections(in tableView: UITableView) -> Int {
-        return statusResultsController.sections.count
+        return 1
     }
 
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return statusResultsController.sections[section].numberOfObjects
+        guard section == 0 else {
+            return 0
+        }
```
Ecarrion, line 148: I don't think this is necesary! :-)

Ecarrion (approved):
Code and tests look good to me! 🚀

shiki:
@allendav Sorry, mine was just a little thing. Didn't want to block this PR in any way. 😅 I see Ernesto already approved this so :shipit:!

WooCommerce/Classes/ViewModels/Order Details/OrderStatusListViewModel.swift
allendav: I'm going to keep it. Xcode warns: "Trailing closure in this context is confusable with the body of the statement; pass as a parenthesized argument to silence this warning"

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
allendav: I'd like to keep it, especially as a protection against future sectioning of statuses

Issue #1795, opened 2020-01-29
Improved Stats: Visitor count can be wrong


rachelmcr:
**Describe the bug**

When I select the current month on the "This Year" tab, I expect the stats to match what I see on the "This Month" tab. However, the Visitors count doesn't match — the visitor count for "This Month" is too high.

It looks like the Visitors count for "This Month" may simply be adding up the total visitors for each day in the month. However, this is not accurate if the same visitor comes to the site on multiple days in the month.

**To Reproduce**
1. Log in to a store with WC Admin where at least one visitor has returned to the store on multiple days in the same month.
2. Enable the Improved Stats feature.
3. Select the "This Year" tab, select the current month, and note the number of visitors displayed.
4. Select the "This Month" tab and note that the number of visitors is different.

**Screenshots**

This Month tab (in January)|This Year tab (January selected)
-|-
<img src="https://user-images.githubusercontent.com/8658164/73373428-828a4b00-42b0-11ea-854a-0c1246aef31c.png">|<img src="https://user-images.githubusercontent.com/8658164/73373455-8ddd7680-42b0-11ea-8d9b-6399afe9de17.png">

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 13.3
- WooCommerce iOS version: 3.5.0.0 TestFlight beta

jaclync:
I noticed incorrect visitor count in "today" tab - it is showing 10, however I didn't visit the site at all today and "this week" shows 1 visitor count 😓 After some debugging, it seems like the "today" tab isn't getting the correct `SiteVisitStats` since both "today" and "this week" end on Sep 30 and one overrides the other in storage. That's why "today" tab is getting the visitor count on "this month" tab.

today | this week | this month
-- | -- | --
![IMG_7717](https://user-images.githubusercontent.com/1945542/94669258-93112a80-0343-11eb-949e-96cb60fe455e.PNG) | ![IMG_7718](https://user-images.githubusercontent.com/1945542/94669268-960c1b00-0343-11eb-8fc8-b30a9aa224c5.PNG) | ![IMG_7719](https://user-images.githubusercontent.com/1945542/94669272-973d4800-0343-11eb-9377-9ec2fffe2967.PNG)

"This month" and "this year > this month" visitor count also don't match:

this month | this year > this month
-- | --
[...]

PR #3531, opened 2021-01-27
Dashboard > visitor stats: storage layer changes - add `timeRange` attribute to `SiteVisitStats` 

jaclync:
Attempted fix for #1795 - storage layer

## Why

I wanted to work on this fix because I can reproduce the issue this week of the year 😆 Right now, we save `SiteVisitStats`'s `granularity` to the period unit like `day`/`day`/`day`/`month` for `Today`/`This Week`/`This Month`/`This Year` respectively. In `StoreStatsV4PeriodViewController` where we show the visitor stats, we query `SiteVisitStats` with `granularity` (each time range's period unit) and `date` (the end date of the time range). This is usually fine, but for certain days of the year the `granularity` and `date` could be the same for two time ranges. For example, "This Week"'s end date for 1/25-31 is 1/31 (for regions that start the week on Mondays) and "This Month"'s end date is also 1/31. In the database, we also only see 3 entries for `SiteVisitStats` in this case.

Screen Shot 2021-01-25 at 2 20 11 PM

To prevent from conflicting data models, I added `timeRange` attribute to `SiteVisitStats` entity so that we can later store `SiteVisitStats` with a unique time range and query with the time range for each tab in Dashboard.

## Changes

This PR only includes the storage layer changes since it is already more than 500 diffs (most diffs are on the new model version):

- Added `timeRange` attribute to `SiteVisitStats` entity
  - `timeRange` is non-optional and default to empty string, this way we don't have to add a mapping model. In the app, we only query data with the given `timeRange` value for each tab so the pre-existing data with empty `timeRange` won't be shown. Pre-existing data will also be replaced by the latest data with the corresponding `timeRange`

## Testing

Please confidence check on app migration:

- Launch the app from `develop` or any other branch not based off of PR branch
- Log in to the app if needed
- Launch the app from PR branch --> the app should run as before, the dashboard stats should look the same as before

Update release notes:

### DIFF
--- a/Storage/Storage.xcodeproj/project.pbxproj
+++ b/Storage/Storage.xcodeproj/project.pbxproj
@@ -1595,7 +1595,7 @@
 				746A9D14214071F90013F6FF /* Model 2.xcdatamodel */,
 				B59E11D920A9D00C004121A4 /* Model.xcdatamodel */,
 			);
-			currentVersion = 0284BD8725BAC75500D00C06 /* Model 43.xcdatamodel */;
+			currentVersion = 029DE8AD25C115860046EFF5 /* Model 44.xcdatamodel */;
 			path = WooCommerce.xcdatamodeld;
 			sourceTree = "<group>";
 			versionGroupType = wrapper.xcdatamodel;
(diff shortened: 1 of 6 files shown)


pmusolino (approved):
LGTM :shipit:

rachelmcr (approved):
Looks good and tests well! 👍 

I was going to mention that someone else should probably also review since I'm new to Core Data, but I see that @pmusolino beat me to it. 😄

PR #3532, opened 2021-01-27
Dashboard: fix visitor stats data on days when the end date is the same for more than one tab

jaclync:
Attempted fix for #1795 
Based on https://github.com/woocommerce/woocommerce-ios/pull/3531

## Why

I wanted to work on this fix because I can reproduce the issue this week of the year 😆 Right now, we save `SiteVisitStats`'s `granularity` to the period unit like `day`/`day`/`day`/`month` for `Today`/`This Week`/`This Month`/`This Year` respectively. In `StoreStatsV4PeriodViewController` where we show the visitor stats, we query `SiteVisitStats` with `granularity` (each time range's period unit) and `date` (the end date of the time range). This is usually fine, but for certain days of the year the `granularity` and `date` could be the same for two time ranges. For example, "This Week"'s end date for 1/25-31 is 1/31 (for regions that start the week on Mondays) and "This Month"'s end date is also 1/31. In the database, we also only see 3 entries for `SiteVisitStats` in this case.

Screen Shot 2021-01-25 at 2 20 11 PM

To prevent from conflicting data models, the previous PR https://github.com/woocommerce/woocommerce-ios/pull/3531 added `timeRange` attribute to `SiteVisitStats` entity so that we can later store `SiteVisitStats` with a unique time range and query with the time range for each tab in Dashboard. This PR set the `timeRange` attribute when saving to storage and query with `timeRange` for UI display.

This PR doesn't close the issue #1795 because there could be other reasons for incorrect visitor data like from the API, which I've seen before but cannot reproduce.

## Changes

- In `Yosemite.StatsStoreV4`, set `timeRange` to `Storage.SiteVisitStats` when upserting
  - It also loads existing object with the same `timeRange` instead `date` since `date` might not be unique for each tab
- In `StoreStatsV4PeriodViewController`, replaced `date` with `timeRange` in `SiteVisitStats` FRC predicate
- Updated unit tests

## Testing

Prerequisites:
1. The store has non-zero visitor stats on various days this month
2. The region of the simulator/device has Monday as the start of the week (e.g. UK)

- Launch the app --> check the visitor stats in `Today`/`This Week`/`This Month`/`This Year` tabs, they should not conflict with each other (e.g. `This Week` is 3 while `This Month` is 0) and are consistent with `wp-admin > Stats`

### `SiteVisitStats` in the database

before | after
-- | --
Screen Shot 2021-01-25 at 2 20 11 PM | Screen Shot 2021-01-27 at 1 27 46 PM

### Dashboard tabs

\ | This Week | This Month | This Year
-- | -- | -- | --
before | Simulator Screen Shot - iPhone 11 - 2021-01-27 at 13 18 58 | Simulator Screen Shot - iPhone 11 - 2021-01-27 at 13 19 00 | Simulator Screen Shot - iPhone 11 - 2021-01-27 at 13 19 01
after | Simulator Screen Shot - iPhone 11 - 2021-01-27 at 13 52 53 | Simulator Screen Shot - iPhone 11 - 2021-01-27 at 13 52 55 | Simulator Screen Shot - iPhone 11 - 2021-01-27 at 13 52 57

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsV4PeriodViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsV4PeriodViewController.swift
@@ -373,9 +373,9 @@ private extension StoreStatsV4PeriodViewController {
         let storageManager = ServiceLocator.storageManager
         let dateFormatter = DateFormatter.Stats.statsDayFormatter
         dateFormatter.timeZone = siteTimezone
-        let predicate = NSPredicate(format: "granularity ==[c] %@ AND date == %@",
+        let predicate = NSPredicate(format: "granularity ==[c] %@ AND timeRange == %@",
                                     timeRange.siteVisitStatsGranularity.rawValue,
-                                    dateFormatter.string(from: currentDate))
+                                    timeRange.rawValue)
         let descriptor = NSSortDescriptor(keyPath: \StorageSiteVisitStats.date, ascending: false)
         return ResultsController(storageManager: storageManager, matching: predicate, sortedBy: [descriptor])
     }
--- a/Networking/Networking/Model/Stats/StatGranularity.swift
+++ b/Networking/Networking/Model/Stats/StatGranularity.swift
@@ -8,41 +8,4 @@ public enum StatGranularity: String, Decodable {
     case week
     case month
     case year
-
-    public var pluralizedString: String {
-        switch self {
-        case .day:
-            return NSLocalizedString("Days", comment: "Plural of 'day' — a statistical unit")
-        case .week:
-            return NSLocalizedString("Weeks", comment: "Plural of 'week' — a statistical unit")
-        case .month:
-            return NSLocalizedString("Months", comment: "Plural of 'month' — a statistical unit")
-        case .year:
-            return NSLocalizedString("Years", comment: "Plural of 'year' — a statistical unit")
-        }
-    }
-
-    public var accessibilityIdentifier: String {
-        return "granularity-\(self.rawValue)"
-    }
-}
-
-// MARK: - StringConvertible Conformance
-//
-extension StatGranularity: CustomStringConvertible {
-
-    /// Returns a user-freindly, localized string describing the stat granularity
-    ///
-    public var description: String {
-        switch self {
-        case .day:
-            return NSLocalizedString("Day", comment: "Statistical unit - a single day")
-        case .week:
-            return NSLocalizedString("Week", comment: "Statistical unit - a single week")
-        case .month:
-            return NSLocalizedString("Month", comment: "Statistical unit - a single week")
-        case .year:
-            return NSLocalizedString("Year", comment: "Statistical unit - a single year")
-        }
-    }
 }
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -4,6 +4,7 @@
 -----
 - [*] Updated copyright notice to WooCommerce
 - [*] Fix: top performers in "This Week" tab should be showing the same data as in WC Admin.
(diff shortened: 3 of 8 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsV4PeriodViewController.swift
```diff
@@ -373,9 +373,9 @@ private extension StoreStatsV4PeriodViewController {
         let storageManager = ServiceLocator.storageManager
         let dateFormatter = DateFormatter.Stats.statsDayFormatter
         dateFormatter.timeZone = siteTimezone
-        let predicate = NSPredicate(format: "granularity ==[c] %@ AND date == %@",
+        let predicate = NSPredicate(format: "granularity ==[c] %@ AND timeRange == %@",
                                     timeRange.siteVisitStatsGranularity.rawValue,
-                                    dateFormatter.string(from: currentDate))
+                                    timeRange.rawValue)
```
pmusolino, line 378: Could the new Predicates be used here since we are changing this NSPredicate? p91TBi-3BJ

pmusolino (approved):
LGTM :shipit: 

Code ✅ just a minor nit
Testing ✅

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsV4PeriodViewController.swift
Ecarrion: I guess we could, but we would need to `import Storage`, how do we feel about that? cc @woocommerce/ios-developers

jaclync: I actually tried that earlier, but importing `Storage` resulted in a bunch of ambiguous types in `StoreStatsV4PeriodViewController` 😞 If not importing `Storage`, I'm not sure how to use the operators like `public func && (p1: TypedPredicate, p2: TypedPredicate) -> CompoundPredicate` 😐 

It'd be nice if we have a separate foundation framework with helpers that are applicable to all layers, but it's a non-trivial task on its own. Would like to hear thoughts from the team!

pmusolino: So don't worry, It looks good to me to proceed in this way 👍

shiki: Sorry, I'm late! 

I agree that we shouldn't import `Storage` here. `WooCommerce` should only reference `Yosemite`. We're still on a year-long crusade of removing `Storage` references from `WooCommerce` and we're not done yet! 😅 

From the [design doc](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/architecture-overview.md): 

> WooCommerce:  Our main target, which is expected to **only** interact with the entire stack thru the Yosemite.framework.

Jaclyn's idea of a separate Foundation framework sounds like a good option to try.

Ecarrion: Extracting those predicates into a separate package should be fairly easy! But The work would be to analyze what other utilities would be a good fit for that new package!

jaclync: Nice, I wasn't sure how much effort it takes to create a separate package/framework! I created a separate issue to extract the utilities in https://github.com/woocommerce/woocommerce-ios/issues/3547 - feel free to update the list of files to share!

jaclync:
Great discussion from the team on sharing the typed predicates! merging now 😄

PR #3552, opened 2021-02-01
Add Product Variations Attributes and Options Cell UI

### LINKED ISSUE #3519: Variations: Attributes and Options Cell UI
Tasks
- [ ] Build and display product attribute and options cell 
- [ ] Make sure this is under a feature flag

cell-ui

rachelmcr:
Closes #3519 

## Description

This PR adds an **Attributes** cell to the Product Variation detail screen, behind a new feature flag for editing product variation attributes (enabled for local development/alpha only). For now, the Attributes cell doesn't do anything when tapped and always displays the options text "Any Color - Any Material."

## Changes

* Adds the `editProductVariationAttributes` feature flag set to `localDeveloper` and `alpha` builds.
* Adds the `customize` icon from gridicons (for the attributes cell icon), and an icon test case to make sure the icon is not nil.
* In `DefaultProductFormTableViewModel`, configures the attributes row. For now the cell displays static detail text (for the attribute options), and the cell is set to expand if that detail text doesn't fit on a single line.
* In `ProductVariationFormActionsFactory`, sets the attributes row to be displayed based on the feature flag. When enabled, the attributes row is displayed below the price settings/warning row and above the variation status (Enabled) row.

### Screenshots

/|Before|After
-|-|-
Variation with price|Simulator Screen Shot - iPhone 12 mini - 2021-02-01 at 15 33 56|Simulator Screen Shot - iPhone 12 mini - 2021-02-01 at 15 29 36
Variation without price|Simulator Screen Shot - iPhone 12 mini - 2021-02-01 at 15 34 00|Simulator Screen Shot - iPhone 12 mini - 2021-02-01 at 15 29 31

## Testing

1. Open a variable product that has existing variations.
2. Tap on Variations.
3. Select a variation that has a price set.
4. Confirm you see the Attributes cell below the Price cell in the variation details.
5. Go back to the variations list.
6. Select a variation with no price set.
7. Confirm you see the Attributes cell below the "Variations without price ..." warning.
8. In `DefaultFeatureFlagService`, set `.editProductVariationAttributes` to false and confirm no Attributes cell appears in the product variation details.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIImage+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIImage+Woo.swift
@@ -134,6 +134,12 @@ extension UIImage {
         UIImage.gridicon(.creditCard)
     }
 
+    /// Customize Icon
+    ///
+    static var customizeImage: UIImage {
+        return UIImage.gridicon(.customize)
+    }
+
     /// Delete Icon
     ///
     static var deleteImage: UIImage {
--- a/WooCommerce/Classes/Extensions/UIImage+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIImage+Woo.swift
@@ -137,7 +137,8 @@ extension UIImage {
     /// Customize Icon
     ///
     static var customizeImage: UIImage {
-        return UIImage.gridicon(.customize)
+        UIImage.gridicon(.customize)
+            .imageFlippedForRightToLeftLayoutDirection()
     }
 
     /// Delete Icon
--- a/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
+++ b/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
@@ -7,6 +7,8 @@ struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .addProductVariations:
             return buildConfig == .localDeveloper || buildConfig == .alpha
+        case .editProductVariationAttributes:
+            return buildConfig == .localDeveloper || buildConfig == .alpha
         case .shippingLabelsRelease1:
             return true
         default:
--- a/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
+++ b/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
@@ -7,8 +7,6 @@ struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .addProductVariations:
             return buildConfig == .localDeveloper || buildConfig == .alpha
-        case .editProductVariationAttributes:
-            return buildConfig == .localDeveloper || buildConfig == .alpha
         case .shippingLabelsRelease1:
             return true
         default:
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
@@ -423,10 +423,7 @@ private extension DefaultProductFormTableViewModel {
         let title = Localization.variationAttributesTitle
         let details = "Any Color - Any Material"
 
-        return ProductFormSection.SettingsRow.ViewModel(icon: icon,
-                                                        title: title,
-                                                        details: details,
-                                                        isActionable: isEditable)
+        return .init(icon: icon, title: title, details: details, isActionable: isEditable)
     }
 
     // MARK: Product downloads only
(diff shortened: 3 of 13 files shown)


WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -134,6 +134,12 @@ extension UIImage {
         UIImage.gridicon(.creditCard)
     }
 
+    /// Customize Icon
+    ///
+    static var customizeImage: UIImage {
+        return UIImage.gridicon(.customize)
```
Ecarrion: I think we need to prepare this image for RTL languages with `imageFlippedForRightToLeftLayoutDirection()`, otherwise the image won't be properly rotated for RTL languages.

super nit: for one-liners, we can drop the `return` keyword. It looks like the project has adopted that convention!

WooCommerce/Classes/System/DefaultFeatureFlagService.swift
```diff
@@ -7,6 +7,8 @@ struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .addProductVariations:
             return buildConfig == .localDeveloper || buildConfig == .alpha
+        case .editProductVariationAttributes:
+            return buildConfig == .localDeveloper || buildConfig == .alpha
```
Ecarrion: Thinking about it, I think we can reuse the `addProductVariations` feature flag(maybe rename it to `productVariations`)

Since this is the only place where the new feature flag will be used I think it's safe to use the existing one and then remove the conditional if we want to release the "edit" functionality first.

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -416,6 +418,17 @@ private extension DefaultProductFormTableViewModel {
         return ProductFormSection.SettingsRow.WarningViewModel(icon: icon, title: title)
     }
 
+    func variationAttributesRow(isEditable: Bool) -> ProductFormSection.SettingsRow.ViewModel {
+        let icon = UIImage.customizeImage
+        let title = Localization.variationAttributesTitle
+        let details = "Any Color - Any Material"
+
+        return ProductFormSection.SettingsRow.ViewModel(icon: icon,
```
Ecarrion: Tip: Since you already declared the type you are going to return in the method signature, you can commit all the namespaces in favor of something like:
```swift
return .ViewModel(icon: icon, title: title, details: details, isActionable: isEditable)
```

WooCommerce/WooCommerceTests/Extensions/IconsTests.swift
```diff
@@ -75,6 +75,10 @@ final class IconsTests: XCTestCase {
         XCTAssertNotNil(UIImage.creditCardImage)
     }
 
+    func testCustomizeImageIsNotNil() {
+        XCTAssertNotNil(UIImage.customizeImage)
```
Ecarrion, line 79: 👏

Ecarrion (commented):
This is looking very good! 👏 
I left a couple of suggestions and tips, let me know what you think!

WooCommerce/Classes/System/DefaultFeatureFlagService.swift
rachelmcr: > I think it's safe to use the existing one and then remove the conditional if we want to release the "edit" functionality first.

Ah, good point! I wanted to make it simple for releasing the edit functionality separately, but this also makes sense.

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
rachelmcr: How do you decide between being more concise like this versus following the style used elsewhere?

It works when I return the init method like this:

```swift
return .init(icon: icon, title: title, details: details, isActionable: isEditable)
```

But since all the related methods include the full namespace I was following the same approach here. Is it better to be consistent with how it's done in other, related methods or go ahead with this more concise style?

WooCommerce/Classes/System/DefaultFeatureFlagService.swift
rachelmcr: I went ahead and reused the `addProductVariations` feature flag, but without renaming it. (I started to rename it, but it felt a little confusing in context since not all of the product variations feature is behind the flag.)

I left a comment on the conditional where it's used for the "edit" functionality, as a reminder if we want to release that before everything else.

rachelmcr:
Thanks for your feedback, @Ecarrion! Ready for another look.

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
Ecarrion: That's a tricky one! I usually always aim for consistency, however, I don't see this as a lack of consistency. I think of consistency more like similar approaches on how to achieve something. EG: Have this function as a static function would be a diff approach. 

Here I see removing the namespace as achieving the same but with less code, so I think we are good to go!

Ecarrion (approved):
Looks and works perfect! 🚀

Issue #3535, opened 2021-01-27
Variations: Link attribute to Product


Ecarrion:
After tapping on the "Next" CTA on the attribute options screen, we should update the product with this new product attribute.

<img width="777" alt="Screen Shot 2021-01-27 at 2 16 56 PM" src="https://user-images.githubusercontent.com/562080/106041988-7d6e0080-60aa-11eb-86bb-9926f24a5122.png">

Tasks: 
- [x] Update product encoding to support attributes
- [x] Show a loading spinner instead of the "Next Button" while the creation request is happening.
- [x] Show an error notice if the creation request fails.
- [x] Connect the product update with the "Next" CTA

PR #3568, opened 2021-02-02
[Variations] Add new option(to attribute) UI handling

### LINKED ISSUE #3527: Variations: Create New Product Attribute Interaction
UI:
create

Tasks :
- [ ] Initially show only "Option Name" text field
- [ ] After hitting the return key, Add that option in the "Options Offered" Section
- [ ] Disable "next" button until at least one option is selected.

Note: next button action will be implemented in another task https://github.com/woocommerce/woocommerce-ios/issues/3535

Ecarrion:
closes #3527 

# Why

Today's turn will be to: Add a new option, to a new attribute. All of these are prerequisites for creating our first variation!

This PR does not create the attribute or option remotely yet(#3535), it only allows to create it visually.

# How
- Update `TextFieldTableViewCell` to provide a callback when the return key is pressed

- Update `AddAttributeOptionsViewController.Row` so that the `.selectedTerms` case has now an associated value to transport the option name

- Update `AddAttributeOptionsViewModel` to add a method so consumer can add new options. This is powered by a new `State` struct that is in charge of holding the view model state. This `state` has an observation that regenerates the `sections` property everytime it changes. This helps us to maintain the view in sync.

- Update `AddAttributeOptionsViewController` to render the options offered section and the next button visibility based on the view model content.

# Demo
add-new-option

# Testing Steps
- Navigate to a variable product that does not have variations yet
- Tap on "Add Variation" cell
- Tap on "Add Variation" Button
- Write an attribute name and tap next
- Write an option name and hit enter/return
- See that the option is moved to the "options offered" section
- See that the next button is enabled

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -33,6 +33,7 @@ final class AddAttributeOptionsViewController: UIViewController {
         registerTableViewCells()
         startListeningToNotifications()
         observeViewModel()
+        renderViewModel()
     }
 }
 
@@ -41,8 +42,6 @@ final class AddAttributeOptionsViewController: UIViewController {
 private extension AddAttributeOptionsViewController {
 
     func configureNavigationBar() {
-        title = viewModel.titleView
-
         navigationItem.rightBarButtonItem = UIBarButtonItem(title: Localization.nextNavBarButton,
                                                            style: .plain,
                                                            target: self,
@@ -76,8 +75,14 @@ private extension AddAttributeOptionsViewController {
     func observeViewModel() {
         viewModel.onChange = { [weak self] in
             guard let self = self else { return }
-            self.tableView.reloadData()
+            self.renderViewModel()
         }
+    }
+
+    func renderViewModel() {
+        title = viewModel.titleView
+        navigationItem.rightBarButtonItem?.isEnabled = viewModel.isNextButtonEnabled
+        tableView.reloadData()
     }
 }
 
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -138,32 +150,36 @@ private extension AddAttributeOptionsViewController {
...
+        case (.termTextField, let cell as TextFieldTableViewCell):
             configureTextField(cell: cell)
-        case let cell as BasicTableViewCell where row == .selectedTerms:
-            configureOption(cell: cell)
-        case let cell as BasicTableViewCell where row == .existingTerms:
-            configureOption(cell: cell)
+        case (let .selectedTerms(name), let cell as BasicTableViewCell):
+            configureOption(cell: cell, text: name)
+        case (.existingTerms, let cell as BasicTableViewCell):
+            configureOption(cell: cell, text: "Work in Progress")
```
Ecarrion, line 159: This was changed because now, `row` has cases with associated types( `.selectedTerms(name: String)` ) and we can't compare with with `==` anymore.

WooCommerce/Classes/ViewRelated/ReusableViews/TextFieldTableViewCell.swift
```diff
@@ -38,6 +39,7 @@ final class TextFieldTableViewCell: UITableViewCell {
         textField.keyboardType = viewModel.keyboardType
         textField.addTarget(self, action: #selector(textFieldDidChange(textField:)), for: .editingChanged)
         textField.addTarget(self, action: #selector(textFieldDidBegin(textField:)), for: .editingDidBegin)
+
```
ealeksandrov: Accidental newline?

ealeksandrov (approved):
Code is good, attributes input UI works well ✅
Nice suite of tests! 👍

WooCommerce/Classes/ViewRelated/ReusableViews/TextFieldTableViewCell.swift
Ecarrion: Indeed!

Issue #3518, opened 2021-01-25
Variations: Update product variation attribute 


Ecarrion:
Tasks: 
- [ ] Add support(encoding) to update a product variation attribute's options remotely
- [ ] Mark the variation as "updatable"(show update button) after the user has selected new options.
- [ ] Make sure the product variation attributes are correctly updated in our storage layer and in previous screens

Issue #3516, opened 2021-01-25
Variations: Populate attributes and attributes option


Ecarrion:
Tasks:
- [ ] Make sure that the UI created in https://github.com/woocommerce/woocommerce-ios/issues/3515 is populated with the correct attributes and attributes options
- [ ] Investigate if we need to include an "Any option" option client side in case it is not returned within the product variation response.

<img width="1392" alt="ui" src="https://user-images.githubusercontent.com/562080/105730582-87520100-5efc-11eb-9c91-41e3b69fc1af.png">

Issue #3515, opened 2021-01-25
Variations: Edit Attributes UI


Ecarrion:
Tasks:
 - [ ] Create the necessary classes to represent the "Attributes" and "Edit Attributes" UI
 - [ ] Add navigation to launch that UI from the product variation screen -> attributes cell

<img width="1392" alt="ui" src="https://user-images.githubusercontent.com/562080/105729726-8371af00-5efb-11eb-9038-b747a831774a.png">

PR #3578, opened 2021-02-03
Add first screen of edit variation attributes

ealeksandrov:
Part of #3515, #3516.

## Description

This PR adds first screen of variation attributes edit flow. This is the screen where you can see all attributes with options.
It doesn't allow to change any attributes yet, but it will display current values.

Navigation is connected to button that is already behind `addProductVariations` feature flag.

## Changes

- Added `AttributePickerViewController` with UITableView to display attributes + options.
- `EditableProductVariationModel` injected at VC init.
- Connected navigation from `ProductFormViewController` to `AttributePickerViewController`
- There is no additional ViewModel layer for screen now, since it simply displays data from `EditableProductVariationModel`, but maybe it'll be introduced later, together with editing and saving.

## Testing

1. Open product with existing variations
2. Open variations details
3. Tap on attributes cell (added in #3519)
4. Observe correct attributes and details on new screen

### Known limitations

- Tapping on attributes doesn't open next screen
- Done button doesn't work yet - part of #3518

## Screenshots

Any | All
--|--
Simulator Screen Shot - iPhone 12 - 2021-02-03 at 14 12 33|Simulator Screen Shot - iPhone 12 - 2021-02-03 at 14 12 47

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
@@ -348,7 +348,12 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>: 
                                                                                formType: viewModel.formType,
                                                                                isAddProductVariationsEnabled: isAddProductVariationsEnabled)
                 show(variationsViewController, sender: self)
-            case .status, .noPriceWarning, .attributes:
+            case .attributes:
+                if let productVariationModel = product as? EditableProductVariationModel {
+                    let attributePickerViewController = AttributePickerViewController(variationModel: productVariationModel)
+                    show(attributePickerViewController, sender: self)
+                }
+            case .status, .noPriceWarning:
                 break
             }
         }
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
@@ -348,11 +348,12 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>: 
                                                                                formType: viewModel.formType,
                                                                                isAddProductVariationsEnabled: isAddProductVariationsEnabled)
                 show(variationsViewController, sender: self)
-            case .attributes:
-                if let productVariationModel = product as? EditableProductVariationModel {
-                    let attributePickerViewController = AttributePickerViewController(variationModel: productVariationModel)
-                    show(attributePickerViewController, sender: self)
+            case .attributes(_, let isEditable):
+                guard isEditable, let productVariationModel = product as? EditableProductVariationModel else {
+                    return
                 }
+                let attributePickerViewController = AttributePickerViewController(variationModel: productVariationModel)
+                show(attributePickerViewController, sender: self)
             case .status, .noPriceWarning:
                 break
             }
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -348,7 +348,12 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
                                                                                formType: viewModel.formType,
                                                                                isAddProductVariationsEnabled: isAddProductVariationsEnabled)
                 show(variationsViewController, sender: self)
-            case .status, .noPriceWarning, .attributes:
+            case .attributes:
```
rachelmcr: I think we'll need a check for `isEditable` here, to make sure `AttributePickerViewController` isn't accessible when the variation is in read-only mode. (For example, when you view the variation details from the order details screen.)

Something like this should work:

```swift
            case .attributes(_, let isEditable):
                guard isEditable else {
                    return
                }
```

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
```diff
@@ -0,0 +1,151 @@
...
+        // TODO-3515: Show options for attribute
+    }
+}
+
+private extension AttributePickerViewController {
+    enum Localization {
+        static let titleView = NSLocalizedString("Attributes", comment: "Edit Product Attributes screen navigation title")
+        static let doneNavBarButton = NSLocalizedString("Done", comment: "Done nav bar button title in Edit Product Attributes screen")
+        static let headerAttributes = NSLocalizedString("Options", comment: "Header of attributes section in Edit Product Attributes screen")
+        static let anyAttributeOption = NSLocalizedString("Any", comment: "Product varition attribute description where the attribute is set to any value.")
```
rachelmcr: tiny nit: There's a small typo in the comment (`Product varition` should be `Product variation`).

rachelmcr (changes requested):
This looks great! We just need to make sure the navigation to `AttributePickerViewController` is restricted in read-only mode, but otherwise this is 👍 .

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
ealeksandrov: Thanks! There is even similar comment in another file that I'll also update:

https://github.com/woocommerce/woocommerce-ios/blob/e69b54614f6dd6084eae35b0db203cc04b636090/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/EditableProductVariationModel.swift#L226

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
ealeksandrov: Agree, will add a `guard` 👍

ealeksandrov: Updated in ab930ce

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
ealeksandrov: Fixed in 965724c

ealeksandrov:
Thanks for review @rachelmcr! Pushed small update.

rachelmcr (approved):
LGTM! :shipit:

PR #3579, opened 2021-02-03
Variations: Delete and reorder options of a new product attribute

### LINKED ISSUE #3528: Variations: New Product Attribute options management
Ref:
management

Tasks:
- [ ] Allow options offered to be sorted
- [ ] When tapping the "X" button, drop the respective option (This should be only locally as it hasn't been created yet)

Ecarrion:
closes #3528 

# Why
Today's turn will be to: Reorder and delete options offered to a new product attribute.

# How

- Update `AddAttributeOptionsViewController` to support reordering by implementing the `moveRowAt` and `canMoveRowAt` methods.

- Update `AddAttributeOptionsViewController` to show an `X` button in the "Options Offered" cell and attack a gesture recognizer to it.

- Update `Section` struct to include an `allowsReorder` property so that the `ViewModel` can dictate which section can be reordered.

- Update `AddAttributeOptionsViewModel` to support `reorder` and `delete` functions by modifying it's internal state.

# Demo
reorder-delete

# Testing Steps

- Navigate to a variable product that does not have any variation yet
- Initiate the "Add Variation" Flow by: Tap on the Add variations button(There are two) and create a new attribute(By adding a new attribute name)
- Add a couple of new options
- Try to reorder and delete some options
- See that everything works correctly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -60,6 +60,7 @@ private extension AddAttributeOptionsViewController {
 
         tableView.dataSource = self
         tableView.delegate = self
+        tableView.isEditing = true
     }
 
     func registerTableViewHeaderSections() {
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -1,5 +1,6 @@
 import UIKit
 import Yosemite
+import WordPressUI
```
ctarda: Do we really need this import? I removed it and the project seems to build as expected

ctarda (approved):
It works great. Options can be reordered (showed in the video) and reordering works as expected  after removing one or more of them


https://user-images.githubusercontent.com/2722505/106790299-5ae76480-6621-11eb-8869-7ab5c415f506.mov

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
Ecarrion: Forgot to remove it! Good catch!

Ecarrion:
Thanks @ctarda for the quick review 🙇

PR #3581, opened 2021-02-03
Shipping Labels (M2): create shipping label form

### LINKED ISSUE #2972: Shipping Labels M2: create shipping label form
Once a [button is added to create a shipping label](https://github.com/woocommerce/woocommerce-ios/issues/2971), we would need to redirect to a new screen when the button is clicked. This screen should display the origin and destination address for the shipping label.

There are five sections in this form:

- Origin address (M2)
- Destination address (M2)
- Package details (M2+M3)
- Carrier & rates (M3)
- Payment (M3)

Screenshot 2020-10-14 at 9 57 33 AM

pmusolino:
Fixes #2972

## Description
As part of M2 of Shipping Labels, I implemented the first screen of the flow that displays the form with all the various steps for creating a shipping label. At the moment, it just displays fake data, and there is no model handled because we currently don't have them. They will be implemented in later PRs. Also, the state machine that handles all the states will be implemented later as soon as we have the first entities.

## Changes
- Modified the `productPlaceholderImage` that now returns the image without the tint color. The tint color is now applied directly where the property is used. More here p91TBi-4ke
- Implemented the cell `ShippingLabelFormStepTableViewCell` that is used in this new screen for showing all the steps in the form. The cell accepts 3 types of state, each of them represents a different state of the cell (row greyed out, row enabled and tappable, row enabled and button continue visible).
- `CreateShippingLabelFormViewController` which represents the view controller of the screen. It accepts an `Order` which will be passed to the view model. Currently, it contains just the table view and the data source methods.
- Implemented `CreateShippingLabelFormViewModel` which represent the view model used in `CreateShippingLabelFormViewController`. For the moment, there is no logic inside, and it returns just sections and rows.

## Testing
This view is currently not connected with any button/action (because the button to start the flow is a WIP). So, if you want to show it, I'll suggest you in `OrderListViewController` under the method tableview `didSelectRowAt`, to substitute `navigationController?.pushViewController(orderDetailsVC, animated: true)` with 
```
let createShipController = CreateShippingLabelFormViewController(order: order)
navigationController?.pushViewController(createShipController, animated: true)
```

## Screenshots
| Light mode            |  Dark mode |
Simulator Screen Shot - iPhone 12 Pro - 2021-02-03 at 17 56 02 | Simulator Screen Shot - iPhone 12 Pro - 2021-02-03 at 17 56 17

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/ProductDetailsTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/ProductDetailsTableViewCell.swift
@@ -54,7 +54,7 @@ private extension ProductDetailsTableViewCell {
     }
 
     func configureProductImageView() {
-        productImageView.image = UIImage.productPlaceholderImage.imageWithTintColor(UIColor.listIcon)
+        productImageView.image = UIImage.productPlaceholderImage
         productImageView.tintColor = .listSmallIcon
         productImageView.contentMode = .scaleAspectFill
         productImageView.clipsToBounds = true
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/ProductDetailsTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/ProductDetailsTableViewCell.swift
@@ -1,6 +1,7 @@
 import UIKit
 import Gridicons
 import Yosemite
+import WordPressUI
 
 
 /// Product Details: Renders a row that displays a single Product.
@@ -53,7 +54,7 @@ private extension ProductDetailsTableViewCell {
     }
 
     func configureProductImageView() {
-        productImageView.image = .productPlaceholderImage
+        productImageView.image = UIImage.productPlaceholderImage.imageWithTintColor(UIColor.listIcon)
         productImageView.tintColor = .listSmallIcon
         productImageView.contentMode = .scaleAspectFill
         productImageView.clipsToBounds = true
@@ -89,7 +90,7 @@ extension ProductDetailsTableViewCell {
     func configure(item: ProductDetailsCellViewModel, imageService: ImageService) {
         imageService.downloadAndCacheImageForImageView(productImageView,
                                                        with: item.imageURL?.absoluteString,
-                                                       placeholder: .productPlaceholderImage,
+                                                       placeholder: UIImage.productPlaceholderImage.imageWithTintColor(UIColor.listIcon),
                                                        progressBlock: nil,
                                                        completion: nil)
 
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
@@ -3,10 +3,10 @@ import UIKit
 
 final class ShippingLabelFormStepTableViewCell: UITableViewCell {
 
-    @IBOutlet private weak var icon: UIImageView!
-    @IBOutlet private weak var title: UILabel!
+    @IBOutlet private weak var iconView: UIImageView!
+    @IBOutlet private weak var titleLabel: UILabel!
     @IBOutlet private weak var stackView: UIStackView!
-    @IBOutlet private weak var body: UILabel!
+    @IBOutlet private weak var bodyLabel: UILabel!
     @IBOutlet private weak var button: UIButton!
     @IBOutlet private weak var separator: UIImageView!
     @IBOutlet private weak var separatorLeadingConstraint: NSLayoutConstraint!
@@ -37,9 +37,9 @@ final class ShippingLabelFormStepTableViewCell: UITableViewCell {
                    body: String?,
                    buttonTitle: String?,
                    onButtonTouchUp: (() -> Void)? = nil) {
-        self.icon.image = icon
-        self.title.text = title
-        self.body.text = body
+        iconView.image = icon
+        titleLabel.text = title
+        bodyLabel.text = body
         button.setTitle(buttonTitle, for: .normal)
         self.onButtonTouchUp = onButtonTouchUp
         configureCellBasedOnState(state)
@@ -55,15 +55,15 @@ private extension ShippingLabelFormStepTableViewCell {
 private extension ShippingLabelFormStepTableViewCell {
     func configureStyle() {
         applyDefaultBackgroundStyle()
-        icon.tintColor = .neutral(.shade100)
+        iconView.tintColor = .neutral(.shade100)
         selectionStyle = .none
         separator.backgroundColor = .systemColor(.separator)
     }
 
     func configureLabels() {
-        title.applyBodyStyle()
-        body.applySubheadlineStyle()
-        body.numberOfLines = 0
+        titleLabel.applyBodyStyle()
+        bodyLabel.applySubheadlineStyle()
+        bodyLabel.numberOfLines = 0
     }
 
     func configureButton() {
@@ -73,23 +73,23 @@ private extension ShippingLabelFormStepTableViewCell {
     func configureCellBasedOnState(_ state: State) {
         switch state {
         case .disabled:
-            icon.alpha = 0.3
-            title.alpha = 0.3
-            body.alpha = 0.3
+            iconView.alpha = 0.3
+            titleLabel.alpha = 0.3
+            bodyLabel.alpha = 0.3
             button.isHidden = true
             separatorLeadingConstraint.constant = Constants.separatorDefaultMargin
             accessoryType = .none
         case .enabled:
-            icon.alpha = 0.6
-            title.alpha = 1.0
-            body.alpha = 0.6
+            iconView.alpha = 0.6
+            titleLabel.alpha = 1.0
+            bodyLabel.alpha = 0.6
             button.isHidden = true
             separatorLeadingConstraint.constant = Constants.separatorDefaultMargin
             accessoryType = .disclosureIndicator
         case .continue:
-            icon.alpha = 1.0
-            title.alpha = 1.0
-            body.alpha = 0.6
+            iconView.alpha = 1.0
+            titleLabel.alpha = 1.0
+            bodyLabel.alpha = 0.6
             button.isHidden = false
             separatorLeadingConstraint.constant = Constants.separatorCustomMargin
             accessoryType = .none
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
@@ -1,5 +1,6 @@
 import UIKit
 
+
 final class ShippingLabelFormStepTableViewCell: UITableViewCell {
 
     @IBOutlet private weak var icon: UIImageView!
@@ -7,6 +8,8 @@ final class ShippingLabelFormStepTableViewCell: UITableViewCell {
     @IBOutlet private weak var stackView: UIStackView!
     @IBOutlet private weak var body: UILabel!
     @IBOutlet private weak var button: UIButton!
+    @IBOutlet private weak var separator: UIImageView!
+    @IBOutlet private weak var separatorLeadingConstraint: NSLayoutConstraint!
 
     private var onButtonTouchUp: (() -> Void)?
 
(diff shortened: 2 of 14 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/ProductDetailsTableViewCell.swift
```diff
@@ -53,7 +54,7 @@ private extension ProductDetailsTableViewCell {
     }
 
     func configureProductImageView() {
-        productImageView.image = .productPlaceholderImage
+        productImageView.image = UIImage.productPlaceholderImage.imageWithTintColor(UIColor.listIcon)
```
jaclync: nit: do we still need to set `imageWithTintColor` if the `UIImageView`'s `tintColor` is set? I remember if a `UIImage` has `.alwaysTemplate` rendering mode (which Gridicons do), the color can be set in `UIImageView.tintColor`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
```diff
@@ -0,0 +1,105 @@
+import UIKit
+
+
+final class ShippingLabelFormStepTableViewCell: UITableViewCell {
+
+    @IBOutlet private weak var icon: UIImageView!
+    @IBOutlet private weak var title: UILabel!
+    @IBOutlet private weak var stackView: UIStackView!
+    @IBOutlet private weak var body: UILabel!
```
jaclync: nit: I was wondering if the `icon`, `title`, and `body` properties would read more clear if they have postfix as the UI element type like `iconView`, `titleLabel`, and `bodyLabel` respectively? this way, we also don't need `self` in the `configure` function.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
```diff
@@ -0,0 +1,105 @@
...
+    @IBOutlet private weak var button: UIButton!
+    @IBOutlet private weak var separator: UIImageView!
```
jaclync: nit: could the separator just be a `UIView`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
```diff
@@ -0,0 +1,105 @@
...
+
+    func configureCellBasedOnState(_ state: State) {
+        switch state {
+        case .disabled:
+            icon.alpha = 0.3
+            title.alpha = 0.3
+            body.alpha = 0.3
+            button.isHidden = true
+            separatorLeadingConstraint.constant = Constants.separatorDefaultMargin
+            accessoryType = .none
```
jaclync: nit: how about we have 6 `let` constants that are set in each case, so that if we ever add a new case we don't forget to set all these values? like:

```swift
let iconAlpha: CGFloat
let titleAlpha: CGFloat
...

switch state {
case .disabled:
    iconAlpha = 0.3
    titleAlpha = 0.3
    ...
case ...
}

icon.alpha = iconAlpha
title.alpha = titleAlpha
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/CreateShippingLabelFormViewController.swift
```diff
@@ -0,0 +1,236 @@
+import UIKit
+import Yosemite
+
+final class CreateShippingLabelFormViewController: UIViewController {
```
jaclync: nit: how about `ShippingLabelFormViewController` to keep the name concise, since "form" usually infers creation/editing? the table view cell is also called `ShippingLabelFormStepTableViewCell` 🙂

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/CreateShippingLabelFormViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+        case let cell as ShippingLabelFormStepTableViewCell where row.type == .shipTo:
+            configureShipTo(cell: cell, row: row)
+        case let cell as ShippingLabelFormStepTableViewCell where row.type == .packageDetails:
+            configurePackageDetails(cell: cell, row: row)
+        case let cell as ShippingLabelFormStepTableViewCell where row.type == .shippingCarrierAndRates:
+            configureShippingCarrierAndRates(cell: cell, row: row)
+        case let cell as ShippingLabelFormStepTableViewCell where row.type == .paymentMethod:
+            configurePaymentMethod(cell: cell, row: row)
+        default:
+            fatalError()
```
jaclync: nit: maybe we can show a message about the cell type and row value?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/CreateShippingLabelFormViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+            break
+        }
+    }
+
+    func configureShipFrom(cell: ShippingLabelFormStepTableViewCell, row: Row) {
+        let state = row.cellState
+        cell.configure(state: state,
```
jaclync: nit: could we just assign `row.cellState` since `state` doesn't seem to be used elsewhere in this function?

jaclync: nit: could we just assign `row.cellState` since `state` doesn't seem to be used elsewhere in this function? (same for other `configure*` functions)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/CreateShippingLabelFormViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+            }
+            else if dataState == .pending && displayMode == .editable {
+                return .continue
+            }
+            return .disabled
+        }
+    }
+
+    /// Each row has a data state
+    enum DateState {
```
jaclync: nit: did you mean `DataState`? 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/CreateShippingLabelFormViewController.swift
```diff
@@ -0,0 +1,236 @@
...
+    struct Section: Equatable {
+        let rows: [Row]
+    }
+
+    struct Row: Equatable {
+        let type: RowType
+        let dataState: DateState
+        let displayMode: DisplayMode
+
+        var cellState: ShippingLabelFormStepTableViewCell.State {
```
jaclync: nit: this could be `fileprivate`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/CreateShippingLabelFormViewModel.swift
```diff
@@ -0,0 +1,26 @@
+import UIKit
+import Yosemite
+
+/// Provides view data for Create Shipping Label, and handles init/UI/navigation actions needed.
+///
+final class CreateShippingLabelFormViewModel {
```
jaclync: nit: similar suggestion as the view controller, maybe `ShippingLabelFormViewModel`?

jaclync (approved):
Nice work Paolo! Sorry I haven't got to make a feature flag and add a CTA in order details behind the feature flag, will do that tomorrow!

From testing (and same in the screenshots you shared), I noticed the separator ends before the disclosure indicator:



maybe it's from `contentView` being shorter from the accessory view?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/ProductDetailsTableViewCell.swift
pmusolino: I used everywhere `imageWithTintColor` since I don't know if there are some cases where we used another `imageWithTintColor` (in that case, the `tintColor` doesn't work like expected). For this specific case, because we set always the same image, I followed your suggestion ;) 4ff484feb58e9cbefb012d37cf74ce5dafed6319

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
pmusolino: Sure, it make sense. Done here 80ca01fcd486e272e0b7fbd09d662f251e35da39

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
pmusolino: Updated here b29eba09d78c52eeea5cbb3afa9267840d2cbbda 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/CreateShippingLabelFormViewController.swift
pmusolino: omg yes! :D

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Cells/ShippingLabelFormStepTableViewCell.swift
pmusolino: It seems to me that in that way we complex the situation without any good benefit 🤔 I don't know honestly if it's an improvement.

pmusolino:
Thank you for all the suggestions @jaclync! I made the changes ;)
Thanks also for discovering the issue about the separator. It seems that if I connect the trailing constraint to the safe area or to the superview, nothing changes because the disclosure indicator moves the content view. Do we have other cases where this happens and where we was able to solve it?

jaclync:
> It seems that if I connect the trailing constraint to the safe area or to the superview, nothing changes because the disclosure indicator moves the content view. Do we have other cases where this happens and where we was able to solve it?

@pmusolino I remember that the product cell in the Products tab (`ProductsTabProductTableViewCell`) also has a custom separator - it looks like the separator's trailing anchor has to be pinned to the trailing anchor of the **cell** instead of `contentView`:

https://github.com/woocommerce/woocommerce-ios/blob/c48c05337330a7b1bd24b41fd275a63febd7bd63/WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift#L134

this is probably easier to set in code, I tried it in xib and it's default to the cell's safe area. I tested adding a disclosure indicator to the `ProductsTabProductTableViewCell` and the separator goes all the way to the cell's trailing edge!

Garance91540 (approved):
Waw so cool to see these coming to life! 
2 small UI feedback:
- There's no right margin for the dividers, they should span until the screen ends
- The icon for the "Ship to" section should be outlined, not filled

Screen Shot 2021-02-08 at 10 21 25 AM

Garance91540 (changes requested):
(approved by mistake)

AliSoftware:
Moving milestone to 6.1 as discussed in Slack as I'm code-freezing 6.0 today

pmusolino:
Thank you @Garance91540 for the review and for noticing the wrong house icon. 

I fixed all the things, and the PR is ready for a second check.

| Light mode             |  Dark mode |
Simulator Screen Shot - iPhone 12 Pro - 2021-02-08 at 13 07 41 | Simulator Screen Shot - iPhone 12 Pro - 2021-02-08 at 13 07 50

Garance91540 (approved):
👏👏👏

PR #3583, opened 2021-02-03
Use rethrows for waitFor()

shiki:
I learned a while ago that `rethrows` exists. Because I didn't know that before, I used `throws` when creating `waitFor()`:

https://github.com/woocommerce/woocommerce-ios/blob/bfc2481e9177e5c07e2de88c461417ac4a7c608b/TestKit/Sources/TestKit/XCTestCase%2BWait.swift#L72-L75

🤦  Now that my eyes have been opened, it's now just `rethrows`:

https://github.com/woocommerce/woocommerce-ios/blob/31a26670f48b100292caaa5dd6209cb7ce912259/TestKit/Sources/TestKit/XCTestCase%2BWait.swift#L72-L75

Callers of `waitFor` like this: 

https://github.com/woocommerce/woocommerce-ios/blob/bfc2481e9177e5c07e2de88c461417ac4a7c608b/Networking/NetworkingTests/Remote/ProductAttributeTermRemoteTests.swift#L34-L38

No longer have to use `try` if the underlying function does not throw:

https://github.com/woocommerce/woocommerce-ios/blob/31a26670f48b100292caaa5dd6209cb7ce912259/Networking/NetworkingTests/Remote/ProductAttributeTermRemoteTests.swift#L34-L38

🤸 

## Testing

The unit tests passing should be enough. 🙂 

## Author Checklist

### DIFF
--- a/Networking/NetworkingTests/Remote/PaymentsGatewayRemoteTests.swift
+++ b/Networking/NetworkingTests/Remote/PaymentsGatewayRemoteTests.swift
@@ -28,7 +28,7 @@ final class PaymentsGatewayRemoteTests: XCTestCase {
         network.simulateResponse(requestUrlSuffix: "payment_gateways", filename: "payment-gateway-list")
 
         // When
-        let result = try waitFor { promise in
+        let result = waitFor { promise in
             remote.loadAllPaymentGateways(siteID: self.sampleSiteID) { result in
                 promise(result)
             }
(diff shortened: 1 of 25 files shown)


Ecarrion (approved):
NICE!!! 🚀 

> A function or method can be declared with the rethrows keyword to indicate that it throws an error only if one of it’s function parameters throws an error.

Should be called `ifthrows` 🤪

shiki:
Thanks for the quick review!

PR #3553, opened 2021-02-01
Fixed typo in enter Store Address screen

CrazyBusyCreative:
Fixes #3551 

Did a s/'Enter the address of *your* WooCommerce store'/Enter the address of *the* WooCommerce store/

FYI - I changed the English versions of the localization strings, but didn't touch the translations.

Update release notes:
- [X] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3594, opened 2021-02-04
Add Localization.md

### LINKED ISSUE #2223: Document Localization
- We'd just need to copy @shiki's work in WPiOS

shiki:
Closes #2223

There's a contributor's PR https://github.com/woocommerce/woocommerce-ios/pull/3553 that could have used this documentation to clarify what needs to be done. 

I copied the Localization file [from WordPress](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/docs/localization.md).

## Testing

1. Find the Localization link in the [home page](https://github.com/woocommerce/woocommerce-ios/tree/docs/add-localization) and tap it. 
2. Review the content and see if it looks alright with you. 

## Author Checklist

### DIFF
--- a/docs/localization.md
+++ b/docs/localization.md
@@ -1,0 +1,103 @@
+# Localization
+
+During development, using [`NSLocalizedString()`](https://developer.apple.com/documentation/foundation/nslocalizedstring) in the code should be enough. You shouldn't need to touch the `Localizable.strings` files manually.
+
+During the release process, `NSLocalizedString` statements are scanned and stored in the `Localizable.strings` file. The file is then uploaded to [GlotPress](https://translate.wordpress.com/projects/woocommerce/woocommerce-ios/) for translation. Before the release build is finalized, all the translations are grabbed from GlotPress and saved back to the `Localizable.strings` files.
+
+## Always add Comments
+
+Always add a meaningful comment. If possible, describe where and how the string will be used. If there are placeholders, describe what each placeholder is. 
+
+```swift
+// Do
+let title = NSLocalizedString("Following %@",
+                              comment: "Title for a notice informing the user that they've successfully followed a site. %@ is a placeholder for the name of the site.")
+```
+
+```swift
+// Avoid
+let title = NSLocalizedString("Following %@", comment: "")
+```
+
+Comments help give more context to translators.
+
+## Do not use Variables
+
+Do not use variables as the argument of `NSLocalizedString()`. The string value will not be automatically picked up. 
+
+```swift
+// Do
+let myText = NSLocalizedString("This is the text I want to translate.", comment: "Put a meaningful comment here.")
+myTextLabel?.text = myText
+```
+
+```swift
+// Don't
+let myText = "This is the text I want to translate."
+myTextLabel?.text = NSLocalizedString(myText, comment: "Put a meaningful comment here.")
+```
+
+## Do not use Interpolated Strings
+
+Interpolated strings are harder to understand by translators and they may end up translating/changing the variable name, causing a crash.
+
+Use [`String.localizedStringWithFormat`](https://developer.apple.com/documentation/swift/string/1414192-localizedstringwithformat) instead.
+
+```swift
+// Do
+let year = 2019
+let template = NSLocalizedString("© %d Acme, Inc.", comment: "Copyright Notice")
+let str = String.localizeStringWithFormat(template, year)
+```
+
+```swift
+// Don't
+let year = 2019
+let str = NSLocalizedString("© \(year) Acme, Inc.", comment: "Copyright Notice")
+```
+
+## Multiline Strings
+
+For readability, you can split the string and concatenate the parts using the plus (`+`) symbol. 
+
+```swift
+// Okay
+NSLocalizedString(
+    "Take some long text here " +
+    "and then concatenate it using the '+' symbol."
+    comment: "You can even use this form of concatenation " +
+        "for extra-long comments that take the time to explain " +
+        "lots of details to help our translators make accurate translations."
+)
+```
+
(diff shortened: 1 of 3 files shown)


docs/localization.md
```diff
@@ -0,0 +1,103 @@
...
+myTextLabel?.text = myText
+```
+
+```swift
+// Don't
+let myText = "This is the text I want to translate."
+myTextLabel?.text = NSLocalizedString(myText, comment: "Put a meaningful comment here.")
+```
+
+## Do not use Interpolated Strings
```
jaclync: Do you think we can add the use of positional placeholders for every variable in a `NSLocalizedString`, from @AliSoftware's p2 post pbMoDN-1Df-p2 ?

jaclync (approved):
Thank you for adding documentation for localization! I often still forget how we write multiline strings. Just had a suggestion on adding positional placeholders to a string format!

docs/localization.md
```diff
@@ -0,0 +1,103 @@
...
+During the release process, `NSLocalizedString` statements are scanned and stored in the `Localizable.strings` file. The file is then uploaded to [GlotPress](https://translate.wordpress.com/projects/woocommerce/woocommerce-ios/) for translation. Before the release build is finalized, all the translations are grabbed from GlotPress and saved back to the `Localizable.strings` files.
+
+## Always add Comments
+
+Always add a meaningful comment. If possible, describe where and how the string will be used. If there are placeholders, describe what each placeholder is. 
+
+```swift
+// Do
+let title = NSLocalizedString("Following %@",
+                              comment: "Title for a notice informing the user that they've successfully followed a site. %@ is a placeholder for the name of the site.")
```
AliSoftware:
```suggestion
let title = NSLocalizedString("Following %1$@",
                              comment: "Title for a notice informing the user that they've successfully followed a site. %1$@ is a placeholder for the name of the site.")
```

docs/localization.md
AliSoftware: +1 for mentioning positional parameters 🙂 

_To be clear in case it wasn't, I'd suggest to have a dedicated `## Always use positional placeholders` section for that (not add a comment into an existing one)_

docs/localization.md
```diff
@@ -0,0 +1,103 @@
...
+## Do not use Interpolated Strings
+
+Interpolated strings are harder to understand by translators and they may end up translating/changing the variable name, causing a crash.
+
+Use [`String.localizedStringWithFormat`](https://developer.apple.com/documentation/swift/string/1414192-localizedstringwithformat) instead.
+
+```swift
+// Do
+let year = 2019
+let template = NSLocalizedString("© %d Acme, Inc.", comment: "Copyright Notice")
```
AliSoftware:
```suggestion
let template = NSLocalizedString("© %1$d Acme, Inc.", comment: "Copyright Notice")
```

docs/localization.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+## Pluralization
+
+GlotPress currently does not support pluralization using the [`.stringsdict` file](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html#//apple_ref/doc/uid/10000171i-CH5-SW10). So, right now, you have to support plurals manually by having separate localized strings.
+
+```swift
+struct PostCountLabels {
+    static let singular = NSLocalizedString("%d Post", comment: "Number of posts displayed in Posting Activity when a day is selected. %d will contain the actual number (singular).")
+    static let plural = NSLocalizedString("%d Posts", comment: "Number of posts displayed in Posting Activity when a day is selected. %d will contain the actual number (plural).")
```
AliSoftware:
```suggestion
    static let singular = NSLocalizedString("%1$d Post", comment: "Number of posts displayed in Posting Activity when a day is selected. %1$d will contain the actual number (singular).")
    static let plural = NSLocalizedString("%1$d Posts", comment: "Number of posts displayed in Posting Activity when a day is selected. %1$d will contain the actual number (plural).")
```

docs/localization.md
shiki: Thanks for the reminder, I added a section in 75d0233.

docs/localization.md
```diff
@@ -21,6 +21,28 @@ let title = NSLocalizedString("Following %@", comment: "")
 
 Comments help give more context to translators.
 
+## Always Use Positional Placeholders
+
+Always include the positional index of parameters. For example, instead of just using `%@` as the placeholder, use `%1$@` instead. Positional placeholders allow translators to change the order of placeholders or repeat them if necessary.
```
AliSoftware: I think it could deserve a quick mention of "apply this rule even if there's only one placeholder", because it's common to think about this only when there are multiple.

```suggestion
Always include the positional index of parameters, even if you only have one placeholder in your string. For example, instead of using `%@` as the placeholder, use `%1$@` instead. Positional placeholders allow translators to change the order of placeholders or repeat them if necessary.
```

+ Add an example, with single placeholder, in the Do / Don't examples snippet below, in addition to the ones you already wrote.

shiki: Thanks for the feedback. I thought it was obvious. 🤷 Fixed.

AliSoftware: I don't see the additional example with a single placeholder I suggested about though 😅  (But maybe I'm overthinking / overdocumenting it 😓 )

shiki: Oh, I thought it was okay because there are already a lot of examples with single placeholders in the doc. So we might be adding too much information that readers would probably just skip anyway.

AliSoftware: That's a good point. Yeah, probably not worth adding too many examples then, you're right 👍

AliSoftware (approved):

shiki:
Thanks, Jaclyn and Olivier!

PR #3596, opened 2021-02-04
Add Unit Test to Protect from Incorrect NSMappingModel Initializer Results

shiki:
We found out a while ago that the `NSMappingModel()` initializer can lie and return the incorrect mapping model (p91TBi-463-p2). 

If you follow the following steps, migrating from 40 to 41 will return the mapping model that should be used for migrating from 39 to 40. 🤦 

1. In `WooCommerceModelV39toV40`:
    a. Delete the `ProductAttributeToProductAttribute` entity mapping.
     b. In the `ProductToProduct` entity mapping, delete the `attributes` relationship mapping.

We're still not sure why this happens. We fixed this issue in https://github.com/woocommerce/woocommerce-ios/pull/3431 by changing the mapping model configuration. But this doesn't change the fact that `NSMappingModel()` can still lie. Who knows if, at some point, this was the cause of our migration issues. 🤷 

## Changes

This PR adds a unit test, [`MappingModelTests`](https://github.com/woocommerce/woocommerce-ios/blob/add/mapping-model-protection/Storage/StorageTests/CoreData/MappingModelTests.swift), that will fail if the `NSMappingModel()` initializer returns the incorrect mapping model. 

## Testing

1. In `WooCommerceModelV39toV40`:
    a. Delete the `ProductAttributeToProductAttribute` entity mapping.
     b. In the `ProductToProduct` entity mapping, delete the `attributes` relationship mapping.
2. Run the `MappingModelTests` unit tests. 

The test will fail because migrating from 40 to 41 will return the (incorrect) **39 to 40** mapping model.

## Author Checklist


pmusolino (approved):
LGTM and thanks for handling this @shiki :shipit: 

Just a thing: is it normal that the text return "From model 40 to 41" here? Do this happens because it's the next migration?

Schermata 2021-02-05 alle 15 41 22

shiki:
Thanks, @pmusolino! That's because I missed a couple of backslashes. Fixed in 107d43544f32f801b1c933ab3843ffe46583fe2a. 

image

Issue #3529, opened 2021-01-26
Variations: Add existing product attribute Management


Ecarrion:
When adding an attribute to a variation, we can select one that already exists.
**Important:** That attribute could be local(id zero) or global(id non-zero)

<img width="1239" alt="existing attributes" src="https://user-images.githubusercontent.com/562080/105901585-92d02580-5feb-11eb-8634-7d5a59636d22.png">

Tasks:
- [x] Render available options on the "Add Options" Section. For global attributes, this should be fetched using the `ProductAttributeTerm` store
- [ ] Tapping on an existing option should move it to the "Options Offered" section
- [ ] When tapping on an "X" if the option is newly created we just drop it, if the option existed before, we should move it down to the "Add Options" Section
- [ ] Make sure Local attributes are also supported!

PR #3598, opened 2021-02-04
Variations: New Attribute -> Sync global attribute options(terms)

Ecarrion:
part of #3529 

# Why
Today's turn will be to: Sync options(Called terms at API level) of a global attribute.

This PR also does not let you manage the newly synced options, that will be done in the next PR.

# How

- Updated `ProductAttribute` entity to add a couple of properties to know if it's a global or a local attribute.

- Updates `AddAttributeOptionsViewModel`: 
    - To encapsulate the two different `init` methods into a single one with a `Source` parameter, as it declares the intention better. 
    - To sync attribute options(terms) when the VM source is a global attribute.
    - To update the `updateSections()` method with the `optionsAdded` section when necessary.
    - To expose a `showGhostTableView ` property to the VC to consume.

- Updates `AddAttributeOptionsViewController`: 
    - To show a ghost table view while the terms are synched
    - To render "options added" with the correct name from the `ViewModel`.

# Demo
sync-global-attributes

# Testing Steps
- Make sure your store has some global attributes
- Tap on an existing attribute.
- See that the gosht table view is displayed
- See that the product options are displayed

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/ProductAttribute.swift
+++ b/Networking/Networking/Model/Product/ProductAttribute.swift
@@ -70,6 +70,19 @@ public struct ProductAttribute: Decodable {
     }
 }
 
+public extension ProductAttribute {
+    /// Returns weather an attribute belongs to a product(local) or to the store(global)
+    ///
+    var isLocal: Bool {
+        attributeID == 0 // Currently the only way to know if an attribute is local is if it has a zero ID
+    }
+
+    /// Returns weather an attribute belongs to a product(local) or to the store(global)
+    ///
+    var isGlobal: Bool {
+        !isLocal
+    }
+}
 
 /// Defines all the ProductAttribute CodingKeys.
 ///
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
@@ -63,7 +63,8 @@ final class AddAttributeOptionsViewModel {
         }
 
         let predicate = NSPredicate(format: "siteID == %lld && attribute.attributeID == %lld", attribute.siteID, attribute.attributeID)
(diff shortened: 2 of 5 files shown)


Networking/Networking/Model/Product/ProductAttribute.swift
```diff
@@ -70,6 +70,19 @@ public struct ProductAttribute: Decodable {
...
+    var isLocal: Bool {
+        attributeID == 0 // Currently the only way to know if an attribute is local is if it has a zero ID
+    }
+
+    /// Returns weather an attribute belongs to a product(local) or to the store(global)
+    ///
+    var isGlobal: Bool {
+        !isLocal
+    }
+}
```
pmusolino, line 85: Good idea to define two variables to identify the type of a Product Attribute 👍

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -108,6 +166,36 @@ private extension AddAttributeOptionsViewModel {
...
+        let rows = state.optionsAdded.map { term in
+            AddAttributeOptionsViewModel.Row.existingTerms(name: term.name)
+        }
+
+        return Section(header: Localization.headerExistingTerms, footer: nil, rows: rows, allowsReorder: false)
+    }
+
+    /// Synchronizes options(terms) for global attributes
+    ///
+    func synchronizeOptions() {
```
pmusolino: nit: since this method synchronize just global attributes, maybe renaming it to `synchronizeGlobalOptions` (or `synchronizeGlobalTerms`)  it's more clear

pmusolino (approved):
Looks good to me :shipit: 

Code ✅ there is just one thing. It seems that we use the naming "terms" and "options" at the same time in the same view model or view controller. Maybe it could be better to uniform all the things to `terms`. 
Testing ✅ It works as expected. Also the ghost table view.

Ecarrion:
> It seems that we use the naming "terms" and "options" at the same time in the same view model or view controller. Maybe it could be better to uniform all the things to terms.

Nice catch @pmusolino, will go over the code to make sure the same terminology is used 👍

PR #3599, opened 2021-02-04
Use the Fastlane API Key

jkmassel:
Fixes code signing when shipping `5.9` to the App Store.

**To Test:**
- Remove the fastlane/deliver username from your `.wcios-env.default` file, replacing it with PILOT_API_KEY_PATH=".configure-files/app_store_connect_fastlane_api_key.json"
- Run `bundle exec fastlane build_and_upload_itc`. Note that it still connects to Apple just fine (it'll fail though because this isn't a new release)

**Note that this change will need to be made in CircleCI environment variables prior to cutting a build, otherwise it will fail.**

Update release notes:


AliSoftware (approved):
Updated the `user.env.example` file too, which might be something you want to do on WP and SN repos as well if you didn't do it yet in the related PRs

PR #3602, opened 2021-02-05
Variations: Edit Attributes

### LINKED ISSUE #3515: Variations: Edit Attributes UI

ealeksandrov:
Closes #3515, #3516.

## Description

This PR adds second screen of variation attributes edit flow (options picker), connects it with attributes list and with data update.

## Changes

- Added `AttributeOptionListSelectorCommand` for `ListSelectorViewController` to display options for picked attribute.
- Added "Any Attribute" options on client side.
- Added view model for `AttributePickerViewController`.
- Added unit tests for `AttributePickerViewModel` and `AttributeOptionListSelectorCommand`.

## Testing

1. Open product with existing variations
2. Open variations details
3. Tap on attributes cell (added in #3519)
4. Tap any attribute and change it
5. Go to previous screen
6. Observe "Done" button appears only if any attribute is changed
7. Tap "Done" and then "Update" to save new attributes

Try full flow with changing to "Any Attribute" and from "Any Attribute".

## Screenshots

Attributes | Options
--|--

## Video

https://user-images.githubusercontent.com/3132438/107057038-0dffbb80-67e4-11eb-89b0-ab92d7b2da40.mp4

Update release notes:


Ecarrion:
Hey @ealeksandrov,

First of all, great PR 🎉 . But before reviewing any further I want to ask you if you have considered using our reusable `ListSelectorViewController`? It may help you reduce the PR complexity quite a bit!

ealeksandrov:
@Ecarrion thank you for pointing to `ListSelectorViewController`, -200 lines after a small refactor 🙂
Ready for review!

ealeksandrov:
1 thing that I've noticed - inconsistency between parent and child pickers. We don't have "Done" button on generic `ListSelectorViewController`. So if you quickly change attribute and hit back button twice - it will **not** save the change. You need to tap/swipe back first, then "done".

"Done" button initially added since it exists in design assets.

To test behaviour with 2 "done" buttons - try commit 34e9d1c.

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
```diff
@@ -0,0 +1,57 @@
+import Foundation
+import Yosemite
+
+/// Command to populate the shipping label paper size list selector
```
Ecarrion: I think you forgot to update this copy!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
```diff
@@ -0,0 +1,57 @@
...
+///
+final class AttributeOptionListSelectorCommand: ListSelectorCommand {
+
+    enum Row: Equatable {
+        case anyAttribute
```
Ecarrion: Should this be `anyOption` or even `any`?

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
```diff
@@ -39,10 +38,16 @@ private extension AttributePickerViewController {
...
-                                                           target: self,
-                                                           action: #selector(doneButtonPressed))
+    func updateDoneButton() {
+        if viewModel.isChanged {
+            navigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .done,
+                                                                target: self,
+                                                                action: #selector(doneButtonPressed))
+        } else {
+            navigationItem.rightBarButtonItem = nil
+        }
```
Ecarrion: nit: I think this would read a bit better as a `guard`
```swift
guard viewModel.isChanged else {
    navigationItem.rightBarButtonItem = nil
}

avigationItem.rightBarButtonItem = UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(doneButtonPressed))
```

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
```diff
@@ -133,14 +136,39 @@ private extension AttributePickerViewController {
...
-        // TODO-3515: Show options for attribute
+    func onAttributeOptionListSelectorCompletion(parentAttribute: ProductAttribute,
+                                                 oldAttribute: ProductVariationAttribute?,
+                                                 selectedRow: AttributeOptionListSelectorCommand.Row?) {
+        let newAttribute: ProductVariationAttribute?
+        if case .option(let selectedOption) = selectedRow {
+            newAttribute = ProductVariationAttribute(id: parentAttribute.attributeID, name: parentAttribute.name, option: selectedOption)
+        } else {
+            newAttribute = nil
+        }
```
Ecarrion: I like to use inline closures for adding initialization code like this.
```swift
let newAttribute: ProductVariationAttribute? = {
    switch selectedRow {
        case let .option(selectedOption):
            return ProductVariationAttribute(id: parentAttribute.attributeID, name: parentAttribute.name, option: selectedOption)
        default:
            return nil
    }
}()
```

That way we indicate to the reader that the code inside de closure is only needed for creating `newAttribute` and does not mess with any other part of the method

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+        variationModel.productVariation.attributes.sorted { $0.id < $1.id } != editableAttributes.sorted { $0.id < $1.id }
+    }
+
+    var resultAttributes: [ProductVariationAttribute] {
+        editableAttributes
+    }
+
+    func selectedOption(for attribute: ProductAttribute) -> ProductVariationAttribute? {
+        return editableAttributes.first(where: { $0.id == attribute.attributeID && $0.name == attribute.name })
+    }
```
Ecarrion: We tried to keep these public variables documented trough our classes, would you mind adding some to it?
Specially to `isChanged` as it does some calculation!

Ecarrion (approved):
Looking good! Just left you some small nitpicks that could make the code read a bit better!

Ecarrion:
@ealeksandrov 

> 1 thing that I've noticed - inconsistency between parent and child pickers. We don't have "Done" button on generic ListSelectorViewController. So if you quickly change attribute and hit back button twice - it will not save the change. You need to tap/swipe back first, then "done".

I think that is how it works on the rest of the app
demo-2

However, I think it's valid point to raise and discuss, would you want to P2 it?

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
ealeksandrov: I don't like `any` because it's too similar to Swift's `Any` when you quickly scan over the code. But I agree on `anyOption` 👍

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
ealeksandrov: I've decided to keep `if-else` flow because I feel that it follows `updateDoneButton()` logic (both `true` and `false` branches are "normal" flow for "updating button" 🙂) and guard would need additional `return` line.

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
ealeksandrov: That's perfect for this case, thanks for sharing! 💯

ealeksandrov:
Thanks for great feedback @Ecarrion! Will merge it now and let's discuss navigation question on p2 later on.

PR #3607, opened 2021-02-05
Variations: First Variation -> Allow existing options to be selected as offered options.

Ecarrion:
Part of #3529

# Why

Today's turn will be to: Allow existing options to be selected as offered options.

This PR only handles global attributes, local attributes will be handled on a following PR.

# How

- Update ViewModel to:
  - Add a new `selectExistingOption(at Index: Int)` method for allowing moving existing options into offered options.
  - Add a new `filterOptionsAdded() -> [ProductAttributeTerms]` method that will give us an array of existing options but filtering the ones that have already been selected.
  
- Update ViewController to:
  - Update the `Section` struct so the VM can determine which section is selectable
  - Wire `didSelectRowAtIndexPath` with the new VM interface
  - Clean the `X` image on the cell when rendering an existing option, to fix a small cell reuse issue.

# Demo
existing-global-attribute-management

# Testing Steps
- Make sure your store has some global attributes
- Tap on an existing attribute.
- Tap one(or more) existing option and see that it moves to the selected options
- Add a new option by writing the name on the text field
- Remove the existing option and see that it moves back to the existing options section
- Remove the new option and see that it disappears.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
@@ -141,6 +141,15 @@ extension AddAttributeOptionsViewModel {
         }
         state.optionsOffered.remove(at: index)
     }
+
+    /// Moves an option from at the specified index to the "Options Offered" section.
+    ///
+    func selectExistingOption(atIndex index: Int) {
+        guard let option = state.optionsAdded[safe: index] else {
+            return
+        }
+        addNewOption(name: option.name)
+    }
 }
 
 // MARK: - Synchronize Product Attribute Options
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -196,6 +211,14 @@ private extension AddAttributeOptionsViewModel {
         state.isSyncing = true
         stores.dispatch(fetchOptions)
     }
+
+    /// Returns a filtered version of `state.optionsAdded` where options that exists in `state.optionsOffered` are removed.
+    ///
+    func filterOptionsAdded() -> [ProductAttributeTerm] {
```
ealeksandrov, line 217: nit: "OptionsAdded" sounds like a data set for "options selected by users" (offered section). Maybe rename it according to section title - `filterOptionsExisting`?

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -164,20 +178,21 @@ private extension AddAttributeOptionsViewModel {
             AddAttributeOptionsViewModel.Row.selectedOptions(name: option)
         }
 
-        return Section(header: Localization.headerSelectedOptions, footer: nil, rows: rows, allowsReorder: true)
+        return Section(header: Localization.headerSelectedOptions, footer: nil, rows: rows, allowsReorder: true, allowsSelection: false)
     }
 
     func createAddedSection() -> Section? {
```
ealeksandrov, line 184: nit: same as below - "added section" doesn't mean "existing options section" on first glance.

ealeksandrov (approved):
Looks good and works well :shipit:

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
Ecarrion: Yeah, good call, I'll rename those!

PR #3610, opened 2021-02-05
Variations: First Variation -> Add local attribute options

### LINKED ISSUE #3529: Variations: Add existing product attribute Management

Ecarrion:
closes #3529 

# Why

Today's turn will be to: Support local attributes and options when creating the first variation.

# How

- Add a new `Option` enum inside our `State` struct to properly differentiate what kind of option are we dealing with, either `.local` or `.global`

- Update the `VM` `init` to populate the `state.optionsAdded` from `attribute.options` when the attribute is local.

- Update the `VM` to work with the new `Option` enum described before.

# Demo
option-existing-local-attribute

# Testing Steps

- On web, go to a variable product without any variation and add one(or more) local attributes.
- On the app, navigate to a variable product with no variations added
- Tap on the local attribute you just created
- Tap one(or more) existing option and see that it moves to the selected options
- Add a new option by writing the name on the text field
- Remove the existing option and see that it moves back to the existing options section
- Remove the new option and see that it disappears.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
@@ -18,14 +18,21 @@ final class AddAttributeOptionsViewModel {
 
     /// Defines the necessary state to produce the ViewModel's outputs.
     ///
-    private struct State {
+    fileprivate struct State {
+        /// Bundles the attributes options that are supported
+        ///
+        enum Option {
+            case local(name: String)
+            case global(option: ProductAttributeTerm)
+        }
+
         /// Stores the options to be offered
         ///
-        var optionsOffered: [String] = []
+        var optionsOffered: [Option] = []
 
         /// Stores options previously added
         ///
-        var optionsAdded: [ProductAttributeTerm] = []
+        var optionsAdded: [Option] = []
 
         /// Indicates if the view model is syncing a global attribute options
         ///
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -18,14 +18,21 @@ final class AddAttributeOptionsViewModel {
...
+
         /// Stores the options to be offered
         ///
-        var optionsOffered: [String] = []
+        var optionsOffered: [Option] = []
 
         /// Stores options previously added
         ///
-        var optionsAdded: [ProductAttributeTerm] = []
+        var optionsAdded: [Option] = []
```
jaclync: nit: I first thought `optionsAdded` is for the options added by the user on this Add Options screen, would it be less confusing if this variable is called like `options`/`allOptions`/`optionsAvailable`? The "Add" action in "AddAttributeOptions*" actually means adding an existing option to be offered, if I understand correctly?

jaclync (approved):
LGTM! :shipit: all the actions worked as expected!

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
Ecarrion: I'll rename those properties in order for it to be more clear! Thanks for flagging that up! :-)

PR #3611, opened 2021-02-07
Use API Key for `deliver`

jkmassel:
Use the App Store Connect API when doing builds based on `fastlane deliver` (which is the _preferred_ method across our projects)

Update release notes:


AliSoftware (commented):
Probably worth:
- re-editing the `user.env.example` to remove unnecessary DELIVER env var, avoiding having that value set in two different places (better have a single source of truth than multiple!)
- maybe even remove the lane that only calls deliver. Because keeping the two different ways (calling `fastlane deliver` directly, vs calling the lane) means we'd also have 2 different things to maintain and keep in sync (which is what led to this discrepancy in the first place), and besides that lane doesn't exist in other repos – and I didn't even knew it existed until yesterday when I looked at those API Key changes as I've always been using `fastlane deliver` directly – so that'd be a good occasion to make progress towards our consistency goal

jkmassel:
> re-editing the user.env.example to remove unnecessary DELIVER env var, avoiding having that value set in two different places (better have a single source of truth than multiple!)

 ✅ in 94d55173e0707f95adfa106937e01eaaff56e19b

> maybe even remove the lane that only calls deliver. Because keeping the two different ways (calling fastlane deliver directly, vs calling the lane) means we'd also have 2 different things to maintain and keep in sync (which is what led to this discrepancy in the first place), and besides that lane doesn't exist in other repos – and I didn't even knew it existed until yesterday when I looked at those API Key changes as I've always been using fastlane deliver directly – so that'd be a good occasion to make progress towards our consistency goal

✅ in dce30c2b1d4e62cb43e7c85edbd9a92099bf921d

Also in f12997979efe794b1f887336fe61b89a05ec5965 – applied it to the `Deliverfile` in order to ensure it works when using `fastlane deliver` instead of the lane.

AliSoftware:
@jkmassel What do you want to do with this PR?

* I see the other fixes you mention are on the `fix/code-signing` branch, not on this `update/deliver-for-app-store-connect-api` branch.
* I also see that f12997979efe794b1f887336fe61b89a05ec5965 is referencing the config file via `../.configure-files` while the similar commit in this PR references it via `./configure-files`. I never remember what the paths are supposed to be relative to, but I think from what we discussed recently, that when referenced from the `Fastfile` (and I assume, similar for the `Deliverfile`?), the `pwd` is `fastlane/`, while when they are from within an action (e.g. from release-toolkit) they are relative to the repo root. So I think f12997979efe794b1f887336fe61b89a05ec5965 is the correct fix, as opposed to what we have in the PR?

In light of this, should we close this PR in favor of a new one merging `fix/code-signing` (once its ready) into `release/6.0`? Or keep both but potentially fix the `.configure-files` path in this PR?

jkmassel:
👋 I've synced the changes into this PR, so we can get rid of the other branch.

The path is correct in this PR – you can verify this by running `bundle exec fastlane deliver download` – and note that it says `Creating authorization token for App Store Connect API` in the output (at that point you could just cancel it if you like).

I think it's ready to land now, but LMK if not!! 😅

.configure-files/project.env.enc
```diff
@@ -1,2 +1,3 @@
...
�r�/D�ӛ םi�Ҋ��i �+~T�58"2�\T1��~H���;^�t��=eoJ�
\ No newline at end of file
+�1^����:8��'E����
���-ث	W����v�Ƕ9�Q
�吳���^�$�>�j�,�=#�^Fl`{�:r:!5��G;�`�9��O��AG��
��a.`� �-
�r�/D�ӛ םi�Ҋ��i �+~T�58"2�\T1��~H���(�'���w��	�M�
+/��)�x�v=�,L��*Ow��ڄ�J��&$�e{r�TlV��
0���B��*N�"-H�
c���PK/�*1���
```
AliSoftware, line 3: This encrypted content reminded me of Q*bert 😆 

image

AliSoftware (changes requested):
This still misses the update and removal of now useless keys in the `fastlane/env/user.env.example` file:

 - The `PILOT_API_KEY_PATH` env var suggested to be set there is now useless now that this value is set in the `project.env` from mobile secrets.
 - The `DELIVER_API_KEY_PATH` env var suggested to be set there is now useless now that this value is provided via the `Deliverfile`.
 - The `SIGH_API_KEY_PATH` (used by `match`) has not yet been moved out of this env file… but we should. i.e. especially providing this value into the `Matchfile` instead (like we did for `DELIVER` and the `Deliverfile`) and then removing it from the `user.env.example`.

jkmassel:
Addressed in 86f5fa5, 6cdc178, and 22d4f9e, thanks!!

.configure-files/project.env.enc
jkmassel: Hopefully just as unintelligible!! 😅

AliSoftware (approved):

AliSoftware:
Merge on green! 🟢 :shipit:

PR #3615, opened 2021-02-08
Add a new Swift Package: Hardware

### LINKED ISSUE #3586: [Mobile Payments] New Swift Package
Setup a new local Swift Package to encapsulate the logic necessary to directly communicate with the Stripe Terminal SDK.

Scope: 
* A new local package
* Ideally, the package will not contain any code at this stage. 

For context: p91TBi-4k7-p2 and p91TBi-4il-p2

ctarda:
Closes #3586 
For context: p91TBi-4k7-p2 and p91TBi-4il-p2

This PR adds a new Swift Package to the project, that @allendav and myself would use as the base to start building our prototype, as mentioned in p91TBi-4k7-p2. This PR does not introduce any user-facing changes.

⚠️  There is one struct in this PR that will get removed as soon as we start actually adding code to the package. But for now, it helps check that the PR passes CI, tests can be run and code in the package can be used. The aforementioned struct is marked as disposable in the headerdocs ⚠️ 

## Changes
* Added a new swift package, named Hardware. Name might be revised later on, as we progress. 
* Made WooCommerce's main target link against Hardware.
* Edited WooCommerce's test plan to run also the unit tests corresponding to Hardware.

## How to test
* Check that CI is ✅ 
* Check out the branch, and run all the tests (i.e. pressing command+U). One way to check that the tests in the package are run might be adding a breakpoint in `HardwareTests.test_that_tests_run` and check that the breakpoint is hit.
* Build and run (i.e. command + R). Check that the project builds. For extra credit, `import Hardware` into any other class in the WooCommerce target (for example, AppDelegate), add something like the following anywhere in that class and check that "Hello World" is printed to the console:
```
        let salutation = Hardware().text
        print("salutation ", salutation)
```

Update release notes:

### DIFF
--- a/Hardware/Package.swift
+++ b/Hardware/Package.swift
@@ -1,0 +1,28 @@
+// swift-tools-version:5.3
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "Hardware",
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
+        .library(
+            name: "Hardware",
+            targets: ["Hardware"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
+        .target(
+            name: "Hardware",
+            dependencies: []),
+        .testTarget(
+            name: "HardwareTests",
+            dependencies: ["Hardware"]),
+    ]
+)
--- a/Hardware/Package.swift
+++ b/Hardware/Package.swift
@@ -5,6 +5,7 @@ import PackageDescription
 
 let package = Package(
     name: "Hardware",
+    platforms: [.iOS(.v13)],
     products: [
         // Products define the executables and libraries a package produces, and make them visible to other packages.
         .library(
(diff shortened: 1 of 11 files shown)


Hardware/Package.swift
```diff
@@ -0,0 +1,28 @@
+// swift-tools-version:5.3
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "Hardware",
```
shiki, line 7: Not sure if this matters. Should we define the `platforms`? 

```suggestion
    name: "Hardware",
    platforms: [.iOS(.v13)],
```

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce.xcscheme
```diff
@@ -120,8 +120,8 @@
       </BuildableProductRunnable>
       <CommandLineArguments>
          <CommandLineArgument
+            argument = ""
             isEnabled = "NO">
-            argument = "mocked-network-layer"
          </CommandLineArgument>
```
shiki: This looks like an unintentional change. This is now empty in the scheme. This does not seem to affect the screenshots tests though. I think it's just used for the development of the Screenshots package.

shiki (approved):
It works! 

All the unit tests ran and I can also **make it fail**: 



---

I left some comments that may need some attention but feel free to merge whenever you're ready.

Hardware/Package.swift
ctarda: Good catch. It doesn't hurt to be explicit.

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce.xcscheme
ctarda: Interesting. I'll revert this. Thanks!

ctarda:
Thanks for the review @shiki I've incorporated your feedback, and I'll merge if/when CI gives the ✅

Issue #3536, opened 2021-01-27
Variations: Attribute List UI


Ecarrion:
After creating a new product attribute(#3535), the user should be redirected to this screen 

<img width="414" alt="Screen Shot 2021-01-27 at 2 26 40 PM" src="https://user-images.githubusercontent.com/562080/106042978-c4a8c100-60ab-11eb-8954-9d38cd86cb14.png">

Tasks:
- [ ] Build UI: Cell + Add attribute button + Done button CTA
- [ ] Connect "Next" CTA(after updating the product) to this screen

PR #3616, opened 2021-02-08
Variations: First Variation -> Update / Create attributes remotely 

### LINKED ISSUE #3535: Variations: Link attribute to Product

Ecarrion:
closes #3535 

# Why
Today's turn will be to: Update/Create attributes with their options.

**Note:** This PR does not navigate anywhere after the remote operation has been finished. That, will be handled in #3536  

# How

- Updated `Product` and `ProductAtrribute` to add `codable` conformance to it.

- On the ViewModel:
  - Add `Product` as a dependency because it's needed to update the product attributes later.
  - Rename `Source` enum to `Attribute` as it represents the intention better.
  - Added a `showUpdateIndicator` property which will be `true` while the product operation is happening.
  - Added a new `updateProductAttributes` method that will trigger the remote update by updating the existing attribute with the selected options or creating the attribute if it didn't exist before.
  
- On the ViewController:
  - Listen to the VM's `showUpdateIndicator` to decide when to show the "next" button and when to show the loading indicator.
  - Call the `ViewModel.updateProductAttributes` when pressing the next button
  - Show an error notice if the product update fails.

# Demo

## Success
App | Web
--- | ---
demo | Screen Shot 2021-02-08 at 3 14 55 PM

## Fail
demo-error

# Testing Steps

- Tap on the "Next" button
- See that a loading spinner appears instead of the next button.
- When the loading finishes, check on the web your product attribute and corroborate that it updated successfully.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/Product.swift
+++ b/Networking/Networking/Model/Product/Product.swift
@@ -508,6 +508,9 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         // Linked Products (Upsells and Cross-sell Products)
         try container.encode(upsellIDs, forKey: .upsellIDs)
         try container.encode(crossSellIDs, forKey: .crossSellIDs)
+
+        // Attributes
+        try container.encode(attributes, forKey: .attributes)
     }
 }
 
--- a/Networking/Networking/Model/Product/ProductAttribute.swift
+++ b/Networking/Networking/Model/Product/ProductAttribute.swift
@@ -3,7 +3,7 @@ import Foundation
 
 /// Represents a ProductAttribute entity.
 ///
-public struct ProductAttribute: Decodable {
+public struct ProductAttribute: Codable {
     public let siteID: Int64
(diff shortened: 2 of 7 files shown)


ealeksandrov (approved):
LGTM :shipit:

On UI side I can hit back and leave the screen while spinner is active. It's fine for user to abandon process if they wish but it should work correctly when forward navigation is added later.

Error state reproduced with internet connection turned off.

Ecarrion:
Thanks @ealeksandrov for the review, I'll add the back button situation as a task so I don't forget!

PR #3620, opened 2021-02-09
Products: Fix dynamic type support for details text

### LINKED ISSUE #3614: Products: Variation list needs dynamic type support for details text
**Describe the bug**

While testing https://github.com/woocommerce/woocommerce-ios/issues/3600 I noticed that the details text in the variations list (each variation's stock status and price) does not adjust to font size changes dynamically. The font size only changes after relaunching the app.

**To Reproduce**
1. Go to the Products tab.
2. Select a variable product.
3. Select Variations and notice you see a list of the product variations.
4. In the device's Settings app, go to Accessibility > Display & Text Size > Larger Text and use the slider to increase the font size.
5. Go back to the app and notice that the details text (stock status and price) for each variation hasn't changed size.
6. Close and relaunch the app, and notice that now the details text is a larger size.

**Screenshots**

Regular font size | Large font size | Large font size after relaunch
-|-|-
Simulator Screen Shot - iPhone 12 - 2021-02-08 at 14 59 16|Simulator Screen Shot - iPhone 12 - 2021-02-08 at 14 58 46|Simulator Screen Shot - iPhone 12 - 2021-02-08 at 15 00 22

**Expected behavior**

The details text should adjust size automatically; it probably just needs `adjustsFontForContentSizeCategory` set to `true`. (We have this for all label styles in `UILabel+Helpers` but this label has extra/conditional attributes so it isn't using those helpers.)

momo-ozawa:
Resolves https://github.com/woocommerce/woocommerce-ios/issues/3614

### Notes

> The details text should adjust size automatically; it probably just needs adjustsFontForContentSizeCategory set to true. (We have this for all label styles in UILabel+Helpers but this label has extra/conditional attributes so it isn't using those helpers.)

I followed this suggestion and set`adjustsFontForContentSizeCategory` to true, but unfortunately this didn't resolve the issue.
 
I'm not totally sure why directly calling `StyleManager.footerLabelFont` instead of `Style.detailsFont` fixes the issue.

### Testing instructions
1. Go to the Products tab.
2. Select a variable product.
3. Select Variations and notice you see a list of the product variations.
4. In the device's Settings app, go to Accessibility > Display & Text Size > Larger Text and use the slider to increase the font size.
5. Go back to the app.
6. ✅ Notice that the details text (stock status and price) for each variation has changed to a larger size.

### Demo
https://user-images.githubusercontent.com/6711616/107309347-9b395d80-6acd-11eb-8c57-6d767d7e5290.mov

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductsTabProductViewModel+ProductVariation.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductsTabProductViewModel+ProductVariation.swift
@@ -22,7 +22,7 @@ private extension EditableProductVariationModel {
         let detailsAttributedString = NSMutableAttributedString(attributedString: stockStatusAttributedString)
         detailsAttributedString.append(NSAttributedString(string: " • ", attributes: [
             .foregroundColor: UIColor.textSubtle,
-            .font: Style.detailsFont
+            .font: StyleManager.footerLabelFont
         ]))
         detailsAttributedString.append(variationStatusOrPriceAttributedString)
         return NSAttributedString(attributedString: detailsAttributedString)
@@ -33,7 +33,7 @@ private extension EditableProductVariationModel {
         return NSAttributedString(string: stockText,
                                   attributes: [
                                     .foregroundColor: UIColor.textSubtle,
-                                    .font: Style.detailsFont
+                                    .font: StyleManager.footerLabelFont
         ])
     }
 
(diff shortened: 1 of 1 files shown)


jaclync (approved):
wow, nice fix! maybe a dynamic `UIFont` isn't meant to be reused?

also, you can tag @woocommerce/ios-developers as reviewer next time and anyone on the team can take a look 🙂 

Code: ✅ 
Testing: ✅ 

before | after
-- | --
Simulator Screen Shot - iPhone 11 - 2021-02-09 at 15 23 27 | Simulator Screen Shot - iPhone 11 - 2021-02-09 at 15 28 15

PR #3622, opened 2021-02-09
Variations: First Variation -> Notify when an attribute is updated

### LINKED ISSUE #3618: Variations: Encapsulate new/edit attribute flow
Think about a lightweight coordinator object to handle the add/edit attribute flow. 

It should: 
- [ ] Start on the `AddAttributeViewController`
- [ ] End on the `AttributeListViewController`
- [ ] Provide a completion block when the block is finished

Needs to be spawed from:
- [ ] Variation list (Empty state and edit attributes)
- [ ] Attribute List

Ecarrion:
closes #3618  

# Why

Today's turn will be to: **Add completion blogs to notify when a product attribute is updated/created**

Note: This PR, temporarily drives you back to the empty state screen, #3536 will be in charge of displaying the correct screen later.

# How

- Add `onCompletion` closures to `AddAttributeViewController` and `AddAttributeOptionsViewController`
- Set that `onCompletion` closure on  `ProductVariationsViewController` and update the `product` property when the flow is finished.

# Demo
demo

# Testing Steps
- Tap on the "Next" button
- See that a loading spinner appears instead of the next button.
- When the loading finishes, see that you are redirected back to the empty state screen.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -21,8 +21,10 @@ final class AddAttributeOptionsViewController: UIViewController {
         self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
     }
 
-    /// Init
+    /// Initializer for `AddAttributeViewController`
     ///
+    /// - Parameters:
+    ///   - onCompletion: Closure to be invoked(with the updated product)  when the update/create attribute operation finishes successfully.
     init(viewModel: AddAttributeOptionsViewModel,
          noticePresenter: NoticePresenter = ServiceLocator.noticePresenter,
          onCompletion: @escaping (Product) -> Void) {
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -21,7 +21,7 @@ final class AddAttributeOptionsViewController: UIViewController {
         self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
     }
 
-    /// Initializer for `AddAttributeViewController`
+    /// Initializer for `AddAttributeOptionsViewController`
     ///
     /// - Parameters:
     ///   - onCompletion: Closure to be invoked(with the updated product)  when the update/create attribute operation finishes successfully.
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -11,6 +11,10 @@ final class AddAttributeOptionsViewController: UIViewController {
 
     private let noticePresenter: NoticePresenter
 
+    /// Closure to be invoked(with the updated product)  when the update/create attribute operation finishes successfully.
+    ///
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -11,17 +11,26 @@ final class AddAttributeOptionsViewController: UIViewController {
 
     private let noticePresenter: NoticePresenter
 
+    /// Closure to be invoked(with the updated product)  when the update/create attribute operation finishes successfully.
```
rachelmcr, line 14: These comments are super helpful! 👏

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -11,17 +11,26 @@ final class AddAttributeOptionsViewController: UIViewController {
...
+    private let onCompletion: (Product) -> Void
+
     /// Keyboard management
     ///
     private lazy var keyboardFrameObserver: KeyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
         self?.handleKeyboardFrameUpdate(keyboardFrame: keyboardFrame)
     }
 
-    /// Init
+    /// Initializer for `AddAttributeViewController`
```
rachelmcr: Should this be `AddAttributeOptionsViewController` (instead of `AddAttributeViewController`)?

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
```diff
@@ -269,19 +277,15 @@ extension AddAttributeViewController {
...
-        navigationController?.pushViewController(addAttributeOptionsVC, animated: true)
-    }
-
-    private func presentAddAttributeOptions(for existingAttribute: ProductAttribute) {
-        let viewModel = AddAttributeOptionsViewModel(product: self.viewModel.product, attribute: .existing(attribute: existingAttribute))
-        let addAttributeOptionsVC = AddAttributeOptionsViewController(viewModel: viewModel)
-        navigationController?.pushViewController(addAttributeOptionsVC, animated: true)
+    /// Presents `AddAttributeOptionsViewController` and passes the same `onCompletion` closure, for our presenterVC  to handle.
+    ///
+    private func presentAddAttributeOptions(attribute: AddAttributeOptionsViewModel.Attribute) {
```
rachelmcr: What do you think of keeping `for` as the argument label for `attribute` here? As in:

```swift
private func presentAddAttributeOptions(for attribute: AddAttributeOptionsViewModel.Attribute) {
```

For me, that makes it easier to understand what's happening when you call it elsewhere, for example in line 187 it could read like a sentence:

```swift
presentAddAttributeOptions(for: .existing(attribute: attribute))
```

rachelmcr (approved):
Looks great and tests well! 👍 

I had a couple questions/suggestions noted below but nothing critical.

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
Ecarrion: Yup, copy paste mistake!

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
Ecarrion: Thanks for suggesting this! English has always been a struggle for me!

rachelmcr (approved):
🎉  :shipit:

Issue #2969, opened 2020-10-14
Woo: Shipping Labels - add endpoint to validate the origin and destination address


anitaa1990:
As part of the [shipping labels project](https://github.com/woocommerce/woocommerce-ios/projects/24), we need to add an endpoint to validate the origin and destination address.

pmusolino:
Related P2 post p91TBi-42G

pmusolino:
- [x] Networking layer
- [x] Yosemite layer

PR #3623, opened 2021-02-09
Networking layer: added Shipping Labels address validation endpoint

pmusolino:
Part of #2969

## Description
As part of Shipping Labels Milestone 2 we need to validate the address (origin and destination) for the Shipping Labels.
In this PR, I implemented the networking layer to verify an address. Unlike the other endpoints in the app, the response can contain the verified address or a list of errors.

## Changes
- `ShippingLabelAddress` now conform to `Codable`.
- Implemented the new model `ShippingLabelAddressVerification` that contains the parameters that should be passed to the validation endpoint.
- Implemented the new model `ShippingLabelAddressValidationResponse` that parses the response of the validation endpoint, that can return the `address` or a list of `errors` (two for the moments).
- Implemented the new model `ShippingLabelAddressValidationError` that parses the `field_errors` under the validation response in case of error during the validation.
- Implemented `ShippingLabelAddressValidationResponseMapper` that convert a dictionary into `ShippingLabelAddressValidationResponse`.
- Implemented the method `addressValidation` under `ShippingLabelRemote` + tests.
- Added two JSON sample responses `shipping-label-address-validation-error.json` and `shipping-label-address-validation-success.json`.
- implemented the address validation result under `MockShippingLabelRemote`.
 
## Testing
- Just CI, for the moment this code is not used and will be useful later.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationError.swift
+++ b/Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationError.swift
@@ -1,0 +1,31 @@
+import Foundation
+
+/// Represents Shipping Label Address Validation Error.
+///
+public struct ShippingLabelAddressValidationError: Equatable {
+    public let address: String?
+    public let general: String?
+
+    public init(address: String?, general: String?) {
+        self.address = address
+        self.general = general
+    }
+}
+
+extension ShippingLabelAddressValidationError: Decodable {
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let address = try container.decodeIfPresent(String.self, forKey: .address)
+        let general = try container.decodeIfPresent(String.self, forKey: .address)
+        self.init(address: address, general: general)
+    }
+}
+
+/// Defines all of the ShippingLabelAddressValidationError CodingKeys
+///
+private extension ShippingLabelAddressValidationError {
+    enum CodingKeys: String, CodingKey {
+        case address
+        case general
+    }
+}
--- a/Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationError.swift
+++ b/Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationError.swift
@@ -16,7 +16,7 @@ extension ShippingLabelAddressValidationError: Decodable {
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         let address = try container.decodeIfPresent(String.self, forKey: .address)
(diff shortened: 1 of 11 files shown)


Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationError.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+
+/// Represents Shipping Label Address Validation Error.
+///
+public struct ShippingLabelAddressValidationError: Equatable {
+    public let address: String?
+    public let general: String?
```
ealeksandrov: Does this property hold error message/description? It's not obvious from its name - maybe rename or add a comment?

Yosemite/YosemiteTests/Mocks/Networking/Remote/MockShippingLabelRemote.swift
```diff
@@ -23,6 +23,10 @@ final class MockShippingLabelRemote {
         let shippingLabelID: Int64
     }
 
+    private struct AddressValidationResultKey: Hashable {
+        let siteID: Int64
+    }
```
ealeksandrov, line 28: Question: that's expected limitation, since addresses don't have additional ids?
From what I understand - it won't allow to map different address validation responses within a same site id. 

I think that validation success and fail will be separated test cases any way, so not a blocker 🙂

ealeksandrov (approved):
Looks great 🚀
Just added couple of clarification nitpicks/questions.

Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationError.swift
ctarda: Good point. The name is definitely not very descriptive. I think renaming it is a good idea

ctarda (approved):
👏 👏 

:shipit:

Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationError.swift
pmusolino: Yes, I tried to follow the model as much as possible because the struct already contains the word error. But for sure I can rename them to `addressError` and `generalError`. 👍

Yosemite/YosemiteTests/Mocks/Networking/Remote/MockShippingLabelRemote.swift
pmusolino: Yes correct. The Address Validation doesn't need any other info, this is the reason why I used just the `siteID`. 👍

pmusolino:
Thank you both for the review 💯

Issue #3557, opened 2021-02-01
Add mock powered placeholder list in settings for discovered readers


allendav:
As an initial step toward reader connection, this issue covers the work needed to display (using a non-final user interface, accessed from the store settings) a list of nearby mock bluetooth readers, updated as a Mock Stripe Terminal SDK discovers them.

Discovery should start on opening the view and end on closing it.

The list should include the name (aka serial number) of the reader and be sorted by signal strength (strongest near the top)

There is no expectation to support actual readers (the discovery will be mocked), persist discovered readers, support detail views or make connections in the scope of this issue.

p91TBi-4i6-p2

allendav:
I don't think we need this anymore. We are well past this.

PR #3624, opened 2021-02-09
Adds a placeholder reader setting VC behind the card present feat flag

allendav:
Partially addresses #3557 

This is the first in a series of pulls for #3557 

To test:
- from the My Store tab, open settings (gear)
- from the Settings view, tap on Manage card readers
- from the Card Reader view, tap on the back arrow

Screenshots:

iPhone XR, Portrait, Dark Mode

iPad 10.2", Portrait, Light Mode

Update release notes:

cc @Garance91540

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -141,11 +141,14 @@ private extension SettingsViewController {
 
         sections = [
             Section(title: selectedStoreTitle, rows: storeRows, footerHeight: CGFloat.leastNonzeroMagnitude),
-            Section(title: nil, rows: [.support], footerHeight: UITableView.automaticDimension),
         ]
 
+        /// Iif we are going to show a card reader section we want a minimal section footer height for the support section above it.
         if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentPayments) {
+            sections.append(Section(title: nil, rows: [.support], footerHeight: CGFloat.leastNonzeroMagnitude))
             sections.append(Section(title: nil, rows: [.cardReaders], footerHeight: UITableView.automaticDimension))
+        } else {
+            sections.append(Section(title: nil, rows: [.support], footerHeight: UITableView.automaticDimension))
         }
 
         if couldShowBetaFeaturesRow() {
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -229,7 +229,7 @@ private extension SettingsViewController {
     func configureCardReaders(cell: BasicTableViewCell) {
         cell.accessoryType = .disclosureIndicator
         cell.selectionStyle = .default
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -139,32 +139,30 @@ private extension SettingsViewController {
...
 
-        let otherSection: Section
-        #if DEBUG
-        otherSection = Section(title: otherTitle, rows: [.appSettings, .wormholy], footerHeight: CGFloat.leastNonzeroMagnitude)
-        #else
-        otherSection = Section(title: otherTitle, rows: [.appSettings], footerHeight: CGFloat.leastNonzeroMagnitude)
-        #endif
+        sections = [
+            Section(title: selectedStoreTitle, rows: storeRows, footerHeight: CGFloat.leastNonzeroMagnitude),
+            Section(title: nil, rows: [.support], footerHeight: UITableView.automaticDimension),
```
ctarda: The distance between the cells for "Help and Support" and "manage Card Readers" seems too high, specially compared to the distance between "Switch Store" and "Help And Support". 



The footerHeight for the support section might need to be `CGFloat.leastNonzeroMagnitude` as well when we present the "Manage card readers" row

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -223,6 +223,12 @@ private extension SettingsViewController {
         cell.textLabel?.text = NSLocalizedString("Help & Support", comment: "Contact Support Action")
     }
 
+    func configureCardReaders(cell: BasicTableViewCell) {
+        cell.accessoryType = .disclosureIndicator
+        cell.selectionStyle = .default
+        cell.textLabel?.text = NSLocalizedString("Manage card readers", comment: "Manage card readers Action")
```
ctarda: The comment passed to NSLocalizedString is meant to help translators figure out the context for that screen. Would it make sense for this to read "Navigates to List of Card Readers screen" (or similar, but better worded)?

ctarda (approved):
👏 👏 

The code looks great. this is how it looks in dark mode:


And this is how it looks with the feature flag removed:




There is a minor issue with the spacing between the new cell and the one above it, but once that's fixed, :shipit: 

Also, you might need to register the new tracks event added. (P91TBi-3Py-p2 for context)

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3624/commits/466de0b9d7cf649217ebc0382809c755f5e62528

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3624/commits/231ca271cc5b2a580c63c90ff89f9d6f5ad90fa3

allendav:
https://github.com/Automattic/tracks-events-registration/pull/450

PR #3625, opened 2021-02-09
[Card Present Payments] Scaffolding CardReaderService

### LINKED ISSUE #3589: [Mobile Payments] Define interface of ReaderService
The Stripe Terminal SDK exposes all its functionality via a class called [Terminal](https://stripe.com/docs/terminal/sdk/ios#initialize)

The SDK also provides a set of protocols to notify clients of relevant changes in state (i.e. DiscoveryDelegate)

Scope:
* Define a protocol that [adapts](https://en.wikipedia.org/wiki/Adapter_pattern) the Stripe Terminal SDK APIs in a way that the implementation details of the direct communication with the Stripe SDK are abstracted out rom the rest of the app.
* Add model objects: i.e. PaymentIntent, Reader, Error handling
* Add as little code as possible to just pass the build

The new abstractions would belong in the swift package introduced in #3586

For context: p91TBi-4k7-p2

ctarda:
Closes #3589 

For reference: p91TBi-4q4-p2

This is a very first iteration of the basic abstractions we are going to use to implement the integration with the Stripe Terminal SDK.

Stripe Terminal Reference: https://stripe.dev/stripe-terminal-ios/docs/index.html

It is important to remember that we do not want to mimic the API of the Stripe Terminal SDK, but to provide an adapter wrapping it, so we want to provide methods in the API to:
* start
* cancel
* create payment intents
* process payment intents
* cancel payment intents
* update the reader (checking for updates and updating the reader is intentionally left out of this PR)

What is a payment item? I'm glad you asked! According to [the documentation](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntent.html):

> A PaymentIntent tracks the process of collecting a payment from your customer. We recommend that you create exactly one PaymentIntent for each order or customer session in your system. You can reference the PaymentIntent later to see the history of payment attempts for a particular session.

> A PaymentIntent transitions through multiple statuses throughout its lifetime and ultimately creates at most one successful charge.

Things to note:
* These are just empty types for now. There is no actual implementation.
* Most of the actual model objects do not declare any properties now. We will get to that later.

## Changes
* Added a protocol, CardReaderService that would be the public facade of our integration with the Stripe Terminal SDK. This protocol is supposed to provide high-level operations and Publishers that clients can observe to inspect the status of the reader.
* Added an empty implementation of the `ConnectionTokenProvider` protocol.
* Added an implementation of the CardReaderService that will end up being the actual integration with the Stripe Terminal SDK. This contains enough (not really functional or useful in any way) code to pass the build.
* Added enough model objects to pass the build. Some of them are empty, some others are enumerations that might be close to their final state.

## How to test
There really isn't much to test. None of this code is executed, but it is something we want to get in as early as possible to start building our prototype (as mentioned in p91TBi-4k7-p2)

Anyway please:
* 👁️  at the code
* Make sure that CI gives the ✅ 

Update release notes:


Hardware/Sources/Hardware/CardReaderEvent.swift
```diff
@@ -0,0 +1,14 @@
+/// The possible events from a connected reader.
+public enum CardReaderEvent {
+    // A card was inserted.
+    case cardInserted
+
+    // A card was removed.
+    case cardRemoved
+
+    // Low battery warning.
+    case lowBattery
```
allendav, line 10: Maybe add lowBatteryResolved. Maybe also add softwareUpdateNeeded and softwareUpToDate. OK to wait for next iteration for all of this.

Hardware/Sources/Hardware/CardReaderService.swift
```diff
@@ -0,0 +1,58 @@
...
+    var paymentStatus: AnyPublisher<PaymentStatus, Never> { get }
+
+    /// The Publisher that emits reader events
+    var readerEvent: AnyPublisher<CardReaderEvent, Never> { get }
+
+    // MARK: - Commands
+
+    /// Starts the service.
+    /// That could imply, for example, that the reader discovery process starts
+    func start()
```
allendav, line 29: In the next iteration of this abstraction, let's have initialization, discovery and connection be things handled separately.

Hardware/Sources/Hardware/CardReaderService.swift
```diff
@@ -0,0 +1,58 @@
...
+    func createPaymentIntent(_ parameters: PaymentIntentParameters) -> Future <PaymentIntent, Error>
+
+    /// Collects a payment method for the given intent.
+    func collectPaymentMethod(_ intent: PaymentIntent) -> Future<PaymentIntent, Error>
+
+    /// Processes a payment after collecting a payment method succeeds.
+    /// - Parameter intent: the payment intent
+    func processPaymentIntent(_ intent: PaymentIntent) -> Future<PaymentIntent, Error>
+
+
```
allendav, line 54: We're probably going to need capturePaymentIntent too https://stripe.com/docs/terminal/payments#capture-payment

Hardware/Sources/Hardware/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -0,0 +1,116 @@
...
+/// The adapter wrapping the Stripe Terminal SDK
+public final class StripeCardReaderService {
+    private let tokenProvider: ConnectionTokenProvider
+
+    private let discoveredReadersSubject = CurrentValueSubject<[CardReader], Never>([])
+    private let connectedReadersSubject = PassthroughSubject<[CardReader], Never>()
+    private let serviceStatusSubject = CurrentValueSubject<CardReaderServiceStatus, Never>(.ready)
+    private let discoveryStatusSubject = CurrentValueSubject<CardReaderServiceDiscoveryStatus, Never>(.idle)
+    private let paymentStatusSubject = CurrentValueSubject<PaymentStatus, Never>(.notReady)
+    private let readerSubject = PassthroughSubject<CardReaderEvent, Never>()
```
allendav: maybe readerEventsSubject ?

Hardware/Sources/Hardware/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -0,0 +1,116 @@
+import Combine
+import StripeTerminal
+
+/// The adapter wrapping the Stripe Terminal SDK
+public final class StripeCardReaderService {
+    private let tokenProvider: ConnectionTokenProvider
+
+    private let discoveredReadersSubject = CurrentValueSubject<[CardReader], Never>([])
+    private let connectedReadersSubject = PassthroughSubject<[CardReader], Never>()
```
allendav: It might be possible for a view to be subscribing to the connectedReaders after a reader has been connected by the service. I think this means the connectedReadersSubject should be a CurrentValueSubject so the view can obtain initial state.

Hardware/Sources/Hardware/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -0,0 +1,116 @@
...
+        discoveryStatusSubject.eraseToAnyPublisher()
+    }
+
+    /// The Publisher that emits the payment status
+    public var paymentStatus: AnyPublisher<PaymentStatus, Never> {
+        paymentStatusSubject.eraseToAnyPublisher()
+    }
+
+    /// The Publisher that emits reader events
+    public var readerEvent: AnyPublisher<CardReaderEvent, Never> {
```
allendav: Same comment as above

allendav (approved):
A few suggestions. After those are considered/handled, feel free to merge and we'll keep iterating. Thank you!

Hardware/Sources/Hardware/CardReaderService.swift
ctarda: Is this something we want to do from the service? This sounds like business logic to me, wouldn't it belong in the Store?

ctarda:
Thanks @allendav Merging, now that CI is ✅

Issue #3537, opened 2021-01-27
Variations: Attribute List Content


Ecarrion:
<img width="414" alt="Screen Shot 2021-01-27 at 2 26 40 PM" src="https://user-images.githubusercontent.com/562080/106044888-55809c00-60ae-11eb-9f96-2352805bc11e.png">

Tasks:
- [ ] Render product attributes
- [ ] Connect "Add Attribute" CTA to the Add attribute screen (to start the add attribute flow) and make sure it goes back here
- [ ] Make sure the "Done" button is only visible from the "first variation" flow.

PR #3627, opened 2021-02-10
Variations: First Variation -> Navigate to EditAttributeList after creating first attribute

### LINKED ISSUE #3536: Variations: Attribute List UI

### LINKED ISSUE #3618: Variations: Encapsulate new/edit attribute flow

Ecarrion:
closes #3536  

# Why

Today's turn will be to: **Navigate to "Edit Attribute List" after creating the first attribute**, this will allow merchants to add more attributes before generating the first variation.

Note: This PR, does not implement any functionality inside the `EditAttribute` screen, this is only the UI, real content will be handled in #3537 

# How

- Adds `EditAttributesViewController` with dummy content.

- Update `ProductVariationsViewController` to present `EditAttributesViewController` after finishing creating/updating the first attribute. A note here, is that the view controllers used to create the attribute(`AddAttributeVC` and `AddAttributeOptionVC`) are being removed before presenting the `EditAttributesVC`.

# Screenshots

Light | Dark
--- | ---
normal | dark-mode

Landscape | Big Font
--- | ---
landscape | big-font

# Demo

demo

# Testing Steps
- Tap on the "Next" button
- When the loading finishes, see that you are redirected to a dummy attribute list screen.
- When tapping back, you should be redirected back to the empty state screen(not the Add Attribute Option Screen)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
@@ -1,0 +1,10 @@
+import UIKit
+
+/// EditAttributesViewController: Displays the list of attributes for a product.
+///
+final class EditAttributesViewController: UIViewController {
+
+    @IBOutlet private weak var addButton: UIButton!
+    @IBOutlet private weak var addButtonSeparator: UIView!
+    @IBOutlet private var tableView: UITableView!
+}
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
@@ -53,10 +53,7 @@ private extension EditAttributesViewController {
     }
 
     func configureRightButton() {
-        let rightBarButton = UIBarButtonItem(title: Localization.done,
-                                             style: .done,
-                                             target: self,
-                                             action: #selector(doneButtonTapped))
+        let rightBarButton = UIBarButtonItem(barButtonSystemItem: .done, target: self, action: #selector(doneButtonTapped))
         navigationItem.setRightBarButton(rightBarButton, animated: false)
     }
 }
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+        configureRightButton()
+    }
+
+    func configureTitle() {
+        title = Localization.title
+    }
+
+    func configureRightButton() {
+        let rightBarButton = UIBarButtonItem(title: Localization.done,
+                                             style: .done,
```
ealeksandrov: Maybe switch to `barButtonSystemItem: .done` to reuse system styling and localizations?

ealeksandrov (changes requested):
Code and new UI looks great 👍

I see the only issue with navigation when you trigger next and then go back immediately:

https://user-images.githubusercontent.com/3132438/107501048-f93a7380-6ba7-11eb-975c-2659203ee531.mp4

What do you think about it? Should we cancel the network operation? Or forward navigation? Or maybe this is too far of edge case? App works well anyway, nothing really breaks 🙂

Ecarrion:
Hey @ealeksandrov, thanks for the review! I've updated the bar button item as per your suggestion.

Regarding the back button problem, I have [this card](https://github.com/woocommerce/woocommerce-ios/projects/29#card-54557890) to not forget about that edge case because I think it's better to deal with it outside of these PR's

Would you mind converting that card into an issue, explain the problem, and provide possible alternatives so the design team can weigh in?

Thanks!

ealeksandrov (approved):
Agree! LGTM 🚀

PR #3632, opened 2021-02-10
Fix "Back Button" behaviour on long press.

Ecarrion:
# Why

Yesterday, I found a piece of code that hides the back button text for the next screen, and that looks and works fine until I discovered that if you long-press the back button(iOS 14+) you will get a list of all the back buttons in order to go back into a specific screen.

For screens where we hide the back button text, it looks broken.

Screen Shot 2021-02-10 at 12 02 04 PM

This PR tries to fix that by updating the `removeNavigationBackBarButtonText()` method to use an empty image, instead of an empty text as the back button companion. Also, this PR makes sure the rest of the app uses that method instead of a custom empty back button.

# Demo
## Before
https://user-images.githubusercontent.com/562080/107544521-50384c80-6b98-11eb-9886-730eabffe8b7.MOV 

## After
https://user-images.githubusercontent.com/562080/107544440-3991f580-6b98-11eb-987d-e42c7da50688.mov

# Testing steps
- Go deep into a flow navigation wise
- Long press back button
- See that the menu list is populated with screen names.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,7 @@
 
 6.1
 -----
-
+- [*] Fixes a bug where long pressing the back button sometimes displayed an empty list of screens.
 
 6.0
 -----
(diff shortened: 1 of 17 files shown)


ctarda (approved):
Nice

I've been playing around for a while, navigating as deep as I could into the app and I couldn't find any spot where the menu is displayed on long press.

:shipit:

PR #3636, opened 2021-02-11
Fix cells separator on variations list

ealeksandrov:
## Description

1-line change. Variations list had 2 sets separators, one is custom view from the cell, another is system from UITableView.
System separator is longer in length, so you can spot a color change in "before" screenshot.
Updated separator style is same as in design assets and main products list.

## Testing

1. Select product with existing variations on products tab
2. Tap on "variations" cell
3. Check separators in variations list

## Screenshots

before | after
--|--

Update release notes:


pmusolino (approved):
LGTM :shipit:

PR #3639, opened 2021-02-11
Variations: Update "Any Attribute" option in AttributeOptionListSelectorCommand

### LINKED ISSUE #3637: Variations: Update "Any Attribute" option in AttributeOptionListSelectorCommand
Based on discussion in https://github.com/woocommerce/woocommerce-ios/pull/3628#issuecomment-776672709 let's update the "Any Attribute" option in `AttributeOptionListSelectorCommand` when editing the attributes in a product variation. 

Instead of "Any Attribute" it should be "Any [attribute name]" (e.g. "Any Color" when choosing options for the Color attribute).

rachelmcr:
Fixes: #3637 

## Description

When editing the attributes in a product variation, after selecting an attribute (e.g. Color) you're taken to a list of all the attribute options. Previously this included a generic "Any Attribute" option; now the generic option is "Any [attribute name]". For example, for the Color attribute the generic option is "Any Color."

## Changes

* In `AttributeOptionListSelectorCommand`, `anyOption` now takes the attribute name as a string to use in the localized `"Any %@"` string.

Before|After
-|-
Simulator Screen Shot - iPhone 12 - 2021-02-11 at 12 27 31|Simulator Screen Shot - iPhone 12 - 2021-02-11 at 12 25 26

## Testing

1. Make sure you have at least one variable product with variations on your store.
2. Go to the Products tab in the app.
3. Select a variable product.
4. Select Variations.
5. Select a product variation.
6. Select Attributes.
7. Select one of the attributes.
8. Confirm you see a list of all the available options for that attribute, including an "Any [attribute name]" (e.g. "Any Color") option at the top of the list.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
@@ -52,7 +52,7 @@ final class AttributeOptionListSelectorCommand: ListSelectorCommand {
 private extension AttributeOptionListSelectorCommand {
     enum Localization {
         static let anyOption = NSLocalizedString(
-            "Any %@",
+            "Any %1$@",
             comment: "Product variation attribute description where the attribute is set to any value.")
     }
 }
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
@@ -6,7 +6,7 @@ import Yosemite
 final class AttributeOptionListSelectorCommand: ListSelectorCommand {
 
     enum Row: Equatable {
-        case anyOption
+        case anyOption(String)
         case option(String)
     }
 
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
```diff
@@ -52,7 +52,7 @@ final class AttributeOptionListSelectorCommand: ListSelectorCommand {
 private extension AttributeOptionListSelectorCommand {
     enum Localization {
         static let anyOption = NSLocalizedString(
-            "Any Attribute",
+            "Any %@",
```
Ecarrion: Here we should use [positional placeholders](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/localization.md#always-use-positional-placeholders)!

Ecarrion (approved):
Looks good! Just a minor detail!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributeOptionListSelectorCommand.swift
rachelmcr: Thanks for catching that! I'll update that and another place where that string is used, to avoid having two different strings for translation.

PR #3640, opened 2021-02-11
Add check for addProductVariations feature flag

ealeksandrov:
## Description

This PR adds back accidentally removed check for `addProductVariations` feature flag.

In test build variations cell looks inactive but still opens to Create Variations flow which is unfinished yet. Check is used for viewmodel (cell appearance) setup, but not for actual tap action.

## Changes

Since creation flow starts only from empty state - prevent opening of `ProductVariationsViewController` if there are no existing variations (and no active `addProductVariations` flag). Therefore empty state can't be reached and creation flow can't be triggered. 

## Testing

0. Disable `addProductVariations` feature flag/run in release mode.
1. Open details for variable product with no variations yet created.
2. Tap on "Variations" cell.

## Screenshot

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
@@ -339,12 +339,9 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>: 
                 ServiceLocator.analytics.track(.productDetailViewGroupedProductsTapped)
                 editGroupedProducts()
                 break
-            case .variations:
+            case .variations(let row):
                 ServiceLocator.analytics.track(.productDetailViewVariationsTapped)
-                guard let product = product as? EditableProductModel else {
-                    return
-                }
-                guard product.product.variations.isNotEmpty || isAddProductVariationsEnabled else {
+                guard let product = product as? EditableProductModel, row.isActionable else {
                     return
                 }
                 let variationsViewController = ProductVariationsViewController(product: product.product,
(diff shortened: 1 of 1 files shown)


Ecarrion (commented):
Nice catch @ealeksandrov!

I'm thinking about the implications of the condition:
```swift
guard product.product.variations.isNotEmpty
```

As it will change the behavior that we previously had. I think a safer approach would be to constraint the visibility of the empty state screen on `ProductVariationsViewController` to listen for the `isAddProductVariationsEnabled` variable.

What do you think?

ealeksandrov:
Do you suggest adding more checks in `ProductVariationsViewController` or moving this check there?
I agree with adding additional check before presenting empty view, but if we remove suggested check here - we'll need to update UI, because cell doesn't look actionable now. Also it will navigate to empty view without anything.

Ecarrion:
@ealeksandrov taking a better look, I see that here is where we indicate that the cell is not actionable.
https://github.com/woocommerce/woocommerce-ios/blob/eff56e7402b38941c1e0714e858fb2e304238126/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/DefaultProductFormTableViewModel.swift#L394

So I think we should reformulate those conditions to something like
```swift
case .variations(let row):
    ServiceLocator.analytics.track(.productDetailViewVariationsTapped)
    guard let product = product as? EditableProductModel, row.isActionable else {
        return
    }
    let variationsViewController = ProductVariationsViewController(product: product.product,
                                                                   formType: viewModel.formType,
                                                                   isAddProductVariationsEnabled: isAddProductVariationsEnabled)
    show(variationsViewController, sender: self)
```

ealeksandrov:
@Ecarrion that's a great suggestion, it's same logic, but cleaner! Pushed update.

Ecarrion (approved):
🎉 🚀 
Also, don't forget to notify our release manager once this is merged!

AliSoftware:
This fix has just been shipped to beta-testers as part of new beta 6.0.0.1 🎉

PR #3642, opened 2021-02-11
[Card Present Payments] Expose CardReaderService through ServiceLocator

ctarda:
Closes #3590 

Expose CardReaderService introduced in #3590 via the Service Locator

## Changes
* Add a getter and setter to the Service Locator
* Add unit tests

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3643, opened 2021-02-11
[Card Present Payments] Add initial implementation of the CardPresentPaymentsStore.

### LINKED ISSUE #3591: [Mobile Payments] Yosemite: Add PaymentsStore
Once we have a way to fetch a mock implementation of the two pieces we need to put together (obtaining a token and attacking the Stripe SDK) it is time to start scaffolding the business logic layer.

Scope:
* Declare a Store (i.e. PaymentsStore) and an Action (i.e. PaymentsAction) in Yosemite.
* Integrate the new Store with StoresManager (behind a feature flag)
* Add just enough code to "initialize" the Reader Service (i.e. grabbing a reference to PassPhraseService and passing it to ReaderService)

ctarda:
Closes #3591 
For reference: p91TBi-4q4-p2

Now that the CardReaderService (#3625) is exposed via the Service Locator (#3642) it is time to add a barebones CardPresentPatmentsStore to Yosemite.

This PR just adds a tag Action and a tag Store, with one method, so that we can close the loop of the first iteration of our prototype, linking UI with the infrastructure to communicate with the external card readers.

The actual implementation of this action and store will be added, iteratively, during the next few weeks, starting with #3650 and #3651

## Changes
* Added an action (`CardPresentPaymentAction`) and a store (`CardPresentPaymentStore`) to Yosemite
* Added the new store to the list of stores available to logged in users, only when the nature flag is active.
* Updated WooCommerce's linked frameworks to remove Hardware (it was added in a previous PR).

## How to test
There isn't much to test for now, as this PR adds types that are not really used for now.
In any case, CI should be ✅ 

Update release notes:


WooCommerce/Classes/Yosemite/AuthenticatedState.swift
```diff
@@ -57,6 +57,16 @@ class AuthenticatedState: StoresManagerState {
...
+        if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentPayments) {
+            let cardReaderStore =  CardPresentPaymentStore(dispatcher: dispatcher,
+                                                           storageManager: storageManager,
+                                                           network: network,
+                                                           cardReaderService: ServiceLocator.cardReaderService)
+            allServices.append(cardReaderStore)
+        }
+
+        services = allServices
+
```
pmusolino: Why not enable the CardPresentPaymentStore always? Just the functionality is behind a feature flag, but I don't think that we should hide the store behind the flag (in theory it should not be called from anywhere else right)?

pmusolino (approved):
LGTM to me :) I only left a comment about why hiding the `CardPresentPaymentStore` behind the feature flag in `AuthenticatedState`.

WooCommerce/Classes/Yosemite/AuthenticatedState.swift
ctarda: That's a good point, Paolo. I'll update and merge.

ctarda:
You are on fire @pmusolino ! Thanks for the review. I'll update the way the store is instantiated, as per your suggestion, and then merge.

PR #3645, opened 2021-02-11
Variations: First Variation -> Render attribute list + create multiple attributes

### LINKED ISSUE #3537: Variations: Attribute List Content

Ecarrion:
closes #3537 

# Why
Today's turn will be to: **Render attribute list content + allow the merchants to keep adding more attributes before creating the first variation** 

Note: This PR, does not implement any functionality inside the EditAttribute screen, this is only the UI, real content will be handled in #3537

# How

- Add `EditAttributesViewModel` to store product info and generate attributes view models for the VC to render, as well as control the done button visibility, which only should be shown when creating the first variation.

- Update `EditAttributesViewController` to render content based on it's View Model.

- Update `EditAttributesViewController` to allow users to keep adding attributes by navigating to the `AddAttributeViewController` when tapping the `Add New Attribute` button.

- Update `AddAttributeViewController` to make sure it deletes the back button text when navigating away from it.

- Update `ProductAttributeStore` to add a workaround when fetching options and terms from a global attribute. This happens on the `AddAttributeViewController` and I'm still working on finding a definitive solution 😇 

# Demo
add-flow-demo

# Testing Steps

- Tap on the "Next" button
- When the loading finishes, see that you are redirected to the attribute list screen.
- Make sure that you can tap the "Add New Attribute" method and add more attributes correctly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
@@ -46,6 +46,7 @@ private extension EditAttributesViewController {
     func configureNavigationBar() {
         configureTitle()
         configureRightButton()
+        removeNavigationBackBarButtonText()
     }
 
     func configureTitle() {
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
@@ -80,7 +80,18 @@ extension EditAttributesViewController {
     }
 
     @objc private func addButtonTapped() {
-        // TODO: Launch add attribute flow and update product upon completion
+        navigateToAddAttributeViewController()
+    }
+
+    /// Navigates to `AddAttributeViewController` and upon completion, update the product and clean the navigation stack
+    ///
+    private func navigateToAddAttributeViewController() {
+        let addAttributeVM = AddAttributeViewModel(product: viewModel.product)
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
```diff
@@ -74,19 +102,13 @@ extension EditAttributesViewController {
 extension EditAttributesViewController: UITableViewDataSource, UITableViewDelegate {
 
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        return 6 // Temporary
+        return viewModel.attributes.count
```
ctarda, line 105: Just a thought. Would it make sense to expose the number of attributes as a property of the view model? Maybe even provide a getter, some short of attributeAt(indexPath), so that the view controller does not need to deal with the fact that attributes are packed in an array?

ctarda (approved):
It works perfect! 

I left a comment, mostly out of curiosity, but :shipit:

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
Ecarrion: I didn't do that because the data structure is very simple and doesn't seem prone to change, so I didn't see much benefit. I could add them if you feel strongly about it tho! :-)

Issue #3654, opened 2021-02-12
Add Time Duration Metric to Orders and Products Loading


shiki:
We'll use it to gather data for p91TBi-4p5-p2.

## Events

- [x] `orders_list_loaded`
    - [ ] ~`page_number` (Int)~ This isn't trivial in Android because Android uses offsets. 
    - [x] `status`: same as before. `nil` if loading all orders. 
    - [x] `total_duration`: (Double) The total duration (seconds) from requesting up to parsing and creating models to be used by the UI. 
- [ ] `orders_list_load_failed` 
    - [ ] TBD
- [ ] products TBD

shiki:
Closing this. The `orders_list_loaded` should be enough for now since we also track `jetpack_tunnel_timeout`. I don't think tracking products would be useful at this stage.

PR #3657, opened 2021-02-12
Add total_duration and page_number to orders_list_loaded Event

shiki:
Part of #3654. 

I updated the `orders_list_event` to include `page_number` and the `total_duration`. This will be used to gather more information for the slow orders reports. 

https://github.com/woocommerce/woocommerce-ios/blob/10dbce2fb4269ea9db8496938202268190bd5179/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift#L113-L118

I also changed the custom property specifications of `WooAnalyticsEvent` so that it will allow other support property types instead of just `String`. 

https://github.com/woocommerce/woocommerce-ios/blob/10dbce2fb4269ea9db8496938202268190bd5179/WooCommerce/Classes/Analytics/WooAnalyticsEventPropertyType.swift#L3-L14

## Testing

1. Navigate to Orders.
2. Confirm that you will see a Tracks event log like this:
    ```
    Tracked orders_list_loaded, properties: [AnyHashable("page_number"): 1, AnyHashable("total_duration"): 1.2141919136047363, AnyHashable("blog_id"): 0, AnyHashable("status"): "processing", AnyHashable("is_wpcom_store"): false]
    ```
3. If you can, please also confirm in Tracks that the custom properties are logged. 

## Author Checklist

### DIFF
--- a/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
@@ -39,7 +39,7 @@ import Foundation
 ///
 public struct WooAnalyticsEvent {
     let statName: WooAnalyticsStat
-    let properties: [String: String]
+    let properties: [String: WooAnalyticsEventPropertyType]
 }
 
 // MARK: - In-app Feedback and Survey
--- a/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Yosemite
 
 /// This struct represents an analytics event. It is a combination of `WooAnalyticsStat` and
 /// its properties.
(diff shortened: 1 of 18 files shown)


WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
```diff
@@ -108,6 +109,14 @@ extension WooAnalyticsEvent {
     static func shippingLabelsAPIRequest(result: ShippingLabelsAPIRequestResult) -> WooAnalyticsEvent {
         WooAnalyticsEvent(statName: .shippingLabelsAPIRequest, properties: ["action": result.rawValue])
     }
+
+    static func ordersListLoaded(totalDuration: TimeInterval, pageNumber: Int, status: OrderStatus?) -> WooAnalyticsEvent {
+        WooAnalyticsEvent(statName: .ordersListLoaded, properties: [
+            "status": status?.slug ?? String(),
```
Ecarrion, line 115: [Q]: Any reason on why using `String()` rather than `""` ?

Ecarrion (approved):
👍 

Good start point to address this!

WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
shiki: I copy-pasted this from the previous code in `OrderListViewController`. So, no good reason. 😅 Is there any difference between using `String()` versus `""`?

Ecarrion: Not according to Apple!
Screen Shot 2021-02-16 at 9 42 24 AM

shiki:
Thanks, Ernesto!

Issue #3619, opened 2021-02-08
Variations: Refresh Product after updating attributes or variations


Ecarrion:
We need to find a way to update the underlying `product` in  `ProductFormViewController` after updates to it happen.

### Tasks
- [x] Add new / Edit  product attribute #3690 
- [x] Remove Product Attribute
- [x] Add new variation #3690
- [x] Remove variation #3764
- [x] Update main product list page on create and delete variations. #3787

Ecarrion:
We also need to make sure that the main product list is updated!

PR #3658, opened 2021-02-14
Variations: Create First Variation

### LINKED ISSUE #3538: Variation: Generating First Variation
When tapping the "Done" CTA of the edit attribute list(#3036), we should generate our first variation.

Screen Shot 2021-01-27 at 2 47 35 PM

Tasks:
- [x] Build Loading Screen
- [x] Create variation(Using `ProuctVariationStore`) while showing the loading screen
- [x] After successful creation, redirect the user to the variation list screen - expose completion block on `EditAttributesViewController`.

Ecarrion:
closes #3538 

# Why 

Today's turn will be to: **Actually create the first variation** and navigate back to the variations list after it.

Note: After creating the first variation, the main product page is not updated. That will be handled In #3619 

# How

- Update `EditAttributesViewModel` to expose a `generateVariation()` method.

- Update `EditAttributesViewController` to:
  - Call the VM `generateVariation` method, when the "Done" button is pressed.
  - Show a loading screen while the variation is being generated
  - Show an error notice if the variation generation fails
  - Expose an `onVariationCreation` closure, to notify back after the variation is generated.

- Update `ProductVariationsViewController` to pop the previously pushed `EditAttributesViewController` after the variation is created.

# Demo

create-variation

# Testing Steps
- Tap on the "Next" button
- When the loading finishes, see that you are redirected to the attribute list screen.
- Tap on the "Done" button to generate the first variation.
- See a loading screen
- See that the user is redirected to the variations list screen and the new variation is visible

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
@@ -147,7 +147,6 @@ private extension EditAttributesViewController {
     enum Localization {
         static let addNewAttribute = NSLocalizedString("Add New Attribute", comment: "Action to add new attribute on the Product Attributes screen")
         static let title = NSLocalizedString("Edit Attributes", comment: "Navigation title for the Product Attributes screen")
-        static let done = NSLocalizedString("Done", comment: "Button title for the Done Action on the navigation bar")
 
         static let generatingVariation = NSLocalizedString("Generating Variation", comment: "Title for the progress screen while generating a variation")
         static let waitInstructions = NSLocalizedString("Please wait while we create the new variation",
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
```diff
@@ -123,5 +148,11 @@ private extension EditAttributesViewController {
         static let addNewAttribute = NSLocalizedString("Add New Attribute", comment: "Action to add new attribute on the Product Attributes screen")
         static let title = NSLocalizedString("Edit Attributes", comment: "Navigation title for the Product Attributes screen")
         static let done = NSLocalizedString("Done", comment: "Button title for the Done Action on the navigation bar")
```
ealeksandrov: nit: this can be removed, since system UIBarButtonItem is used for "done".

ealeksandrov (approved):
Great work 👏 Excited to finally see full variation creation!

Ecarrion:

PR #3664, opened 2021-02-15
Variations: Add Attribute screen updates

### LINKED ISSUE #3526: Variations: Add Attribute screen updates
Screen Shot 2021-01-26 at 12 06 28 PM

Tasks:
- [x] Update "Attribute Name" placeholder copy
> New Attribute Name

- [x] Update "To create a variation" placeholder copy
> To create a variation, you’ll need to set its attributes (ie “Color”, “Size”) first

- [ ] ~Update "Existing attribute" section title copy~
> ~OR TYPE TO SELECT EXISTING ATTRIBUTE~

- [x] Disable "Next" Button while the attribute text field is empty. Consider navigating to next screen when tapping `return`

rachelmcr:
Fixes: #3526 

## Description

Updates the Add Attribute screen placeholder text. Also updates the keyboard so it displays a "next" button (instead of "return") that navigates to the next screen when tapped.

## Changes

* Text field placeholder updated to `New Attribute Name`.
* Text field footer text updated to `To create a variation, you’ll need to set its attributes (i.e. “Color,” “Size”) first`.
* `TextFieldTableViewCell` now supports an optional `returnKeyType` to change the keyboard "return" button.
* `TextFieldTableViewCell.ViewModel` in `AddAttributeViewController` sets the `returnKeyType` and `onTextDidReturn` to support navigating to the next screen when the "next" button is tapped.

## Testing

1. Before you start, make sure you have a variable product on your store with no variations.
2. In the app, go to the Products tab.
3. Select your variable product.
4. Select "Add variations."
5. Select "Add Variation."
6. Confirm the placeholder text updates appear as expected on the Add Attribute screen.
7. Tap into the new attribute name field and confirm that tapping "next" on the keyboard does nothing (except dismiss the keyboard) if there is no text in the text field.
8. Enter a new attribute name, tap "next" on the keyboard, and confirm it proceeds to the next screen.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeViewController.swift
@@ -323,7 +323,7 @@ private extension AddAttributeViewController {
     enum Localization {
         static let titleView = NSLocalizedString("Add attribute", comment: "Add Product Attribute screen navigation title")
         static let nextNavBarButton = NSLocalizedString("Next", comment: "Next nav bar button title in Add Product Attribute screen")
-        static let titleCellPlaceholder = NSLocalizedString("Attribute name",
+        static let titleCellPlaceholder = NSLocalizedString("New Attribute Name",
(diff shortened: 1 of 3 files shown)


pmusolino (approved):
LGTM :shipit:

PR #3670, opened 2021-02-16
Variations: Fix a big when creating the first attribute and first variation

Ecarrion:
# Why

I spotted a bug that a new variation wasn't being rendered properly after being generated when it was created using a new attribute.

bug

This happened because the `ProductVariationsViewController` only updated its `product` property after creating a variation but did not update its `attributes` property.

# How

To fix it, I'm converting all the stored properties that derived from `product` to be computed properties, so they are always up to date.

# Demo
create-first-attribute

# Testing Steps
- Create a New Variable product
- Create a Variation by creating a new attribute
- After generating the variation see that it renders properly in the variation list screen.

Note: After creating the first variation, the main product page is not updated. That will be handled In #3619

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -251,6 +251,7 @@ private extension AddAttributeOptionsViewController {
                                                             if let text = text {
                                                                 self?.viewModel.addNewOption(name: text)
                                                             }
+
                                                          }, inputFormatter: nil,
                                                          keyboardType: .default)
         cell.configure(viewModel: viewModel)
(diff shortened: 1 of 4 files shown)


jaclync (approved):
LGTM! :shipit: 

Some other observations:

- In the screen where we can add options for an attribute, if I just tap enter without typing anything an empty option row is created. The empty options end up being removed in the backend, but maybe we can also skip creating an option row when the name is empty?



- If I create a local attribute, then add another global attribute with some selected options, the generated variation only has the local attribute. It's probably a TODO?

Ecarrion:
@jaclync first of all, thanks for the extra testing, highly appreciated! 🙇 

> In the screen where we can add options for an attribute, if I just tap enter without typing anything an empty option row is created. The empty options end up being removed in the backend, but maybe we can also skip creating an option row when the name is empty?

Sure, since it was quite simple I added the fix and a test in this same PR https://github.com/woocommerce/woocommerce-ios/pull/3670/commits/760d6764c7b785c623be954833e51965e355a9d0

> If I create a local attribute, then add another global attribute with some selected options, the generated variation only has the local attribute. It's probably a TODO?

Yes, this is a known issue that the product in other VCs is not getting updated, I'm working on it on https://github.com/woocommerce/woocommerce-ios/issues/3619

PR #3671, opened 2021-02-16
Variations: “Edit Attributes” Button

### LINKED ISSUE #3194: Variations: “Edit Attributes” Button

#3180 allows adding variations if the variable product does not have existing variations. 

If the product has existing variations, then we should show an Edit button in the variations list. The button allows the user to edit the **attributes and options** and generate new variations again. That flow starts in #3181. 

Screen Shot 2021-01-27 at 2 57 15 PM

## Goals

- [ ] Add `...` menu.
- [ ] Show the Edit Attributes button from the Options menu. 
- [ ] Tapping on Edit Attributes will forward the user to the attributes list screen

---

Android https://github.com/woocommerce/woocommerce-android/issues/3206

rachelmcr:
Closes: #3194 

## Description

If a variable product has existing variations, a `...` menu is shown with an "Edit Attributes" button (currently behind the `addProductVariations` feature flag). The button navigates to the Edit Attributes screen, allowing the user to edit the attributes and options and generate new variations again. 

## Changes

In `ProductVariationsViewController`:

* The `...` menu ("more options" button) is configured and added to the navigation bar on the condition that the product has variations and the `addProductVariations` feature flag is enabled.
* The "more options" action sheet is configured and presented with an "Edit Attributes" button and a "Cancel" button.

## Testing

**Setup:** Make sure you have at least two variable products on your store, one with variations and one without any variations.

Without variations:
1. Go to the Products tab.
2. Select the variable product with no variations.
3. Select "Add variations."
4. Confirm no `...` appears in the navigation bar.

With variations:
1. Go to the Products tab.
2. Select the variable product with variations.
3. Select "Variations."
4. Tap the `...` menu in the top right.
5. Confirm you can tap "Cancel" to close the action sheet.
6. Tap the `...` menu again.
7. Select "Edit Attributes" and confirm it opens the "Edit Attributes" screen.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -148,7 +148,7 @@ final class ProductVariationsViewController: UIViewController {
 //
 private extension ProductVariationsViewController {
 
-    /// Set the title.
+    /// Set the title and navigation buttons.
     ///
     func configureNavigationBar() {
         removeNavigationBackBarButtonText()
@@ -156,6 +156,20 @@ private extension ProductVariationsViewController {
             "Variations",
             comment: "Title that appears on top of the Product Variation List screen."
         )
+        if product.variations.isNotEmpty {
+            configureMoreOptionsButton()
+        }
+    }
+
+    /// Configure More Options button.
+    ///
+    func configureMoreOptionsButton() {
+        let moreButton = UIBarButtonItem(image: .moreImage,
+                                         style: .plain,
+                                         target: self,
+                                         action: #selector(presentMoreOptionsActionSheet(_:)))
+        moreButton.accessibilityLabel = Localization.moreButtonLabel
+        navigationItem.setRightBarButton(moreButton, animated: false)
     }
 
     /// Apply Woo styles.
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -443,6 +457,28 @@ private extension ProductVariationsViewController {
...
 
+// MARK: Action Sheet
+//
+private extension ProductVariationsViewController {
+    @objc private func presentMoreOptionsActionSheet(_ sender: UIBarButtonItem) {
+        let actionSheet = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
+        actionSheet.view.tintColor = .text
+
+        let editAttributesAction = UIAlertAction(title: Localization.editAttributesAction, style: .default) { [weak self] _ in
+            self?.navigateToEditAttributeViewController()
```
Ecarrion: Here, we call `navigateToEditAttributeViewController` which inside calls 
```swift
EditAttributesViewModel(product: product, allowVariationCreation: true)
```

For the edit flow, the `EditAttributesViewModel` should not allow the variation creation, as the way to create attributes is by tapping the `Add Variation` button once there is at least one variation.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -148,14 +148,28 @@ final class ProductVariationsViewController: UIViewController {
...
-    /// Set the title.
+    /// Set the title and navigation buttons.
     ///
     func configureNavigationBar() {
         removeNavigationBackBarButtonText()
         title = NSLocalizedString(
             "Variations",
             comment: "Title that appears on top of the Product Variation List screen."
         )
+        if product.variations.isNotEmpty && isAddProductVariationsEnabled {
```
Ecarrion, line 159: I wonder why we don't have a VM for this VC 🤔 
*Nothing to change here tho, just rambling*

Ecarrion (changes requested):
Requested changes, just because there is a needed change regarding the flow.
The code is 💯

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
rachelmcr: I'd be very interested in learning more about when we do or don't think a VM is needed!

rachelmcr:
@Ecarrion Thanks for catching that difference in the edit flow! Ready for another look.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
Ecarrion: That could be a great and wide discussion, but lately, my quick take is:

> Add code to a VM, If there is any logic that you think it's worth unit test.

In this particular example, there is no "easy" way to test that this code is working as expected. But in a VM this would be fairly simple. 

```swift
func test_more_button_appears_when_the_preoduct_is_not_empty_anf_feature_is_enabled() {
   let product = createNonEmptyProduct()
   let VM = ViewModel(product: Product, isFeatureEnabled: true)
   XCTAssertEquals(vm.showMoreButton, true)
}
```

And the same for all the other combinations!

Ecarrion (approved):
Great! Testing and code 🚀

Issue #2973, opened 2020-10-14
Shipping Labels M2: add option to validate shipping label address details


anitaa1990:
Once the [shipping label addresses are displayed](https://github.com/woocommerce/woocommerce-ios/issues/2972), we would need to:
- [x] parse the origin address of the store from the site settings.
- [x] for populating the origin address with the correct first and last name of the store owner, we need to explicitly ask for `first_name,last_name` parameters in `loadAccountSettings` method and add the new properties to `AccountSettings` entity.
- [x] call API to validate the origin address. If the address is not valid, the next steps in the creation flow should be greyed out.
- [x] If the origin address is valid, call API to validate the destination address. If the destination address is not valid, the next steps in the creation flow should be greyed out.
- [x] When the **Continue** button is clicked on the origin address section, we need to redirect to the edit address screen.  
- [x] If the validation fails, we need to display an error warning banner at the top of the screen.
- [x] Provide option to use the address as it is.
- [x] If the API response returns a suggested address, we need to display the suggested response and allow the users to select the suggested address.
- [x] Open Apple Maps tapping the error banner on top, if there is an address that can't be validated.
- [x] Add loader during address validation API request.
- [x] Show errors when the name, the company or other essential fields like the address are empty.
- [x] Top Banner autolayout issues

<img width="663" alt="Screenshot 2020-10-14 at 10 43 38 AM" src="https://user-images.githubusercontent.com/22608780/95946456-32392600-0e0a-11eb-8f4d-95a1c9322feb.png">
<img width="663" alt="Screenshot 2020-10-14 at 10 43 54 AM" src="https://user-images.githubusercontent.com/22608780/95946458-3402e980-0e0a-11eb-94db-ddf26cab0f93.png">
<img width="500" alt="Screenshot 2020-10-14 at 10 44 03 AM" src="https://user-images.githubusercontent.com/22608780/95946460-35341680-0e0a-11eb-86a1-afa7dc313287.png">

pmusolino:
Extra task: for populating the origin address with the correct first and last name of the store owner, we need to explicitly ask for `first_name,last_name` parameters in `loadAccountSettings` method and add the new properties to `AccountSettings` entity.

PR #3672, opened 2021-02-16
Shipping Labels: parsing of the store address from Site Settings

pmusolino:
Part of #2973

## Description
As part of Shipping Labels M2, during the shipping label creation, we need to show the store address as the origin address. In this PR, I included this new information coming from the Site Settings endpoint `/settings/general/`, which now are properly parsed under a new class called `SiteAddress`.

## Changes
- Removed the old class `SiteCountry` which was useful just for parsing the country, and substituted it with a new class `SiteAddress` which parses all the fields that we need, included the country.
- Implemented the unit tests for this new class.

## Testing
- Look at the code.
- Make sure CI pass.

Update release notes:


ctarda (approved):
💯 

Code looks great and tests pass locally and on CI, so :shipit:

PR #3617, opened 2021-02-08
Pull the Stripe Terminal SDK as a dependency of the Hardware Package

ctarda:
Closes #3595 
For context: p91TBi-4k7-p2 and p91TBi-4il-p2

This PR pulls the Stripe Terminal SDK specifically [the tag that supports SPM](https://github.com/stripe/stripe-terminal-ios/issues/61) as a dependency of Hardware. 

Hardware is a new (yet empty) Package introduced in #3615 

The dependency tree would be:

WooCommerce
    |
    | -> Hardware
            |
            | -> Stripe Terminal SDK

## Changes
* Added the Stripe Terminal SDK as a dependency to Hardware
* Import the Stripe Terminal into a struct declared in Hardware, just to prove that the build passes.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3676, opened 2021-02-16
Duplicate Info.plist, adding special keys for Stripe Terminal to Debug-Info.plist

### LINKED ISSUE #3593: [Mobile Payments] Update Info.plist. Enable location services, required background modes and bluetooth permissions

ctarda:
Closes #3593 

Second attempt at adding the keys required by the Stripe Terminal SDK to Info.plist.

See #3656  and the discussion in the comments. That discussion justifies the changes to this PR.

Also see p1613141952407700-slack-mobile-woo

## Changes
* Duplicated the Info.plist file. 
* Assigned Debug-Info.plist to Debug builds and Release-Info.plist to release builds
* Added the keys required by the Stripe Terminal SDK to Debug-Info.plist: https://stripe.com/docs/terminal/sdk/ios#configure

## Testing
* 👀 on the Info.plist file, in particular the one for release builds.
* Remove the app from the simulator.
* Checkout the branch, build and run on the simulator.
* Check that the app is installed, and it displayed the proper name.

cc: @allendav  and @AliSoftware 

Update release notes:


allendav:
diffing the two files yields the [expected deltas](https://github.com/woocommerce/woocommerce-ios/pull/3656#issuecomment-780115333)

```
diff WooCommerce/Resources/Debug-Info.plist WooCommerce/Resources/Release-Info.plist
64,69d63
< 	<key>NSLocationWhenInUseUsageDescription</key>
< 	<string>Location access is required in order to accept payments.</string>
< 	<key>NSBluetoothPeripheralUsageDescription</key>
< 	<string>Bluetooth access is required in order to connect to supported card readers.</string>
< 	<key>NSBluetoothAlwaysUsageDescription</key>
< 	<string>This app uses Bluetooth to connect to supported card readers.</string>
78d71
< 		<string>bluetooth-central</string>
```

allendav (commented):
For the WooCommerce Run Scheme, Debug build configuration builds and launches successfully.
For the WooCommerce Profile Scheme, Release build configuration builds and launches successfully.
As noted [above](https://github.com/woocommerce/woocommerce-ios/pull/3676#issuecomment-780141716), the expected diffs are in the Debug plist.
Do we need to do anything for the other Schemes and Build Configurations, however?

AliSoftware:
As said in Slack, I'm a bit puzzled about all this:

 - The real goal would be to have the `Info.plist` keys _when Stripe SDK is included_, and not have the keys when we don't link against the Stripe SDK. Not sure how that relates to "Debug vs Release" – as opposed to "SDK included or not"? or did I miss something?
 - If we end up integrating the `Hardware` submodule/subproject, and thus the StripeSDK as a dependency of it, in all configurations and unconditionally, as solutions/attempts like https://github.com/woocommerce/woocommerce-ios/pull/3617 or https://github.com/woocommerce/woocommerce-ios/pull/3682 seems to suggest (i.e. the SDK would be linked in all cases regardless of Debug/Release), shouldn't we apply the same about the keys anyway

In other words, **following the syllogism** that "if SDK is there, then keys in Info.plist should be added; SDK is always linked (at least in the current attempts/solutions to integrate it in the project). **Therefore, keys should always be added to `Info.plist`**"…

---

And, alternatively, if we _do_ want to only conditionally integrate the SDK (and therefore the Info.plist keys), so that both options (with StripeSDK+Keys or without) live in parallel in `develop` – and we only enable the SDK+Keys+Flag all at once when working on the feature, all together in sync –, then we should start the brainstorming on how to do that conditional integration of the SDK first (will it be a separate configuration in addition to Debug/Release? Some kind of separate scheme? A trick in the Build Settings to add `--framework StripeSDK` to the linker flags conditionally based on a flag? Other? We need to discuss and brainstorm ideas and options about it in a P2 if we want to go that route imho) as that's probably the most delicate to do, and only then apply the same logic to the Info.plist case (or not, because even if the keys are required _if_ the SDK is _linked_, the opposite is probably not true and if the SDK is not linked, the keys can still remain in Info.plist anyway… I think?)

---

In any case, given the debate and attempts already showed that this might be more complicated than we thought, a P2 might be welcome anyway imho.

PR #3678, opened 2021-02-17
Order Details: Fix product image not loading

### LINKED ISSUE #3550: Photos of products on OrderDetails don't load correctly
**Describe the bug**
Photos of products on a hold/canceled/failed order are not being loaded correctly.

**To Reproduce**
1. Go to a failed/on hold/pending order, 
2. Tap “Details” below the product list inside the order,
3. See that thumbnails are not loading.

**Screenshots**
IMG_243653C31286-1

**Expected behavior**
- All photos should load.

**Mobile Environment**
Please include:
- Device: iPhone 12 Pro
- iOS version: 14.2
- WooCommerce iOS version: 4.9

momo-ozawa:
Resolves #3550 

### Description
- Fixed a bug in Order Details where the thumbnail image of a product wasn't being loaded correctly.

### Error
- We were getting a `Kingfisher.KingfisherError.ImageSettingErrorReason.emptySource` for this method:
https://github.com/woocommerce/woocommerce-ios/blob/3632d0a0dd6f594e5bb3cdf78701c97dd69db106/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Product%20List%20Section/Product%20Details/PickListTableViewCell.swift#L91-L95

### Actual source of error
- The `products` property wasn't being set in `ProductListViewController`. As a result, the `imageURL` we pass in `downloadAndCacheImageForImageView` was nil, which is why we were getting the error above.

### How to test
1. Go to the Orders tab and tap on a order with an order status of failed / on hold / pending 
2. Tap “Details” below the product list inside the order
3. ✅ Confirm that the thumbnail images are loaded correctly

### Screenshot
Before | After
--- | ---
Simulator Screen Shot - iPhone 12 Pro - 2021-02-17 at 14 35 28 | Simulator Screen Shot - iPhone 12 Pro - 2021-02-17 at 14 31 58

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/PickListTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/PickListTableViewCell.swift
@@ -92,7 +92,12 @@ extension PickListTableViewCell {
                                                        with: item.imageURL?.absoluteString,
                                                        placeholder: UIImage.productPlaceholderImage.imageWithTintColor(UIColor.listIcon),
                                                        progressBlock: nil,
-                                                       completion: nil)
+                                                       completion: { [weak productImageView] (image, error) in
+                                                           guard image != nil else {
+                                                               return
+                                                           }
+                                                           productImageView?.contentMode = .scaleAspectFill
+                                                       })
         name = item.name
         quantity = item.quantity
 
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/PickListTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/PickListTableViewCell.swift
@@ -92,12 +92,7 @@ extension PickListTableViewCell {
                                                        with: item.imageURL?.absoluteString,
                                                        placeholder: UIImage.productPlaceholderImage.imageWithTintColor(UIColor.listIcon),
                                                        progressBlock: nil,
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/PickListTableViewCell.swift
```diff
@@ -92,7 +92,12 @@ extension PickListTableViewCell {
                                                        with: item.imageURL?.absoluteString,
                                                        placeholder: UIImage.productPlaceholderImage.imageWithTintColor(UIColor.listIcon),
                                                        progressBlock: nil,
-                                                       completion: nil)
+                                                       completion: { [weak productImageView] (image, error) in
+                                                           guard image != nil else {
+                                                               return
+                                                           }
+                                                           productImageView?.contentMode = .scaleAspectFill
+                                                       })
```
ealeksandrov: Do you think we need this together with similar call in `setupImageView()` below?

ealeksandrov (approved):
Thanks for a fix! 👏
Just added a minor question about setting content mode in 2 places.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/PickListTableViewCell.swift
momo-ozawa: Ah! Good catch -- I totally missed that!
Fixed in 066b238 😄

PR #3684, opened 2021-02-17
Shipping Labels: populated fields in the address form screen

pmusolino:
Part of #2973 

## Description 
As part of Shipping Labels M2, during the shipping label creation, we need to display the store address as the origin address, and we also need to display this info in the address form screen. In this PR, now the address will be shown in the Address Form Screen, and also the origin address will be generated (the name is still a WIP. For populating the origin address with the correct first and last name of the store owner, we need to explicitly ask for `first_name,last_name` parameters in `loadAccountSettings` method and add the new properties to `AccountSettings` entity. This will be part of a separate PR).

## Changes
- Display the address info in `ShippingLabelAddressFormViewController` cells.
- Generated the origin address in `ShippingLabelFormViewModel`.

## Testing
2. Open an order detail.
4. Press the button "Continue" in the Create Shipping Label form.
5. The new screen for validating an address should be presented with all the information. Since at this moment just the "Ship From" option is enabled, you will be able to see the address and the info of your store.

## Screenshot

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -13,13 +13,13 @@ final class ShippingLabelFormViewModel {
 
     init(originAddress: Address?, destinationAddress: Address?) {
         self.originAddress = ShippingLabelFormViewModel.fromAddressToShippingLabelAddress(address: originAddress) ??
-            ShippingLabelFormViewModel.generateDefaultOriginAddress()
+            ShippingLabelFormViewModel.getDefaultOriginAddress()
         self.destinationAddress = ShippingLabelFormViewModel.fromAddressToShippingLabelAddress(address: destinationAddress)
     }
 
     var sections: [Section] {
-        let shipFrom = Row(type: .shipFrom, dataState: .validated, displayMode: .editable)
-        let shipTo = Row(type: .shipTo, dataState: .pending, displayMode: .editable)
+        let shipFrom = Row(type: .shipFrom, dataState: .pending, displayMode: .editable)
+        let shipTo = Row(type: .shipTo, dataState: .pending, displayMode: .disabled)
         let packageDetails = Row(type: .packageDetails, dataState: .pending, displayMode: .disabled)
         let shippingCarrierAndRates = Row(type: .shippingCarrierAndRates, dataState: .pending, displayMode: .disabled)
         let paymentMethod = Row(type: .paymentMethod, dataState: .pending, displayMode: .disabled)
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -29,6 +30,25 @@ final class ShippingLabelFormViewModel {
...
+        // and add the new properties to AccountSettings entity.
+        let address = Address(firstName: ServiceLocator.stores.sessionManager.defaultAccount?.displayName ?? "",
+                lastName: "", company: ServiceLocator.stores.sessionManager.defaultSite?.name ?? "",
+                address1: SiteAddress().address,
+                address2: SiteAddress().address2,
+                city: SiteAddress().city,
+                state: SiteAddress().state,
+                postcode: SiteAddress().postalCode,
+                country: SiteAddress().countryName ?? "",
+                phone: "",
```
rachelmcr, line 47: Is the phone something we can also retrieve from the back end, or does this always have to be entered manually?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -29,6 +30,25 @@ final class ShippingLabelFormViewModel {
...
 extension ShippingLabelFormViewModel {
+    // We generate the default origin address using the information
+    // of the logged Account and of the website.
+    private static func getDefaultOriginAddress() -> ShippingLabelAddress? {
+        // TODO: 2973 - for populating the origin address with the correct first and last name of the store owner
+        // we need to explicitly ask for first_name,last_name parameters in loadAccountSettings method
+        // and add the new properties to AccountSettings entity.
+        let address = Address(firstName: ServiceLocator.stores.sessionManager.defaultAccount?.displayName ?? "",
+                lastName: "", company: ServiceLocator.stores.sessionManager.defaultSite?.name ?? "",
+                address1: SiteAddress().address,
```
rachelmcr, line 41: This looks fine, but I noticed in testing that the address isn't always generated correctly. It looks like it's an issue with how we're actually setting `SiteAddress().address` here:

https://github.com/woocommerce/woocommerce-ios/blob/59ab55ba14463c84bc4f6b9cef0eea0b0e4963ab/WooCommerce/Classes/Tools/SiteAddress.swift#L53-L57

We're finding each value with `setting.settingID.contains(settingID)`, but looking for the key `woocommerce_store_address` can return the value for `woocommerce_store_address_2` instead, if it comes first. So we can end up with the wrong value set for `address1` here.

rachelmcr (approved):
These changes look good! I have a few questions but they're mostly unrelated to the changes in this PR.

I think the phone number field should probably be `Phone` instead of `Phones`:

pmusolino:
> These changes look good! I have a few questions but they're mostly unrelated to the changes in this PR.
> I think the phone number field should probably be Phone instead of Phones

Good catch. I think that for some reason the original design has a typo because on the web it's called just `Phone`. I updated it here ed86945a5352784a8ff35ff296aee7f9f59d54f9 🙇

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: For the moment the source address will not contain the phone number, because it's not information we have on the API side.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: Good catch Rachel! 💯 I substituted `containts` with `isEqual` 👍 0a7b4cce1dc2726e9261792d261af1e3f3e61b51

PR #3686, opened 2021-02-17
Add view model for ProductVariationsViewController

rachelmcr:
Related discussion: https://github.com/woocommerce/woocommerce-ios/pull/3671#discussion_r576917870

## Description

This PR introduces the view model `ProductVariationsViewModel` for `ProductVariationsViewController`, and accompanying unit tests. For now the view model is very small, just introducing the `showMoreButton` variable to control whether the `...` more button appears (when the product has variations and the `addProductVariations` feature flag is enabled).

## Changes

* New view model `ProductVariationsViewModel` with `showMoreButton`.
* `ProductVariationsViewController` is updated to use the new view model.
* New unit tests to check that the more button is only displayed when expected.

(This is my first time adding a new view model — please share any thoughts, big or nitpicky, about how it's done!)

## Testing

* Check the code and confirm all unit tests pass.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/WooCommerceTests/ViewRelated/Products/Variations/ProductVariationsViewModelTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Products/Variations/ProductVariationsViewModelTests.swift
@@ -1,0 +1,42 @@
+import XCTest
+@testable import WooCommerce
+
+final class ProductVariationsViewModelTests: XCTestCase {
+    func test_more_button_appears_when_product_is_not_empty_and_addProductVariations_feature_is_enabled() {
+        // Arrange
+        let variations: [Int64] = [101, 102]
+        let product = MockProduct().product(variations: variations)
+        let viewModel = ProductVariationsViewModel(product: product, isAddProductVariationsEnabled: true)
+
+        // Assert
+        XCTAssertEqual(viewModel.showMoreButton, true)
+    }
+
+    func test_more_button_does_not_appear_when_product_is_not_empty_and_addProductVariations_feature_is_disabled() {
+        // Arrange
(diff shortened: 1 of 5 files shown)


WooCommerce/WooCommerceTests/ViewRelated/Products/Variations/ProductVariationsViewModelTests.swift
```diff
@@ -0,0 +1,42 @@
+import XCTest
+@testable import WooCommerce
+
+final class ProductVariationsViewModelTests: XCTestCase {
+    func test_more_button_appears_when_product_is_not_empty_and_addProductVariations_feature_is_enabled() {
+        // Arrange
+        let variations: [Int64] = [101, 102]
+        let product = MockProduct().product(variations: variations)
```
Ecarrion: nit: I would say that the `MockProduct()` is deprecated.
We should try to use `Product().copy(variations: [variations])` instead! :-)

Ecarrion (approved):
Nothing to complain about the code! This looks great!
Just left a minor nitpick for one of our old legacy classes.

WooCommerce/WooCommerceTests/ViewRelated/Products/Variations/ProductVariationsViewModelTests.swift
rachelmcr: Thanks! I updated the tests to use `Product()` instead in https://github.com/woocommerce/woocommerce-ios/pull/3686/commits/679b6c80be4bda0f620b1cdda9bc3e7d4fe0970c.

PR #3682, opened 2021-02-17
Move Hardware from a Swift Package to a subproject 

ctarda:
Closes #3681 
For context, see: paaHJt-1gq#comment-3828-p2

The original setup (where `>` indicates a dependency) was:
WooCommerce > Hardware > Stripe Terminal SDK

It turns out there seems to be a limitation in the Swift Package Manager that prevents embedded frameworks from being codesigned, which in turn was making the app impossible to run on devices.

This PR migrates Hardware from a Swift Package to a subproject (like Yosemite or Networking) that builds a framework.

Also, the Stripe Terminal SDK is pulled in via Cocoapods

## Changes
* Removed the Hardware Swift Package
* Add a new subproject, called Hardware
* Move the code in the Hardware Swift Package to the new Hardware Subproject
* Add the Stripe Terminal SDK to the Podfile.
* Add a new section to the Podfile to setup the Hardware subproject

## Testing
* Checkout the branch, run `bundle exec pod install`
* Build and run. Make sure the project runs as expected.
* Add a breakpoint to line  of `HardwareTests.swift`. Run all the WooCommerce tests (i.e. press command+u) and make sure that the breakpoint in HardwareTests is hit. 
* Build and run on a device. The app should run on a device properly

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3687, opened 2021-02-17
[Mobile Payments] Remove Hardware and dependency with the StripeTerminal SDK

### LINKED ISSUE #3681: [Mobile Payments] Remove Hardware and Stripe SDK
We have hit some limitations and difficult solutions that we need to sort out. Be need to find the optimal way to:
* Pull the Stripe Terminal SDK in a way that it is only used in Debug builds
* Add the keys required by the Stripe Terminal SDK to the Info.plist that are needed for development, in a way that only affect debug builds.

For all that, we have decided to the integration of the Stripe Terminal SDK to a feature branch for now. 

Scope: 
* removing all traces of the integration with the Stripe Terminal SDK from the project.

ctarda:
Closes #3681 

For more context: p1613553100463900-slack-mobile-woo

We need to rethink the way we implement the integration with the Stripe Terminal SDK in a way that it is safe for production builds. We do not want to risk prompting users for Bluetooth or location permissions at this point.

After several discussions and several attempts at finding the optimal way to do so, we have decided to remove the integration from develop, and continue development in a feature branch. (see, for example, #3656 and #3682)

## Changes
* Remove Hardware package and its contents.
* Remove dependency with Stripe Terminal SDK.
* Remove CardPresentPaymentStore and CardPresentPaymentAction.
* Remove instantiation of CardPresentPaymentsStore from AuthenticatedState.
* Remove HardwareTests from WooCommerce’s testing plan.

## How to test
* Check out the branch, run `bundle exec pod install`
* Build the project. Make sure the project builds without issues on the simulator
* Run the tests. Make sure all the tests pass.
* Build the project on a device. There shouldn't be any errors

Update release notes:


AliSoftware:
After checking out the branch of this PR and doing `rake dependencies` (which includes `bundle exec pod install`), it seems I still had a `Hardware/` folder at the root of the repository, even if it now only contained `.swiftpm/` subfolder with `xcode/` subfolder with some xcuserdata.

Since it seemed it was only local and a `git clean -dxn` suggested that it wasn't tracked by git anymore anyway (as expected), I `rm -rf Hardware` from my local working copy, and closed and reopened the Xcode workspace. But the "Hardware" blue folder still remained in Xcode's Project Inspector (the left pane) – with that entry now in red as the folder wasn't there anymore.
I had to select the red "Hardware" entry in the Project navigator and hit the delete key to remove it to get rid of it… even though doing this didn't make any change in any files tracked by git (i.e. `git diff` was still empty after doing this).

So TL;DR: people might need to clean some local, untracked leftovers after this PR lands…

---

Other than this, tested the PR by building on simulator and also running the tests and all worked, but wasn't able to test building on actual device, because my iPhone is on iOS 14.4 and I still use Xcode 12.2 which doesn't have the symbols for it, and haven't had time to download and update my setup (it's been a while since I launched Xcode and/or build the project, not being a Code Wrangler working on the project itself), so I'll let someone else do the test on device.

allendav:
<del>Can we remove the `Hardware` folder from the repo too for complete-ness?</del>

Nevermind. Looks like it was a local glitch. The folder is now not present.

allendav (approved):
I was able to install both WooCommerce Debug and Release on my iPad. I also had the phantom Hardware folder initially persist (and git rm -rf report that it failed to remove it) but it eventually apparently nonetheless removed properly.

I did note an empty, red Vendors group in the Project navigator

ctarda:
I'll merge now so that this gets into the next release. I kind of recall having similar "phantom" files/folders in Tumblr a while back, and the solution seemed to be running `git clean -somethingsomething`. I do not recall the exact format of the flags, but it was -fx as fas as I recall, I'll look that up.

AliSoftware:
> I'll merge now so that this gets into the next release. I kind of recall having similar "phantom" files/folders in Tumblr a while back, and the solution seemed to be running `git clean -somethingsomething`. I do not recall the exact format of the flags, but it was -fx as fas as I recall, I'll look that up.

I gave that command in my earlier comment above. I use `git clean -dxn` to dry-run the operation (the `-n` flag) and see what it _would_ have cleaned (just in case it shows some local files I forgot to save/commit and that it would revert/remove while I wouldn't want), then if all looks good, `git clean -dxf` (`-f` = force, the opposite of `-n` dry run)

PR #3690, opened 2021-02-17
Variations: Update product details screen after generating a variation

### LINKED ISSUE #3619: Variations: Refresh Product after updating attributes or variations

Ecarrion:
closes #3619 

# Why

[Creating a variation is already implemented](#3658), but screens like the product detail screen weren't notified of such creation, hence not updating its UI.

This PR makes sure that the `ProductFormViewController` is updated with the new variation after generating a new variation(and attributes)

# How

- Update `EditAttributesViewController` to provide an updated product(adding the new variation id) after generating a variation. Before, only the new variation was provided.

- Modify `AddAttributeViewController`, `EditAttributesViewController`, and `VariationsViewController` to provide closures for when the  product gets updated(with thew new variation and attributes) .

- Update `ProductFormViewController` to listen to the `VariationsViewController. onProductUpdate` closure

- Update `ProductFormViewModel` to correctly update the underlying product with the new variation, without losing any pending change that the merchant hasn't committed yet.

# Demo

https://user-images.githubusercontent.com/562080/108268770-04a71500-713b-11eb-87af-313adf460e56.mov

# Testing Steps

- Create a New Variable product
- Create a Variation by creating one or more attributes
- After generating the variation see that it renders properly in the variation list screen.
- When tapping back, see that the product details screen has information about the variation.

Update release notes:


WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/ProductFormViewModelTests.swift
```diff
@@ -153,6 +153,23 @@ final class ProductFormViewModelTests: XCTestCase {
         // Assert
         XCTAssertFalse(canDeleteProduct)
     }
+
+    func test_update_variations_updates_original_product_while_mantaining_pending_changes() throws {
```
jaclync: super nit:

```suggestion
    func test_update_variations_updates_original_product_while_maintaining_pending_changes() throws {
```

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/EditAttributesViewModelTests.swift
```diff
@@ -90,6 +90,35 @@ final class EditAttributesViewModelTests: XCTestCase {
...
+
+        let viewModel = EditAttributesViewModel(product: product, allowVariationCreation: true, stores: mockStores)
+
+        // When
+        let result: Result<Product, Error> = waitFor { promise in
+            viewModel.generateVariation { result in
+                promise(result)
+            }
+        }
+
```
jaclync, line 117: nit:
```suggestion
        // Then
```

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
```diff
@@ -15,7 +15,11 @@ final class EditAttributesViewController: UIViewController {
 
     /// Assign this closure to be notified after a variation is created.
     ///
-    var onVariationCreation: ((ProductVariation) -> Void)?
+    var onVariationCreation: ((Product) -> Void)?
```
jaclync, line 18: nit: maybe we can also change the variable name passed to this closure? right now it's still called `variation` 🙂

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
```diff
@@ -15,7 +15,11 @@ final class EditAttributesViewController: UIViewController {
...
+
+    /// Assign this closure to be notified after an attribute  is created.
+    ///
+    var onAttributeCreation: ((Product) -> Void)?
```
jaclync, line 22: non-blocking thought: it might be nice having a Combine publisher (observable) for `Product` and passing it along the way for subscription on changes 😄 or keeping the closure, but for any product changes like `var onProductUpdate: ((Product) -> Void)?`

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -251,6 +251,25 @@ extension ProductFormViewModel {
...
+        // If the product doesn't have any pending changes, we can safely override the original product
+        guard hasUnsavedChanges() else {
+            return resetProduct(newOriginalProduct)
+        }
+
+        // If the product has pending changes, we need to override the `originalProduct` first and the `living product` later with a saved copy.
+        // This is because, overriding `originalProduct` also overrides the `living product`.
+        let productWithChanges = EditableProductModel(product: product.product.copy(attributes: newProduct.attributes,
+                                                                                    variations: newProduct.variations))
+        resetProduct(newOriginalProduct)
```
jaclync, line 271: nit: just wondering, if there are unsaved changes on the product, does it make sense to make the original product based on `originalProduct` so that we call `resetProduct(EditableProductModel(product: originalProduct.product.copy(attributes:...variations...)))` here?

It might not make a difference UI-wise, since `originalProduct` is mostly used for showing the "Update" CTA and there are already outstanding changes from the new attributes and variations. Just thought it might be more consistent keeping the `originalProduct` 😄

jaclync (approved):
Working great! The variation creation flow is really coming together 😃 Just some non-blocking nits for your consideration!

jaclync (approved):
Working great! The variation creation flow is really coming together 😃 Just some non-blocking nits for your consideration!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
Ecarrion: I prefer to leave it as it is because the closure is only invoked when a variation is created. I changed the associated variable because we don't really need the variation(as it's only an `ID` inside `Product`).

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
Ecarrion: I kind of need them to be different, because the consumer `ProductVariationsViewController.navigateToEditAttributeViewController()` perform different actions when an attribute or variations are created.

Ecarrion: I could have used combine 🤔 but for these simpler callbacks, I didn't see much benefit 😅

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
Ecarrion: Nice, I think it does not make a difference in practice but it does read conceptually better!
Thanks for pointing this out!

Ecarrion:
Thanks for the review @jaclync :-)

Issue #3675, opened 2021-02-16
[Mobile Payments] Add unit test suit for integration with Stripe SDK to the app target


ctarda:
We attempted to uni test the boundary of the integration with the Stripe Terminal SDK (that is `StripeCardReaderService`) by adding a test class to the Hardware package that would target that class.

However, the Terminal SDK throws an assertion if there is no Info.plist with a certain set of keys. Ww tried to provide that Info.plist as part of the Hardware package's tests bt failed to workaround the issue.

We asked the fine folks building the Stripe Terminal SDK and they confirmed that, yes, an Info.plist s necessary:

p1613069557084200-slack-woo-stripe-external

So the solution is:
* once #3656 is merged, add the unit tests covering StripeCardReaderService to the main app's unit testing target

Scope:
* add one test to cover the discovery process in StripeCardReaderService, running against the SDK's test mode
* If CardPresentPaymentStore has been added to the project at this point, add a test to cover the discovery process in that store.

PR #3691, opened 2021-02-17
[Mobile Payments] Model objects for CardReader and PaymentIntent

### LINKED ISSUE #3649: [Mobile Payments] Model objects
In #3625 we just declared tag types to be able to pass the build. 

Now it is time to look into the actual data we get from the Stripe Terminal SDK, what we need to provide to the SDK and to start populating the interface of the model objects declared in Hardware

Scope: 
* Enough code to implement CardReader, PaymentIntent, and Charge. These objects model some of their counterparts in the [Stripe Terminal SDK](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPReader.html)
* Initializers to instantiate our domain objects with their counterparts in StripeTerminal.

Out of scope:
* Unit tests. We can not test our model objects directly, because for that we would need to create mock objects on the StripeTerminal side of the boundary. However, the initializers of the model objects in StripeTerminal are marked as [NS_UNAVAILABLE](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPReader.html#/c:objc(cs)SCPReader(im)init). We can not even instantiate subclasses.

ctarda:
Closes #3649 

⚠️  This PR is not against develop, but against the future branch for Mobile Payments ⚠️ 

We continue building our prototype. As a follow up to #3643, we need some model objects. 
For reference: p91TBi-4q4-p2

Scope:
* Enough code to implement CardReader, PaymentIntent, and Charge. These objects model some of their counterparts in the Stripe Terminal SDK. For more info and APIs: https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPReader.html
Initializers to instantiate our domain objects with their counterparts in StripeTerminal.

Out of scope:
* Unit tests. We can not test our model objects directly, because for that we would need to create mock objects on the StripeTerminal side of the boundary. However, the initializers of the model objects in StripeTerminal are marked as NS_UNAVAILABLE. We can not even instantiate subclasses.
* Unit tests will come in a later PR, in the form of integration tests. We will use the test mode of the SDK to simulate calls to the methods that return these model objects which we will then proceed to unit test.

## Changes
* Added model objects for CardReader, PaymentIntent, Charge and enums to back up the different statuses (i.e. PaymentIntentStatus, ChargeStatus...)

## How to test
* Look at the code. 
* This code is not executed for now, so this PR does not add any new behaviour
* CI should be ✅ 

Update release notes:


Hardware/Hardware/CardReader/CardReader.swift
```diff
@@ -1,7 +1,18 @@
...
+    public let softwareVersion: String?
+    public let batteryLevel: Float?
+    public let readerType: CardReaderType
+}
+
+
+extension CardReader: Equatable {
+    public static func ==(lhs: CardReader, rhs: CardReader) -> Bool {
+        lhs.serial == rhs.serial
+    }
```
Ecarrion: Hmm... I think implementing equatable on value-types by only checking one property could lead to some errors in the future.  

What do you think about just adding a method, `hasSameSerial()` or similar.

Hardware/Hardware/CardReader/Charge.swift
```diff
@@ -0,0 +1,27 @@
...
+
+    /// Metadata associated with the charge.
+    let metadata: [AnyHashable: Any]?
+}
+
+
+extension Charge: Equatable {
+    public static func ==(lhs: Charge, rhs: Charge) -> Bool {
+        return lhs.identifier == rhs.identifier
+    }
```
Ecarrion: Same here, for value-types, I think this should be something like [`Identifiable`](https://developer.apple.com/documentation/swift/identifiable) and not `Equatable`. WDYT?

Ecarrion (commented):
Looks good in general, but left you some concerns about the `equality` implementation

Hardware/Hardware/CardReader/CardReader.swift
ctarda: > I think implementing equatable on value-types by only checking one property could lead to some errors in the future.

I'm curious, why? What difference would it make to add that layer of indirection?

Hardware/Hardware/CardReader/Charge.swift
ctarda: Oh, like use Identifiable as a way to decouple identity from state? Yep, that makes sense, I'll update, thanks!

Hardware/Hardware/CardReader/CardReader.swift
Ecarrion: To me, the expectation when I do a `==` over two value-types is that that object will behave the same as the other because they are "equals". Here, by comparing only one property, we can't have that expectation about it. Operations would likely have different outcomes because the object can have different values.

Additionally, I can see ourselves getting some false positives on unit tests because someone used `XCTAssertEqual` expecting the objects to be exactly the same.

ctarda:
Thanks for the review @Ecarrion I just pushed an update that I hope will address your concerns (and hopefully fixes the tests 😊 )

This is ready for another round.

Ecarrion (approved):
🚀

ctarda:
I had to push one empty test to HardwareTest.swift in order to pass the build on CI. Proper tests will come soon in #3675 

Merging now!

PR #3692, opened 2021-02-17
Fix Failing Stats Availability Network Request

shiki:
Ref p4a5px-2If-p2#comment-11882.

I found out that the [`checkStatsAvailability`](https://github.com/woocommerce/woocommerce-ios/blob/59ab55ba14463c84bc4f6b9cef0eea0b0e4963ab/Yosemite/Yosemite/Stores/AvailabilityStore.swift#L40-L54) action was always failing because we were passing an incorrect date format (not ISO-8601). 

This PR fixes that. 

@jaclync Can you confirm if the failing API call is not intentional? 

## Testing

1. Use WC > 4.0. 
1. Add a breakpoint here:
    https://github.com/woocommerce/woocommerce-ios/blob/de627f40debb3897d4cc6f93654e35f15343bdb6/Yosemite/Yosemite/Stores/AvailabilityStore.swift#L50
2. Run the app and log in. 
3. Open the Stats page. 
4. Confirm the breakpoint was hit and that the `error` object is `nil`. 

## Author Checklist

### DIFF
--- a/Networking/Networking/Remote/OrderStatsRemoteV4.swift
+++ b/Networking/Networking/Remote/OrderStatsRemoteV4.swift
@@ -10,23 +10,25 @@ public final class OrderStatsRemoteV4: Remote {
     ///   - siteID: The site ID
     ///   - unit: Defines the granularity of the stats we are fetching (one of 'hourly', 'daily', 'weekly', 'monthly', or 'yearly')
     ///   - latestDateToInclude: The latest date to include in the results.
-    ///     This string is ISO8601 compliant
     ///   - completion: Closure to be executed upon completion.
     ///
     /// Note: by limiting the return values with the `_fields` param, we shrink the response size by over 90%! (~40kb to ~3kb)
     ///
     public func loadOrderStats(for siteID: Int64,
                                unit: StatsGranularityV4,
-                               earliestDateToInclude: String,
-                               latestDateToInclude: String,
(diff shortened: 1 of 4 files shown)


jaclync (approved):
Thank you @shiki for fixing this! I noticed the `before`/`after` param issue for the availability check before, was going to work on this the other day but then the requests stopped failing 😓 

Code: ✅ - thanks for making the interface better by passing the earliest/latest dates as `Date`!
Testing: ✅ 

\ | before | after
-- | -- | --
URL | https://public-api.wordpress.com/rest/v1.1/jetpack-blogs/161199960/rest-api/?json=true&path=/wc-analytics/reports/revenue/stats%26_method%3Dget%26interval%3Dyear%26after%3D635309042.795412%26before%3D635309042.795412%26per_page%3D41 | https://public-api.wordpress.com/rest/v1.1/jetpack-blogs/161199960/rest-api/?json=true&path=/wc-analytics/reports/revenue/stats%26_method%3Dget%26after%3D2021-02-18T10%3A52%3A48%26per_page%3D41%26before%3D2021-02-18T10%3A52%3A48%26interval%3Dyear
Response Code | 400 | 200
Response Body | {"error" : "rest_invalid_param", "message" : "Invalid parameter(s): after, before"} | {"data" : { ... }}

shiki:
Thank you for the quick review, Jaclyn! 🤸

PR #3694, opened 2021-02-18
Order Details: Fix Product Variations showing empty data if attribute equals Any

### LINKED ISSUE #3648: Product Variations cells in Order Detail shows empty data
Product Variations with Attributes equals to `Any` is not shown correctly inside the Order Detail screen because the cell will show just some commas. The Product variation detail shows them correct.

| Order Detail (look at the cell)             |  Product Variation detail |
Simulator Screen Shot - iPhone 12 Pro - 2021-02-12 at 12 11 58 | Simulator Screen Shot - iPhone 12 Pro - 2021-02-12 at 12 11 56

momo-ozawa:
Resolves https://github.com/woocommerce/woocommerce-ios/issues/3648

### Description
- Fixed a bug where the Order Detail shows empty data if the attribute equals `Any`

### How to test
1. Create a variable product and set the variation attribute(s) to `Any`
2. On wp-admin, create an order for the product you created in step 1
3. Open the order you just created in the app
4. ✅ The product variation cell should display `Any {attributeName}` instead of empty data
5. Tap "Details" then tap on the product
6. ✅ The product variation detail should display `Any {attributeName}` (regression testing)

### Screenshot

Before (Order Detail) | After (Order Detail) | After (Product Variation Detail, no regression)
--- | --- | ---
Simulator Screen Shot - iPhone 12 Pro - 2021-02-18 at 15 56 07 | Simulator Screen Shot - iPhone 12 Pro - 2021-02-18 at 17 47 27 | Simulator Screen Shot - iPhone 12 Pro - 2021-02-18 at 17 47 35

### Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
@@ -27,8 +27,12 @@ struct VariationAttributeViewModel {
         self.value = value
     }
 
-    init(variationAttribute: ProductVariationAttribute) {
-        self.init(name: variationAttribute.name, value: variationAttribute.option)
+    init(orderItemAttribute: OrderItemAttribute) {
+        self.init(name: orderItemAttribute.name, value: orderItemAttribute.value)
+    }
+
+    init(productVariationAttribute: ProductVariationAttribute) {
+        self.init(name: productVariationAttribute.name, value: productVariationAttribute.option)
     }
 }
 
@@ -122,7 +126,7 @@ struct ProductDetailsCellViewModel {
                   positiveTotal: formatter.convertToDecimal(from: item.total)?.abs() ?? NSDecimalNumber.zero,
                   positivePrice: item.price.abs(),
                   skuText: item.sku,
-                  attributes: item.attributes.map { VariationAttributeViewModel(name: $0.name, value: $0.value) })
+                  attributes: item.attributes.map { VariationAttributeViewModel(orderItemAttribute: $0) })
     }
 
     /// Aggregate Order Item initializer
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
```diff
@@ -1,15 +1,37 @@
...
+    }
+}
+
+private extension VariationAttributeViewModel {
+    enum Localization {
+        static let anyAttributeFormat =
+            NSLocalizedString("Any %1$@", comment: "Format of a product variation attribute description where the attribute is set to any value.")
+    }
+}
+
```
momo-ozawa: I wasn't 100% sure where I should put this model... open to any suggestions! 🙏

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
```diff
@@ -1,15 +1,37 @@
 import Foundation
 import Yosemite
 
+protocol AnyAttributeDisplayable {
+    var name: String { get set }
+    var value: String { get set }
+    var nameOrValue: String { get }
+}
```
Ecarrion: Do you have in mind some functionality for this protocol?  I don't see it being used anywhere! :-)

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
```diff
@@ -1,15 +1,37 @@
...
+
+struct VariationAttributeViewModel: AnyAttributeDisplayable {
+
+    var name: String = ""
+
+    var value: String = ""
```
Ecarrion: Do you have any reason to have these properties as `var`s. If they are not going to be mutated I think having them as `let`s express the intention better to the reader!

Ecarrion (commented):
Thanks for this @momo-ozawa! I just left a couple of simple questions! :-)

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
momo-ozawa: Good point, I'll remove the protocol!

WooCommerce/Classes/ViewModels/ProductDetailsCellViewModel.swift
momo-ozawa: Fixed in d68c3d3!

momo-ozawa:
@Ecarrion Thanks for reviewing! Ready for another look 😄

Ecarrion (approved):
🚀

PR #3696, opened 2021-02-18
Shipping Labels: display the correct store owner full name in the Origin Shipping Address

pmusolino:
Part of #2973

## Description
As part of Shipping Labels M2, we should fetch and display the correct full name of the store owner in the origin address. After having implemented the first name and last name in the `AccountSettings` entity, in this PR we fetch the last Account Settings from the stored entity and I also fixed an issue to the `ReadOnlyConvertible` entity that was not properly handling `firstName` and `lastName`, not updating them in storage. I also added a test for this case.

## Testing
2. Open an order detail.
4. Press the button "Continue" in the Create Shipping Label form.
5. The new screen for validating an address should be presented with all the information. -> You should see the `name` field populated with the correct data (first name and last name of your account).

## Screenshots

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -12,8 +12,15 @@ final class ShippingLabelFormViewModel {
     let destinationAddress: ShippingLabelAddress?
 
     init(originAddress: Address?, destinationAddress: Address?) {
+        let accountSettings = ShippingLabelFormViewModel.getStoredAccountSettings()
+        let company = ServiceLocator.stores.sessionManager.defaultSite?.name
+        let defaultAccount = ServiceLocator.stores.sessionManager.defaultAccount
+
         self.originAddress = ShippingLabelFormViewModel.fromAddressToShippingLabelAddress(address: originAddress) ??
-            ShippingLabelFormViewModel.getDefaultOriginAddress()
+            ShippingLabelFormViewModel.getDefaultOriginAddress(accountSettings: accountSettings,
+                                                               company: company,
+                                                               siteAddress: SiteAddress(),
+                                                               account: defaultAccount)
         self.destinationAddress = ShippingLabelFormViewModel.fromAddressToShippingLabelAddress(address: destinationAddress)
     }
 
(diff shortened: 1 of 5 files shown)


Yosemite/Yosemite/Model/Storage/AccountSettings+ReadOnlyConvertible.swift
```diff
@@ -6,11 +6,13 @@ import Storage
...
 extension Storage.AccountSettings: ReadOnlyConvertible {
 
-    /// Updates the Storage.AccountSettings with the a ReadOnly.
+    /// Updates the Storage.AccountSettings with the ReadOnly.
     ///
     public func update(with accountSettings: Yosemite.AccountSettings) {
         userID = accountSettings.userID
         tracksOptOut = accountSettings.tracksOptOut
+        firstName = accountSettings.firstName
+        lastName = accountSettings.lastName
```
jaclync, line 15: ah, these are easy to miss!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -12,8 +12,15 @@ final class ShippingLabelFormViewModel {
     let destinationAddress: ShippingLabelAddress?
 
     init(originAddress: Address?, destinationAddress: Address?) {
+        let accountSettings = ShippingLabelFormViewModel.getStoredAccountSettings()
+        let company = ServiceLocator.stores.sessionManager.defaultSite?.name
+        let defaultAccount = ServiceLocator.stores.sessionManager.defaultAccount
```
jaclync, line 17: wow, the default origin address information are scattered in many places 😅

Yosemite/YosemiteTests/Stores/AccountStoreTests.swift
```diff
@@ -169,6 +169,31 @@ class AccountStoreTests: XCTestCase {
...
+                promise(.success(account))
+            }
+            accountStore.onAction(action)
+        }
+
+        // Then
+        let account = try result.get()
+        XCTAssertEqual(self.viewStorage.countObjects(ofType: Storage.AccountSettings.self), 1)
+        XCTAssertEqual(account?.firstName, "Dem 123")
+        XCTAssertEqual(account?.lastName, "Nines")
```
jaclync: nit: what do you think of asserting on the whole `AccountSettings` (it already implements `Equatable`) so that we don't forget to assert on any new attributes in the future?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -69,4 +77,12 @@ extension ShippingLabelFormViewModel {
                                                         postcode: address.postcode)
         return shippingLabelAddress
     }
+
+    static func getStoredAccountSettings() -> AccountSettings? {
+        let storageManager = ServiceLocator.storageManager
+
+        let resultsController = ResultsController<StorageAccountSettings>(storageManager: storageManager, sortedBy: [])
+        try? resultsController.performFetch()
+        return resultsController.fetchedObjects.first
```
jaclync, line 86: non-blocking thought 💬 I wish we don't have to couple the view model with the storage layer here - the alternative is to retrieve the account settings via Yosemite layer like `AccountStore`, the downside is the result comes back asynchronously 😐

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -12,8 +12,15 @@ final class ShippingLabelFormViewModel {
...
+        let accountSettings = ShippingLabelFormViewModel.getStoredAccountSettings()
+        let company = ServiceLocator.stores.sessionManager.defaultSite?.name
+        let defaultAccount = ServiceLocator.stores.sessionManager.defaultAccount
+
         self.originAddress = ShippingLabelFormViewModel.fromAddressToShippingLabelAddress(address: originAddress) ??
-            ShippingLabelFormViewModel.getDefaultOriginAddress()
+            ShippingLabelFormViewModel.getDefaultOriginAddress(accountSettings: accountSettings,
+                                                               company: company,
+                                                               siteAddress: SiteAddress(),
+                                                               account: defaultAccount)
```
jaclync, line 23: nit: maybe we could have some unit tests on the initial origin and destination address now that there is quite some logic involved?

jaclync (approved):
LGTM 🚢 

Code: ✅ just some non-blocking thoughts!
Testing: ✅ the name is displayed correctly:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: Unfortunately, yes, and we also have to do several magical things under the hood to transform the things into the information we need (I'm talking about the transformations between the various entities). 🛸

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: I've thought about it, but I think this brings extra complexity in handling the things that maybe we can avoid right now.  This flow will be full of API requests, and having the information asynchronously will lead to many problems in verifying the address in the future, it is practically necessary to create a queue of calls, to make sure that the source address that is being verified is complete (and that therefore the information has been taken from the API). I would say not to discard this option, and to evaluate it as soon as we have a complete view of the flow for verifying the address, so as to understand if it is easily integrated, or not necessary. 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: I started implementing them in a later PR 😅

Yosemite/YosemiteTests/Stores/AccountStoreTests.swift
pmusolino: It's a good suggestion, I applied the change here ebec70f303f66dfb92d37c274de0986951384564 👍

PR #3698, opened 2021-02-18
Variations: Add More Variations

### LINKED ISSUE #3539: Variations: Creating more variations
While in the variation list(non-empty) screen, after tapping the "Add Variation" button, we should create a new variation with the existing product attributes.

Screen Shot 2021-01-27 at 2 52 57 PM

Tasks:
- [x] Show loading screen
- [x] Create variation 
- [x] Update screen with new variation

Ecarrion:
closes #3539 

# Why

After creating our first variation, we need to give the merchant the ability to generate more variations using the product attributes(previously added when creating the first variation).

# How

- Extract variation creation into a `GenerateVariationUseCase` to it can be reused in `EditAttributesViewModel` and in `ProductVariationsViewController`

- Call the `GenerateVariationUseCase` from the `ProductVariationsViewModel`

- Update `ProductVariationsViewController` to call the `generateVariation()` method and present a loading screen whit the action is being performed.

# Demo

create-more-variations

# Testing Steps

- Navigate into a variable product that has variations
- Tap in the variations row
- Tap on "Add Variation" button
- See the loading screen appears
- See the new variation in the list with no options selected (Any xxx)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewModel.swift
@@ -49,19 +49,8 @@ extension EditAttributesViewModel {
     /// Generates a variation in the host site using the product attributes
     ///
     func generateVariation(onCompletion: @escaping (Result<Product, Error>) -> Void) {
-        let action = ProductVariationAction.createProductVariation(siteID: product.siteID,
-                                                                   productID: product.productID,
-                                                                   newVariation: createVariationParameter()) { [weak self] result in
-            guard let self = self else { return }
-
-            // Convert the variationResult into a productResult by appending the variationID to the variations array
-            let productResult = result.map { variation in
-                self.product.copy(variations: self.product.variations + [variation.productVariationID])
-            }
-
-            onCompletion(productResult)
-        }
-        stores.dispatch(action)
+        let useCase = GenerateVariationUseCase(product: product, stores: stores)
+        useCase.generateVariation(onCompletion: onCompletion)
     }
 }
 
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewModel.swift
```diff
@@ -49,19 +49,8 @@ extension EditAttributesViewModel {
...
-            // Convert the variationResult into a productResult by appending the variationID to the variations array
-            let productResult = result.map { variation in
-                self.product.copy(variations: self.product.variations + [variation.productVariationID])
-            }
-
-            onCompletion(productResult)
-        }
-        stores.dispatch(action)
+        let useCase = GenerateVariationUseCase(product: product, stores: stores)
+        useCase.generateVariation(onCompletion: onCompletion)
```
jaclync, line 53: nice!

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -567,6 +573,23 @@ extension ProductVariationsViewController: SyncingCoordinatorDelegate {
...
         ServiceLocator.stores.dispatch(action)
     }
+
+    /// Creates a variation and presents a loading screen while it is created.
+    ///
+    private func createVariation() {
+        let progressViewController = InProgressViewController(viewProperties: .init(title: Localization.generatingVariation,
+                                                                                    message: Localization.waitInstructions))
+        present(progressViewController, animated: true)
+        viewModel.generateVariation { [onProductUpdate, noticePresenter] result in
```
jaclync, line 583: ooh does the capture list `[onProductUpdate, noticePresenter]` mean that they are strongly captured?

jaclync (approved):
Working nicely! :shipit: The fullscreen modal felt a bit much personally (since I still has to select the options and set a price for the variation later), but we don't really have a less prominent waiting UI

I also created a separate issue for a minor copy update that I noticed from testing 😄 https://github.com/woocommerce/woocommerce-ios/issues/3702

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
Ecarrion: Yes! I don't see any reference cycle so I think it's sage to capture those strongly.

Ecarrion:
Thanks for the review @jaclync

Issue #3509, opened 2021-01-22
Mobile Payments M1: Epic: Connecting BBPOS Chipper 2X BT Readers


allendav:
This epic contains the work needed to discover, connect and forget BBPOS Chipper 2X BT card readers for US merchants. Whenever possible we want to leverage the terminal mock to avoid having to introduce the location serivces and bluetooth capabilities prematurely to develop

- [x] Add feature flag
- [x] Add remote for connection token
- [x] Fetch whether a site has a WCPay account
- [x] Fetch site's current WCPay mode (test or live or not active)
- [ ] Fetch whether site has ordered a reader
- [x] Add entity (or UserDefaults) for known readers
- [x] Add reader discovery delegate (service) and terminal mock
- [x] Add terminal delegate to service
- [x] Add settings view for readers
- [x] Add discovered readers to settings view
- [x] Add connect-to-reader to settings view
- [x] Add disconnect-and-forget-reader to settings view
- [x] Add reader SW update view
- [ ] Add reader status bar subview
- [ ] Add buy-a-reader view
- [x] Enable location services and bluetooth in plist

ctarda:
Closing. This one seems outdated.

PR #3699, opened 2021-02-19
[Mobile Payments] Add Partial Card Reader Settings View Model and Connect Views

allendav:

- Adds a view for merchants connecting to a card reader for the first time. This is the first of many pull requests for the card reader settings user interface screens.
- Partially addresses "Add settings view for readers" for #3509
- Only addresses the "Connect Your Card Reader" screen
- Needs Dark Mode improvements #3738

To Test:

- Navigate to Settings (gear, top right)
- Select Manage Card Readers
- Tap on Connect Card Reader
- Note the alert (this is a temporary alert - FancyAlert will be in a subsequent PR)
- Dismiss the alert
- Tap on the Learn More row
- Note the woocommerce.com/payments page loads
- Return to the Woo app
- Tap on the left arrow
- Note you are returned to the settings screen

Screengrabs:

Sorry this PR is over 500 lines

@Garance91540 - very interested in your thoughts - I tried to pick up image and font sizes and margins from figma, but let me know if I missed anything

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectViewController.swift
```diff
@@ -0,0 +1,8 @@
+import UIKit
+
+class CardReaderSettingsConnectViewController: UIViewController {
```
ctarda: I understand this is still a placeholder. 

Unless you really want to subclass this (and in that case I'd argue there might be other solutions worth exploring) I would suggest defaulting to marking classes as final.

Back in the day [Apple suggested declaring classes as final also had performance implications](https://developer.apple.com/swift/blog/?id=27). I am not sure if that still holds true, though.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,9 +1,47 @@
 import UIKit
+import Combine
+import OSLog
 
 class CardReaderSettingsViewController: UIViewController {
+    private var viewmodel: CardReaderSettingsViewModel
```
ctarda: Have you considered passing the view model as a parameter of the initializer?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,9 +1,47 @@
...
+    private var previousChildView: CardReaderSettingsActiveChildView = .none
+    private var subscriptions = Set<AnyCancellable>()
+
+    required init?(coder: NSCoder) {
```
ctarda: Are we going to initialize this VC programatically?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,9 +1,47 @@
...
+        self.viewmodel = CardReaderSettingsViewModel()
+        super.init(coder: coder)
+
+        os_log("In CRSVC init")
```
ctarda: Ah! The cool stuff! We use DDLog directly, if this log is for yourself, to trace that the VC is initialized while building its implementation, then maybe a print is enough?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -0,0 +1,31 @@
+import Foundation
+import Combine
+import OSLog
+
+struct CardReader {
```
ctarda, line 4: I was going to comment "Have you tried to consume Yosemite.CardReader?", which made me realize that we might have lost that model object when I merged develop into our feature branch yesterday. I'll submit a PR to bring it back soon

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: Are you thinking we should support being able to test the VC by passing in a mock VM?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: No. Not at this time. This was leftover from a previous approach. Will remove.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: Yep - this was just for me to monitor lifecycle as I built out the foundation. Will remove.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectViewController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3699/commits/46c1941a4b75b3595b156f57b130316575113c0e

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3699/commits/46c1941a4b75b3595b156f57b130316575113c0e

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3699/commits/46c1941a4b75b3595b156f57b130316575113c0e

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
```diff
@@ -0,0 +1,31 @@
...
+        sharedInit()
+    }
+
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        sharedInit()
+    }
+
+    private func sharedInit() {
+        Bundle.main.loadNibNamed("CardReaderSettingsConnectView", owner: self, options: nil)
```
ctarda: There are some helpers in UIView+Helpers that you might want to look into, that would help with this

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
```diff
@@ -0,0 +1,31 @@
...
+        addSubview(contentView)
+        contentView.frame = self.bounds
+        contentView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
```
ctarda: Would it be possible to declare this using auto layout constraints instead of frames and autoresizing masks?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
```diff
@@ -0,0 +1,31 @@
+import UIKit
+
+final class CardReaderSettingsConnectView: UIView {
```
ctarda: This looks bit more like a view controller than a view to me. I'm curious, what would be the benefit of declaring this as a view?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,9 +1,91 @@
...
+    private var viewmodel = CardReaderSettingsViewModel()
+    private var subView: UIView?
+    private var alert: UIAlertController?
 
-class CardReaderSettingsViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
         configureNavigation()
+
+        self.viewmodel = CardReaderSettingsViewModel()
```
ctarda, line 19: it might be interesting to explore injecting this view model trough the constructor.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,9 +1,91 @@
...
+            noop()
+        case .updateRequired:
+            noop()
+        case .updating:
+            noop()
+        }
+    }
+
+    private func addCleanSlateView() {
+        let connectView = CardReaderSettingsConnectView(frame: view.frame)
```
ctarda: I see why it's a view now... I am curious, considering CardReaderSettingsConnectView seems to only wrap a button, would it make sense to just add the button here?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,9 +1,91 @@
 import UIKit
+import Combine
+
+final class CardReaderSettingsViewController: UIViewController {
+    private var subscriptions = Set<AnyCancellable>()
+    private var viewmodel = CardReaderSettingsViewModel()
+    private var subView: UIView?
+    private var alert: UIAlertController?
```
ctarda, line 10: I am not sure it is necessary to retain the alert view controller, specially if it is a FancyAlertViewController.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,9 +1,91 @@
...
+    private func addListView() {
+        // TODO Implement
+    }
+
+    private func noop() {
+        // TODO cleanup
+    }
+
+    private func addSearchingModal() {
+        // TODO Use FancyAlert instead
```
ctarda, line 90: Oh, ok. I should have reached this point before commenting about the FancyAlertViewController 😊

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: I am retaining it so, if/when the VM says it can go (e.g. searching finds a reader), the VC can dismiss it

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
```diff
@@ -0,0 +1,197 @@
+import UIKit
+
+final class CardReaderSettingsConnectView: NSObject {
```
ctarda, line 3: Do we really need to extend NSObject? 

The name suffix (`view`) kind of suggest to me this would be a view, and it seems more like a controller/viewmodel?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
```diff
@@ -0,0 +1,197 @@
...
+        cell.selectionStyle = .none
+    }
+
+    private func configureImage(cell: ImageTableViewCell) {
+        cell.detailImageView?.image = UIImage(named: "card-reader-connect")
+        cell.hideSeparator()
+        cell.selectionStyle = .none
+    }
+
+    private func configureHelp1(cell: NumberedListItemTableViewCell) {
```
ctarda: Would it make sense to name these methods in a way that indicate what each help button does? Maybe primaryHelp, secondaryHelp...

ctarda (approved):
This looks great @allendav !

I would suggest adding a ticket to inject the view model into the view controller, and add unit tests to both of them. 

Other than that, :shipit: !

🚀

Garance91540:
Hi @allendav ! Nice work! 

> @Garance91540 - very interested in your thoughts - I tried to pick up image and font sizes and margins from figma, but let me know if I missed anything

Is the PR ready for design review? FYI, in general when this is the case, the team tags me as a reviewer so I'm notified and don't forget to take a look :)

 If yes, here are a few notes:

- Are we going to implement the new Settings section in a separate PR? (not urgent)

- Manage card reader screen:
Screen Shot 2021-02-25 at 11 25 48 AM

- Scanning for reader
Screen Shot 2021-02-25 at 11 26 08 AM

Are we going to support the custom dialogs in a separate PR?

Lastly, when you get the chance, I'd be curious to see dark mode designs as well!

allendav:
@Garance91540 wrote:

> Are we going to implement the new Settings section in a separate PR? (not urgent)

Yes! https://github.com/woocommerce/woocommerce-ios/issues/3746

> Manage card reader screen

Will be fixed under https://github.com/woocommerce/woocommerce-ios/issues/3745

> Are we going to support the custom dialogs in a separate PR?

Yep! Coming soon!

> Lastly, when you get the chance, I'd be curious to see dark mode designs as well!

Yep! I need to fix a couple things: https://github.com/woocommerce/woocommerce-ios/issues/3738

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
allendav: [Copied from slack] NSObject is an interesting one. I only included it because I wasn't able to implement table data source without it. And as you note, its not quite a view and not quite a viewmodel

I mean, its basically a TableView except that the tableview is provided by the controller, and since all views in the card settings are table views and it avoids rough animation that could happen by tearing down and re-creating a table view

allendav:
@ctarda wrote:

> I would suggest adding a ticket to inject the view model into the view controller, and add unit tests to both of them.

Can do! https://github.com/woocommerce/woocommerce-ios/issues/3744

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
allendav: Yep! Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3699/commits/6d9903a53c2aa579399e2de43657d9c805077611

PR #3701, opened 2021-02-19
Update secrets in release 6.0

mokagio:
Done via `bundle exec fastlane run configure_update`.

You can test this manually by:

- Checking out `release/16.7`
- Making the `appstore_code_signing` lane public (that is, remove `private`)
- Running `bundle exec fastlane appstore_code_signing`
- Verifying it fails

Then 

- Check out this branch
- Run `bundle exec fastlane run configure_apply`
- Verifying it succeeds
- Making the `appstore_code_signing` lane public (that is, remove `private`)
- Running `bundle exec fastlane appstore_code_signing`
- Verifying it succeeds


jkmassel (approved):
I ran into a probably-unrelated `fastlane` issue with this (it may need to be updated?)

To test this anyway, I verified that the pinned hash is the correct one in the secrets repo, and verified that the decrypted file is identical to that of WPiOS using `md5 google_cloud_keys.json.enc`

Issue #3494, opened 2021-01-19
Products with a decimal number don't display order details for quantity


rossanafmenezes:
**Describe the bug**

Much like this issue here in the Android Repo [2838](https://github.com/woocommerce/woocommerce-android/issues/2838) this user has an extension to show the order quantity of a product in decimal. When they open the app it shows a message saying that "it was impossible to connect to the store" and you cannot access any data through the app, it's not only affecting the info of that
product but any data of the whole store. And when they change back the stock to a non-decimal value, the problem disappears.

**To Reproduce**

1 - Install WPC Product Quantity for WooCommerce (user has the Premium one but I tested with the regular version and was able to reproduce the issue
2 -   Go to a non-virtual product inventory
3 - add a decimal value to the inventory (I changed from 6 to 600.5)
4 - Go to the app and try to refresh the product screen

![Screenshot 2021-01-19 at 9 09 20 AM](https://user-images.githubusercontent.com/30724184/105046237-d7315380-5a36-11eb-85d0-60345bc547bf.png)

![IMG_0026](https://user-images.githubusercontent.com/30724184/105046264-e0babb80-5a36-11eb-9edf-32249082f4d2.jpg)

See the unable to Refresh message at the bottom. Got this message all across the app. 

Once you change the inventory back to a whole number the message disappears and the app refreshes and show the new inventory number (600 in this case)

![Screenshot 2021-01-19 at 9 09 26 AM](https://user-images.githubusercontent.com/30724184/105046555-3d1ddb00-5a37-11eb-9a3e-27248c541ec0.png)

![IMG_0027](https://user-images.githubusercontent.com/30724184/105046631-54f55f00-5a37-11eb-9597-a56e45b2454e.jpg)

**Expected behavior**

The product should be displayed on the order details screen with the correct product quantity in decimal.

**Isolating the problem (mark completed items with an [x]):**
- [x ] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x ] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x ] I can reproduce this bug consistently using the steps above.

**Mobile Environment**
Please include:

- Device: iPad 7th Generation
- iOS version: 14.2
- WooCommerce iOS version: 5.4.0.2 (before updating) and 5.7.0.3

User's info
3655729-zen
 IOS version is 13.1.3
User Woo version 5.6.0.2

**WordPress Environment**
<details>
[...]

PR #3705, opened 2021-02-19
Products: Support products and variations with decimal stock quantities

rachelmcr:
Related issue: #3494 
Related PR: https://github.com/woocommerce/woocommerce-ios/pull/3707

## Description

The WooCommerce Core API only supports stock quantities with `Int` or `nil` values. However, certain third-party extensions can enable decimal (non-integer) stock quantities. When products have decimal stock quantities, the list of products on the Products tab doesn't load (current behavior in `develop`).

This PR adds support for decimal stock quantities to be displayed in the app, so the product list will load.

Note: This does not add full support for editing products with decimal stock quantities (which requires an API change). For now, we will restrict products with decimal stock quantities to be read-only in the app; that change is in https://github.com/woocommerce/woocommerce-ios/pull/3707 along with a release note about the read-only support. Internal ref: p91TBi-4iF-p2#comment-3111

## Changes

* Adds a new Core Data model 46, which changes `ProductVariation.stockQuantity` from `Int64` to `Decimal`. (`Product.stockQuantity` is already stored as a `String` so it doesn't require any Core Data changes.)
* Adds a mapping model for Model 45 to Model 46.
* Adds a unit test for the Model 45 to Model 46 migration, confirming that the attribute type is changed as expected.
* Updates `stockQuantity` throughout the app from `Int64` to `Decimal`.

Before|After - Products|After - Product Variations
-|-|-
Simulator Screen Shot - iPhone 12 - 2021-02-19 at 16 26 33|Simulator Screen Shot - iPhone 12 - 2021-02-19 at 16 25 16|Simulator Screen Shot - iPhone 12 - 2021-02-19 at 16 46 12

## Testing

1. Switch to `develop`. Clean, build, and launch the app.
2. Log in and open the Products tab to fetch products. Open a variable product and open Variations to fetch the variations for that product.
3. On your store, enable decimal stock quantities by installing a third-party extension on your store like [WPC Product Quantity](https://wordpress.org/plugins/wpc-product-quantity/).
4. Update the inventory for at least one product and one product variation to have a decimal stock quantity.
5. Open the app, refresh the Products tab, and notice the products list doesn't load ("Unable to refresh list").
6. Kill the app.
7. Switch to this branch. Clean, rebuild, and launch the app.
8. Open the Products tab and confirm the products list now loads with no errors, showing the correct stock quantity for your product with a decimal stock quantity.
9. Open a variable product with at least one variation with a decimal stock quantity.
10. Open "Variations" and confirm the variations list loads, showing the correct stock quantity for the variation with a decimal stock quantity.

## Submitter Checklist

Update release notes:


Networking/Networking/Model/Product/Product.swift
```diff
@@ -463,7 +463,11 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         try container.encode(sku, forKey: .sku)
         try container.encode(manageStock, forKey: .manageStock)
         try container.encode(soldIndividually, forKey: .soldIndividually)
-        try container.encode(stockQuantity, forKey: .stockQuantity)
+
+        // API currently only accepts integer values for stock quantity
+        let integerStockQuantity = Int(truncating: stockQuantity as NSDecimalNumber? ?? 0)
```
Ecarrion, line 468: [Small Tip] you can use `map`, to convert an optional type without having to deal with coalescing operators, which makes the code a bit hard to read.
```swift
let integerStockQuantity = stockQuantity.map { NSDecimalNumber(decimal: $0).intValue }
```

Ecarrion (approved):
Test 🚀 
Code: 🚀 

This would be very good for merchants and for the "product load" problems.
I really like the fact that we are shipping whats possible while the API gets improved.

Networking/Networking/Model/Product/Product.swift
rachelmcr: Thanks, that's a great tip! I wasn't familiar with this as a way to avoid dealing with coalescing operators.

I'll leave this as-is for now (in this PR), since https://github.com/woocommerce/woocommerce-ios/pull/3717 changes how we do this encoding.

PR #3706, opened 2021-02-19
[Tooling] fix AppStore metadata for release 6.0

AliSoftware:
## What?

Update the `download_metadata.swift` script to allow it to be run from anywhere, and remove extra `metadata/` folder commited at root accidentally a couple of sprints ago.

## Why?

While doing the release for WC 6.0, I encountered an issue with the translations, as Release Notes were not up-to-date in GlotPress.

I reached out to #i18n channel [p1613754174122400-slack-C02AED43D] so they update it, then ran `fastlane/download_metadata.swft` to re-download them, and tried again. But the HTML preview still showed the old, untranslated ones, so I was confused. Took me a couple of minutes to realise that this was because I ran the `fastlane/download_metadata.swft` script from the root of the repository, while I was instead supposed to `cd fastlane && ./download_metadata.swift`.

* So I did a quick fix to the Swift script to allow it to be run from anywhere, by computing the `baseFolder` relative to the path of the script itself.
* Then ran the script again, which updated the metadata of the 4 locales which were missing a translation from the previous run (before #i18n's sync), hence the 4 _new_ files in this PR.
* But then processed to removing the extra files that I previously created accidentally at the root of the repository (in `<reporoot>/metadata`)… and realized that those files were in fact previously commited 😱 . So even in `develop` we have had both `metadata/*` and `fastlane/metadata/*` folders tracked by git apparently, while the `fastlane/metadata/*` ones are supposed to be the source of truth. They probably have been commited accidentally by me the last time I ran into the exact same issue and made the exact same mistake a couple of sprints ago. So I removed them from git, hence the many deleted files in this PR.

## Review

While reviewing this, I'd appreciate if you could double-check that the metadata we _keep_ (and which are not visible in that PR's diff) seems the most up-to-date to you, and that we didn't previously accidentally manually update the ones from the root of the repo (that we're deleting here) at some point… while leaving the ones in `fastlane/metadata` (that we keep, not shown in PR) out of sync and behind the new values.


jkmassel (approved):
I tested this by running `./fastlane/download_metadata.swift` and `cd fastlane && ./download_metadata.swift`. 

Both ran successfully and didn't dirty the repo

PR #3707, opened 2021-02-19
Products: Make products and variations with decimal stock quantities read-only

rachelmcr:
Related: #3494
Depends on: #3705 

## Description

The WooCommerce Core API only supports stock quantities with `Int` or `nil` values. However, certain third-party extensions can enable decimal (non-integer) stock quantities. In #3705 we added support for loading products and variations with decimal stock quantities in the app.

This PR restricts products and variations with decimal stock quantities to be read-only. Any products or variations without decimal stock quantities can still be edited as usual. (We currently can't edit products with decimal stock quantities, since that requires an API change.)

## Changes

* Adds a computed property `hasDecimalStockQuantity` to `Product` and `ProductVariation`.
* Wherever an editable product or variation may be loaded in the app, `hasDecimalStockQuantity` is used to force a ready-only mode for that product or variation:
   * `ProductsViewController` to open details for selected product
   * `ProductVariationsViewController` to open details for selected variation
   * `ProductLoaderViewController` to open details for selected product or variation (used to load details from My Store > Top Performers list and Orders > Order details)

## Testing

1. On your store, enable decimal stock quantities by installing a third-party extension on your store like [WPC Product Quantity](https://wordpress.org/plugins/wpc-product-quantity/).
2. Update the inventory for at least one product and one product variation to have a decimal stock quantity.
3. Create a new order for products/variations with and without decimal stock quantities.

**Products**
1. In the app, go to the Products tab.
2. Select a product that **doesn't** have a decimal stock quantity.
3. Confirm the product details open and they are editable.
4. Go back to the product list and select a product that **does** have a decimal stock quantity.
5. Confirm the product details open and they are read-only.
6. Go back to the product list and select a variable product with a mix of decimal and non-decimal stock quantities.
7. Select "Variations."
8. Select a variation that **doesn't** have a decimal stock quantity.
9. Confirm the variation details open and they are editable.
10. Go back to the variation list and select a variation that **does** have a decimal stock quantity.
11. Confirm the variation details open and they are read-only.

**Top Performers**
1. In the app, go to the My store tab.
2. Go to the Top Performers list and select a product/variation that **doesn't** have a decimal stock quantity.
3. Confirm the product details open and they are editable.
4. Select a product/variation that **does** have a decimal stock quantity.
5. Confirm the product details open and they are read-only.

**Order Details**
1. In the app, go to the Orders tab.
2. Select the order you created for products/variations with and without decimal stock quantities.
3. Select a product/variation that **doesn't** have a decimal stock quantity.
4. Confirm the product details open and they are read-only. (Products in order details should always be read-only.)
5. Select a product/variation that **does** have a decimal stock quantity.
6. Confirm the product details open and they are read-only.

## Submitter Checklist

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/Product.swift
+++ b/Networking/Networking/Model/Product/Product.swift
@@ -109,6 +109,14 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         return ProductTaxStatus(rawValue: taxStatusKey)
     }
 
+    /// Whether the product has a decimal (non-integer) stock quantity.
+    /// Can be used to determine if the product should be loaded as read-only.
+    /// Related issue: https://github.com/woocommerce/woocommerce-ios/issues/3494
+    ///
+    public var hasDecimalStockQuantity: Bool {
+        stockQuantity?.exponent ?? 0 < 0
+    }
+
     /// Product struct initializer.
     ///
     public init(siteID: Int64,
--- a/Networking/Networking/Model/Product/ProductVariation.swift
+++ b/Networking/Networking/Model/Product/ProductVariation.swift
@@ -55,6 +55,17 @@ public struct ProductVariation: Codable, GeneratedCopiable, Equatable {
 
     public let menuOrder: Int64
 
+    /// Computed Properties
+    ///
+
+    /// Whether the product variation has a decimal (non-integer) stock quantity.
+    /// Can be used to determine if the product variation should be loaded as read-only.
+    /// Related issue: https://github.com/woocommerce/woocommerce-ios/issues/3494
+    ///
+    public var hasDecimalStockQuantity: Bool {
+        stockQuantity?.exponent ?? 0 < 0
+    }
+
     /// ProductVariation struct initializer.
     ///
     public init(siteID: Int64,
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,9 @@
 *** PLEASE FOLLOW THIS FORMAT: [<priority indicator, more stars = higher priority>] <description> [<PR URL>]
 
+6.2
+-----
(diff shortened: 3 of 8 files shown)


Ecarrion (commented):
Code & Testing looks good, but I wonder if there is a better solution. 🤔 

----

What do you think about modifying if the inventory cell is actionable/editable and only prevent the user modify that part of the product? (We may need to update the encoding to not encode decimal numbers)

Hint: I think here is where that `actionable` logic for that cell resides.

https://github.com/woocommerce/woocommerce-ios/blob/a06ba7534a1b1f566e45dfe0150df70164c6953d/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/DefaultProductFormTableViewModel.swift#L198-L221

----

If the above is not possible, should we at least show some kind of message saying that the product is on read-only mode because the store is using some unsupported plugins?
cc @woocommerce/mobile-designers

rachelmcr:
Thanks for the suggested alternative! I definitely like the idea of some messaging (like a top banner on the product details screen?) about why the product is read-only.

> What do you think about modifying if the inventory cell is actionable/editable and only prevent the user modify that part of the product? (We may need to update the encoding to not encode decimal numbers)

I agree, this would be a lot nicer. I tried this approach earlier and had some trouble getting it to work, but it's possible I wasn't updating the encoding correctly — I might have given up too early on that. I'll give it another try!

Ecarrion:
> I agree, this would be a lot nicer. I tried this approach earlier and had some trouble getting it to work, but it's possible I wasn't updating the encoding correctly — I might have given up too early on that. I'll give it another try!

Great, feel free to ping me if you need any assistance!

rachelmcr:
Making that change (modifying on whether the inventory cell is actionable/editable) ended up being easier than I thought! I'll close this in favor of https://github.com/woocommerce/woocommerce-ios/pull/3717.

Issue #3711, opened 2021-02-19
Crash: UIKitCore: -[UIApplication _runWithMainScene:transitionContext:completion:] + 2020


astralbodies:
This crash is not hitting Sentry. Initially had a few people look at it, but we're not really sure what's going on here. 

<img width="1452" alt="2021-02-19_15-23-43" src="https://user-images.githubusercontent.com/373903/108563067-7a8bb780-72c6-11eb-9562-1a6b7e93d660.png">

jaclync:
Example full log below, it seems like Sentry was able to detect this crash (the main thread reached `SentryCrashMonitor_CPPException.cpp:179`) but somehow it isn't visible in Sentry 🤔 

[This Developer Forum post](https://developer.apple.com/forums/thread/128810) has the closest stack trace as this crash. If our crash is from the same cause, it appears to be from `Application windows are expected to have a root view controller at the end of application launch`.

```
Date/Time:           2021-02-15 20:38:31.6074 +0100
Launch Time:         2021-02-15 20:38:31.2653 +0100
OS Version:          iPhone OS 14.4 (18D52)
Release Type:        User
Baseband Version:    1.42.03
Report Version:      104

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Triggered by Thread:  0

Last Exception Backtrace:
0   CoreFoundation                	0x187fa986c __exceptionPreprocess + 220 (NSException.m:199)
1   libobjc.A.dylib               	0x19cfc4c50 objc_exception_throw + 60 (objc-exception.mm:565)
2   CoreFoundation                	0x187eaf000 +[NSException raise:format:arguments:] + 100 (NSException.m:146)
3   Foundation                    	0x18924391c -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 132 (NSException.m:231)
4   UIKitCore                     	0x18a962ff4 -[UIApplication _runWithMainScene:transitionContext:completion:] + 2020 (UIApplication.m:4191)
5   UIKitCore                     	0x189fb8c74 -[_UISceneLifecycleMultiplexer completeApplicationLaunchWithFBSScene:transitionContext:] + 152 (_UISceneLifecycleMultiplexer.m:436)
[...]

PR #3713, opened 2021-02-22
Make sure window's root view controller is set in all app states

jaclync:
Attempted fix for crash that is only in Xcode #3711 

## Why

We started seeing a crash that started in release 5.9 in Xcode, but this crash is somehow not visible in Sentry ([example full stack trace](https://github.com/woocommerce/woocommerce-ios/issues/3711#issuecomment-783264982)).

After some research online, [this Developer Forum post](https://developer.apple.com/forums/thread/128810) has the closest stack trace as this crash. If our crash is from the same cause, it appears to be from `Application windows are expected to have a root view controller at the end of application launch`.

I looked into the most suspicious PR https://github.com/woocommerce/woocommerce-ios/pull/3498 that I worked on for release 5.9, where I set the window's `rootViewController` based on the app state. In one scenario where `isLoggedIn` and `needsDefaultStore` are both `false`, it just does nothing. Even though this is a highly unlikely scenario where the app is logged out while it already has a default store, `needsDefaultStore` is from UserDefaults and it could be out of our control / expectation. Also, we've done a bunch of testing in release 5.9 and on `develop` and haven't been able to reproduce this crash yet - it could be because of this particular app state that it got in.

## Changes

This PR thus handled this unexpected case and added a logging to Sentry, which we will remove if this turns out to be the cause of the crash.

## Testing

Please confidence check with the following scenarios:

### Log in to a store, then log out

- Launch the app logged out --> the login UI is shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Select a store (if more than one) and tap "Continue" --> the tab bar UI should be shown
- Relaunch the app --> the tab bar UI should be shown
- Log out from the app --> the login UI should be shown

### Log in with credentials, then restart login flow

- Launch the app logged out --> the login UI should be shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Tap "Try another account" --> the login UI should be shown

### Log in with credentials, force close the app, then continue with a store

- Launch the app logged out --> the login UI should be shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Force close the app and relaunch (relaunch the app from Xcode while the app is still on the store picker screen) --> the store picker should be shown
- Select a store (if more than one) and tap "Continue" --> the tab bar UI should be shown

### Log in with credentials, force close the app, then restart login flow

- Launch the app logged out --> the login UI should be shown
- Enter login credentials associated with at least one WC store via any method you choose (email/password or magic link) --> the store picker should be shown
- Force close the app and relaunch (relaunch the app from Xcode while the app is still on the store picker screen) --> the store picker should be shown
- Tap "Try another account" --> the login UI should be shown (known issue: there is a glitch when it animates to the login UI https://github.com/woocommerce/woocommerce-ios/issues/2888)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
@@ -1,6 +1,7 @@
 import Combine
 import UIKit
 import Yosemite
+import class AutomatticTracks.CrashLogging
 
 /// Coordinates app navigation based on authentication state: tab bar UI is shown when the app is logged in, and authentication UI is shown
 /// when the app is logged out.
@@ -40,12 +41,14 @@ final class AppCoordinator {
                 switch (isLoggedIn, needsDefaultStore) {
                 case (false, true):
                     self.displayAuthenticator()
+                case (false, false):
+                    // TODO-3711: let's delete the logging once we verify the crash is caused by this case.
+                    CrashLogging.logMessage("Unexpected app state where `isLoggedIn` and `needsDefaultStore` are both `false`", level: .info)
+                    self.displayAuthenticator()
                 case (true, true):
                     self.displayStorePicker()
                 case (true, false):
                     self.displayLoggedInUI()
-                default:
-                    break
                 }
                 self.isLoggedIn = isLoggedIn
             }


pmusolino (commented):
Thank you Jaclyn for opening this PR. 

I wasn't able to replicate this case:
> Force close the app and relaunch (relaunch the app from Xcode while the app is still on the store picker screen) --> the store picker should be shown

Every time that I force close the app, and relaunch it, I return to the initial screen where I should decide if enter the store address or continue with WordPress.com. I think this isn't expected, right?

AliSoftware:
I'm cutting the `release/6.1` in a few hours, so if you want this PR to be included in 6.1 it would be nice to get it merged by then. Otherwise we could postpone it to get it to 2nd beta later, or to 6.2

AliSoftware:
I've cut the `release/6.1` branch already, BUT I haven't triggered the first beta yet.

Please update the target branch of this PR to `release/6.1` if you want it to get included in WCiOS 6.1. If this is then merged into `release/6.1` by the time I trigger the first 6.1.0.0 beta tomorrow, then all good and it will be part of it. If not, I'll probably do the beta 1 tomorrow anyway and do a 2nd beta later in the week, once this PR has landed in `release/6.1`.

jaclync:
Thanks for the quick review @pmusolino!

> Every time that I force close the app, and relaunch it, I return to the initial screen where I should decide if enter the store address or continue with WordPress.com. I think this isn't expected, right?

Did you manually force close the app, and relaunch it? Because [the store picker deauthenticates when the app is terminating](https://github.com/woocommerce/woocommerce-ios/blob/a06ba7534a1b1f566e45dfe0150df70164c6953d/WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift#L290-L294), this flow only works if the app is not receiving the `applicationWillTerminate(_:)` notification. This is only possible if we relaunch the app by rebuilding the app from Xcode while the app is still in the foreground. I just tested it again in iPhone 11 iOS 14.1 simulator and was able to repro this flow, lemme know if you still have issues!

pmusolino:
> Did you manually force close the app, and relaunch it? Because the store picker deauthenticates when the app is terminating, this flow only works if the app is not receiving the applicationWillTerminate(_:) notification. This is only possible if we relaunch the app by rebuilding the app from Xcode while the app is still in the foreground. I just tested it again in iPhone 11 iOS 14.1 simulator and was able to repro this flow, lemme know if you still have issues!

Got it! I was testing it closing the app in the simulator, and not force closing it from Xcode 👍 

LGTM :shipit:

pmusolino (approved):
:shipit:

PR #3717, opened 2021-02-22
Products: Make inventory settings read-only for decimal stock quantities

rachelmcr:
Related: #3494
Depends on: #3705

## Description

The WooCommerce Core API only supports stock quantities with Int or nil values. However, certain third-party extensions can enable decimal (non-integer) stock quantities. In #3705 we added support for loading products and variations with decimal stock quantities in the app.

This PR makes a product or variation's inventory settings read-only if it has decimal stock quantities, since the API currently doesn't accept decimal stock quantities. Other settings (and all settings for products or variations without decimal stock quantities) can still be edited as usual.

## Changes

* Adds a computed property `hasIntegerStockQuantity` to the `Product` and `ProductVariation` Networking models.
* Networking only encodes the stock quantity if `hasIntegerStockQuantity` is true. This prevents decimal quantities from being sent to the API, while allowing other settings to be updated as usual.
* The property `hasIntegerStockQuantity` is then used in `ProductFormActionsFactory` and  `ProductVariationFormActionsFactory` to determine if the Inventory settings cell is editable.
* Adds unit tests to check that the inventory settings cell is not editable when a product or variation has a decimal stock quantity.

/|Integer stock quantity|Decimal stock quantity
-|-|-
Product|Simulator Screen Shot - iPhone 12 - 2021-02-22 at 16 57 46|Simulator Screen Shot - iPhone 12 - 2021-02-22 at 16 57 36
Product variation|Simulator Screen Shot - iPhone 12 - 2021-02-22 at 16 58 00|Simulator Screen Shot - iPhone 12 - 2021-02-22 at 16 57 56

**Editing product settings with/without decimal stock quantities**

## Testing

**Setup**

1. On your store, enable decimal stock quantities by installing a third-party extension on your store like [WPC Product Quantity](https://wordpress.org/plugins/wpc-product-quantity/).
2. Update the inventory for at least one product and one product variation to have a decimal stock quantity.
3. Create a new order for products/variations with and without decimal stock quantities.

**Products**

1. In the app, go to the Products tab.
2. Select a product that **doesn't** have a decimal stock quantity.
3. Confirm the product details open and they are all editable, including Inventory.
4. Go back to the product list and select a product that **does** have a decimal stock quantity.
5. Confirm the product details open and Inventory is not editable, while other product settings are.
6. Make a change to the product settings and confirm you can update the product without any changes to the decimal stock quantity.

**Product Variations**

1. Go back to the product list and select a variable product with a mix of decimal and non-decimal stock quantities.
2. Select "Variations."
3. Select a variation that **doesn't** have a decimal stock quantity.
4. Confirm the variation details open and they are all editable, including Inventory.
5. Go back to the variation list and select a variation that **does** have a decimal stock quantity.
6. Confirm the variation details open and Inventory is not editable, while other variation settings are.
7. Make a change to the variation settings and confirm you can update the variation without any changes to the decimal stock quantity.

## Submitter Checklist

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/Product.swift
+++ b/Networking/Networking/Model/Product/Product.swift
@@ -109,6 +109,13 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         return ProductTaxStatus(rawValue: taxStatusKey)
     }
 
+    /// Whether the product variation has an integer stock quantity.
+    /// Decimal (non-integer) stock quantities currently aren't accepted by the Core API.
+    /// Related issue: https://github.com/woocommerce/woocommerce-ios/issues/3494
+    public var hasIntegerStockQuantity: Bool {
+        stockQuantity?.exponent ?? 0 >= 0
+    }
+
     /// Product struct initializer.
     ///
     public init(siteID: Int64,
@@ -465,8 +472,9 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         try container.encode(soldIndividually, forKey: .soldIndividually)
 
         // API currently only accepts integer values for stock quantity
-        let integerStockQuantity = Int(truncating: stockQuantity as NSDecimalNumber? ?? 0)
-        try container.encode(integerStockQuantity, forKey: .stockQuantity)
+        if hasIntegerStockQuantity {
+            try container.encode(stockQuantity, forKey: .stockQuantity)
+        }
 
         try container.encode(backordersKey, forKey: .backordersKey)
         try container.encode(stockStatusKey, forKey: .stockStatusKey)
--- a/Networking/Networking/Model/Product/Product.swift
+++ b/Networking/Networking/Model/Product/Product.swift
@@ -109,11 +109,15 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         return ProductTaxStatus(rawValue: taxStatusKey)
     }
 
-    /// Whether the product variation has an integer stock quantity.
+    /// Whether the product has an integer (or nil) stock quantity.
     /// Decimal (non-integer) stock quantities currently aren't accepted by the Core API.
     /// Related issue: https://github.com/woocommerce/woocommerce-ios/issues/3494
     public var hasIntegerStockQuantity: Bool {
-        stockQuantity?.exponent ?? 0 >= 0
+        guard let stockQuantity = stockQuantity else {
+            return true
+        }
+
+        return stockQuantity.isInteger
     }
 
     /// Product struct initializer.
--- a/WooCommerce/WooCommerceTests/ViewRelated/Products/Actions Factory/ProductVariationFormActionsFactory+ReadonlyVariationTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Products/Actions Factory/ProductVariationFormActionsFactory+ReadonlyVariationTests.swift
@@ -54,6 +54,22 @@ final class ProductVariationFormActionsFactory_ReadonlyVariationTests: XCTestCas
         let expectedBottomSheetActions: [ProductFormBottomSheetAction] = []
         XCTAssertEqual(factory.bottomSheetActions(), expectedBottomSheetActions)
     }
+
+    func test_variation_with_decimal_stock_quantities_has_read_only_inventory() {
+        // Arrange
+        let productVariation = Fixtures.variationWithDecimalStockQuantity
+        let model = EditableProductVariationModel(productVariation: productVariation)
+
+        // Action
+        let factory = ProductVariationFormActionsFactory(productVariation: model, editable: true)
+
+        // Assert
+        let expectedSettingsSectionActions: [ProductFormEditAction] = [.priceSettings(editable: true),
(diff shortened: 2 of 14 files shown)


Networking/Networking/Model/Product/Product.swift
```diff
@@ -109,6 +109,13 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         return ProductTaxStatus(rawValue: taxStatusKey)
     }
 
+    /// Whether the product variation has an integer stock quantity.
```
jaclync: super nit:

```suggestion
    /// Whether the product has an integer stock quantity.
```

Networking/Networking/Model/Product/Product.swift
```diff
@@ -109,6 +109,13 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
...
+    /// Decimal (non-integer) stock quantities currently aren't accepted by the Core API.
+    /// Related issue: https://github.com/woocommerce/woocommerce-ios/issues/3494
+    public var hasIntegerStockQuantity: Bool {
+        stockQuantity?.exponent ?? 0 >= 0
```
jaclync: nit: TIL'ed [`Double.exponent`](https://developer.apple.com/documentation/swift/double/1848471-exponent) though I'm not sure if I fully understand how this is used to check if a double is an integer 😅  do you think we could use a potentially easier solution like [this SO answer](https://stackoverflow.com/a/28448059/9185596)?

To reuse the logic with `ProductVariation`, we could also make an extension for `Double` like `Double.isInteger: Bool`?

WooCommerce/WooCommerceTests/ViewRelated/Products/Actions Factory/ProductVariationFormActionsFactory+ReadonlyVariationTests.swift
```diff
@@ -54,6 +54,22 @@ final class ProductVariationFormActionsFactory_ReadonlyVariationTests: XCTestCas
         let expectedBottomSheetActions: [ProductFormBottomSheetAction] = []
         XCTAssertEqual(factory.bottomSheetActions(), expectedBottomSheetActions)
     }
+
+    func test_variation_with_decimal_stock_quantities_has_read_only_inventory() {
```
jaclync, line 58: nice test cases! 💯

jaclync (approved):
Nice work supporting products with decimal stock quantity! :shipit: My test site already has a plugin ["All in One Product Quantity for WooCommerce"](https://wordpress.org/plugins/product-quantity-for-woocommerce/) that I tested decimal order quantity before, and I was able to use this plugin for decimal stock quantity to reproduce the product/variation list loading error

\ | before | after - list | after - details
-- | -- | -- | --
product | Simulator Screen Shot - iPhone 11 - 2021-02-23 at 11 11 48 | Simulator Screen Shot - iPhone 11 - 2021-02-23 at 11 12 15 | Simulator Screen Shot - iPhone 11 - 2021-02-23 at 11 12 31
variation | Simulator Screen Shot - iPhone 11 - 2021-02-23 at 11 11 01 | Simulator Screen Shot - iPhone 11 - 2021-02-23 at 11 12 53 | Simulator Screen Shot - iPhone 11 - 2021-02-23 at 11 13 01

Networking/Networking/Model/Product/Product.swift
rachelmcr: I chose this approach because `stockQuantity` is a `Decimal` type so we can check the exponent directly (versus converting it to a `Double` to do other math on it). It's based on [floating point arithmetic](https://en.wikipedia.org/wiki/Floating-point_arithmetic), where a negative exponent essentially means you're moving the decimal point to the left (to create a non-integer). For example, the exponent `-3` here:

> 12345 * 10^-3 = 12.345

Whereas an integer has a 0 or positive exponent:

> 12345 * 10^0 = 12345

or

> 12345 * 10^3 = 12345000

Do you think it'd make sense to use this with a comment explaining the math behind it? If that still seems like it'd be unclear, I can switch it out for a solution like the one you found.

Either way, great idea to make an extension to reuse the logic — I'll work on that!

rachelmcr: After thinking this over, I moved the existing `exponent` check into a `Decimal.isInteger` extension in https://github.com/woocommerce/woocommerce-ios/pull/3717/commits/0ea9a52a207202ebe91a25720d16dd6bb47ce8e9. I added an explanation there of how the arithmetic works, which hopefully should help with clarity. Let me know if you think it could still be improved!

I also added a unit test in https://github.com/woocommerce/woocommerce-ios/pull/3717/commits/5b522c9790debb448d1679d59901dec480d816a2 to confirm it works with a variety of potential values.

jaclync: The explanation you added is super helpful and easy to understand! I was confused by the example in [the Apple doc](https://developer.apple.com/documentation/swift/double/1848471-exponent) where the radix is 2 😅 

The extension changes look great with a unit test! 💯

PR #3724, opened 2021-02-22
Variations: Edit attribute options(offered) for generating variations

### LINKED ISSUE #3540: Variations: Finish editing attribute
After finishing updating an attribute this should be updated remotely.

Screen Shot 2021-01-27 at 3 18 20 PM

Tasks: 
- [x] Navigate to edit options from attribute list
- [x] Render selected options
- [x] Enable next button when changes are made
- [x] Update product attribute code to support editing an attribute.
- [x] Redirect the user to the attribute list screen

Ecarrion:
closes #3540 

# Why

After being able to create variations(and attributes), we need to add the ability to edit those attributes so merchants can keep creating and updating variations on mobile.

This PR gives the ability to merchants to edit the product attribute options offered.

# How

- Update `EditAttributesViewController` to navigate the merchant to the `AddAttributeOptionsViewController` after tapping on an attribute row.

- Update `AddAttributeOptionsViewModel` to preselect(show in the options offered section) all the attributes that the merchant selected in past sessions(reside in `product.attributes[index].options`)

- Update `AddAttributeOptionsViewModel.isNextButtonEnabled` to not enable the next button if the previously selected options are the same as the options about to submit.

# Demo

Update Attribute

# Testing Steps

- Go to a variable product with variations
- Tap on the variations row
- Tap on the more menu(...) and select edit attributes
- Tap on an attribute to edit.
- See that the next button is disabled and that there are some options pre-selected
- Edit the attribute by adding, removing or re-ordering an option.
- Tap Next
- Go back to the product variation detail screen and see that the variations row is updated with the new data

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
@@ -136,12 +136,13 @@ final class AddAttributeOptionsViewModel {
         switch attribute {
         case .new:
             updateSections()
-        case let .existing(attribute) where attribute.isGlobal:
-            synchronizeGlobalOptions(of: attribute)
-        case let .existing(attribute) where attribute.isLocal:
-            populateLocalOptions(of: attribute)
-        default:
-            break
+        case let .existing(attribute):
+            preselectCurrentOptions(of: attribute)
+            if attribute.isGlobal {
+                synchronizeGlobalOptions(of: attribute)
+            } else {
+                populateLocalOptions(of: attribute)
+            }
         }
     }
 }
@@ -265,6 +266,14 @@ private extension AddAttributeOptionsViewModel {
         return Section(header: Localization.headerExistingOptions, footer: nil, rows: rows, allowsReorder: false, allowsSelection: true)
     }
 
+    /// Pre-selects options of a product attribute.
+    ///
+    func preselectCurrentOptions(of attribute: ProductAttribute) {
+        state.selectedOptions = attribute.options.map { option in
+            .local(name: option)
+        }
+    }
+
     /// Synchronizes options for global attributes
     ///
     func synchronizeGlobalOptions(of attribute: ProductAttribute) {
@@ -278,6 +287,8 @@ private extension AddAttributeOptionsViewModel {
         stores.dispatch(fetchOptions)
     }
 
+    /// Populates existing options for a local attribute.
+    ///
     func populateLocalOptions(of attribute: ProductAttribute) {
         state.existingOptions = attribute.options.map { option in
             .local(name: option)
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
@@ -251,7 +251,17 @@ private extension AddAttributeOptionsViewModel {
             AddAttributeOptionsViewModel.Row.selectedOptions(name: option.name)
         }
 
-        return Section(header: Localization.headerSelectedOptions, footer: nil, rows: rows, allowsReorder: true, allowsSelection: false)
+        // Only allow reorder for local attributes(existing and new)
+        let allowsReorder: Bool = {
+            switch attribute {
+            case .new:
+                return true
+            case let .existing(productAttribute):
+                return productAttribute.isLocal
+            }
+        }()
+
+        return Section(header: Localization.headerSelectedOptions, footer: nil, rows: rows, allowsReorder: allowsReorder, allowsSelection: false)
     }
 
     func createExistingSection() -> Section? {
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -265,6 +267,14 @@ private extension AddAttributeOptionsViewModel {
         return Section(header: Localization.headerExistingOptions, footer: nil, rows: rows, allowsReorder: false, allowsSelection: true)
     }
 
+    /// Pre-selects options of a product attribute.
+    ///
+    func preselectCurrentOptions(of attribute: ProductAttribute) {
+        state.selectedOptions = attribute.options.map { option in
+            .local(name: option)
```
jaclync, line 284: does this local option also work for a global attribute since this is also called in case of `attribute.isGlobal` above?

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
```diff
@@ -125,6 +125,21 @@ extension EditAttributesViewController {
         }
         show(addAttributeViewController, sender: self)
     }
+
+    /// Navigates to `AddAttributeOptionsViewController` to provide delete/rename/edit-options functionality.
+    /// Upon completion, update the product and pop the view controller.
+    ///
+    func navigateToEditAttribute(_ attribute: ProductAttribute) {
```
jaclync: nit: could this be `private`?

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
```diff
@@ -17,7 +17,7 @@ final class EditAttributesViewController: UIViewController {
     ///
     var onVariationCreation: ((Product) -> Void)?
 
-    /// Assign this closure to be notified after an attribute  is created.
+    /// Assign this closure to be notified after an attribute  is created or updated.
```
jaclync, line 20: nit: maybe this closure could be called `onAttributeUpdate`? 😄 up to you!

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/EditAttributesViewModelTests.swift
```diff
@@ -58,6 +58,22 @@ final class EditAttributesViewModelTests: XCTestCase {
...
+        // Given
+        let attribute = sampleAttribute(name: "attr", options: ["Option 1", "Option 2"])
+        let attribute2 = sampleAttribute(name: "attr-2", options: ["Option 3", "Option 4"])
+        let product = Product().copy(attributes: [attribute, attribute2])
+
+        // When
+        let viewModel = EditAttributesViewModel(product: product, allowVariationCreation: false)
+
+        // Then
+        XCTAssertTrue(viewModel.attributes.isNotEmpty)
```
jaclync: nit: how about asserting on `viewModel.attributes` is equal to 2?

jaclync (commented):
Looks great! I just have a question about global attributes in `AddAttributeOptionsViewModel`, not sure if it's related to this issue where reordered options aren't saved for a global attribute? (it works for local attributes)


https://user-images.githubusercontent.com/1945542/108799968-cbecbd00-75cc-11eb-8974-8bd7056a85c2.mov

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/EditAttributesViewController.swift
Ecarrion: Totally!

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
Ecarrion: Yes, that works as we only send a string as the option name to the API.

Ecarrion:
> not sure if it's related to this issue where reordered options aren't saved for a global attribute?

Nice catch @jaclync, It seems like the API is not listening to the global attributes options sort order.

**Sent**
body-sent

**Received**
product-recieved


Also, I noticed that the website does not allow us to choose an order for global attributes options.

https://user-images.githubusercontent.com/562080/108854256-ac966580-75b5-11eb-9e7d-e097c0b235a7.mov

----

For that reason, [I've decided to remove](https://github.com/woocommerce/woocommerce-ios/pull/3724/commits/cd6c950dd5ac0de8e902d7743f4c59598b31de21) the sort ability for global attribute options in the app.
allow-reorder

cc @ThomazFB @Garance91540

jaclync (approved):
TIL'ed global attributes don't allow reordering in core as well, and the ordering is always alphabetical! I tested again and it's working as expected (no reordering for global attributes), LGTM 🚀

PR #3725, opened 2021-02-22
[Mobile Payments] Set up the testing strategy for Card Present Payments

### LINKED ISSUE #3675: [Mobile Payments] Add unit test suit for integration with Stripe SDK to the app target

ctarda:
Closes #3675 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by setting up the basis for our testing strategy.

Testing an integration with an SDK that provides an integration with external hardware is turning out to be challenging. For some of the challenges we have encountered so far, and the ways we have found to either overcome or workaround them, please see the markdown file in docs/stripe-tests.md (which is part of the diff of this PR)

But in a nutshell, the strategy is...

## Changes
* Add unit tests to the testing bundle of Hardware to unit tests the model objects that `CardReaderService` exposes. This requires a way to workaround the fact that the Stripe SDK's model objects can not be instantiated by us directly.
* Add a test suite for integration tests to WooCommerce. The readme.md included in the diff explains the rationale behind adding those tests in the app. This suite only provides one test for now, but we want the test to be part of the prototype as early as possible, if only to "mark the spot" where integration tests should live.
* Add a couple of unit tests to Yosemite, to test the only method implemented so far of the `CardPresentPaymentStore`. This two tests also are partially part of this PR to "mark the spot"

## How to test
* Look at the code. 
* Read the rationale behind the testing strategy in `docs/stripe-tests.md` (part of the PR diff)
* Wait for CI to ✅ 
* Ideally, checkout the branch, run `bundle exec pod install` and run all the tests (command + u)

Update release notes:


docs/stripe-tests.md
```diff
@@ -0,0 +1,79 @@
...
+
+### Unit tests
+
+The `CardreaderService` protocol exposes a set of model object also declared in `Hardware`: i.e. `CardReader`, `Charge`, `PaymentIntent`. These objects map to their counterparts declared in the [Stripe Terminal SDK](https://stripe.dev/stripe-terminal-ios/docs/index.html).
+
+The model objects declared in the Stripe Terminal SDK can not be initialized by us directly, because all their initialisers are annotated as NS_UNAVAILABLE. The way we have found to workaround that in a way that we can instantiate our own model objects with mocked Stripe model objects is by making the objects provided by Stripe implement a protocol that we declare, and then typing the parameter passed to our model objects as that protocol.
+
+That translates into something like this:
+
+```
```
jaclync: nit: I think the code could be syntax highlighted with Swift:

```suggestion
```swift
```

WooCommerce/WooCommerceTests/Stripe Integration Tests/StripeCardReaderIntegrationTests.swift
```diff
@@ -0,0 +1,53 @@
...
+@testable import WooCommerce
+@testable import Hardware
+
+
+/// Integration tests for the integration with the Stripe Terminal SDK.
+/// We want to reach into the ServiceLocator in all tests
+/// because these are integration test, and we do not want to mock anything,
+/// at this point, other than the actual hardware
+final class StripeCardReaderIntegrationTests: XCTestCase {
+    private var cancellables: Set<AnyCancellable>!
```
jaclync: nit: how about initializing it with an empty set to avoid the `!` here?

WooCommerce/WooCommerceTests/Stripe Integration Tests/StripeCardReaderIntegrationTests.swift
```diff
@@ -0,0 +1,53 @@
...
+                return
+            }
+
+            // There should be at least one non nil reader
+            guard let _ = readers.first else {
+                XCTFail()
+                return
+            }
+
+            // We blisfully ignore the actual values received (for now)
```
jaclync: super nit:
```suggestion
            // We blissfully ignore the actual values received (for now)
```

Yosemite/YosemiteTests/Stores/CardPresentPaymentStoreTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Mock Card Reader Service: In memory
+    private var mockCardReaderService: MockCardReaderService!
+
+    override func setUp() {
+        super.setUp()
+        dispatcher = Dispatcher()
+        storageManager = MockStorageManager()
+        network = MockNetwork()
+        mockCardReaderService = MockCardReaderService()
```
jaclync, line 38: we might want to reset these in `tearDown`?

Yosemite/YosemiteTests/Stores/CardPresentPaymentStoreTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    /// Verifies that CardPresentPaymentAction.startCardReaderDiscovery hits the `start` method in the service.
+    ///
+    func test_start_discovery_action_hits_start_in_service() {
+        let cardPresentStore = CardPresentPaymentStore(dispatcher: dispatcher,
+                                                       storageManager: storageManager,
+                                                       network: network,
+                                                       cardReaderService: mockCardReaderService)
+
+        //let expectation = self.expectation(description: "Start discovery")
```
jaclync: nit: is this line still needed?

Hardware/Hardware/CardReader/StripeCardReader/Charge+Stripe.swift
```diff
@@ -4,12 +4,29 @@ extension Charge {
...
-        self.description = charge.description
+        self.description = charge.stripeDescription
         self.metadata = charge.metadata
     }
 }
+
+/// The initializers of SCPCharge are annotated as NS_UNAVAILABLE
+/// So we can not create instances of that class in our tests.
+/// A workaround is declaring this protocol, which matches the parts of
+/// SCPCharge that we are interested in, make Crahge implement it,
```
jaclync, line 20: nit:

```suggestion
/// SCPCharge that we are interested in, make Charge implement it,
```

I also like these protocols to surface the properties we care about 👍

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
```diff
@@ -4,13 +4,32 @@ extension PaymentIntent {
...
+/// SCPPaymentIntent that we are interested in, make Reader implement it,
+/// and initialize Harware.CardReader with a type conforming to it.
+protocol StripePaymentIntent {
+    var stripeId: String { get }
+    var created: Date { get }
+    var status: StripeTerminal.PaymentIntentStatus { get }
+    var amount: UInt { get }
+    var currency: String { get }
+    var metadata: [AnyHashable: Any]? { get }
+    var charges: [SCPCharge] { get }
```
jaclync, line 31: nit: would it make sense to use the protocol here as well?
```suggestion
    var charges: [StripeCharge] { get }
```

Hardware/HardwareTests/CardReaderTests.swift
```diff
@@ -0,0 +1,61 @@
+import XCTest
+@testable import Hardware
+
+/// Tests the mapping between CardReader and SCPReader
+final class CardReaderTests: XCTestCase {
+    func test_card_reader_maps_serial_number() {
+        let mockReader = MockStripeCardReader.bbpos()
+        let cardReader = CardReader(reader: mockReader)
+
+        XCTAssertEqual(cardReader.serial, mockReader.serialNumber)
```
jaclync, line 10: nit: would be nice to have the Given/When/Then annotations so that it's easier to follow the scenario, especially for more complicated test cases with async results 🙂 (same suggestion for the other test cases)

```suggestion
        // Given
        let mockReader = MockStripeCardReader.bbpos()
        
        // when
        let cardReader = CardReader(reader: mockReader)

        // Then
        XCTAssertEqual(cardReader.serial, mockReader.serialNumber)
```

Yosemite/YosemiteTests/Mocks/CardPresentPayments/MockCardReaderService.swift
```diff
@@ -0,0 +1,81 @@
...
+    func connect(_ reader: Hardware.CardReader) -> Future<Void, Error> {
+        Future() { promise in
+            // To be implemented
+        }
+    }
+
+    func disconnect(_ reader: Hardware.CardReader) -> Future<Void, Error> {
+        Future() { promise in
+            // This will be removed. We just want to pretend we are doing a roundtrip to the SDK for now.
+            DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
```
jaclync, line 50: just wondering, would this add 2 seconds to each test case that expects the result? if so, maybe a shorter time would help reduce the test running time?

Yosemite/YosemiteTests/Stores/CardPresentPaymentStoreTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+        let action = CardPresentPaymentAction.startCardReaderDiscovery { discoveredReaders in
+            //
```
jaclync, line 61: nit: maybe this line and `discoveredReaders` param aren't necessary?

Hardware/HardwareTests/PaymentIntentTests.swift
```diff
@@ -0,0 +1,60 @@
+import XCTest
+@testable import Hardware
+
+/// Tests the mapping between PaymentIntent and SCPPaymentIntent
+final class PaymentIntentTests: XCTestCase {
+    func test_intent_maps_id() {
+        let mockIntent = MockStripePaymentIntent.mock()
```
jaclync: nit: it seems like each test case uses the same intent, maybe we could have a private property in this class like `private let mockIntent = MockStripePaymentIntent.mock()` for each test case to use?

Hardware/HardwareTests/PaymentIntentTests.swift
```diff
@@ -0,0 +1,60 @@
...
+    func test_intent_maps_charges() {
+        let mockIntent = MockStripePaymentIntent.mock()
+        let intent = PaymentIntent(intent: mockIntent)
+
+        // Very indirect test, that doesn't really test much.
+        // It is not possible to instantiate a SCPCharge, which is what
+        // would be needed to instantiate a mock intent.
+        // For now, we will rely on counting charges as a way
+        // to chek that at least both SCPPaymentIntent and
+        // PaymentIntent reference the same number of charges 🤷
```
jaclync, line 52: would the use of `StripeCharge` protocol help with this case?

WooCommerce/WooCommerceTests/Stripe Integration Tests/StripeCardReaderIntegrationTests.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /// The observed behaviour is that when discovery starts, the Stripe SDK will
+    /// call into its delegate with an empty array of "discovered readers".
+    /// Later on, it will call into its delegate again everytime it discovers a reader
+    /// passing the full list of readers discovered
+    func test_start_discovery_updates_discovered_readers_at_least_twice() {
+        let discoveredReaders = expectation(description: "Discovered Readers publishes first, an empty array, and then the actual reader(s) discovered")
+
+        let readerService = ServiceLocator.cardReaderService
```
jaclync, line 26: I'm guessing this is the actual Stripe card reader service? if so, does it depend on some connection for reader discovery or it's default to run in "test mode" as in the doc?

WooCommerce/WooCommerceTests/Stripe Integration Tests/StripeCardReaderIntegrationTests.swift
```diff
@@ -0,0 +1,53 @@
...
+
+        readerService.discoveredReaders.sink { completion in
+            discoveredReaders.fulfill()
+        } receiveValue: { readers in
+            // The Stripe Terminal SDK published an empty list of discovered readers first
+            // and it will continue publishing as new readers are discovered.
+            // So we ignore the first call to receiveValue, and perform the test on the first call to
+            // receive value that is receiving a non-empty array.
```
jaclync: nit: if we always want to drop the first call, can also use [`dropFirst(1)`](https://developer.apple.com/documentation/combine/fail/dropfirst(_:))?

docs/stripe-tests.md
```diff
@@ -0,0 +1,79 @@
...
+
+The Stripe Terminal SDK provides a "test mode". This test mode allows us to run the terminal SDK in the simulator, without connecting to actual external hardware.
+
+In order to run, no matter if it is in test mode or production mode, the Stripe Terminal SDK asserts that [certain keys are available](https://stripe.com/docs/terminal/sdk/ios#configure) in the Info.plist file. 
+
+That, in turn, means that tests against the Stripe SDK will crash unless those tests are run against an app. So to workaround this, we moved the integration tests to the WooCommerceTests suite. See `StripeCardReaderIntegrationTests`
+
+## Business logic
+
+The business logic of the integration is encapsulated in `CardPresentPaymentStore`. This logic is unit tested in the YosemiteTests, like any other Storein the codebase
```
jaclync: nit:
```suggestion
The business logic of the integration is encapsulated in `CardPresentPaymentStore`. This logic is unit tested in the YosemiteTests, like any other Store in the codebase
```

jaclync (approved):
Thanks for documenting the reasons behind the workarounds and the gotchas testing the Stripe Terminal SDK 💳  I just had some questions out of curiosity and nit suggestions, nothing blocking!

docs/stripe-tests.md
ctarda: TIL, thanks!

Hardware/HardwareTests/PaymentIntentTests.swift
ctarda: Interesting. My first impulse was to reply "I would prefer not to, to avoid sharing state between tests" but the mock is immutable so there is no danger of having one test affecting other... Updating!

Hardware/HardwareTests/PaymentIntentTests.swift
ctarda: Good question, and that's something I tried as well. But the limitation is in the way SCPPaymentIntent is declared. One of its public properties is of type [SCPCharge], so we can not workaround that dependency (or at least I couldn't find a way to workaround it) in a way that we can instantiate an object that could pass as SCPPaymentIntent and has a property of type [Charge] instead of [SCPCharge]. Wow, that's a mouthful! 😛

WooCommerce/WooCommerceTests/Stripe Integration Tests/StripeCardReaderIntegrationTests.swift
ctarda: That's correct. For now, it is running, hardcoded, in "test mode". Whenever we switch that flag and move the app into "production mode", we'd have to update these tests to set a card reader service that still attacks the SDK test mode.

WooCommerce/WooCommerceTests/Stripe Integration Tests/StripeCardReaderIntegrationTests.swift
ctarda: Good call!

ctarda:
Thanks for the thorough review @jaclync 🙇 

I've pushed an update to incorporate your feedback, and I tried to answer your questions the best I could 😊 

I'll wait for CI to give me the ✅  and then merge.

Hardware/HardwareTests/PaymentIntentTests.swift
jaclync: ah I see! the only way I can think of is for `StripePaymentIntent` protocol to have a property of a different name (like `var stripeCharges: [Charge] { get }`) and return `charges` in `extension StripeTerminal.PaymentIntent: StripePaymentIntent` - not sure how much value this abstraction helps though!

PR #3727, opened 2021-02-23
Update secrets

mokagio:
Done via `bundle exec fastlane run cofigure_update`.

Internal ref paaHJt-1Qj-p2.

Update release notes:


loremattei (approved):
LGTM!
GitHub says that the encoded files are empty, but I checked out the branch locally and they looks ok.

AliSoftware:
Strange empty diff of env files indeed, but checked them too too like Lore using xxd and they look good, so merging.

Thanks @mokagio 🙌

Issue #3193, opened 2020-11-23
Variations: Allow Deletion of a Single Variation


shiki:
Refs p91TBi-3Oc-p2, p91TBi-2Ur-p2.

Allow deletion of a product variation.

<img src="https://user-images.githubusercontent.com/198826/100024490-1f6c2680-2da4-11eb-9633-ffb536b17370.png" width="500">

## Goals

- [x] Show the more options icon on the right side. 
- [x] Show a Delete option. 
- [x] Tapping on Delete will delete the variation. XD 
- [x] If there are no variations left, navigate the user to the empty list state screen https://github.com/woocommerce/woocommerce-ios/issues/3195
- [x] If there are still variations left after the deletion, navigate the user to the Variations list. 

---

Android: https://github.com/woocommerce/woocommerce-android/issues/3212

PR #3728, opened 2021-02-23
Variations: single variation deletion - Networking, Storage and Yosemite

ealeksandrov:
Part of https://github.com/woocommerce/woocommerce-ios/issues/3193.

## Description

This PR adds Networking, Storage and Yosemite integrations to allow deleting a single variation.

Related API doc: https://woocommerce.github.io/woocommerce-rest-api-docs/#delete-a-product-variation

## Changes

1. Adds `deleteProductVariation` case to `ProductVariationAction`
2. Adds handling for new action in `ProductVariationStore` (+ tests)
3. Adds local storage function to remove a single variation
4. Adds networking for deleting a single variation (+ tests)

## Test

Nothing is exposed to the UI here, so just check the code and unit tests status.
(UI is added in https://github.com/woocommerce/woocommerce-ios/pull/3730, feel free to test actions from there.)

Update release notes:

### DIFF
--- a/Storage/Storage/Tools/StorageType+Deletions.swift
+++ b/Storage/Storage/Tools/StorageType+Deletions.swift
@@ -33,8 +33,7 @@ public extension StorageType {
     /// Deletes single stored Product Variation for the provided siteID and productVariationID.
     ///
     func deleteProductVariation(siteID: Int64, productVariationID: Int64) {
-        guard let productVariation = loadProductVariation(siteID: siteID,
-                                                          productVariationID: productVariationID) else {
+        guard let productVariation = loadProductVariation(siteID: siteID, productVariationID: productVariationID) else {
             return
         }
 
--- a/Storage/Storage/Tools/StorageType+Deletions.swift
+++ b/Storage/Storage/Tools/StorageType+Deletions.swift
@@ -30,6 +30,17 @@ public extension StorageType {
         }
     }
 
+    /// Deletes single stored Product Variation for the provided siteID and productVariationID.
+    ///
+    func deleteProductVariation(siteID: Int64, productVariationID: Int64) {
+        guard let productVariation = loadProductVariation(siteID: siteID,
+                                                          productVariationID: productVariationID) else {
+            return
+        }
+
+        deleteObject(productVariation)
+    }
+
     /// Deletes all of the stored Product Shipping Class models for the provided siteID.
     ///
     func deleteProductShippingClasses(siteID: Int64) {
--- a/Yosemite/Yosemite/Actions/ProductVariationAction.swift
+++ b/Yosemite/Yosemite/Actions/ProductVariationAction.swift
@@ -28,4 +28,8 @@ public enum ProductVariationAction: Action {
     /// Requests the variations in a specified Order that have not been fetched.
     ///
     case requestMissingVariations(for: Order, onCompletion: (Error?) -> Void)
+
+    /// Delete an existing ProductVariation.
+    ///
+    case deleteProductVariation(siteID: Int64, productID: Int64, variationID: Int64, onCompletion: (Result<ProductVariation, ProductUpdateError>) -> Void)
 }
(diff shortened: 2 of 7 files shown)


Storage/Storage/Tools/StorageType+Deletions.swift
```diff
@@ -30,6 +30,17 @@ public extension StorageType {
         }
     }
 
+    /// Deletes single stored Product Variation for the provided siteID and productVariationID.
+    ///
+    func deleteProductVariation(siteID: Int64, productVariationID: Int64) {
+        guard let productVariation = loadProductVariation(siteID: siteID,
+                                                          productVariationID: productVariationID) else {
```
Ecarrion: super nit: I think this could be in the same line!

Yosemite/Yosemite/Actions/ProductVariationAction.swift
```diff
@@ -28,4 +28,8 @@ public enum ProductVariationAction: Action {
     /// Requests the variations in a specified Order that have not been fetched.
     ///
     case requestMissingVariations(for: Order, onCompletion: (Error?) -> Void)
+
+    /// Delete an existing ProductVariation.
+    ///
+    case deleteProductVariation(productVariation: ProductVariation, onCompletion: (Result<ProductVariation, ProductUpdateError>) -> Void)
```
Ecarrion: Do you think it's better to have the result as `Result<Error, ProductUpdateError>`? as the deleted variation won't have any usage.

Yosemite/Yosemite/Stores/ProductVariationStore.swift
```diff
@@ -202,6 +204,25 @@ private extension ProductVariationStore {
...
+
+    /// Deletes the product variation.
+    ///
+    func deleteProductVariation(productVariation: ProductVariation, onCompletion: @escaping (Result<ProductVariation, ProductUpdateError>) -> Void) {
+        remote.deleteProductVariation(siteID: productVariation.siteID,
+                                      productID: productVariation.productID,
+                                      variationID: productVariation.productVariationID) { [weak self] result in
+            guard let self = self else {
+                return
+            }
```
Ecarrion: nit: we are writing these as one-liners lately.
Ref: https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/coding-style-guide.md#braces

Yosemite/Yosemite/Stores/ProductVariationStore.swift
```diff
@@ -202,6 +204,25 @@ private extension ProductVariationStore {
...
+            switch result {
+            case .failure(let error):
+                onCompletion(.failure(ProductUpdateError(error: error)))
+            case .success(let productVariation):
+                self.deleteStoredProductVariation(siteID: productVariation.siteID, productVariationID: productVariation.productVariationID)
+                onCompletion(.success(productVariation))
```
Ecarrion: nit: I think having the success case first and the failure case later it's a bit more aligned with the rest of the project!

Ecarrion (approved):
Looks great! 
Just some minor suggestions and tips!

Yosemite/Yosemite/Actions/ProductVariationAction.swift
ealeksandrov: That's actually a deeper question - why does API return deleted product and product variation? Since I don't know the answer - preferred to pass it as is 🙂 Also to keep consistency with existing "delete product" action.

In the end after I have everything connected in UI - I agree that there is no need to pass around value here.

I think using `Error` as a type for `Success` breaks logic of [Result type](https://developer.apple.com/documentation/swift/result). And it still needs some success value.

But I see `Result<Void, ProductUpdateError>` as a good fit, maybe you even meant this one 🙂

ealeksandrov:
Thank you for review @Ecarrion! I've pushed minor change for Result type, please take a quick look to confirm.

Yosemite/Yosemite/Actions/ProductVariationAction.swift
Ecarrion: >  why does API return deleted product and product variation?

I think that's just for some REST conventions.

> But I see Result<Void, ProductUpdateError> as a good fit, maybe you even meant this one

haha Yea, I totally meant that one, made a mistake while typing :-)

Ecarrion (approved):
🚀

PR #3730, opened 2021-02-23
Variations: single variation deletion - UI

ealeksandrov:
Part of https://github.com/woocommerce/woocommerce-ios/issues/3193.
Based on https://github.com/woocommerce/woocommerce-ios/pull/3728.

## Description

This PR adds UI integration for delete action added in https://github.com/woocommerce/woocommerce-ios/pull/3728.

## Changes

1. Enables delete action by updating `canDeleteProduct` in `ProductVariationFormViewModel`
2. Adds `ProductVariationAction.deleteProductVariation` call in `ProductVariationFormViewModel` (+ tests)
3. Moves confirmation alerts and progress views into helper extension on `ProductFormViewController` because variations deletion need different text - products are moved to trash, but variations are deleted ([reference in API doc](https://woocommerce.github.io/woocommerce-rest-api-docs/#delete-a-product-variation)).

## Test

1. Open details for product with existing variations
2. Open variations list
3. Tap on variation
4. Tap "•••" in navbar, pick "Delete"
5. Observe that variation in successfully removed

### Known limitations

Empty state is missing after removal of all variations, also parent product details is not updated. Will be fixed in separate PR.

## Video

https://user-images.githubusercontent.com/3132438/108848960-b12d3f00-75f2-11eb-9f5b-dd37c6e82c45.mp4

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController+Helpers.swift
```diff
@@ -67,4 +67,80 @@ extension ProductFormViewController {
...
+    ///
+    func presentVariationConfirmationDeleteAlert(completion: @escaping (_ isConfirmed: Bool) -> ()) {
+        let title = NSLocalizedString("Remove variation",
+                                      comment: "Title of the alert when a user is deleting a variation")
+        let body = NSLocalizedString("Are you sure you want to remove this variation?",
+                                     comment: "Body of the alert when a user is deleting a variation")
+        let cancelButton = NSLocalizedString("Cancel",
+                                             comment: "Cancel button on the alert when the user is cancelling the action on deleting a variation")
+        let confirmButton = NSLocalizedString("Remove",
+                                              comment: "Confirmation button on the alert when the user is deleting a variation")
```
Ecarrion: Do you think we can move these strings(and all the others) to the `Localization` enum as we do on other view controllers?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
```diff
@@ -265,9 +265,24 @@ extension ProductVariationFormViewModel {
         storesManager.dispatch(updateAction)
     }
 
-    func deleteProductRemotely(onCompletion: @escaping (Result<EditableProductModel, ProductUpdateError>) -> Void) {
-        // no-op
+    func deleteProductRemotely(onCompletion: @escaping (Result<EditableProductVariationModel, ProductUpdateError>) -> Void) {
+        let deleteAction = ProductVariationAction.deleteProductVariation(productVariation: productVariation.productVariation) { [weak self] result in
+            guard let self = self else {
+                return
+            }
```
Ecarrion: Nit: 
```suggestion
            guard let self = self else { return }
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -685,35 +665,47 @@ private extension ProductFormViewController {
...
+                self?.deleteProduct()
+            }
+        }
+    }
+
+    func deleteProduct() {
+        self.viewModel.deleteProductRemotely { [weak self] result in
+            guard let self = self else {
+                return
+            }
```
Ecarrion: nit
```suggestion
            guard let self = self else { return }
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -685,35 +665,47 @@ private extension ProductFormViewController {
...
+            switch result {
+            case .failure(let error):
+                DDLogError("⛔️ Error deleting Product: \(error)")
+
+                // Dismisses the in-progress UI then presents the error alert.
+                self.navigationController?.dismiss(animated: true) { [weak self] in
+                    self?.displayError(error: error)
                 }
+            case .success:
```
Ecarrion: nit:
Switch `failure` and `success` order? 😄

Ecarrion (approved):
Looks good, just a few minor nitpicks!

Regarding:
> Empty state is missing after removal of all variations, also parent product details is not updated. Will be fixed in separate PR.

I would have either add everything on this PR or -as you were asking before- add this behind a feature toggle. The reason being that if this gets released, functionality will appear broken to the merchants.

If you plan to submit the next PR soon enough it shouldn't be a problem tho.

ealeksandrov:
Thanks for the review @Ecarrion!

> add this behind a feature toggle

I've decided to do exactly this. Hopefully it'll be fixed soon, so we can remove the flag.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController+Helpers.swift
ealeksandrov: I felt that this extension is already feels like a list of strings, but extraction allowed to simplify a logic a bit and it looks cleaner now (dfbc904) 🎉

PR #3732, opened 2021-02-24
Remove Sentry logging for the unexpected app state

jaclync:
Cleanup for #3711 

## Why

In a previous PR https://github.com/woocommerce/woocommerce-ios/pull/3713 as an attempted fix for crash #3711, I added an info-level logging for the unexpected app state. Now that we know from the Sentry log that it's possible to reach this path, we can remove this logging for the next release. I'm still following up possible leads (like if I see anyone on the team in the logs) to understand how this path might be reached.

## Changes

Removed logging for the unexpected app state.

## Testing

Please confidence check on logging out and in:

- Log out from the app
- Log in to a store --> the whole flow should work like before

Update release notes:


pmusolino (approved):
LGTM :shipit:

Issue #3734, opened 2021-02-24
[Mobile Payments] Error handling. Make service propagate errors


ctarda:
Implement proper error handling

For now, to continue moving forward, we have just thrown a couple of errors typed as CardReaderServiceError. 

We need to populate that error object properly, in a way that also provides an indication of what went wrong and where.

Issue #3715, opened 2021-02-22
[Mobile Payments] Pair with a single reader. Mark II: Yosemite


ctarda:
After #3714 is landed into the feature branch, implement the next layer: Yosemite.

Scope:
* Add enough code to CardPresentPaymentsAction and CardPresentPaymentsStore to pair a single reader.
* Add an integration test
* This would require implementing a mechanism to transform model objects declared in Yosemite (i.e. Yosemite.CardReader) to their counterparts in Hardware (i.e. Hardware.CardReader). At this point that's pretty much a 1:1 relationship but it will not be like that for too long.

PR #3735, opened 2021-02-24
[Mobile Payments] Pair a reader, Mark I: CardReaderService

### LINKED ISSUE #3714: [Mobile Payments] Pair with one single reader. Mark I: service
Now that we have the skeleton of the logic to discover readers, it is time to start pairing with a single reader.

For this ticket, we look only at the Hardware Layer

Scope:
* Implement enough code in StripeCardReaderService to pair with a given instance of Networking.CardReader
* That pairing would happen against the test model of the Stripe Terminal SDK, to facilitate review
* Add at least one integration test

ctarda:
Closes #3714 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by adding enough code to the CardReaderService to attack the Stripe Terminal SDK (in test mode) and request a connection to a given reader.

Testing this is very tricky. There are indications for settings up and running the integration tests (that cover this use case) in the diff of the README.md that accompanies the PR. But, to be honest, the easiest way to test this will be when #3715  is implemented, which is something I plan on doing right away.

## Changes
* Add enough code to StripeCardReaderService to pair with a card reader. Because we do not want to expose the model objects provided by the Stripe Terminal SDK to the rest of the codebase, we cache the discovered readers, of type `StripeTerminal.Reader`, until the connection process is complete. [The documentation suggests this is possible](https://stripe.dev/stripe-terminal-ios/docs/Protocols/SCPDiscoveryDelegate.html#/c:objc(pl)SCPDiscoveryDelegate(im)terminal:didUpdateDiscoveredReaders:):
> You should not cache the SCPReader objects in the readers array. In other words, after calling connectReader: with a discovered reader, you should not attempt to cache and re-use the Reader object

* Added a skeleton of an enumeration to model errors. This will be fully implemented in #3734 
* An integration test to trigger discovery, then trigger a connection to the first discovered reader, and then observes if the `connectedReader` property of CardReaderTerminal is updated. Instructions for setting up to run this test are available in the readme file that is part of the diff. If the test is setup successfully, the simulator will ask for permission to access Bluetooth. There is no way to recover, at this point, if the request is not granted. We'll get there though...

## How to test
This one is tricky. There are instructions in the diff to setup a testing site and obtain a connection token to run the integration test. 

* Checkout the branch
* run `bundle exec pod install`
* Activate StripeCardReaderIntegrationTests
* Run StripeCardReaderIntegrationTests (see instructions in stripe-tests.md)

Update release notes:


Hardware/Hardware/CardReader/CardReaderService.swift
```diff
@@ -28,6 +28,9 @@ public protocol CardReaderService {
     /// That could imply, for example, that the reader discovery process starts
     func start()
 
+    /// Cancels the discovery process.
+    func cancelDiscovery()
```
ctarda, line 32: Needed if we want to run more than one integration test. If the Stripe Terminal SDK is in discovery mode, and we start a new test, the second test will crash.

allendav:
> we cache the discovered readers, of type StripeTerminal.Reader, until the connection process is complete

This should be safe. Stripe does the same in their example: https://github.com/stripe-samples/sample-terminal-ios-app/blob/9ab501aa6fc990f194650d90551872e33015e321/rocketrides-pilot/ReaderConnectionViewController.swift#L107

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderCache.swift
```diff
@@ -0,0 +1,28 @@
+import StripeTerminal
+
+/// In memory, volatile cache
+final class StripeCardReaderCache {
```
allendav: Maybe name it StripeCardReaderDiscoveryCache to emphasize its ephemeral-ness?

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -73,13 +78,19 @@ extension StripeCardReaderService: CardReaderService {
         // Enough code to pass a test
         discoveryCancellable = Terminal.shared.discoverReaders(config, delegate: self, completion: { error in
             if let error = error {
-                print("discoverReaders failed: \(error)")
+                print("***** discoverReaders failed: \(error)")
```
allendav: Glad I'm not the only person that does things like this :)

Hardware/Hardware/CardReader/StripeCardReader/WCPayTokenProvider.swift
```diff
@@ -5,7 +5,7 @@ import StripeTerminal
 /// Tokens can be "live" or "test", depending on wheter WCPay
 /// is set for live or test mode.
 public final class WCPayTokenProvider: ConnectionTokenProvider {
-    private let mockToken = "mock_token"
+    private let mockToken = "get a token from your test site"
```
allendav: Maybe `get a one-time use token...`

docs/stripe-tests.md
```diff
@@ -74,6 +74,20 @@ In order to run, no matter if it is in test mode or production mode, the Stripe
 
 That, in turn, means that tests against the Stripe SDK will crash unless those tests are run against an app. So to workaround this, we moved the integration tests to the WooCommerceTests suite. See `StripeCardReaderIntegrationTests`
 
+Integration tests that go just beyond discovering readers are not easy to run (at this stage). The reason is that the Terminal SDK needs a valid connection token when it attempts to connect to a reader, even if the SDK is in test mode.
+
+There are some prerequistes for running some of the integration tests:
+
+* Install WooCommerce Payments on a WooCommerce site. Do not run the setup assisntant yet.
```
allendav: assistant

docs/stripe-tests.md
```diff
@@ -74,6 +74,20 @@ In order to run, no matter if it is in test mode or production mode, the Stripe
...
+* Install [the dev plugin](https://github.com/Automattic/woocommerce-payments-dev-tools) by downloading the repo as a zip and uploading it to the site as a new plugin, and activate it.
```
allendav: This (woocommerce-ios) repo is public. Should we refer to a private repo? As an alternative, the end user could add this to their wp-config.php or somewhere.

```php
define( 'WCPAY_DEV_MODE', true );
```

docs/stripe-tests.md
```diff
@@ -74,6 +74,20 @@ In order to run, no matter if it is in test mode or production mode, the Stripe
...
+* After activating the Dev Tools plugin, look in wp-admin all the way at the bottom of the sidebar menu for WCPay Dev.
+* Uncheck Proxy WPCOM requests
+* Run the WooCommerce Payments setup assistant. It should indicate that it is running in "TEST DATA" mode. That would allow entering mock data.
+
+At the time of writing this, we do not have the infrastructure to obtain that connection token by performing an authenicated request to a WooCommerce store. So integration tests that requir a connection token will require to edit WCPayTokenProvider and update the mock token with a value obtainied from your store's wp-admin. In order to do that you would need to download and install [this plugin](https://github.com/Automattic/woocommerce-payments-card-reader-token-helper).
```
allendav: I can make this repo public if needed

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
ctarda: rofl. It me. All the time.

docs/stripe-tests.md
ctarda: I think for now, until we extract networking and implement the request to fetch the token properly, it should be fine? I'll ask

allendav (approved):
Screen Shot 2021-02-24 at 12 55 16 PM


:shipit: after CI completes

Issue #3629, opened 2021-02-10
Variations: non-blocking navigation on attribute creation process


ealeksandrov:
Consider the scenario where, while creating an attribute, the user can tap the back button as the screen is not blocked.
App works well and nothing really breaks.

https://user-images.githubusercontent.com/3132438/107529519-00737880-6bcc-11eb-96dc-84ce588298e7.mp4

Should we cancel the network operation? Or forward navigation? Or maybe this is too far of edge case?

Ecarrion:
What do you think @Garance91540 @adamzelinski?

One option I can think of is to show the same loading screen that covers the whole screen, like when we are updating a product.

Garance91540:
> One option I can think of is to show the same loading screen that covers the whole screen, like when we are updating a product.

Yes maybe that's best, so it's consistent with our other loading states and we don't introduce a whole new interaction

Ecarrion:
@Garance91540 any specific wording?

@ealeksandrov Would you mind creating a separate issue(and close this one) with the screens, texts, and whatever else is needed for someone to work on this?

Garance91540:
Designs here:
https://d.pr/i/XUpJVv

Ecarrion:
@Garance91540 while playing the app I noticed that the tags screen works in the same way and I think people have not found this to be a problem.

https://user-images.githubusercontent.com/562080/107774407-f311d780-6d0c-11eb-9af4-164fd3301d89.MP4

Should we still proceed with the change?

Garance91540:
Hi @Ecarrion, I tested the tag flow before and it didn't confuse me to be able to go back while my changes were being saved. It's probably also less of a friction since we are not disrupting the user flow with a full screen loading screen. So...let's keep things as such 😂 (loader in the top nav)

Ecarrion:
Thanks, @Garance91540 for your time!
@ealeksandrov let's close this for the time being. We'll re-assess if merchants start complaining!

PR #3736, opened 2021-02-24
Variations: Remove product attributes

### LINKED ISSUE #3212: Variations: Renaming an Attribute
Ref p91TBi-3Oc-p2.

From #3183, users would be presented with the options management screen (#3182). On that screen, the user should be able to rename the attribute. 

## Goals 

- [x] Add a Rename button in the options menu.
- [x] Build and Present the Rename Attribute screen.
- [x] Tapping on Done will rename the attribute. 
- [x] Add Tracking https://github.com/woocommerce/woocommerce-ios/issues/3703

---

Android: https://github.com/woocommerce/woocommerce-android/issues/3214

Ecarrion:
closes #3212

# Why

This PR allows the merchant to remove a product attribute, removing an attribute affects the product's variations in **Core**, so on each attribute update, we need to re-fetch the variations.

Also, when removing the last attribute, we need to show the user the empty state screen to allow them to trigger the first variation flow.

# How

- In `AddAttributeOptionsViewModel`:
  - Added `showMoreButton` property for VC to consume
  - Added `removeCurrentAttribute` that removes the attribute remotely.

- In `AddAttributeOptionsViewController`: 
  - Add "More" button, together with the existing "Next" Button
  - Present action sheet with remove option when tapping the "More" button
  - Present confirmation alert when selecting the remove action
  - Call `ViewModel.removeCurrentAttribute` and notify the parent controller 

- Update `EditAttributesViewController` to provide a `AddAttributeOptionsViewModel` with edit mode.

- Update `ProductVariationsViewModel` to:
  - Add an `shouldShowEmptyState()` method
  - Remove product dependency: Before this PR, there were two sources for the product(VM and VC) that can go out of sync. We need to stick with one(The VC) until the VC is refactored to use `MVVM`

- Update `ProductVariationsViewController` to: 
  - Configure its right button item based on the `VM` 
  - Display or hide the empty state screen when the product attributes are updated
  - Re-fetch variations when the product attributes are updated
  - Make sure the user is on the right screen after deleting the last variation. (EditAttributes after deleting/updating an attribute. VariationsList after deleting the last attribute)
  
  # Demo
  
Delete Any | Delete Last
--- | --- 
remove-attribute | remove-last-attribute

# Testing Steps
- Go to a product with variations and at least two attributes.
- Navigate to the variations list
- Tap on the "more" button and go to edit attributes
- Tap on an attribute
- Tap on the "more" button and see that an action sheet with a "remove" option appears.
- Tap on the "remove" action and see that an alert confirmation appears.
- Tap on the "remove" confirmation and see that you are brought back to the edit attribute screen and that the attribute was removed.
**Repeat process with all other attributes**
- Once the last attribute is removed, see that you are taken back to the variation list screen with the empty state on top of it.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -390,7 +390,7 @@ extension AddAttributeOptionsViewController {
             case let .success(product):
                 self.onCompletion(product)
             case let .failure(error):
-                // TODO: Show error notice
+                self.noticePresenter.enqueue(notice: .init(title: Localization.removeAttributeError, feedbackType: .error))
                 DDLogError(error.localizedDescription)
             }
         }
@@ -434,6 +434,8 @@ private extension AddAttributeOptionsViewController {
                                                             comment: "Placeholder of cell presenting the title of the new attribute option.")
         static let updateAttributeError = NSLocalizedString("The attribute couldn't be saved.",
                                                             comment: "Error title when trying to update or create an attribute remotely.")
+        static let removeAttributeError = NSLocalizedString("The attribute couldn't be removed.",
+                                                            comment: "Error title when trying to remove an attribute remotely.")
 
         static let removeAction = NSLocalizedString("Remove", comment: "Title for removing an attribute in the edit attribute action sheet.")
         static let cancelAction = NSLocalizedString("Cancel", comment: "Title for canceling the edit attribute action sheet.")
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -84,7 +84,7 @@ private extension AddAttributeOptionsViewController {
     }
 
     func createNextButton(enabled: Bool) -> UIBarButtonItem {
-        let button = UIBarButtonItem(title: Localization.nextNavBarButton, style: .plain, target: self, action: #selector(nextButtonPressed(_:)))
+        let button = UIBarButtonItem(title: Localization.nextNavBarButton, style: .plain, target: self, action: #selector(moreButtonPressed(_:)))
         button.isEnabled = enabled
         return button
     }
@@ -346,7 +346,21 @@ extension AddAttributeOptionsViewController {
     }
 
     @objc private func moreButtonPressed(_ sender: UIBarButtonItem) {
-        // TODO: Show Edit action sheet
+        let actionSheet = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
+        actionSheet.view.tintColor = .text
+
+        let removeAction = UIAlertAction(title: Localization.removeAction, style: .default) { [weak self] _ in
+            // TODO: Remove
+        }
+        actionSheet.addAction(removeAction)
+
+        let cancelAction = UIAlertAction(title: Localization.cancelAction, style: .cancel)
+        actionSheet.addAction(cancelAction)
+
+        let popoverController = actionSheet.popoverPresentationController
+        popoverController?.barButtonItem = sender
+
+        present(actionSheet, animated: true)
     }
 }
 
@@ -387,5 +401,8 @@ private extension AddAttributeOptionsViewController {
                                                             comment: "Placeholder of cell presenting the title of the new attribute option.")
         static let updateAttributeError = NSLocalizedString("The attribute couldn't be saved.",
                                                             comment: "Error title when trying to update or create an attribute remotely.")
+
+        static let removeAction = NSLocalizedString("Remove", comment: "Title for removing an attribute in the edit attribute action sheet.")
+        static let cancelAction = NSLocalizedString("Cancel", comment: "Title for canceling the edit attribute action sheet.")
     }
 }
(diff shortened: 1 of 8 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -333,6 +344,57 @@ extension AddAttributeOptionsViewController {
...
+
+        let popoverController = actionSheet.popoverPresentationController
+        popoverController?.barButtonItem = sender
+
+        present(actionSheet, animated: true)
+    }
+
+    /// Presents a confirmation alert and removes the attribute if the merchant confirms it.
+    ///
+    @objc private func presentRemoveAttributeConfirmation() {
```
jaclync: nit: `@objc` is probably not necessary since it's called from a closure?
```suggestion
    private func presentRemoveAttributeConfirmation() {
```

WooCommerce/WooCommerceTests/ViewRelated/Products/Variations/ProductVariationsViewModelTests.swift
```diff
@@ -4,40 +4,104 @@ import Yosemite
...
+        // When
+        let showMoreButton = viewModel.showMoreButton(for: product)
 
-        // Assert
-        XCTAssertEqual(viewModel.showMoreButton, true)
+        // Then
+        XCTAssertTrue(showMoreButton)
+    }
+
+    func test_more_button_does_not_appear_when_product_has_variations_does_not_have_attributes_and_addProductVariations_feature_is_enabled() {
```
jaclync, line 20: nice coverage on these edge cases!

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewModel.swift
```diff
@@ -17,22 +13,30 @@ final class ProductVariationsViewModel {
...
+extension ProductVariationsViewModel {
+    /// Defines the empty state screen visibility
+    ///
+    func shouldShowEmptyState(for product: Product) -> Bool {
+        product.variations.isEmpty || product.attributes.isEmpty
+    }
+
+    /// Defines if the More Options button should be shown
+    ///
+    func showMoreButton(for product: Product) -> Bool {
```
jaclync: nit: would it be more consistent with naming the "should show" functions with `shouldShowMoreButton(for:)`?

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -92,6 +92,8 @@ final class ProductVariationsViewController: UIViewController {
 
     private var product: Product {
         didSet {
+            configureRightButtonItem()
+            configureEmptyState()
```
jaclync: nit: usually we use `configure*` for one-time setup, and `update*` for setup that could change so that it's easier to tell if a setup stays the same 🙂

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -471,6 +486,18 @@ private extension ProductVariationsViewController {
...
+
+    /// Refreshes the product variations list and navigates to the appropriate screen.
+    /// Navigates back to edit attribute screen if the product has attributes.
+    /// Navigates back to variations list if the product doesn't has attributes.
+    ///
+    private func onAttributesUpdate(editAttributesViewController: UIViewController) {
+        // Refresh variations because updating an attribute updates the product variations.
+        syncingCoordinator.synchronizeFirstPage()
+
+        let viewControllerToShow = product.attributes.isNotEmpty ? editAttributesViewController : self
```
jaclync, line 498: this is considerate, saving users one tap!

jaclync (approved):
Working nicely! :shipit: 

Some minor UI/UX observations:

- The "Remove" action in the action sheet could be destructive style, like in the confirmation alert?


- After confirming to remove an attribute, I can still navigate back to the previous screen to "cancel" this deletion - it's probably fine? 🤔 

https://user-images.githubusercontent.com/1945542/109098277-fae56900-775b-11eb-859d-d64ccf20c061.mov

- After removing one or more attributes, the product could end up having **many** duplicate variations. Do you think we could show some "duplication" UI - like some warning text in the variation rows, or a section of duplicated variations? cc @woocommerce/mobile-designers

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
Ecarrion: Copy-paste mistake 😇

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
Ecarrion: Nice, I wasn't aware of that convention!

Ecarrion:
Thanks for the review @jaclync 

----
> The "Remove" action in the action sheet could be destructive style, like in the confirmation alert?

The designs didn't have that style but I agree it looks better!

destructive-action

----

> After confirming to remove an attribute, I can still navigate back to the previous screen to "cancel" this deletion - it's probably fine? 🤔

We discussed [that problem](https://github.com/woocommerce/woocommerce-ios/issues/3629#issuecomment-779099393), and decided it's not a big problem, for now!

----

> After removing one or more attributes, the product could end up having many duplicate variations. Do you think we could show some "duplication" UI - like some warning text in the variation rows, or a section of duplicated variations? 

Let me think a little bit about it [outside of this PR](https://github.com/woocommerce/woocommerce-ios/projects/29#card-55679574)!

Issue #3703, opened 2021-02-19
Variations: Tracks


Ecarrion:
The following events should be logged in the app.

### Variations

- [x] add_first_variation_button_tapped
params: product_id

- [x] add_more_variations_button_tapped
params: product_id

- [x] create_product_variation
params: product_id

- [x] create_product_variation_success
params: product_id, time

- [x] create_product_variation_fail 
params: product_id, error_description, time

- [x] create_product_variation_button_tapped
params: product_id, product_variation_id

### Product Attributes

- [x] edit_product_attributes_button_tapped
params: product_id

- [x] add_product_attribute_button_tapped
params: product_id

- [x] update_product_attribute
params: product_id

- [x] update_product_attribute_success
params: product_id, time

- [x] update_product_attribute_fail
params: product_id, error_description, time

- [x] rename_product_attribute_button_tapped
params: product_id

- [x] remove_product_attribute_button_tapped
params: product_id

### Product Variation Attributes

- [x] edit_product_variation_attribute_options_row_tapped
params: product_id, product_variation_id

- [x] edit_product_variation_attribute_options_done_button_tapped
params: product_id, product_variation_id

Ecarrion:
@rachelmcr could you please close this issue when the work on https://github.com/woocommerce/woocommerce-ios/issues/3212 is complete?

Thanks!

PR #3747, opened 2021-02-25
Variations: Track events

Ecarrion:
Part of #3703 

# Why

Since the variations feature is almost(few tasks missing) done, now it's time to add tracking to events defined in #3703.

# How 
- Define event ids in `WooAnalyticStats`
- Define events method in `WooAnalyticsEvent`
- Update `ViewControllers` and `ViewModels` to add the tracking calls.

Note: In `ProductFormViewController` I used `ServiceLocator.analytics` instead of injecting it as a dependency, because the rest of the view controller already uses `ServiceLocator.analytics` and I think it's best to keep consistency and refactor them in one go eventually.

# Screenshot
Screen Shot 2021-02-25 at 2 54 27 PM

# Testing Steps

Go through the add and edit variation flow and see the events being tracked in the console for the following actions:
- Tap Add Variation button when there aren't variations
- Create one or more attributes
- Create attribute success or fail
- Create variation
- Create variation success or fail
- Create more variations button tapped
- Edit attributes button tapped
- Add attribute button tapped
- Delete Attribute button tapped
- Update variations attribute options row tapped
- Update variations attribute options done button tapped
- Remove product variation button tapped

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -375,7 +379,9 @@ extension AddAttributeOptionsViewController {
 
         alertController.addCancelActionWithTitle(Localization.cancelAction)
         alertController.addDestructiveActionWithTitle(Localization.removeAction) { [weak self] _ in
-            self?.removeCurrentAttribute()
+            guard let self = self else { return }
+            self.analytics.track(event: WooAnalyticsEvent.Variations.removeAttributeButtonTapped(productID: self.viewModel.product.productID))
+            self.removeCurrentAttribute()
```
rachelmcr, line 384: I noticed that when we remove an attribute, it fires `remove_product_attribute_button_tapped` but also these events:

* `create_product_attribute`
* `create_product_attribute_success`

Is that expected? I see it's because `removeCurrentAttribute()` calls `performProductUpdate()` (which fires the `create_*` events). But in this case, the product update isn't creating anything.

If we want to track the update (and the update success/fail) when the attribute is removed, would it make sense to use a more general name (e.g. `update_product_attribute` instead of `create_product_attribute`)? If we want to keep track of what kind of update is done, we could pass params like `create`/`remove` depending on what the update is for.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -672,6 +673,7 @@ private extension ProductFormViewController {
                     return
                 }
 
+                self?.trackVariationRemoveButtonTapped()
                 self?.showVariationDeletionProgress()
                 self?.deleteProduct()
```
rachelmcr, line 678: Something I noticed while deleting variations:

The tracks event for `trackVariationRemoveButtonTapped()` above works, but then when we call `deleteProduct()` here it fires `product_detail_product_deleted`. That is an existing tracks event, and I think we try not to change tracks events that are in use (right?), but is that sufficiently clear when we're deleting a variation versus a product?

rachelmcr (approved):
Code looks good, and I confirmed all the expected tracks events are fired with the expected parameters.

I had a couple questions about event names but the events themselves are all working. 👍

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
Ecarrion: That's probably a good idea. I'll make that change!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
Ecarrion: > That is an existing tracks event, and I think we try not to change tracks events that are in use (right?)

That's right!

> but is that sufficiently clear when we're deleting a variation versus a product?

When they added that metric I don't think they care about what kind of product was being deleted. So I think we are safe here, as our event is enough for us!

Issue #3751, opened 2021-02-26
Large titles in the main tabs


jaclync:
Design ref: p91TBi-3AK-p2
Spike: p91TBi-45c-p2

The goal is to enable large titles on the four main tabs: dashboard, orders, products, and reviews. Only the first view controller has large title, while the rest do not by default. This paves the way for future iterations on the "dashboard" tab replacement.

- [x] Enable large titles behind a feature flag, with scrolling animation available only in products and reviews tab (since dashboard and orders tabs have multiple scroll views and a workaround is required) https://github.com/woocommerce/woocommerce-ios/pull/3752
- [x] Hidden scroll view workaround for dashboard and orders tab https://github.com/woocommerce/woocommerce-ios/pull/3763
- [x] Turn on feature flag for beta release when there are no issues found during development https://github.com/woocommerce/woocommerce-ios/pull/3763
- [x] [Design enhancements](https://user-images.githubusercontent.com/28450084/109961834-c0934300-7cea-11eb-9c0e-bed6dfbbf5fa.png): https://github.com/woocommerce/woocommerce-ios/pull/3763#issuecomment-790572165 (fixed in https://github.com/woocommerce/woocommerce-ios/pull/3794)
  - [x] Update navigation bar background color
  - [x] Remove navigation bar bottom border
  - [x] Update navigation bar items to use SF Symbols
  - [x] Products and Orders tab: move search CTA from left to right
  ~~- [ ] Products tab: update Add CTA to a FAB (requires FAB component from WPiOS) - out of scope~~

- [ ] Remove feature flag after launching https://github.com/woocommerce/woocommerce-ios/pull/6225

PR #3752, opened 2021-02-26
Enable large titles behind a feature flag (scrolling animation is only available in Products and Reviews tab)

jaclync:
Part of #3751 

## Why

Design calls for large titles in all of the four main tabs, which is already implemented in Android. This PR enables large titles behind a feature flag, and the title animation from scrolling only works in the Products and Reviews tab. Since the Dashboard and Orders tabs have multiple scroll views, we need to implement a workaround separately for to support title animation from scrolling (there's no API to inform the navigation bar on the scroll offset when there is more than one scroll view in the same level of view hierarchy).

More details are in the spike post: p91TBi-45c-p2

## Changes

- Created a feature flag `largeTitles`, which is enabled in debug or alpha build
- Updated navigation bar styles in https://github.com/woocommerce/woocommerce-ios/commit/d1e258d2299235575b630155684421a4f8520ff7
- Reviews tab: pin top edge to superview instead of safe area, otherwise large title is not shown after navigation back from review details https://github.com/woocommerce/woocommerce-ios/commit/f4d32d96fc340583155adb0c1cef8c621ce84923 (whenever we are not seeing large titles when we're supposed to, it's possible the scroll view's top edge is pinned to the safe area instead of superview)
- Created `WooTabNavigationController: UINavigationController` that is used for each Woo tab, and enables large title for the first view controller. The following view controllers in the navigation stack do not have large title by default but could override the new `func preferredLargeTitleDisplayMode() -> UINavigationItem.LargeTitleDisplayMode` to customize the large title display mode. Thanks to @Ecarrion for this simpler solution!

### Feature flag on

- Launch the app from Xcode (debug build), and log in if needed --> each tab should have large title
- Go to the products tab, and scroll the product list --> the title should animate with scrolling
- Tap on a product while large title is shown, optionally navigate to some other screens, then navigate back to the products tab --> the products tab should still have large title
- Go to the reviews tab, and scroll the review list --> the title should animate with scrolling
- Tap on a review while large title is shown, optionally navigate to some other screens, then navigate back to the reviews tab --> the reviews tab should still have large title
- Go to the dashboard tab, and scroll in each stats tab --> there is no title animation (this will be implemented in another subtask)
- Go to the orders tab, and scroll in the processing or all tab --> there is no title animation (this will be implemented in another subtask)

### Feature flag off

- In `DefaultFeatureFlagService`, return `false` for `largeTitles` feature flag, and launch the app. Log in if needed --> there should not be large title in all four tabs, and the app should look as before

## Example screenshots

\ | dashboard | orders | products | reviews
-- | -- | -- | -- | --
light | https://user-images.githubusercontent.com/1945542/109253406-b1178400-782a-11eb-8a9d-d531b2fc2e54.mov | https://user-images.githubusercontent.com/1945542/109253419-bb398280-782a-11eb-8cb6-e26a33f43667.mov | https://user-images.githubusercontent.com/1945542/109253531-fb006a00-782a-11eb-8989-5fa49f27bc57.mov | https://user-images.githubusercontent.com/1945542/109253576-0e133a00-782b-11eb-998f-2f7d6b8865dc.mov
dark | https://user-images.githubusercontent.com/1945542/109253208-3fd7d100-782a-11eb-9973-a0d72867203a.mov | https://user-images.githubusercontent.com/1945542/109253252-58e08200-782a-11eb-9be6-fd86b62c7500.mov | https://user-images.githubusercontent.com/1945542/109253265-61d15380-782a-11eb-8d7f-4540aea5c3c1.mov | https://user-images.githubusercontent.com/1945542/109253353-934a1f00-782a-11eb-9990-d3d755fc7920.mov

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/WooTabNavigationController.swift
+++ b/WooCommerce/Classes/System/WooTabNavigationController.swift
@@ -17,7 +17,7 @@ final class WooTabNavigationController: UINavigationController {
     }
 
     override var preferredStatusBarStyle: UIStatusBarStyle {
-        ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) ? .default: StyleManager.statusBarLight
+        ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) ? .default : StyleManager.statusBarLight
     }
 }
 
--- a/WooCommerce/Classes/System/WooTabNavigationController.swift
+++ b/WooCommerce/Classes/System/WooTabNavigationController.swift
@@ -1,0 +1,43 @@
+import UIKit
+
+/// Navigation controller for a Woo tab, shown as the root view controller of one of the tab bar.
+/// The first view controller always has large title, while the following view controllers in the navigation stack do not have large title by default.
+/// The following view controllers can override `preferredLargeTitleDisplayMode` function to change the large title display mode.
+final class WooTabNavigationController: UINavigationController {
+    init() {
+        super.init(nibName: nil, bundle: nil)
+        if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) {
+            navigationBar.prefersLargeTitles = true
+            delegate = self
+        }
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override var preferredStatusBarStyle: UIStatusBarStyle {
+        ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) ? .default: StyleManager.statusBarLight
+    }
+}
+
+extension WooTabNavigationController: UINavigationControllerDelegate {
+    // The first view controller always has large title, while the following view controllers in the navigation stack do not have large title by default.
+    func navigationController(_ navigationController: UINavigationController, willShow viewController: UIViewController, animated: Bool) {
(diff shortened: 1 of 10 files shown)


WooCommerce/Classes/System/WooTabNavigationController.swift
```diff
@@ -0,0 +1,43 @@
...
+            delegate = self
+        }
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override var preferredStatusBarStyle: UIStatusBarStyle {
+        ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) ? .default: StyleManager.statusBarLight
```
ealeksandrov: nit:
```suggestion
        ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) ? .default : StyleManager.statusBarLight
```

ealeksandrov (approved):
LGTM 🚀

WooCommerce/Classes/System/WooTabNavigationController.swift
jaclync: updated in https://github.com/woocommerce/woocommerce-ios/pull/3752/commits/e2ca1b1a61fe1b4b8fd47323ca64488ff547225c!

PR #3753, opened 2021-02-26
Update state flow on ReviewsViewController

ealeksandrov:
## Description

This PR fixes endless loading state on empty Reviews tab.

## Changes

1. Removes UI state switch in `viewWillAppear`
2. Adds resync call in `viewWillAppear` after UI state check instead of model check - so it won't trigger additional reload while model is empty, but synchronization is already in process.

### To reproduce the bug:

1. Open reviews tab.
2. On first open it should correctly show loading animation, then switch to empty state.
3. Switch to another tab and then back to reviews.
4. Never-ending loading animation will appear.
5. Try pull-to refresh to force resync and correctly switch to empty state.

### To validate the fix

1. Open reviews tab.
2. Wait until empty state.
3. Switch to another tab and then back to reviews.
4. Observe correct switch from loading -> empty.

## Video

https://user-images.githubusercontent.com/3132438/109297724-84239b00-7843-11eb-90fd-16c698774c8d.mp4

https://user-images.githubusercontent.com/3132438/109297733-85ed5e80-7843-11eb-8665-0c16c4e704cd.mp4

Update release notes:


pmusolino (approved):
LGTM :shipit: 

Just a nit: I think it's worth adding a line in the `RELEASE-NOTES` file about it. 👍

PR #3754, opened 2021-02-26
Shipping Labels: address formatting in the Shipping Label Form and loader indicator in the Address Validation Form

pmusolino:
Part of #2973 

## Description
In this PR I applied some minor fixes, plus some improvements to the shipping label form. In particular, in this PR now the formatted address will be shown in the Shipping Label Form, for the origin and destination address. Also, the done button for validating the address now the done button will be transformed in a loader indicator during the address validation request.

## Changes
- Added the new property `isTrivialNormalization` in `ShippingLabelAddressValidationResponse` entity because it will be useful in the near future for implementing the possibility for the user to choose the suggested address.
- Implemented an extension of the `ShippingLabelAddress` entity, where a method returns the `name`, `company`, and `address`. Basically combines the various components of a ShippingLabelAddress.
- Added a completion block to `ShippingLabelAddressFormViewController` that will return the latest shipping label address updated to the previous view controller.
- Now in the `ShippingLabelAddressFormViewController` the done button will be transformed in a loader indicator during the address validation request.
- Fixed the logic of the top banner view that was wrong.
- Fixed the keyboard type for some fields in the shipping label address form.
- Fixed the state for the `address` field, which now can be in the error state.
- Showed the formatted address (origin and destination) in the Shipping Label Form.
- Handle origin and destination address value changes in `ShippingLabelFormViewModel` + tests.

## Testing
2. Open an order detail.
4. In the Shipping Label Form -> you should see the formatted address for the origin and the destination address.
5. Press the button "Continue" in the Shipping Label form (under the origin address).
6. Try to validate an address pressing the Done button on top. -> The done button should be transformed in a loader indicator during the address validation request.
7. Try to edit the various field of the address. -> Make sure that every field has the right keyboard type (standard or numeric pad).

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Model/ShippingLabelAddress+Woo.swift
+++ b/WooCommerce/Classes/Model/ShippingLabelAddress+Woo.swift
@@ -13,7 +13,7 @@ extension ShippingLabelAddress {
     var fullNameWithCompanyAndAddress: String {
         var output: [String] = [fullNameWithCompany]
 
-        if let formattedPostalAddress = formattedPostalAddress, formattedPostalAddress.isEmpty == false {
+        if let formattedPostalAddress = formattedPostalAddress, formattedPostalAddress.isNotEmpty {
             output.append(formattedPostalAddress)
         }
 
@@ -35,10 +35,10 @@ private extension ShippingLabelAddress {
     var fullNameWithCompany: String {
         var output: [String] = []
 
-        if name.isEmpty == false {
+        if name.isNotEmpty {
             output.append(name)
         }
-        if company.isEmpty == false {
+        if company.isNotEmpty {
             output.append(company)
         }
 
@@ -49,7 +49,7 @@ private extension ShippingLabelAddress {
     /// Per US Post Office standardized rules for address lines. Ref. https://pe.usps.com/text/pub28/28c2_001.htm
     ///
     var combinedAddress: String {
-        guard address2.isEmpty == false else {
+        guard address2.isNotEmpty else {
             return address1
         }
 
--- a/WooCommerce/Classes/Model/ShippingLabelAddress+Woo.swift
+++ b/WooCommerce/Classes/Model/ShippingLabelAddress+Woo.swift
@@ -1,0 +1,72 @@
+import Foundation
+import Contacts
+import Yosemite
+
+
+// Yosemite.ShippingLabelAddress Helper Methods
+//
+extension ShippingLabelAddress {
+
+    /// Returns the `name`, `company`, and `address`. Basically this var combines the
+    /// various components of a ShippingLabelAddress.
+    ///
+    var fullNameWithCompanyAndAddress: String {
+        var output: [String] = [fullNameWithCompany]
+
+        if let formattedPostalAddress = formattedPostalAddress, formattedPostalAddress.isEmpty == false {
+            output.append(formattedPostalAddress)
+        }
+
+        return output.joined(separator: "\n")
+    }
+}
+
+private extension ShippingLabelAddress {
+
+    /// Returns the Postal Address, formated and ready for display.
+    ///
+    var formattedPostalAddress: String? {
+        return postalAddress.formatted(as: .mailingAddress)
+    }
+
+    /// Returns the `name` and `company` (on a new line). If either the `name` or `company` is empty,
+    /// then a single line is returned containing the other value.
+    ///
+    var fullNameWithCompany: String {
+        var output: [String] = []
+
+        if name.isEmpty == false {
+            output.append(name)
+        }
+        if company.isEmpty == false {
+            output.append(company)
+        }
+
+        return output.joined(separator: "\n")
+    }
+
+    /// Returns the two Address Lines combined (if there are, effectively, two lines).
+    /// Per US Post Office standardized rules for address lines. Ref. https://pe.usps.com/text/pub28/28c2_001.htm
+    ///
+    var combinedAddress: String {
+        guard address2.isEmpty == false else {
+            return address1
+        }
+
+        return address1 + " " + address2
+    }
+
+    /// Returns a CNPostalAddress with the receiver's properties
+    ///
+    var postalAddress: CNPostalAddress {
+        let address = CNMutablePostalAddress()
+        address.street = combinedAddress
+        address.city = city
+        address.state = state
+        address.postalCode = postcode
+        address.country = country
+        address.isoCountryCode = country
+
+        return address
+    }
+}
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -53,6 +53,7 @@ final class ShippingLabelAddressFormViewController: UIViewController {
         configureNavigationBar()
         configureMainView()
         configureTableView()
+        observeViewModel()
         configureConfirmButton()
         keyboardFrameObserver.startObservingKeyboardFrame(sendInitialEvent: true)
     }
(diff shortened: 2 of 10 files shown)


WooCommerce/Classes/Model/ShippingLabelAddress+Woo.swift
```diff
@@ -0,0 +1,72 @@
...
+    /// Returns the `name` and `company` (on a new line). If either the `name` or `company` is empty,
+    /// then a single line is returned containing the other value.
+    ///
+    var fullNameWithCompany: String {
+        var output: [String] = []
+
+        if name.isEmpty == false {
+            output.append(name)
+        }
+        if company.isEmpty == false {
```
Ecarrion: I've seen that we sometimes use `var == false` and sometimes `!var` through our codebase.
Should we be following any style here, I search on our style guide and couldn't find any clear direction, WDYT?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -94,7 +107,7 @@ private extension ShippingLabelAddressFormViewController {
     }
 
     func updateTopBannerView() {
-        topBannerView.isHidden = viewModel.addressValidationError?.generalError != nil
+        topBannerView.isHidden = viewModel.addressValidationError?.generalError == nil
```
Ecarrion: I think we can turn this into
```swift
topBannerView.isHidden = !viewModel.shouldShowTopBannerView
```
or
```swift
topBannerView.isHidden = !viewModel.hasAddresValidationError()
```

What do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -111,13 +124,19 @@ private extension ShippingLabelAddressFormViewController {
 private extension ShippingLabelAddressFormViewController {
 
     @objc func doneButtonTapped() {
-        // TODO: after the validation, return to the previous controller passing the new address.
+        configureRightButtonItemAsLoader()
```
Ecarrion: Tip:
If you move this logic into the VM like:

https://github.com/woocommerce/woocommerce-ios/blob/dd84c1d3064f8c065cc293079619ccc399a6d0d9/WooCommerce/Classes/ViewRelated/Products/Variations/Add%20Attributes/AddAttributeOptionsViewModel.swift#L84-L88

You can add unit tests to corroborate this behavior! 

https://github.com/woocommerce/woocommerce-ios/blob/dd84c1d3064f8c065cc293079619ccc399a6d0d9/WooCommerce/WooCommerceTests/ViewRelated/Variations/Add%20Attributes/AddAttributeOptionsViewModelTests.swift#L340-L363

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -107,29 +107,37 @@ private extension ShippingLabelFormViewController {
...
-                                                                                     type: .origin,
-                                                                                     address: shippingLabelAddress)
-            self.navigationController?.pushViewController(shippingAddressValidationVC, animated: true)
+            let shippingAddressVC = ShippingLabelAddressFormViewController(siteID: self.viewModel.siteID,
+                                                                           type: .origin,
+                                                                           address: shippingLabelAddress,
+                                                                           completion: { [weak self] (shippingLabelAddress) in
+                                                                            guard let self = self else { return }
+                                                                            self.viewModel.handleOriginAddressValueChanges(address: shippingLabelAddress)
+                                                                           })
```
Ecarrion, line 120: I hate this Xcode indentation! 🤪

Ecarrion (dismissed):
Lookking good, just left some small suggestions, let me know what you think?

WooCommerce/Classes/Model/ShippingLabelAddress+Woo.swift
pmusolino: We don't have a specific rule for this, but since all these properties here are not optionals, I think that using `isNotEmpty` is the best option. I apply the changes here edd7cb2ce5b8695132689569879da78cf84bb7f8

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
pmusolino: It makes sense to me. I applied the changes here e5c58a2ba3646b367d625f1a133da831bc30299e

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
pmusolino: Do you mean to introduce a state variable for managing the various states (in this moment the states are just two: `isUpdating` and `loaded`)?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
pmusolino: Same, at the same time there is no other way to stay in 160 characters :/

pmusolino:
@Ecarrion ready for a second look 🙇

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
Ecarrion: Yes, adding that state into the VM so it can be tested!

Ecarrion: The state can be stored into a single var,  `private var: isUpdating: Bool`
And then 
```swift
var showUpdateIndicator: Bool { 
    state.isUpdating 
} 
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
```diff
@@ -14,6 +14,10 @@ final class ShippingLabelAddressFormViewModel: NSObject {
 
     private let stores: StoresManager
 
+    var shouldShowTopBannerView: Bool {
```
Ecarrion, line 38: I think the name should be `shouldHideTopBannerView`, or we need to change the condition to `!=` right?

pmusolino: I'd say to maintain `shouldShowTopBannerView`, so I updated the conditions 085b2de6ab6e4d77be7581bfb9406fbe7d425802

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
pmusolino: Done 5c2a64c94c7a289e2e4f853e289c77657ef33164 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -64,10 +65,17 @@ private extension ShippingLabelAddressFormViewController {
 
     func configureNavigationBar() {
         title = viewModel.type == .origin ? Localization.titleViewShipFrom : Localization.titleViewShipTo
+        removeNavigationBackBarButtonText()
+        if viewModel.showLoadingIndicator {
+            configureRightButtonItemAsLoader()
+        }
+        else {
```
Ecarrion: nit:
```suggestion
        } else {
```

Ecarrion (approved):
🚀

pmusolino:
Thank you so much @Ecarrion for your detailed review!

PR #3757, opened 2021-02-26
Variations: Add back button confirmation on AttributePickerViewController

### LINKED ISSUE #3638: Variations: Add "discard changes" alert in AttributePickerViewController
Based on discussion in p1612977972011800-slack-CGPNUU63E and p91TBi-4rc-p2#comment-3318 we need to add a check for unsaved changes and show "Discard changes" alert.

Example:

https://github.com/woocommerce/woocommerce-ios/blob/4b748523cf63dc2b296eb5f4e647cea9b6cd62b7/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/Linked%20Products/LinkedProductsViewController.swift#L197-L222

Ecarrion:
fix #3638 

# Why

During testing, we noticed that the flow of configuring a "variation attributes options" was not presenting a confirmation action sheet when navigating back before saving changes.

**Note: This has to be tested on a device or on a simulator with a version lower than 13.4`.**

# How

- Override back button and swipe back intercepts to present the confirmation when there are pending changes.

# Demo
back-button

# Testing Steps
- Navigate to a specific variation
- Tap on the attributes row
- Select an option for any attribute
- Tap back twice(option picker and attribute picker)
- See the confirmation action sheet being presented.

Update release notes

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
@@ -29,6 +29,7 @@ final class AttributePickerViewController: UIViewController {
         registerTableViewHeaderSections()
         registerTableViewCells()
         configureTableView()
+        handleSwipeBackGesture()
     }
 }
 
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/AttributePickerViewController.swift
```diff
@@ -170,6 +188,12 @@ private extension AttributePickerViewController {
         tableView.reloadData()
         updateDoneButton()
     }
+
+    private func presentBackNavigationActionSheet() {
```
ealeksandrov: nit: extension is already private
```suggestion
    func presentBackNavigationActionSheet() {
```

ealeksandrov (approved):
LGTM :shipit:

Ecarrion:
@AliSoftware This has been merged as a beta-fix for 6.1!

AliSoftware:
This fix has been included in the new beta 6.1.0.1 🎉

PR #3761, opened 2021-02-26
[Mobile Payments] Add connected reader view

allendav:
Ready for review.

Includes
- mocked connected reader (that is not persisted)
- light mode AND dark mode

Does not include:
- any changes to the Settings or "Connect your card reader" screens
- any changes to nav bar title (still says "Card Readers" - should say "Manage Card Reader"
- the final form of the modals for "Scanning for readers" or "Found reader" or "Connecting to reader" - for now simple Alerts are being used. FancyAlerts that match the designs will be added in a subsequent PR. This was done to keep this PR under 400 lines.
- unit tests - those will come when we integrate with the CardReaderService in a subsequent PR

To test:
- Navigate to Settings (Gear), then Manage Card Readers
- Tap on "Connect card reader" button
- A "Scanning for Readers" alert will appear. Wait 3 seconds.
- A "Found reader" alert will replace it. Click on "Connect to Reader." Wait 3 seconds.
- A "Connecting to Reader" alert will replace it. Wait 3 seconds.
- A "Connected Readers" list will appear with the mocked reader. Note that its battery level declines 1% every 3 seconds.
- Click on the Disconnect button. You'll be returned to the Connect card reader view.

A light mode movie and a dark mode still:

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
@@ -144,6 +144,21 @@ private enum Row: CaseIterable {
         }
     }
 
+    var height: CGFloat {
+        switch self {
+        case .connectHeader,
+             .connectButton:
+            return UITableView.automaticDimension
+        case .connectImage:
+            return 250
+        case .connectHelpHintChargeReader,
+             .connectHelpHintTurnOnReader,
+             .connectHelpHintEnableBluetooth,
+             .connectLearnMore:
+            return 70
+        }
+    }
+
     var reuseIdentifier: String {
         return type.reuseIdentifier
     }
@@ -177,16 +192,7 @@ extension CardReaderSettingsConnectView: UITableViewDataSource {
 extension CardReaderSettingsConnectView: UITableViewDelegate {
     func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
         let row = rowAtIndexPath(indexPath)
-        if row == .connectImage {
-            return 250
-        }
-        if row == .connectHelpHintChargeReader || row == .connectHelpHintTurnOnReader || row == .connectHelpHintEnableBluetooth {
-            return 70
-        }
-        if row == .connectLearnMore {
-            return 70
-        }
-        return UITableView.automaticDimension
+        return row.height
     }
 
     func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
@@ -55,13 +55,11 @@ final class CardReaderSettingsConnectView: NSObject {
             "Connect your card reader",
             comment: "Settings > Manage Card Reader > Prompt user to connect their first reader"
         )
-        cell.hideSeparator()
         cell.selectionStyle = .none
     }
 
     private func configureImage(cell: ImageTableViewCell) {
         cell.detailImageView?.image = UIImage(named: "card-reader-connect")
-        cell.hideSeparator()
         cell.selectionStyle = .none
     }
 
@@ -73,7 +71,6 @@ final class CardReaderSettingsConnectView: NSObject {
         cell.itemTextLabel?.text = NSLocalizedString(
             "Make sure card reader is charged",
             comment: "Settings > Manage Card Reader > Connect > Hint to charge card reader")
-        cell.hideSeparator()
         cell.selectionStyle = .none
     }
 
@@ -85,7 +82,6 @@ final class CardReaderSettingsConnectView: NSObject {
         cell.itemTextLabel?.text = NSLocalizedString(
             "Turn card reader on and place it next to mobile device",
             comment: "Settings > Manage Card Reader > Connect > Hint to power on reader")
-        cell.hideSeparator()
         cell.selectionStyle = .none
     }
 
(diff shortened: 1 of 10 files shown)


Garance91540:
🎉 Connecting readers complete!

Garance91540 (approved):

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
```diff
@@ -190,6 +183,9 @@ extension CardReaderSettingsConnectView: UITableViewDelegate {
         if row == .connectHelpHintChargeReader || row == .connectHelpHintTurnOnReader || row == .connectHelpHintEnableBluetooth {
             return 70
         }
+        if row == .connectLearnMore {
```
ctarda: I'm just curious, but is there any particular reason to have this separate branch, instead of collapsing this into something like:
`if row == .connectHelpHintChargeReader || row == .connectHelpHintTurnOnReader || row == .connectHelpHintEnableBluetooth || row == .connectLearnMore`?

Which, at that point, I would suggest extracting the if... to a separate method so it would read like `if cellisFixedHeigh(for row:)` or similar

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
```diff
@@ -0,0 +1,127 @@
...
+        return cell
+    }
+}
+
+// MARK: - UITableViewDelegate Conformance
+//
+extension CardReaderSettingsConnectedReaderView: UITableViewDelegate {
+    func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        let row = rowAtIndexPath(indexPath)
+        if row == .connectedReader {
```
ctarda: This might end up being problematic when supporting Dynamic Type, as the cell is of a fixed height. 

If I recall correctly, the best way to implement support for self-resizing cells is returning .automaticDimension in tableView:heightForRowAt:indexPath and returning the specific height (in this case 60) in tableView:estimatedHeightForRowAt:indexPath.

I think it's ok if we defer support for Dynamic Type to a separate ticket, but it's something we have to implement.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -23,35 +24,91 @@ enum CardReaderSettingsViewActiveAlert {
...
-        self.foundReader = nil
-        self.connectedReader = nil
+        activeView = .connectYourReader
+        activeAlert = .none
+        knownReaders = []
+        foundReader = nil
+        connectedReader = nil
+        timer = nil // TODO Remove
+
+        // TODO Remove - simulates a declining battery level with a timer
```
ctarda, line 48: And yet, it's a great way to test that the view gets updated 💯

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderTableViewCells/ConnectedReaderTableViewCell.swift
```diff
@@ -0,0 +1,6 @@
+import UIKit
+
+class ConnectedReaderTableViewCell: UITableViewCell {
```
ctarda: Should this class be final?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderTableViewCells/ConnectedReaderTableViewCell.swift
```diff
@@ -0,0 +1,6 @@
...
+    @IBOutlet weak var serialNumberLabel: UILabel!
```
ctarda, line 4: The pattern we use is:
* outlets are private
* there is a `configure()` method where you pass an object, and the cell takes care of populating its outlets.

ctarda (approved):
👏 It works great, and the code looks really good.


https://user-images.githubusercontent.com/2722505/109659549-c8f64d00-7b35-11eb-8a03-5048a4355e3a.mov

There are a couple of things that we might want to defer to separate PRs, none of them are blockers, imo.
:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectView.swift
allendav: I didn't want to group the unrelated views with each other is all. Can change.

allendav: Changed to be a property of the Row Enum in https://github.com/woocommerce/woocommerce-ios/pull/3761/commits/2022ea4b09e87cc0c85f6d122cdde68746dd2406
This will also allow `heightForRowAt` to move into a superclass in a follow on PR.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
allendav: https://github.com/woocommerce/woocommerce-ios/issues/3770

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderTableViewCells/ConnectedReaderTableViewCell.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3761/commits/da11c75b16fd220869ccc27624622a62f23169ff

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderTableViewCells/ConnectedReaderTableViewCell.swift
allendav: https://github.com/woocommerce/woocommerce-ios/issues/3771

PR #3764, opened 2021-03-01
[Variations] Update parent product after variation deletion

### LINKED ISSUE #3193: Variations: Allow Deletion of a Single Variation

ealeksandrov:
Closes https://github.com/woocommerce/woocommerce-ios/issues/3193, https://github.com/woocommerce/woocommerce-ios/issues/3619.

## Description

This PR fixes adds parent product updating after variation deletion.

## Changes

1. Adds `onVariationDeletion` stored closure in `ProductVariationFormViewModel`
2. Adds local product update in `onVariationDeletion` callback in `ProductVariationsViewController`

## Test

1. Open product with existing variations.
2. Open variation details
3. Tap "•••" in navbar, pick "Delete"
4. Repeat deletion for all variations
5. Observe correct empty state instead of simple empty table
6. Go back on parent product details and check that variations cell has "Add variations" text

## Video

https://user-images.githubusercontent.com/3132438/109465837-d9e18880-7a79-11eb-8839-02cfabf192e0.mp4

https://user-images.githubusercontent.com/3132438/109465846-dd750f80-7a79-11eb-8794-2ea03b547589.mp4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
@@ -57,6 +57,10 @@ final class ProductVariationFormViewModel: ProductFormViewModelProtocol {
         }
     }
 
+    /// Assign this closure to get notified when the variation is deleted.
+    ///
+    var onVariationDeletion: ((ProductVariation) -> Void)?
+
     private let allAttributes: [ProductAttribute]
     private let parentProductSKU: String?
     private let productImageActionHandler: ProductImageActionHandler
@@ -272,6 +276,7 @@ extension ProductVariationFormViewModel {
         let deleteAction = ProductVariationAction.deleteProductVariation(productVariation: productVariation.productVariation) { result in
             switch result {
             case .success:
+                self.onVariationDeletion?(self.productVariation.productVariation)
                 onCompletion(.success(()))
             case .failure(let error):
                 onCompletion(.failure(error))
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
```diff
@@ -272,6 +276,7 @@ extension ProductVariationFormViewModel {
         let deleteAction = ProductVariationAction.deleteProductVariation(productVariation: productVariation.productVariation) { result in
             switch result {
             case .success:
+                self.onVariationDeletion?(self.productVariation.productVariation)
```
jaclync: nit: maybe weak self just in case?

jaclync: nit: could this be unit tested? like mocking `ProductVariationAction.deleteProductVariation` to return success and asserting the callback is called, while the error case doesn't

jaclync (approved):
LGTM (just some nit suggestions for your consideration)! 🚢

ealeksandrov:
Thanks for the review @jaclync!
I've added `weak self` and `onVariationDeletion` checks to existing viewmodel unit tests 👍

PR #3695, opened 2021-02-18
Variations: Add error state for syncing global attribute options

ealeksandrov:
Closes https://github.com/woocommerce/woocommerce-ios/issues/3601.

## Description

This PR adds error state on attribute options screen.

## Changes

1. Switch `isSyncing` from `Bool` to `SyncState` enum.
2. Add `.failed` case for `SyncState`
3. Display `EmptyStateViewController` with error description and retry action for `.failed` case in `AddAttributeOptionsViewController`.

`EmptyStateViewController` picked instead of non-blocking `Notice` because we need to block user from adding new options until existing data is properly synced.

## Test

1. Open variable product, with existing variations or not
2. Tap "variations" (or "add variation")
3. For no variations:
a. Tap "Add Variation"
For existing variations:
a. Tap "•••" in navbar, pick "Edit Attributes"
b. Tap "Add New Attribute"

4. Turn off internet connection
5. Select global attribute
6. Observe error state
7. Turn on internet connection and tap "Retry"
8. Observe correctly synced options list

## Screenshots

before | after
--|--
Simulator Screen Shot - iPhone 12 - 2021-02-18 at 12 37 39|Simulator Screen Shot - iPhone 12 - 2021-02-18 at 12 38 35

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3765, opened 2021-03-01
Fix error state crash on AddAttributeOptionsViewController

ealeksandrov:
## Description

This PR fixes crash introduced in https://github.com/woocommerce/woocommerce-ios/pull/3695.

## Changes

Mostly a rollback of refactor from 56d4a38. Calling `configure` on `EmptyStateViewController` too early leads to a crash because IBOutlets are nil at this point. Moved `configure` to be called after adding view to hierarchy.

## Test

1. Open variable product, with existing variations or not
2. Tap "variations" (or "add variation")
3. For no variations:
a. Tap "Add Variation"
For existing variations:
a. Tap "•••" in navbar, pick "Edit Attributes"
b. Tap "Add New Attribute"
4. Turn off internet connection
5. Select global attribute
6. Observe error state (without a crash 🙂)
7. Turn on internet connection and tap "Retry"
8. Observe correctly synced options list

Update release notes:


jaclync (approved):
I was able to reproduce the crash without turning off the internet connection (from testing [the other PR](https://github.com/woocommerce/woocommerce-ios/pull/3764)) and now it doesn't crash anymore! :shipit:

PR #3768, opened 2021-03-01
Variations: Add UI for rename attributes flow

rachelmcr:
Related: #3212 

## Description

This PR adds the UI for renaming variation attributes:

* Adds a Rename button in the options menu.
* Builds and presents the Rename Attribute screen.

This only adds the UI; the changes to actually perform the update (to rename the attribute) will come in another PR.

## Changes

* Updates `AddAttributeOptionsViewController` to add a Rename button to the `...` menu action sheet. Tapping the Rename button navigates to the Rename Attribute screen.
* New `RenameAttributesViewController`, `RenameAttributesViewController.xib`, and `RenameAttributesViewModel` for the Rename Attribute screen:
   * Rename Attribute screen displays a single text field for the attribute name, prefilled with the current attribute name.
   * Done button is disabled if there is no text in the text field (to ensure the attribute has a name).
   * For now, tapping Done just navigates back. (This will be updated to perform the update in another PR.)
   * Checks for unsaved changes and presents "discard changes" action sheet when navigating back.

Rename button|Rename Attribute screen
-|-
Simulator Screen Shot - iPhone 12 - 2021-03-01 at 11 31 56|Simulator Screen Shot - iPhone 12 - 2021-03-01 at 11 32 00

## Testing

1. Go to the Products tab.
2. Select a variable product.
3. Select Variations to open the variations list.
4. Tap the `...` menu and select Edit Attributes.
5. Select an attribute to open the Attribute Options screen.
6. Tap the `...` menu and select Rename.
7. Confirm the Rename Attribute screen opens as expected, with the text field prefilled with the current name of the attribute you selected.
8. Confirm you can enter a new attribute name in the text field, and the Done button is disabled if you remove all the text there.

## Submitter Checklist

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -365,6 +370,18 @@ extension AddAttributeOptionsViewController {
         present(actionSheet, animated: true)
     }
 
+    /// Navigates to `RenameAttributesViewController`
+    ///
+    private func navigateToRenameAttribute() {
+        guard let attributeName = self.viewModel.titleView else {
```
Ecarrion: nit: This is an unnecessary `self` since we are not inside a closure!

Ecarrion: I think there is no reason for `titleView` to be optional (I know it was like that before this PR 😛)
And, since we are using it for more than the title view, probably is best if we rename the property to `attributeName` or something like that, what do you think?

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        configureNavigationBar()
+        configureMainView()
+        configureTableView()
+        enableDoneButton(true)
+    }
+
+    override func viewWillDisappear(_ animated: Bool) {
```
Ecarrion: nit: I think we can delete this

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+// MARK: - Navigation actions handling
+
+extension RenameAttributesViewController {
+    @objc private func doneButtonTapped() {
+        // TODO: Update the attribute name if it has changed
+        // For now, just navigate back
+        self.navigationController?.popViewController(animated: true)
```
Ecarrion: nit: This is an unnecessary self since we are not inside a closure!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+    func configureAttributeName(cell: TextFieldTableViewCell) {
+        cell.accessoryType = .none
+
+        let placeholder = Localization.placeholder
+
+        let cellViewModel = TextFieldTableViewCell.ViewModel(text: viewModel.attributeName,
+                                                         placeholder: placeholder,
+                                                         onTextChange: { [weak self] newAttributeName in
+                                                            self?.viewModel.handleAttributeNameChange(newAttributeName)
+                                                            self?.enableDoneButton(self?.viewModel.newAttributeName != nil)
```
Ecarrion: I think it's worth to move this logic inside the `ViewModel` and provide an output like `shouldEnableDoneBotton`, that way we can add a test for it!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewController.swift
```diff
@@ -0,0 +1,228 @@
...
+                                                         },
+                                                         onTextDidBeginEditing: nil,
+                                                         onTextDidReturn: { [weak self] _ in
+                                                            self?.doneButtonTapped()
```
Ecarrion, line 175: Neat!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewModel.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    let attributeName: String
+
+    private(set) var newAttributeName: String?
+
+    init(attributeName: String) {
+        self.attributeName = attributeName
+    }
+
+}
```
Ecarrion: A few comments here and there, just to keep consistency with the rest of the project would be nice :-)

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewModel.swift
```diff
@@ -0,0 +1,31 @@
...
+
+// MARK: - Actions
+extension RenameAttributesViewModel {
+
+    func handleAttributeNameChange(_ name: String?) {
+        guard name != nil && name?.isNotEmpty == true else {
+            newAttributeName = nil
+            return
+        }
```
Ecarrion: I think you can drop this guard and just assign the content, or am I missing something? 😅

Ecarrion (commented):
Nice PR @rachelmcr, it looks very good! 
I left you some small suggestions and I have one extra question!

What would be the reason to use a table view to power this screen? I think that having some simple views with a simple layout could reduce the complexity a lot, what do you think?

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewModel.swift
rachelmcr: Oh, yep! I was struggling to check for an empty text field to enable the Done button, so I was using this to check for `nil`. But I realized I can just assign the string here and then check for an empty string to enable the Done button:

```swift
self?.enableDoneButton(self?.viewModel.newAttributeName != "")
```

rachelmcr:
Thank you for the feedback! It all makes sense and I'm working on those changes. About this:

> What would be the reason to use a table view to power this screen? I think that having some simple views with a simple layout could reduce the complexity a lot, what do you think?

I mostly did it that way because we already have the reusable `TextFieldTableViewCell` and by making it a table view the footer text is also easy to add. I'd seen it done this way in some other fairly simple screens, and was familiar with the setup for using a table view, so I thought it would be a reasonable approach. That said, I could rework it to not use a table view if you think that'd be better!

Ecarrion:
> That said, I could rework it to not use a table view if you think that'd be better!

I think it would have been simpler, would require less code, and will be easier to maintain, but up to you, both approaches are correct!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewModel.swift
rachelmcr: Added in https://github.com/woocommerce/woocommerce-ios/pull/3768/commits/e7265f4d118e7335e751d803d29415121f675f83

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewController.swift
rachelmcr: Done in https://github.com/woocommerce/woocommerce-ios/pull/3768/commits/bb36f3908b1c20488ea4fe04366b9785a2658d5b! I realized I'd forgotten to add unit tests in this PR so I added those in https://github.com/woocommerce/woocommerce-ios/pull/3768/commits/6b3eb8dcf406268439ebf104aae8a3f76323e457. Thanks for the nudge!

rachelmcr:
@Ecarrion Could you give it another look? I made quite a few changes based on your feedback, and added unit tests.

I'm thinking that for now I'll merge it using a table view like this, since it works, so we can keep the variations project moving forward. But I agree with the reasoning behind doing something simpler — updating the screen to something simpler could be a good hack week task for me next week.

Ecarrion (approved):
Looks good to me! Thanks for the changes!

Issue #2975, opened 2020-10-14
Shipping Labels M2: add a link to order details to view information about shipping label creation


anitaa1990:
We need to provide a link below the **Create shipping label** to find out how to create shipping labels. This will only be displayed if the **Create shipping label button** is displayed.

## Subtasks

- [ ] Show the info link in order details https://github.com/woocommerce/woocommerce-ios/pull/3769
- [ ] Navigate to the new info screen with a link "Learn more" to https://woocommerce.com/products/shipping/ (we can probably reuse `EmptyStateViewController` for this screen)

<img src="https://user-images.githubusercontent.com/22608780/95947008-59dcbe00-0e0b-11eb-89f7-9f2c8e5e081d.png" width="300"/>. <img src="https://user-images.githubusercontent.com/22608780/95947034-63febc80-0e0b-11eb-9fbd-bca696ed0793.png" width="300"/>

jaclync:
@woocommerce/mobile-designers I'd like to confirm some design differences between the shipping label creation info screen and another use case of a reusable UI component:

This is how the design looks like for shipping label creation info screen:

<img width="245" alt="Screen Shot 2021-03-02 at 5 19 46 PM" src="https://user-images.githubusercontent.com/1945542/109626426-822b4600-7b7b-11eb-85ae-28bfabe1fbdd.png">

Since this screen looks a lot like the empty state UI like in the orders tab (I hardcoded to show the empty state), I was wondering if we could reuse the design here to be more consistent (screenshots below). 

Using the empty state UI link style | Orders tab empty state
-- | --
![Simulator Screen Shot - iPhone 11 - 2021-03-02 at 17 14 34](https://user-images.githubusercontent.com/1945542/109626785-e948fa80-7b7b-11eb-8cba-c575496900a6.png) | ![Simulator Screen Shot - iPhone 11 - 2021-03-02 at 17 17 15](https://user-images.githubusercontent.com/1945542/109626800-ecdc8180-7b7b-11eb-8200-1ef138dc8f32.png)


The differences between the dev screenshot and design are:

- The details text color ("Cut the post office line...") is dark (design) vs. gray (empty state component)
- The "Learn more" CTA is a link (design) vs. a primary button (empty state component)

Do you think it's better making the two design customizations above, or following the current empty state text color & button style? I can implement the design customizations, but just wanted to confirm before doing so!

Garance91540:
[...]

Issue #3574, opened 2021-02-02
Shipping Labels: fix icon alignment and update font size for the(i) note on the Order Detail screen


designsimply:
<blockquote>On the Order Details screen:
<br>– can we vertically align the info icon with the link "Don't know how to print from your phone?"?
<br>– can we update the font size of the link to “Subhead”?</blockquote>
<sup>(internal reference: p5T066-1K4-p2#comment-6955 ht @Garance91540)</sup>

<br><img width="420" alt="Screen Shot 2021-02-01 at 14 58 26" src="https://user-images.githubusercontent.com/1119271/106660021-bdc2f800-655c-11eb-8515-33ab35116efd.png">
<sup>Tested with WCiOS 5.9.0.0.</sup>

Issue #3279, opened 2020-12-03
Order Details: "Details" screen duplicates information from the main order screen. Not needed?


eshurakov:
**Details**
When an order is in [failed, on hold, pending] state, there is a "details" row under the list of products on the order screen. Details screen shows same information that you can already get from the main order screen.

Maybe details screen can be removed?

|||
|-|-|
|![IMG_DBE70A8261EC-1](https://user-images.githubusercontent.com/54751/100980656-de170c00-3545-11eb-972b-71ba43a624af.jpeg)|![IMG_DBE70A8261EC-2](https://user-images.githubusercontent.com/54751/100980678-e4a58380-3545-11eb-97bf-8c11dc463586.jpeg)|

**Mobile Environment**
Please include:
- Device: iPhone
- iOS version: iOS 14.2
- WooCommerce iOS version: 5.6

eshurakov:
If details screen is removed, issue https://github.com/woocommerce/woocommerce-ios/issues/3274 will also be fixed.

jaclync:
From [G's comment](https://github.com/woocommerce/woocommerce-ios/pull/3769#pullrequestreview-601910258), we are good to remove the "Details" row in order details and the product list screen, which presents the same information as the product list in order details. cc @woocommerce/mobile-designers 

I'll work on this shortly!

jaclync:
We no longer have the "details" row in the products section anymore, I also checked [`OrderDetailsDataSource` code to confirm](https://github.com/woocommerce/woocommerce-ios/blob/4eb7ba59270a9d899b0a19a04a07f14640f2fa05/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsDataSource.swift#L902-L950). Closing!

<img src="https://user-images.githubusercontent.com/1945542/158960101-bde6e695-6222-474f-8e69-917709da4794.png" width="300" />

PR #3769, opened 2021-03-02
Order Details: show shipping label creation info row when the order is eligible

jaclync:
Part of #2975 

## Why

For users who are new to shipping label creation in mobile, we want to show an info link for [more details](https://user-images.githubusercontent.com/22608780/95947034-63febc80-0e0b-11eb-9fbd-bca696ed0793.png). This PR implemented the first part of the task - adding an info link row to order details if the order is eligible for shipping label creation.

### Pre-existing issues in `ImageAndTitleAndTextTableViewCell`

I noticed that footnote style wasn't configured correctly in `ImageAndTitleAndTextTableViewCell`, since the title label font is configured in multiple places. There's some tech debt in the cell to have different configuration styles https://github.com/woocommerce/woocommerce-ios/issues/3419, which I hope to address this sprint. For this PR, I fixed the title font style in https://github.com/woocommerce/woocommerce-ios/commit/0956a9ce7bd0113ee1b8141345cc75758eef7037.

### Showed shipping label creation info link row in `OrderDetailsDataSource`

- Added a new `Row.shippingLabelCreationInfo(showsSeparator: Bool)` enum case for the info link row
- Since there are two states where the info link is shown (whether "Mark Order Complete" CTA is visible), some associated values have to be used to configure cells up to design. This complicates the switch statement for cell configuration, lemme know if you know of a better way to unwrap the associated values!

### Order that is processing

- Go to the orders tab
- In the Processing tab, tap on an order --> an info link row should be shown below the "Mark Order Complete" CTA

### Order that is not processing and not refunded

- Go to the orders tab
- In the All Orders tab, tap on an order that is not processing and not refunded --> an info link row should be shown below the "Create Shipping Label" CTA

## Example screenshots

Notes for design reviewers: I didn't see design where "Mark Order Complete" CTA isn't visible, so I placed the info link below the "Create Shipping Label" CTA for now. Please lemme know if you have different thoughts!

\ | eligible for shipping label creation - processing order | eligible for shipping label creation - non-processing non-refunded order | not eligible for shipping label creation - processing order | not eligible for shipping label creation - non-processing non-refunded order
-- | -- | -- | -- | --
dark | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 32 31 | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 32 22 | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 35 32 | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 35 29
light | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 33 00 | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 32 50 | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 35 43 | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 11 35 40

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -266,6 +266,11 @@ private extension OrderDetailsDataSource {
             configureShippingLabelDetail(cell: cell)
         case let cell as ImageAndTitleAndTextTableViewCell where row == .shippingNotice:
             configureShippingNotice(cell: cell)
+        case let cell as ImageAndTitleAndTextTableViewCell where row == .shippingLabelCreationInfo(showsSeparator: true),
+             let cell as ImageAndTitleAndTextTableViewCell where row == .shippingLabelCreationInfo(showsSeparator: false):
+            if case .shippingLabelCreationInfo(let showsSeparator) = row {
+                configureShippingLabelCreationInfo(cell: cell, showsSeparator: showsSeparator)
+            }
         case let cell as ImageAndTitleAndTextTableViewCell where row == .shippingLabelPrintingInfo:
             configureShippingLabelPrintingInfo(cell: cell)
         case let cell as LeftImageTableViewCell where row == .addOrderNote:
@@ -288,10 +293,13 @@ private extension OrderDetailsDataSource {
             configureAggregateOrderItem(cell: cell, at: indexPath)
         case let cell as ButtonTableViewCell where row == .shippingLabelCreateButton:
             configureCreateShippingLabelButton(cell: cell, at: indexPath)
-        case let cell as ButtonTableViewCell where row == .markCompleteButton(style: .primary):
-            configureMarkCompleteButton(cell: cell, buttonStyle: .primary)
-        case let cell as ButtonTableViewCell where row == .markCompleteButton(style: .secondary):
-            configureMarkCompleteButton(cell: cell, buttonStyle: .secondary)
+        case let cell as ButtonTableViewCell where row == .markCompleteButton(style: .primary, showsBottomSpacing: true),
+             let cell as ButtonTableViewCell where row == .markCompleteButton(style: .primary, showsBottomSpacing: false),
+             let cell as ButtonTableViewCell where row == .markCompleteButton(style: .secondary, showsBottomSpacing: true),
+             let cell as ButtonTableViewCell where row == .markCompleteButton(style: .secondary, showsBottomSpacing: false):
+            if case .markCompleteButton(let style, let showsBottomSpacing) = row {
+                configureMarkCompleteButton(cell: cell, buttonStyle: style, showsBottomSpacing: showsBottomSpacing)
+            }
         case let cell as ButtonTableViewCell where row == .shippingLabelReprintButton:
             configureReprintShippingLabelButton(cell: cell, at: indexPath)
         case let cell as OrderTrackingTableViewCell where row == .tracking:
@@ -461,6 +469,25 @@ private extension OrderDetailsDataSource {
         cell.hideSeparator()
     }
 
+    private func configureShippingLabelCreationInfo(cell: ImageAndTitleAndTextTableViewCell, showsSeparator: Bool) {
+        cell.update(with: .imageAndTitleOnly(fontStyle: .footnote),
+                    data: .init(title: Title.shippingLabelCreationInfoAction,
+                                image: .infoOutlineFootnoteImage,
+                                imageTintColor: .systemColor(.secondaryLabel),
+                                numberOfLinesForTitle: 0,
+                                isActionable: false,
+                                showsSeparator: showsSeparator))
+
+        cell.selectionStyle = .default
+
+        cell.accessibilityTraits = .button
+        cell.accessibilityLabel = Title.shippingLabelCreationInfoAction
+        cell.accessibilityHint =
+            NSLocalizedString("Tap to show instructions on how to create a shipping label on the mobile device",
+                              comment:
+                                "VoiceOver accessibility hint for the row that shows instructions on how to print a shipping label on the mobile device")
+    }
+
     private func configureShippingLabelDetail(cell: WooBasicTableViewCell) {
         cell.bodyLabel?.text = Footer.showShippingLabelDetails
         cell.applyPlainTextStyle()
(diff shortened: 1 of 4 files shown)


ghost:
The design implementation looks great to me @jaclync however maybe we could name it something other than just 'details' do you think 'shipping details' would work @Garance91540

jaclync:
> The design implementation looks great to me @jaclync however maybe we could name it something other than just 'details' do you think 'shipping details' would work @Garance91540

Hey @woocommerce/mobile-designers! The "Details" here actually means "order's product details" and feels quite redundant to me, since it just navigates to another screen of the same list of products as in order details:

order details | tapping "Details" in the Product(s) section
-- | --
Simulator Screen Shot - iPhone 11 - 2021-03-02 at 14 32 47 | Simulator Screen Shot - iPhone 11 - 2021-03-02 at 14 32 49

I'm not sure why we have this separate product list screen in the first place 🤔

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -461,6 +469,25 @@ private extension OrderDetailsDataSource {
...
+                                numberOfLinesForTitle: 0,
+                                isActionable: false,
+                                showsSeparator: showsSeparator))
+
+        cell.selectionStyle = .default
+
+        cell.accessibilityTraits = .button
+        cell.accessibilityLabel = Title.shippingLabelCreationInfoAction
+        cell.accessibilityHint =
+            NSLocalizedString("Tap to show instructions on how to create a shipping label on the mobile device",
```
pmusolino: Maybe it is superfluous to specify “on mobile device”?

WooCommerce/Classes/ViewRelated/ReusableViews/ButtonTableViewCell.swift
```diff
@@ -62,7 +62,7 @@ private extension ButtonTableViewCell {
     }
 }
 
-private extension ButtonTableViewCell {
+extension ButtonTableViewCell {
```
pmusolino, line 65: Q: Is there a reason why this is no more `private`?

pmusolino (commented):
Looks good to me @jaclync!
I just noticed an issue on big screens (like on iPhone 12 pro) where the label and the icon are not vertically aligned.

Garance91540 (approved):
Can't wait to see this button in the app for real!!

Everything looks good except as @jaclync mentioned we shouldn't be seeing the "Details" row in there. I'm assuming it's a legacy from when we had the "Details" screen on the Order screen?

https://d.pr/i/05913c

jaclync:
> Everything looks good except as @jaclync mentioned we shouldn't be seeing the "Details" row in there. I'm assuming it's a legacy from when we had the "Details" screen on the Order screen?
https://d.pr/i/05913c

@Garance91540 yup let's remove this "Details" row! We already have an issue for this removal https://github.com/woocommerce/woocommerce-ios/issues/3279, which I'll work on after this PR.

jaclync:
> I just noticed an issue on big screens (like on iPhone 12 pro) where the label and the icon are not vertically aligned.

@pmusolino I think this misalignment is from the dynamic type to support various font sizes. If you change the system font size, the icon and text might look aligned in some sizes? This is a tricky issue, since the text could be any number of lines and I haven't found a good way to vertically align the center of the icon and the **first line** of the text. We also have an existing issue for this https://github.com/woocommerce/woocommerce-ios/issues/3574 - I'll see how we can align this better separately!

WooCommerce/Classes/ViewRelated/ReusableViews/ButtonTableViewCell.swift
jaclync: Before this PR, we only need to customize a `ButtonTableViewCell`'s bottom spacing to 0pt. However, in this PR we could be showing the default bottom spacing (20pt) or 0pt in `OrderDetailsDataSource.configureMarkCompleteButton`. I thought it'd be better using the same default value from the cell in case we change it in the future. Lemme know if you have other thoughts!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
jaclync: Makes sense! Updated in https://github.com/woocommerce/woocommerce-ios/pull/3769/commits/0125fea0e2b728d7ba5b1fb19a7ed1192a689d01

jaclync:
@pmusolino would you like to take another look at the updated PR? 😄

pmusolino:
> @pmusolino I think this misalignment is from the dynamic type to support various font sizes. If you change the system font size, the icon and text might look aligned in some sizes? This is a tricky issue, since the text could be any number of lines and I haven't found a good way to vertically align the center of the icon and the first line of the text. We also have an existing issue for this #3574 - I'll see how we can align this better separately!

This is right, but probably the text is aligned on top of the icon, so with small text sizes, it seems misaligned. If you align it at the center of the icon it should fix all the issues with all the text sizes.

Also with large titles there is a small vertical misalignment

jaclync:
> Also with large titles there is a small vertical misalignment

@pmusolino did you mean the alignment here (just wanted to confirm, since I wasn't sure if by "large titles" you meant the recent navigation bar large titles change):



> This is right, but probably the text is aligned on top of the icon, so with small text sizes, it seems misaligned. If you align it at the center of the icon it should fix all the issues with all the text sizes.

The tricky part is that the icon height and text line height aren't always the same, so we can't easily align them vertically. Vertically aligning the icon and text at the center would be a design change all over the app. @woocommerce/mobile-designers do you think we're okay with this change? Like in the screenshot above, the icon and text (single or multiple lines) are center aligned vertically.

pmusolino:
> @pmusolino did you mean the alignment here (just wanted to confirm, since I wasn't sure if by "large titles" you meant the recent navigation bar large titles change)

Yes exactly

> The tricky part is that the icon height and text line height aren't always the same, so we can't easily align them vertically. Vertically aligning the icon and text at the center would be a design change all over the app.

Right, but probably moving the label 4 pixel below, will improve the appearance with every text size. cc @woocommerce/mobile-designers

Garance91540:
> @woocommerce/mobile-designers do you think we're okay with this change? Like in the screenshot above, the icon and text (single or multiple lines) are center aligned vertically.

Sounds good!

jaclync:
> Right, but probably moving the label 4 pixel below, will improve the appearance with every text size.

@pmusolino Let's go with the vertical center alignment as G suggested! since the design change affects other parts of the app, I plan to work on it separately https://github.com/woocommerce/woocommerce-ios/issues/3574 for more thorough testing. Also, this PR would unblock other work like https://github.com/woocommerce/woocommerce-ios/issues/3279 and https://github.com/woocommerce/woocommerce-ios/pull/3785. Lemme know if you think there are other issues to address for this PR!

jaclync:
@pmusolino do you think this PR is good to merge? (it hasn't got an approval from code review 😄 )

pmusolino (approved):
LGTM :shipit:

PR #3772, opened 2021-03-02
[Mobile payments] Fetch connection token from WCPay-enabled stores

### LINKED ISSUE #3737: [Mobile Payments] Extract Networking Foundation
For context: p91TBi-4rr-p2

The Stripe Terminal SDK needs to perform an authenticated network request to an endpoint already implemented in WooCommerce Payments.

in order to provide support for that, we need to extract the facilities that we have in Networking now to a separate module. 

This is starting to become a priority now, as we need it in order to run end to end tests

ctarda:
I will document this better before submitting a formal PR (if we think it deserved to be submitted as a formal PR) but for now, this brief description will have to do.

## The difficult solution
* We want to fetch the [connection token required](https://stripe.com/docs/terminal/sdk/ios#connection-token) by the Stripe Terminal SDK 
* We want to do so by leveraging the existing infrastructure that allows us to submit signed requests via the Jetpack tunnel
* We want to do so in a way that does not require us to blow up the current architecture of the app. I thought we would have to do so by breaking Networking down in two modules (for reference: p91TBi-4rr-p2), but it turns out there might be a better way, which is...

1. Hardware exposes a public protocol, called CardReaderConfigProvider. This protocol declares one method, with an async API, that returns a token
2. There is a private class declared in CardPresentPaymentStore, called WCPayTokenProvider that implements RemoteConfigProvider. 
3. This class attacks a new Remote (WCPayRemote) to obtain the connection token
4. We pass an instance of WCPayTokenProvider to the CardReaderService (typed as CardReaderConfigProvider)
5. CardReaderService wraps that instance passed to it into an implementation of ConnectionTokenProvider, which is what the Stripe Terminal SDK expects. 

In a nutshell, what we do is encapsulate the call to WCPayRemote in an object that then we pass to the card reader service, which wraps that object into another object that implements the protocol that the Stripe Terminal SDK expects.

### Networking
* Add a new Remote to Networking, called WCPayRemote. This could (maybe should) be where we end up implementing all support for APIs specific to WCPay, so maybe there is a better name for it.
* Add a new model object to Networking, to model the connection token returned by WCPay
* Add a new Mapper to networking, to parse the response to requests to `payments/connection_tokens`

### Hardware
* Add a new public protocol, called CardReaderConfigProvider. Implementations of this protocol will figure out a way to obtain a token.
* Update the signature of the start() method in CardReaderService to pass an instance of CardReaderConfigProvider as a parameter.
* Update the implementation of the StripeCardReaderService to call Terminal.setTokenProvider when a call to start discovery is initiated.
* An implementation of ConnectionTokenProvider (the protocol required by the Stripe Terminal SDK) called DefaultConnectionTokenProvider. This class uses the RemoteConfigProvider passed to the service as a parameter to actually fetch the token.

### Yosemite
* Declare, in CardPresentPaymentStore.swift, a private class called WCPayTokenProvider that implements 
RemoteConfigProvider. This class attaches the new remote (WCPayRemote) to obtain a connection token. Why do we want to declare this implementation of RemoteConfigProvider here? Because the way the token is obtained is tightly coupled to the fact that we are using WCPay for it.

### UI
Just to have an entry point to the implementation, and test that it works  CardReaderSettingsViewModel will trigger the discovery, fetching a token, when tapping the "Connect card reader" button in Settings > manage Card readers

# Trade offs

Things I identify as upsides of this solution:
* We maintain the encapsulation of the integration with the Stripe SDK. Outside of Yosemite, no-one knows that we are dealing with Stripe.
* We do not need to break Networking down into two modules. This keeps things simple, and maintains symmetry between the integration of Card Present Payments and the rest of the codebase. There is value in removing "special cases"

Things I do not love:
* The solution is a bit complex. Not terribly complex, but it's not the simplest thing to explain.

## How to test
This requires better testing instructions before submitting as a proper PR, but for now:

* Checkout the branch
* Build and run. Navigate to Settings > Manage card readers > tap Connect Card Reader
* There should be a log in the console, immediately after tapping the button, similar to this:
```
==== new card readers discovered begins ====
new readers:  []
==== new card readers discovered ends   ====
```

After a few seconds, and a round trip to the backend and to the Terminal SDK, there should be a second trace like this:

```
==== new card readers discovered begins ====
new readers:  [Hardware.CardReader(serial: "WPE-SIMULATOR-1", vendorIdentifier: Optional("SIMULATOR"), name: Optional("Simulated WisePOS E"), status: Hardware.CardReaderStatus(connected: true, remembered: false), softwareVersion: Optional("0.0.0.1"), batteryLevel: nil, readerType: Hardware.CardReaderType.notSupported), Hardware.CardReader(serial: "P400-SIMULATOR-1", vendorIdentifier: Optional("SIMULATOR"), name: Optional("Simulated Verifone P400"), status: Hardware.CardReaderStatus(connected: true, remembered: false), softwareVersion: Optional("3.0.1.17"), batteryLevel: nil, readerType: Hardware.CardReaderType.counterTop)]
==== new card readers discovered ends   ====
```

Those two readers are provided by our integration, according to this sequence:
* Fetch a token from the backend
* Ask the Terminal SDK to start the discovery process, passing it a CardReaderNetworkingAdapter that the SDK can use to fetch the token.
* Parse the response from the SDK, transform it into model objects owned by us. 
* Propagate those objects up to the CardPresentPaymentStore
* Which in turn pushed them up to the UI layer.

Update release notes:


Hardware/Hardware/CardReader/CardReaderNetworkAdapter.swift
```diff
@@ -0,0 +1,7 @@
+/// Abstraction provided by Hardware so that clients of this library
+/// can model a way to provide a connection token.
+/// It is meant to abstract an implementation of the [adapter pattern](https://en.wikipedia.org/wiki/Adapter_pattern)
+/// This name is terrible. I can't think of a better one though
+public protocol CardReaderNetworkingAdapter {
+    func fetchToken(completion: @escaping(String?, Error?) -> Void)
```
allendav: The response returned by `wp-json/wc/v3/payments/connection_tokens` includes not just a token string (in `secret`, but also a `test_mode` boolean. We can just do with the token string for now, but we will probably want to capture that `test_mode` property as well in the future. Maybe open a separate issue for that.

As for a name, CRNA is fine, or perhaps CardReaderConnectionTokenProvider

Hardware/Hardware/CardReader/CardReaderService.swift
```diff
@@ -26,7 +26,7 @@ public protocol CardReaderService {
 
     /// Starts the service.
     /// That could imply, for example, that the reader discovery process starts
-    func start()
+    func start(_ adapter: CardReaderNetworkingAdapter)
```
allendav: I appreciate it is following the adapter pattern, but I wonder if tokenProvider might be a better argument name

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -62,13 +53,15 @@ extension StripeCardReaderService: CardReaderService {
 
     // MARK: - CardReaderService conformance. Commands
 
-    public func start() {
+    public func start(_ adapter: CardReaderNetworkingAdapter) {
```
allendav: Same comment as above

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -62,13 +53,15 @@ extension StripeCardReaderService: CardReaderService {
...
         // This is enough code to pass a unit test.
         // The final version of this method would be completely different.
         // But for now, we want to start the discovery process using the
         // simulate reader included in the Stripe Terminal SDK
         // https://stripe.com/docs/terminal/integration?country=CA&platform=ios&reader=p400#dev-test
 
+        setNetworkingAdapter(adapter)
```
allendav: Maybe setTokenProvider ?

Networking/Networking/Mapper/WCPayConnectionTokenMapper.swift
```diff
@@ -0,0 +1,26 @@
...
+        let decoder = JSONDecoder()
+
+        return try decoder.decode(WCPayConnectionTokenEnvelope.self, from: response).token
+    }
+}
+
+
+// WCPayConnectionTokenEnvelope Disposable Entity
+///
+/// `Load connection Token` endpoint returns the requested order refund document in the `data` key. This entity
```
allendav: copypasta

Hardware/Hardware/CardReader/CardReaderNetworkAdapter.swift
allendav: Nevermind `test_mode` I see it below

Networking/Networking/Remote/WCPayRemote.swift
```diff
@@ -0,0 +1,22 @@
...
+///
+public class WCPayRemote: Remote {
+
+    /// Loads a WCPay connection token for a given site ID and parses the rsponse
+    /// - Parameters:
+    ///   - siteID: Site for which we'll fetch the WCPay Connection token.
+    ///   - completion: Closure to be executed upon completion.
+    public func loadConnectionToken(for siteID: Int64,
+                                    completion: @escaping(WCPayConnectionToken?, Error?) -> Void) {
+        let path = "payments/connection_tokens"
```
allendav, line 14: Interesting! So we don't need to include wc/v3 here. TIL.

allendav:
Added a few comments above. Got expected results:

```
==== just for testing
discovered card readers == []
//// just for testing
2021-03-02 13:46:24.528660-0800 WooCommerce[41503:1444635] TracksService sendQueuedEvents completed. Sent 1 events.
==== just for testing
discovered card readers == [Hardware.CardReader(serial: "WPE-SIMULATOR-1", vendorIdentifier: Optional("SIMULATOR"), name: Optional("Simulated WisePOS E"), status: Hardware.CardReaderStatus(connected: true, remembered: false), softwareVersion: Optional("0.0.0.1"), batteryLevel: nil, readerType: Hardware.CardReaderType.notSupported), Hardware.CardReader(serial: "P400-SIMULATOR-1", vendorIdentifier: Optional("SIMULATOR"), name: Optional("Simulated Verifone P400"), status: Hardware.CardReaderStatus(connected: true, remembered: false), softwareVersion: Optional("3.0.1.17"), batteryLevel: nil, readerType: Hardware.CardReaderType.counterTop)]
//// just for testing
```

Hardware/Hardware/CardReader/CardReaderNetworkAdapter.swift
ctarda: 🤔 that's a good point. Maybe we should rename this to something that indicates that it provides "backend configuration parameters"? CardReaderRemoteConfigAdapter? That's too long for my liking, but maybe RemoteConfigAdapter? RemoteConfigProvider?

Hardware/Hardware/CardReader/CardReaderService.swift
ctarda: lol. Yes, I think provider would be better https://github.com/woocommerce/woocommerce-ios/pull/3772#discussion_r586361003

Networking/Networking/Mapper/WCPayConnectionTokenMapper.swift
ctarda: 😊

ctarda:
Thanks for taking a look @allendav 

My takeaway is that we might want to rename CardReaderNetworkingAdapter to RemoteConfigurationProvider (or similar) and rename all the parameters of that type as well. Thoughts?

allendav:
> My takeaway is that we might want to rename CardReaderNetworkingAdapter to RemoteConfigurationProvider (or similar) and rename all the parameters of that type as well. Thoughts?

Let's avoid Remote since that's a noun with a different meaning in our architecture already. How about CardReaderConfigProvider?

ctarda:
> Let's avoid Remote since that's a noun with a different meaning in our architecture already. How about CardReaderConfigProvider?

Updated!

allendav (approved):
Tests fine. Was able to trigger discovery successfully from the UI. Thanks for addressing the naming bits too - naming is hard :)

:shipit: after CircleCI completes

PR #3774, opened 2021-03-02
[Settings] Add new store and app settings sections, show store name instead of user

### LINKED ISSUE #3746: [Settings] Re-organize sections and rows
Light mode:

Dark mode:

Related Android issue: https://github.com/woocommerce/woocommerce-android/issues/3628

cc @Garance91540

allendav:
Closes #3746 

A figma file is available

Noteworthy changes:
- The selected store's name is shown in place of the URL. The URL is shown in place of the user account name.
- A new store settings section with a manage card reader row has been added (only visible with the card reader feature flag however)
- Help and support row has been moved with send feedback into a new Help & Feedback section
- A new App Settings section has been added. Privacy Settings has been moved into it.

Unchanged:
- About the App section
- Other section

Screenshots:

Light Mode:

Dark Mode:

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -24,6 +24,13 @@ final class SettingsViewController: UIViewController {
         return ServiceLocator.stores.sessionManager.defaultAccount?.displayName ?? String()
     }
 
+    /// Main Site's Name
+    ///
+    private var siteName: String {
+        let nameAsString = ServiceLocator.stores.sessionManager.defaultSite?.name as String?
+        return nameAsString ?? String()
+    }
+
     /// Main Site's URL
     ///
     private var siteUrl: String {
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -24,6 +24,13 @@ final class SettingsViewController: UIViewController {
         return ServiceLocator.stores.sessionManager.defaultAccount?.displayName ?? String()
     }
 
+    /// Main Site's Name
+    ///
+    private var siteName: String {
+        let nameAsString = ServiceLocator.stores.sessionManager.defaultSite?.name as String?
+        return nameAsString ?? String()
```
allendav, line 31: Concern: Should we provide a non-empty default?

ghost (changes requested):
This looks great @allendav it looks like we're missing the 16px padding from the heading to the first row when the select store section is hidden. Could we please add that in?

Screen Shot 2021-03-03 at 12 40 35 pm

Screen Shot 2021-03-03 at 12 44 46 pm

ealeksandrov:
@adamzelinski that is not a first row on a screenshot, but "scrolled to bottom" state. On some screen sizes "card reader" row will be positioned right on top.

Here's whole screen on iPhone 12:

https://user-images.githubusercontent.com/3132438/109780686-47c3b680-7c18-11eb-8b4a-b186ea3b7b09.mp4

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
ealeksandrov: But what can be the reason for missing `defaultSite`? If it's not a common case - empty title is fine.

ealeksandrov (approved):
LGTM :shipit:

allendav:
@adamzelinski wrote:

> This looks great @allendav it looks like we're missing the 16px padding from the heading to the first row when the select store section is hidden. Could we please add that in?

That's - as @ealeksandrov mentioned - just a result of me scrolling down for the 2nd screengrab so you could see the bottom rows

Edit: If that makes sense, I'll merge after you've approved. Thanks!

ghost (approved):

ghost:
Sorry about that @allendav looks great!

PR #3775, opened 2021-03-03
[Mobile payments] Add fancy alert for scanning for card readers

allendav:
Work in progress. Replaces the iOS Alert displayed when scanning for card readers with a richer experience per the design

A figma file is available.

Update 2021-Mar-5 @Garance91540 @adamzelinski @ctarda - I finished hooking this up only to be disappointed :P The FancyAlertViewController available from WordPressUI (and which we use a little bit elsewhere in the app) doesn't really match what y'all would like:

alert-side-by-side

I am thinking the views are sufficiently different in order and style that we might want a separate view for these. What do you think @ctarda ? Here's a snippet of the various alerts from figma:

Screen Shot 2021-03-05 at 4 05 38 PM

Update release notes:


ghost (approved):

WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -58,6 +58,12 @@ extension UIImage {
             .applyTintColor(.placeholderImage)!
     }
 
+    /// Card Reader Scanning Image
+    ///
+    static var cardReaderScanningImage: UIImage {
```
ctarda, line 63: We usually add a unit test to cover images. I have personally been bitten in the past by runtime crashes due to assets being missed or typos in the asset name. 

You can look at the existing tests in IconsTests

ctarda (commented):
It's looking great! My only comment would be about adding a test to ensure the icon exists in the bundle.

👏

WooCommerce/Classes/Extensions/UIImage+Woo.swift
allendav: Thanks! Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3775/commits/091b86d65eb8063e61c236a40c9a887d595efac1

allendav:
@Garance91540 @adamzelinski @ctarda - please see update in description above. I don't think the FancyAlerts from WordPressUI are going to work so I think I'll be rolling our own for this. What do y'all think?

Garance91540:
> so I think I'll be rolling our own for this. What do y'all think?

Sounds good to me! With have so many of those coming out for this project (including the tutorial dialogs) that I think it definitely will be worth it

ctarda:
@Garance91540 @allendav @adamzelinski  Sorry if this has been already discussed, but could we continue moving forward **for now** with the style provided by the FancyAlertViewController?

Rolling our own dialogs, specially if we want to support the tutorials is not a trivial task. Could we implement those new dialogs+tutorials in the next iteration of the prototype?

Garance91540:
Hi Cesar! 

> @Garance91540 @allendav @adamzelinski Sorry if this has been already discussed, but could we continue moving forward for now with the style provided by the FancyAlertViewController?

sounds good. Two questions regarding the FancyAlertViewController, 
- is it possible to make the scrim screen a bit darker so the background screen isn't blending with the dialog? 
- is it possible to remove the divider between the dialog text and the CTA?
visual here: https://d.pr/i/pfYwO6

ctarda:
👋  Garance!

> sounds good. 
👍 We could prioritize this (tutorials and re-styling of modal alerts) to be the very first thing we do right after we have a first iteration of the UI for collecting a payment in place. I think that could be ideally in two or three sprint's time, but @allendav is the best position to know how far we might be from that.

> is it possible to make the scrim screen a bit darker so the background screen isn't blending with the dialog?

I don't think it is, at least not out of the box.

> is it possible to remove the divider between the dialog text and the CTA?
visual here: https://d.pr/i/pfYwO6

Yes, that is possible. It does reduce the vertical spacing a lot though, but I guess we could workaround that:

ctarda:
@allendav I am not sure of the status of this PR. Are you waiting for me to give a 👍 ?

Garance91540:
Hi @ctarda 👋

I just saw this PR again, and on the last screenshot you shared, the part of the modal with text + CTA looks good, but we are missing top and bottom vertical padding on the image section: https://d.pr/i/Bq5D6r

allendav:
Closing without merging. Will resume alerts on a fresh branch next week.

allendav:
> I am not sure of the status of this PR. Are you waiting for me to give a 👍 

Sorry I missed this. We'll revisit these UI changes in a new PR/new branch later since out of the box Fancy Alerts aren't quite what we need, but we have a lot of Fancy Alert logic throughout our view controllers in our app. I'll summarize in a post shortly so we can come to a decision.

PR #3781, opened 2021-03-03
Variations: Update attribute name remotely after rename

rachelmcr:
Related: #3212 

## Description

This is a followup to https://github.com/woocommerce/woocommerce-ios/pull/3768, to perform the remote update after a variation attribute is renamed. After renaming an attribute and tapping the "Done" button, the previous (attribute options) screen reflects the new name in the navigation bar. Tapping "Next" performs the remote update for the attribute (the new attribute name and any changes to the attribute options).

This PR also limits renaming attributes to only local attributes, to reflect the behavior in WP Admin. (Global attributes can't be renamed while editing an individual product.)

## Changes

* Adds a closure to `RenameAttributesViewController`, invoked when the Done button is tapped.
* Adds `allowsRename` to `AddAttributeOptionsViewModel` to check if the attribute is an existing local attribute. This is used to control whether the Rename menu item appears in the more (...) menu.
* Changes `attributeName` to `currentAttributeName` in `AddAttributeOptionsViewModel`. Now, `currentAttributeName` is set to the attribute `name` by default, and set to the new attribute name after the rename. That is used to set the navigation bar title, check whether the Next button should be enabled, and update the attribute name when the Next button is tapped.
* Adds unit tests.

## Testing

1. Make sure you have a variable product with both local and global attributes.
2. Go to the Products tab.
3. Select a variable product.
4. Select Variations to open the variations list.
5. Tap the ... menu and select Edit Attributes.
6. Select a global attribute to open the Attribute Options screen.
7. Tap the ... menu and confirm there is no Rename option.
8. Go back to the Edit Attributes screen.
9. Select a local attribute to open the Attribute Options screen.
10. Tap the ... menu and select Rename.
11. On the Rename Attribute screen, enter a new name for the attribute and tap Done.
12. Confirm you are returned to the Attribute Options screen, with the new name in the navigation bar and the Next button enabled.
13. Tap the Next button and confirm the new attribute name is saved remotely.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -149,7 +149,7 @@ private extension AddAttributeOptionsViewController {
     }
 
     func renderViewModel() {
-        title = viewModel.attributeName
+        title = viewModel.currentAttributeName
         configureRightButtonItem()
         tableView.reloadData()
 
@@ -419,8 +419,13 @@ extension AddAttributeOptionsViewController {
     /// Navigates to `RenameAttributesViewController`
     ///
     private func navigateToRenameAttribute() {
-        let viewModel = RenameAttributesViewModel(attributeName: self.viewModel.attributeName)
-        let renameAttributeViewController = RenameAttributesViewController(viewModel: viewModel)
+        let viewModel = RenameAttributesViewModel(attributeName: self.viewModel.currentAttributeName)
+        let renameAttributeViewController = RenameAttributesViewController(viewModel: viewModel) { [weak self] updatedAttributeName in
+            // Sets new attribute name and refreshes the view to reflect the change
+            self?.viewModel.currentAttributeName = updatedAttributeName
+            self?.renderViewModel()
+            self?.navigationController?.popViewController(animated: true)
+        }
         show(renameAttributeViewController, sender: nil)
     }
 
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -149,7 +149,7 @@ private extension AddAttributeOptionsViewController {
     }
 
     func renderViewModel() {
-        title = viewModel.currentAttributeName
+        title = viewModel.getCurrentAttributeName()
         configureRightButtonItem()
         tableView.reloadData()
 
@@ -421,11 +421,10 @@ extension AddAttributeOptionsViewController {
     /// Navigates to `RenameAttributesViewController`
     ///
     private func navigateToRenameAttribute() {
-        let viewModel = RenameAttributesViewModel(attributeName: self.viewModel.currentAttributeName)
+        let viewModel = RenameAttributesViewModel(attributeName: self.viewModel.getCurrentAttributeName())
         let renameAttributeViewController = RenameAttributesViewController(viewModel: viewModel) { [weak self] updatedAttributeName in
-            // Sets new attribute name and refreshes the view to reflect the change
+            // Sets new attribute name
             self?.viewModel.setCurrentAttributeName(updatedAttributeName)
-            self?.renderViewModel()
             self?.navigationController?.popViewController(animated: true)
         }
         show(renameAttributeViewController, sender: nil)
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
@@ -149,7 +149,7 @@ private extension AddAttributeOptionsViewController {
     }
 
     func renderViewModel() {
-        title = viewModel.getCurrentAttributeName()
+        title = viewModel.getCurrentAttributeName
         configureRightButtonItem()
         tableView.reloadData()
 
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -53,20 +53,14 @@ final class AddAttributeOptionsViewModel {
...
     ///
-    var attributeName: String {
-        switch attribute {
-        case .new(let name):
-            return name
-        case .existing(let attribute):
-            return attribute.name
-        }
-    }
+    private(set) var currentAttributeName: String
```
Ecarrion: Since this is now part of the state of the mutable state of the view model, I think you can move it inside the state struct. This would allow us to remove the `self?.renderViewModel()` that you added in the `navigateToRenameAttribute` method to make sure that the `VC` is always rendering the latest `VM` state.

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewController.swift
```diff
@@ -397,10 +397,12 @@ extension AddAttributeOptionsViewController {
         let actionSheet = UIAlertController(title: nil, message: nil, preferredStyle: .actionSheet)
         actionSheet.view.tintColor = .text
 
-        let renameAction = UIAlertAction(title: Localization.renameAction, style: .default) { [weak self] _ in
-            self?.navigateToRenameAttribute()
+        if viewModel.allowsRename {
```
Ecarrion, line 400: Nice!

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -93,6 +87,17 @@ final class AddAttributeOptionsViewModel {
...
 
+    /// Defines if the attribute can be renamed (existing local attributes only)
+    ///
+    var allowsRename: Bool {
+        switch attribute {
+        case .new:
+            return false
+        case let .existing(productAttribute):
+            return productAttribute.isLocal
+        }
```
Ecarrion, line 98: Beatiful!

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -285,7 +297,7 @@ extension AddAttributeOptionsViewModel {
         // Name has to be considered, because local attributes are zero-id based.
         let updatedAttributes: [ProductAttribute] = {
             var attributes = product.attributes
-            attributes.removeAll { $0.attributeID == newAttribute.attributeID && $0.name == newAttribute.name }
+            attributes.removeAll { $0.attributeID == attribute.attributeID && $0.name == attribute.name }
```
Ecarrion, line 308: Tricky! It's probably worth it to add a comment explaining that we use `attribute` and not `newAttribute` to make sure the old name is the one that is matched, because the new name could not match!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewController.swift
```diff
@@ -81,9 +85,10 @@ private extension RenameAttributesViewController {
 
 extension RenameAttributesViewController {
     @objc private func doneButtonTapped() {
-        // TODO: Update the attribute name if it has changed
-        // For now, just navigate back
-        navigationController?.popViewController(animated: true)
+        guard let newAttributeName = viewModel.newAttributeName else {
+            return onCompletion(viewModel.attributeName)
+        }
+        onCompletion(newAttributeName)
```
Ecarrion: I think this can be reduced to
```swift
onCompletion(viewModel.newAttributeName ?? viewModel.attributeName)
```
or rename `viewModel.attributeName` to `viewModel.originalAttributeName` and then create a new `viewModel.attributeName`(that you can unit test) that has that merge operation.
```swift
onCompletion(viewModel.attributeName)
```

Ecarrion (commented):
This is looking very good, just left you a few suggestions! Let me know what you think!

WooCommerce/Classes/ViewRelated/Products/Variations/Edit Attributes/RenameAttributesViewController.swift
rachelmcr: Ah, yes! I thought this approach seemed unnecessarily complicated but I forgot I could just use `??` here.

But I really like the idea of unit testing this and having `viewModel.attributeName` separate from `viewModel.originalAttributeName` for clarity, so I've gone with that suggestion. I made that change in https://github.com/woocommerce/woocommerce-ios/pull/3781/commits/4e04677777b8adf4ee0bf028266dfe999b08b7c3.

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
rachelmcr: Comment added in https://github.com/woocommerce/woocommerce-ios/pull/3781/commits/d8f7de5cafd75a2e4c06ca34a267ed2b08d757bb.

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
rachelmcr: Oh, good point, thanks! I made that change in https://github.com/woocommerce/woocommerce-ios/pull/3781/commits/d061bcbc4383f0eee553c0db753c93a52fa312ff. The changes work, but let me know if there are better conventions for that (especially things like initializing `currentAttributeName` and the setter/getter methods since the state struct is entirely private).

rachelmcr:
Thanks for the feedback @Ecarrion! This is ready for another look.

WooCommerce/Classes/ViewRelated/Products/Variations/Add Attributes/AddAttributeOptionsViewModel.swift
```diff
@@ -227,6 +233,18 @@ extension AddAttributeOptionsViewModel {
...
+    func setCurrentAttributeName(_ newName: String) {
+        state.currentAttributeName = newName
+    }
+
+    /// Gets the current attribute name
+    ///
+    func getCurrentAttributeName() -> String {
+        return state.currentAttributeName
+    }
+
```
Ecarrion: I think this could be modeled better as a computed property like `allowsRename` since we are not performing any transformation but just accessing a value!

Ecarrion (approved):
Code and tests looks 🚀 

Just a small nit!

Issue #3744, opened 2021-02-25
[Mobile Payments] Inject CardReaderSettingsViewModel, Add Unit Tests


allendav:
Feedback on https://github.com/woocommerce/woocommerce-ios/pull/3699

allendav:
Resolved by https://github.com/woocommerce/woocommerce-ios/issues/4051

PR #3784, opened 2021-03-03
[Mobile Payments] Connect card reader discovery completion to UI

allendav:
Ready for review.

Note: This does not change the UI. This only replaces the timer based simulations with actual service data.

To test:

- Settings
- Manage card readers
- Connect card reader
- Scanning for readers
- Found reader: Connect to reader
- Observe Connected Reader
- Disconnect

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
@@ -3,6 +3,20 @@ import Yosemite
 
 final class CardReaderSettingsConnectedReaderView: NSObject {
 
+    private enum DefaultStrings: String {
+        case noSerialNumber
+
+        var userFacingString: String {
+            switch self {
+            case .noSerialNumber:
+                return NSLocalizedString(
+                    "Unknown",
+                    comment: "Settings > Manage Card Reader > Connected Reader > Displayed for card readers without a serial number"
+                )
+            }
+        }
+    }
+
     /// A simple model for this "ViewModel" - just a reference to the CardReaderSettingsViewModel connected reader
     var connectedReader: CardReader?
 
@@ -47,11 +61,11 @@ final class CardReaderSettingsConnectedReaderView: NSObject {
 
         let batteryLabelFormat = NSLocalizedString(
             "%1$@%% Battery",
-            comment: "Settings > Manage Card Reader > Connected Reader >> Battery level as a percentage"
+            comment: "Settings > Manage Card Reader > Connected Reader > Battery level as a percentage"
         )
 
         cell.batteryLevelLabel?.text = String.localizedStringWithFormat(batteryLabelFormat, batteryLevelString)
-        cell.serialNumberLabel?.text = connectedReader?.serial ?? "Unknown"
+        cell.serialNumberLabel?.text = connectedReader?.serial ?? DefaultStrings.noSerialNumber.userFacingString
         cell.selectionStyle = .none
     }
 
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
@@ -2,24 +2,7 @@ import UIKit
 import Yosemite
 
 final class CardReaderSettingsConnectedReaderView: NSObject {
-
-    private enum DefaultStrings: String {
-        case noSerialNumber
-
-        var userFacingString: String {
-            switch self {
-            case .noSerialNumber:
-                return NSLocalizedString(
-                    "Unknown",
-                    comment: "Settings > Manage Card Reader > Connected Reader > Displayed for card readers without a serial number"
-                )
-            }
-        }
-    }
-
-    /// A simple model for this "ViewModel" - just a reference to the CardReaderSettingsViewModel connected reader
-    var connectedReader: CardReader?
-
+    var viewModel = CardReaderViewModel()
     var onPressedDisconnect: (() -> ())?
 
     private var rows = [Row]()
@@ -39,10 +22,6 @@ final class CardReaderSettingsConnectedReaderView: NSObject {
         ]
     }
 
-    public func update(reader: CardReader) {
-        connectedReader = reader
-    }
-
     private func configure(_ cell: UITableViewCell, for row: Row, at indexPath: IndexPath) {
         switch cell {
         case let cell as ConnectedReaderTableViewCell where row == .connectedReader:
@@ -55,17 +34,8 @@ final class CardReaderSettingsConnectedReaderView: NSObject {
     }
 
     private func configureConnectedReader(cell: ConnectedReaderTableViewCell) {
-        let batteryLevel = connectedReader?.batteryLevel ?? 1.0
-        let batteryLevelPercent = Int(100 * batteryLevel)
-        let batteryLevelString = NumberFormatter.localizedString(from: batteryLevelPercent as NSNumber, number: .decimal)
-
-        let batteryLabelFormat = NSLocalizedString(
-            "%1$@%% Battery",
-            comment: "Settings > Manage Card Reader > Connected Reader > Battery level as a percentage"
-        )
-
-        cell.batteryLevelLabel?.text = String.localizedStringWithFormat(batteryLabelFormat, batteryLevelString)
-        cell.serialNumberLabel?.text = connectedReader?.serial ?? DefaultStrings.noSerialNumber.userFacingString
+        cell.serialNumberLabel?.text = viewModel.displayName
+        cell.batteryLevelLabel?.text = viewModel.batteryStatus
         cell.selectionStyle = .none
     }
 
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
@@ -1,4 +1,5 @@
 import UIKit
+import Yosemite
 
 final class CardReaderSettingsConnectedReaderView: NSObject {
 
@@ -50,7 +51,7 @@ final class CardReaderSettingsConnectedReaderView: NSObject {
         )
 
         cell.batteryLevelLabel?.text = String.localizedStringWithFormat(batteryLabelFormat, batteryLevelString)
-        cell.serialNumberLabel?.text = connectedReader?.serialNumber ?? "Unknown"
+        cell.serialNumberLabel?.text = connectedReader?.serial ?? "Unknown"
         cell.selectionStyle = .none
     }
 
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
```diff
@@ -50,7 +51,7 @@ final class CardReaderSettingsConnectedReaderView: NSObject {
         )
 
         cell.batteryLevelLabel?.text = String.localizedStringWithFormat(batteryLabelFormat, batteryLevelString)
-        cell.serialNumberLabel?.text = connectedReader?.serialNumber ?? "Unknown"
+        cell.serialNumberLabel?.text = connectedReader?.serial ?? "Unknown"
```
ctarda: I presume you are already aware, because I've seen you do it in previous PRs, but just in case, all user facing strings need to be localized (in this case "Unknown". 

We usually declare those as static properties to a private enumeration (an enumeration so that it can not be instantiated)

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -129,7 +130,7 @@ final class CardReaderSettingsViewController: UIViewController {
 
     private func addFoundReaderModal() {
         // TODO Use FancyAlert instead
-        let foundReaderName = self.viewmodel.foundReader?.serialNumber ?? ""
+        let foundReaderName = self.viewmodel.foundReader?.serial ?? ""
```
ctarda: Just a though, not a suggestion: this logic (returning the serial or an empty string/placeholder) could be pushed down to the view model?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -2,11 +2,6 @@ import Foundation
 import Combine
 import Yosemite
 
-struct CardReader {
```
ctarda, line 5: Yay! Integrate all the things!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -44,63 +36,43 @@ final class CardReaderSettingsViewModel: ObservableObject {
...
-        timer = Timer.scheduledTimer(timeInterval: 3, target: self, selector: #selector(dummyFoundReader), userInfo: nil, repeats: false)
     }
 
-    // TODO Remove
-    @objc private func dummyFoundReader() {
-        foundReader = CardReader(serialNumber: "CHB204909001234", batteryLevel: 0.885)
-        activeAlert = .foundReader
-    }
+    private func discoveredReader(cardReaders: [CardReader]) -> Void {
+        if activeAlert == .foundReader {
```
ctarda: The usual (swiftly) idiom for early returns is `guard`. To word this differently, I think most iOS engineers would expect this to be:
`guard activeAlert != .foundReader else { return }`

Personally, I am not sure that's actually more readable, but I guess it's kind of the standard?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: OMG absolutely!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3784/commits/e8747e94d0a363bd4f5ddbb4e0b0d8f502979884

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3784/commits/b7384077caaf1454e3c5bd58550704a402817399 (among other things)

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
allendav: So, I took this suggestion a step further, since I'll need these strings elsewhere too (e.g. the fancy alerts) and I now have CardReaderViewModels for connected and found readers created by CardReaderSettingsViewModel - see changes in https://github.com/woocommerce/woocommerce-ios/pull/3784/commits/b7384077caaf1454e3c5bd58550704a402817399

Thoughts? I think it is much cleaner this way.

allendav:
@ctarda - in https://github.com/woocommerce/woocommerce-ios/pull/3784/commits/13a67ef068984dc0297263d7c4dc8b6f55a3d95e I've added dispatching the connect action and it works nicely. I did notice the P400 simulator is returning a nil batteryLevel (which makes sense since the P400 doesn't have a battery.)

Screen Shot 2021-03-04 at 3 59 48 PM

@Garance91540 - how about we display "Ready" or "Powered On" there for P400s instead of a battery status?



Also @ctarda - FYI - as you can see above - the very first time I connected, I got a location services prompt. I expect I'll get it again if I clean my simulator.

Lastly - @ctarda - I'd like to pick up the changes in #3788 once that merges - you added some enum helpers there that I'd like to use

ctarda:
@allendav  Yes, the simulator will ask for permissions the first time, and only the fist time as far as I can tell, you attempt to connect to a reader. 

Erasing All Content and Settings in the simulator will reset that and the next time you attempt a pairing permissions will be requested again. 

I have the gut feeling #3788  is going to take a [while to merge](https://github.com/woocommerce/woocommerce-ios/pull/3788#issuecomment-791406775), so if it is blocking you, can you defer using those enums you mentioned to a separate PR so that we can get this one in?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -1,12 +1,13 @@
...
 import Combine
+import Yosemite
 
 final class CardReaderSettingsViewController: UIViewController {
 
     @IBOutlet weak var tableView: UITableView!
 
     private var subscriptions = Set<AnyCancellable>()
-    private var viewmodel = CardReaderSettingsViewModel()
+    private var viewModel = CardReaderSettingsViewModel()
```
allendav, line 10: We'll inject this VM in a subsequent PR

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
```diff
@@ -128,19 +129,19 @@ final class CardReaderSettingsViewController: UIViewController {
     }
 
     private func addFoundReaderModal() {
-        // TODO Use FancyAlert instead
-        let foundReaderName = self.viewmodel.foundReader?.serialNumber ?? ""
+        // TODO Use FancyAlert instead - all these strings will be moved there
+        let foundReaderName = viewModel.foundReadersViewModels[0].displayName
```
allendav, line 133: We'll handle multiple found readers in a subsequent PR

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -24,36 +19,82 @@ enum CardReaderSettingsViewActiveAlert {
     case updating
 }
 
+struct CardReaderViewModel {
```
allendav, line 22: This public struct may prove useful for other places, like alerts, so that we are only formatting display names and battery status strings in one place.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -24,36 +19,82 @@ enum CardReaderSettingsViewActiveAlert {
...
+    @Published var foundReadersViewModels: [CardReaderViewModel]
 
-    var knownReaders: [CardReader]
-    var foundReader: CardReader?
-
-    var timer: Timer? // TODO Remove
-    var batteryTimer: Timer? // TODO Remove
+    private var connectedReader: CardReader? {
+        didSet {
+            updateConnectedReaderViewModel()
```
allendav, line 69: Update the connectedReader, get a published update to the VM. Same thing below for foundReaders

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -62,80 +103,84 @@ final class CardReaderSettingsViewModel: ObservableObject {
...
     }
 
+    /// Views can call this method to connect to a found reader.
     func connect() {
-        // TODO dispatch an action to connect.
         activeAlert = .connecting
 
-        // TODO Remove - simulates connecting with a timer
-        timer = Timer.scheduledTimer(timeInterval: 3, target: self, selector: #selector(dummyConnectedReader), userInfo: nil, repeats: false)
+        let action = CardPresentPaymentAction.connect(reader: foundReaders[0], onCompletion: { [weak self] result in
```
allendav, line 143: When we handle more than one found reader, we'll need to update this. We'll need to decide what to pass to connect to "pick" the right reader.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
```diff
@@ -1,10 +1,8 @@
 import UIKit
+import Yosemite
 
 final class CardReaderSettingsConnectedReaderView: NSObject {
-
-    /// A simple model for this "ViewModel" - just a reference to the CardReaderSettingsViewModel connected reader
-    var connectedReader: CardReader?
-
+    var viewModel = CardReaderViewModel()
```
ctarda, line 5: Sorry. I missed this in previous passes. If the viewModel is not going to be modified, it's probably better if it's declared as `let` or even better `private let`

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewController.swift
ctarda: same comment here about declaring it as let instead of var

ctarda (approved):
Works great. 

:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsConnectedReaderView.swift
allendav: Hmmmm. Good point! I want to change all the ViewModels to be passed in and will address this in #3744

Issue #3741, opened 2021-02-25
[Mobile Payments] Error handling. Catch Service errors in Yosemite


ctarda:
The commands in the API of CardReaderService return Promises that can fail with an error.

We are ignoring those errors for now. But we need to:
* Instantiate domain error objects specific to Yosemite
* Wrap the errors thrown by the service into those Yomseite errors (i.e. as underlying error)
* Propagate the Yosemite errors up to the UI

PR #3788, opened 2021-03-04
[Mobile Payments] Cancel discovery mark II: Yosemite (and a bit of Hardware)

### LINKED ISSUE #3742: [Mobile Payments] Yosemite: Cancel reader discovery
The CardReaderService already provides support for cancelling the discovery process.

Scope:
* Implement support for cancellation in Yosemite

ctarda:
Closes #3742 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by implementing logic in Yosemite to cancel reader discovery. This is not strictly necessary to move on to the next step (processing a payment) but it's nice to provide now for UI integration.

## Caveats
* We are not doing any error handling yet (there #3734  and #3741 for that), and everything happens in the happy place
* It could be argued that there is a bit of redundancy in the API of the service, in the way the discovery status is exposed. We are still exploring where this is going, so I would suggest not considering that as final at this point.
* This is still a bit difficult to test.

## Changes
* Added logic to StripeCardReaderService to manage the state of the discovery status. This is the most basic, barebones implementation of the state handling. I am not 100% sure that we end up exposing the discovery status, so I am not sure there is a point in complicating this more now.
* Added an action to CardPresentPaymentAction
* Added logic to back that action in CardPresentPaymentStore
* A minimal update to the UI to provide a manual test.

### The easy way:
* checkout the branch, run `bundle exec pod install`.
* look at the code, make sure the tests in CardPresentPaymentStore pass (in fact, a run of the full test suit might be a better approach)

### The manual way (requires configuring your test site)
Eventually (soon), this would be the only way to test any new PRs related to card present payments.
* Make sure your site is setup for testing payments. There is some information [in the documentation about running integration tests](https://github.com/woocommerce/woocommerce-ios/blob/feature/stripe-terminal-sdk-integration/docs/stripe-tests.md#integration-tests) and a more specific guide in P91TBi-4BH-p2)
* Navigate to settings > Manage Card Readers > tap "Connect Card Reader"
* Look at the console. You should see a trace like this in the console:

```
==== new card readers discovered begins ====
new readers:  []
==== new card readers discovered ends   ====
===== cancellation begins ====
new status  discovering
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
```

What is happening? First, when reader discovery begins, we receive an empty collection of readers (there have been zero readers discovered). Right after that, the service changes status to "discovering", until the cancellation command is completed (the Stripe SDK cancels discovery asynchronously). That's when the service changes status to idle. This change happens twice, as expected [according to the documentation](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(im)discoverReaders:delegate:completion:), as there are two callbacks being called (the one when cancellation is completed and the one corresponding to the discovery delegate)

I understand these testing instructions are not great, but it will be much easier to test anything related to payments very soon.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -75,13 +75,13 @@ extension StripeCardReaderService: CardReaderService {
          *
          *Note that if discoverReaders is canceled, the completion block will be called with nil (rather than an SCPErrorCanceled error).
          */
-        discoveryCancellable = Terminal.shared.discoverReaders(config, delegate: self, completion: { error in
+        discoveryCancellable = Terminal.shared.discoverReaders(config, delegate: self, completion: { [weak self] error in
             guard let error = error else {
-                self.switchStatusToIdle()
-                return
+                return self?.switchStatusToIdle()
+
             }
 
-            self.internalError(error)
+            self?.internalError(error)
         })
     }
 
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -9,7 +9,7 @@ public final class StripeCardReaderService: NSObject {
     private let discoveredReadersSubject = CurrentValueSubject<[CardReader], Never>([])
     private let connectedReadersSubject = CurrentValueSubject<[CardReader], Never>([])
     private let serviceStatusSubject = CurrentValueSubject<CardReaderServiceStatus, Never>(.ready)
-    private let discoveryStatusSubject = CurrentValueSubject<CardReaderServiceDiscoveryStatus, Never>(.idle)
+    private let discoveryStatusSubject = PassthroughSubject<CardReaderServiceDiscoveryStatus, Never>()
     private let paymentStatusSubject = CurrentValueSubject<PaymentStatus, Never>(.notReady)
     private let readerEventsSubject = PassthroughSubject<CardReaderEvent, Never>()
 
@@ -17,6 +17,8 @@ public final class StripeCardReaderService: NSObject {
     /// see
     ///  https://stripe.dev/stripe-terminal-ios/docs/Protocols/SCPDiscoveryDelegate.html#/c:objc(pl)SCPDiscoveryDelegate(im)terminal:didUpdateDiscoveredReaders:
     private let discoveredStripeReadersCache = StripeCardReaderDiscoveryCache()
+
+    private var isDiscovering = false
 }
 
 
@@ -37,7 +39,7 @@ extension StripeCardReaderService: CardReaderService {
     }
 
     public var discoveryStatus: AnyPublisher<CardReaderServiceDiscoveryStatus, Never> {
-        discoveryStatusSubject.eraseToAnyPublisher()
+        discoveryStatusSubject.removeDuplicates().eraseToAnyPublisher()
     }
 
     /// The Publisher that emits the payment status
@@ -68,6 +70,7 @@ extension StripeCardReaderService: CardReaderService {
             simulated: true
         )
 
+        print("|||| Service. calling start in service ", self)
         switchStatusToDiscovering()
 
         /**
@@ -96,10 +99,10 @@ extension StripeCardReaderService: CardReaderService {
          * discovering mode before attempting a cancellation
          *
          */
-        guard discoveryStatusSubject.value == .discovering else {
+        guard isDiscovering else {
             return
         }
-
+        print("||||||| Service calling cancel in service", self)
         discoveryCancellable?.cancel { [weak self] error in
             guard let error = error else {
                 self?.switchStatusToIdle()
@@ -238,14 +241,17 @@ private extension StripeCardReaderService {
 private extension StripeCardReaderService {
     func switchStatusToIdle() {
         updateDiscoveryStatus(to: .idle)
+        isDiscovering = false
     }
 
     func switchStatusToDiscovering() {
         updateDiscoveryStatus(to: .discovering)
+        isDiscovering = true
     }
 
     func switchStatusToFault() {
         updateDiscoveryStatus(to: .fault)
+        isDiscovering = false
     }
 
     func updateDiscoveryStatus(to newStatus: CardReaderServiceDiscoveryStatus) {
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -77,8 +77,8 @@ extension StripeCardReaderService: CardReaderService {
          */
         discoveryCancellable = Terminal.shared.discoverReaders(config, delegate: self, completion: { [weak self] error in
             guard let error = error else {
-                return self?.switchStatusToIdle()
-
+                self?.switchStatusToIdle()
+                return
             }
 
             self?.internalError(error)
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -9,7 +9,7 @@ public final class StripeCardReaderService: NSObject {
     private let discoveredReadersSubject = CurrentValueSubject<[CardReader], Never>([])
     private let connectedReadersSubject = CurrentValueSubject<[CardReader], Never>([])
     private let serviceStatusSubject = CurrentValueSubject<CardReaderServiceStatus, Never>(.ready)
-    private let discoveryStatusSubject = PassthroughSubject<CardReaderServiceDiscoveryStatus, Never>()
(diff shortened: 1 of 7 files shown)


Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -68,20 +68,45 @@ extension StripeCardReaderService: CardReaderService {
...
+         *Note that if discoverReaders is canceled, the completion block will be called with nil (rather than an SCPErrorCanceled error).
+         */
         discoveryCancellable = Terminal.shared.discoverReaders(config, delegate: self, completion: { error in
-            if let error = error {
-                print("discoverReaders failed: \(error)")
-            } else {
-                print("discoverReaders succeeded")
+            guard let error = error else {
+                self.switchStatusToIdle()
+                return
```
Ecarrion: nit: I think we can do here `return self. switchStatusToIdle()`

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -68,20 +68,45 @@ extension StripeCardReaderService: CardReaderService {
...
             }
+
+            self.internalError(error)
```
Ecarrion: sanity check: Is it ok for us to capture `self` here and in the guard above?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -59,15 +59,25 @@ final class CardReaderSettingsViewModel: ObservableObject {
...
+
+        let discoveryCancellationAction = CardPresentPaymentAction.cancelCardReaderDiscovery { status in
+            // TODO. To be implemented
+            // Leaving these prints here to help test the PR
+            print("===== cancellation begins ====")
+            print("new status ", status)
+            print("===== cancellation ends ====")
+        }
+
+        ServiceLocator.stores.dispatch(discoveryCancellationAction)
```
Ecarrion, line 83: Is the intention here to launch the discovery action and immediately cancel the action? or am I missing something?

Yosemite/YosemiteTests/Mocks/CardPresentPayments/MockCardReaderService.swift
```diff
@@ -41,10 +43,18 @@ final class MockCardReaderService: CardReaderService {
     func start(_ configProvider: CardReaderConfigProvider) {
         didHitStart = true
         didReceiveAConfigurationProvider = true
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {[weak self] in
```
Ecarrion, line 52: Is this required or just for testing, a comment will help to new-comers like me!

Yosemite/YosemiteTests/Mocks/CardPresentPayments/MockCardReaderService.swift
```diff
@@ -29,9 +29,11 @@ final class MockCardReaderService: CardReaderService {
     }
 
     var didHitStart = false
+    var didHitCancel = false
```
Ecarrion: Maybe a comment on how these would be used is also good!

Ecarrion (dismissed):
Left you some comments. mainly because I haven't reviewed PRs for this project in a while!
@allendav feel free to take another look!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
ctarda: This is just a completely made up flow to start the discovery process and trigger a cancellation right away.

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
ctarda: I'll weakify self, which, as far as I understand, will make the return in one line mentioned above not possible

ctarda:
Thanks for the review @Ecarrion I just pushed an update to clarify your concerns adding some comments. 

@allendav this one seems to be all yours now

allendav (commented):
Tests fine with one concern - if you click on Keep Searching in the dummy flow:

keep-searching

you will see what appears to be double firing of cancellation action callbacks:

```
===== cancellation begins ====
new status  discovering
===== cancellation ends ====
===== cancellation begins ====
new status  discovering
===== cancellation ends ====
===== cancellation begins ====
new status  discovering
===== cancellation ends ====
==== new card readers discovered begins ====
new readers:  []
==== new card readers discovered ends   ====
===== cancellation begins ====
new status  discovering
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
===== cancellation begins ====
new status  idle
===== cancellation ends ====
```

each time you Keep Searching, it appears to double again

ctarda:
@allendav 
> each time you XXXXXX, it appears to double again

Good catch! 

That (things breaking the second time you do them) usually points to a memory leak. In this case though, there does not seem to be any leaks. 

Adding a breakpoint or a trace to CardPaymentStore.startCardReaderDiscovery and CardPaymentStore.cancelCardReaderDiscovery proves that those methods are called only once. So the next step is looking into the service. Do we really have one instance of it or do we have more than one? Is the issue really a memory leak, or are the publishers dispatching values more than once?

Adding a breakpoint (or another print) in StripeCardReaderService.start and StripeCardReaderService.cancelDiscovery proves that both methods are called only once as well. 

So that points to the discoveryStatusSubject publishing several times. So I went on to explore that, and it turned out that is not the issue either.

What is happening is that CardPresentPaymentStore.cancelCardReaderDiscovery is [attaching a new subscriber to the discoveryStatus](https://github.com/woocommerce/woocommerce-ios/pull/3788/files#diff-90899ba331903ac8c2f604f84fcf455e71f7f759cd02ba98e13745d4af71394dR68) publisher every time we call that method. That happens in line 68 of CardPaymentStore.cancelCardReaderDiscovery. Oops 🤷 

I pushed a commit with updated traces to the console. After four taps to the "connect reader" button, the full console log, adding traces to the service, store and view model is:

```
|||| Service. calling start in service  <Hardware.StripeCardReaderService: 0x6000038368f0>
///// status received  discovering
===== View model - cancellation callback ====
View model. new status received  discovering
///// status received  discovering
===== View model - cancellation callback ====
View model. new status received  discovering
///// status received  discovering
===== View model - cancellation callback ====
View model. new status received  discovering
**** Store. starting discovery*
==== View model new card readers discovered callback ====
View Model  - new readers:  []
**** Store. cancelling discovery*
||||||| Service calling cancel in service <Hardware.StripeCardReaderService: 0x6000038368f0>
///// status received  idle
===== View model - cancellation callback ====
View model. new status received  idle
///// status received  idle
===== View model - cancellation callback ====
View model. new status received  idle
///// status received  idle
===== View model - cancellation callback ====
View model. new status received  idle
///// status received  idle
===== View model - cancellation callback ====
View model. new status received  idle
```

Notice how there is only one hit to the service and the store for starting and cancelling discovery.

At this point, there are several options:
* make sure the publisher is [removing duplicates](https://developer.apple.com/documentation/combine/publisher/removeduplicates()), which does not solve the problem but I think it's good to have anyway, so I added it to this particular publisher.
* cancel any previous subscribers every time we call cancelDiscovery. Or at least not attach any new ones when calling cancelDiscovery. There are two downsides here I can see: if we did this, there would only be one subscriber active at any given moment, which I don't know how bad it is, but most importantly, it sounds like a lot of logic in the store (track cancellables per high-level method, and decide which ones are cancelled and when), which sounds like error prone, and like something defies the whole point of using Combine.
* get rid of Combine now that it's not too late. Switching the service to block-based APIs is still easy. We could leave Combine confined to the view model <-> view controller relationship. Considering that from the CardPresentPaymentStore up to the UI the API is block based anyway, and considering as well that the rest of the app is block based, this seems like the best solution to me. It would make the code familiar to everyone else. We would not need to this now, we could get this PR in and then migrate the service APIs right after hack week.

Thoughts?

allendav:
> cancelCardReaderDiscovery is attaching a new subscriber to the discoveryStatus publisher every time we call that method

That's what I guessed was happening. Thanks for digging into it

> At this point, there are several options:

I am ok with any of those options. Even removing Combine at that level. It might be best for this service to work like the other services at this time, and cut our Combine teeth at a safer level (the UI). I'm ok with whichever way you want to proceed.

allendav (approved):
So, "At this point, there are several options" notwithstanding, I am OK with moving forward with this since it works and unlocks capabilities for the UI and rest of the app - i.e. gets us to other unknowns faster.

A separate issue should be opened to decide on the option of course

PR #3789, opened 2021-03-04
Shipping Labels: completed the suggested address screen and ended the addresses validation flow

pmusolino:
Part of #2973

Shipping Labels: completed the suggested address screen and ended the addresses validation flow.

## Description
This is one of the last PR about the Shipping Labels addresses validation flow. The other PR will be focusing on completing small things, still in todo under #2973.
In this PR the entire logic and design for the suggested address screen are completed, and you will be able to test the entire address validation flow.

## Changes
- `ShippingLabelSuggestedAddressViewController`: completed the logic for managing the two statuses (suggested or address) in the suggested address screen. Handled the actions of the two buttons inside the screen.
- `ShippingLabelAddressFormViewController`: handled the action of the confirm button.
- `ShippingLabelFormViewController`: handled the action on top of the cell when tapped in the editable + validated status.

## Testing
2. Open an order detail.
5a. If the address is valid, you should see the **Ship To** cell enabled with the **Continue** button visible. Inside the **Ship From** cell the Continue button should disappear.
5b. If the address is not valid, you should see the editing screen, where you can validate the address, or use it as is.
5c. If the address is valid, but there is a suggested address, you should see the suggested address screen, where you can select the original address, the suggested address, or editing it.
6. Repeat the procedure with the **Ship To** cell.

## GIF

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
@@ -129,24 +129,24 @@ private extension ShippingLabelSuggestedAddressViewController {
         switch selectedAddress {
         case .entered:
             onCompletion(address)
-            self.navigationController?.popViewController(animated: true)
+            navigationController?.popViewController(animated: true)
         case .suggested:
             onCompletion(suggestedAddress)
-            self.navigationController?.popViewController(animated: true)
+            navigationController?.popViewController(animated: true)
         }
     }
 
     @objc func didTapEditAddressButton() {
         switch selectedAddress {
         case .entered:
-            displayEditAddressFormVC(siteID: siteID, address: address, type: type)
+            displayEditAddressFormVC(address: address, type: type)
         case .suggested:
-            displayEditAddressFormVC(siteID: siteID, address: suggestedAddress, type: type)
+            displayEditAddressFormVC(address: suggestedAddress, type: type)
         }
 
     }
 
-    func displayEditAddressFormVC(siteID: Int64, address: ShippingLabelAddress?, type: ShipType) {
+    func displayEditAddressFormVC(address: ShippingLabelAddress?, type: ShipType) {
         let shippingAddressVC = ShippingLabelAddressFormViewController(siteID: siteID,
                                                                        type: type,
                                                                        address: address,
@@ -155,7 +155,7 @@ private extension ShippingLabelSuggestedAddressViewController {
                                                                         self.onCompletion(newShippingLabelAddress)
                                                                         self.navigationController?.popViewController(animated: true)
                                                                        })
-        self.navigationController?.pushViewController(shippingAddressVC, animated: true)
+        navigationController?.pushViewController(shippingAddressVC, animated: true)
     }
 }
 
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
@@ -34,6 +34,13 @@ final class ShippingLabelSuggestedAddressViewController: UIViewController {
     typealias Completion = (_ address: ShippingLabelAddress?) -> Void
     private let onCompletion: Completion
 
+    private var selectedAddress: SelectedAddress = .suggested {
+        didSet {
+            tableView.reloadData()
+            configureUseAddressButton()
+        }
+    }
+
     /// Init
     ///
     init(type: ShipType, address: ShippingLabelAddress?, suggestedAddress: ShippingLabelAddress?, completion: @escaping Completion) {
@@ -81,6 +88,7 @@ private extension ShippingLabelSuggestedAddressViewController {
         registerTableViewCells()
 
         tableView.dataSource = self
+        tableView.delegate = self
 
         // Configure header container view
         let headerContainer = UIView(frame: CGRect(x: 0, y: 0, width: Int(tableView.frame.width), height: 0))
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -202,20 +220,34 @@ private extension ShippingLabelFormViewController {
...
+                                                                                                validated: true)
+                                        }
                                                                        })
         self.navigationController?.pushViewController(shippingAddressVC, animated: true)
     }
 
-    func displaySuggestedAddressVC(address: ShippingLabelAddress?, suggestedAddress: ShippingLabelAddress?) {
-        let vc = ShippingLabelSuggestedAddressViewController(type: .origin,
+    func displaySuggestedAddressVC(address: ShippingLabelAddress?, suggestedAddress: ShippingLabelAddress?, type: ShipType) {
+        let vc = ShippingLabelSuggestedAddressViewController(siteID: self.viewModel.siteID,
```
jaclync: nit: `self` might not be necessary?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -202,20 +220,34 @@ private extension ShippingLabelFormViewController {
         let shippingAddressVC = ShippingLabelAddressFormViewController(siteID: self.viewModel.siteID,
                                                                        type: type,
                                                                        address: address,
-                                                                       completion: { [weak self] (newShippingLabelAddress) in
+                                                                       completion: { [weak self]
+                                                                        (newShippingLabelAddress) in
```
jaclync: nit: any reasons for breaking the original line to two lines?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
```diff
@@ -99,18 +109,56 @@ private extension ShippingLabelSuggestedAddressViewController {
...
+        switch selectedAddress {
+        case .entered:
+            displayEditAddressFormVC(siteID: siteID, address: address, type: type)
+        case .suggested:
+            displayEditAddressFormVC(siteID: siteID, address: suggestedAddress, type: type)
+        }
+
+    }
+
+    func displayEditAddressFormVC(siteID: Int64, address: ShippingLabelAddress?, type: ShipType) {
```
jaclync: nit: if the site ID stays the same throughout this screen, maybe the `siteID` parameter isn't necessary?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
```diff
@@ -99,18 +109,56 @@ private extension ShippingLabelSuggestedAddressViewController {
...
+        let shippingAddressVC = ShippingLabelAddressFormViewController(siteID: siteID,
+                                                                       type: type,
+                                                                       address: address,
+                                                                       completion: { [weak self] (newShippingLabelAddress) in
+                                                                        guard let self = self else { return }
+                                                                        self.onCompletion(newShippingLabelAddress)
+                                                                        self.navigationController?.popViewController(animated: true)
+                                                                       })
+        self.navigationController?.pushViewController(shippingAddressVC, animated: true)
```
jaclync: nit: `self` probably not needed? 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
```diff
@@ -99,18 +109,56 @@ private extension ShippingLabelSuggestedAddressViewController {
...
 }
 
+// MARK: - Actions
+//
+private extension ShippingLabelSuggestedAddressViewController {
+    @objc func didTapUseAddressButton() {
+        switch selectedAddress {
+        case .entered:
+            onCompletion(address)
+            self.navigationController?.popViewController(animated: true)
```
jaclync: nit: `self` might not be needed?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
```diff
@@ -99,18 +109,56 @@ private extension ShippingLabelSuggestedAddressViewController {
...
+        case .suggested:
+            onCompletion(suggestedAddress)
+            self.navigationController?.popViewController(animated: true)
```
jaclync: nit: `self` might not be needed? 🙂

jaclync (approved):
Nice work! :shipit: 

Just some observations from testing:

- I haven't seen any suggested address for destination address for several orders, where the destination address could be slightly modified like the postal code in the US. Does the suggested address also work for destination address?
- Scenario 5c: is there any way to validate the address remotely and get the suggested address again after selecting the address entered once?
- The warning text about the address not verified seems to be truncated (iPhone 11 simulator iOS 14.1):


- Right now I can remove all the address fields and use this address as entered. We might want to block this when the "Required" fields are empty? I see a subtask in #2973 to `Show errors when the name or the company is empty.`, it looks like there are more required fields and the company field is optional.

pmusolino:
Thank you Jaclyn for all the suggestions and the review! There were all those `self`s because those methods were extrapolated later before they were nested in some closures.

> I haven't seen any suggested address for destination address for several orders, where the destination address could be slightly modified like the postal code in the US. Does the suggested address also work for destination address?

Yes it works also for the destination address, the API call is the same, and the only difference is that we send the parameter `origin` or `destination` to the API request, but I think there are no differences under the hood... or maybe not? 😄

> Scenario 5c: is there any way to validate the address remotely and get the suggested address again after selecting the address entered once?

Yes, but in our flow, it does not seem to be expected. One reason is that you could probably end up in a loop.

> The warning text about the address not verified seems to be truncated (iPhone 11 simulator iOS 14.1):

This seems to be a bug in the `TopBannerView`, because I found that bug also in other scenarios (like in Products). The next week I'm going to work on it during the Hack Week, I'll try to fix it.

> Right now I can remove all the address fields and use this address as entered. We might want to block this when the "Required" fields are empty? I see a subtask in #2973 to Show errors when the name or the company is empty., it looks like there are more required fields and the company field is optional.

I updated the subtask because it was mentioning just Name and Company.  👍

PR #3794, opened 2021-03-05
Large titles design enhancements

jaclync:
Part of #3751 

## Why

After enabling large titles for the main tabs, there are a couple of design enhancements we can make:

## Changes

- Navigation bar appearance: updated background color to be the same as the list view, and hid the bottom border only when large title is always shown https://github.com/woocommerce/woocommerce-ios/commit/bb51a8800f81a35273d8d299534eb979d029c963 https://github.com/woocommerce/woocommerce-ios/commit/85fd3bc946eec110704a823a09488059dccf3eea
- Update `UIImage` icons in navigation bar to use [SF Symbols](https://developer.apple.com/design/human-interface-guidelines/sf-symbols/overview/) (thankfully we're iOS 13+!) https://github.com/woocommerce/woocommerce-ios/commit/ab55332bfbb70de66c8f6d2f741cff5f6650c048
- Orders and Products tab: moved the search CTA from left to right https://github.com/woocommerce/woocommerce-ios/commit/83e1cc1c0f5eb8928dd55d0a1fab85ec2bf05ae0

## Testing

- Launch the app and log in if needed --> the main tabs should look up to design, while the rest of the screens look like before except for the navigation bar background color change

## Example screenshots

@woocommerce/mobile-designers you can also try the installable build #19090!

\ | light - before | light - after | dark - before | dark - after
-- | -- | -- | -- | --
Dashboard | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 14 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 12 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 37 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 32
Orders | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 17 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 15 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 39 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 35
Products | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 22 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 20 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 41 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 37
Reviews | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 26 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 23 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 05 43 | Simulator Screen Shot - iPhone 11 - 2021-03-05 at 18 07 40

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIImage+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIImage+Woo.swift
@@ -121,10 +121,10 @@ extension UIImage {
         return UIImage.gridicon(.cross)
     }
 
-    /// Cog Icon
+    /// Gear Icon - used in `UIBarButtonItem`
     ///
-    static var cogImage: UIImage {
-        return UIImage.gridicon(.cog)
+    static var gearBarButtonItemImage: UIImage {
+        return UIImage(systemName: "gear", withConfiguration: Configurations.barButtonItemSymbol)!
     }
 
     /// Comment Icon
@@ -439,10 +439,16 @@ extension UIImage {
         return UIImage.gridicon(.plus)
     }
 
-    /// Search Icon
+    /// Plus Icon - used in `UIBarButtonItem`
     ///
-    static var searchImage: UIImage {
-        return UIImage.gridicon(.search)
+    static var plusBarButtonItemImage: UIImage {
+        return UIImage(systemName: "plus", withConfiguration: Configurations.barButtonItemSymbol)!
+    }
+
+    /// Search Icon - used in `UIBarButtonItem`
+    ///
+    static var searchBarButtonItemImage: UIImage {
+        return UIImage(systemName: "magnifyingglass", withConfiguration: Configurations.barButtonItemSymbol)!
             .imageFlippedForRightToLeftLayoutDirection()
     }
 
@@ -605,5 +611,9 @@ private extension UIImage {
 
     enum Metrics {
         static let defaultWooLogoSize = CGSize(width: 30, height: 18)
+    }
+
+    enum Configurations {
+        static let barButtonItemSymbol = UIImage.SymbolConfiguration(pointSize: 22, weight: .regular, scale: .medium)
     }
 }
--- a/WooCommerce/Classes/Extensions/UINavigationBar+Appearance.swift
+++ b/WooCommerce/Classes/Extensions/UINavigationBar+Appearance.swift
@@ -12,7 +12,8 @@ extension UINavigationBar {
         if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) {
             let appearance = UINavigationBarAppearance()
             appearance.configureWithOpaqueBackground()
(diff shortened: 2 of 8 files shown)


WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -606,4 +612,8 @@ private extension UIImage {
     enum Metrics {
         static let defaultWooLogoSize = CGSize(width: 30, height: 18)
     }
+
+    enum Configurations {
+        static let barButtonItemSymbol = UIImage.SymbolConfiguration(pointSize: 22, weight: .regular, scale: .medium)
```
pmusolino, line 617: nit: isn't better to not customize the size here, but just where is needed? In this way, the icon is not reusable in other points of the app, if needed.

pmusolino (approved):
LGTM :shipit: 
Just a minor nit/question.

ghost (approved):

WooCommerce/Classes/Extensions/UIImage+Woo.swift
jaclync: I think symbol images work a bit differently from pre-existing images from an asset or Gridicons library, since the image initialization requires `UIImage.SymbolConfiguration` and I haven't found an API to update its `pointSize`/`weight`/`scale` after initialization.

When we start replacing existing Gridicons usage in the future, we could consider passing a `UIImage.SymbolConfiguration` in `UIImage+Woo` with a static function instead and some default configurations are provided. For this PR, I thought I'd keep changes only to the affected navigation bar button items. What do you think?

pmusolino: Probably the best way, If I understood correctly the problem, is to not use a `UIBarButtonItem`, but a UIButton that will be initialized with a specific size that respects the size that we want.

> For this PR, I thought I'd keep changes only to the affected navigation bar button items. What do you think?

Yes I agree, I don't see any other elements that are affected by this at the moment 👍

jaclync: > Probably the best way, If I understood correctly the problem, is to not use a UIBarButtonItem, but a UIButton that will be initialized with a specific size that respects the size that we want.

Did you mean using a `UIButton` inside a `UIBarButtonItem` so that the image is more customizable and reusable? It sounds worth trying when we are replacing more images with SF symbols  👍

jaclync:
Thanks for the code & design review! Merging now to have the updated look for illustration updates 🙂

PR #3795, opened 2021-03-05
Filter attributes for variations

ealeksandrov:
## Description

^ Notice "Used for variations" setting.

This PR removes non-variable attributes usage in variations-related part.

Related discussion - p1614883468085000-slack-CGPNUU63E

## Changes

1. Adds `attributesForVariations` computed var on `Product`
2. Switches all variations stuff to use `attributesForVariations`
3. Updates tests

## Test

1. On `wp-admin` add an attribute without "used for variations" option
2. Generate couple of variations
3. Variations cell on a product should not have non-variable attribute listed
4. Open generated variations list on a device, it should match with `wp-admin`

## Screenshots

before | after
--|--

### after

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/Product.swift
+++ b/Networking/Networking/Model/Product/Product.swift
@@ -109,6 +109,10 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         return ProductTaxStatus(rawValue: taxStatusKey)
     }
 
+    public var attributesForVariations: [ProductAttribute] {
+        attributes.filter { $0.variation }
+    }
+
     /// Whether the product has an integer (or nil) stock quantity.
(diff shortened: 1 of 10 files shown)


Networking/Networking/Model/Product/Product.swift
```diff
@@ -109,6 +109,10 @@ public struct Product: Codable, GeneratedCopiable, Equatable {
         return ProductTaxStatus(rawValue: taxStatusKey)
     }
 
+    public var attributesForVariations: [ProductAttribute] {
```
Ecarrion, line 115: nit: It would be nice to add a comment and explain why this is needed to future readers.

Ecarrion (approved):
Amazing! 
Looks and works great me! 🚀

PR #3800, opened 2021-03-06
Fix running `match` locally

jkmassel:
**To Test:**
- Run `bundle exec fastlane update_certs_and_profiles` on this branch

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -648,7 +648,7 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
     match(
       type: "enterprise",
       team_id: get_required_env("INT_EXPORT_TEAM_ID"),
-      readonly: options[:readonly] || is_ci,
+      readonly: true,
       app_identifier: ["com.automattic.alpha.woocommerce"])
   end
 
@@ -656,7 +656,7 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
     match(
       type: "appstore",
       team_id: get_required_env("EXT_EXPORT_TEAM_ID"),
-      readonly: options[:readonly] || is_ci,
+      readonly: true,
       app_identifier: ["com.automattic.woocommerce"])
   end
 


fastlane/Fastfile
```diff
@@ -648,15 +648,15 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
     match(
       type: "enterprise",
       team_id: get_required_env("INT_EXPORT_TEAM_ID"),
-      readonly: options[:readonly] || is_ci,
+      readonly: true,
```
mokagio, line 651: Playing captain obvious: this prevents us from regenerating these provisioning profiles from here.

I think this is good, the extra friction of having to edit the `Fastfile` might prevent unintended changes done while hitting keys in a rush, like when someone (me) does `git add .` instead of `git add -p`...

One suggestion I have is to extract this in a constant, so that if we'll  ever had the need to regenerate the profiles from here, it'll be easier to do it.

Having the value in a constant would also add a chance to document the rationale behind this value.

```ruby
# At the time of writing, the Enterprise Developer Account doesn't support logging in via
# API key. To avoid confusing errors when syncing then provisioning profiles locally because
# match can't access the Enterprise account, let's prevent it from trying to access it in the
# first place.
MATCH_READONLY = true
```

mokagio (approved):
Tried it locally, all good. My suggestion does not block merging this; leaving it up to you.

shiki (approved):
I have no idea how this fixes the error but it works! 👏 

To document, this was the error I was getting that was fixed by this PR: 

```
/wooios/develop/vendor/bundle/ruby/2.6.0/gems/fastlane-2.170.0/match/lib/match/runner.rb:73:in `run': [!] undefined method `in_house?' for nil:NilClass (NoMethodError)
```

fastlane/Fastfile
jkmassel: > One suggestion I have is to extract this in a constant, so that if we'll ever had the need to regenerate the profiles from here, it'll be easier to do it.

My thought on this is actually that we should probably have a way of doing this outside the project. We have about one year before it affects us – I guess we'll see if we get to it by then? 😅

jkmassel:
> I have no idea how this fixes the error but it works!

For posterity, this fixes the issue by not requiring a login to App Store Connect as part of the process. It's not required for these instances – the keys can simply be downloaded from Google Cloud Storage. By saying `readonly: false`, we avoid communicating with App Store Connect at all, which simplifies things and works around the error.

Issue #2971, opened 2020-10-14
Shipping Labels M2: check if the order is eligible for shipping label creation


anitaa1990:
- [ ] Add a new button in the Order detail screen to Create Shipping Label. This button will only be displayed if:
- WooCommerce Services is installed and active on the store.
- the store is in the US. 
- the destination address for the order is in the US. 
- the order has physical products or a combination of physical and virtual products.
- [x] If CTA is displayed, also show `Learn more about creating labels with your phone` info row at the end of the Products section

<img width="300" src="https://user-images.githubusercontent.com/22608780/95943338-1ed68c80-0e03-11eb-8774-97987bd5e040.png">

rachelmcr:
Completed in #4143.

PR #3810, opened 2021-03-10
Yosemite: add placeholder action for checking shipping label creation eligibility for an order

jaclync:
Part of #2971 

## Why

The [shipping label creation eligibility endpoint](https://github.com/Automattic/woocommerce-services/pull/2334) is still in code review, and I thought I could add a placeholder Yosemite action that keeps the current feature flag logic in the meantime. When the endpoint is completed, we can then add the Networking logic to check shipping label creation eligibility from the new `/creation_eligibility` endpoint.

## Changes

In Yosemite layer:
- Added `ShippingLabelAction.checkCreationEligibility` for checking whether an order is eligible for shipping label creation
- Implemented the action that returns the given feature flag value

In app layer:
- Added a function to dispatch `ShippingLabelAction.checkCreationEligibility` in `OrderDetailsViewModel`, and set the eligibility flag in `OrderDetailsDataSource` on completion
- In `OrderDetailsViewController`, called the above view model function in `viewWillAppear` and pull to refresh (same as other sync actions)

### Feature flag on

- Launch the app from Xcode (debug build)
- Go to the orders tab
- Tap on an order --> there should be a "Create Shipping Label" CTA

### Feature flag off

- In `DefaultFeatureFlagService`, hard code to return `false` for `shippingLabelsRelease2` feature flag
- Launch the app
- Go to the orders tab --> there should be no "Create Shipping Label" CTA

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -40,7 +40,7 @@ final class OrderDetailsDataSource: NSObject {
 
     /// Whether the order is eligible for shipping label creation.
     ///
-    var isEligibleForShippingLabelCreation: Bool = ServiceLocator.featureFlagService.isFeatureFlagEnabled(.shippingLabelsRelease2)
+    var isEligibleForShippingLabelCreation: Bool = false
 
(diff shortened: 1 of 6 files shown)


ealeksandrov (approved):
Great idea about abstracting API call away with placeholder for now :+1:
LGTM 🚢

Issue #3096, opened 2020-11-03
[HACK Week] FakeKit ™️


Ecarrion:
# Problem

There is an inconvenience when creating a new Networking model which is that creating these models for tests is very tedious.

To circumvent this the project defines a series of "Mock<Model>" objects that help in creating these models in a simpler way.

The problem is that these Mock classes have to be created manually and can't be used between targets.

# Proposed solution

Create `FakeKit ™️` framework on SPM that will store all of these mock objects.

# Tasks

- [x] Fix the problem that does not let us run `rake generate` due to SPM #2864 
- [x] Create a new framework on SPM (and decide a better name?)
- [x] Analyze if fake is a better fit than mock
- [x] Autogenerate fake/mock objects on the new framework
- [ ] Integrate new fakes and delete old mocks
- [x] Update Documentation

shiki:
Another concern that I have is that in some of our unit tests, it's not immediately clear **what data** is being tested. Consider this test for example:

https://github.com/woocommerce/woocommerce-ios/blob/0221c16caae9888b20462d3d673b448f29eca559/WooCommerce/WooCommerceTests/ViewRelated/OrderDetailsViewModelTests.swift#L40-L48

It is not clear whether the test will fail if someone changed `sampleOrderCreatedInCurrentYear()` later. What data is important in `sampleOrderCreatedInCurrentYear()`? Do we really need all this pre-filled information? 

https://github.com/woocommerce/woocommerce-ios/blob/0221c16caae9888b20462d3d673b448f29eca559/WooCommerce/WooCommerceTests/Tools/MockOrders.swift#L68-L93

I think we would do better if we only had [MockOrders.empty()](https://github.com/woocommerce/woocommerce-ios/blob/0221c16caae9888b20462d3d673b448f29eca559/WooCommerce/WooCommerceTests/Tools/MockOrders.swift#L8), which defines the minimum default values, and use `.copy()` to define the test requirements instead: 

```swift
    func testFormattedDateCreatedDoesNotIncludeTheYearForDatesWithTheSameYear() {
        let order = MockOrders().empty().copy(dateCreated: Date())

        let viewModel = OrderDetailsViewModel(order: order)
        let expectedYearString = DateFormatter.yearFormatter.string(from: order.dateCreated)

        let formatted = viewModel.formattedDateCreated

        XCTAssertFalse(formatted.contains(expectedYearString))
    }
```

[...]

PR #3813, opened 2021-03-10
[Hack Week] Introduce "Fakes" framework

Ecarrion:
part of #3096 

# Why

There is an inconvenience when creating a new `Networking` model which is that instantiating these models for tests is very tedious.

To circumvent this, the project defines a series of "Mock" objects that help in creating these models in a simpler way.

The problem is that these "Mock" classes have to be created manually and can't be used between targets.

# How

This PR solves the problem stated above by creating a new **Fakes.framework**, the idea behind this framework is to provide an insolated package that contains "fake" objects for our `Networking` models. These models will be autogenerated using the same strategy as the [Copiable](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/copiable.md#generating-copiable-methods) pattern.

## FAQ

**Why a framework?** So we can reuse the fake objects through all test targets, as well as making sure that these objects are not part of any release target.

**Why not an SPM package?** I tried to create them as an SPM package but I found several blockers due to dependencies, P2 coming soon!

**How do I create a new fake object?** While creating your networking model, you can make it conform to the `GeneratedFakeable` protocol, then run `rake generate` to codegen the fake object. **Tip:** Also make it conform to `GeneratedCopiable` to be able to modify it too. A possible improvement would be to create a new protocol that encloses those two `GeneratedXxxx` protocols.

**How are fake values defined?** There is a [`Fake.swift`](https://github.com/woocommerce/woocommerce-ios/blob/issue/3096-new-spm-project/Fakes/Fakes/Fake.swift) file that defines all fake values for most common data types. Also, default values for enums default to their first enum case.

**Where is the codegen code defined?** The template resides in [`Fakes.swifttemplate`](https://github.com/woocommerce/woocommerce-ios/blob/issue/3096-new-spm-project/CodeGeneration/Fakes/Fakes.swifttemplate)

**How would a test look like?**
```swift
func test() {
     // Given
     let initialProduct = Product.fake()
     let expectedProduct = initialProduct.copy(name: "new-name")
     let ViewModel = ViewModel(product: initialProduct)
     
     // When
     viewModel.updateName("new_name")
     
     // Then
     XCTAssertEqual(viewModel.product, expectedProduct)
}
```

**Will this work for all of our networking models?** Yes, I've already [codegened](https://github.com/woocommerce/woocommerce-ios/pull/3814)  `fakes` for all of our networking models to make sure this approach is complete.

## Notes

- I've made `Product` and its `enums` conform to `GeneratedFakeable` so you can see what code is being generated. I will update the rest of the models in a next PR.

- The PR looks big, but most of the code is the new `Fakes.xcodeproj/project.pbxproj`(311) file in the [autogenerated file](https://github.com/woocommerce/woocommerce-ios/blob/issue/3096-new-spm-project/Fakes/Fakes/Fakes.generated.swift)(217).

# Testing Steps
- Go to any networking model and make it conform to `GeneratedFakeable`
- Run `rake generate` from the project's root path.
- Search your new fake object int `Fakes.generated.swift`
- See that `Fakes` framework compiles successfully.

Update release notes:


ealeksandrov (approved):
Generation of networking fakes works well 🚀

Ecarrion:

PR #3816, opened 2021-03-12
[Hack Week] Delete `MockProduct` class

Ecarrion:
part of #3096 

# Why

Using `MockProduct().product()` is very convenient, but when reading the test is not very clear what is being tested.

https://github.com/woocommerce/woocommerce-ios/blob/c0c953752c36b750dd9a80a62c87465415df3222/WooCommerce/WooCommerceTests/ViewRelated/Products/Edit%20Product/Downloadable%20Files/ProductDownloadFileViewModelTests.swift#L9-L23

It is not clear whether the test will fail if someone changed `MockProduct().product()` later. What data is important in `MockProduct().product()`? Do we really need all this pre-filled information?

**Note:** We had two `MockProduct` classes, one in `Yosemite` and one in `WooCommerce` targets.

----

That problem can be mitigated by replacing `MockProduct().product()` with our new `Product.fake().copy(attrs:...)` where it would be clear what it is being tested.

# How 

- Replace `MockProduct().product()` with `Product.fake()`
- Added a new `ProductFactory` type(only available on tests) that serves specifically-configured products that are reused in two or more test files.

# Testing Steps
- Make sure the CI passes

Update release notes:


Networking/Networking/Model/Product/ProductDownload.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 /// Represents a ProductDownload entity.
 ///
-public struct ProductDownload: Codable, GeneratedFakeable {
+public struct ProductDownload: Codable, GeneratedFakeable, GeneratedCopiable {
```
ealeksandrov: Q: Was that added just for consistency? I don't see any usage of `ProductDownload.copy()`

ealeksandrov (approved):
Nice improvement 👏
Build and unit tests are ✅

Networking/Networking/Model/Product/ProductDownload.swift
Ecarrion: You are right, I end up not needing it! Removed in https://github.com/woocommerce/woocommerce-ios/pull/3816/commits/58c30f4d688327084439fe6f2bdaf69adeb12ead
Thanks!

PR #3820, opened 2021-03-12
[HACK Week]Add "Fakes.framework" documentation

### LINKED ISSUE #3096: [HACK Week] FakeKit ™️

Ecarrion:
closes #3096

# Why

Now that the `Fakes.framework` is [integrated](#3813) in the project, is more than fair to add propper documentation to our repository so old and newcomers can understand better how we work.

# Testing Steps

See [live view](https://github.com/woocommerce/woocommerce-ios/blob/3d303fd8c143c33977a580665ce982bed7d5a5da/docs/README.md) and tap on the fakeable links.

Update release notes:


docs/README.md
```diff
@@ -18,6 +18,9 @@
     - [Copiable](copiable.md)
         - [Generating Copiable Methods](copiable.md#generating-copiable-methods)
         - [Modifying The Copiable Code Generation](copiable.md#modifying-the-copiable-code-generation)
+    - [Fakeable](fakeable.md)
+        - [Generating Fake Methods](fakeable.md#generating-fake-methods)
+        - [Modifying Fakes Code Generation](fakeable.md#modifying-the-fakeable-code-generation)
```
shiki: Let's also add these to the **main README** page → https://github.com/woocommerce/woocommerce-ios/tree/issue/3096-documentation#-documentation. 

I think we should just remove this `docs/README.md` file since it's repetitive. Or at least change `docs/README.md` to say “The index is at `/README.md`”.  I could not decide during the time I created this file. 😅

shiki (approved):
I just saw the implementation and want to commend you for the [`Fake.swift`](https://github.com/woocommerce/woocommerce-ios/blob/issue/3096-documentation/Fakes/Fakes/Fake.swift) implementation. That's quite nifty! Excellent work! 👏 

Thank you for documenting it! 

---

The content looks great. I just have one minor quest. 🙂 Merge whenever you're ready!

docs/README.md
Ecarrion: Ha I totally missed that the content was repeated, I also think deleting this file is better!

PR #3822, opened 2021-03-12
Refactor Rename Attributes VC to simplify and remove table view

rachelmcr:
Related: https://github.com/woocommerce/woocommerce-ios/pull/3768#pullrequestreview-600913106

## Description

This is an attempt to simplify `RenameAttributesViewController` by removing the table view and replacing it with a simplified layout that includes a text field and label. I created the layout in `TextFieldWithFooterView` so it could be reused in other VCs that are similarly simple (e.g. `ProductSlugViewController`).

This simplifies the view controller, but `TextFieldWithFooterView` does a fair amount of work to recreate a design that is based on table views, including some duplication of the logic in `TextFieldTableViewCell`. This was a useful exercise for me, but I'm curious if other think this refactor is worth it?

## Changes

* Adds `TextFieldWithFooterView`, which recreates the layout you'd get with a table view that has a single `TextFieldTableViewCell` and a footer. It puts the text field in a container view to provide an edge-to-edge background and border around the text field.
* Removes the table view from `RenameAttributesViewController` and replaces it with `TextFieldWithFooterView`.

Before|After
-|-
Before|After

## Testing

1. Go to the Products tab.
2. Select a variable product.
3. Select Variations to open the variations list.
4. Tap the ... menu and select Edit Attributes.
5. Select a local attribute to open the Attribute Options screen.
6. Tap the ... menu and select Rename.
7. Confirm the Rename Attribute screen appears with the expected layout and behavior (can rename the selected attribute).

## Submitter Checklist

Update release notes:


rachelmcr:
@Ecarrion If you get a chance sometime, I'd be interested in getting your quick thoughts on these changes. Is this the sort of approach you were thinking of in your [earlier feedback](https://github.com/woocommerce/woocommerce-ios/pull/3768#pullrequestreview-600913106)? Do you think this refactor is worth it? (I'm curious to get thoughts on the overall approach before requesting a full review.)

Ecarrion:
Hey @rachelmcr, sorry for the delay in looking at this 😇  

I really like the approach, it feels simpler to me! A couple of nits would be 
- I wouldn't have exposed a VM from the view, as it not really very UIKit-ish. I would just provide de closure or the delegate.
- I think it would be very useful if we go a step further and provide a reusable VC too. That way, places like `ProductSlugViewController` can be replaced by this.

rachelmcr:
Thanks for taking a look! I put this on the back burner for now, but I'll see if I can circle back next sprint and incorporate your feedback.

rachelmcr:
I ran out of time to come back to this during this cool-down sprint, so for now I'm going to close this PR just so it doesn't sit around and get stale. I'll make a note to come back to this idea in a future sprint when I have more time.

Issue #3802, opened 2021-03-08
Add "Variable product" to product creation


Ecarrion:
Now that we can add variations, we should be able to create a variable product from scratch

PR #3832, opened 2021-03-16
Variations: Fix issues when updating product attributes

Ecarrion:
# Why

While working on [adding a variable product from scratch](https://github.com/woocommerce/woocommerce-ios/issues/3802) I realized that the add variations had a couple of problems.

1. If you tap back after creating an attribute(no variations) the variation list screen fails to display the empty variation state screen

2. If you add modifications to your product, then create an attribute or variation, the UI will show that there are changes pending to commit, but the changes have already been committed remotely.

3. If you updated **only** the product name, and then went out to create an attribute or variation. After coming back, the name was not correctly rendered.

# How

1. Fixed by not removing the empty state on the `didLeave.noResultsPlaceholder` state and let the empty state only be controller by the `VM`. The issue was happening because `didLeave.noResultsPlaceholder` occurs when refreshing/going into `.synced` state.

2. Fixed by passing the original product, rather than the modified product to the "Product Variations" module.

3. Fixed by observing any name change and only reloading the screen if the change happened while the screen was not visible. This in order to prevent any table update while the merchant is typing a new product name.

# Demo
Before | After
--- | ---
bad | good

# Testing Steps
- Open the app and go to a variable product
- Modify the name
- Tap on variations cell and create an attribute or a variation
- Tap back until the product screen 
- See that the name is still modified and the update button is visible
- Tap back until the product list screen
- See that the name has the original product name, refresh the list and verify the same.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 6.3
 -----
 - [*] Fix: In landscape orientation, all backgrounds on detail screens and their subsections now extend edge-to-edge. [https://github.com/woocommerce/woocommerce-ios/pull/3808]
+- [*] Fix: Creating an attribute or a variation no longer saves your product pending changes. [https://github.com/woocommerce/woocommerce-ios/pull/3832]
 - [*] Enhancement/fix: image & text footnote info link rows are now center aligned in order details reprint shipping label info row and reprint screen. [https://github.com/woocommerce/woocommerce-ios/pull/3805]
 
 6.2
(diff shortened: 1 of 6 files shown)


ealeksandrov (approved):
LGTM :shipit:
Is this fix scheduled for 6.3, after initial variations release in currently frozen 6.2? I think it's fine since the issue is far from obvious or blocking, but want to confirm.

Ecarrion:

> Is this fix scheduled for 6.3, after initial variations release in currently frozen 6.2? 

Yes, I'm pointing 6.3 as I consider these problems more on the edge-case side!

Issue #3050, opened 2020-10-22
Display the product type cell also in the Product creation screen


pmusolino:
We always should display the product type selected (even if in read-only) in the product details page during the product creation if possible. It gives users feedback on what they previously selected (1. select product type on bottom sheet > 2. see product type selected on product details page). Discussion here p5T066-1zC#comment-6119

<img width="203" alt="Screen Shot 2020-10-22 at 16 51 34" src="https://user-images.githubusercontent.com/495617/96906319-eafe0580-1499-11eb-8154-6b0cf24f5b3b.png">

jaclync:
Just wanted to share why product type was hidden in the first place (ref p91TBi-3f7-p2 and p91TBi-3dS#comment-1649 is the discussion) - the reason was because we don't support adding a variable product now, we decided at the time to hide the product type until we support variable product creation in M5.

PR #3836, opened 2021-03-17
Variations: Create Variable Product

### LINKED ISSUE #3802: Add "Variable product" to product creation

Ecarrion:
closes #3802 #3050 

# Why

Since now we can add variations to a variable product, it's time to enable the ability to create a variable product(with variations) from scratch.

# How

Unfortunately, this wasn't as straightforward. 😓 

In order to add attributes or variations to a new product, we need to create that product as a draft first. This is done in `EditAttributeOptionsViewModel` right after updating the product attributes. Here we check if the product exists remotely to know if we should update the product or create the product as a draft first.

That draft creation has some cascade consequences like:

- The `resultsController` in `ProductVariationViewController` is now invalid since the product has a different ID now. This is fixed by recreating and reconfiguring the `resultsController` with using the new product information.

- The `formType` of `ProductForm` screen and `ProductVariation` screen has to change from `.add` to `.edit` in order to reflect properly the current state of the product. This is done both in `ProductFormViewModel` and in `ProductVariationsViewModel`

# Demo

https://user-images.githubusercontent.com/562080/111397454-589c0f80-868f-11eb-8d82-ad31aefe816f.mov

# Testing Steps
- Open the app and go to the product tab
- Tap on the "+" button and choose variable product
- See that the product type is now visible in the add product form.
- Create a product with attributes and variations as you see fit.
- See that the product form is now in edit mode.
- See that the product has the draft status
- Update the product to published status if you wish.

Update release notes:


rachelmcr:
> In order to add attributes or variations to a new product, we need to create that product as a draft first.

I wonder if we need to make this more explicit to users, rather than just doing it behind the scenes. Otherwise, it isn't obvious why the product is saved as a draft (versus other new products that aren't) and why it has an Update button instead of a Publish button. If users don't understand that, these steps to publish it might not be clear:

> See that the product has the draft status
> Update the product to published status if you wish.

@woocommerce/mobile-designers what do you think?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -353,6 +358,17 @@ extension ProductFormViewModel {
             }
         }
     }
+
+    /// Updates the internal `formType` when a product changes from to to a saved status.
```
rachelmcr: nit: Typo - "when a product changes from **to** to a saved status" (should that be "from **new** to a saved status"?)

rachelmcr (approved):
The changes and new tests look good to me! 👍  

My only concern is the [draft save behavior I mentioned above](https://github.com/woocommerce/woocommerce-ios/pull/3836#issuecomment-801000827), and whether there's a clearer way we can communicate to users what's happening there.

Ecarrion:
Thanks for the review @rachelmcr! 🙇 

> My only concern is the draft save behavior I mentioned above, and whether there's a clearer way we can communicate to users what's happening there.

We discussed that on CGPNUU63E-slack-p1615825991003000 and decided to create the project as a draft silently as Core has the same behavior.

> why it has an Update button instead of a Publish button.

Also on CGPNUU63E-salck-p1615942199006800 we are discussing adding a publish button for drafts products which will help mitigate that concern!

PR #3837, opened 2021-03-17
Shipping Labels: local validation of address fields

pmusolino:
Part of #2973 

## Description
As part of the Shipping Labels M2, we should validate some fields of the Shipping Label Address also locally, in particular, we want to make sure that the user can't proceed without adding some fields that are mandatory, like name, address, city, country, ecc...
I fixed also an issue with the color of the activity indicator that substitute the "Done" button in the navigation bar when pressed. 

## Changes
- `Row` now conform to `Equatable` because we want to manage different type of errors in `fieldError` case.
- The `validateAddress` method allow us to validate an address locally or both locally + remotely.
- Refactor of how `sections` are managed and refreshed. Before, we built sections on the fly, when they were requested by the view controller. This could cause a misalignment with the state of the view model. Sections are now updated when the `State` undergoes a change.
- Updated unit tests.

## Testing
2. Open an order detail.
5. Go to the editing address screen, and try to delete one of the mandatory fields (like `name`). Press the Done button. -> You should see an error under the name field. 
6. The same things should happen pressing the button `Use address as Entered`, with the only difference that the address will be validated only locally, checking that the mandatory fields are fulfilled.

## Screenshots

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -422,7 +422,7 @@ private extension ShippingLabelAddressFormViewController {
                                                       comment: "Error showed in Shipping Label Address Validation for the address field")
         static let missingCity = NSLocalizedString("City missing",
                                                    comment: "Error showed in Shipping Label Address Validation for the city field")
-        static let missingPostcode = NSLocalizedString("Postcode invalid",
+        static let missingPostcode = NSLocalizedString("Postcode missing",
                                                        comment: "Error showed in Shipping Label Address Validation for the postcode field")
         static let missingState = NSLocalizedString("State missing",
                                                     comment: "Error showed in Shipping Label Address Validation for the state field")
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -113,9 +113,8 @@ private extension ShippingLabelAddressFormViewController {
     }
 
     func registerTableViewCells() {
-        for row in Row.allCases {
-            tableView.registerNib(for: row.type)
-        }
+            tableView.registerNib(for: TitleAndTextFieldTableViewCell.self)
+            tableView.registerNib(for: BasicTableViewCell.self)
     }
 
     func observeViewModel() {
@@ -144,7 +143,7 @@ private extension ShippingLabelAddressFormViewController {
 private extension ShippingLabelAddressFormViewController {
 
     @objc func doneButtonTapped() {
-        viewModel.validateAddress { [weak self] (success, error) in
+        viewModel.validateAddress(onlyLocally: false) { [weak self] (success, error) in
             guard let self = self else { return }
             if success {
                 self.onCompletion(self.viewModel.address)
@@ -154,8 +153,13 @@ private extension ShippingLabelAddressFormViewController {
     }
 
     @objc func confirmButtonTapped() {
-        onCompletion(viewModel.address)
-        navigationController?.popViewController(animated: true)
+        viewModel.validateAddress(onlyLocally: true) { [weak self] (success, error) in
+            guard let self = self else { return }
+            if success {
+                self.onCompletion(self.viewModel.address)
+                self.navigationController?.popViewController(animated: true)
+            }
+        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
```diff
@@ -75,45 +80,126 @@ final class ShippingLabelAddressFormViewModel: NSObject {
...
+        return errors
+    }
+}
+
+// MARK: - Remote Validation API
+extension ShippingLabelAddressFormViewModel {
+
+    /// Validate the address locally and remotely. If `onlyLocally` is equal `true`, the validation will happens just locally.
+    ///
+    func validateAddress(onlyLocally: Bool, onSuccess: ((Bool, ShippingLabelAddressValidationError?) -> ())? = nil) {
```
Ecarrion: What do you think about returning a result here?
```swift
Result<Void, ShippingLabelAddressValidationError>
```

I think it models the intention better.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
```diff
@@ -19,7 +19,8 @@ final class ShippingLabelAddressFormViewModel: NSObject {
     let type: ShipType
     private(set) var address: ShippingLabelAddress?
     private(set) var addressValidationError: ShippingLabelAddressValidationError?
-    private(set) var isAddressValidated: Bool = false
+    private(set) var isAddressLocallyValidated: Bool = false
+    private(set) var isAddressFullyValidated: Bool = false
```
Ecarrion: Are these meant to be used later? I only see them being used in tests?
In case you do need them, maybe an enum could help you model this better.

```swift
enum Validation {
  case none
  case local
  case remote
}
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
```diff
@@ -75,45 +80,126 @@ final class ShippingLabelAddressFormViewModel: NSObject {
...
+                rows.insert(.fieldError(.state), at: rows.index(after: index))
+            }
         }
-        return [Section(rows: rows)]
+        else if localErrors.contains(.country) {
+            if let index = rows.firstIndex(where: { $0 == .country }) {
+                rows.insert(.fieldError(.country), at: rows.index(after: index))
+            }
+        }
+        sections = [Section(rows: rows)]
```
Ecarrion: Is it possible to not have these on `if else` conditions?
I tested the feature and its a bit annoying to have the name as an error, fix it, tap done, and then have another field with an error.

It will be nice to get all errors at once, what do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -274,8 +278,24 @@ private extension ShippingLabelAddressFormViewController {
...
+        case .address:
+            errorMessage = viewModel.addressValidationError?.addressError ?? Localization.invalidAddress
+        case .city:
+            errorMessage = Localization.invalidCity
+        case .postcode:
+            errorMessage = Localization.invalidPostcode
+        case .state:
+            errorMessage = Localization.invalidState
+        case .country:
+            errorMessage = Localization.invalidCountry
```
Ecarrion: I think these names are a bit misleading. Do you think it makes sense to have something like "Missing {attribute}" instead of "Invalid {attribute}"

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
```diff
@@ -75,45 +80,126 @@ final class ShippingLabelAddressFormViewModel: NSObject {
...
+            isAddressLocallyValidated = false
+            state.isLoading = false
+            onSuccess?(false, nil)
+            return
+        }
+        isAddressLocallyValidated = true
+
+        if onlyLocally {
+            onSuccess?(true, nil)
+            return
```
Ecarrion: nit: I think you can write this as 
```swift
return onSucess?(...)
```
Which looks a tad bit nicer!

Ecarrion (commented):
Good job @pmusolino, Left you some code and UX suggestions, let me know what you think!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
pmusolino: You are probably right, I wasn't sure of the behavior because on the design we always have the case with just one field on error state. I'll change it 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
pmusolino: Unfortunately, it doesn't work, because in this case, the function expects a non-void return value 🤕

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
pmusolino: Definitely better 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
pmusolino: There is just one thing that can be confusing. Since `ShippingLabelAddressValidationError` should conform to `Error`, there are some cases where the return is a failure, but without an error. What do you expect to see in that case?

Ecarrion: Can `ShippingLabelAddressValidationError` have an `.empty` enum case?

That way you could return something like
```swift
onCompletion(.failure(.empty))
```

Does that make sense?

pmusolino: Consider that `ShippingLabelAddressValidationError` is a network model, which will be parsed from the error response in the address validation, so it contains other data and by default is not an enum 💭

Ecarrion: You deff know better than me since I don't know the module that well but I see two options.

- Create an error enum to encapsulate the network response into a view model error
```swift
enum AddressValidationError: Error {
    case .local // Or a empty or a better name
    case .remote(ShippingLabelAddressValidationError)
}
```

- Just make the error optional in the result signature
```swift
Result<Void, ShippingLabelAddressValidationError?>
```

pmusolino: I valuated the first option, but I'd like to avoid another wrapped, because still the `ShippingLabelAddressValidationError` is a wrapper of an error, and it can be confusing to go up the chain. But if on the contrary, you think it makes things more understandable, I do the change 👍 

About the second one, is it possible to make Error optional in a `Result`? I remember that it's a signature, and if try it, I get this error ```Type 'ShippingLabelAddressValidationError?' (aka 'Optional<ShippingLabelAddressValidationError>') does not conform to protocol 'Error'```. Maybe I'm missing something since I'm at the end of my day and I'm toasted? 😸

Ecarrion: >  But if on the contrary, you think it makes things more understandable, I do the change 👍

I think it does add clarity as the current signature allows you to have the possibility to have a successful validation with errors, but ultimately it's up to you!

pmusolino: Ready for a second check ;) 8cd4788ae406c41428d164735772feb50fe06592

Ecarrion (approved):
Thanks for this Paolo!
🚀

pmusolino:
Thanks to you for the review @Ecarrion 🙇

Issue #3839, opened 2021-03-17
[Mobile Payments] Epic: Does the site support card present payments


allendav:
As part of mobile payments, we need to properly wire up checking whether the WooCommerce site has a card present payment compatible plugin active, account status, etc. Some intersection with #3555 

- [x] Networking: get plugins (to see if wcpay active) #3844
- [x] Networking: get wcpay account (if any) #3881
- [ ] Storage?: persist plugin list, persist card present capable accounts (more in the sense of a medium timeframe cache to reduce fetches since plugin list and account status change infrequently)
- [ ] Yosemite: Actions to fetch plugins list, fetch card present capable accounts
- [ ] WooCommerce/UI: Invite user to enroll, complete setup, show account status
- [ ] Performance?: Advise user if site is capable of supporting card present payments

Ecarrion:
Storage/Yosemite for plugin list will be handled here https://github.com/woocommerce/woocommerce-ios/issues/3921

AliSoftware:
Moved this issue's milestone to 6.9, as code freeze for 6.8 is happening today.

ctarda:
Closing, as this is completed

PR #3844, opened 2021-03-17
[Mobile payments] Add networking - plugins

allendav:
Partially addresses #3839

Why are we doing this?
- This will allow us to, as part of determining if a site supports card present payments, get a list of active and inactive plugins.

Note:
- This is in support of card-present-payments, but since it doesn't directly depend on card-present-payments, I've elected to submit this against the `develop` branch, and not `feature/stripe-terminal-sdk-integration`

What it does:
- Adds networking layer support for fetching the plugins on a site, both active and inactive

How fast is it?
- Not very. On low-end shared hosting, the plugins endpoint typically takes 4-5 seconds to respond.

To test:
- Ensure SitePluginsMapperTests pass
- Ensure SitePluginsRemoteTests pass
- If you'd like to test with a real request against a real site, a quick test would be to modify `Yosemite/Yosemite/Stores/SettingStore.swift` as follows:

```diff
diff --git a/Yosemite/Yosemite/Stores/SettingStore.swift b/Yosemite/Yosemite/Stores/SettingStore.swift
index cb5ac8b18..7cd7726d3 100644
--- a/Yosemite/Yosemite/Stores/SettingStore.swift
+++ b/Yosemite/Yosemite/Stores/SettingStore.swift
@@ -8,6 +8,7 @@ import Storage
 public class SettingStore: Store {
     private let siteSettingsRemote: SiteSettingsRemote
     private let siteAPIRemote: SiteAPIRemote
+    private let sitePluginsRemote: SitePluginsRemote
 
     private lazy var sharedDerivedStorage: StorageType = {
         return storageManager.newDerivedStorage()
@@ -16,6 +17,7 @@ public class SettingStore: Store {
     public override init(dispatcher: Dispatcher, storageManager: StorageManagerType, network: Network) {
         self.siteSettingsRemote = SiteSettingsRemote(network: network)
         self.siteAPIRemote = SiteAPIRemote(network: network)
+        self.sitePluginsRemote = SitePluginsRemote(network: network)
         super.init(dispatcher: dispatcher, storageManager: storageManager, network: network)
     }
 
@@ -62,6 +64,9 @@ private extension SettingStore {
                 onCompletion(nil)
             }
         }
+        sitePluginsRemote.loadPlugins(for: siteID) { [weak self] (plugins, error) in
+
+        }
     }
```
and then set a breakpoint inside that completion. When the app launches and fetches settings, it will also fetch plugins and show them in the watch window.

Update release notes:

### DIFF
--- a/Networking/Networking/Mapper/SitePluginsMapper.swift
+++ b/Networking/Networking/Mapper/SitePluginsMapper.swift
@@ -13,7 +13,6 @@ struct SitePluginsMapper: Mapper {
     ///
     func map(response: Data) throws -> [SitePlugin] {
         let decoder = JSONDecoder()
-        decoder.dateDecodingStrategy = .formatted(DateFormatter.Defaults.dateTimeFormatter)
         decoder.userInfo = [
             .siteID: siteID
         ]
--- a/Networking/Networking/Mapper/SitePluginsMapper.swift
+++ b/Networking/Networking/Mapper/SitePluginsMapper.swift
@@ -1,0 +1,36 @@
+import Foundation
+
+/// Mapper: SitePlugins
+///
+struct SitePluginsMapper: Mapper {
+
+    /// Site Identifier associated to the plugins that will be parsed.
+    /// We're injecting this field via `JSONDecoder.userInfo` because the remote endpoints don't return the SiteID in the plugin endpoint.
+    ///
+    let siteID: Int64
+
+    /// (Attempts) to convert a dictionary into [SitePlugin].
+    ///
+    func map(response: Data) throws -> [SitePlugin] {
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .formatted(DateFormatter.Defaults.dateTimeFormatter)
+        decoder.userInfo = [
+            .siteID: siteID
+        ]
+
+        return try decoder.decode(SitePluginsEnvelope.self, from: response).plugins
+    }
+}
+
+
+/// SitePluginsEnvelope Disposable Entity:
+/// The plugins endpoint returns the document within a `data` key. This entity
+/// allows us to do parse all the things with JSONDecoder.
+///
+private struct SitePluginsEnvelope: Decodable {
+    let plugins: [SitePlugin]
+
+    private enum CodingKeys: String, CodingKey {
+        case plugins = "data"
+    }
+}
--- a/Networking/Networking/Model/SitePlugin.swift
+++ b/Networking/Networking/Model/SitePlugin.swift
@@ -120,27 +120,6 @@ private extension SitePlugin {
     }
 }
 
-// MARK: - Equatable Conformance
-//
-extension SitePlugin: Equatable {
-    public static func == (lhs: SitePlugin, rhs: SitePlugin) -> Bool {
-        return lhs.plugin == rhs.plugin &&
-            lhs.status == rhs.status &&
-            lhs.name == rhs.name &&
-            lhs.pluginUri == rhs.pluginUri &&
-            lhs.author == rhs.author &&
-            lhs.authorUri == rhs.authorUri &&
-            lhs.descriptionRaw == rhs.descriptionRaw &&
-            lhs.descriptionRendered == rhs.descriptionRendered &&
-            lhs.version == rhs.version &&
-            lhs.networkOnly == rhs.networkOnly &&
-            lhs.requiresWPVersion == rhs.requiresWPVersion &&
-            lhs.requiresPHPVersion == rhs.requiresPHPVersion &&
-            lhs.textDomain == rhs.textDomain
-    }
-}
-
-
 // MARK: - Decoding Errors
 //
 enum SitePluginError: Error {
--- a/Networking/Networking/Model/SitePlugin.swift
+++ b/Networking/Networking/Model/SitePlugin.swift
@@ -5,7 +5,7 @@ import Foundation
 public struct SitePlugin: Decodable, GeneratedFakeable {
     public let siteID: Int64
     public let plugin: String                   // e.g. woocommerce/woocommerce (i.e. [folder/]main php file)
-    public let status: String                   // e.g. active, inactive, ...
+    public let status: SitePluginStatusEnum     // i.e. .active | .networkActive | .inactive
     public let name: String                     // e.g. WooCommerce
     public let pluginUri: String                // e.g. https://woocommerce.com/
     public let author: String                   // e.g. Automattic
@@ -23,7 +23,7 @@ public struct SitePlugin: Decodable, GeneratedFakeable {
     public init(
         siteID: Int64,
         plugin: String,
-        status: String,
+        status: SitePluginStatusEnum,
         name: String,
         pluginUri: String,
         author: String,
@@ -61,7 +61,7 @@ public struct SitePlugin: Decodable, GeneratedFakeable {
 
         let container = try decoder.container(keyedBy: CodingKeys.self)
         let plugin = try container.decode(String.self, forKey: .plugin)
-        let status = try container.decodeIfPresent(String.self, forKey: .status) ?? ""
+        let status = try container.decode(SitePluginStatusEnum.self, forKey: .status)
         let name = try container.decodeIfPresent(String.self, forKey: .name) ?? ""
         let pluginUri = try container.decodeIfPresent(String.self, forKey: .pluginUri) ?? ""
         let author = try container.decodeIfPresent(String.self, forKey: .author) ?? ""
@@ -120,9 +120,9 @@ private extension SitePlugin {
     }
 }
 
-// MARK: - Comparable Conformance
+// MARK: - Equatable Conformance
 //
-extension SitePlugin: Comparable {
+extension SitePlugin: Equatable {
     public static func == (lhs: SitePlugin, rhs: SitePlugin) -> Bool {
         return lhs.plugin == rhs.plugin &&
             lhs.status == rhs.status &&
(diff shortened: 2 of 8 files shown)


Networking/Networking/Mapper/SitePluginsMapper.swift
```diff
@@ -0,0 +1,36 @@
...
+    /// Site Identifier associated to the plugins that will be parsed.
+    /// We're injecting this field via `JSONDecoder.userInfo` because the remote endpoints don't return the SiteID in the plugin endpoint.
+    ///
+    let siteID: Int64
+
+    /// (Attempts) to convert a dictionary into [SitePlugin].
+    ///
+    func map(response: Data) throws -> [SitePlugin] {
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .formatted(DateFormatter.Defaults.dateTimeFormatter)
```
pmusolino: nit: I noticed that there are no dates to decode, so the date decoding strategy might be superfluous 👍

Networking/Networking/Model/SitePlugin.swift
```diff
@@ -0,0 +1,156 @@
+import Foundation
+
+/// Represents a specific plugin entity for a specific site.
+///
+public struct SitePlugin: Decodable, GeneratedFakeable {
+    public let siteID: Int64
+    public let plugin: String                   // e.g. woocommerce/woocommerce (i.e. [folder/]main php file)
+    public let status: String                   // e.g. active, inactive, ...
```
pmusolino: nit: could be better to use an `enum` here if we already know all the statuses

Networking/Networking/Model/SitePlugin.swift
```diff
@@ -0,0 +1,156 @@
...
+
+    enum DescriptionCodingKeys: String, CodingKey {
+        case raw      = "raw"
+        case rendered = "rendered"
+    }
+}
+
+// MARK: - Comparable Conformance
+//
+extension SitePlugin: Comparable {
```
pmusolino: Do we really need the `Comparable` conformance? I think that the `Equatable` conformance should be enough, but perhaps there are cases that I have not considered where `Comparable` is needed?

Networking/Networking/Remote/SitePluginsRemote.swift
```diff
@@ -0,0 +1,30 @@
...
+///
+public class SitePluginsRemote: Remote {
+
+    /// Retrieves all of the `SitePlugin`s for a given site.
+    ///
+    /// - Parameters:
+    ///   - siteID: Site for which we'll fetch the plugins.
+    ///   - completion: Closure to be executed upon completion.
+    ///
+    public func loadPlugins(for siteID: Int64, completion: @escaping ([SitePlugin]?, Error?) -> Void) {
```
pmusolino: nit: It could be better to use the return type `Result<[SitePlugin], Error>` here, which seems perfect for this use case and that we started to use by some months in our codebase.

Networking/NetworkingTests/Remote/SitePluginsRemoteTests.swift
```diff
@@ -0,0 +1,55 @@
...
+        network.removeAllSimulatedResponses()
+    }
+
+    // MARK: - Load plugins tests
+
+    /// Verifies that loadPlugins properly parses the sample response.
+    ///
+    func testLoadPluginsProperlyReturnsPlugins() {
+        let remote = SitePluginsRemote(network: network)
+        let expectation = self.expectation(description: "Load site plugins")
```
pmusolino: nit: also here we started to abandon the `expectation` and the `wait(for: [expectation], timeout: Constants.expectationTimeout)` methods, and we started to use this shape (the code below is not tested 🗡️ ):

```
// Given
        let remote = SitePluginsRemote(network: network)
        network.simulateResponse(requestUrlSuffix: "plugins", filename: "plugins")

        // When
        let result: Result<[SitePlugin], Error> = waitFor { promise in
            remote.loadPlugins(for: sampleSiteID) { result in
                promise(result)
            }
        }

        // Then
        let sitePlugins = try XCTUnwrap(result.get().plugins)
        let errors = try result.get().errors
        XCTAssertEqual(sitePlugins?.count, 5)
        XCTAssertNil(errors)
```

Networking/NetworkingTests/Remote/SitePluginsRemoteTests.swift
```diff
@@ -0,0 +1,55 @@
...
+    ///
+    override func setUp() {
+        network.removeAllSimulatedResponses()
+    }
+
+    // MARK: - Load plugins tests
+
+    /// Verifies that loadPlugins properly parses the sample response.
+    ///
+    func testLoadPluginsProperlyReturnsPlugins() {
```
pmusolino:
```suggestion
    func test_loadPlugins_properly_returns_plugins() {
```

Networking/NetworkingTests/Remote/SitePluginsRemoteTests.swift
```diff
@@ -0,0 +1,55 @@
...
+            XCTAssertEqual(sitePlugins?.count, 5)
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: Constants.expectationTimeout)
+    }
+
+    /// Verifies that loadGeneralSettings properly relays Networking Layer errors.
+    ///
+    func testLoadPluginsProperlyRelaysNetwokingErrors() {
```
pmusolino:
```suggestion
    func test_loadPlugins_properly_relays_netwoking_errors() {
```

pmusolino (approved):
Thank you @allendav for having introduced this new endpoint, which will be really useful, not only for mobile payments! 
The PR looks good to me, I just left some nits 💯

Networking/Networking/Model/SitePlugin.swift
allendav: Agreed. There are three: active, inactive and network-active. I'll refactor.

https://github.com/WordPress/WordPress/blob/35e1b34f8ee4fcfc8f5e12d48f665b8d599c6e89/wp-includes/rest-api/endpoints/class-wp-rest-plugins-controller.php#L61

Networking/Networking/Model/SitePlugin.swift
allendav: Hmmmm. Probably not. I'll delete it.

Networking/NetworkingTests/Remote/SitePluginsRemoteTests.swift
allendav: Ended up doing more like this:

```
        // When
        let result: Result<[SitePlugin], Error> = waitFor { promise in
            remote.loadPlugins(for: self.sampleSiteID) { result in
                promise(result)
            }
        }

        // Then
        switch result {
        case .success(let plugins):
            XCTAssertEqual(plugins.count, 5)
        case .failure(let error):
            XCTAssertNil(error)
        }
```

Networking/NetworkingTests/Remote/SitePluginsRemoteTests.swift
allendav: Updated in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/8e320fa88e27122e36d57a3de6e396bc6ec0ce37

Networking/NetworkingTests/Remote/SitePluginsRemoteTests.swift
allendav: Updated in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/8e320fa88e27122e36d57a3de6e396bc6ec0ce37

Networking/NetworkingTests/Remote/SitePluginsRemoteTests.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/8e320fa88e27122e36d57a3de6e396bc6ec0ce37

Networking/Networking/Remote/SitePluginsRemote.swift
allendav: Good idea! Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/8e320fa88e27122e36d57a3de6e396bc6ec0ce37

Networking/Networking/Model/SitePlugin.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/8e320fa88e27122e36d57a3de6e396bc6ec0ce37

Networking/Networking/Model/SitePlugin.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/8e320fa88e27122e36d57a3de6e396bc6ec0ce37

Networking/Networking/Mapper/SitePluginsMapper.swift
allendav: Removed in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/8e320fa88e27122e36d57a3de6e396bc6ec0ce37

allendav:
@pmusolino - would you take a look and let me know what you think? The changes were small, but not trivial

Networking/Networking/Model/SitePlugin.swift
pmusolino: I think that you can use just the `Equatable` conformance without declaring a custom `func ==` 😉 because, in theory, you should use all the fields, like the `author` (maybe there are two plugins with the same name, but with a different author? I don't know if this is possible).

pmusolino:
@allendav LGTM :shipit: I just added a new suggestion about the `Equatable` conformance.

Networking/Networking/Model/SitePlugin.swift
allendav: TIL. Deleted it in https://github.com/woocommerce/woocommerce-ios/pull/3844/commits/2dcffd48d2bc4143f6fdfd095b8a7e20076072d1

PR #3827, opened 2021-03-15
[Mobile Payments] Create Payment Intent

ctarda:
Closes #3748 and #3749 

⚠️ This PR is against the feature branch implementing support for Card Present Payments, not against develop ⚠️

🔴🔴 Testing this PR requires a hardware reader and setting up a store for testing🔴🔴

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by implementing logic in Yosemite and Hardware to create a Payment Intent on device. 

The significance of a Payment Intent is, hopefully, explained in [the README accompanying the PR](https://github.com/woocommerce/woocommerce-ios/blob/issue/3748-intent-service/docs/HARDWARE.md).

## Changes
* Implemented `createPaymentIntent` in StripeCardReaderService to implement just enough functionality to attack the Stripe Terminal SDK and create a PaymentIntent. This does not include error handling for now. #3741 and #3747 will cover error handling.
* Implements support in `CardPresentPaymentAction` and `CardPresentPaymentStore` for a new high-level action to collect a payment. The reason why the high level action exposed to the app is "collect a payment" instead of just "create an intent" is covered [in the README that is part of the PR diff](https://github.com/woocommerce/woocommerce-ios/blob/issue/3748-intent-service/docs/HARDWARE.md).
* Adds a variable to `OrderDetailsViewModel` that exposes if an order is eligible for card present payments. At this point, that variable relies only on the feature flag, but eventually this is where we would check if the user is a WCPay user and the order is eligible.
* Adds a button to the toolbar of the OrderDetailsViewController only for testing purposes. This button is available in all orders for now, and triggers the "collect payment" high level action exposed by Yosemite The button will be removed and implemented properly in #3826 

## How to test
Testing is complicated because:
1. Requires setting up a store for [Card Present Payment Testing](https://github.com/woocommerce/woocommerce-ios/blob/issue/3748-intent-service/docs/stripe-tests.md#integration-tests) (also P91TBi-4BH-p2 for more specific instructions)  
2. Requires an external hardware reader.

The process to test this PR requires interacting with the UI while running the app on a device, and also looking at the Xcode console for some logs.
1. Checkout the branch. Run `bundle exec pod install`. 
2. Build and run on a device. 
3. Switch an external card reader on. The only card reader supported by now is the BBPOS Chipper 2X BT.
4. Log in to an account that matches a store that has been configured for testing.
5. Navigate to Settings in the app. (Tap the gear button in the top right)
[...]

PR #3847, opened 2021-03-18
[Mobile Payments] Process a payment

### LINKED ISSUE #3750: [Mobile Payments] Process a PaymentIntent. Mark I: CardReaderService
Implement logic to process a Payment Intent

Scope:
* At this point, if we do not have proper error handling, or a skeleton of it, in place, implement only the happy path

### LINKED ISSUE #3825: [Mobile Payments] Collect Payment Methods
Step 2 of processing a payment:
https://stripe.com/docs/terminal/payments#collect-payment

### LINKED ISSUE #3841: [Mobile Payments] Process a PaymentIntent. Mark II: Yosemite

ctarda:
Closes #3825, #3750, #3841

🔴🔴 Testing this PR requires a hardware reader and setting up a store for testing🔴🔴

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by implementing logic in Yosemite and Hardware to process a payment.

This is a two step process, on top of the previous step implemented in PR #3827 (creating a Payment Intent on device)

After creating a payment intent, we request collecting a payment method (which basically implies tapping the card, or inserting it into the card reader) and then process the payment.

## Changes
* Modified the CardReaderService API, to return a Payment Intent only when the payment has been processed, instead of at every step of the sequence (e.g. after collecting a payment method).
* Modified `collectPayment` in CardPresentPaymentStore to trigger the three operations (create intent, collect payment method and process payment) in sequence
* Add code to OrderDetailsViewModel and OrderDetailsViewController to trigger the payment capture process. For testing purposes, we added a button to the toolbar, to keep things simple.

There are several logs left in place. Some of them indicate functionality that we do have to implement, like:

```
==== did request reader input
SCPReaderInputOptions(rawValue: 7)
7
//// did request reader input
==== did request display message
5
//// did request display message
```

These tells us that the reader expects us to show user-facing messages asking users to tap / insert a card. There will be tickets added soon for these.

We also log the final payment intent, after the payment processing is completed:

```
==== Yosemite log for testing. Payment intent after processing payment
pi_REDACTED
//// payment intent processed 
```

## How to test

2. Build and run on a device. 
8. Tap "Connect card reader"
10. Tap "Connect to reader"
12. Navigate to an order (any order).
13. Tap the money emojis in the top right corner
14. Wait until the reader beeps, one small square blue led turns on, and tap a card.
15. Wait until the reader beeps.
16. Wait until the UI presents an alert showing success or error. 

The following video hopefully helps illustrate the process:

https://user-images.githubusercontent.com/2722505/111704327-54016380-8815-11eb-9feb-05c78f3a9eae.mov

We tried to provide enough UI so that the test can be actually provide visual feedback, without going down the rabbit hole of fully implementing error handling

Update release notes:


Hardware/Hardware/CardReader/CardReaderService.swift
```diff
@@ -45,14 +45,13 @@ public protocol CardReaderService {
 
     /// Creates a PaymentIntent
     /// - Parameter parameters: the intent's parameters
-    func createPaymentIntent(_ parameters: PaymentIntentParameters) -> Future <PaymentIntent, Error>
+    func createPaymentIntent(_ parameters: PaymentIntentParameters) -> Future <Void, Error>
```
allendav, line 48: I love this so much! The PI is mostly meaningless to the rest of the app. Keeping it in CardReaderService instead of passing it around, accepting it again, feels a lot safer.

Hardware/Hardware/CardReader/CardReaderService.swift
```diff
@@ -45,14 +45,13 @@ public protocol CardReaderService {
...
-    /// Collects a payment method for the given intent.
-    func collectPaymentMethod(_ intent: PaymentIntent) -> Future<PaymentIntent, Error>
+    /// Collects a payment method.
+    func collectPaymentMethod() -> Future<Void, Error>
 
-    /// Processes a payment after collecting a payment method succeeds.
-    /// - Parameter intent: the payment intent
-    func processPaymentIntent(_ intent: PaymentIntent) -> Future<PaymentIntent, Error>
+    /// Captures a payment after collecting a payment method succeeds.
+    func processPayment() -> Future<PaymentIntent, Error>
```
allendav: Not sure we need to pass out the entire PI here - we just need the ID as far as we know at this point. If you want to proceed this way, that's fine - I just worry about the surface area making refactoring harder later if we discover we've bound to properties and methods more liberally than we should have.

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -224,6 +263,22 @@ extension StripeCardReaderService: DiscoveryDelegate {
 }
 
 
+// MARK: - ReaderDisplayDelegate.
+extension StripeCardReaderService: ReaderDisplayDelegate {
+    public func terminal(_ terminal: Terminal, didRequestReaderInput inputOptions: ReaderInputOptions = []) {
```
allendav, line 273: Maybe add a comment that didRequestReaderInput means the reader wants the buyer to tap/swipe/insert a card/device (since the value is a bit hard to grok on its own)

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -224,6 +263,22 @@ extension StripeCardReaderService: DiscoveryDelegate {
...
+        print("==== did request reader input")
+        print(inputOptions)
+        print(inputOptions.rawValue)
+        print("//// did request reader input")
+    }
+
+    public func terminal(_ terminal: Terminal, didRequestReaderDisplayMessage displayMessage: ReaderDisplayMessage) {
```
allendav, line 283: Maybe add a comment that didRequestReaderDisplayMessage means the reader has a message for us to display to the buyer

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -442,17 +442,19 @@ extension OrderDetailsViewModel {
         stores.dispatch(deleteTrackingAction)
     }
 
-    func collectPayment() {
+    func collectPayment(onCompletion: @escaping (Result<Void, Error>) -> Void) {
         // Mock Payment. We will have to instantiate these parameters with
         // the proper amount, currency and readable descripitions later.
+        // We could pull the amount from the order but in test mode the Stripe Terminal SDK fails to process amounts that do not end in 00
         let paymentParameters = PaymentParameters(amount: 100,
```
allendav, line 449: Is this $100 or $1? If the former, could we change this to  something like $1 or $5 just in case we accidentally do this live - I generally like to test WooCommerce with dollar amounts that wouldn't stress me out if they were to turn out to be live charges.

Hardware/Hardware/CardReader/CardReaderService.swift
ctarda: Let's pass the id only, we can always revert later on if we find we need to. I'll update.

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
ctarda: I am not sure I follow. Eventually we are going to have to update this to pass the order's total amount 

[The documentation says](https://stripe.com/docs/terminal/payments#create-payment): 
> Use test amounts to try producing different results. An amount ending in 00 results in an approved payment.

I don't know if that answers your question though?

ctarda:
Thanks for the first pass @allendav . This is ready for review

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: No worries. I'm asking Karl if the amount is in cents and how that might vary by currency code

allendav (approved):
Tests fine. Love the emojis. Probably can't release it that way :trollface: 

That said,  :shipit: 

Screen Shot 2021-03-19 at 2 18 59 PM

PR #3848, opened 2021-03-19
Shipping Labels: open Apple Maps or Call the Customer if an error occurs in the address validation form

### LINKED ISSUE #2973: Shipping Labels M2: add option to validate shipping label address details

pmusolino:
Closes #2973

## Description
The last part of the address validation form in Shipping Labels is to show the actions for opening the address in Apple Maps if not found to call the customer.
In this PR, I added the two actions in the top banner view, and I implemented two extensions for opening Apple Maps or for calling the user.

## Testing
Note: _It would be preferable to test on the device, as it will not be possible to call the customer's phone number from the simulator._
2. Open an order detail.
5. Go to the editing address screen and make sure the address is not valid remotely. -> You should see the top banner view with two buttons "Open Map" and "Contact Customer".
6. Tapping "Open Map" you should see that Apple Maps will be opened. If the address is not found on Apple Maps, you should see an error notice.
7. Tapping "Contact Customer" you should be able to call it. If the phone is not valid or the device doesn't support phone calls, you should see an error notice.

## Screenshots

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Tools/MapsHelper.swift
+++ b/WooCommerce/Classes/Tools/MapsHelper.swift
@@ -1,6 +1,5 @@
 import MapKit
 
-
 final class MapsHelper {
 
     /// The method accept a string of the address info you already have, and open Apple Maps on that address if found.
--- a/WooCommerce/Classes/Tools/MapsHelper.swift
+++ b/WooCommerce/Classes/Tools/MapsHelper.swift
@@ -1,0 +1,41 @@
+import MapKit
+
+
+final class MapsHelper {
+
+    /// The method accept a string of the address info you already have, and open Apple Maps on that address if found.
+    ///
+    static func openAppleMaps(address: String?, completion: @escaping (Result<Void, MapsHelperError>) -> Void) {
+        guard let address = address else {
+            completion(.failure(.locationNotFound))
+            return
+        }
+        CLGeocoder().geocodeAddressString(address) { (placemarksOptional, error) -> Void in
+              if let placemarks = placemarksOptional {
+                DDLogInfo("First address found in Apple Maps: \(String(describing: placemarks.first))")
+                if let location = placemarks.first?.location {
+                  let query = "?ll=\(location.coordinate.latitude),\(location.coordinate.longitude)"
+                  let path = "http://maps.apple.com/" + query
+                  if let url = URL(string: path) {
+                    UIApplication.shared.open(url, options: [:], completionHandler: nil)
+                    completion(.success(()))
+                  } else {
+                    completion(.failure(.constructURL))
+                  }
+                } else {
+                    completion(.failure(.locationNotFound))
+                }
+              } else {
+                completion(.failure(.locationNotFound))
+              }
+            }
+    }
+
+    enum MapsHelperError: Error {
+        // Could not construct url.
+        case constructURL
+
+        // Could not get a location from the geocode request.
+        case locationNotFound
+    }
+}
--- a/WooCommerce/Classes/Tools/MapsHelper.swift
+++ b/WooCommerce/Classes/Tools/MapsHelper.swift
@@ -10,24 +10,24 @@ final class MapsHelper {
             return
         }
         CLGeocoder().geocodeAddressString(address) { (placemarksOptional, error) -> Void in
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/Tools/MapsHelper.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    /// The method accept a string of the address info you already have, and open Apple Maps on that address if found.
+    ///
+    static func openAppleMaps(address: String?, completion: @escaping (Result<Void, MapsHelperError>) -> Void) {
+        guard let address = address else {
+            completion(.failure(.locationNotFound))
+            return
+        }
+        CLGeocoder().geocodeAddressString(address) { (placemarksOptional, error) -> Void in
+              if let placemarks = placemarksOptional {
```
ctarda: This is a personal preference, but I tend to find nested if/elses difficult to read. Maybe separating the first if as a `guard let placemarks = placemarksOptional else { completion(.failure(.locationNotFound)) }` could help remove one else later on, kind of flatten things one level?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -18,7 +18,21 @@ final class ShippingLabelAddressFormViewController: UIViewController {
...
+            MapsHelper.openAppleMaps(address: self?.viewModel.address?.formattedPostalAddress) { [weak self] (result) in
+                switch result {
+                case .success:
+                    break
+                case .failure:
+                    self?.displayAppleMapsErrorNotice()
+                }
+            }
+        } contactCustomerPressed: { [weak self] in
+            if !PhoneHelper.callPhoneNumber(phone: self?.viewModel.address?.phone) {
```
ctarda: This seems to be failing for empty phone numbers

ctarda (approved):
There seems to be one issue with the phone number validation. In my case, the phone number is empty, but that allows tapping the "contact customer" button and presents an alert with an empty phone number:


Other than that, it works fine. There is an error notice presented on non-valid addresses:


And:
 

 :shipit: . The validation of empty phone numbers could be fixed before merging or deferred to a separate ticket.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
pmusolino: Good catch, fixed here 991e3160aa4d8368fcb9f6b0b14e84a7a11fee5b

PR #3796, opened 2021-03-05
Add app version to the in-app survey metadata for all types of…

angelinaolmedo7:
…surveys, along with corresponding tests

**Description**
I tagged app version in the metadata for the two cases that lacked it (`.productsM5Feedback`, `.shippingLabelsRelease1Feedback`), and updated the test for product feedback (`test_it_loads_the_correct_product_feedback_survey()`) accordingly. 
Additionally, `.shippingLabelsRelease1Feedback` did not appear to have a test associated with it, so one was created. Please ensure the new test, `test_it_loads_the_correct_shipping_labels_release_survey()`, matches the desired naming conventions. 

Fixes #3777 

**Type of change**
 New feature (non-breaking change which adds functionality)

**How Has This Been Tested?**
The following tests have been added to or updated in SurveyViewControllerTests.swift to ensure the changes work as expected: 
`test_it_loads_the_correct_product_feedback_survey()`
`test_it_loads_the_correct_shipping_labels_release_survey()`

**Checklist:**
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

**Update release notes:**
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3851, opened 2021-03-19
Add variations announcement banner

ealeksandrov:
## Description

This PR adds a banner on top of Products List asking user feedback about Product Variations feature.

Following beta test feedback from: p5T066-21p-p2#comment-7623

## Changes

- Updated `ProductsTopBannerFactory` to have different content depending on banner type
- Added `productsVariations` FeedbackType to be stored in settings
- Added `productsVariationsFeedback` source for `SurveyViewController`
- Added `products_variations` feedback context for analytics
- Updated `ProductsViewController` to show Variations banner only after ProductsM5

**Note:** in code there is a little confusion between M4 and M5 for products. Analytics context string is `products_m4` while some of the code refers to this feedback as M5. There was no change applied, M4 naming is just more exposed now out of banner code due to parameters injection. Maybe worth additional rename to M5?

## Test

1. Open products list
2. If products M5 feedback banner is displayed - dismiss it and restart the app
3. Observe request for variations feedback
4. Dismiss it and make sure it won't appear again

## Screenshot

Update release notes:

### DIFF
--- a/Storage/Storage/Model/FeedbackType.swift
+++ b/Storage/Storage/Model/FeedbackType.swift
@@ -9,6 +9,10 @@ public enum FeedbackType: String, Codable {
     ///
     case productsM5
 
+    /// Identifier for the Products Variations.
+    ///
+    case productsVariations
+
     /// identifier for the shipping labels m1 feedback survey
     ///
     case shippingLabelsRelease1
--- a/Storage/Storage/Model/FeedbackType.swift
+++ b/Storage/Storage/Model/FeedbackType.swift
@@ -5,10 +5,6 @@ public enum FeedbackType: String, Codable {
     ///
     case general
 
-    /// Identifier for the Products M5: Linked Products, Downloadable Files, Trashing.
-    ///
-    case productsM5
-
     /// Identifier for the Products Variations.
     ///
     case productsVariations
--- a/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
@@ -61,6 +61,8 @@ extension WooAnalyticsEvent {
         /// Shown in products banner for Milestone 4 features. New product banners should have
         /// their own `FeedbackContext` option.
         case productsM4 = "products_m4"
+        /// Shown in products banner for Variations release.
+        case productsVariations = "products_variations"
         /// Shown in shipping labels banner for Milestone 1 features.
         case shippingLabelsRelease1 = "shipping_labels_m1"
     }
--- a/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
@@ -58,9 +58,6 @@ extension WooAnalyticsEvent {
     public enum FeedbackContext: String {
         /// Shown in Stats but is for asking general feedback.
         case general
-        /// Shown in products banner for Milestone 4 features. New product banners should have
-        /// their own `FeedbackContext` option.
-        case productsM4 = "products_m4"
         /// Shown in products banner for Variations release.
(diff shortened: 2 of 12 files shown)


AliSoftware:
@ealeksandrov  I'm code-freezing 6.3 today, so moved this PR's milestone to 6.4.

Feel free to ping me if you need it to be included in 6.3 anyway (in which case we can re-target it to the `release/6.3` branch and make a new beta once it's merged)

Storage/Storage/Model/FeedbackType.swift
```diff
@@ -9,6 +9,10 @@ public enum FeedbackType: String, Codable {
     ///
     case productsM5
```
pmusolino: I think that productsM5 can be removed from here since they are enabled for everyone.

WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
```diff
@@ -61,6 +61,8 @@ extension WooAnalyticsEvent {
         /// Shown in products banner for Milestone 4 features. New product banners should have
         /// their own `FeedbackContext` option.
         case productsM4 = "products_m4"
```
pmusolino: Do it make sense to remove products m4 from here?

pmusolino (commented):
The test LGTM 👍 
Just, I don't think that we want to maintain products M4 logic for the banner, since the features in M4 are enabled by months now.

ealeksandrov:
@pmusolino thank you for review!
Can you clarify on products M4/M5 banner - do you suggest removing it completely? All that settings storage and analytics context are required for it to work.

pmusolino:
@ealeksandrov I think that they are obsolete since we are not showing any more the banner for M4 and M5, but if you think that would be good to leave them until M6, it's ok for me.

ealeksandrov:
@pmusolino I didn't know history of products screen well, but now I understand and agree that M5 banner is obsolete 🙂
Removed M5 banner, ready for another look.

WooCommerce/Classes/ViewRelated/Survey/SurveyViewController.swift
```diff
@@ -74,11 +74,10 @@ extension SurveyViewController {
                     .tagPlatform("ios")
                     .tagAppVersion(Bundle.main.bundleVersion())
 
-            case .productsM5Feedback:
-                return WooConstants.URLs.productsM4Feedback
+            case .productsVariationsFeedback:
+                return WooConstants.URLs.productsFeedback
                     .asURL()
                     .tagPlatform("ios")
-                    .tagProductMilestone("5")
```
ealeksandrov, line 81: Maybe we need `product-milestone` = 6 for variations here?

cc @shiki who worked on previous version of this banner

shiki: I don't exactly remember why we have this tag. Do you remember, @pmusolino? 

What I can deduce is that this would be useful to identify what product-related features are available when the merchant submitted the feedback. Let's say, the feedback is “I cannot add variations” and the tagged milestone is 5. With that, we'll know that the feedback is no longer valid because we have the new variations feature available now. 

But since app versions are now included in the tag (https://github.com/woocommerce/woocommerce-ios/pull/3796), I don't think this is necessary anymore. 

Also, we stopped using Milestone numbers at some point. 🤷

ealeksandrov: Removed `tagProductMilestone` completely in 358d045, since it's unused now.

pmusolino: I don't remember why we are using this tag 🤔 But it seems superfluous now.

pmusolino (approved):
LGTM :shipit:

PR #3852, opened 2021-03-21
[Mobile Payments] Add "Collect Payment" button to order details view

### LINKED ISSUE #3826: [Mobile Payments] Add "Collect Payment" button to order details view
For context: p91TBi-4Af-p2

Add a new button to OrderDetailsViewController that triggers the functionality to Collect a Payment

ctarda:
Closes #3826 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by implementing logic to display a "Collect Payment" button in the order details view.

For a non elegible order:

And for an elegible order:

| Light appearance | Dark appearance |
| ------ | ------ |
|  |  |

## Changes
* Update OrderDetalsViewController, OrderDetailsViewModel and OrderDetailsDataSource, to add a "Collect Payment" button that should be visible only on orders with "Cash On Delivery" as payment method.

## Testing
⚠️ Tapping the button will crash the app, unless a card reader has been paired first. This is expected (for now) ⚠️

* Checkout the branch.
* Navigate to an order with payment method "cash on delivery". 
* Notice the button labeled "Collect Payment" in the payment section.
* If testing with a hardware reader, connect to a reader first (Settings> Manage Card Readers > Connect card reader and follow the prompts) Then navigate back to the order and tap Collect Payment. That should trigger the payment collection flow

Update release notes:


WooCommerce/Classes/Model/OrderPaymentMethod.swift
```diff
@@ -0,0 +1,33 @@
+/// Order Payment methods
+enum OrderPaymentMethod: RawRepresentable {
+    /// Cash on Delivery
+    case cod
```
allendav, line 4: Please add something like `none` - in the REST API response - referring to an empty string for `_payment_method`. This is the payment method automatically assigned to Manually Created Orders and I think it is likely those orders will be something merchants use with Card Present payments (i.e. someone comes in the shop or calls on the phone, they enter the order in front of them and then the customer pays in person with their card when they get the goods.)

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -409,6 +409,19 @@ extension OrderDetailsViewModel {
...
     }
 
+    func checkCardPaymentEligibility(onCompletion: (() -> Void)? = nil) {
+        // For now, this defaults to the feature flag,
+        // combined with the order payment method being cash on delivery
+        // We also need to take into account wheter the store is enrolled into WCPay.
+        // This is synchronous for now, but it will be async in the nearly future:
+        // https://github.com/woocommerce/woocommerce-ios/issues/3828
+        // We need to add a unit test for this, after the logic is settled
+        dataSource.isEligibleForCardPresentPayment = ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentPayments) &&
```
allendav, line 424: Could we add "TODO" to make this easier to find later?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -409,6 +409,19 @@ extension OrderDetailsViewModel {
...
+            OrderPaymentMethod(rawValue: order.paymentMethodID) == .cod
```
allendav: Same comment above about allowing card present payments when paymentMethod is empty

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav, line 424: We should only allow payment on orders with status "pending payment" or "on hold"

We should never allow payment on orders with status "cancelled", "processing," "refunded" or "completed" 

We may want to allow payment on orders with status "failed" or "authentication required" - this could be useful if the card they attempted first was declined and they wish to try again with a card present payment. Or, if you prefer, it would also make sense to disallow these two as well for now and open an issue to settle this later.

ctarda: I think it would be easier if we go with pending payment and on hold for now, until we have more clarity.
Added an issue: #3857

ctarda: I am curious., why not on orders with status "processing"? If I order something on a store (web), that and select "cash on delivery" that'd be the status the order is in, as far as I can tell.

ctarda:
@allendav  I just pushed an update incorporating your feedback, with one question about the mapping between eligibility of the orders and their status

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: I'm surprised! Our [docs](https://docs.woocommerce.com/document/managing-orders/) say:

> Processing — Payment received (paid) and stock has been reduced; order is awaiting fulfillment.

But sure enough, `cod` surprisingly leaves the order in `processing` - this looks like an area of some contention and I think there is a bug in core: https://github.com/woocommerce/woocommerce/pull/15268/files - I'll investigate and follow up with core

In the meantime, yeah, I'd allow `processing` for `cod` too

ctarda: I'll push a quick update for that

allendav: https://github.com/woocommerce/woocommerce/issues/29471

ctarda:
Alright, pushed an other one to show the button in orders with "Processing" status as well. *Now* I'm out.

allendav (approved):
Tests fine. We can defer the COD/processing debate to a subsequent PR. :shipit:

PR #3853, opened 2021-03-22
Use http api calls to trigger release jobs

loremattei:
This PR updates our release tools to trigger any CI release job via an API call instead of relying on the automatic trigger on tags.
This way, it will be easier and cleaner to handle any failure during the release build and, moreover, it'll be possible to move more steps of the releases to CI (for example, validation and lint steps that can fail by design).

Note that in order to use the release tools, users should configure a new token for CircleCI on their local environment, following these steps:

- Go to https://app.circleci.com/settings/user/tokens.
- Create a Personal API Token.
- Add `CIRCLE_CI_AUTH_TOKEN=<Your Token>` to your `~/.wcios-env.default` (you can check out the updated fastlane/env/user.env-example for a template).

This PR needs an updated version of `release-toolkit`, so it's now pointing to the required hash in `develop`. After approval, I'll release a new version of `release-toolkit` and update this PR before merging.

**Test**

Full testing is tricky because it's tight to the release cycle. Anyway, it's possible to test the critical part of it this way:

- Checkout a new `release/test` branch from this one.
- Add a dummy commit (the build will fail, so adding a dummy commit will avoid the failure to be reported to this PR).
- `bundle install`
- Run `bundle exec fastlane trigger_beta_build branch_to_build:"release/test"`
- Verify that the release build job has been started on CircleCI. The job will fail during the upload to AppStore because the current version is already present. This is fine.
- Delete the test branch.

**Review**
I asked the review to @AliSoftware because he is currently the release manager of WooCommerce iOS. All the code is already in production in WPiOS, so it should be fine, but there may always be some nuances with porting that I missed, so I think that understanding the changes will help if something happens during the next release.

Update release notes:


.circleci/config.yml
```diff
@@ -6,6 +6,17 @@ orbs:
   git: wordpress-mobile/git@1.0
   slack: circleci/slack@3.4.2
 
+parameters:
+  beta_build:
+    type: boolean
+    default: false 
+  release_build:
+    type: boolean 
+    default: false
```
AliSoftware, line 15: Same Q as https://github.com/wordpress-mobile/WordPress-Android/pull/14289#discussion_r604285556 : can't we merge those 2 parameters into one?

.circleci/config.yml
```diff
@@ -162,13 +173,27 @@ jobs:
...
+                  bundle exec fastlane build_and_upload_release skip_confirm:true create_gh_release:true
+                no_output_timeout: 60m
+      - when: 
+          condition: << pipeline.parameters.release_build >>
+          steps:
+            - run:
+                name: Build
+                command: | 
+                  bundle exec fastlane build_and_upload_release skip_confirm:true create_gh_release:true
+                no_output_timeout: 60m
```
AliSoftware, line 216: Those 2 steps are running the exact same lane with same parameters, so in practice they could be mutualized as a single `run` step with a `condition: { or: [<< pipeline.parameters.beta_build >>, << pipeline.parameters.release_build >> ] }`.

Which in turn means we could even simplify this even further to:

1. move the condition on the whole "Release Build" workflow like you did for https://github.com/wordpress-mobile/WordPress-Android/pull/14289/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R496-R500
2. Even mutualize those 2 params into a single one

fastlane/Fastfile
```diff
@@ -48,6 +48,9 @@ ENV["DOWNLOAD_METADATA"]="fastlane/download_metadata.swift"
 ENV["PROJECT_ROOT_FOLDER"]="./"
 ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
 
+# Other defines used across multiple lanes
+REPOSITORY_NAME="WordPress-Android"
```
AliSoftware: `woocommerce-ios`

fastlane/Fastfile
```diff
@@ -211,40 +217,77 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+  #
+  #####################################################################################
+  lane :trigger_beta_build do | options |
+    circleci_trigger_job(
+      circle_ci_token: ENV["CIRCLE_CI_AUTH_TOKEN"], 
+      repository: REPOSITORY_NAME, 
+      branch: options[:branch_to_build],
+      job_params: {"beta_build" => true}
+    )
+  end
```
AliSoftware: If we end up unifying the 2 parameters (since they end up executing the same steps in CI), one of these can go as we'd also unify those 2 lanes in one.

fastlane/Pluginfile
```diff
@@ -6,7 +6,7 @@ group :screenshots, optional: true do
   gem 'rmagick', '~> 3.2.0'
 end
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.14.0'
+gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', ref: 'aefec479766afe93ef30696c1c0ca82e91574ecf'
```
AliSoftware, line 9: As you mentioned in your PR description, this needs to be updated once we do a new release. Just highlighting this as a comment so we don't forget before merge.

AliSoftware (changes requested):

.circleci/config.yml
loremattei: Yes, we could. I preferred to keep them separate because in some apps we have different steps for beta and release (e.g. on WPiOS we publish to AppCenter on beta, but not on release) and so I opted to keep the same infrastructure on every app. 
This doesn't add a lot of code to maintain (it's just a parameter. I'm usually against adding a ton of code for things that could be done in 2 lines just for the sake of making it future proof :-) ), but make all the code ready to support specific steps for the two cases.

.circleci/config.yml
loremattei: Same as above

fastlane/Fastfile
loremattei: Good catch. 
There are still a few things to tweak and test. I plan to finish this and to put it out of the "draft" state tomorrow 😅

.circleci/config.yml
AliSoftware: Right, makes sense I guess. But if we want to make all the apps have same infra, we should still uniformize between doing it as 2 steps each with their `when` condition (like in this PR) vs how you did it in the WPAndroid PR (where the "Release Build" job has just one common step, and the condition (with `or` on both params) is on the workflow, not the step.
I don't really have a strong preference on one way or the other, but if the goal is to keep the same infra on every app, we should use the same in both 😉

loremattei: Yeah, right! After reviewing the Android one, I think it doesn't make sense on Android because the build steps are quite different. I've updated the WPAndroid PR.

loremattei:
@AliSoftware this should be ready now (pending the release-toolkit update)

fastlane/Fastfile
```diff
@@ -211,40 +217,79 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+  desc "Builds and uploads for distribution"
   lane :build_and_upload_release do | options |
-    final_tag = (is_ci() == true) ? ios_validate_ci_build() : false
-    create_release = (final_tag && is_ci()) || options[:create_gh_release]
-
     ios_build_prechecks(skip_confirm: options[:skip_confirm], 
       internal: false,
       external: true)
     ios_build_preflight()
     build_and_upload_itc(skip_prechecks: true, skip_confirm: options[:skip_confirm], create_release: create_release)
```
AliSoftware: `create_release` variable is now undefined at that point.
Should we make it an `option[:create_release]` and pass it at call site? or decide to always create a GH release from now on?

fastlane/Fastfile
```diff
@@ -211,40 +217,79 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane trigger_beta_build [branch_to_build:<branch_name>]
+  #
+  #####################################################################################
+  lane :trigger_beta_build do | options |
+    circleci_trigger_job(
+      circle_ci_token: ENV["CIRCLE_CI_AUTH_TOKEN"], 
+      repository: REPOSITORY_NAME, 
+      organization: "woocommerce",
```
AliSoftware: Would make sense to DRY this as `ORG_NAME='woocommerce'` – same as how you _did_ DRY the `REPOSITORY_NAME`.

fastlane/Fastfile
```diff
@@ -211,40 +217,79 @@ ENV["APP_STORE_STRINGS_FILE_NAME"]="AppStoreStrings.pot"
...
+  # -----------------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane trigger_release_build [branch_to_build:<branch_name>]
+  #
+  #####################################################################################
+  lane :trigger_release_build do | options |
+    circleci_trigger_job(
+      circle_ci_token: ENV["CIRCLE_CI_AUTH_TOKEN"], 
+      repository: REPOSITORY_NAME, 
+      organization: "woocommerce",
```
AliSoftware: … so that it can be reused here too.

AliSoftware (changes requested):
Runtime failure on `create_release` undefined variable

fastlane/Fastfile
AliSoftware: Possible alternative to improve all that: we could also automatically derive that from `git remote get-url origin` so that we don't have to explicitly declare the value, making this all more portable (easier to copy/paste that part of the Fastfile from one repo to another) 🤷

fastlane/Fastfile
loremattei: Yeah, good catch.

fastlane/Pluginfile
AliSoftware: Reminder to update ☝️  to at least 0.15 before merging 🙂

AliSoftware (approved):
Worked as expected! 🎉  ([CI build here](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/11074/workflows/57eb83ba-b862-449b-a83f-43c803cc0dff))

Approving, but don't forget to update the release-toolkit reference to a tag before merging 😉

loremattei:
Yup! Thank you @AliSoftware. 
`release-toolkit` has been updated to `0.17.1` and CI is green, so I'm merging this PR.

PR #3856, opened 2021-03-23
[Mobile Payments] Model CardReaderServiceErrors

### LINKED ISSUE #3734: [Mobile Payments] Error handling. Make service propagate errors

ctarda:
Closes #3734 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by implementing logic to set the foundation for proper error handling. The diff contains changes to the documentation that explain the rationale behind the implementation.

## Changes
* Expand the pre-existing `CardReaderServiceError` so that it wraps an error that indicates the reason why a given operation has failed
* These underlying errors are a 1:1 map with the error codes provided by [the Stripe Terminal SDK](https://stripe.dev/stripe-terminal-ios/docs/Enums/SCPError.html). Not every single error code provided by the Stripe SDK is mapped to a domain error. Some of them did not look like would provide much value at this point. We can always add new cases to UnderlyingError later on as we see fit though.
* Update the documentation (Hardware.md) with a mention to how the errors are modelled.

## Testing
There are no user-facing changes here, so testing is a matter of 👀  on the code and the changes to the  documentation and waiting for CI to give the ✅ 
Checking out the branch and building and running wouldn't hurt.

Update release notes:


Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -127,17 +127,15 @@ extension StripeCardReaderService: CardReaderService {
 
     public func createPaymentIntent(_ parameters: PaymentIntentParameters) -> Future<Void, Error> {
         return Future() { promise in
-            // Contains enough implementation just to pass a test on the happy path.
-            // We are not doing any proper error handling yet, but for now we
-            // will propagate an error specific to this operation (.intentCreation)
             Terminal.shared.createPaymentIntent(parameters.toStripe()) { [weak self] (intent, error) in
                 guard let self = self else {
-                    promise(Result.failure(CardReaderServiceError.intentCreation))
+                    promise(Result.failure(CardReaderServiceError.intentCreation()))
```
Ecarrion: nit: Can you omit the `Result` namespace? or does the compiler gets angrier?

Hardware/Hardware/CardReader/StripeCardReader/UnderlyingError+Stripe.swift
```diff
@@ -0,0 +1,84 @@
+import StripeTerminal
+/// Maps Stripe SDK specific errors to domain errors:
+/// the mapping is done according to the error codes documented here:
+/// https://stripe.dev/stripe-terminal-ios/docs/Enums/SCPError.html
+extension UnderlyingError {
+    static func make(with stripeError: NSError) -> Self {
```
Ecarrion: I think having this as an initializer, could read a bit swifter, what do you think?
Also you could receive the stripe error and cast it here, to avoid doing the cast to `NSError` in every call site.

Ecarrion (approved):
Just a couple of nits!

Hardware/Hardware/CardReader/StripeCardReader/UnderlyingError+Stripe.swift
ctarda: Good points!

ctarda:
Thanks for the review @Ecarrion . I just pushed an update to incorporate your feedback 🙇

Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -1,20 +1,152 @@
 /// Models errors thrown by the CardReaderService.
-/// This is doing the bare minimum for now.
-/// Proper error handling is coming in
-/// https://github.com/woocommerce/woocommerce-ios/issues/3734
+/// It identifies the interaction with the card reader
+/// where the error was thrown.
+/// It is important to mark the specific interaction
+/// because some operations (like, for example, procesing a payment)
```
allendav: processing

Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -1,20 +1,152 @@
...
+/// require of three interactions with the SDK in sequence, and any error
```
allendav: <del>of</del>

Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -1,20 +1,152 @@
...
 
     /// Error thrown while capturing a payment
-    case capturePayment
+    case capturePayment(underlyingError: UnderlyingError = .internalServiceError)
+}
+
+
+/// Underlying error. Models the specific error that made a given
+/// interaction with the SDK fail.
+public enum UnderlyingError: Error {
```
allendav, line 28: I think we should consider separate enums for each "underlying error" for each `CardReaderServiceError` case.  If I am writing discovery and connection logic, I shouldn't need to be aware of the underlying error cases for payment method creation.

Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -1,20 +1,152 @@
...
+    /// The service is busy executing another command. The service can only execute a single command at a time.
+    case busy
```
allendav, line 30: So, some of these errors are because we coded something wrong, e.g. busy, or because the user did something wrong, e.g. left a card in the reader from an earlier transaction.

Instead of having the caller have to figure out what to do in each case, what if the enumeration provided a pair of localized user facing string for cases where there is something the user could do, like a whats-wrong-string "Card Left in Reader" and a suggested how-to-maybe-fix-it-action "Please remove your card and try again"

If there is nothing specific the user could do, we could use the whats-wrong-string to bubble up a unique code or something to the end user so they can tell it to us when reporting the error, e.g. "An Unexpected Error Occurred (10223)" and a generic "Please try connecting to the card reader again" how-to-maybe-fix-it-action if any

Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -1,20 +1,152 @@
...
-    case connection
+    case connection(underlyingError: UnderlyingError = .internalServiceError)
 
     /// Error thrown while creating a payment intent
-    case intentCreation
+    case intentCreation(underlyingError: UnderlyingError = .internalServiceError)
 
     /// Error thrown while collecting payment methods
-    case paymentMethod
+    case paymentMethod(underlyingError: UnderlyingError = .internalServiceError)
```
allendav: The other enum case names are event or timeframe like, e.g. during "discovery", during "connection", during "intentCreation" but paymentMethod isn't like that. Could we take this opportunity to rename this case to something like `paymentMethodCollection` ?

Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -1,20 +1,152 @@
...
 
     /// Error thrown while capturing a payment
-    case capturePayment
+    case capturePayment(underlyingError: UnderlyingError = .internalServiceError)
```
allendav: Similarly, could we make this more like the naming order of the others and make this 'paymentCapture` ?

Hardware/Hardware/CardReader/CardReaderServiceError.swift
ctarda: Consider the case where we are chaining several operations in sequence, for example when processing a payment, which requires to create an intent, collect a payment method and then process the intent.

If we have a single error with an underlying error, we can chain those operations like this (as we are doing in the CardPresentPaymentStore at the moment):

```
        cardReaderService.createPaymentIntent(parameters)
            .flatMap {
                self.cardReaderService.collectPaymentMethod()
            }.flatMap {
                self.cardReaderService.processPayment()
            }.sink { error in
            switch error {
            case .failure(let error):
                onCompletion(.failure(error))
            case .finished:
                onCompletion(.success(()))
            }
        } receiveValue: { intent in
            print("==== Yosemite log for testing. Payment intent after processing payment")
            print(intent)
            print("//// payment intent processed ")
            // TODO. Initiate final step. Update order. Submit intent id to backend.
            // Deferred to https://github.com/woocommerce/woocommerce-ios/issues/3825
            onCompletion(.success(()))
        }.store(in: &cancellables)
```

And then let any layer on top of that decide how to react to that error, because it can be aware of the operation where it failed, and the reason why it failed.

allendav: > e.g. error will equal CardReaderServiceError.intentCreation(.whateverfailedinthesdk)

I think the downside is if I want to do something with .whateverfailedinthesdk I'll need to cherry pick cases for ifs, right? Or, god forbid, a full switch. Not having all the errors in one enum reduces the number of enums the UI writer needs to consider writing ifs for (or handling in a comprehensive switch)

I'd like to avoid ending up with the user seeing a default case "An unexpected error occurred"-"OK" and then we get HE's hit with that and we don't have a clear idea what they ran in to

Hardware/Hardware/CardReader/CardReaderServiceError.swift
ctarda: Right. I logged a separate issue for adding user-facing strings to these underlying errors: #3868 

The idea would be that UnderlyingError provides a description that could be used as default for that specific error. 

Later on, the UI could "override" that specific error description if needed.

Hardware/Hardware/CardReader/CardReaderServiceError.swift
ctarda: >  I'd like to avoid ending up with the user seeing a default case "An unexpected error occurred"-"OK" and then we get HE's hit with that and we don't have a clear idea what they ran in to

Absolutely! There is no reason to do that. Still, the way to avoid that is by giving the UI information as granular as possible, right? 😄

Hardware/Hardware/CardReader/CardReaderServiceError.swift
allendav: WFM. Thank you!

Hardware/Hardware/CardReader/CardReaderServiceError.swift
allendav: Yes, and making it as easy as possible to get that granular error information (and any potential corrective action) in front of the users. Looking forward to seeing how this plays out as we continue wiring up the UI.

Thanks!

ctarda:
Thanks for the review @allendav 

I just pushed an update to incorporate your feedback (fixing typos and renamed the error case for collecting the payment method)

As we discussed, I've logged #3868 to pass a description as well. That will give us the granularity we want in the UI to:
- Decide how we present each and every type of error.
- Decide what error description we want to provide to the user (if it is the default one bubbled up from Hardware or a custom message specific to a given UI component)

allendav (approved):
:shipit:

PR #3860, opened 2021-03-24
Improve Jetpack query parameters syntax

### LINKED ISSUE #1980: Improve Jetpack Proxy query parameters syntax
Currently when requests are sent to the remote site, the URL parameters for the endpoint are encoded in the `path` parameter to the WordPress.com endpoint for the Jetpack proxy. There are also cases where the parameters for the remote site are set as actual parameters on the proxy endpoint. The proxy endpoint will pass these parameters along in most cases, but we've seen some strange behavior because of it.

Because of this, let's change how we send parameters to the remote site. Instead of appending them to the `path` property, bundle them all in a JSON object, URL-encode it, and use the `query` parameter on the `rest-api` endpoint. This works for POST requests as well as the body can contain JSON that isn't URL-encoded.

Android: https://github.com/woocommerce/woocommerce-android/issues/2069

ealeksandrov:
Closes https://github.com/woocommerce/woocommerce-ios/issues/1980.

## Description

This PR moves wrapped Jetpack proxy request parameters from `path` to JSON-formatted `query` item.
Related discussion: p91TBi-2wf-p2

## Test

1. Verify that networking works well in the app (like getting all pages of products list)
2. Try POST and DELETE actions - like product update and trash (should be unaffected by PR)
3. Open Wormholy debug and check `query` parameter in calls to `rest-api` endpoint

Update release notes:

### DIFF
--- a/Networking/Networking/Requests/JetpackRequest.swift
+++ b/Networking/Networking/Requests/JetpackRequest.swift
@@ -91,8 +91,12 @@ private extension JetpackRequest {
     var dotcomParams: [String: String] {
         var output = [
             "json": "true",
-            "path": jetpackPath + "&_method=" + method.rawValue.lowercased() + jetpackQueryParams
+            "path": jetpackPath + "&_method=" + method.rawValue.lowercased()
         ]
+
+        if let jetpackQueryParams = jetpackQueryParams {
+            output["query"] = jetpackQueryParams
+        }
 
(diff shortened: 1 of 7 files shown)


pmusolino (approved):
LGTM, I didn't notice any issue in GET/POST/DELETE requests.
:shipit:

PR #3846, opened 2021-03-18
Products: Update navigation bar button behaviour

Ecarrion:
fix #3838 

# Why

Some time ago, when the app could not create products, It made sense to have an update button to commit changes when the product was being modified.

Later, we allowed merchants to create products and added a "publish" button to indicate that the new product was being published/created.

While this works well, there are several unhandled or not-ideal scenarios, as the merchant can change the product type, product status, and product content at any time.

Two example inconveniences / unhandled cases:

- A. There is no easy way to publish an un-published product 
- B. Creating a product with a different status(other than publish) gives the impression that the product will be published when the merchant only wants to save it.

A | B
--- | ---
publish-draft-product | new-pending-product-before

This PR updates the button rendering mechanism to makes sure those cases are handled to be able to show meaningful button actions to the merchant.

# How

- Added `actionButtons` variable and `canShowPublishOption` function to `ProductFormViewModelProtocol`
- Update `ProductFormViewModel` and `ProductVariationFormViewModel` to provide buttons based on a specific set of **rules**
- Update `ProductFormViewController` to render the buttons based on the `ViewModel` indications.
- Update `ProductFormViewController` to render show(or not) a publish option in the "more" action sheet based on the view model
- Rename the "Update" Button copy for "Save".

## Rules
**When creating a product:**
- Add a "Publish" button in the navigation bar when the product has a `.publish` status.
- Add a "Save" button in the navigation bar when the product has a status different than `.publish` and add a "Publish" option in the "more" action sheet.

**When editing a product**
- Add a "Publish" button in the navigation bar when the product has an un-published status.
- If the un-published product has modifications, show the "Save" button and add a "Publish" option in the "more" action sheet.
- For variations, show the "Save" button only when there are changes to the variation.

**When editing a product**
- Only show the "more" action button when required.

# <a name="testing-steps">Testing Steps</a>

- Go to the product tab and tap "+" to create a  product
- See that the "Publish" button is visible
- Change product status to something different than "Published"
- See that the "Save" button is visible
- Tap on the "more" button and see the "Publish" option is visible.

new-product

----

- Go to a variation from a variable product
- See that nor "save" or "publish" buttons are visible
- Modify the variation
- See that the save button is visible.

variation

----

- Go to an existing non-published product (draft or pending review)
- See that the "publish" button is visible
- Make any modification
- See that only the and "save" button are visible
- Tap on the "more" button and see the "Publish" option is visible.

draft-product

----

- Go to an order and tap on one of its products
[...]

PR #3866, opened 2021-03-24
Fix Unable to Delete Variations

shiki:
It looks like this is a regression caused by #3846. The changes in there did not include the `.more` as an option for the action buttons. 

### Before 

https://github.com/woocommerce/woocommerce-ios/blob/3bd504763609fa37b60e16d78c45674f5b7ef2df/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/ProductVariationFormViewModel.swift#L66-L74

### After

https://github.com/woocommerce/woocommerce-ios/blob/f07b64f15a27e4e7b18aea206ce7d5c7ff004c70/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/ProductVariationFormViewModel.swift#L66-L82

### After Ants Existed 🐜 🐜 🐜 

https://user-images.githubusercontent.com/198826/112392090-7c2e0e00-8cbe-11eb-8e0b-45f6c7d50bcf.mp4

## Assumptions

Based on 6.2, we always show the Delete button. So I kept the same behavior here. 

## Testing

1. Navigate to a variation. 
2. Confirm the ants button is visible. 
3. Tap ants. 🐜 🐜 🐜 
4. Tap Delete. Confirm deletion still works. It should! If not, OMG!

## Author Checklist

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
@@ -67,9 +67,9 @@ final class ProductVariationFormViewModel: ProductFormViewModelProtocol {
     var actionButtons: [ActionButtonType] {
         switch (formType, hasUnsavedChanges()) {
         case (.edit, true):
-            return [.save]
+            return [.save, .more]
         default:
-            return []
+            return [.more]
         }
     }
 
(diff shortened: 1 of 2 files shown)


Ecarrion (approved):
Thanks for this @shiki. I totally missed it! 😇

shiki:
Thanks for the quick review! Also, I'm glad that I fixed this correctly. I can still code! Yay! 😅

PR #3867, opened 2021-03-24
Login: New Prologue Screens

### LINKED ISSUE #3824: Login: New Prologue screens
Internal ref: p91TBi-33r-p2

Update the current Prologue screen with new illustrations in a carousel to promote the following features:

* Stats
* Order Management
* Products
* Reviews

rachelmcr:
Fixes: https://github.com/woocommerce/woocommerce-ios/issues/3824

## Description

Updates the current Prologue screen with new illustrations in a carousel to promote the following features:

- Stats
- Order Management
- Products
- Reviews

## Changes

The new Prologue screen and carousel has a few parts:

* `LoginPrologueViewController.swift` is still the main VC, with an updated background image and colors, and unused methods removed. `LoginPrologueViewController.xib` was also updated with these changes. (The RBG values for the color **Gray 0** are also updated to the correct values for this shade.)
* `LoginProloguePageViewController.swift` is the UIPageViewController, with a custom page control, embedded as a child view of `LoginPrologueViewController`. (This is modeled after [the UIPageViewController in WordPressAuthenticator](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/develop/WordPressAuthenticator/Signin/LoginProloguePageViewController.swift).)
* `LoginProloguePages.swift` contains the carousel's unique content (images and text) in `LoginProloguePageType` and shared styles/layout in `LoginProloguePageTypeViewController`.

### Screenshots

/|Stats|Order Management|Products|Reviews
-|-|-|-|-
Light|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 32 39|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 32 41|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 32 43|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 32 46
Dark|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 33 06|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 33 08|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 33 10|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 33 12

iPhone SE|iPhone 11|iPad Pro (11")
-|-|-
Simulator Screen Shot - iPhone SE (2nd generation) - 2021-03-26 at 13 32 18|Simulator Screen Shot - iPhone 11 Pro - 2021-03-26 at 13 32 39|Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2021-03-26 at 13 31 30

## Testing

2. Build and run the app, and confirm the initial (prologue) screen shows a carousel containing four pages with the new illustrations.
3. Confirm the layout displays all of the images/text, on both small and large device sizes.
4. Confirm the carousel and illustrations are reversed if the app is using a right-to-left language (e.g. Arabic, Hebrew).

## Submitter Checklist

Update release notes:


Garance91540 (changes requested):
Hi Rachel! Thank you for working on this, they look so much better than what we currently have <3

Looking at your screenshots, I realized I hadn't included designs for tablets...Oupsie, I just added them to the Figma file:

Screen Shot 2021-03-25 at 4 23 57 PM

rachelmcr:
Thanks @Garance91540! I updated the layout to match the iPad design as closely as I could. I had trouble exactly matching the design without really altering the iPhone layout, so let me know if this is ok or if it needs more work:



(I also updated all the screenshots in the PR description above, so they reflect the latest changes.)

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
```diff
@@ -0,0 +1,112 @@
+import UIKit
+
+/// Displays the Login Prologue carousel, populated with `LoginProloguePageTypeViewController` pages.
+///
+final class LoginProloguePageViewController: UIPageViewController {
+
+    private var pages: [UIViewController] = []
+
+    private var pageControl: UIPageControl!
```
Ecarrion: Can we get rid of this implicitly unwrapped optional? Maybe turning this into a lazy variable?

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        dataSource = self
+        delegate = self
+
+        LoginProloguePageType.allCases.forEach({ type in
+            pages.append(LoginProloguePageTypeViewController(pageType: type))
+        })
```
Ecarrion: What do you think about changing this to something like:
```swift
let pages: [UIViewController]  = {
    LoginProloguePageType.allCases.map { LoginProloguePageTypeViewController(pageType: $0) }
}()
```

That way the pages array can become immutable as it is intended.

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        setViewControllers([pages[0]], direction: .forward, animated: false)
```
Ecarrion: Maybe adding something like
```swift
if let firstPage = pages.first {
    setViewControllers([firstPage], direction: .forward, animated: false)
}
```
Could avoid any crash in the future

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        addPageControl()
+    }
+
+    // MARK: Page Control Setup
+    //
+    private func addPageControl() {
+        let newPageControl = UIPageControl()
+        newPageControl.currentPageIndicatorTintColor = .gray(.shade5)
+        newPageControl.pageIndicatorTintColor = .wooCommercePurple(.shade50)
+        newPageControl.transform = CGAffineTransform(scaleX: 0.8, y: 0.8)
```
Ecarrion: What do you think about extracting this magic number into a constant enum, to be easily changed and referenced?
Also, is this change due to the design?

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        newPageControl.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(newPageControl)
+
+        NSLayoutConstraint.activate([
+            newPageControl.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            newPageControl.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: -10)
```
Ecarrion: Same, about moving this into the enum 😇

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// MARK: - UIPageViewControllerDelegate Conformance
+//
+extension LoginProloguePageViewController: UIPageViewControllerDelegate {
+    func pageViewController(_ pageViewController: UIPageViewController,
+                            didFinishAnimating finished: Bool,
+                            previousViewControllers: [UIViewController],
+                            transitionCompleted completed: Bool) {
+        let toVC = previousViewControllers[0]
```
Ecarrion: I think this is worth to be included in the guard statement below.
```swift
guard let toViewController = previousViewControllers.first, ... {
}
```

WooCommerce/Classes/Authentication/Prologue/LoginProloguePages.swift
```diff
@@ -0,0 +1,130 @@
...
+// MARK: - View Controller
+
+/// Simple container for each page of the login prologue carousel.
+///
+class LoginProloguePageTypeViewController: UIViewController {
+    private let stackView = UIStackView()
+    private let titleLabel = UILabel()
+    private let imageView = UIImageView()
+
+    private var pageType: LoginProloguePageType!
```
Ecarrion: I don't think this needs to be an Implicitly unwrapped optional, right?

WooCommerce/Classes/Authentication/Prologue/LoginProloguePages.swift
```diff
@@ -0,0 +1,130 @@
...
+        view.addSubview(stackView)
+
+        // Stack view layout
+        stackView.axis = .vertical
+        stackView.alignment = .center
+        stackView.spacing = 40
+
+        // Reduce centerYAnchor constraint priority to ensure the bottom margin has higher priority, so stack view is fully visible on shorter devices
+        let verticalCentering = stackView.centerYAnchor.constraint(equalTo: view.centerYAnchor, constant: 103)
+        verticalCentering.priority = UILayoutPriority(999)
```
Ecarrion: nit: you can do `verticalCentering.priority = required - 1` in case the default required priority changes from 1000 in the future.

WooCommerce/Classes/Authentication/Prologue/LoginProloguePages.swift
```diff
@@ -0,0 +1,130 @@
...
+
+        // Set constraints
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            verticalCentering,
+            stackView.bottomAnchor.constraint(lessThanOrEqualTo: view.bottomAnchor, constant: -57),
```
Ecarrion: Probably worth moving all these magic numbers no a constant enum, and provide a comment if necessary.

Ecarrion (commented):
Nice work @rachelmcr!
I left you some suggestions to gain a bit more consistency with the rest of the project. 
Also, I must admit that seeing so many magic numbers makes me a bit anxious coz I can't be sure if it will scale well a cross all screen sizes 😁

rachelmcr:
Thanks for the review!

> Also, I must admit that seeing so many magic numbers makes me a bit anxious coz I can't be sure if it will scale well a cross all screen sizes 😁

I feel the same! I think in some of our other VCs there are similar magic numbers but they are less obvious because they're set in a `.xib` file, instead of in code. That said, as I work through your feedback, I'll try to think if there are ways to remove some of these magic numbers.

Garance91540 (approved):
👏

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
rachelmcr: > Also, is this change due to the design?

Yeah, I was trying to size the page control according to the design. Is that ok? If we don't usually do that, and want to just have the default page control size, we can remove this line.

rachelmcr:
@Ecarrion Ready for another look! I moved the layout constants into enums, but I couldn't come up with any better ways to avoid those constants without deviating further from the designs. If there's another way to think about the constraints that you think would be more reliable, I'm definitely open to trying it; I might be unaware of a better way to construct it.

I've tested across a variety of devices (and it's a bit simpler because we don't support landscape on phones), and so far I think this approach works ok. There are some layout issues with the very largest dynamic type font sizes, but no functional issues.

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
```diff
@@ -4,9 +4,11 @@ import UIKit
 ///
 final class LoginProloguePageViewController: UIPageViewController {
 
-    private var pages: [UIViewController] = []
+    private let pages: [UIViewController] = {
+        LoginProloguePageType.allCases.map { LoginProloguePageTypeViewController(pageType: $0) }
+    }()
 
-    private var pageControl: UIPageControl!
+    private lazy var pageControl = UIPageControl()
```
Ecarrion: Can't this be `private let pageControl = UIPageControl()` and just configure it below, and not re-assign it?

Ecarrion (approved):
Looks great! Thanks for taking the time on this! 🚀

WooCommerce/Classes/Authentication/Prologue/LoginProloguePageViewController.swift
rachelmcr: 🤦  Yep, I don't know why I didn't do it that way the first time around! I'll make that change.

PR #3872, opened 2021-03-25
Unify navigation bar back button styles

### LINKED ISSUE #3760: Back button management
I've noticed that there is a certain disparity when dealing with back button styles across the app.

We should unify them all to only show the back arrow `<` and no text whatsoever.
Additionally, it will be nice if we can find a way to automate that setup.

Disparities:

  Title Text  | No Text | Placeholder Text
--- | --- | ---
Screen Shot 2021-02-26 at 2 46 46 PM | Screen Shot 2021-02-26 at 2 47 04 PM | Screen Shot 2021-02-26 at 2 46 21 PM

Ecarrion:
fix #3760 

# Why

We have spotted that the back button style was not consistent across the app. This happens because the back button configuration is handled in each view controller, which makes it easy for the dev to forget.

 Title Text  | No Text | Placeholder Text
--- | --- | ---
Screen Shot 2021-02-26 at 2 46 46 PM | Screen Shot 2021-02-26 at 2 47 04 PM | Screen Shot 2021-02-26 at 2 46 21 PM

This PR fixes that by placing the back button style management in our `WooNavigationController` so any view controller managed by it gets configured with the correct back button.

# How
- Update our `removeNavigationBackBarButtonText()` method to use the new iOS14+ API for back buttons.
- Replace usages of `UINavigationController` in favor of `WooNavigationController`
- Update `WooNavigationController` to configure the back button of managed VCs.
- Update `WooNavigationController` to forward the navigation delegate events(as it captures the navigation delegate), for cases where this class is used as a parent class. EG: `WooTabNavigationController`
- Remove the usage of `removeNavigationBackBarButtonText()` throughout VCs

# Testing Steps
- Smoketest the app on IOS 14 and make sure the back button does not have any text and is consistent across the app.
- Smoketest the app on IOS 13 and make sure the back button does not have any text and is consistent across the app.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIViewController+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UIViewController+Helpers.swift
@@ -15,7 +15,11 @@ extension UIViewController {
     /// Removes the text of the navigation bar back button in the next view controller of the navigation stack.
     ///
     func removeNavigationBackBarButtonText() {
-        navigationItem.backBarButtonItem = UIBarButtonItem(image: UIImage(), style: .plain, target: nil, action: nil)
+        if #available(iOS 14.0, *) {
+            navigationItem.backButtonDisplayMode = .minimal
+        } else {
+            navigationItem.backBarButtonItem = UIBarButtonItem(image: UIImage(), style: .plain, target: nil, action: nil)
+        }
     }
 
     /// Show the X close button or a custom close button with title on the left bar button item position
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,7 @@
 
 6.4
 -----
-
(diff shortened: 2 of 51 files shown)


WooCommerce/Classes/Extensions/UIViewController+Helpers.swift
```diff
@@ -15,7 +15,11 @@ extension UIViewController {
     /// Removes the text of the navigation bar back button in the next view controller of the navigation stack.
     ///
     func removeNavigationBackBarButtonText() {
-        navigationItem.backBarButtonItem = UIBarButtonItem(image: UIImage(), style: .plain, target: nil, action: nil)
+        if #available(iOS 14.0, *) {
+            navigationItem.backButtonDisplayMode = .minimal
```
ealeksandrov, line 19: 👏

ealeksandrov (approved):
Great work unifying this in a single place! :shipit:

Ecarrion:

PR #3876, opened 2021-03-26
Update CircleCI script to fix wrong iOS sim usage

ealeksandrov:
## Description

I've noticed that our UI tests jobs (for "iPhone 11" and "iPad Air 4") both output ([example here](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/10897/workflows/e25833d5-4260-4506-98b0-f86ebc7bd617/jobs/19663/parallel-runs/0/steps/0-108)):
```
Found simulator "iPhone 8 (13.5)"
```
It's not even iOS 14! 

After a small investigation I found similar issue with a fix already applied in WordPress-iOS:  https://github.com/wordpress-mobile/WordPress-iOS/pull/15261. Check for additional details and discussion there, since this PR is almost exact duplicate. Thanks @guarani!

## Changes

- Removed `ios/boot-simulator` and `ios/wait-for-simulator` steps
- Instead added `device` and `deployment_target_version` arguments to fastlane actions `run_tests` and `multi_scan`
- Cleaned up parameters everywhere to keep Xcode version, devices and iOS versions list in single place on top

## Test

1. Open "Build Tests" job
  a. Check "Build for Testing" step -> "Summary for scan" section -> ensure that the `device` parameter is "iPhone 11"
  b. Ensure the job completes successfully
2. Open "Unit Tests" job
  a. Expect to no longer see "Boot simulator" or "Wait for simulator" steps
  b. Check "Run Unit Tests" step -> search for "SimDevice" -> ensure value is `iPhone 11 ... iOS 14.0`
  c. Ensure job completes successfully
3. Open "UI Tests" jobs for iPhone and iPad
  a. Expect to no longer see "Boot simulator" or "Wait for simulator" steps
  b. Check "Run UI Tests" step -> search for "SimDevice" -> ensure value is `iPhone 11 ... iOS 14.0` and `iPad Air (4th generation) ... iOS 14.0` according to job device
  c. Ensure job completes successfully
4. Open "Installable Build" job
  a. Ensure the job completes successfully

Update release notes:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -210,7 +210,7 @@ workflows:
           requires: [ "Build Tests" ]
       # Always run UI tests on develop and release branches
       - UI Tests:
-          name: UI Tests (iPhone 11)
+          name: UI Tests (iPhone)
           <<: *iphone_test_device
           post-to-slack: true
           requires: [ "Build Tests" ]
@@ -220,7 +220,7 @@ workflows:
                 - develop
                 - /^release.*/
       - UI Tests:
-          name: UI Tests (iPad Air 4th generation)
+          name: UI Tests (iPad)
           <<: *ipad_test_device
           post-to-slack: true
           requires: [ "Build Tests" ]
@@ -243,11 +243,11 @@ workflows:
           <<: *iphone_test_device
           requires: [ "Hold" ]
       - UI Tests:
-          name: Optional UI Tests (iPhone 11)
+          name: Optional UI Tests (iPhone)
           <<: *iphone_test_device
           requires: [ "Build Tests" ]
       - UI Tests:
-          name: Optional UI Tests (iPad Air 4th generation)
+          name: Optional UI Tests (iPad)
           <<: *ipad_test_device
           requires: [ "Build Tests" ]
   Installable Build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,17 @@ orbs:
   git: wordpress-mobile/git@1.0
   slack: circleci/slack@3.4.2
 
+xcode_version: &xcode_version
+  xcode-version: "12.0.0"
+
+iphone_test_device: &iphone_test_device
+  device: iPhone 11
+  ios-version: "14.0"
+
+ipad_test_device: &ipad_test_device
+  device: iPad Air (4th generation)
+  ios-version: "14.0"
+
 commands:
   fix-image:
     steps:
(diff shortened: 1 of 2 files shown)


AliSoftware:
Overall the code LGTM, thanks for tackling this! 👍 👌 

I'll wait for you to (1) undraft this PR when ready (2) list the test steps (probably copy/paste from the WP PR) in the PR description, before submitting an actual GH review 🙂 (including maybe create a dummy commit that would make the test fail on purpose to validate the "test tailed" slack messages work too?)

ealeksandrov:
Thanks for hints @AliSoftware! PR is ready for review now.

I've waited in draft to see output of all jobs. Probably I should have assigned reviewers only after it's green and ready 😅
Slack triggering workflow is here: https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/10899/workflows/5423195a-a856-486d-99e3-ac1dfbb8bfcb, message is here: p1616764775224400-slack-C6H8C3G23

One small difference is that WordPress-iOS test runs have `Booted` state of sim, while Woo-iOS has `Shutdown` state. No impact on tests, but maybe there are some other differences missed between test setups.

.circleci/config.yml
```diff
@@ -199,7 +221,7 @@ workflows:
                 - /^release.*/
       - UI Tests:
           name: UI Tests (iPad Air 4th generation)
```
mokagio: To test this PR, I changed the iPhone version from iPhone 11 to iPhone 12. When I went to unblock the UI tests, I was surprised for a second they mentioned "iPhone 11".

I'd suggest removing the device name from the job name.


```suggestion
          name: UI Tests iPad
```

.circleci/config.yml
```diff
@@ -218,14 +240,15 @@ workflows:
                 - develop
                 - /^release.*/
       - Build Tests:
+          <<: *iphone_test_device
           requires: [ "Hold" ]
       - UI Tests:
           name: Optional UI Tests (iPhone 11)
```
mokagio: As mentioned above

```suggestion
          name: Optional UI Tests iPhone
```

mokagio (approved):
I tested it with [a commit that changed the iPhone device to "iPhone 12"](https://github.com/woocommerce/woocommerce-ios/commit/affbcf5f5f6b768c662da2fd4b08a9300a0a806c). I didn't take the time to change the iOS version.

✅ I can see "Build Tests" uses the device value I [specified](https://github.com/woocommerce/woocommerce-ios/commit/affbcf5f5f6b768c662da2fd4b08a9300a0a806c) in `iphone_test_device`

image

✅ So did the "Unit Tests" step

image

✅ The UI tests attempted to use the given devices and OS (more below on the attempted bit)

✅ The Installable Build worked, but that had no dependency on the changes from this PR 👍 

Approving this to avoid delays. I'd like the suggestion to remove the device name from the UI tests job name to be considered before merging, though.

---

The UI tests for the iPhone 12 failed to find the Simulator version

image

I _think_ that's because the GM build didn't have the image given that it was released before the new devices were announced? I did not take the time to verify this, though.

Given Apple has released Xcode 12.4 a while back, I think we should open a followup PR that upgrades CI to use 12.4.

ealeksandrov:
@mokagio thanks for review and testing! I've removed hardcoded devices names in bd49fb7.

I agree about separate Xcode version bump (and maybe iOS too).

PR #3879, opened 2021-03-26
[Mobile Payments] Add support for metadata to PaymentIntentParameters. Add an extra layer of validation to the parameters passed to the Stripe Terminal SDK

### LINKED ISSUE #3835: [Mobile Payments] Add support for metadata to PaymentIntentParameters
We will need to support at least passing the orderID when creating a PaymentIntent

### LINKED ISSUE #3877: [Mobile Payments] Validate length of descriptions in PaymentIntentParameters
The property statementDescriptor in StripeTerminal.PaymentIntentParameters must comply to a set of validation rules:

https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)statementDescriptor

The issue with this is that if we pass a string that does not pass that validation, the SDK fires an assertion and the process fails.

One potential solution is a mix of using a property wrapper in Hardware.PaymentIntentParameters to enforce those constraints, combined with some string manipulation when generating the StripeTerminal.PaymentIntentParameters, removing special characters and limiting the length of the string

ctarda:
Closes #3877 and #3835

The integration with the Stripe Terminal SDK provides a way to create Payment Intents on device. In order to create a Payment Intent we need to provide an instance of a class ([SCPaymentIntent](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html)) that wraps a set of parameters. 

The Stripe Terminal SDK imposes certain constrains on some of those parameters, for example:
* [Currency](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)currency) must be lowercase and meet the ISO standard
* The [descriptor that will shown in statements](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)statementDescriptor) cannot contain certain characters and must be of a certain length.
* The [email](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)receiptEmail) passed to generate a receipt must be a valid email address

Passing data that does not comply to those restrictions will make the creation of a PaymentIntent fail. 

So this PR does not completely solve the problem of data validation, which will need to be performed on the UI anyway (for example, validating that an email address provided by the user is valid) but adds one more extra layer of protection.

In other words, this might not be technically needed, and we surely should not rely on validating data this low in the stack, but I guess it does not hurt to have some extra protection. And property wrappers are fun.

## Changes
* Add support for a metadata dictionary. This might be needed to pass some extra data like a zip code.
* Add three property wrappers: one that validates an email address, another one that forces a string to be lowercase, and a third one that removes a set of characters and trims a string.
* Add some tests

## Testing.
* There are no user facing changes, so 👀  on the code and waiting for CI to ✅ could be enough.
* Processing a payment would not hurt. For instructions on how to process a payment, see #3847 

Update release notes:


Hardware/Hardware/CardReader/Lowercase.swift
```diff
@@ -0,0 +1,23 @@
+/// A property wrapper to return a string as lowercased.
+/// If there is a way to enforce that the content of the string is
+/// a valid ISO three-letter currency code, this would be a good place to add it
+/// We do this to add an extra layer of validation to the values that we
+/// pass to the Stripe Terminal SDK when creating a payment intent:
+/// https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)currency
+@propertyWrapper
```
allendav: Could we use https://developer.apple.com/documentation/foundation/nslocale/1417834-isocurrencycodes ?

Edit: We'd still have to handle converting to lowercase of course.

Hardware/Hardware/CardReader/StatementDescriptor.swift
```diff
@@ -0,0 +1,40 @@
+/// A property wrapper to return a string where:
+/// - The characters <>"' are replaced.
+/// - The maximum length of the string is 22 characters
+/// We do this to add an extra layer of validation to the values that we
+/// pass to the Stripe Terminal SDK when creating a payment intent:
+/// https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)statementDescriptor
+/// If we pass strings longer than 22 characters or that contain any of the
+/// characters that are not allowed, the Stripe Terminal SDK will crash.
+@propertyWrapper
```
allendav, line 9: On the web side, we found it useful to trim leading (or trailing) spaces too - for some reason people would enter those and then the descriptor string would get truncated in unexpected ways

Hardware/Hardware/CardReader/Email.swift
```diff
@@ -0,0 +1,31 @@
...
+            return validate(email: value) ? value : nil
+        }
+        set {
+            value = newValue
+        }
+    }
+
+    private func validate(email: Value?) -> Bool {
+        guard let email = email else { return false }
+        let emailRegEx = "[A-Z0-9a-z._%+-]+@[A-Za-z0-9.-]+\\.[A-Za-z]{2,64}"
```
allendav, line 28: Maybe add comment pointing at http://emailregex.com/

allendav (changes requested):
A couple of suggested changes. PaymentIntentParametersTests pass.

ctarda:
Thanks for the review @allendav . This is ready for another round:
* I renamed Lowercase.swift to CurrencyCode.Swift, which seems to reflect the intent better
* The conversion to the corresponding Stripe model object will fail now if the currency code is not declared in Locale.isoCurrencyCodes
* And the service will emit an error in that case as well.

allendav (approved):
Thank you for making the changes. All the tests under `PaymentIntentParametersTests` pass.

:shipit:

ctarda:
Thanks for following up @allendav 🙇  Merging now!

PR #3881, opened 2021-03-27
[Mobile Payments] Add support for querying the WCPay account endpoint

allendav:
**Ready for Review - Don't let the size scare you. Roughly half of it is JSON responses for the tests.**

Partially addresses https://github.com/woocommerce/woocommerce-ios/issues/3839

## Changes

Adds a remote for fetching selected WCPay account details from the self-hosted site. This is needed in order to know
- if we should allow the merchant to scan for readers or not (if they don't have an account, token fetching will fail)
- what statement descriptor to use in payment intents (see https://github.com/woocommerce/woocommerce-ios/pull/3879 )
- if the merchant has an account, but it has been rejected (i.e. such that we should not attempt card present payments)
- if the merchant has an account, but it has been restricted (i.e. such that we should not attempt card present payments - and we could display a prompt encouraging them to rectify that)
- what currencies the site supports
- what country the underlying stripe account is in

Note: Restricts query to _fields we need
Note: Requires https://github.com/Automattic/woocommerce-payments/pull/1403 to actually retrieve account

## Testing Instructions

- Ensure WCPayRemoteTests (now 13) all pass

## Advanced Testing Instructions
- Since this is the networking layer, there isn't a way to trigger this from the UI (yet), but if you really want to...
- Install a WooCommerce Payments plugin using this branch https://github.com/woocommerce/woocommerce-ios/issues/3839
- Try any of the following: A site with WCPay active but no on-boarding completed, a site with WCPay account using a dev mode/test account only, or a site with WCPay active using a live for-reals $$$$ account
- Hack something like SettingStore.swift to set a breakpoint to see the response for your site and the mapping to WCPayAccount, e.g.

```diff
diff --git a/Yosemite/Yosemite/Stores/SettingStore.swift b/Yosemite/Yosemite/Stores/SettingStore.swift
index cb5ac8b18..4c2cdd811 100644
--- a/Yosemite/Yosemite/Stores/SettingStore.swift
+++ b/Yosemite/Yosemite/Stores/SettingStore.swift
@@ -8,6 +8,7 @@ import Storage
 public class SettingStore: Store {
     private let siteSettingsRemote: SiteSettingsRemote
     private let siteAPIRemote: SiteAPIRemote
+    private let wcPayRemote: WCPayRemote
 
     private lazy var sharedDerivedStorage: StorageType = {
         return storageManager.newDerivedStorage()
@@ -16,6 +17,7 @@ public class SettingStore: Store {
     public override init(dispatcher: Dispatcher, storageManager: StorageManagerType, network: Network) {
         self.siteSettingsRemote = SiteSettingsRemote(network: network)
         self.siteAPIRemote = SiteAPIRemote(network: network)
+        self.wcPayRemote = WCPayRemote(network:network)
         super.init(dispatcher: dispatcher, storageManager: storageManager, network: network)
     }
 
@@ -62,6 +64,9 @@ private extension SettingStore {
                 onCompletion(nil)
             }
         }
+        wcPayRemote.loadAccount(for: siteID) { [weak self] (result) in
+
+        }
     }
 
     /// Synchronizes the product site settings associated with the provided Site ID (if any!).
```

Update release notes:


Networking/Networking/Remote/WCPayRemote.swift
```diff
@@ -19,4 +18,34 @@ public class WCPayRemote: Remote {
...
+// MARK: - Constants!
+//
+private extension WCPayRemote {
+
+    enum AccountParameterKeys {
+        static let fields: String = "_fields"
+    }
+
+    enum AccountParameterValues {
+        static let fieldValues: String = "status, has_pending_requirements, has_overdue_requirements, current_deadline, statement_descriptor, store_currencies, country"
```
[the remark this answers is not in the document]
allendav: I'll fix this

Networking/Networking/Mapper/WCPayAccountMapper.swift
```diff
@@ -0,0 +1,47 @@
...
+        /// Unfortunately other properties use other formats for dates, but we
+        /// can cross that bridge when we need those decoded.
+        decoder.dateDecodingStrategy = .secondsSince1970
+
+        /// Detect the exceptional case where we got a response of data:[] indicating
+        /// that the plugin is active but the merchant has not on-boarded (and therefore
+        /// has no account.)
+        if let _ = try? decoder.decode(WCPayNullAccountEnvelope.self, from: response) {
+            return WCPayAccount(status: .noAccount)
+        }
```
shiki: Does this mean that the API returns an empty JSON array instead of `null`? 

If this becomes commonplace, we might want to consider having a generic [field-level decoder](https://stackoverflow.com/a/51617547). Not sure if this is feasible though. 😅

Networking/Networking/Model/WCPayAccount.swift
```diff
@@ -0,0 +1,94 @@
...
+
+    /// The public initializer for WCPay Account.
+    ///
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let status = try container.decode(WCPayAccountStatusEnum.self, forKey: .status)
+        let hasPendingRequirements = try container.decode(Bool.self, forKey: .hasPendingRequirements)
+        let hasOverdueRequirements = try container.decode(Bool.self, forKey: .hasOverdueRequirements)
+        /// current_deadline may contain "null" instead of a unix timestamp integer value, so we use decodeIfPresent to handle that properly
+        let currentDeadline = try container.decodeIfPresent(Date.self, forKey: .currentDeadline) ?? WCPayAccount.nullDeadline
```
shiki: ❓  Is there a reason why we don't have the `WCPayAccount.currentDeadline` defined as optonal? 

This is what I receive from the API: 

```json
{
	"data": {
		"has_pending_requirements": false,
		"has_overdue_requirements": false,
		"current_deadline": null,
		"status": "complete",
		"statement_descriptor": "SITE_ADDRESS_OR_SOMETHING",
		"store_currencies": {
			"default": "usd",
			"supported": ["usd"]
		},
		"country": "US"
	}
}
```

I think it makes sense to define `currentDeadline` as optional. But maybe we have a good reason not to do that. If so, we should probably document it. 🙂 

```swift
public let currentDeadline: Date?
```

Networking/Networking/Model/WCPayAccount.swift
```diff
@@ -0,0 +1,94 @@
...
+///
+public struct WCPayAccount: Decodable {
+    public static let nullDeadline = Date(timeIntervalSince1970: 0)
+
+    public let status: WCPayAccountStatusEnum
+    public let hasPendingRequirements: Bool
+    public let hasOverdueRequirements: Bool
+    public let currentDeadline: Date
+    public let statementDescriptor: String
+    public let defaultCurrency: String
```
shiki, line 13: It might be worth it to document what format this is in. If I see this, I would wonder that. Is it uppercase? Is it in lowercase? Or maybe an example value would be nice. 🙂 Or a link to a public API reference? Do we have that?

Networking/Networking/Model/WCPayAccount.swift
```diff
@@ -0,0 +1,94 @@
...
+            defaultCurrency: defaultCurrency,
+            supportedCurrencies: supportedCurrencies,
+            country: country
+        )
+    }
+
+    /// Convenience initializer for the exceptional case where the self-hosted
+    /// site has WCPay active, but the merchant has not on-boarded
+    ///
+    public init(status: WCPayAccountStatusEnum) {
```
shiki: I would have expected that there's a specific `WCPayAccountStatusEnum` value for it. Not `.completed` but `.activeButNotOnboarded`? 😅 

Or is this `.noAccount`? It looks like this is only used in `WCPayAccountMapper`. Perhaps you want to consider having a specific definition for this like: 

```swift
public extension WCPayAccount {
    static let noAccount = WCPayAccount(
        status: .noAccount,
        hasPendingRequirements: false,
        hasOverdueRequirements: false,
        currentDeadline: WCPayAccount.nullDeadline,
        statementDescriptor: "",
        defaultCurrency: "",
        supportedCurrencies: [],
        country: ""
    )
}
```

And then in `WCPayAccountMapper`:

```swift
        if let _ = try? decoder.decode(WCPayNullAccountEnvelope.self, from: response) {
            return WCPayAccount.noAccount
        }
```

In this way, it is less easy to define something like this:

```swift
WCPayAccount(status: .completed)
```

... where it's `.completed` but the expected values (e.g. `defaultCurrency`) are empty.

Networking/Networking/Model/WCPayAccountStatusEnum.swift
```diff
@@ -0,0 +1,100 @@
+import Foundation
+
+/// Represents all of the possible Site Plugin Statuses in enum form
```
shiki, line 3: Excellent documentation!

Networking/Networking/Model/WCPayAccountStatusEnum.swift
```diff
@@ -0,0 +1,100 @@
...
+///
+public enum WCPayAccountStatusEnum: Decodable, Hashable, GeneratedFakeable {
```
shiki, line 5: I think it's fine to name this as `WCPayAccountStatus` if you like. 🙂

Networking/Networking/Model/WCPayAccountStatusEnum.swift
```diff
@@ -0,0 +1,100 @@
...
+    /// card present payments.
+    case complete
+
+    /// restricted: This state occurs when there is required business information missing from the account.
+    /// If hasOverdueRequirements is also true, then the deadline for providing that information HAS PASSED.
+    /// The merchant will probably NOT be able to collect card present payments.
+    /// Otherwise, if hasPendingRequirements is true, then the deadline for providing that information has not yet passed.
+    /// The deadline is available in currentDeadline. The merchant will probably be able to collect card present payments
+    /// until the deadline.
+    /// Otherwise, if neither hasOverdueRequirements nor hasPendingRequirements is true, then the account is under
```
shiki: If you like, you can also use backticks to identify the references to code. 

```suggestion
    /// Otherwise, if neither `hasOverdueRequirements` nor `hasPendingRequirements` is true, then the account is under
```

Helps with reading I think. 

image

Networking/Networking/Model/WCPayAccountStatusEnum.swift
```diff
@@ -0,0 +1,100 @@
...
+/// Enum containing all possible account status keys
+///
+private enum Keys {
+    static let complete               = "complete"
+    static let restricted             = "restricted"
+    static let rejectedFraud          = "rejected.fraud"
+    static let rejectedTermsOfService = "rejected.terms_of_service"
+    static let rejectedListed         = "rejected.listed"
+    static let rejectedOther          = "rejected.other"
+    static let noAccount              = "NOACCOUNT" /// The status field will be not present in the API response when no account has been set up.
```
shiki: Maybe you would like to consider moving the comment above so that it's visible in Xcode's helper. 🙂 

```suggestion
    /// The status field will be not present in the API response when no account has been set up.
    static let noAccount              = "NOACCOUNT" 
```

Networking/Networking/Remote/WCPayRemote.swift
```diff
@@ -19,4 +18,37 @@ public class WCPayRemote: Remote {
...
+
+    enum AccountParameterKeys {
+        static let fields: String = "_fields"
+    }
+
+    enum AccountParameterValues {
+        static let fieldValues: String = """
+            status,has_pending_requirements,has_overdue_requirements,current_deadline,
+            statement_descriptor,store_currencies,country
+            """
```
shiki: 🛑  We would most probably want to remove the line break here. This is what gets encoded in the `parameters` dictionary in `loadAccount()`:

```
(lldb) po parameters
▿ 1 element
  ▿ 0 : 2 elements
    - key : "_fields"
    - value : "status,has_pending_requirements,has_overdue_requirements,current_deadline,\nstatement_descriptor,store_currencies,country"
```

Notice the `\n`. 😬 

We can remove the line break by adding a backslash:

```suggestion
        static let fieldValues = """
            status,has_pending_requirements,has_overdue_requirements,current_deadline,\
            statement_descriptor,store_currencies,country
            """
```

Networking/NetworkingTests/Remote/WCPayRemoteTests.swift
```diff
@@ -53,4 +53,269 @@ final class WCPayRemoteTests: XCTestCase {
...
+        switch result {
+        case .success(let account):
+            XCTAssertEqual(account.status, .complete)
+            XCTAssertEqual(account.statementDescriptor, "MY.FANCY.US.STORE")
+            XCTAssertEqual(account.defaultCurrency, "usd")
+            XCTAssertEqual(account.supportedCurrencies, ["usd"])
+            XCTAssertEqual(account.country, "US")
+        case .failure(let error):
+            XCTAssertNil(error)
+        }
```
shiki, line 82: One way we can make this more intentional is:

1. Answer “What is the expected status of `result`”? Successful or failure? 
2. Get the actual success value and test that. 

In this way, we don't need to check for `.failure`, which clarifies that we only expected the result to be successful. Here's what I think that can look like: 

```suggestion
        XCTAssertTrue(result.isSuccess)

        let account = try result.get()
        XCTAssertEqual(account.status, .complete)
        XCTAssertEqual(account.statementDescriptor, "MY.FANCY.US.STORE")
        XCTAssertEqual(account.defaultCurrency, "usd")
        XCTAssertEqual(account.supportedCurrencies, ["usd"])
        XCTAssertEqual(account.country, "US")
```

P.S. We'd need to make the test function `throws`. 😄

Networking/NetworkingTests/Remote/WCPayRemoteTests.swift
```diff
@@ -53,4 +53,269 @@ final class WCPayRemoteTests: XCTestCase {
...
+                promise(result)
+            }
+        }
+
+        // Then
+        switch result {
+        case .success(let account):
+            XCTAssertNil(account)
+        case .failure(let error):
+            XCTAssertNotNil(error)
```
shiki, line 318: Just noting that the `error` is not exactly what we expect. It's a decoding error instead of the error represented by `wcpay-account-error.json`. 

```
(lldb) po error
▿ DecodingError
  ▿ typeMismatch : 2 elements
    - .0 : Swift.String
    ▿ .1 : Context
      ▿ codingPath : 2 elements
        - 0 : CodingKeys(stringValue: "data", intValue: nil)
        - 1 : CodingKeys(stringValue: "status", intValue: nil)
      - debugDescription : "Expected to decode String but found a number instead."
      - underlyingError : nil
```

I think we're supposed to be using `network.simulateError()` for the mock. But we'd need to make a change in there to simulate the JSON response too. That's best done outside of this PR. 🙂

shiki (approved):
Hey, @allendav! This looks good to me. I left a **non-blocking** question (❓) and a change request that I think we should do (🛑 ). Everything else is just suggestions or random comments that you can ignore. 😅 

Feel free to merge when you're happy with whatever changes you make. 🙂 

---

By the way, thank you very much for writing a detailed PR description. 🙏 I tested this manually and it works nicely!

Networking/Networking/Mapper/WCPayAccountMapper.swift
allendav: > Does this mean that the API returns an empty JSON array instead of null?

Yes exactly. It's an odd one :P

Networking/Networking/Remote/WCPayRemote.swift
allendav: Eeek! Thank you for catching that! Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3881/commits/eda4b4fd272f3b0a4f03023af1d8f934e5e78eca

Networking/Networking/Model/WCPayAccount.swift
allendav: Perfect! Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3881/commits/11a4c1de8f44217e1d8ee138a08989e7d1e99140

Networking/Networking/Model/WCPayAccountStatusEnum.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3881/commits/a544444a49d41e5f8ba83d906ac36435bd614a93

Networking/Networking/Model/WCPayAccountStatusEnum.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3881/commits/b14c02d7dea186086d7c96b87bccd3f76ea5eefd

Networking/Networking/Model/WCPayAccount.swift
allendav: That's so much safer. Thank you. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3881/commits/57944812068829c295103ea9cac128e33f69878f

Networking/Networking/Model/WCPayAccount.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/commit/a544444a49d41e5f8ba83d906ac36435bd614a93

Networking/NetworkingTests/Remote/WCPayRemoteTests.swift
allendav: Will fix in https://github.com/woocommerce/woocommerce-ios/issues/3900

Networking/NetworkingTests/Remote/WCPayRemoteTests.swift
allendav: Good idea - opened a new issue in https://github.com/woocommerce/woocommerce-ios/issues/3900

PR #3783, opened 2021-03-03
Variations: Accessibility details

Ecarrion:
closes #3704

# Why

Since the variations project is almost finished, it's time to review a few accessibility details before releasing it to the public.

# How

This PR makes the following changes

- Update variations list screen to properly support big fonts - the details text in each cell didn't spawn over two lines - 
- Update variations list screen to cover all edges in landscape - the table view was pinned to the safe area-
- Update add attribute screen to cover all edges in landscape - the table view was pinned to the safe area-
- Update add attribute options screen to cover all edges in landscape - the table view was pinned to the safe area-
- Updates loading indicator view color when updating attributes - Now that we have large titles(behind a feature flag), the navigation bar color changed - 

# Screenshots

Variations big font before | Variations big font after
--- | ---
variations-list-before | variations-list-after

Variations landscape before | Variations landscape after
--- | ---
variation-list-landscape-before | variation-list-landscape-after

Add Attribute landscape before | Add Attribute landscape after
--- | ---
add-attribute-before | add-attribute-after

Add Options landscape before | Add Options landscape after
--- | ---
add-option-before | add-options-after

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3896, opened 2021-03-30
Large Titles: Update navigation bar loading color

### LINKED ISSUE #3803: Variations: Update nav loading indicator color
We should update the navigation bar loading indicator color that is shown when updating attributes, to some kind of gray.

ref: https://github.com/woocommerce/woocommerce-ios/pull/3783#issuecomment-792733670

Ecarrion:
fix #3803

# Why

[It was identified](https://github.com/woocommerce/woocommerce-ios/pull/3783#issuecomment-792733670) that the loading indicators in the large title navigation bar could appear as actionable due to its purple color.

This PR fixes that bu changing the color to a grey one.

# Demo

loading

# Testing Steps
- Go to a variable product
- Try to add or edit an attribute option.
- See the gray loading indicator in the navigation bar while the remote operation happens.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Styles/UIColor+SemanticColors.swift
+++ b/WooCommerce/Classes/Styles/UIColor+SemanticColors.swift
@@ -266,7 +266,7 @@ extension UIColor {
     ///
     static var navigationBarLoadingIndicator: UIColor {
         if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) {
-            return .gray(.shade60)
+            return .systemGray
         } else {
             return .white
         }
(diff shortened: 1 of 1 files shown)


rachelmcr (approved):
This works as described! 👍 

I'm wondering about the broader use of this indicator color, though. I noticed we only use the `navigationBarLoadingIndicator` color in two places right now (this attributes screen and in the shipping label flow). Most other indicators don't customize the color and are a default gray, or the `FancyAnimatedButton` uses a plain `.gray` color. Do we want to use this shade of gray in particular? And if so, should we look at making it consistent everywhere?

Ecarrion:
Nice catch @rachelmcr, I've updated it to use `.systemGray` so it looks unmodified from the rest of activity indicators!

Issue #2974, opened 2020-10-14
Shipping Labels M2 release: add option to add package details for a shipping label 


anitaa1990:
Once the [shipping label addresses are displayed and validated](https://github.com/woocommerce/woocommerce-ios/issues/2973), we would need to provide the option to add package details for the shipping label.

- [x] Clicking on the **Continue** button on the Package details section, displays a list of products associated with the order under a single package. 
- [x] Manage the logic for showing the list of products associated with the order under a single package.

<img width="582" alt="Screenshot 2020-10-14 at 10 46 42 AM" src="https://user-images.githubusercontent.com/22608780/95946622-8e9c4580-0e0a-11eb-87f8-7030a4e6c057.png">

**Note that the option to move a product to a different package or create a new package will be taking place in Milestone 3**

PR #3899, opened 2021-04-02
Shipping Label Package Details: managed the logic for generating the list of items that will be shipped inside an order

pmusolino:
Part of #2974 

## Description
As part of Shipping Labels M2, in this PR I managed the logic for showing the list of items associated with the order.
Since the order items do not contain all the info we need, like the weight of every product or if one of the items is a virtual product, we should fetch them from a network request or, if already stored, from Core Data. I implemented the `ShippingLabelPackageDetailsResultsControllers` that contains the two results controllers for fetching/monitoring products and product variations that will be used for matching the info of every item and for generating the items rows.
The logic will be used in the future also for generating the items that will be sent through the network requests.

## Changes
- Implemented an extension of `Decimal` that returns the int value of a decimal value. + unit tests
- `ShippingLabelPackageDetailsViewModel`: since we need to add the results controller for fetching the stored products and product variations, this model is now a class and no more a struct. This means also that the class conforms to `ObservableObject` and the property `itemsRows` is now a `@Published` property which will be observed by `ShippingLabelPackageDetails`. Updated also the logic for generating the items rows, creating an element in the array for every item, and removing virtual products.
- `ShippingLabelPackageDetails`: updated based on changes in the view model.
- `ShippingLabelPackageDetailsResultsControllers`: implemented this new class that contains two results controllers for fetching/monitoring products and product variations that will be used in the Package Details screen.
- `ShippingLabelPackageNumberRow`: fixed the leading/trailing padding.
- Implemented unit tests for `ShippingLabelPackageDetailsViewModel`.

## Testing
2. Open an order detail.
8. The items rows should display an element for every item. If there is an item with quantity 3, you should see 3 different items. If virtual products are present, you should not see these products. You should also see the weight of every product, if available.

## Screenshots

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/Decimal+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/Decimal+Helpers.swift
@@ -1,0 +1,22 @@
+import Foundation
+
+extension Decimal {
+
+    /// Returns the int value of a decimal.
+    ///
+    var intValue: Int {
+        NSDecimalNumber(decimal: whole).intValue
+    }
+
+    private func rounded(_ roundingMode: NSDecimalNumber.RoundingMode = .down, scale: Int = 0) -> Self {
+        var result = Self()
+        var number = self
+        NSDecimalRound(&result, &number, scale, roundingMode)
+        return result
+    }
+
+    private var whole: Self { rounded( self < 0 ? .up : .down) }
+
+    private var fraction: Self { self - whole }
+
+}
--- a/WooCommerce/Classes/Extensions/Decimal+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/Decimal+Helpers.swift
@@ -2,7 +2,7 @@ import Foundation
 
 extension Decimal {
 
-    /// Returns the int value of a decimal.
+    /// Returns the int value of a decimal. We ensure we round down our Decimal before converting it to an Int, using NSDecimalRound.
     ///
     var intValue: Int {
         NSDecimalNumber(decimal: whole).intValue
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -43,12 +43,12 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.stores = stores
         self.storageManager = storageManager
         weightUnit = shippingSettingsService.weightUnit
-        instantiateResultsControllers()
+        configureResultsControllers()
         syncProducts()
         syncProductVariations()
     }
 
-    private func instantiateResultsControllers() {
+    private func configureResultsControllers() {
         resultsControllers = ShippingLabelPackageDetailsResultsControllers(siteID: order.siteID,
                                                                            orderItems: order.items,
            onProductReload: { [weak self] (products) in
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -84,6 +84,10 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
                 product = products.first { $0.productID == item.productID }
             }
             if product?.virtual == false || productVariation?.virtual == false {
+                
+                /// We do not consider fractional quantities because the backend will return always int values for the quantity.
+                /// We are also showing items only when the quantity is > 1, because in that case we are not considering it a valid value.
+                ///
                 for _ in 0..<item.quantity.intValue {
                     let attributes = item.attributes.map { VariationAttributeViewModel(orderItemAttribute: $0) }
                     let weightUnitA: String = weightUnit ?? ""
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -67,7 +67,7 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         itemsRows = generateItemsRows()
     }
 
-    /// Generate the items rows, creating an element in the array for every item (eg. if there are an item with quantity 3,
+    /// Generate the items rows, creating an element in the array for every item (eg. if there is an item with quantity 3,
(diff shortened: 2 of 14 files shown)


WooCommerce/Classes/Extensions/Decimal+Helpers.swift
```diff
@@ -0,0 +1,22 @@
+import Foundation
+
+extension Decimal {
+
+    /// Returns the int value of a decimal.
```
allendav: Please describe how the fraction is treated in the comment too (up/down rounding)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -1,50 +1,148 @@
...
+                productVariation = productVariations.first { $0.productVariationID == item.variationID }
+            }
+            else {
+                product = products.first { $0.productID == item.productID }
+            }
+            if product?.virtual == false || productVariation?.virtual == false {
+                for _ in 0..<item.quantity.intValue {
+                    let attributes = item.attributes.map { VariationAttributeViewModel(orderItemAttribute: $0) }
+                    let weightUnitA: String = weightUnit ?? ""
+                    let subtitle = Localization.subtitle(weight: isVariation ? productVariation?.weight : product?.weight, attributes: attributes) + weightUnitA
```
allendav: The weight unit should probably be part of the localization formatter too, not appended afterwards, especially for RTL languages.

I'm not sure they change from language to language, but core localizes them too: https://github.com/woocommerce/woocommerce/blob/b446417be69602ffce3d9023bd2b3eef39318451/includes/admin/settings/class-wc-settings-products.php#L363-L366

WooCommerce/WooCommerceTests/Extensions/DecimalWooTests.swift
```diff
@@ -0,0 +1,18 @@
...
+///
+final class DecimalWooTests: XCTestCase {
+
+    func test_intValue_returns_the_expected_result() {
+        XCTAssertEqual(Decimal(string: "1234.0")?.intValue, 1234)
+        XCTAssertEqual(Decimal(string: "100.123456789123")?.intValue, 100)
+        XCTAssertEqual(Decimal(string: "200")?.intValue, 200)
+        XCTAssertEqual(Decimal(string: "2")?.intValue, 2)
+        XCTAssertEqual(Decimal(string: "0")?.intValue, 0)
+        XCTAssertEqual(Decimal(string: "0.2323921301")?.intValue, 0)
```
allendav, line 18: Since your extension has logic for negative numbers, you may want to add tests for that too

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -1,50 +1,148 @@
...
+            var productVariation: ProductVariation?
+
+            if isVariation {
+                productVariation = productVariations.first { $0.productVariationID == item.variationID }
+            }
+            else {
+                product = products.first { $0.productID == item.productID }
+            }
+            if product?.virtual == false || productVariation?.virtual == false {
+                for _ in 0..<item.quantity.intValue {
```
allendav, line 91: If my order contains a fractional quantity of an item (e.g. 3.5) and we get a row for each whole item (3), should we not also get a row for the fractional remainder (0.5)?

Also, maybe explain in a comment why we are row-izing items with quantity != 1?

allendav:
A thought: I see weight units as Strings throughout this PR. Could weight units be an enumeration instead? That could make localization easier, and when the weight of a package is passed to the labels endpoint, you may need to convert between units to get to a unit supported by the service. An enumeration will make that safer.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
allendav: That said, I also think there should be a space between the weight and the unit, e.g. `1 lb`

Maybe we'd benefit from a formatter for weight like we have for [prices](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift).

allendav:
I couldn't find this in this PR, but when there is only 1 item (row) the singular form "item" should be used:

Screen Shot 2021-04-02 at 2 27 39 PM

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
pmusolino: Good question. I double-checked this on Core, and it seems that the quantity is always an int value. Also if you create an Order from the core (so on the admin side) and you add a fractional quantity, it will be converted to an int value. So the real question is why `quantity` is considered a Decimal value. I opened an issue about it https://github.com/woocommerce/woocommerce-ios/issues/3908

> Also, maybe explain in a comment why we are row-izing items with quantity != 1?

Done ✅

pmusolino: Discussion about this, here https://a8c.slack.com/archives/C6H8C3G23/p1617703217370100

pmusolino: So, after some investigations, it seems that I handled this particular case in the right way. The web client considers a quantity like `3.5` as `3 items`. 

Screen Shot 2021-04-06 at 12 12 47

pmusolino:
Thanks for the review @allendav!

I made all the changes and I also added a weight formatter as suggested by you, which for the moment is a super easy implementation since we don't have so many details about the weight. It's just a number and a unit, but it can become more complex in the near future. 4926303116490c45a83f95c4b386a9a34cf8fdae

> A thought: I see weight units as Strings throughout this PR. Could weight units be an enumeration instead? That could make localization easier, and when the weight of a package is passed to the labels endpoint, you may need to convert between units to get to a unit supported by the service. An enumeration will make that safer.

I believe that we will not have to send the unit of the weight since what we use in the app is taken from the settings of the shipping labels. This is a thing that it's useful just on the visual side, and it's not used for any calc on our side. Since the weight units can be dynamic (maybe on the backend side they will support other units in the future), do you think that it makes sense to transform it into an enum?

allendav:
> I believe that we will not have to send the unit of the weight since what we use in the app is taken from the settings of the shipping labels.

That makes sense. I was thinking like we have to for payments where we have to send currency directly to Stripe. But you're right - all labels stuff goes through our backend.

allendav (approved):
Thanks for the changes! I cannot get Xcode to finish building anything today for some reason, but let's not have that hold this up. Once the tests pass, :shipit:

pmusolino:
Thank you so much @allendav! 💯

PR #3907, opened 2021-04-05
Storage: Use a shared child context for writing operations

### LINKED ISSUE #3819: Stores & Shared Derived Storage
We have discovered some inconsistencies when stores need to add content saved or modified in another store.

One possible solution is to use a shared derived storage for all of our store's writes. 

initial discussion: C6H8C3G23-slack-p1613053833352500
possible approach: https://gist.github.com/shiki/53a00c4a2cb91827570b32ad2af559db
CoreData RFC: paaHJt-1I7-p2

Ecarrion:
fix #3819 

# Why

There have been some lengthy discussions about the potential concurrency problems that our current writing mechanism in our storage layer has.

To sum it up, since each **store** creates a unique writing child context, writing concurrently on those can be problematic as they don't know each other until sync time.

I recently hit this problem when two different write child contexts had merge problems because both updated the same object.

The solution I'm implementing here is to have a single shared writer context, so operations can be done sequentially on it, in a background thread. 

FYI, this is inline with the Core Data modernization RFC, that platform9 presented.

# How

- Replace the method `newDerivedStorage()` for a `lazy var writerDerivedStorage`. The reason I'm choosing to not provide a `newDerivedStorage()` anymore is because, after this change, there is no usage for it and we want to discourage creating a new child context.
- Remove workaround for variations project in `ProductAttributeStore`
- Delete tests that proved our concurrency problem.

# Testing Steps
- Just smoke test the app, make changes to products, orders, etc and make sure everything is working as expected.

Update release notes:


pmusolino (approved):
Looks good to me. I did a smoke test on Orders, Products, and Reviews and I didn't notice any issue.
I also ran all the unit tests locally, and they are ✅.

:shipit:

Ecarrion:
Thanks for the review Paolo!

Issue #3843, opened 2021-03-17
[Mobile Payments] Propagate request display message


ctarda:
During the process to collect payment methods and process a payment intent, the reader might call into a delegate method called `terminal:didRequestReaderDisplayMessage:`.

This method will receive an enumeration that represents the display messages that a reader may request be displayed by your app.

We need to map this enumeration to user-facing strings and propagate those strings to the UI

PR #3925, opened 2021-04-07
[Mobile Payments] Prepare for providing feedback of payment collection process on the UI

### LINKED ISSUE #3842: [Mobile Payments] Propagate reader input options
When the process to collect a payment method begins, the Stripe Terminal SDK will call into a delegate method called `terminal:didRequestReaderInput`, passing an enumeration.

This enumeration represents all of the input methods available to your user when the reader begins waiting for input.

We would need to:
* match the values in the enumeration to user-facing strings
* propagate those strings to the UI

ctarda:
Closes #3842 , #3843 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by implementing enough code to propagate certain events dispatched by the card reader to the UI. In particular:

* When the card reader requests a card to be inserted / tapped / swiped.
* When the card reader requests the card to be removed.
* When the card reader requests any string to be presented on screen.

These events are generated by the Stripe Terminal SDK. For some of those events, the Stripe Terminal SDK also provides a user facing string (see [this](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(cm)stringFromReaderInputOptions:) and [this](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(cm)stringFromReaderDisplayMessage:)). We are passing those strings to the UI as well, although, in my opinion, and from what I've seen in my tests, the 
value they provide is debatable. 

## Changes
* Add a couple of cases to CardReaderEventType
* Add a new model object CardReaderEvent that wraps a CardReaderEventType and a user facing string
* Re-implement the methods declared in the conformance of StripeCardReaderService to ReaderDisplayDelegate so that they send new events via the reader events subject
* Update CardPresentPaymentStore to observe those events and update the API of the collectPayment method to propagate the events to the UI layer. See #3926 for some known issues in the implementation of th CardPresentPaymentStore.
* Add just enough code to OrderDetailsViewModel and OrderDetailsViewController to present some feedback on the UI, so that testing the PR is a bit easier. This code will definitely be removed later on, and the APIs are not final. This is part of the PR only so that there is visible feedback on the UI of the changes.

## How to test

2. Build and run on a device. 
8. Tap "Connect card reader"
10. Tap "Connect to reader"
12. Navigate to an order (any order with payment method Cash On Delivery).
13. Scroll down to the "Collect Payment" button
14. Tap the button. Follow the instructions on screen

This is a video of the "before" (notice there is only one alert at the end of the process):
https://user-images.githubusercontent.com/2722505/111704327-54016380-8815-11eb-9feb-05c78f3a9eae.mov

And this is how it should look now (notice there is an alert prompting to tap the card, and that said alert is dismissed after tapping the card):

https://user-images.githubusercontent.com/2722505/113864644-24d97480-9779-11eb-93a8-df4e33c25c93.mov

Update release notes:


Hardware/Hardware/CardReader/CardReaderEvent.swift
```diff
@@ -1,5 +1,26 @@
+/// An event dispatched by  a connected reader.
```
Ecarrion:
```suggestion
/// An event dispatched by a connected reader.
```

Hardware/Hardware/CardReader/CardReaderEvent.swift
```diff
@@ -1,5 +1,26 @@
...
+/// It wraps an event type and an optional user-facing message
+public struct CardReaderEvent {
+    public let type: CardReaderEventType
+    public let message: String?
+
+    init(type: CardReaderEventType, message: String? = nil) {
```
Ecarrion: Should this also be public?

Hardware/Hardware/CardReader/CardReaderEvent.swift
```diff
@@ -1,5 +1,26 @@
...
-public enum CardReaderEvent {
+public enum CardReaderEventType {
+    /// The reader begins waiting for input.
+    /// The app should prompt the customer to present a payment method
+    case waitingForInput
+
+    /// Request that a prompt be displayed in the app.
+    /// For example, if the prompt is SwipeCard,
+    /// the app should instruct the user to present the card again by swiping it.
+    case displayMessage
```
Ecarrion: Do you think it expresses the intention better to have this enum with the message as an associated value, instead of having an optional message in the `CardReaderEvent` type?

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -156,6 +156,8 @@ extension StripeCardReaderService: CardReaderService {
             }
 
             Terminal.shared.collectPaymentMethod(activeIntent, delegate: self) { (intent, error) in
+                self?.sendReaderEvent(CardReaderEvent(type: .cardRemoved))
```
Ecarrion: Q: Is it ok to send the event before checking for the error or intent?

Ecarrion (commented):
Looks good to me, but left some general questions/suggestions.
I won't approve it because I have no ways to test it yet :-(

cc @allendav

Hardware/Hardware/CardReader/CardReaderEvent.swift
ctarda: Not really. This can only be initialized within the integration with the Stripe SDK, so I think it's better if it stays internal. Am I missing something?

Hardware/Hardware/CardReader/CardReaderEvent.swift
ctarda: I am not sure if it conveys the intention better, but it sure makes things easier for clients, and I don't see any downside to it... I'll update thanks!

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
ctarda: In this case (and only this case), it does. This is the event that notifies the UI that the card has been tapped / swiped. Internally, that means that the operation to collect a payment method has already a response from the SDK (which then can be either an error or success)

ctarda:
Thanks for the review @Ecarrion 🙇  I just pushed an update to incorporate you feedback. @allendav , you are next!

Hardware/Hardware/CardReader/CardReaderEvent.swift
```diff
@@ -1,5 +1,14 @@
...
+    /// The app should prompt the customer to present a payment method
+    case waitingForInput(String)
+
+    /// Request that a prompt be displayed in the app.
+    /// For example, if the prompt is SwipeCard,
+    /// the app should instruct the user to present the card again by swiping it.
+    case displayMessage(String)
+
     // A card was inserted.
     case cardInserted
```
allendav, line 13: Will `cardInserted` also fire for tapped or swiped cards? If so, please update comment.
Either way, please make it `///` (not `//`) so xcode displays it like the others you are adding

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -156,6 +156,8 @@ extension StripeCardReaderService: CardReaderService {
             }
 
             Terminal.shared.collectPaymentMethod(activeIntent, delegate: self) { (intent, error) in
+                self?.sendReaderEvent(.cardRemoved)
```
allendav, line 166: Please add a comment explaining why we are sending this event and if it applies to swiped or tapped cards (or only inserted ones)

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -460,7 +460,11 @@ extension OrderDetailsViewModel {
         stores.dispatch(deleteTrackingAction)
     }
 
-    func collectPayment(onCompletion: @escaping (Result<Void, Error>) -> Void) {
+    /// This does not have to be the final API. For now, it is good enough to show
```
allendav: Please consider adding a TODO so we remember to look at this later?

allendav (changes requested):
A couple comments on the code, plus...

During testing I observed the initial alert prompting me to Swipe / Insert / Tap. I tested all three flows:

For the tap flow, tapping the card led to a "success - the payment has been processed" alert. Holding the card to the reader longer than necessary didn't change that flow. I think that's ok.

For the swipe flow,  swiping the card led to a "success - the payment has been processed" alert.

For the swipe flow, I tried to swipe it "incorrectly" but couldn't get anything other than success. Not sure there is a problem there. I suspect it was processing it as tapped. It might be helpful for debugging to log which flow happened?

For the insert flow, I noted the "Remove Card" alert, but I was surprised when that alert didn't close automatically on removing the card. I also didn't see the success alert ever fire. This seems bad. Please double check?

allendav (approved):
:shipit:

ctarda:
Thanks for the feedback @allendav I'll merge as soon as I get the ✅  from CI

Issue #3864, opened 2021-03-24
[Mobile Payments] Pass testing flag to StripeCardReaderService


ctarda:
It turns out that if we initialize the Stripe Terminal SDK with a configuration like this:

```
        let config = DiscoveryConfiguration(
            discoveryMethod: .bluettoothProximity,
            simulated: true
        )
```

we should be able to run unit tests with the Terminal in test mode, and returning the readers that we need.

PR #3931, opened 2021-04-08
[Mobile Payments] Bring PaymentIntent back and hide implementation details of payment collection

### LINKED ISSUE #3927: [Mobile Payments] Bring PaymentIntent back
We removed the PaymentIntent model object as part of a previous PR, but we are going to need it back in order to support rendering receipts on device.

### LINKED ISSUE #3930: [Mobile Payments] Hide implementation details of payment collection
We are leaking the internal implementation of the Stripe Terminal SDK by exposing the three different methods that take part of the payment collection process. 

We only want one public method that can be provided with a PaymentIntentsParameters and returns a Future that will be fulfilled with a success or error after the full pipeline has been completed.

ctarda:
Closes #3927, #3930

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by rolling back some changes we did [in previous PRs](https://github.com/woocommerce/woocommerce-ios/pull/3847/commits/47cd4acd4be6f6ca862c552630ac99746ba7067f) and doing a bit of cleanup in the service API.

This PR does not introduce any user facing changes.

## Changes
* Revert the removal of PaymentIntent, and associated objects and tests. A few weeks back we removed this object, but after looking into what is necessary to render receipts, it is clear that we need it back. (p91TBi-4Ob-p2 for reference)
* Avoid leaking some of the implementation details of the Stripe Terminal SDK, in particular, the fact that processing a payment requires creating an intent, then collecting a payment method for that intent, and finally capturing the payment. Now, all that is done by the service via a new public method that takes the intent parameters as a parameter and returns a publisher that will complete with the final PaymentIntent or an error. In a nutshell, we moved most of the implementation of CardPresentPaymentStore.collectPayment to a method in StripeCardReaderService. 
* Removed a few tests that are just not possible now. #3864 , which goes next, will open up testing again.

## How to test

2. Build and run on a device. 
8. Tap "Connect card reader"
10. Tap "Connect to reader"
12. Navigate to an order (any order with payment method Cash On Delivery).
13. Scroll down to the "Collect Payment" button
14. Tap the button. Follow the instructions on screen

The result should look like this:

https://user-images.githubusercontent.com/2722505/114028174-2de44780-9846-11eb-9155-bab847453f3e.mov

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/PaymentIntent.swift
+++ b/Hardware/Hardware/CardReader/PaymentIntent.swift
@@ -9,7 +9,9 @@ public struct PaymentIntent: Identifiable {
 
     /// When the PaymentIntent was created
     public let created: Date
+
     ///The amount to be collected by this PaymentIntent, provided in the currency’s smallest unit.
+    /// e.g. USD$5.00 should have amount = 500 and currency = 'usd'
     /// - see: https://stripe.com/docs/currencies#zero-decimal
     public let amount: UInt
 
--- a/Hardware/Hardware/CardReader/PaymentIntent.swift
+++ b/Hardware/Hardware/CardReader/PaymentIntent.swift
@@ -1,0 +1,24 @@
+/// A PaymentIntent tracks the process of collecting a payment from your customer.
+/// We would create exactly one PaymentIntent for each order
+public struct PaymentIntent: Identifiable {
+    /// Unique identifier for the PaymentIntent
+    public let id: String
+
+    /// The status of the Payment Intent
+    public let status: PaymentIntentStatus
+
+    /// When the PaymentIntent was created
+    public let created: Date
+    ///The amount to be collected by this PaymentIntent, provided in the currency’s smallest unit.
+    /// - see: https://stripe.com/docs/currencies#zero-decimal
+    public let amount: UInt
+
+    /// The currency of the payment.
+    public let currency: String
+
+    /// Set of key-value pairs attached to the object.
+    public let metadata: [AnyHashable: Any]?
+
+    // Charges that were created by this PaymentIntent, if any.
+    public let charges: [Charge]
+}
--- a/Hardware/Hardware/CardReader/PaymentIntentStatus.swift
+++ b/Hardware/Hardware/CardReader/PaymentIntentStatus.swift
@@ -1,0 +1,9 @@
+///The possible statuses for a PaymentIntent.
+public enum PaymentIntentStatus {
+    case requiresPaymentMethod
+    case requiresConfirmation
+    case requiresCapture
+    case processing
+    case canceled
+    case succeeded
+}
--- a/Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
@@ -1,0 +1,35 @@
+import StripeTerminal
+
+extension PaymentIntent {
+
+    /// Convenience initializer
+    /// - Parameter intent: An instance of a StripeTerminal.PaymentIntent
+    init(intent: StripePaymentIntent) {
+        self.id = intent.stripeId
+        self.status = PaymentIntentStatus.with(status: intent.status)
+        self.created = intent.created
+        self.amount = intent.amount
+        self.currency = intent.currency
+        self.metadata = intent.metadata
+        self.charges = intent.charges.map { .init(charge: $0) }
+    }
+}
+
+
+/// The initializers of StripeTerminal.PaymentIntent are annotated as NS_UNAVAILABLE
+/// So we can not create instances of that class in our tests.
+/// A workaround is declaring this protocol, which matches the parts of
+/// SCPPaymentIntent that we are interested in, make Reader implement it,
+/// and initialize Harware.CardReader with a type conforming to it.
+protocol StripePaymentIntent {
+    var stripeId: String { get }
+    var created: Date { get }
+    var status: StripeTerminal.PaymentIntentStatus { get }
+    var amount: UInt { get }
+    var currency: String { get }
+    var metadata: [AnyHashable: Any]? { get }
+    var charges: [SCPCharge] { get }
+}
+
+
+extension StripeTerminal.PaymentIntent: StripePaymentIntent { }
--- a/Hardware/Hardware/CardReader/StripeCardReader/PaymentIntentStatus+Stripe.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/PaymentIntentStatus+Stripe.swift
@@ -1,0 +1,25 @@
+import StripeTerminal
+
+extension PaymentIntentStatus {
+
+    /// Factory Method to initialize PaymentItemStatus with StripeTerminal's PaymentIntentStatus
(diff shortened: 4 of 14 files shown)


Hardware/Hardware/CardReader/PaymentIntent.swift
```diff
@@ -0,0 +1,24 @@
...
+public struct PaymentIntent: Identifiable {
+    /// Unique identifier for the PaymentIntent
+    public let id: String
+
+    /// The status of the Payment Intent
+    public let status: PaymentIntentStatus
+
+    /// When the PaymentIntent was created
+    public let created: Date
+    ///The amount to be collected by this PaymentIntent, provided in the currency’s smallest unit.
```
allendav, line 13: For great success, please add something like

e.g. USD$5.00 should have amount = 500 and currency = 'usd'

Hardware/Hardware/CardReader/PaymentIntentStatus.swift
```diff
@@ -0,0 +1,9 @@
+///The possible statuses for a PaymentIntent.
+public enum PaymentIntentStatus {
+    case requiresPaymentMethod
+    case requiresConfirmation
+    case requiresCapture
+    case processing
+    case canceled
```
allendav, line 7: TIL https://www.grammarly.com/blog/canceled-vs-cancelled/#:~:text=Canceled%20or%20cancelled%20is%20the,British%20English%20and%20other%20dialects.

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntentStatus+Stripe.swift
```diff
@@ -0,0 +1,25 @@
+import StripeTerminal
+
+extension PaymentIntentStatus {
+
+    /// Factory Method to initialize PaymentItemStatus with StripeTerminal's PaymentIntentStatus
+    /// - Parameter readerType: an instance of PaymentIntentStatus, declared in StripeTerminal
```
allendav: readerType -> status

Hardware/HardwareTests/MockStripePaymentIntent.swift
```diff
@@ -0,0 +1,28 @@
...
+
+extension MockStripePaymentIntent: StripePaymentIntent {}
+
+extension MockStripePaymentIntent {
+    static func mock() -> Self {
+        MockStripePaymentIntent(stripeId: "id",
+                                created: Date(),
+                                status: .succeeded,
+                                amount: 100,
+                                currency: "USD",
```
allendav, line 24: Should this be `usd` ?

docs/HARDWARE.md
```diff
@@ -77,6 +77,8 @@ Collecting a payment is a three step process that needs to be performed in this
 
 For more details, see [Stripe's documentation](https://stripe.com/docs/terminal/payments)
 
+This process is abstracted away by the CardrReaderService, meaning that clients of the service do not need to know that the process requires three steps.
```
allendav: CardReaderService

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
```diff
@@ -0,0 +1,35 @@
...
+/// and initialize Harware.CardReader with a type conforming to it.
+protocol StripePaymentIntent {
+    var stripeId: String { get }
+    var created: Date { get }
+    var status: StripeTerminal.PaymentIntentStatus { get }
+    var amount: UInt { get }
+    var currency: String { get }
+    var metadata: [AnyHashable: Any]? { get }
+    var charges: [SCPCharge] { get }
+}
```
allendav, line 32: FWIW, eventually we are going to need to create and push a customer too. It helps with fraud detection and is an obligation with have with Stripe. Perhaps open an issue

https://stripe.com/docs/api/payment_intents/create#create_payment_intent-customer

allendav, line 32: Similarly, maybe open an issue for receipt email

https://stripe.com/docs/api/payment_intents/create#create_payment_intent-receipt_email

allendav, line 32: And for subscription support, an issue for 

https://stripe.com/docs/api/payment_intents/create#create_payment_intent-setup_future_usage

allendav, line 32: We're also expected to provide the customer's address and, if it differs from the address on the credit card, a shipping address too

Please open an issue for

https://stripe.com/docs/api/payment_intents/create#create_payment_intent-shipping

Hardware/HardwareTests/MockStripePaymentIntent.swift
ctarda: Not really. It does not really matter, because It's a mock, but if it is uppercase it also helps test that the property wrapper on the currency actually returns that as lowercase.

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
ctarda: Yes. There are also other fields we will need to add later, namely those that will be necessary for generating receipts

ctarda: Let's get into receipts when we get to that point. All the fields necessary to generate a receipt will need to be added here anyway, and that's when we'l have full clarity of what needs to be done

ctarda: I think we should defer all support for subscriptions for now, until we reach the point where we actually need to support them.

ctarda: Same with other fields, let's defer all things receipt related until we reach the point were we are in a position to generate it.

ctarda:
Thanks for taking look @allendav . I pushed an update to incorporate your feedback, and logged #3936 for updating PaymentIntent with the fields required for receipts.

Ready for another round!

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
allendav: Sure! Although this isn't receipt related. It's a level 2/3 data requirement.

allendav: Absolutely.

allendav: The intention was to make sure we had a clear, comprehensive backlog of subscriptions work for MVP+x

allendav (approved):
Tests well, but I did notice that it is not yet using the amount from the order (the order I tested with was for $6.10, not $1.00). I assume you're aware of that :)

Screen Shot 2021-04-08 at 1 13 23 PM

ctarda:
Yes, we are initiating the capture with a hard coded value. I think we have a ticket to update that (on mobile now and I can't check) but I'll make sure to double check and log a ticket for it if we don't. Actually, that's make a fun short PR for tomorrow Friday 😜

PR #3935, opened 2021-04-08
Fix image loading for some filenames/urls

### LINKED ISSUE #2789: Product Images are not showed if they contain Cyrillic characters in the link
## Describe the bug
Rebecca [reported](https://a8c.slack.com/archives/C6H8C3G23/p1599648229152800) a user who is unable to view images in products. After some investigations, it seems that we are unable to download/render images that contain Cyrillic characters in the image link.

## Screenshots
The first two images contain Cyrillic characters in the link, the last one no, and in fact, it’s showed correctly.

### LINKED ISSUE #3928: Images with special characters in filenames fail to load
## Description

iOS app fails to preview ([1](https://github.com/woocommerce/woocommerce-ios/blob/14ced0b655757025532350c4e3bd281a064b84d8/WooCommerce/Classes/Tools/ImageService/DefaultImageService.swift#L61), [2](https://github.com/woocommerce/woocommerce-ios/blob/14ced0b655757025532350c4e3bd281a064b84d8/WooCommerce/Classes/ViewRelated/Products/Media/DefaultProductUIImageLoader.swift#L57), [3](https://github.com/woocommerce/woocommerce-ios/blob/14ced0b655757025532350c4e3bd281a064b84d8/WooCommerce/Classes/ViewRelated/Products/Media/Media%2BWPMediaAsset.swift#L8), ...) and [map](https://github.com/woocommerce/woocommerce-ios/blob/f7f51638a0c85bcf58ec5d4c1970995045750d47/Networking/Networking/Model/Media/Media.swift#L50) image URLs with non-encoded special characters.

Support request: p4a5px-2Jd-p2

### Possible solution

Add `.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)` on all affected strings on iOS side.

## Test

1. Create an image with non-latin characters in name, like `test-图像.png`
2. Upload it to Media Library
3. Create a new product and assign test image to it
4. Open mobile app, switch to products tab
5. Check product icon in the list
6. Open product details
7. Check product image in the header
8. Tap "Add Image" icon -> "Add Photos" -> "WP Media Library"
9. Check that media library loads all data and displays test image from step 1

## Screenshots

|Example image in web admin media library|
|---|
|Screenshot 2021-04-07 at 20 37 09|

products | product | media lib
---|---|---
products|product|media lib

ealeksandrov:
Closes https://github.com/woocommerce/woocommerce-ios/issues/2789, https://github.com/woocommerce/woocommerce-ios/issues/3928.

## Description

iOS app fails to preview and map product image URLs with non-encoded special characters - like Chinese or Cyrillic.
This PR adds percent encoding on all URL inits related to product images.

## Test

1. Create an image with non-latin characters in name, like `test-图像.png` or `тест.png`
2. Upload it to Media Library
3. Create a new product and assign test image to it
4. Open mobile app, switch to products tab
5. Check product icon in the list
6. Open product details
7. Check product image in the header
8. Tap "Add Image" icon -> "Add Photos" -> "WP Media Library"
9. Check that media library loads all data and displays test image from step 1

## Screenshots

|Example image in web admin media library|
|---|
||

/ | before | after
--|--|--
Products list |Simulator|Simulator
Product details|Simulator|Simulator
Media Library |Simulator|Simulator

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Media/Media.swift
+++ b/Networking/Networking/Model/Media/Media.swift
@@ -47,6 +47,7 @@ extension Media: Decodable {
         let date = try container.decodeIfPresent(Date.self, forKey: .date) ?? Date()
         let fileExtension = try container.decodeIfPresent(String.self, forKey: .fileExtension) ?? ""
         let mimeType = try container.decodeIfPresent(String.self, forKey: .mimeType) ?? ""
+        // Decoding String instead of URL for `src`, since it may have non-Latin chars and URL init can fail without prior percent encoding
         let src = try container.decodeIfPresent(String.self, forKey: .src) ?? ""
         let name = try container.decode(String.self, forKey: .name)
         let alt = try container.decodeIfPresent(String.self, forKey: .alt)
--- a/Networking/Networking/Model/Media/Media.swift
+++ b/Networking/Networking/Model/Media/Media.swift
@@ -47,7 +47,7 @@ extension Media: Decodable {
         let date = try container.decodeIfPresent(Date.self, forKey: .date) ?? Date()
         let fileExtension = try container.decodeIfPresent(String.self, forKey: .fileExtension) ?? ""
         let mimeType = try container.decodeIfPresent(String.self, forKey: .mimeType) ?? ""
-        let src = try container.decodeIfPresent(URL.self, forKey: .src)?.absoluteString ?? ""
+        let src = try container.decodeIfPresent(String.self, forKey: .src) ?? ""
         let name = try container.decode(String.self, forKey: .name)
         let alt = try container.decodeIfPresent(String.self, forKey: .alt)
         let height = try container.decodeIfPresent(Double.self, forKey: .height)
--- a/Networking/NetworkingTests/Remote/MediaRemoteTests.swift
+++ b/Networking/NetworkingTests/Remote/MediaRemoteTests.swift
@@ -34,7 +34,7 @@ final class MediaRemoteTests: XCTestCase {
         remote.loadMediaLibrary(for: sampleSiteID) { mediaItems, error in
             XCTAssertNil(error)
             XCTAssertNotNil(mediaItems)
-            XCTAssertEqual(mediaItems!.count, 5)
+            XCTAssertEqual(mediaItems?.count, 5)
             expectation.fulfill()
         }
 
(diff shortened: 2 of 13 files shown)


Networking/Networking/Model/Media/Media.swift
```diff
@@ -47,7 +47,7 @@ extension Media: Decodable {
         let date = try container.decodeIfPresent(Date.self, forKey: .date) ?? Date()
         let fileExtension = try container.decodeIfPresent(String.self, forKey: .fileExtension) ?? ""
         let mimeType = try container.decodeIfPresent(String.self, forKey: .mimeType) ?? ""
-        let src = try container.decodeIfPresent(URL.self, forKey: .src)?.absoluteString ?? ""
+        let src = try container.decodeIfPresent(String.self, forKey: .src) ?? ""
```
allendav, line 51: Perhaps add a comment here that we cannot decode URL directly since the image src URL may contain non Latin characters that have not been percent encoded.

allendav (approved):
Tests well, including unit tests. One comment requested. I was a little surprised at the use of `.urlQueryAllowed` instead of `.urlPathAllowed` but since we are passing the entire URL including the protocol, to avoid encoding the `:` after the protocol the use of `.urlQueryAllowed` seems to work even when no query `?` is present in the URL.

:shipit: 

IMG_1960
IMG_1959

ealeksandrov:
Thank you for review @allendav!

> I was a little surprised at the use of .urlQueryAllowed instead of .urlPathAllowed

The most amount of work in this PR was deciding about URL parsing and encoding 🙂
Best solution would be to use `URLComponents` which applies different rules on each component, but we can't use it in scope of parsing plain string.

On a few initial test cases `.urlQueryAllowed` charset worked well. I've compared `.urlQueryAllowed` and `.urlPathAllowed` character sets and found that second is a subset of a first.
Essentially, `.urlQueryAllowed == .urlPathAllowed + [";", "?"]`.

I don't think we want to percent-encode `?` and `;`, so `urlQueryAllowed` is preferable?

P.S. I should've added this decision in initial PR desc, sorry for confusion 😅

allendav:
> I don't think we want to percent-encode ? and ;, so urlQueryAllowed is preferable?

Agreed.

> I should've added this decision in initial PR desc, sorry for confusion

All good. It was fun to learn about this.

PR #3939, opened 2021-04-09
[Mobile Payments] Charge actual order total

### LINKED ISSUE #3937: [Mobile Payments] Charge actual order total
So far we have been capturing a payment with a hardcoded amount. 

Update OrderDetailsViewModel to pass the actual order total amount to CardPresentPaymentAction.collectPayment

ctarda:
Closes #3937 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by charging the actual total amount of an order.

So far we have been capturing a payment with a hardcoded amount. This PR adds logic to charge the actual total order.

## Changes
* Update PaymentIntentParams to take an NSDecimal, instead of a UInt to model the charge amount. The reason for that is that a UInt is specific to Stripe. We already have mechanisms in place to format the order total amount as an NSDecimal, so it makes sense to do it here as well.
* Transform the amount to the currency’s smallest unit, [as required by Stripe](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)amount), in the internal method (internal to Hardware) that transform the intent parameters into what Stripe requires.

## Testing

2. Add a breakpoint in line 109 of CardPresentPaymentStore
Screen Shot 2021-04-09 at 06 20 31

3. Build and run on a device. 
4. Switch an external card reader on. The only card reader supported by now is the BBPOS Chipper 2X BT.
5. Log in to an account that matches a store that has been configured for testing.
6. Navigate to Settings in the app. (Tap the gear button in the top right)
7. In the "Store Settings" section, tap "Manage Card Reader".
8. Make sure the card reader is on and the blue light is blinking (we don't do any error handling yet, so it's important that things are setup for the happy path)
9. Tap "Connect card reader"
10. Wait until the app suggests one reader to connect to.
11. Tap "Connect to reader"
12. Make sure the reader light turns solid blue.
13. Navigate to an order (any order with payment method Cash On Delivery).
14. Scroll down to the "Collect Payment" button
15. Tap the button. Follow the instructions on screen

When the breakpoint is triggered, `po intent` and check that the amount matches the expectation.
Screen Shot 2021-04-09 at 06 23 07

For example, for an order with a total of 210.00, the amount in the intent should read 21000

Update release notes:


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -108,16 +108,6 @@ final class OrderDetailsViewModel {
         }
     }
 
-    /// Indicates if payment for the order can be collected using an external card reader
```
ctarda, line 111: This was probably lingering from a previous PR. It is not used anywhere, so off it goes

Yosemite/Yosemite/Model/Model.swift
```diff
@@ -98,6 +98,7 @@ public typealias StoredProductSettings = Networking.StoredProductSettings
 public typealias CardReader = Hardware.CardReader
 public typealias CardReaderEvent = Hardware.CardReaderEvent
 public typealias CardReaderServiceDiscoveryStatus = Hardware.CardReaderServiceDiscoveryStatus
+public typealias CardReaderServiceError = Hardware.CardReaderServiceError
```
ctarda, line 101: This was going to be necessary eventually, so that the UI can switch on the errors emitted by the card reader, so we might as well do it now.

Ecarrion (approved):
Code looks ok to me! My only question would be if it's possible to use `Decimal` instead of `NSDecimalNumber`? It is a bit less powerful but the syntax is nicer to read and understand!

Note: I haven't tested this PR due to the lack of reader hardware!

ctarda:
Thanks for the review @Ecarrion . I pushed a commit to type that property as Decimal, and to be completely honest I am not sure it looks better: we have to cast what `convertToDecimal` in the currency formatter provides as NSDecimal to Decimal, and then build yet another NSDecimalNumber before passing the value to Stripe (which requires it as a Uint). But it works, so 🤷 

Anyway, this PR is blocking further progress, so I'll merge for now. We can always come back to this later if we decide to.

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntentParameters+Stripe.swift
```diff
@@ -9,7 +9,13 @@ extension Hardware.PaymentIntentParameters {
             return nil
         }
 
-        let returnValue = StripeTerminal.PaymentIntentParameters(amount: self.amount, currency: self.currency)
+        /// The amount of the payment needs to be provided in the currency’s smallest unit.
+        ///https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)amount
+        let amountInSmallestUnit = amount * 100
```
allendav, line 14: This will work for most currencies for the time being.  If / when we get to Japanese Yen we will need to change this. For example: https://wordpress.org/support/topic/stripe-krw-and-jpy-currency-issue/

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntentParameters+Stripe.swift
```diff
@@ -9,7 +9,13 @@ extension Hardware.PaymentIntentParameters {
             return nil
         }
 
-        let returnValue = StripeTerminal.PaymentIntentParameters(amount: self.amount, currency: self.currency)
+        /// The amount of the payment needs to be provided in the currency’s smallest unit.
+        ///https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)amount
```
allendav, line 13: Space needed after ///

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntentParameters+Stripe.swift
ctarda: Good point. Logged #3948 for it

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntentParameters+Stripe.swift
ctarda: Will update in an upcoming PR

Issue #3886, opened 2021-03-29
Add-Ons: Parse and store product add-on metadata


Ecarrion:
The product detail response includes add-on in its meta-data object.

We need to parse and store that add-on metadata in order to cross-reference it later with the order detail metadata

As this is a temporary solution, we shouldn't spend much time trying to parse this metadata into a very complex structure as it will be deleted later.

```json
{
  "meta_data": [
    {
      "id": 6471,
      "key": "_product_addons",
      "value": [
        {
          "name": "Topping",
          "title_format": "label",
          .
          .
          .
       }
    }
  ]
}
```

PR #3942, opened 2021-04-09
Add-Ons: Parse add-ons from product metadata

Ecarrion:
part of #3886

# Why

As part of showing requested add-ons for an order, we first need to parse the add-ons from the product response.

This PR takes care of converting that add-on info into a `ProductAddOn` networking entity. Which also adds some new entities as dependencies:

- ProductAddOn
- ProductAddOnOption
- AddOnType
- AddOnTitleFormat
- AddOnRestrictionsType
- AddOnPriceType
- AddOnDisplay

# How

- The add-on parsing is fairly straightforward, we are just declaring the necessary `decodable` structs to let swift do its thing.

- The complicated bit is on accessing the add-on JSON objects from the product `metadata`, as the `metadata` is not a typed object and it is wrapped in not ideal JSON structure(Array with objects, where the "key" key is "_product_addons" and its "value" key has the array of add-ons).

To isolate that complexity I've created a `ProductAddOnEnvelope` type that provides a `resolve()` function to try to extract those add-ons.

```swift
"meta_data": [
 {
  "id": 6469,
  "key": "_last_editor_used_jetpack",
  "value": "classic-editor"
},
{
  "id": 6471,
  "key": "_product_addons",
  "value": [
    {
      "description": "Pizza topping",
      "title_format": "label",
      "required": 0,
    }
  ]
}
]
```

- Ran `rake generate` to generate `fakes` and `copiable` methods for the new types

# Testing Steps
- I've added a new test case to check that the above JSON parses correctly, so making sure the test pass should be enough for now! 
 

Update release notes:


Networking/Networking/Model/Product/ProductAddOnEnvelope.swift
```diff
@@ -0,0 +1,77 @@
...
+///  "key": "_product_addons",
+///  "value": [
+///    {
+///      "description": "Pizza topping",
+///      "title_format": "label",
+///      "required": 0,
+///    }
+///  ]
+///}
+///]
```
ealeksandrov: nit: JSON snippet starts with 1 space left interval (from ///), but then it slips to 0 interval.

Networking/Networking/Model/Product/ProductAddOnEnvelope.swift
```diff
@@ -0,0 +1,77 @@
...
+        let jsonDecoder = JSONDecoder()
+        return try addOnJsonObjects.map { try decode(addOnJsonObject: $0, using: jsonDecoder) }
+    }
+
+    /// Searches for addOns json objects inside the product metadata
+    /// Returns nil if none are found.
+    ///
+    private func getAddOnsJsonObjects() -> [AnyDictionary]? {
+        guard let envelopeDictionary = getAddOnsEnvelopeDictionary() else {
+            return []
```
ealeksandrov: Maybe `return nil`?

ealeksandrov (approved):
Looks good, just couple of small nits 🚀 
Thank you for adding many helpful comments!

Ecarrion:
Thanks for the review @ealeksandrov 🙇

PR #3955, opened 2021-04-12
Shipping Label M2: add new package screen that works like a middleware

### LINKED ISSUE #3911: Shipping Labels M2: add the new package screen that works like a middleware
As part of Shipping Labels M2, we should implement a screen that works like a middleware, where the user from the menu on top can select one of the options between "Custom Package" and "Service Package".
I imagine a SwiftUI view, that contains the switch and that shows below a SwiftUI subview for the custom package or for the service package.

 
Schermata 2021-04-06 alle 12 55 19

pmusolino:
Closes #3911 

## Description
As part of M2, I implemented a screen that works like a middleware, and that for the moment will be useful in the next PR for showing the list of Service Packages, but in the future (in the M4) will be useful also for showing the screen for adding a Custom Package. So, I implemented also a reusable component called `SegmentedView` which will be shown on top of the view (for the moment is hidden) for selecting the custom package or the service package menu.

## Testing
2. Open an order detail.
8. Press the "Package selected" row. -> You should see the new empty screen "Add New Package".
9. Remove the `.hidden()` method in `ShippingLabelAddNewPackage.swift`, re-launch the app, and redo the entire testing procedure. -> This time in the Add New Package screen you should see the `SegmentedView` on top.

## Screenshots

| Add New Package container view             |  Add New Package container view with SegmentedView displayed |
Simulator Screen Shot - iPhone 12 Pro - 2021-04-12 at 17 09 58 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-12 at 17 13 36

Update release notes:


WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/SegmentedView.swift
```diff
@@ -0,0 +1,63 @@
...
+                            .foregroundColor(Color(.clear))
+                    }
+                }
+            }
+        }
+    }
+
+    func getContentView(_ index: Int) -> some View {
+        return views[index]
+            .frame(maxHeight: .infinity)
```
allendav, line 47: I'm curious, why `.infinity` and not something like the device height?

allendav (approved):
Tests well. One question. :shipit:

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/SegmentedView.swift
pmusolino: That's a good question. The main reason is that since the SegmentedView was implemented with the reuse in mind, and since a developer can add inside anything (not only text), we don't know the height of the content. So there may be cases where:
1) The `SegmentedView` is contained inside a `ScrollView`, so in that case, it's wrong to declare the `maxHeight` equal to the height of the device, maybe the content of the `SegmentedView` is greater than the device height.
2) The content inside can be dynamic, and we don't know the maximum height. Imagine a simple Text that will have different sizes based on the settings of their device.

pmusolino:
Thank you @allendav for the review 🙇

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/SegmentedView.swift
allendav: Thanks @pmusolino !

PR #3956, opened 2021-04-12
AddOns: Cross-reference order attributes with product add-ons

### LINKED ISSUE #3887: Add-Ons: Cross-reference order metadata with product metadata
With #3885 and #3886 implemented, we can create a use case that cross-reference order metadata and product add-on metadata to output an object that contains only the order add-on information.

This would be used later to render the order add-on information on a separate screen.

Ecarrion:
closes #3887 

# Why

As you may know(p91TBi-4MK-p2), add-on information for orders is disguised as raw order item attributes. In order to know which attribute from an order is an add-on, we need to cross-reference the attribute name with the add-on name.

This PR creates a `UseCase` whose responsibility is to filter order item's attributes that are real add-on attributes. 

# How

- Added `AddOnCrossreferenceUseCase` that receives an `AggregatedorderItem` and a `Product`. This use case has an `addOnsAttributes()` method that infers the `AggregatedorderItem.attributes` names and checks if they exist on the product add-on's list.

- The method that infers the `AggregatedorderItem.attributes` names is a simple heuristic that takes a string with the format `"add-on-title (add-on-price)"` and takes out the last part of the string by splitting it via the `" ("` token.

# Testing Steps
- I've added new tests, so make sure the CI is green!

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/AddOns/AddOnCrossreferenceUseCase.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/AddOns/AddOnCrossreferenceUseCase.swift
@@ -1,0 +1,1 @@
+import Foundation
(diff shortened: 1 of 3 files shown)


pmusolino (approved):
Looks good to me 💯 

Code ✅ 
CI ✅

Ecarrion:
Thanks(again!) for the review @pmusolino :-)

Issue #3891, opened 2021-03-30
Present a clearer message if the site is password-protected, XML-RPC is blocked, or cannot be reached


shiki:
Found in https://github.com/woocommerce/woocommerce-ios/issues/3883#issuecomment-809819474
Solutions discussed in p91TBi-4Pz-p2

## Describe the bug

The app responds with an indistinct message (“Not a WordPress site”) when logging in with a site that has one of these conditions:

- [ ] Is password-protected via a basic HTTP auth. (moved to #3883)
- [x] If the site blocks XML-RPC. #3957 
- [x] If the device is offline #3957 
- [x] If the site is HTTP only #3957 

### Password-protected

I performed similar steps with my site:

1. Create a WooCommerce site that is connected to Jetpack.
2. I added a basic HTTP auth to the site. [Here](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/) is an example for Nginx.  Ask @shiki if you need help setting up a site. 
3. Log in by clicking on "Enter your Store Address" and enter the site address.

Here is the result on Android but it's the same on iOS: 

<img src="https://user-images.githubusercontent.com/198826/112916639-2e4e4700-90be-11eb-89a2-fd5d1daa8311.png" width="300">

### XML-RPC 

This will probably work → [How to Disable XML-RPC in WordPress](https://www.wpbeginner.com/plugins/how-to-disable-xml-rpc-in-wordpress/)

## Expected behavior

I would expect a more specific message so that I can troubleshoot further. 

**Isolating the problem (mark completed items with an [x]):**
- [x] I have deactivated other plugins and confirmed this bug occurs when only WooCommerce plugin is active.
- [x] This bug happens with a default WordPress theme active, or [Storefront](https://woocommerce.com/storefront/).
- [x] I can reproduce this bug consistently using the steps above.

## Mobile Environment

Please include:
- Device: iPhone 11 simulator
- iOS version: 14
- WooCommerce iOS version: a0ebe53

## WordPress Environment
<details>
```
Copy and paste the system status report from **WooCommerce > System Status** in WordPress admin.
```
</details>

--- 

Android: https://github.com/woocommerce/woocommerce-android/issues/3767

designsimply:
Answering a question [from another thread on a related issue](https://github.com/woocommerce/woocommerce-ios/issues/3883#issuecomment-810421656):

[...]

PR #3957, opened 2021-04-12
Improve Login Error Handling 

shiki:
Fixes part of #3891.  This improves the error messages that the merchant receives when logging in with a site with an offline device, or an invalid SSL certificate, or a blocked XML-RPC. 

As we discovered in #3891  we default to showing “The website is not a WordPress site...”. That message is not specific and does not help our customers or support folks in troubleshooting further. This PR improves some of that. 

### Offline

This is using the default UI and the default implementation of handling offline messages provided by the [WP Authenticator library](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS). How does this look? 

Before | After 
--------|-------
        |       

### Blocked XML-RPC

Blocking XML-RPC is common. Some hosting providers also block it by default. This UI uses the default provided by the WP Authenticator library. The message is coming from the WP Auth library too. And there [are quite a lot of them](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/4062408fce1596b4744449fd950f9d47ddd4d7eb/WordPressKit/WordPressOrgXMLRPCValidator.swift#L16-L34). Should we display this differently because they're quite technical? 

Before | After 
--------|-------
|

### Invalid SSL Certificate

In p91TBi-4Pz-p2, we confirmed that we want to keep the requirement of WooCommerce sites having valid SSL certificates. We currently don't display that information. I added an error message when the merchant tries to log in with a site that does not have SSL. Please suggest better wording! 

Before | After 
--------|-------
|

## Technical Changes ⚙️ 

The WPKit changes in https://github.com/wordpress-mobile/WordPressKit-iOS/pull/382 fixed the problem of WPKit incorrectly returning `NSURLErrorAppTransportSecurityRequiresSecureConnection` (`-1022`) for most unexpected login errors. Now that we have that fixed, the default offline and XML-RPC error messages (shown above) come for free. 🎉 

I then changed the [`AuthenticationError` detection](https://github.com/woocommerce/woocommerce-ios/pull/3957/files#diff-d6518838fddd72ad3679fdd85061c39f0c727d2efec93f40d4a8339c5b7617f8) to preserve the error messages that we currently handle like the `.notValidAddress`. 

I also added a new error screen if the site is not using SSL, [`NoSecureConnectionErrorViewModel`](https://github.com/woocommerce/woocommerce-ios/blob/issue/3891-fix-offline-messaging/WooCommerce/Classes/Authentication/Navigation%20Exceptions/NoSecureConnectionErrorViewModel.swift). 

### Unit Tests

I'm not sure how to add unit tests for these changes. The [`AuthenticationManager.make(with:)`](https://github.com/woocommerce/woocommerce-ios/blob/059da38b74f6581b914f4d42fa4c2cd89682b7cc/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L405) seems like a good candidate but it's just a `switch` so I wasn't sure if it's worth it. Open to suggestions! 

## <a name="testing">Testing</a> 🧪 

1. Log out.
2. Tap “Enter Your Store Address”.
3. Enter a site that does not have HTTPS. If you don't have one, you can use http://httponly.peregrinator.net. The password is available in the Secret Store.
4. Confirm that an error screen is shown: “A secure connection to the site could not be made...”.
5. Go offline.
6. Enter the site URL again.
7. Confirm that the error “The Internet connection appears to be offline.” is displayed. 
8. Go online.
9. Enter a site with blocked XML-RPC. You can use https://do.wpmt.co/ddos. Press Continue.
10. Confirm that the error “... We received a 403 error...” is displayed.
11. Go back to the Login Prologue.
12. Tap “Enter Your Store Address”. 
13. Enter a self-hosted WooCommerce site that has HTTPS. 
14. Confirm that an error is shown: “To use this app, ... you'll need to have the Jetpack plugin installed...”. Or if you activated Jetpack on the site, you should be asked for your WordPress.com password.

## Known Side Effect

There is a bug from WPAuth that is made possible by this change. See https://github.com/woocommerce/woocommerce-ios/issues/4034. I spent some time looking into it but I'm not quite sure whether to fix it in WPAuth or in WC. 

I think it's fine to merge this since the bug only happens when the merchant previously had an XML-RPC error. I'll fix it next. 

## Author Checklist


shiki:
@woocommerce/mobile-designers Would you mind checking the designs of these error messages and let me know what you think?

ghost:
These look really good to me. @Garance91540 do you have any concerns?

Garance91540 (approved):
Looks awesome! I only noted a typo for the error "a secure connexcion..." with an extra punctuation dot
https://d.pr/i/7ghGpY

Is it possible to make sure Android gets the same improvements?

shiki:
Thank you, @adamzelinski and @Garance91540! 

> Looks awesome! I only noted a typo for the error "a secure connexcion..." with an extra punctuation dot

Fixed! 

> Is it possible to make sure Android gets the same improvements?

This is tracked in https://github.com/woocommerce/woocommerce-android/issues/3767. I'll see if I can do it myself. If not, I'll ask someone else. 😅

RELEASE-NOTES.txt
```diff
@@ -4,6 +4,7 @@
 -----
 - [*] Fix: Product variations only support at most one image, so we won't show an option to add a second one. [https://github.com/woocommerce/woocommerce-ios/pull/3994]
 - [*] Fix: The screen to select images from the Media Library would sometimes crash when the library had a specific number of images. [https://github.com/woocommerce/woocommerce-ios/pull/4003]
+- [*] Improved error messages for logins.
```
ealeksandrov: nit: add link to this PR?

ealeksandrov (approved):
This is a very important improvement for login flow, great work! 👏 
I've got expected error messages and code looks good :shipit:

RELEASE-NOTES.txt
shiki: Lol can you tell that I don't know how to contribute code anymore? 🤣 Thanks for catching it! Fixed in 8ff08d8e0e7c888e205447a310a18b8110b8ae24.

Issue #3966, opened 2021-04-15
Shipping Labels M2: manage the `syncing` state in Package List


pmusolino:
Using the ghost cells https://github.com/woocommerce/woocommerce-ios/issues/3953, we should manage the `syncing` state in the package list screen, displaying 3 empty ghost cells like in other screens (look at Order list or Product list).

<img width="675" alt="114556965-8150fe00-9c69-11eb-8919-c546a38cff56" src="https://user-images.githubusercontent.com/495617/114865388-7675a500-9df2-11eb-8214-381013a9fe9a.png">

Issue #3965, opened 2021-04-15
Shipping Labels M2: handle the `error` state in package list + add pull to refresh


pmusolino:
Since the Package List screen has 3 different states: `syncing`, `results`, `error`, we should handle the error state, and offer a way to refresh the content of the view with a pull to refresh action.

<img width="675" alt="114556965-8150fe00-9c69-11eb-8919-c546a38cff56" src="https://user-images.githubusercontent.com/495617/114865209-38788100-9df2-11eb-8e67-b59a3a80ed20.png">

Issue #3964, opened 2021-04-15
Shipping Labels M2: handle the selected value in the Package List


pmusolino:
As part of Shipping Labels M2, after having implemented the package list screen, https://github.com/woocommerce/woocommerce-ios/issues/3958 we should provide a method to handle the selected package in `ShippingLabelPackageList` view, and add a `Done` button with a callback on top of `ShippingLabelAddNewPackage` screen.

<img width="675" alt="114556965-8150fe00-9c69-11eb-8919-c546a38cff56" src="https://user-images.githubusercontent.com/495617/114864942-e6cff680-9df1-11eb-9c79-22e1846a3e7f.png">

PR #3967, opened 2021-04-15
Shipping Labels M2: display the Package List

### LINKED ISSUE #3958: Shipping Labels M2: display the service package list
Clicking on the package name text field opens a new screen that displays a list of default packages for the store.

Schermata 2021-04-13 alle 15 03 28

pmusolino:
Closes #3958  

## Description
In this PR I implemented the screen for displaying the Package List in the Shipping Labels.
All the components are implemented in SwiftUI. The flow was implemented using a view that contains another subview. Wanting to be more specific, `ShippingLabelAddNewPackage` is the container of the reusable `ShippingLabelPackageList` view. I implemented it in this way because in the future we will use the `ShippingLabelPackageList` component for the Package Creation Flow. So, the `ShippingLabelPackageList` will be embedded in another view where the user can create a Custom Package or select an existing package.

## Changes
- Implemented the `ShippingLabelPackageList` view that displays a list of Custom Packages and Predefined Packages. I also implemented the view model `ShippingLabelPackageListViewModel` and the tests. The view model takes care of managing the state of the view and the selected package.
- I made some small UI changes to the existing `ShippingLabelAddNewPackage` view, removing the unused (for the moment) `SegmentedView`. I also made some changes to the view model, which now takes care of the States and of fetching the package details. I added the unit tests also for this view model.
- I implemented a new reusable view in SwiftUI, called `SelectableItemRow` that displays a title, a subtitle, and the checkmark on the left side when selected.

## Testing
2. Open an order detail.
8. Press the "Package selected" row. -> You should see the new screen with the list of the existing packages, with Custom Packages on top if any.

For the moment the `error` and the `syncing` states are not managed and will be part of other PRs. https://github.com/woocommerce/woocommerce-ios/issues/3965 https://github.com/woocommerce/woocommerce-ios/issues/3966 Also the handle of the value will be part of another PR https://github.com/woocommerce/woocommerce-ios/issues/3964

| Package List (Light mode)             |  Package List (Dark mode)  |
Simulator Screen Shot - iPhone 12 Pro - 2021-04-15 at 14 26 30 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-15 at 14 27 19

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Add New Package/Service Package List/ShippingLabelPackageList.swift
```diff
@@ -0,0 +1,92 @@
...
+struct ShippingLabelPackageList: View {
+    @ObservedObject private var viewModel: ShippingLabelPackageListViewModel
+
+    var body: some View {
+        VStack(spacing: 0) {
+
+            /// Custom Packages
+            ///
+            if viewModel.customPackages.count > 0 {
+                ListHeaderView(text: Localization.customPackageHeader, alignment: .left)
```
rachelmcr: Would it make sense to use `Localization.customPackageHeader.uppercased()` here (similar to the headers for Predefined Packages below) instead of relying on the string `Localization.customPackageHeader` to be in uppercase?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Add New Package/Service Package List/ShippingLabelPackageList.swift
```diff
@@ -0,0 +1,92 @@
...
+        static let dividerPadding: CGFloat = 48
+    }
+}
+
+struct ShippingLabelPackageList_Previews: PreviewProvider {
+    static var previews: some View {
+        let storeOptions = ShippingLabelStoreOptions(currencySymbol: "$", dimensionUnit: "in", weightUnit: "oz", originCountry: "US")
+
+        let customPackages = [
+            ShippingLabelCustomPackage(isUserDefined: true, title: "Box", isLetter: true, dimensions: "3 x 10 x 4", boxWeight: 10, maxWeight: 11),
```
rachelmcr: nit: For consistency/readability, these parameters could each be on their own line (like the other `ShippingLabelCustomPackage`s below).

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/SelectableItemRow.swift
```diff
@@ -0,0 +1,49 @@
...
+                Text(title)
+                    .font(.body)
+                Text(subtitle)
+                    .font(.footnote)
+                    .foregroundColor(Color(.textSubtle))
+            }.padding([.trailing], Constants.vStackPadding)
+            Spacer()
+        }
+        .padding([.top, .bottom], Constants.hStackPadding)
+        .frame(minHeight: Constants.height)
```
rachelmcr, line 27: I noticed that when `SelectableItemRow` is used, only the part of the row with text in it is tappable.

Based on [this article](https://www.hackingwithswift.com/quick-start/swiftui/how-to-control-the-tappable-area-of-a-view-using-contentshape), it looks like we can make the whole row tappable by adding `contentShape()`. Adding something like this to the `HStack` should work:

```swift
.frame(minHeight: Constants.height)
.contentShape(Rectangle())
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Add New Package/ShippingLabelAddNewPackage.swift
```diff
@@ -1,32 +1,31 @@
...
         static let title = NSLocalizedString("Add New Package", comment: "Add New Package screen title in Shipping Label flow")
-        static let customPackage = NSLocalizedString("Custom Package", comment: "Custom Package menu in Shipping Label Add New Package flow")
-        static let servicePackage = NSLocalizedString("Service Package", comment: "Service Package menu in Shipping Label Add New Package flow")
     }
 }
 
 struct ShippingLabelAddNewPackage_Previews: PreviewProvider {
     static var previews: some View {
-        ShippingLabelAddNewPackage()
+        ShippingLabelAddNewPackage(siteID: 1234)
```
rachelmcr: I can't think of a straightforward way to do this, but do you think it'd be possible to fake the site in a way that will make this SwiftUI preview work? As it is, the preview just shows a blank screen. It isn't critical, but it'd be nice to have the preview working.

rachelmcr (commented):
This is really nice! One thing I'm not sure about, though, is the container view:

> Wanting to be more specific, ShippingLabelAddNewPackage is the container of the reusable ShippingLabelPackageList view. I implemented it in this way because in the future we will use the ShippingLabelPackageList component for the Package Creation Flow.

In the designs, the package list is displayed on a screen titled “Package Selected.” In this implementation, the screen is titled “Add New Package” (which is part of the creation flow design). Should we have two different container views that we can use depending on which flow we're in?

I also noticed that the screen title "Add New Package" is also showing up in the very bottom left (see my screenshots); do you know why that's happening?

I have a few other comments below, but overall this is working really well for me. A few screenshots from testing with different package lists:

Custom & Predefined packages | Predefined only | Custom only
-|-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 16 00 00|Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 16 08 46|Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 16 09 51

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/SelectableItemRow.swift
pmusolino: Good catch! Fixed 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Add New Package/ShippingLabelAddNewPackage.swift
pmusolino: I thought about this thing, but this would involve providing a method to access data from the outside (therefore without going through the Yosemite layer), and which would be used only for the preview, and I didn't particularly like the solution. 🤔

pmusolino:
Thank you @rachelmcr for the review! I fixed all the things that you discovered 👍 

> In the designs, the package list is displayed on a screen titled “Package Selected.” In this implementation, the screen is titled “Add New Package” (which is part of the creation flow design). Should we have two different container views that we can use depending on which flow we're in?

You are right, I fixed it and renamed the `ShippingLabelAddNewPackage` view as `ShippingLabelPackageSelected`. 

> I also noticed that the screen title "Add New Package" is also showing up in the very bottom left (see my screenshots); do you know why that's happening?

I tried to investigate why this happens, but I wasn't able to find a quick solution, so I opened an issue here https://github.com/woocommerce/woocommerce-ios/issues/3995 I'll fix it before shipping the M2.

Can you do a second check, please? 🙇

rachelmcr (approved):
Looks great! 🎉   :shipit:

PR #3968, opened 2021-04-16
Add-Ons: Create view to render order add-ons

### LINKED ISSUE #3893: Add-Ons: Create view to render add-ons
After addons are identified in #3887 We should display them on a single table similar to:

Screen Shot 2021-04-15 at 8 07 34 AM

This view should navigate from the button added in #3889

Ecarrion:
closes #3893

# Why

This PR makes sure the user can navigate from the order item cell to a screen where the product add-ons are rendered in a very simple list.

Note: This is my first PR using `SwiftUI` and I'm trying to put what I consider "best" practices(from my limited experience) here: paNNhX-cY-p2.

# How

- Updated `OrderDetailsDataSource` to provide the necessary callback when the "View add-ons" button is pressed with the correct add-on attributes.

- Created a `WooStyleModifiers` file to migrate functions from `UILabel+Helpers` and `UIButton+Helpers` as they are needed in `SwiftUI` views.

- Created `OrderAddOnListI1View` which is a `SwiftUI` view that renders one or more `OrderAddOnI1View` views based on it's view model content.

- Created `OrderAddOnListI1ViewModel` which is in charge of receiving order item attributes and converting them to view models that can be consumed by the `OrderAddOnListI1View`. Here the workaround magic happens. 

- Created `OrderAddOnI1ViewModel` which represents a single add-on, to be consumed by `OrderAddOnListI1View`.

- Created `OrderAddOnsListViewController` as a hosting controller of `OrderAddOnListI1View`

- Updated `OrderDetailsViewController` to launch the newly created `OrderAddOnsListViewController`

# Screenshots

Light | Dark | Big Font
--- | --- | ---
regular | dark | big font

# Testing Steps

**Prerequisites:** Have your site with the add-ons plugin installed and configured.

- Make an order with an item with add-ons
- Launch the app and navigate to that order
- See that a new screen is presented with the add-on's information.

Update release notes:


WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
```diff
@@ -0,0 +1,53 @@
...
+            .font(.body)
+            .foregroundColor(Color(.text))
+    }
+}
+
+extension View {
+    func bodyStyle() -> some View {
+        self.modifier(BodyStyle())
+    }
+}
```
pmusolino: What do you think about creating a single `View` extension, with all the methods inside? It seems repetitive to add the same extension for each method, and it makes reading more difficult.

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
```diff
@@ -0,0 +1,53 @@
+import Foundation
+import SwiftUI
+
+/// Woo style modifiers.
+/// Migrate them from `UILabel+Helpers` or  `UIButton+helpers` as needed.
+///
+
+// MARK: Body Style
+struct BodyStyle: ViewModifier {
```
pmusolino, line 9: I like a lot your proposal of the view modifiers. I will update my existing code to start using them.

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+/// Hosting controller that wraps an `AddOnsListView`
+///
+class OrderAddOnsListViewController: UIHostingController<OrderAddOnListI1View> {
+    init(viewModel: OrderAddOnListI1ViewModel) {
+        super.init(rootView: OrderAddOnListI1View(viewModel: viewModel))
+        self.title = viewModel.title
+        addCloseNavigationBarButton()
+    }
+
+    @objc required dynamic init?(coder aDecoder: NSCoder) {
```
pmusolino: nit: `@objc` doesn't seem needed.

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+/// Renders a list of add ons
+///
+struct OrderAddOnListI1View: View {
```
pmusolino, line 21: Isn't better to move this View to a different file?

pmusolino, line 21: It iss not totally clear to me because in the name there is the `I1` 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
+import UIKit
+import SwiftUI
+import Yosemite
+
+/// Hosting controller that wraps an `AddOnsListView`
+///
+class OrderAddOnsListViewController: UIHostingController<OrderAddOnListI1View> {
```
pmusolino: nit
```suggestion
final class OrderAddOnsListViewController: UIHostingController<OrderAddOnListI1View> {
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+/// Renders a list of add ons
+///
+struct OrderAddOnListI1View: View {
+
+    /// Static view model to populate the view content
+    let viewModel: OrderAddOnListI1ViewModel
```
pmusolino, line 24: nit
```suggestion
    private let viewModel: OrderAddOnListI1ViewModel
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    /// Discussion: Due to the inability of customizing the `List` separators. We have opted to simulate the list behaviour with a `ScrollView` + `VStack`.
+    ///          We expect performance to be acceptable as normally an order does not have too many add-ons.
+    ///          A future improvement can be to use a `LazyVStack` when iOS-14 becomes our minimum target.
```
pmusolino: nit: there is some extra space in front of the comment that can be removed 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+    ///
+    let viewModel: OrderAddOnI1ViewModel
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 10) {
+            Divider()
+
+            Text(viewModel.title)
+                .headlineStyle()
+                .padding(.leading)
```
pmusolino: nit: add also the trailing padding, or a long text will touch the right margin.
```suggestion
                .padding([.leading, .trailing])
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+            }
+            .padding([.leading, .trailing])
+
+            Divider()
+        }
+        .background(Color(.basicBackground))
+    }
+}
+
+#if DEBUG
```
pmusolino, line 74: nit: not needed

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+
+struct OrderAddOnView_Previews: PreviewProvider {
+    static var previews: some View {
+        OrderAddOnListI1View(viewModel: .init(addOns: [
+            OrderAddOnI1ViewModel(id: 1, title: "Topping", content: "Pepperoni", price: "$3.00"),
+            OrderAddOnI1ViewModel(id: 2, title: "Topping", content: "Salami", price: "$2.00"),
+            OrderAddOnI1ViewModel(id: 3, title: "Soda", content: "3", price: "$6.00"),
+            OrderAddOnI1ViewModel(id: 4, title: "Instructions", content: "Leave it in the front door", price: "")
+        ]))
+        .environment(\.colorScheme, .light)
```
pmusolino, line 84: nit: not needed because by default it is always `light` 👍

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/AddOns/OrderAddOnListI1Tests.swift
```diff
@@ -0,0 +1,28 @@
+import XCTest
+import Fakes
+
+@testable import WooCommerce
+@testable import Yosemite
+@testable import Networking
```
pmusolino: nit: this import seems not needed

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnListViewModel.swift
```diff
@@ -0,0 +1,76 @@
+import Foundation
+import Yosemite
+
+/// ViewModel for `OrderAddOnListI1View`
+///
+class OrderAddOnListI1ViewModel {
```
pmusolino: nit
```suggestion
final class OrderAddOnListI1ViewModel {
```

pmusolino (approved):
Congrats for opening your first PR with SwiftUI 👏 

Testing ✅ 
Code ✅ just some nits, but it looks good!

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
Ecarrion: I was on the fence here, at the end I choose to have the modifier and the extension close to each other to be able to avoid unnecessary scrolling. but I'm fine either way!

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
Ecarrion: Xcode autocompleted that for me, will remove!

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
Ecarrion: Since `OrderAddOnsListViewController` is very minimal and this is its view I think it makes sense to stay there 😅

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
Ecarrion: Is not private, so I can make use of the automatic memberwise-initializer!

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
Ecarrion: Nice catch!

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
Ecarrion: If we don't add the compiler tag the preview type would be included in our production binary, right? or am I missing something?

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
Ecarrion: I rather leave it to be able to change it to dark easily!

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnsListViewController.swift
Ecarrion: The `i1` is for iteration 1(workaround)

Issue #3940, opened 2021-04-09
Shipping Labels M2: display the last package selected


pmusolino:
After discussing with @anitaa1990, it seems that we should show the last package selected in the "Package Details screen", and we need to fetch the Shipping Label Settings for it.
There is some work already done, and probably we just need to update the model `ShippingLabelSettings` and update the Storage layer for this.

rachelmcr:
The last package selected comes from the WCS Settings endpoint `/wc/v1/connect/account/settings` (`last_box_id` in `userMeta` in the response). This is different from the Shipping Label Settings we already support (which are per order).

To display the last package selected we'll need:

- [x] Support for `GET /wc/v1/connect/account/settings` in the Networking layer #3983 
- [x] New entity representing WCS Settings in the Storage layer, to save the last package ID (`last_box_id`). #3996
- [x] Support for WCS Settings in the Yosemite layer #4016

I'm also checking to see if `last_box_id` includes details about the last custom package selected, or if it only returns service packages (since custom packages only have a `name`, not an `id`).

rachelmcr:
@hichamboushaba clarified a couple points about how we're doing this on the Android side. (Thanks!)

Handling both service and custom packages:

> For custom packages, `last_box_id` returns the name of the package, so to make things clearer on our side, we decided to add a property id to the custom package class in the business model layer, and copy the name value to it. Then we just compare the id from our class to what last_box_id returns.

When to fetch the endpoint:

> We also fetch the endpoint when the form is loaded (before showing any steps), and we cache the result while the form is open. We do so because the result contains the payment methods that we need also to decide whether the payment step is required or not, if a payment method is already selected, the user shouldn't have to re-select it.

PR #3983, opened 2021-04-16
Shipping Labels: Networking layer for account settings endpoint

rachelmcr:
Part of: #3940 

## Description

This PR adds support in the Networking layer for the Shipping Label account settings endpoint (`/wc/v1/connect/account/settings`). We currently need this endpoint to fetch the last selected package for shipping labels, and it will also be used in the future for handling payment methods.

Note: The settings returned in this endpoint are different from other Shipping Label settings we already support. Our existing support is for shipping labels linked to specific orders. The settings in this endpoint are account-level (such as user capabilities for editing payment settings and methods) and store-level (such as store owner details and payment methods).

Future PRs will add support in the Yosemite and Storage layers.

## Changes

* Adds `ShippingLabelRemote.loadShippingLabelAccountSettings` to make the API request to fetch the account settings.
* Adds the mapper `ShippingLabelAccountSettingsMapper` to handle the API response.
* Adds two new models `ShippingLabelAccountSettings` and `ShippingLabelPaymentMethod` and an enum `ShippingLabelPaymentCardType`.
   * **Question:** Should we also inject `userID` into the `ShippingLabelAccountSettings` model, since it includes account-specific settings?
* Adds mocks and unit tests for remote success/failure responses in `ShippingLabelRemoteTests`, and unit tests for the mapper parsing in `ShippingLabelAccountSettingsMapperTests`.

## Testing

Confirm tests pass in CI (endpoint is not yet used in app).

## Submitter Checklist

Update release notes:

### DIFF
--- a/Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
+++ b/Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
@@ -1,0 +1,60 @@
+import Foundation
+
+/// The supported card types for Shipping Label Payment Methods.
+///
+public enum ShippingLabelPaymentCardType: GeneratedFakeable {
+    case amex
+    case discover
+    case mastercard
+    case visa
+    case paypal
+}
+
+/// RawRepresentable Conformance
+extension ShippingLabelPaymentCardType: RawRepresentable {
+    /// Designated Initializer.
+    ///
+    public init(rawValue: String) {
+        switch rawValue {
+        case Keys.amex:
+            self = .amex
+        case Keys.discover:
+            self = .discover
+        case Keys.mastercard:
+            self = .mastercard
+        case Keys.visa:
+            self = .visa
+        case Keys.paypal:
+            self = .paypal
+        default:
+            assertionFailure("Unexpected value for `ShippingLabelPaymentCardType`: \(rawValue)")
+            self = .amex
+        }
+    }
+
+    /// Returns the current Enum Case's Raw Value
+    ///
+    public var rawValue: String {
+        switch self {
+        case .amex:
+            return Keys.amex
+        case .discover:
+            return Keys.discover
+        case .mastercard:
+            return Keys.mastercard
+        case .visa:
+            return Keys.visa
+        case .paypal:
+            return Keys.paypal
+        }
+    }
+}
+
+/// Contains the supported ShippingLabelPaymentCardType values.
+private enum Keys {
+    static let amex = "amex"
+    static let discover = "discover"
+    static let mastercard = "mastercard"
+    static let visa = "visa"
+    static let paypal = "paypal"
+}
--- a/Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
+++ b/Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
@@ -2,59 +2,10 @@ import Foundation
 
 /// The supported card types for Shipping Label Payment Methods.
 ///
-public enum ShippingLabelPaymentCardType: GeneratedFakeable {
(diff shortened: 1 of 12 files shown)


Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+
+/// The supported card types for Shipping Label Payment Methods.
+///
+public enum ShippingLabelPaymentCardType: GeneratedFakeable {
```
ealeksandrov: We can use simple decodable enum here instead of custom `RawRepresentable` conformance.
Any reasons against that I'm missing?

```suggestion
public enum ShippingLabelPaymentCardType: String, Decodable, GeneratedFakeable {
```

It will also require a change in `Decodable` extension for `ShippingLabelPaymentMethod`:
```swift
let cardType = try? container.decode(ShippingLabelPaymentCardType.self, forKey: .cardType)
```

ealeksandrov (approved):
Looks good!
Just one question about `ShippingLabelPaymentCardType` to check and 🚀

> Should we also inject userID into the ShippingLabelAccountSettings model, since it includes account-specific settings?

Could we possibly have a situation where we query for stored `ShippingLabelAccountSettings` with same siteID, but with a different user? As far as I know, when user logs out - we delete all stored data. So at least for now - it doesn't matter. For logical completeness we can still add it, you're right that it is account related 🙂

Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
```diff
@@ -0,0 +1,60 @@
...
+            self = .discover
+        case Keys.mastercard:
+            self = .mastercard
+        case Keys.visa:
+            self = .visa
+        case Keys.paypal:
+            self = .paypal
+        default:
+            assertionFailure("Unexpected value for `ShippingLabelPaymentCardType`: \(rawValue)")
+            self = .amex
```
ealeksandrov: Is there a possibility that we can get new kind of CardType here? maybe add `.unknown` case for that? Or even `.unknown(String)`.

ProductStatus for example:
https://github.com/woocommerce/woocommerce-ios/blob/0f15625548d2ef2859cf6baf5dce97d22e4f543e/Networking/Networking/Model/Product/ProductStatus.swift#L11

Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
rachelmcr: Oh yes, that's much simpler, thanks! I don't think there's any reason not to, at least at this point, so I made that change in f1b2d53.

Networking/Networking/Model/ShippingLabel/Enums/ShippingLabelPaymentCardType.swift
rachelmcr: No, there shouldn't be any possibility of an unexpected card type here. This isn't a core type that could be modified by another extension; it's only the card types that our Shipping & Tax extension supports, [defined here](https://github.com/Automattic/woocommerce-services/blob/c97384bc78254bd484485f3f781917af605788c2/client/extensions/woocommerce/woocommerce-services/views/label-settings/label-payment-method.js#L18-L25). So we should be ok without a fallback case, as far I understand.

rachelmcr:
> Could we possibly have a situation where we query for stored ShippingLabelAccountSettings with same siteID, but with a different user? As far as I know, when user logs out - we delete all stored data. So at least for now - it doesn't matter.

That's what I was thinking. I don't think there's any expected situation where we'd query for different users and the same site, except after logging out and in again with a different user. I'll leave it as-is at this point since it shouldn't be needed. Thanks!

Issue #3912, opened 2021-04-06
Coupon management (task 1): networking layer


pmusolino:
Coupons Management is one of the core functionalities that we want to offer to our users, and that is part of the Order Creation process. Also if the Coupons Management is part of the Order Creation, we'd like to maintain the flow separated. In this way, we can allow users to customize them from the app settings or from other points of the app.

The first task will be to implement the networking layer for fetching/adding/editing/delete coupons remotely, plus the creation of the models, based on our [API documentation](https://woocommerce.github.io/woocommerce-rest-api-docs/#coupons).

<img width="939" alt="Schermata 2021-04-06 alle 16 50 32" src="https://user-images.githubusercontent.com/495617/113732272-bd301480-96f9-11eb-9b30-02c966580838.png">

joshheald-a8c:
Coupon Model

- [x] Add fields
- [x] Codable
- [x] Comparible
- [x] GeneratedCopiable
- [x] GeneratedFakeable

Field list:
- id
- code
- amount
- date_created_gmt
- date_modified_gmt
- discount_type
- description
- date_expires_gmt
- usage_count
- individual_use
- product_ids
- excluded_product_ids
- usage_limit
- usage_limit_per_user
- limit_usage_to_x_items
- free_shipping
- product_categories
- excluded_product_categories
- exclude_sale_items
- minimum_amount
- maximum_amount
- email_restrictions
- used_by

joshheald-a8c:
CouponListMapper
- [x] Add coupons-all.json
- [x] Create Mapper
- [x] Test list decoding
- [x] Test siteId injection
- [x] Test date decoding strategy
- [x] Test field decoding

joshheald-a8c:
CouponsRemote

- [x] loadAllCoupons method
- [x] deleteCoupon method
- [x] test error relay

joshheald-a8c:
CouponMapper

- [x] Add coupon.json
- [x] Create mapper
- [x] Test object decoding
- [x] Test siteId injection
- [x] Test date decoding strategy
- [x] Test field decoding
- [x] Consider Refactor to share between List and individual mapper

itsmeichigo:
Closing this issue in favor of the new Coupons project.

PR #3985, opened 2021-04-16
Coupon Management: Networking - Coupon model

joshheald-a8c:
Part of #3912

## Description
Adds the Coupon model according to the API docs, except for a few fields which would add confusion and are unlikely to be required, e.g. the dates in the site’s timezone.

Synthesised `Codable` conformance is used as far as possible, which gives us more compiler-provided safety, e.g. around key decoding correctness. Overrides for keys are only given where they're strictly required, and they're in camel case rather than snake case to support using a decoder set to `.convertFromSnakeCase`.

`siteId` is a `private(set) var`, rather than a `let` as we would ideally like. This is required so that we can inject the `siteId` in the mapper, while still benefitting from the synthesised `decode` method. Injection in the mapper will be done via `.copy()`, not mutation, and future improvements in Swift may eventually let us make this a `let` again.

I'll put full details of this approach in a P2 and link to it from here, as it is a slight departure from the established convention. I think that doing it this way would make us less likely to make mistakes while writing model property names out in slightly different ways in several places through the file and tests.

## Testing
The model is not in use yet, so no applicable testing other than the unit tests, and CI once it's merged to the main repo.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Coupon.swift
+++ b/Networking/Networking/Model/Coupon.swift
@@ -1,0 +1,106 @@
+import Foundation
+
+
+// MARK: - Coupon
+
+/// Represents a Coupon entity: https://woocommerce.github.io/woocommerce-rest-api-docs/?shell#coupons
+///
+public struct Coupon {
+    /// `siteId` should be set on a copy in the Mapper as it's not returned by the API.
+    /// Using a default here gives us the benefit of synthesised codable conformance.
+    /// `private(set) var` is required so that `siteId` will still be on the synthesised`init` which `copy()` uses
+    private(set) var siteId: Int64 = 0
+    let couponId: Int64
+    /// The coupon code for use at checkout
+    let code: String
+    /// Discount provided by the coupon, used whether the `discountType` is a percentage or fixed amount type.
+    let amount: String
+    /// Date the coupon was created, in GMT (UTC)
+    let dateCreated: Date
+    /// Date the coupon was modified (or created), in GMT (UTC)
+    let dateModified: Date
+    /// Determines the type of discount that will be applied. Options: `.percent` `.fixedCart` and `.fixedProduct`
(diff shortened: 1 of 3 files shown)


Networking/Networking/Model/Coupon.swift
```diff
@@ -0,0 +1,106 @@
...
+    /// If `true`, this coupon will not be applied to items that have sale prices
+    public let excludeSaleItems: Bool
+    /// Minimum order amount that needs to be in the cart before coupon applies
+    public let minimumAmount: String
+    /// Maximum order amount allowed when using the coupon
+    public let maximumAmount: String
+    /// Email addresses of customers who are allowed to use this coupon, which may include * as wildcard
+    public let emailRestrictions: [String]
+    /// Email addresses of customers who have used this coupon
+    public let usedBy: [String]
```
pmusolino: nit: also if it adds some extra lines, would you like to add a return after every property? I think it becomes more readable.

pmusolino (approved):
Congrats for opening your first PR on `WCiOS` @joshheald-a8c  🥳 

The code looks good to me :shipit: just a minor nit. I'll merge it after you will fix it ;) 

>  I thought I could spend the rest of today reviewing your PR from yesterday and catching up on some p2s and other things.

Since I don't want to block this PR, and since I'm merging it in a separate branch `feat/coupon-management`, we can discuss these aspects in the P2 post, and from what comes out, we can correct it later.

Networking/Networking/Model/Coupon.swift
joshheald-a8c: Thanks! Yep, definitely will do that too, will be a little easier to do it after the other networking PRs are merged.

joshheald-a8c: Done 😊

PR #3986, opened 2021-04-16
Coupon Management: Storage - Add Coupon Core Data model

### LINKED ISSUE #3913: Coupon management (task 2): storage layer
The second task of the Coupon management will be to implement the storage layer, storing in Core Data the coupons of every website when they are fetched.

Here P91TBi-48i you can find one of our internal guides on how to generate Core Data Files.

joshheald-a8c:
Closes #3913

## Description
This PR adds the Core Data classes to represent `Coupon` and its relationships with `Product` and `ProductCategory`, migrating those to include the new relationships.

## Testing
The model is not in use yet, so no applicable testing other than the unit tests, and CI once it's merged to the main repo.

Update release notes:

### DIFF
--- a/Storage/Storage.xcodeproj/project.pbxproj
+++ b/Storage/Storage.xcodeproj/project.pbxproj
@@ -34,6 +34,8 @@
 		02DA64172313C26400284168 /* StatsVersionBySite.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02DA64162313C26400284168 /* StatsVersionBySite.swift */; };
 		02DA64192313C2AA00284168 /* StatsVersion.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02DA64182313C2AA00284168 /* StatsVersion.swift */; };
 		02EAB6D72480A86D00FD873C /* CrashLogger.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02EAB6D62480A86D00FD873C /* CrashLogger.swift */; };
+		03DCB80F2629AFE700C8953D /* Coupon+CoreDataClass.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03DCB80D2629AFE700C8953D /* Coupon+CoreDataClass.swift */; };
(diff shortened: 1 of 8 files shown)


pmusolino (commented):
Thanks for opening this PR @joshheald-a8c 🙇 
It's not clear to me why the `Coupon` entity has a relationship with `Product` and `ProductCategory`.
I understood that we have some properties in the `Coupon` model that are related to products or to some product categories, but these are only IDs. I think that with this type of relationship, we will be unable to store a Coupon without fetching and storing also the related Products and Product Categories, which means that we should do a lot of networking requests just for fetching all the models that we need. Am I wrong, or have I misinterpreted everything?

joshheald-a8c:
Thanks @pmusolino. I wasn't totally sure of the implications of this but based on what I was seeing elsewhere in other models, similar arrays of IDs were represented as relationships to their respective models, so I thought the same thing would be right here as well.

If you think it'd be better to store a Transformable `[Int64]` here, I can do that instead. I definitely don't think we should fetch all the products and categories when we make a coupon model. It's likely that if we ever display this data it'll be in a detail screen somewhere, and it would be appropriate to fetch the required products there if needed, but we wouldn't want to fetch them on the list screen unless we had some pretty advanced filtering, or something like that. The inverse references from products and categories seem unlikely to be useful as well, and were only added to silence Xcode warnings.

I think it's best on reflection to go with `[Int64]` for these four fields, do you agree?

pmusolino:
> If you think it'd be better to store a Transformable [Int64] here, I can do that instead. I definitely don't think we should fetch all the products and categories when we make a coupon model. It's likely that if we ever display this data it'll be in a detail screen somewhere, and it would be appropriate to fetch the required products there if needed, but we wouldn't want to fetch them on the list screen unless we had some pretty advanced filtering, or something like that. The inverse references from products and categories seem unlikely to be useful as well, and were only added to silence Xcode warnings. 

Yes, I agree. For our specific case with coupon management, we don't want to fetch all the products or the product categories. If we will allow changing the products or the product categories in the future, we will create a flow that will fetch/display/return the appropriate products and categories based on the list of IDs. Otherwise, in the Yosemite layer, we should include a way to fetch also the products and the categories for every coupon. It could be a nightmare (also for the data consumption of our users).

pmusolino (approved):
Thanks @joshheald-a8c for making the changes! :shipit:

PR #3987, opened 2021-04-16
Coupon Management: Networking - Mapper for list of coupons

joshheald-a8c:
Part of #3912
Please ensure #3985 is merged before this PR

## Description
This PR adds the CouponListMapper, which is responsible for creating Coupon model objects from the JSON representation received from the API. Since `Coupon` is a root model, the mapper adds the siteID into all the responses we get from the API after parsing. As far as possible, JSON decoding is delegated to the synthesised `Codable` conformance of `Coupon`, by setting appropriate options on the `JSONDecoder` we use.

## Testing
The mapper is not in use yet, so no applicable testing other than the unit tests, and CI once it's merged to the main repo.

Update release notes:


Networking/NetworkingTests/Mapper/CouponListMapperTests.swift
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    /// Verifies that the `siteID` is added in the mapper, to all results, because it's not provided by the API endpoint
+    ///
+    func test_CouponsList_map_includes_siteID_in_parsed_results() throws {
+        let coupons = try mapLoadAllCouponsResponse()
+        guard coupons.count > 0 else {
+            XCTFail("No coupons parsed")
+            return
+        }
```
pmusolino: nit: a shorter and implicit version can be
```
XCTAssertTrue(coupons.count > 0)
```

Networking/NetworkingTests/Mapper/CouponListMapperTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        XCTAssertEqual(coupon.usageLimitPerUser, 3)
+        XCTAssertEqual(coupon.limitUsageToXItems, 10)
+        XCTAssertEqual(coupon.freeShipping, true)
+        XCTAssertEqual(coupon.productCategories, [123, 435, 232])
+        XCTAssertEqual(coupon.excludedProductCategories, [908])
+        XCTAssertEqual(coupon.excludeSaleItems, false)
+        XCTAssertEqual(coupon.minimumAmount, "5.00")
+        XCTAssertEqual(coupon.maximumAmount, "500.00")
+        XCTAssertEqual(coupon.emailRestrictions, ["*@a8c.com", "someone.else@example.com"])
+        XCTAssertEqual(coupon.usedBy, ["someone.else@example.com", "person@a8c.com"])
```
pmusolino: It would be better to create an expected entity and compare the two entities.
In this way, we do not risk creating a regression or missing a test if in the future we will add new fields to `Coupon`.

Networking/NetworkingTests/Mapper/CouponListMapperTests.swift
```diff
@@ -0,0 +1,122 @@
...
+        XCTAssertEqual(coupon.usageLimitPerUser, nil)
+        XCTAssertEqual(coupon.limitUsageToXItems, nil)
+        XCTAssertEqual(coupon.freeShipping, false)
+        XCTAssertEqual(coupon.productCategories, [])
+        XCTAssertEqual(coupon.excludedProductCategories, [])
+        XCTAssertEqual(coupon.excludeSaleItems, false)
+        XCTAssertEqual(coupon.minimumAmount, "0.00")
+        XCTAssertEqual(coupon.maximumAmount, "0.00")
+        XCTAssertEqual(coupon.emailRestrictions, [])
+        XCTAssertEqual(coupon.usedBy, [])
```
pmusolino: Like above, it would be better to compare two entities.

Networking/NetworkingTests/Mapper/CouponListMapperTests.swift
```diff
@@ -0,0 +1,122 @@
...
+    ///
+    func mapLoadAllCouponsResponse() throws -> [Coupon] {
+        return try mapCoupons(from: "coupons-all")
+    }
+
+    /// Returns the CouponsMapper output from `coupons-minimal.json`
+    ///
+    func mapLoadMinimalCouponsResponse() throws -> [Coupon] {
+        return try mapCoupons(from: "coupons-minimal")
+    }
```
pmusolino: Is there a reason for having two JSON?

pmusolino (commented):
Good job @joshheald-a8c.
I left just some suggestions and one question 👍

Networking/NetworkingTests/Mapper/CouponListMapperTests.swift
joshheald-a8c: Good point, will change these

Networking/NetworkingTests/Mapper/CouponListMapperTests.swift
joshheald-a8c: This was so we could separate test the correct handling of nulls, but I could certainly add the JSON object from `coupons-minimal.json` to `coupons-all.json`. Only downside of doing this is that were the optional test to start failing, so would earlier tests unrelated to handling nulls correctly. That's not such a big deal, so if we'd prefer to minimise the number of JSON files I'll just move it over 😊

pmusolino: I think that for this specific case you can move the coupon-minimal content as the third coupon in `coupons-all.json`

pmusolino (approved):
LGTM 👏 @joshheald-a8c

PR #3989, opened 2021-04-16
Coupon Management: Networking - Mapper for a single coupon

joshheald-a8c:
Part of issue #3912 
Please ensure #3988 is merged before this PR

## Description
This PR adds `CouponMapper` for a single coupon. Similarly to `CouponListMapper` it is responsible for creating a `Coupon` model object from the JSON representation received from the API. Since `Coupon` is a root model, the mapper adds the siteID into all the responses we get from the API after parsing. As far as possible, JSON decoding is delegated to the synthesised `Codable` conformance of `Coupon`, by setting appropriate options on the `JSONDecoder` we use.

The `JSONDecoder` required is common to any mapper that decodes a coupon, so it's moved out to an extension on `Coupon` to allow it to be shared.

## Testing
The mapper is not in use yet, so no applicable testing other than the unit tests, and CI once it's merged to the main repo.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #3990, opened 2021-04-16
Coupon Management: Networking - Add deleteCoupon endpoint method to Remote

joshheald-a8c:
Part of issue #3912 
Please ensure #3989 is merged before this PR

## Description
This PR adds the `deleteCoupon` endpoint method to `CouponsRemote`, which will required by the coupon list management screen to enable swipe to delete, and/or the coupon edit screen to include a delete button. This makes `DELETE` requests to `/coupons/{id}`, and returns the deleted coupon. 

The requests made by the endpoint always set `force=true` as a parameter, which is required as there is no bin functionality for Coupons. If this parameter is not set, the deletion will not do anything, and an error will be returned by the API.

## Testing
The endpoint is not in use yet, so no applicable testing other than the unit tests, and CI once it's merged to the main repo.

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/CouponsRemote.swift
+++ b/Networking/Networking/Remote/CouponsRemote.swift
@@ -32,6 +32,29 @@ public final class CouponsRemote: Remote {
 
         enqueue(request, mapper: mapper, completion: completion)
     }
+
+    // MARK: - Delete Coupon
+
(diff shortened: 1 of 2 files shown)


pmusolino (approved):
LGTM :shipit:

PR #3993, opened 2021-04-19
Shipping Labels M2: Show different address verification banner text and actions based on address type

### LINKED ISSUE #3918: Shipping Labels M2: Wrong banner text shown for address that can't be verified
**Describe the bug**

When an origin or destination address can't be verified while creating a shipping label, we display a banner at the top of the edit shipping address screen explaining the problem. However, the banner should have different text depending on whether the address is the origin address or the destination address. Right now, we're always showing the text meant for a destination address that can't be verified. (This includes text about contacting the customer about the address, which doesn't make sense for the origin address.)

**To Reproduce**

Before testing, make sure your WooCommerce store address is invalid (e.g. it uses the wrong postcode). You can edit this in wp-admin under WooCommerce > Settings > General (`/wp-admin/admin.php?page=wc-settings`).
In the app, go to the Orders > Processing tab.
Select an order.
Select "Create Shipping Label."
Select "Continue" in the "Ship from" section.
On the suggested address screen, select "Edit Address" to open the edit address screen. -> Notice the banner text doesn't match the design.

**Screenshots**

**Expected behavior**

When editing the origin ("Ship from") address, the banner text should read:

> We were unable to automatically verify the origin address. 

image

rachelmcr:
Fixes: #3918 

## Description

When an origin or destination address can't be verified while creating a shipping label, we display a banner at the top of the edit shipping address screen explaining the problem. Previously, we showed the same text and actions regardless of the address type (origin or destination).

This PR updates the banner so it shows different text and actions depending on whether it's for the origin or destination address.

## Changes

This updates `ShippingLabelAddressTopBannerFactory` to take `ShippingLabelAddressVerification.ShipType` (origin or destination) as a parameter and use it to determine what text and actions to display in the banner:

* The origin banner is a simple text banner (no actions) with relevant message.
* The destination includes text and actions prompting the user to view the address on Apple Maps or contact the customer to verify the address.

Origin|Destination
-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-04-19 at 11 32 39|Simulator Screen Shot - iPhone 11 Pro - 2021-04-19 at 11 40 30

## Testing

2. Open an order detail.
4. Press the button "Continue" in Create Shipping Label form, under the Ship From cell.
5. Go to the editing address screen and make sure the address is not valid remotely. -> You should see the top banner view with the message "We were unable to automatically verify the origin address." and no action buttons.
6. Fix the address and tap "Done" to return to the "Create Shipping Label" screen.
7. Press the button "Continue" under the Ship To cell.
8. Go to the editing address screen and make sure the address is not valid remotely. -> You should see the top banner view with the message "We were unable to automatically verify the shipping address. View on Apple Maps or try contacting the customer to make sure the address is correct." and two buttons "Open Map" and "Contact Customer".

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -18,7 +18,7 @@ final class ShippingLabelAddressFormViewController: UIViewController {
     /// Top banner that shows a warning in case there is an error in the address validation.
     ///
     private lazy var topBannerView: TopBannerView = {
-        let topBanner = ShippingLabelAddressTopBannerFactory.addressErrorTopBannerView { [weak self] in
+        let topBanner = ShippingLabelAddressTopBannerFactory.addressErrorTopBannerView(shipType: viewModel.type) { [weak self] in
             MapsHelper.openAppleMaps(address: self?.viewModel.address?.formattedPostalAddress) { [weak self] (result) in
                 switch result {
                 case .success:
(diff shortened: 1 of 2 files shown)


Ecarrion (approved):
👍 🚀

PR #3994, opened 2021-04-19
Remove option to add more than one image for variations

### LINKED ISSUE #3932: Variations: Remove add new image icon
Variable products can only have one image, but the UX makes it look like we could add multiple images by providing an "Add new image" icon next to the current image.

Task:

- [ ] Remove the "Add new image" on the variable product form once an image has been added

ios-delete-image-icon

koke:
Closes #3932

## Description

Since variable products only allow at most one image, this prevents the button to add more images to a variation when there is one already.

## Testing

- A product variation with no image should show the option to add a product image
- A product variation with an image should show the option to replace/remove the image, but not add a new one
- A regular product should still allow adding more than one product image

Update release notes:


pmusolino (approved):
LGTM, and congrats @koke for opening your first PR on WCiOS! 🥳

Just one thing: is it normal that in the main product that contains all the variation, the button for adding multiple images is still visible?

koke:
> is it normal that in the main product that contains all the variation, the button for adding multiple images is still visible?

I think so. A product with variations can have multiple images:

Screen Shot 2021-04-20 at 12 18 34

Then it seems the store will show all the available images but will switch to the one specific to the variation when you select it:

variations

PR #3996, opened 2021-04-19
Shipping Labels: Storage layer for account settings endpoint

rachelmcr:
Part of: #3940

## Description

This PR adds support in the Storage layer for the Shipping Label account settings endpoint (`/wc/v1/connect/account/settings`). We currently need this endpoint to fetch the last selected package for shipping labels, and it will also be used in the future for handling payment methods.

Support for the Networking layer was added in https://github.com/woocommerce/woocommerce-ios/pull/3983, and Yosemite changes will come in a future PR.

## Changes

* Adds Core Data Model 48 with new entities `ShippingLabelAccountSettings` and `ShippingLabelPaymentMethod`.
* Adds generated `ShippingLabelAccountSettings+CoreDataProperties` and `ShippingLabelPaymentMethod+CoreDataProperties` subclasses.
* Adds a `StorageType` extension method for loading shipping label account settings, and associated unit test.

## Testing

Confirm tests pass in CI.
This is the first time I'm adding new entities to the storage layer, so I also appreciate a close look at the code changes!

## Submitter Checklist

Update release notes:

### DIFF
--- a/Storage/Storage.xcodeproj/project.pbxproj
+++ b/Storage/Storage.xcodeproj/project.pbxproj
@@ -141,6 +141,10 @@
 		B5DC3CAA20D1B2FE0063AC41 /* WooCommerce.xcdatamodeld in Resources */ = {isa = PBXBuildFile; fileRef = B59E11D820A9D00C004121A4 /* WooCommerce.xcdatamodeld */; };
 		B5FD111E21D4046E00560344 /* OrderSearchResults+CoreDataProperties.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5FD111C21D4046E00560344 /* OrderSearchResults+CoreDataProperties.swift */; };
 		B5FD111F21D4046E00560344 /* OrderSearchResults+CoreDataClass.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5FD111D21D4046E00560344 /* OrderSearchResults+CoreDataClass.swift */; };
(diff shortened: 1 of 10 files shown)


pmusolino (commented):
Thanks for working on this @rachelmcr! 
I noticed that in the two models just a few properties are optional, but in the storage model, you made all the properties optional. I think that it is probably better to reflect the networking model unless there is a specific reason to keep everything optional. What do you think?

rachelmcr:
> I noticed that in the two models just a few properties are optional, but in the storage model, you made all the properties optional. I think that it is probably better to reflect the networking model unless there is a specific reason to keep everything optional. What do you think?

Ah, that was an oversight; thanks for catching it! I'll fix that.

rachelmcr:
@pmusolino Ready for another look! I noticed that the generated NSManagedObject subclasses still have some of the properties as optional, although the model itself doesn't. From what I've read that's expected but let me know if we usually handle those differently.

pmusolino:
@rachelmcr do you regenerated them from scratch? In the previous model (like `Product` I didn't see any optional property if it's not already declared as optional)

rachelmcr:
@pmusolino Yes, I regenerated them under Editor > Create NSManagedObject subclass, and used the contents of those newly generated files. For `Product`, I noticed that in the past it has been regenerated with more optional properties (like in 2241f1c) but those changes were later reverted and the file was updated manually after that, retaining those non-optional properties.

Should I go ahead and update these files manually to remove the optionals for the properties that aren't optional in the model entity? Or is it better to just keep the automatically generated file to make it easier to update in the future?

pmusolino:
After some research, it seems that ["Optional" means something different in Core Data](https://stackoverflow.com/a/33552046/7241994) than it does to Swift (TIL).
In my opinion, it makes sense to maintain them optional because we don't want to maintain manually these classes if some properties will change in the future.

pmusolino (approved):
:shipit:  LGTM!

PR #3997, opened 2021-04-19
[Mobile Payments] Add model objects with metadata necessary to print receipts

### LINKED ISSUE #3936: [Mobile Payments] [Receipts] Update PaymentIntent to add fields needed for receipts
Update PaymentIntent, adding the fields necessary for printing receipts compliant

ctarda:
Closes #3936 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by adding the fields needed to print a receipt.

Receipts can be customized, as long as we list some required information. When accepting in-person payments with EMV chip cards, card networks require us to include several fields on the receipts we provide to customers. Those fields are inluded as part of the PaymentIntent object (to be more specific, the CardReaderDetails object), once the payment is processed. For more information, see Stripe's documentation: [Custom receipts](https://stripe.com/docs/terminal/checkout/receipts#custom)

So this PR added the necessary fields to `Charge`, so that [Charges](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPCharge.html) associated to a [Payment Intent](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntent.html) contain the information needed to print a receipt. Sorry it went over 500 lines, but in between docs, changes to the project files, headerdocs and tests, it has ballooned a bit, and it was not pragmatic to break it down further.

## Changes
The PR adds the following model objects:
* PaymentMethod. An enumeration that models the payment method associated with the charge. This enumeration can have an associated value for one of its cases:
* CardPresentDetails. The value associated to the case .cardPresent of PaymentMethod. This objects represents the details from a transaction using a card_present payment method. Among other properties, it contains:
* ReceiptDetails. Receipt details associated with a card present transaction. This is the data that we actually need to render into the receipt. 
* CardBrand. An enumeration to model the type of card. 
* Extension to all the mentioned model objects, to provide initializers that map between the model objects provided by Stripe (which we do not want to expose to the rest of the codebase) an our domain objects.
* Some unit tests to cover those mappings, when possible.

Note how there is not a 1:1 correspondence between our model objects and Stripe's. Some objects the Stripe SDK make sense in a context where said objects need to be consumed from Obj-C code (e.g. SCPaymentMethod and SCPaymentMethodType), while for us, being all Swift, there are better ways to model that data (e.g. hardware.PaymentMethod)

## How to test

* Checkout the branch, run `bundle exec pod install`, just for kicks and giggles.
* **Add break point to line 113 of CardPresentPaymentStore** :

Screen Shot 2021-04-19 at 13 38 24

* Build and run on a device. 
* Tap "Connect card reader"
* Tap "Connect to reader"
* Navigate to an order eligible for Card Present Payments (it has to be a store enrolled to WCPay, and the order needs to have as payment method associated  Cash on Delivery or none).
* If everything goes well, the payment should be processed and the breakpoint added to CardPresentPaymentStore should be triggered.
* `po intent`

The resulting instance of PaymentIntent returned by the card reader service after the payment has been processed should look like this:

```
▿ PaymentIntent
  - id : "pi_REDACTED"
  - status : Hardware.PaymentIntentStatus.requiresCapture
  ▿ created : 2021-04-19 17:38:05 +0000
    - timeIntervalSinceReferenceDate : 640546685.0
  - amount : 21000
  - currency : "usd"
  ▿ metadata : Optional<Dictionary<AnyHashable, Any>>
    - some : 0 elements
  ▿ charges : 1 element
    ▿ 0 : Charge
      - id : "ch_REDACTED"
      - amount : 21000
      - currency : "usd"
      - status : Hardware.ChargeStatus.succeeded
      ▿ description : Optional<String>
        - some : "Receipt description."
      ▿ metadata : Optional<Dictionary<AnyHashable, Any>>
        - some : 0 elements
      ▿ paymentMethod : Optional<PaymentMethod>
        ▿ some : PaymentMethod
          ▿ presentCard : 1 element
            ▿ details : CardPresentDetails
              - last4 : "9999"
              - expMonth : 12
              - expYear : 2021
              ▿ cardholderName : Optional<String>
                - some : "TEST/STRIPE"
              - brand : Hardware.CardBrand.visa
              - fingerprint : "REDACTED"
              ▿ generatedCard : Optional<String>
                - some : "pm_REDACTED"
              ▿ receipt : Optional<ReceiptDetails>
                ▿ some : ReceiptDetails
                  - applicationPreferredName : "Stripe Credit"
                  - dedicatedFileName : "A000000003101001"
                  - authorizationResponseCode : "3030"
                  - applicationCryptogram : "36E4873E616952E7"
                  - terminalVerificationResults : "0000000000"
                  - transactionStatusInformation : "0000"
                  ▿ accountType : Optional<String>
                    - some : "credit"
              ▿ emvAuthData : Optional<String>
                - some : "8A023030"
```

Notice the object tree structure: PaymentIntent > Charge > PaymentMethod > CardPresentDetails > ReceiptDetails

* Also, CI should be ✅ 

Update release notes:


Hardware/Hardware/CardReader/CardPresentDetails.swift
```diff
@@ -0,0 +1,31 @@
+/// An object representing details from a transaction using a card_present payment method.
+public struct CardPresentDetails {
+    /// The last 4 digits of the card.
+    public let last4: String
+
+    /// The card’s expiration month. 1-indexed (i.e. 1 == January)
+    public let expMonth: Int
+
+    /// The card’s expiration year.
+    public let expYear: Int
```
allendav, line 10: I'm old. Please explicitly include "4 digits" in the comment :P

Hardware/Hardware/CardReader/StripeCardReader/CardPresentDetails+Stripe.swift
```diff
@@ -0,0 +1,39 @@
...
+        self.brand = CardBrand(brand: details.brand)
+        self.fingerprint = details.fingerprint
+        self.generatedCard = details.generatedCard
+        self.receipt = Hardware.ReceiptDetails(receiptDetails: details.receipt)
+        self.emvAuthData = details.emvAuthData
+    }
+}
+
+
+/// The initializers of StripeTerminal.CardPresentDetails are annotated as NS_UNAVAILABLE
```
allendav, line 21: What would you think about asking Stripe to change this in a future version of the SDK?

docs/HARDWARE.md
```diff
@@ -110,4 +113,8 @@ public enum CardReaderServiceError: Error {
 
 `UnderlyingError` is another enumeration that abstracts out the [error codes provided by the Stripe Terminal SDK](https://stripe.dev/stripe-terminal-ios/docs/Enums/SCPError.html), mapping them to domain errors. 
 
-This way, clients of the service can switch on the CardReaderServiceError in order to understand what part of the process wehn wrong, and then extract the underlying error to understand why the operation failed.
\ No newline at end of file
+This way, clients of the service can switch on the CardReaderServiceError in order to understand what part of the process wehn wrong, and then extract the underlying error to understand why the operation failed.
```
allendav: wehn when

docs/HARDWARE.md
```diff
@@ -110,4 +113,8 @@ public enum CardReaderServiceError: Error {
...
+
+### Printing receipts.
+
+Receipts can be customized, as long as we list some required information. When accepting in-person payments with EMV chip cards, card networks require us to include several fields on the receipts we provide to customers. Those fields are inluded as part of the PaymentIntent object (to be more specific, the CardReaderDetails object), once the payment is processed. For more information, see Stripe's documentation: [Custom receipts](https://stripe.com/docs/terminal/checkout/receipts#custom)
```
allendav: inluded included

allendav (approved):
Tests fine. I was able to set the breakpoint and see my payment intent. Exciting to see another step forward :)
All CardPresentDetailsTests also passed
A handful of comments for you to consider.

Hardware/Hardware/CardReader/StripeCardReader/CardPresentDetails+Stripe.swift
ctarda: I've taught about it... I think it makes sense, from their point of view, to seal these objects, specially CardReader, so that we can not instantiate our own and pass them back to the SDK. If I try to put myself in their shoes, as a developer of an SDK that integrates with something that takes people's money, I would probably seal my model objects as much as I could. That definitely makes things more difficult for us, in order to unit test things, but it is also fair to admin that we might be trying to do something (unit test everything) that the SDK is not meant for. 

So... I don't know.

ctarda:
Thanks for the review @allendav  🙇🏼  I just pushed an update to incorporate your feedback, I will merge as soon as CI goes  ✅

PR #3998, opened 2021-04-19
[Mobile Payments] Add networking for capturing WCPay payment intents (and updating order status)

### LINKED ISSUE #3970: [Mobile Payments] [Closing the loop] Capture Payment Intent. Mark I: Networking
For context: p91TBi-4VZ-p2

Implement integration with `wc/v3/payments/orders/{orderID}/capture` to submit the payment intent ID captured using the Stripe Terminal SDK

See also #3971

allendav:
Closes #3970 

Changes:
- Adds a new networking request for `captureOrderPayment` to `WCPayRemote`
- `captureOrderPayment` will send a request through Jetpack to the self-hosted site to instruct the site to capture the payment intent created by the Stripe Terminal SDK and update the order to reflect payment completion
- Note that the `payments/orders/\(sampleOrderID)/capture` endpoint does not yet exist in the WooCommerce Payments plugin - see https://github.com/Automattic/woocommerce-payments/issues/1622

To Test:
- Since the endpoint doesn't exist yet, the appropriate test is to ensure all the new tests added to WCPayRemoteTests pass 

Update release notes:


Networking/Networking/Remote/WCPayRemote.swift
```diff
@@ -35,6 +35,30 @@ public class WCPayRemote: Remote {
...
+    /// - Parameters:
+    ///   - siteID: Site for which we'll capture the payment.
+    ///   - orderID: Order for which we are capturing the payment.
+    ///   - paymentIntentID: Stripe Payment Intent ID created using the Terminal SDK.
+    ///   - completion: Closure to be run on completion.
+    public func captureOrderPayment(for siteID: Int64,
+                               orderID: Int64,
+                               paymentIntentID: String,
+                               completion: @escaping (Result<WCPayPaymentIntent, Error>) -> Void) {
+        let path = "payments/orders/" + String(orderID) + "/capture"
```
ctarda: We tend to declare the paths in a private enum. Also, I would say that the preferred way to compose the path would be by interpolating strings, so in the end, this line could look like:

let path = "\(Paths.ordersPath)/\(orderID)/\(Paths.capture)"

ctarda (approved):
Nicely done. One comment, and it's definitely not a blocker.

:shipit:

Networking/Networking/Remote/WCPayRemote.swift
allendav: Can do

allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/3998/commits/5680ccc1511a204bfaf30acc8364f44b8ab3c2ff

PR #3999, opened 2021-04-20
Merge 6.5 code freeze into develop

jkmassel:
Merges the `6.5` code freeze into `develop`, including version number changes and translation updates


mokagio (commented):
:shipit:

mokagio (approved):

fastlane/Fastfile
```diff
@@ -74,9 +74,9 @@ GH_ORG_NAME="woocommerce"
     ios_bump_version_release()
     new_version = ios_get_app_version()
     ios_update_release_notes(new_version: new_version)
+    get_prs_list(repository:GHHELPER_REPO, tag:"#{new_version}", report_path:"#{File.expand_path('~')}/wcios_prs_list_#{old_version}_#{new_version}.txt")
```
AliSoftware, line 77: @jkmassel I'm a bit puzzled with this change, since the new action expects `milestone`, not `tag`, for the new parameter (see [change in #227](https://github.com/wordpress-mobile/release-toolkit/pull/227/files#diff-c3c92f31a3aed155aab1aba9315c428344a5079f853257ac495f27c81c12b317R52-R53)) 🤔  How did that even work then if you used the wrong param name?

AliSoftware: Update after chatting about it in DM, what likely happened is:
 - The lane failed when Jeremy ran it the first time, because it was still using the old call site and parameters for `get_prs_list` (the one from the now deleted line 79 in that PR)
 - So instead of reverting all the side effects from the steps of that lane that _did_ run  (bump version, update release notes, branch proection, frozen tag), fix it and start all over, which would have been quite a lot to revert just for that fix, Jeremy decided it was not worth the trouble and just ran the `get_prs_list` action manually on the side, then updated the `Fastfile` like seen in that PR's diff to reflect what he did manually, but without re-running the lane with those `Fastfile` changes he made
 - This is why those changes were not actually tested/run at that time, and why the typo in the new parameter names (`tag:` instead of `milestone:`) was not spotted.

All this means that this will need to be updated again before next code freeze. (@JavonDavis will try to take care of this later today)

Issue #3894, opened 2021-03-30
Add-Ons: Add outdated notice.


Ecarrion:
Task

- [x] Add beta feature banner view
- [x] Add disclaimer below list content

<img width="641" alt="Screen Shot 2021-04-22 at 7 35 13 AM" src="https://user-images.githubusercontent.com/562080/115715103-72d7a600-a33d-11eb-8201-f113b76feb6b.png">

PR #4001, opened 2021-04-20
Add-Ons: Add update disclaimer to order add-ons

Ecarrion:
part of #3894

# Why

As our add-ons screen won't have edit capabilities right now, this PR adds a small disclaimer informing the merchant that they can do any edit on the web dashboard.

# How

- Added a new `footnoteStyle()` modifier
- Updated VM to provide the "edit disclaimer"
- Updated view to render the disclaimer: Creates a new `OrderAddOnNoticeView` which is added in `OrderAddOnListI1View`

Note: There is a small workaround where texts inside a `ScrollView` do not have the ideal height. 
Adding the `fixedSize()` modifier, seems to fix the issue. Ref: paNNhX-d2-p2

# Screenshot

# Testing Steps
**Prerequisites:** Have your site with the add-ons plugin installed and configured.

- Make an order with an item with add-ons
- Launch the app and navigate to that order
- See that a new screen is presented with the add-on's information with the info disclaimer.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
+++ b/WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
@@ -30,6 +30,14 @@ struct HeadlineStyle: ViewModifier {
     }
 }
 
+struct FootnoteStyle: ViewModifier {
+    func body(content: Content) -> some View {
+        content
+            .font(.footnote)
+            .foregroundColor(Color(.textSubtle))
+    }
+}
+
 // MARK: View extensions
 extension View {
     func bodyStyle() -> some View {
@@ -42,5 +50,9 @@ extension View {
 
     func headlineStyle() -> some View {
         self.modifier(HeadlineStyle())
+    }
+
+    func footnoteStyle() -> some View {
+        self.modifier(FootnoteStyle())
     }
 }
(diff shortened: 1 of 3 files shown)


koke (approved):
I'm wondering if the disclaimer should be tappable and take you to the dashboard, but this looks good to me.

The `fixedSize` issue seems annoying, but I guess SwiftUI is still in its early days and that's the price to pay 😅 I wonder if there's a way to turn that into a more specific modifier that includes a more extensive explanation of why that is needed. And if that ever gets fixed in SwiftUI, it would be easier to search for all the instances and remove the workaround.

Ecarrion:
@koke 

> I'm wondering if the disclaimer should be tappable and take you to the dashboard, but this looks good to me.

Good point! @adamzelinski What do you think about it?

> The fixedSize issue seems annoying, but I guess SwiftUI is still in its early days and that's the price to pay 😅 I wonder if there's a way to turn that into a more specific modifier that includes a more extensive explanation of why that is needed. And if that ever gets fixed in SwiftUI, it would be easier to search for all the instances and remove the workaround.

This is a very good point, [let me explore a little bit more](https://github.com/woocommerce/woocommerce-ios/projects/35#card-59442185) and see how often places this is needed before I turn this into a public modifier.

Issue #3982, opened 2021-04-16
[Mobile Payments] [Saved Receipts] Mark II 1/2: Render a receipt


ctarda:
For context: p91TBi-4KR-p2

Receipts need to be rendered to screen and to a printer. 

We want to find a way to render to both destinations without duplicating the logic that renders the receipt

Issue #3981, opened 2021-04-16
[Mobile Payments] [Saved Receipts] Print Receipts. Mark III: Receipt details view


ctarda:
For context: p91TBi-4KR-p2

A view / view controller that renders a receipt and offers a print button.

This is where a pre-existing receipt can be inspected

Issue #3979, opened 2021-04-16
[Mobile Payments] [Receipts] Print Receipts. Mark II: Yosemite


ctarda:
For context: p91TBi-4KR-p2

Business logic to initiate a printout of a receipt associated to an order.

Issue #3978, opened 2021-04-16
[Mobile Payments] [Receipts] Print Receipts. Mark IV: Iterate on layout of printed receipt


ctarda:
For context: p91TBi-4KR-p2

Improve the layout of the receipt

As noted here: https://github.com/woocommerce/woocommerce-ios/pull/4002#discussion_r617950620

We should let go of the concept of "pages:

PR #4002, opened 2021-04-20
[Mobile Payments] Printing receipts. Mark I: Hardware

### LINKED ISSUE #3977: [Mobile Payments] [Receipts] Print receipts. Mark I: Hardware
For context: p91TBi-4KR-p2

Declare interface of the service that will use to print receipts. 

Provide a barebones implementation that just dumps the receipt data to AirPrint

ctarda:
Closes #3977 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by adding the scaffolding for printing receipts to AirPrint.

This PR adds a new interface to Hardware, and a barebones implementation of that interface that integrates with AirPrint. It also adds a couple of new model objects to wrap all the information we need to add to the receipt.

For more context about what needs to go into a receipt, please check [Stripe's documentation](https://stripe.com/docs/terminal/checkout/receipts#custom).

https://user-images.githubusercontent.com/2722505/115439566-b666ce80-a1dc-11eb-9008-5d1dcf6124b6.mov

## Changes
* Add ReceiptPrinterService. Expose this service via the ServiceLocator.
* Add an implementation of ReceiptPrinterService that prints using AirPrint. This class will be iterated on while working on issues #3982 and #3978
* Add a barebones implementation of a subclass of UIPrintPageRenderer that will be responsible for actually render the content of the receipt. To be fully implemented in #3978 
* Add a couple of model objects to wrap the information we need to pass to the ReceiptPrinterService.

## How to test

1. Requires setting up a store for Card Present Payment Testing (also P91TBi-4BH-p2 for more specific instructions)

* Build and run on a device.
* Tap "Connect card reader"
* Tap "Connect to reader"
* At that point, and only in order to test this PR, a preview of the receipt should be presented. At this point, the preview only shows a page number and the actual amount and currency of the order that was just charged using the card reader. The amount and currency should match the Order's.

Note: this is not the actual flow that will be implemented in the final beta. There is work pending in tickets #3979, #3982, #3981 #3978

Update release notes:


Hardware/Hardware/Printer/AirPrintReceipt/AirPrintReceiptPrinterService.swift
```diff
@@ -0,0 +1,28 @@
+import UIKit
+
+/// Barebones Implementation of the ReceiptPrinterService that integrates with AirPrint
+/// Will be iterated in https://github.com/woocommerce/woocommerce-ios/issues/3982
```
ctarda, line 4: This exists, at this point, pretty much so that we can actually test that the receipt is being generated. The proper implementation will come in a later PR

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -0,0 +1,53 @@
+import UIKit
+
+/// Renders a receipt in an AirPrint enabled printer.
+/// To be properly implemented in https://github.com/woocommerce/woocommerce-ios/issues/3978
+final class ReceiptRenderer: UIPrintPageRenderer {
```
ctarda, line 5: This also needs to be iterated on.

Hardware/Hardware/Printer/ReceiptContent.swift
```diff
@@ -0,0 +1,11 @@
+/// Models the content of the receipt.
+/// To be fully implemented in https://github.com/woocommerce/woocommerce-ios/issues/3978
+public struct ReceiptContent {
+    public let lineItems: [ReceiptLineItem]
+    public let paymentIntent: PaymentIntent
```
ctarda, line 5: While working on #3979 I've realized everything would be easier if we don't pass the full PaymentIntent here, and we pass a different type that contains only what we need to render the receipt. 

That way we don't need to persist PaymentIntent to Storage. 

So this will change in #3979

ctarda:
Unit tests pass locally 🤷🏼

ctarda:
And now tests pass on CI again 🤷🏼

Hardware/Hardware/Printer/AirPrintReceipt/AirPrintReceiptPrinterService.swift
allendav: Do we need a ReceiptPrinterService or just a PrinterService - and could that be shared with shipping label printing?

We could define a renderer protocol, implement it for receipts (and labels, and whatever in the future like inventory bar codes) and pass the renderer in as a dependency to a generic PrinterService

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -0,0 +1,53 @@
...
+    private let lines: [ReceiptLineItem]
+    private let paymentIntent: PaymentIntent
+
+    private let attributes: [NSAttributedString.Key: Any] = [.font: UIFont(name: "HelveticaNeue", size: 48) as Any]
+
+    init(content: ReceiptContent) {
+        self.lines = content.lineItems
+        self.paymentIntent = content.paymentIntent
```
allendav, line 13: I'm not sure receipts should know about paymentIntents. Eventually we will need to support receipts for other means of payment.

Have you considered passing structured data instead to the renderer? i.e. arrays of key:values that should be output in each section?

allendav: Nevermind. I just saw your related comment below that this will change.

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -0,0 +1,53 @@
...
+
+        super.init()
+
+        configureHeaderAndFooter()
+
+        configureFormatter()
+    }
+
+    override func drawHeaderForPage(at pageIndex: Int, in headerRect: CGRect) {
+        let pageNumberString = NSString(string: "Order receipt. Page \(pageIndex + 1)")
```
allendav, line 23: Needs localization.

Continuous roll receipt printers won't have page numbers. Do we need to open an issue to handle formatting for continuous roll vs sheet paper receipts?

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -0,0 +1,53 @@
...
+        pageNumberString.draw(in: headerRect, withAttributes: attributes)
+    }
+
+    override func drawContentForPage(at pageIndex: Int, in contentRect: CGRect) {
+        let printOut = NSString(string: "Total charged: \(paymentIntent.amount / 100) \(paymentIntent.currency.uppercased())")
```
allendav, line 28: Final receipts should probably show itemized order list, taxes, etc.?

Hardware/Hardware/Printer/ReceiptContent.swift
allendav: I don't see the required fields for EMV. I assume those come in a later PR?

allendav (approved):
Tests fine. Several suggestions (some of which it looks like you've already considered yourself) below for later PRs to consider. :shipit:

Hardware/Hardware/Printer/ReceiptContent.swift
ctarda: Those are declared as properties of PaymentIntent's Charge, modelling what Stripe does. In #4010 this model object flattens a bit, and the EMV fields are a property of ReceptDetails

Hardware/Hardware/Printer/AirPrintReceipt/AirPrintReceiptPrinterService.swift
ctarda: I also considered that, but I thought it would be better to not reach for more than what we need for the beta at this point. We can always rename this service later if we add support for Shipping Labels.

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
ctarda: Yep. It's changed in #4010

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
ctarda: Good point! I added this just to print something for now, to visualize that it actually works. I added a note to the ticket we have for iterating on the renderer (#3978), so that we remove page numbers.

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
ctarda: Yep, I would expect so. That was the idea behind passing an array of "LineItem", so that we can provide that information.

Hardware/Hardware/Printer/AirPrintReceipt/AirPrintReceiptPrinterService.swift
ctarda: That being said I'll rename to PrinterService right away

PR #4003, opened 2021-04-20
Prevent crash when the media library has exactly the same amount of items as the request page size

### LINKED ISSUE #3369: WordPressMediaLibraryPickerDataSource.retrieveMedia upperBound EXC_BREAKPOINT
Sentry Issue: [WOOCOMMERCE-IOS-1CF0](https://sentry.io/organizations/a8c/issues/2053522094/?referrer=github_integration)

```
EXC_BREAKPOINT: Exception 6, Code 4500452, Subcode 8
  File "WordPressMediaLibraryPickerDataSource.swift", in upperBound
  File "WordPressMediaLibraryPickerDataSource.swift", line 193, in WordPressMediaLibraryPickerDataSource.updateMediaItems
  File "WordPressMediaLibraryPickerDataSource.swift", line 158, in WordPressMediaLibraryPickerDataSource.sync
  File "<compiler-generated>", in WordPressMediaLibraryPickerDataSource.sync
  File "WordPressMediaLibraryPickerDataSource.swift", line 169, in WordPressMediaLibraryPickerDataSource.retrieveMedia
...
(21 additional frame(s) were not displayed)
```

Create by @shiki

koke:
Fixes #3369

## Description

When the media library contains exactly the amount of items defined as the page size (currently 25) the app would crash. In theory, this would also happen with any exact multiple of that page size.

The problem is that when we have a full page and request the next one, the response will have no items. The previous code tried to generate a range of indices to update that was invalid because the end was lower than the start. Instead, if there is no new data to update, we can bail early and skip the callback.

## Testing

1. Ensure your media library has exactly 25 images
2. Create or tap into a product, then tap Add a product image
3. Tap Add Photos > WP Media Library
4. The app shouldn't crash

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 6.6
 -----
 - [*] Fix: Product variations only support at most one image, so we won't show an option to add a second one. [https://github.com/woocommerce/woocommerce-ios/pull/3994]
+- [*] Fix: The screen to select images from the Media Library would sometimes crash when the library had a specific number of images. [https://github.com/woocommerce/woocommerce-ios/pull/4003]
 
 6.5
 -----
--- a/WooCommerce/Classes/ViewRelated/Products/Media/WordPressMediaLibraryPickerDataSource.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Media/WordPressMediaLibraryPickerDataSource.swift
@@ -177,6 +177,12 @@ private extension WordPressMediaLibraryPickerDataSource {
     /// Appends the new media items to the existing media items.
     ///
     func updateMediaItems(_ newMediaItems: [Media], pageNumber: Int, pageSize: Int) {
+        // If the response contains no new items, there is nothing to update.
+        // We return early since the code would generate an invalid range of indices to update otherwise.
+        guard newMediaItems.isNotEmpty else {
+            return
+        }
+
         let pageFirstIndex = syncingCoordinator.pageFirstIndex
         let mediaStartIndex = (pageNumber - pageFirstIndex) * pageSize
         let mediaStartIndexOfTheNextPage = (pageNumber + 1 - pageFirstIndex) * pageSize


Ecarrion (approved):
Nice catch! 🚀

designsimply:
@koke @AliSoftware can we move this into the 6.5 release because it came up in beta testing and we talked about it as something that might affect new users disproportionately and cause extra frustration for them even if it may not affect a HUGE number of users overall. If it's too late in the code freeze or you feel strongly 6.6 would be better, please let me know! 

<sup>(internal reference: p5T066-2ci-p2#comment-8053)</sup>

AliSoftware:
I'm ok to include it in 6.5 if you want to; but if we do, be aware that we'll probably not make an intermediate new beta with that fix before the final version — since I'll submit the final version to Apple as soon as tomorrow. So it would be included in the final 6.5 without having gone thru beta testers. Are you good with that @designsimply ?

---

[EDIT] Though I just realized that maybe exceptionally we can send the final build that I'll submit to Apple tomorrow… also to beta testers, so they get it around their Saturday morning and have the week-end to test it and raise any last minute issue if any, allowing us to cancel the release before Monday in case an emergency/blocker is raised in the meantime. Not sure it's worth the trouble though, as opposed to just include it into final 6.5 without having this fix go thru beta testers?

koke:
I have a PR ready at https://github.com/woocommerce/woocommerce-ios/pull/4070 back porting this to 6.5 if you want to do it.

designsimply:
@AliSoftware apologies for not seeing this sooner. As a retroactive reply, yes I'd be ok to have shipped it in 6.5 but if it didn't make it in then I don't think it's worth a hotfix but rather we'll test it as part of 6.6 beta testing.

AliSoftware:
> @AliSoftware apologies for not seeing this sooner. As a retroactive reply, yes I'd be ok to have shipped it in 6.5 but if it didn't make it in then I don't think it's worth a hotfix but rather we'll test it as part of 6.6 beta testing.

@designsimply it _did_ make it to 6.5 — see @koke 's PR https://github.com/woocommerce/woocommerce-ios/pull/4070 above which was indeed merged (on the day before I did the final submission to Apple, so it indeed made it) 🙂

PR #4004, opened 2021-04-20
[Mobile Payments] Capture Payment Intent

### LINKED ISSUE #3971: [Mobile Payments] [Closing the loop] Capture Payment Intent. Mark II: Yosemite
For context: p91TBi-4VZ-p2

Related issues: #3970

allendav:
Closes #3971 

This PR leverages the remote added for #3970 to capture the payment intent created by the Stripe Terminal SDK.

⚠️ As the WCPay endpoint does not actually exist yet, this capture will, for now, always fail ⚠️

## How to Test

- Ensure all automated tests continue to pass

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
+++ b/Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
@@ -110,6 +110,25 @@ private extension CardPresentPaymentStore {
             // https://stripe.dev/stripe-terminal-ios/docs/Enums/SCPPaymentIntentStatus.html#/c:@E@SCPPaymentIntentStatus@SCPPaymentIntentStatusRequiresCapture
             // TODO. Persist PaymentIntent, so that we can use it later to print a receipt
             // Deferred to https://github.com/woocommerce/woocommerce-ios/issues/3825
+
+            // Note: Here we transition from the Stripe Terminal Payment Intent
+            // to the WCPay backend managed Payment Intent, which we need
+            // to use to capture the payment.
+
+            self.remote.captureOrderPayment(for: siteID, orderID: orderID, paymentIntentID: intent.id) { result in
+                switch result {
+                case .success(let intent):
+                    if intent.status == .succeeded {
+                        onCompletion(.success(()))
+                    } else {
+                        DDLogDebug("Unexpected payment intent status \(intent.status) after attempting capture")
+                        onCompletion(.failure(CardReaderServiceError.paymentCapture()))
+                    }
+                case .failure(let error):
+                    onCompletion(.failure(error))
+                }
+            }
+
             onCompletion(.success(()))
         }.store(in: &cancellables)
 
--- a/Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
+++ b/Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
@@ -118,18 +118,17 @@ private extension CardPresentPaymentStore {
             self.remote.captureOrderPayment(for: siteID, orderID: orderID, paymentIntentID: intent.id) { result in
                 switch result {
                 case .success(let intent):
-                    if intent.status == .succeeded {
-                        onCompletion(.success(()))
-                    } else {
+                    guard intent.status == .succeeded else {
                         DDLogDebug("Unexpected payment intent status \(intent.status) after attempting capture")
                         onCompletion(.failure(CardReaderServiceError.paymentCapture()))
+                        return
                     }
+                    onCompletion(.success(()))
                 case .failure(let error):
                     onCompletion(.failure(error))
+                    return
                 }
             }
-
-            onCompletion(.success(()))
         }.store(in: &cancellables)
 
         // Observe status events fired by the card reader


Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -110,6 +110,25 @@ private extension CardPresentPaymentStore {
...
+            // to use to capture the payment.
+
+            self.remote.captureOrderPayment(for: siteID, orderID: orderID, paymentIntentID: intent.id) { result in
+                switch result {
+                case .success(let intent):
+                    if intent.status == .succeeded {
+                        onCompletion(.success(()))
+                    } else {
+                        DDLogDebug("Unexpected payment intent status \(intent.status) after attempting capture")
+                        onCompletion(.failure(CardReaderServiceError.paymentCapture()))
```
allendav, line 123: @ctarda - or should CardPresentPaymentStore have its own Error enumeration here?

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -110,6 +110,25 @@ private extension CardPresentPaymentStore {
             // https://stripe.dev/stripe-terminal-ios/docs/Enums/SCPPaymentIntentStatus.html#/c:@E@SCPPaymentIntentStatus@SCPPaymentIntentStatusRequiresCapture
             // TODO. Persist PaymentIntent, so that we can use it later to print a receipt
             // Deferred to https://github.com/woocommerce/woocommerce-ios/issues/3825
+
+            // Note: Here we transition from the Stripe Terminal Payment Intent
+            // to the WCPay backend managed Payment Intent, which we need
+            // to use to capture the payment.
+
+            self.remote.captureOrderPayment(for: siteID, orderID: orderID, paymentIntentID: intent.id) { result in
```
ctarda, line 118: Not a blocker, but Would it make sense to extract this to a separate private method? It would help us avoid nesting this call into the receiveValue block.

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -110,6 +110,25 @@ private extension CardPresentPaymentStore {
...
+                switch result {
+                case .success(let intent):
+                    if intent.status == .succeeded {
```
ctarda: This looks a bit more "swift" to me:

```
guard intent.status == .succeed else {
onFailure(...)
return
}

onCompletion(.sucess)
```

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
ctarda: What about passing the error you are passing now (paymentCapture) with a new case in the underlyingError (e.g. .remoteServiceError)? In any case, this looks like something that could be merged as is and be deferred to a subsequent ticket, what do you think?

ctarda (approved):
Two suggestions for you to consider about a more "swiftier" approach. Other than that, :shipit:

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
allendav: So... I was thinking about this a bit and it reminded me of something I considered yesterday too.... there is a nice chaining of `createPaymentIntent`, `collectPaymentMethod`, `processPayment` already in place in `StripeCardReaderService` `capturePayment` where it might make sense to have intent capture instead. However, we’d have to do something similar for what we did for tokens since we don’t want to introduce a dependency between `StripeCardReaderService` and Networking. We’d need a “PaymentIntentCaptureProvider” or some such.

So, if I were going to move the call to self.remote.captureOrderPayment to its own private function, I’d be inclined to consider going the whole way to moving it and adding it to that chain. What do you think? I can open an issue for whichever path we choose (and commit this as it is for now)

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4004/commits/ad72e0190f0c72e064d51457bedb083f650d9496

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
allendav: I don't know. I'm still thinking a separate enumeration for the payment store is more appropriate. I'll merge this and I've opened the following ticket to explore further once we better understand the various ways the plugin and WCPay backend might fail to capture. https://github.com/woocommerce/woocommerce-ios/issues/4012

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
ctarda: I have been thinking about this on my side too and I think that, if we changed it, we should go the other way around. 

I logged a ticket earlier today for adding a new store to manage the business logic related to WCPay (#4011 ) I think we might need to consider to pull this to that store. 

Maybe we can chat about this at another time? 😊

allendav: Sounds good.

Issue #3975, opened 2021-04-16
[Mobile Payments] [Saved Receipts]  Save and present saved receipts. Mark III: UI


ctarda:
For context: p91TBi-4KR-p2

Add a "See receipts" button to the order details view

PR #4010, opened 2021-04-21
[Mobile Payments] Printing receipts. Mark II: Yosemite

### LINKED ISSUE #3979: [Mobile Payments] [Receipts] Print Receipts. Mark II: Yosemite

ctarda:
Closes #3979 

We continue building our prototype (for reference, p91TBi-4q4-p2), this time, by adding a new store to manage the business logic related to receipts.

https://user-images.githubusercontent.com/2722505/115587100-453a2080-a29b-11eb-82ea-1458dff2f4c7.mov

In this first iteration, all we do is print a receipt, but soon (e.g. #3975) we will need to add support for persisting the data necessary to generate receipts for any eligible order.

Tradeoffs:
* The action to generate and print the receipt is triggered automatically after a payment succeeds. This is temporary, and it is not the final, expected behaviour. But it helps test this PR.

This PR starts where #4002 ends:

## Changes
* Add a new Store (ReceiptStore) and a new Action (ReceiptAction) to encapsulate the business logic of loading pre-existing receipts from Storage, and coordinate printing receipts to the ReceiptPrinterService introduced in PR #4002 
* Rename the model object that was passed as parameter to he ReceiptPrinterService. This is motivated because...
* We abandon the pre-exisitng idea of persisting PaymentIntent. Instead, we will persist only the information we need to generate a receipt. This is something that might have to be rolled back later. We'll see.
* Some unit tests.

## How to test

1. Requires setting up a store for Card Present Payment Testing (also P91TBi-4BH-p2 for more specific instructions)

* Build and run on a device.
* Tap "Connect card reader"
* Tap "Connect to reader"
* At that point, and only in order to test this PR, a preview of the receipt should be presented. At this point, the preview only shows a page number and the actual amount and currency of the order that was just charged using the card reader. The amount and currency should match the Order's.

Note: this is not the actual flow that will be implemented in the final beta. There is work pending in tickets  #3982, #3981 #3978

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardPresentDetails.swift
+++ b/Hardware/Hardware/CardReader/CardPresentDetails.swift
@@ -29,3 +29,17 @@ public struct CardPresentDetails {
     /// (Only applicable to EMV payments) The authorization data from the card issuer.
     public let emvAuthData: String?
 }
+
+extension CardPresentDetails: Equatable {
+    public static func ==(lhs: CardPresentDetails, rhs: CardPresentDetails) -> Bool {
+        return lhs.last4 == rhs.last4 &&
+            lhs.expMonth == rhs.expMonth &&
+            lhs.expYear == rhs.expYear &&
+            lhs.cardholderName == rhs.cardholderName &&
+            lhs.brand == rhs.brand &&
+            lhs.fingerprint == rhs.fingerprint &&
+            lhs.generatedCard == rhs.generatedCard &&
+            lhs.receipt == rhs.receipt &&
+            lhs.emvAuthData == rhs.emvAuthData
+    }
+}
--- a/Hardware/Hardware/Printer/ReceiptParameters.swift
+++ b/Hardware/Hardware/Printer/ReceiptParameters.swift
@@ -1,6 +1,14 @@
+/// Encapsulates the information necessary to print a receipt for a
+/// card present payment
 public struct ReceiptParameters {
+    /// The total amount
     public let amount: UInt
+
+    /// The currency
     public let currency: String
+
+    /// Metadata provided by the payment processor. Contains the
+    /// to be added to the receipt required by the card networks.
     public let cardDetails: CardPresentDetails
 
     public init(amount: UInt, currency: String, cardDetails: CardPresentDetails) {
--- a/Hardware/Hardware/Printer/ReceiptParameters.swift
+++ b/Hardware/Hardware/Printer/ReceiptParameters.swift
@@ -1,4 +1,4 @@
-public struct ReceiptRegulatoryInfo {
+public struct ReceiptParameters {
     public let amount: UInt
     public let currency: String
     public let cardDetails: CardPresentDetails
--- a/Hardware/Hardware/Printer/ReceiptParameters.swift
+++ b/Hardware/Hardware/Printer/ReceiptParameters.swift
@@ -9,9 +9,9 @@ public struct ReceiptParameters {
 
     /// Metadata provided by the payment processor. Contains the
     /// to be added to the receipt required by the card networks.
-    public let cardDetails: CardPresentDetails
+    public let cardDetails: CardPresentTransactionDetails
 
-    public init(amount: UInt, currency: String, cardDetails: CardPresentDetails) {
+    public init(amount: UInt, currency: String, cardDetails: CardPresentTransactionDetails) {
         self.amount = amount
         self.currency = currency
         self.cardDetails = cardDetails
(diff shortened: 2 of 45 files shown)


Hardware/Hardware/CardReader/CardPresentDetails.swift
```diff
@@ -29,3 +29,17 @@ public struct CardPresentDetails {
     /// (Only applicable to EMV payments) The authorization data from the card issuer.
     public let emvAuthData: String?
 }
+
+extension CardPresentDetails: Equatable {
```
allendav: I should have thought of this before, but `CardPresentTransactionDetails` might be a clearer name for this struct. I had to think about it for a few seconds (and look at the heading comment again) to remember what this struct's responsibility was.

Hardware/Hardware/Printer/ReceiptParameters.swift
```diff
@@ -0,0 +1,19 @@
+/// Encapsulates the information necessary to print a receipt for a
+/// card present payment
+public struct ReceiptParameters {
```
allendav: Maybe `CardPresentReceiptParameters` instead - we might want to keep the generic ReceiptParameters available for the future

Hardware/Hardware/Printer/ReceiptParameters.swift
```diff
@@ -0,0 +1,19 @@
...
+    /// The total amount
+    public let amount: UInt
+
+    /// The currency
+    public let currency: String
+
+    /// Metadata provided by the payment processor. Contains the
```
allendav: the -> data

Yosemite/Yosemite/Model/Payments/PaymentIntent+ReceiptParameters.swift
```diff
@@ -0,0 +1,17 @@
+import Hardware
+
+extension PaymentIntent {
+    /// Maps a PaymentIntent into an struct that contains only the data we need to
+    /// render a receips.
```
allendav: receips -> receipt

Yosemite/Yosemite/Model/Payments/PaymentIntent+ReceiptParameters.swift
```diff
@@ -0,0 +1,17 @@
...
+    /// - Returns: an optional struct containing all the data that needs to go into a receipt
+    func receiptParameters() -> ReceiptParameters? {
```
allendav: We probably will need order ID too so that a receipt can access the items in the order?

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -131,7 +135,8 @@ private extension CardPresentPaymentStore {
                         onCompletion(.failure(CardReaderServiceError.paymentCapture()))
                         return
                     }
-                    onCompletion(.success(()))
+                    // This is always failing at this point. 
+                    //onCompletion(.success(receiptParameters))
```
allendav, line 139: I'll have this branch ready today for moving forward: https://github.com/Automattic/woocommerce-payments/tree/try/capture-endpoint

Yosemite/Yosemite/Model/Payments/PaymentIntent+ReceiptParameters.swift
allendav: Nevermind. I see the action has the orderID already

allendav (approved):
Tests fine. A few suggestions. :shipit:

Hardware/Hardware/CardReader/CardPresentDetails.swift
ctarda: Makes sense, I'll rename now

Hardware/Hardware/Printer/ReceiptParameters.swift
ctarda: Sounds good

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
ctarda: Awesome, thanks. I had to comment this out in #4015, otherwise the whole process is reported as an error (which is what we want, only, we don't want it now)

ctarda:
Thanks for the review @allendav . I've renamed those structs fixed the typos, and will merge as soon as CI give me the ✅

PR #4017, opened 2021-04-22
Add-Ons: Add beta indicator banner

Ecarrion:
Part of #3894  

# Why

This PR does 3 things(Sorry, I know it should do only one thing!)

1. Update the info label below the add-on list to have a new copy and to have a bit more padding.

2. Add a background color/view to matches the design better

3. Show a beta indicator top banner. Button actions will be handled later

# How
1. Just changed the copy in the localization menu and added some top padding to `OrderAddOnNoticeView`

2. Update `OrderAddOnListI1View` to have a solid color as a background view, so it can expand beyond the safe areas and cover the whole screen.

3. To read the whole story: paNNhX-dd-p2

	- Created `TopBannerWrapperView` in order to have a view that wraps a `TopBannerView` but allows us to have a custom `intrinsicContentSize` without having to modify the legacy of `TopBannerView`. This is needed because `SwiftUI` is unable to render a `TopBannerView` instance due to how its constraints are built. 
	
	- Created a `OrderAddOnTopBanner` which is a `struct` that wraps a `UIKitView` using the `UIViewRepresentable` protocol, to allow `SwiftUI` to represent it. The `UIKit` view that this wraps is a `TopBannerWrapperView` already configured to show the "beta banner" content as well as provide functions to get notified when the action buttons are pressed.
	
	- Update `OrderAddOnListI1View` to include the new `OrderAddOnTopBanner` created in the step above.
	
# Demo

demo

# Testing steps

**Prerequisites:** Have your site with the add-ons plugin installed and with at least one order with add-ons.

- See that there is a "beta feature" banner
- Tap on the chevron button and make sure the banner collapses/expands accordingly
 
Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnTopBanner.swift
```diff
@@ -0,0 +1,79 @@
...
+    private var onGiveFeedback: Callback? = nil
+
+    /// Closure to be invoked when the "Dismiss" button is pressed.
+    ///
+    private var onDismiss: Callback? = nil
+
+    /// Create a view with the desired `width`. Needed to calculate a correct view `height` later.
+    ///
+    init(width: CGFloat) {
+        self.bannerWrapper = TopBannerWrapperView(width: width)
```
pmusolino: nit: `self` is not needed

pmusolino (commented):
The resize after the rotation in landscape mode is not working correctly; in fact, the height of the banner will be the same as when it is in portrait mode. Is there a way to recalculate the banner height after the screen rotation?


| Portrait             |  Landscape |
Simulator Screen Shot - iPhone 12 Pro - 2021-04-26 at 12 45 24 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-26 at 12 40 18

| Portrait with long text |  Landscape with long text |
Simulator Screen Shot - iPhone 12 Pro - 2021-04-26 at 12 41 26 | Simulator Screen Shot - iPhone 12 Pro - 2021-04-26 at 12 41 28

Ecarrion:
Hey, @pmusolino Thanks for the review and for spotting this problem.

It has been fixed in https://github.com/woocommerce/woocommerce-ios/pull/4017/commits/39bb3df9ea78a18bebc9cff471384853fb928e78

Basically, I was doing bad things by storing a class in a `SwiftUI` type. I've changed it to the recommended way of using a `Coordinator` that to preserve any state across the layout passes. Will post a P2 with the whole story! 😇 

Screen Shot 2021-04-26 at 8 47 25 AM

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnTopBanner.swift
```diff
@@ -0,0 +1,94 @@
...
+    ///
+    private let width: CGFloat
+
+    /// Closure to be invoked when the "Give Feedback" button is pressed.
+    ///
+    private var onGiveFeedback: Callback? = nil
+
+    /// Closure to be invoked when the "Dismiss" button is pressed.
+    ///
+    private var onDismiss: Callback? = nil
```
koke, line 17: I am wondering why these are optional instead of part of the initializer. If the view is always going to show these two buttons, should it ensure there is a callback to handle them?

Ecarrion: Just for syntactic sugar.

After looking at some native `SwiftUI` components, I decided to add the closure as a modifier rather than as a parameter as it reads a little bit better.

```swift
var body: some view {
    OrderAddOnTopBanner(width: geometry.size.width, onGiveFeedback: {
         // Give Feedback
    }, onDismiss: {
         // Dismiss
    })
        .bodyStyle()
	    .fixedSize()
}
```

VS

```swift
var body: some view {
    OrderAddOnTopBanner(width: geometry.size.width)
        .onDismiss {
            //dismiss
        }
        .onGiveFeedback {
            // Give Feedback
        }
        .bodyStyle()
        .fixedSize()
}
```

I could go either way if you feel strongly about it tho!

pmusolino (approved):
Looks good to me 👍

PR #4022, opened 2021-04-23
Activity Indicator in SwiftUI

pmusolino:
Since we need to use a progress indicator in SwiftUI, but we are still supporting iOS 13, we can't use the `ProgressView` introduced in SwiftUI with iOS 14. For this reason, I introduced a custom component called `ActivityIndicatorView` which is a wrapper around the UIKit component called `UIActivityIndicatorView` which allows us to display an activity indicator and to decide if we want to start or stop the animation.

## How to use it
```
ActivityIndicatorView(isAnimating: .constant(true), style: .large)
```

## Testing
For the moment the component is still not used, so just look at the preview in SwiftUI.

## Screenshot
Schermata 2021-04-23 alle 11 52 57

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
@@ -1,0 +1,24 @@
+import SwiftUI
+
+/// This is a wrapper view for using the UIActivityIndicator in SwiftUI. The native progress indicator is not available under iOS 13.
+///
+struct ActivityIndicatorView: UIViewRepresentable {
+    @Binding var isAnimating: Bool
+    let style: UIActivityIndicatorView.Style
+
+    func makeUIView(context: UIViewRepresentableContext<ActivityIndicatorView>) -> UIActivityIndicatorView {
+        return UIActivityIndicatorView(style: style)
+    }
+
+    func updateUIView(_ uiView: UIActivityIndicatorView, context: UIViewRepresentableContext<ActivityIndicatorView>) {
+        isAnimating ? uiView.startAnimating() : uiView.stopAnimating()
+    }
+}
+
+struct ActivityIndicatorView_Previews: PreviewProvider {
+
+    static var previews: some View {
+        ActivityIndicatorView(isAnimating: .constant(true), style: .large)
+            .previewLayout(.fixed(width: 100, height: 100))
+    }
+}
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicator.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicator.swift
@@ -2,23 +2,23 @@ import SwiftUI
 
 /// This is a wrapper view for using the UIActivityIndicator in SwiftUI. The native progress indicator is not available under iOS 13.
 ///
-struct ActivityIndicatorView: UIViewRepresentable {
+struct ActivityIndicator: UIViewRepresentable {
     @Binding var isAnimating: Bool
     let style: UIActivityIndicatorView.Style
 
-    func makeUIView(context: UIViewRepresentableContext<ActivityIndicatorView>) -> UIActivityIndicatorView {
+    func makeUIView(context: UIViewRepresentableContext<ActivityIndicator>) -> UIActivityIndicatorView {
         return UIActivityIndicatorView(style: style)
     }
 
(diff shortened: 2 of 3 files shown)


WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
```diff
@@ -0,0 +1,24 @@
+import SwiftUI
+
+/// This is a wrapper view for using the UIActivityIndicator in SwiftUI. The native progress indicator is not available under iOS 13.
+///
+struct ActivityIndicatorView: UIViewRepresentable {
+    @Binding var isAnimating: Bool
```
Ecarrion: Since we are all new to `SwiftUI` I think some comments explaining how `Binding` works and how to pass a binding property would be nice. What do you think?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
```diff
@@ -0,0 +1,24 @@
...
+    func makeUIView(context: UIViewRepresentableContext<ActivityIndicatorView>) -> UIActivityIndicatorView {
+        return UIActivityIndicatorView(style: style)
+    }
+
+    func updateUIView(_ uiView: UIActivityIndicatorView, context: UIViewRepresentableContext<ActivityIndicatorView>) {
+        isAnimating ? uiView.startAnimating() : uiView.stopAnimating()
+    }
+}
+
+struct ActivityIndicatorView_Previews: PreviewProvider {
```
Ecarrion: I always wonder, should this be under a `#if DEBUG` compiler flag? 
If not, this type would end up in our production binary, right?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
```diff
@@ -0,0 +1,24 @@
+import SwiftUI
+
+/// This is a wrapper view for using the UIActivityIndicator in SwiftUI. The native progress indicator is not available under iOS 13.
+///
+struct ActivityIndicatorView: UIViewRepresentable {
```
Ecarrion: I think `SwiftUI` tends to drop the `View` postfix for the types.
`ActivityIndicator` seems a bit more aligned with their naming conventions. What do you think?

Ecarrion (commented):
Looks good to me. Left a couple of suggestions/questions.
Also, I think it's crazy that `SwiftUI` hasn't ported the `ActivityIndicatorView` already. :rage4:

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
pmusolino: This is no more needed in Xcode 12, [Apple strips automatically the code under the PreviewProvider from the production binary](https://web.archive.org/web/20190911210340/https://developer.apple.com/documentation/xcode_release_notes/xcode_11_release_notes#3384944).
aUkVp

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
pmusolino: I agree, I changed the name e7a0b72cab56ff5918c92327a66403a6403b3156

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
pmusolino: I think it's a bit complex to explain how binding works and do it everywhere because it's really essential in SwiftUI. You will hardly find a view without `@Binding` properties declared. It could get a bit repetitive

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ActivityIndicatorView.swift
Ecarrion: Oh nice!

Ecarrion (approved):

pmusolino:
Thanks for the review @Ecarrion !

koke:
I'm late to this PR, but I'm wondering what's the use case for the activity indicator not animating, and if it would be better to always have it on. When there is no activity I imagine we'd be more likely to hide the view than disable it

Issue #3915, opened 2021-04-06
Coupon management (task 3): Yosemite layer


pmusolino:
After having implemented the [networking layer](https://github.com/woocommerce/woocommerce-ios/issues/3912) and the [storage layer](https://github.com/woocommerce/woocommerce-ios/issues/3913), we should make in communication them implementing the Yosemite layer, which will be exposed to the Woo app.

itsmeichigo:
Closing this issue in favor of the new Coupons project.

PR #4025, opened 2021-04-23
Coupon Management: Yosemite - Action and storage helper methods

joshheald-a8c:
Part of #3915 

## Description
This PR adds the prerequisites for implementation of the `CouponStore`:
* Storage loading and deletion helper methods
* ReadOnlyConvertible conformance for `Coupon`
* The `synchroniseCoupons` action, designed for use in paginated table views/list selectors
* Required type aliases for using the different types of `Coupon` within Yosemite

## Testing
The endpoint and storage is not in use yet, so no applicable testing other than the unit tests, and CI once it's merged to the main repo.

Update release notes:


Storage/StorageTests/Tools/StorageTypeDeletionsTests.swift
```diff
@@ -0,0 +1,62 @@
...
+
+    func test_deleteCoupon_by_siteID_deletes_all_coupons() throws {
+        // Given
+        let coupon1 = storage.insertNewObject(ofType: Coupon.self)
+        coupon1.siteID = sampleSiteID
+        coupon1.couponID = 1
+
+        let coupon2 = storage.insertNewObject(ofType: Coupon.self)
+        coupon2.siteID = sampleSiteID
+        coupon2.couponID = 2
```
pmusolino, line 34: I think that here you should check that first the two objects exist in the Storage after adding them, otherwise the test, also if you delete the coupons, will be always green.

Storage/StorageTests/Tools/StorageTypeExtensionsTests.swift
```diff
@@ -973,4 +973,36 @@ class StorageTypeExtensionsTests: XCTestCase {
         // Then
         XCTAssertEqual(labelSettings, storedLabelSettings)
     }
+
+    func test_loadCoupon_by_siteID_couponID() throws {
```
pmusolino, line 977: Minor nit:
```suggestion
    func test_loadCoupon_by_siteID_and_couponID() throws {
```

Yosemite/Yosemite/Actions/CouponAction.swift
```diff
@@ -0,0 +1,20 @@
...
+    /// - `pageNumber`: page of results based on the `pageSize` provided. 1-indexed.
+    /// - `pageSize`: number of results per page.
+    /// - `onCompletion`: invoked when the sync operation finishes.
+    ///     - `result.success(Bool)`: value indicates whether there are further pages to retrieve.
+    ///     - `result.failure(Error)`: error indicates issues syncing the specified page.
+    ///
+    case synchronizeCoupons(siteID: Int64,
+                            pageNumber: Int,
+                            pageSize: Int,
+                            onCompletion: (Result<Bool, Error>) -> Void)
```
pmusolino, line 19: What do you think about having a tuple of a list of Coupons and a boolean that indicates whether there are further pages to retrieve? 
Something like that `onCompletion: (Result<(coupons: [Coupon], hasNextPage: Bool), Error>) -> Void)`

Yosemite/Yosemite/Model/Storage/Coupon+ReadOnlyConvertible.swift
```diff
@@ -0,0 +1,66 @@
...
+        minimumAmount = coupon.minimumAmount
+        maximumAmount = coupon.maximumAmount
+        emailRestrictions = coupon.emailRestrictions
+        usedBy = coupon.usedBy
+    }
+
+    /// Returns a ReadOnly (`Networking.Coupon`) version of the `Storage.Coupon`
+    ///
+    public func toReadOnly() -> Yosemite.Coupon {
+        return Yosemite.Coupon(siteID: siteID,
```
pmusolino: Minor nit
```suggestion
        return .init(siteID: siteID,
```

pmusolino (approved):
Looks good to me, just some minor nits 👍

Yosemite/Yosemite/Actions/CouponAction.swift
joshheald-a8c: I took cues from the existing `synchronize` actions in designing this one, and found that none of the paginated actions include the model objects in the `onCompletion` arguments. There are a few non-paginated actions which do, and actions for some paginated endpoints which have a `retrieve` method that may return the models, e.g. on `[ProductAction](https://github.com/woocommerce/woocommerce-ios/blob/e8f9a518a50c1d7200ce86c6d6122320da658335/Yosemite/Yosemite/Actions/ProductAction.swift#L37)`.

If we passed in the coupons, would you expect the completion handler to receive only the coupons from the page requested, or all coupons retrieved so far? I would think just the ones from the page specified, personally, but that list seems like it would be of limited use to the view.

I'm not sure that this would be the right way to go for the `synchronizeCoupons` action, as returning the results seems like a side effect. If there are future uses of the store which may benefit from getting the results back, I think it might be better to add a `retrieveCoupons` action, which provides the model objects to the completion handler. That said, perhaps we'd always want to store coupons we'd gone to the trouble of fetching from the API, so synchronize would still be the right method to call in that situation too. 

My understanding is that accessing model objects for display is usually a two part process: firstly fetch and store the model (synchronize) and secondly use a `FetchedResultsController` to retrieve them from storage for display in a table view. Perhaps, for other use cases which require the results in response to an action instead of via a `FetchedResultsController`, we could also provide Yosemite methods for getting models from Storage, which could be used within completion handlers to compose the required behaviour that way. Have I understood the approach correctly here, wrt `FetchedResultsController`s?

For the coupon management screen, I don't think we need the completion handler to receive the array of coupons, because its `FetchedResultsController` will handle getting the models from Core Data, but I'm wary that I might be missing the point that you're making with this suggestion, so I'm definitely happy to discuss further.

Yosemite/Yosemite/Model/Storage/Coupon+ReadOnlyConvertible.swift
joshheald-a8c: Using `.init` gives the error `The compiler is unable to type-check this expression in reasonable time; try breaking up the expression into distinct sub-expressions`

However it doesn't need the context of `Yosemite`, so I've removed that to shorten it a bit.

Yosemite/Yosemite/Actions/CouponAction.swift
pmusolino: > For the coupon management screen, I don't think we need the completion handler to receive the array of coupons, because its FetchedResultsController will handle getting the models from Core Data, but I'm wary that I might be missing the point that you're making with this suggestion, so I'm definitely happy to discuss further.
 

It's totally ok to left it as you implemented it, it makes totally sense to use the `FetchedResultsController`, also because, maybe, we want to show cached coupons before fetching the new ones. 💯

pmusolino:
Thanks @joshheald-a8c! I'm going to merge it ;)

PR #4026, opened 2021-04-23
Coupon Management: Yosemite - CouponStore synchronizeCoupons method

joshheald-a8c:
Part of #3915 
Please ensure #4025 is merged before this PR

## Description
This PR adds the initial implementation of the `CouponStore`, containing what it needs to enable us to list coupons either in a table view, or a `PaginatedListSelectorViewController`. This requires only the `synchronizeCoupons` method, which inserts or updates coupons retrieved from the `Remote`, deleting any previously stored coupons as it does so, to avoid showing coupons which were previously deleted on another device.

## Testing
The endpoint and storage are not in use yet, so no applicable testing other than the unit tests, and CI once it's merged to the main repo.

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/CouponsRemote.swift
+++ b/Networking/Networking/Remote/CouponsRemote.swift
@@ -1,8 +1,20 @@
 import Foundation
 
+/// Protocol for `CouponsRemote` mainly used for mocking.
+///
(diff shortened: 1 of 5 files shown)


pmusolino (approved):
Nothing to report! Great work here @joshheald-a8c  :shipit:

PR #4027, opened 2021-04-23
Add-ons: Top banner button handling

### LINKED ISSUE #3894: Add-Ons: Add outdated notice.

Ecarrion:
closes #3894

# Why

After #4017, where I introduced a beta top banner view on the order-add on screen this PR is in charge of handling action buttons(Give feedback and dismiss) of that banner.

- The "Give Feedback" button will launch a survey feedback
- The "Dismiss" button will hide the banner for the screen session.

# How

- Updated `WooConstants` with the add-on survey URL under the `orderAddOnI1Feedback` property.

- Added a new `.addOnsI1` case on `SurveyViewController.source`

- Created `Survey.swift` to wrap `SurveyCoordinatingController` for `SwiftUI`.

- Updated `OrderAddOnListI1ViewModel` to conform to `ObservableObject`. This is needed in order for the view to observe changes from this object using the `@ObservedObject` property wrapper.

- Updated `OrderAddOnListI1ViewModel` to add two new `@Published` properties. `shouldShowSurvey` and `shouldShowBetaBanner` which controls when the survey should be presented and when the banner should be rendered.

- Created a new view modifier/extension to render(or not) a view when a condition is met. paNNhX-dg-p2

- Updated `OrderAddOnListI1View` to render content conditionaly according to the view model properties.

# Demo

https://user-images.githubusercontent.com/562080/115928495-ccc69180-a44b-11eb-8948-e4aa841d95dd.mov

# Testing Steps
**Prerequisites:** Have your site with the add-ons plugin installed and with at least one order with add-ons.

- See that there is a "beta feature" banner
- Tap on the "Give Feedback" button and make sure you are able to fill the survey
- After dismissing the survey, tap on the "Dismiss" button and make sure the banner is hidden.

Note: The next time you open an order's add-ons, the banner should be visible again.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/View Modifiers/View+Conditionals.swift
+++ b/WooCommerce/Classes/View Modifiers/View+Conditionals.swift
@@ -1,0 +1,14 @@
+import SwiftUI
+
+extension View {
+    /// Renders a view if the provided  `condition` is met.
+    /// If the `condition` is not met, an instance of `EmptyView` will be used in place of the receiver view.
+    ///
+    func renderedIf(_ condition: Bool) -> some View {
+        guard condition else {
+            return AnyView(EmptyView())
+        }
+        return AnyView(self)
+    }
+}
+
--- a/WooCommerce/Classes/View Modifiers/View+Conditionals.swift
+++ b/WooCommerce/Classes/View Modifiers/View+Conditionals.swift
@@ -2,15 +2,12 @@ import SwiftUI
 
 extension View {
     /// Renders a view if the provided  `condition` is met.
-    /// If the `condition` is not met, an instance of `EmptyView` will be used in place of the receiver view.
+    /// If the `condition` is not met, an `nil`  will be used in place of the receiver view.
     ///
-    func renderedIf(_ condition: Bool) -> some View {
-        Group {
-            if condition {
-                self
-            } else {
-                EmptyView()
-            }
+    func renderedIf(_ condition: Bool) -> Self? {
+        guard condition else {
+            return nil
         }
+        return self
     }
 }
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/View Modifiers/View+Conditionals.swift
```diff
@@ -0,0 +1,13 @@
+import SwiftUI
+
+extension View {
+    /// Renders a view if the provided  `condition` is met.
+    /// If the `condition` is not met, an instance of `EmptyView` will be used in place of the receiver view.
+    ///
+    func renderedIf(_ condition: Bool) -> some View {
```
pmusolino: I like this view modifier 👍

pmusolino: Is it probably better to rename it to `renderIf(_ condition: Bool)`?

pmusolino (approved):
LGTM :shipit:  just a minor nit

WooCommerce/Classes/View Modifiers/View+Conditionals.swift
Ecarrion: I was looking at `SwiftUI` API naming and saw examples like "hidden()" instead of "hide()".
That's why I choose "renderedIf()" instead of "renderIf()"

Makes sense?

pmusolino: Makes sense ;)

Issue #3916, opened 2021-04-06
Coupon management (task 4): list all the coupons


pmusolino:
As part of the Coupons management, we should implement a screen that displays the list of all the coupons available.

We should manage:
- [ ] Pull to refresh.
- [ ] Pagination.
- [ ] Cells.
- [ ] Tap on cells.
- [ ] State in case of error or in case there are no coupons.
- [ ] "Add" button in the navigation bar.
- [ ] Implement a swipe for deleting the single coupon. 
- [ ] Dark mode.

<img width="436" alt="Schermata 2021-04-06 alle 17 10 44" src="https://user-images.githubusercontent.com/495617/113734520-b0142500-96fb-11eb-995a-a681dc116592.png">

joshheald-a8c:
- [x]  Pull to refresh.
- [x]   Pagination.
- [x]   Cells.
- [x]   Tap on cells.
- [x]   State in case of error or in case there are no coupons.
- [x]   "Add" button in the navigation bar.
- [ ]   Implement a swipe for deleting the single coupon.
- [ ]   Dark mode.

itsmeichigo:
Closing this issue in favor of the new Coupons project.

PR #4032, opened 2021-04-26
Shipping Labels M2: Prevent button flashing when a different address is selected

### LINKED ISSUE #3914: Shipping Labels M2: Primary button on suggested address screen has a flash of black text when selecting a different address
**Describe the bug**

On the Shipping Labels suggested address screen (`ShippingLabelSuggestedAddressViewController`) the primary button text changes when you select a different address. When it changes, there is a flash of black text before the text is shown as white.

**Expected behavior**

I'd expect the button text to always be white, without that flash of black text.

**To Reproduce**

1. Before testing, make sure you have an order with an address that is slightly different from the suggested mailing address. (For example, enter an incomplete street name "Market" instead of "Market St.")
2. Go to the Orders > Processing tab.
3. Select an order.
4. Select "Create Shipping Label."
5. Select "Continue" in the "Ship from" section.
6. If the "Ship from" address is accepted, select "Continue" in the "Ship to" section. -> If the order's shipping address is different from the suggested address, you should land on the suggested address screen at this point.
7. Select between "Address Entered" and "Address Suggested" and watch the text on the primary button change as you make your selection. -> The text on the button flashes black as it changes.

**Screenshots**

koke:
Closes #3914

## Why

As described in #3914, the primary button would briefly turn black when a different address was selected.

## How

The issue was that `configureUseAddressButton()` was being called every time the address changed to update the label, but this was also configuring all the appearance for the button.

The `applyPrimaryButtonStyle` does a number of things, but the problematic sequence is that it calls `titleLabel?.applyHeadlineStyle()` first, which sets the color to black, then `setTitleColor(.primaryButtonTitle, for: .normal)`, which sets the right color.
Since setting the title does so with animations enabled, we can see the color changing to black then white (if we call `applyPrimaryButtonStyle` but not `setTitle`, there is also no flashing.

To solve this, I have decoupled the initial styling of the button (which shouldn't change), from updating the label with a new `updateUseAddressButton()` method.

The button still animates the label change, but I believe that's the desired behavior.

wcios-suggested-address-flash

## How to test

I followed the steps in  #3916:

> 1. Before testing, make sure you have an order with an address that is slightly different from the suggested mailing address. (For example, enter an incomplete street name "Market" instead of "Market St.")
> 1. Go to the Orders > Processing tab.
> 1. Select an order.
> 1. Select "Create Shipping Label."
> 1. Select "Continue" in the "Ship from" section.
> 1. If the "Ship from" address is accepted, select "Continue" in the "Ship to" section. -> If the order's shipping address is different from the suggested address, you should land on the suggested address screen at this point.
> 1. Select between "Address Entered" and "Address Suggested" and watch the text on the primary button change as you make your selection. -> The text on the button flashes black as it changes.

Update release notes:

☝🏽 I understand the feature is not yet released, so I think a bug fix shouldn't make it into the release notes?

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Suggested Address/ShippingLabelSuggestedAddressViewController.swift
@@ -38,7 +38,7 @@ final class ShippingLabelSuggestedAddressViewController: UIViewController {
     private var selectedAddress: SelectedAddress = .suggested {
         didSet {
             tableView.reloadData()
-            configureUseAddressButton()
+            updateUseAddressButton()
         }
     }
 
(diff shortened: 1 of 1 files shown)


pmusolino (approved):
LGTM :shipit: thank you @koke for fixing this!

PR #4020, opened 2021-04-23
Shipping Labels M2: handle the selected value in the package list

pmusolino:
Closes #3964 
Closes #3966
Closes #3965 

## Description
In this PR there are a lot of deletions because I removed a lot of superfluous code after the refactoring, and now the data of the Package Flow are shared between the views as environment property, making the code lighter and more understandable.
I also handled the value selected in the Package List screen and moved the fetching of the existing packages from the remote endpoint in the main view model.

## Changes
- Fixed the `WeightFormatter`, which now is able to handle the case when the weight is an empty string + tests.
- Moved some Shipping Labels classes under a new folder.
- Deleted the `ShippingLabelPackageListViewModel` because it is no more needed.
- Deleted the middleware view `ShippingLabelPackageSelected` (and its view model) that will be reimplemented in M4, since now it is useless.
- Now the `ShippingLabelPackageList` view takes the `ShippingLabelPackageDetailsViewModel` that is declared as `EnvironmentObject`, and that will be updated dynamically in the entire Package Selection flow.
- `ShippingLabelPackageDetails` now handle the selected package, showing the ID of the selected package. The final value will be shown later and will be part of another PR.
- In `ShippingLabelPackageDetails` I moved the `sampleOrder()` method under the view model.
- `ShippingLabelPackageDetailsViewModel`: massive refactoring, adding the fetching of the existing packages from API and the selected package + unit tests.
- `TitleAndValueRow`: fixed the selection, adding the content shape.

## Testing
1. Launch the app in debug mode (because of the feature flag).
2. Open an order detail.
3. Press the button "Create Shipping Label".
4. Press the button "Continue" in Create Shipping Label form, under the **Ship From** cell. Confirm the address.
5. Press the button "Continue" in Create Shipping Label form, under the **Ship To** cell. Confirm the address.
6. Press the button "Continue" in Create Shipping Label form, under the **Package Details** cell.
7. You should see the new screen that shows the Package Details screen and all the items that will be shipped.
8. Press the "Package selected" row. -> You should see the new screen with the list of the existing packages, with Custom Packages on top if any.
9. Select a package, and press "Done". -> You should see the selected package ID under the "Package Selected" row.

## GIF

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #2132, opened 2020-04-13
Failing unit tests setting Arabic language


pmusolino:
We have some failing tests in `DoubleWooTests`, `NSDecimalNumberWooTests` and `ReviewViewModelTests` when the language in the simulator is configured to Arabic.

<img width="559" alt="Schermata 2020-04-13 alle 11 53 29" src="https://user-images.githubusercontent.com/495617/79111853-99075500-7d7d-11ea-876c-2fbc2af79033.png">

pmusolino:
Update.
I re-tested this issue, and now I also see 15 failing tests in `CurrencyFormatterTests`

<img width="402" alt="Schermata 2020-09-04 alle 15 34 01" src="https://user-images.githubusercontent.com/495617/92245063-14e87200-eec4-11ea-9567-0dcb9ea93dfb.png">

rachelmcr:
These tests are failing because we are making assertions on localized, human-readable strings and Arabic uses different numerals. Any time we convert a number to a string with `NumberFormatter().string` it localizes the string, which returns e.g. `١,١٧` instead of `1,17` (equivalent strings but with different numerals).

Do we intend for our tests to pass in any supported locale? Or do we only mean to test them in English?

I'm wondering because in `CurrencyFormatterTests`, for example, we set English as the locale in a number of tests using this constant:

```swift
private let sampleLocale = Locale(identifier: "en")
```

If we want to force the tests to always run with `EN` as the locale, we could add that to the unit test plan configuration (under `UnitTests.xctestplan` > Configurations > Localization). That way, regardless of the simulator's language, the tests will always be run in English without having to force it in specific tests.

If we want these unit tests to work in all languages, though, we'll need to account for cases where the localized string uses different numerals than English. One possibility would be to set the expected value based on the simulator locale, taking into account the different expected result with Arabic. For example, in `testDecimalSeparatorIsLocalized` in `CurrencyFormatterTests`:

```swift
let expectedResult = Locale.current.languageCode == "ar" ? "١,١٧" : "1,17"
```

PR #4035, opened 2021-04-27
Update WeightFormatter to use MeasurementFormatter

koke:
The current implementation basically joins the value and unit with a space.

This works in many cases, but not for all locales. RTL languages will need the components reversed, some languages (like Chinese) prefer omitting the space between the number and the unit, and many others translate the unit. Additionally, some regions like Spain use a comma for the decimal separator instead of a period.

By using MeasurementFormatter when possible, we get all these localizations for free.

When the weight unit isn't recognized, we fall back to the previous behavior.

One noticeable difference is that Woo seems to use `lbs` as the abbreviation for pounds, while iOS uses `lb`. I think this should be OK, but could use a second opinion.

weight-formatter

cc @pmusolino since I spotted this while reviewing #4020 and I think there will be conflicts with that one.

Update release notes:


shiki:
@koke It looks like the conflicts have arrived. 👇  😅 We probably should get those fixed first. 

Not assigning myself to review this yet in case someone gets to it first.

koke:
@shiki thanks for the heads up, I've resolved the conflict and pushed those changes

WooCommerce/Classes/Tools/WeightFormatter.swift
```diff
@@ -1,30 +1,69 @@
...
-        return weight + " " + weightUnit
+        return formatWithFormatter(weight: weightValue, unit: unit)
     }
 }
 
-// MARK: - Utils
 private extension WeightFormatter {
-    func unwrapWeight(weight: String?) -> String {
-        return (weight == nil || weight?.isEmpty == true) ? "0" : (weight ?? "0")
+    func unit(symbol weightUnit: String) -> UnitMass? {
```
shiki, line 37: Side note: This `weightUnit` `String` should probably have been converted into an `enum` somewhere. Then again, maybe a plugin can make its own unit. 🙈

WooCommerce/WooCommerceTests/Tools/WeightFormatterTests.swift
```diff
@@ -5,20 +5,34 @@ final class WeightFormatterTests: XCTestCase {
 
     func test_weightFormat_returns_expected_values() {
```
shiki, line 6: Just a note that these tests fail when using an Arabic simulator. Not sure if we want to fix this right now or as a debt later. 

image

shiki (approved):
👏 :shipit:

WooCommerce/WooCommerceTests/Tools/WeightFormatterTests.swift
koke: I accidentally left my simulator in Spanish yesterday on a different PR and a dozen tests were failing because of that. I think we should address the issue separately. I see at least #2382 and #2132 related to this. The easiest solution is probably to force English/US as the System Language/Region in the test plan configuration.

PR #4038, opened 2021-04-27
Add-ons: Fix add-on cross-reference for variable products 

### LINKED ISSUE #4008: Add-ons: Missing add-ons in variable products
**Describe the bug**

The app is not showing add-ons on variable products

**To Reproduce**
1. Create a variable product with an add-on.
2. Place an order for that product, including the add-on.
3. View the order in the app.
4. The "View add-ons" button is missing

**Screenshots**

Order on the web:

Screen Shot 2021-04-21 at 12 17 49

Order in the app:

Simulator Screen Shot - iPhone 12 Pro - 2021-04-21 at 12 17 55

**Mobile Environment**
Please include:
- Device: Simulator
- iOS version: 
- WooCommerce iOS version:  `develop` (488d423)

Ecarrion:
fix #4008 

# Why

It was reported that variable product orders were not displaying add-ons correctly.
After a bit of investigation, I discovered two problems.

1. I was cross-referencing the order attributes with the product variation add-ons, but the product variation does not have add-on information, the main product has it, so we need to always cross-reference against the main product.

2. I noticed that an add-on without a price does not come with the expected format of `"add-on-title (add-on-price)"` and is just only `"add-on-title"`.

# How 

1. Changed `productOrVariationID` for `productID` for our cross-reference comparison.

2. Exit early and return the raw attribute name, when we detect it does not match our name format assumptions.

# Demo

https://user-images.githubusercontent.com/562080/116251798-9f703100-a734-11eb-8d3f-22820b4ccc54.mov

# Testing Steps

**Prerequisites:** Have your site with the add-ons plugin installed

On Core:
- Set a variable product and create an add-on without a price.
- Create an order with that variable product and that add-on.

On App:
- See that there is a "View Add-Ons" button.
- Tap on the button and see that it display the add-ons without a price appropriately.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/AddOns/AddOnCrossreferenceUseCase.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/AddOns/AddOnCrossreferenceUseCase.swift
@@ -32,6 +32,11 @@ struct AddOnCrossreferenceUseCase {
         let splitToken = " ("
         let components = attribute.name.components(separatedBy: splitToken)
 
+        // If name the format does not match our format assumptions, return the raw name.
+        guard components.count > 1 else {
+            return attribute.name
+        }
+
         // In case there are more `" ("` occurrences in the string, drop the last one assuming its the add-on price,
         // and join the remaining components to keep the original name integrity
         return components.dropLast().joined(separator: splitToken)
--- a/WooCommerce/Classes/ViewModels/Order Details/AddOns/AddOnCrossreferenceUseCase.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/AddOns/AddOnCrossreferenceUseCase.swift
@@ -32,7 +32,7 @@ struct AddOnCrossreferenceUseCase {
         let splitToken = " ("
         let components = attribute.name.components(separatedBy: splitToken)
 
-        // If name the format does not match our format assumptions, return the raw name.
+        // If name does not match our format assumptions, return the raw name.
         guard components.count > 1 else {
             return attribute.name
         }
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -751,7 +751,7 @@ private extension OrderDetailsDataSource {
     /// Returns an `empty` array if we can't find the product associated with order item or if the `addOns` feature is disabled.
     ///
     private func filterAddOns(of item: AggregateOrderItem) -> [OrderItemAttribute] {
-        guard let product = products.first(where: { $0.productID == item.productOrVariationID }), showAddOns else {
+        guard let product = products.first(where: { $0.productID == item.productID }), showAddOns else {
             return []
         }
         return AddOnCrossreferenceUseCase(orderItem: item, product: product).addOnsAttributes()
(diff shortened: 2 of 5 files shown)


WooCommerce/Classes/ViewModels/Order Details/AddOns/AddOnCrossreferenceUseCase.swift
```diff
@@ -32,6 +32,11 @@ struct AddOnCrossreferenceUseCase {
         let splitToken = " ("
         let components = attribute.name.components(separatedBy: splitToken)
 
+        // If name does not match our format assumptions, return the raw name.
```
shiki, line 35: Great job on all the comments in this `UseCase`.

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnListViewModel.swift
```diff
@@ -44,9 +44,15 @@ final class OrderAddOnListI1ViewModel: ObservableObject {
     /// The `attribute.name` comes in the form of "add-on-title (add-on-price)". EG: "Topping (Spicy) ($30.00)"
     ///
     private static func addOnName(from attribute: OrderItemAttribute) -> String {
-        attribute.name.components(separatedBy: " (") // "Topping (Spicy) ($30.00)" -> ["Topping", "Spicy)", "$30.00)"]
-            .dropLast()                              // ["Topping", "Spicy)", "$30.00)"] -> ["Topping", "Spicy)"]
-            .joined(separator: " (")                 // ["Topping", "Spicy)"] -> "Topping (Spicy)"
+        let components = attribute.name.components(separatedBy: " (") // "Topping (Spicy) ($30.00)" -> ["Topping", "Spicy)", "$30.00)"]
```
shiki, line 47: Since this looks like an important logic, should we consider merging this with `AddOnCrossreferenceUseCase.extractAddOnName` somehow? Not sure if it's worth it if we're going to replace this later on anyway. 🤷

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -751,7 +751,7 @@ private extension OrderDetailsDataSource {
     /// Returns an `empty` array if we can't find the product associated with order item or if the `addOns` feature is disabled.
     ///
     private func filterAddOns(of item: AggregateOrderItem) -> [OrderItemAttribute] {
-        guard let product = products.first(where: { $0.productID == item.productOrVariationID }), showAddOns else {
+        guard let product = products.first(where: { $0.productID == item.productID }), showAddOns else {
```
shiki, line 754: I changed this back to using `productOrVariationID` and the unit tests passed. Should we protect this with a unit test? Or is this temporary?

shiki (approved):
Works as advertised.  🎉 I can confirm that the View Add-ons button is now visible. The `develop` branch did not show it.

WooCommerce/Classes/ViewRelated/Orders/Order Details/AddOns/OrderAddOnListViewModel.swift
Ecarrion: Yeah, I expect all of this to be temporary. Also, the code here, while similar, is meant to be only for displaying/cosmetic purposes so I don't want to mix it with the cross-reference one!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
Ecarrion: Because of the nature of `OrderDetailsDataSource` I've found it's quite hard to add test to this logic :-/ :shame: on me

Ecarrion:
Thanks for the review @shiki!

PR #4039, opened 2021-04-27
Add subtitle and update copies on empty state view of Variations list

### LINKED ISSUE #3903: Variations: Update empty state screen
Ref: p91TBi-4O0-p2

If the product doesn't have variations and doesn't have attributes:
- Add subtitle Copy 
- Update button text

no-attrs-no-variations

Note: Leave the current state when the product doesn't have variations but does have attributes.

itsmeichigo:
Fix #3903 

## Description

This PR changes empty state view of Variations list screen:
- Updates `emptyStateConfig` with a content string for `detail` to show subtitle.
- Updates copies of title and button.

## Problem
I can see that the current font of subtitle text doesn't really match with the expected design. Also the button size is slightly different too. But I'm keeping both of them as-is for now as they seem to be out of scope of issue #3903. Let me know if the UI needs further adjustments - I'd need detailed specifications for the metrics and constraints too (it's hard to tell from the screenshot).

## Testing
This screen can be reached by either of these two flows:
- Select Product tabs -> Add new variable product
- Select Product tabs -> Add new variable product -> Add a new variation with attributes -> Delete that variation

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -18,7 +18,7 @@ final class ProductVariationsViewController: UIViewController {
         let message = NSAttributedString(string: Localization.emptyStateTitle, attributes: [.font: EmptyStateViewController.Config.messageFont])
         return .withButton(message: message,
                            image: .emptyBoxImage,
-                           details: "",
+                           details: Localization.emptyStateSubtitle,
                            buttonTitle: Localization.emptyStateButtonTitle) { [weak self] _ in
                             self?.createVariationFromEmptyState()
                            }
@@ -733,8 +733,11 @@ private extension ProductVariationsViewController {
     }
 
     enum Localization {
-        static let emptyStateTitle = NSLocalizedString("Add your first variation", comment: "Title on the variations list screen when there are no variations")
-        static let emptyStateButtonTitle = NSLocalizedString("Add Variation", comment: "Title on add variation button when there are no variations")
+        static let emptyStateTitle = NSLocalizedString("Create your first variation",
+                                                       comment: "Title on the variations list screen when there are no variations")
+        static let emptyStateSubtitle = NSLocalizedString("To add a variation, you'll need to set its attributes (ie \"Color\", \"Size\") first",
+                                                          comment: "Subtitle on the variations list screen when there are no variations")
+        static let emptyStateButtonTitle = NSLocalizedString("Add Attributes", comment: "Title on add variation button when there are no variations")
         static let addNewVariation = NSLocalizedString("Add Variation", comment: "Action to add new variation on the variations list")
         static let moreButtonLabel = NSLocalizedString("More options", comment: "Accessibility label to show the More Options action sheet")
         static let editAttributesAction = NSLocalizedString("Edit Attributes", comment: "Action to edit the attributes and options used for variations")
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -16,9 +16,9 @@ final class ProductVariationsViewController: UIViewController {
     ///
     private var emptyStateConfig: EmptyStateViewController.Config {
         let message = NSAttributedString(string: Localization.emptyStateTitle, attributes: [.font: EmptyStateViewController.Config.messageFont])
-        let productHasAttributes = product.attributesForVariations.isNotEmpty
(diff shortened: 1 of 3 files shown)


Ecarrion (changes requested):
Hi @itsmeichigo, thanks for working on this.

However, I spotted one main problem with the PR. 😁 

According to #3903, the copy changes should apply only when the product has no attributes and has no variations.

Right now, if a product has attributes but no variation the new copies don't make much sense.

create-variations

Let me know if you need further clarification. :-)

itsmeichigo:
Thank you @Ecarrion for spotting the issue - I have updated the changes to take the edge case into consideration. The empty state should be as expected now.

itsmeichigo:
Thank you @Ecarrion for spotting the issue - I have updated my changes to take the edge case into consideration. The empty state view should work as expected now:

Product has neither variations nor attributes:



Product has attributes but no variations:

itsmeichigo:
I have also added a check whether to show guide to create attributes on `ProductVariationsViewModel` and use it to update the empty state screen accordingly. Let me know if that makes sense!

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -14,15 +14,18 @@ final class ProductVariationsViewController: UIViewController {
...
         return .withButton(message: message,
                            image: .emptyBoxImage,
-                           details: "",
-                           buttonTitle: Localization.emptyStateButtonTitle) { [weak self] _ in
+                           details: subtitle,
+                           buttonTitle: buttonTitle) { [weak self] _ in
                             self?.createVariationFromEmptyState()
                            }
-    }()
+    }
```
Ecarrion: Since this configuration now depends on a view model value, I think that having a function that creates it will express the intention better, as you can make the dependencies explicit.

```swift
func createEmptyStateConfig(showAttributeGuide: Bool) -> EmptyStateViewController.Config {
  /// code
}
```

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewModel.swift
```diff
@@ -48,4 +48,10 @@ extension ProductVariationsViewModel {
     func shouldShowMoreButton(for product: Product) -> Bool {
         product.attributesForVariations.isNotEmpty
     }
+
+    /// Defines if empty state screen should show guide for creating attributes
+    ///
+    func shouldShowCreatingAttributesGuide(for product: Product) -> Bool {
+        product.variations.isEmpty && product.attributesForVariations.isEmpty
+    }
```
Ecarrion: I really like to have this kind of logic in the view model. Especially because it allows us to test it easily.

Could you add some unit tests to it? :-)

Ecarrion (changes requested):
Nice improvements! Functionality now seems correct 🎉 
I left a couple of suggestions to keep improving this further.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
itsmeichigo: I think this is a nice suggestion! Will surely update it.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewModel.swift
itsmeichigo: Absolutely!

itsmeichigo: Hey @Ecarrion - I was writing the tests and found a case I have not covered: when the product has variation but no attribute (i.e all the attributes are removed). How should we display the empty state in this case? I'm currently not showing attribute guide:

Ecarrion: Wonderful that you discovered an edge case when writing tests! 🎉 
For the case when the product has "variations but not attributes" I think we should show the "Attribute Guide" as tapping the button will direct you to create attributes.

Can you please update the PR accordingly?
Thanks!

Ecarrion (commented):
Changes are looking 🔝 

Lets just add the changes discussed in https://github.com/woocommerce/woocommerce-ios/pull/4039#discussion_r622445563 and we should be good to go! 🚀

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewModel.swift
itsmeichigo: Thanks for confirming, the test has been updated 🙂

Ecarrion (approved):
Looks good to me! 🚀

PR #4040, opened 2021-04-27
Update product variation list buttons

### LINKED ISSUE #3904: Variations: Update Variations list Buttons
Ref: p91TBi-4O0f-p2

Update variations list screen to:
- Update "Add Variation" button copy
- Move "Edit attributes" action from the navigation bar to a button below the "Add variation" button
- Update buttons to match the designs.

update-variations-list

itsmeichigo:
Closes #3904 

## Description

This PR updates UI of product variation list screen:
- Updated Add Variation button title to Generate New Variation and change its style to primary.
- Removed right navigation button item along with the corresponding action sheet.
- Moved Edit Attributes button out as a separate view beneath Generate New Variation button.
- Moved the top banner to below the buttons following design.

## Testing
This screen can be reached with these steps:
- Select product tab
- Choose to add a new variable product
- Choose variations and create a new variation
- Observe the resulting variations list

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -300,14 +300,14 @@ private extension ProductVariationsViewController {
         let buttonContainer = UIView()
         buttonContainer.backgroundColor = .listForeground
 
-        let editAttributesButton = UIButton()
-        editAttributesButton.translatesAutoresizingMaskIntoConstraints = false
-        editAttributesButton.setTitle(title, for: .normal)
-        editAttributesButton.addTarget(self, action: actionSelector, for: .touchUpInside)
-        stylingHandler(editAttributesButton)
+        let topButton = UIButton()
+        topButton.translatesAutoresizingMaskIntoConstraints = false
+        topButton.setTitle(title, for: .normal)
+        topButton.addTarget(self, action: actionSelector, for: .touchUpInside)
+        stylingHandler(topButton)
 
-        buttonContainer.addSubview(editAttributesButton)
-        buttonContainer.pinSubviewToSafeArea(editAttributesButton, insets: insets)
+        buttonContainer.addSubview(topButton)
+        buttonContainer.pinSubviewToSafeArea(topButton, insets: insets)
 
         if hasBottomBorder {
             let separator = UIView.createBorderView()
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -94,6 +94,7 @@ final class ProductVariationsViewController: UIViewController {
         didSet {
             viewModel.updatedFormTypeIfNeeded(newProduct: product)
 
+            configureRightButtonItem()
             resetResultsController(oldProduct: oldValue)
             updateEmptyState()
             onProductUpdate?(product)
@@ -177,6 +178,22 @@ private extension ProductVariationsViewController {
             "Variations",
             comment: "Title that appears on top of the Product Variation List screen."
         )
+        configureRightButtonItem()
+    }
+
+    /// Configure right button item.
+    ///
+    func configureRightButtonItem() {
+        guard viewModel.shouldShowMoreButton(for: product) else {
+            return navigationItem.rightBarButtonItem = nil
+        }
+
+        let moreButton = UIBarButtonItem(image: .moreImage,
+                                         style: .plain,
+                                         target: self,
+                                         action: #selector(presentMoreOptionsActionSheet(_:)))
+        moreButton.accessibilityLabel = Localization.moreButtonLabel
+        navigationItem.setRightBarButton(moreButton, animated: false)
     }
 
     /// Apply Woo styles.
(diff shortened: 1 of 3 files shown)


Ecarrion (changes requested):
Thanks for this @itsmeichigo!

I reviewed the changes and I don't think we can get rid of the "more button" altogether. We are supposed to still show it in the empty state screen when the product does not have variations but has attributes. Like:


Screen Shot 2021-04-27 at 4 02 14 PM


Also, I noticed that there is a missing divider on the first row. It is probably easy to miss because of the color of the banner view. But it's quite noticeable when that banner is gone.

Screen Shot 2021-04-27 at 4 02 14 PM

itsmeichigo:
Hi @Ecarrion, I have updated my changes to show the missing separator and put the More button back when a product has attributes but no variations (screenshots below). Please review this again - thanks!

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -287,10 +291,26 @@ private extension ProductVariationsViewController {
         addVariationButton.translatesAutoresizingMaskIntoConstraints = false
         addVariationButton.setTitle(Localization.addNewVariation, for: .normal)
         addVariationButton.addTarget(self, action: #selector(addButtonTapped), for: .touchUpInside)
-        addVariationButton.applySecondaryButtonStyle()
+        addVariationButton.applyPrimaryButtonStyle()
 
         buttonContainer.addSubview(addVariationButton)
-        buttonContainer.pinSubviewToSafeArea(addVariationButton, insets: .init(top: 16, left: 16, bottom: 16, right: 16))
+        buttonContainer.pinSubviewToSafeArea(addVariationButton, insets: .init(top: 16, left: 16, bottom: 8, right: 16))
```
Ecarrion: I that these insets are now duplicated. I think it would be good to abstract it in our `Settings` enum. What do you think?

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewModel.swift
```diff
@@ -46,6 +46,6 @@ extension ProductVariationsViewModel {
     /// Defines if the More Options button should be shown
     ///
     func shouldShowMoreButton(for product: Product) -> Bool {
-        product.attributesForVariations.isNotEmpty
+        product.variations.isEmpty && product.attributesForVariations.isNotEmpty
```
Ecarrion, line 49: I think this change is causing tests to fail. Could you make sure tests are correctly updated?
Test: `test_more_button_appears_when_product_is_not_empty()`

Ecarrion (changes requested):
Changes are looking good 🎉 
I left you one tiny suggestion and a request to update our unit tests 😇

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewModel.swift
itsmeichigo: 🤦‍♀️ My bad! I'll check the test again.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
itsmeichigo: Actually, the insets are subtly different between the two buttons - if you look closely you can see that the Add Variation button's bottom space is 8 points, and the Edit Attribute button's top space is 8 points. I did this to make sure the space between the two buttons is 16 points.

But I get the point that there's code duplication and that smells - so I committed a new change to make top button configuration logic reusable - please check that out and let me know what you think!

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -269,36 +268,56 @@ private extension ProductVariationsViewController {
...
+        configureTopButton(title: Localization.addNewVariation,
+                           insets: .init(top: 16, left: 16, bottom: 8, right: 16),
+                           actionSelector: #selector(addButtonTapped),
+                           stylingHandler: { $0.applyPrimaryButtonStyle() })
+
+        configureTopButton(title: Localization.editAttributesAction,
+                           insets: .init(top: 8, left: 16, bottom: 16, right: 16),
+                           hasBottomBorder: true,
+                           actionSelector: #selector(editAttributesTapped),
+                           stylingHandler: { $0.applySecondaryButtonStyle() })
```
Ecarrion: Nit: Since these functions create separate buttons I think they should be renamed to something like 

```swift
addTopButton(title: String,
             insets: UIEdgeInsets,
             hasBottomBorder: Bool = false,
             actionSelector: Selector,
             stylingHandler: (UIButton) -> Void)
```

The current name makes me think that I'm configuring the same top button twice.

What do you think?

WooCommerce/WooCommerceTests/ViewRelated/Products/Variations/ProductVariationsViewModelTests.swift
```diff
@@ -3,7 +3,20 @@ import XCTest
...
+        let viewModel = ProductVariationsViewModel(formType: .edit)
+
+        // When
+        let showMoreButton = viewModel.shouldShowMoreButton(for: product)
+
+        // Then
+        XCTAssertTrue(showMoreButton)
+    }
+
+    func test_more_button_does_not_appear_when_product_is_not_empty() {
```
Ecarrion, line 19: Thanks for adding a new test case!

Ecarrion (commented):
Left a minor name change suggestion and then we should be good to go! 🚀

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
itsmeichigo: Good catch! Let me update that.

Ecarrion (approved):
👏👏

Issue #4049, opened 2021-04-27
[Mobile Payments] Card Reader Settings v2-3: Create CardReaderUnknownViewModel


allendav:
Create CardReaderUnknownViewModel and add to CardReaderSettingsPresentingViewController array of ViewModel.

`shouldShow` returns true if both CardReaderService.connectedReaders and CardReaderService.knownReaders are both empty. Unit test should exercise `shouldShow`

Edit (Apr 28): Don't pass CardReaderService. Have the dependency be on CardPresentPaymentStore

Issue #3559, opened 2021-02-01
Persist chosen reader across app launches and connect to it when seen


allendav:
Building on #3558 this issue will require creating an entity and store to persist a chosen reader. Then, during discovery, if we see a “known” reader, we should stop discovery and connect to that reader. Lastly, forgetting a reader should also purge it from persistence.

p91TBi-4i6-p2

allendav:
@ctarda  - proposed sprints and tickets:

```
MS1 Ticket1 Add feature flag

MS1 Ticket2 Add ReaderID to AppSettingsAction, AppSettingsStore
- Add addKnownReaderID to AppSettingsAction ala something like
	addKnownReader(readerID: String, onCompletion: (Result<Void, Error>) -> Void)
- Add forgetKnownReaderID to AppSettingsAction ala something like
	forgetKnownReader(readerID: String, onCompletion: (Result<Void, Error>) -> Void)
- Add subscribeKnownReaderIDs to AppSettingsAction ala something like
    - subscribeKnownReaders(onCompletion: @escaping ([String]) -> Void))

func subscribeKnownReadersID(onCompletion: @escaping ([String]) -> Void) {
        self.knownReaderIDs.sink { _ in
        } receiveValue: { readerIDs in
            onCompletion(readerIDs)
        }.store(in: &cancellables)
    }

MS1 Ticket3 Integrate AppSettings ReaderID with CardReaderSettingsUnknownViewModel and CardReaderSettingsConnectedViewModel
- When we connect to a reader, CardReaderSettingsUnknownViewModel should dispatch addKnownReader
- When CardReaderSettingsConnectedViewModel disconnects from a reader, dispatch forgetKnownReader
- Change CardReaderSettingsUnknownViewModel to dispatch the observeKnownReaders from AppSettingsStore not CardPresentPaymentStore

MS2 Ticket1 Create new CardReaderSettingsKnownViewModel to handle auto-connect if a known reader is discovered
- Add CardReaderSettingsKnownViewModel
    - why? we don’t want to pollute CardReaderSettingsUnknownViewModel with a second concern, and the flows differ enough when we know about a card reader versus when we don’t. This also opens up the opportunity to display tailored content for merchants that have already connected to a card reader successfully vs those doing it for the first time
    - shouldShow = true if noKnownReaders is false and noConnectedReaders is true
    - has a CardReaderSettingsKnownViewModelDiscoveryState
[...]

PR #4060, opened 2021-04-27
[Mobile Payments] Add (empty) knownReaders

### LINKED ISSUE #4047: [Mobile Payments] Card Reader Settings v2-1: Card Reader Service Known Readers
Add a publisher that emits a list of knownReaders. It will emit an empty list initially.

Edit: See also https://github.com/woocommerce/woocommerce-ios/issues/3559

allendav:
Closes #4047 

- Adds a `loadKnownReaders` action to our CardPresentPaymentsStore. We'll need this for the viewmodel to be added in #4049
- Note:  Returns an empty array presently. Actually "knowing" and "unknowing" readers will come in #3559 

To test:
- No user facing functionality
- Simply confirm all unit tests continue to pass

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardReaderService.swift
+++ b/Hardware/Hardware/CardReader/CardReaderService.swift
@@ -10,10 +10,6 @@ public protocol CardReaderService {
     /// The Publisher that emits the connected readers
     var connectedReaders: AnyPublisher<[CardReader], Never> { get }
 
-    /// The Publisher that emits the list of known readers. Known readers are readers that the merchant has
-    /// connected to in the past.
-    var knownReaders: AnyPublisher<[CardReader], Never> { get }
-
     /// The Publisher that emits the service status
     var serviceStatus: AnyPublisher<CardReaderServiceStatus, Never> { get }
 
--- a/Hardware/Hardware/CardReader/CardReaderService.swift
+++ b/Hardware/Hardware/CardReader/CardReaderService.swift
@@ -10,6 +10,10 @@ public protocol CardReaderService {
     /// The Publisher that emits the connected readers
     var connectedReaders: AnyPublisher<[CardReader], Never> { get }
 
+    /// The Publisher that emits the list of known readers. Known readers are readers that the merchant has
+    /// connected to in the past.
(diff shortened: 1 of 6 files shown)


Hardware/Hardware/CardReader/CardReaderService.swift
```diff
@@ -10,6 +10,10 @@ public protocol CardReaderService {
     /// The Publisher that emits the connected readers
     var connectedReaders: AnyPublisher<[CardReader], Never> { get }
 
+    /// The Publisher that emits the list of known readers. Known readers are readers that the merchant has
+    /// connected to in the past.
+    var knownReaders: AnyPublisher<[CardReader], Never> { get }
```
allendav: So, there are things in `CardReader` that might not make sense to persist, like battery level. But there are a few things that are interesting. Since we are not adding or removing known readers in this PR, that's ok. A future PR should be able to safety chose a different structure here with minimal impact to the subscribing code.

ctarda:
I am curious. As far as I recall, the concept of "known reader" is not something that the terminal SDK provides. 

How do we decide if a card reader is known? It seems to me like this logic (deciding what is a "known reader") should belong in the business logic layer. What am I missing?

Hardware/Hardware/CardReader/CardReaderService.swift
ctarda: > Known readers are readers that the merchant has connected to in the past.

How do we know the merchant has already connected to a particular reader?

allendav: Good point. I've captured this in https://github.com/woocommerce/woocommerce-ios/issues/4047#issuecomment-828580028 and we'll move this concept to the store instead

allendav:
@ctarda - I've updated this PR to define an action instead of modifying the CardReaderService. Please let me know what you think. https://github.com/woocommerce/woocommerce-ios/pull/4060/commits/282a3bd7dbd5a33fd66e01d0849fde036499b80e

ctarda (approved):
Thanks for iterating on this @allendav 

:shipit: 🚀

PR #4061, opened 2021-04-28
Add helper text for product variations in Product images screen

### LINKED ISSUE #3933: Variations: Add image helper text
It's not very clear that variable products can have only one image, to help with possible confusion
, let's add a helper above the photo list screen.

Current | Expected(Android design)
--- | ---
ios-no-text | android-text

itsmeichigo:
Closes #3933 

## Description
This PR updates the Product images screen by adding a helper text about the image list of product variations. This means the Product images screen should stay intact for products that allow multiple photos.

| Normal products | Product variations |
| ------- | ------- |
|  |  |

## Solution
- Embed the content of top container view (the view containing top button and separator view) inside a vertical stack view. This makes it easier to show / hide the helper text.
- Insert the helper view below the separator view. This is basically a UIView containing a UILabel that allows for adding margins around the label.
- Configure the helper view by checking conditions: (1) whether the product does not allow multiple images and (2) the product is of type variable. If these 2 conditions are met, the helper text should be visible.

## Discussion
Since the helper text content is now specific to product variation, I had to check whether product is of type variable. If in the future the text is more generic, the check for allowing multiple images is sufficient.

## Testing
This can be checked by visiting the Product images through different flows:
- When editing products, select any image in the list or choose to add new image in the main edit screen. The resulting image list screen should not show the helper text.
- Create or edit a variable product, choose a variation to add or edit its photo. The helper text should be shown in this case. 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Media/ProductImagesViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Media/ProductImagesViewController.swift
@@ -13,7 +13,7 @@ final class ProductImagesViewController: UIViewController {
     @IBOutlet private weak var imagesContainerView: UIView!
     @IBOutlet private weak var helperContainerView: UIView!
     @IBOutlet private weak var helperLabel: UILabel!
-    
+
     private let siteID: Int64
     private let productID: Int64
     private let product: ProductFormDataModel
@@ -96,6 +96,7 @@ final class ProductImagesViewController: UIViewController {
         configureNavigation()
         configureAddButton()
         configureAddButtonBottomBorderView()
+        configureHelperViews()
         configureImagesContainerView()
         configureProductImagesObservation()
         handleSwipeBackGesture()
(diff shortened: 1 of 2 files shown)


ealeksandrov (approved):
LGTM :shipit:

ealeksandrov (approved):
Still LGTM :shipit:

Ecarrion:
@woocommerce/mobile-designers We only have an android design for this change.

android-text

Could you confirm if the font/style @itsmeichigo is proposing is ok? or if you would want something different?

Screen Shot 2021-04-28 at 10 20 58 AM

Thanks!

Ecarrion (approved):
Changes look good to me too! 🎉 

Sorry @ealeksandrov for dismissing your review. I miss-tapped the button when trying to request a review from the design team! 🤦 

Let's wait on their opinion before merging!

Garance91540 (commented):
Hi there! One small change request: can we make the text size iOS/Footnote?

Thanks! https://d.pr/i/BpPM4G

itsmeichigo:
Thank you @Garance91540 - below is the screenshot of the screen after the font update. However, the copy is different from your design, should I update that as well?

Garance91540:
Yes if possible let's update the copy too, and the right and left padding (should be 16px). Thanks!

Issue #3919, opened 2021-04-06
Shipping Labels M2: Invalid address banner seems to always be shown on edit address screen (even with valid address)


rachelmcr:
**Describe the bug**

I'm not 100% sure that this is an issue (I might be doing something wrong with the addresses I'm using) but I think we're always showing the invalid address banner on the edit address screen, even if the address _is_ valid.

The way the banner was implemented in https://github.com/woocommerce/woocommerce-ios/pull/3731, it should only appear if there is a problem with the address. However, even after I select a suggested address, if I go to the edit address screen I see that top banner.

**To Reproduce**

1. Go to the Orders > Processing tab.
2. Select an order.
3. Select "Create Shipping Label."
4. Select "Continue" in the "Ship from" section.
5. If the "Ship from" address is accepted, the "Continue" button will disappear. (If the "Ship from" address isn't accepted, you'll end up on the suggested address screen. Select "Use Suggested Address" — this will take you back to the Create Shipping Label screen.)
6. Tap on the "Ship from" section to open the edit address screen.
7. Notice that even though you have a valid, accepted origin address, the top banner appears.

**Screenshots**

<img src="https://user-images.githubusercontent.com/8658164/113737492-f9fa0c80-96f5-11eb-8507-1cf5107a5068.gif" width="300px">

koke:
It does look like a bug to me. What I'm seeing is that the error shows up because the address is only being validated when you tap on the Done button. But even then, if you come back to the screen, it will show the banner again, since it's a new view that hasn't validated.

I'm still not fully understanding the validation flow, but I'll investigate more

PR #4063, opened 2021-04-28
Shipping Labels M2: Refactor address validation result to encode validation errors as Swift errors

koke:
Started as part of #3919, but these changes are independent from the bug fix, although they will make the future solution simpler.

## Why

The address validation API returns a response that might either contain an address (with possible suggestions), or an error if the validation failed. The current implementation maps this JSON response literally, and exposes these fields as optionals in a `ShippingLabelAddressValidationResponse` model.

This means that the app code needs to do a lot of `Optional` wrangling to deal with the validation response, which currently happens in two different places, which involved duplicate logic.

## How

The networking layer already uses `Result` to deal with errors, so this PR will move that logic for dealing with all the possible response combinations from the UI to the Network layer.

- The existing `ShippingLabelAddressValidationResponse` remains as an internal model whose job is basically parsing the response and turning it into a `Result<ShippingLabelAddressValidationSuccess, ShippingLabelAddressValidationError>`
- There is a new `ShippingLabelAddressValidationSuccess` model, which contains the data in case the validation was successful. This includes both the address and the `isTrivialNormalization` flag.
- The logic in the view models is then simplified, as it only needs to deal with the potential `Result` cases. To differentiate validation errors from generic network errors, it checks whether the returned error is a `ShippingLabelAddressValidationError`.
- Additionally, the view model handling of results was rewritten using pattern matching

Update release notes:


Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationResponse.swift
```diff
@@ -2,26 +2,29 @@ import Foundation
...
-        self.isTrivialNormalization = isTrivialNormalization
+    init(address: ShippingLabelAddress?, errors: ShippingLabelAddressValidationError?, isTrivialNormalization: Bool?) {
+        if let errors = errors {
+            result = .failure(errors)
+        } else if let address = address,
+                  let isTrivialNormalization = isTrivialNormalization {
+            result = .success(.init(address: address, isTrivialNormalization: isTrivialNormalization))
+        } else {
+            // This case should never happen, but that's not guaranteed.
+            // We'll treat the absence of both an address and errors as an error with no message.
```
pmusolino, line 20: In reality it can happen for all the errors related to HTTP, like all the errors `500`

Networking/Networking/Mapper/ShippingLabelAddressValidationResponseMapper.swift
```diff
@@ -6,9 +6,12 @@ import Foundation
 struct ShippingLabelAddressValidationResponseMapper: Mapper {
```
pmusolino: nit: maybe it's worth renaming this mapper as `ShippingLabelAddressValidationSuccessMapper`?

pmusolino (approved):
Looks good to me :shipit: thanks @koke for working on this! 
Just some nits, but it's ready to be 🚢 ed

Networking/Networking/Mapper/ShippingLabelAddressValidationResponseMapper.swift
koke: Good catch, that makes sense

Networking/Networking/Model/ShippingLabel/ShippingLabelAddressValidationResponse.swift
koke: I don't think so, since this initializer is only ever called from the JSON decoding one, which means we did get a JSON response and not a network error.

Issue #4051, opened 2021-04-27
[Mobile Payments] Card Reader Settings v2-5: cardReadersV2WasPressed


allendav:
Add a new menu item (in debug builds) to Dashboard SettingsViewController. Have it fire cardReadersV2WasPressed which should instantiate CardReaderSettingsPresentingViewController and have it:

- register the CardReaderUnknownViewModel
- register the CardReaderConnectedViewModel
- immediately evaluate view to present (first viewModel to return shouldShow true)
- nitially it will display just the name of the ViewController it would present

Edit (Apr 28): Instead of having CardReaderSettingsPresentingViewController register the viewmodels, have those viewmodels instantiated outside CardReaderSettingsPresentingViewController and passed into it. Dependency injection will support unit testing.

Issue #4058, opened 2021-04-27
[Mobile Payments] Card Reader Settings v2-12: Remove v1 Dashboard SettingsViewController cardReadersWasPressed


allendav:
Remove v1 Dashboard SettingsViewController cardReadersWasPressed and just keep the new v2 one

PR #4066, opened 2021-04-28
[Mobile Payments] Card Reader Settings V2 2: Define ViewModel protocols, add Presenting View Controller

### LINKED ISSUE #4048: [Mobile Payments] Card Reader Settings v2-2: Define ViewModel and PresentingViewController
Next step in Card Reader Settings v2: Create VM protocol and CardReaderSettingsPresentingViewController. CardReaderSettingsPresentingViewController has an array of ViewModel, String tuples. The String will be used to hold the name of the ViewController related to the ViewModel that CardReaderSettingsPresentingViewController should present when the viewModels shouldShow returns true

allendav:
Closes #4048

This is the next step in launching a more SOLID approach to card reader settings. It:

- Adds a protocol for view models to implement and a presenting view controller that accepts an array of a struct of those viewmodels and their related views (class names) to show
- Protocol defines `shouldShow` which will be used in #4051 to actually determine which view to push

To test:
- Navigate to Settings, note the new V2 row. Press it. That's all for now.
- Navigate to Settings, tap on the old row. Press it. It should work as before

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentedViewViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentedViewViewModel.swift
@@ -4,5 +4,7 @@ protocol CardReaderSettingsPresentedViewModel {
     func shouldShow() -> Bool
 }
 
-/// A tuple containing a reference to the view model for a presented view and the related view controller's class name.
-typealias CardReaderSettingsViewModelAndView = (CardReaderSettingsPresentedViewModel, String)
+struct CardReaderSettingsViewModelAndView {
+    var viewModel: CardReaderSettingsPresentedViewModel
+    var viewIdentifier: String
+}
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentedViewViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentedViewViewModel.swift
@@ -1,0 +1,8 @@
+import Foundation
+
+protocol CardReaderSettingsPresentedViewModel {
+    func shouldShow() -> Bool
+}
+
+/// A tuple containing a reference to the view model for a presented view and the related view controller's class name.
+typealias CardReaderSettingsViewModelAndView = (CardReaderSettingsPresentedViewModel, String)
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
@@ -1,0 +1,42 @@
+import Foundation
+import UIKit
+
+final class CardReaderSettingsPresentingViewController: UIViewController {
+
+    /// An array of viewModels and related view classes
+    private var viewModelsAndViews = [CardReaderSettingsViewModelAndView]()
+
+    /// Set our dependencies
+    func configure(viewModelsAndViews: [CardReaderSettingsViewModelAndView]) {
+        self.viewModelsAndViews = viewModelsAndViews
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        configureBackground()
+        configureNavigation()
+    }
+}
+
+// MARK: - View Configuration
+//
+private extension CardReaderSettingsPresentingViewController {
+
+    private func configureBackground() {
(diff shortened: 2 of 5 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -379,6 +391,15 @@ private extension SettingsViewController {
         show(viewController, sender: self)
     }
 
+    func cardReadersV2WasPressed() {
+        ServiceLocator.analytics.track(.settingsCardReadersTapped)
+        guard let viewController = UIStoryboard.dashboard.instantiateViewController(ofClass: CardReaderSettingsPresentingViewController.self) else {
+            fatalError("Cannot instantiate `CardReaderSettingsPresentingViewController` from Dashboard storyboard")
+        }
+        viewController.configure(viewModelsAndViews: [])
```
allendav, line 399: In a follow on issue, I think we'll want to define a DependencyContainer object which registers each of the viewmodels (and related view class names) and can serve as a source for the viewModelsAndViews here.

No need for SettingsViewController to know all those details.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -253,6 +259,12 @@ private extension SettingsViewController {
         cell.textLabel?.text = NSLocalizedString("Manage Card Reader", comment: "Navigates to Card Reader management screen")
     }
 
+    func configureCardReadersV2(cell: BasicTableViewCell) {
+        cell.accessoryType = .disclosureIndicator
+        cell.selectionStyle = .default
+        cell.textLabel?.text = NSLocalizedString("Manage Card Reader (V2)", comment: "Navigates to Card Reader management screen")
```
allendav, line 265: We'll remove the "V2" the same time we remove the original implementation in #4058

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentedViewViewModel.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+
+protocol CardReaderSettingsPresentedViewModel {
+    func shouldShow() -> Bool
+}
+
+/// A tuple containing a reference to the view model for a presented view and the related view controller's class name.
```
ctarda: Would you consider declaring this a a struct? Structs are super cheap, and you could make it implement Equatable/Comparable which might help clean code that looks up into the array of these later.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
```diff
@@ -0,0 +1,42 @@
+import Foundation
+import UIKit
+
+final class CardReaderSettingsPresentingViewController: UIViewController {
+
+    /// An array of viewModels and related view classes
+    private var viewModelsAndViews = [CardReaderSettingsViewModelAndView]()
+
+    /// Set our dependencies
+    func configure(viewModelsAndViews: [CardReaderSettingsViewModelAndView]) {
```
ctarda, line 10: I am not sure if this makes actual sense, but if if were possible to inject this via the constructor, we could get away with declaring viewModelsAndViews as a let instead of a var

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
ctarda: Makes sense. Also, it might make sense to encapsulate the details abut how the view controller is instantiated in a static method in the view controller itself.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
ctarda: Whoops, storyboards. Scratch my comment out please 😊

ctarda (approved):
👏🏼 👏🏼 A couple of questions / suggestions, but nothing blocking.

:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentedViewViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4066/commits/d333edc07151bed8e136534cd197070871ceff57

PR #4070, opened 2021-04-29
[6.5 Backport] Prevent crash when the media library has exactly the same amount of items as the request page size

### LINKED ISSUE #3369: WordPressMediaLibraryPickerDataSource.retrieveMedia upperBound EXC_BREAKPOINT

koke:
Backporting #4003 to 6.5, copying the description from there:

Fixes #3369

## Description

When the media library contains exactly the amount of items defined as the page size (currently 25) the app would crash. In theory, this would also happen with any exact multiple of that page size.

The problem is that when we have a full page and request the next one, the response will have no items. The previous code tried to generate a range of indices to update that was invalid because the end was lower than the start. Instead, if there is no new data to update, we can bail early and skip the callback.

## Testing

1. Ensure your media library has exactly 25 images
2. Create or tap into a product, then tap Add a product image
3. Tap Add Photos > WP Media Library
4. The app shouldn't crash

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,6 +7,7 @@
 6.5
 -----
 - [*] Fix: Product images with non-latin characters in filenames now will load correctly and won't break Media Library. [https://github.com/woocommerce/woocommerce-ios/pull/3935]
+- [*] Fix: The screen to select images from the Media Library would sometimes crash when the library had a specific number of images. [https://github.com/woocommerce/woocommerce-ios/pull/4070]
 
 6.4
 -----
--- a/WooCommerce/Classes/ViewRelated/Products/Media/WordPressMediaLibraryPickerDataSource.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Media/WordPressMediaLibraryPickerDataSource.swift
@@ -177,6 +177,12 @@ private extension WordPressMediaLibraryPickerDataSource {
     /// Appends the new media items to the existing media items.
     ///
     func updateMediaItems(_ newMediaItems: [Media], pageNumber: Int, pageSize: Int) {
+        // If the response contains no new items, there is nothing to update.
+        // We return early since the code would generate an invalid range of indices to update otherwise.
+        guard newMediaItems.isNotEmpty else {
+            return
+        }
+
         let pageFirstIndex = syncingCoordinator.pageFirstIndex
         let mediaStartIndex = (pageNumber - pageFirstIndex) * pageSize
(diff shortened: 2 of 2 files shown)


RELEASE-NOTES.txt
```diff
@@ -7,6 +7,7 @@
 6.5
 -----
 - [*] Fix: Product images with non-latin characters in filenames now will load correctly and won't break Media Library. [https://github.com/woocommerce/woocommerce-ios/pull/3935]
+- [*] Fix: The screen to select images from the Media Library would sometimes crash when the library had a specific number of images. [https://github.com/woocommerce/woocommerce-ios/pull/4070]
```
AliSoftware, line 10: Note: The Release Notes won't get picked up when doing a PR targeting `release/*`, because those release notes have already been sent to Editorial middle of last week, then to our translation vendor last Sunday.

I'm ok keeping that change still (so we have the history of what was shipped in 6.5 for ourselves internally), just wanted to point it out to be sure you knew.

AliSoftware (approved):
Haven't tested it but looking at the code and given it's the same as https://github.com/woocommerce/woocommerce-ios/pull/4003 which has been tested and approved, I'm confident approving this directly.

RELEASE-NOTES.txt
koke: I was wondering about that, and what will happen when the 6.5 branch gets merged into develop. The idea was to have that item in 6.5 just to have the history of the changes, but now it might appear in both after merging? 😬

AliSoftware: Yeah it will appear in both indeed 🤔  🤷  🙃

PR #4071, opened 2021-04-29
Shipping Labels M2: Fix address validation banner showing by default even if there are no errors

### LINKED ISSUE #3919: Shipping Labels M2: Invalid address banner seems to always be shown on edit address screen (even with valid address)

koke:
Fixes #3919 
Requires #4063 to be merged first

## Why

As reported on #3919, the address form will show a validation error when presented even if the address is correct. The cause is that it seems to assume that it was presented because there was an error.

In reality there are two scenarios where this might get presented:
1. If there is an error validating after the user taps Continue on the "Create Shipping Label" screen
2. If the validation has been successful previously but the user taps on the address row again (for instance, to edit a valid address)

There were two problems specifically:

1. The view model was configuring the alert as visible by default, and only updating its visibility after performing address validation when the user tapped on Done.
2. There was no way for the previous screen to communicate the result of a previous validation.

## How

This PR adds the ability for the address view to be initialized with an error, so the view can differentiate the two scenarios.
Also, it will update the state of the alert when it's loaded, and hide it if there is no known validation error.

One potential improvement I didn't add here was that, since the previous view already knows if the address was validated successfully, the address screen could skip validation if the address hasn't changed. This seems useful but also a different issue from fixing the original bug.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -52,8 +52,14 @@ final class ShippingLabelAddressFormViewController: UIViewController {
 
     /// Init
     ///
-    init(siteID: Int64, type: ShipType, address: ShippingLabelAddress?, completion: @escaping Completion) {
-        viewModel = ShippingLabelAddressFormViewModel(siteID: siteID, type: type, address: address)
+    init(
+        siteID: Int64,
+        type: ShipType,
+        address: ShippingLabelAddress?,
+        validationError: ShippingLabelAddressValidationError?,
+        completion: @escaping Completion
+    ) {
+        viewModel = ShippingLabelAddressFormViewModel(siteID: siteID, type: type, address: address, validationError: validationError)
         onCompletion = completion
         super.init(nibName: nil, bundle: nil)
     }
@@ -69,6 +75,7 @@ final class ShippingLabelAddressFormViewController: UIViewController {
         configureTableView()
         observeViewModel()
         configureConfirmButton()
+        updateTopBannerView()
         keyboardFrameObserver.startObservingKeyboardFrame(sendInitialEvent: true)
     }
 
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -52,8 +52,14 @@ final class ShippingLabelAddressFormViewController: UIViewController {
 
     /// Init
     ///
-    init(siteID: Int64, type: ShipType, address: ShippingLabelAddress?, completion: @escaping Completion) {
-        viewModel = ShippingLabelAddressFormViewModel(siteID: siteID, type: type, address: address)
+    init(
+        siteID: Int64,
+        type: ShipType,
+        address: ShippingLabelAddress?,
+        validationError: ShippingLabelAddressValidationError?,
```
rachelmcr, line 59: Just out of curiosity, is there a reason why we wouldn't give this parameter a default `nil` value? I was thinking that would simplify things so we wouldn't have to explicitly set it to `nil` in the various places where the address is successfully validated.

rachelmcr (approved):
Looks good and tests well! :shipit: 

In the process of testing I noticed an issue with the banner, unrelated to this PR, reported in https://github.com/woocommerce/woocommerce-ios/issues/4095.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
koke: Not really. I think I initially did that, but then it felt a bit better to do it explicitly. In the end I didn't have a strong preference towards either option in this case.

koke:
> In the process of testing I noticed an issue with the banner, unrelated to this PR

Oh good catch! I think I spotted that once but I completely forgot about it 😬

PR #4072, opened 2021-04-29
Coupon Management: List coupons - add view placeholder

joshheald-a8c:
Part of #3916

## Description
This PR adds the placeholder view and feature flag for the Manage Coupons list view.

## Testing
Open the alpha or local developer version of the app, and navigate to the settings menu.
A "Manage Coupons" row should be present in the Store Settings section
Tapping this row takes the user to a "Coming soon" screen

Doing the above on a release build should not show the "Manage Coupons" row in settings.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Analytics/WooAnalyticsStat.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalyticsStat.swift
@@ -121,6 +121,7 @@ public enum WooAnalyticsStat: String {
     case settingsSelectedStoreTapped            = "settings_selected_site_tapped"
     case settingsContactSupportTapped           = "main_menu_contact_support_tapped"
     case settingsCardReadersTapped              = "settings_card_readers_tapped"
+    case settingsCouponManagementTapped         = "settings_coupon_management_tapped"
 
     case settingsBetaFeaturesButtonTapped       = "settings_beta_features_button_tapped"
     case settingsBetaFeaturesProductsToggled    = "settings_beta_features_products_toggled"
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/Analytics/WooAnalyticsStat.swift
```diff
@@ -121,6 +121,7 @@ public enum WooAnalyticsStat: String {
     case settingsSelectedStoreTapped            = "settings_selected_site_tapped"
     case settingsContactSupportTapped           = "main_menu_contact_support_tapped"
     case settingsCardReadersTapped              = "settings_card_readers_tapped"
+    case settingsCouponManagementTapped         = "settings_coupon_management_tapped"
```
pmusolino: I think that you can remove this type of tracking. Usually, we add the tracking at the end of a project and we discuss it also with Android developers. So feel free to ignore everything regarding the tracking. 👍

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
```diff
@@ -0,0 +1,17 @@
+import UIKit
+
+class CouponManagementViewController: UIViewController {
```
pmusolino: nit: 
```suggestion
final class CouponManagementViewController: UIViewController {
```

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
```diff
@@ -0,0 +1,17 @@
...
+        configureNavigation()
+    }
+}
+
+// MARK: - View Configuration
+//
+private extension CouponManagementViewController {
+
+    func configureNavigation() {
+        title = NSLocalizedString("Coupons", comment: "Coupon management coupon list screen title")
```
pmusolino: What do you think about moving this localized string under an enum called `Localization`? We use it in many points of the app.

pmusolino (approved):
Looks good to me, just 3 small nits 🚢

pmusolino:
Thanks @joshheald-a8c for the changes 💯 I'm going to merge it right now.

PR #4073, opened 2021-04-29
Shipping Labels: Display the last selected package in Package Details

### LINKED ISSUE #3940: Shipping Labels M2: display the last package selected

rachelmcr:
Closes: #3940

## Description

This PR sets the shipping label package as the last selected package (from the API) by default. This displays the package name in the Package Selected field in the Package Details screen and sets it as the selected package (so it is already selected when you open the packages list).

## Changes

* Adds `syncShippingLabelAccountSettings` to the shipping label form screen, to sync the shipping label account settings. These settings include the last selected package ID, as well as other settings we'll use later in the shipping label flow (including payment methods).
* Moves `syncPackageDetails` to the shipping label form screen (from package details) so that package details are already synced when we open the Package Details screen, and passes the response to the Package Details VM.
* Adds `accountSettingsResultsController` to retrieve the shipping label account settings from storage on the Package Details screen.
* Sets the selected package ID to the last selected package ID with `setDefaultPackage`, so the last selected package is selected and displayed.
* Updates the Package Details screen to display `selectedPackageName` — previously it displayed `selectedPackageID`, which isn't the same as the name (title) for predefined packages.

## Testing

Setup:

* If you haven't purchased a shipping label on your store before, the Package Selected field will be blank. You can test first with this scenario and then test after purchasing a shipping label.
* You can follow the internal testing guide at PCYsg-wev-p2 to purchase a shipping label before testing. Alternately, you can use a proxy app like Charles Proxy and edit the API response from the `/wc/v1/connect/account/settings` endpoint to fake a last selected package value; with this approach, enter one of your existing package IDs as the `last_box_id` in `userMeta` in that response.

To test:

2. Open an order detail.
4. Press the button "Continue" in Create Shipping Label form, under the `Ship From` cell. Confirm the address.
5. Press the button "Continue" in Create Shipping Label form, under the `Ship To` cell. Confirm the address.
6. Press the button "Continue" in Create Shipping Label form, under the `Package Details` cell.
7. Confirm the `Package Selected` field on the Package Details screen displays your last selected package by default.
8. Tap on Package Selected and confirm the last selected package is selected in the package list.
9. Select a different package and tap Done. Confirm `Package Selected` now shows the new package selection.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -48,6 +48,20 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
     /// The id of the selected package. Defaults to last selected package, if any.
     ///
     @Published var selectedPackageID: String?
+
+    /// The title of the selected package, if any.
+    ///
+    var selectedPackageName: String {
+        if let selectedCustomPackage = selectedCustomPackage {
+            return selectedCustomPackage.title
+        }
+        else if let selectedPredefinedPackage = selectedPredefinedPackage {
+            return selectedPredefinedPackage.title
+        }
+        else {
+            return ""
+        }
+    }
     @Published private(set) var selectedCustomPackage: ShippingLabelCustomPackage?
     @Published private(set) var selectedPredefinedPackage: ShippingLabelPredefinedPackage?
 
@@ -58,6 +72,7 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
     }
 
     init(order: Order,
+         packagesResponse: ShippingLabelPackagesResponse?,
          formatter: CurrencyFormatter = CurrencyFormatter(currencySettings: ServiceLocator.currencySettings),
          stores: StoresManager = ServiceLocator.stores,
          storageManager: StorageManagerType = ServiceLocator.storageManager,
@@ -69,10 +84,11 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.stores = stores
         self.storageManager = storageManager
         self.weightUnit = weightUnit
+        self.packagesResponse = packagesResponse
         configureResultsControllers()
         syncProducts()
         syncProductVariations()
-        syncPackageDetails()
+        setDefaultPackage()
     }
 
     private func configureResultsControllers() {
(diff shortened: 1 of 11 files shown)


koke:
I tried testing this (without any previous purchases), and when I tap on Package Selected I just get a blank screen, not sure yet why 🤔 

Simulator Screen Shot - iPhone 12 - 2021-04-30 at 08 46 34

koke:
> when I tap on Package Selected I just get a blank screen

OK, I just realized I did not have any packages set up in the backend

Screen Shot 2021-04-30 at 10 04 53

I imagine we'll have to add some explanatory empty state for that case, right?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -45,9 +45,23 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
...
     @Published private(set) var itemsRows: [ItemToFulfillRow] = []
 
-    /// The id of the selected package, if any
+    /// The id of the selected package. Defaults to last selected package, if any.
     ///
     @Published var selectedPackageID: String?
+
+    /// The title of the selected package, if any.
+    ///
+    var selectedPackageName: String {
```
koke, line 54: Not really specific to this PR, but I found a bit surprising to have two `selected` properties for the custom and predefined packages, since technically the model could get in a state where both are populated. It would be nice to have a base `ShippingLabelPackage` prototype for both types to simplify the logic a bit.

koke (approved):
After setting up my store correctly 😅 I tested this and works as expected 🚢

rachelmcr:
> OK, I just realized I did not have any packages set up in the backend
> I imagine we'll have to add some explanatory empty state for that case, right?

I was thinking that we'd handle this in Milestone 4 when we add support for creating packages in the app. But now that you mention it, I'm not sure if we'll be releasing this to users before M4 and in any case I'm not sure if we have a design for the empty state. I opened https://github.com/woocommerce/woocommerce-ios/issues/4092 to keep track of that.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
rachelmcr: Agreed! I wanted to do that to simplify the logic for this, but it ended up being a little too much to do all at once. It would be nice to do that, though.

PR #4081, opened 2021-04-29
[Mobile Payments] Add VM for the no-known no-connected readers state

### LINKED ISSUE #4049: [Mobile Payments] Card Reader Settings v2-3: Create CardReaderUnknownViewModel

allendav:
Closes #4049 

**Note: This is against feature/stripe-terminal-sdk-integration, not develop**

Changes:
- This is the first viewmodel in a series of viewmodels for card reader settings
- Modifies the protocol to make the "shouldShow" a readonly property
- Adds a didChangeShouldShow optional callback that can be used by the VM to tell its creator shouldShow changed
- Adds a new action to subscribe to changes in the list of connected readers
- Fires actions to subscribe to the list of known readers and the list of connected readers
- Updates shouldShow based on those subscriptions
- Unit tests, including a new MockCardPresentPaymentsStoresManager

Rationale:
- It might have been interesting to use Combine and have our VMs Publish shouldShow, but Swift doesn't allow for @Published vars in protocols (yet) https://danielbernal.co/combine-and-protocols-in-swift/
- Conceivably, if we started having stores Publish things, it would be easy to modify a VM like what I have here to Observe things
- I am assuming this idea of having a completion called repeatedly will actually result in the completion being called repeatedly (or at least as long as the VM exists). Or is there something special I need to do that I missed to ensure that happens?

Work not yet done:
- Add this VM to the array of ViewModels passed to CardReaderSettingsPresentingViewController - I'm going to defer that to #4051 since this PR is already > 200 lines

To test:
- Ensure all unit tests pass, especially CardReaderSettingsUnknownViewModelTests

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewModel.swift
@@ -1,0 +1,59 @@
+import Foundation
+import Yosemite
+
+final class CardReaderSettingsUnknownViewModel: CardReaderSettingsPresentedViewModel {
+    var shouldShow: Bool
+    var didChangeShouldShow: ((Bool) -> Void)?
+
+    private var noConnectedReaders: Bool
+    private var noKnownReaders: Bool
+
+    init(didChangeShouldShow: ((Bool) -> Void)?) {
+
+        // Default shouldShow to true, since we don't know if there are any known or connected readers
+        // TODO: Decide if we want a flag to reflect this initial state and use it to drive a loading indicator
+
+        self.shouldShow = true
+        self.didChangeShouldShow = didChangeShouldShow
+        self.noConnectedReaders = true
+        self.noKnownReaders = true
+
+        beginObservation()
+    }
+
+    /// Dispatches actions to the CardPresentPaymentStore so that we can monitor changes to the list of known
+    /// and connected readers.
+    ///
+    private func beginObservation() {
+
+        // This completion should be called repeatedly as the list of known readers changes
+        let knownAction = CardPresentPaymentAction.observeKnownReaders() { [weak self] readers in
+            self?.noKnownReaders = readers.isEmpty
+            self?.reevaluateShouldShow()
+        }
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewModel.swift
```diff
@@ -0,0 +1,59 @@
...
+        // This completion should be called repeatedly as the list of known readers changes
+        let knownAction = CardPresentPaymentAction.observeKnownReaders() { [weak self] readers in
+            self?.noKnownReaders = readers.isEmpty
+            self?.reevaluateShouldShow()
+        }
+        ServiceLocator.stores.dispatch(knownAction)
+
+        // This completion should be called repeatedly as the list of connected readers changes
+        let connectedAction = CardPresentPaymentAction.observeConnectedReaders() { [weak self] readers in
+            self?.noConnectedReaders = readers.isEmpty
```
ctarda: This is a personal preference, so please take it as that (meaning, feel free to ignore this)

I find the self? a bit annoying, specially where there are several lines attacking self? so I tend to do something like 

```
self.oneThing()
self.anotherThing()
```

I could argue that's also a bit of a "performance" improvement, as the execution would short-circuit if there is no valid self at that point, but meh, that'd be a pretty weak argument anyway. 🤷🏼

allendav: Updated in https://github.com/woocommerce/woocommerce-ios/pull/4081/commits/06a4be7721bde38a2117f137a6b4b08874bb237f

allendav:
@ctarda this is ready for another round

ctarda (approved):
:shipit:

PR #4084, opened 2021-04-30
Coupon Management: Move Settings screen strings to Localization enum

joshheald-a8c:
## Description
Since I'd added an inline NSLocalizedString to the settings screen in #4072, and the code review feedback included adding a Localization enum on the new screen I added, I thought it would be worthwhile to also make this change within the settings screen.

## Testing
This has no user-visible changes, it's a refactor only.

Open the app, and go to the settings screen
Check that all strings are present as they were before this change.
Tap Log out, and check that the alert strings are also still correct there.

Update release notes:


pmusolino (approved):
Thank you so much @joshheald-a8c for this extra effort ;) 

LGTM :shipit:

PR #4096, opened 2021-04-30
Documentation: Add info about rake generate to Networking doc

rachelmcr:
## Description

Although we already have documentation about how to use `rake generate` to generate `fake()` and `copy()` methods, there isn't any reference to those docs from higher-level docs. I overlooked that step when making Networking changes, since I wasn't aware it was needed, so I'd suggest adding links from the Networking doc just for awareness.

Any suggested changes to the placement/wording are welcome!

## Submitter Checklist

Update release notes:

### DIFF
--- a/docs/NETWORKING.md
+++ b/docs/NETWORKING.md
@@ -54,6 +54,8 @@ Model objects declared in `Networking` are immutable, and modelled as value type
 
 Model objects should implement `Comparable`.
 
+Model objects should conform to `GeneratedFakeable` and `Copiable`. [Fakeable](fakeable.md) and [Copiable](copiable.md) methods should be generated automatically with `rake generate`.
+
 ## Unit tests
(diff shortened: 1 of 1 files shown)


docs/NETWORKING.md
```diff
@@ -54,6 +54,8 @@ Model objects declared in `Networking` are immutable, and modelled as value type
 
 Model objects should implement `Comparable`.
 
+Model objects should conform to `GeneratedFakeable` and `Copiable`. [Fakeable](fakeable.md) and [Copiable](copiable.md) methods should be generated automatically with `rake generate`.
```
Ecarrion:
```suggestion
Model objects should conform to `GeneratedFakeable` and `GeneratedCopiable`. [Fakeable](fakeable.md) and [Copiable](copiable.md) methods should be generated automatically with `rake generate`.
```

Ecarrion (approved):

Ecarrion:
Thanks for improving the docs!

PR #4098, opened 2021-04-30
Merge release/6.5 final into develop

AliSoftware:
6.5.0.2 was just submitted to Apple.

The diff of this PR includes #4070 which was merged too close to today's submission to have time to be included in a new beta.
(that fix was a backport of something that had already been fixed and tested in `develop`)

---

Worth also noting that I ran into an issue during the automation of this release (`finalize_release`): see [this comment on Gio's PR in the toolkit](https://github.com/wordpress-mobile/release-toolkit/pull/236#issuecomment-830226372).

TL;DR is that translations were not commited and thus not included in the build `6.5.0.1` that I submitted initially. Only realized that those changes were left uncommited in my working copy _after_ I've submitted the build to Apple, so I had to:

 - "revoke from review" in the App Store
 - commit the translation changes that were left uncommited in my working copy (9648dba45fa28d196d593aa0b25fcf764ab216c6)
 - then manually bump the version to 6.5.0.2 (7b7dc28fcbda442b84a66f05697627376268fc5e)
 - delete the draft GitHub release that was previously created
 - trigger another release build via `fl trigger_release_build branch_to_build:"release/6.5"`
 - finally replace the build 6.5.0.1 with build 6.5.0.2 in the AppStore before resubmitting for review.

This bug is likely to happen again in all the subsequent releases we do, both on WCiOS, WPiOS and SNiOS, so we should fix this ASAP in the toolkit for next sprint (see Gio's PR)


oguzkocer (approved):
Thank you for the detailed explanation of the extra circumstances @AliSoftware!

PR #4103, opened 2021-05-02
Coupon Management - Yosemite: CouponStore registration

joshheald-a8c:
Part of #3915 

## Description
The implementation of Coupons in Yosemite was incomplete, as the CouponStore was not registered with the AuthenticatedState stores manager.

This meant that any CouponActions which were dispatched would never be handled.

## Testing
None applicable yet, as the Coupons entity is not yet in use.

Update release notes:


pmusolino:
:shipit:

pmusolino (approved):
:shipit:

PR #4105, opened 2021-05-03
[Tooling] Fix code_freeze lane

AliSoftware:
`code_freeze` lane was broken during partial manual fix last release after the breaking change in release toolkit.

Updating to latest release toolkit with fix of the call site for `get_prs_list`.


AliSoftware:
Force-merging since this is required to do my code-freeze and will be tested during that step anyway

PR #4106, opened 2021-05-03
Merge release/6.6 into develop

AliSoftware:
Code freeze completed, first beta out.

### Shipped Beta built without new .strings file commit

Note that due to once again bumping into [our toolkit bug about strings failing to commit](https://github.com/wordpress-mobile/release-toolkit/pull/236#issuecomment-830226372), the [first beta that was built by CI](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/11545/workflows/0ddf8ac7-7eb0-45cb-9288-bdf3ebc776b3/jobs/20930) (the one that went out to beta testers) was actually built from commit 2188395 – so not including 894a8f8 that I pushed to the remote manually only after the `circleci_trigger_job` action fired.

This should not be a big deal for this time, given we happen to use the English copy as key in our `.strings` file so any key that was missing in 2188395 (and that only got inserted in 894a8f8) will just fallback to the key which happens to be the same as the expected English copy anyway. Which is why I didn't bother doing the whole dance and creating a new build after 894a8f8 after all; but still not really nice to have that happen…

### Manual tag

Since our process now doesn't create the tag automatically anymore (tags are now only created when we publish GitHub releases, which we don't do for betas on iOS repos), I created and pushed the `6.6.0.0` tag manually. I created it on 2188395 (and not on the tip of this branch 894a8f8) to reflect the point mentioned above and tag what was _really_ sent to testers.


WooCommerce/Resources/en.lproj/Localizable.strings
```diff
@@ -147,6 +147,9 @@
 /* No comment provided by engineer. */
 "A refund will not be issued to the customer. You will need to manually issue the refund through Stripe." = "A refund will not be issued to the customer. You will need to manually issue the refund through Stripe.";
 
+/* Message explaining that the site may have an invalid SSL certificate. */
+"A secure connection to the site could not be made. Please make sure that your site has a valid SSL certificate." = "A secure connection to the site could not be made. Please make sure that your site has a valid SSL certificate.";
+
```
AliSoftware, line 152: Note: the changes in that file were not commited at the time the CI was triggered and the 6.6.0.0 beta was built – due to the current bug on toolkit about strings file commit. See PR description for details.

loremattei (approved):
LGTM. 
Thanks for the detailed explanation of the issue!
I agree that, this time, it doesn't affect the beta in any way and that manually committing the strings to this branch and merging it back to `develop` works for GlotPress.

PR #4108, opened 2021-05-03
[Mobile Payments] Card Reader Settings 2-4: Add VM for when a reader is connected

### LINKED ISSUE #4050: [Mobile Payments] Card Reader Settings v2-4: Create CardReaderConnectedViewModel
Create CardReaderConnectedViewModel and add to CardReaderSettingsPresentingViewController array of ViewModel. VM accepts siteID and CardReaderService. shouldShow returns true if CardReaderService.connectedReaders is not empty. Unit test exercises shouldShow

Edit (Apr 28): Don't pass CardReaderService. Have the dependency be on CardPresentPaymentStore

allendav:
Closes #4050

**Note: This is against feature/stripe-terminal-sdk-integration, not develop**

Changes:
- This is the next viewmodel in a series of viewmodels for card reader settings
- Updates shouldShow based on whether there are connected readers
- Adds unit tests

Work not yet done:
- Add this VM to the array of ViewModels passed to CardReaderSettingsPresentingViewController - I'm going to defer that to #4051

To test:
- Ensure all unit tests pass, especially CardReaderSettingsConnectedViewModelTests

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -3,18 +3,15 @@ import Yosemite
 
 final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedViewModel {
 
-    var shouldShow: CardReaderSettingsTriState
+    var shouldShow: CardReaderSettingsTriState = .isUnknown
     var didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?
 
-    private var didGetConnectedReaders: Bool
-    private var connectedReaders: [CardReader]
+    private var didGetConnectedReaders: Bool = false
+    private var connectedReaders = [CardReader]()
 
     init(didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?) {
 
         self.didChangeShouldShow = didChangeShouldShow
-        self.shouldShow = .isUnknown
-        self.didGetConnectedReaders = false
-        self.connectedReaders = []
 
         beginObservation()
     }
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -1,0 +1,62 @@
+import Foundation
+import Yosemite
+
+final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedViewModel {
+
+    var shouldShow: CardReaderSettingsTriState
+    var didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?
+
+    private var didGetConnectedReaders: Bool
+    private var connectedReaders: [CardReader]
+
+    init(didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?) {
+
+        self.didChangeShouldShow = didChangeShouldShow
+        self.shouldShow = .isUnknown
+        self.didGetConnectedReaders = false
+        self.connectedReaders = []
+
+        beginObservation()
+    }
+
+    /// Dispatches actions to the CardPresentPaymentStore so that we can monitor changes to the list of
+    /// connected readers.
+    ///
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+import Yosemite
+
+final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedViewModel {
+
+    var shouldShow: CardReaderSettingsTriState
```
ctarda: It seems like this is meant to be mutable only from within the context of this class, right? If it is, would you consider declaring it as a `private(set) var`?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -0,0 +1,62 @@
...
+    var didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?
+
+    private var didGetConnectedReaders: Bool
+    private var connectedReaders: [CardReader]
```
ctarda: It would be possible to initialise these properties here, next to their declaration:
`private var didGetConnectedReaders = falsea
`private var connectedReaders = []`

And remove the initialisation from the constructor. I am not sure if that’s better though 😊

ctarda (approved):
:shipit: Looks good!

A couple of (optional) comments related to Swift specific idioms.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: > It seems like this is meant to be mutable only from within the context of this class, right?

Yup! 100% correct.

> If it is, would you consider declaring it as a `private(set) var`?

Doesn't it "inherit" that from the protocol CardReaderSettingsPresentedViewModel ?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: Sure! I'll change `CardReaderSettingsUnknownViewModel` that way too. I have no strong feelings either way

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
ctarda: Nope. Conformance to protocols is interesting… You can declare a var in a protocol like this:
```
var aThing: String { get }
```

And then conform to it as ` let aThing: String = “whatnot”`

In this particular case, if the var declared in the protocol is read only, and the class that conforms to the protocol is read-write, conformance is still granted, and the var in the class is mutable. Fun times.

allendav: LOL. TIL. Will fix.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4108/commits/2e04020badf09266341640b9150430f168b126bb

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4108/commits/e2599a34a37b3541fb4cbff1585fcc7dda12a64f

PR #4109, opened 2021-05-04
Coupon Management: List - CouponListViewController

joshheald-a8c:
Part of #3916 

## Description
Adds a basic list of coupons to the coupon list view controller:
 - loading state
 - load and display the first page of coupons (only) 
 - an unfinished empty state (requires correct copy, image, and button implementation).

This is clearly a work in progress, given the above description!

## Testing
Open Settings, then tap "Manage Coupons" in the list.
Observe the loading state, and then that coupon cells replace the loading state if any are present on the store.
If there are no coupons, an (unfinished) empty state is displayed.
If there is an error, this is not yet displayed to the user, and an empty list is shown.

### Coupon list

https://user-images.githubusercontent.com/80399824/116983320-d8268200-acc1-11eb-8740-bb27e991ccba.mp4

### Empty state

https://user-images.githubusercontent.com/80399824/116983561-2cc9fd00-acc2-11eb-86f5-6f30a43f8893.mp4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIImage+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIImage+Woo.swift
@@ -168,6 +168,12 @@ extension UIImage {
             .imageFlippedForRightToLeftLayoutDirection()
     }
 
+    /// Empty Coupons Icon
+    ///
+    static var emptyCouponManagementListImage: UIImage {
+        return UIImage(named: "woo-error")!
+    }
+
     /// Empty Products Icon
     ///
     static var emptyProductsImage: UIImage {
--- a/WooCommerce/Classes/Extensions/UIImage+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIImage+Woo.swift
@@ -168,12 +168,6 @@ extension UIImage {
             .imageFlippedForRightToLeftLayoutDirection()
     }
 
-    /// Empty Coupons Icon
-    ///
-    static var emptyCouponManagementListImage: UIImage {
-        return UIImage(named: "woo-error")!
-    }
-
     /// Empty Products Icon
     ///
     static var emptyProductsImage: UIImage {
--- a/WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
@@ -1,0 +1,170 @@
+import Foundation
+import Yosemite
+import protocol Storage.StorageManagerType
+import class AutomatticTracks.CrashLogging
+
+
+struct CouponListCellViewModel {
+    var title: String
+    var subtitle: String
+    var accessiblityLabel: String
+}
+
+enum CouponListState {
+    case initialized // Nothing to do here
+    case loading // View should show ghost cells
+    case empty // View should display the empty state
+    case failed // View should display error in Notice
+    case coupons // View should display the contents of `couponViewModels`
+}
+
+final class CouponManagementListViewModel {
+    /// onChange: Closure to inform View that the state of the ViewModel has changed
+    ///
+    private let onChange: (CouponListState) -> ()
+
+    /// couponViewModels: ViewModels for the cells representing Coupons
+    ///
+    var couponViewModels: [CouponListCellViewModel] = []
+
+    /// siteID: siteID of the currently active site, used for fetching and storing coupons
+    ///
+    private let siteID: Int64
+
+    /// resultsController: provides models from storage used for creation of cell ViewModels
+    ///
+    private let resultsController: ResultsController<StorageCoupon>
+
+    /// syncingCoordinator: Keeps tracks of which pages have been refreshed, and
+    /// encapsulates the "What should we sync now" logic.
+    ///
+    private let syncingCoordinator: SyncingCoordinatorProtocol
+
+    /// storesManager: provides the store for handling actions
+    ///
+    private let storesManager: StoresManager
+
+    /// storageManager: provides the storage for the results controller to fetch from
+    ///
+    private let storageManager: StorageManagerType
+
+    // MARK: - Initialization and setup
+    //
+    init(siteID: Int64,
+         syncingCoordinator: SyncingCoordinatorProtocol = SyncingCoordinator(),
+         storesManager: StoresManager = ServiceLocator.stores,
+         storageManager: StorageManagerType = ServiceLocator.storageManager,
+         onChange: @escaping (CouponListState) -> ()) {
+        self.siteID = siteID
+        self.onChange = onChange
+        self.syncingCoordinator = syncingCoordinator
+        self.storesManager = storesManager
+        self.storageManager = storageManager
+        self.resultsController = Self.createResultsController(siteID: siteID,
+                                                              storageManager: storageManager)
+        configureSyncingCoordinator()
+        configureResultsController()
+        onChange(.initialized)
+    }
+
+    private static func createResultsController(siteID: Int64,
+                                                storageManager: StorageManagerType) -> ResultsController<StorageCoupon> {
+        let predicate = NSPredicate(format: "siteID == %lld", siteID)
+        let descriptor = NSSortDescriptor(keyPath: \StorageCoupon.dateCreated,
+                                          ascending: true)
+
+        return ResultsController<StorageCoupon>(storageManager: storageManager,
+                                                matching: predicate,
+                                                sortedBy: [descriptor])
+    }
+
+    /// Setup: Results Controller
+    ///
+    private func configureResultsController() {
+        resultsController.onDidChangeContent = updateViewModelState
+        resultsController.onDidResetContent = updateViewModelState
+
+        do {
+            try resultsController.performFetch()
+        } catch {
+            CrashLogging.logError(error)
+        }
+    }
+
+    /// Setup: Syncing Coordinator
+    ///
+    private func configureSyncingCoordinator() {
+        syncingCoordinator.delegate = self
+    }
+
+    /// Builds coupon cell view models if coupons are present, and notifies view controller of what to display
+    ///
+    func updateViewModelState() {
+        if resultsController.numberOfObjects > 0 {
+            buildCouponViewModels()
+            onChange(.coupons)
+        } else {
+            onChange(.empty)
+        }
+    }
+
+    private func buildCouponViewModels() {
+        couponViewModels = resultsController.fetchedObjects.map({ coupon in
+            return CouponListCellViewModel(title: coupon.code,
+                                           subtitle: coupon.description,
+                                           accessiblityLabel: coupon.description)
+        })
+    }
+
+
+    // MARK: - ViewController actions
+    //
+    /// The ViewController calls `viewDidLoad` to notify the view model it's ready to recieve results
+    ///
+    func viewDidLoad() {
+        onChange(.loading)
+        syncingCoordinator.synchronizeFirstPage(reason: nil, onCompletion: nil)
+    }
+
+    /// The ViewController may use this method to retrieve a coupon for navigation purposes
+    ///
+    func coupon(at indexPath: IndexPath) -> Coupon? {
+        return resultsController.safeObject(at: indexPath)
+    }
+}
+
+
+// MARK: - SyncingCoordinatorDelegate
+//
+extension CouponManagementListViewModel: SyncingCoordinatorDelegate {
+    /// Syncs the specified page of coupons from the API
+    /// - Parameters:
+    ///   - pageNumber: 1-indexed page number
+    ///   - pageSize: Number of coupons per page
+    ///   - reason: A string originating from a call to the coordinator's sync request methods,
+    ///   to identify the type of sync required
+    ///   - onCompletion: Completion handler to call passing whether the sync was successful
+    func sync(pageNumber: Int,
+              pageSize: Int,
+              reason: String?,
+              onCompletion: ((Bool) -> Void)?) {
+        let action = CouponAction
+            .synchronizeCoupons(siteID: siteID,
+                                pageNumber: pageNumber,
+                                pageSize: pageSize) { [weak self] result in
+                guard let self = self else { return }
+
+                switch result {
+                case .failure(let error):
+                    DDLogError("⛔️ Error synchronizing coupons: \(error)")
+                    self.onChange(.failed)
+                case .success:
+                    DDLogInfo("Synchronized coupons")
+                }
+
+                onCompletion?(result.isSuccess)
+        }
+
+        storesManager.dispatch(action)
+    }
+}
--- a/WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
@@ -11,16 +11,33 @@ struct CouponListCellViewModel {
 }
 
 enum CouponListState {
+    case initialized // ViewModel ready to recieve actions
     case loading // View should show ghost cells
     case empty // View should display the empty state
-    case failed // View should display error in Notice
     case coupons // View should display the contents of `couponViewModels`
 }
 
 final class CouponManagementListViewModel {
-    /// onChange: Closure to inform View that the state of the ViewModel has changed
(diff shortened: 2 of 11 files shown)


joshheald-a8c:
@pmusolino Sorry this one has gotten a little long, since it's over 500 lines. I considered splitting out the empty/error state stuff to reduce it, but in the end I thought it was better to include them together so you can see the approach now and let me know if you think I should make changes to it.

In particular, I'm starting to think about the paging and whether my original plan was flawed by not using `PaginatedListViewControllerStateCoordinator` either directly on the view controller, or as a part of the view model's interface. The `didLeave` and `didEnter` approach it has perhaps gives a bit more context to correctly update the view, compared to the more implicit interpretation the view has to make of the single piece of state change information that the coupon list view model currently exposes to it. I.e. currently, the clearing down of placeholder cells and the empty state has to happen at different places in `viewModelDidChange`.

Would love to get your thoughts on this in particular.

WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -168,6 +168,12 @@ extension UIImage {
             .imageFlippedForRightToLeftLayoutDirection()
     }
 
+    /// Empty Coupons Icon
+    ///
+    static var emptyCouponManagementListImage: UIImage {
+        return UIImage(named: "woo-error")!
+    }
+
```
ealeksandrov: For all images here we also add a test case to `IconsTests` to make sure all assets included.
And since it's a temp placehloder - maybe just reuse `errorImage` directly for now?

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
```diff
@@ -0,0 +1,170 @@
...
+
+
+struct CouponListCellViewModel {
+    var title: String
+    var subtitle: String
+    var accessiblityLabel: String
+}
+
+enum CouponListState {
+    case initialized // Nothing to do here
```
ealeksandrov: Why do we need `initialized` state? Can we merge it with empty?

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
```diff
@@ -0,0 +1,170 @@
...
+    /// Setup: Syncing Coordinator
+    ///
+    private func configureSyncingCoordinator() {
+        syncingCoordinator.delegate = self
+    }
+
+    /// Builds coupon cell view models if coupons are present, and notifies view controller of what to display
+    ///
+    func updateViewModelState() {
+        if resultsController.numberOfObjects > 0 {
```
ealeksandrov: nit: `isEmpty == false` would fit better here

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
```diff
@@ -1,25 +1,200 @@
 import UIKit
+import WordPressUI
 
 final class CouponManagementViewController: UIViewController {
+
+    @IBOutlet weak var tableView: UITableView?
```
ealeksandrov: Can we make it `private`?

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
```diff
@@ -1,25 +1,200 @@
...
 }
 
+
 // MARK: - View Configuration
 //
 private extension CouponManagementViewController {
 
     func configureNavigation() {
-        title = Localization.title
+        self.title = Localization.title
```
ealeksandrov: nit: no need for `self`

ealeksandrov (changes requested):
Hey @joshheald-a8c!

I've tested and confirm that I can see existing coupons in a list.
But when I don't have any coupons - I don't see empty state, it just loads endlessly (while I get `Synchronized coupons` in a console).

https://user-images.githubusercontent.com/3132438/117136121-9b858400-adb0-11eb-8c93-4669e85d5ce8.mp4

> whether my original plan was flawed by not using PaginatedListViewControllerStateCoordinator

`PaginatedListViewControllerStateCoordinator` may be a good idea, since it provides a dedicated callback to manage state of a view - I think I would choose it. Of course it's possible to update state handling in your viewmodel with pagination integration in mind.

> either directly on the view controller, or as a part of the view model's interface

I would also keep resultsController in a VC since this screen doesn't have much logic to encapsulate it in a VM (since we already have dedicated solutions for data fetching). There is just view manipulation in response to different contents types.

Also one thing that caught my attention - there are a few autolayout errors when I open coupons lists (`Unable to simultaneously satisfy constraints.`)

WooCommerce/Classes/Extensions/UIImage+Woo.swift
joshheald-a8c: Thanks for the heads up. I have an image now but to avoid this PR getting bigger I'll switch to errorImage for the time being and update the empty state in a future PR, with tests for the new image.

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
joshheald-a8c: If we want to go get rid of it, it might be better to merge it with loading, because briefly showing an empty state first could be misleading...

`initialized` was added so that we could represent all the possible states of the screen: when we first init, we're not yet loading, we don't know whether there are coupons or not, etc. It's a habit from the RxSwift architecture we used for a while at my current job, definitely not something we need to keep here.

It's not used like this in this example, but it can be helpful to inform the view controller that the view model is ready for use in some situations. I'll remove it for now and we can add it back if it turns out it's needed.

joshheald-a8c:
Thanks for your review @ealeksandrov 👍 

I've fixed up the nits, smaller changes, and runtime constraint warnings just now. Further changes still to make on the other stuff and I'll let you know when I'm done with those...

> But when I don't have any coupons - I don't see empty state, it just loads endlessly (while I get `Synchronized coupons` in a console).
> `PaginatedListViewControllerStateCoordinator` may be a good idea, since it provides a dedicated callback to manage state of a view - I think I would choose it. Of course it's possible to update state handling in your viewmodel with pagination integration in mind.

I think so too, especially given the issue you've found with the empty state. I'll switch to using the PaginatedListCoordinator with another commit later on, and resolve that empty state bug at the same time.

> I would also keep resultsController in a VC since this screen doesn't have much logic to encapsulate it in a VM (since we already have dedicated solutions for data fetching). There is just view manipulation in response to different contents types.

This one I'm less sure on, and I'm going to think more about it... if it's in the VC, it'll be harder to test the cell view model creation, which will have more logic around it as I move through the work here.

joshheald-a8c:
@ealeksandrov when you have time, I'd really appreciate a follow up review on this please :)

> I'll switch to using the PaginatedListCoordinator with another commit later on, and resolve that empty state bug at the same time.
So, I tried this, and ended up deciding against using it directly. I found that when I moved to try implementing pull to refresh, and loading subsequent pages, using the existing coordinator meant for either: 
1. complicated state translation code in the view model, or
2. complicated untestable state handling code in the view controller

This was due to `.syncing` representing three distinct states: first page load, subsequent page loads, and refreshing after loading data.

Instead, I've opted to take inspiration from the implementation of the existing coordinator, with the intention to extend it to cover these three states explicitly in a later commit. I think that should be done as a refactor after the coupon work is merged, because it will touch unrelated screens and we'd run the risk of causing either merge conflicts, or accidentally breaking code which is added after the coupon branch was taken.

This has also meant a small change to my initial plan, where I now have two closures on the view model (didEnterState, didLeaveState) to update the state of the view rather than one (onStateChange). This means we no longer have to rely on interpreting one state to mean that we've left another, and really cleans up the view controller.

What do you think of these changes?

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
joshheald-a8c: I ended up bringing this back. Having an explicit change from `.initialized` to `.loading` gives the view controller an explicit `didEnter` state change after it tells the view model to start loading. This gives a straightforward way to start showing the placeholder cells, without the risk of showing the empty state briefly.

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
```diff
@@ -0,0 +1,198 @@
+import Foundation
+import Yosemite
+import protocol Storage.StorageManagerType
+import class AutomatticTracks.CrashLogging
+
+
+struct CouponListCellViewModel {
+    var title: String
+    var subtitle: String
+    var accessiblityLabel: String
```
ealeksandrov: nit:
```suggestion
    var accessibilityLabel: String
```

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
```diff
@@ -0,0 +1,198 @@
...
+            return CouponListCellViewModel(title: coupon.code,
+                                           subtitle: coupon.description,
+                                           accessiblityLabel: coupon.description)
+        })
+    }
+
+
+    // MARK: - ViewController actions
+    //
+    /// The ViewController calls `viewDidLoad` to notify the view model it's ready to recieve results
```
ealeksandrov: nit:
```suggestion
    /// The ViewController calls `viewDidLoad` to notify the view model it's ready to receive results
```

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    func handleCouponSyncResult(result: Result<Bool, Error>) {
+        switch result {
+        case .failure(let error):
+            DDLogError("⛔️ Error synchronizing coupons: \(error)")
+
+        case .success:
+            DDLogInfo("Synchronized coupons")
+        }
```
ealeksandrov, line 176: nit: `success` case should come first, before failure

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -38,6 +38,11 @@ final class SettingsViewController: UIViewController {
         return urlAsString?.hostname() ?? String()
     }
 
+    private var siteID: Int64 {
+        let siteID = ServiceLocator.stores.sessionManager.defaultStoreID
+        return siteID ?? 0
+    }
```
ealeksandrov: Since it's used only in a single place, what do you think about moving it there? Also fallback to 0 will break things on CouponManagementViewController, maybe add guard with error message instead?

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
```diff
@@ -1,25 +1,205 @@
...
+
+        emptyStateViewController.view.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(emptyStateViewController.view)
+
+        NSLayoutConstraint.activate([
+            emptyStateViewController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            emptyStateViewController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            emptyStateViewController.view.topAnchor.constraint(equalTo: view.topAnchor),
+            emptyStateViewController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        ])
```
ealeksandrov: nit: there is `pinSubviewToAllEdges` helper that can simplify this

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
```diff
@@ -1,25 +1,205 @@
...
+
+        static let emptyStateMessage = NSLocalizedString(
+            "No coupons yet",
+            comment: "The text on the placeholder overlay when there are no coupons on the coupon management list")
+
+        static let emptyStateDetails = NSLocalizedString(
+            "Market your products by adding a coupon to offer your customers a discount.",
+            comment: "The details on the placeholder overlay when there are no coupons on the coupon management list")
     }
+
```
ealeksandrov: very nit: here I would prefer 2 braces without newline between. And there are more cases of similar spaced closing braces above.

I've spend some time to find a rule about it and I think it aligns with https://github.com/raywenderlich/swift-style-guide/#spacing:

> There should be no blank lines after an opening brace or before a closing brace.

(don't assume strict style requirements, I've just noticed inconsistency between VC and VM looking side-by-side 🙂)

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
```diff
@@ -1,25 +1,205 @@
 import UIKit
+import WordPressUI
 
 final class CouponManagementViewController: UIViewController {
+
+    @IBOutlet private weak var tableView: UITableView?
```
ealeksandrov: Any reason for `IBOutlet` to be optional?

ealeksandrov (approved):
Nice work @joshheald-a8c!
I can see empty, loading and content state fine ✅
State logic works well and looks good ✅

I've added more smaller nits, since I've looked more line-by-line this time, but there is nothing critical.
To catch more typos - check `Edit > Format > Spelling and Grammar` in Xcode, I've learned about it recently and it's awesome 😄

Ping me after you push last bits (if you wish) and ready to merge it.

Following our `initialized` case discussion - I've tried to replace it with both `empty` and `coupons` and it worked well for me, no glitches or anything. And there is no logic around it, it's just initial placeholder value 🤔
I'll let you decide on keeping it in next iteration of this, maybe you'll find it useful in a more complete flow.

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
joshheald-a8c: Thanks, good spot!

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementListViewModel.swift
joshheald-a8c: And again 🦅 👁️

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
joshheald-a8c: I've added a pop up error message if there is no store ID, with an action to let the user select the store. Was this what you meant for the error message? I think it's probably better than a `fatalError` in this instance, since it should be user-recoverable really.
coupon-list-opening-error

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
joshheald-a8c: Thanks!

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
joshheald-a8c: Yeah, the inconsistency wasn't great. I've removed the excess newlines and prefer it without them too. 

I think I've taken different cues from different existing files as I've been working, but hadn't spotted that in the RW style guide. Thanks for finding that 🙇 it's a well stated guideline, nice and clear.

WooCommerce/Classes/ViewRelated/Coupons/CouponManagementViewController.swift
joshheald-a8c: Nope, just something which was useful earlier on in writing it but not needed any more. Removed 😃

joshheald-a8c:
@ealeksandrov Thanks for your review 👍 
I've fixed up the remaining nits and improvements you identified, so whenever you're ready to take a look, that would be awesome 😁

ealeksandrov (approved):
LGTM :shipit:

BTW, that's a nice implementation of a error for "please select a store" - I was thinking about simple log into console 😅

Issue #3895, opened 2021-03-30
Add-Ons: Move the feature flag to the beta features screen


Ecarrion:
Once the workaround is complete, we should put this semi-feature behind the beta-features screen. As this is not yet our final form! ™️

PR #4117, opened 2021-05-05
Extract Copiable & Fakeable types out of Networking

Ecarrion:
# Why

While I was developing a solution for #3895, I found myself in need of updating `GeneralAppSettings` with a new property to be able to persis some new settings. 

As `GeneralAppSettings` is a struct we can't mutate it like a class, so I instantly tried to adopt our `copiable` pattern on it.

To my surprise, `GeneralAppSettings` lives in the `Storage` layer and our `copiable` types lives in the `Networking` layer. 

As I don't wanted to couple those two layers in order to use some handy code-gen, this PR takes care of extracting the `copiable` and `fakeable` types out of the `Networking` layer and add them into a new `Codegen` SPM package, so it can be reused safely across targets.

# How

- Created a new `Codegen` SPM package, with the contents of the `copiable` and `fakeable` types. Sourcery templates and configuration files have also been moved under that directory.

- Updated sourcery configuration files with the new template's path.

- Added a new sourcery configuration files to support `copiable` types from the `Storage` layer, as it will be needed for #3895 

- Updated the `copiable` swifttemplate to guarantee that the generated file will have all the required imports a copiable type needs.

- Update our target(`WooCommerce`, `Yosemite`, `Networking`, `Storage`) to link against the new `Codegen` dynamic library.

- Update entities with the necessary `import Codegen`

# Testing steps
- Run `rake generate` and see that the script doesn't fail.

Update release notes:


koke:
I still haven't used these enough to have a solid opinion, but my impression is that both `Copiable` and `Fakeable` are tools to work with immutable models. The first would be more useful for manipulating actual models, while the second seems more like a testing helper to create mocks.

I think the fact that they are mostly generated with Sourcery should stay as an implementation detail, and it might make more sense to have a package with a name that reflects that these are tools to handle immutable models.

I don't think that's a blocker to merge this though, just something to think about.

Rakefile
```diff
@@ -177,19 +177,19 @@ end
 
 desc "Run all code generation tasks"
 task :generate do
```
koke, line 179: Unrelated to this PR, but I just noticed this task should depend on `:dependencies` since it uses a command from `Pods`

koke (approved):
I'm not super familiar with Sourcery yet, but this seems to work fine. I'm having some thoughts about these tools, but they don't necessarily apply to where the code lives (other than maybe the naming of the package), so I'll save them for a future post 👍🏽 

Even for the naming, I think Codegen might be fine in its current state, but maybe this can eventually evolve into a `Model` framework.

koke:
Also, adding a side note here. I thought about suggesting we add a `typealias` for the protocols to avoid so many `import Codegen` instances, like we do for the models, but then the code generation stopped working. If you have a `typealias GeneratedFakeable = Codegen.GeneratedFakeable` in your module, Sourcery will see the protocol with the prefix and `types.based["GeneratedCopiable"]` will only match those types that conform to the `typealias` and not the original.

I don't think we need to solve this right now, but I found that interesting to know.

Ecarrion:
Thanks for the review @koke!
Looking forward to your post to keep improving these tools!

AliSoftware:
Just putting this here for reference and history:

This change introduced an issue when archiving/packaging the app for TestFlight, due to a bug in Xcode 12.4 which always configures framework added in the link phase to also be in the "embed framworks" phase.
Meaning that the new `Codegen` framework was not only linked to `Storage`, `Networking`, `Fakes` and `Yosemite` as we wanted it to, but also embedded inside those frameworks (i.e Codegen was nested inside a `Framworks/` folder within the `Storage.framework` and the 3 others), while nested frameworks is not allowed and rejected during TestFlight/AppStore validation.

This Xcode 12.4 bug was fixed in Xcode 12.5 so we had to [use Xcode 12.5 to fix the `pbxproj`](https://github.com/woocommerce/woocommerce-ios/pull/4215) and be able to configure `Codegen` to be _linked **but** not embedded_ to those 4 frameworks. The fix made using Xcode 12.5's UI luckily seems to persist when we open the `pbxproj` back in Xcode 12.4, allowing us to still build the app and upload betas to TestFlight with 12.4 and via CI. 

But we might want to be on the lookout in case Xcode 12.4 re-introduce that bug when someone makes a (possibly unrelated) change to the `pbxproj` later and Xcode 12.4 regenerates a bogus `pbxproj` in the process… 🤞

Issue #3921, opened 2021-04-06
Plugins: Show plugin list in settings


Ecarrion:
The end goal is to synchronize plugins for a site and display them in the app via the settings module.

Requirements:
- Fetch site plugin list using the `wp/v2/plugins` endpoint. p91TBi-4ND-p2#comment-3883
- Store a site's plugin list.
- Create a view to display all synced ~sites~ plugins via the settings module.
- Make sure plugins are being kept in sync automatically. 
- Allow to manually re-sync all plugins.

Settings | Plugin List | Error State
--- | --- | ---
<img width="375" alt="iOS-settings" src="https://user-images.githubusercontent.com/562080/113867828-9c58d500-9774-11eb-93ad-fa4b1d3d02ab.png"> | <img width="375" alt="iOS-settings-Plugins" src="https://user-images.githubusercontent.com/562080/116894394-9c26ea80-abf7-11eb-8260-a1d26a143712.png"> | <img width="341" alt="err-rstate" src="https://user-images.githubusercontent.com/562080/117006689-5bd67380-acae-11eb-9cb8-312f216f595b.png">

Asset: 
[img-sign in-jetpack needed-light.pdf](https://github.com/woocommerce/woocommerce-ios/files/6421199/img-sign.in-jetpack.needed-light.pdf)
[img-sign in-jetpack needed-dark.pdf](https://github.com/woocommerce/woocommerce-ios/files/6426971/img-sign.in-jetpack.needed-dark.pdf)

Note: We can omit the `+` and the `>` CTA for now as that is outside the scope of this task.

itsmeichigo:
Hi @Ecarrion, I have one small question regarding one of the requirements - please help clarify:

> Create a view to display all synced sites via the settings module.

I'm not sure the purpose of this feature. The "Switch Store" row can redirect users to their lists of stores (sites), and this screen seems to serve the purpose of showing synced sites. Do we really need another screen just to list the sites?

If I'm misunderstanding something here - please let me know too.

Ecarrion:
Hey @itsmeichigo, good catch, that was a copy-paste mistake 😇 
It should be:
> Create a view to display all synced plugins via the settings module.

Does that make better sense?

itsmeichigo:
Thanks, that's perfect! 🙂

Ecarrion:
I have updated this issue with the following requirement.

[...]

PR #4118, opened 2021-05-05
Plugins: Add new entity SitePlugin to Storage

### LINKED ISSUE #4112: Plugins: Update Storage layer
## Tasks

- [ ] Add SitePlugin model
- [ ] Add unit test for migration
- [ ] Update MIGRATIONS.md
- [ ] Add feature flag & merge to develop

itsmeichigo:
~closes~ Part of #4112 

## Description
As part of #3921 - in preparation for syncing plugins, a SitePlugin entity is needed to persist the data to Storage.

## Solution
- Added new data model version 50
- Added new entity SitePlugin
- Updated StorageType+Extension with function to load plugins by site ID

## Testing
- Tests were added to check migration and plugin loading so make sure CI succeeds.
- App should launch successfully with new the model version.

Update release notes:


Ecarrion (changes requested):
Hey @itsmeichigo! I think some core-data files are pointing to an absolute path as the project does not compile. Could you update it so I can review it?

> error: Build input files cannot be found: '/Users/huongdo/Documents/GitHub/ichigo-woocommerce-ios/Storage/Storage/Model/SitePlugin+CoreDataProperties.swift', '/Users/huongdo/Documents/GitHub/ichigo-woocommerce-ios/Storage/Storage/Model/SitePlugin+CoreDataClass.swift' (in target 'Storage' from project 'Storage')

Storage/Storage/Model/SitePlugin+CoreDataProperties.swift
```diff
@@ -0,0 +1,26 @@
...
+    @NSManaged public var pluginUri: String?
+    @NSManaged public var author: String?
+    @NSManaged public var authorUri: String?
+    @NSManaged public var descriptionRaw: String?
+    @NSManaged public var descriptionRendered: String?
+    @NSManaged public var version: String?
+    @NSManaged public var networkOnly: Bool
+    @NSManaged public var requiresWPVersion: String?
+    @NSManaged public var requiresPHPVersion: String?
+    @NSManaged public var textDomain: String?
```
Ecarrion: If these are marked as non-optional in the data model. I think we can remove the optional(?) annotation. What do you think?

Storage/Storage/Tools/StorageType+Extensions.swift
```diff
@@ -514,4 +514,12 @@ public extension StorageType {
         let predicate = \AddOnGroup.siteID == siteID && \AddOnGroup.groupID == groupID
         return firstObject(ofType: AddOnGroup.self, matching: predicate)
     }
+
+    /// Returns all stored plugins for a provided `siteID`.
+    ///
+    func loadPlugins(siteID: Int64) -> [SitePlugin] {
+        let predicate = \SitePlugin.siteID == siteID
```
Ecarrion, line 521: Do you have any opinions on this way to build predicates?

Storage/Storage/Tools/StorageType+Extensions.swift
```diff
@@ -514,4 +514,12 @@ public extension StorageType {
         let predicate = \AddOnGroup.siteID == siteID && \AddOnGroup.groupID == groupID
         return firstObject(ofType: AddOnGroup.self, matching: predicate)
     }
+
+    /// Returns all stored plugins for a provided `siteID`.
+    ///
+    func loadPlugins(siteID: Int64) -> [SitePlugin] {
```
Ecarrion, line 520: FYI, think you will also need functions to load an specific plugin. It's fine to add them later, when you really need them tho.

Storage/StorageTests/Tools/StorageTypeExtensionsTests.swift
```diff
@@ -1019,4 +1019,21 @@ class StorageTypeExtensionsTests: XCTestCase {
...
+        let plugin2 = storage.insertNewObject(ofType: SitePlugin.self)
+        plugin2.name = "AAA"
+        plugin2.siteID = sampleSiteID
+
+        // When
+        let storedPlugins = try XCTUnwrap(storage.loadPlugins(siteID: sampleSiteID))
+
+        // Then
+        XCTAssertEqual(storedPlugins, [plugin2, plugin1])
+    }
```
Ecarrion: Do you think it's worth it to create a `SitePlugin` with a different `siteID` so we can test that those sites are not being loaded?

Ecarrion (changes requested):

Storage/Storage/Model/SitePlugin+CoreDataProperties.swift
itsmeichigo: Good catch! I usually keep the variables as-is to avoid having to change them again when there is any update in the future. But since it is less likely for changes to happen in this case (the API seems stable), I think it's a good idea to make the variables non-optional and keep code clearer when working with them later on.

Storage/Storage/Tools/StorageType+Extensions.swift
itsmeichigo: I really like the idea of TypedPredicate! This is a great way to ensure type safety while also makes code more Swifty and readable than the traditional string queries.

Storage/Storage/Tools/StorageType+Extensions.swift
itsmeichigo: Sure! I think another function to load a plugin by `siteID` and `name` should be sufficient - what do you think? I'm not aware of other possible needs for querying the plugins though, so let me know if you have other suggestions!

Storage/StorageTests/Tools/StorageTypeExtensionsTests.swift
itsmeichigo: This is another great idea 💯 I think it's good to have more tests to check for possible issues. My current test may be a bit optimistic: if the `loadPlugins` function returns all the plugins in the storage, this issue would not be caught in this test. I'll add another test to make sure this does not happen.

itsmeichigo:
@Ecarrion I have pushed more changes handling your suggestions, please help review again. Thanks!

Ecarrion (approved):
Thanks for this @itsmeichigo it looks 💯 

FYI, I'm not closing the issue you reference in this PR because I see that it also includes the "feature flag" task.

Issue #3926, opened 2021-04-07
[Mobile Payments] Optimizing use of Combine in CardPresentPaymentStore


ctarda:
At this point CardPresentPaymentStore is over retaining the Futures it gets from the CardReaderService.

When we dispatch an action declared in CardPresentPaymentStore we instantiate one cancellable (or more than one in some methods) and then store it in a set of Cancellables.

We never clear that set of Cancellables though, so every time we dispatch a method in the store we will be adding new cancellable to he existing ones. if we dispatch the same method twice in a row, then we'll have duplicated results. 

Since we are basically interested in four streams (card reader discovery, connection status, payment collection and card reader events), the simplest solution might be getting rid of the set of cancellables, and declare four different Cancellables that we override every time we dispatch the corresponding method in the Store.

koke:
So far, I've started investigating the part of the store that deals with reader discovery and connection.

My understanding of the potential states for discovery and connection looks something like this:

![Card Present Payments - Reader connection](https://user-images.githubusercontent.com/8739/117136453-adfbbf80-ada8-11eb-8b13-0bec1dd425a2.png)

Also, there are some assumptions I got from looking at the code, documentation, and the sample app:
- You have to set the token provider before doing anything else with the SDK.
- The SDK will stay in discovery mode until it's canceled, a reader is connected, or an error happens. I'm not sure which kind of errors might happen, but one I experienced in the sample app was that the token provider did not provide a valid token. Another was that the app tried to start discovery while the SDK was already busy with another operation.
- We should not cache any Reader value beyond the call to `connect`. This means that connecting to a Reader always requires a discovery process before.
- We should ensure no reader is connected before starting a discovery process. This, combined with the previous fact, means that only one reader can be connected at a time.
- We should ensure no discovery process is already active before starting discovery.
- The SDK provides a `connectedReader` and `connectionStatus` as a source of truth.

The current sequence of events for the simple case where a single reader is discovered and connected looks like this:

![stripe-discovery](https://user-images.githubusercontent.com/8739/117137134-9113bc00-ada9-11eb-803a-9d210915bfb6.png)


@allendav @ctarda does this match your understanding of how the SDK works?

The issues I see with the current implementation are:

- The `store(in:)` method we use to store cancellables is more suited to an object with a finite lifecycle, so when that object is deallocated, it cancels any outstanding subscription. The store will live as long as the app is running, so those will never be cancelled. We probably need to keep track of each individual subscription more explicitly.
[...]

PR #4120, opened 2021-05-05
[Mobile Payments] Refactor discoveredReaders publisher to clean up when discovery ends

koke:
Part of #3926

I'm having a hard time testing these with the UI being a work in progress and not having a physical reader yet, but here's a first step towards better subscription management.

# Why

There's a much larger explanation in this comment https://github.com/woocommerce/woocommerce-ios/issues/3926#issuecomment-832628627, but the existing implementation is holding its subscriptions forever, which might lead to unexpected behavior and performance issues.

This PR tackles just one of the instances in `CardPresentPaymentStore` where this happens: starting the discovery process.

# How

Instead of keeping a subscription to `discoveredReaders` indefinitely, this now does a few things differently:
- `StripeCardReaderService` will send a completion event when it stops being in discovery mode.
- `StripeCardReaderService` will replace its internal `discoveredReadersSubject` when this happens with a new one, so new subscriptions can still happen if the discovery process starts again
- `CardPresentPaymentStore` will use a `[Sink](https://developer.apple.com/documentation/combine/subscribers/sink)` subscriber instead of the `sink` operator, which doesn't return a cancellable, automatically manages the subscription, and cleans up when it completes.

# To test

I've tested that the discovery and connection process still works, but I haven't been able to test many edge cases. Sometimes the UI is failing when it gets more than one update from the simulated card reader which makes it impossible to test.

Unless you have a card reader, change the `start` method of `StripeCardReaderService` to use this configuration:
```swift
  let config = DiscoveryConfiguration(
            discoveryMethod: .bluetoothScan,
            simulated: true
        )
```

Then add a quick implementation to stop discovery. In `CardReaderSettingsViewModel.stopSearch()` add:
```swift
ServiceLocator.stores.dispatch(CardPresentPaymentAction.cancelCardReaderDiscovery(onCompletion: { _ in }))
``` 

Finally, in `CardReaderSettingsViewModel.startSearch` add some logging to the completion block of `startCardReaderDiscovery`:

```swift
            print("📶 Discovered \(cardReaders.count) readers")
```

1. Build and run the app
2. Go to Settings > Manage Card Reader
3. Tap Connect Card Reader
4. As soon as the searching alert appears, tap cancel
5. Repeat 3-4 a few times.
6. Ensure the console doesn't start printing duplicated `Discovered %d readers` lines after doing this a few times.

Note: You might see an alert briefly pop up, as the simulated reader comes in very quickly as you are canceling the search.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -6,7 +6,7 @@ public final class StripeCardReaderService: NSObject {
 
     private var discoveryCancellable: StripeTerminal.Cancelable?
 
-    private let discoveredReadersSubject = CurrentValueSubject<[CardReader], Never>([])
+    private var discoveredReadersSubject = CurrentValueSubject<[CardReader], Never>([])
     private let connectedReadersSubject = CurrentValueSubject<[CardReader], Never>([])
     private let serviceStatusSubject = CurrentValueSubject<CardReaderServiceStatus, Never>(.ready)
     private let discoveryStatusSubject = CurrentValueSubject<CardReaderServiceDiscoveryStatus, Never>(.idle)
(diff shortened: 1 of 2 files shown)


ctarda (approved):
Tests perfect! :shipit:

PR #4123, opened 2021-05-05
[Mobile Payments] Card Reader Settings V2-5: Register viewmodels

### LINKED ISSUE #4051: [Mobile Payments] Card Reader Settings v2-5: cardReadersV2WasPressed

allendav:
Closes #4051

Changes:
- This is the next step in Card Reader Settings v2.
- We add a new protocol `CardReaderSettingsPrioritizedViewModelsProvider` - objects that conform to this protocol will be able to tell consumers (like a presenting view controller) which of many viewmodels should have priority at a given moment.
- We add `CardReaderSettingsViewModelsOrderedList` which 1) conforms to that protocol, 2) instantiates the viewmodels we created in previous PRs and 3) figures out which viewmodel is prioritized by calling `shouldShow` on each
- We build out `CardReaderSettingsPresentingViewController` a little further to accept a `CardReaderSettingsPrioritizedViewModelsProvider` and then update a temporary label with the identifier of the view controller it would present (based on which viewmodel should be prioritized according to the provider passed in

To test:
- Launch the app
- Navigate to Settings (Gear) and then Manage Card Readers V2
- Ensure it shows “CardReaderSettingsUnknownViewController”
- Navigate back to Settings (Gear)
- Navigate to Manage Card Readers (not V2)
- Connect your reader
- Navigate to Manage Card Readers V2
- Ensure it shows “CardReaderSettingsConnectedViewController”

Work not yet done
- Add unit tests for `CardReaderSettingsViewModelsOrderedList` to ensure it prioritizes viewmodels as expected

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
@@ -3,18 +3,32 @@ import UIKit
 
 final class CardReaderSettingsPresentingViewController: UIViewController {
 
+    @IBOutlet weak var temporaryLabel: UILabel!
+
     /// An array of viewModels and related view classes
-    private var viewModelsAndViews = [CardReaderSettingsViewModelAndView]()
+    private var viewModelsAndViews: CardReaderSettingsPrioritizedViewModelsProvider?
 
     /// Set our dependencies
-    func configure(viewModelsAndViews: [CardReaderSettingsViewModelAndView]) {
+    func configure(viewModelsAndViews: CardReaderSettingsPrioritizedViewModelsProvider) {
         self.viewModelsAndViews = viewModelsAndViews
+        self.viewModelsAndViews?.onPriorityChanged = onViewModelsPriorityChange
     }
 
     override func viewDidLoad() {
         super.viewDidLoad()
         configureBackground()
         configureNavigation()
+        configureInitialState()
+    }
+
+    private func configureInitialState() {
+        onViewModelsPriorityChange(viewModelAndView: viewModelsAndViews?.priorityViewModelAndView)
+    }
+
+    private func onViewModelsPriorityChange(viewModelAndView: CardReaderSettingsViewModelAndView?) {
+        // For now, just update the label with the view identifer we should display
+        // A later PR will actually add presented views
+        temporaryLabel.text = viewModelAndView?.viewIdentifier ?? "Loading"
     }
 }
 
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsViewModelsOrderedList.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsViewModelsOrderedList.swift
@@ -48,12 +48,12 @@ final class CardReaderSettingsViewModelsOrderedList: CardReaderSettingsPrioritiz
     }
 
     private func reevaluatePriorityViewModelAndView() {
-        let newPriorityViewModelAndView = viewModelsAndViews.first(
+        guard let newPriorityViewModelAndView = viewModelsAndViews.first(
             where: { $0.viewModel.shouldShow == .isTrue }
-        )
-
-        if newPriorityViewModelAndView != priorityViewModelAndView {
-            priorityViewModelAndView = newPriorityViewModelAndView
+        ), newPriorityViewModelAndView != priorityViewModelAndView else {
+            return
         }
+
+        priorityViewModelAndView = newPriorityViewModelAndView
     }
 }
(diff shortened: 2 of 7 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
```diff
@@ -3,18 +3,32 @@ import UIKit
...
+
     /// An array of viewModels and related view classes
-    private var viewModelsAndViews = [CardReaderSettingsViewModelAndView]()
+    private var viewModelsAndViews: CardReaderSettingsPrioritizedViewModelsProvider?
 
     /// Set our dependencies
-    func configure(viewModelsAndViews: [CardReaderSettingsViewModelAndView]) {
+    func configure(viewModelsAndViews: CardReaderSettingsPrioritizedViewModelsProvider) {
         self.viewModelsAndViews = viewModelsAndViews
+        self.viewModelsAndViews?.onPriorityChanged = onViewModelsPriorityChange
```
allendav, line 14: @ctarda - I am a little concerned that this will make a strong reference preventing this view controller from being destroyed. I'll double check that but your input is especially welcome.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
```diff
@@ -3,18 +3,32 @@ import UIKit
...
+    }
+
+    private func configureInitialState() {
+        onViewModelsPriorityChange(viewModelAndView: viewModelsAndViews?.priorityViewModelAndView)
+    }
+
+    private func onViewModelsPriorityChange(viewModelAndView: CardReaderSettingsViewModelAndView?) {
+        // For now, just update the label with the view identifer we should display
+        // A later PR will actually add presented views
+        temporaryLabel.text = viewModelAndView?.viewIdentifier ?? "Loading"
```
allendav, line 31: This "loading" string is not localized because this is temporary and will be removed in an upcoming PR

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
ctarda: I don't see any reference cycles here, at least not in plain sight. Let's keep an eye on this anyway, we can always refer back here if we notice something weird going on

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsViewModelsOrderedList.swift
```diff
@@ -0,0 +1,59 @@
...
+        /// And then immediately get a priority view if possible
+        reevaluatePriorityViewModelAndView()
+    }
+
+    private func onDidChangeShouldShow(_ : CardReaderSettingsTriState) {
+        reevaluatePriorityViewModelAndView()
+    }
+
+    private func reevaluatePriorityViewModelAndView() {
+        let newPriorityViewModelAndView = viewModelsAndViews.first(
```
ctarda: I am not sure if this is overthinking it too much, but `first` should return an optional, so maybe this could be written as (pseudocode)
```
guard let newThingy = thingies.first(), newThingy!= oldThingy else { return }

oldThingy = newThingy
```

It feels more "familiar" from a Swift point of view.

ctarda (approved):
https://user-images.githubusercontent.com/2722505/117202415-c7f2cd80-adbb-11eb-99b8-06040d7a845c.MP4



:shipit: 

It work as advertised. One suggestion to consider about the syntax of a method, but other than that 🚀

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsViewModelsOrderedList.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4123/commits/ec97ab87df9ba979a66affb8fc681c35edfa6f19

koke: In the case where `first` doesn't find any matching model and returns `nil`, shouldn't this replace the existing `priorityViewModelAndView` with that `nil`? In that case, I think the first attempt was correct.

allendav: Ooooh. Yeah. I'll fix this in the next PR. Thank you @koke !

Issue #4033, opened 2021-04-26
[Mobile Payments] [Receipts] Iterate on layout of printed receipt


ctarda:
Approximate the layout of the receipt to this:

<img width="336" alt="Screen Shot 2021-05-04 at 05 06 22" src="https://user-images.githubusercontent.com/2722505/116982212-83214680-ac96-11eb-9867-d22253a6867a.png">

* Add icons for credit card brands
* Approximate text styles to the designs
* Approximate header to designs.
* Add section about questions and merchant contact details.
* Add number of items for each line.

For context: p91TBi-4KR-p2

koke:
I started looking into those card icons, and I think they look nice on screen or larger paper, but I'm wondering if they would be readable at all on B/W small roll paper 🤔 

![label-1-output](https://user-images.githubusercontent.com/8739/117624875-b1f75b00-b175-11eb-8229-6cf9581e6a9e.jpg)

koke:
> Add section about questions and merchant contact details

Where should this data come from? I don't see any phone number set up in the backend. I see an email in Settings > Email > "From" address, but I'm not sure that's what we want to use for receipts.

koke:
Added another bullet point for number of items, which is missing now:

![Screen Shot 2021-05-11 at 11 09 43](https://user-images.githubusercontent.com/8739/117790463-a9705480-b249-11eb-9f9f-17781093cd38.png)

![Simulator Screen Shot - iPhone 12 - 2021-05-11 at 11 09 38](https://user-images.githubusercontent.com/8739/117790450-a5dccd80-b249-11eb-96f1-a56f2b6cb540.png)

koke:
@ctarda is this good to close for this iteration? I've logged #4250 separately, and maybe @allendav can open new issues for things found during testing with an actual printer.

ctarda:
Yep, @koke ! It sounds like this one can be closed. Closing now!

PR #4127, opened 2021-05-06
[Mobile Payments] Sample Receipt Printer App

koke:
Developed as part of #4033, but not really specific to that issue.

In the current state, working on receipts can be time-consuming. To test changes, you need to run the app, go to settings and make sure a reader is connected, then go through the payment flow for an order and print a receipt. This adds plenty of friction when you try to understand how the UIKit printing system actually works, and tweak the layout and receipt design accordingly.

I created a new SwiftUI app which exposes some options to create a sample receipt, then gives you the option to print, email, or preview it (only the printing is implemented for now).

So far, the only option is the number of items to include so we can play with the length of the receipt, but we can introduce more options in the future, like type of card

Simulator Screen Shot - iPhone 12 - 2021-05-06 at 15 59 14 |  Simulator Screen Shot - iPhone 12 - 2021-05-06 at 15 53 58
-|-

Update release notes:


allendav (approved):
This is a great idea! Thank you for creating this @koke

No suggestions on code. Tests well. :shipit:

PR #4128, opened 2021-05-06
[Mobile Payments] Implement support for emailing a receipt right after a payment has been captured

### LINKED ISSUE #4014: [Mobile Payments] [Receipts] Email receipts

ctarda:
Closes #4014 

https://user-images.githubusercontent.com/2722505/117327291-6df81380-ae60-11eb-9793-70405cd0a8fc.mov

## Changes
* Make ReceiptRenderer public. Not ideal, but, in my opinion, not that terrible either. That being said, we could maybe declare two different "receipt renderers", one for printed receipts and another one for emailed receipts, and extract the code that generates the body of the receipt as HTML to an aggregated entity. Let's see how it goes.
* Move the "footer" (which was pretty much just the two fields we need to include for regulatory reasons) to the body of the receipt.
* Add support for a new Action to the ReceiptsStore to generate the body of the receipt as HTML
* Updates to OrderDetailsViewController and OrderDetailsViewModel to trigger that action, add localized strings, present the mail controller... All the good stuff.

## How to test
Testing requires:

* Build and run on a device.
* Tap "Connect card reader"
* Tap "Connect to reader"
* Tap the "Email Receipt" button
* The standard iOS email composing sheet is presented. If there is a customer email address on file, it should be pre-populated with it. The subject should read "Your receipt from STORE_NAME". The receipt should contain, at this point, a list of the order items, the total, and two fields we need to include into the receipt for regulatory reasons.

Update release notes:


Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -2,7 +2,7 @@ import UIKit
 
 /// Renders a receipt in an AirPrint enabled printer.
 /// To be properly implemented in https://github.com/woocommerce/woocommerce-ios/issues/3978
-final class ReceiptRenderer: UIPrintPageRenderer {
+public final class ReceiptRenderer: UIPrintPageRenderer {
```
ctarda, line 5: As I mentioned in the PR description, we could probably extract what htmlContent() does to a separate class, and make that new class public instead of this one

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalSuccess.swift
```diff
@@ -47,7 +47,7 @@ final class CardPresentModalSuccess: CardPresentPaymentsModalViewModel {
 
     func didTapSecondaryButton(in viewController: UIViewController?) {
         emailReceiptAction()
-        viewController?.dismiss(animated: true)
+        viewController?.dismiss(animated: false)
```
ctarda, line 50: Probably not strictly necessary, but it seemed to make the presentation of the email controller less janky in my tests 🤷🏼

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -21,6 +21,20 @@ final class PaymentCaptureOrchestrator {
                                      onClearMessage: onClearMessage,
                                      onCompletion: onCompletion)
     }
+
+    func printReceipt(for order: Order, params: CardPresentReceiptParameters) {
```
ctarda, line 25: This was done directly in the view controller, so it's time to move it here

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -118,6 +118,22 @@ final class OrderDetailsViewModel {
         }
     }
 
+    /// The customer's email address, if available
+    ///
+    var customerEmail: String? {
```
ctarda, line 123: A couple of new fields for the email controller

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -109,10 +96,28 @@ private extension ReceiptRenderer {
                         <h3>\(Localization.summarySectionTitle.uppercased())</h3>
                         \(summaryTable())
                     </p>
+                    <p>
+                        \(regulatoryItems())
```
allendav: Could we rename as requiredItems ?

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -133,6 +138,20 @@ private extension ReceiptRenderer {
...
+    private func regulatoryItems() -> String {
+        guard let emv = parameters.cardDetails.receipt else {
+            return "<br/>"
+        }
+
+        /// According to the documentation, only `Application name` and `AID`
+        /// are required in the US.
+        /// https://stripe.com/docs/terminal/checkout/receipts#custom
+        return """
+            \(Localization.applicationName): \(emv.applicationPreferredName)<br/>
```
allendav, line 151: What prevents the applicationPreferredName or dedicatedFileName from containing html tags and/or < or > ? We should probably be filtering these strings to strip tags and brackets.

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -109,10 +96,28 @@ private extension ReceiptRenderer {
...
+        headerHeight = Constants.headerHeight
+        footerHeight = Constants.footerHeight
+    }
+
+    private func configureFormatter() {
+        let formatter = UIMarkupTextPrintFormatter(markupText: htmlContent())
+        formatter.perPageContentInsets = .init(top: Constants.headerHeight, left: Constants.marging, bottom: Constants.footerHeight, right: Constants.marging)
+
+        addPrintFormatter(formatter, startingAtPageAt: 0)
+    }
```
allendav, line 120: I didn't catch it before, but same question for line.title : how do we avoid tags or brackets breaking the generated html?

allendav:
I'm ok with this as an interim approach, but we miss out on a lot of receipt logic that WooCommerce core has and that extensions our merchants might be using also have to customize receipts that WooCommerce core generates. Is this an interim approach?

allendav (approved):
Tests fine. I was able to collect payment for an order and get to a draft email with the correct details from the order filled in. :shipit: A few comments on the code to consider.

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
ctarda: Should we strip them or scape them? I mean, if we remove tags and brackets we would be changing the actual content of the string... I've logged a ticket to look into this in more detail: #4137

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
ctarda: We don't, I missed this. I've logged #4137 for it

ctarda:
> I'm ok with this as an interim approach, but we miss out on a lot of receipt logic that WooCommerce core has and that extensions our merchants might be using also have to customize receipts that WooCommerce core generates. Is this an interim approach?

I don't follow. We are trying to approach the receipts from the point of view of this: p91TBi-4KR-p2 How are extensions or core a concern here?

PR #4130, opened 2021-05-06
[Mobile Payments] Pass customer email when creating a payment intent

### LINKED ISSUE #4129: [Mobile Payments] Pass customer email when initiating a charge
Context: p91TBi-55F-p2#comment-4515

We have all the infrastructure in place to pass the customer email address to the Stripe Terminal SDK when creating the initial Payment Intent.

However, we are not actually passing that value. 

We should pass the email when initializing PaymentParameters at the very beginning of the sequence that captures a payment

ctarda:
Closes #4129 

## Changes
* Pass customer email when initiating a payment capture, as discussed in p91TBi-55F-p2#comment-4515

## How to test
**TL:DR; Add a breakpoint to StripeCardReaderService in line 236. The breakpoint should be hit when we start capturing a payment, and a `po parameters` should show the customer email address as part of the payload we pass to the Stripe Terminal SDK.**

Testing requires:

* Build and run on a device.
* Tap "Connect card reader"
* Tap "Connect to reader"
* Tap the "Email Receipt" button
* The standard iOS email composing sheet is presented. If there is a customer email address on file, it should be pre-populated with it. The subject should read "Your receipt from STORE_NAME". The receipt should contain, at this point, a list of the order items, the total, and two fields we need to include into the receipt for regulatory reasons.

Update release notes:


allendav (approved):
Breakpoint shows email where expected. :shipit:

PR #4132, opened 2021-05-06
Plugins: Update Yosemite with SitePluginAction and SitePluginStore

### LINKED ISSUE #4113: Plugins: Update Yosemite layer
## Tasks
- [ ] Add PluginAction
- [ ] Add PluginStore & implement syncSitePlugins
- [ ] Extend Storage plugin model to convert to and from Networking model

itsmeichigo:
Closes #4113 

## Description
As part of #3921, Yosemite layer needs to be updated with new action and store for synchronizing site plugins. Synchronization should take care of all the following:
- Fetch plugins from remote.
- Store plugins locally.
- Update plugins details if they were already stored earlier.
- Remove any stale plugins (plugins that have been stored but no longer exist on remote).

## Solution
- New action `SitePluginAction`. This action only has option to synchronize plugins for the scope of #3921.
- New store `SitePluginStore` to handle synchronization.
- Exported `Networking.SitePlugin` and `Storage.SitePlugin` symbols for access from UI layer without needing to import Networking and Storage modules.
- New helper function to remove stale Storage SitePlugin entities.
- Also: generated copiable method for SitePlugin for convenient testing.

## Testing
Updated test for `StorageType+Deletion` extension and added tests for `SitePluginStorer`, so CI need to pass.

Update release notes:


Storage/Storage/Tools/StorageType+Deletions.swift
```diff
@@ -118,4 +118,15 @@ public extension StorageType {
             deleteObject($0)
         }
     }
+
+    /// Deletes all of the stored `StalePlugin` entities with a specified `siteID` whose name is not included in `currentPluginNames` array.
```
Ecarrion:
```suggestion
    /// Deletes all of the stored `SitePlugin` entities with a specified `siteID` whose name is not included in `currentPluginNames` array.
```

Storage/Storage/Tools/StorageType+Deletions.swift
```diff
@@ -118,4 +118,15 @@ public extension StorageType {
...
+    ///
+    func deleteStalePlugins(siteID: Int64, currentPluginNames: [String]) {
```
Ecarrion:
```suggestion
    func deleteStalePlugins(siteID: Int64, activePluginNames: [String]) {
```
I think `active` expressed the intention better than `current`. What do you think?

Yosemite/Yosemite/Stores/SitePluginStore.swift
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// MARK: - Network request
+//
+private extension SitePluginStore {
+    func synchronizeSitePlugins(siteID: Int64, completionHandler: @escaping (Result<Void, Error>) -> Void) {
+        remote.loadPlugins(for: siteID) { [weak self] result in
+            guard let self = self else {
+                return
+            }
```
Ecarrion: Nit: we allow ourselves to [write these](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/coding-style-guide.md#braces) as a one liner!

Yosemite/Yosemite/Stores/SitePluginStore.swift
```diff
@@ -0,0 +1,94 @@
...
+                    return plugin
+                }
+                return storage.insertNewObject(ofType: StorageSitePlugin.self)
+            }()
+
+            storagePlugin.update(with: readonlyPlugin)
+
+            // remove stale plugins
+            let currentPluginNames = readonlyPlugins.map(\.name)
+            storage.deleteStalePlugins(siteID: siteID, currentPluginNames: currentPluginNames)
```
Ecarrion: I think this is performed for each plugin right?
Should we run it only when we finish upserting the list of new plugins?

Yosemite/YosemiteTests/Stores/SitePluginStoreTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        // Then
+        XCTAssertTrue(result.isSuccess)
+        XCTAssertEqual(viewStorage.countObjects(ofType: StorageSitePlugin.self), 5) // number of plugins in json file
+    }
+
+    func test_synchronizeSitePlugins_removes_stale_plugins_correctly() {
+        // Given
+        let stalePlugin = SitePlugin.fake().copy(siteID: sampleSiteID, name: "Stale Plugin")
```
Ecarrion: How do you feel about these functions to create fake objects for tests?

Yosemite/YosemiteTests/Stores/SitePluginStoreTests.swift
```diff
@@ -0,0 +1,78 @@
...
+        let result: Result<Void, Error> = waitFor { promise in
+            let action = SitePluginAction.synchronizeSitePlugins(siteID: self.sampleSiteID) { result in
+                promise(result)
+            }
+            store.onAction(action)
+        }
+
+        // Then
+        XCTAssertTrue(result.isSuccess)
+        XCTAssertEqual(viewStorage.countObjects(ofType: StorageSitePlugin.self), 5) // number of plugins in json file
```
Ecarrion, line 77: I think you could also check that the old plugin does not exist.
```swift
XCTAssertNil(viewStorage.loadPlugin(siteID: stalePlugin.siteID, name: stalePlugin.name))
```

Ecarrion (changes requested):
Looks very good in general! 🎉 
But left you some comments to address!

Storage/Storage/Tools/StorageType+Deletions.swift
itsmeichigo: I'm not quite happy with `current` either, but `active` does not seem right since plugins can be active/inactive, so the name can be ambiguous. How about `valid`? 🤷‍♀️

Ecarrion: I think we can go with `installed` plugins!

Yosemite/Yosemite/Stores/SitePluginStore.swift
itsmeichigo: Good catch! This really should not happen 🤦‍♀️

Yosemite/YosemiteTests/Stores/SitePluginStoreTests.swift
itsmeichigo: I like the convenience of these functions - I've been using [Fakery](https://github.com/vadymmarkov/Fakery) in my unit tests at work but I prefer the code generation solution we have here - there's no need for 3rd party libraries.

itsmeichigo: Oh actually we're still using [Sourcery](https://github.com/krzysztofzablocki/Sourcery)... but with Sourcery Pro out now, we may use the extension to replace it.

itsmeichigo:
Thanks, @Ecarrion for your review! I have added my changes, please check again if they're all good.

Storage/Storage/Tools/StorageType+Deletions.swift
```diff
@@ -119,11 +119,11 @@ public extension StorageType {
         }
     }
 
-    /// Deletes all of the stored `StalePlugin` entities with a specified `siteID` whose name is not included in `currentPluginNames` array.
+    /// Deletes all of the stored `SitePlugin` entities with a specified `siteID` whose name is not included in `currentPluginNames` array.
```
Ecarrion:
```suggestion
    /// Deletes all of the stored `SitePlugin` entities with a specified `siteID` whose name is not included in `instaledPluginNames` array.
```
😅

Yosemite/Yosemite/Stores/SitePluginStore.swift
```diff
@@ -85,10 +83,10 @@ private extension SitePluginStore {
...
 
             storagePlugin.update(with: readonlyPlugin)
-
-            // remove stale plugins
-            let currentPluginNames = readonlyPlugins.map(\.name)
-            storage.deleteStalePlugins(siteID: siteID, currentPluginNames: currentPluginNames)
         }
+
+        // remove stale plugins
+        let currentPluginNames = readonlyPlugins.map(\.name)
```
Ecarrion:
```suggestion
        let installedPluggins = readonlyPlugins.map(\.name)
```

Yosemite/YosemiteTests/Stores/SitePluginStoreTests.swift
Ecarrion: Good to know, I haven't explored in detail `Sourcery Pro` yet 😅

Ecarrion (approved):
🚀 

Just a few super nits and we should be good to go!

itsmeichigo:
I've replaced all the occurrences of the old name in the new commit 😅

Ecarrion (approved):

PR #4134, opened 2021-05-06
Add-ons: Add track events

Ecarrion:
# Why

As the order add-on workaround gets ready to be released, this PR adds some event tracking to the feature.

- orderAddOnsViewed
- betaFeaturesSwitchToggled

# How

- In the beta settings screen, fire `betaFeaturesSwitchToggled` when the switch is toggled.
- In the order add-on list screen, fire `orderAddOnsViewed` when the screen appears.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
@@ -279,3 +279,25 @@ extension WooAnalyticsEvent {
         }
     }
 }
+
+// MARK: - Variations
+//
+extension WooAnalyticsEvent {
+    // Namespace
+    enum OrderDetailAddOns {
+        /// Common event keys
+        ///
+        private enum Keys {
+            static let state = "state"
+            static let addOns = "add-ons"
+        }
+    }
+
+    static func betaFeaturesSwitchToggled(isOn: Bool) -> WooAnalyticsEvent {
+        WooAnalyticsEvent(statName: .settingsBetaFeaturesOrderAddOnsToggled, properties: ["state": isOn])
+    }
+
+    static func orderAddOnsViewed(addOnNames: [String]) -> WooAnalyticsEvent {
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/Analytics/WooAnalyticsEvent.swift
```diff
@@ -279,3 +279,25 @@ extension WooAnalyticsEvent {
         }
     }
 }
+
+// MARK: - Variations
```
rachelmcr: nit: Update this comment to refer to add-ons.

```suggestion
// MARK: - Order Detail Add-Ons
```

rachelmcr (commented):
Looks pretty good! I tested and confirmed the Tracks events are fired at the expected times. I have a tiny nit and a couple questions:

> The tracks events must be validated in the Tracks system.

How should I check this? I checked the Tracks Events Directory and don't see these events in the list (even when I select `All` in the events filters). Not sure if I'm looking in the wrong place for this.

> Verify the internal tracks spreadsheet has also been updated.

I noticed that for `*_settings_beta_features_order_add-ons_toggled` the spreadsheet lists the properties as `state="on|off"`. Should that be `state=true|false` since we're passing a boolean for that property? Or should the event properties actually be `on` and `off` instead of `true` and `false` here?

Ecarrion:
Hey @rachelmcr, thanks for the review.

> How should I check this? I checked the Tracks Events Directory and don't see these events in the list (even when I select All in the events filters). Not sure if I'm looking in the wrong place for this.

I normally follow the `Validating Tracks Events` post in field-guide. And see that the events appear. I didn't validate them before your review tho 😅 

> I noticed that for *_settings_beta_features_order_add-ons_toggled the spreadsheet lists the properties as state="on|off". Should that be state=true|false since we're passing a boolean for that property? Or should the event properties actually be on and off instead of true and false here?

Good catch. I've updated the property in code.

rachelmcr (approved):
Looks great! :shipit: 

> I normally follow the Validating Tracks Events post in field-guide. And see that the events appear.

That's how I checked and I'm still not seeing the events there, but I must be doing something wrong. I'll ping you on Slack to clarify.

Issue #4114, opened 2021-05-04
Plugins: Update UI layer


Ecarrion:
## Tasks
- [ ] Update DefaultStoresManager to sync plugins
- [ ] Update Settings screen for an entry point to the plugin list
- [ ] Add PluginListViewModel
- [ ] Add PluginListViewController
- [ ] Build SwiftUI refresh control (if possible)

itsmeichigo:
Thanks @Ecarrion 🙂

itsmeichigo:
@Ecarrion  I think this issue can be closed now right?

PR #4135, opened 2021-05-07
Plugins: Add new feature flag and add new entry point for plugin list on Settings screen

### LINKED ISSUE #4112: Plugins: Update Storage layer

itsmeichigo:
closes #4112 
part of #4114
Implements #3921 

## Description
In order to let users find the list of plugins added to their selected sites, a new entry point to the Plugins screen should be added to Settings screen.

This feature is not intended to be released to the wild yet so I also added a feature flag to make it visible in only local and alpha builds.

## Solution
- Added new case `sitePlugins` to  `FeatureFlag` enum
- Updated `DefaultFeatureFlagService` to enable site plugins feature only for local and alpha builds.
- Updated Settings screen with new section Plugins. This section contains only one row Plugins to redirect to new Plugins screen (the action is pending for implementation coming soon).
- The Plugins section is only available when `sitePlugins` feature flag is enabled.

## Testing
When built in local and alpha environments, the app should show Settings screen with Plugins section like screenshot below:

P.s: `SettingsViewController` has not yet been migrated to MVVM pattern so it cannot be unit tested for now. It is desirable to handled this migration soon with dependency injection of `FeatureFlagService` to double check for visibility of beta features like this one, as we sure cannot test the app in production environment.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -381,7 +381,9 @@ private extension SettingsViewController {
         }
     }
 
-    func sitePluginsWasPressed() {}
+    func sitePluginsWasPressed() {
+        // TODO: Pending implementation for issue #4114
+    }
 
     func supportWasPressed() {
         ServiceLocator.analytics.track(.settingsContactSupportTapped)
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -167,7 +167,7 @@ private extension SettingsViewController {
 
         // Plugins
         if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.sitePlugins) {
-            sections.append(Section(title: pluginsTitle, rows: [], footerHeight: UITableView.automaticDimension))
+            sections.append(Section(title: pluginsTitle, rows: [.plugins], footerHeight: UITableView.automaticDimension))
         }
 
         // Store Settings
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -241,6 +252,13 @@ private extension SettingsViewController {
         )
     }
 
+    func configurePlugins(cell: BasicTableViewCell) {
+        cell.selectionStyle = .default
+        cell.accessoryType = .disclosureIndicator
+        cell.textLabel?.text = NSLocalizedString("Plugins", comment: "Navigates to Plugins screen."
+        )
```
Ecarrion: Nit: This should go in the previous line.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -363,6 +381,8 @@ private extension SettingsViewController {
         }
     }
 
+    func sitePluginsWasPressed() {}
```
Ecarrion: nit: Probably better to add a `TODO: To be implemented later #GH-issue` so we communicate with the rest of the team why this method is empty.

Ecarrion (changes requested):
Looks great. I'm requesting changes just for a few nits! 😅 

> P.s: SettingsViewController has not yet been migrated to MVVM pattern so it cannot be unit tested for now. It is desirable to handled this migration soon with dependency injection of FeatureFlagService to double check for visibility of beta features like this one, as we sure cannot test the app in production environment.

How do you feel about creating an issue for this so we can prioritize it accordingly?

itsmeichigo:
@Ecarrion I have fixed the formatting issues as per your suggestions 👍 And yes, I'll create a separate issue regarding testability for the SettingsViewController!

Edit: the issue is [here](https://github.com/woocommerce/woocommerce-ios/issues/4142)!

Ecarrion (approved):
🔝 

I think #4112 can be closed now, right?

itsmeichigo:
@Ecarrion - yes please close #4112, thanks!

PR #4140, opened 2021-05-07
[Mobile Payments] Disconnect from reader

### LINKED ISSUE #4031: [Mobile Payments] Implement disconnection from reader

ctarda:
Closes #4031 

Add support for disconnecting from the currently connected reader:

https://user-images.githubusercontent.com/2722505/117460522-99d7cf80-af1a-11eb-9052-a9db2eb8dba2.MP4

## Changes
* Implemented `disconnect()` in the integration with the Stripe Terminal SDK (StripeCardReaderService)
* Added an action to CardPresentPaymentAction and its implementation to CardPresentPaymentStore.
* No new tests. Testing the integration is still blocked pending #3864 
* Touched the current UI to integrate with the new action. More about that in a comment in the actual diff.

## How to test
Testing requires:

* Build and run on a device.
* Tap "Connect card reader"
* Tap "Connect to reader"
* Tap the "Disconnect" button. 
* The UI should flip to its initial state. The reader led should start blinking again. 
* Try to connect and disconnect a couple more times, just for kicks and giggles.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
@@ -162,10 +162,19 @@ final class CardReaderSettingsViewModel: ObservableObject {
 
     /// Views can call this method to disconnect from the connected reader.
     func disconnectAndForget() {
-        // TODO dispatch an action to disconnect.
-        connectedReader = nil
-        activeView = .connectYourReader
-        activeAlert = .none
+        let action = CardPresentPaymentAction.disconnect { [weak self] result in
+            guard let self = self else {
+                return
+            }
+
+            if result.isSuccess {
+                self.connectedReader = nil
+                self.activeView = .connectYourReader
+                self.activeAlert = .none
+            }
+        }
+
+        ServiceLocator.stores.dispatch(action)
     }
 
     /// Called when a reader has been connected to.
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
@@ -163,15 +163,13 @@ final class CardReaderSettingsViewModel: ObservableObject {
     /// Views can call this method to disconnect from the connected reader.
     func disconnectAndForget() {
         let action = CardPresentPaymentAction.disconnect { [weak self] result in
-            guard let self = self else {
+            guard let self = self, result.isSuccess else {
                 return
             }
 
-            if result.isSuccess {
-                self.connectedReader = nil
-                self.activeView = .connectYourReader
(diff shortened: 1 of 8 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -162,10 +162,19 @@ final class CardReaderSettingsViewModel: ObservableObject {
 
     /// Views can call this method to disconnect from the connected reader.
     func disconnectAndForget() {
-        // TODO dispatch an action to disconnect.
-        connectedReader = nil
-        activeView = .connectYourReader
-        activeAlert = .none
+        let action = CardPresentPaymentAction.disconnect { [weak self] result in
```
ctarda, line 165: This UI is temporary and this view model may or may not be thrown away soon. 

Based on that, I didn't add any user facing errors. If this view model is meant to remain then let's log a new ticket to present an error here, in the unlike event of a failure when disconnecting.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReaders/CardReaderSettingsViewModel.swift
```diff
@@ -162,10 +162,19 @@ final class CardReaderSettingsViewModel: ObservableObject {
...
+            guard let self = self else {
+                return
+            }
+
+            if result.isSuccess {
```
koke: Not super important, but maybe that condition could also go in the `guard` above

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -101,6 +103,19 @@ private extension CardPresentPaymentStore {
...
+        cardReaderService.disconnect().sink(receiveCompletion: { error in
+            switch error {
+            case .failure(let error):
+                onCompletion(.failure(error))
+            default:
+                break
+            }
+        }, receiveValue: { (result) in
+            onCompletion(.success(result))
+        }).store(in: &cancellables)
```
koke: I'd suggest using a `Subscribers.Sink` subscriber instead of storing the cancellable, like the one in `startCardReaderDiscovery` from #4120. Since this is using a `Future`, it will send a completion right after sending its only value, so we can get automatic cleanup with `Sink`.

koke (approved):
Tested and it works fine. I notice an issue that I think it just hasn't been implemented yet, but when you connect a reader, navigate away from settings and go to the card reader management screen again, it doesn't know a reader is connected.

I left one suggestion about the cancellable, but we could also tackle that as part of #3926

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
ctarda: Good catch, thanks!

ctarda:
Thanks for the review @koke I just pushed an update to incorporate our feedback (and great point about the Subscriber).

I'll merge as soon as ci is ✅

PR #4144, opened 2021-05-07
[Mobile Payments] 2-6 Add No Known No Connected Readers ViewController

### LINKED ISSUE #4052: [Mobile Payments] Card Reader Settings v2-6: CardReaderUnknownViewController
CardReaderUnknownViewController:
- Accepts CardReaderUnknownViewModel
- Learn more does nothing
- Connect Card Reader Button does nothing
- Have  CardReaderSettingsPresentingViewController instantiate and push it properly (using the shouldShow loop)
- Have CardReaderSettingsPresentingViewController detect when the user pops the presentedVC and have CardReaderSettingsPresentingViewController dismiss itself too

allendav:
Closes #4052

Changes:

- This is the next step in Card Reader Settings v2. In this step we add the first child view controller as a pattern to follow. We also add a bit of the second for grins.
- Note especially the code in `onViewModelsPriorityChanged` which adds and removes the child view controller as shouldShow dictates
- Note that the no-known no-connected view controller section and rows code is largely copied from the v1 implementation

To test:

- Launch the app
- Navigate to Settings (Gear) and then Manage Card Readers V2
- Ensure it shows the familiar "connect your reader" view - note that connecting the button comes in the next PR
- Navigate back to Settings (Gear)
- Navigate to Manage Card Readers (not V2)
- Connect your reader
- Navigate to Manage Card Readers V2
- Ensure it shows that the connected view is under construction (coming in 2-10)

Things to catch up on:
- Row separators should be suppressed
- Row height should be managed

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
@@ -57,6 +57,7 @@ private extension CardReaderSettingsConnectedViewController {
     }
 
     func configureTable() {
+        tableView.rowHeight = UITableView.automaticDimension
         tableView.dataSource = self
         tableView.delegate = self
         tableView.reloadData()
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
@@ -1,0 +1,187 @@
+import Foundation
+import UIKit
+
+/// This view controller is used when a reader is currently connected. It assists
+/// the merchant in updating and/or disconnecting from the reader, as needed.
+///
+/// TODO: This is just a placeholder for now. The implementation of this view controller will
+/// begin in earnest with #4056
+///
+final class CardReaderSettingsConnectedViewController: UIViewController {
+
+    /// Main TableView
+    ///
+    @IBOutlet weak private var tableView: UITableView!
+
+    /// ViewModel
+    ///
+    private var viewModel: CardReaderSettingsConnectedViewModel?
+
+    /// Table Sections to be rendered
+    ///
+    private var sections = [Section]()
+
+    // MARK: - Overridden Methods
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        registerTableViewCells()
+        configureNavigation()
+        configureSections()
+        configureTable()
+    }
+
+    func configure(viewModel: CardReaderSettingsConnectedViewModel) {
+        self.viewModel = viewModel
+    }
+}
+
+// MARK: - View Configuration
+//
+private extension CardReaderSettingsConnectedViewController {
+
+    /// Set the title and back button.
+    ///
+    func configureNavigation() {
+        title = Localization.title
+    }
+
+    /// Setup the sections in this table view
+    ///
+    func configureSections() {
+        sections = [Section(title: nil,
+                            rows: [
+                                .temporaryHeader
+                            ])]
+    }
+
+    func configureTable() {
+        tableView.dataSource = self
+        tableView.delegate = self
+        tableView.reloadData()
+    }
+
+    /// Register table cells.
+    ///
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    /// Cells currently configured in the order they appear on screen
+    ///
+    func configure(_ cell: UITableViewCell, for row: Row, at indexPath: IndexPath) {
+        switch cell {
+        case let cell as TitleTableViewCell where row == .temporaryHeader:
+            configureHeader(cell: cell)
+        default:
+            fatalError()
```
ctarda, line 81: 💣 !

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -0,0 +1,187 @@
...
+        cell.selectionStyle = .none
+    }
+}
+
+// MARK: - Convenience Methods
+//
+private extension CardReaderSettingsConnectedViewController {
+
+    func rowAtIndexPath(_ indexPath: IndexPath) -> Row {
+        return sections[indexPath.section].rows[indexPath.row]
```
ctarda, line 96: I always have mixed feelings here. On the one hand, there is a potential crash here if indexPath.section is greater that the number of sections. Which shouldn't have, but...

On the other hand, guarding against it might be too much. I'll leave that out to you 😊

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -0,0 +1,187 @@
...
+    func numberOfSections(in tableView: UITableView) -> Int {
+        return sections.count
+    }
+
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return sections[section].rows.count
+    }
+
+    func tableView(_ tableView: UITableView, heightForHeaderInSection section: Int) -> CGFloat {
+        return UITableView.automaticDimension
```
ctarda, line 113: Nice

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -0,0 +1,187 @@
...
+    }
+
+    func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
+        return UITableView.automaticDimension
```
ctarda, line 117: Nice aussi. Since both header and footer support autosizing, would you consider doing the same for the cells? 

Wherever you configure the table view, you could set `        tableView.rowHeight = UITableView.automaticDimension`

ctarda (approved):
it works great:


https://user-images.githubusercontent.com/2722505/117496500-239c9280-af45-11eb-8f21-22f0908ad6a7.MP4

A couple of suggestions for you to consider. Support for autosizing cells, in particular, can be tricky, so that's 100% deferrable.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: I could guard for that here, but I couldn't find a way to return `nil` from `cellForRowAt` in Swift. I'm willing to change it later if you have any hints :)

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4144/commits/7fc5fb81612016e63aa14cb6626d054dce13b96a

PR #4146, opened 2021-05-08
Plugins: Sync site plugins automatically

itsmeichigo:
Part of #4114 and #3921

## Description
To handle automatic synchronization of site plugins, the data has to be synced in 3 scenarios: 
- After user is logged in and select a site they own.
- When app launches if user is already authenticated.
- After user change selected site.

These 3 cases are all covered when restoring session site, so this is the perfect place to trigger synchronization of plugins.

## Solution
- Create a SitePluginStore instance and retain it in AuthenticatedState
- Add new private method to sync site plugin in `restoreSessionSiteIfPossible` method of `DefaultStoresManager`.
- Since there's a chance of silent failure when trying to dispatch an action whose accompanying store hasn't been retained, I also added an `assertionFailure` to warn about the issue. This does not change the main behavior of the Dispatcher, so I think it's appropriate to include it in this PR.

## Testing
It's not yet possible to write tests to check for appropriate triggering of actions in `DefaultStoresManager`. For now, a quick test would be to add a breakpoint in `onAction` method of `SitePluginStore`, make sure that the application pauses there when running the above 3 scenarios.

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Base/Dispatcher.swift
+++ b/Yosemite/Yosemite/Base/Dispatcher.swift
@@ -69,7 +69,12 @@ public class Dispatcher {
     public func dispatch(_ action: Action) {
         assertMainThread()
 
-        processors[action.identifier]?.onAction(action)
+        // Avoid silent failure when a store is not retained
+        guard let processor = processors[action.identifier] else {
+            assertionFailure("⛔️ No processor found for \(action.identifier)!")
+            return
+        }
+        processor.onAction(action)
     }
 }
 
--- a/Yosemite/Yosemite/Base/Dispatcher.swift
+++ b/Yosemite/Yosemite/Base/Dispatcher.swift
@@ -71,8 +71,7 @@ public class Dispatcher {
 
         // Avoid silent failure when a store is not retained
         guard let processor = processors[action.identifier] else {
-            assertionFailure("⛔️ No processor found for \(action.identifier)!")
-            return
+            return DDLogWarn("⚠️ No processor found for \(action.identifier)!")
         }
         processor.onAction(action)
     }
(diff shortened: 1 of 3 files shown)


Yosemite/Yosemite/Base/Dispatcher.swift
```diff
@@ -69,7 +69,12 @@ public class Dispatcher {
     public func dispatch(_ action: Action) {
         assertMainThread()
 
-        processors[action.identifier]?.onAction(action)
+        // Avoid silent failure when a store is not retained
+        guard let processor = processors[action.identifier] else {
+            assertionFailure("⛔️ No processor found for \(action.identifier)!")
+            return
```
Ecarrion: Pro tip: Since `assertion failure` returns `Void` under the hood, you can do
```suggestion
            return assertionFailure("⛔️ No processor found for \(action.identifier)!")

```

And keep your guard statements more concise!

Yosemite/Yosemite/Base/Dispatcher.swift
```diff
@@ -69,7 +69,12 @@ public class Dispatcher {
     public func dispatch(_ action: Action) {
         assertMainThread()
 
-        processors[action.identifier]?.onAction(action)
+        // Avoid silent failure when a store is not retained
+        guard let processor = processors[action.identifier] else {
```
Ecarrion, line 73: Nice! I remember being bitten by this in the past!

Ecarrion (approved):
👏 

Works like a charm!

Ecarrion (changes requested):
Just noticed that there is a crash due to the assertion failure that this PR introduces. 

I believe this is because it is possible for the  `deauthenticatedState` to receive and ignore actions.

itsmeichigo:
Oops! I just updated the `assertionFailure` with just `DDLogWarn` to avoid the crash, but still give some warning just in case someone needs the reminder. What do you think?

Ecarrion (approved):
> Oops! I just updated the assertionFailure with just DDLogWarn to avoid the crash, but still give some warning just in case someone needs the reminder. What do you think?

I think that is fair! 👍

PR #4147, opened 2021-05-09
[Mobile Payments] Reset Stripe Terminal SDK when logging out and/or switching stores

### LINKED ISSUE #4111: [Mobile Payments] Support for multiple stores
The StripeTerminal framework requires a connection token to work, which I understand is site-specific.

We should ensure that users with multiple stores set up in the app show the right availability for payments and use the correct connection token.

ctarda:
Closes #4111 

This PR starts where #4140 ends. As pointed out [in this comment](https://github.com/woocommerce/woocommerce-ios/issues/4111#issuecomment-831909896) the Stripe Terminal SDK [supports multiple accounts and provides a mechanism for switching accounts](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(im)clearCachedCredentials).

## Changes
* Add a new action to CardPresentPaymentAction to reset the payments integration, and its implementation in CardPresentPaymentStore. This action calls disconnect and clearCacheCredentials in sequence.
* There be dragons though. `disconnect ` would crash [if called before the Terminal has been provided with a way to fetch connection tokens](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(cpy)shared). So we need to add a bit of state tracking to StripeCardReaderService.

## How to test.
Testing requires:

* Checkout the branch, build and run on device.
* Add a breakpoint to StripeCardReaderService, line 159.
Screen Shot 2021-05-09 at 15 29 47

* Log in, and log out immediately, without doing any operations related to payments. The breakpoint added to StripeCardReaderService should not be hit.
* Log in to a store that does not support payments. Log out or switch to a store that supports payments, the breakpoint should not be hit.
* Log in, switch to a store that supports payments. Connect to a reader. Switch stores, and the breakpoint should be hit. Switch back to the original store, attempt to pair to a reader again. It should work.
* Log in to a store that support payments. Connect to a reader. Log out. The breakpoint should be hit.
* Any variations and combinations of the above that you could think of.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -134,7 +134,6 @@ extension StripeCardReaderService: CardReaderService {
     }
 
     public func clear() {
-        // 🧹
         Terminal.shared.clearCachedCredentials()
     }
 
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -20,8 +20,6 @@ public final class StripeCardReaderService: NSObject {
     private let discoveredStripeReadersCache = StripeCardReaderDiscoveryCache()
 
     private var pendingSoftwareUpdate: ReaderSoftwareUpdate?
-
-    private var initialized: Bool = false
 }
 
 
@@ -63,7 +61,6 @@ extension StripeCardReaderService: CardReaderService {
 
     public func start(_ configProvider: CardReaderConfigProvider) {
         setConfigProvider(configProvider)
-        initialized = true
 
         let config = DiscoveryConfiguration(
             discoveryMethod: .bluetoothProximity,
@@ -112,13 +109,13 @@ extension StripeCardReaderService: CardReaderService {
     }
 
     public func disconnect() -> Future<Void, Error> {
-        return Future() { [weak self] promise in
+        return Future() { promise in
             // Throw an error if the SDK has not been initialized.
             // This prevent a crash when logging out or switching stores before
             // the SDK has been initialized.
             // Why? https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(cpy)shared
             // `Before accessing the singleton for the first time, you must first call setTokenProvider: and setDelegate:.`
-            guard self?.initialized == true else {
+            guard Terminal.hasTokenProvider() else {
                 promise(.failure(CardReaderServiceError.disconnection()))
                 return
             }
@@ -152,7 +149,7 @@ extension StripeCardReaderService: CardReaderService {
         // the SDK has been initialized.
         // Why? https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(cpy)shared
         // `Before accessing the singleton for the first time, you must first call setTokenProvider: and setDelegate:.`
-        guard initialized == true else {
+        guard Terminal.hasTokenProvider() else {
             return
         }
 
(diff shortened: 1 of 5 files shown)


Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -185,6 +187,14 @@ private extension CardPresentPaymentStore {
             onProgress(progress)
         }.store(in: &cancellables)
     }
+
+    func restart() {
+        cardReaderService.disconnect().sink(receiveCompletion: { [weak self] _ in
+            self?.cardReaderService.clear()
+        }, receiveValue: { [weak self] value in
+            self?.cardReaderService.clear()
+        }).store(in: &cancellables)
```
koke, line 198: Shouldn't it be enough to call `clear` on completion?

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -20,6 +20,8 @@ public final class StripeCardReaderService: NSObject {
     private let discoveredStripeReadersCache = StripeCardReaderDiscoveryCache()
 
     private var pendingSoftwareUpdate: ReaderSoftwareUpdate?
+
+    private var initialized: Bool = false
```
koke: I think we might be able to replace this with a call to `Terminal.hasTokenProvider`?

koke:
I don't have a second store yet to test this one, so that'll take a while longer unless someone else does it first. 

My main concern here is that I'm not sure we're clearing everything correctly. I understand the `[clearCachedCredentials](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPTerminal.html#/c:objc(cs)SCPTerminal(im)clearCachedCredentials)` method does exactly what the name says, but we aren't clearing the config provider. Any subsequent action from `Terminal` could ask the provider for a new token that I understand would be associated to the old account. I don't see a way to set the token provider to `nil` or "deinitialize" the terminal, but maybe it needs an empty provider until `start` gets called again.

This doesn't seem likely to cause problems right now as we always reset the config provider right before starting discovery, but more generally I'm a bit worried that the responsibility of keeping state seems a bit blurry between the store, service, and Terminal's own internal state.

Yosemite/Yosemite/Actions/CardPresentPaymentAction.swift
```diff
@@ -41,10 +41,18 @@ public enum CardPresentPaymentAction: Action {
...
+    /// Update card reader firmware.
     case startCardReaderUpdate(onProgress: (Float) -> Void,
                         onCompletion: (Result<Void, Error>) -> Void)
 
+    /// Restarts the card present payments system
+    /// This might imply, but not be limited to:
+    /// 1. Disconnect from a connected reader
+    /// 2. Clear all credentials, cached data
+    /// 3. Reset all status indicators
+    case restart
```
koke: Maybe `reset` would be a better term since this isn't calling `start` again, and if I understand the architecture correctly, the store handling this will be deinitialized and replaced with a new one when we re-authenticate

ctarda:
> Any subsequent action from Terminal could ask the provider for a new token that I understand would be associated to the old account. I don't see a way to set the token provider to nil or "deinitialize" the terminal, but maybe it needs an empty provider until start gets called again.

Once start gets called again (which will need to be in order to start a new discovery process, the token provider will be set again. That happens indirectly, though. When CardPresentPaymentStore calls into cardReaderService.start() it always passes a new instance of the class that provides he logic to support fetching a new token

ctarda:
Thanks for the review @koke 
I pushed an update to rename the action, remove the boolean flag from the service and to remove an extra call to clear

@allendav could you take a look at this one too, please?

allendav (approved):
Nice catch. I switched back and forth between WCPay sites and non WCPay sites with and without the reader connected. The breakpoint was hit when expected and the app never crashed. :shipit:

ctarda:
Thanks folks. I'll merge for now, I just have too much work in flight. We can revisit the service lifecycle if we think we need to.

PR #4151, opened 2021-05-10
Plugins: Add Plugin List screen

itsmeichigo:
Part of #4114 and #3921 
Please make sure that #4146 is merged before testing this.

## Description
To display the list of installed plugins for a selected site, a new view controller needs to be created and showed from the Plugins entry point in the Settings screen. 

## Solution
I chose MVVM pattern and UIKit to build the Plugin List screen since there's no native support for refresh control in SwiftUI and current solutions aren't really performant and straightforward.

### PluginListViewModel
- The view model is initialized with the injection of siteID, stores manager, and storage manager for better testability.
- Added a private results controller to fetch plugins from storage and exposes details about sections and rows used for the plugins table view's data source.
- Added a method `activate` to trigger fetching data.
- Added a method `resyncPlugins` to trigger syncing plugins manually.
- The `resync` method updates plugin list states based on the result of the synchronization.

### PluginListViewController
- Added the view controller to the Dashboard storyboard. 
- For convenient dependency injection, the storyboard extension was updated to [accept `creator` block](https://developer.apple.com/documentation/uikit/uistoryboard/3173166-instantiateviewcontrollerwithide?language=objc).
- Triggered activation of the view model on `viewDidLoad`.
- Added refresh control for manual resyncing.
- Updated different states for the screen: error state view and ghost animation when loading.

## Testing
Unit tests were added to test the table view's section and row details returned from the view model. There are also tests to check for correct action dispatching when triggering the `resyncPlugins` method.

When #4146 is merged to develop, building this PR should result in complete features of #3921.

## Screenshots
| Error State | Loading State | Normal State |
|------------|---------------|--------------|
||||

## Design Review needed
@adamzelinski - there are some problems with the design that I need your opinion on:
- I'm using the default empty state controller for the error state view, which looks a bit different from design. Can you review the screenshot above and let me know if that is alright, or should I update the screen to strictly look like your design?
- The list background color is also a bit different from design. I'm using system table background color, and I'm not sure which color to use to match the design. Let me know if I should change to some other color.

Update release notes:


itsmeichigo:
Closing this to split into smaller PRs!

PR #4152, opened 2021-05-10
Shipping Labels: Add Codegen import for GeneratedFakeable

rachelmcr:
## Description

This fixes the build error in `develop` after https://github.com/woocommerce/woocommerce-ios/pull/4138 was merged. It adds a `Codegen` import to support the changes to `fakeable` types from https://github.com/woocommerce/woocommerce-ios/pull/4117.

## Testing

* Confirm the app builds and tests pass.

## Submitter Checklist

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4153, opened 2021-05-10
Fix storage tests inconsistent order

ealeksandrov:
## Description

This fixes inconsistency of Storage unit tests noticed (but not introduced) in https://github.com/woocommerce/woocommerce-ios/pull/4152

I can reproduce this error for around a half of my test runs.

image

## Test

- Just check CI status.

Update release notes:

### DIFF
--- a/Storage/StorageTests/Tools/StorageTypeDeletionsTests.swift
+++ b/Storage/StorageTests/Tools/StorageTypeDeletionsTests.swift
@@ -24,9 +24,9 @@ class StorageTypeDeletionsTests: XCTestCase {
     func test_deleteStaleAddOnGroups_does_not_delete_active_addOns() throws {
         // Given
         let initialGroups: [AddOnGroup] = [
-            createAddOnGroup(groupID: 123),
-            createAddOnGroup(groupID: 1234),
-            createAddOnGroup(groupID: 12345)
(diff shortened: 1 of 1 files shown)


Storage/StorageTests/Tools/StorageTypeDeletionsTests.swift
```diff
@@ -33,7 +33,7 @@ class StorageTypeDeletionsTests: XCTestCase {
         storage.deleteStaleAddOnGroups(siteID: sampleSiteID, activeGroupIDs: [123, 1234])
 
         // Then
-        let activeGroups = storage.loadAddOnGroups(siteID: sampleSiteID)
+        let activeGroups = storage.loadAddOnGroups(siteID: sampleSiteID).sorted { $0.groupID < $1.groupID }
```
Ecarrion: Thanks for catching this @ealeksandrov.

The `loadAddOnGroups` method is meant to sort items by name. I think that, by passing a name to  `createAddOnGroup` we will be reflecting a more real use-case for future readers. What do you think?

ealeksandrov:
Thanks for suggestion, @Ecarrion! I agree, that's better than a manual sort - pushed an update.

Ecarrion (approved):
🚀 
Thanks again @ealeksandrov

PR #4155, opened 2021-05-10
Plugins: Redirect to Plugin List from Settings screen

itsmeichigo:
Part of #4114 and #3921 

## Description
To display the list of installed plugins for a selected site, a new view controller needs to be created and showed from the Plugins entry point in the Settings screen. 

## Solution
- I chose MVVM pattern and UIKit to build the Plugin List screen since there's no native support for refresh control in SwiftUI and current solutions aren't really performant and straightforward.
- The view model is injected into PluginListViewController upon initialization. Since I added the view controller to the Dashboard storyboard, I also updated the storyboard extension to [accept `creator` block](https://developer.apple.com/documentation/uikit/uistoryboard/3173166-instantiateviewcontrollerwithide?language=objc), which is a good place for dependency injection.
- The view model is incomplete, please ignore it for now.

## Testing
When tapping on the Plugins entry point on the Settings screen, a new screen should be pushed. This screen is currently empty and will be updated in the next PR.

## Note
I left a `TODO` where the tap on Plugins entry point is handled - I'm not sure if we need to track this tap action. Since every other actions in the Settings screen is tracked, I think it makes sense to do that for Plugins also. Let me know if this is necessary and if it is, I'd like to know the name of the event that needs to be tracked.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIStoryboard+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIStoryboard+Woo.swift
@@ -23,9 +23,10 @@ extension UIStoryboard {
 //
 extension UIStoryboard {
     /// Returns a view controller from a Storyboard assuming the identifier is the same as the class name.
+    /// Accepts an optional creator to inject additional dependencies.
     ///
-    func instantiateViewController<T: NSObject>(ofClass classType: T.Type) -> T? {
+    func instantiateViewController<T: UIViewController>(ofClass classType: T.Type, creator: ((NSCoder) -> T?)? = nil) -> T {
         let identifier = classType.classNameWithoutNamespaces
-        return instantiateViewController(withIdentifier: identifier) as? T
+        return instantiateViewController(identifier: identifier, creator: creator)
     }
 }
--- a/WooCommerce/Classes/Extensions/UIStoryboard+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIStoryboard+Woo.swift
@@ -23,10 +23,9 @@ extension UIStoryboard {
 //
 extension UIStoryboard {
     /// Returns a view controller from a Storyboard assuming the identifier is the same as the class name.
-    /// Accepts an optional creator to inject additional dependencies.
     ///
-    func instantiateViewController<T: UIViewController>(ofClass classType: T.Type, creator: ((NSCoder) -> T?)? = nil) -> T {
+    func instantiateViewController<T: NSObject>(ofClass classType: T.Type) -> T? {
         let identifier = classType.classNameWithoutNamespaces
-        return instantiateViewController(identifier: identifier, creator: creator)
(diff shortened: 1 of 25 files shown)


WooCommerce/Classes/Extensions/UIStoryboard+Woo.swift
```diff
@@ -23,9 +23,10 @@ extension UIStoryboard {
...
     /// Returns a view controller from a Storyboard assuming the identifier is the same as the class name.
+    /// Accepts an optional creator to inject additional dependencies.
     ///
-    func instantiateViewController<T: NSObject>(ofClass classType: T.Type) -> T? {
+    func instantiateViewController<T: UIViewController>(ofClass classType: T.Type, creator: ((NSCoder) -> T?)? = nil) -> T {
         let identifier = classType.classNameWithoutNamespaces
-        return instantiateViewController(withIdentifier: identifier) as? T
+        return instantiateViewController(identifier: identifier, creator: creator)
     }
 }
```
Ecarrion: Hey @itsmeichigo, we have been actively trying to migrate away from storyboards. Probably our fault if this is not well documented :-(

Could you please re-work it to use Xibs or Code?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -0,0 +1,20 @@
+import UIKit
+
+class PluginListViewController: UIViewController {
```
Ecarrion: Nit: This can be `final`

Ecarrion: Also, could you add a quick description of the class?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
```diff
@@ -0,0 +1,13 @@
...
+
+final class PluginListViewModel {
+
+    /// Reference to the StoresManager to dispatch Yosemite Actions.
+    ///
+    private let storesManager: StoresManager
+
+    init(storesManager: StoresManager = ServiceLocator.stores) {
+        self.storesManager = storesManager
+    }
```
Ecarrion, line 12: +1 for setting things up for dependency injection!

Ecarrion (changes requested):
I think you forgot to merge develop back onto this branch because the `assertionFailure` on `Dispatcher` is still present. 😅

itsmeichigo:
Thanks @Ecarrion, I have merged changes from develop and migrated the view controller to use xib instead!

Ecarrion (approved):
Looks good to me! Thanks for changing from Storyboards to Xibs.

PR #4156, opened 2021-05-10
Plugins: Add table view and populate data to Plugin List screen

itsmeichigo:
Part of #4114 and #3921
Please make sure that #4155 is merged before testing this PR.

## Description
To show plugins on the Plugin List screen, a table view needs to be added to the view controller. Data needs to be populated to the table view using ResultsController.

### View Model
- For better testability, the view model's dependencies were updated with site ID and storage manager.
- Added ResultsController to fetch plugin data. Plugins are fetched by site ID, grouped by status, and sorted by status and name.
- Added `activate` method to trigger fetching data and receive `onDataChanged` block to notify the method's caller when data is changed.
- Added an extension to handle data source for plugin list. This uses the output of the results controller to determine section and row details for the list.
- A unit test was added to validate the output of the plugin list data source.

## View Controller
- Added a grouped table view to the view controller.
- Updated navigation title
- Populate the table view with data source from the  view model
- Activate the view model on `viewDidLoad` to trigger fetching data

## Testing
- The unit test added for the view model needs to pass.
- When #4155 is merged, selecting the Plugins entry point on the Settings screen should show a list of plugins grouped by sections like the screenshot below.

## Design Review needed
For the table view background, I'm using the default system background color for table views. This is a bit different from the original design. @adamzelinski can you please confirm if this color is alright; otherwise please let me know the correct color to use on this screen. Thanks!

Update release notes:


ghost:
Hey @itsmeichigo, this is looking correct to me we use iOS system background primary light for these I think the value is #F2F2F7 so if that's what you have it should be correct.

Ecarrion:
FYI: It's fine if you point your PR to a branch(and not develop) to keep the change-set easy to review!
Once the base PR is merged, GH will automatically change the base branch!

itsmeichigo:
Hey Ernesto - this PR has been updated to use xib, it should be ready for review again now, thanks!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListCell/PluginTableViewCell.swift
```diff
@@ -0,0 +1,23 @@
...
+
+         descriptionLabel.applySecondaryBodyStyle()
+         descriptionLabel.numberOfLines = 2
+     }
+
+     func update(viewModel: PluginListCellViewModel) {
+         nameLabel.text = viewModel.name
+         descriptionLabel.text = viewModel.description
+     }
+ }
```
Ecarrion: We already have a bunch of basic table views to reuse in the app. EG `HeadlineLabelTableViewCell` or `TitleBodyTableViewCell`.

I think it may be better to update any of those to be configurable to our needs rather than creating yet another table view cell, what do you think?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -17,6 +19,61 @@ final class PluginListViewController: UIViewController {
...
+        return UITableView.automaticDimension
+    }
+}
+
+// MARK: - UITableViewDelegate conformance
+//
+extension PluginListViewController: UITableViewDelegate {
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+    }
```
Ecarrion: I don't think these cells should be selectable right now, as it does not add anything to the user. What do you think?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
```diff
@@ -1,13 +1,108 @@
...
+    /// StorageManager to load plugins from storage
+    ///
+    private let storageManager: StorageManagerType
+
+    /// Results controller for the plugin list
+    ///
+    private lazy var resultsController: ResultsController<StorageSitePlugin> = {
+        let predicate = NSPredicate(format: "siteID = %ld", self.siteID)
+        let nameDescriptor = NSSortDescriptor(keyPath: \StorageSitePlugin.name, ascending: true)
+        let statusDescriptor = NSSortDescriptor(keyPath: \StorageSitePlugin.status, ascending: true)
```
Ecarrion, line 26: Is there a specific case on why we need to sort by status too?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
```diff
@@ -1,13 +1,108 @@
...
+         storesManager: StoresManager = ServiceLocator.stores,
+         storageManager: StorageManagerType = ServiceLocator.storageManager) {
+        self.siteID = siteID
         self.storesManager = storesManager
+        self.storageManager = storageManager
+    }
+
+    /// Start fetching plugin data from local storage.
+    ///
+    func activate(onDataChanged: @escaping () -> Void) {
```
Ecarrion: Activate reads a bit generic to me.
What do you think about: `fetchPlugins()`

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
```diff
@@ -1,13 +1,108 @@
...
+        }
+        let pluginStatus = SitePluginStatusEnum(rawValue: rawStatus)
+        let sectionTitle: String
+        switch pluginStatus {
+        case .active:
+            sectionTitle = NSLocalizedString("Active Plugins", comment: "Title for table view section of active plugins")
+        case .inactive:
+            sectionTitle = NSLocalizedString("Inactive Plugins", comment: "Title for table view section of inactive plugins")
+        case .networkActive:
+            sectionTitle = NSLocalizedString("Network Active Plugins", comment: "Title for table view section of network active plugins")
```
Ecarrion: We normally have these strings in a separate enum like:
```swift
private extension PluginListViewModel {
   enum Localization {
       static let sectionTitle = NSLocalizedString("Active Plugins", comment: "Title for table view section of active plugins")
   }
}
```

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
```diff
@@ -1,13 +1,108 @@
...
+    func numberOfRows(inSection sectionIndex: Int) -> Int {
+        resultsController.sections[safe: sectionIndex]?.objects.count ?? 0
+    }
+
+    /// View model for the table view cell at specified index path.
+    ///
+    func cellModelForRow(at indexPath: IndexPath) -> PluginListCellViewModel {
+        let plugin = resultsController.object(at: indexPath)
+        // The description can sometimes contain HTML tags
+        // so it's best to be extra safe by removing those tags
```
Ecarrion: 👍

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
```diff
@@ -0,0 +1,69 @@
...
+        XCTAssertEqual(viewModel.titleForSection(at: 0), "Active Plugins")
+        XCTAssertEqual(viewModel.titleForSection(at: 1), "Inactive Plugins")
+        XCTAssertEqual(viewModel.numberOfRows(inSection: 0), 1)
+        XCTAssertEqual(viewModel.numberOfRows(inSection: 1), 2)
+        XCTAssertEqual(viewModel.cellModelForRow(at: IndexPath(row: 0, section: 0)).name, activePlugin.name)
+        XCTAssertEqual(viewModel.cellModelForRow(at: IndexPath(row: 0, section: 0)).description, "Lorem ipsum random HTML content")
+        XCTAssertEqual(viewModel.cellModelForRow(at: IndexPath(row: 0, section: 1)).name, inactivePlugin2.name)
+        XCTAssertEqual(viewModel.cellModelForRow(at: IndexPath(row: 1, section: 1)).name, inactivePlugin1.name)
+    }
+}
```
Ecarrion: What do you think about splitting tests into multiple ones?

Eg:
```
test_number_of_sections_is_correct_for_active_and_inactive_plugins
test_number_of_sections_is_correct_for_active_and_no_inactive_plugins
test_number_of_sections_is_correct_for_no_active_and_inactive_plugins
test_cell_view_model_contains_correct_information
test_cell_view_model_extracts_html_tags
test_number_of_rows_is_equal_to_available_plugins
```

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -17,6 +19,61 @@ final class PluginListViewController: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        configureNavigation()
+        configureTableView()
+        viewModel.activate { [weak self] in
+            self?.tableView.reloadData()
+        }
```
Ecarrion: Probably also good to have it in a `configureViewModel` method to keep consistency! 😅

Ecarrion (changes requested):
Great work @itsmeichigo, left you some comments and suggestions.

Also, I spotted these UI Issues


Here you can see that upon rotation, the table view does not expand to the edges.

rotation

Here you can see how in accessibility sizes, the title of the plugin is capped to two lines. I think the title should remain uncaped. Also the trailing dots(`...` ) indicating that the description is trimmed is not visible.

accibility-fonts

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
itsmeichigo: I wanted to make it sound like this method is meant to be called once when setting up the view model (with the configuration of onDataChanged callback). `fetchPlugins()` sounds like it should be triggered more than once and can be mistaken with `resyncPlugins` method. What do you think?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
itsmeichigo: It makes sense to keep the cell not selectable for now - I was keeping it there as I thought the detail screen should be implemented soon before this Plugins feature is released to public. But yeah it's better to just keep things concise and focused 👍

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListCell/PluginTableViewCell.swift
itsmeichigo: I actually did look at `HeadlineLabelTableViewCell` but not sure if updating it can require regression tests to all other screens that use it so I chose the safer solution. But I'll take a look again and try to make changes that don't require change to other parts of the app.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
itsmeichigo: NSFetchedResultsController requires that if we group data by sections then the first sort descriptor should have the same key path as the `sectionNameKeyPath`, otherwise the sorting will get messed up if we add sort descriptor with different key path. I found a reference for it [here](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/CoreData/nsfetchedresultscontroller.html) (this should be more transparent in the NSFetchedResultsController document though).

Ecarrion: That's probably a good annotation to documented on comments.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
Ecarrion: Hmm, I see your point. However, I think activate is still too generic. What do you think about 
```swift
observePlugins(onDataChanged: @escaping () -> Void)
```

itsmeichigo: This sounds great! Will update with your suggestion, thanks!

itsmeichigo:
@Ecarrion About the UI issues:
- I think it makes better sense to set constraints to safe area instead of super view, because I don't want contents to be obscured by devices' curves. Plus, I just double-checked and see that the same behavior is applied to other screens: Settings, Products, Reviews etc.
- I can update the labels' line break mode to truncate tail, but I believe that it makes sense to keep the titles in maximum 2 lines as the UI design does show cases where the title gets 2 lines:

Ecarrion:
Hey @itsmeichigo 

> I think it makes better sense to set constraints to safe area instead of super view, because I don't want contents to be obscured by devices' curves.

Content won't be obscured because `UITableView` adjusts it's insets to consider safe areas, even if you pin the view to the super view.

> Plus, I just double-checked and see that the same behavior is applied to other screens: Settings, Products, Reviews, etc.

These are screens that we are due to update, all the new screens expand through the extent of the screen.

Look the settings screen for example.
Screen Shot 2021-05-11 at 10 49 36 PM



> I can update the labels' line break mode to truncate tail, but I believe that it makes sense to keep the titles in maximum 2 lines as the UI design does show cases where the title gets 2 lines:
> 

What I meant is that the title should not have the `two line` limit. It should have a `zero line` limit. If not the screen will be useless for users who use accessibility fonts.

EG. I have no clue what plugin is this 😅 

Screen Shot 2021-05-11 at 10 57 43 PM


The description label, should have the `two line` limit + the `...` indicator when needed.

itsmeichigo:
@Ecarrion Thanks for clarifying! I did not know about UITableView adjusting its contents based on safe areas so that's one `TIL` for me 😄

With your explanation I now understand why it's necessary to keep titles with zero line limit. Accessibility is something that really needs more attention, I never really focused on that in my earlier work so this is another thing to be mindful about 👍

itsmeichigo:
Hi @Ecarrion - I have pushed changes as per your suggestions. The UI should look correct by now:
| Normal mode | Zoomed in mode |
|---------------|------------------|
|Simulator Screen Shot - iPod touch (7th generation) - 2021-05-12 at 12 08 16|Simulator Screen Shot - iPod touch (7th generation) - 2021-05-12 at 11 16 19|

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -17,6 +19,59 @@ final class PluginListViewController: UIViewController {
...
+        configureNavigation()
+        configureTableView()
+        configureViewModel()
+    }
+}
+
+// MARK: - UI Configurations
+private extension PluginListViewController {
+    func configureNavigation() {
+        title = NSLocalizedString("Plugins", comment: "Title of the Plugin List screen")
```
Ecarrion: I think this also should go into a `Localization` enum 😅

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -17,6 +19,59 @@ final class PluginListViewController: UIViewController {
...
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        viewModel.numberOfRows(inSection: section)
+    }
+
+    func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(HeadlineLabelTableViewCell.self, for: indexPath)
+        let cellModel = viewModel.cellModelForRow(at: indexPath)
+        cell.update(style: .bodyWithLineLimit(count: 2),
+                    headline: cellModel.name,
+                    body: cellModel.description)
```
Ecarrion: nit: I think this can go in one line. We normally spawn over multiple lines when going beyond the `swiftlint` limit.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
```diff
@@ -1,13 +1,120 @@
...
+        case .active:
+            sectionTitle = Localization.activeSectionTitle
+        case .inactive:
+            sectionTitle = Localization.inactiveSectionTitle
+        case .networkActive:
+            sectionTitle = Localization.networkActiveSectionTitle
+        case .unknown:
+            sectionTitle = "" // This case should not happen
+        }
+        return sectionTitle.capitalized
```
Ecarrion: nit: This code is just for figuring out the section title but If I'm reading the code I would have to inspect any single case to know for sure that there is not extra logic to side-effect in any of the switch cases. To help the reader, I'd like to scope these calculations into an inline closure, like:

```swift
let sectionTitle: String = {
    switch pluginStatus {
    case .active:
        return Localization.activeSectionTitle
    case .inactive:
        return Localization.inactiveSectionTitle
    case .networkActive:
        return Localization.networkActiveSectionTitle
    case .unknown:
        return "" // This case should not happen
   }
}()
```

What do you think about this approach?

Ecarrion (changes requested):
This is looking very good! 
Tests are much more readable now and we have reduced complexity by extending `HeadlineLabelTableViewCell` 🎉 

One minor thing I noticed, is that the plugin names could also include HTML tags, so it's probably good to remove them too!

Screen Shot 2021-05-12 at 8 02 15 AM

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
itsmeichigo: I think this is pretty neat 👍 Will certainly adopt this approach in my future work!

itsmeichigo:
@Ecarrion Thanks a lot for the supportive suggestions! I have updated the code, hopefully it looks good now 🙂

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -28,7 +28,7 @@ final class PluginListViewController: UIViewController {
 // MARK: - UI Configurations
 private extension PluginListViewController {
     func configureNavigation() {
-        title = NSLocalizedString("Plugins", comment: "Title of the Plugin List screen")
+        title = viewModel.pluginListTitle
```
Ecarrion, line 31: 👍

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -383,8 +383,7 @@ private extension SettingsViewController {
     func sitePluginsWasPressed() {
         // TODO: do we need analytics to track tap here?
         guard let siteID = ServiceLocator.stores.sessionManager.defaultStoreID else {
-            DDLogError("⛔️ Cannot find ID for current site to load plugins for!")
-            return
+            return DDLogError("⛔️ Cannot find ID for current site to load plugins for!")
```
Ecarrion, line 386: 🎉

Ecarrion (approved):
This looks good to me! Great work @itsmeichigo

PR #4158, opened 2021-05-10
Plugins: Manually re-synchronize plugins

itsmeichigo:
Part of #4114 and #3921
This PR's source branch is based on #4156 so please make sure that PR is merged before testing this.

## Description
This PR handles the final feature for #3921 - the manual synchronization of plugins. This adds the pull-to-refresh feature and also handles loading and error states of the plugin list.

### View Model
- Added `PluginListState` enum, which includes the states of the plugin list: syncing, results, and error.
- Added `resyncPlugins` method. This method dispatches the `synchronizePlugins` action and updates plugin states based on the output of the action.
- Added unit tests to validate behaviors of the `resyncPlugins` method.

### View Controller
- Added Pull to Refresh support by adding refresh control to the table view and triggering the view model's `resyncPlugins` method.
- Added ghost animation to indicate the loading process.
- Added error state view for when resyncing fails. 

## Testing
- Unit tests for PluginListViewModel need to pass.
- When #4156 is merged, building this PR should result in complete features of the plugin list 🎉 

## Screenshots
| Loading State | Error State |
|---------------|------------|
|Simulator Screen Shot - iPhone 12 Pro - 2021-05-10 at 17 45 50|Simulator Screen Shot - iPhone 12 Pro - 2021-05-09 at 20 57 29|

## Design Review needed
@adamzelinski I need your reviews on the screenshots above:
- Is the loading state looks okay? This was not included in the original design so I'm not sure if it's up to the team's standards.
- I'm using the empty state view's style for the error state as it makes sense to keep these screens consistent. However it looks a bit different from your original design - let me know if it needs updating.

Update release notes:


ghost:
Hey @itsmeichigo, these screens are looking great so far.
Just one thing I noticed is the loading state. I see you have a spinner there. Is this just for pull to refresh? If so that's fine but if you're not pulling to refresh there should be no spinner.

Thanks!

itsmeichigo:
@adamzelinski yes the refresh control is just for the pull to refresh feature. The list does not show this or the ghost animation when first loaded as data is synchronized automatically on app launch. Thanks for confirming!

WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -384,6 +384,13 @@ extension UIImage {
         return UIImage(named: "woo-wp-no-site")!.imageFlippedForRightToLeftLayoutDirection()
     }
 
+    /// Plugins error
+    ///
+    static var pluginListError: UIImage {
+        return UIImage(named: "woo-plugins-error")!
+            .imageFlippedForRightToLeftLayoutDirection()
```
Ecarrion: Thanks for taking RTL layouts into consideration!
nit: I think this can go in one line.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -45,6 +66,33 @@ private extension PluginListViewController {
...
+                guard let self = self else { return }
+                switch state {
+                case .syncing:
+                    self.removeErrorStateView()
+                    self.tableView.startGhostAnimation(style: .wooDefaultGhostStyle)
+                case .results:
+                    self.tableView.stopGhostAnimation()
+                case .error:
+                    self.tableView.stopGhostAnimation()
+                    self.displayErrorStateView()
```
Ecarrion: Let's make sure to use a different tableView for the ghosting effect. This can be one of the causes of the invalid number of rows crash (e.g. wordpress-mobile/WordPress-iOS#12974). It's better if we just avoid it. It's also good practice IMO.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -75,3 +123,57 @@ extension PluginListViewController: UITableViewDataSource {
...
+
+        errorStateView.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(errorStateView)
+
+        NSLayoutConstraint.activate([
+            errorStateView.leadingAnchor.constraint(equalTo: view.safeLeadingAnchor),
+            errorStateView.trailingAnchor.constraint(equalTo: view.safeTrailingAnchor),
+            errorStateView.topAnchor.constraint(equalTo: view.safeTopAnchor),
+            errorStateView.bottomAnchor.constraint(equalTo: view.safeBottomAnchor)
+        ])
```
Ecarrion: Tip: I think you can use
```swift
view.pinSubviewToSafeArea(errorStateView)
```

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
```diff
@@ -51,6 +55,23 @@ final class PluginListViewModel {
...
+            guard let self = self else { return }
+            switch result {
+            case .success:
+                self.pluginListState = .results
+            case .failure:
+                self.pluginListState = .error
+            }
+            onComplete()
+        }
+        storesManager.dispatch(action)
```
Ecarrion: While I like the fact that you are using `Combine` 🎉 
I think there may be a bit of overengineering in here, especially since we are mixing two paradigms. Reactive chains and normal completion closures.

I think a simpler approach would be to have something like:

```swift
func resyncPlugins(onComplete: @escaping (Result<Void, Error>) -> Void) {
   // code
}
```

And then call it from the `VC` like:

```swift
@objc func resyncPlugins() {
    
    removeErrorStateView()
    gosthTableView.startGhostAnimation()
    
    viewModel.resyncPlugins { [weak self] result in
        
        self.gosthTableView.stopGhostAnimation()
        
        
        if result.isFailure {
             self.displayErrorStateView()
             self.refreshControl.endRefreshing()
        }
    }
}
```

This, while simpler also contains the state in a single place. What do you think?

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
```diff
@@ -174,11 +174,183 @@ class PluginListViewModelTests: XCTestCase {
...
+        let activePlugin = SitePlugin.fake().copy(
+            siteID: sampleSiteID,
+            status: .active,
+            name: "AAA"
+        )
+        insert(activePlugin)
+
+        let viewModel = PluginListViewModel(siteID: sampleSiteID, storageManager: storageManager)
+        let expect = expectation(description: "Section info is updated correctly.")
+        viewModel.activate {
```
Ecarrion: FYI: we have a handy `waitFor()` function to keep these tests and wait times simpler. 

https://github.com/woocommerce/woocommerce-ios/blob/03f86b6656fce7397880fbf99260ecad2d8e4ce8/Yosemite/YosemiteTests/Stores/AddOnGroupStoreTests.swift#L44-L49

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
```diff
@@ -174,11 +174,183 @@ class PluginListViewModelTests: XCTestCase {
...
+            invokedSynchronizePlugins = true
+            invokedSynchronizePluginsWithSiteID = siteID
+            DispatchQueue.main.asyncAfter(deadline: .now() + completionDelay) { [weak self] in
+                if self?.shouldSucceed == true {
+                    onCompletion(.success(()))
+                } else {
+                    onCompletion(.failure(MockPluginError.mockError))
+                }
+            }
+        }
```
Ecarrion: I think we have a `MockStoresManager` that can help you achieve the same as this class.

EG:
https://github.com/woocommerce/woocommerce-ios/blob/4fe952211777fad134084cbc19d4db229a37c8c9/WooCommerce/WooCommerceTests/ViewRelated/Orders/Issue%20Refund/RefundConfirmationViewModelTests.swift#L109-L122

Ecarrion (changes requested):
Great Work @itsmeichigo.
I have some suggestions, let me know what you think.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewModel.swift
itsmeichigo: I think it also works if I remove the `.syncing` state and end refreshing when I receive new state events for `.result` or `.error`. Yet your block keeps state handling within the only scope that changes it so it's definitely a better solution!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
itsmeichigo: This is clearly something I'm missing out on. Thanks for the reference!

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
itsmeichigo: I don't think `waitFor()` works for this test case though, since I'm triggering `observePlugins()` and expect the `onDataChanged` block to get called after storage is updated. Using `waitFor()` will block the thread so I cannot reach that point where I intend to update storage. But I can certainly use it to test action dispatching! 💯

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
itsmeichigo: Great idea! I have migrated to this and it looks so much neater 🎉

itsmeichigo:
Thanks for the suggestions @Ecarrion 🙇‍♀️ I have pushed some changes and I think this is ready for review again.

WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -384,6 +384,12 @@ extension UIImage {
         return UIImage(named: "woo-wp-no-site")!.imageFlippedForRightToLeftLayoutDirection()
     }
 
+    /// Plugins error
+    ///
+    static var pluginListError: UIImage {
+        return UIImage(named: "woo-plugins-error")!.imageFlippedForRightToLeftLayoutDirection()
+    }
```
Ecarrion, line 391: Could you add a test for this in `IconsTests.swift` please? :-)

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -36,17 +58,59 @@ private extension PluginListViewController {
...
 
+    func configureGhostTableView() {
+        ghostTableView.registerNib(for: HeadlineLabelTableViewCell.self)
+        ghostTableView.translatesAutoresizingMaskIntoConstraints = false
+        ghostTableView.backgroundColor = .listBackground
+        ghostTableView.isScrollEnabled = false
+        ghostTableView.isHidden = true
+
+        view.addSubview(ghostTableView)
+        view.pinSubviewToSafeArea(ghostTableView)
```
Ecarrion: I think this also needs to expand beyond safe-areas, otherwise, you could see the real table behind it!


https://user-images.githubusercontent.com/562080/118050809-9b076300-b345-11eb-9e44-586ba2bffcc3.mov

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
```diff
@@ -174,11 +174,153 @@ class PluginListViewModelTests: XCTestCase {
...
+
+        // When
+        updateStorage(with: activePlugin.copy(status: .inactive))
+
+        // Then
+        waitForExpectations(timeout: Constants.expectationTimeout) { _ in
+            XCTAssertEqual(viewModel.numberOfSections, 1)
+            XCTAssertEqual(viewModel.titleForSection(at: 0), "Inactive Plugins")
+        }
+    }
```
Ecarrion: > I don't think waitFor() works for this test case though, since I'm triggering observePlugins() and expect the onDataChanged block to get called after storage is updated. Using waitFor() will block the thread so I cannot reach that point where I intend to update storage. But I can certainly use it to test action dispatching! 💯

Looking at this a bit I think we can improve it the following way. 

I'm not a fan of `observePlugins()` to also trigger the `performFetch` as the name does not imply it.

And, there is nothing in the `ViewModel` that prevents us from calling methods like `numberOfSections` before observing the plugins.

For that reason I think we should move the perform fetch to the `ResultsController` creation like:

```swift
private lazy var resultsController: ResultsController<StorageSitePlugin> = {
    let resultsController = // Code to create the results controller

    do {
        try resultsController.performFetch()
    } catch {
        DDLogError("⛔️ Error fetching plugin list!")
    }

    return resultsController
}()

```

After that, we could write simpler tests where the `Given, When, Then` are more clear, like:
```swift
func test_section_info_is_correct_after_plugin_status_is_updated() {
    // Given
    let activePlugin = SitePlugin.fake().copy(
        siteID: sampleSiteID,
        status: .active,
        name: "AAA"
    )
    insert(activePlugin)

    let viewModel = PluginListViewModel(siteID: sampleSiteID, storageManager: storageManager)
    XCTAssertEqual(viewModel.numberOfSections, 1)
    XCTAssertEqual(viewModel.titleForSection(at: 0), "Active Plugins")


    // When
    let pluginsChanged: Bool = waitFor { promise in
        viewModel.observePlugins {
            promise(true)
        }

        self.updateStorage(with: activePlugin.copy(status: .inactive))
    }

    // Then
    XCTAssertTrue(pluginsChanged)
    XCTAssertEqual(viewModel.numberOfSections, 1)
    XCTAssertEqual(viewModel.titleForSection(at: 0), "Inactive Plugins")
}
```

What do you think?

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
```diff
@@ -174,11 +174,153 @@ class PluginListViewModelTests: XCTestCase {
...
+
+    func test_cellModel_is_correct_after_plugin_is_deleted() {
+        // Given
+        let plugin1 = SitePlugin.fake().copy(
+            siteID: sampleSiteID,
+            status: .active,
+            name: "AAA"
+        )
```
Ecarrion: nit: I think all of these plugin creations, can go in the same line. They are a bit more readable to me that way 😅

Ecarrion (changes requested):
Hey @itsmeichigo, a few more details and I think we should we good to go!

WooCommerce/Classes/Extensions/UIImage+Woo.swift
itsmeichigo: Sure! I did not know this file exist!

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/Plugins/SitePluginListViewModelTests.swift
itsmeichigo: I think this is a great idea! Separation of concerns is what I really need to work on, thanks for pointing that out 🙏

itsmeichigo:
@Ecarrion Please do another review for my new changes, thanks!

Ecarrion (approved):
This looks good to me @itsmeichigo 
Well done! 👏

Issue #4160, opened 2021-05-10
[Mobile Payments] Update WCPayAccount.init(from: Decoder)  to parse value for CPP eligibility flag


ctarda:
Update the code hat is parsing the WCPayAccount object to taken account the flag based by the server, that indicates if a store is eligible for card present payments. 

This depends on https://github.com/Automattic/woocommerce-payments-server/issues/725

If that PR to WCPay server didn't get merged on time, we could even hardcode a set of eligible siteIDs into this method.

PR #4161, opened 2021-05-10
[Mobile Payments] Show "Collect Payment" and "Manage Card Readers" only for eligible accounts

### LINKED ISSUE #3828: [Mobile Payments] Update OrderDetailsViewModel with logic to decide if an order is a candidate for card present payment collection
At this moment, OrderDetailsViewModel exposes a boolean property that indicates if the order is a candidate for card present payments. 

That logic is just a feature flag for now, but we should take into account whether the store is enrolled into WCPay and the order is elegible for collecting card present payments

ctarda:
Closes #3828 . 

Well, not really, but reduces the scope of presenting functionality related to CPP to parsing a boolean flag in the Networking layer (#4160)

## Changes
* Added a flag to WCPayAccount to signal if the account is enrolled into Card Present Payments. This flag is hardcoded to true at the moment because
a) We are waiting for this to happen: https://github.com/Automattic/woocommerce-payments-server/issues/725
b) we like to live dangerously.
* Added logic to OrderDetailsViewModel and SettingsViewController to fetch the WCPayAccount and then show or hide the "Collect Payment" button (in OrderDetailsViewModel) or the "Manage Card Settings" section (in SettingsViewController)

But wait! That requires a network request every time we present Settings or we navigate into the Order Details screen! 
True, it does, partially because WCPayAccount is not persisted to the coredata store yet, but also partially because we are going to have to refresh that account information anyway, every time we present the order details screen or the settings view, to make sure we don't show buttons e should not show. So 🤷🏼 

## How to test.
Testing requires:

Testing this PR does not require a hardware reader.

* Checkout the branch, build and run
* Navigate to Settings. If the store has WCPay installed, the "Store Settings" section should be visible. It takes a second.
* Navigate to a COD order. The "Collect Payment " button should be visible.
* In WCPayAccount, line 40, change the value of the flag `canCollectPayments` to false.

Screen Shot 2021-05-10 at 16 21 01

* Build and run again. Navigate to Settings. There should not be a "Store Settings" section 
* Navigate to a COD order. There should not be a "Collect Payment" button

Update release notes:

### DIFF
--- a/Networking/Networking/Model/WCPayAccount.swift
+++ b/Networking/Networking/Model/WCPayAccount.swift
@@ -15,6 +15,8 @@ public struct WCPayAccount: Decodable {
     /// A two character country code, e.g. `US`
     /// See https://stripe.com/docs/api/accounts/object#account_object-country
     public let country: String
+    /// A boolean flag indicating if this Account can collect card present payments
+    public let canCollectPayments: Bool
 
     public init(
         status: WCPayAccountStatusEnum,
@@ -34,6 +36,8 @@ public struct WCPayAccount: Decodable {
         self.defaultCurrency = defaultCurrency
         self.supportedCurrencies = supportedCurrencies
         self.country = country
+        /// Hardcoded until support for this property in WCPay is available
+        self.canCollectPayments = true
     }
 
     /// The public initializer for WCPay Account.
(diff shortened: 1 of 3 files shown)


allendav (approved):
It works, but there is a 2-5 second delay between when I enter Settings and when the menu item appears for a store that supports card present payments. Could we open an issue to cache True values to accelerate things for sites that have successfully passed the test?

ctarda:
> Could we open an issue to cache True values to accelerate things for sites that have successfully passed the test?

Sure! That being said... this is unavoidable, as far as I understand? We have to check that the store is eligible every time we present the Settings, to avoid showing the feature to folks that should don't see it. How do you think we should do this?

PR #4165, opened 2021-05-11
[Mobile Payments] Add card brand icons to receipts

### LINKED ISSUE #4033: [Mobile Payments] [Receipts] Iterate on layout of printed receipt

koke:
Part of #4033

This PR adds icons for each of the card brands to be used in receipts.

I've copied these icons from the woocommerce-payments plugin. I ran into some rendering issues with the current version of them, since the optimized SVGs introduced in https://github.com/Automattic/woocommerce-payments/pull/1078 didn't render correctly on iOS, so I used the original SVGs from https://github.com/Automattic/woocommerce-payments/pull/305.

Because this will be used only for HTML for now, I think using an asset catalog would have complicated things, so I'm bundling the svg files directly, then reading the svg data and embedding that in the CSS directly. This simplifies things as we don't need to deal with URLs for printing, and should work better for emails as well.

## To test

The easiest way to test is to use the new SampleReceiptPrinter app target, and try to print a sample receipt. You can tweak `SampleContent.swift` to choose a different card brand.

Screen Shot 2021-05-11 at 11 08 27

Update release notes:


Hardware/Hardware/CardReader/CardBrand.swift
```diff
@@ -82,3 +82,40 @@ private extension CardBrand {
...
+    private class _Bundle {}
+
+    private var iconURL: URL! {
+        Bundle(for: CardBrand._Bundle.self)
+            .url(forResource: iconName, withExtension: "svg")
+    }
+
+    /// Icon that represents the brand in SVG format
+    /// Since we mostly use these in HTML for receipts, we are reading the files directly instead of using UIImage
+    var iconData: Data! {
```
koke, line 98: Generally, I'm not a fan of force-unwrapping things, but I added a test to ensure every possible case has non-empty data. This should never be nil, event we can't enforce it at compile-time.

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -86,6 +86,15 @@ public extension ReceiptRenderer {
                     table { background-color:#F5F5F5; width:100%; color: #707070 }
                     table td:last-child { width: 30%; text-align: right; }
                     table tr:last-child { color: #000000; }
+                    .card-icon {
+                       width: 24px;
+                       height: 17px;
+                       vertical-align: bottom;
```
koke, line 92: The alignment isn't working perfectly, but since I'm tweaking the font sizes further in an upcoming PR, we can leave it like this for now.

Hardware/Hardware/CardReader/CardBrand.swift
ctarda: That works. FWIW, it's a similar approach to what we do in UIImage+Woo

ctarda (approved):
This is definitely an improvement! 👏🏼 

This is how it looks on an email:



:shipit:

koke:
Thanks, the email looks nice 🎉

PR #4166, opened 2021-05-11
[Mobile Payments] Improve layout of printed receipts for roll paper

koke:
Part of #4033

This PR improves layout and typography to approximate the design in #4033 better.

The biggest change is that I've stopped using headers and footers hoping that it works fine in roll paper, although that might not be great with regular paper. The assumption here is that roll paper will print each page right after the previous one without distinction, which might work until we can actually calculate the receipt height, but this should be tested when we get some hardware.

Other improvements include tweaks in the typography, spacing, and title alignment.

Some sample receipts printed with 1 and 30 items in 4" Roll and US Letter (for some reason the US Letter pages are printed in reverse order, I'm not sure if this is the Printer Simulator, and I don't have any AirPrint capable printers).

receipts

## To test

The easiest way to test is to use the new SampleReceiptPrinter app target, and try to print a sample receipt.

Screen Shot 2021-05-11 at 11 54 54

Update release notes:

### DIFF
--- a/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
+++ b/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
@@ -6,21 +6,6 @@ public final class ReceiptRenderer: UIPrintPageRenderer {
     private let lines: [ReceiptLineItem]
     private let parameters: CardPresentReceiptParameters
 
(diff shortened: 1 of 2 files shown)


koke:
I just noticed the amount doesn't look right, looking into it 🤔

ctarda (approved):
That's much nicer!

This is how it looks on an email:


:shipit:

PR #4168, opened 2021-05-11
[Mobile Payments] Pass statement descriptor when initiating a payment capture

### LINKED ISSUE #3938: [Mobile Payments] [Receipts] Pass a proper Receipt description and StatementDescription when initiating charge
At the point, we are still hardcoding the email and receipt descriptions passed to PaymentIntentParams

The statement descriptor is part of the response to WPPayRemote.loadAccount. p91TBi-4KR-p2#comment-4303

Depends on #4011

ctarda:
Closes #3938 

## Changes
* Remove the previously hardcoded receiptDescription and statementDesxription
* Pass a generic string containing the store name as [receiptDescription](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)stripeDescription)
* Pass WCPayAccount.statementDescriptor as [statementDescription](https://stripe.dev/stripe-terminal-ios/docs/Classes/SCPPaymentIntentParameters.html#/c:objc(cs)SCPPaymentIntentParameters(py)statementDescriptor) This is the one we really care about as it is the one that might be in the customer's credit card statement.

## How to test
Testing requires:

* Checkout the branch, build and run
* Add a breakpoint to StripeCardReaderService, line 163
Screen Shot 2021-05-11 at 06 11 19
* Navigate to an eligible order, tap "Collect Payment"
* The breakpoint should be hit. `po parameters` should print something similar to this:

```
PaymentIntentParameters(amount: 200, _currency: Hardware.CurrencyCode(value: "USD"), receiptDescription: Optional("Receipt from wooctarda”), _statementDescription: Hardware.StatementDescriptor(value: Optional("CTARDAWOO.WPCOMSTAGING")), _receiptEmail: Hardware.EmailSwift.String(value: Optional("ctarda@automattic.com")), metadata: Optional([AnyHashable("paymentintent.storename"): Optional("wooctarda")]))
```

Notice receiptDescription and statementDescription. Actual values would depend on the store configuration, but data should be there. 

Update release notes:


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -134,6 +134,8 @@ final class OrderDetailsViewModel {
         return String.localizedStringWithFormat(Localization.emailSubjectWithStoreName, storeName)
     }
 
+    private var paymentsAccount: WCPayAccount? = nil
```
ctarda, line 137: I went "stateless" first, in https://github.com/woocommerce/woocommerce-ios/pull/4168/commits/f27ca1e864a4191a016f43c49a6a27937fbbd72b, and fetched the account right before starting the process to collect the payment. But:
1. that made the process longer, and also required a new modal view before the "tap reader" message, which made the whole process feel even longer
2. we are already fetching the account when showing the order details view, in order to show/hide the "Collect Payment" button.
3. if, for some reason, this reference was nil when we start collecting the payment, we would just passing nils, which would not be terrible.

So I went stateful 🤷🏼

allendav (approved):
Tests fine. The statement descriptor for my store was reflected in the stripe dashboard:

Screen Shot 2021-05-11 at 11 54 07 AM

I am a little worried about passing the entire Account object around, as that might lead to broad coupling down the road, but let's proceed.

ctarda:
> I am a little worried about passing the entire Account object around, as that might lead to broad coupling down the road, but let's proceed.

I don't disagree 😉

PR #4169, opened 2021-05-11
[Mobile Payments] Include quantity in receipt lines

koke:
Part of #4033

This PR ensures each line in the receipt includes the quantity. The amount shown in the receipt is per item, and so the numbers won't add up to the total amount if the quantity isn't displayed.

Screen Shot 2021-05-11 at 13 10 25

I've tweaked the sample content generation a bit to simplify, but otherwise this is just passing extra information already present for the order item.

## To test

The easiest way to test is to use the new SampleReceiptPrinter app target, and try to print a sample receipt.

Update release notes:

### DIFF
--- a/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
+++ b/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
@@ -111,7 +111,7 @@ private extension ReceiptRenderer {
     private func summaryTable() -> String {
         var summaryContent = "<table>"
         for line in lines {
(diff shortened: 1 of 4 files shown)


ctarda (approved):
Nice! Loving the approach of submitting small, focused and frequent PRs 😄 

This is how it looks on email:



The first receipt is for an order that has a Shipping Fee of 200 dollars. I am not sure if that's something we should add to the receipts as a line item (shipping fees and one time fees). But if we decide to add those, that can be a separate PR. 

:shipit:

koke:
Good catch. Probably there's more things that should be there like discounts 🤔

PR #4170, opened 2021-05-11
[Mobile Payments] Settings 2-8b: Swap out scanning alert with new modal

### LINKED ISSUE #4163: [Mobile Payments] Card Reader Settings 2-8b: Switch to New Modals
Replace `UIAlerts` with new modals

See https://github.com/woocommerce/woocommerce-ios/pull/4162#discussion_r629720990

allendav:
Closes #4163 

Changes:
- Replaces UIAlert displayed when scanning for readers with new modal

Unfinished:
- Matching styles in design exactly

To test:
- Open Settings > Manage Card Readers V2 while no reader is connected
- Tap on Connect button
- Ensure new modal appears
- Tap on Cancel
- Ensure alert disappears
- You should also go through the payments flow and ensure all those modals appear correct since some tweaks were made to bottom label handling in the modals

Update release notes:

Screen Shot 2021-05-11 at 10 55 46 AM


WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -336,6 +336,12 @@ extension UIImage {
         return UIImage(named: "woo-payments-card")!
     }
 
+    /// Card Reader with radio waves
+    ///
+    static var cardReaderScanning: UIImage {
```
ctarda, line 341: Could you add a test for this in IconsTests, please?

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.swift
```diff
@@ -136,7 +136,7 @@ private extension CardPresentPaymentsModalViewController {
             configureActionButtonsView()
         }
 
-        if !shouldShowActionButtons() {
+        if shouldShowBottomLabels() {
```
ctarda, line 139: Good catch

ctarda (approved):
Code looks good, but please add a test for the new image before merging, to make sure that we do not add a potential crash.

Speaking the image, it does not work great on dark mode. I think this can be deferred to a separate PR.


The functionality is there, it works as expected, and none of the pre-existing modals seems to break.

:shipit:

WooCommerce/Classes/Extensions/UIImage+Woo.swift
allendav: Added test and fixed dark mode in https://github.com/woocommerce/woocommerce-ios/pull/4170/commits/c82390ebee53d6bdbce28a84b3b31d8bad376c6e

PR #4172, opened 2021-05-11
[Mobile Payments] Card Reader Settings 2-9: Hook up card reader connect

### LINKED ISSUE #4055: [Mobile Payments] Card Reader Settings v2-9: Discovered readers
Hook up discovered readers completion in CardReaderUnknownViewController
Add I found reader X alert showing / not to CardReaderUnknownViewModel
Add connect action to CardReaderUnknownViewModel
Add “keep searching”

allendav:
Closes #4055 

Note: This PR is against feature/stripe-terminal-sdk-integration not develop

Note: This is the last PR in the Card Reader Settings v2 effort before we remove the v1 code ( 2-12 #4058 )

Changes:
- Adds new images for connecting and finding readers, including dark mode images
- Adds new modals for a found reader and connecting to a reader
- Allows the user to connect to a found reader, or continue searching

To test:
- Settings > Manage Card Readers v2 > Connect Card Reader
- Power on your reader
- The modal should change from a "Scanning for reader" modal to a "Do you want to connect to..." modal
- Tap Connect to Reader
- The view should change to show your connected reader

Loose ends:
- I am having a little trouble with the keep searching flow and multiple readers. I'll resolve this during the review

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsAlerts.swift
```diff
@@ -16,6 +16,24 @@ final class CardReaderSettingsAlerts {
         from.present(newAlert, animated: true)
     }
 
+    func connectingToReader(from: UIViewController) {
```
ctarda, line 14: I am just curious, would it be possible to update the model without dismissing it?  It seems like there are maybe too many presentations and dismissals. 

In theory, we could do something like:
```
        let viewModel = connectingToReader()
       if modalController != nil {
          modalComtroller.setViewModel(viewModel)
      }
```

That would update the modal being currently presented without dismissing it first.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewModel.swift
```diff
@@ -51,29 +55,93 @@ final class CardReaderSettingsUnknownViewModel: CardReaderSettingsPresentedViewM
...
+            self?.didDiscoverReaders(cardReaders: cardReaders)
         }
         ServiceLocator.stores.dispatch(action)
     }
 
+    /// Alert the user we have a found reader
+    ///
+    func didDiscoverReaders(cardReaders: [CardReader]) {
+        /// If we are already presenting a foundReader alert to the user, ignore the found reader
+        guard discoveryState != .foundReader else {
```
ctarda, line 84: We could do these two in just one guard:

```
guard discoveryState != .fundReader, cardReaders.count > 0 else ...
```

That might or might not be more readable though. I don't have any strong feelings one way or another 🤷🏼

ctarda (approved):
It works great! 


https://user-images.githubusercontent.com/2722505/117897861-aac56f80-b291-11eb-99aa-660ca1bd3b91.MP4

One idea for you to consider, regarding the way the modal views are updated, and how, in theory, it should be possible to update those without actually dismissing them. 

Just for fun, I also tested to disconnect from the reader and connecting again, and shutting down the reader while it is connected, and those two cases still work fine.

:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsAlerts.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4172/commits/bd40e9a748889e2eae038e3e342b2d66290d4967 - looks better without those extra animations

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewModel.swift
allendav: I'd like to keep this one as is, since the cases I am guarding against are somewhat different

PR #4177, opened 2021-05-12
[Mobile Payments] Saved receipts mark II: Yosemite

### LINKED ISSUE #3974: [Mobile Payments] [Saved Receipts]  Save and present saved receipts. Mark II: Yosemite
For context: p91TBi-4KR-p2

Add business logic to fetch a saved receipt associated to an order.

UPDATE: This might not be necessary if we end up fetching the application_preferred_name and dedicated_file_name from an endpoint

ctarda:
Closes #3974 

This is a throwaway solution, that could get us to beta 1 if we do not have an endpoint to fetch the receipt data on time. But at least, it gives us the chance to move forward with the UI side of the solution.

## Changes
* Added actions to save and read receipt metadata.
* Made all the model objects we use to model the receipt metadata conform to Codable
* Added an implementation that saves and reads the receipt metadata to a plist file. Ideal? No. Again, this is something we might (and hopefully will) throw away in the coming days.
* I didn't add any tests. Let's wait a couple of days for that.

## How to test
* Pair with a reader
* Navigate to an order
* Capture a payment. Notice there will be a print out in the console after the order has ben captures with a message like this when running on the simulator:
```
new receipt saved: open /Users/ctarda/Library/Developer/CoreSimulator/Devices/90C3D678-6F22-4DD2-873D-9BB5C7B79217/data/Containers/Data/Application/0388C091-8303-4648-A69C-34206563BE63/Documents/order-id-894-receipt.plist
```

And like this when running on device:
```
new receipt saved: open /var/mobile/Containers/Data/Application/240FD6C4-5871-44C7-8CEB-BB4A6830636E/Documents/order-id-893-receipt.plist
```

* You can copy that and paste it in Terminal to open the new plist file.
* Now navigate out of the order, go to all orders > navigate back into the previous order. There should be a trace like this:
```
====== receipt parameters for order Optional(893)
CardPresentReceiptParameters(amount: 20000, currency: "usd", date: 2021-05-12 14:46:49 +0000, storeName: Optional("wooctarda"), cardDetails: Hardware.CardPresentTransactionDetails(last4: "9999", expMonth: 12, expYear: 2021, cardholderName: Optional("TEST/STRIPE"), brand: Visa, fingerprint: "REDACTED", generatedCard: Optional("pm_REEDACTED"), receipt: Optional(Hardware.ReceiptDetails(applicationPreferredName: "Stripe Credit", dedicatedFileName: "REDACTED", authorizationResponseCode: "3030", applicationCryptogram: "REDACTED", terminalVerificationResults: "REDACTED", transactionStatusInformation: "0000", accountType: Optional("credit"))), emvAuthData: Optional("REDACTED")))
////// recipt parameters loaded
```

That's it. That's all for now. Next steps are #3975 and #3981

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
@@ -39,6 +39,16 @@ final class PaymentCaptureOrchestrator {
 
         ServiceLocator.stores.dispatch(action)
     }
+
+    func saveReceipt(for order: Order, params: CardPresentReceiptParameters) {
+//        let action = ReceiptAction.print(order: order, parameters: params)
+//
+//        ServiceLocator.stores.dispatch(action)
+
+        let action = ReceiptAction.saveReceipt(order: order, parameters: params)
+
+        ServiceLocator.stores.dispatch(action)
+    }
 }
 
 
@@ -86,19 +96,19 @@ private extension PaymentCaptureOrchestrator {
             onCompletion(.failure(error))
         case .success(let paymentIntent):
             submitPaymentIntent(siteID: order.siteID,
-                                orderID: order.orderID,
+                                order: order,
                                 paymentIntent: paymentIntent,
                                 onCompletion: onCompletion)
         }
     }
 
     func submitPaymentIntent(siteID: Int64,
-                             orderID: Int64,
+                             order: Order,
                              paymentIntent: PaymentIntent,
                              onCompletion: @escaping (Result<CardPresentReceiptParameters, Error>) -> Void) {
         let action = WCPayAction.captureOrderPayment(siteID: siteID,
-                                                     orderID: orderID,
-                                                     paymentIntentID: paymentIntent.id) { result in
+                                                     orderID: order.orderID,
+                                                     paymentIntentID: paymentIntent.id) { [weak self] result in
 
             guard let receiptParameters = paymentIntent.receiptParameters() else {
                 let error = CardReaderServiceError.paymentCapture()
@@ -111,6 +121,7 @@ private extension PaymentCaptureOrchestrator {
 
             switch result {
             case .success:
+                self?.saveReceipt(for: order, params: receiptParameters)
                 onCompletion(.success(receiptParameters))
             case .failure(let error):
                 onCompletion(.failure(error))
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
@@ -41,10 +41,6 @@ final class PaymentCaptureOrchestrator {
     }
 
     func saveReceipt(for order: Order, params: CardPresentReceiptParameters) {
-//        let action = ReceiptAction.print(order: order, parameters: params)
-//
-//        ServiceLocator.stores.dispatch(action)
-
         let action = ReceiptAction.saveReceipt(order: order, parameters: params)
 
         ServiceLocator.stores.dispatch(action)
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
@@ -417,15 +417,15 @@ extension OrderDetailsViewModel {
     }
 
     func syncSavedReceipts(onCompletion: ((Error?) -> ())? = nil) {
-        let action = ReceiptAction.loadReceipt(order: order) { result in
+        let action = ReceiptAction.loadReceipt(order: order) { [weak self] result in
             switch result {
             case .success(let parameters):
-                print("====== recipt parameters loaded")
+                print("====== receipt parameters for order \(String(describing: self?.order.orderID))")
                 print(parameters)
                 print("////// recipt parameters loaded")
                 onCompletion?(nil)
             case .failure:
-                print("==== no receipt. moving on")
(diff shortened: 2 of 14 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -416,6 +416,32 @@ extension OrderDetailsViewModel {
         stores.dispatch(action)
     }
 
+    func syncSavedReceipts(onCompletion: ((Error?) -> ())? = nil) {
+        let action = ReceiptAction.loadReceipt(order: order) { [weak self] result in
+            switch result {
+            case .success(let parameters):
+                print("====== receipt parameters for order \(String(describing: self?.order.orderID))")
```
ctarda, line 423: These prints are here for testing the PR and will be removed in an upcoming PR.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -39,6 +39,16 @@ final class PaymentCaptureOrchestrator {
 
         ServiceLocator.stores.dispatch(action)
     }
+
+    func saveReceipt(for order: Order, params: CardPresentReceiptParameters) {
+//        let action = ReceiptAction.print(order: order, parameters: params)
```
allendav: Did you mean to leave these commented-out lines here?

Yosemite/Yosemite/Stores/ReceiptStore.swift
```diff
@@ -56,4 +62,66 @@ private extension ReceiptStore {
...
+        return try FileManager.default.url(for: .documentDirectory,
+                                                           in: .userDomainMask,
+                                                           appropriateFor: nil,
+                                                           create: false)
+                .appendingPathComponent(fileName(order: order))
+                .appendingPathExtension("plist")
+    }
+
+    func fileName(order: Order) -> String {
+        "order-id-\(order.orderID)-receipt"
```
allendav: This will be fine I think for now, but we should probably support multiple receipts in the future. For example, if a in-person card payment buyer gets a refund, they usually ask for a receipt for that too.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
ctarda: Nope. I did not 😊

Yosemite/Yosemite/Stores/ReceiptStore.swift
ctarda: True, good point

koke: Oh, another thing to keep in mind is multiple stores, and more generally when we should remove these files from the device.

ctarda: Yep. I can add the storeid to the file name as well. Removing the files... as soon as we have an endpoint that we can use to fetch this data

Yosemite/Yosemite/Stores/ReceiptStore.swift
```diff
@@ -56,4 +62,66 @@ private extension ReceiptStore {
         let renderer = ReceiptRenderer(content: content)
         onContent(renderer.htmlContent())
     }
+
+    func loadReceipt(order: Order, onCompletion: @escaping (Result<CardPresentReceiptParameters, Error>) -> Void) {
+
+        guard let outputURL = try? fileURL(order: order) else {
+            let error = ReceiptStoreError.fileNotFound
```
koke, line 70: When I was testing this earlier I went into the wrong order and opened one that didn't have a receipt and the error wasn't coming from here, but from the decoder not being able to load data at the given URL. I think the `fileURL` method will succeed as long as the URL is valid regardless of the file existing or not. Maybe we want to add an extra check for that, if we care about differentiating missing receipts from wrong ones.

ctarda: Good catch, thanks!

ctarda:
@allendav @koke  I just pushed an update to:
* add the store id to the final name
* check that the file actually exists before attempting to load and parse it

Yosemite/Yosemite/Stores/ReceiptStore.swift
```diff
@@ -56,4 +62,67 @@ private extension ReceiptStore {
...
+
+        guard let outputURL = try? fileURL(order: order),
+              FileManager.default.fileExists(atPath: outputURL.path) else {
+            let error = ReceiptStoreError.fileNotFound
+            onCompletion(.failure(error))
+            return
+        }
+
+        guard let receiptContent: ReceiptContent = try? fileStorage.data(for: outputURL) else {
+            DDLogError("⛔️ Unable to load receipt metadata for order: \(order.orderID)")
```
allendav: Could we make this an info or at most a warn? Most orders won't have in-person payment receipts

allendav (approved):
:shipit: A few suggestions to consider. Tests fine. I see everything I'd expect in the file:

plist

And in the console on returning to that order:

console

Yosemite/Yosemite/Stores/ReceiptStore.swift
ctarda: Sure, I'll make it a warning. This should only be reached when the file exists and we fail to parse it though.

allendav: I got it on an older order:

```
2021-05-12 10:04:13.483040-0700 WooCommerce[3172:2030342] ⛔️ Unable to load receipt metadata for order: 8295
==== no receipt parameters for order Optional(8295). moving on
```

PR #4179, opened 2021-05-12
[Mobile Payments] Simplify card reader connection result handling

koke:
Part of #3926 

The existing data flow for connecting a reader had a bit of split responsibilities. `CardPaymentStore` was subscribing both to the Future result of `connect()` and the `connectedReaders` property.

Instead, we can pass the reader object as part of the successful completion of the `connect()` future and use that instead. I took the opportunity to replace the existing array arguments with a single instance of `CardReader`, since you can't connect to more than one reader at once. I think we could do the same to `connectedReaders`, but that's used in the newer settings that I haven't been able to test yet, and I think @allendav mentioned changing that anyway.

## To test

See that you can discover, connect and disconnect readers as usual.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardReaderService.swift
+++ b/Hardware/Hardware/CardReader/CardReaderService.swift
@@ -36,7 +36,7 @@ public protocol CardReaderService {
 
     /// Connects to a card reader
     /// - Parameter reader: The card reader we want to connect to.
-    func connect(_ reader: CardReader) -> Future <Void, Error>
+    func connect(_ reader: CardReader) -> Future <CardReader, Error>
 
     /// Disconnects from the currently connected reader
     func disconnect() -> Future <Void, Error>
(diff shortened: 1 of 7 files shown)


ctarda (approved):
It seems to work fine with the V1 UI. I am not sure if there might be some expectations @allendav might have for the work he is doing to finalize that UI.

So, it is a :shipit: from me, but maybe it would be a good idea to test again, after #4172 is merged?

allendav:
I will retest with 4172 asap

allendav:
Re-tested after merging these changes into 4172 's branch. No changes were needed in that branch to accommodate this and reader connection continues to work, mostly because I don't (yet) care about the connect completion but I am instead counting on observed connected readers to change.

We should allow for in-progress-connection cancellation (we don't yet) if possible, and we should handle connection errors. I'll open tickets for that.

PR #4180, opened 2021-05-12
[Mobile Payments] Return a future when canceling card reader discovery

koke:
Part of #3926 

This time I'm getting rid of the cancellable from `cancelCardReaderDiscovery`. Instead of observing the `discoveryStatus` value, we return a `Future` that will complete when the discovery process is actually canceled, or pass through the error if there is one.

I've update the integration tests to avoid some warnings, but those aren't running by default and were failing anyway. I'm wondering if we want to fix those, remove them, or leave them for now.

## To test

The current settings don't implement the cancel button yet, so try a quick implementation for `CardReaderSettingsViewModel.stopSearch()`:

```swift
    func stopSearch() {
        ServiceLocator.stores.dispatch(CardPresentPaymentAction.cancelCardReaderDiscovery(onCompletion: { [weak self] _ in
            self?.activeAlert = .none
        }))
    }
```

Then try to start discovering readers and immediately hit cancel.

I imagine this can be tested more extensively as the feature is developed.

Update release notes:


allendav:
> The current settings don't implement the cancel button yet

Take a peek at CardReaderSettingsUnknownViewModel

```
    /// Dispatch a request to cancel reader discovery
    ///
    func cancelReaderDiscovery() {
        discoveryState = .notSearching
        updateProperties()
        didUpdate?()

        let action = CardPresentPaymentAction.cancelCardReaderDiscovery() { _ in
        }
    }
```

that is hooked up to the modals as of about yesterday

allendav:
I'm not yet sure why, but I cannot connect to a card reader with this branch. I get the following in Xcode console:

```
2021-05-13 13:56:21.656045-0700 WooCommerce[4545:2454534] [CoreBluetooth] XPC connection invalid
```

and the reader will not connect. It shows a unexpected color sequence on its power LED too: purple -> yellow -> green -> off (and then repeats)


https://user-images.githubusercontent.com/1595739/118188088-b9786780-b3f4-11eb-8a7b-99151396e318.MOV

koke:
I'm seeing the same issue in the `feature/stripe-terminal-sdk-integration`, so it doesn't seem specific to this PR. Can either of you verify this on an actual reader? I'm still trying to figure out what's going on. From what I can see we're doing everything correctly, but the Stripe SDK will be stuck discovering when the process starts again after a first cancellation.

koke:
I think I found the issue, at least to what I can reproduce with the simulated reader. The SDK, service, and store were actually working fine, but the UI wouldn't present a new alert after an existing one had been dismissed.

ctarda (approved):
I just tested the most recent commit:

1. Started discovery.
2. Cancelled.
3. Started discovery again. (no more flashing less in the reader)
4. Connected to a reader.
5. Disconnected (tapping the disconnect button)
6. Started discovery again. 
7. Connected again.
8. Turned the reader off. The UI got updated.

:shipit:

allendav:
@koke wrote:

> I think I found the issue, at least to what I can reproduce with the simulated reader. The SDK, service, and store were actually working fine, but the UI wouldn't present a new alert after an existing one had been dismissed.

Is this still a problem? Does this happen in the flow initiated within settings or the flow initiated within orders? Or both?

koke:
> Is this still a problem? Does this happen in the flow initiated within settings or the flow initiated within orders? Or both?

1b2104727ef662b40634b985f3e611b1d83c0f4c should have taken care of it, although now that you mention it, I didn't test the flow initiated with orders

koke:
I just tested the flow from orders and it's also working correctly

PR #4185, opened 2021-05-13
[Mobile Payments] Use Sink subscriber for resetting card reader service

koke:
Part of #3926 

This removes the need to store a cancellable for the `reset` action by using `Subscribers.Sink` instead of the `sink` operator.

## To test

Follow the testing steps in #4147

Update release notes:


ctarda (approved):
Tested well, I switched stores and logged in and out and things work the way they should.

:shipit:

PR #4186, opened 2021-05-13
[Mobile Payments] Improve publisher subscription management for payment collection

koke:
Part of #3926 

This one improves the cancellable management for the `collectPayment` action.

In this case, managing the subscription gets slightly more complicated, since we have:
1. Two different callback methods.
2. A `readerEvents` publisher that we want to subscribe as long as the payment is in progress.
3. A `capturePayment` method that performs several steps sequentially.

My approach here has been to first declare that `CardReaderService.capturePayment` behaves like a future. I don't think there is a good way to enforce this, or at least not one that wouldn't overcomplicate things, so for now we can state that in the documentation and make sure it behaves like that. This wasn't strictly necessary for the cancellable, but we only want to call the completion callback once, so this makes that more explicit.

For the `readerEvents` subscription, I've stored the cancellable in a local variable, which gets captured by the completion block, which will call `cancel()` on completion.

For the `capturePayment` method, I've switched it to `Subscribers.Sink` like others.

## To test

I'm not sure how to test all the possible payment flows without a physical reader, but I've tried completing the payment flow for a couple of orders with the simulated reader, and everything seemed to work fine.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardReaderService.swift
+++ b/Hardware/Hardware/CardReader/CardReaderService.swift
@@ -46,6 +46,7 @@ public protocol CardReaderService {
     func clear()
 
     /// Captures a payment after collecting a payment method succeeds.
+    /// The returned publisher will behave as a Future, eventually producing a single value and finishing, or failing.
     func capturePayment(_ parameters: PaymentIntentParameters) -> AnyPublisher<PaymentIntent, Error>
 
     /// Cancels a a PaymentIntent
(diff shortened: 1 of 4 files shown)


ctarda (approved):
Nice!

I just tested with an external reader and the payment capture completes perfectly, and the order gets updated, changes status and updates the notes. 

So everything seems to be working perfect.

:shipit:

PR #4188, opened 2021-05-13
[Mobile Payments] Improve combine subscriptions for card reader software updates

koke:
Part of #3926 

We're almost done 😅. This was the trickiest of all the subscriptions, but also the most interesting.

In this case, we have data coming at `StripeCardReaderService` from two different sources:
- The completion block for `Terminal.installUpdate(_:delegate:completion)` will tell us when the update is finished, or if there is an error
- The `terminal(_:didUpdateDiscoveredReaders:)` delegate method will be called periodically with progress updates.

Previously, the service would offer this two sources separately, but the problem description fits a `Publisher` perfectly: we want to send progress events, and complete when the update is done.

It took some effort to combine these two sources, but I think the end result is not too complicated. The main difficulty is that Combine doesn't offer many operators to take completion events into account, so rather than writing a custom publisher, I managed to find a solution with a boolean publisher. I tried to document this in the code but, as a summary:

- We keep the previous `Future` internally as a way to wrap the `installUpdate` completion in a way that it can be used with Combine.
- However, it now has an `Output` type of `Bool`, so we can prepend an initial `false` value. If we didn't do this, the `Future` would only emit its first event when the update was finished, and the rest wouldn't be useful to forward progress events.
- We map this `Bool` state into another publisher:
  - If the value is `false`, we return the existing `softwareUpdateSubject` so we can forward progress updates
  - If the value is `true`, we are done with the update and return an `Empty` publisher
- We use `map` and `switchToLatest` because `flatMap` would not complete until all the mapped publishers had completed, and `softwareUpdateSubject` never completes.

## To test

I've only been able to use the simulated reader, but my steps were:

1. Go to Settings > Manage Card Readers
2. Tap Connect to reader and connect to the first reader that shows up
3. Tap the 🕵️‍♂️ button to look for updates, make sure the console shows a `=== there is an update available`
4. Tap the 🚀 button to install the update.

The console should show a bunch of `==== progress  0.08` messages, and when it's done, a `=== the update has been completed. Moving on!`

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardReaderService.swift
+++ b/Hardware/Hardware/CardReader/CardReaderService.swift
@@ -58,5 +58,8 @@ public protocol CardReaderService {
 
     /// Triggers a software update. This method requires that checkForUpdates
     /// has been completed successfully
-    func installUpdate() -> Future<Void, Error>
+    ///
+    /// The returned publisher will periodically publish the fraction of progress during the software update
+    /// and it will complete when it's finished, unless there is any error.
+    func installUpdate() -> AnyPublisher<Float, Error>
 }
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -239,20 +239,22 @@ extension StripeCardReaderService: CardReaderService {
         }
     }
 
-    public func installUpdate() -> Future<Void, Error> {
-        return Future() { [weak self] promise in
-            guard let self = self,
-                  let pendingUpdate = self.pendingSoftwareUpdate else {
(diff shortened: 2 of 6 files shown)


ctarda (approved):
This looks much better than what we had.

This is what I see in the console (I tested only on the simulator, because the card reader can not be downgraded, and I am also scared to brick it)

Checking for updates:
```
=== there is an update available
=== version  1.00.03.32-SZZZ_Generic_v37-300001
=== estimated time  betweenOneAndTwoMinutes
=== check for updates completed
```

Updating:
```
==== progress  0.0
==== progress  0.0
==== progress  0.01
......
==== progress  0.97
==== progress  0.98
==== progress  0.99
=== the update has been completed. Moving on!
```
:shipit:

PR #4189, opened 2021-05-13
[Mobile Payments] Pair to reader when collecting a payment and a reader is not paired

### LINKED ISSUE #4062: [Mobile Payments] Launch pairing flow if users tap "Collect Payment" but there is not paired reader
1. Navigate to an eligible order, without pairing a reader first.
2. Tap "collect payment"

The SDK will throw an assertion because it has not been initialized at this point.

ctarda:
Closes #4062

Present the UI to discover and connect to card readers (v2) when an attempt to collect a payment is initiated without being connected to a reader

## Changes
* OrderDetailsViewModel will check if there is a reader connected
* If there is no reader connected, OrderDetailsViewController will present the UI to discover and connect.
* Once we have connected to a reader, the UI for connection is dismissed, and we initiate the process to collect the payment.

## How to test
Testing requires:

2. WCPay 2.4.0

* Checkout the branch
* Navigate to an order eligible for in-person payments, without connecting to a reader.
* Tap "Collect payment". There should be a fullscreen modal presenting the UI to discover and connect. Once the process is completed, the process to collect a payment should be initiated.
* Navigate to a different eligible order. Now there there is a reader connected, tapping "Collect payment" should start the flow to collect the payment directly.

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -611,6 +622,17 @@ private extension OrderDetailsViewController {
         }
     }
 
+    private func connectToCardReader() {
```
allendav, line 626: 😁

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -611,6 +622,17 @@ private extension OrderDetailsViewController {
...
+        guard let viewController = UIStoryboard.dashboard.instantiateViewController(ofClass: CardReaderSettingsPresentingViewController.self) else {
+            fatalError("Cannot instantiate `CardReaderSettingsPresentingViewController` from Dashboard storyboard")
+        }
+
+        let viewModelsAndViews = CardReaderSettingsViewModelsOrderedList()
```
allendav, line 631: This is an interesting spot where you might decide to have a different set of viewmodels, since you don't need to show the connected view for example. Don't need to do this now, just a thought for the future.

ctarda: Ah, that's a good point, I love this.

allendav (approved):
Tests beautifully. :shipit:

PR #4194, opened 2021-05-13
[Mobile Payments] Make view models responsible for subscription to connectedReaders

### LINKED ISSUE #3926: [Mobile Payments] Optimizing use of Combine in CardPresentPaymentStore

koke:
Fixes: #3926 
⚠️ Includes changes from #4180, to be merged first. The actual changes are in f9b1ef5

This is the last cancellable to get rid of for #3926, and possibly the most controversial 😬 

## Yosemite, Core Data, and Combine

The existing implementation had an `observeConnectedReaders` action that the store hooked into the service, calling its completion block each time the `connectedReaders` publisher sent a new value.

All the other cancellables so far were created by actual actions (connect, discover,...) that would eventually complete, but in this case, the view models want to observe just the connected reader, detached from the connect and disconnect actions. Because of this, the responsibility to cancel that subscription would fall on these view models which want to be notified. When the user navigates away from settings, they would get deallocated and stop listening for changes.

One way to shoehorn this requirement into the existing architecture would be to have the action's callback return the publisher instead, but that feels like we're bending the architectures too much.

When I look at [our documentation](https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/architecture-overview.md#yosemiteframework), it has this to say about actions:

> Lightweight entities expected to contain anything required to perform a specific task.
> Usually implemented by means of Swift enums, but can be literally any type that conforms to the Action protocol.
> 
> *Allowed* to have a Closure Callback to indicate Success / Failure scenarios.
> 
> **NOTE:** Success callbacks can return data, but the "preferred" mechanism is via the EntityListener or
> ResultsController tools.

I would argue that "observing the connected readers" is not an action: it doesn't perform any task, so there's no result to send back to the caller.

However, any of the suggested alternatives (`ResultsController` or `EntityListener`) won't work, since they're both based on Core Data objects. Our architecture was clearly designed to leverage Core Data for all reads, and it doesn't accommodate any model that's not a managed object. Now we have a new `Hardware` target, where there is data that needs to flow upwards but not necessarily go through Core Data, and there's a gap in our architecture.

I believe the simplest solution for now, since there's a single use case, is for the view models to use the `ServiceLocator` to access the publishers on the `CardReaderService` directly. There is no perfect solution here, but without more than a single use case, adding extra layers of abstraction seems a bit much. If we start needing this more, maybe we can evaluate a better solution.

## Sharing mocks for testing

I was stuck for a good while trying to figure out how to share `MockCardReaderService` from the `YosemiteTests` to the `WooCommerceTests`. I didn't find any good solutions but then I saw that `MockCardReader` was already present in both targets with slightly different implementations. I'm not proud of this, but I just did the same and copied that file to the `WooCommerceTests` target. I'd love to find a better solution for sharing mocks across targets/projects, but I don't have one yet.

## To test

I used the V2 card settings with a simulated reader, and tried to hit many different configurations:

- Go to settings (should not be connected) > Connect to a reader > Disconnect > Go back
- Go to settings (should not be connected) > Connect to a reader > Go back
- Go to settings (should be connected) > Disconnect > Go back
- Go to settings (should not be connected) > Connect to a reader > Disconnect > Go back

Update release notes:


ctarda:
> However, any of the suggested alternatives (ResultsController or EntityListener) won't work, since they're both based on Core Data objects. Our architecture was clearly designed to leverage Core Data for all reads, and it doesn't accommodate any model that's not a managed object. Now we have a new Hardware target, where there is data that needs to flow upwards but not necessarily go through Core Data, and there's a gap in our architecture.

I don't see it that way. There is clearly a gap, but it is in functionality. 

One of the features that we have left out of beta 1 is reconnecting automatically to known readers (that is, reconnecting to readers that we have connected in the past).

That's one of the reasons why we want to go through a layer in between: soon, there is data that we are passing between service and UI that we will need to store to the core data store and fetch from it.

> I would argue that "observing the connected readers" is not an action: it doesn't perform any task, so there's no result to send back to the caller.

I don't disagree. I do think though that, given that we are going to need to persist data, which, again, it's something we have been actively avoiding so far, is a tradeoff that makes it worth.

There is one more thing to consider. We will need to persist receipt data, payment intents, and several other entities. So we can not get completely rid of the stores and attack the service directly for every single use case. Again, it's a tradeoff, but I think we should have a consistent approach to consuming the services APIs for all the operations that it provides.

ctarda:
Or.... What if...

What if we encapsulated reader discovery behind a separate interface. Let's say spin off a new protocol, let's call it `CardReaderDiscovery`, or whatever makes sense. This new protocol exposes publishers for the readers, and is exposed via the ServiceLocator as well, and view models consume it directly.

We leave the rest of the operations in `CardReaderService` (start discovery, all things related to payments) as they are, and we access them via the current store. That way this functionality also gets linked to the Authenticated State.

Both interfaces are implemented by StripeCardReaderService.

We would rely on what CardReaderDiscovery provides only for the discovery UI, and then abstract all other functionality behind the store (e.g. remembering we have connected to a given reader, cancelling the discovery process after a given timeout...). 

Because the thing is, there is business logic related to collecting payments and integrating all that with WCPay, and I don't think we should get rid of it.

allendav:
The changes in this PR make sense to me and address a nagging concern I had about the settings view models. But I also like the layer that @ctarda has proposed in https://github.com/woocommerce/woocommerce-ios/pull/4194#issuecomment-840793102 and I'm inclined to say let's do that AND then have the view models subscribe from there?

ctarda:
If we end up splitting the interface in two, we could rename `CardReaderService` to `PaymentsService` as well. That would make the intent more clear, maybe?

koke:
> We will need to persist receipt data, payment intents, and several other entities. So we can not get completely rid of the stores and attack the service directly for every single use case.

To clarify here, I'm not suggesting we use this for every case. If we end up persisting something, leveraging Core Data and `EntityListener`/`ResultsController` seems like the most reasonable approach. What I'm advocating for is that not every single piece of data that flows upwards will have to be persisted, in this case the list of connected readers.

> What if we encapsulated reader discovery behind a separate interface. Let's say spin off a new protocol, let's call it CardReaderDiscovery...

Sorry, I understand what you are describing, but I don't understand why. What would be the benefit of doing that? If the new protocol is also going to be accessed through `ServiceLocator`, and implemented by the same object, I don't see what we get from it (other than simplifying the mocks for testing). It makes sense in term of grouping functionality, but IMO doesn't really address the gap in data flow.

In [WordPressFlux](https://github.com/wordpress-mobile/WordPress-iOS/tree/develop/WordPressFlux), we have the concept of "queries" that you can register in the flow to observe data instead of watching Core Data objects directly, which I think would fit what you're describing (having Yosemite expose the publishers instead of Hardware). However, that's possible because the UI layer has direct access to the stores instead of communicating indirectly though actions only.

> there is business logic related to collecting payments and integrating all that with WCPay, and I don't think we should get rid of it

This comment also confuses me, and makes me thing this proposal was understood as something else. The extent of my proposal is this PR, and I don't want to change any of the data flow beyond that. The business logic would still be contained in the store, and we'd access the service directly only to read the state.

koke:
I think I might have found a good solution in 1e78ddaf4. This is more in line with my ideal solution, and it turned out to be less complex than I anticipated.

I agree that it would be better if something in Yosemite exposed the publisher for the connected readers, instead of going straight from the UI to the service in Hardware. So far, we didn't have a way to do that without going through Core Data, but I added a new method for stores to expose publishers. The UI can't access a store's instance directly, but I found that using key paths make for a nice and concise API.

I also noticed that `StoresManager` already exposes a few accessors, and it seemed like a nice place to add `connectedReaders`:

- The UI would request the publisher from `ServiceLocator.stores.connectedReaders`
- The Store Manager will combine the publisher coming from the store when authenticated, with a fallback one that emits an empty value, when not authenticated. It will request this method via the new `StoresManagerState.publisher(keyPath:)` method.
- The store exposes a `connectedReaders` property with a publisher, leaving room for future customization of that publisher behavior, which would still contain that logic within the store.

docs/YOSEMITE.md
```diff
@@ -34,6 +34,8 @@ Each subclass of `Store` registers a processor with the `Dispatcher`. To registe
 
 That set of actions is declared in an implementation of the `Action` protocol. We will look at `Action` in details in the next section, but for now,  it is important to note that there is a one to one relationship between each subclass of `Store` and a corresponding implementation of the `Action` protocol. (i.e. for `AccountStore` we would also have a `AccountAction`).
 
+Stores can also expose `Combine` publishers for read-only access to data that's not persisted in Core Data. For accessing managed objects, the preferred method is to rely on `EntityListener` or `ResultsControllers`. Any publishers exposed by a store can be accessed through the `publishers(keyPath:)` method of the `StoresManagerState`.
```
koke: Writing this, I'm now wondering if it makes sense to also add a `publisher(keyPath:)` method to the Store Manager so the UI layer can request a key path directly.

Right now, exposing `connectedReaders` as a shorthand reads more nicely, but if this become useful in more scenarios, then the store manager might start to have too many properties. 🤔 

```swift
// With publisher
ServiceLocator.stores.publisher(keyPath: \CardPresentPaymentStore.connectedReaders)
// With shorthand
ServiceLocator.stores.connectedCardReaders
```

ctarda:
> Sorry, I understand what you are describing, but I don't understand why. What would be the benefit of doing that? If the new protocol is also going to be accessed through ServiceLocator, and implemented by the same object, I don't see what we get from it (other than simplifying the mocks for testing). It makes sense in term of grouping functionality, but IMO doesn't really address the gap in data flow.

We would abstract different functionality behind different interfaces, one to be consumed directly by the UI, another one to be consumed only through a store. The UI should not have direct access to, for example, the functionality necessary to collect payments. 

> is that not every single piece of data that flows upwards will have to be persisted, in this case the list of connected readers.

I agree. We need to find a good place for this, and for other upcoming pieces of data that won't be persisted. Would like to point out also, that persistence is not the most important reason for finding a place for this type of data. The lack of business logic associated to it is.

> I also noticed that StoresManager already exposes a few accessors, and it seemed like a nice place to add connectedReaders:

StoresManager is meant to coordinate the interaction with the Stores, I don't think exposing card readers there is at the same level of abstraction. 

> Writing this, I'm now wondering if it makes sense to also add a publisher(keyPath:) method to the Store Manager so the UI layer can request a key path directly.

I like that though. It does not expose anything specific to card readers.

What about adding that publisher(keyPath:) method to SessionManagerProtocol instead of StoresManager? SessionManagerProtocol abstracts session-specific things, I guess we could consider readers something session-specific.

koke:
Just leaving a note here that, while testing this, I noticed there seems to be some sort of retain cycle with the V2 settings, as the subscriptions aren't being cleared correctly. I checked with the memory graph debugger, and after opening and closing settings several times, there were several instances of the view controllers and models, all holding subscriptions to the publishers.

koke:
> What about adding that publisher(keyPath:) method to SessionManagerProtocol instead of StoresManager? 

I saw that one, and it feels like a better fit, although the `SessionManager` doesn't access state or stores right now. I didn't see a clear way to make SessionManager aware of data coming from stores without changing the current design too much. Any ideas for that one?

Maybe the store manager can inject a `publishers` callback when initializing the SessionManager (giving up the `.standard` shorthand), and that would forward every call to the state?

koke:
That said, the documentation for `SessionManager` says:

> SessionManager provides persistent storage for Session-Y Properties

And it seems like all the properties there are persisted across sessions, either in UserDefaults or the keychain. I would consider connected readers session-specific, but I probably wouldn't say the same about the things already in `SessionManager` 😅

ctarda:
>  I didn't see a clear way to make SessionManager aware of data coming from stores without changing the current design too much. Any ideas for that one?

I am a bit lost here. Help me understand why we'd need SessionManager to know about the Stores.

My understanding was that all we wanted to expose is the publishers for subscribing to changes to connected readers. Is that correct?

The way I see it, we would have two interfaces (humour me here for a second). One of them exposes whatever is needed for subscribing to those publishers (or even the actual publishers themselves). This interface is unknown to Stores, and know only to SessionManager. The other interface is known only to the Stores. Both interfaces are exposed separately by the ServiceLocator, and both are implemented by the same class. 

That was we can separate the two concerns that we want to expose separately (one being "observing available card readers" and the other one being "operations and commands related to external card readers")

How does that look?

koke:
> Help me understand why we'd need SessionManager to know about the Stores

The `publisher` method is meant to expose publishers that are properties of a store without having a direct reference to that store. If we don't want those publishers to live within a store and access the Hardware code directly, then I don't think we need the `publisher` method.

> My understanding was that all we wanted to expose is the publishers for subscribing to changes to connected readers. Is that correct?

Yes, that's all we need for now.

> This interface is unknown to Stores, and know only to SessionManager

So, if I understand this correctly, the SessionManager would get the data directly from the StripeCardReaderService (through a different protocol). But how does the SessionManager gets access to this? Directly from `ServiceLocator.cardReaderService` (or the equivalent for a new protocol)?

> This interface is unknown to Stores, and know only to SessionManager

I understand that we don't want things in the UI calling code for payments directly, nor we want any of the actions related to connection called directly either. However, if `ServiceLocator.cardReaderService` already exists today and is not a problem, I'm not sure how splitting the functionality helps with this, unless we restrict access to `ServiceLocator.cardReaderService` to the store only.

I would understand having a new read-only protocol to monitor connected readers if we managed to lock down the rest of the code so that only the store gets to call it. Maybe I can give that a try next week

ctarda:
>  how does the SessionManager gets access to this? Directly from ServiceLocator.cardReaderService (or the equivalent for a new protocol)?

Yes.

> However, if ServiceLocator.cardReaderService already exists today and is not a problem, I'm not sure how splitting the functionality helps with this

True, there is no way we can prevent the UI from using a global dependency. We would need to rely on a bit of convention over configuration.

To me, we have two different concerns. Two different things. That grants two different interfaces, even though there is no way we can enforce that one of them is not used by the UI.

> having a new read-only protocol to monitor connected readers if we managed to lock down the rest of the code so that only the store gets to call it. Maybe I can give that a try next week

I don't think that's really necessary. We can rely on convention for that.

At this point though, I would like to circle back to the origin of the discussion, to your first argument, because I feel like we might be over-complicating things a bit.

Let's assume for a second that we go with two interfaces. The one exposing the publishers for card readers can be consumed directly from the UI via `ServiceLocator.connectedReaders` or whatever way we call it, and uses it directly to subscribe to those publishers. 

The other interface is exposed through the service locator as well, and we trust folks to consume it only through Stores. After all, that's exactly what is happening now.

This would give us what we want, without having to reach deep into the application architecture this close to the beta 1 release.

How does that sound?

koke:
I'm going to close this one for now because it bothers me having a draft PR on hold for this long. If/when we decide to resume work on it, we can always reopen and have GitHub restore the branch.

PR #4196, opened 2021-05-13
[Mobile Payments] Card Reader Settings: Allow for priority viewmodelandview to be nil

### LINKED ISSUE #4195: [Mobile Payments] Fix reevaluatePriorityViewModelAndView to allow it to return nil when needed
Right now the guarding will prevent nil from being returned even when appropriate (i.e. we transition from at least one view model shouldShow to no viewmodels sholdShow - this can happen imperceptibly briefly as it turns out.)

Props @koke 

Re https://github.com/woocommerce/woocommerce-ios/pull/4123#discussion_r627098865

allendav:
Fixes #4195 

Changes:
- Restores the possibility of `reevaluatePriorityViewModelAndView` returning nil (i.e. when no viewmodel thinks it should show)
- On further inspection, this can happen imperceptibly briefly as we advance through connecting a reader

To test:
- Settings > Manage Card Readers V2
- Connect a reader
- Disconnect it

Update release notes:


ctarda (approved):
:shipit:

PR #4202, opened 2021-05-14
[Mobile Payments] Prevent retain cycles when presenting Card Reader Settings V2

### LINKED ISSUE #4201: [Mobile Payments] Possible retain cycle in V2 settings
As I was testing #4194, I noticed that subscriptions to publishers from V2 settings weren't cleared up correctly. Since I'm storing those in a `Set` of cancellables, they should automatically go away when the view model is deallocated, which should happen in theory when the user navigates away from the screen.

As I navigated back and forth, I'd see several subscription receiving events for the same discovery events.

To reproduce this with the Memory Graph Debugger:

1. Edit the scheme, and enable "Malloc Stack Logging" in the Diagnostics tab. I think this is only recommended to detect leaks, and Xcode isn't flagging this as a Leak for some reason, but it might help

Screen Shot 2021-05-14 at 13 21 26

2. Run the app, go to the V2 settings and back a few times
3. Switch to the Memory Graph Debugger

Screen Shot 2021-05-14 at 13 25 55

4. Notice several instances of the related View Controllers and Models

Screen Shot 2021-05-14 at 13 26 36

koke:
Fixes #4201

This PR tweaks a few callbacks in the Card Reader Settings V2 files to prevent retain cycles.

## To test

Follow the instructions in #4201, and make sure no extra instances of the `CardReader*` classes are kept in memory after navigating away.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
@@ -12,7 +12,9 @@ final class CardReaderSettingsPresentingViewController: UIViewController {
     /// Set our dependencies
     func configure(viewModelsAndViews: CardReaderSettingsPrioritizedViewModelsProvider) {
         self.viewModelsAndViews = viewModelsAndViews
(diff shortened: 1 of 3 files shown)


ctarda (approved):
Good catch! 

Tested, didn't notice any leaks. 

:shipit:

allendav:
Thank you @koke ! Adding this to my personal reference sheet as a solid example

PR #4203, opened 2021-05-14
Merge release/6.6 final into develop

AliSoftware:
Build 6.6.0.1 just got submitted for review to Apple today.


loremattei (commented):
@AliSoftware I don't see any new translation which looks weird to me 🤔 Is this expected?

AliSoftware:
@loremattei Nice catch, thanks!

This is due to a bug in the toolkit, which fails to glob the `.strings` files to `git add` and thus create an empty commit instead of commiting them. I spot and fixed that issue manually last 2 times, but accidentally missed it this time around. This bug has been recently fixed in the toolkit since, but WCiOS hasn't been updated to point to it yet (I'll update it as part of the upcoming Sprint). 

I've updated the translations, and discarded the build in ADC and will resubmit. Please re-review those updated PRs (develop and trunk) in the meantime 🙂

AliSoftware:
The CI failures are because the build of UI tests was cancelled (by me triggering the release build, I assume)

loremattei (approved):
Looks better now :-) 
:shipit:

Issue #4077, opened 2021-04-29
[Mobile Payments] Closing the Loop: Add customer name, email, order, order_key and other info to intent metadata


allendav:
Both the WCPay server and WCPay client rely on certain bits of metadata to be added by when creating payment intents.

The fields are, at this time:

- `customer_name`: e.g. "Allen Snook"
- `customer_email`
- `site_url`: e.g. "https://staging.marsiswaiting.com"
- `order_id`: e.g. 6140
- `order_key`: e.g. wc_order_0000000000000
- `payment_type`: e.g. `single`

The site_url is the URL of the store. The order_key is available from order metadata in _order_key. The [payment_type](https://github.com/Automattic/woocommerce-payments/blob/e099744fb8cead46fe23ec25d785e40827947c1f/includes/constants/class-payment-type.php#L20-L23) is `single` for one-time payments and `recurring` for subscription payments.

The WCPay server cares about order_key. It should not be omitted.

The client relies certain bits to display transaction details. Without it, some of the transaction columns in the list view are empty and the transaction detail view in wp-admin fails to load properly at all.

![missing](https://user-images.githubusercontent.com/1595739/116614823-47c9f500-a8ef-11eb-9c21-65c243316260.png)

See also #3863 

cc @ctarda @malinajirka

ctarda:
We can break this down in two. First, for beta one, we can add all the metadata, except for `order_key`. 

As soon as we merge into develop and touching the core data store is safe, we'd need to add the order_key to the Order object.

allendav:
Moved back to "To do" until we are ready to do the remaining step: order key

allendav:
Closing this issue - we have a separate issue now just for order key: #4282

PR #4208, opened 2021-05-14
[Mobile Payments] Add first set of metadata to the payment intent

allendav:
Partially addresses #4077 

This PR adds all the metadata except for order key. That will come in a later PR.

There are two ways to see this working:
- Set a breakpoint in `PaymentCaptureOrchestrator.swift` to halt after metadata is set. Ensure you see the expected values for your order after you begin the capture payment flow, i.e.:

breakpoint

- Or, load the transaction in the Stripe dashboard after switching to that account's view:

dashboard2

- Also, optionally ensure that the transaction can be rendered, including the order ID, in wp-admin using this branch (note: the other metadata values are NOT displayed in this view): https://github.com/Automattic/woocommerce-payments/pull/1840

- Lastly, please run the newly added unit tests

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/PaymentIntent.swift
+++ b/Hardware/Hardware/CardReader/PaymentIntent.swift
@@ -66,16 +66,16 @@ public extension PaymentIntent {
 }
 
 public extension PaymentIntent {
-    enum PaymentTypes {
+    enum PaymentTypes: String {
         /// A payment that IS NOT associated with an order containing a subscription
         /// This key is also used by the plugin when it creates payment intents.
         ///
-        public static let single = "single"
+        case single = "single"
 
         /// A payment that IS associated with an order containing a subscription
         /// This key is also used by the plugin when it creates payment intents.
         ///
-        public static let recurring = "recurring"
+        case recurring = "recurring"
     }
 }
 
@@ -88,7 +88,7 @@ public extension PaymentIntent {
                           siteURL: String? = nil,
                           orderID: Int64? = nil,
                           orderKey: String? = nil,
-                          paymentType: String? = nil
+                          paymentType: PaymentTypes? = nil
     ) -> [AnyHashable: Any] {
         var metadata = [AnyHashable: Any]()
 
@@ -117,7 +117,7 @@ public extension PaymentIntent {
         }
 
         if paymentType != nil {
-            metadata[PaymentIntent.MetadataKeys.paymentType] = paymentType
+            metadata[PaymentIntent.MetadataKeys.paymentType] = paymentType?.rawValue
         }
 
         return metadata
--- a/Hardware/Hardware/CardReader/PaymentIntent.swift
+++ b/Hardware/Hardware/CardReader/PaymentIntent.swift
@@ -92,33 +92,13 @@ public extension PaymentIntent {
     ) -> [AnyHashable: Any] {
         var metadata = [AnyHashable: Any]()
 
-        if store != nil {
-            metadata[PaymentIntent.MetadataKeys.store] = store
-        }
-
-        if customerName != nil {
-            metadata[PaymentIntent.MetadataKeys.customerName] = customerName
-        }
-
-        if customerEmail != nil {
-            metadata[PaymentIntent.MetadataKeys.customerEmail] = customerEmail
-        }
-
-        if siteURL != nil {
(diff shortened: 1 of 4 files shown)


Hardware/Hardware/CardReader/PaymentIntent.swift
```diff
@@ -30,5 +30,96 @@ public extension PaymentIntent {
...
+        /// This key is also used by the plugin when it creates payment intents.
+        ///
+        public static let recurring = "recurring"
+    }
+}
+
+// MARK: - Convenience Initializers
+//
+public extension PaymentIntent {
+    static func initMetadata(store: String? = nil,
```
ctarda, line 85: Just for kicks and giggles, this could be a good candidate for a [builder](https://en.wikipedia.org/wiki/Builder_pattern). Not sure it's worth the effort though

allendav: I was hoping there was a Swift-ier way. I'll look at builder for a future PR? Thankfully the unit tests will catch if/when a builder doesn't build it the same way :)

koke: I think this looks fine as it is, but if you are looking for something more Swifty, you could have a Metadata struct that then converts to a dictionary. Also, to simplify, it's OK to assign a `nil` value to a dictionary: it will clear the value for that key if it exists, or it will be ignored if it doesn't.

```swift
public extension PaymentIntent {
    struct Metadata {
        let store: String?
        let customerName: String?
        let customerEmail: String?
        //...
        
        var asDictionary: [String: Any] {
            var metadata = [String: Any]()
            metadata[Keys.store] = store
            metadata[Keys.customerName] = customerName
            metadata[Keys.customerEmail] = customerEmail
            // ...
            return metadata
        }

        /// Metadata Keys
        enum Keys {
            public static let store = "paymentintent.storename"
            public static let customerName = "customer_name"
            public static let customerEmail = "customer_email"
            // ...
        }
    }
}
```

A possibly swiftier way would be to use `Codable` and a [dictonary encoder](https://elegantchaos.com/2018/02/21/decoding-dictionaries-in-swift.html), but bringing in a new dependency or writing our own encoder seems too much for something like this.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -116,13 +116,35 @@ private extension PaymentCaptureOrchestrator {
...
+            customerName = firstName
+        }
+
+        if let lastName = order.billingAddress?.lastName {
+            if customerName == nil {
+                customerName = lastName
+            } else {
+                customerName? += " " + lastName
+            }
+        }
```
koke, line 131: This might be OK considering that we're launching only in the US first, but I wonder if we need to [localize full names](https://www.w3.org/International/questions/qa-personal-names) better. Luckily, it seems Apple provides a [PersonNameComponentsFormatter](https://developer.apple.com/documentation/foundation/personnamecomponentsformatter) for that.

koke (approved):
Code looks good to me and the metadata seems fine when testing on breakpoints. I don't know how to access the Stripe dashboard though

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: I agree! This is (presently) how the backend does it as well. I'll open an issue here and there

Hardware/Hardware/CardReader/PaymentIntent.swift
allendav: > it's OK to assign a nil value to a dictionary: it will clear the value for that key if it exists, or it will be ignored if it doesn't

Removed unneeded `nil` checks in https://github.com/woocommerce/woocommerce-ios/pull/4208/commits/b76acdc243e9a1a251f8ad0f2a22b3aab6f2022f

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: https://github.com/woocommerce/woocommerce-ios/issues/4217

PR #4209, opened 2021-05-17
Bump deployment target to iOS 14.0

koke:
As discussed in p91TBi-51t-p2, we are bumping the required iOS version to 14.0.

This will happen in the 6.8 release, set to be released June 14.

## Changes

- [x] Update all the targets to iOS 14.0
- [x] `SKStoreReviewController.requestReview()` was deprecated in iOS 14

## To test

I'd recommend general testing of the app.
The only thing that has changed besides the target is the app review prompts, so maybe do some extra testing there. I had to modify the code to make sure I got the prompt and make sure `AppRatingManager.shouldPromptForAppReview(section:)` returned `true`

Update release notes:


pmusolino:
I tested it and checked if everything is in place. I have not noticed any oddities and it seems all right!

pmusolino (approved):
LGTM :shipit:

koke:
🎉 Thanks for the review!

PR #4218, opened 2021-05-18
[Mobile Payments] Add code to suppress Wallet during card capture

### LINKED ISSUE #4079: [Mobile Payments] Disable Apple Pay on the device during card capture
Observed (again) today on my iPhone XR and BBPOS Chipper 2XBT during payment capture. I thought I had written this issue up already but alas I apparently did not.

Otherwise, your phone will attempt to complete the payment for you if the card reader is within a few inches of the phone when you start card capture.

See https://developer.apple.com/documentation/passkit/pkpasslibrary/1617078-requestautomaticpasspresentation

Props Karl/Stripe for the suggested fix.

Note:

> This method requires a special entitlement from Apple. If the entitlement isn’t present, the request fails with a PKPassLibrary result. For more information, see developer.apple.com/apple-pay/.

cc @ctarda @AliSoftware

allendav:
Closes #4079 

Changes:
- NOTE: I had to download the latest provisioning profile (July 8) for this to work.
- Adds the entitlement for wallet suppression for Debug, Release and Alpha
- Disables wallet when payment collection begins
- Restores wallet presentation when payment collection completed

To test:
- On a phone with Wallet set up (e.g. Apple Card or a bank debit or credit card)
- Start the Woo app
- Position the card reader directly next to the phone
- Enter an order and begin payment collection
- Ensure the phone's wallet does NOT appear
- Complete payment
- Launch the Woo app on another phone or tablet and bring the first device to the 2nd devices reader
- Ensure the first phone's wallet DOES appear. Cancel out of the wallet
- Repeat the above process, but instead of completing the payment cancel it and ensure the wallet DOES reappear when the 1st device is in proximity to the 2nd device's reader

Noteworthy:

The following appears on the screen the first time the merchant attempts to collect payment. This is expected behavior for this entitlement:

cc @ctarda @astralbodies 

Update release notes:


WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -19,6 +21,16 @@ final class PaymentCaptureOrchestrator {
             onCompletion(.failure(CardReaderServiceError.paymentCapture()))
             return
         }
+
+        if !PKPassLibrary.isSuppressingAutomaticPassPresentation() {
+            PKPassLibrary.requestAutomaticPassPresentationSuppression() { result in
+                guard result == .success else {
+                    DDLogWarn("Automatic Pass (Apple Pay) Presentation Suppression failed. Reason: \(result.rawValue)")
```
Ecarrion: Just got here from the p2 where the entitlement was approved. I know this is not yet ready for review but do you think we should also log this to sentry? It seems like an error we would want to fix sooner than later.

allendav: Yes, lets

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -10,6 +12,8 @@ final class PaymentCaptureOrchestrator {
 
     private let celebration = PaymentCaptureCelebration()
 
+    private var walletSuppressionRequestToken = PKSuppressionRequestToken(0)
```
koke: Is there any guarantee that `0` won't ever be returned? An optional `PKSuppressionRequestToken` might be safer here.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -35,6 +45,7 @@ final class PaymentCaptureOrchestrator {
                                                                     break
                                                                 }
                                                              }, onCompletion: { [weak self] result in
+                                                                self?.allowWalletPresentation()
```
AliSoftware: What if the user sends the app into background / foreground between the `collectPayment` call and its completion? Especially, if the user initiates a payment, then goes into background, and soon after (i.e. when app code is still running and hasn't been suspended by the OS yet) the transaction completes and the completion block is called?

Asking especially given p91TBi-5xP-p2 mentioned this: 

> * Do not call the suppression API, if `– [UIApplication applicationState:]` is in the Background state.

Which means that I think it's worth double-checking that test/use case

AliSoftware:
> @AliSoftware - does what I did for the entitlement files look right to you? I've tested this successfully on Debug, but I'm not sure about Alpha - is there a way to test it and Release too?

I'll let your new Release Manager @jkmassel take over this, but it should be possible to do an alpha build to AppCenter on demand to test this (even though from what I see in the diff the integration seems ok to me, but we never know with code signing and entitlements, so worth testing indeed)

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: I'm pretty sure there is no way for a transaction to complete while the app is in background, but I will test this.

ctarda: I think you are already aware of this, but it might be necessary to call allowWalletPresentation earlier as well, if the operator or the customer cancels before tapping/inserting/swapping the card.

Other than that, I just tested on a device and it seems to work well.

allendav: Hey @ctarda - I ran into something interesting while hooking up allowing pass presentation when the payment collection is cancelled. Would like to run this by you because it is a slightly involved change needed:

As you've already seen, in this PR I’ve added pass presentation suppression to the responsibilities of PaymentCaptureOrchestrator.collectPayment. 

I’ve also already added re-allowing pass presentation to the onCompletion handler PaymentCaptureOrchestrator.collectPayment passes to the action creator collectPayment, and that re-enables pass presentation just fine when the payment is collected.

That leaves re-allowing pass presentation when payment collection is cancelled.

I was surprised to see that CardPresentModalTapCard queues up cancelPayment directly. That unfortunately leaves the orchestrator out of the loop. I think that’s not a good idea.

I’d like to change CardPresentModalTapCard to NOT queue up CardPresentPaymentAction.cancelPayment but to instead call a completion that would be passed to its init

I’d like to change OrderDetailsViewController (which has the paymentAlerts) to pass into CardPresentModalTapCard init a completion that calls OrderDetailsViewModel.cancelPayment. In this way, the orchestrator is aware of the cancellation and can do things like allowing pass presentation. This also decouples the tap-card-modal from the store.

What do you think?

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: Added sentry in https://github.com/woocommerce/woocommerce-ios/pull/4218/commits/784c6cb0cefd01a19e1aa6665d837b985f499c22

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4218/commits/784c6cb0cefd01a19e1aa6665d837b985f499c22

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: @AliSoftware - I've added a check to avoid calling the API if the app is in background in https://github.com/woocommerce/woocommerce-ios/pull/4218/commits/784c6cb0cefd01a19e1aa6665d837b985f499c22

allendav: @AliSoftware - FWIW, starting payment collection and then immediately backgrounding the app and then tapping the payment card to the reader causes payment failure. I don't think it is likely the merchants will background the app in the middle of the payment collection tap window. I think it is a corner case.

We could monitor, I suppose, for this corner case and wait until the app is foreground again to then attempt to allow passes again, but I think that would add unjustified complexity and likely exceed Apple's expectations. I think it is sufficient that we re-enable pass presentation if the payment is completed or cancelled in the foreground. Do you concur?

AliSoftware: I think we should choose one take or the other but not a mix of the two:
- Either we consider that it's unlikely that the vendor would put the app in background, and remove the `guard` checks you added in the previous commit, and just not handle that edge case at all
- Or handle the case to be sure to follow Apple's guidelines, which means we should then not just `return` in that `guard` statement and do nothing if in background, but instead register for the `ApplicationStateChanged` notification at that point (or any similar solution), to queue a block that would do the call only when the app gets back in foreground instead of immediately in those cases.

Otherwise if we only implement it halfway, we'll encounter corner cases, where if the app is backgrounded between the payment collection modal and its validation, the pass presentation will never be restored after payment because the call to restore it would have happened while in background but you returned and exit early in that case, instead of queuing that call for when you get back in the app.


I agree that this might add complexity both in the implement and the testing scenarios, and I'm actually not sure what Apple expects us to do in such scenarios in case we really needed to handle that complexity anyway (if we were in the middle of a transaction, should we always restore the pass presentation when the app goes in background and suppress it again when we get back in foreground? Or is that supposed to be handled by the OS? Why do they have this rule/restriction about background, is it just to prevent abuse? Or would it create actual issues for the end users?).
So maybe it's just simpler to not handle that corner case altogether at that stage?

allendav: @AliSoftware wrote:

> Either we consider that it's unlikely that the vendor would put the app in background, and remove the guard checks you added in the previous commit, and just not handle that edge case at all... or....

I think it is VERY unlikely the merchant would start payment collection and background the app. I'll revert the checks for background.

ctarda: > I’d like to change OrderDetailsViewController (which has the paymentAlerts) to pass into CardPresentModalTapCard init a completion that calls OrderDetailsViewModel.cancelPayment. In this way, the orchestrator is aware of the cancellation and can do things like allowing pass presentation

@allendav Yes, that does make sense to me. I don't think that should be block this PR for going in though? I mean, that sounds like something that can be tackled in a different PR, imo.

ctarda (approved):
This looks like it is ready to cover most of the uses cases, and I think we could go ahead and merge it, noting whatever things we need to do as a follow up for later.

This one is a :shipit: from me.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: OK. I'll remove the background check for @AliSoftware and open a separate ticket for restoring pass presentation if payment collection is cancelled

Edit: Separate ticket for cancellation: https://github.com/woocommerce/woocommerce-ios/issues/4649

allendav: Removed background guards in https://github.com/woocommerce/woocommerce-ios/pull/4218/commits/0ddb45a917b1aead3948fc6206aa4f518cff5df4

koke: > I think it is VERY unlikely the merchant would start payment collection and background the app

I'm not sure about this. I agree it's unlikely that a merchant will intentionally switch tasks while collecting payment, but there are many cases where the app might go to the background, like receiving a phone call, or accidentally dragging from the bottom of the screen initiating the app switcher, even if the gesture is canceled.

allendav: I'll take a second look at this as part of #4649

PR #4226, opened 2021-05-19
Add quick check for nested frameworks

AliSoftware:
Following the issue we had with TestFlight validation and nested frameworks introduced by the Xcode 12.4 bug (see paNNhX-ee-p2 for details), this check will hopefully catch if we encounter such a configuration issue again later.

## To test

I already tested this myself (see comments below) but it could be nice to validate the check works on other's machines too.
To test this:

- Checkout this branch, open the WooCommerce workspace in Xcode 12.4 and build the project
- Check that you don't get any error during the build. In particular, you might want to go to the Build Log tab and expand the `Run custom shell script 'Check for nested frameworks'` step at the end and check that it says `✅ No nested framework found, you're good to go!`
- Select the `Yosemite` project in the workspace, go to the "General" tab, and in the "Frameworks and Libraries" section, change the line for `Codegen` from "Do Not Embed" to "Embed & Sign"
- Build the project again, and check that you get a build error telling you all about the nested framework issue.

### DIFF
--- a/Scripts/check-nested-frameworks.sh
+++ b/Scripts/check-nested-frameworks.sh
@@ -12,7 +12,12 @@ if [ -z "$NESTED_FMKS_DIRS" ]; then
 else
     echo "❌ Found nested \`Frameworks\` folder inside frameworks of final bundle."
     for fmk_dir in $NESTED_FMKS_DIRS; do
+        # Extract the name of the parent framework containing the nested ones
         parent_fmk=$(basename $(dirname $fmk_dir) .framework)
+        # Extract the list of frameworks nested inside that parent framework. In the next command:
+        #  * `-depth 1` is to avoid logging cases of "C nested in B itself nested in A" (2+ levels of nesting), since C nested in B will already be logged when looping on fmk_dir=B, so no need to log it during fmk_dir=A too.
+        #  * The `sed` command removes the leading `./` in the paths returned by `find`, then quote the results in backticks for nicer formatting in final message.
+        #  * The `tr` command joins all the lines (= found frameworks) with a `,`. Note that this will result in an extra comma at the end of the list too, but we'll get rid of that in the final message using ${nested_fmks%,} bash substitution.
         nested_fmks=$(cd "${fmk_dir}" && find . -name '*.framework' -depth 1 | sed "s:^./\(.*\)$:\`\1\`:" | tr '\n' ',')
         echo "error: Found nested frameworks in ${fmk_dir} -- Such a configuration is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework(s) ${nested_fmks%,} within the \`${parent_fmk}\` Xcode project which links to them. You might need to use Xcode 12.5 to fix this, due to an Xcode 12.4 bug – see paNNhX-ee-p2"
     done
--- a/Scripts/check-nested-frameworks.sh
+++ b/Scripts/check-nested-frameworks.sh
@@ -1,0 +1,20 @@
+#!/bin/sh -euo pipefail
+
+# Nested frameworks (i.e. having a Frameworks/ folder inside *.app/Frameworks/.framework) is invalid and will make the build be rejected during TestFlight validation.
+# This can happen especially due to an Xcode 12.4 UI bug when linking binary frameworks to the project which always embed the binary (and if you try to change to Do Not Embed it also removes it from linked libraries)
+# This is a bug in Xcode 12.4 UI that is fixed in Xcode 12.5, so fixing nested frameworks to "Do Not Embed" can be fixed using Xcode 12.5, while still continue using Xcode 12.4 for development after the fix.
+
+# This script is intended to be used as a Build Phase on the main app target, as the very last build phase (and especially after the "Embed Frameworks" phase)
+
+NESTED_FMKS_DIRS=$(find "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}" -name Frameworks -depth 2) 
+if [ -z "$NESTED_FMKS_DIRS" ]; then
+    echo "✅ No nested framework found, you're good to go!"
+else
+    echo "❌ Found nested \`Frameworks\` folder inside frameworks of final bundle."
+    for fmk_dir in $NESTED_FMKS_DIRS; do
+        parent_fmk=$(basename $(dirname $fmk_dir) .framework)
+        nested_fmks=$(cd "${fmk_dir}" && find . -name '*.framework' -depth 1 | sed "s:^./\(.*\)$:\`\1\`:" | tr '\n' ',')
+        echo "error: Found nested frameworks in ${fmk_dir} -- Such a configuration is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework(s) ${nested_fmks%,} within the \`${parent_fmk}\` Xcode project which links to them. You might need to use Xcode 12.5 to fix this, due to an Xcode 12.4 bug – see paNNhX-ee-p2"
+    done
+    exit 1
+fi 
--- a/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
+++ b/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
@@ -6010,8 +6010,8 @@
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
-			shellPath = /bin/sh;
-			shellScript = "# Nested frameworks (i.e. having a Frameworks/ folder inside *.app/Frameworks/.framework) is invalid and will make the build be rejected during TestFlight validation. But this can happen especially due to an Xcode 12.4 UI bug when linking binary frameworks to the project which always embed the binary (bug that is fixed in Xcode 12.5)\n\nNESTED_FMKS=$(find \"${TARGET_BUILD_DIR}\"/WooCommerce.app/Frameworks/*.framework/Frameworks -name '*.framework') \nif [ -z \"$NESTED_FMKS\" ]; then\n    echo \"✅ No nested framework found, you're good to go!\"\nelse\n    for f in $NESTED_FMKS; do\n        nested=$(basename $f)\n        parent=$(basename $(dirname $(dirname $f)) .framework)\n        echo \"error: Found nested framework in $f -- Such a configuration is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework \\`$nested\\` within the \\`$parent\\` Xcode project containing it. You might need to use Xcode 12.5 to fix this, due to an Xcode 12.4 bug – see paNNhX-ee-p2\"\n    done\n    exit 1\nfi \n";
+			shellPath = "/bin/sh -euo pipefail";
+			shellScript = "${PROJECT_DIR}/../Scripts/check-nested-frameworks.sh\n";
 			showEnvVarsInLog = 0;
 		};
 		57CCFFD4249D2A5700825FCF /* SwiftLint */ = {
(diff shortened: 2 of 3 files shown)


koke:
I think we can add this as a Run Script phase, so we catch the error locally and not just on CI.

This seems to work, although I'm not too familiar with build environment vars to know if `TARGET_BUILD_DIR` is the right one:

```sh
NESTED_FMKS=$(find "${TARGET_BUILD_DIR}"/WooCommerce.app/Frameworks/*.framework -name Frameworks) 
if [ -n "$NESTED_FMKS" ]; then
    echo "error: Found some nested frameworks inside Frameworks/*/Frameworks subdirectories. Such a configuration is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework in the Xcode project of the parent framework containing it"
    echo $NESTED_FMKS
    exit 1
fi 
```

AliSoftware:
Oh that's a good point and good idea 👍  Yeah will try as a build phase indeed!

AliSoftware:
Updated the PR to use a Script Build Phase, and this is what it looks like in Xcode when it finds nested frameworks:
image

[And as a bonus, here's the expected build failure on CI too when the fix hasn't been applied yet](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/11992/workflows/58a04641-f49c-4ec6-9a9a-f6ba3e13a760/jobs/21761):
```
▸ Running script 'Check for nested frameworks'

❌  error: Found nested framework in /Users/distiller/project/DerivedData/Build/Products/Debug-iphonesimulator/WooCommerce.app/Frameworks/Networking.framework/Frameworks/Codegen.framework -- Such a configura
tion is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework `Codegen.framework` within the `Networking` Xcode project containing it. You might need to use Xcode 12.5 to fix this, due to an Xcode 12.4 bug – see paNNhX-ee-p2

❌  error: Found nested framework in /Users/distiller/project/DerivedData/Build/Products/Debug-iphonesimulator/WooCommerce.app/Frameworks/Storage.framework/Frameworks/Codegen.framework -- Such a configuration is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework `Codegen.framework` within the `Storage` Xcode project containing it. You might need to use Xcode 12.5 to fix this, due to an Xcode 12.4 bug – see paNNhX-ee-p2

❌  error: Found nested framework in /Users/distiller/project/DerivedData/Build/Products/Debug-iphonesimulator/WooCommerce.app/Frameworks/Yosemite.framework/Frameworks/Codegen.framework -- Such a configuration is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework `Codegen.framework` within the `Yosemite` Xcode project containing it. You might need to use Xcode 12.5 to fix this, due to an Xcode 12.4 bug – see paNNhX-ee-p2
```

I'm gonna rebase that branch on top of the fix now to double-check that the script goes green and doesn't error anymore once fixed.

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
```diff
@@ -5994,6 +5995,25 @@
...
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "# Nested frameworks (i.e. having a Frameworks/ folder inside *.app/Frameworks/.framework) is invalid and will make the build be rejected during TestFlight validation. But this can happen especially due to an Xcode 12.4 UI bug when linking binary frameworks to the project which always embed the binary (bug that is fixed in Xcode 12.5)\n\nNESTED_FMKS=$(find \"${TARGET_BUILD_DIR}\"/WooCommerce.app/Frameworks/*.framework/Frameworks -name '*.framework') \nif [ -z \"$NESTED_FMKS\" ]; then\n    echo \"✅ No nested framework found, you're good to go!\"\nelse\n    for f in $NESTED_FMKS; do\n        nested=$(basename $f)\n        parent=$(basename $(dirname $(dirname $f)) .framework)\n        echo \"error: Found nested framework in $f -- Such a configuration is invalid and will be rejected by TestFlight. Please fix by choosing 'Do Not Embed' for the nested framework \\`$nested\\` within the \\`$parent\\` Xcode project containing it. You might need to use Xcode 12.5 to fix this, due to an Xcode 12.4 bug – see paNNhX-ee-p2\"\n    done\n    exit 1\nfi \n";
```
AliSoftware: @koke Seems like using `$TARGET_BUILD_DIR` was the right decision, according to [my favorite Build Settings reference](https://pewpewthespells.com/blog/buildsettings.html#target_build_dir) 👍 :
> Run Script build phases that operate on product files of the target that defines them should use the value of this build setting.

AliSoftware:
And this is what it looks like after rebasing on top of `develop` after the commit that got the fix 🎉 
image

AliSoftware:
Retrospective P2 post here for reference: paNNhX-ee-p2

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
```diff
@@ -5605,6 +5605,7 @@
 				B7A94351C1ADC31EA528B895 /* [CP] Embed Pods Frameworks */,
 				CE1445302188ED0300A991D8 /* Zendesk Strip Frameworks */,
 				59DF5B32D7C07692EC8B4C59 /* [CP] Copy Pods Resources */,
+				095040D72655531C001D08FA /* Check for nested frameworks */,
```
mokagio, line 5608: Shameless plug for suggestion out of the scope of this PR. When having multiple custom run script build phases, I find it useful to track them in a dedicated ["main" script](https://mokacoding.com/blog/better-build-phase-scripts/#group-multiple-scripts-in-a-master-script) and only run that one.

Just leaving it out there...

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
mokagio: Following up on the comment above, having the run script in an actual script also makes it simpler to read and edit it, as well as having a sensible diff to look at as opposed to a _single very long line_ (`:xcode:` :trollface: )

mokagio (approved):
Default behavior:

Screen Shot 2021-05-20 at 11 27 41 am

Hacked behavior to force failure:

Screen Shot 2021-05-20 at 11 29 19 am

👍 :shipit:

---

_One nitpick_

I found the error output from `find` before the successful setup message confusing. "_Why am I seeing something that looks like an error when the log tells me everything is fine?_ 🤔"

Have you considered forwarding `find`'s output to `/dev/null`?

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
AliSoftware: That's a nice article but I'm having trouble seeing how it would apply here, given that we have only that single script having to be run at the very very end of the list of Build phases and especially after the embed frameworks phase — so isolated from any other script we might otherwise have in the list of Build Phases?

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
AliSoftware: Yeah I agree that I could have extracted that script into a file in Scripts/….sh, to make it more readable and diff-able 👍
That's actually something I usually do so not sure why I didn't that time around 😺

AliSoftware:
> I found the error output from `find` before the successful setup message confusing. "_Why am I seeing something that looks like an error when the log tells me everything is fine?_ 🤔"
>
> Have you considered forwarding `find`'s output to `/dev/null`?

@mokagio I can't forward the output of find to `/dev/null` because that would defeat the purpose of having the list of nested frameworks as an output of this command when some are found. But I should be able to `|| true` after the command to avoid this case though, will try that.

koke:
I think the error on success is because the `*.framework/Frameworks` comes from shell expansion and not find, so when they don't exist, we are calling `find` with an invalid path.

I think this should work instead:
```sh
find "${TARGET_BUILD_DIR}"/WooCommerce.app/Frameworks -name Frameworks -depth 2
```

AliSoftware:
> I think the error on success is because the `*.framework/Frameworks` comes from shell expansion and not find, so when they don't exist, we are calling `find` with an invalid path.
> 
> I think this should work instead:
> 
> ```shell
> find "${TARGET_BUILD_DIR}"/WooCommerce.app/Frameworks -name Frameworks -depth 2
> ```

This is now fixed – and also extracted in a dedicated script file as @mokagio suggested 🎉 

I had to tweak the script a bit more so that the error messages were still nice with this new result format of the new `find` command in all situations… but this comes out pretty nice!

all-green-case
errors-case

Scripts/check-nested-frameworks.sh
```diff
@@ -0,0 +1,20 @@
...
+# This script is intended to be used as a Build Phase on the main app target, as the very last build phase (and especially after the "Embed Frameworks" phase)
+
+NESTED_FMKS_DIRS=$(find "${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}" -name Frameworks -depth 2) 
+if [ -z "$NESTED_FMKS_DIRS" ]; then
+    echo "✅ No nested framework found, you're good to go!"
+else
+    echo "❌ Found nested \`Frameworks\` folder inside frameworks of final bundle."
+    for fmk_dir in $NESTED_FMKS_DIRS; do
+        parent_fmk=$(basename $(dirname $fmk_dir) .framework)
+        nested_fmks=$(cd "${fmk_dir}" && find . -name '*.framework' -depth 1 | sed "s:^./\(.*\)$:\`\1\`:" | tr '\n' ',')
```
AliSoftware, line 21: * `-depth 1` here is so that if `Networking.framework` is embed in `Yosemite` but `Codegen` is also embed in `Networking`, we don't want to print about the 2nd level ie the  `WooCommerce.app/Frameworks/Yosemite.framework/Frameworks/Networking.framework/Frameworks/Codegen.framework` – since the nesting of Codegen within Networking will already be printed as a separate error
* The `sed` call is to get rid of the `./` at the beginning of each output line of `find . -name …`, and also to quote each line between a pair of backticks
* The `tr` call is to join all the lines with a `,`. This will add an extra `,` at the end of the list, but we'll get rid of this by using bash substitution `${nested_fmks%,}` when printing the list line 17, to get rid of the suffixed comma.

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
mokagio: Oh I see. Fair enough. I saw the "Zendesk Strip Framework" in the diff and also noticed the call to Swiftlint in the project. But that's true: grouping all the scripts into a single one does force them to execute all at the same point of the build phase, which might not be desirable.

Nevermind 👍

Scripts/check-nested-frameworks.sh
mokagio, line 21: Super useful comment! Thank you for adding it (and for extracting the script 🙇‍♂️ )

```suggestion
        # `-depth 1` here is so that if `Networking.framework` is embed in `Yosemite` but 
        # `Codegen` is also embed in `Networking`, we don't want to print about the 2nd level 
        # ie the
        # `WooCommerce.app/Frameworks/Yosemite.framework/Frameworks/Networking.framework/Frameworks/Codegen.framework` 
        # – since the nesting of Codegen within Networking will already be printed as a 
        # separate error.
        #
        # The `sed` call is to get rid of the `./` at the beginning of each output line of 
        # `find . -name …`, and also to quote each line between a pair of backticks.
        # 
        # The `tr` call is to join all the lines with a `,`. This will add an extra `,` at the 
        # end of the list, but we'll get rid of this by using bash substitution 
        # `${nested_fmks%,}` when printing the list line 17, to get rid of the suffixed comma.
        nested_fmks=$(cd "${fmk_dir}" && find . -name '*.framework' -depth 1 | sed "s:^./\(.*\)$:\`\1\`:" | tr '\n' ',')
```

mokagio (approved):
Default:

Screen Shot 2021-05-21 at 10 38 35 am

Hacked to emulate issue:

image

Scripts/check-nested-frameworks.sh
AliSoftware: Thanks! Reworded that comment a bit and pushed those explanations in eff300ed5cdd361da4915553572447e7d32cd9b3

Issue #4085, opened 2021-04-30
Carrier and rates selection


pmusolino:
Add the ability for the user to select the shipping carrier in the shipping label creation flow.

Feel free to split this issue into sub-tasks.

Tasks:
- [x] Add the endpoint `POST: wc/v1/connect/label/{{order_id}}/rates` in Networking
- [x] Implement the Yosemite Layer
- [x] Implement the UI
- [x] Manage the error state
- [x] Top banner when the user pay a flat free for the shipping label.
- [x] Save/handle selection 
- [x] When users select “signature required”, all prices update with signature required options

<img width="914" alt="Schermata 2021-04-30 alle 12 29 17" src="https://user-images.githubusercontent.com/495617/116683322-b2426a00-a9af-11eb-88ec-aaa60f6274f3.png">

Issue #4095, opened 2021-04-30
Shipping Labels: Address validation banner doesn't display at full height after tapping Done button


rachelmcr:
**Describe the bug**

When creating a shipping label, the address validation banner shows up as expected (full height, all text visible) when it appears immediately when the Edit Address screen appears.

However, if you edit the address (using an address that can't be validated), and tap Done, the address validation banner only displays as a single line with the text cut off.

**To Reproduce**

1. Start with an order that has an address that can't be automatically validated. An easy way to test this is with an order that doesn't have a shipping address set at all.
2. On the Orders tab, open the order details.
3. Press the button "Create Shipping Label".
4. Press the button "Continue" in Create Shipping Label form, under the `Ship From` cell. Confirm the address.
5. Press the button "Continue" in Create Shipping Label form, under the `Ship To` cell. This should open a screen to edit the address, with the validation banner displayed as expected.
6. Enter the required address details, but again use an address that can't be automatically validated.
7. Tap Done. Notice that the address validation banner appears again, but not at full height and only a single line of text appears.

**Screenshots**

Validation banner - as expected | Validation banner - after tapping Done
-|-
![Simulator Screen Shot - iPhone 11 Pro - 2021-04-30 at 13 37 56](https://user-images.githubusercontent.com/8658164/116696905-647f2f00-a9ba-11eb-976d-c140de697db7.png)|![Simulator Screen Shot - iPhone 11 Pro - 2021-04-30 at 13 35 01](https://user-images.githubusercontent.com/8658164/116696909-6517c580-a9ba-11eb-97f6-f0931347a1a2.png)

Issue #4092, opened 2021-04-30
Shipping Labels: Empty state for Package Selected (packages list) screen


rachelmcr:
We need some empty state message for the Package Selected screen, when there are no packages set up on the backend.

@koke noted in https://github.com/woocommerce/woocommerce-ios/pull/4073#issuecomment-829892893 that if you don't have any packages set up on your store's backend, the Package Selected screen (which usually shows the packages list) shows a blank screen:

<img src="https://user-images.githubusercontent.com/8739/116658705-ae531f80-a990-11eb-94fc-f910b1887ee8.png" width="300px">

In wp-admin:

![image](https://user-images.githubusercontent.com/8658164/116685741-c041bc00-a9aa-11eb-84cf-eb48182a4d95.png)

pmusolino:
Thanks for discovering this issue @koke and thanks for opening the issue @rachelmcr. I'm moving this to M3 👍

rachelmcr:
I took a closer look at this (preparing to ask for design input) when I realized it might not be needed:

* For M3, we're targeting stores that already have at least one package (see p91TBi-2om-p2):

> Stores with WCS installed, payment method already added, with at least one custom package already created and shipping within the United States.

* For M4, we'll support the custom package creation flow. At that point, the Package Selected screen won't be shown right away if there aren't any packages on the store — instead, selecting the Package Selection field on the Package Details screen will take you into the package creation flow.

As long as we make sure to check for existing packages to determine if a site is eligible for shipping label creation in M3, users won't see this screen in an empty state.

rachelmcr:
I don't think we need an empty state screen here for the M3 release.

How I tested:

1. Started with a test store with WooCommerce Shipping & Tax set up, except no packages added under WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Set the `shippingLabelsM4` feature flag to `false` (to only test features included in the M3 release).
3. Opened an eligible order on the Orders tab of the app.
4. Confirmed that the "Create Shipping Label" button doesn't appear on the order.
5. Added a package in wp-admin, opened the order again, and confirmed the "Create Shipping Label" button now appears.

[...]

Issue #3995, opened 2021-04-19
Shipping Labels: fix navigation title showed in TabBar in Package List


pmusolino:
As found by @rachelmcr here https://github.com/woocommerce/woocommerce-ios/pull/3967 there is a bug in the Package List screen implemented in SwiftUI. It seems that the navigation title is showed also in the tab bar as `UITabBarButtonLabel`.
It is probably caused by the fact that the SwiftUI view is originally presented by a View Controller.

![Schermata 2021-04-19 alle 17 08 12](https://user-images.githubusercontent.com/495617/115262219-9da2de00-a134-11eb-8dae-32240f906859.png)

pmusolino:
After a more in-depth investigation, the only way I've found to solve it effectively is to embed the second view in the `UIHostingController` as well and make the navigation happen between the view controllers. This is clearly not a "lightweight" solution, because it would also mean exposing the model data to the view controllers, and finding a way to get the views in SwiftUI to communicate with the view controllers.

I continue to investigate.

Ecarrion:
Do we know why it happens? Or is that still a mystery?

pmusolino:
No idea why it happens. It's like if there is a conflict with our TabBar. The text remains inside the TabBar also on screens that are not related to SwiftUI, and I wasn't able to find someone else with the same issue. @Ecarrion

pmusolino:
I made some progress.
In the initial `UIHostingController`, we always set the title in this way:
```
navigationItem.title = "My title"
```

Looking in detail at the view hierarchy it seems like a new tab item is added in the tabbar.

For this reason, I changed it to `navigationController?.title = Localization.navigationBarTitle` and now when I push the SwiftUI View, that tab item is still there, but when I pop back to the original view, the tab item disappear.

Ecarrion:
@pmusolino I spent a little bit on this and got nowhere 🤷 

However, I noticed that we could mitigate this if we hide the `TabBar` when pushing the shipping label creating flow 🤷 🤷 

I think we should try to properly fix it, but if not, We can go with the workaround. What do you think?

Branch: https://github.com/woocommerce/woocommerce-ios/compare/issue/3995-possible-fix?expand=1
Demo:

https://user-images.githubusercontent.com/562080/116469021-37eeda00-a837-11eb-8262-a51321c98dbc.mov

pmusolino:
Thank you @Ecarrion for spending time investigating this issue. 🙇‍♂️ 

[...]

PR #4229, opened 2021-05-19
Shipping Labels: basic UI for Carriers and Rates

pmusolino:
Part of #4085 

## Description
In this PR I implemented a basic UI for the Carriers and Rates screen, implementing:
- The main view for Carriers and Rates implemented in SwiftUI.
- A simple view model for Carriers and Rates that fetch the carriers and rates from the remote API, and manage the sync status. 
- The `ShippingLabelCarrierRow` displays the information of every single carrier.
- The selected package is now a computed property in `ShippingLabelFormViewModel`.
- `ShimmerModified` applies an animated `shimmer` effect to any SwiftUI view. In our specific case, it will be used with the `redacted` modifier for showing a ghost cell.
- Added the 3 carriers logos in the app (UPS, USPS, FedEx).

## Testing
7. You should see the screen that shows the Package Details screen and all the items that will be shipped.
8. Press the "Package selected" row. 
9. Select a package, then press "Done".
10. Add a Total package weight, press "Done".
11. The "Shipping Carrier and Rates" cell should be enabled now. Press it.
12. You should see the new screen "Carrier and Rates". It should display first the ghost cells, then the data fetched from the API (the data showed are still not formatted).

## GIF

Update release notes:


shiki:
@pmusolino I was trying to review this but I think I'm doing something wrong or my site could be incorrectly configured. Or I incorrectly followed the instructions. 🤣  

The “Package Selected” seems to be empty for me. 



Also, there's a weird “Package Selected” label at the bottom:



What do you think I'm doing wrong? 

---

Anyway, I'm leaving this for review for someone else so you get unblocked cause clearly I made a mistake somewhere. 😆 But if no one gets to it tomorrow, we can figure it out and I can continue reviewing.

pmusolino:
Thanks for trying to test the shipping labels @shiki 😄 

> The “Package Selected” seems to be empty for me.

Probably you don't have any packages set up on your store's backend. If you configure them, you should see them in the app. This wouldn't happen when we will ship the entire M3. Related issue: https://github.com/woocommerce/woocommerce-ios/issues/4092

> Also, there's a weird “Package Selected” label at the bottom:

We already know that issue which seems a bug on the Apple side (I filled a radar for it) 🗡️ and we tried different roads to fix it https://github.com/woocommerce/woocommerce-ios/issues/3995 I'll try different things since now we dropped iOS 13.

shiki:
> Probably you don't have any packages set up on your store's backend. 

@pmusolino Ah that was it! 🤦 Will continue reviewing...

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRow.swift
```diff
@@ -0,0 +1,67 @@
+import SwiftUI
+
+struct ShippingLabelCarrierRow: View, Identifiable {
+
+    let id = UUID()
+    let title: String
+    let subtitle: String
+    let price: String
+    let image: UIImage?
```
shiki: I still can't get over with the idea that only publicly accessible properties can be part of the automatic `struct` initializer. I want properties to have clear access levels whenever possible. But I understand the convenience. It's Swift that should get better. 😅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRow.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+}
+
+private extension ShippingLabelCarrierRow {
+    enum Constants {
+        static let zStackWidth: CGFloat = 48
+        static let vStackSpacing: CGFloat = 8
+        static let hStackSpacing: CGFloat = 25
+        static let hStackPadding: CGFloat = 10
+        static let height: CGFloat = 60
```
shiki: ⚠️  Should this be `minHeight`?

```suggestion
        static let minHeight: CGFloat = 60
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRow.swift
```diff
@@ -0,0 +1,67 @@
...
+            .previewLayout(.fixed(width: 375, height: 60))
+            .previewDisplayName("With Image")
+
+        ShippingLabelCarrierRow(title: "UPS Ground",
+                                subtitle: "3 business days",
+                                price: "$2.49",
+                                image: nil)
+            .previewLayout(.fixed(width: 375, height: 60))
+            .previewDisplayName("Without Image")
+    }
```
shiki, line 84: Do you think it's useful to have a large font size preview as well? 

```suggestion
        ShippingLabelCarrierRow(title: "UPS Ground",
                                subtitle: "3 business days",
                                price: "$2.49",
                                image: nil)
            .frame(maxWidth: 375, minHeight: 60)
            .previewLayout(.sizeThatFits)
            .environment(\.sizeCategory, .accessibilityExtraExtraExtraLarge)
            .previewDisplayName("Large Font Size Roar!")

    }
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRow.swift
```diff
@@ -0,0 +1,67 @@
...
+        .padding([.leading, .trailing], Constants.padding)
+        .frame(minHeight: Constants.height)
+        .contentShape(Rectangle())
+    }
+}
+
+private extension ShippingLabelCarrierRow {
+    enum Constants {
+        static let zStackWidth: CGFloat = 48
+        static let vStackSpacing: CGFloat = 8
```
shiki, line 42: There might be some opportunity to standardize the constants a bit in the future. Mainly following our standard Layout. P9MsLS-1g-p2

We probably would end up not needing some of the constants. 

Not sure if it's worth it though. 🤷 

```suggestion
    /// Global constants conforming to Woo Mobile layout standards
    /// (P9MsLS-1g-p2)
    enum Layout {
        /// Base length for spacing, gaps, padding.
        /// The name could be better XD
        static let baseLength: CGFloat = 8
    }

    enum Constants {
        /// We probably wouldn't even need these constants
        static let zStackWidth = Layout.baseLength * 6
        static let vStackSpacing = Layout.baseLength

```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRates.swift
```diff
@@ -0,0 +1,60 @@
+import SwiftUI
+import Yosemite
+
+struct ShippingLabelCarriersAndRates: View {
```
shiki: ⚠️  I got a bit confused by this because there's also a `ShippingLabelCarriersAndRates` model. Should we consider renaming this to something else? `ShippingLabelCarriersAndRatesView` maybe? Should we standardize views to always have a `*View` suffix?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+         originAddress: ShippingLabelAddress,
+         destinationAddress: ShippingLabelAddress,
+         packages: [ShippingLabelPackageSelected],
+         stores: StoresManager = ServiceLocator.stores) {
+        self.order = order
+        self.originAddress = originAddress
+        self.destinationAddress = destinationAddress
+        self.packages = packages
+        self.stores = stores
+        syncCarriersAndRates()
```
shiki: We need an easy `viewDidLoad` in SwiftUI 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+    var ghostRows: [ShippingLabelCarrierRow] {
+        var tempRows: [ShippingLabelCarrierRow] = []
+        for _ in 0..<3 {
+            let row = ShippingLabelCarrierRow(title: "Ghost title",
+                                              subtitle: "Ghost subtitle",
+                                              price: "Ghost price",
+                                              image: nil)
+            tempRows.append(row)
+        }
+        return tempRows
```
shiki: ℹ️  This one also works. A bit shorter and no need for a temp variable. 

```suggestion
        Array(0..<3).map { _ in
            ShippingLabelCarrierRow(title: "Ghost title",
                                    subtitle: "Ghost subtitle",
                                    price: "Ghost price",
                                    image: nil)
        }

```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+private extension ShippingLabelCarriersAndRatesViewModel {
+
+    func syncCarriersAndRates() {
+        syncStatus = .loading
+        let action = ShippingLabelAction.loadCarriersAndRates(siteID: order.siteID,
+                                                              orderID: order.orderID,
+                                                              originAddress: originAddress,
+                                                              destinationAddress: destinationAddress,
+                                                              packages: packages) { [weak self] (result) in
+            switch result {
```
shiki: ⚠️  I'm curious why we're not using a `guard let self = self else { return }` here. 🤔

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
shiki: Though I actually wonder if `ghostRows` even belongs in the VM since it's a UI implementation and not necessarily “data” that the VM is responsible for. In that sense, I also wonder if `ShippingLabelCarrierRow` itself should be returned by the `ViewModel` since that is a view (SwiftUI) and not data per se. I think I would have expected the VM to just _emit_ `ShippingLabelCarrierRate` models. And it's the view's responsibility to `map()` those models into `ShippingLabelCarrierRow`.

Just thinking out loud. No strong opinions at the moment. 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+                var tempRows: [ShippingLabelCarrierRow] = []
+                for rate in response.defaultRates {
+                    let row = ShippingLabelCarrierRow(title: rate.title,
+                                                     subtitle: "\(rate.deliveryDays) business days",
+                                                     price: "\(rate.retailRate)",
+                                                     image:
+                                                        UIImage(named: "shipping-label-ups-logo")!)
+                    tempRows.append(row)
+                }
+                self?.rows = tempRows
```
shiki: ℹ️  Same as above, looks like this can be:

```suggestion
                self?.rows = response.defaultRates.map { rate in
                    ShippingLabelCarrierRow(title: rate.title,
                                            subtitle: "\(rate.deliveryDays) business days",
                                            price: "\(rate.retailRate)",
                                            image: UIImage(named: "shipping-label-ups-logo")!)
                }

```

shiki: ⚠️  Also the `UIImage()` is temporary I think? I did not see that mentioned in #4085.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -274,6 +274,23 @@ private extension ShippingLabelFormViewController {
...
+              let selectedPackage = viewModel.selectedPackage else {
+            return
+        }
+
+        let vm = ShippingLabelCarriersAndRatesViewModel(order: viewModel.order,
+                                                        originAddress: originAddress, destinationAddress: destinationAddress,
+                                                        packages: [selectedPackage])
+
+        let hostingVC = UIHostingController(rootView: ShippingLabelCarriersAndRates(viewModel: vm))
+        hostingVC.title = Localization.navigationBarTitleCarriersAndRates
```
shiki: ⚠️  Perhaps we can consider moving this inside `ShippingLabelCarriersAndRatesViewModel` so "Carrier and Rates" configurations are located in a single place. 

```swift 
struct ShippingLabelCarriersAndRatesViewModel {
    let preferredViewControllerTitle = NSLocalizedString()
}

hostingVC.title = vm.preferredViewControllerTitle
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -26,6 +26,37 @@ final class ShippingLabelFormViewModel {
...
+        for customPackage in packagesResponse.customPackages {
+            if customPackage.title == selectedPackageID {
+                return ShippingLabelPackageSelected(boxID: customPackage.title,
+                                                    length: customPackage.getLength(),
+                                                    width: customPackage.getWidth(),
+                                                    height: customPackage.getHeight(),
+                                                    weight: customPackage.getWidth(),
+                                                    isLetter: customPackage.isLetter)
+            }
+        }
```
shiki: ⚠️ I think we should use an `if let` here instead so it's easier to see that this is an early exit. 

```suggestion
        if let customPackage = packagesResponse.customPackages.first(where: { $0.title == selectedPackageID }) {
            return ShippingLabelPackageSelected(boxID: customPackage.title,
                                                length: customPackage.getLength(),
                                                width: customPackage.getWidth(),
                                                height: customPackage.getHeight(),
                                                weight: customPackage.getWidth(),
                                                isLetter: customPackage.isLetter)
        }

```

What do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -26,6 +26,37 @@ final class ShippingLabelFormViewModel {
     private(set) var destinationAddress: ShippingLabelAddress?
     private(set) var packagesResponse: ShippingLabelPackagesResponse?
     private(set) var selectedPackageID: String?
+    var selectedPackage: ShippingLabelPackageSelected? {
```
shiki, line 29: ⚠️ I don't see this covered in `ShippingLabelFormViewModelTests`. Should we add a test for this? 😁

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -26,6 +26,37 @@ final class ShippingLabelFormViewModel {
...
+                if predefinedPackage.id == selectedPackageID {
+                    return ShippingLabelPackageSelected(boxID: predefinedPackage.id,
+                                                        length: predefinedPackage.getLength(),
+                                                        width: predefinedPackage.getWidth(),
+                                                        height: predefinedPackage.getHeight(),
+                                                        weight: predefinedPackage.getWidth(),
+                                                        isLetter: predefinedPackage.isLetter)
+                }
+            }
+        }
```
shiki: Not quite sure how to make this more expressive. It's tricky. 😄 

This is what I got so far but I haven't tested it! 

```swift
        // The lazy is critical for performance
        if let predefinedPackage = packagesResponse.predefinedOptions.lazy.compactMap({ option in
            option.predefinedPackages.first(where: { $0.id == self.selectedPackageID })
        }).first {
            return ShippingLabelPackageSelected(boxID: predefinedPackage.id,
                                                length: predefinedPackage.getLength(),
                                                width: predefinedPackage.getWidth(),
                                                height: predefinedPackage.getHeight(),
                                                weight: predefinedPackage.getWidth(),
                                                isLetter: predefinedPackage.isLetter)
        }
```

I'm not convinced this is better. 🤣 And I don't think it even works.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ShimmerModifier.swift
```diff
@@ -0,0 +1,90 @@
+import SwiftUI
+
+/// A view modifier that applies an animated "shimmer" to any view,
+/// ideal to be used with the `redacted` modifier.
+/// Author: https://github.com/markiv/SwiftUI-Shimmer
```
shiki, line 5: Does this mean that we would need to include their license too?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
```diff
@@ -0,0 +1,82 @@
+import UIKit
+import SwiftUI
+import Yosemite
+
+final class ShippingLabelCarriersAndRatesViewModel: ObservableObject {
```
shiki: ⚠️ Are we going to add a unit test for this later?

shiki (commented):
This is looking good! I have some questions/suggestions that could probably use your attention (⚠️ ). And some suggestions (ℹ️ ). The rest are mostly philosophical questions and feel free to ignore them. 😄  

Not approving yet because I noticed that there's one that could use a unit test. But let me know if that's something you wanted to do later. Or I'm wrong and it's not even necessary. 😅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRow.swift
pmusolino: It would be great to have a standard for the Constants, but at the same time, just some spacing properties are always the same. For example, sometimes for the leading/trailing padding we use 8, or 16, or 20. We should also use a format that should be easily understandable, where we don't risk to broke anything modifying a constant in the future. It would be ideal for me to write a P2 post about this, to discuss this thing with the team. What do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRates.swift
pmusolino: It seems that Apple changed the way we should use the `*View` suffix. In fact, in SwiftUI all the components don't have the `*View` suffix: it's `VStack` not a `VStackView`. We used some `*View` suffix in the past for some SwiftUI views in our app, but probably we should update them.
I renamed it `ShippingLabelCarriers` because the rates are part of the carriers. 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
pmusolino: There is no reason not to, so I updated my code! 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
pmusolino: Yes all the data are still temporary, and they will be handled/displayed in a future PR 👍

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ShimmerModifier.swift
pmusolino: Probably you are right. Also if it's a [MIT License](https://github.com/markiv/SwiftUI-Shimmer/blob/main/LICENSE) (the most permissive) we should also include the original license. Given that, probably it's better to include this small piece of code using SPM? What do you think?

We can still include the license also doing a copy/paste of the license on top of the file.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
pmusolino: For sure! I didn't include the unit tests because the data will change in the next PRs.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: I tried this version, a "combo" between my version and yours 😄 

```
        for option in packagesResponse.predefinedOptions {
            if let predefinedPackage = option.predefinedPackages.first(where: { $0.id == selectedPackageID }) {
                    return ShippingLabelPackageSelected(boxID: predefinedPackage.id,
                                                        length: predefinedPackage.getLength(),
                                                        width: predefinedPackage.getWidth(),
                                                        height: predefinedPackage.getHeight(),
                                                        weight: predefinedPackage.getWidth(),
                                                        isLetter: predefinedPackage.isLetter)
            }
        }
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
pmusolino: If it sounds good to you, I can implement a super simple view model that will be pass to `ShippingLabelCarrierRow`, in this way the `ShippingLabelCarriersViewModel` will returns just the view model of the row, and the `ShippingLabelCarriers` will be responsible to map those models into `ShippingLabelCarrierRow`.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRow.swift
shiki: Sounds good! Only if you have time though! I didn't want to add another thing to your plate. It's not a strong opinion for me at the moment since I'd need more experience with SwiftUI. 😄

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ShimmerModifier.swift
shiki: Including it as an SPM would probably be best. As long as we don't break our builds again. 😄 A separate SPM also informs us that “this is third-party code and we shouldn't touch it”. 

Totally cool if you'd like to do this later and leave it as is for now.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
shiki: Alrighty! In subsequent PRs is fine. I just wasn't sure if it was intentional or not. Sorry about that! 🙂

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRates.swift
shiki: I see! Probably something I have to get used to then.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
shiki: That sounds better and seems to fit with how we used to do stuff in UIKit. Our `UITableViewCells` are always passed a `Model` of some sort. That simple `RowViewModel` could also have something like:

```swift
struct RowViewModel {
    /// If true, then use the fancy skeleton view
    let isGhost: Bool
}
```

Just a possibly stupid idea from me. 😄 

Totally fine to do this later when you have time. I'm not aware of the time constraints so don't wanna disrupt your flow. 🙂

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriers.swift
```diff
@@ -30,13 +30,22 @@ struct ShippingLabelCarriers: View {
                 }
             }
         }
+        .navigationTitle(Localization.titleView)
```
shiki, line 33: Oh, wow! That's so much better! 😄

shiki (approved):
Re-reviewed and re-tested. Works great! Thanks, @pmusolino. I liked the “combo”. 🤣 

:shipit: 

---

Out of scope: I also noticed this “notice” which does not scale well.

pmusolino:
Thank you @shiki for your review 🙇 

> Out of scope: I also noticed this “notice” which does not scale well.

Yes, we already know that there is an issue on the top banner https://github.com/woocommerce/woocommerce-ios/issues/4095 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersAndRatesViewModel.swift
pmusolino: No problem at all, I'll do that right now ;)

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ShimmerModifier.swift
pmusolino: I'll handle it in a different PR ;)

Issue #4082, opened 2021-04-30
[Mobile Payments] Add Tracks events


ctarda:
Add Tracks events. 

For context: p91TBi-52G-p2

koke:
All events have been added, validated, and registered.

PR #4233, opened 2021-05-19
[Mobile Payments] Tracks events for collecting payment

koke:
Part of #4082 

## Events included

- card_present_collect_payment_tapped: The user tapped on the Collect Payment button to collect a card payment.
- card_present_collect_payment_canceled: The user canceled the payment process.
- card_present_collect_payment_failed: The payment process failed.
- card_present_collect_payment_success: The payment was completed successfully.

I tested that all the events are triggered and received on the backend.

One question I have is whether we should add more detail on some of the errors. For instance, for `paymentDeclinedByPaymentProcessorAPI`, the documentation says to inspect `SCPProcessPaymentError.requestError` 
for more details, which could include useful information:

> ...requestError = Error Domain=com.stripe-terminal Code=9020 "Allowable number of PIN tries exceeded." UserInfo={com.stripe-terminal:StripeAPIDocUrl=https://stripe.com/docs/error-codes/card-declined...

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
@@ -78,8 +78,7 @@ private extension PaymentCaptureOrchestrator {
                                     captureResult: Result<PaymentIntent, Error>,
                                     onCompletion: @escaping (Result<CardPresentReceiptParameters, Error>) -> Void) {
         switch captureResult {
-        case .failure:
-            let error = CardReaderServiceError.paymentCapture()
+        case .failure(let error):
             onCompletion(.failure(error))
         case .success(let paymentIntent):
             submitPaymentIntent(siteID: order.siteID,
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -78,8 +78,7 @@ private extension PaymentCaptureOrchestrator {
                                     captureResult: Result<PaymentIntent, Error>,
                                     onCompletion: @escaping (Result<CardPresentReceiptParameters, Error>) -> Void) {
         switch captureResult {
-        case .failure:
-            let error = CardReaderServiceError.paymentCapture()
+        case .failure(let error):
```
koke, line 81: @ctarda was there a reason to replace the received error with a new one? This was erasing all the details and logging an `internalServiceError` instead

ctarda: Not that I recall. Go figure 🤷🏼

ctarda (approved):
:shipit: 

I can see the events going out. 

I don't have a great answer to your question about logging errors as part of the event. My first  inclination is that, being errors related to payments, the less details we log, the better. WE don't want to log any sensitive information.

On the other hand, SCPProcessPaymentError.requestError contains information that Stripe provides, so I guess we can assume it's safe to log?

Do you think that question might be worth a p2?

PR #4240, opened 2021-05-20
[Mobile Payments] Parse and rename eligibility flag

### LINKED ISSUE #4160: [Mobile Payments] Update WCPayAccount.init(from: Decoder)  to parse value for CPP eligibility flag

ctarda:
Closes #4160 

There is ongoing work to pass down the eligibility flag from WCPay; https://github.com/Automattic/woocommerce-payments-server/pull/813

To be completely honest, I am not sure if submitting this now is actually a good thing, give we'd still need to come back to this in #4241 . My thinking was that it's better to get a bit closer to the final stage now and leave one line (in #4241) for later. But who knows.

The feature flag is hardcoded to true for now, but I've logged #4241 to revert that as soon as https://github.com/Automattic/woocommerce-payments-server/pull/813 is available

## Changes
* Parse the eligibility flag using the proper key, to integrate with the work done on the backend.
* Rename the eligibility flag to match the name it has on the backend.

I excluded unit tests, will add those with #4241 

## How to test
Testing requires:

2. WCPay 2.4.0

* Checkout the branch
* Connect to a reader
* Navigate to an eligible order. The "Collect Payment" button should be available in debug builds
* Navigate to Settings. The "Manage Card Readers" section should be available in debug builds.

Update release notes:


Networking/Networking/Model/WCPayAccount.swift
```diff
@@ -37,7 +38,7 @@ public struct WCPayAccount: Decodable {
         self.supportedCurrencies = supportedCurrencies
         self.country = country
         /// Hardcoded until support for this property is available in WCPay
-        self.canCollectPayments = true
+        self.isCardPresentEligible = true
```
ctarda, line 41: This is what we'd need to change in #4241 , and set the property to the value passed as parameter, instead f hardcoding it to true

koke:
> should be available in debug builds

Is there something making the feature not available in release builds? Should we test for that?

ctarda:
@koke yes, there is a feature flag: `cardPresentPayments`. The entry points to the feature in the UI ("Collect payment" button  and "Manage Card Readers" section in Settings) should be gated with that feature flag as well

koke (approved):
🤦🏽 I was looking at the line of code where it's checked and somehow missed it. Tested the change and the feature is still available

Issue #4232, opened 2021-05-19
[Mobile Payments] Reader discovery fails silently if Bluetooth permission is denied


koke:
To reproduce:

1. On a fresh install, deny the Bluetooth permission. Otherwise go to System Settings > Woo, and disable Bluetooth
2. Try to connect to a card reader

<img src="https://user-images.githubusercontent.com/8739/118857321-32d1f780-b8d8-11eb-9d9e-7fcfce3443e7.png" width="400">

The call to `Terminal.shared.discoverReaders` immediately calls completion with this error:

```Error Domain=com.stripe-terminal Code=3200 "Generic bluetooth error." UserInfo={NSLocalizedDescription=Generic bluetooth error., com.stripe-terminal:Message=Generic bluetooth error.}```

However, the UI still appears to be "Scanning for reader"

PR #4244, opened 2021-05-20
[Mobile Payments] Tracks events for reader discovery and connection

koke:
Fixes of #4082

## Events included

- card_reader_discovery_tapped: The user taps on Connect Card Reader and the app starts looking for readers to connect to
- card_reader_discovery_failed: The discovery process fails. This is likely because the SDK was busy with another command.
- card_reader_discovery_reader_discovered: The app found an available reader we can connect to.
- card_reader_connection_tapped: The user tapped on the button to connect to a reader that was discovered.
- card_reader_connection_failed: The attempt to connect to a card reader failed.
- card_reader_connection_success: The attempt to connect to a card reader was successful.
- card_reader_disconnect_tapped: The user asked to disconnect a card reader.

## To test

An easy way to test the failure states is to temporarily disable Bluetooth on your device. Keep in mind that doing that might leave the app thinking it's still discovering readers when it's not (#4232).

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardReaderService.swift
+++ b/Hardware/Hardware/CardReader/CardReaderService.swift
@@ -5,7 +5,7 @@ public protocol CardReaderService {
 
     // MARK: - Queries
     /// The publisher that emits the list of discovered readers whenever the service discovers a new reader.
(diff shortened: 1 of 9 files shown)


ctarda (approved):
:shipit:

PR #4245, opened 2021-05-20
Shipping Labels: substituted `VStack` with `LazyVStack` in SwiftUI views

pmusolino:
Since we dropped iOS 13, now we can use `LazyVStack` instead of `VStack` on Shipping Labels views implemented in SwiftUI. The three standard stack views, `HStack`, `VStack`, and `ZStack`, all load their contained view hierarchy when they display, and loading large numbers of views all at once can result in slow runtime performance. Using `LazyVStack` the stack is “lazy,” in that the stack view doesn’t create items until it needs to render them onscreen.

I also updated the method `navigationBarTitle` with `navigationTitle` available on iOS 14.

## Testing
Just check that everything in the Shipping Label flow works like before.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/ShippingLabelPackageList.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/ShippingLabelPackageList.swift
@@ -7,8 +7,7 @@ struct ShippingLabelPackageList: View {
 
     var body: some View {
         ScrollView {
-            /// `List` doesn't allow us to change easily some UI things, like the separators. So, we used `VStack`.
-            VStack(spacing: 0) {
+            LazyVStack(spacing: 0) {
 
                 /// Custom Packages
                 ///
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/ShippingLabelPackageList.swift
```diff
@@ -42,7 +41,7 @@ struct ShippingLabelPackageList: View {
             .background(Color(.systemBackground))
         }
         .background(Color(.listBackground))
-        .navigationBarTitle(Text(Localization.title), displayMode: .inline)
+        .navigationTitle(Localization.title)
```
Ecarrion, line 44: [Q] Why this change by mistake? 😅

Ecarrion (approved):
LGTM!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/ShippingLabelPackageList.swift
pmusolino: Because of this `I also updated the method navigationBarTitle with navigationTitle available on iOS 14.`. It's just a new way to declare the bar title in iOS 14. 👍

PR #4246, opened 2021-05-21
[Mobile Payments] Include battery level on reader connection event

koke:
Part of #4082

In #4244 I forgot to include the battery level as a property in `card_reader_connection_success`.

If the reader does not have a battery, or the battery level is unknown, it will be `nil` so we won't set the property.

## To test

Connect a reader, and ensure the battery level is included in the event looking at the console:

> 🔵 Tracked card_reader_connection_success, properties: [AnyHashable("blog_id"): 191646963, AnyHashable("battery_level"): Optional(1.0), AnyHashable("is_wpcom_store"): false]

Update release notes:


ctarda (approved):
Screen Shot 2021-05-21 at 05 45 26

:shipit:

PR #4249, opened 2021-05-21
Removed ShimmerModifier class and included SwiftUI-Shimmer using SPM

### LINKED ISSUE #4248: Include SwiftUI-Shimmer using SPM
As discussed [here](https://github.com/woocommerce/woocommerce-ios/pull/4229#discussion_r636264165) it would be better to include SwiftUI-Shimmer as a separate package using Swift Package Manager, which will include also the MIT License and also informs us that “this is third-party code and we shouldn't touch it”.

pmusolino:
Closes #4248 

## Description
As discussed [here](https://github.com/woocommerce/woocommerce-ios/pull/4229#discussion_r636264165) it's better to include `SwiftUI-Shimmer` as a separate package using Swift Package Manager, which will include also the MIT License and also informs us that “this is third-party code and we shouldn't touch it”. In this PR, I removed our existing class `ShimmerModifier` and I included [SwiftUI-Shimmer](https://github.com/markiv/SwiftUI-Shimmer) using SPM.

## Testing
7. You should see the screen that shows the Package Details screen and all the items that will be shipped.
8. Press the "Package selected" row. 
9. Select a package, then press "Done".
10. Add a Total package weight, press "Done".
11. The "Shipping Carrier and Rates" cell should be enabled now. Press it.
12. You should see the new screen "Carrier and Rates". It should display the ghost cells like before using SwiftUI-Shimmer, then the data fetched from the API.

Update release notes:


Ecarrion (approved):
Everything, including the Codegen package configuration, looks good.

PR #4251, opened 2021-05-21
Update processing orders badge logic

ealeksandrov:
## Description

Previously `MainTabViewModel` used action to get amount of processing orders from network, process callback and display it. Action underneath saved data to cache but never used it. At network level it used action `countOrders` that is a duplicate of `retrieveOrderStatuses`.

This PR updates logic to listen on Core Data cache for `OrderStatus` change (to reuse fresh data from other `retrieveOrderStatuses` calls). It will still make own `retrieveOrderStatuses` requests to keep flow working in exact same way as previously. Unifying these actions will allow smarter caching and data reuse.

Unused actions, networking calls and models will be removed in a [separate PR](https://github.com/woocommerce/woocommerce-ios/pull/4252).

## Test

1. Launch app and observe correct badge number for processing orders
2. Make a new order from the web
3. If you're testing on a simulator (no push) - switch to orders tab to trigger a refresh
4. Check that badge count is updated

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/MainTabViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/MainTabViewModel.swift
@@ -33,8 +33,8 @@ final class MainTabViewModel {
     /// Setup: ResultsController for `processing` OrderStatus updates
     ///
     func configureOrdersStatusesListener(for siteID: Int64) {
-        statusResultsController = updateStatusResultsController(siteID: siteID)
-        configureResultsController()
+        statusResultsController = createStatusResultsController(siteID: siteID)
+        configureStatusResultsController()
     }
 
     /// Callback to be executed when this view model receives new data
@@ -65,14 +65,14 @@ private extension MainTabViewModel {
 
     /// Construct `ResultsController` with `siteID`
     ///
-    func updateStatusResultsController(siteID: Int64) -> ResultsController<StorageOrderStatus> {
+    func createStatusResultsController(siteID: Int64) -> ResultsController<StorageOrderStatus> {
         let predicate = NSPredicate(format: "siteID == %lld AND slug == %@", siteID, OrderStatusEnum.processing.rawValue)
         return ResultsController<StorageOrderStatus>(storageManager: ServiceLocator.storageManager, matching: predicate, sortedBy: [])
     }
 
     /// Connect hooks on `ResultsController` and query cached data
     ///
-    func configureResultsController() {
+    func configureStatusResultsController() {
         statusResultsController?.onDidChangeObject = { [weak self] (updatedOrdersStatus, _, _, _) in
             self?.processBadgeCount(updatedOrdersStatus)
         }
--- a/WooCommerce/Classes/ViewModels/MainTabViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/MainTabViewModel.swift
@@ -20,8 +20,21 @@ final class MainTabViewModel {
 
     private let storesManager: StoresManager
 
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewModels/MainTabViewModel.swift
```diff
@@ -42,39 +55,68 @@ final class MainTabViewModel {
...
 
-        let action = OrderAction.countProcessingOrders(siteID: siteID) { [weak self] orderCount, error in
-            if error != nil {
-                return
+        let action = OrderStatusAction.retrieveOrderStatuses(siteID: siteID) { result in
+            if case let .failure(error) = result {
+                DDLogError("⛔️ Could not successfully fetch order statuses for siteID \(siteID): \(error)")
             }
-
-            self?.processBadgeCount(orderCount)
```
ealeksandrov, line 62: Now we don't process immediate callback, we listen for Core Data updates to get any related changes.
This network request will be throttled later as a part of networking improvements.

WooCommerce/Classes/ViewModels/MainTabViewModel.swift
```diff
@@ -42,39 +55,68 @@ final class MainTabViewModel {
...
         requestBadgeCount()
     }
 }
 
 
 private extension MainTabViewModel {
+
+    /// Construct `ResultsController` with `siteID`
+    ///
+    func updateStatusResultsController(siteID: Int64) -> ResultsController<StorageOrderStatus> {
```
Ecarrion: Since this returns a new object, probably `createStatusResultsController` is a better fit. What do you think?

WooCommerce/Classes/ViewModels/MainTabViewModel.swift
```diff
@@ -42,39 +55,68 @@ final class MainTabViewModel {
...
+        let predicate = NSPredicate(format: "siteID == %lld AND slug == %@", siteID, OrderStatusEnum.processing.rawValue)
+        return ResultsController<StorageOrderStatus>(storageManager: ServiceLocator.storageManager, matching: predicate, sortedBy: [])
+    }
+
+    /// Connect hooks on `ResultsController` and query cached data
+    ///
+    func configureResultsController() {
```
Ecarrion: Another nit, probably `configureStatusResultsController` makes the code easier to read!

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
```diff
@@ -369,6 +369,9 @@ private extension MainTabBarController {
             return
         }
 
+        // Update view model with `siteID` to query correct Orders Status
+        viewModel.configureOrdersStatusesListener(for: siteID)
```
Ecarrion, line 373: Can you explain a little bit why this is needed here? Should we remove the setup on init?

Ecarrion (commented):
Testing ✅ 
Only one question about the configuration of the results controller!

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
ealeksandrov: That's because `MainTabBarController` and its `MainTabViewModel` never deallocate. It is always the same instance when we switch stores (resulting different `siteID`s) and even users.

ealeksandrov:
Nice catch on update/create, @Ecarrion! I've pushed some refactoring and added a comment about configuration.

Ecarrion (approved):
🚀

PR #4253, opened 2021-05-21
[Mobile Payments] Undo CPP merge

ctarda:
Don't mind me, just testing a thing...

Update release notes:


ctarda:
Closing. It looks like we are not going to need this right away, and even if we did, we can recreate it from a checklist

PR #4262, opened 2021-05-24
[Mobile Payments] Ensure amounts in receipts always show two decimal numbers

### LINKED ISSUE #4250: [Mobile Payments] Receipt amounts should include decimals
As was pointed out in https://github.com/woocommerce/woocommerce-android/pull/4070, the current receipt implementation doesn't always show decimals.

I thought we could use `CurrencyFormatter` directly, but then we'd have to move some of the receipt rendering from `Hardware` to `WooCommerce`.

koke:
Fixes #4250 

We now use a number formatter for receipt amounts. Ideally, we'd use the existing `CurrencyFormatter` but that's not available at the Hardware/Yosemite layer where the receipts are generated, so a simpler solution like this will work for now.

If there's a chance that we'll end up generating these in the backend, I don't think it's worth re-architecting things too much for this, considering we'll only support US stores for now.

## To test

Generate payment for an order with the amounts being integers, ensure the receipt shows `.00` for each line item and total amount.

Screen Shot 2021-05-24 at 12 37 18

Update release notes:

### DIFF
--- a/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
+++ b/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
@@ -75,7 +75,7 @@ public extension ReceiptRenderer {
                         <h1>\(receiptTitle)</h1>
                         <h3>\(Localization.amountPaidSectionTitle.uppercased())</h3>
                         <p>
-                            \(parameters.amount / 100) \(parameters.currency.uppercased())
+                            \(formattedAmount()) \(parameters.currency.uppercased())
                         </p>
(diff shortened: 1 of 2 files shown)


ealeksandrov (commented):
Works as described, code looks good 🚀

One question - we also have "Amount paid" section on top with same `/100` logic, have you considered it to be updated in same way?

allendav:
This will be ok for now, but if/when we add support for currencies like Japanese Yen, which have no decimals, we'll need to reconsider

koke:
> we also have "Amount paid" section on top

Good catch, I fixed that one too in 8a3b525dd
Screen Shot 2021-05-25 at 08 47 13

ealeksandrov (approved):
:shipit:

PR #4277, opened 2021-05-25
[Mobile Payments] Observe connected readers correctly when initiating payment

### LINKED ISSUE #4220: [Mobile Payments] Payment fails when connecting directly from orders
To reproduce:

1. Without connecting to a reader first, go to an order that can take payments
2. Tap on Collect Payment. The screen to connect a reader should show up.
3. Tap Connect Card Reader.
4. When the reader is discovered, tap Connect to Reader.
5. There is an error saying the payment failed.

I haven't been able to reproduce 100% of the times, so there might be some specific timing factors here.

A second attempt at collecting payment, once the reader is connected, works without a problem.

https://user-images.githubusercontent.com/8739/118790833-6db33b80-b896-11eb-9715-f4a9237cf323.mp4

### LINKED ISSUE #4224: [Mobile Payments] Connecting a reader from payments leaves the app in an odd state and sometimes crashes
I'm not sure how to explain what's going on, but I recorded a video. When you connect to a reader directly from an Order details, there are a few issues:

1. The list of connected readers with the option to disconnect is briefly visible after connecting, before the modal is dismissed [00:16, 00:48]. Possibly related to #4220 
2. When you go to settings and tap on Disconnect, a new modal (Connect your card reader) is presented, on top of an identical one [00:26] . The button in the presented one does nothing [00:28]. When you dismiss it, the one underneath is functional [00:30]. However, on the first run, connecting to a reader crashed the app.
3.  On the second run, I didn't press the connect button on the presented modal and dismissed it immediately. This didn't crash the app. However, it immediately presented the modal to collect payment [01:03], even though I was in the app settings screen.

https://user-images.githubusercontent.com/8739/118800816-9d674100-b8a0-11eb-988f-c52c7d974db5.mp4

koke:
Fixes #4220
Fixes #4224 

## Why

As mentioned on https://github.com/woocommerce/woocommerce-ios/issues/4224#issuecomment-847713764, the previous implementation was not canceling the subscription to `connectedReaders`, and callbacks for (dis)connection were getting triggered, even if there was no attempt to collect payment in progress

## How

We still don't have a good way to have Yosemite expose publishers, and the UI needs to have access to the subscription so it can cancel it when the modal is dismissed.

For now, I have the view model talk get the `connectedReaders` publisher directly from the card service through `ServiceLocator`. Now `OrderDetailsViewModel.cardReaderAvailable()` exposes a publisher to the view controller that guarantees an initial empty value if there is no reader connected, and a completion when a reader is finally connected.

The View Controller can update the UI accordingly with these two cases, and also cancel the observation if the user dismisses the reader connection modal.

## To test

I added a couple `.print()` operators at the start and end of the operator chain in `OrderDetailsViewModel.cardReaderAvailable()` to verify that the subscription lifecycle was what I expected.

1. Go to an order, tap on Collect Payment, a modal should be presented asking to connect a reader.
2. Dismiss the modal. The subscription to the publisher should be canceled.
3. Tap on Collect Payment, and connect a reader. The "Reader is ready" alert should be visible.
4. Cancel payment.
5. Turn off the reader. Nothing should happen
6. Turn on the reader.
7. Tap on collect payment, and connect a reader. The "Reader is ready" alert should be visible.
8. Tap your card. The payment should complete correctly.
9. Find another order, tap Collect Payment. The "Reader is ready" alert should be visible without prompting to connect a reader.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
@@ -533,7 +533,7 @@ extension OrderDetailsViewModel {
     /// Returns a publisher that emits an initial value if there is no reader connected and completes as soon as a
     /// reader connects.
     func cardReaderAvailable(onCompletion: @escaping (AnyPublisher<[CardReader], Never>) -> Void) {
-        let action = CardPresentPaymentAction.isReadyToCollectPayment { publisher in
+        let action = CardPresentPaymentAction.checkCardReaderConnected { publisher in
             onCompletion(publisher)
         }
 
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
@@ -3,6 +3,7 @@ import UIKit
 import Gridicons
 import Yosemite
 import MessageUI
+import Combine
 
 final class OrderDetailsViewModel {
     private let paymentOrchestrator = PaymentCaptureOrchestrator()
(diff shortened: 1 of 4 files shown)


koke:
@ctarda I moved some of the logic of dealing with callbacks and publishers from the VC to the VM, but I think your suggestion is looking good, and seems to work fine in testing 🎉

ctarda:
@koke Nice! Then let's approve each other! 😉

ctarda (approved):
:shipit: 

On a side note, I like how we iterated on the PR on the fly. We should all pair more...

PR #4279, opened 2021-05-25
Guard against invalid json objects when parsing product add-ons

### LINKED ISSUE #4205: ProductAddOnEnvelope: NSInvalidArgumentException: Invalid type in JSON write (__SwiftValue)
Sentry Issue: [WOOCOMMERCE-IOS-1CZD](https://sentry.io/organizations/a8c/issues/2393713606/?referrer=github_integration)

```
NSInvalidArgumentException: Invalid type in JSON write (__SwiftValue)
  File "<compiler-generated>", in ProductAddOnEnvelope.revolve
  File "Product.swift", line 388, in Product.init
  File "<compiler-generated>", in Product
  File "<compiler-generated>", in ProductListEnvelope.init
  File "<compiler-generated>", in ProductListEnvelope.init
...
(56 additional frame(s) were not displayed)
```
---

Created by @shiki

As of May 14. Events: 6, users: 2

Ecarrion:
closes #4205 

# Why

There appears to be some crashes -WOOCOMMERCE-IOS-1CZD- (Only happening to 2 users so far) when parsing add-ons from the product metadata.

I couldn't replicate the crash, but the crash log indicates that it happens on some kind of json write
``` 
Invalid type in JSON write (__SwiftValue) - ProductAddOnEnvelope.revolve
``` 

My guess is that the call where we map a product-metadada json object to a `ProductAddOn` entity is not handling well a particularly mal-formed JSON object and crashing without proper error handling.
```swift
try JSONSerialization.data(withJSONObject: addOnJsonObject, options: .fragmentsAllowed)
```

# How

My solution consists in asking `JSONSerialization` if the JSON object is valid and throwing a custom error if it isn't.

```swift
guard JSONSerialization.isValidJSONObject(addOnJsonObject) else {
     throw ProductAddOnEnvelopeError.invalidJsonObject(addOnJsonObject)
}
```

This error is then logged onto tracks where we can monitor for the malformed JSON.
https://github.com/woocommerce/woocommerce-ios/blob/3115f4c5e1a05c646f957e594cdc75a345de1a88/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift#L724-L727

# Testing Steps
None, as we can't reproduce the error.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/ProductAddOnEnvelope.swift
+++ b/Networking/Networking/Model/Product/ProductAddOnEnvelope.swift
@@ -71,7 +71,21 @@ internal struct ProductAddOnEnvelope: Decodable {
     /// Converts an addOnJsonObject(`Dictionary`) to a `ProductAddOn` entity.
     ///
     private func decode(addOnJsonObject: AnyDictionary, using decoder: JSONDecoder) throws -> ProductAddOn {
+        // It appears to be unexpected crashes when parsing some JSON objects that are not handled inside subsequent `try JSONSerialization.data` call.
+        // https://github.com/woocommerce/woocommerce-ios/issues/4205
+        guard JSONSerialization.isValidJSONObject(addOnJsonObject) else {
+            throw ProductAddOnEnvelopeError.invalidJsonObject(addOnJsonObject)
+        }
+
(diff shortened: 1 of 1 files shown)


ctarda (approved):
I can’t reproduce either, but the approach seems solid to me. :shipit:

AliSoftware:
This just got included in new beta 6.7.0.1

PR #4289, opened 2021-05-26
[Mobile Payments] Show an alert if the app can't use Bluetooth

### LINKED ISSUE #4232: [Mobile Payments] Reader discovery fails silently if Bluetooth permission is denied

koke:
Fixes #4232

Currently, tapping on Connect reader will attempt to start discovery even if the user has declined the Bluetooth permission. The stripe SDK would return an error, but we weren't handling errors at all in discovery.

This PR introduces a new alert to show any errors related to discovery, although it will only show a detailed message for Bluetooth permission errors.

## How

Initially I tried to have the service call `switchStatusToFault` and pretend there was an error coming from the API. However, the permission check is synchronous, so doing that would send a failed completion to `discoveredReaders` before the store had subscribed to it, and it would re-initialize it, so when the store subscribed, it would have the initial empty value.

Instead of trying to fit this into the Combine flow, I made the call to `start` throw an error.

I also took the opportunity to simplify a couple things for state management in `CardReaderSettingsUnknownViewModel`.

## To test

1. Go to System Settings > Woo, and disable Bluetooth
2. Back in the Woo app, try to connect to a reader
3. An alert showing an error should display:

Any other error would also display the alert without the explanation. For instance, with the reader powered off and bluetooth available, start reader discovery, then enable airplane mode:

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -1,5 +1,6 @@
 import Combine
 import StripeTerminal
+import CoreBluetooth
 
 /// The adapter wrapping the Stripe Terminal SDK
 public final class StripeCardReaderService: NSObject {
@@ -62,13 +63,17 @@ extension StripeCardReaderService: CardReaderService {
 
     // MARK: - CardReaderService conformance. Commands
 
-    public func start(_ configProvider: CardReaderConfigProvider) {
+    public func start(_ configProvider: CardReaderConfigProvider) throws {
         setConfigProvider(configProvider)
 
         let config = DiscoveryConfiguration(
             discoveryMethod: .bluetoothProximity,
             simulated: false
         )
+
+        guard CBCentralManager.authorization != .denied else {
+            throw CardReaderServiceError.bluetoothDenied
+        }
 
         switchStatusToDiscovering()
 
--- a/Hardware/Hardware/CardReader/CardReaderService.swift
+++ b/Hardware/Hardware/CardReader/CardReaderService.swift
@@ -29,7 +29,7 @@ public protocol CardReaderService {
 
     /// Starts the service.
     /// That could imply, for example, that the reader discovery process starts
-    func start(_ configProvider: CardReaderConfigProvider)
+    func start(_ configProvider: CardReaderConfigProvider) throws
 
     /// Cancels the discovery process.
     func cancelDiscovery() -> Future <Void, Error>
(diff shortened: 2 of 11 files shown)


Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -62,14 +63,18 @@ extension StripeCardReaderService: CardReaderService {
...
-    public func start(_ configProvider: CardReaderConfigProvider) {
+    public func start(_ configProvider: CardReaderConfigProvider) throws {
         setConfigProvider(configProvider)
 
         let config = DiscoveryConfiguration(
             discoveryMethod: .bluetoothProximity,
             simulated: false
         )
 
+        guard CBCentralManager.authorization != .denied else {
```
koke, line 74: There is also a `.restricted` case which is interesting. The docs say:

> In this state, the user can’t change the Bluetooth authorization status, possibly due to active restrictions such as parental controls

I'm not handling this currently since I imagine it's unlikely merchants will encounter that, and I understand that the person using the app might not be able to do anything about it.

ctarda (approved):
This is fantastic @koke  Thanks!

I see the alert going off after removing permissions:


Also in airplane mode


And after re-enabling permissions:


:shipit:

ctarda:
Merging on behalf of the author to trigger a new installable build of the PR into develop

allendav:
Sorry to be late to this, but could we replace, even in a future issue/PR

Bluetooth permission was denied

with either

Go to Settings > Bluetooth and Enable Bluetooth (in the event system wide Bluetooth has been disabled)

or

Go to Settings > Privacy > Bluetooth to Enable Bluetooth for WooCommerce (in the event this application specifically is not permitted)

See also https://developer-mdn.apple.com/design/human-interface-guidelines/ios/app-architecture/settings/ which recommends a button to take the user to the appropriate place.

> Provide shortcuts to Settings when appropriate. If your app includes text that directs users to Settings, such as “Go to Settings > MyApp > Privacy > Location Services,” provide a button that opens that location automatically.

koke:
I think that makes sense, the modal didn't seem to accommodate a longer message, but with #4292 fixing the text clipping maybe there's a way to make it better

PR #4292, opened 2021-05-26
[Mobile Payments] Fix text being clipped in modal alerts

ctarda:
Closes #4176

| Before | After |
| ---- | ---- |
|  |  |
|  |  |

## Changes
* Change the modal alert `distribution` on rotation.

## How to test

Testing requires:

1. Setting up a store for Card Present Payment Testing (also P91TBi-4BH-p2 for more specific instructions)
2. WCPay 2.4.0

* Checkout the branch, build and run on device.
* Attempt to connect to a reader. Check the modal alert presents text, while discovering, and the text is not trimmed. 
* Rotate the device

Update release notes:
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #4222, opened 2021-05-19
[Mobile Payments] Button label truncated when payment fails


koke:
When a payment fails, the text in the modal button is truncated:

![Screen Shot 2021-05-19 at 11 40 48](https://user-images.githubusercontent.com/8739/118791925-75271480-b897-11eb-8221-fccf64749ce6.png)

To reproduce, try to make a test purchase with an amount ending in `.75` (or any of the other [error decimals](https://stripe.com/docs/terminal/testing#physical-test-card)).

allendav:
The first part of the fix is to change from truncate in the middle to word wrap, and I poked at this a bit today with that but the NUXButton really doesn't look good with multiple lines of text (text gets close to button edge). It seems to be overriding the insets or constraining the height or both.

I'll keep trying

<img src="https://user-images.githubusercontent.com/1595739/119205290-bb71a480-ba4c-11eb-9eea-a6f8da0c995c.PNG" width=50% />

PR #4302, opened 2021-05-27
[Mobile Payments] Improve Bluetooth permission error

koke:
Part of #4232 
A continuation of #4289 

As suggested by @allendav in https://github.com/woocommerce/woocommerce-ios/pull/4289#issuecomment-849094781, this improves the messaging around Bluetooth permissions a bit, adding an explanation and a direct action to open device settings.

The primary button shows clipped text in Spanish, but we'll handle that in #4222
Spanish | English
-|-
Screen Shot 2021-05-27 at 10 01 14 | Screen Shot 2021-05-27 at 10 04 03

## To test

I tested this applying the fix in #4292 to see the full message, but I'm not including that code in this PR.

1. Go to System Settings > Woo, and disable Bluetooth
2. Back in the Woo app, try to connect to a reader
3. An alert showing an error should display.
4. Tapping dismiss should close the alert. Tapping Open Device Settings should take you to the Woo iOS settings where you can enable bluetooth

It seems that when you toggle the Bluetooth permission setting and come back to the app, iOS kills the app and relaunches it. I would have tried to detect the state when the app becomes active again, but there's no point in doing that if it will get killed. It doesn't seem like we support UIKit's state preservation either?

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalScanningFailed.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalScanningFailed.swift
@@ -10,18 +10,18 @@ final class CardPresentModalScanningFailed: CardPresentPaymentsModalViewModel {
     /// A closure to execute when the primary button is tapped
     private let primaryAction: () -> Void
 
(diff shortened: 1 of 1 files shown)


ctarda (approved):
:shipit:

PR #4318, opened 2021-05-28
Add-ons: Allow product encoding to continue after add-ons decoding fail

Ecarrion:
# Why

#4279 fixed a crash when parsing add-ons with malformed metadata, but I just realized that while that prevents the crash, it also prevents the products list to successfully continue its decoding, which results in an empty screen for the user.

Since add-ons is an optional plugin and the feature is behind a beta toggle, this PR makes sure that even if the add-on encoding fails, the product encoding will continue.

# Testing steps
- Run the app and see that the product list is displayed normally.

Update release notes:


rachelmcr (approved):
LGTM! :shipit:

Issue #4269, opened 2021-05-24
Add-ons: Populate product add-on list view


Ecarrion:
With #4267 and #4268 ready, we should render all of the product available add-ons on the product add-on list view.

Issue #4267, opened 2021-05-24
Add-ons: Create product add-on list view


Ecarrion:
Create a simple view to hold all of the product's add-ons views.

The specific add-on views will be done in a separate task.

<img width="402" alt="add-on-list" src="https://user-images.githubusercontent.com/562080/119417509-8cad3580-bcbb-11eb-848d-5902ae710369.png">

Tasks:
- [x] Create main view
- [x] Create template view model
- [x] Disclaimer at the bottom not needed.

PR #4322, opened 2021-05-28
Add-ons: Add "Add-on" row in product detail screen

### LINKED ISSUE #4266: Add-ons: Add "Product Add-ons" button
In the product detail screen, we need to show a "Product Add-ons" button like:

main-button

Considerations:
- Button should be under a feature flag.
- Button should only be visible if the product has add-ons.

Ecarrion:
closes #4266 

# Why

The final goal is to give the ability to the merchant to see what add-ons a product can have.

This PR adds a "Product Add-ons" row to the product screen in order for the merchant to initiate that flow(to be implemented later #4267 #4269)

Note:
 - The row will be visible when the product has add-ons and when the "add-on beta feature switch" has been enabled.

# How

- Update `DefaultProductFormTableViewModel` to provide a view model for the specific `add-on` row.

- Update `ProductFormActionsFactory` to include the `add-on` row when the product has add-ons and the feature is enabled.

- Update `ProductFormViewModel` to fetch the `add-on` beta feature switch state and recreate the form rows if necessary.

# Screenshots

No add-ons | Add-ons disabled | Add-ons enabled
--- | --- | ---
no-addons | add-ons-disabled | add-ons

# Testing

**Prerequisites:** 
- Have your site with the add-ons plugin installed and with at least one product with add-ons.
- Enable the "Add-Ons" feature from the beta features screen in settings.

**Steps:**
- Launch the app and navigate to a product with add-ons
- See a "Product Add-ons" row

Update release notes:


pmusolino (approved):
Looks good to me :shipit: 

I just noticed a thing not specifically related to this PR. The "View Add-Ons" switch under the Experimental Features menu is enabled also if the feature flag `addOnsI1` is set to `false`. I think that it means that our users will be able to see this experimental feature menu also if the feature is still a work in progress. In that case, I think we should show the experimental feature just if the `addOnsI1` is enabled.

Ecarrion:
> I just noticed a thing not specifically related to this PR. The "View Add-Ons" switch under the Experimental Features menu is enabled also if the feature flag addOnsI1 is set to false. 

That's ok. We have already shipped viewing add-ons on an order detail. That's why the beta-feature switch is available to everyone.

And this new feature(add-ons on product form) is being developed under `addOnsI1` and under the beta-feature switch.

Issue #4333, opened 2021-05-31
[Mobile Payments] Add support for "Several Readers Found"


ctarda:
Even though we have UI that suggest that we would be able to handle multiple card readers, there is no logic to do so.

This comment in line 100 of CardReaderSettingsUnknownViewModel tells the story:
`        /// TODO: Add another viewmodel and view to handle multiple discovered readers.`

ctarda:
@allendav  could you please prioritize this over your other tasks?

ctarda:
This is pretty much what is missing.

<img width="203" alt="Screen Shot 2021-05-31 at 17 24 36" src="https://user-images.githubusercontent.com/2722505/120243266-25503200-c235-11eb-98eb-e83cfca2366b.png">

ctarda:
@allendav  let's see what comes out of this discussion first: p91TBi-5fB-p2

ctarda:
@allendav crisis adverted for now. p91TBi-5fB#comment-4849-p2 I will move this to the post-beta 1 backlog for now

allendav:
@ctarda sounds good - I also added my thoughts in comment 4854

allendav:
@ctarda 

I've looked at what I think it would take to do this and I am proposing adding a new modal to CardReaderSettingsAlerts that presents the list. This will also require changes to the CardPresentPaymentsModalViewModel protocol to also include an optional array of tappable items (found readers) and CardReaderConnectionController to use that modal instead of the "found one reader" modal. We'll also have to stop using bluetoothProximity and switch to bluetoothScan in order to see multiple readers.

I'll wrap it all in a feature flag and implement it over the course of roughly 5 PRs:

Testing will require ensuring this works properly both in the settings context but also in the collect-payment context.

- PR 1
    - Feature flag
- PR 2
    - change to bluetoothscan
    - change CardReaderConnectionController foundReader to array
    - update CardReaderConnectionController onConnectToReader to connect to the selected reader
- PR 3
    - extend CardPresentPaymentsModalViewModel protocol to also include optional tableviewcell array
    - Add foundMultipleReaders to CardReaderSettingsAlertsProvider protocol
    - Add foundMutlipleReaders to CardReaderSettingsAlerts
    - add CardPresentModalFoundMultipleReaders conforming to CardPresentPaymentsModalViewModel
    - Add foundMultipleReaders to CardReaderSettingsAlertsProvider protocol
[...]

PR #4334, opened 2021-05-31
[Mobile Payments] Fix Tapping "Keep Searching" when discovering readers dismisses the modal and nothing else happens

### LINKED ISSUE #4330: [Mobile Payments] Tapping "Keep Searching" when discovering readers dismisses the modal and nothing else happens
1. Navigate to Settings > Manage Card Readers
2. Tap Connect Card Reader
3. When a reader is found, tap Keep Searching.
4. The alert is dismissed, nothing else happens, and the reader gets into a state where it needs restarting
5. Tapping Connect Card Reader again dos nothing.

ctarda:
Closes #4330 

This is an issue I found today while taking screenshots for the UI review.

In a nutshell, the "Keep Searching" functionality when discovering readers does not work, partially because a UI glitch, and partially because it is not fully implemented.

This PR solves the UI side of it. #4333 should take care of the rest.

https://user-images.githubusercontent.com/2722505/120240903-b6240f00-c22f-11eb-8f59-d870f1dc5195.MP4

## Changes
* Call completion block on primary block after dismissal has been completed.
* Do not dismiss on secondary button. That makes possible to launch the discovery UI again.

## How to test
* Build and run on device.
* Navigate to settings > Manage Card Readers
* Tap Connect Card Reader
* When a reader is found, tap "Keep Searching"
* A new modal "searching readers would be presented". This is the correct UI, but it is not actually searching for other readers (that's what #4333 should solve)
* Tap Cancel in the modal.
* Tap Connect Card Reader again.
* When a new card reader is discovered, tap "Connect" and it should work

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalFoundReader.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalFoundReader.swift
@@ -35,13 +35,13 @@ final class CardPresentModalFoundReader: CardPresentPaymentsModalViewModel {
     }
 
     func didTapPrimaryButton(in viewController: UIViewController?) {
-        connectAction()
-        viewController?.dismiss(animated: true, completion: nil)
+        viewController?.dismiss(animated: true, completion: { [weak self] in
+            self?.connectAction()
+        })
     }
 
     func didTapSecondaryButton(in viewController: UIViewController?) {
         continueSearchAction()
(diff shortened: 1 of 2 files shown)


koke (approved):
That seems to work nicely. I don't have a second reader but what I tried:
1. Start connecting
2. Turn on the reader, wait until discovered
3. Tap Keep Searching
4. Turn off the reader, then turn it back on. Reader is discovered again
5. Tap connect

PR #4335, opened 2021-06-01
[Mobile Payments] Only show the option to email a receipt if the device can send email

### LINKED ISSUE #4225: [Mobile Payments] Don't show option to email if we can't send it
Currently, we show the option to email a receipt, even if mail isn't configured on the device. We do check `MFMailComposeViewController.canSendMail()`, but only when the button is pressed, so it's visible but it does nothing.

To reproduce, you can try on the simulator, or go to System Settings > Mail, and disable all of your mail accounts before testing.

koke:
Fixes #4225 

This will check whether the device can send mail before showing the option to email a receipt, which would silently fail before.

Instead of adding more complicated logic to `CardPresentModalSuccess`, I've created a new `CardPresentModalSuccessWithoutEmail` with two buttons.

## To test

The easiest way to test this is to go to System Settings > Mail, and temporarily disable all your email accounts. Then try to collect a payment. At the end of the successful payment, it should look like this:

With email available | With no email
-|-
Screen Shot 2021-06-01 at 12 02 53|Screen Shot 2021-06-01 at 12 02 17

The "No thanks" button should dismiss the modal in both cases, and the "Email receipt" should show the email composer with the receipt:

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalSuccessWithoutEmail.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalSuccessWithoutEmail.swift
@@ -1,0 +1,68 @@
+import UIKit
+
+/// Modal presented when the payment has been collected successfully
+final class CardPresentModalSuccessWithoutEmail: CardPresentPaymentsModalViewModel {
+
+    /// Amount charged
+    private let amount: String
+
+    /// Closure to execute when primary button is tapped
+    private let printReceiptAction: () -> Void
+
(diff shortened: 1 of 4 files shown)


ctarda (approved):
Works great, and it's pretty clean. Totally optional, and definitely not a blocker, we have unit tests for other view models backing the CPP modals. Not like there is much to test, but those can cover that the properties that are meant to be nil are actually nil 🤷🏼 

:shipit:

koke:
Good call about the tests. I originally went to update the one for `CardPresentModalSuccess` but when I changed course I forgot to write a new test case. I'll add those before merging

PR #4339, opened 2021-06-01
[Mobile Payments] Restart reader discovery when tapping "Keep Searching"

### LINKED ISSUE #4338: [Mobile Payments] Keep searching should restart the search
When a reader is found, we show a modal popover with two buttons: Connect and Keep Searching

At this moment, the Keep Searching button does not do anything.

As discussed here: p91TBi-5fB-p2#comment-4849

for now, we will restart the search, and leave showing a list of other readers for a later stage

ctarda:
Closes #4338

This is the continuation to #4334. If that PR took care of the UI side of the issue with the "Keep Searching" button, this PR, after discussing the scope for the "Keep Searching" button for beta 1 (p91TBi-5fB-p2#comment-4849) introduces the following change:

## Changes
* Keep Searching now triggers a restart of the reader discovery process. That means that it first cancel the previous discovery, and then starts a new one. Which has turned out to be fun because there seems to be something a bit off in the Terminal SDK hen it comes to discovery cancellation: the completion block that signals that cancellation is completed seems to be executed before the SDK has actually transitioned to an idle state.
* I added a good old "dispatch to the next cycle" to what happens when the Terminal SDK executes the completion block for the cancellation of the discovery process. There might be a more Combine-est way to do that, but I have no shame 🤷🏼 

## How to test
This one is tricky because I have only one reader. I assume that if it works with one reader, it does not matter which reader it is.

Testing requires:

2. WCPay 2.4.0

It should not matter if this is tested on device or on the simulator, but it would be better to test on device and with an actual reader.

1. Navigate to Settings > Manage Card Readers
2. Tap Connect to Card Reader
3. When a reader is discovered, tap Keep Searching. 
4. The modal UI should transition to the "searching readers" state
5. After a while, the app should find a new reader
6. At this point, tap either Connect (and the app should connect to the reader, and it should be possible to capture a payment) or Keep Searching again, which would take you back to step 4
7. At any moment in the process, feel free to tap Cancel and start again. Things should still work
8. Also, once a connection to a reader is completed, try disconnecting from it and going back to step 2.

Fun times.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -112,20 +112,24 @@ extension StripeCardReaderService: CardReaderService {
                 return promise(.success(()))
             }
 
-            // This completion block, apparently, is called when
-            // the SDK has not really transitioned to an idle state
+            // The completion block for cancel, apparently, is called when
+            // the SDK has not really transitioned to an idle state.
             // Clients might need to dispatch operations that rely on this completion block
             // to start a second operation on the card reader.
             // (for example, starting an operation after discovery has been cancelled)
             //
             self?.discoveryCancellable?.cancel { [weak self] error in
-                guard let error = error else {
-                    self?.switchStatusToIdle()
-                    return promise(.success(()))
-                }
+                // Horrible, terrible workaround.
+                // And yet, it is the classic "dispatch to the next run cycle".
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
+                    guard let error = error else {
+                        self?.switchStatusToIdle()
+                        return promise(.success(()))
+                    }
 
-                self?.internalError(error)
-                promise(.failure(error))
+                    self?.internalError(error)
+                    promise(.failure(error))
+                }
             }
         }
     }
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -121,7 +121,7 @@ extension StripeCardReaderService: CardReaderService {
             self?.discoveryCancellable?.cancel { [weak self] error in
                 // Horrible, terrible workaround.
                 // And yet, it is the classic "dispatch to the next run cycle".
-                DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
+                DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
                     guard let error = error else {
                         self?.switchStatusToIdle()
                         return promise(.success(()))
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -112,11 +112,18 @@ extension StripeCardReaderService: CardReaderService {
                 return promise(.success(()))
             }
 
+            // This completion block, apparently, is called when
+            // the SDK has not really transitioned to an idle state
+            // Clients might need to dispatch operations that rely on this completion block
+            // to start a second operation on the card reader.
+            // (for example, starting an operation after discovery has been cancelled)
+            //
             self?.discoveryCancellable?.cancel { [weak self] error in
                 guard let error = error else {
                     self?.switchStatusToIdle()
                     return promise(.success(()))
                 }
+
                 self?.internalError(error)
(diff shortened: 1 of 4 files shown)


Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -112,13 +112,24 @@ extension StripeCardReaderService: CardReaderService {
...
+            // to start a second operation on the card reader.
+            // (for example, starting an operation after discovery has been cancelled)
+            //
             self?.discoveryCancellable?.cancel { [weak self] error in
-                guard let error = error else {
-                    self?.switchStatusToIdle()
-                    return promise(.success(()))
+                // Horrible, terrible workaround.
+                // And yet, it is the classic "dispatch to the next run cycle".
+                DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
```
koke: Can we maybe confirm with Stripe if there's a better way to do this?
I enabled Stripe's verbose logging and compared it to our calls, and there's already just 120ms between the cancelable's callback, and what might be the actual internal state update. This is in a relatively dedicated testing device with nothing else running, so I wouldn't be surprised if a more real-world device crosses the 200ms boundary. If we can't find a better way, I would give it a bit more to be safe, maybe one second.

```
2021-06-02 10:18:30.603202+0200 WooCommerce[16043:8119290] 💳 cancelDiscovery called
2021-06-02 10:18:30.603648+0200 WooCommerce[16043:8119290] 💳 cancelDiscovery future called
2021-06-02 10:18:30.610444+0200 WooCommerce[16043:8120605] 💳 l app_id=com.automattic.woocommerce command=discoverReaders completed=0 event=cancel logpoint_level=info scope=SCPCancelable sdk_version=1.4.0 time=1622621910605
2021-06-02 10:18:30.614259+0200 WooCommerce[16043:8120605] 💳 l app_id=com.automattic.woocommerce event=stopBTScan logpoint_level=info scope=SCPBbposBluetoothProximityAdapter sdk_version=1.4.0 time=1622621910610
2021-06-02 10:18:30.618669+0200 WooCommerce[16043:8119290] 💳 discoveryCancellable cancel callback
2021-06-02 10:18:30.618951+0200 WooCommerce[16043:8120605] 💳 l app_id=com.automattic.woocommerce command=discoverReaders event=cancelationSucceeded logpoint_level=info scope=SCPCancelable sdk_version=1.4.0 time=1622621910615
2021-06-02 10:18:30.732585+0200 WooCommerce[16043:8120598] 💳 l app_id=com.automattic.woocommerce command_in_progress=none event=state_change logpoint_level=info scope=SCPTerminal sdk_version=1.4.0 time=1622621910726
2021-06-02 10:18:30.736947+0200 WooCommerce[16043:8120598] 💳 e app_id=com.automattic.woocommerce event=discoverReaders logpoint_level=info scope=SCPTerminal sdk_version=1.4.0 time=1622621910733 total_time_ms=6390
2021-06-02 10:18:30.739693+0200 WooCommerce[16043:8120598] 💳 x domain=terminal_request event=discoverreaders outcome=success scope=scpterminal time=1622621910738
2021-06-02 10:18:30.835547+0200 WooCommerce[16043:8119290] 💳 discoveryCancellable cancel callback after 0.2 second
```

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewModel.swift
```diff
@@ -7,6 +7,7 @@ enum CardReaderSettingsUnknownViewModelDiscoveryState {
     case failed(Error)
     case foundReader
     case connectingToReader
+    case cancellingSearch
```
koke: I'm thinking this could be confusing in the future, since you won't go into this state when you actually tap cancel. How about `restartingSearch` instead?

koke (approved):
I left a couple of suggestions, but I think this works fine 👍🏽

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
ctarda: Bumped the delay to 1 second

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewModel.swift
ctarda: Done!

PR #4326, opened 2021-05-31
[Mobile Payments] Prevent clipping amount in errors on landscape

koke:
This allows the amount in payment errors to reduce its font size so it fits on the space available when the phone is on landscape mode.

## To test

Attempt a payment with an amount ending in `.75`, then rotate your phone to landscape when the payment fails.

Before:
Screen Shot 2021-05-31 at 12 13 25

After:
Screen Shot 2021-05-31 at 12 17 14

The button clipping is being tracked separately in #4222 

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4340, opened 2021-06-02
[Mobile Payments] Scale down button labels when text doesn't fit

### LINKED ISSUE #4222: [Mobile Payments] Button label truncated when payment fails

koke:
Fixes #4222 

Similar to #4326, this allows the buttons in modals to reduce the font size by 50% at most so that the text can fit. I didn't see a way to do this for buttons directly from Interface Builder, so I did it with code.

## To test

To test this, I modified one of the modals to have some sample long text in the buttons. Then also tested the original payment failed modal in #4222 .

Update release notes:


ctarda (approved):
It looks great! 

:shipit:

PR #4342, opened 2021-06-02
Shipping Labels: populated the total weight automatically in packages step

### LINKED ISSUE #4090: Shipping Labels: populate the weight automatically in packages step
As @anitaa1990 mentioned in #[3810](https://github.com/woocommerce/woocommerce-android/pull/3810), when creating shipping labels, wp-admin pre-populate the weight field if the products' weight is defined, we should do the same.

pmusolino:
Fixes #4090 by fetching and populating the weight of the package for every item.

## Testing
1. Modify products and product variations by specifying different weights.
2. Create an order using these products, and make sure it satisfies the SL creation conditions.
3. Open the order in the app.
4. Click on Create Shipping Label.
5. Pass the origin and shipping addresses steps.
6. Open the package details screen.
7. Confirm that the shown "Total package weight" is the correct one.

## Screenshot

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -104,15 +104,18 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
             guard let self = self else { return }
             self.products = products
             self.itemsRows = self.generateItemsRows()
+            self.setTotalWeight()
         }, onProductVariationsReload: { [weak self] (productVariations) in
             guard let self = self else { return }
             self.productVariations = productVariations
             self.itemsRows = self.generateItemsRows()
+            self.setTotalWeight()
         })
 
         products = resultsControllers?.products ?? []
         productVariations = resultsControllers?.productVariations ?? []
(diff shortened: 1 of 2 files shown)


rachelmcr (commented):
This looks good and tests well with different products (regular product with weight, variation with weight, product without a set weight) as long as I follow the expected user flow. However, I found a few cases that we might want to consider:

1. When none of the items in the order have a set weight, or the user manually enters `0` as the weight, the Done button is still enabled and I can save those package details. However, in WP Admin the total weight field doesn't allow this:
Screen Shot 2021-06-03 at 11 21 57 AM

2. If I manually update the total weight, and tap Done, I see that the weight I entered is saved. However, if I tap the Package Details cell again we use `setTotalWeight()` to recalculate the weight again, so the previous change I made to the weight isn't displayed. That doesn't seem like a common set of steps for someone to take, but do you think we should consider it?


3. This isn't specific to the changes in this PR, but I noticed that we don't validate the weight entered in the `Total package weight` field. I can manually enter an invalid value such as `1..5` or text like `zero` and it's accepted.

pmusolino:
Thank you @rachelmcr!
I fixed the case 1 and the case 3 in this commit: 52d063362dcb7858e7a6f77b8976e5dd13d1ce23
And because on the web when the panel will be closed, they don't re-calculate the total weight if it's already set, I updated our logic: a9c5657d85506f8be69b0fb2acfc0299b7dd9ff0

Ready for another pass! 👍

rachelmcr (approved):
Code looks good and it tests well. Thanks for making those changes! :shipit:

Issue #4313, opened 2021-05-27
Variations: Changes when creating a new variation


Ecarrion:
We are going to be doing some changes when creating the first variation.

<img width="1173" alt="creating-first-variation" src="https://user-images.githubusercontent.com/562080/119867226-46d2b600-bee3-11eb-94a6-b5dacfe29a8d.png">

## Task

- [x] After creating the first variation, redirect the user to the main product screen
- [x] After creating the first variation, present a notice informing the user that a variation was created.
- [x] If some variations don't have prices, show the "add price to your variations" cell indicator

PR #4347, opened 2021-06-02
Variation: Show "No Price" warning on product screen

Ecarrion:
part of #4313 

Why:

As part of the Variation changes, we are introducing a no-price warning when a variable product has variations without price.

# How

- Update `ProductFormActionsFactory` to decide when the `noPriceInfoWarning` row should be visible.

- Update `ProductFormViewModel` to observe the product variations and check if any of them is missing the price.

- Update `ProductFormViewModel` to provide and observable when a variation price changes. This will be consumed by `ProductFormViewController` in order to reload its table.

- Update relevant tests

# Demo
https://user-images.githubusercontent.com/562080/120552788-4e033380-c3bd-11eb-9a9b-2939e32d27f4.mov

# Testing steps

- Launch the app and navigate to a variable product
- Navigate to one of its variations, set its price to "" and save.
- Navigate back to the main product and see the "warning" row.
- Navigate that variation, set its price to some valid value.
- Navigate back to the main product and see no "warning" row.

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormActionsFactory.swift
```diff
@@ -157,9 +168,13 @@ private extension ProductFormActionsFactory {
         let shouldShowReviewsRow = product.reviewsAllowed
         let canEditProductType = formType != .add && editable
         let canEditInventorySettingsRow = editable && product.hasIntegerStockQuantity
+        let variationsHaveNoPriceSet = variationsPrice == .notSet
+        let productHasNoPriceSet = variationsPrice == .unknown && product.product.variations.isNotEmpty && product.product.price.isEmpty
+        let shouldShowNoPriceWarningRow = canEditProductType && (variationsHaveNoPriceSet || productHasNoPriceSet)
```
shiki: ⚠️  Would it be okay if we enclose this in a block so it's clearer which variables are used where? 

```suggestion
        let shouldShowNoPriceWarningRow: Bool = {
            let variationsHaveNoPriceSet = variationsPrice == .notSet
            let productHasNoPriceSet = variationsPrice == .unknown && product.product.variations.isNotEmpty && product.product.price.isEmpty
            return canEditProductType && (variationsHaveNoPriceSet || productHasNoPriceSet)
        }()
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductVariationFormViewModel.swift
```diff
@@ -26,6 +26,12 @@ final class ProductVariationFormViewModel: ProductFormViewModelProtocol {
         isUpdateEnabledSubject
     }
 
+    /// Emits a void value informing when there is a new variation price state available
+    var newVariationsPrice: Observable<Void> {
+        // No-op
+        PublishSubject<Void>()
+    }
```
shiki: Looks like this can also just be 

```suggestion
    var newVariationsPrice: Observable<Void> = PublishSubject<Void>()
```

I don't think we'll save a lot of performance if we create it lazily.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -501,28 +504,39 @@ private extension ProductFormViewController {
...
-            self?.onEditProductNameCompletion(newName: name ?? "")
-        }, onStatusChange: { [weak self] isEnabled in
-            self?.onEditStatusCompletion(isEnabled: isEnabled)
-        }, onAddImage: { [weak self] in
-            self?.eventLogger.logImageTapped()
-            self?.showProductImages()
-        })
-        tableView.dataSource = tableViewDataSource
-        tableView.reloadData()
+        reconfigureDataSource(tableViewModel: tableViewModel, statuses: productImageActionHandler.productImageStatuses)
```
shiki, line 521: Not sure if this would be any better. Since `tableViewModel` and `productImageActionHandler` are instance variables, maybe those can be accessed by `reconfigureDataSource()` directly.

```suggestion
        reconfigureDataSource()
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -531,6 +545,7 @@ private extension ProductFormViewController {
         }, onStatusChange: { [weak self] isEnabled in
             self?.onEditStatusCompletion(isEnabled: isEnabled)
         }, onAddImage: { [weak self] in
+            self?.eventLogger.logImageTapped()
```
shiki, line 548: Good find!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -20,6 +22,11 @@ final class ProductFormViewModel: ProductFormViewModelProtocol {
         isUpdateEnabledSubject
     }
 
+    /// Emits a void value informing when there is a new variation price state available
+    var newVariationsPrice: Observable<Void> {
+        newVariationsPriceSubject
+    }
```
shiki, line 28: Do you think we should change this to use Combine instead? The `Observable` is deprecated. Just curious. 😄

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -432,6 +452,27 @@ extension ProductFormViewModel {
 
         formType = .edit
     }
+
```
shiki: Note that this is in a “Remote Actions” section. We might want to move these somewhere else.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -501,28 +504,39 @@ private extension ProductFormViewController {
         }
     }
 
+    /// Updates table rows when the price of the underlying variations change.
+    /// Needed to show/hide the `.noPrinceWarning` row.
```
shiki, line 508: Thanks for the useful comment!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -452,6 +493,26 @@ private extension ProductFormViewModel {
...
+        newVariationsPriceSubject.send(())
+    }
+
+    /// Calculates the variations price state for the current fetched variations.
+    ///
+    func calculateVariationPriceState() -> ProductFormActionsFactory.VariationsPrice {
+        // If there are no fetched variations we can't be sure of it's price state
+        guard variationsResultsController.fetchedObjects.isNotEmpty else {
+            return .unknown
+        }
```
shiki: ⚠️ It's more performant to use `isEmpty` because the `fetchedObjects` runs the mapping to readonly objects which cause more Core Data fetches. 

https://github.com/woocommerce/woocommerce-ios/blob/7b74933a8f4719e2d459e46dd9d8d3e475f1618e/Yosemite/Yosemite/Tools/ResultsController.swift#L184-L192

```suggestion
        guard !variationsResultsController.isEmpty else {
            return .unknown
        }
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
```diff
@@ -452,6 +493,26 @@ private extension ProductFormViewModel {
...
+
+        let someMissingPrice = variationsResultsController.fetchedObjects.contains { $0.regularPrice.isNilOrEmpty }
```
shiki, line 483: This is just a thought exercise.

Not sure if this is going to work but if it will, it will probably perform better. We _could_ include this `contains` predicate in the `ResultsController` predicate instead. 

```suggestion
        let predicate = NSPredicate(format: "product.siteID = %ld AND product.productID = %ld AND (product.regularPrice == nil OR product.regularPrice == '')", product.siteID, product.productID)
```

This _should_ return only the list of variations that don't have prices. And then we'd be able to use `isEmpty` and no mapping will ever happen. 

The problem with this is there's no way to check the other condition above, if the product has variations or not. So it's not enough. 

Something to think about. Our `ResultsController` API is unwieldy and could be better. Sometimes I wish I can just run SQL. 🤪

shiki (approved):
Tested and it works! 

There are only a couple of things I feel strongly about, marked with ⚠️ . The others are just questions, suggestions, and ramblings. 😄 

Merge away when you're satisfied! 🚢

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
Ecarrion: I do think we should update it to Combine. But I didn't want to have combine & our observables merged in the same file. I think we should tackle the migration as a separate task. 

I've created the issue https://github.com/woocommerce/woocommerce-ios/issues/4379

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModel.swift
Ecarrion: Nice tip! Thanks

Ecarrion:
Thanks for the review @shiki 🙇

PR #4350, opened 2021-06-03
Add-ons: Add Variations Attributes row

### LINKED ISSUE #3702: Product form > Variations: "1 options" is shown for an attribute with one option
**Describe the bug**

When there is only one option for a variation attribute, it's currently shown as "1 options" for the attribute in the product form. Let's show the singular format in this case.

**To Reproduce**

Prerequisite: the store has a variable product that has an attribute with only one option

- Go to the products tab
- Tap on a variable product that has an attribute with only one option --> the attribute shows "1 options"

**Screenshots**

**Expected behavior**

Let's show "1 option" in this case.

**Mobile Environment**
Please include:
- Device: iPhone 11 sim
- iOS version: iOS 14.1
- WooCommerce iOS version: 5.0.0

### LINKED ISSUE #4312: Variations: Add "Variation Attributes" button
We are going to be doing some changes for editing attributes.

product attributes

## Tasks

- [x] Show "Variations Attribute" row when the product has at least one attribute

- [x] Make that button to navigate to the "Edit Attributes" screen

- [x] Remove buttons to edit attributes from the "Create Variation" flow. (Variations list, empty variations)

Ecarrion:
closes #4312
fixes #3702

# Why

As part of the Variations changes, we are trying to make the variation flow simpler. This PR takes out the "Product Variation Attributes" section that was inside the "Variations List" screen and makes it accessible from the "Product Form" screen.

# How

- Update `ProductAttribute` to conform to `GeneratedCopiable`

- Update `ProductFormActionsFactory` to include the `.attributes` actions for the variable products when the product has at least one attribute.

- Update `DefaultProductFormTableViewModel` to properly render the `.attributes` row and update the `.variations` row to show the variations count.

- Update `ProductFormViewController` to navigate to the "Edit Attributes" screen after the new "Variation Attributes" row is tapped.

- Update `ProductVariationsViewController` to remove the ability to edit product attributes for that screen.

# Demo

https://user-images.githubusercontent.com/562080/120580636-275df080-c3ef-11eb-8fc2-4f7d68d51694.mov

# Testing Steps

- Launch the app and navigate to a product with variations

- See that there is a "Variations" row stating how many variations the product has.

- See that there is a "Variations Attributes" row.

- Tap on the "Variations Attributes" row and see that it navigates to the "Edit Attributes" screen.

- Edit/Add/Delete an attribute

- Go back the to the main product screen, see that the "Variations Attribute" row is updated.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
@@ -92,6 +92,8 @@ private extension DefaultProductFormTableViewModel {
                 return .linkedProducts(viewModel: linkedProductsRow(product: product, isEditable: editable), isEditable: editable)
             case .noPriceWarning:
                 return .noPriceWarning(viewModel: noPriceWarningRow())
+            case .attributes(let editable):
+                return .attributes(viewModel: productVariationsAttributesRow(product: product.product, isEditable: editable), isEditable: editable)
             default:
                 assertionFailure("Unexpected action in the settings section: \(action)")
                 return nil
@@ -420,6 +422,18 @@ private extension DefaultProductFormTableViewModel {
         return ProductFormSection.SettingsRow.WarningViewModel(icon: icon, title: title)
     }
 
+    func productVariationsAttributesRow(product: Product, isEditable: Bool) -> ProductFormSection.SettingsRow.ViewModel {
+        let icon = UIImage.variationsImage
+        let title = Localization.productVariationAttributesTitle
+
+        let format = NSLocalizedString("%1$@ (%2$ld options)", comment: "Format for each Product attribute")
+        let details = product.attributesForVariations
+            .map({ String.localizedStringWithFormat(format, $0.name, $0.options.count) })
+            .joined(separator: "\n")
+
+        return ProductFormSection.SettingsRow.ViewModel(icon: icon, title: title, details: details, isActionable: isEditable)
+    }
+
     func variationAttributesRow(productVariation: EditableProductVariationModel, isEditable: Bool) -> ProductFormSection.SettingsRow.ViewModel {
         let icon = UIImage.customizeImage
         let title = Localization.variationAttributesTitle
@@ -573,6 +587,10 @@ private extension DefaultProductFormTableViewModel {
         static let variationStatusTitle =
             NSLocalizedString("Enabled",
                               comment: "Title of the status row on Product Variation main screen to enable/disable a variation")
+
+        // Product Variations Attributes
+        static let productVariationAttributesTitle = NSLocalizedString("Variations Attributes",
+                                                                       comment: "Title of the variations attributes row on Product screen")
 
         // Variation attributes
         static let variationAttributesTitle = NSLocalizedString("Attributes", comment: "Title of the attributes row on Product Variation main screen")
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -420,6 +407,18 @@ private extension DefaultProductFormTableViewModel {
         return ProductFormSection.SettingsRow.WarningViewModel(icon: icon, title: title)
     }
 
+    func productVariationsAttributesRow(product: Product, isEditable: Bool) -> ProductFormSection.SettingsRow.ViewModel {
+        let icon = UIImage.customizeImage
+        let title = Localization.productVariationAttributesTitle
+
+        let format = NSLocalizedString("%1$@ (%2$ld options)", comment: "Format for each Product attribute")
```
rachelmcr: nit: Maybe add this string to the `Localization` enum.

rachelmcr: Since we're working on this section now, it might also be worth updating this string to fix https://github.com/woocommerce/woocommerce-ios/issues/3702? We could check for when `options.count` is 1 and use a different string in that case.

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -569,11 +568,32 @@ private extension DefaultProductFormTableViewModel {
...
+                    return NSLocalizedString("%1$ld variation",
+                                             comment: "Format for the variations detail row in singular form. Reads, `1 variation`")
+                default:
+                    return NSLocalizedString("%1$ld variations",
+                                             comment: "Format for the variations detail row in plural form. Reads, `2 variations`")
+                }
+            }()
+
+            return String.localizedStringWithFormat(format, count)
+        }
```
rachelmcr, line 588: This is a really clear way to handle the singular/plural cases! ⭐  I wish we could properly handle plurals for languages (like Russian) that have different plural forms (not just 1 or many), but until we can do that this is great. (We don't have an open issue for supporting different plural forms, but it's the same issue described in https://github.com/wordpress-mobile/WordPress-iOS/issues/6327.)

rachelmcr (approved):
LGTM! Tests well with various cases:

* New variable product
* Existing variable product with no variations or attributes
* Existing variable product with attributes but no existing variations
* Existing variable product with one or more variations/attributes

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
Ecarrion: [Stringdict](https://samwize.com/2018/07/19/how-stringsdict-and-nslocalizedstring-works/) would be the tool to help us with that but I think it is not supported by Glotpress yet :-(

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
Ecarrion: Oh, I wasn't aware of that issue.
Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4350/commits/85ac1576277e12c0f26068834e145194cd071a5b

Screen Shot 2021-06-03 at 7 13 17 AM

PR #4351, opened 2021-06-03
[Mobile Payments] Ensure we don't start another reader discovery process until the previous one is fully canceled

koke:
Part of #4338 

After #4339, I started noticing that sometimes if you start discovering readers and cancel, the app gets into a broken state where it says it can't start discovering readers because it's already discovering readers. On top of that, the error alert had "Bluetooth permission required" as the title.

To reproduce the issue, it's a bit hit and miss, but cancel a search in progress and try to start a new one within a second.

https://user-images.githubusercontent.com/8739/120641526-fa075600-c473-11eb-9adb-ea52d8d89c58.mp4

## How

#4339 introduced a delayed call to completion to give `StripeTerminal` enough time to clean up its internal state. However, the UI doesn't wait for this and optimistically dismisses the modal immediately, leaving the option to tap on "Connect Card Reader" while the previous cancelation is still in progress.

Internally, the discovery cancelation happens fairly quickly, but since we have a 1 second delay, you can reach a scenario where the user taps start, the discovery process starts successfully because Stripe's internal state is cleared already, and then a second later, the completion block for cancelation triggers and sets the service's state to `idle`.

To avoid this issue, I've introduced a lock to make sure we only start a process if we're not canceling one, and vice versa. Instead of waiting indefinitely, there's a 1 second timeout, so there's still a chance of the app complaining that it's busy if the user tries to start too quickly, but at least dismissing that alert and trying again will work.

Finally, I've updated the modal alerts to handle Bluetooth permissions and other errors separately.

Bluetooth error | Other error
-|-
Screen Shot 2021-06-03 at 14 07 16|Screen Shot 2021-06-03 at 14 06 46

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalBluetoothRequired.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalBluetoothRequired.swift
@@ -2,7 +2,7 @@ import UIKit
 import Yosemite
 
 /// Modal presented on error
-final class CardPresentModalScanningFailed: CardPresentPaymentsModalViewModel {
+final class CardPresentModalBluetoothRequired: CardPresentPaymentsModalViewModel {
 
     /// The error returned by the stack
     private let error: Error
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsAlerts.swift
```diff
@@ -1,4 +1,5 @@
 import UIKit
+import Hardware
```
ctarda: I would suggest importing Yosemite instead. Yosemite maps 
`public typealias CardReaderServiceError = Hardware.CardReaderServiceError
`

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalBluetoothRequired.swift
```diff
@@ -0,0 +1,72 @@
+import UIKit
+import Yosemite
+
+/// Modal presented on error
+final class CardPresentModalBluetoothRequired: CardPresentPaymentsModalViewModel {
```
ctarda, line 5: Would a new unit test for this make sense?

ctarda (approved):
Looks good to me, and seems to work well. 

One suggestion about preventing an import of Hardware, and another one about adding a test.

:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsAlerts.swift
koke: Ah, I thought about adding the type alias, but I should have checked if it was already there.

PR #4353, opened 2021-06-03
[Mobile Payments] Improve UI to retry failed payments

### LINKED ISSUE #4227: [Mobile Payments] Add a "no thanks" button to the error modal.
Otherwise we would get trapped in an endless loop of retrying.

koke:
Fixes #4227 

I started trying to add a "No thanks" button to the payment error alert, but soon noticed that the try again button was not working. I tested this with an order amount ending in `.75`, which simulates the payment being declined by the payment processor.

It looks like the retry operation was trying to cancel any payment in progress before starting again, but the callback method on `paymentCancellable.cancel` was not being called, presumably because the payment collection was completed successfully and the error came on the `processPayment` stage.

With this PR, the `paymentCancellable` is cleared when `collectPaymentMethod` completes, and `cancelPaymentIntent` will only cancel the payment intent if there is no `paymentCancellable` to cancel. The [documentation](https://stripe.com/docs/api/payment_intents/cancel) says:

> A PaymentIntent object can be canceled when it is in one of these statuses: requires_payment_method, requires_capture, requires_confirmation, or requires_action.

In my testing, the active intent had the `requires_payment_method` status, which I understand it means we should be canceling it, but I still get confused around payment intents and collection, so I could use a second opinion here.

Besides that, I modified the alert to show the "No thanks" button to dismiss, now that the retry button would get you into an infinite loop.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -214,25 +214,28 @@ extension StripeCardReaderService: CardReaderService {
                 return
             }
 
+            let cancelPaymentIntent = { [weak self] in
+                Terminal.shared.cancelPaymentIntent(activePaymentIntent) { (intent, error) in
+                    if let error = error {
+                        let underlyingError = UnderlyingError(with: error)
+                        promise(.failure(CardReaderServiceError.paymentCancellation(underlyingError: underlyingError)))
+                    }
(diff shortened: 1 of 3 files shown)


Ecarrion (approved):
Code & explanation looks good to me.

Couldn't test it because my store is not yet enrolled in the Beta program and also the PR does not have testing steps 😅

ctarda:
Merging on behalf of @koke

PR #4365, opened 2021-06-03
Do Not Send Sign-up Emails if the Email Address Is Unknown

### LINKED ISSUE #3962: Automatic Sign-ups Should Be Disabled
## Problem

In https://github.com/woocommerce/woocommerce-ios/issues/3146, we prevented the app from automatically creating a WordPress.com account. But at some point, most probably in https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/574, that change was reversed. The app now allows automatic sign-ups again. 

Tested in recent `develop`: 4fe9522. 

https://user-images.githubusercontent.com/198826/114770987-4d3d1200-9d29-11eb-9154-91843717151f.mp4

Android still works as expected. 

## Expected

I would expect the functionality shown in #3146 (PR https://github.com/woocommerce/woocommerce-ios/pull/3286) to be restored. 

## Tasks

- [ ] This could probably use an automated test. Hopefully, it's feasible.

---

FYI @designsimply This is probably one of those critical scenarios that we should keep track too.

shiki:
The WPAuth dependency for this has been merged: https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/595. 

Fixes #3962.  

## Findings

We do not allow automatic sign-up emails. If the entered email address is not recognized, we used to display this custom error message:

However, this regressed when https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/574 was merged. And eventually got “fixed” again when https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/590 was merged. Funny how life works. It wasn't a complete fix though because it doesn't show the custom UI. It only shows an alert:

## Solution

In https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/595, I restored the callback that was removed by https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/574. 

I also changed how we detect an unknown email. Previously, we were using the `7` error code which matches [`WordPressComRestApiError.unknown`](https://github.com/wordpress-mobile/WordPressKit-iOS/blob/aa59a1627455e0c6b6deb0bf48fdab38e4f27ce1/WordPressKit/WordPressComRestApi.swift#L7-L27). But based on the definition of `.unknown`, it's literally "unknown" and can be pretty much anything. 

https://github.com/woocommerce/woocommerce-ios/blob/95aa04f23e58cd20d32ca7b5115462626c765698/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L399

I changed it to be the same as what [WPAuth does](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/6a463a807be8a042d450d69f883d6399c01f09fd/WordPressAuthenticator/Unified%20Auth/View%20Related/Get%20Started/GetStartedViewController.swift#L403), which is detecting the underlying error code coming from the API. This is probably safer. 

https://github.com/woocommerce/woocommerce-ios/blob/72659b2215751445d98b5bc61dd226d9b4f001c4/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L428-L434

## Testing

Please test that the app will now display the custom UI. 

Log in with an email that is not registered in WPCOM. The app should respond with the custom UI saying “It looks like this email is not associated with...”.

## Author Checklist


WooCommerce/Classes/Authentication/AuthenticationManager.swift
```diff
@@ -390,13 +390,32 @@ extension AuthenticationManager: WordPressAuthenticatorDelegate {
     }
 }
 
+// MARK: - ViewModel Factory
```
shiki, line 393: I made this accessible so I could add more unit tests. I was thinking of testing the view controllers [exposed by `handleError`](https://github.com/woocommerce/woocommerce-ios/blob/72659b2215751445d98b5bc61dd226d9b4f001c4/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L206-L214)  and use [the `viewModel`](https://github.com/woocommerce/woocommerce-ios/blob/72659b2215751445d98b5bc61dd226d9b4f001c4/WooCommerce/Classes/Authentication/Navigation%20Exceptions/ULErrorViewController.swift#L10-L12) to test. But I didn't want to make that public. That seems worse to me. 

Let me know if there's a better way to do this.

ctarda (approved):
Tests perfect.


https://user-images.githubusercontent.com/2722505/120834963-ba9f3f00-c531-11eb-811a-8600208161ee.mov



:shipit:

shiki:
Thank you, César!

PR #4372, opened 2021-06-03
Shipping Labels: Disable payment method editing for non-store owners

### LINKED ISSUE #4087: Shipping Labels Payments: disable editing for non-store owners
If the user creating the shipping label is not the store owner, they can't change the payment method, so we need to display the below screen to let them know that a payment method is already in the file and that they can't edit it.

Feel free to split this task into sub-tasks.

Handle these two permissions from `ShippingLabelAccountSettings`:

- [ ] `canManagePayments` — permission for the user to manage the payment methods
- [ ] `canEditSettings` — whether the user can edit non-payment settings (e.g. email receipts setting); this permission is always true for now.

rachelmcr:
Closes: #4087

## Description

If the user creating the shipping label is not the store owner, they can't change the payment method. This PR disables the payment method fields and displays a banner to non-store-owners to explain why that can't be edited.

## Changes

* Adds a banner `ShippingLabelPaymentMethodsTopBanner` with informational text about why the payment methods are disabled.
* Adds conditionals to `bodyStyle()` and `footnoteStyle()` in `WooStyleModifiers`, so the text color can be conditionally set based on whether the view is enabled or disabled.
* Updates `SelectableItemRow` and `TitleAndToggleRow` to set the body and footnote styles based on whether the view is enabled or disabled.
* Updates `ShippingLabelPaymentMethodsViewModel` to include variables for the permissions to edit the payment methods (store owners only) and edit the email receipts setting (all users), based on the shipping label account settings.
* Updates `ShippingLabelPaymentMethods` to add the banner and disable the fields on that screen, based on the user's permissions. Also adds a SwiftUI preview for a fully disabled state. (The diff for this file shows a lot of lines changed, but many of those are just indentation changes because the `ScrollView` is now embedded in `GeometryReader`.)

## Testing

1. Make sure you have one or more payment methods set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Create an administrator account in your store (separate from the store owner account) if you don't already have one.
3. Launch the app in debug mode (because of the feature flag).
4. Log in to the app using your non-store-owner account.
5. Open an order detail that is eligible for the shipping label, where the customer paid a flat fee for the shipping.
6. Tap the button "Create Shipping Label".
7. Pass all the steps until you reach Payment Method.
8. Tap the Payment Method cell to open the Payment Method screen.
9. Confirm that editing the payment methods is disabled, and there is a warning to explain it.
10. Confirm that you can toggle the email receipts setting.

## Submitter Checklist

Update release notes:


WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/SelectableItemRow.swift
```diff
@@ -5,20 +5,21 @@ struct SelectableItemRow: View {
     let title: String
     let subtitle: String
     let selected: Bool
+    @Environment(\.isEnabled) var isEnabled
```
pmusolino, line 8: Interesting, I didn't know there was an environment property to understand if the view is enabled or not.

pmusolino (approved):
LGTM :shipit: 

Code ✅ 
Testing ✅

PR #4373, opened 2021-06-03
[Mobile Payments] Add order number to payment intent description

### LINKED ISSUE #4206: [Mobile Payments] More meaningful descriptions for payment intents
Right now we create payment intents with `description` set to

`Receipt from {Name of Store}`

This appears in receipts from Stripe as:

receipt

And appears in our internal dashboard as:

receipt-from

It would be better to have this say something like

`Online payment for order 4567 for {Name of Store}`

See also related issue for web payments at https://github.com/Automattic/woocommerce-payments/issues/1807

allendav:
Closes #4206 

Changes:
- Adds `order.number` to the payment intent description
- Why not orderID? `order.number` is a comma-less string making it more suitable than the Int64 which, when localized, gains grouping commas)

To test:
- Complete a card present payment
- Open the internal WooCommerce Payments dashboard and ensure the transaction (when viewing as the merchant) shows the new description with the order number, i.e.:

ordernumber

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
@@ -157,18 +157,19 @@ private extension PaymentCaptureOrchestrator {
 
         return PaymentParameters(amount: orderTotal as Decimal,
                                                   currency: order.currency,
-                                                  receiptDescription: receiptDescription(),
+                                                  receiptDescription: receiptDescription(orderNumber: order.number),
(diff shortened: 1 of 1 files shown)


koke:
> Why not orderID? order.number is a comma-less string making it more suitable than the Int64 which, when localized, gains grouping commas)

The grouping comas come from us explicitly using `String.localizedStringWithFormat(_:_:)`, which we could replace with `String.init(format:_:)`. That said, I'm seeing that we generally use `number` for display across the app and `orderID` for more API/programatic use, so your solution makes sense.

The code looks good but I don't know how to access that dashboard to test this.

ctarda (approved):
Test great. :shipit:

PR #4380, opened 2021-06-04
Better diffs on Unit Tests

### LINKED ISSUE #3779: Replace xcpretty with trainer
While trying to implement p91TBi-4AQ-p2 to get better diffs on Xcode unit tests, we noticed that there is a problem with our CI where only the first line of the error message is being displayed. 

@AliSoftware noticed that the reason for this is because `xcpretty` which is part of our Fastlane setup is ignoring the rest of the error message lines when prettifying the output. 

The suggestion is to replace `xcpretty` with `trainer` and see if that fixes our problem.

Ecarrion:
closes #3779 

# Why

In p91TBi-4AQ-p2 we agreed on integrating [Difference](https://github.com/krzysztofzablocki/Difference) to easily identify what is different when a `XCTAssertEqual` fails with two complex types.

Before | After
---- | ----
bad-error | good-error

And while that worked well, we found out that `xcpretty` which is used on our CI, does not work well with multi-line test errors.

To circumvent that problem, this PR apart from integrating [Difference](https://github.com/krzysztofzablocki/Difference), adds test report generation using [Trainer](https://github.com/fastlane-community/trainer) which is a tool that generates reports from `xcresults` files that does not have problems with multi-lines test messages.

# How

- Install `Difference` as a dependency of `TestKit` using SPM.

- Adds shadow `XCTAssertEqual` function that uses `Difference` inside.

- Adds trainer as a dependency in our `Pluginfile`

- Update `test_without_building` lane to generate reports using trainer while deleting reports generated by `multi_scan`

# Tests

- CI with test errors: https://circleci.com/gh/woocommerce/woocommerce-ios/22576

- CI with no errors: https://circleci.com/gh/woocommerce/woocommerce-ios/22578

Update release notes:


fastlane/Fastfile
```diff
@@ -826,6 +826,34 @@ end
...
+      next if File.extname(report_path) != ".xcresult"
+      trainer(path: report_path, fail_build: false)
+    end
   end
+
+  # -----------------------------------------------------------------------------------
+  # Deletes html & junit reports generated by scan.
+  # -----------------------------------------------------------------------------------
+  def delete_scan_reports(report_files)
+    report_files.each do | report_path |
```
AliSoftware: 💄 nit: the rubocop linter and common guidelines advocate for no spaces around `|` bars:
```suggestion
    report_files.each do |report_path|
```

fastlane/Fastfile
```diff
@@ -826,6 +826,34 @@ end
...
+      next if File.extname(report_path) == ".xcresult"
+      File.delete(report_path)
+    end
+  end
```
AliSoftware: 💄 nit: "No newline at end of file" is considered bad convention

fastlane/Fastfile
```diff
@@ -826,6 +826,34 @@ end
...
+
+  # -----------------------------------------------------------------------------------
+  # Geneate trainer reports for each xcresult report file.
+  # -----------------------------------------------------------------------------------
+  def generate_trainer_reports(report_files)
+    report_files.each do | report_path |
+      next if File.extname(report_path) != ".xcresult"
+      trainer(path: report_path, fail_build: false)
+    end
```
AliSoftware: 💄 style nit:
```suggestion
    report_files.reject { |path| File.extname(report_path) != '.xcresult' }.each do |report_path|
      trainer(path: report_path, fail_build: false)
    end
```

.circleci/config.yml
```diff
@@ -99,6 +99,8 @@ jobs:
       - run:
           name: Run Unit Tests
           command: bundle exec fastlane test_without_building name:UnitTests try_count:3 device:'<< parameters.device >>' ios-version:'<< parameters.ios-version >>'
+      - store_test_results:
+          path: build/results
       - ios/save-xcodebuild-artifacts:
           result-bundle-path: build/results
```
AliSoftware: ⚠️  Seems that those 2 steps are kinda "conflicting", as the second one already includes the first one, as you can see [in its definition in the CircleCI orb here](https://github.com/wordpress-mobile/circleci-orbs/blob/trunk/src/ios/orb.yml#L355-L365).

Which not only leads to an unnecessary step, but also to the second one not finding any report once the first passed:
image

fastlane/Fastfile
```diff
@@ -826,6 +826,34 @@ end
...
+      result_bundle: true,
+      fail_build: false
     )
+
+    # Generate reports with trainer and delete scan ones.
+    generate_trainer_reports(summary[:report_files])
+    delete_scan_reports(summary[:report_files])
+
+    # Manually fail build if test have failed.
+    UI.test_failure!("Tests have failed") if summary[:failed_testcount] > 0
```
AliSoftware: 💄 nit
```suggestion
    UI.test_failure!("#{summary[:failed_testcount]} tests have failed") if summary[:failed_testcount] > 0
```

.circleci/config.yml
AliSoftware: 😢  Also sadly, CircleCI doesn't seem to show multiline errors properly either in its "Tests" tab:
image

Not sure if it's worth spending too much time here since we plan to migrate away from CircleCI soon-ish so that multiline issue might be solved there… or not

@jkmassel any idea how BuildKite might show this kind of test result in their UI? I don't think they have a dedicated "Test Results" tab like CircleCI does, right, so we will either have to log the XML files in stdout with the risk of being truncated, or just upload the test result XMLs as artefacts and let people download them to inspect failures?

---

In any case, it would be nice if we could find a way to print those multiline error messages from the XML to stdout, so that (1) they are visible in CircleCI log output directly with nice formatting (unlike Test tab which ignores newlines apparently) (2) we'll be ready to also be able to see them once we migrate to BuildKite (or another service that doesn't have a Tests tab like CCI). Sadly `trainer` can only generate JUnit XML and no other format like HTML or plain text (maybe with a simple-enough XSLT stylesheet and calling `xsltproc` we could get a nice output? 🤔 🤔 )

I'm ok to keep that for later and only keep the Tests tab without multiline support for now (I'm guessing this was a HACK week project and week is now over anyway… 😛 )

TestKit/Sources/TestKit/Assertions.swift
```diff
@@ -47,3 +48,19 @@ public func assertThat<T>(_ subject: Any?, isAnInstanceOf expectedType: T.Type,
...
+/// Overrides regular `XCTAssertEqual` to show a formatted difference when `expected` and  `received` are not equal,
+///
+public func XCTAssertEqual<T: Equatable>(_ expected: @autoclosure () throws -> T,
+                                         _ received: @autoclosure () throws -> T,
+                                         file: StaticString = #filePath,
+                                         line: UInt = #line) {
+    do {
+        let expected = try expected()
+        let received = try received()
+        XCTAssertTrue(expected == received, "Found difference for \n" + diff(expected, received).joined(separator: ", "), file: file, line: line)
```
koke: I noticed that, by using this, the test failure will start with `XCTAssertTrue failed`, which might be a bit confusing. I think we can use [custom failures with XCTIssue](https://developer.apple.com/videos/play/wwdc2020/10687/), but that requires access to the `XCTestCase` object, so it might require a base class instead of a top-level function 😞

AliSoftware (commented):
Thanks for tackling this, great idea!

The general code LGTM
 - Left a couple of style nits, take them or leave them.
 - I'm pretty sure the step to store tests should be removed as already included in the orb command
 - A bit disappointed by CircleCI not handling multilines apparently either. Left some ideas and reflections on how we could try other solutions to circumvent this limitation, but I'm ok to merge as is and let us get back to those details in a later HACK week

TestKit/Sources/TestKit/Assertions.swift
```diff
@@ -47,3 +48,19 @@ public func assertThat<T>(_ subject: Any?, isAnInstanceOf expectedType: T.Type,
                   file: file,
                   line: line)
 }
+
+/// Overrides regular `XCTAssertEqual` to show a formatted difference when `expected` and  `received` are not equal,
+///
+public func XCTAssertEqual<T: Equatable>(_ expected: @autoclosure () throws -> T,
```
koke: Overriding a system function name seems a bit tricky. I noticed that if you don't `import TestKit` in your tests, it will use the `XCTest` one. Even if it requires a lot of search and replace, I think we might be better off with a more explicit `assertEqual`, which is more consistent with the other functions in this file.

.circleci/config.yml
Ecarrion: Hey @AliSoftware I added that `store_test_results` because the `ios/save-xcodebuild-artifacts` step looks for reports in the `build/reports` folder wich does not exists. You can see that any other PR [does nothing](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/12376/workflows/11293f6f-a769-47b4-b58d-5701526555c6/jobs/22604) in that step. 

Do you think I should remove the `- ios/save-xcodebuild-artifacts:` step?

koke (commented):
Overall, I love this idea, and it can make the results much clearer. However, I think the Difference library is too eager assuming every struct should be treated like a dictionary, and will show you differences even in private properties, and even when you might have a simple description for a complex object. This is unless your type implements `CustomReflectable` to provide its simpler representation, but I don't think we'd want that in most cases.

I tried a quick example to show what I meant:

```swift
func test_sample() {
    struct CurrencyPair: CustomStringConvertible, Equatable {
        private let quote: String
        private let base: String

        init(_ symbol: String) {
            let components = symbol.split(separator: "/")
            quote = String(components[0])
            base = String(components[1])
        }

        var description: String {
            return [quote, base].joined(separator: "/")
        }
    }

    let expected = CurrencyPair("USD/EUR")
    let received = CurrencyPair("GBP/EUR")

    XCTAssertEqual(expected, received, "XCTest variant")
    XCTAssertEqual(expected, received)
}
```

Screen Shot 2021-06-11 at 16 54 17

Another issue to be aware of, is that it Difference compare every children, even if they don't participate in the equality checks. I don't imagine this causing a lot of trouble in our tests, but it could lead to some confusing scenarios in theory.

Screen Shot 2021-06-11 at 17 06 23


I'd go back to my earlier comment, and make this an explicit opt-in method rather than trying to override `XCTAssertEqual`.

I was trying [Mocha](https://mochajs.org) for a personal project recently, and something like `assertDeepEqual` might be a good match for this.

I haven't looked much into the Trainer reporting and CI integration, I'll let @AliSoftware comment on that

.circleci/config.yml
AliSoftware: Ah good point I missed the difference between `build/results` and `build/reports` there 😅 

We should keep that `save-xcodebuild-artefacts` step as it also upload crash reports that might have been encountered during test runs and other stuff so I think it makes more sense to keep it. So either we keep the solution you used here (and one of the 2 steps will not find anything), or you can check if it's easy to change the location for the trainer reports from `build/results` to `build/reports` if trainer provides that option. If not, another alternative could be to add a `mv build/results build/reports` step to do that 😛

Ecarrion: I think what we can do is to make `multi_scan` [to output](https://github.com/woocommerce/woocommerce-ios/blob/develop/fastlane/Fastfile#L814) in `build/reports` would that be ok?

AliSoftware: wfm

Ecarrion:
@koke Thanks for the review 🙇 

We discussed a little bit in p91TBi-4AQ-p2 about having a separate function or overriding `XCTAssertEqual`. 

I'm not a fan of having two different functions to achieve the same because often one will go unused or forgotten, but I wasn't aware of those limitations you just mentioned.

While I think that the cases you expose could be considered edge cases, If you feel strongly about them I will turn this into an explicit opt-in function and see how the adoption rate goes. 

What do you think?

TestKit/Sources/TestKit/Assertions.swift
Ecarrion: Nice I didn't knew about this!

TestKit/Sources/TestKit/Assertions.swift
Ecarrion: Will give it a try.

Ecarrion:
@AliSoftware 
Thanks for all the ruby style suggestions. You can tell I'm not a ruby dev 😛
I've also changed the `multi_scan` output to `build/reports` so that everything can be uploaded in the `ios/save-xcodebuild-artifacts` step and it [seems to be working](https://circleci.com/gh/woocommerce/woocommerce-ios/22624) correctly!

Screen Shot 2021-06-11 at 11 44 09 AM


@koke I took your suggestions for a spin to see how they would look. I really like the new `XCTIssue` API. Let me know what you think.

koke (approved):
> @koke I took your suggestions for a spin to see how they would look

Oh an extension makes perfect sense, I missed that possibility. I guess that's what I get for doing reviews on a Friday afternoon 😂 

The only thing I'd suggest, to mimic the messaging from XCTest a bit more closely is to add which assertion method failed, and let the developer specify an optional message, now that it's not a problem for overriding. Following XCTest patterns, that would be `assertEqual failed: Found difference ... - {optional message}`

Other than that, the assertion part looks good to me ✨

AliSoftware (approved):
Apart from the fact that CircleCI still doesn't handle multilines in its Test tab which kinda defeats the purpose of `Difference` output, LGTM.

_We can always submit a subsequent PR if we find the time to find a solution for that, especially once we are done migrating to our new CI (paaHJt-22H-p2), so let's merge this in the meantime especially since it's useful in Xcode itself already_

Issue #4314, opened 2021-05-28
[Mobile Payments] Fetch WCPayAccount from Storage in Settings and OrderDetails


ctarda:
At this moment, we fire a network request to fetch the WCPayAccount object in order to check merchant eligibility in:
* Settings (to show the "Manage Card Reader" button
* Order details (to show the "Collect Payment" button)

Once we have #4305 in, we should fetch that object from the local caches

PR #4381, opened 2021-06-04
[Mobile Payments] Add PaymentGatewayAccountStore

allendav:
Partially addresses #4314

This PR builds on #4349 (which added PaymentGatewayAccount to CoreData).

This PR adds a new store - PaymentGatewayAccountStore - that knows how to get PaymentGatewayAccounts from Storage and convert WCPayAccounts from the WCPayRemote into them.

A subsequent PR will change SettingsViewController and OrderDetails to use a PaymentGatewayAccount ResultsController, this store, and remove the WCPayStore

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -46,7 +46,7 @@ final class SettingsViewController: UIViewController {
 
     /// ResultsController: Loads Sites from the Storage Layer.
     ///
-    private let sitesResultsController: ResultsController<StorageSite> = {
+    private let resultsController: ResultsController<StorageSite> = {
         let storageManager = ServiceLocator.storageManager
         let predicate = NSPredicate(format: "isWooCommerceActive == YES")
         let descriptor = NSSortDescriptor(key: "name", ascending: true)
@@ -57,30 +57,6 @@ final class SettingsViewController: UIViewController {
     /// Sites pulled from the results controlelr
     ///
     private var sites = [Yosemite.Site]()
-
-    /// Payment Gateway Account ResultsController
-    ///
-    private var gatewayAccountResultsController: ResultsController<StoragePaymentGatewayAccount>? = {
-        guard let siteID = ServiceLocator.stores.sessionManager.defaultStoreID else {
-            DDLogError("⛔️ Cannot find ID for current site to load accounts for!")
-            return nil
-        }
-
-        let storageManager = ServiceLocator.storageManager
-        let predicate = NSPredicate(format: "siteID == %lld", siteID)
-
-        return ResultsController(storageManager: storageManager, matching: predicate, sortedBy: [])
-    }()
-
-    /// Accounts pulled from the results controller
-    ///
-    private var gatewayAccounts: [PaymentGatewayAccount] {
-        guard gatewayAccountResultsController != nil else {
-            return []
-        }
-
-        return gatewayAccountResultsController?.fetchedObjects ?? []
-    }
 
     /// Store Picker Coordinator
     ///
@@ -95,9 +71,11 @@ final class SettingsViewController: UIViewController {
     override func viewDidLoad() {
         super.viewDidLoad()
 
-        configureResultsControllers(onReload: { [weak self] in
+        checkAvailabilityForPayments() { [weak self] in
             self?.refreshViewContent()
-        })
+        }
+
+        refreshResultsController()
         configureNavigation()
         configureMainView()
         configureTableView()
@@ -134,33 +112,35 @@ private extension SettingsViewController {
         tableView.delegate = self
     }
 
-    func configureResultsControllers(onReload: @escaping () -> Void) {
-        configureSiteResultsController(onReload: onReload)
-        configureAccountResultsController(onReload: onReload)
+    private func refreshResultsController() {
+        try? resultsController.performFetch()
+        sites = resultsController.fetchedObjects
     }
 
-    private func configureSiteResultsController(onReload: @escaping () -> Void) {
-        try? sitesResultsController.performFetch()
-        sites = sitesResultsController.fetchedObjects
-    }
-
-    private func configureAccountResultsController(onReload: @escaping () -> Void) {
-        /// Bail early if the card present payments feature is disabled
-        ///
-        guard ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentPayments) else {
+    func checkAvailabilityForPayments(onCompletion: @escaping () -> Void) {
+        guard let siteID = self.siteID else {
+            canCollectPayments = false
+            onCompletion()
             return
         }
 
-        gatewayAccountResultsController?.onDidChangeContent = {
-            onReload()
+        let action = WCPayAction.loadAccount(siteID: siteID) { [weak self] result in
+            guard let self = self else {
+                return
+            }
+
+            switch result {
+            case .failure:
+                self.canCollectPayments = false
+            case .success(let account):
+                self.canCollectPayments = ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentPayments) &&
+                    account.isCardPresentEligible
+            }
+
+            onCompletion()
         }
 
-        gatewayAccountResultsController?.onDidResetContent = {
-            // TODO - is this correct for onDidResetContent?
-            onReload()
-        }
-
-        try? gatewayAccountResultsController?.performFetch()
+        ServiceLocator.stores.dispatch(action)
     }
 
     func configureTableViewFooter() {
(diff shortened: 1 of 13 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -58,6 +58,20 @@ final class SettingsViewController: UIViewController {
     ///
     private var sites = [Yosemite.Site]()
 
+    /// Payment Gateway Account ResultsController
+    ///
+    private let gatewayAccountResultsController: ResultsController<StoragePaymentGatewayAccount> = {
+        let storageManager = ServiceLocator.storageManager
+        /// TODO - resolve how I want to manage the siteID here. Should I just fetch all gatewayAccounts and filter by siteID later perhaps?
```
ctarda: I'm curious, what would be the benefit of fetching all the payment gateways?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -58,6 +58,20 @@ final class SettingsViewController: UIViewController {
...
+        let predicate = NSPredicate(format: "siteID == %lld", siteID)
```
ctarda: Extra super mage nitpick comments, but siteID is probably an optional at this point, so maybe the results controller would hav to be optional as well?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -112,35 +124,35 @@ private extension SettingsViewController {
...
-                self.canCollectPayments = ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentPayments) &&
-                    account.isCardPresentEligible
-            }
-
-            onCompletion()
+            onReload()
         }
 
-        ServiceLocator.stores.dispatch(action)
+        try? gatewayAccountResultsController.performFetch()
```
ctarda: We'd probably need to do something like:
`gatewayAccounts = gatewayAccountResultsController.fetchedObjects`

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -162,10 +174,31 @@ private extension SettingsViewController {
...
+            canCollectPayments = false
+            return
+        }
+
+        canCollectPayments = false
+
+        /// At the moment, we only expect one gateway account, but in the futue we may have multiple gateway accounts
+        /// each capable of supporting card present payments, so let's do this right now
+        ///
+        for gatewayAccount in gatewayAccounts {
```
ctarda: I am sure there is a way to do this via a combination of map and reduce. Or something like 

`canCollectPayments = gatewayAccounts.filter { $0.isCardPresentEligible == true}.isNotEmpty`

I am not sure that would make things much better though 😛

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
allendav: Nah, it would be better to only return the gatewayAccount(s) for the site. I was just thinking out loud.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
koke: There's [`Array.contains(where:_)`](https://developer.apple.com/documentation/swift/array/2297359-contains):

```swift
canCollectPayments = gatewayAccounts.contains(where: \.isCardPresentEligible)
```

Yosemite/Yosemite/Stores/PaymentGatewayAccountStore.swift
```diff
@@ -0,0 +1,84 @@
+import Foundation
+import Networking
+import Storage
+
+/// Implements `PaymentGatewayAccountAction` actions
+///
+public final class PaymentGatewayAccountStore: Store {
```
ctarda, line 7: Sorry if you have already considered this. Also, this is a question 😊 

So far, our UI was consuming a WCPayStore, which was pretty much used to capture the payment intent and load the account. 

Are you planning on keeping these two stores (WCPayStore and PaymentGatewayAccountStore) separate? Will those two be merged into one?

Yosemite/Yosemite/Actions/PaymentGatewayAccountAction.swift
```diff
@@ -0,0 +1,10 @@
+import Foundation
+import Networking
+
+/// Defines all of the `actions` supported by the `PaymentGatewayAccountStore`.
+///
+public enum PaymentGatewayAccountAction: Action {
+    /// Retrieves and stores payment gateway accounts for the provided `siteID`
+    ///
+    case loadAccounts(siteID: Int64, onCompletion: (Result<[PaymentGatewayAccount], Error>) -> Void)
```
ctarda: Do we really need to pass the accounts in the completion block? As far as I know, all we need for now is to be able to tell if a store is enrolled into the beta, and later, if it has WCPay installed. Those could be the actual actions?

ctarda: Also, in theory, once your store updates the core data store with a new value, the ResultsController in your UI should be notified of the update

Yosemite/Yosemite/Stores/PaymentGatewayAccountStore.swift
allendav: Long term plan is for PaymentGatewayAccountStore to be the only one - to allow us to support a variety of processors (not jus WCPay). Introducing PaymentGatewayAccountStore is the first step toward eventually decommissioning WCPayStore

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
allendav: In https://github.com/woocommerce/woocommerce-ios/pull/4381/commits/775566b5dcc26310f255e389af0a617e923a744c I removed this change - I'm deferring view controller changes to the last pull in this series.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
allendav: In https://github.com/woocommerce/woocommerce-ios/pull/4381/commits/775566b5dcc26310f255e389af0a617e923a744c I removed this change - I'm deferring view controller changes to the last pull in this series.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
allendav: In https://github.com/woocommerce/woocommerce-ios/pull/4381/commits/775566b5dcc26310f255e389af0a617e923a744c I removed this change - I'm deferring view controller changes to the last pull in this series.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
allendav: Can do! In https://github.com/woocommerce/woocommerce-ios/pull/4381/commits/775566b5dcc26310f255e389af0a617e923a744c I removed this change - I'm deferring view controller changes to the last pull in this series.

Yosemite/Yosemite/Model/Payments/WCPayAccount+PaymentGatewayAccount.swift
```diff
@@ -0,0 +1,21 @@
+import Hardware
+
+public extension WCPayAccount {
+    /// Maps a WCPayAccount into the PaymentGatewayAccount struct
+    ///
+    func toPaymentGatewayAccount(siteID: Int64) -> PaymentGatewayAccount { // TODO can we add siteID to WCPayAccount?
```
allendav, line 6: WCPayAccount doesn't have a siteID right now. Since we aren't storing it, I'm not sure it is appropriate to add it to it, and that this `toPaymentGatewayAccount` doing it makes sense, but I am open to y'alls thoughts

Yosemite/Yosemite/Actions/PaymentGatewayAccountAction.swift
allendav: Good idea. Changed it to simply `Result<Void, Error>`

Yosemite/Yosemite/Stores/PaymentGatewayAccountStore.swift
ctarda: Sounds good!

Yosemite/Yosemite/Model/Payments/WCPayAccount+PaymentGatewayAccount.swift
ctarda: It seems to me like there should be a relationship between WCPayAccount and a site, right? It's something we can add later. I guess.

ctarda (approved):
Thanks for iterating on this @allendav 

Given the "next steps" you have laid out in the PR description, I think we can merge this now. We can continue iterating as we move forward.

:shipit:

PR #4386, opened 2021-06-08
Update Fastlane's snapshot integration to version 1.25

mokagio:
See https://github.com/Automattic/simplenote-ios/pull/1316

For testing, I run `bundle install && bundle exec fastlane take_screenshots devices:'iPhone 11 Pro Max' languages:en-US`. But, I got stuck on a "start the device; can't start the device; try again" loop which kept going till it failed.

I'm not sure the issue is due to this change. Keen to know how that works on your end @jkmassel.


jkmassel (approved):
I wouldn't expect it to matter, but this is a nice improvement, so let's :shipit:

mokagio:
> I wouldn't expect it to matter, but this is a nice improvement, so let's :shipit:

Thanks.

I think Fastlane `snapshot` may bark at you if you don't have the file version it expects. 🤷‍♂️

PR #4387, opened 2021-06-08
Improve application log viewer

koke:
## Why

The old `ApplicationLogDetailViewController` has not kept up with the times. The current implementation is not that different from [the initial one in 2013](https://github.com/wordpress-mobile/WordPress-iOS/blob/037f7469e07985c2c874299c06a9d043f76e958e/WordPress/Classes/ActivityLogDetailViewController.m), and it has some issues:
- Most of the times that you go to that view, you either want to share the log, or read the latest lines
- The scroll view doesn't extend to the full width, so it's hard to drag the scroll bar
- Some layout guides might be wrong because dragging the scroll bar (when you can grab it), doesn't always get you to the bottom. I've noticed this more often in the WordPress app, where logs tend to be much longer.
- When you load the view, the text view is not scrolled at the top, it has a small offset.

https://user-images.githubusercontent.com/8739/121161371-36103180-c84d-11eb-8937-85006f34e512.mov

## What changed

Instead of dumping all the text into a text view, the new log viewer uses SwiftUI's List to show each line. We do a little bit of parsing to extract the log timestamp into a smaller header, so the log message is front and center. I think this will also make it easier to implement a searching feature if that seems useful enough.

New list design | Scroll to bottom | Button hides when not needed
-|-|-
|scroll-to-bottom|button-appear

## How

Implementing this with SwiftUI was relatively easy, but reading the scroll position is still a little rough. I had to use the `onAppear`/`onDisappear` modifiers on cells to keep track of whether the last cell was visible. This method is not completely reliable as they often get fired more than necessary (it seems there's a fix on iOS 15 for this, see paNNhX-et-p2), and the `onDisappear` method seems to be called when the cell is disposed/reused, not necessarily the moment it goes off-screen. The first problem was solved by using Combine and debouncing events each 100ms. The second problem isn't solved, but the result for this particular use case isn't too bad: the button doesn't hide exactly when the last cell scrolls off-screen, but it will hide eventually as you keep scrolling up.

## To test

1. Go to App Settings > Help & Support > View Application Log.
2. Pick the current log or any other.
3. The log should load with the new view scrolled to the top
4. Tapping on the arrow down button should scroll to the end of the log
5. Dragging the scroll bar should work as expected
6. Tapping the share button should allow to share the log (I tested with Airdrop)

Update release notes:


Garance91540 (approved):
👏

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,103 @@
+import SwiftUI
+
+/// Hosting controller that wraps an `ApplicationLogDetailView`
+///
+final class ApplicationLogDetailViewController: UIHostingController<ApplicationLogDetailView> {
+    init(with contents: String, for date: String) {
+        let viewModel = ApplicationLogViewModel(logText: contents)
```
Ecarrion: I've seen some places in the app where is preferred to pass the already created `ViewModel`
I think we should follow a common standard but there is none at the moment? What do you think?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,103 @@
...
+        super.init(rootView: ApplicationLogDetailView(viewModel: viewModel))
+        self.title = date
+        viewModel.present = { [weak self] vc in
+            self?.present(vc, animated: true, completion: nil)
+        }
```
Ecarrion: I wonder if instead of using completion blocks we could provide Combine publishers? 🤔

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,103 @@
...
+
+    required dynamic init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+struct ApplicationLogDetailView: View {
+    @ObservedObject var viewModel: ApplicationLogViewModel
+
+    private let dateFormatter: DateFormatter = {
```
Ecarrion: ⚠️  Do you think it's worth passing this as a parameter or adding the `@State` property wrapper to it? I wonder how many times this gets re-created with the view? 🤔

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,103 @@
...
+    var body: some View {
+        ScrollViewReader { scrollProxy in
+            List(viewModel.lines) { line in
+                VStack(alignment: .leading, spacing: 6) {
+                    line.date
+                        .flatMap(dateFormatter.string(for:))
+                        .map {
+                            Text($0)
+                                .footnoteStyle()
+
```
Ecarrion: nit: extra blank line

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,103 @@
...
+        formatter.timeStyle = .medium
+        return formatter
+    }()
+
+    var body: some View {
+        ScrollViewReader { scrollProxy in
+            List(viewModel.lines) { line in
+                VStack(alignment: .leading, spacing: 6) {
+                    line.date
+                        .flatMap(dateFormatter.string(for:))
```
Ecarrion: ⚠️  What's the reason for having this logic here instead of on the view model?
If we move it there we could have the `dateFormatter` there where it won't get recreated, right?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,103 @@
...
+                action: {
+                    viewModel.showShareActivity()
+                }, label: {
+                    Image(systemName: "square.and.arrow.up")
+            }))
+    }
+
+    func scrollToBottomButton(_ action: @escaping () -> Void) -> some View {
+        Group {
+            if viewModel.buttonVisible {
```
Ecarrion, line 58: nit: a `guard let` could make this easier to read with one less level of branching!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,103 @@
...
+                .padding()
+                .transition(.move(edge: .bottom))
+            }
+        }
+        .animation(.easeInOut(duration: 0.1))
+    }
+}
+
+private extension View {
+    func storeAppearance(on target: Binding<Bool>, if condition: @autoclosure () -> Bool) -> some View {
```
Ecarrion: ⚠️ I think some comments on why is this needed would be highly appreciated by the team!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewController.swift
```diff
@@ -212,20 +212,14 @@ private extension ApplicationLogViewController {
     func logFileWasPressed(in row: Int) {
         let logFileInfo = logFiles[row]
 
-        let identifier = ApplicationLogDetailViewController.classNameWithoutNamespaces
-        guard let appLogDetailVC = UIStoryboard.dashboard.instantiateViewController(identifier: identifier) as? ApplicationLogDetailViewController else {
-            DDLogError("Error: attempted to instantiate ApplicationLogDetailViewController. Instantiation failed.")
-            return
-        }
```
Ecarrion, line 219: ⚠️  Are we missing a storyboard/nib to remove?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewController.swift
```diff
@@ -212,20 +212,14 @@ private extension ApplicationLogViewController {
...
         do {
             let contents = try String(contentsOfFile: logFileInfo.filePath)
             let date = dateFormatter.string(from: logFileInfo.creationDate )
-            appLogDetailVC.logText = contents
-            appLogDetailVC.logDate = date
+            let appLogDetailVC = ApplicationLogDetailViewController(with: contents, for: date)
+            navigationController?.pushViewController(appLogDetailVC, animated: true)
```
Ecarrion: nit: show()
```suggestion
            show(appLogDetailVC, sender: self)
```

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewModel.swift
```diff
@@ -0,0 +1,130 @@
...
+    let date: Date?
+    let text: String
+
+    private let dateFormatter: DateFormatter = {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "YYYY-MM-dd HH:mm:ss:SSS"
+        return dateFormatter
+    }()
+
+    private let datePrefixPattern = try! NSRegularExpression(pattern: "^(\\d{4}/\\d{2}/\\d{2} \\d{2}:\\d{2}:\\d{2}\\:\\d{3})  (.*)")
```
Ecarrion: ⚠️  What is this pattern mattiching? 
Also, could we get rid of the `try!` 😅

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewModel.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    func isLastLine(_ line: ApplicationLogLine) -> Bool {
+        line.id == lastLineID
+    }
+
+    func showShareActivity() {
+        let activityVC = UIActivityViewController(activityItems: [logText], applicationActivities: nil)
+        activityVC.modalPresentationStyle = .popover
+        activityVC.excludedActivityTypes = assembleExcludedSupportTypes()
+        present?(activityVC)
```
Ecarrion: ⚠️  I'm not sure if these view models should be aware of UIKit types.
Do you think having something like an `enum` representing the navigation cases for then sending it in the closure or in a publisher is an overkill?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewModel.swift
```diff
@@ -0,0 +1,130 @@
...
+         * which would be wrong, because the allActivityTypes listed above are stored as enum types.
+         */
+        let supportedTypes = NSSet(objects: UIActivity.ActivityType.copyToPasteboard,
+                                            UIActivity.ActivityType.mail,
+                                            UIActivity.ActivityType.airDrop)
+
+        // Now you can downcast to Set, because the type was preserved on init of the NSSet.
+        activityTypes.minus(supportedTypes as! Set<AnyHashable>)
+
+        return activityTypes.allObjects as! [UIActivity.ActivityType]
```
Ecarrion: I know this is legacy code but I still have nightmares when I see `!` in production code!

Ecarrion (commented):
Left you some questions, mostly `nits` and some marked with ⚠️ that I think are good discussions to have once we dive more into `SwiftUI`

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewModel.swift
koke: I didn't try to update this code at first, but this was probably written when `Set` was much less powerful. I've been able to rewrite it now without all the weird casting 🎉

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewController.swift
koke: Good catch, removed in 3d5621bbb

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewModel.swift
koke: Added some annotation in a724eda62.

`try!` and force unwrapping always makes me uncomfortable, but I think it's sometimes OK. As long as unit tests are passing, this is very unlikely to fail, and if it does, a crash will alert us about it quickly. I see this type of thing comparable to the force-unwrapping we do in our image extensions in `UIImage+Woo`.

The alternative I see used in other places in the app would be to ignore the error, maybe log a message, and act as if the date was invalid instead of the regular expression. This might not crash, but I imagine it could be months before we realized something was off.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
koke: Moved this to the view model instead

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
koke: I agree that's better, moved it in d2eaf9839

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
koke: I don't think that's possible in a [result builder transformation](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html#ID635):

> You can’t use break, continue, defer, guard, or return statements, while statements, or do-catch statements in the code that a result builder transforms.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
koke: I think 9d24e767a should make this better. I still haven't found a way to conditionally apply modifiers that feels nice though 😞

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
koke: I don't have a preference, but if that's what we're doing elsewhere, it makes more sense to do it that way. Added it in 8585d9643

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewController.swift
koke: 🤯 how did I not know about `show()`?! The documentation says it's been around since iOS 8. I'm so very confused 🙃

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewModel.swift
koke: I was hesitant to invest more time in refactoring this, but finally I decided to go all in and implemented a SwiftUI share sheet in b665286e3, which makes for a much nicer API

koke:
Thanks for the thorough review! I think I didn't miss any comment, ready for another look.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
Ecarrion: TIL!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,106 @@
...
+
+    var body: some View {
+        ScrollViewReader { scrollProxy in
+            List(viewModel.lines) { line in
+                VStack(alignment: .leading, spacing: 6) {
+                    line.dateText
+                        .map {
+                            Text($0)
+                                .footnoteStyle()
+                        }
```
Ecarrion: I think this can become
```suggestion
                    Text(line.dateText)
                        .footnoteStyle()
```
Or did I miss something? 😅

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,106 @@
...
+
+/// Hosting controller that wraps an `ApplicationLogDetailView`
+///
+final class ApplicationLogDetailViewController: UIHostingController<ApplicationLogDetailView> {
+    init(viewModel: ApplicationLogViewModel) {
+        super.init(rootView: ApplicationLogDetailView(viewModel: viewModel))
+        self.title = viewModel.title
+        viewModel.present = { [weak self] vc in
+            self?.present(vc, animated: true, completion: nil)
+        }
```
Ecarrion: I think this is no longer needed!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
```diff
@@ -0,0 +1,106 @@
...
+                    viewModel.shareSheetVisible.toggle()
+                }, label: {
+                    Image(systemName: "square.and.arrow.up")
+                })
+                .sharePopover(isPresented: $viewModel.shareSheetVisible) {
+                    ShareSheet(
+                        activityItems: viewModel.activityItems,
+                        excludedActivityTypes: Array(viewModel.excludedActivityTypes)
+                    )
+                }
```
Ecarrion, line 52: Neat!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApplicationLogViewModel.swift
```diff
@@ -0,0 +1,154 @@
...
+        pattern: """
+    ^                                       # Match start of the string only
+        (                                   # First capture group for the timestamp
+            \\d{4}/\\d{2}/\\d{2}            # Date component (YYYY-MM-dd)
+            \\s                             # A single whitespace
+            \\d{2}:\\d{2}:\\d{2}\\:\\d{3}   # Time component (HH:mm:ss:SSS)
+        )                                   # End of first capture group
+        \\s+                                # At least one whitespace character, discard this
+        (.*)                                # Second capture group: rest of text
+""",
```
Ecarrion, line 34: 🤩

Ecarrion (approved):
Looks good! 🚀

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Help/ApllicationLogDetailView.swift
koke: `dateText` is an optional so that won't work. However, I think I got confused about what syntax is allowed in a result builder, and a regular `if let` seems to work, and it's clearer. I replaced it in c8a33d739

PR #4391, opened 2021-06-09
Remove one unused screen from UI tests

mokagio:
While working on integrating [`ScreenObject`](https://github.com/Automattic/ScreenObject) as part of the 2nd HACK week for 2021, I bumped into this unused file.

While I was at it, I added a scheme to run the UI tests from Xcode. I'm surprised there wasn't one already, which makes me think there must be a way to run them that I'm not aware of. If so, what is it? 😅🙏 I noticed there's a test plan for the UI tests, but it seems annoying to have to select it manually in the scheme settings or do a long press on the run button 🤔 Tip: you can cycle through the schemes with <kbd>Ctrl Cmd [</kbd> and <kbd>Ctrl Cmd ]</kbd>. Demo: https://www.loom.com/share/d2838c8d7caf4f668052379ec58f8cd1.

**To test:** build the WooCommerceScreenshots scheme for testing, which builds both UI and screenshot generation tests. As you can see, I run the UI tests in CI and they passed 👌 


thehenrybyrd:
> While working on integrating ScreenObject as part of the 2nd HACK week for 2021, I bumped into this unused file.

I initially wondered if we were using this for login, but I see we're using PrologueScreen now, so I don't see any problems!

> While I was at it, I added a scheme to run the UI tests from Xcode. I'm surprised there wasn't one already, which makes me think there must be a way to run them that I'm not aware of. If so, what is it? 

I always press the play button and then "switch and run tests". Thanks for the cycling tip!



> To test: build the WooCommerceScreenshots scheme for testing, which builds both UI and screenshot generation tests. As you can see, I run the UI tests in CI and they passed 👌

I was able to build WooCommerceScreenshots without any issues.

thehenrybyrd (approved):

rachelmcr:
> While I was at it, I added a scheme to run the UI tests from Xcode. I'm surprised there wasn't one already, which makes me think there must be a way to run them that I'm not aware of. If so, what is it?

You can run the UI tests using the `WooCommerce` scheme and selecting the UI test plan (e.g. in the test navigator or the Product menu > Test Plan). I originally set things up with the same scheme and separate test plans so we could build a single scheme in CI and reuse it for all of the test runs (instead of having to do separate builds for unit and UI test runs). I didn't mind having to manually select the test plan when I did local runs, but setting up a separate scheme to make that easier seems reasonable!

mokagio:
Thanks for the context @rachelmcr 😄 

The two options are not mutually exclusive, so I guess it's just a matter of personal dev workflow. I'm fond of my clickety mechanical keyboard and try to do as much work through sequences of keyboard shortcuts and commands, maybe to too much of an extreme 🤷 



> I originally set things up with the same scheme and separate test plans so we could build a single scheme in CI and reuse it for all of the test runs (instead of having to do separate builds for unit and UI test runs)

This makes total sense. Great use case for test plans 👍

PR #4392, opened 2021-06-09
Fetch ScreenObject Swift package and use it in UI tests  for PrologueScreen 

mokagio:
This adds the new [`ScreenObject`](https://github.com/Automattic/ScreenObject) Swift package to the project and uses it instead of `BaseScreen` in the `PrologueScreen`. I picked that screen because it was part touched in the login tests flow which we are currently working on (so there's fresher knowledge of them).

To test, run the `LoginTests` `XCTestCase` from the WooCommerceScreenshots scheme with the mocks (`rake mocks`) running on the terminal, it should build and pass.

I also run the UI tests in CI and they passed.

Screen Shot 2021-06-15 at 9 14 55 pm

---

~~This PR is still a draft, but I would appreciate your [30% feedback](https://www.linkedin.com/pulse/5-30-90-feedback-why-matters-manley-/) on it.~~

### DIFF
--- a/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -24,7 +24,7 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": "trunk",
-          "revision": "68d92f082bf56358f5dba2021e1f5f81107d123e",
+          "revision": "df8ee4983e674b583b546bb0640ec83428be2465",
           "version": null
         }
       },
--- a/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -24,7 +24,7 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": "trunk",
-          "revision": "3d93c36483884db3fab88aea5c181ce75ceeac1d",
+          "revision": "d4ad399fb4777d390f359ce20865051320bf30b4",
           "version": null
         }
       },
--- a/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -20,6 +20,15 @@
         }
       },
       {
+        "package": "ScreenObject",
+        "repositoryURL": "https://github.com/Automattic/ScreenObject",
+        "state": {
+          "branch": "trunk",
+          "revision": "3d93c36483884db3fab88aea5c181ce75ceeac1d",
+          "version": null
+        }
+      },
+      {
         "package": "SwiftUI-Shimmer",
         "repositoryURL": "https://github.com/markiv/SwiftUI-Shimmer",
         "state": {
--- a/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -24,7 +24,7 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": "trunk",
-          "revision": "d4ad399fb4777d390f359ce20865051320bf30b4",
+          "revision": "68d92f082bf56358f5dba2021e1f5f81107d123e",
           "version": null
         }
       },
--- a/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -23,9 +23,9 @@
         "package": "ScreenObject",
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
-          "branch": "trunk",
+          "branch": null,
           "revision": "df8ee4983e674b583b546bb0640ec83428be2465",
-          "version": null
+          "version": "0.1.0"
         }
       },
       {
--- a/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
+++ b/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
@@ -8425,8 +8425,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/Automattic/ScreenObject";
 			requirement = {
-				branch = trunk;
-				kind = branch;
+				kind = upToNextMajorVersion;
+				minimumVersion = 0.1.0;
 			};
 		};
 		45455E302657C3F300BBB0C4 /* XCRemoteSwiftPackageReference "SwiftUI-Shimmer" */ = {
--- a/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
+++ b/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
@@ -494,6 +494,8 @@
 		31F92DE125E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31F92DE025E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift */; };
 		31FE28C225E6D338003519F2 /* LearnMoreTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31FE28C125E6D338003519F2 /* LearnMoreTableViewCell.swift */; };
 		31FE28C825E6D384003519F2 /* LearnMoreTableViewCell.xib in Resources */ = {isa = PBXBuildFile; fileRef = 31FE28C725E6D384003519F2 /* LearnMoreTableViewCell.xib */; };
+		3FF2247326706AA3008FFA87 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3FF2247226706AA3008FFA87 /* ScreenObject */; };
+		3FF2248A267073AE008FFA87 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3FF22489267073AE008FFA87 /* ScreenObject */; };
 		4506BD712461965300FE6377 /* ProductVisibilityViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4506BD6F2461965300FE6377 /* ProductVisibilityViewController.swift */; };
 		4506BD722461965300FE6377 /* ProductVisibilityViewController.xib in Resources */ = {isa = PBXBuildFile; fileRef = 4506BD702461965300FE6377 /* ProductVisibilityViewController.xib */; };
 		4508BF622660E34A00707869 /* ShippingLabelCarrierAndRatesTopBanner.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4508BF612660E34A00707869 /* ShippingLabelCarrierAndRatesTopBanner.swift */; };
@@ -2668,6 +2670,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				247CE89C2583402A00F9D9D1 /* Embassy in Frameworks */,
+				3FF2247326706AA3008FFA87 /* ScreenObject in Frameworks */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
@@ -2676,6 +2679,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				24C5AC7625A53021008FD769 /* Embassy in Frameworks */,
+				3FF2248A267073AE008FFA87 /* ScreenObject in Frameworks */,
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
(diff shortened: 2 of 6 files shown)


WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -10,6 +10,15 @@
...
       },
+      {
+        "package": "ScreenObject",
+        "repositoryURL": "https://github.com/Automattic/ScreenObject",
+        "state": {
+          "branch": "trunk",
+          "revision": "68d92f082bf56358f5dba2021e1f5f81107d123e",
+          "version": null
+        }
+      },
```
mokagio, line 30: Interesting to see that WooCommerce has the `Package.resolved` file. As I mentioned in the [WordPress PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/16648/files#r647153613) we're not using it in Simplenote.

Looking at it with fresh eyes here. I think tracking it is the appropriate thing to do. This looks to me like the `Podfile.lock` or `package-lock.json` equivalent for the Xcode managed dependency resolution. The Xcode project is the source of truth for the configuration and this is the snapshot of the dependencies as resolved from that configuration that the project expects.

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
```diff
@@ -484,6 +484,8 @@
 		31F92DE125E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31F92DE025E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift */; };
 		31FE28C225E6D338003519F2 /* LearnMoreTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31FE28C125E6D338003519F2 /* LearnMoreTableViewCell.swift */; };
 		31FE28C825E6D384003519F2 /* LearnMoreTableViewCell.xib in Resources */ = {isa = PBXBuildFile; fileRef = 31FE28C725E6D384003519F2 /* LearnMoreTableViewCell.xib */; };
+		3FF2247326706AA3008FFA87 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3FF2247226706AA3008FFA87 /* ScreenObject */; };
+		3FF2248A267073AE008FFA87 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3FF22489267073AE008FFA87 /* ScreenObject */; };
```
mokagio, line 498: Two entries because one is linked against the UI tests framework and the other against the screenshots framework.

WooCommerce/WooCommerceScreenshots/WooCommerceScreenshots.swift
```diff
@@ -1,5 +1,6 @@
-import XCTest
 import Embassy
+import ScreenObject
+import XCTest
```
mokagio, line 3: This is at the bottom to keep imports sorted alphabetically 🤓

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
```diff
@@ -1,20 +1,27 @@
-import Foundation
```
mokagio, line 1: Importing `Foundation` is unnecessary because there is no direct call to a Foundation specific type, such as `URL`.

I wonder why the Xcode Swift new file template still has `import Foundation` in it.

Besides, I'm pretty sure that XCTest brings Foundation with it.

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
```diff
@@ -1,20 +1,27 @@
...
     static let siteAddressButton = "Prologue Self Hosted Button"
 }
 
-final class PrologueScreen: BaseScreen {
-    private let continueButton: XCUIElement
-    private let siteAddressButton: XCUIElement
+final class PrologueScreen: ScreenObject {
+    private let continueButtonGetter: (XCUIApplication) -> XCUIElement = {
+        $0.buttons[ElementStringIDs.continueButton]
+    }
```
mokagio, line 12: In this implementation, I experimented with computing elements at runtime every time, instead of only once.

Now that I've sat on it for a bit, I wonder if that's a good idea or not.

The reason I went for it in the first place was to ensure we always had up to date elements, but I do wonder if that's really necessary given `XCUIElement` is a `NSObject` subclass, that is, an object and therefore assigned by reference. 🤔

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
```diff
@@ -1,20 +1,27 @@
...
-        continueButton = XCUIApplication().buttons[ElementStringIDs.continueButton]
-        siteAddressButton = XCUIApplication().buttons[ElementStringIDs.siteAddressButton]
+    private var continueButton: XCUIElement { expectedElement }
+    private var siteAddressButton: XCUIElement { siteAddressButtonGetter(app) }
 
-        super.init(element: continueButton)
+    init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(
+            expectedElementGetter: { $0.buttons[ElementStringIDs.continueButton] },
+            app: app
```
mokagio, line 23: This is more of a comment about `ScreenObject` that this PR, but the two sort of go hand in hand since the PR exercises this framework that is still in flux.

Here I'm passing a closure to get the element from the app and the app itself because I felt uncomfortable about the possibility of passing an element from one app and then a different app instance. By using the closure instead, we can ensure that `ScreenObject` will look for the element in the same app it stores.

Now that I say this all out loud, I do wonder whether `ScreenObject` even needs to store an `app` instance? It makes sense for its subclasses because they may need to access multiple elements, but maybe the base class can get away without it?

WooCommerce/WooCommerceUITests/Tests/LoginTests.swift
```diff
@@ -13,7 +13,7 @@ final class LoginTests: XCTestCase {
 
     // Login with Store Address and log out.
     func test_site_address_login_logout() {
-        let prologue = PrologueScreen().selectSiteAddress()
+        let prologue = try! PrologueScreen().selectSiteAddress()
```
shiki: Is there a way to make this fail with a prettier error message of what happened? 😅

mokagio: Hey @shiki! I'm glad you caught this `try!` 💪 

It's totally _not_ something I would endorse in the final implementation. Here, I took a shortcut just to get the code to compile.

Extra details from the [matching PR on WordPress iOS](https://github.com/wordpress-mobile/WordPress-iOS/pull/16648#discussion_r647155154):

> I force tried (`try!`) here because I didn't want to keep traversing the call graph. In the "final" implementation, I'd like not to have any `try!`. The reason I'm saying this is because XCTest knows what to do when a `try` `throws` and reports it as a test failure instead of a crash.
>
> _Shameless plug: I wrote of such an instance [today](https://mokacoding.com/blog/how-to-test-async-await-code-in-swift/)._

shiki: > Here, I took a shortcut just to get the code to compile.

Ah make sense! Sorry for the drive-by comment. 😅 I was just curious. This PR looks interesting!

WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
jkmassel: > Looking at it with fresh eyes here. I think tracking it is the appropriate thing to do. This looks to me like the Podfile.lock or package-lock.json equivalent for the Xcode managed dependency resolution. The Xcode project is the source of truth for the configuration and this is the snapshot of the dependencies as resolved from that configuration that the project expects.

Correct!

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
jkmassel: IMHO fetching everything every time is preferable because some screens may not _always_ contain every UI element (consider a table view that might not contain an actual tableview if there's no data, but instead have a nice picture in the center).

Also from a purely aesthetic standpoint, I personally prefer `continueButton` over `continueButtonGetter` 🤷

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
```diff
@@ -1,20 +1,27 @@
...
-    init() {
-        continueButton = XCUIApplication().buttons[ElementStringIDs.continueButton]
-        siteAddressButton = XCUIApplication().buttons[ElementStringIDs.siteAddressButton]
+    private var continueButton: XCUIElement { expectedElement }
+    private var siteAddressButton: XCUIElement { siteAddressButtonGetter(app) }
 
-        super.init(element: continueButton)
+    init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(
+            expectedElementGetter: { $0.buttons[ElementStringIDs.continueButton] },
```
jkmassel, line 22: Presumably this replaces `isLoaded`? I like that a lot

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
jkmassel: IMHO it doesn't hurt to have it since subclasses are likely to need it? Also should help folks avoid using globals to access the app instance all the time?

jkmassel (commented):
This LGTM to me so far!

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
mokagio: Thanks 😄 

This is the code that `isLoaded` uses to fetch the element to use for its check under the hood. See step one [here](https://github.com/Automattic/ScreenObject/blob/df8ee4983e674b583b546bb0640ec83428be2465/Sources/ScreenObject/ScreenObject.swift#L28) and step two [here](https://github.com/Automattic/ScreenObject/blob/df8ee4983e674b583b546bb0640ec83428be2465/Sources/ScreenObject/ScreenObject.swift#L18).

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
mokagio: I don't remember where I wrote about it, but I worry that by running a query on every access, we'll make the tests considerably slower, because of how slow querying `XCUIApplication` is.

But I do agree with your point, which is exactly why I went for computing in the first place.

🤔 I wonder if there's a good way to measure this.

---

Regarding `continueButton` vs `continueButtonGetter`. Here, it's called `-Getter` because otherwise it would name-clash with the actual `continueButton` `XCUIElement` below.

Looking at the code, I also noticed that the `-Getter`s are unused, which is a mistake on my side. A leftover from my experimented mentioned below, on which I then took a step back (as you can see in the [source](https://github.com/Automattic/ScreenObject/blob/df8ee4983e674b583b546bb0640ec83428be2465/Sources/ScreenObject/ScreenObject.swift) for screen object.

I do agree on this with you, too. If we were to go down the "always compute" path, I would have something like this

```swift
// PrologueScreen.swift
var continueButton: XCUIElement { expectedElement }
var otherButton: XCUIElement { app.buttons["other button"] }
// where `app` is the instance property inherited from the superclass, as discussed

// ScreenObject.swift
var expectedElement: XCUIElement { expectedElementGetter(app) }
```

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
mokagio: Very good point. I came to a similar conclusion eventually. `ScreenObject` currently is [storing and exposing `app` as a `let`](https://github.com/Automattic/ScreenObject/blob/trunk/Sources/ScreenObject/ScreenObject.swift#L9-L11).

mokagio:
Bumped from 7.0 to 7.1 because I don't have time to work on this right now. And anyway, this is not user facing stuff so no dramas.

mokagio:
Having to bump milestone again, had no time to finish this yet.

mokagio:
@pachlava @thehenrybyrd @jkmassel @shiki I finally got the time to work on this and updated it. It's now ready to review. I enabled auto-merge, if you're happy with the changes.

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
jkmassel: Now that I'm looking at this again, I wonder how much boilerplate we're adding with the `foo`/`fooGetter` pattern here? Is there a way to avoid this?

WooCommerce/WooCommerceUITests/Screens/Settings/SettingsScreen.swift
```diff
@@ -27,7 +27,7 @@ final class SettingsScreen: BaseScreen {
     }
 
     @discardableResult
-    func logOut() -> PrologueScreen {
+    func logOut() throws -> PrologueScreen {
```
jkmassel, line 30: Nice – it makes sense to have `throws` be applied to any screen method – every single one has to perform IO (to the XCUITest framework).

And no more `try!` 😁

jkmassel (approved):
This still looks solid – I have one concern about too much boilerplate (maybe we could use a property wrapper for this?), but it doesn't have to be addressed immediately.

WooCommerce/WooCommerceUITests/Screens/Login/PrologueScreen.swift
mokagio: Good catch. I didn't clean up properly before enabling auto-merge 🤦‍♂️ 

👉 https://github.com/woocommerce/woocommerce-ios/pull/4770

PR #4395, opened 2021-06-09
Fix Login Not Detecting that the Site is Connected

### LINKED ISSUE #4034: Login Does not Detect That the Site is Connected
This happens if https://github.com/woocommerce/woocommerce-ios/pull/3957 is merged. But it only happens if the user ended up with an XML-RPC error. 

When performing the following steps, the app will incorrectly say “It looks like <SITE> is connected to a different account”. 

1. Log in with a site address that has a blocked XML-RPC. For example, https://do.wpmt.co/ddos. 
2. After the error is displayed, log in with a valid site address. 
3. The app will present the “...is connected to a different account error”.

shiki:
Fixes #4034. 

This just pulls in the WPAuth changes that fix #4034. See https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/pull/596. The findings and solutions are explained in there. 

During the site address entry flow, if there is an error that is displayed below the text field, the subsequent flows may no longer allow the user to log in. Also, the Continue button no longer shows the activity indicator. It looks like the app is doing nothing. Here is a video of this bug: 

https://user-images.githubusercontent.com/198826/121392440-a29e3580-c90c-11eb-84ab-5eae02ba617c.mov

And here is how it looks like after the WPAuth changes are merged:

https://user-images.githubusercontent.com/198826/121393211-5e5f6500-c90d-11eb-9830-6e99ffafe9bb.mov

## Testing

1. Log in with a site address. Use a domain that has XML-RPC errors. For example, "do.wpmt.co/ddos". 
2. Wait for the error to show up. 
3. Change the address to a valid WooCommerce site URL. 
4. Continue entering credentials
5. Confirm that you can successfully log in.

## Author Checklist

- [x] Update WPAuth to point to correct tag (assuming the WPAuth PR is accepted and merged).


ealeksandrov (approved):
Works well. Thanks for digging into it and fixing! :shipit:

shiki:
Thank you, Evgeny!

PR #4396, opened 2021-06-10
Hack Week: improve the loading state in the Order Detail

pmusolino:
As discussed here p91TBi-5f1, we'd like to improve the loading state of the Order Detail screen.
So, I implemented the solution proposed by @Garance91540 here: p91TBi-5f1#comment-4949.

One of the main issues that we want to solve, is to communicate to our users, displaying a refresh control on top of the table view, that the screen is still loading its content.

#### Case 1
1. Open an Order Detail.
2. You should see the detail of the order, and an activity indicator on top (the same we use for the pull to refresh action).
3. The activity indicator should be hidden when all the requests finish.
4. The pull to refresh action should work like before.
5. If you try to open a sub-screen starting from Order Detail when you do a "pop" action, you should see the indicator at the top again.

#### Case 2
1. Open an Order Detail.
2. Open a sub-screen starting from Order Detail (like tapping "View Billing Information") 
3. Tap back.
4. You should see the indicator at the top again, while we are refreshing the data.

## Video

https://user-images.githubusercontent.com/495617/121526997-3bf74580-c9fa-11eb-80df-62cd1830a20f.mp4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
@@ -53,10 +53,6 @@ final class OrderDetailsViewController: UIViewController {
     ///
     private var cardReaderAvailableSubscription: Combine.Cancellable? = nil
 
-    /// Loading view displayed while the order is loading
-    ///
-    private let loadingView = LoadingView(waitMessage: "Order is loading...")
-
     // MARK: - View Lifecycle
 
     /// Create an instance of `Self` from its corresponding storyboard.
@@ -76,7 +72,6 @@ final class OrderDetailsViewController: UIViewController {
         configureEntityListener()
         configureViewModel()
         updateTopBannerView()
-        loadingView.showLoader(in: view)
 
         // FIXME: this is a hack. https://github.com/woocommerce/woocommerce-ios/issues/1779
         reloadTableViewSectionsAndData()
@@ -84,8 +79,10 @@ final class OrderDetailsViewController: UIViewController {
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+
+        programaticallyBeginRefreshing()
         syncEverything { [weak self] in
-            self?.loadingView.hideLoader()
+            self?.refreshControl.endRefreshing()
         }
     }
 
@@ -276,7 +273,7 @@ private extension OrderDetailsViewController {
 
 // MARK: - Action Handlers
 //
-extension OrderDetailsViewController {
+private extension OrderDetailsViewController {
 
     @objc func pullToRefresh() {
         ServiceLocator.analytics.track(.orderDetailPulledToRefresh)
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -278,10 +273,32 @@ private extension OrderDetailsViewController {
 
 // MARK: - Action Handlers
 //
-extension OrderDetailsViewController {
+private extension OrderDetailsViewController {
 
     @objc func pullToRefresh() {
         ServiceLocator.analytics.track(.orderDetailPulledToRefresh)
+        refreshControl.beginRefreshing()
```
allendav, line 297: Do we need to be concerned if the user pulls to refresh while a programmatic refresh is happening?

allendav:
@ctarda - it would be nice to trigger this after collecting a payment - what do you think?

Garance91540 (changes requested):
Ah Paolo I misunderstood when you asked if we should use the same indicator! I still want us to display the banner, but I just thought you wanted to use this indicator instead of the one I used in my designs. https://d.pr/i/bBX0gR

As pull to refresh and loading are different (although close) actions, I don't think they should be displayed the same way. (The banner appears as the page is loading automatically vs the spinner appears when you trigger a refresh).

allendav (approved):
Tests nicely. One question about simultaneous manual and programatic refreshes. (Which I tried but it didn't seem to be a problem nor precipitate an error.)

IMG_0074

pmusolino:
Thank you both @allendav and @Garance91540. I'll update the design using the suggestion of Garance.
@Garance91540 do you think that in the case of dark mode enabled, we should use the white background for the banner and the black for the text?

pmusolino:
I updated the design, implementing the top loader view as suggested by @Garance91540, which can be used as a banner and that is embedded inside a stack view where we embed also the table view. 

Can you re-check it, please? @Garance91540 @allendav 🙇 

## Screenshots
| Light Mode            |  Dark Mode |
Simulator Screen Shot - iPhone 12 Pro - 2021-06-11 at 16 46 28 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-11 at 16 46 42



## Video
https://user-images.githubusercontent.com/495617/121707239-a7164a00-cad6-11eb-8a9c-8011cf011737.mp4

Garance91540 (approved):
Looking neat!

Garance91540:
Now that we created this component, I wonder if we can generalize its usage to any update that happens after screens have been loaded, for example for actions that happen on the web and update the app automatically. One that I have in mind that I think Jirka or Allen brought up is this one: 

https://cloudup.com/c1LrT9DG-6b

If I'm on the Order List and I receive an order, the new order will pop out automatically, but I have no feedback that this just happened. Could we add the banner in this instances?

pmusolino:
@Garance91540 In the order list, we update the content every time the user navigates between different screens (since we have no way, at the moment, to understand if there are new orders compared to those previously downloaded). In that case, we would show the loader every time that screen loads.

ctarda:
I just wanted to drop a drive-by comment... This is great work and a significant improvement @pmusolino . Thanks for doing this.

allendav (approved):
Tests fine. I confirmed the pull-to-refresh is as it always has been (a spinner) and the initial load now looks like:

PR #4413, opened 2021-06-14
[Mobile Payments] Add currency to order total in modals when capturing a payment

### LINKED ISSUE #4356: [Mobile Payments] [UI] Modals do not reflect currency
Brought up in the review of #4331

https://d.pr/i/bHk4yn

We present the total amount to be charged in several modals, but it does not include the currency

ctarda:
Closes #4356 

| Before | After |
| ---- | ---- |
|  |  |
|  |  |
|  |  |

## Changes
Format the total amount being charged using the store currency settings. 

## How to test
* Attempt to charge an order. 
* Notice the amount, and I the amount includes the currency.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
@@ -594,9 +594,14 @@ private extension OrderDetailsViewController {
     }
 
     private func collectPaymentForCurrentOrder() {
+        let currencyFormatter = CurrencyFormatter(currencySettings: ServiceLocator.currencySettings)
+        let currencyCode = ServiceLocator.currencySettings.currencyCode
+        let unit = ServiceLocator.currencySettings.symbol(from: currencyCode)
+        let value = currencyFormatter.formatAmount(viewModel.order.total, with: unit) ?? ""
+
         paymentAlerts.readerIsReady(from: self,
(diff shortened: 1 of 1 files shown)


ghost (approved):

ealeksandrov (approved):
LGTM :shipit:

ctarda:
Thanks for the review @ealeksandrov 🙇🏼

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -594,9 +594,14 @@ private extension OrderDetailsViewController {
     }
 
     private func collectPaymentForCurrentOrder() {
+        let currencyFormatter = CurrencyFormatter(currencySettings: ServiceLocator.currencySettings)
+        let currencyCode = ServiceLocator.currencySettings.currencyCode
```
allendav, line 598: Orders can have different currency than the store. Admittedly they don't usually, but they can. Please open an issue to use the currency from the order.

https://github.com/woocommerce/woocommerce/blob/b19500728b4b292562afb65eb3a0c0f50d5859de/includes/abstracts/abstract-wc-order.php#L279-L287

BTW, separate but related -- for card-present payments -- we may eventually need to use the PaymentGatewayAccount supportedCurrencies property to determine whether or not we can collect a payment. I've opened issue #4425 for that.

Issue #4126, opened 2021-05-06
Add-Ons: De-group items with different add-ons


Ecarrion:
**Note: This can't be implemented until the product list and the package list are separated p91TBi-566-p2**

## Describe the bug

Orders with multiple items(same product) but with different add-on information are grouped. By grouping them we can't show the "View Add-On" button for the correct items.

## To Reproduce
1. Make an order with one product with add-ons and with the same product without add-ons.
2. Go the order in the app
3. See that items are grouped and the "view add-on" is sometimes visible

## Expected behavior
- Items that have different add-on information are not grouped and the "view add-on" button can be shown for the specific item.

[Possible fix](https://github.com/woocommerce/woocommerce-ios/compare/issue/differentiate-items-with-add-ons?expand=1)

## Screenshots

Current | Expected
--- | ---
<img width="400" alt="before-1" src="https://user-images.githubusercontent.com/562080/117303388-42663080-ae42-11eb-8089-72d4e74e8deb.png"> | <img width="399" alt="after-1" src="https://user-images.githubusercontent.com/562080/117303383-41350380-ae42-11eb-831c-ac36fe540bd5.png">

## Mobile Environment
Please include:
- Device: Simulator
- iOS version: 14.2

Ecarrion:
Fixed in #4445

PR #4416, opened 2021-06-15
Include attributes for aggregating order items

### LINKED ISSUE #4412: Order Details → Variations With Different Attributes are Combined Together
**Describe the bug**
When a customer orders multiple items of a variable product but with different attributes, the app only displays that variation once in the Order Details.

This basically makes the merchant unable to fulfill the order properly. 

- Same Android issue: https://github.com/woocommerce/woocommerce-android/issues/3811
- PR in Android: https://github.com/woocommerce/woocommerce-android/pull/4193
- Android discussion: p1623648152420100-slack-C6H8C3G23

**To Reproduce**

1. Create a variable product. 
2. Add “Color” as an attribute with “Blue|Green|Yellow” as the values. 
3. Create a **single** variation of that variable product. Choose “Any” as the attribute option. 
4. Create a single order for the product. 
  - Add 1 Blue
  - Add 2 Green
  - Add 5 Yellow
5. Check out and submit the order. 
6. In the app, view the order. You'll notice that the order details only shows one line of the product, with `8` as the quantity. And the attribute looks like the first color that you picked (e.g. Blue). 

**Screenshots**

This is the order in WCAdmin: 

image

This is what is shown in the app (`develop` 431b134b8). 

**Expected behavior**

**Mobile Environment**
Please include:
- Device: Simulator
- iOS version: 14.4
- WooCommerce iOS version: `develop` 431b134b8

**WordPress Environment**
<details>
```
```
</details>

koke:
Fixes #4412

In #4412, it was reported that when a customer orders multiple items of a variable product but with different attributes, the app only displays that variation once in the Order Details.

I think the main cause is that we are aggregating order items based only on product and variation ID so, if there are multiple items of the same product with a single variation, but different attributes, they will be grouped as one.

The fix for this seems pretty simple, which makes me wonder if there is a lot more to it that I'm missing, so please review accordingly. I'm not too familiar with this area of the app yet.

## To test

Follow the steps in #4412:

1. Create a variable product.
2. Add “Color” as an attribute with “Blue|Green|Yellow” as the values.
3. Create a single variation of that variable product. Choose “Any” as the attribute option.
4. Create a single order for the product.
    - Add 1 Blue
    - Add 2 Green
    - Add 5 Yellow
5. Check out and submit the order.
6. In the app, view the order. It should show three different items, one for each color purchased.

On the web:

Screen Shot 2021-06-15 at 09 43 35

Before | After
-|-
Screen Shot 2021-06-15 at 09 44 15|Screen Shot 2021-06-15 at 09 44 59

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,10 @@
 *** PLEASE FOLLOW THIS FORMAT: [<priority indicator, more stars = higher priority>] <description> [<PR URL>]
 
+7.0
+-----
+- [*] Fix: Orders for a variable product with different configurations of a single variation will now show each order item separately. [https://github.com/woocommerce/woocommerce-ios/pull/4416]
+
+
 6.9
 -----
 - [*] Order Detail: now we display a loader on top, to communicate that the order detail view has not yet been fully loaded. [https://github.com/woocommerce/woocommerce-ios/pull/4396]
--- a/WooCommerce/Classes/ViewModels/Order Details/Aggregate Order Items/AggregateOrderItem.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/Aggregate Order Items/AggregateOrderItem.swift
@@ -65,6 +65,7 @@ extension AggregateOrderItem: Hashable {
     public func hash(into hasher: inout Hasher) {
         hasher.combine(productID)
         hasher.combine(variationID)
+        hasher.combine(attributes)
     }
 }
 


Ecarrion:
Hi @koke thanks for working on this. 

Unfortunately, we can't merge it yet because this change breaks the packages bought section on the order details screen. 

The problem was documented in p91TBi-566 and the issue is being tracked in https://github.com/woocommerce/woocommerce-ios/issues/4126

Ecarrion (commented):
Hi @koke, thanks for working on this. 

Unfortunately, we can't merge it yet because this change breaks the packages bought section on the order details screen. 

The problem was documented in p91TBi-566 and the issue is being tracked in https://github.com/woocommerce/woocommerce-ios/issues/4126

PR #4417, opened 2021-06-15
[Tooling] Run rubocop on Fastfile

AliSoftware:
### Why?

As part of trying to ensure I didn't leave any other typo in the ruby code while working on https://github.com/woocommerce/woocommerce-ios/pull/4404, I ran `rubocop` on the `Fastfile`.

I can confirm (\cc @mokagio) that if there is a Ruby syntax error in the Fastfile, like a missing parenthesis, rubocop will error and fail to parse the file, so at least that was a good test. Rubocop doesn't warn if I accidentally use `FIle` instead of `File` tho, as it only validates syntax (balancing of parens, of do…end, etc) but does not check if classes and methods exists (for all it knows there could be a class named `FIle` in one of our Gem 🤷 ).

Since I ran `rubocop --auto-correct` while I was at it, to reduce the noise when checking the output, I figured I might as well push and open a PR for it.

### Notes

This PR **does NOT fix ALL the rubocop violations**. Only the ones that got fixed by `--auto-correct`. There are still some violations in the code of our Fastfile, mainly `Naming/VariableName` to suggest us to use snake_case, `Style/StringConcatenation`, `Lint/UselessAssignment`, `Style/MutableConstant` and `BlockLength`/`LineLength` violations that might be worth looking into at some point in the future, but I didn't plan to focus on this today, just to commit the auto-correct I happened to do.

* This PR sits on top of https://github.com/woocommerce/woocommerce-ios/pull/4404. They can be merged in any order, GitHub will auto-update the target of this PR if 4404 to `develop` gets merged first
* It might be worth turning off whitespaces in the diff for easier comparison


Podfile
```diff
@@ -119,7 +115,7 @@ def networking_pods
   pod 'CocoaLumberjack', '~> 3.5'
   pod 'CocoaLumberjack/Swift', '~> 3.5'
 
-  pod 'Sourcery', '~> 1.0.3', :configuration => 'Debug'
+  pod 'Sourcery', '~> 1.0.3', configuration: 'Debug'
```
mokagio, line 118: Oh, now I'm curious to see what we use Sourcery for 🤓

Rakefile
```diff
@@ -1,104 +1,103 @@
...
+XCODE_SCHEME = 'WooCommerce'
+XCODE_CONFIGURATION = 'Debug'
 
 require 'fileutils'
 require 'tmpdir'
 require 'rake/clean'
 require 'yaml'
 require 'digest'
-PROJECT_DIR = File.expand_path(File.dirname(__FILE__))
+PROJECT_DIR = __dir__
```
mokagio, line 11: ❤️

Scripts/update-translations.rb
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env ruby
-# encoding: utf-8
```
mokagio, line 2: RuboCop removed this because UTF-8 is the default encoding in [Ruby 2+](https://www.ruby-lang.org/en/news/2013/02/24/ruby-2-0-0-p0-is-released/).

fastlane/Fastfile
```diff
@@ -548,100 +541,97 @@ GH_ORG_NAME="woocommerce"
...
-      target_files: files,
-      locales: metadata_locales,
-      source_locale: "en-US",
-      download_path: "./fastlane/metadata"
-    )
+    gp_downloadmetadata(project_url: 'https://translate.wordpress.com/projects/woocommerce%2Fwoocommerce-ios%2Frelease-notes/',
+                        target_files: files,
+                        locales: metadata_locales,
+                        source_locale: 'en-US',
+                        download_path: './fastlane/metadata')
```
mokagio, line 631: Is the Ruby standard to organize the code like this, or is this just a result of the first param being "inline" with the method name, do you know?

I find this better to read

```suggestion
    gp_downloadmetadata(
      project_url: 'https://translate.wordpress.com/projects/woocommerce%2Fwoocommerce-ios%2Frelease-notes/',
      target_files: files,
      locales: metadata_locales,
      source_locale: 'en-US',
      download_path: './fastlane/metadata'
    )
```

Rationale: methods with names of different length. will indent in the same way:

```ruby
short(a: 1,
      b: 2)

long_long_long_long(a: 1,
                    b: 2)
```

vs 

```ruby
short(
  a: 1,
  b: 2
)

long_long_long_long(
  a: 1,
  b: 2
)
```

fastlane/Fastfile
```diff
@@ -711,130 +700,130 @@ GH_ORG_NAME="woocommerce"
...
   end
-
 end
 
-  def fastlane_directory()
-    File.expand_path File.dirname(__FILE__)
-  end
+def fastlane_directory
+  __dir__
+end
```
mokagio: This is beyond the scope of this PR and what RuboCop can auto correct, but, eventually, it would be good to have this and all the other methods returning constant values as actual constants 😄 

```suggestion
FASTLANE_DIRECTORY = __dir__
```

We'll get there 💪

mokagio (approved):
This LGTM. Thank you for doing this 💘 

As CI [points out](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/12449/workflows/167981c4-5179-4f93-982b-5eda355bc725/jobs/22744), the `Podfile` hash in the `Podfile.lock` has changed because of the changes here, but it's not tracked in the repo.

I committed the fix and pushed it to this branch directly, because it's a no-brainer fix.

Podfile
AliSoftware: There's a dedicated Codegen target in the Xcode project if you are curious and want to take a look 😛

fastlane/Fastfile
AliSoftware: I agree I prefer your way too.

I think/hope this is mostly to align how the existing first line was indented initially; what happens if you fix the indentation manually and try to rerun rubocop?

fastlane/Fastfile
AliSoftware: Good point and nice catch

AliSoftware: Actually probably fixing this one is worth it in this PR 🤔

fastlane/Fastfile
mokagio: I sent the first argument on a new line then run `rubocop -a`.

```diff
     gp_downloadmetadata(
       project_url: 'https://translate.wordpress.com/projects/woocommerce%2Fwoocommerce-ios%2Frelease-notes/',
-                        target_files: files,
-                        locales: metadata_locales,
-                        source_locale: 'en-US',
-                        download_path: './fastlane/metadata')
+      target_files: files,
+      locales: metadata_locales,
+      source_locale: 'en-US',
+      download_path: './fastlane/metadata'
+    )
```

So, yeah, RuboCop fixes it in the way we both like. It also moved the closing `)` to a new line by itself, which was nice.

fastlane/Fastfile
mokagio: 🤷‍♂️ up to you

fastlane/Fastfile
AliSoftware: 🎉 👌

jkmassel: This is way better, thanks!!

Gemfile
```diff
@@ -1,11 +1,11 @@
 source 'https://rubygems.org'
 
-gem 'rake'
 gem 'cocoapods', '~> 1.10'
-gem 'xcpretty-travis-formatter'
 gem 'dotenv'
-gem 'xcode-install'
 gem 'fastlane', '~> 2'
+gem 'rake'
```
jkmassel: Feels weird not to have version constraints on any of these. While we're reordering, WDYT about adding them?

jkmassel (changes requested):
I left one suggestion for a thing we should _probably_ do while we're at it

In a perfect world I'd love a rubocop rule that enforces our preferred multiline argument passing format, but that could be for another day :)

Gemfile
AliSoftware: Good point, will add that

AliSoftware: Fixed in 91ccf8a

fastlane/Fastfile
AliSoftware: @mokagio Fixed in f6121fe

Feel free to give it a quick check and merge the PR yourself if all looks good.

AliSoftware:
@jkmassel Can I have an approval now that your query was addressed? I'd love to have this merged soon-ish because these changes conflicts with changes I want to do about my hotfix lanes work (and any PR started from develop touching the Fastfile will likely end up in conflicts too until this is merged)

jkmassel (approved):
I think it'd still be nice to have version constraints on `dotenv`, `xcode-install`, and `xcpretty-travis-formatter` (why do we need this last one? We maybe don't?)

Don't want to block though 🙂

AliSoftware:
Oh I missed that you also flagged the others, only addressed `rake`. But yeah I think this can go into a separate PR given it's unrelated.

(Note that for the 3 you mentioned 2 are already dependencies of fastlane itself and one seems obsolete indeed. We could investigate this at some point I'll keep it in mind.)

PR #4419, opened 2021-06-15
Refactor SettingsAction to use Result

ealeksandrov:
## Description

This PR updates SettingsAction + related Store and Remote to use `Result`, so the state of response is more limited with less optional parameters.

## Test

2. Run app to trigger `RequirementsChecker` and verify there are no errors.

---

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/SiteAPIRemote.swift
+++ b/Networking/Networking/Remote/SiteAPIRemote.swift
@@ -1,5 +1,4 @@
 import Foundation
-import Alamofire
 
 
 /// Site API: Remote Endpoints
@@ -13,7 +12,7 @@ public class SiteAPIRemote: Remote {
     ///   - siteID: Site for which we'll fetch the API settings.
     ///   - completion: Closure to be executed upon completion.
     ///
-    public func loadAPIInformation(for siteID: Int64, completion: @escaping (Swift.Result<SiteAPI, Error>) -> Void) {
+    public func loadAPIInformation(for siteID: Int64, completion: @escaping (Result<SiteAPI, Error>) -> Void) {
         let path = String()
         let parameters = [ParameterKeys.fields: ParameterValues.fieldValues]
         let request = JetpackRequest(wooApiVersion: .none, method: .get, siteID: siteID, path: path, parameters: parameters)
--- a/Networking/Networking/Remote/SiteAPIRemote.swift
+++ b/Networking/Networking/Remote/SiteAPIRemote.swift
@@ -13,7 +13,7 @@ public class SiteAPIRemote: Remote {
     ///   - siteID: Site for which we'll fetch the API settings.
     ///   - completion: Closure to be executed upon completion.
     ///
-    public func loadAPIInformation(for siteID: Int64, completion: @escaping (SiteAPI?, Error?) -> Void) {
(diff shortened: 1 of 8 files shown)


Networking/Networking/Remote/SiteAPIRemote.swift
```diff
@@ -13,7 +13,7 @@ public class SiteAPIRemote: Remote {
     ///   - siteID: Site for which we'll fetch the API settings.
     ///   - completion: Closure to be executed upon completion.
     ///
-    public func loadAPIInformation(for siteID: Int64, completion: @escaping (SiteAPI?, Error?) -> Void) {
+    public func loadAPIInformation(for siteID: Int64, completion: @escaping (Swift.Result<SiteAPI, Error>) -> Void) {
```
shiki: The `import Alamofire` above is not actually necessary. If we remove that, then we can remove the `Swift.` prefix here. 

```suggestion
    public func loadAPIInformation(for siteID: Int64, completion: @escaping (Result<SiteAPI, Error>) -> Void) {
```

Networking/NetworkingTests/Remote/SiteAPIRemoteTests.swift
```diff
@@ -22,34 +22,39 @@ class SiteAPIRemoteTests: XCTestCase {
 
     /// Verifies that loadAPIInformation properly parses the sample response.
     ///
-    func testLoadGeneralSettingsProperlyReturnsParsedSettings() {
+    func test_loadAPIInformation_properly_returns_parsed_settings() throws {
```
shiki, line 25: Thanks for updating the test names and adding Given-When-Then! 🙏

Yosemite/Yosemite/Stores/SettingStore.swift
```diff
@@ -82,9 +82,9 @@ private extension SettingStore {
     /// Retrieves the site API information associated with the provided Site ID (if any!).
     /// This call does NOT persist returned data into the Storage layer.
     ///
-    func retrieveSiteAPI(siteID: Int64, onCompletion: @escaping (SiteAPI?, Error?) -> Void) {
-        siteAPIRemote.loadAPIInformation(for: siteID) { (siteAPI, error) in
-            onCompletion(siteAPI, error)
+    func retrieveSiteAPI(siteID: Int64, onCompletion: @escaping (Result<SiteAPI, Error>) -> Void) {
+        siteAPIRemote.loadAPIInformation(for: siteID) { result in
+            onCompletion(result)
```
shiki: I think this can be: 

```suggestion
        siteAPIRemote.loadAPIInformation(for: siteID, completion: onCompletion)
```

🤪

shiki (approved):
LGTM! I just left unimportant suggestions. 🤪

Yosemite/Yosemite/Stores/SettingStore.swift
ealeksandrov: Wow, nice catch! 😄 Fixed in 4ecacfd.

Networking/Networking/Remote/SiteAPIRemote.swift
ealeksandrov: Right! Fixed in 2a2f3fa.
I'll make a note to self to revisit other remotes to remove unnecessary imports.

ealeksandrov:
Thank you for review and great suggestions @shiki!

Issue #4088, opened 2021-04-30
Shipping labels: view order summary and Woo discount


pmusolino:
Display Shipping Label order summary, and the Woo discount.

- [x] Add the UI of the order summary and discount.
- [x] Calculate the price and discount from the selected shipping carriers.
- [x] Display the banner when the user taps the info icon on WooCommerce Discount.

<img width="595" alt="Schermata 2021-04-30 alle 12 47 28" src="https://user-images.githubusercontent.com/495617/116685215-444b7200-a9b2-11eb-8952-ada6aebbfd0c.png">

Issue #4343, opened 2021-06-02
Handle different font sizes in Shipping Label Form Creation


pmusolino:
It seems that some labels in the "Create shipping label" screen are not properly resizing when the font changes.

<img width="405" alt="120341014-224c5480-c2bc-11eb-85ba-15bb241a5e3e" src="https://user-images.githubusercontent.com/495617/120497378-a25ae300-c3be-11eb-9636-9cca25f06956.png">

PR #4420, opened 2021-06-15
Shipping Labels: order summary and woo discount section

pmusolino:
Part of #4088 

## Description
As part of Shipping Labels M3, in this PR I implemented the order summary that displays the Subtotal, the WooCommerce discount (if available), the Order Total, a switch for completing the order, and the button for purchasing the shipping label.
I've implemented a new cell `ShippingLabelSummaryTableViewCell` that displays all the information needed when the user fulfilled all the fields in the Shipping Labels form.

## Testing
5. Complete all the fields, and try to select the more expensive option under "Shipping Carriers and Rates" (in theory, this makes your shipping label eligible for our WooCommerce discount).
6. You should be able to see the "Shipping Label Order Summary" section at the end of the Shipping Label Form, and the "WooCommerce Discount" row should be visible.
7. Try to change the "Shipping Carriers and Rates" option, selecting the cheapest one. -> You shouldn't be able to see anymore the discount row.

PS: There is no guarantee that the cheapest carrier will not guarantee a discount, and the reverse is also true, an expensive carrier does not guarantee a discount. So do several tests if you are not able to see the discount row.

## Screenshots
| Order Summary with discount             |  Order Summary without discount |
Simulator Screen Shot - iPhone 12 Pro - 2021-06-15 at 16 53 45 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-15 at 16 55 48

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -224,6 +224,73 @@ final class ShippingLabelFormViewModel {
...
+    }
+
+    /// Returns, if available, the discount under the Order Summary.
+    ///
+    func getDiscount() -> String? {
+        guard let selectedRate = selectedRate else {
+            return nil
+        }
+
+        var rate: Double = selectedRate.retailRate - selectedRate.rate
```
rachelmcr: I noticed that the discount appears as a positive number if I select an expensive shipping rate without adding any signature option. It looks like `retailRate` and `rate` should be switched here:

```suggestion
        var rate: Double = selectedRate.rate - selectedRate.retailRate
```

rachelmcr (commented):
For the most part, this looks and works great! I confirmed the discount line is shown/hidden as expected, and the summary section only appears when all of the form sections are complete.

There's a small issue with the WooCommerce discount showing as a positive value when a signature option isn't selected for the shipping rate (see suggested code change below).

I also noticed a couple issues with accessibility and localization, but we could probably address those separately. (We should probably add a task to the Shipping Labels M3 project to test things like accessibility and localization through the whole flow; what do you think?) Here's what I noticed:

* Accessibility: With large font sizes, "WooCommerce Discount" gets cut off. (I also noticed that with _really_ large font sizes, the discounted amount gets pushed off the screen.) Should the titles wrap when they don't fit on one line?



* Localization: With right-to-left languages (such as Arabic), the titles aren't aligned correctly in the summary cell rows. Looking just at the fully translated rows, you can see that the titles are left-aligned instead of right-aligned:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: Good catch!

pmusolino:
Thank you @rachelmcr for your review.

> There's a small issue with the WooCommerce discount showing as a positive value when a signature option isn't selected for the shipping rate (see suggested code change below).

Fixed 👍 

> I also noticed a couple issues with accessibility and localization, but we could probably address those separately. (We should probably add a task to the Shipping Labels M3 project to test things like accessibility and localization through the whole flow; what do you think?)

We already have an issue opened about the Shipping Label Form with large font sizes: https://github.com/woocommerce/woocommerce-ios/issues/4343 we will fix it all together when we will work on this issue. 👍 

About the RTL languages issue, good catch, I fixed it 🙌

rachelmcr (approved):
Looks good! :shipit:

PR #4421, opened 2021-06-15
Refactor AccountAction, AccountStore and AccountRemote to use Result

ealeksandrov:
## Description

This PR updates AccountAction + related Store and Remote to use Result, so the state of response is more limited with less optional parameters.

## Test

2. Run app and do a cycle of logout-login, verify that account actions work well.

---

Update release notes:


ctarda (approved):
This looks much cleaner

I logged in and out, switched stores, and things seem to work properly.

:shipit:

PR #4411, opened 2021-06-14
[Mobile Payments] UI review for reader connect instructions

koke:
Part of #4361, this updates:

- Headline style for the header
- Illustration for both light and dark mode

I tried changing the numbered items background, but I'm having issues with dark mode, so I'm putting up this PR with the items already completed that are more critical

## To test

Go to Settings > Manage Card Readers

Light|Dark
-|-
Screen Shot 2021-06-14 at 12 40 07|Screen Shot 2021-06-14 at 12 40 17

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #4361, opened 2021-06-03
[Mobile payments] [UI] Update instructions to connect to a reader


ctarda:
Brought up during the review of #4331. https://github.com/woocommerce/woocommerce-ios/pull/4331#issuecomment-853989884

<img width="712" alt="120666304-c38a0500-c48c-11eb-9981-8445eafb50c7" src="https://user-images.githubusercontent.com/2722505/120691447-77cb6180-c474-11eb-9b95-4ebcd1a2a7b4.png">

The comments in red would be what we should try really hard to get into beta 1. The rest can wait, but since we are at it...

koke:
Putting this back in the To Do column, as the remaining items will be implemented better after #4449

PR #4422, opened 2021-06-15
[Mobile Payments] UI Improvements to card reader instructions, part 2

### LINKED ISSUE #4374: [Mobile Payments] Connect Your Card Reader Background mismatch on iPad on dark mode
odd-background

koke:
Part of #4361
Fixes #4374 4374

This continues the work in #4411, and addresses the following feedback:

- Use the `info-outline` gridicon for the _learn more_ icon. Apparently, storyboards can't use images coming from Gridicons, so I left that empty and set it in code, but please let me know if there's a better way.
- Use footnote style for the _learn more_ text. I'm not sure I appreciate the difference, but the style is set.
- Use correct background colors for the whole screen, as pointed out in https://github.com/woocommerce/woocommerce-ios/pull/4411#pullrequestreview-684026224. This felt hacky because I had to apply the background to each cell individually, but we're using a table view to build a view that doesn't look like a table, so I guess it's OK.
- Remove the underline from the link in _learn more_

There are still a few items in #4361 that haven't been addressed yet, as I'm having a harder time re-learning auto layout after all this time 😅. I'm not sure if the fact that we're using a table view is problematic here, or I just haven't figured those out yet.

## To test

I've tested this on Light/Dark mode, and connecting from Settings/connecting from tapping Collect Payment within an order. Depending on how you get to this screen, it is presented in a navigation controller (from settings) or in a modal (from orders). The `systemBackground` applied basically makes the views transparent, so the modal will have the default background color, while the one with the navigation controller will have a black background.

### From settings (Light)

Before|After
-|-
before-nav-light|after-nav-light

### From settings (Dark)

Before|After
-|-
before-nav-dark|after-nav-dark

### From orders (Light)

Before|After
-|-
before-modal-light|after-modal-light

### From orders (Dark)

Before|After
-|-
before-modal-dark|after-modal-dark

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
@@ -192,29 +192,34 @@ private extension CardReaderSettingsUnknownViewController {
 
     private func configureHeader(cell: HeadlineTableViewCell) {
         cell.headlineLabel?.text = Localization.connectYourCardReaderTitle
+        cell.backgroundColor = .systemBackground
         cell.selectionStyle = .none
     }
 
     private func configureImage(cell: ImageTableViewCell) {
         cell.detailImageView?.image = UIImage(named: "card-reader-connect")
+        cell.backgroundColor = .systemBackground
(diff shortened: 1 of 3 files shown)


Ecarrion (commented):
Looks good to me! 🚀

Garance91540 (approved):
👏

Issue #3948, opened 2021-04-09
[Mobile Payments] Improve the logic that converts the order amount into the currency's smallest unit.


ctarda:
Multiplying by 100, as we do in PaymentIntentParameters+Stripe might be problematic in certain locales

ctarda:
Closing for now. Will re-open when needed (whenever we start supporting currencies that require it)

PR #4441, opened 2021-06-16
[Mobile Payments] Avoid loss of cents on receipt totals

### LINKED ISSUE #4434: [Mobile Payments] Card Present Receipts Shows Incorrect Order Total
Props @malinajirka - For orders whose totals are not whole dollars, Card Present Receipts shows the wrong amount. For example, on this sale of a $2.00 product:

Screen Shot 2021-06-16 at 6 47 09 AM

$0.21 in taxes was omitted from the receipt and receipt total:

Screen Shot 2021-06-16 at 6 46 50 AM

This is a must fix for beta @ctarda

allendav:
Fixes #4434 

To Test:
- Add an item with a price ending in something other than .00 to your cart and order it
- Proceed to pay for it with card present payments
- Ensure the correct amount is shown in both printed and email receipts
- Avoid using prices that trigger payment declines: https://stripe.com/docs/terminal/testing#physical-test-card

Also considered:
- I would have also liked to have added a unit test for this, but that would have required investment in HTML parsing and since we are exploring other options for receipt rendering right now, I elected not to,

Before:

After:

Update release notes:


Gemfile.lock
```diff
@@ -155,6 +155,8 @@ GEM
       trainer
       xcodeproj
       xctest_list (>= 1.2.1)
+    fastlane-plugin-trainer (0.4.1)
```
ctarda: There was some discussion on slack about why this happens, but there should not be changes in the Gemfile.lock in the diff

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -111,9 +111,15 @@ private extension ReceiptRenderer {
     private func formattedAmount() -> String {
         // We should use CurrencyFormatter instead for consistency
         let formatter = NumberFormatter()
-        formatter.minimumFractionDigits = 2
-        formatter.maximumFractionDigits = 2
-        return formatter.string(for: parameters.amount / 100) ?? ""
+
+        let fractionDigits = 2 // TODO - support non cent currencies like JPY
```
ctarda: Would you mind linking to #3948 here please?

Hardware/HardwareTests/ReceiptRendererTests.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  ReceiptRendererTests.swift
+//  HardwareTests
+//
+//  Created by Allen Snook on 6/16/21.
+//
+
+import XCTest
+
+class ReceiptRendererTests: XCTestCase {
```
ctarda: It looks like we don't actually need these tests?

ctarda (approved):
Thanks for attacking this @allendav 

It tests well. A couple of comments about changes that probably shouldn't have made it into the diff, the one about the Gemfile is particularly important.

:shipit:

Gemfile.lock
allendav: Fixed with https://github.com/woocommerce/woocommerce-ios/pull/4441/commits/2a2e667bd9c708bb6491987a34ee069199101522

Hardware/HardwareTests/ReceiptRendererTests.swift
allendav: Oops! My bad. Yeah, I started building them out and then realized that we 1) don't have a Swifty way to parse HTML and 2) plan on removing the receipt implementation if we can anyhow. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4441/commits/3d5c24f4bc76a4f4862ddf9aa23b7021bca4928e

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
allendav: Done in https://github.com/woocommerce/woocommerce-ios/pull/4441/commits/47446b1ac08a8570c149545b45214b2d9da7599c

PR #4443, opened 2021-06-17
Update tracks to 0.8.5

koke:
This PR updates the Tracks library to the latest version.

Since the newer versions have made some big changes, I had to make a few changes in our logging stack to adjust for that. Mainly, as the Tracks library doesn't offer a shared instance anymore, I created a new `CrashLoggingStack` to be accessed through `ServiceLocator` so the rest of the app can use it in a similar fashion as before.

A big problem with that approach was that, when Core Data initialization fails, it will try to log an error to Sentry, which tries to get the current user. But since the StoresManager is still initializing, the app would crash without being able to send the error. For this, I've changed the logic a bit and replaced the previous `logMessageAndWait` with a more explicit `logFatalErrorAndExit`. When this method is calling, we know we're about to crash, so we can minimize any attempt to gather extra metadata. In our case, it sets a flag in `WCCrashLoggingDataProvider` so `currentUser` immediately returns `nil`, instead of trying to find the current account.

## To test

Since this is a big change, I'd appreciate suggestions on what else to test, but so far I've modified the code to force an error in CoreDataManager and verified that it tries to send the error to Sentry before crashing.

I also checked that tracks events are being sent, by setting breakpoints in `TracksServiceRemote`, and that the app generally works fine.

Update release notes:


koke:
Fixing tests before marking it ready... 😬

ctarda:
If it helps, I did some smoke testing, and the app seems to work as expected. Tracks events are being sent, and I can't see any logical traps in the code. I did not hit approve because there wasn't any reviewers requested though. 

Anyway, I think it wouldn't hurt if there was an extra pair of eyes taking a look as well.

WooCommerce/Classes/Tools/Logging/WooCrashLoggingStack.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    func logError(_ error: Error) {
+        crashLogging.logError(error)
+    }
+
+    func logError(_ error: Error, userInfo: [String: Any]? = nil, level: SeverityLevel = .error) {
+        crashLogging.logError(error, userInfo: userInfo, level: sentrySeverity(with: level))
+    }
+
+    func logFatalErrorAndExit(_ error: Error, userInfo: [String: Any]? = nil, level: SeverityLevel = .error) -> Never {
```
AliSoftware: Feels a bit odd that the name of this method contains `fatalError` but also accepts a `SeverityLevel`. Which means that we can for example call `logFatalErrorAndExit(…, userInfo: nil, level: .info)` too, which I'm not sure makes sense (and even if it made sense in some scenario, it would feels a bit confusing 🤔 )

WooCommerce/Classes/ServiceLocator/ServiceLocator.swift
```diff
@@ -52,12 +52,16 @@ final class ServiceLocator {
...
-    private static var _storageManager = CoreDataManager(name: WooConstants.databaseStackName, crashLogger: SentryCrashLogger())
+    private static var _storageManager = CoreDataManager(name: WooConstants.databaseStackName, crashLogger: crashLogging)
 
     /// Cocoalumberjack DDLog
     ///
     private static var _fileLogger: Logs = DDFileLogger()
 
+    /// Crash Logging Stack
+    ///
+    private static var _crashLogging: CrashLoggingStack = WooCrashLoggingStack()
```
AliSoftware, line 63: For my curiosity: why not using a `private(set)` to have the getter public but the setter private (instead of using `_crashLogging` here and `var crashLogging { _crashLogging }` on lines 156-161)?
```suggestion
    private(set) static var crashLogging: CrashLoggingStack = WooCrashLoggingStack()
```

WooCommerce/Classes/Tools/Logging/WooCrashLoggingStack.swift
```diff
@@ -0,0 +1,145 @@
...
+        return BuildConfiguration.current.rawValue
+    }
+
+    @objc func updateCrashLoggingSystem(_ notification: Notification) {
+        /// Bumping this call to a later run loop is a little bit hack-y, but because the `StoresManager` fires the events
+        /// we're interested as part of its initialization, we need to wait for that initalization to be complete before
+        /// taking action – otherwise the application will deadlock.
+        DispatchQueue.main.async {
+            ServiceLocator.crashLogging.setNeedsDataRefresh()
+        }
```
AliSoftware, line 124: What about if the app is about to crash in the current iteration of the runloop (and will thus never execute the dispatched code), is that a risk/problem?

AliSoftware (commented):
I have only reviewed by reading the code on GitHub, didn't try to run it.

I think @jkmassel is a bit more knowledgeable in Tracks so might give more insights about potential gotchas or not with those changes?

WooCommerce/Classes/Tools/Logging/WooCrashLoggingStack.swift
koke: Agreed, I added the fatal to this method towards the end and didn't realize the level was still there.

WooCommerce/Classes/Tools/Logging/WooCrashLoggingStack.swift
koke: This was already there and I only tried to adapt the current logic to the new API: https://github.com/woocommerce/woocommerce-ios/pull/4443/files#diff-e97506f77d6983646e4da3f634625fce913b97e6e19cf430a8248ab5ed23ac62

I don't understand all the reasoning behind this, but it seems to me that the worst case scenario here is that a crash would have the wrong (to an extent) username attached to it.

WooCommerce/Classes/ServiceLocator/ServiceLocator.swift
koke: I was following the pattern of how everything else does this in `ServiceLocator`. I think that might result in a cleaner implementation, but if we go for it we should do it for all the methods in `ServiceLocator`

jkmassel (approved):
This approach is considerably better, and I have no concerns with it.

:shipit:

PR #4444, opened 2021-06-17
Fix Stats tabs not ordered correctly in RTL layout

### LINKED ISSUE #3934: Stats: tabs text and order are reversed in RTL languages
While testing the app in Arabic, I noticed that when I first open the app, and after the loading view is dismissed, the tabs of the stats are reversed, same as their content:

If I sent the app to background, and re-open it, they will show in the correct order:

I confirmed that this happens with other RTL languages as well.

dvdchr:
Fixes #3934.

This bug originates from one of the third-party dependencies, XLPagerTabStrip. When the total length of the tabs surpasses the screen width – such that the tab strip is horizontally scrollable – the built-in RTL adjustments from `UICollectionView` does not work somehow. 

Previously, we implemented a workaround by "flipping" the tabs container, and then flipping each individual tab again to make sure the text is not reversed. However, the first flip happens in `traitCollectionDidChange` method, which is not guaranteed to be called when the app is opened.

## Changes
Moved the transformation logic from `traitCollectionDidChange` to `configureView`, which gets called once during `viewDidLoad`. This ensures that the first flip for the tabs container always occurs.

Before | After
-- | --
3934_before | 3934_after

## To Test

1. Change device language to any RTL languages. 💡 **Pro Tip:** when testing from simulators, you can force the app to be displayed in RTL layout with _any language_ by adding `-NSForceRightToLeftWritingDirection YES` as launch argument.
2. Launch WooCommerce app, and log in to an account. 
3. On the Stats page, verify that the first highlighted tab should be the right-most tab, with the title "Today".

## Author Checklist

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 7.0
 -----
 - [*] If the Orders, Products, or Reviews lists can't load, a banner now appears at the top of the screen with links to troubleshoot or contact support. [https://github.com/woocommerce/woocommerce-ios/pull/4400, https://github.com/woocommerce/woocommerce-ios/pull/4407]
+- [*] Fix: Stats tabs are now displayed and ordered correctly in RTL languages. [https://github.com/woocommerce/woocommerce-ios/pull/4444]
 
 6.9
 -----
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
@@ -57,18 +57,6 @@ final class StoreStatsAndTopPerformersViewController: ButtonBarPagerTabStripView
         ensureGhostContentIsAnimated()
     }
 
-    // MARK: - RTL support
-
-    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        super.traitCollectionDidChange(previousTraitCollection)
-        /// ButtonBarView is a collection view, and it should flip to support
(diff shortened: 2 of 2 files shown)


dvdchr:
Additional note: XLPagerTabStrip also forces the contents (i.e. view controllers) to be lined up in LTR layout. This means that even though we've fixed the order of the tabs in RTL, content transitions are still performed directionally to the right. Here's a short video to better illustrate this:

https://user-images.githubusercontent.com/1299411/122379627-3a48f700-cf91-11eb-9b5b-1785fea622e1.mov

(Note: this also happens in the Order page.)

I'm not used to RTL layouts, but _I think_ the correct behavior here is to transition to the left – basically following the order of the tabs. Feel free to correct me if I got this wrong! 😄 

Anyways, since this is out of scope from this PR, I'll check if this has been raised before previously, and if not, I'll probably go ahead and create another issue for this.

RELEASE-NOTES.txt
```diff
@@ -3,6 +3,7 @@
 7.0
 -----
 - [*] If the Orders, Products, or Reviews lists can't load, a banner now appears at the top of the screen with links to troubleshoot or contact support. [https://github.com/woocommerce/woocommerce-ios/pull/4400, https://github.com/woocommerce/woocommerce-ios/pull/4407]
+- [*] Fix: Stats tabs are now displayed and ordered correctly in RTL languages. [https://github.com/woocommerce/woocommerce-ios/pull/4444]
```
ctarda, line 6: 🥳

ctarda (approved):
Nice one @dvdchr !

I tested with RTL Pseudo language. 

| RTL Pseudo language | 🇨🇦 English |
| ---- | ---- |
|  |  |

:shipit:

PR #4445, opened 2021-06-17
Display Order Items and Shipping Labels Packages as separate sections in the Order Detail Screen

### LINKED ISSUE #4412: Order Details → Variations With Different Attributes are Combined Together

### LINKED ISSUE #4436: Display Order items and Shipping Labels packages as separate sections in the Order Detail screen
As discussed here https://github.com/woocommerce/woocommerce-ios/issues/4412 and here https://github.com/woocommerce/woocommerce-ios/issues/4126 we need to display the order items (products) in a section that is always visible also if there are some shipping labels. We basically need to separate the packages bought section from the items ordered section. p91TBi-566#comment-4510

Schermata 2021-06-16 alle 17 33 43

pmusolino:
Fixes #4436  
Fixes #4412

## Description
In this PR, I tried to fix two different issues that we encountered in the Order Detail screen.
The first issue was about displaying the Products section and the Shipping Label Packages sections at the same time, because previously, if there was an existing package, replaced the products. This change will allow us to solve this issue in the future https://github.com/woocommerce/woocommerce-ios/issues/4126 and this one #4412, which I solved following the code implemented previously by @koke here #4416.
This implementation, changed a little bit the Shipping Label Package section because now the items are not showed inside the section, but by tapping on them you will see a new screen showing the package items.
I also removed the cell "Details" under Products, because it was superfluous.

#### Case 1
1. Open an order detail with partially refunded items.
2. Make sure that everything appears like before.

#### Case 2
1. Open an order, that is eligible for Shipping Label Creation.
2. Make sure to see the shipping label creation button under the Order Items.

#### Case 3
1. Open an order, with existing Shipping Label Packages.
2. Make sure you are able to see the order items in the order detail screen, without any button below (for example, you shouldn't see the shipping label creation button).
3. Make sure that you are able to see the shipping label packages under the order items section.

#### Case 4
1. Create a variable product.
2. Add “Color” as an attribute with “Blue|Green|Yellow” as the values.
3. Create a single variation of that variable product. Choose “Any” as the attribute option.
4. Create a single order for the product.
5. Add 1 Blue
6. Add 2 Green
7. Add 5 Yellow
8. Check out and submit the order.
9. In the app, view the order. It should show three different items, one for each color purchased.

## Screenshots
| New Shipping Label Package section in Order Detail             |  Shipping Label Package Items detail screen |
Simulator Screen Shot - iPhone 12 Pro - 2021-06-17 at 13 33 42 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-17 at 13 33 49

Variable Products Before | Variable Products Before After
-|-
Screen Shot 2021-06-15 at 09 44 15|Screen Shot 2021-06-15 at 09 44 59

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -215,6 +215,13 @@ final class OrderDetailsDataSource: NSObject {
         }
         return shippingLabelOrderItemsAggregator.orderItem(of: shippingLabel, at: indexPath.row)
     }
+
+    func shippingLabelOrderItems(at indexPath: IndexPath) -> [AggregateOrderItem] {
+        guard let shippingLabel = shippingLabel(at: indexPath) else {
+            return []
+        }
+        return shippingLabelOrderItemsAggregator.orderItems(of: shippingLabel)
+    }
 }
 
 
@@ -310,8 +317,8 @@ private extension OrderDetailsDataSource {
             configureRefund(cell: cell, at: indexPath)
         case let cell as TwoColumnHeadlineFootnoteTableViewCell where row == .netAmount:
             configureNetAmount(cell: cell)
-        case let cell as ProductDetailsTableViewCell where row == .shippingLabelProduct:
-            configureShippingLabelProduct(cell: cell, at: indexPath)
+        case let cell as WooBasicTableViewCell where row == .shippingLabelProducts:
+            configureShippingLabelProducts(cell: cell, at: indexPath)
         case let cell as ProductDetailsTableViewCell where row == .aggregateOrderItem:
             configureAggregateOrderItem(cell: cell, at: indexPath)
         case let cell as ButtonTableViewCell where row == .collectCardPaymentButton:
@@ -487,8 +494,7 @@ private extension OrderDetailsDataSource {
 
     private func configureCollectPaymentButton(cell: ButtonTableViewCell, at indexPath: IndexPath) {
         cell.configure(style: .primary,
-                       title: Titles.collectPayment,
-                       bottomSpacing: ButtonTableViewCell.Constants.defaultBottomSpacing) {
+                       title: Titles.collectPayment) {
             self.onCellAction?(.collectPayment, indexPath)
         }
         cell.hideSeparator()
@@ -558,23 +564,33 @@ private extension OrderDetailsDataSource {
                                 "VoiceOver accessibility hint for the row that shows instructions on how to print a shipping label on the mobile device")
     }
 
-    private func configureShippingLabelProduct(cell: ProductDetailsTableViewCell, at indexPath: IndexPath) {
-        cell.selectionStyle = .default
-
-        guard let shippingLabel = shippingLabel(at: indexPath),
-              let orderItem = shippingLabelOrderItemsAggregator.orderItem(of: shippingLabel, at: indexPath.row) else {
+    private func configureShippingLabelProducts(cell: WooBasicTableViewCell, at indexPath: IndexPath) {
+        guard let shippingLabel = shippingLabel(at: indexPath) else {
             assertionFailure("Cannot access shipping label and/or order item at \(indexPath)")
             return
         }
 
-        let addOns = filterAddOns(of: orderItem)
-        let itemViewModel = ProductDetailsCellViewModel(aggregateItem: orderItem,
-                                                        currency: order.currency,
-                                                        hasAddOns: addOns.isNotEmpty)
-        cell.configure(item: itemViewModel, imageService: imageService)
-        cell.onViewAddOnsTouchUp = { [weak self] in
-            self?.onCellAction?(.viewAddOns(addOns: addOns), nil)
-        }
+        let orderItems = shippingLabelOrderItemsAggregator.orderItems(of: shippingLabel)
+        let singular = NSLocalizedString("%d item",
+                                         comment: "For example: `1 item` in Shipping Label package row")
+        let plural = NSLocalizedString("%d items",
+                                       comment: "For example: `5 items` in Shipping Label package row")
+        let itemsText = String.pluralize(orderItems.count, singular: singular, plural: plural)
+
+        cell.bodyLabel?.text = itemsText
+        cell.applySecondaryTextStyle()
+        cell.accessoryType = .disclosureIndicator
+        cell.selectionStyle = .default
+        cell.hideSeparator()
+
+        cell.accessibilityTraits = .button
+        cell.accessibilityLabel = NSLocalizedString(
+            "View items in this shipping label package.",
+            comment: "Accessibility label for the items inside a Shipping Label package"
+        )
+
+        cell.accessibilityHint = NSLocalizedString("Show the items included inside this shipping label package.",
+        comment: "VoiceOver accessibility hint, informing the user that the button can be used to view the items inside the shipping label package")
     }
 
     private func configureShippingLabelTrackingNumber(cell: ImageAndTitleAndTextTableViewCell, at indexPath: IndexPath) {
@@ -623,6 +639,7 @@ private extension OrderDetailsDataSource {
     private func configureReprintShippingLabelButton(cell: ButtonTableViewCell, at indexPath: IndexPath) {
         cell.configure(style: .secondary,
                        title: Titles.reprintShippingLabel,
+                       topSpacing: 0,
                        bottomSpacing: 8) { [weak self] in
             guard let self = self else { return }
             guard let shippingLabel = self.shippingLabel(at: indexPath) else {
@@ -685,7 +702,7 @@ private extension OrderDetailsDataSource {
     private func configureMarkCompleteButton(cell: ButtonTableViewCell,
                                              buttonStyle: ButtonTableViewCell.Style,
                                              showsBottomSpacing: Bool) {
-        let bottomSpacing: CGFloat = showsBottomSpacing ? ButtonTableViewCell.Constants.defaultBottomSpacing : 0
+        let bottomSpacing: CGFloat = showsBottomSpacing ? ButtonTableViewCell.Constants.defaultSpacing : 0
         cell.configure(style: buttonStyle, title: Titles.markComplete, bottomSpacing: bottomSpacing) { [weak self] in
             self?.onCellAction?(.markComplete, nil)
         }
@@ -899,9 +916,7 @@ extension OrderDetailsDataSource {
                     rows = [.shippingLabelRefunded, .shippingLabelDetail]
                     headerStyle = .primary
                 } else {
-                    let orderItemsCount = shippingLabelOrderItemsAggregator.orderItems(of: shippingLabel).count
-                    rows = Array(repeating: .shippingLabelProduct, count: orderItemsCount)
-                        + [.shippingLabelReprintButton, .shippingLabelPrintingInfo, .shippingLabelTrackingNumber, .shippingLabelDetail]
+                    rows = [.shippingLabelProducts, .shippingLabelReprintButton, .shippingLabelPrintingInfo, .shippingLabelTrackingNumber, .shippingLabelDetail]
                     let headerActionConfig = PrimarySectionHeaderView.ActionConfiguration(image: .moreImage) { [weak self] sourceView in
                         self?.onShippingLabelMoreMenuTapped?(shippingLabel, sourceView)
                     }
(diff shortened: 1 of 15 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -571,23 +564,33 @@ private extension OrderDetailsDataSource {
...
-                                                        hasAddOns: addOns.isNotEmpty)
-        cell.configure(item: itemViewModel, imageService: imageService)
-        cell.onViewAddOnsTouchUp = { [weak self] in
-            self?.onCellAction?(.viewAddOns(addOns: addOns), nil)
-        }
+        let orderItems = shippingLabelOrderItemsAggregator.orderItems(of: shippingLabel)
+        let singular = NSLocalizedString("%d item",
+                                         comment: "For example: `1 item` in Shipping Label package row")
+        let plural = NSLocalizedString("%d items",
+                                       comment: "For example: `5 items` in Shipping Label package row")
```
Ecarrion: nit: I think we should use `%1$@` in those formats.

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -571,23 +564,33 @@ private extension OrderDetailsDataSource {
...
+        let itemsText = String.pluralize(orderItems.count, singular: singular, plural: plural)
```
Ecarrion: TIL

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -258,28 +258,21 @@ extension OrderDetailsViewModel {
...
             let navigationController = WooNavigationController(rootViewController: printingInstructionsViewController)
             viewController.present(navigationController, animated: true, completion: nil)
-        case .shippingLabelProduct:
-            guard let item = dataSource.shippingLabelOrderItem(at: indexPath), item.productOrVariationID > 0 else {
+        case .shippingLabelProducts:
+            let shippingLabelItems = dataSource.shippingLabelOrderItems(at: indexPath)
+
+            let identifier = AggregatedProductListViewController.classNameWithoutNamespaces
+            guard let productListVC = UIStoryboard.orders.instantiateViewController(identifier: identifier) as? AggregatedProductListViewController else {
+                DDLogError("Error: attempted to instantiate AggregatedProductListViewController. Instantiation failed.")
```
Ecarrion: I was under the impression that we are moving away from Storyboards. What was the reason to use it for this screen?

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -258,28 +258,21 @@ extension OrderDetailsViewModel {
...
                 return
             }
-            let loaderViewController = ProductLoaderViewController(model: .init(aggregateOrderItem: item),
-                                                                   siteID: order.siteID,
-                                                                   forceReadOnly: true)
-            let navController = WooNavigationController(rootViewController: loaderViewController)
-            viewController.present(navController, animated: true, completion: nil)
+            productListVC.viewModel = self
+            productListVC.items = shippingLabelItems
+            viewController.navigationController?.show(productListVC, sender: nil)
```
Ecarrion: nit: I think it can be `viewController.how(productListVC, sender: nil)`

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
```diff
@@ -0,0 +1,148 @@
...
+/// The list shows the product name and the quantity.
+///
+final class AggregatedProductListViewController: UIViewController {
+    @IBOutlet private var tableView: UITableView!
+
+    var viewModel: OrderDetailsViewModel!
+    private var products: [Product]? = []
+    var items: [AggregateOrderItem] = []
+
+    private let imageService: ImageService = ServiceLocator.imageService
```
Ecarrion: I really think we need to turn this into a normal VC with nib, to be able to inject these variables.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
```diff
@@ -0,0 +1,148 @@
...
+//
+private extension AggregatedProductListViewController {
+
+    func itemAtIndexPath(_ indexPath: IndexPath) -> AggregateOrderItem {
+        return items[indexPath.row]
+    }
+
+    func lookUpProduct(by productID: Int64) -> Product? {
+        return products?.filter({ $0.productID == productID }).first
+    }
```
Ecarrion: I think this logic should be inside the view model. That would allow us to add some tests to it.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
```diff
@@ -0,0 +1,148 @@
...
+    /// If the "view add-ons" feature is disabled an empty array will be returned.
+    ///
+    func itemAddOnsAttributes(item: AggregateOrderItem) -> [OrderItemAttribute] {
+        guard let product = lookUpProduct(by: item.productID), viewModel.dataSource.showAddOns else {
+            return []
+        }
+
+        let globalAddOns = viewModel.dataSource.addOnGroups
+        return AddOnCrossreferenceUseCase(orderItemAttributes: item.attributes, product: product, addOnGroups: globalAddOns).addOnsAttributes()
+    }
```
Ecarrion: I think we should not show the "View Add-ons" button on the shipped items screen because we can't really know if a shipped item contained an add-on.

Could we remove all the code associated with it?

Screen Shot 2021-06-17 at 8 43 57 AM

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
```diff
@@ -0,0 +1,148 @@
...
+
+
+// MARK: - Configuration
+//
+private extension AggregatedProductListViewController {
+
+    /// Setup: Main View
+    ///
+    func configureMainView() {
+        title = String.pluralize(items.count, singular: Localization.titleSingular, plural: Localization.titlePlural)
```
Ecarrion: Is it ok to group the items count in the here and in package cell, when we actually shipped more items? 😅 

4 items | 22 items
---- | ----
Screen Shot 2021-06-17 at 8 46 09 AM | Screen Shot 2021-06-17 at 8 46 19 AM

Ecarrion (changes requested):
Hi Paolo! Thanks for tackling this!

Left some comments, but I'm mostly worried about using storyboards & the "view add-on" button!

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
pmusolino: The reason is that the similar view controller that works with `OrderItem` instead of `AggregatedOrderItem` is in the storyboard. I, therefore, preferred to conform and use the same approach by inserting it in the storyboard.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
pmusolino: I'm going to convert it into xib 👍 I did it just to conform to the existing view controllers that are in the `Orders` storyboard.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
pmusolino: Removed 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
pmusolino: I'm not sure about the approach that we want to use here. cc @Garance91540 @adamzelinski 
By the way, I added this header on top of the detail screen, because it's probably not clear what the number on the right indicates.

pmusolino:
Thank you @Ecarrion for the review! Ready for a second pass. I fixed all the things suggested, and moved the `AggregatedProductListViewController` under a xib file.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
```diff
@@ -0,0 +1,155 @@
...
+
+/// A simple Product List presented from the Order Details screen -> Shipping Labels Packages.
+///
+/// The list shows the product name and the quantity.
+///
+final class AggregatedProductListViewController: UIViewController {
+    @IBOutlet private var tableView: UITableView!
+
+    private var viewModel: OrderDetailsViewModel
+    private var products: [Product]? = []
```
Ecarrion: I don't think we need this anymore, do we?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
```diff
@@ -0,0 +1,155 @@
...
+    private var items: [AggregateOrderItem]
+
+    private let imageService: ImageService = ServiceLocator.imageService
```
Ecarrion: nit: A small improvement would be to inject this with a default parameter, so we make dependencies clear.

Ecarrion (approved):
Looks good 🚀 
Just some small nits!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
Garance91540: Hi @pmusolino let's display the total number of items in the Header (so if the package has 3 of the same product, it should read "3 items"). And good idea for adding "qty", it's clearer that way!

Garance91540 (approved):
👏

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
pmusolino: Good catch, removed 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
pmusolino: Done ✅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Product List Section/Product Details/AggregatedProductListViewController.swift
pmusolino: Thank you @Garance91540, I updated the logic 👍

PR #4446, opened 2021-06-17
Add-Ons: Adds ProductAddOn view

### LINKED ISSUE #4268: Add-ons: Create specific add-on view
To be rendered from #4267

This view should render any type of product add-on description in a simple way.

Considerations:
- Show type
- Show Name
- Show description
- Show general price if no options
- Show options as a list
    - Show Name
    - Show Price

Ecarrion:
closes #4268 

# Why

The goal is to show the list of possible product add-ons for a product.

Screen Shot 2021-06-17 at 9 50 35 AM

This PR adds a `ProductAddOn` view type to be able to use it later with real data in `ProductAddOnsListViewController`

# How

- Added `ProductAddOn`

- Added `ProductAddOnViewModel`

- Added `SubheadlineStyle` view modifier. 

# Screenshots

Screen Shot 2021-06-17 at 3 15 24 PM

# Testing

Since this view is not yet integrated, you can go to `ProductAddOn.swift` and play with the preview code.

Update release notes:


koke (approved):
I found a bit odd that you can initialize the model both with a price (that will go unused) and a set of options. Maybe we could model this as an enum, or even two different views.

That said, the view is simple enough that it doesn't feel like a blocker, so I'd say it's good to go 👍🏽

Ecarrion:
> Maybe we could model this as an enum, or even two different views.

Thanks for the review @koke, I don't think that difference is worth the duplication for two different views. But after I confirm that combination is not possible(API allows it 🤷) I'll update the VM!

PR #4447, opened 2021-06-17
Fix magic link flow not opening 3rd party mail apps

### LINKED ISSUE #3272: Magic link flow: can't open 3rd party mail app
**Describe the bug**
When using 3rd party email apps, “open mail” button in the magic link login flow shows an alert instead of opening a mail client.
I looked around a bit and I believe that all the schemes from [here](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/1470051a4a8cee41fcb15ca4e5cfbd4c28455c2c/WordPressAuthenticator/Resources/SupportedEmailClients/EmailClients.plist) needs to be added to `LSApplicationQueriesSchemes` key in `Info.plist` of the Woo app. 

We use `canOpenURL` method to check if the apps are installed, but that call will fail If the schema is not present in `LSApplicationQueriesSchemes`

Reference: https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/LaunchServicesKeys.html

**To Reproduce**
1. Device should be setup with 3rd party mail app
2. Go through magic link flow
3. Tap on "open mail" button
4. See alert instead of a picker with installed mail apps

**Expected behavior**
Tappin on "open mail" button should present a picker with installed mail apps

**Mobile Environment**
Please include:
- Device: iPhone / iPad
- iOS version: iOS 14.2
- WooCommerce iOS version: 5.6

dvdchr:
Fixes #3272.

This change adds several mail app schemes that are [supported by WordPressAuthenticator](https://github.com/wordpress-mobile/WordPressAuthenticator-iOS/blob/1470051a4a8cee41fcb15ca4e5cfbd4c28455c2c/WordPressAuthenticator/Resources/SupportedEmailClients/EmailClients.plist) to `LSApplicationQueriesSchemes` in the `Info.plist`. 

**Note: This change is not tested yet!** I'm unable to test because my personal device is on 14.5 while my Xcode is 12.4 - I'm still downloading Xcode 12.5 🙃. And apparently, it's not possible to install additional Mail apps on the simulator. 

I'll give it another go once I have successfully downloaded Xcode 12.5. Meanwhile, any help testing this would be appreciated! 🙂 

Before | After
-- | --
**Description:** An alert is shown instead of options of supported mail apps. | **Description:** A sheet is presented, prompting the user to select one of the mail apps.
3272_before | IMG_E39C8E51B4AB-1

## To Test

1. Build & run the app on a device with a third-party mail app installed. Note that mail apps are not available in Simulator.
2. Ensure that the app is installed in a clean, logged out state. Then log in with WordPress.com account flow.
3. Select the magic link option.
4. Tap the Open Mail button, and verify that the sheet is shown (instead of an alert).

## Author Checklist

### DIFF
--- a/WooCommerce/Resources/Info.plist
+++ b/WooCommerce/Resources/Info.plist
@@ -54,6 +54,12 @@
 	<key>LSApplicationQueriesSchemes</key>
 	<array>
 		<string>org-appextension-feature-password-management</string>
+		<string>googlegmail</string>
+		<string>airmail</string>
+		<string>ms-outlook</string>
+		<string>readdle-spark</string>
+		<string>ymail</string>
+		<string>fastmail</string>
 	</array>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>


ealeksandrov (approved):
It shows Gmail in action sheet and it opens fine :shipit:
(BTW, when I tried same on `develop` - there was no alert, just Apple Mail opened on button tap)

dvdchr:
Thank you for the help @ealeksandrov! 👍  Sorry for being unclear on the alert condition. Previously, the alert _should_ appear when Apple Mail is uninstalled, and a third-party mail app (that's supported by WordpressAuthenticator) is set as default (and only) mail app. Instead of presenting a sheet, it defaults to showing the alert instead.

PR #4454, opened 2021-06-18
Hot fix: loading banner always visible in the Order Detail screen just in the release build

pmusolino:
As reported by Garance here p5T066-2p9#comment-8784 in our release 6.9 we have a bug, where the new banner introduced for showing a loading banner at the top of the Order Detail screen remain always visible, plus it's possible to do a pull to refresh action while the banner is visible.

## Demo of the bug
https://user-images.githubusercontent.com/495617/122586768-5d40dd00-d05d-11eb-91e5-d1e74d60ca90.MP4

## The bug
It turns out that the bug about the banner not disappearing only occurred on the testflight build, and not in debug mode.
Long short story: the `syncSavedReceipts` method did not return anything in case of an error (an error that occurred only in the release build due to the feature flag not enabled), and this meant that the `syncEverything` method never returned all the responses of the callbacks.

## Testing
Since it occurs just in the release build, you can simulate it by substituting the code of `static var current: BuildConfiguration` under the file `BuildConfiguration.swift` with 
```
    static var current: BuildConfiguration {
        return .appStore
    }
```

1. Open an order detail.
2. Make sure the loading banner on top is visible.
3. Make sure you can't do a pull to refresh while the banner remains visible.
4. Make sure the banner after a bit disappears, and now you are able to do a pull to refresh action.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
@@ -95,6 +95,11 @@ final class OrderDetailsViewController: UIViewController {
         super.viewWillAppear(animated)
         syncEverything { [weak self] in
             self?.topLoaderView.isHidden = true
+
+            /// We add the refresh control to the tableview just after the `topLoaderView` disappear for the first time.
+            if self?.tableView.refreshControl == nil {
+                self?.tableView.refreshControl = self?.refreshControl
+            }
         }
     }
 
@@ -132,7 +137,6 @@ private extension OrderDetailsViewController {
         tableView.estimatedSectionHeaderHeight = Constants.sectionHeight
         tableView.estimatedRowHeight = Constants.rowHeight
         tableView.rowHeight = UITableView.automaticDimension
-        tableView.refreshControl = refreshControl
 
         tableView.dataSource = viewModel.dataSource
     }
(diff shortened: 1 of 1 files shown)


pmusolino:
@jkmassel JFYI this PR will be merged in release 6.9

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -95,6 +95,11 @@ final class OrderDetailsViewController: UIViewController {
         super.viewWillAppear(animated)
         syncEverything { [weak self] in
             self?.topLoaderView.isHidden = true
+
+            /// We add the refresh control to the tableview just after the `topLoaderView` disappear for the first time.
+            if self?.tableView.refreshControl == nil {
+                self?.tableView.refreshControl = self?.refreshControl
+            }
```
Ecarrion, line 102: Should we then take the refresh control out in `viewWillDissapear`?
I mean, in order to not allow the loading + pulling when the user goes away and then back to this screen.

Ecarrion (approved):
Thanks for tackling this @pmusolino!
Just left a minor cconsideration!

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
pmusolino: I think not, because doing so would mean that when the user returns to the screen, we will re-add the control to the view, and he would be allowed to redo pull to refresh while there is already a list of API calls in the background.

pmusolino: I'm going to merge it now since it's a hot fix, but we can continue the discussion and eventually if something comes out, I can open a new PR 👍

Issue #4089, opened 2021-04-30
Shipping Labels: purchase labels


pmusolino:
As part of Shipping Labels M3, add the missing endpoints and UI for purchasing the label after pressing the button "Purchase shipping label", showing the progress view and the view for print the shipping label (this screen was already implemented, but need to be customized for the shipping label purchase flow)

- [x] Add Networking support for purchase endpoint: `POST: /wc/v1/connect/label/{{orderId}}` #4423
- [x] Add Yosemite support for purchase endpoint #4458
- [x] Add Networking support for label status endpoint: `GET /connect/label/{order_id}/{label_id1},{label_id2}...` #4448
- [x] Add Yosemite support for label status endpoint #4458
- [x] UI for progress view (we already have this component) #4475
- [x] UI for print shipping label view (we can reuse the existing one) #4562
- [x] UI for failed purchase #4569
- [x] Hook up the "Mark this order complete and notify the customer" toggle. #4562
- [x] Hook up the "Purchase shipping label" button to make the purchase #4475

<img width="521" alt="Schermata 2021-04-30 alle 12 51 19" src="https://user-images.githubusercontent.com/495617/116685617-f2571c00-a9b2-11eb-8763-9fdd7a3518c3.png">

rachelmcr:
A few notes from the reviews on https://github.com/woocommerce/woocommerce-ios/pull/4475 (things to fix):

- [x] I also noticed that some parsing errors when purchasing the label:

```
<> Mapping Error: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "tracking", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
```
- [x] I just noticed that in the `Print Shipping Label` screen the back button pop to the previous screen, so we are allowed to create an identical shipping label. In this case, we should pop to the order detail screen.
- [x] It seems that the newly purchased shipping label is somehow not connected to the order, as it does not appear under the order, and it is possible to buy a new shipping label. -> Tracking this in #4568

PR #4458, opened 2021-06-21
Shipping Labels: Yosemite layer for shipping label purchases

rachelmcr:
Part of: #4089

## Description

This PR implements the Yosemite layer for purchasing shipping labels. It combines two remote endpoints to initiate the purchase and then poll for the label status until the purchase is complete, returning the purchased label(s) or an error.

This flow chart outlines the purchase flow:

(internal ref: p91TBi-4K3-p2)

## Changes

* Adds the action `purchaseShippingLabel` to `ShippingLabelAction` and `ShippingLabelStore`. This encapsulates the entire flow outline above.
* Adds a private method `pollLabelStatus` to poll the label status according to the flow chart above. This is called in `purchaseShippingLabel`.
* Adds an Error enum `LabelPurchaseError`, to identify the types of errors that could result from a successful response from `ShippingLabelRemote.checkLabelStatus`.
* Adds unit tests for `purchaseShippingLabel`.

It would be nice to abstract the polling behavior into a reusable class, for other places where we might want to poll the API like this. I thought I'd get feedback on this approach first, though — is this a reasonable way to do the polling (using `DispatchQueue`) or is there another way to do this that might be better?

## Testing

Confirm tests pass in CI (Yosemite action is not yet used in app).
Confirm the code matches the flow chart above and the tests are checking the right things.

## Submitter Checklist

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/ShippingLabelStore.swift
+++ b/Yosemite/Yosemite/Stores/ShippingLabelStore.swift
@@ -428,10 +428,16 @@ private extension ShippingLabelStore {
                 }
 
                 // If no errors but status is not PURCHASED for all labels, poll again after delay
-                else {
+                else if maxRetries > 0 {
                     DispatchQueue.main.asyncAfter(deadline: .now() + delay) { [weak self] in
                         self?.pollLabelStatus(withDelayInSeconds: delay, maxRetries: maxRetries - 1, siteID: siteID, orderID: orderID, labelIDs: labelIDs, completion: completion)
                     }
+                }
+
+                // If there are no retries left
+                else {
+                    DDLogError("⛔️ Shipping label purchase for order \(orderID) still in progress")
+                    completion(.failure(LabelPurchaseError.purchaseIncomplete))
                 }
 
             case .failure(let error):
@@ -455,5 +461,7 @@ private extension ShippingLabelStore {
     enum LabelPurchaseError: Error {
         /// API returns a `PURCHASE_ERROR` status for a label
         case purchaseErrorStatus
+        /// Label purchase not complete after polling the backend
+        case purchaseIncomplete
     }
 }
--- a/Yosemite/Yosemite/Stores/ShippingLabelStore.swift
+++ b/Yosemite/Yosemite/Stores/ShippingLabelStore.swift
@@ -69,6 +69,14 @@ public final class ShippingLabelStore: Store {
             synchronizeShippingLabelAccountSettings(siteID: siteID, completion: completion)
         case .updateShippingLabelAccountSettings(let siteID, let settings, let completion):
             updateShippingLabelAccountSettings(siteID: siteID, settings: settings, completion: completion)
+        case .purchaseShippingLabel(let siteID, let orderID, let originAddress, let destinationAddress, let packages, let emailCustomerReceipt, let completion):
+            purchaseShippingLabel(siteID: siteID,
+                                  orderID: orderID,
+                                  originAddress: originAddress,
+                                  destinationAddress: destinationAddress,
+                                  packages: packages,
+                                  emailCustomerReceipt: emailCustomerReceipt,
+                                  completion: completion)
         }
     }
 }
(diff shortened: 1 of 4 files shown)


Yosemite/Yosemite/Stores/ShippingLabelStore.swift
```diff
@@ -212,6 +220,49 @@ private extension ShippingLabelStore {
...
+
+            // Poll the status of the label purchases from the response above
+            // with a delay of 1 second each time, with a maximum of 3 retries
+            self.pollLabelStatus(withDelayInSeconds: 1.0,
+                                 maxRetries: 3,
+                                 siteID: siteID,
+                                 orderID: orderID,
+                                 labelIDs: labelPurchaseIDs,
+                                 completion: completion)
+        }
```
pmusolino: Should we do this just in case of success? Looking at the flow chart, yes.

I don't think it makes sense to start polling, regardless of the result of the previous call. Furthermore, we could happen in the event that the first call for purchase becomes very slow, that the `pollLabelStatus` responds sooner.
And if the purchase API request fails, we would still make an unnecessary call. I'm wrong?

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
```diff
@@ -355,4 +406,72 @@ private extension ShippingLabelStore {
...
+
+                // If any label has PURCHASE_ERROR status, return error and stop polling
+                else if labels.contains(where: { $0.status == .purchaseError }) {
+                    DDLogError("⛔️ Error purchasing shipping label for order \(orderID)")
+                    completion(.failure(LabelPurchaseError.purchaseErrorStatus))
+                }
+
+                // If no errors but status is not PURCHASED for all labels, poll again after delay
+                else if maxRetries > 0 {
+                    DispatchQueue.main.asyncAfter(deadline: .now() + delay) { [weak self] in
```
pmusolino, line 432: I think it makes sense to use `DispatchQueue` here. 👍

pmusolino (commented):
It looks good to me, and thanks for showing the flow chart, which was really useful to understand the entire process.

I only left a comment on one thing that I think needs to be resolved, as we would be polling regardless of the result of the shipping label purchase request.

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
rachelmcr: Oh yes, thanks for catching that! I was mistaken about the flow when I wrote it that way. That polling should be only in the success case; I've changed that in f598fae.

rachelmcr:
Thanks for the review, @pmusolino! I fixed the polling behavior so this is ready for another look.

Also, kudos to @hichamboushaba for the excellent flow chart 🎉

pmusolino (approved):
:shipit: 👏

Yosemite/Yosemite/Stores/ShippingLabelStore.swift
allendav: 5 (2+1+1+1) seconds seems short to me - my experience buying labels through wp-admin (on shared hosting, admittedly) seems at least that long, if not more. cc @DanReyLop

rachelmcr: The idea here was to match the web client's polling behavior (more discussion in p91TBi-4K3-p2). The delay should be between when we get the response and when we next poll, so the total time could be longer than 5 seconds. If I've misunderstood any of that, or if we think we need to lengthen the delay, I can certainly change this!

DanReyLop: 5 seconds is definitely too low. But, as Rachel pointed out, when including 3 `iOS -> WP -> WCS server` round-trips, the actual time will be significantly higher. It still feels unnecessarily low though, there's no downside to retrying 5 times or so instead of 3.

I see that the logic in the web client [is exactly the same](https://github.com/Automattic/woocommerce-services/blob/dbeed610f336f8db7995743765dbcd0ff2e18474/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js#L757-L777), so whatever is decided it should be applied in both codebases.

PR #4459, opened 2021-06-21
Use Xcode 12.5 in CI

mokagio:
Updates the Fastlane and CircleCI config file to use Xcode 12.5 and iOS 14.4. Also adds and `.xcversion` file, which we'll use in the future as the single source of truth for the Xcode toolchain version.

**To test:** If CI is green, that's all we need because this change affects CI only.

Update release notes:


Yosemite/YosemiteTests/Tools/Media/MediaTypeTests.swift
```diff
@@ -61,11 +61,17 @@ final class MediaTypeTests: XCTestCase {
...
             "video/mp4", "video/x-msvideo", "video/mpeg", "video/3gpp", "video/3gpp2",
             // 3gpp/3gpp2 audio is considered as video on iOS
             "audio/3gpp", "audio/3gpp2"
         ]
+        if #available(iOS 14.4, *) {
+            videoMimeTypes += [
+                "video/webm", // iOS 14.4 supports WebM
+                "audio/webm"  // audio/webm also results as a video type
+            ]
+        }
```
mokagio: These tests started failing because of `video/webm` and `audio/webm`. I used `#available(iOS 14.4, *)` to ensure a smooth transition for folks that are still using Xcode < 12.5

.circleci/config.yml
```diff
@@ -9,21 +9,21 @@ orbs:
 parameters:
   beta_build:
     type: boolean
-    default: false 
+    default: false
```
mokagio, line 12: Sorry for the trailing whitespace noise. I thought I'd leave it in here since this PR is quite small.

Yosemite/YosemiteTests/Tools/Media/MediaTypeTests.swift
mokagio: @jaclync GitHub tells me you created this file? Does this change make sense to you?

Is this something that could impact the production code, too?

AliSoftware (approved):
Changes LGTM! 🎉 

I'll let Jaclyn answer your question about WebM changes (tho they make sense to me) but otherwise, :shipit:

Yosemite/YosemiteTests/Tools/Media/MediaTypeTests.swift
pmusolino: Jaclyn is AFK, then I reviewed the code in her place.

It seems that we can exclude `video/webm` and `audio/webm` if possible for every system version, since also if we will be able to upload it, they will not be saved into a WooCommerce product. I did the test also on the web, and I can confirm that it seems a format not supported by WooCommerce/WordPress.

About the video with `webm` extension:
- It's uploaded correctly.
- It will be not added to the product gallery.
- Also from Wordpress, if you try to add it inside the description of the product, from Safari the video is not reproducible.

mokagio: I see... 🤔 I might open a new PR on top of this one to address this change, just to keep the conversation around it localized.

Thanks @pmusolino!

_Stay tuned_

jkmassel: @mokagio just wanted to follow up on this – AFAIK we still need to refactor this test, right?

Yosemite/Yosemite/Tools/Media/MediaType.swift
```diff
@@ -28,6 +28,19 @@ public enum MediaType {
...
+        // interpreting both as video types, but we don't want the app to allow users to upload
+        // them because WooCommerce/WordPress don't support that format, yet.
+        //
+        // The UT type identifier for both video/webm and audio/webm is "org.webmproject.webm"
+        //
+        // See https://github.com/woocommerce/woocommerce-ios/pull/4459/files#r656194065
+        guard "\(fileUTI)" != "org.webmproject.webm" else {
+            self = .other
+            return
+        }
```
mokagio, line 42: As discussed [here](https://github.com/woocommerce/woocommerce-ios/pull/4459#discussion_r656194065), even though iOS now supports WebM audio and video, we don't want the app to import them.

I initially thought to use a `if #available(iOS 14.4, *)` check, but then decided to use this one because:

1. It ~~should work~~ [works](https://github.com/woocommerce/woocommerce-ios/pull/4503) fine on Xcode < 12.5
2. We don't have any reason right now to support Xcode < 12.5

mokagio: @jkmassel @pmusolino @jaclync I refactored (= change of implementation, same behavior, as shown by the tests not having changed in this diff and still passing) as suggested by Paolo [here](https://github.com/woocommerce/woocommerce-ios/pull/4459#discussion_r656194065).

Can I ask you to have another look? Thanks!

pmusolino: Looks good to me @mokagio, and thanks for explaining all the details in the comment 👍

pmusolino (approved):
💯

PR #4461, opened 2021-06-22
Address new warnings from Xcode 12.5

mokagio:
Migrating to Xcode 12.5 (#4459) introduced two new warnings:

> Using '`class`' keyword for protocol inheritance is deprecated; use '`AnyObject`' instead

and 

> Future versions of Swift reserve the word 'await'; if this name is unavoidable, use backticks to escape it

This PR addresses them.

@jaclync, @shiki, @Ecarrion: GitHub suggested you as reviewers 😅 cc @jkmassel as release manager.

_I based this on `xcode-12.5` for ease of review. As an aside, these changes should still work if rebased onto `develop` without the ones from that branch._

---

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. _N.A._

### DIFF
--- a/TestKit/Sources/TestKit/XCTestCase+Wait.swift
+++ b/TestKit/Sources/TestKit/XCTestCase+Wait.swift
@@ -72,7 +72,7 @@ extension XCTestCase {
     public func waitFor<ValueType>(file: StaticString = #file,
                                    line: UInt = #line,
                                    timeout: TimeInterval = 5.0,
-                                   await: @escaping (_ promise: (@escaping (ValueType) -> Void)) throws -> Void) rethrows -> ValueType {
+                                   awaiting: @escaping (_ promise: (@escaping (ValueType) -> Void)) throws -> Void) rethrows -> ValueType {
         let exp = expectation(description: "Expect promise to be called.")
 
         var receivedValue: ValueType? = nil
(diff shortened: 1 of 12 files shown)


TestKit/Sources/TestKit/XCTestCase+Wait.swift
```diff
@@ -72,7 +72,7 @@ extension XCTestCase {
     public func waitFor<ValueType>(file: StaticString = #file,
                                    line: UInt = #line,
                                    timeout: TimeInterval = 5.0,
-                                   await: @escaping (_ promise: (@escaping (ValueType) -> Void)) throws -> Void) rethrows -> ValueType {
+                                   awaiting: @escaping (_ promise: (@escaping (ValueType) -> Void)) throws -> Void) rethrows -> ValueType {
```
mokagio, line 75: The compiler recommends replacing the `await` word, to avoid confusion with the new `await` keyword in Swift 5.5's async/await. The other option would have been to escape it with backticks, but I think replacing makes more sense here.

Ecarrion (approved):
Works well on my end! 👍

mokagio:
Thanks @Ecarrion 🙇 🙌 

I'll hold off merging this till #4459 is in, because there's a bit of discussion in there and I don't want the diff to become heavier.

PR #4331, opened 2021-05-31
[Mobile Payments] UI review. Mark I

ctarda:
Closes #4324 

We have recently reached the point where we have been able to implement our first iteration of the designs. 

This PR would help check the current status of the UI and perform a first round of review.

We are still ironing out some issues in landscape mode, so for this first review let's do only portrait and phone. We will submit separate PRs for landscape phone, and portrait and landscape tablet.

Please, @Garance91540 @adamzelinski , let's use this PR to point out inconsistencies and issues in the UI layer, so that we can log tickets for them and prioritize them for the code freeze in June 25. With that in mind, some indication of the severity would help us prioritize, in case we don't have time to fix every single thing on time (something like "to fix before beta 1" or "can wait until right after beta 1" would be very helpful).

# Videos.

1. Discover and connect to the first reader available:

https://user-images.githubusercontent.com/2722505/120219403-3125ff00-c209-11eb-97d9-7f3a3b046e9b.MP4

2. Collect a payment and print a receipt (when there is a reader connected):

https://user-images.githubusercontent.com/2722505/120219514-587ccc00-c209-11eb-93d5-10b20e340b2f.MP4

3. Connect to a reader and then collect the payment when there is no reader connected:

https://user-images.githubusercontent.com/2722505/120219661-91b53c00-c209-11eb-9a97-45edf17f351f.MP4

# Screenshots:

1. Screen 1: Instructions to connect to a reader:

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

2. Screen 2: Searching for readers:

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

3. Screen 3: Reader found:

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

4. Screen 4: List of readers:

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

5. Screen 5: Order details (collect payment button)

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

6. Screen 6. Collect payment, step 1. 

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

7. Screen 7. Collect payment, step 2. 

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

8. Screen 8. Collect payment, step 3. 

| Light Appearance | Dark Appearance |
| ---- | ---- |
|  |  |

9. Screen 9. Success. 

| Light Appearance | Dark Appearance |
| ---- | ---- |
[...]

PR #4467, opened 2021-06-22
[Mobile Payments] Force height of modal views to 382px

### LINKED ISSUE #4360: [Mobile Payments] [UI] Height of modal views should be 382x280
Brought up during the review of #4331

We have different modal sizes along the flow, which might be a bit unsettling as the modals will seem to "jump" from screen to screen. We should have a unique modal size all along the flow (382px by 280px) (can wait but should be addressed soon if possible) https://github.com/woocommerce/woocommerce-ios/pull/4331#issuecomment-853989884

It should doable but can be deferred to after beta 1

ctarda:
Closes #4360 

Brought up during the review of #4331

We have different modal sizes along the flow, which might be a bit unsettling as the modals will seem to "jump" from screen to screen. We should have a unique modal size all along the flow (382px by 280px) https://github.com/woocommerce/woocommerce-ios/pull/4331#issuecomment-853989884

I don't loooove the solution, and I don't expect it to work perfectly on every single combination of device and rotation, but it seems to work reasonably well. I am going for a "let's get closer to our goal" more than a "let's fix every single use case". 

The modal views can contain several different elements, some of them are not always visible, and some of them can span for more than one line. The only way I can think of to fit all that into a fixed height view that is declared in a single xib is by doing something like what this PR does. 

| Before | After |
| ---- | ---- |
|  |  |
|  |  | 
|  |  |
|  |  |
|  |  |

## Changes
* Force the modal to be 382 px tall and width to 280. 
* Updated vertical compression resistance priorities of several elements.

## How to test
1. Set up a store for In-Person Payments:  P91TBi-5ee-p2
2. Connect to a reader.
3. Navigate to an order eligible for In-Person Payments
4. Go thought the flow to collect a payment.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.xib
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.xib
@@ -132,7 +132,7 @@
                         <constraint firstAttribute="bottom" secondItem="jIt-xb-rrN" secondAttribute="bottom" constant="32" id="lVZ-pu-Fmb"/>
                         <constraint firstAttribute="height" priority="250" constant="581" id="n7H-7U-izh"/>
                         <constraint firstItem="jIt-xb-rrN" firstAttribute="centerX" secondItem="8Tg-Q2-wkH" secondAttribute="centerX" id="p6Y-jJ-KuU"/>
-                        <constraint firstAttribute="width" relation="greaterThanOrEqual" priority="250" constant="280" id="uaz-2N-7M4"/>
+                        <constraint firstAttribute="width" priority="250" constant="280" id="uaz-2N-7M4"/>
                         <constraint firstAttribute="trailing" secondItem="jIt-xb-rrN" secondAttribute="trailing" constant="20" id="ufT-Yf-gsb"/>
                     </constraints>
                 </view>
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.xib
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.xib
@@ -24,6 +24,7 @@
                 <outlet property="topSubtitleLabel" destination="aVs-Lf-QTr" id="uKV-ox-li8"/>
                 <outlet property="topTitleLabel" destination="oDv-zz-CTp" id="wIT-XT-OIO"/>
                 <outlet property="view" destination="i5M-Pr-FkT" id="sfx-zR-JGt"/>
+                <outlet property="widthConstraint" destination="uaz-2N-7M4" id="93C-Ou-yrH"/>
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.xib
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="17701" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="18122" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
     <device id="retina6_1" orientation="portrait" appearance="light"/>
     <dependencies>
-        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="17703"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="18093"/>
```
ctarda, line 5: I used Xcode 12.5 for this, therefore we get this diff back

koke (approved):
It seems to work fine on portrait. On landscape, not so much:

Screen Shot 2021-06-24 at 11 52 55

But to be fair, the previous implementation was also not working great on landscape:

Screen Shot 2021-06-24 at 12 06 28

I really think we should only support portrait orientation in phones for most things in the app 😅

ctarda:
Argh. This never ends... 🤦🏼 

I've logged #4474 to looks into the issues on landscape as a follow-up. I'll take it right after this

Garance91540 (approved):
Hi Cesar! 

Are you thinking of updating these modals to match designs in the future, and having the current solution temporary? In that case, that's all good to me! But in the final version, we should really have each components match the designs (for example it's a bit disturbing to see the image assets changing size from one modal to the other, they should always be the same):
Screen Shot 2021-06-24 at 5 27 35 PM

ctarda:
Hi @Garance91540 Yes, our plan is to iterate on the modals again. 

Just to confirm, what we want is to centre the content of the modal vertically within a 382 by 280 container, is that correct?

Garance91540:
> Just to confirm, what we want is to centre the content of the modal vertically within a 382 by 280 container, is that correct?

Hi Cesar, the modal is vertically centered on the screen, but each component of the modal has a specific padding that are detailed in the Figma file

For example, these are the padding requirements for the "Getting ready to collect payments" modal

Screen Shot 2021-06-29 at 11 24 33 AM

PR #4470, opened 2021-06-23
Refactor Stats action (store and networking) to use Result

ealeksandrov:
## Description

This PR updates StatsActionV4 + related Store and Remotes to use Result, so the state of response is more limited with less optional parameters.

## Test

2. Run app and verify stats works well.

---

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
@@ -136,11 +136,12 @@ private extension StoreStatsAndTopPerformersViewController {
                            siteTimezone: timezoneForSync,
                            timeRange: vc.timeRange,
                            latestDateToInclude: latestDateToInclude) { [weak self] result in
-                if case let .failure(error) = result {
+                switch result {
+                case .success:
+                    self?.trackStatsLoaded(for: vc.granularity)
+                case .failure(let error):
                     DDLogError("⛔️ Error synchronizing order stats: \(error)")
                     syncError = error
-                } else {
-                    self?.trackStatsLoaded(for: vc.granularity)
                 }
                 group.leave()
             }
@@ -367,13 +368,14 @@ private extension StoreStatsAndTopPerformersViewController {
                                                           earliestDateToInclude: earliestDateToInclude,
                                                           latestDateToInclude: latestDateToInclude,
                                                           onCompletion: { result in
-                                                            if case let .failure(error) = result {
-                                                                DDLogError("⛔️ Dashboard (Top Performers) — Error synchronizing top earner stats: \(error)")
-                                                            } else {
+                                                            switch result {
+                                                            case .success:
                                                                 ServiceLocator.analytics.track(.dashboardTopPerformersLoaded,
                                                                                                withProperties: [
                                                                                                 "granularity": timeRange.topEarnerStatsGranularity.rawValue
                                                                                                ])
+                                                            case .failure(let error):
+                                                                DDLogError("⛔️ Dashboard (Top Performers) — Error synchronizing top earner stats: \(error)")
                                                             }
                                                             onCompletion?(result)
         })
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
@@ -135,8 +135,8 @@ private extension StoreStatsAndTopPerformersViewController {
             self.syncStats(for: siteID,
                            siteTimezone: timezoneForSync,
                            timeRange: vc.timeRange,
-                           latestDateToInclude: latestDateToInclude) { [weak self] error in
-                if let error = error {
+                           latestDateToInclude: latestDateToInclude) { [weak self] result in
+                if case let .failure(error) = result {
                     DDLogError("⛔️ Error synchronizing order stats: \(error)")
                     syncError = error
                 } else {
@@ -149,8 +149,8 @@ private extension StoreStatsAndTopPerformersViewController {
             self.syncSiteVisitStats(for: siteID,
                                     siteTimezone: timezoneForSync,
                                     timeRange: vc.timeRange,
-                                    latestDateToInclude: latestDateToInclude) { error in
-                if let error = error {
+                                    latestDateToInclude: latestDateToInclude) { result in
+                if case let .failure(error) = result {
                     DDLogError("⛔️ Error synchronizing visitor stats: \(error)")
                     syncError = error
                 }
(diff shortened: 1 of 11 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
```diff
@@ -135,8 +135,8 @@ private extension StoreStatsAndTopPerformersViewController {
             self.syncStats(for: siteID,
                            siteTimezone: timezoneForSync,
                            timeRange: vc.timeRange,
-                           latestDateToInclude: latestDateToInclude) { [weak self] error in
-                if let error = error {
+                           latestDateToInclude: latestDateToInclude) { [weak self] result in
+                if case let .failure(error) = result {
```
pmusolino: nit: probably it's better to use a switch with `success` and `failure` here since it's the approach that we used elsewhere.

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
```diff
@@ -149,8 +149,8 @@ private extension StoreStatsAndTopPerformersViewController {
             self.syncSiteVisitStats(for: siteID,
                                     siteTimezone: timezoneForSync,
                                     timeRange: vc.timeRange,
-                                    latestDateToInclude: latestDateToInclude) { error in
-                if let error = error {
+                                    latestDateToInclude: latestDateToInclude) { result in
+                if case let .failure(error) = result {
```
pmusolino, line 154: nit: same here, I suggest using the `switch`

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
```diff
@@ -161,8 +161,8 @@ private extension StoreStatsAndTopPerformersViewController {
             self.syncTopEarnersStats(for: siteID,
                                      siteTimezone: timezoneForSync,
                                      timeRange: vc.timeRange,
-                                     latestDateToInclude: latestDateToInclude) { error in
-                if let error = error {
+                                     latestDateToInclude: latestDateToInclude) { result in
+                if case let .failure(error) = result {
```
pmusolino, line 166: nit: same here, I suggest using the `switch`

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
```diff
@@ -320,18 +320,18 @@ private extension StoreStatsAndTopPerformersViewController {
...
         let earliestDateToInclude = timeRange.earliestDate(latestDate: latestDateToInclude, siteTimezone: siteTimezone)
         let action = StatsActionV4.retrieveStats(siteID: siteID,
                                                  timeRange: timeRange,
                                                  earliestDateToInclude: earliestDateToInclude,
                                                  latestDateToInclude: latestDateToInclude,
                                                  quantity: timeRange.maxNumberOfIntervals,
-                                                 onCompletion: { error in
-                                                    if let error = error {
+                                                 onCompletion: { result in
+                                                    if case let .failure(error) = result {
```
pmusolino, line 332: nit: same here, I suggest using the `switch`

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
```diff
@@ -341,16 +341,17 @@ private extension StoreStatsAndTopPerformersViewController {
...
-                            onCompletion: ((Error?) -> Void)? = nil) {
+                            onCompletion: ((Result<Void, Error>) -> Void)? = nil) {
         let action = StatsActionV4.retrieveSiteVisitStats(siteID: siteID,
                                                           siteTimezone: siteTimezone,
                                                           timeRange: timeRange,
-                                                          latestDateToInclude: latestDateToInclude) { error in
-                                                            if let error = error {
+                                                          latestDateToInclude: latestDateToInclude,
+                                                          onCompletion: { result in
+                                                            if case let .failure(error) = result {
```
pmusolino, line 351: nit: same here, I suggest using the `switch`

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
```diff
@@ -359,22 +360,23 @@ private extension StoreStatsAndTopPerformersViewController {
...
+                             onCompletion: ((Result<Void, Error>) -> Void)? = nil) {
         let earliestDateToInclude = timeRange.earliestDate(latestDate: latestDateToInclude, siteTimezone: siteTimezone)
         let action = StatsActionV4.retrieveTopEarnerStats(siteID: siteID,
                                                           timeRange: timeRange,
                                                           earliestDateToInclude: earliestDateToInclude,
-                                                          latestDateToInclude: latestDateToInclude) { error in
-                                                            if let error = error {
+                                                          latestDateToInclude: latestDateToInclude,
+                                                          onCompletion: { result in
+                                                            if case let .failure(error) = result {
```
pmusolino: nit: same here, I suggest using the `switch`

Yosemite/Yosemite/Stores/AvailabilityStore.swift
```diff
@@ -43,12 +43,12 @@ private extension AvailabilityStore {
...
                               earliestDateToInclude: Date(),
                               latestDateToInclude: Date(),
-                              quantity: 1) { (_, error) in
-                                if let error = error as? DotcomError, error == .noRestRoute {
-                                    onCompletion(false)
-                                } else {
-                                    onCompletion(true)
-                                }
+                              quantity: 1) { result in
+            if case let .failure(error) = result, error as? DotcomError == .noRestRoute {
```
pmusolino: nit: same here, I suggest using the `switch`

pmusolino (approved):
Code looks good to me and CI is ✅ 👍 thanks for handling this tech debt. 

I just left some comments about the use of `switch` where possible (in some cases you already used it).

WooCommerce/Classes/ViewRelated/Dashboard/Stats v4/StoreStatsAndTopPerformersViewController.swift
ealeksandrov: Do you mean something like this? That's 4 lines instead of 1 🙂
```suggestion
                switch result {
                case .success:
                    break
                case .failure(let error):
```

ealeksandrov:
Thanks for review @pmusolino! I've updated 3 instances of `if case let` to `switch case` - since we have different cases there and it's more readable 👍 

But I'm not sure about 1-line `error` bindings that are left. They are not really `switch`ing since there are no different actions for cases - they are about getting a single case with var. If we refactor them to switch-case it will be 1->4 lines with empty `break` case.

I wasn't able to find anything in code style guidelines. But there is an article with same conclusion that I have: https://useyourloaf.com/blog/swift-if-case-let/

```swift
if case let .failure(error) = result {
    // error logging code
}
// vs
switch result {
case .success:
    break
case .failure(let error):
    // error logging code
}
```

What do you think about it?

ealeksandrov:
I'll merge it for now, so we we can continue discussing style rules and fix them separately 🙂

Issue #3963, opened 2021-04-15
Error synchronizing: "Unable to load content" and "Unable to refresh list" when logged in to a Jetpack site as a user with the Editor role


designsimply:
To prepare:

<sup>TL;DR have a Jetpack site with one Administrator Jetpack-connected account and one Editor Jetpack-connected account.</sup>

* Set up a WooCommerce store hosted somewhere other than WordPress.com.
* Install, activate, and connect Jetpack using a WordPress.com account.
* Go to WP Admin > Users and create an Editor account using a different WordPress.com account email.
* Log out of both WordPress.com and WP Admin.
* Log back in to WP Admin as the user you just created.
* Go to WP Admin > Jetpack Dashboard > Account connection and click the "Link to WordPress.com" link.
* Connect to the WordPress.com with the email address of the Editor account in WP Admin.
* You should end up with something that looks similar to [this screenshot](https://user-images.githubusercontent.com/1119271/114798908-b6d21600-9d53-11eb-936b-3604021753f3.png).

To test: 

1. Log in to the WooCommerce app as a Jetpack-connected user with an Editor role.
1. Browse around in the app.

**Result:** if I log in to the WooCommerce app with a user that is Jetpack-connected but has an Editor role, the app will show "Unable to load content" and "Unable to refresh list" and the app log says "Error synchronizing" multiple times.  ([1m33s](https://cloudup.com/c0NWSqCI4Xw))

```
2021/04/14 19:58:42:104  â›”ï¸ Dashboard (Top Performers) â€” Error synchronizing top earner stats: Dotcom Error: [woocommerce_rest_cannot_view] Sorry, you cannot list resources.
2021/04/14 19:58:42:104  â›”ï¸ Error synchronizing top earners stats: Dotcom Error: [woocommerce_rest_cannot_view] Sorry, you cannot list resources.
```

<details>
<summary>Full app logs for the timestamp matching the video</summary>

```
2021/04/14 19:58:01:812  ðŸ”µ Tracked unified_login_interaction, properties: [AnyHashable("click"): "submit", AnyHashable("source"): "default", AnyHashable("flow"): "wordpress_com", AnyHashable("step"): "start"]
2021/04/14 19:58:02:630  ðŸ”µ Tracked unified_login_step, properties: [AnyHashable("source"): "default", AnyHashable("step"): "start", AnyHashable("flow"): "login_password"]
2021/04/14 19:58:10:565  Received OAuth2 response: {
[...]

PR #4473, opened 2021-06-24
Auth Role Checks: Part 1 - Add User-related networking and store classes

dvdchr:
## Summary

Refs #3963 

This is the first out of three parts that make up the changes related to role authentication (for more details, refer to #3963). Also, note that this PR targets `issue/3963-auth-role-error` branch. 

Here's a summary of this change:
- Added a new `User` model, with its own `Remote`, `Mapper`, `Store`, and `Action`. The `User` model represents a site-specific user information (e.g. `roles` being the most important one). As for the name, I previously considered `SiteUser` and `SiteAccount`, but decided to go ahead with `User` for now. Let me know if you prefer otherwise!
- The user information is fetched via WP Core v2 API, through this endpoint: `/sites/$site_id/users/me`. It's also important that the request includes `{"context": "edit"}`, since the `roles` property is only visible through that context.
- The key information to fetch from the `User` model is the `roles`, and this is why the model is not persisted in the local database – there's no use case for it as of now. We want to make sure that only users with `administrator` or `shop_manager` roles can get into the main part of the app. 
  - In the later pull requests, this information will be stored locally in user defaults in a form of boolean, just to cache the last check result.
  - Ideally, we almost always want the latest state to ensure that the user doesn't stay blocked for too long after their role has been updated by those in charge.

## To Test

There's no UI or flow to test yet, so:

- Run unit tests on Networking module, and make sure everything is 🆗  .
- Run tests on Yosemite module, and make sure everything is 🆗 .

## Author Checklist 
Update release notes:
- [x] Add unit tests where applicable.


Yosemite/YosemiteTests/Stores/UserStoreTests.swift
```diff
@@ -0,0 +1,57 @@
...
+            let action = UserAction.retrieveUser(siteID: self.testSiteID) { result in
+                promise(result)
+            }
+            store.onAction(action)
+        }
+
+        // Then
+        XCTAssertTrue(result.isSuccess)
+    }
+
```
ealeksandrov, line 56: What about additional test case for failure without simulated response (or simulated error)?

Networking/NetworkingTests/Remote/UserRemoteTests.swift
```diff
@@ -0,0 +1,40 @@
...
+            }
+        }
+
+        // Then
+        XCTAssertTrue(result.isSuccess)
+        let user = try XCTUnwrap(result.get())
+        XCTAssertEqual(user.siteID, siteID)
+    }
+
+
```
ealeksandrov, line 42: Same as Network tests - can we add simple failure test case too?

Networking/NetworkingTests/Mapper/UserMapperTests.swift
```diff
@@ -0,0 +1,42 @@
...
+            XCTFail()
+            return
+        }
+
+        XCTAssertEqual(user.localID, 1)
+        XCTAssertEqual(user.username, "test-username")
+        XCTAssertEqual(user.firstName, "Test")
+        XCTAssertEqual(user.lastName, "User")
+        XCTAssertEqual(user.email, "example@example.blog")
+        XCTAssertEqual(user.nickname, "User")
```
ealeksandrov: This won't fail on mistake if you map `lastName` and `nickname` with same data field 🙂
Since we're testing mapping here I suggest unique values.

ealeksandrov (approved):
That's very important bug, thanks for taking it on! 🙇‍♂️
Code looks good (also great comments!) :shipit:

I've left some minor nits about tests for your consideration.

Also I've noticed inconsistent newline use before closing braces (`}`). Per [style guide](https://github.com/raywenderlich/swift-style-guide#spacing) we don't need them.
This rule also means we shouldn't have space after opening braces as well, but the project already has it everywhere in class declarations and I personally prefer some top spacing too 🤔

Networking/NetworkingTests/Mapper/UserMapperTests.swift
dvdchr: Good catch! Will update this shortly.

Networking/NetworkingTests/Remote/UserRemoteTests.swift
dvdchr: Will do! 👍

Yosemite/YosemiteTests/Stores/UserStoreTests.swift
dvdchr: Yeah, makes sense! Will update this shortly.

dvdchr:
> ... inconsistent newline use before closing braces
> ... we shouldn't have space after opening braces as well

Whoops, sorry about this! 😅  I missed this one. Will update this shortly. As for the newline on the top, I think I'll clean it up for now to conform to the agreed upon rule. Thanks for the review, @ealeksandrov! 

-----
**Update: Changes updated to match the style guide!**

Networking/NetworkingTests/Mapper/UserMapperTests.swift
dvdchr: Fixed!

Networking/NetworkingTests/Remote/UserRemoteTests.swift
dvdchr: Negative test added!

Yosemite/YosemiteTests/Stores/UserStoreTests.swift
dvdchr: Negative test added!

PR #4475, opened 2021-06-24
Shipping Labels: UI layer to make purchase and display progress and label printing views

rachelmcr:
Part of: #4089

## Description

This PR hooks up the "Purchase Label" button in the Create Shipping Label screen. When tapped, the button starts the shipping label purchase, displays a progress view, and then displays the existing print shipping label screen when the purchase is complete.

Future PRs will customize the print shipping label screen (currently it is designed only for reprinting labels) and implement an error screen in cases where the purchase fails.

## Changes

ShippingLabelFormViewController:

* Adds `displayPurchaseProgressView()` to display the progress view (using `InProgressViewController`).
* Adds `displayPrintShippingLabelVC()` to display the print shipping label screen (using `ReprintShippingLabelCoordinator`/`ReprintShippingLabelViewController` for now).
* Hooks up the "Purchase Label" button to display the progress view, make the purchase, and display the print shipping label view if the purchase is successful.

ShippingLabelFormViewModel:

* Adds `purchaseLabel()` to put together the details required for the label purchase, make the purchase, and save the purchased label details when successful.

## Testing

Make sure you have configured the WooCommerce WCShip plugin, and you have one or more payment methods set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
Launch the app in debug mode (because of the feature flag).
Open an order detail that is eligible for the shipping label.
Tap the button "Create Shipping Label".
Complete all the fields in the form.
In the Order Summary section, tap the button "Purchase Label".
Confirm an in-progress view appears that says "Purchasing Label".
When the purchase is complete, if it is successful, confirm the "Print Shipping Label" screen appears.

## Submitter Checklist

Update release notes:


pmusolino:
I tested it @rachelmcr, but I get these errors in the console, and I wasn't able to complete the entire process.
I tested it also a second time, and I get always the same error.

```
2021-06-25 12:05:20.481266+0200 WooCommerce[3977:99781] <> Mapping Error: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "tracking", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
2021-06-25 12:05:23.068428+0200 WooCommerce[3977:99781] <> Mapping Error: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "tracking", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
2021-06-25 12:05:25.691167+0200 WooCommerce[3977:99921] <> Mapping Error: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "tracking", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
2021-06-25 12:05:28.297227+0200 WooCommerce[3977:99922] <> Mapping Error: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "tracking", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
2021-06-25 12:05:28.297917+0200 WooCommerce[3977:99922] ⛔️ Error checking shipping label status for order 1567: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "tracking", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
```

rachelmcr:
Hm, that's odd! I thought I had `tracking` set up to handle null values, but maybe I missed something there. I'll take a look, thanks!

pmusolino:
If it can help, let me know if you need access to my testing website @rachelmcr 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -71,6 +71,10 @@ final class ShippingLabelFormViewModel {
     ///
     var shippingLabelAccountSettings: ShippingLabelAccountSettings?
 
+    /// Shipping Label Purchase
+    ///
+    var purchasedShippingLabel: ShippingLabel?
```
Ecarrion: nit: I think this can be `private(set)` right?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -500,6 +504,34 @@ extension ShippingLabelFormViewModel {
         case validationError(ShippingLabelAddressValidationError)
         case genericError(Error)
     }
+
+    func purchaseLabel(onCompletion: @escaping ((Bool) -> Void)) {
```
Ecarrion: What do you think about returning `Result<Void, Error>` that way we can switch over a semantic meaning(success, error) rather than a boolean that can mean anything! 

Also since this method is public it would be nice to add some docs to it!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -500,6 +504,34 @@ extension ShippingLabelFormViewModel {
...
+        guard let originAddress = originAddress,
+              let destinationAddress = destinationAddress,
+              let selectedPackage = selectedPackage,
+              let selectedRate = selectedRate,
+              let accountSettings = shippingLabelAccountSettings else {
+            return
```
Ecarrion, line 550: ⚠️  I think this should also invoke the completion block right?

Ecarrion (commented):
Flow worked for me but I also noticed that some parsing errors when purchasing the label.

```
<> Mapping Error: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labels", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "tracking", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
```

Apart from that only left one concern marked with ⚠️ 
But the rest looks good!

rachelmcr:
@pmusolino @Ecarrion Thanks to both of you for looking at this one! I'm a little stuck with this mapping error and I wonder if you have advice for how to proceed.

I haven't been able to reproduce the repeated error and failed purchase that Paolo got, but I do see a single mapping error show up when I make a successful purchase, and if I understand correctly, here's what's happening:

* When we start the purchase (with `ShippingLabelRemote. purchaseShippingLabel`), we expect to get back a `ShippingLabelPurchase` in the response, so [we use the ShippingLabelPurchaseMapper](https://github.com/woocommerce/woocommerce-ios/blob/1ad7361547e0788d48d6bd1f0ab74ee3748f2a14/Networking/Networking/Remote/ShippingLabelRemote.swift#L267). That `ShippingLabelPurchase` is a subset of the full details we know we need for a complete shipping label (represented with the `ShippingLabel` model), and it's what we get as long as the shipping label purchase isn't complete. (We talked about adding that `ShippingLabelPurchase` model in Slack, internal ref: p1623678486441500-slack-C6H8C3G23.)
* When we start polling for the label status (with `ShippingLabelRemote.checkLabelStatus`), [we use the ShippingLabelStatusMapper](https://github.com/woocommerce/woocommerce-ios/blob/1ad7361547e0788d48d6bd1f0ab74ee3748f2a14/Networking/Networking/Remote/ShippingLabelRemote.swift#L291), which expects a `ShippingLabel` because when the purchase is complete that's what it will return. But if the purchase hasn't yet completed, it will only return the data in a `ShippingLabelPurchase`, which causes that mapping error.

Because `ShippingLabelRemote.checkLabelStatus` could return either a `ShippingLabelPurchase` or a `ShippingLabel`, is there a good way to handle that potentially different response? My first thought is to update the `ShippingLabel` model to allow for some optional values, but keep those values as non-optional in Storage — that way when we're handling an API response we recognize that the label data could be incomplete, but when we're upserting a `ShippingLabel` in storage we know we expect all those values to be there. Does that make sense or is there another way to do this?

rachelmcr:
I'm going to go ahead and try fixing the mapping error with this approach:

> My first thought is to update the ShippingLabel model to allow for some optional values, but keep those values as non-optional in Storage

And then we can take a look at this again, and see if there are remaining issues that need to be resolved.

rachelmcr:
I have a fix for that mapping error in 33f7f6a, but let me know if you think there's a better way to do this. It's a little awkward because there are places in the UI where we rely on that model, so now those properties are optional in the UI as well. I'm not sure if there's a better way to handle it given that we don't expect those to be optional properties for a complete shipping label after a successful purchase.

Ecarrion:
Hi @rachelmcr! sorry for taking so long to answer this, totally missed the first notification 🤦.

If the response could be one or the other, could we not create an enum similar to 
```swift
enum LabelPollingResponse {
   case pending(ShippingLabelPurchase)
   case purchased(ShippingLabel)
   
   var commonField: String {
       switch self {
           // return correct value
       }
   }
}
```

And then make `ShippingLabelRemote.checkLabelStatus` return that enum?
After the label is purchased, you can continue with the `ShippingLabel` concrete type for the subsequent screens.


Also, since that seems to be unrelated to this PR I think it would be safer to solve it in a different PR/Issue.

Also cc @pmusolino in case I'm missing some context

pmusolino:
@rachelmcr do we have a case where we do not have a `rate`, or have we made `rate` optional just for safety? I think showing always zero when the rate is `null` can be dangerous because, in theory, we will allow users to buy shipping labels at a wrong rate (not sure if in this case, API allows us to continue).

rachelmcr:
@Ecarrion Thank you for that suggestion! That solution makes a lot of sense and didn't occur to me at all. It would be a lot safer than making those `ShippingLabel` properties optional. (@pmusolino The `rate` isn't returned until the label purchase is complete, which is why I made it optional, but you're right that it could be potentially problematic. I don't think it would matter for the purchase itself, but it could do weird things in other places.)

It sounds like the best thing here would be to revert my last commit (where I changed the `ShippingLabel` model), and then fix the mapping error in a followup PR as Ernesto suggested.

pmusolino:
@rachelmcr thanks Rachel for explaining this. If it's not long and complex to solve, I would fix the mapping error before modifying the `ShippingLabel` model. Otherwise, we can take it as is, and fix it later.

pmusolino:
Could we proceed with the review? @rachelmcr 🙇

rachelmcr:
@pmusolino Yes, sorry for the delay! I reverted the changes to the `ShippingLabel` model (in b6f9419) so it's ready to be reviewed again, and I'll fix that mapping error in another PR.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -281,7 +281,17 @@ private extension ShippingLabelFormViewController {
...
             // TODO: Handle order completion
         } onButtonTouchUp: {
-            // TODO: Purchase Label action
+            self.displayPurchaseProgressView()
+            self.viewModel.purchaseLabel { [weak self] result in
+                switch result {
+                case .success:
+                    self?.displayPrintShippingLabelVC()
+                case .failure:
+                    // TODO: Implement and display error screen for purchase failures
```
pmusolino, line 290: Should we add this task as part of the list? https://github.com/woocommerce/woocommerce-ios/issues/4089

pmusolino (approved):
Looks good to me 👏 

I have just two non-blocking questions (we can handle them in later PRs):
1) I just noticed that in the `Print Shipping Label` screen the back button pop to the previous screen, so we are allowed to create an identical shipping label. In this case, we should pop to the order detail screen.
2) It seems that the newly purchased shipping label is somehow not connected to the order, as it does not appear under the order, and it is possible to buy a new shipping label.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
rachelmcr: Yep, it's there! This is the `UI for failed purchase` task in that list.

rachelmcr:
Thanks for your help and review on this!

> I just noticed that in the Print Shipping Label screen the back button pop to the previous screen, so we are allowed to create an identical shipping label. In this case, we should pop to the order detail screen.

Yeah, I noticed that too while we were discussing this PR — I'll fix that behavior as I work on customizing that Print Shipping Label screen.

> It seems that the newly purchased shipping label is somehow not connected to the order, as it does not appear under the order, and it is possible to buy a new shipping label.

Oh hm, that's odd. I thought I saw my purchased labels appear in the order detail, but I might have been mistaken. I'll make a note of that on the issue for this task, to make sure I check on that.

I see there are some merge conflicts now, so I'll resolve those and then merge this one.

PR #4476, opened 2021-06-24
Add-Ons: Fix add-on tests by using a custom currency formatter

Ecarrion:
# Why

Tests were not using a fixed currency formatter which made them inconsistent when being run in setups with different locates.

# How

It has been fixed by using a fixed currency formatter for tests.

Update release notes:


pmusolino (approved):
Thank you for fixing this @Ecarrion :shipit:

Issue #4464, opened 2021-06-22
Force the country to US in Shipping Labels


pmusolino:
Before releasing the Milestone 3, we should force the Country in the Shipping Label Address (`Ship To` - `Ship From` sections) to `US` (since in this milestone we support just US), and we should allow the user to select the state from a list.

Feel free to split this issue into subtasks.

Tasks:
- [x] Implement the networking layer for supporting `/countries`. [Doc](https://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-continents)
- [x] Implement the storage layer for storing the countries.
- [x] Implement the Yosemite layer for fetching and storing the countries.
- [x] Implement the screen that displays the list of states, tapping on the `State` field in the Shipping Label address validation.

PR #4477, opened 2021-06-24
Shipping Labels: storage and yosemite layers for fetching and storing the countries

pmusolino:
Part of #4464

## Description
This PR adds the Storage and the Yosemite layers for fetching and storing the Countries, that will be used in Shipping Label Address Validation. All those lines of code are because of the new model n°53 introduced in Core Data.

## Testing
Since the code implemented here is still not used in any UI, just CI is enough.

Update release notes:


Yosemite/Yosemite/Stores/DataStore.swift
```diff
@@ -0,0 +1,99 @@
...
+            guard let self = self else { return }
+
+            switch result {
+            case .failure(let error):
+                completion(.failure(error))
+            case .success(let countries):
+                self.insertCountriesInBackground(countries: countries) {
+                    completion(.success(countries))
+                }
+            }
```
ealeksandrov, line 61: nit: success (primary) case should be first in a switch

Yosemite/Yosemite/Actions/DataAction.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+// MARK: - DataAction: Defines the actions that allows you to view all types of data available.
+//
+public enum DataAction: Action {
```
ealeksandrov, line 5: Can we add clarification in a comment that it's about `/data/` APIs? Maybe say "WC data" with a link to API doc?
Otherwise this name and description look very generic.

Yosemite/Yosemite/Stores/DataStore.swift
```diff
@@ -0,0 +1,99 @@
+import Foundation
+import Networking
+import Storage
+
+// MARK: - DataStore
+//
+public final class DataStore: Store {
```
ealeksandrov, line 7: Same as `DataAction` above - name without context is too generic and confusing, consider clarifying in a comment, maybe with docs link.

ealeksandrov (approved):
Looks good! Just added some minor comments about clarifying scope of new Action/Store. :shipit:

PR #4478, opened 2021-06-24
Auth Role Checks: Part 2 - Add user interface for role error page

dvdchr:
## Summary

Refs #3963, see also: `KW82M1xZbvqdkuoQ71Q5le-fi-1352%3A0`

This is the second out of 3 parts that make up the changes related to role authentication (for more details, refer to #3963). Also, note that this PR targets `issue/3963-auth-role-error` branch.

Here's a summary of the change:

- Added illustration assets, view controller, and the view model counterpart for the role error page.
  - A note on why the Role Error page needs its own view stack: is because they are *slightly* different than the existing error pages; there's a slightly different component based on the designs. 
  - Secondly, it's slightly different from how `ULErrorViewController` is used. The Role Error page does not interact with WordPressAuthenticator module. 
  - Finally, there's also some additional business logic involved (apart from simple navigation actions from tapping the action buttons), like making a network request to check the current user's role eligibility.
- The Role Error page is not hooked anywhere – it is currently unreachable. It will be hooked properly into the flow in the last pull request.
- Some parts in the `RoleErrorViewModel` file will be left empty (e.g. the Retry button logic). This is intentional, and the implementation will be added in later parts as implementation from the previous parts become available once they are merged.

---
**Note:** Please excuse the large changeset warning 😅 . More than half of it comes from the XIB and unit test file.

---
## Screenshots 👀 

• | Light 🌝  | Dark 🌚 
-- | -- | --
iPhone, Portrait | 4478_iphone_portrait_light | 4478_iphone_portrait_dark
iPhone, Landscape | 4478_iphone_landscape_light | 4478_iphone_landscape_dark
iPad, Portrait | 4478_ipad_portrait_light | 4478_ipad_portrait_dark
iPad, Landscape | 4478_ipad_landscape_light | 4478_ipad_landscape_dark

## Remaining Tasks

- [x] Import light & dark illustrations into assets.
- [x] Create the XIB file.
- [x] Create the view controller counterpart, and hook it up with the XIB file.
- [x] Stylize the view components to match the design.
- [x] Create the view model, and hook it up to the view controller.
- [x] Implement action for the link button to open the web page. 
- [x] Handle other layout cases (landscape, tablet, etc.)
- [x] Add tests where applicable.
- [x] Review accessibility.
- [x] Add screenshots from the user interface implementation.
- [x] Complete the To Test section below.

### 1. Testing the UI

Currently, there's no path present to reach the new role error page. For the sake of just testing the layout, you'd have to intercept with the navigation flow temporarily. Modify `AppCoordinator` like so:

```swift
// AppCoordinator.swift:L38

// add these lines to always show the role error page.
let viewModel = RoleErrorViewModel(displayName: "John Doe", roles: ["editor"])
let viewController = RoleErrorViewController(viewModel: viewModel)
let navigationController = WooNavigationController(rootViewController: viewController)
setWindowRootViewControllerAndAnimateIfNeeded(navigationController)
self.isLoggedIn = isLoggedIn
return
```
When running the app again, you should see the role error page pop up first thing.

### 2. Running the unit tests

Make sure all tests are 🟢 . New tests are added in `RoleErrorViewModelTests.swift`.

## Author Checklist

- [x] Add unit tests where applicable.


WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewModel.swift
```diff
@@ -0,0 +1,85 @@
...
+                                                    + "Presented when the user has tries to switch to a store with incorrect permissions.")
+
+    static let primaryButtonTitle = NSLocalizedString("Retry",
+                                                      comment: "Action button that will recheck whether user has sufficient permissions to manage the store."
+                                                        + "Presented when the user tries to switch to a store with incorrect permissions.")
+
+    static let secondaryButtonTitle = NSLocalizedString("Log In With Another Account",
+                                                        comment: "Action button that will restart the login flow."
+                                                        + "Presented when logging in with a site address that does not have a valid Jetpack installation")
+}
```
dvdchr: This is something new that I'm trying. Instead of wrapping it in an `enum` or `struct`, doing it this way kinda shortens the usage like so:
```swift
text = .errorMessageText
```

compared to:
```swift
text = Localization.errorMessageText
```
Let me know if you have any thoughts about this! 🙂

WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
```diff
@@ -0,0 +1,161 @@
...
+    func refreshTitleLabels() {
+        // these are the only components that may change on runtime, e.g. the role changed
+        // to something else, but still incorrect; or the user updated their display name.
+        titleLabel.text = viewModel.nameText
+        subtitleLabel.text = viewModel.roleText
+    }
+
+    func displayWebContent(for url: URL) {
+        let safariViewController = SFSafariViewController(url: url)
+        present(safariViewController, animated: true, completion: nil)
```
koke: I see that we already use `SFSafariViewController` directly in several places, but there's also `WebviewHelper.launch(_:with:)` as a helper to show a URL.

WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewModel.swift
```diff
@@ -0,0 +1,102 @@
...
+
+    func didTapAuxiliaryButton() {
+        output?.displayWebContent(for: linkDestinationURL)
+    }
+
+    // MARK: Private Helpers
+
+    /// try to "humanize" the roles. e.g., "shop_manager" to "Shop Manager".
+    /// given multiple roles, it will all be joined with a comma separator.
+    private func humanizedRolesText(from roles: [String]) -> String {
```
koke, line 72: I don't see a way to get localized role descriptions from the API, but I think we should explicitly handle the core roles and localize them, specially since the other copy in the screen will localize the referenced roles.

This function might still be useful for any roles that we don't recognize

koke (commented):
The UI looks good, but I think we should localize the current roles if we can

WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
dvdchr: Done! Thanks for the tip. I completely missed the existence of `WebviewHelper` 😅

WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewModel.swift
dvdchr: > ... handle the core roles and localize them,
> This function might still be useful for any roles that we don't recognize

Good point! 👍 I'll make the changes on #4492 since I've refactored that function and also introduced `User+Roles` to sort of encapsulate the roles validation stuff.

koke (approved):
>  I'll make the changes on #4492
Sounds good to me 🚢

PR #4479, opened 2021-06-24
[Mobile Payments] Unbreak modals in landscape just a tad

### LINKED ISSUE #4474: [Mobile Payments] Modals in landscape
Noticed during the review of 4467. https://github.com/woocommerce/woocommerce-ios/pull/4467#pullrequestreview-691575598

The modals are off in landscape.

ctarda:
Closes #4474 

During the review of #4467 we noticed that some modals were severely broken in landscape. Such is life.

This PR attempts to unbreak those modals a little bit. Yes, the width is still not constant, but we are getting closer.

| Before | After |
| ---- | ---- |
|  |  |
|  |  |
|  |  |

## Changes
* Changed some constraints and changed the distribution property of the stack view in landscape mode

## How to test
1. Set up a store for In-Person Payments:  P91TBi-5ee-p2
2. Connect to a reader.
3. Navigate to an order eligible for In-Person Payments
4. Go thought the flow to collect a payment.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.swift
@@ -63,16 +63,15 @@ final class CardPresentPaymentsModalViewController: UIViewController {
     private func resetHeightAndWidth() {
         if traitCollection.containsTraits(in: UITraitCollection(verticalSizeClass: .compact)) {
             mainStackView.axis = .horizontal
-            mainStackView.distribution = .fillProportionally
             heightConstraint.constant = Constants.modalWidth
             widthConstraint.constant = Constants.modalHeight
         } else {
             mainStackView.axis = .vertical
-            mainStackView.distribution = .fill
             heightConstraint.constant = Constants.modalHeight
             widthConstraint.constant = Constants.modalWidth
         }
 
+        mainStackView.distribution = .fill
         heightConstraint.priority = .defaultHigh
         widthConstraint.priority = .defaultHigh
     }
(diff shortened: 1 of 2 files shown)


allendav:
Tested on iPad (10.2") and iPhone XR. Some button labels were cut off, and some images missing or scaled very small

iPhone

iPhone5
iPhone4
iPhone3
iPhone2
iPhone1

iPad

iPad3
iPad2
iPad1

ghost:
I think having 3 columns in the landscape modals could cause a few problems down the road and it looks like there will be problems with text and button cutoffs. Hiding the illustrations might be the way to go for these. I put together a quick concept where we could have 2 buttons full width and then the 3rd action below. Also option for having the illustration or without. Let me know your thoughts @ctarda @allendav @Garance91540 

Screen Shot 2021-06-28 at 5 45 47 pm

ctarda:
@adamzelinski How would you render the error messages in landscape mode?

Error messages might contain, from top to bottom, two lines of text, an illustration, an error message (that will most likely span across two lines) and one or two buttons. An example in portrait mode could be this:

ghost:
@ctarda I think something like this would work

Screen Shot 2021-06-29 at 4 18 32 pm

Garance91540:
Hi everyone! Aaaah landscape mode, always a tough one 😅. Since we are losing a significant amount of real estate in that mode, I think less is more here. 

Removing the illustration on landscape mode is my preferred option:  

Screen Shot 2021-06-29 at 2 48 39 PM

Garance91540 (changes requested):
Yes it's a bit dry, but I think it keeps the information displayed clear and prevents any confusion

ctarda:
Closing for now.

PR #4486, opened 2021-06-28
Fix: new model WooCountry and StateOfAWooCountry and new core data model n°54

pmusolino:
Part of #4464 

I discovered that we already have a declaration of a `Country` entity in `WordPressKit`, so when I started to use our Country entity in the WooCommerce target of the app, I get this compiler error
schermata_2021-06-25_alle_18 27 03

To solve this issue, I renamed the `Country` entity to `WooCountry` and the `StateOfACountry` to `StateOfAWooCountry`. I also introduced a new core data model n° 54 with these changes.

It would be great to merge this in our release 7.0.0.

## Testing
- Just CI

Update release notes:


ealeksandrov (approved):
Change is straightforward rename and LGTM.
But before merging please consider possible https://github.com/wordpress-mobile/WordPressKit-iOS fix, that is discussed in p1624878613012800-slack-C70PAM3RA

PR #4492, opened 2021-06-29
Auth Role Checks: Part 3 - Add role eligibility use case

dvdchr:
Refs #3963
Depends on #4619 👈 please review this first! :)
Depended by #4493 

## Summary

This is the third out of (apparently) 4 parts that makes up the changes related to role authentication (for more details, refer to #3963). Note that the contents of this PR will not affect the current state of the app, as it hasn't modified application flow (yet). Changes to the flow will be introduced in the last PR.

Here's a summary of the changes:

- Adds a `RoleEligibilityUseCase`, a component that encapsulates the logic behind role eligibility checking and provides support for persisting any error information to be read on every app launch later on.
- The error information is encapsulated within `EligibilityErrorInfo` structure, which saves the name and roles of the user that failed the eligibility check. This is to be used for display purposes in the role error UI.
- On persistence: The error information is stored in user defaults because the data is quite simple, and the usage is quite limited. The use case only persists error information when:
  1. role eligibility check fails, and 
  2. there can only be one or zero error information at any time (zero meaning that the previous check was successful _for the current store_; this will be explained in the last PR).
- Added a lot of tests!

---
**Note:** This PR targets `task/3963-role-error-ui` branch as this depends that PR. The changes would be too large to review if it were to target the original issue branch, `issue/3963-auth-role-error`. Once `task/3963-role-error-ui` is merged, the base branch for this PR will be updated to target `issue/3963-auth-role-error` – or even `develop` instead.

---

## Related PRs

- #4473 
- #4478 
- #4619
- #4493 

## To Test

- Since there are no UI changes, make sure all the tests are 🟢 . 

## Author Checklist

- [x] I have added tests where necessary.


WooCommerce/Classes/Authentication/RoleEligibilityUseCase.swift
```diff
@@ -0,0 +1,159 @@
...
+                // i.e., if the user is previously ineligible, then assume they are still ineligible now.
+                return
+            }
+
+            // store the user information locally.
+            self.lastErrorInfo = errorInfo
+        }
+    }
+
+    func checkEligibility(for storeID: Int64, completion: @escaping (RoleEligibilityError?) -> Void) {
```
ealeksandrov: What do you think about using Result here? Maybe `.success` will better communicate meaning of nil error?

```suggestion
    func checkEligibility(for storeID: Int64, completion: @escaping (Result<Void, RoleEligibilityError>) -> Void) {
```

WooCommerce/Classes/Model/EligibilityErrorInfo.swift
```diff
@@ -0,0 +1,62 @@
+import Foundation
+
+/// Convenience type that encapsulates display information for user ineligible for managing the store.
```
ealeksandrov, line 3: 👏

ealeksandrov (approved):
LGTM :shipit:
Just one API readability idea for your consideration.

WooCommerce/Classes/Authentication/RoleEligibilityUseCase.swift
dvdchr: Thank you for the suggestion! 👍  I've noticed before, that in the codebase we kinda have two styles to tackle _errorable_ void results – either with an optional error parameter, or with a `Void, Error` result type (as you've suggested).

However, after receiving similar feedback [here](https://github.com/woocommerce/woocommerce-ios/pull/4493#discussion_r662583600), I guess it makes sense to use the result type instead. Will update this soon!

dvdchr: Resolved in [f2e1ccc](https://github.com/woocommerce/woocommerce-ios/pull/4492/commits/f2e1cccf92c216a97dcf4ab30806c9024fd2330e)! Sorry it took a while. 🙂

dvdchr:
Next steps: I'll merge `issue/3963-auth-role-error` branch (containing Part 1 and Part 2 already) to `develop` first, so it wouldn't create such a large integration branch. Once Part 1 and Part 2 is merged to develop, this will be merged to develop next! I'll get it done this week.

shiki:
Thank you for dedicating more time to this, @dvdchr! 🙏

PR #4494, opened 2021-06-29
Add stats tab basic throttling

ealeksandrov:
## Description

This PR adds 2 improvements on Stats tab sync logic.
1. It prevents triggering new sync unless previous finished.
2. It adds 30 s delay before new refresh can be triggered. This applies only for tab switch update, manual pull-to-refresh will still trigger sync anyway.

## Test

1. Run proxy or add couple of `print("sync")` in `StoreStatsAndTopPerformersViewController` -> `syncStats`/`syncSiteVisitStats`/`syncTopEarnersStats`.
2. Check that stats loaded successfully.
3. Try pull-to-refresh, observe it triggers networking.
4. Try switching to products and back - observe no sync networking (there will be 1 request checking availability - subject for a different PR)
5. Wait >30 s and switch tabs - observe stats networking (12+ API requests)

---

Update release notes:


Ecarrion (approved):
Simple but seems to do the job! 🚀 

A small nitpick would be to encapsulate the logic & state in some kind of `SyncingState` to make it a bit more maintainable!

ealeksandrov:
Thanks for review @Ecarrion! I agree that separate entity for sync/throttling logic would be nice, but stats availability check here gives hard time integrating generic implementation. I've tried single solution for both stats and orders and it ate a lot of time with not-very-nice result 😄
I think new app-wide features availability check will help simplify this and refactor later!

shiki:
I just tested this. Great work, @ealeksandrov! 🎉 

It looks like there are still a few more network requests happening when switching tabs. But definitely significantly less than before. What's the reason why these weren't included in the throttling? It might be good to document these for the future. 

https://user-images.githubusercontent.com/198826/125147304-a5e13880-e0e7-11eb-848b-e2d133a9b03a.mov

PR #4495, opened 2021-06-29
Handle optional `delivery_days` field in Shipping Labels Carrier and Rates

pmusolino:
It seems that the field `delivery_days` returned in Shipping Labels Carrier and Rates sometimes can be `null`. So, in this PR I changed the property to make it optional.

Discussion here: C026HH75ECA/p1624969958004300

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRowViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRowViewModel.swift
@@ -47,7 +47,7 @@ struct ShippingLabelCarrierRowViewModel: Identifiable {
 
         title = rate.title
         let formatString = rate.deliveryDays == 1 ? Localization.businessDaySingular : Localization.businessDaysPlural
-        subtitle = String(format: formatString, rate.deliveryDays)
+        subtitle = String(format: formatString, rate.deliveryDays?.description ?? "-")
 
         let currencyFormatter = CurrencyFormatter(currencySettings: currencySettings)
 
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarrierRowViewModel.swift
```diff
@@ -47,7 +47,7 @@ struct ShippingLabelCarrierRowViewModel: Identifiable {
 
         title = rate.title
         let formatString = rate.deliveryDays == 1 ? Localization.businessDaySingular : Localization.businessDaysPlural
-        subtitle = String(format: formatString, rate.deliveryDays)
+        subtitle = String(format: formatString, rate.deliveryDays?.description ?? "-")
```
rachelmcr: This works, but what do you think about checking if `rate.deliveryDays` is null and, in that case, hiding the subtitle in the `ShippingLabelCarrierRow` view entirely? It looks like that's what we do in wp-admin:

Screen Shot 2021-06-29 at 3 55 36 PM

Otherwise, we end up with "- business days" which is a little unclear:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -211,7 +211,7 @@ final class ShippingLabelFormViewModel {
         let price = currencyFormatter.formatAmount(Decimal(rate)) ?? ""
 
         let formatString = selectedRate.deliveryDays == 1 ? Localization.businessDaySingular : Localization.businessDaysPlural
-        let shippingDays = String(format: formatString, selectedRate.deliveryDays)
+        let shippingDays = String(format: formatString, selectedRate.deliveryDays?.description ?? "-")
```
rachelmcr: Same as above; what about not displaying `shippingDays` if `selectedRate.deliveryDays` is null?

rachelmcr (commented):
Code looks good and works as expected; I just have a question about what we want to show to users in cases where `delivery_days` is `null`.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: It makes more sense. I updated the code here 88bac43ca1d914226f29bbb98f01b4e6daf7f224
👍

rachelmcr (approved):
That change looks great!
Code ✅ 
Testing ✅

PR #4499, opened 2021-06-30
[Mobile Payments] Add actions for remembering, forgetting and recalling card readers

allendav:

This PR adds the actions needed for the feature. Although there are no user facing changes there are unit test changes and new unit tests to cover the actions.

Note: `rake generate` was required for sourcery copiable

The sequence of planned PRs can be found here: https://github.com/woocommerce/woocommerce-ios/issues/3559#issuecomment-868872729

The next PR will define a protocol that a subsequent PR will create a conforming object for which will initiate these actions.

To test:
- Nothing user facing to test
- Ensure linting passes
- Ensure new unit tests pass in AppSettingsStoreTests_CardReaderSettings
- Ensure existing unit tests pass especially GeneralAppSettingsTests and InAppFeedbackCardVisibilityUseCaseTests

Related:

p91TBi-5t5-p2

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/AppSettingsStore.swift
+++ b/Yosemite/Yosemite/Stores/AppSettingsStore.swift
@@ -157,6 +157,12 @@ public class AppSettingsStore: Store {
             setOrderAddOnsFeatureSwitchState(isEnabled: isEnabled, onCompletion: onCompletion)
         case .loadOrderAddOnsSwitchState(onCompletion: let onCompletion):
             loadOrderAddOnsSwitchState(onCompletion: onCompletion)
+        case .rememberCardReader(cardReaderID: let cardReaderID, onCompletion: let onCompletion):
+            rememberCardReader(cardReaderID: cardReaderID, onCompletion: onCompletion)
+        case .forgetCardReader(cardReaderID: let cardReaderID, onCompletion: let onCompletion):
+            forgetCardReader(cardReaderID: cardReaderID, onCompletion: onCompletion)
+        case .loadCardReaders(onCompletion: let onCompletion):
+            loadCardReaders(onCompletion: onCompletion)
         }
     }
 }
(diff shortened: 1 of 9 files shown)


Yosemite/Yosemite/Stores/AppSettingsStore.swift
```diff
@@ -247,6 +253,61 @@ private extension AppSettingsStore {
...
+//
+private extension AppSettingsStore {
+    /// Remember the given card reader (to support automatic reconnection)
+    /// where `cardReaderID` is a String e.g. "CHB204909005931"
+    ///
+    func rememberCardReader(cardReaderID: String, onCompletion: (Result<Void, Error>) -> Void) {
+        do {
+            let settings = loadOrCreateGeneralAppSettings()
+
+            guard !settings.knownCardReaders.contains(cardReaderID) else {
```
koke, line 266: One thing that might simplify this a bit and remove the need for these checks is to store a `Set` instead of an `Array`, unless we actually want to store the order in which the readers were remembered.

```swift
  8> let knownReaders = Set(["a", "b"]) 
knownReaders: Set<String> = 2 values {
  [0] = "a"
  [1] = "b"
}
  9> knownReaders.subtracting(["a"]) 
$R5: Set<String> = 1 value {
  [0] = "b"
}
 10> knownReaders.subtracting(["d"]) 
$R6: Set<String> = 2 values {
  [0] = "a"
  [1] = "b"
}
 11> knownReaders.union(["a"]) 
$R7: Set<String> = 2 values {
  [0] = "a"
  [1] = "b"
}
 12> knownReaders.union(["c"]) 
$R8: Set<String> = 3 values {
  [0] = "a"
  [1] = "b"
  [2] = "c"
}
```

koke (approved):
Code looks good to me. I wonder if we should actually store all known readers or just the last one we connected to. What would happen if a Bluetooth scan shows more than one known reader? Even if we want to remember more than one, do we want to prioritise the most recent one?

allendav:
@koke wrote:

> Code looks good to me. I wonder if we should actually store all known readers or just the last one we connected to.

My first inclination was to only store one, but the consensus seemed to be to store multiple.

> What would happen if a Bluetooth scan shows more than one known reader? Even if we want to remember more than one, do we want to prioritise the most recent one?

Yes, we may, or we may want to rely on RSSI if available. I think by doing it as an array we can keep our options open here - and prefer readers newly added to the array over earlier ones.

Yosemite/Yosemite/Stores/AppSettingsStore.swift
allendav: Let's stick with Array since the position of the reader in the Array gives us a proxy for "added most recently, ok?

PR #4504, opened 2021-06-30
Fix issue with interactive pop gesture in WooNavigationController

### LINKED ISSUE #4116: Order Details & Settings: Swiping left to right doesn't let you go back to the order list
**Describe the bug**
When in a navigation view, swiping from left to right should allow you to move back up one level, like tapping the back button. That gesture does not seem to work on Order Details and the Settings screens.

**To Reproduce**
1. Go to an order detail or the settings screen.
2. Try to swipe from left to right to go back a screen.

This behavior does work on the Product Details screen.

**Expected behavior**
The left to right swipe gesture should let you go backwards in the navigation stack.

**Mobile Environment**
Please include:
- Device: iPhone 12 Pro Max, iPhone 11 Pro
- iOS version: 14.5
- WooCommerce iOS version: 6.5 & 6.6 (TestFlight)

itsmeichigo:
# Description
This PR fixes #4116: Currently interactive pop gesture is not working on Settings and Order Details screens. In other words, swiping on the left edge of these screens cannot trigger popping of the view controllers, making it feel less intuitive.

# Solution
It's noticeable that both the Order Details and Settings screens use custom navigation controllers with customized UI styling using delegates. `WooNavigationControllerDelegate` is used to forward some events of `UINavigationControllerDelegate` to children delegate to handle the customization. 

There are suspicious events that shouldn't be forwarded / customized: `navigationController(_:animationControllerFor:from:to:)` and `navigationController(_:interactionControllerFor:)`, as Apple Documentation says, "Implement this delegate method when you want to provide a custom animated transition between view controllers as they are added to or removed from the navigation stack." 

In our case, there's no need for custom transitions, and since `nil` is returned for the animation controller, this breaks the default behavior of the interactive pop gesture. These delegate implementations are therefore removed to fix the issue.

Note: this bug also happens on Product Review screen for similar reason.

# Testing 
- Order Details and Settings screens' UI should stay unchanged
- It should be possible to swipe to dismiss these screens.

Update release notes:


WooCommerce/Classes/System/WooNavigationController.swift
```diff
@@ -63,27 +62,6 @@ private class WooNavigationControllerDelegate: NSObject, UINavigationControllerD
...
-    ///
-    func navigationController(_ navigationController: UINavigationController,
-                              animationControllerFor operation: UINavigationController.Operation,
-                              from fromVC: UIViewController,
-                              to toVC: UIViewController) -> UIViewControllerAnimatedTransitioning? {
-        forwardDelegate?.navigationController?(navigationController,
-                                               animationControllerFor: operation,
-                                               from: fromVC,
-                                               to: toVC)
-    }
```
pmusolino, line 86: It seems that this code was added here https://github.com/woocommerce/woocommerce-ios/pull/3872 by Ernesto, but I'm not sure it is currently used.
@Ecarrion as far as you remember, removing this code, does something unexpected happen in other parts of the app?

Ecarrion: Interesting problem!

I totally missed that providing an empty implementation for these methods breaks the interactive pop gesture. 🤦 

These methods are implemented and forwarded to allow subclasses of `WooNavigationController.swift` to respond to `UINavigationControllerDelegate` as needed.

By deleting these two methods we won't allow subclasses to provide a custom animation/interaction, which is something we don't need at the moment, thankfully.

As solving this problem is not trivial, I think the quickest win is to delete these two methods and to add a comment on why they are not present anymore. As well as linking/creating a GH issue for this problem.

What do you think @pmusolino @itsmeichigo ?

itsmeichigo: @Ecarrion So sorry for not explicitly having mentioned you in this issue earlier 🙇‍♀️ I actually believe that since the implementation was not intended, it's safe to just remove it for now. In the future if we wish to customize transition animations, it'll be part of the task and not what we have to worry about for now. Does that make sense?

Ecarrion: That makes sense! But I still think we should add a comment explaining why those methods are missing and probably best to create an issue with the `tech debt` label so we can monitor and remember it.

itsmeichigo: @Ecarrion I honestly don't think this is a tech debt because we don't need that for now. I mean those delegate methods are optional, so the person who takes care of the transition animation in the future should be able to add them in as part of the task.

pmusolino: It makes sense to me to remove the code and about the comment, it costs us nothing to do it. I'm not sure about the issue, since, like said by @itsmeichigo, it's not a tech debt and I don't think we should track it.

itsmeichigo: FYI I just added the extra comment in c753050 🙂

Ecarrion: I say it's tech debt because let's imagine the following scenario:

- We need to create a new navigation controller (Subclass of `WooNavigationController`) with some custom interactions.

- Dev will create the subclass and will implement those two methods to provide a custom interaction.

- It will not work, because the methods are not being forwarded from `WooNavigationController`

- Dev will add and forward those methods in `WooNavigationController` but that will make all other subclasses like `WooTabNavigationController` to break and lose it's native pop gesture interaction.

- Dev will realize that our current solution does not scale and will have to find alternatives.

To me, that sounds like tech debt 😅. And creating an issue explaining the situation will probably save a day or two for any future team member. 🪙🪙

itsmeichigo: Thanks @Ecarrion for the clarification 💯 This is really something to be aware of - I have created the issue here: #4524.

itsmeichigo:
Thank you @pmusolino and @Ecarrion for the comments! Can you give a 👍 so I can merge this?

Ecarrion (approved):

astralbodies:
Thanks for fixing this @itsmeichigo!! 🥳

PR #4505, opened 2021-06-30
Fix frozen animation on lists / scroll views when switching tabs

### LINKED ISSUE #2999: The pull to refresh stops to work properly when there is no internet connection
## Describe the bug
When there is no internet connection, the pull to refresh loader will be loading indefinitely, and if you change tab (for example from Order to Products), the animation will be stop.

## To Reproduce
1. Run the app.
2.  Enable the Network Link Conditioner on your mac, setting the profile 100% Loss.
3. Click on the Orders tab, and try to pull to refresh.
4. Tap Products tab, and try to pull to refresh.
5. Tap Reviews tab, and try to pull to refresh.
6. Now go to one of these tabs, and you can notice that the pull to refresh icon is still visible, and the animation stopped.

## Mobile Environment
Please include:
- Device: iPhone 11 Pro simulator
- iOS version: 14.0
- WooCommerce iOS version: -

itsmeichigo:
# Description
This PR fixes #2999 - currently when pull to refresh on Dashboard, Order List, Product List and Review List screens and then switch to a different tab, their refresh controls all get frozen.

I found [a similar issue](https://github.com/wordpress-mobile/WordPress-iOS/issues/1653) in the WordPress iOS - it was decided there that this is a minor framework issue and would not be fixed. 

# Solution
The suspected cause of this issue is that `isRemovedOnCompletion` is default to true for `CAAnimation` instances, which is why animations on both the refresh control and the placeholders are stopped after switching tabs.

This issue can be solved by checking on `viewWillAppear` to see if the control's `isRefreshing` is still true - then we can trigger `endRefreshing` and then begin the refresh again to fix the incomplete animation. This workaround will have to be applied to all the list residing on tabs: Product List, Review List, Order List, Dashboard.

Product List and Review List screens both display placeholders when pull-to-refresh is triggered, and these animations also get frozen when switching tabs. I added fix to re-apply the placeholders to make the animations work again. Order List is not showing placeholder when refreshing data, so I'm keeping it as-is (may need to consult Design team to unify this behavior).

# Discussion
The solution introduces duplicated code and is not clean enough IMHO, but it helps with getting around the issue since we cannot access the animation layer through public APIs. Please comment if you have suggestions for a reusable solution. 

# Screenshot
https://user-images.githubusercontent.com/5533851/123954391-8a34ae80-d9d2-11eb-8483-4ef7372b6ed6.mov

# Testing
- UI of all these screen should stay unchanged when pull-to-refresh is not triggered.
- When pull-to-refresh is triggered and switch tab fast and then get back, animation of the refresh control and placeholder (if any) should not be frozen.

Update release notes:


WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -186,6 +186,15 @@ final class ProductsViewController: UIViewController {
...
             displayRatingPrompt()
         }
+
+        // Fix any incomplete animation of the refresh control
+        // when switching tabs mid-animation
+        refreshControl.resetAnimation(in: tableView) { [unowned self] in
+            // ghost animation is also removed after switching tabs
+            // show make sure it's displayed again
+            self.removePlaceholderProducts()
+            self.displayPlaceholderProducts()
```
pmusolino, line 196: Probably I miss the context, but I'm not sure this is needed, since they are separate things. It should continue to work also without removing and re-adding the placeholder, maybe.

WooCommerce/Classes/ViewRelated/Reviews/ReviewsViewController.swift
```diff
@@ -153,6 +153,15 @@ final class ReviewsViewController: UIViewController {
...
             displayRatingPrompt()
         }
+
+        // Fix any incomplete animation of the refresh control
+        // when switching tabs mid-animation
+        refreshControl.resetAnimation(in: tableView) { [unowned self] in
+            // ghost animation is also removed after switching tabs
+            // show make sure it's displayed again
+            self.removePlaceholderReviews()
+            self.displayPlaceholderReviews()
```
pmusolino, line 163: same question here :)

pmusolino (approved):
Looks good to me, and thanks for fixing this @itsmeichigo 🥳 

I just left a question 👍

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
itsmeichigo: I agree that this has nothing to do with the main issue - but as I was testing with the app I saw that when I triggered pull-to-refresh on Product List and Review List screens, the table views were displayed with placeholders; then when I switched to other tabs and back again, the animations of the placeholders were gone. 

This issue is not obvious if you have more than the maximum number of rows for the placeholder, however in my case I saw this: 


The empty rows lost their animated overlay, and one single row for cached review showed up - which looks weird to me. I think the cause of this issue is similar to the refresh control - the `CAAnimation`s are removed upon the view controllers' disappearance. 

Let me know what you think about this - thanks!

itsmeichigo: (I should probably have included this in the description of the PR though) 😅

pmusolino:
> I agree that this has nothing to do with the main issue - but as I was testing with the app I saw that when I triggered pull-to-refresh on Product List and Review List screens, the table views were displayed with placeholders; then when I switched to other tabs and back again, the animations of the placeholders were gone.

Ohh got it :shipit:

PR #4508, opened 2021-06-30
[Mobile Payments] Fix modals in landscape

### LINKED ISSUE #4474: [Mobile Payments] Modals in landscape

ctarda:
Closes #4474

This is still not a perfect solution for the modals in Portrait mode, but it seems to provide a good enough solution for modals in landscape (on phones) and all orientations on tablets. This PR also prevents the illustration from changing sizes.

As mentioned here: https://github.com/woocommerce/woocommerce-ios/pull/4479#issuecomment-870571015 we want to hide the illustration and change the orientation of the action buttons in landscape mode.

I think we should target the frozen build for this, and then iterate again on portrait in a separate PR.

There are still two issues: 
* the width of the modal when a reader is found is wider than it should
* the modal in portrait mode shrinks a bit after the first modal (the one with the cancel button). I haven't been able to figure out why yet

| Before | After |
| ---- | ---- |
|  |  |
|  |  |
|  |  |

In portrait, notice how there is still a change in height:

But on the bright side, things look better on iPad as well:

## Changes
* Fixed width and height for illustration
* Updated the way the action buttons are laid out, wrapping the primary and secondary button in their own stack view
* When the orientation changes, change the axis of the stack view wrapping the buttons and hide the illustration, instead of changing the axis of the outermost stack view 

## How to test
* Try to connect to a reader and collect a payment, both in portrait and landscape orientations, both on iPad and iPhone

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.swift
@@ -11,6 +11,7 @@ final class CardPresentPaymentsModalViewController: UIViewController {
 
     @IBOutlet weak var containerView: UIView!
     @IBOutlet weak var mainStackView: UIStackView!
+    @IBOutlet weak var primaryActionButtonsStackView: UIStackView!
     @IBOutlet private weak var topTitleLabel: UILabel!
     @IBOutlet private weak var topSubtitleLabel: UILabel!
     @IBOutlet private weak var bottomTitleLabel: UILabel!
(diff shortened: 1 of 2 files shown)


koke (approved):
Even with the missing bits, this is so much better than the previous version, and at least now it's usable in landscape 🚢🚢🚢

Garance91540 (approved):
Looking good for the beta! The landscape modals ideally would need to have a fixed height and width as well, but I'll share more details on this during testing.

PR #4512, opened 2021-06-30
Variations Hide price-warning row borders

### LINKED ISSUE #4450: Variation: Remove price banner border.
Could we remove the top and bottom dividers of the Price alert banner?

screen-shot-2021-06-16-at-4 04 35-pm

reported by @Garance91540

Ecarrion:
fix #4450

# Why 

It was reported that the no-price banner row had some visible cell separators that made it look off. This PR removes those separators.

# How

Each cell has its own bottom separator, but the requirement is to remove both separators for the no-`price-warning` cell. 

To accomplish that, apart from removing the `no-price-warning` bottom separator, I'm removing the upper cell bottom separator, which at this moment can only be the `variations` cell for variable products or the `price` cell for a product variation.

The way to pass that information is done via the `ProductFormEditAction` enum all the way through the cell datasource.

# Screenshots

Before | After
--- | ---
bad |  good

# Testing steps

- Create a new variable product
- Create a new product variation (don't set the price)
- See that the `no-price-warning` cell does not have borders.

Update release notes:


ealeksandrov (approved):
Such multi-layer solution for a small UI change 🙂
LGTM :shipit:

PR #4498, opened 2021-06-29
[Mobile Payments] Add feature flag for card present known reader work

allendav:
This PR is the first in a series of PRs to add automatic reconnection to known card readers for in-person payments.

This PR simply adds the feature flag for the feature. No user facing changes nor unit test changes.

The sequence of planned PRs can be found here: https://github.com/woocommerce/woocommerce-ios/issues/3559#issuecomment-868872729 (The next PR will add actions to AppSettingsStore to add a known reader ID, forget a known reader ID and get all the known reader IDs.)

To test:
- This PR simply adds the flag. No changes to automated tests, nothing user facing to test.
- Best you can do is ensure linting and unit tests pass.

Related:

p91TBi-5t5-p2

Update release notes:
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4517, opened 2021-06-30
[Mobile Payments] Add KnownReadersProvider protocol and KnownReadersStoredList which implements it

allendav:

This PR defines a protocol for a known readers provider layer and an implementation that conforms to it, powered by the AppSettingsStore changes introduced in https://github.com/woocommerce/woocommerce-ios/pull/4499

Although there are no user facing changes there are unit test changes to cover the work.

Note that I combined both the protocol and the implementation in a single PR instead of two as originally planned

The sequence of planned PRs can be found here: https://github.com/woocommerce/woocommerce-ios/issues/3559#issuecomment-868872729

The next PR will pass this "known readers list" provider into the card reader settings view models (depending on the feature flag added in https://github.com/woocommerce/woocommerce-ios/pull/4498 )

To test:

Nothing user facing to test
Ensure linting passes
Ensure new unit tests pass

Related:

p91TBi-5t5-p2

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownReadersProvider.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownReadersProvider.swift
@@ -7,12 +7,9 @@ import Combine
 ///
 protocol CardReaderSettingsKnownReadersProvider {
 
-    /// We can't use a @Published wrapper for knownReaders as part of a protocol declaration, but we can
-    /// require its synthesized methods explicitly
+    /// Publishes changes to the known reader list.
     ///
-    var knownReaders: [String] { get }
-    var knownReadersPublished: Published<[String]> { get }
-    var knownReadersPublisher: Published<[String]>.Publisher { get }
+    var knownReaders: AnyPublisher<[String], Never> { get }
 
     /// Remember a reader. Asynchronous.
     ///
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownReadersProvider.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownReadersProvider.swift
@@ -1,0 +1,24 @@
+import Foundation
+import Combine
+
+/// Defines a protocol for wrappers around AppSettingsStore for "Known" Card Readers to confom to
+/// Allows us to support multiple implementations/dependency injection, and avoid introducing
+/// Combine into the Storage layer
+///
+protocol CardReaderSettingsKnownReadersProvider {
+
+    /// We can't use a @Published wrapper for knownReaders as part of a protocol declaration, but we can
+    /// require its synthesized methods explicitly
+    ///
+    var knownReaders: [String] { get }
+    var knownReadersPublished: Published<[String]> { get }
+    var knownReadersPublisher: Published<[String]>.Publisher { get }
+
+    /// Remember a reader. Asynchronous.
+    ///
+    func rememberCardReader(cardReaderID: String)
+
+    /// Forget a reader. Asynchronous.
+    ///
+    func forgetCardReader(cardReaderID: String)
+}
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownReadersProvider.swift
```diff
@@ -0,0 +1,24 @@
...
+import Combine
+
+/// Defines a protocol for wrappers around AppSettingsStore for "Known" Card Readers to confom to
+/// Allows us to support multiple implementations/dependency injection, and avoid introducing
+/// Combine into the Storage layer
+///
+protocol CardReaderSettingsKnownReadersProvider {
+
+    /// We can't use a @Published wrapper for knownReaders as part of a protocol declaration, but we can
+    /// require its synthesized methods explicitly
```
koke: This seems to be working but I don't think it's using the API as intended. The documentation for [Published](https://developer.apple.com/documentation/combine/published) says:

> The @Published attribute is class constrained. Use it with properties of classes, not with non-class types like structures

I understand that `@Published` is meant to be used on properties of an `ObservableObject`. The source for Combine isn't available, but looking at [OpenCombine](https://github.com/OpenCombine/OpenCombine/blob/master/Sources/OpenCombine/ObservableObject.swift#L52), an `ObservableObject` will use reflection to make sure every `@Published` property calls `objectWillChange` in the parent object.

If we don't need the `ObservableObject` features, I think a `CurrentValueSubject<[String], Never>` might do the same job with simpler code. We should probably expose it as a type-erased `AnyPublisher<[String], Never>`, since we don't want anything external sending values to the subject.

If we need outside access to both the publisher and the current value, then maybe we still need two properties.

I put up a quick diff to illustrate my suggestion:
https://gist.github.com/koke/80e760a5c76ac56080abbeb14437a711

allendav: Thanks @koke - I think you're right. This could be a risky approach. We do need the current value (for when the viewmodels initialize) and to be notified of changes - I think the type erased CurrentValueSubject you've proposed is exactly what we need. I'll update the PR. Thanks especially for the diff.

allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4517/commits/992361eaf8315e314396552dcbcc447a008679a3

Note: I also added a test to confirm that the initial subscription triggers an update as well - and that's the only situation where we need the current value apart from an update - so I think this should satisfy the needs of the remaining steps too.

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/CardReaderSettings/CardReaderSettingsKnownReadersStoredListTests.swift
```diff
@@ -0,0 +1,84 @@
...
+
+/// Mock constants
+///
+private struct TestConstants {
+    static let mockReaderID = "CHB204909005931"
+}
+
+final class CardReaderSettingsKnownReadersStoredListTests: XCTestCase {
+
+    func test_subscribing_publishes_initial_value() {
```
koke: I added another test in 249d703fe since I don't think this one fully covered the initial value case. In `CardReaderSettingsKnownReadersStoredList`, we initialize the `knownReadersSubject` with an empty array, so even if we forgot to `loadReaders` or it failed for some reason, that subject would always have an empty initial value anyway.

In 249d703fe, I added another case to test whether the initial value was correct when there were previously known reader, which was passing correctly 😌

koke (approved):

koke (commented):
I was looking at #4526 and I think I've changed my mind about this one...

The source of truth for the known readers is coming from `StoresManager`, but that isn't publishing updates. Instead, `CardReaderSettingsKnownReadersStoredList` is the class that's creating the publisher, and forcing a reload on every change. However, this only works if every subscriber to `knownReaders` is using the same instance.

This is what is happening in #4526, as there's only one instance being initialized by `CardReaderSettingsViewModelsOrderedList` and passed around to the other view models, but it's not a guarantee.

In the alternate case where each view model initializes its own `CardReaderSettingsKnownReadersStoredList`, they would only get updates for the remember/forget actions that they trigger.

In summary, I think `CardReaderSettingsKnownReadersStoredList` should only be used as a shared instance exposed somewhere in the `ServiceLocator` hierarchy. Or we should move the publishing of updates to a lower layer.

PR #4526, opened 2021-07-01
[Mobile Payments] Add Known Readers Provider to Settings VM Ordered List and pass into VMs

allendav:
Work in progress. I'll rebase this against develop after #4517 is merged to develop

This PR takes the known readers provider implementation introduced in #4517 and adds it to CardReaderSettingsViewModelsOrderedList, which then passes it, if the feature flag is set, into the VMs that need it (both existing VMs do)

CardReaderSettingsUnknownViewModel uses the KnownReadersProvider to tell it to remember a reader that has been explicitly disconnected from.

And CardReaderSettingsConnectedViewModel uses the KnownReadersProvider to tell it to forget a reader that has been explicitly disconnected from.

The next PR will add a new VM that will manage automatically connecting to a known reader during discovery.

The sequence of planned PRs can be found here: #3559 (comment)

To test:

- Ensure linting passes
- Ensure unit tests pass
- Ensure that after connecting to a reader, that the general-app-settings.plist file on the device includes lines similar to the following:

```
        <key>knownCardReaders</key>
        <array>
                <string>CHB204839002034</string>
        </array>
```

- Ensure that after disconnecting from the reader, that the general-app-settings.plist file on the device knownCardReaders lines look more like this:

```
        <key>knownCardReaders</key>
        <array/>
```

- Modify CardReaderSettingsViewModelsOrderedList to NOT initialize knownReadersProvider (i.e. thereby seeing the behavior when the feature flag is NOT present) and ensure you can connect and disconnect from a card reader without any problems.

- To view the general-app-settings.plist file, use Xcode Devices and Simulators View for your device, tap on Woo and then the gear below it and select Download Container. Within the container, dig down into AppData/Documents and there you'll find it.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -11,7 +11,7 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     private var connectedReaders = [CardReader]()
     private var knownReadersProvider: CardReaderSettingsKnownReadersProvider?
 
-    var connectedReaderSerialNumber: String?
+    var connectedReaderID: String?
     var connectedReaderBatteryLevel: String?
 
     init(didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?, knownReadersProvider: CardReaderSettingsKnownReadersProvider? = nil) {
@@ -39,12 +39,12 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
 
     private func updateProperties() {
         guard connectedReaders.count > 0 else {
-            connectedReaderSerialNumber = nil
+            connectedReaderID = nil
             connectedReaderBatteryLevel = nil
             return
         }
 
-        connectedReaderSerialNumber = connectedReaders[0].serial
+        connectedReaderID = connectedReaders[0].id
 
         guard let batteryLevel = connectedReaders[0].batteryLevel else {
             connectedReaderBatteryLevel = Localization.unknownBatteryStatus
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -58,11 +60,19 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     ///
     func disconnectReader() {
         ServiceLocator.analytics.track(.cardReaderDisconnectTapped)
-        let action = CardPresentPaymentAction.disconnect() { result in
+
+        let readerSerialNumberToDisconnect = connectedReaderSerialNumber
```
ctarda: I am not sure I understand this block. If I understand correctly, disconnect is always called, but forgetCardReader is only called, after disconnect is completed, if the serial number is not nil?

Is there a reason why we can't guard against the serial number of the connected reader and call disconnect only when the serial number is not nil? If not, could the call to forgetCardReader be encapsulated in a separate method? This new method would be the one that guards against a nil serial number.

The reason I ask this is because this line seems a bit out of place in the context of this codebase: 
`[readerSerialNumberToDisconnect, weak self]`

ctarda (approved):
It works well. I made an observation for your consideration though.

:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: Hmmm. My thoughts were to forget a card reader only after successfully disconnecting from it, to avoid the possibility of introducing an odd state - where we are connected to a reader, but it is not longer known/remembered.

> Is there a reason why we can't guard against the serial number of the connected reader and call disconnect only when the serial number is not nil?

Perhaps, but the view model would then need to know what's important in that CardReader that was passed to it. I think that detail should remain known only to the lower levels.

In fact, as I type that, I'm going to rename the "Serial number" variables to use simply "ID" like I do elsewhere. The fact that we are using the serial number as an ID is something I should be getting from the CardReader object. I think I'll add/use a helper on that object to get the ID.

> If not, could the call to forgetCardReader be encapsulated in a separate method? This new method would be the one that guards against a nil serial number.

Can do

allendav: OK. I opted for simply forgetting the reader whenever disconnect is called. In the very unlikely event disconnect fails, the next time the merchant connects to the reader we'll save the readerID anyways. The unusual `[readerSerialNumberToDisconnect, weak self]` is now no longer needed.

I also cleaned up the use of serial number to be simply "ID"

Changes made in https://github.com/woocommerce/woocommerce-ios/pull/4526/commits/66195bc396839ad43d340cf9dfa2b0e6546f9326

PR #4527, opened 2021-07-02
[Mobile Payments] Prevents crash when reconnecting to a reader

### LINKED ISSUE #4521: [Mobile Payments] Crash when connecting readers on landscape
**To Reproduce**

1. Launch the app in landscape orientation on iPhone (I can't reproduce on iPad)
2. Go to Settings > Manage Card Readers
3. Tap on Connect Card Reader
4. Turn on your reader, and connect to it when prompted
5. Tap on Disconnect
6. Go back to the settings screen and enter Manage Card Readers again
7. Tap on Connect Card Reader and connect when prompted
8. 💥 

```
2021-07-01 11:06:36.950064+0200 WooCommerce[25473:8541413] *** Assertion failure in -[UITableView _createPreparedCellForGlobalRow:withIndexPath:willDisplay:], UITableView.m:15065
2021-07-01 11:06:36.953494+0200 WooCommerce[25473:8541413] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'UITableView dataSource is not set'
*** First throw call stack:
(0x19522e754 0x1a9cf57a8 0x1951303f8 0x196504714 0x197e4ae58 0x197e16890 0x197e16d44 0x197e34b8c 0x19815da0c 0x1985e4670 0x1985e4b54 0x1985f978c 0x19853b5e4 0x1985677f4 0x198568b20 0x1951a75e0 0x1951a1704 0x1951a1cb0 0x1951a1360 0x1ac7df734 0x197c1c584 0x197c21df4 0x10451e3f8 0x194e5dcf8)
libc++abi: terminating with uncaught exception of type NSException
```

This only seems to happen on iPhone in landscape mode.

**Mobile Environment**
Please include:
- Device: iPhone XS
- iOS version: 14.6
- WooCommerce iOS version: Reproduced on develop and 7.0 from TestFlight

koke:
Fixes #4521 

This was a hard one to track, and I don't understand all the moving pieces, but the essence of it is that, when we use `htmlToAttributedString` to create the attributed string for the "Learn more" footer, this will internally load webkit. While it's loading, the NSAttributedString internals will return control of the execution to the run loop, which might start running async blocks of code.

In this particular case, this can lead to situations where the table view is recreating its cells, and a callback to one of the StripeTerminal methods gets executed and attempts to update the UI. I didn't find the specific problem, but we would likely have several instances where there was a UI callback while the UI was in an inconsistent state.

This commit doesn't solve the issue perfectly, as it would still happen the first time the cell is rendered, but at least any subsequent calls will be cached. At least this fixes the crash, although we might want to reconsider this approach in the future, and maybe use something simpler like markdown for links in translations, or process them at build time.

## To test

1. Launch the app in landscape orientation on iPhone (I can't reproduce on iPad)
2. Go to Settings > Manage Card Readers
3. Tap on Connect Card Reader
4. Turn on your reader, and connect to it when prompted
5. Tap on Disconnect
6. Go back to the settings screen and enter Manage Card Readers again
7. Tap on Connect Card Reader and connect when prompted
8. The app shouldn't crash now

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
@@ -405,7 +405,7 @@ private extension CardReaderSettingsUnknownViewController {
             comment: "Settings > Manage Card Reader > Connect > A button to begin a search for a reader"
         )
 
-        static var learnMore: NSAttributedString = {
+        static let learnMore: NSAttributedString = {
             let learnMoreText = NSLocalizedString(
                 "<a href=\"https://woocommerce.com/payments\">Learn more</a> about accepting payments with your mobile device and ordering card readers",
                 comment: "A label prompting users to learn more about card readers with an embedded hyperlink"
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
@@ -405,7 +405,7 @@ private extension CardReaderSettingsUnknownViewController {
             comment: "Settings > Manage Card Reader > Connect > A button to begin a search for a reader"
         )
 
-        static var learnMore: NSAttributedString {
+        static var learnMore: NSAttributedString = {
             let learnMoreText = NSLocalizedString(
                 "<a href=\"https://woocommerce.com/payments\">Learn more</a> about accepting payments with your mobile device and ordering card readers",
                 comment: "A label prompting users to learn more about card readers with an embedded hyperlink"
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
```diff
@@ -405,7 +405,7 @@ private extension CardReaderSettingsUnknownViewController {
             comment: "Settings > Manage Card Reader > Connect > A button to begin a search for a reader"
         )
 
-        static var learnMore: NSAttributedString {
+        static var learnMore: NSAttributedString = {
```
Ecarrion: I think this can be an `static let` right?

Ecarrion (approved):
What a weird bug, I followed the slack thread and I'm glad we have a solution although it does seem concerning for the future. Do you think adding a comment here explaining some of this discovery would help?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownViewController.swift
koke: Yes, that's better 😁

Issue #4136, opened 2021-05-07
Bring back Order fulfillment screens


anitaa1990:
There was some feedback from users that wanted to have a confirmation screen after clicking on `Mark order complete` button in Order detail. We had initially removed that screen from both platforms, since it only provided duplicate information already available in the Order Detail.

Ref: p91TBi-53v

PR #4528, opened 2021-07-02
Add Review Order screen (previously known as Fulfill Order)

itsmeichigo:
Part of #4136 

# Description
This PR aims to add new Review Order screen that should appears after selecting "Mark Order Complete" button on Order Detail screen. The objective is to let user take a final look at the order details and decide if they're ready to notify users that order is fulfilled.

# Solution
- The old Fulfill Order screen was removed in March, so this task is to bring the similar screen back. Since the product list is a bit different and the old approach was plain MVC, I chose to rewrite the feature with MVVM for better testability and readability.
- To keep the PR small, this PR only includes the setup of new Review Order screen and product section. To avoid blocking other teammate, I'm putting the changes behind a feature flag named `reviewOrder`. This flag will be removed and the release notes will be updated when the feature is complete.

# Screenshot

# Testing

Update release notes:


itsmeichigo:
Closing to create smaller PRs 🚫

PR #4529, opened 2021-07-02
Fixed the package weight in Shipping Label Creation form

### LINKED ISSUE #4514: Shipping Labels: Suggested rates use the wrong package weight
**Describe the bug**

I noticed that the rates we're offering in the iOS app don't match the same rates we offer in the Android app or web client.

It looks like we're sending the wrong weight for the selected package to the endpoint for getting the carriers and rates. Instead of using the package weight that's entered in "Package Details," we're setting the weight as the package width [here](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Shipping%20Labels/Create%20Shipping%20Label%20Form/ShippingLabelFormViewModel.swift#L51) and [here](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Shipping%20Labels/Create%20Shipping%20Label%20Form/ShippingLabelFormViewModel.swift#L61).

**To Reproduce**

5. Fill out all sections of the Create Shipping Label form, until you get to "Shipping Carrier and Rates".
6. Open "Shipping Carrier and Rates" and notice what rates you are offered there. (You can follow the same steps in the Android app or web client and notice that different rates are offered there.)
7. Go back to the "Package Details" section and change the package weight. Check the "Shipping Carrier and Rates" section again and notice that the rates haven't changed.

**Expected Behavior**

We should make sure that we're sending the correct weight for the selected package, and that the offered carriers and rates match other client behavior.

pmusolino:
Fixes #4514 

## Description
@rachelmcr noticed that the rates we're offering in the iOS app don't match the same rates we offer in the Android app or web client.
The reason was that we were sending the wrong weight for the selected package to the endpoint for getting the carriers and rates. Instead of using the package weight that's entered in "Package Details," we're setting the weight as the package width.

## Testing
5. Fill out all sections of the Create Shipping Label form, until you get to "Shipping Carrier and Rates".
6. Open "Shipping Carrier and Rates" and notice what rates you are offered there. (You can follow the same steps in the Android app or web client and notice that different rates are offered there.)
7. Go back to the "Package Details" section and change the package weight. Check the "Shipping Carrier and Rates" section again and notice that the rates have changed.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -43,12 +43,14 @@ final class ShippingLabelFormViewModel {
             return nil
         }
 
+        let weight = Double(truncating: (Decimal(string: totalPackageWeight ?? "0") as  NSNumber?) ?? 0)
+
         if let customPackage = packagesResponse.customPackages.first(where: { $0.title == selectedPackageID }) {
             return ShippingLabelPackageSelected(boxID: customPackage.title,
                                                 length: customPackage.getLength(),
                                                 width: customPackage.getWidth(),
                                                 height: customPackage.getHeight(),
-                                                weight: customPackage.getWidth(),
+                                                weight: weight,
                                                 isLetter: customPackage.isLetter)
         }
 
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -43,12 +43,14 @@ final class ShippingLabelFormViewModel {
             return nil
         }
 
+        let weight = Double(truncating: (Decimal(string: totalPackageWeight ?? "0") as  NSNumber?) ?? 0)
```
Ecarrion: Not sure if I understanding this completely, but I think you can get way with a simpler API like:

```suggestion
        let weight = Double(totalPackageWeight ?? "0") ?? .zero
```

Ecarrion (approved):
Just left a minor suggestion!

pmusolino:
Thank you @Ecarrion for the review 👍

PR #4532, opened 2021-07-02
Variations: Remove unnecessary no-price warning banner

### LINKED ISSUE #4451: Variations: Remove redundant "No Price" banner
Could we remove the price alert banner on the Variations list screen (sorry I know this was in the original designs)? I think it’s probably enough that we show the banner on the parent product page + the “No price set” button. With the banner, the Variation list screen becomes a bit too crowded

screen-shot-2021-06-16-at-4 07 44-pm

reported by @Garance91540

Ecarrion:
fixes #4451

# Why

During testing, it was reported that the "no price warning" banner in the variation list screen was redundant because we also say that a variation does not have a price within each cell. 

This PR removes that banner.

# How 

Just 🔪 it!

# Screenshots

Before | After 
--- | ---
before | after

# Testing Steps

- Create a new variable product

- Add a variation

- Navigate to the variations list screen and confirm that the warning banner is not visible.

Update release notes:


pmusolino (approved):
LGTM :shipit:

PR #4533, opened 2021-07-04
Big titles solution

fernandofgfer:
#4533 
**Why**
In this PR I'm fixing the issue of the big titles on the Home Screen, specifically the subtitle problem.
The aim of this feature is to bring the possibility to show big store names, adding a subtitle with a lower font.

**How**
To solve this, I've added a new label that is hidden or shown depends on the scroll

**Final result**
Light Mode
Dark Mode

**Tests steps**
Go to the home screen and check how the subtitle disappear or appear with scroll interaction

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -15,12 +15,42 @@ final class DashboardViewController: UIViewController {
     private let dashboardUIFactory: DashboardUIFactory
     private var dashboardUI: DashboardUI?
 
+    private var isDefaultSiteNameAvailable: Bool {
```
Ecarrion: The way this variable is used, in `titleName` and in `configureSubtitle` I think the name is a bit misleading. 

What do you think about `shouldShowStoreNameAsSubtitle`?

Also, adding some comments explaining how that is decided would be good for future mantainers! :-)

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -15,12 +15,42 @@ final class DashboardViewController: UIViewController {
...
+        return ServiceLocator.stores.sessionManager.defaultSite?.name != nil && ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles)
+    }
+
+    private var titleName: String {
+        if !isDefaultSiteNameAvailable {
+            return ServiceLocator.stores.sessionManager.defaultSite?.name ?? Localization.title
+        }
+        return Localization.title
```
Ecarrion: I think this is a perfect use case for the `guard` pattern!
```suggestion
        guatd shouldShowStoreNameAsSubtitle {
            return Localization.title
        }
        return ServiceLocator.stores.sessionManager.defaultSite?.name ?? ""
```

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -15,12 +15,42 @@ final class DashboardViewController: UIViewController {
...
         return UIView(frame: .zero)
     }()
 
+    private lazy var storeNameLabel: UILabel = {
+        let label = UILabel()
+        label.numberOfLines = 1
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.font = UIFont.subheadline
+        label.backgroundColor = .listForeground
+        label.textColor = UIColor.text
```
Ecarrion: To keep styles aligned, we try to use some reusable style methods that reside in `UILabel+Helpers`. EG `applySubheadlineStyle()`

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -78,7 +108,31 @@ private extension DashboardViewController {
...
 
     func configureTitle() {
-        navigationItem.title = ServiceLocator.stores.sessionManager.defaultSite?.name ?? Localization.title
+        navigationItem.title = titleName
+    }
+
+    func configureSubtitle() {
+        if isDefaultSiteNameAvailable {
+            containerView.backgroundColor = .listForeground
+            storeNameLabel.text = ServiceLocator.stores.sessionManager.defaultSite?.name ?? Localization.title
```
Ecarrion: I've seen this pattern repeats quite a lot now! I think it would be good if we can extract it to its own variable.

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -15,12 +15,42 @@ final class DashboardViewController: UIViewController {
...
+
     // MARK: Subviews
 
     private lazy var containerView: UIView = {
         return UIView(frame: .zero)
     }()
 
+    private lazy var storeNameLabel: UILabel = {
+        let label = UILabel()
+        label.numberOfLines = 1
```
Ecarrion: I think this line is not needed as it is the default!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -15,12 +15,42 @@ final class DashboardViewController: UIViewController {
...
+        label.backgroundColor = .listForeground
+        label.textColor = UIColor.text
+        return label
+    }()
+
+    private lazy var stackView: UIStackView = {
+        let view = UIStackView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.backgroundColor = .listForeground
+        view.layoutMargins = UIEdgeInsets(top: 0, left: navigationController?.navigationBar.layoutMargins.left ?? 0, bottom: 0, right: 0)
```
Ecarrion: I think we need to use `directionalLayoutMargins` to properly support RTL lenguages.
```suggestion
        view.layoutMargins = UIEdgeInsets(top: 0, left: navigationController?.navigationBar.directionalLayoutMargins.leading ?? 0, bottom: 0, right: 0)
```

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -78,7 +108,31 @@ private extension DashboardViewController {
     }
 
     func configureTitle() {
-        navigationItem.title = ServiceLocator.stores.sessionManager.defaultSite?.name ?? Localization.title
+        navigationItem.title = titleName
+    }
+
+    func configureSubtitle() {
+        if isDefaultSiteNameAvailable {
```
Ecarrion: I think this would read better if we use the early return pattern.
```swift
guard let !isDefaultSiteNameAvailable {
   return
}

// rest of the code
```

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -126,6 +180,25 @@ private extension DashboardViewController {
 extension DashboardViewController: DashboardUIScrollDelegate {
     func dashboardUIScrollViewDidScroll(_ scrollView: UIScrollView) {
         hiddenScrollView.updateFromScrollViewDidScrollEventForLargeTitleWorkaround(scrollView)
+        showOrHideSubtitle(offset: scrollView.contentOffset.y)
+    }
+
+    private func showOrHideSubtitle(offset: CGFloat) {
+        if ServiceLocator.stores.sessionManager.defaultSite?.name != nil {
```
Ecarrion: This is also a good fit for the early return pattern!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -126,6 +180,25 @@ private extension DashboardViewController {
...
+            if offset > stackView.frame.height {
+                storeNameLabel.isHidden = true
+            }
```
Ecarrion: I think this can be
```swift
storeNameLabel.isHidden = offset > stackView.frame.height
```
right?

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -126,6 +180,25 @@ private extension DashboardViewController {
...
+            if offset < -stackView.frame.height {
+                UIView.transition(with: storeNameLabel, duration: 0.2,
```
Ecarrion: It would be nice to have this transition on a constant enum, see `LoadingView.Layout` to allow us to easily find it and modify it in the future.

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -126,6 +180,25 @@ private extension DashboardViewController {
...
+                                  options: .showHideTransitionViews,
+                                  animations: { [weak self] in
+                                    guard let self = self else {
+                                        return
+                                    }
```
Ecarrion: Tip: According to our style guide, this can be a one-liner! :-)

Ecarrion (changes requested):
Left you some comments @fernandofgfer, let me know what you think!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Good improvement, more readable!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Totally agree!!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Thanks for the tip!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: I've used the new variable `shouldShowStoreNameAsSubtitle `

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Yep xD

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Changed it to be more readable!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Agree!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Nice improvement!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Yep. Will be better if someone has to modify it in the future

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Nice to know! Thanks for the tip!

Ecarrion:
Also, something I forgot to mention earlier, could you please make sure that the subtitle responds appropriately to accessibility font sizes?

This is how it looks right now.
Screen Shot 2021-07-07 at 10 30 12 AM

fernandofgfer:
> Also, something I forgot to mention earlier, could you please make sure that the subtitle responds appropriately to accessibility font sizes?
> 
> This is how it looks right now.
> Screen Shot 2021-07-07 at 10 30 12 AM

Hi! Thanks for remembering this accessibility problem, I will add it in the next test steps to have in mind. I think this problem is fixed in one of my last commits'UI improvements', using the property `adjustsFontForContentSizeCategory = true`
Captura de pantalla 2021-07-07 a las 22 38 21

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -8,19 +8,53 @@ import Yosemite
 //
 final class DashboardViewController: UIViewController {
 
+    // MARK: Constants
+    enum Constants {
+        static let animationDuration = 0.2
+    }
```
Ecarrion: I think we already have a constants extension at the end of this file, could we move this down, please?

https://github.com/woocommerce/woocommerce-ios/blob/858a992cc766d41a0ab50a9dd9a872521b970ba6/WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift#L214-L221

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
Ecarrion: I think you missed this comment! :-)

Ecarrion (changes requested):
This is looking good! Just a couple of minor changes and we should be good to go!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Sure, moved!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
fernandofgfer: Yes, I didn't realize about this comment. It's fixed!

Ecarrion (approved):
Looks good now! 🎉

PR #4534, opened 2021-07-05
[Order Fulfilment] Add Review Order screen (previously known as Fulfill Order)

itsmeichigo:
Part 1 of #4136. [Part 2](https://github.com/woocommerce/woocommerce-ios/pull/4535), [Part 3](https://github.com/woocommerce/woocommerce-ios/pull/4539), Part 4 to come soon.

# Description
This PR aims to add new Review Order screen that should appears after selecting "Mark Order Complete" button on Order Detail screen. The objective is to let user take a final look at the order details and decide if they're ready to notify users that order is fulfilled.

# Solution
- The old Fulfill Order screen was removed in March, so this task is to bring the similar screen back. Since the product list is a bit different and the old approach was plain MVC, I chose to rewrite the feature with MVVM for better testability and readability.
- To keep the PR small, this PR only includes the setup of new Review Order screen. To avoid blocking other teammate, I'm putting the changes behind a feature flag named `reviewOrder`. This flag will be removed and the release notes will be updated when the feature is complete.
- ReviewOrderViewModel accepts order and product list from Order details. It also holds reference to a stores manager and storage manager to later handle adding tracking and fulfilling order.
- ReviewOrderViewController accepts injection of its view model and has a table view to display details of order for review.

# Screenshot

# Testing
- Navigate to Orders tab
- Select Mark Order Complete
- You should then be navigated to new screen named Review Order with an empty list for now.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
+++ b/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
@@ -17,6 +17,8 @@ struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .cardPresentKnownReader:
             return buildConfig == .localDeveloper || buildConfig == .alpha
+        case .reviewOrder:
+            return false
         default:
             return true
         }
--- a/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
+++ b/WooCommerce/Classes/System/DefaultFeatureFlagService.swift
@@ -18,7 +18,7 @@ struct DefaultFeatureFlagService: FeatureFlagService {
         case .cardPresentKnownReader:
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .reviewOrder:
-            return false
+            return buildConfig == .localDeveloper
         default:
             return true
         }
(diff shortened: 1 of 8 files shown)


WooCommerce/Classes/System/DefaultFeatureFlagService.swift
```diff
@@ -17,6 +17,8 @@ struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .cardPresentKnownReader:
             return buildConfig == .localDeveloper || buildConfig == .alpha
+        case .reviewOrder:
+            return buildConfig == .localDeveloper
```
allendav, line 21: Why not .alpha too?

allendav:
In the future, please include in the description ( or #4136 ) descriptions of the other PRs to come. It is harder to evaluate this PR without knowing how it fits into the overall feature.

allendav (approved):
Tests fine. We'll want unit tests for the new viewmodel once it does something interesting. :shipit:

WooCommerce/Classes/System/DefaultFeatureFlagService.swift
itsmeichigo: Yes normally I'd add .alpha but in this case I plan to remove it as soon as I get the feature done - so I think just keeping it local is enough.

itsmeichigo:
Thank you @allendav for reviewing!

PR #4535, opened 2021-07-05
[Order Fulfillment][Part 2] Update Review Order screen with Products section

itsmeichigo:
Part 2 of #4136
For more context: [part 1](https://github.com/woocommerce/woocommerce-ios/pull/4534), [part 3](https://github.com/woocommerce/woocommerce-ios/pull/4539), [part 4](https://github.com/woocommerce/woocommerce-ios/pull/4549), [part 5](https://github.com/woocommerce/woocommerce-ios/pull/4567).

# Description
This PR follows up #4534 (please makes sure to review that PR first) and updates Review Order screen with Product section.

# Solution
- ReviewOrderViewModel was updated with new subtypes `Section` and `Row` to define sections and rows of the review order table view.
- The view model acts as data source for the review order table view, so it was updated to config displayed data for the Product section including header title and cell view model.
- Product cell view models require info about add-ons, so the view model was updated to receive `showAddOns` config from Order Details screen. Add-ons are also queried to check if they're available on the products.
- ReviewOrderViewController was updated to config its table view's datasource and delegate. Action to show add-ons was also implemented.

# Screenshot

# Testing
- Navigate to Orders tab
- Select Mark Order Complete
- You should then be navigated to Review Order screen with only one section named Products.
- If `showAddOns` config is enabled, the displayed product cells should show option to see add-ons appropriately (like in the gif above).

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Review Order/ReviewOrderViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Review Order/ReviewOrderViewModel.swift
@@ -38,6 +38,77 @@ extension ReviewOrderViewModel {
     }
 }
 
+// MARK: - Section and row types for Review Order table view
+//
+extension ReviewOrderViewModel {
+    struct Section {
+        /// Section types for Review Order screen
+        ///
+        enum Category: CaseIterable {
+            case products
+            case customerInformation
+            case tracking
+        }
+
+        /// Category of the section
+        ///
+        let category: Category
+
+        /// Rows in the section
+        ///
+        let rows: [Row]
+
+        /// UITableViewHeaderFooterView type for each section
+        ///
+        var headerType: UITableViewHeaderFooterView.Type {
+            switch category {
+            case .products:
+                return PrimarySectionHeaderView.self
+            case .customerInformation, .tracking:
+                return TwoColumnSectionHeaderView.self
+            }
+        }
+
+        init(category: Category, rows: [Row]) {
+            self.category = category
+            self.rows = rows
+        }
+    }
+
+    /// Row types for Review Order screen
+    ///
+    enum Row {
+        case orderItem(item: OrderItem)
+        case customerNote(text: String)
+        case shippingAddress(address: Address)
+        case shippingMethod(method: String)
+        case billingDetail
+        case tracking
+        case trackingAdd
+
+        /// UITableViewCell type for each row type
+        ///
+        var cellType: UITableViewCell.Type {
+            switch self {
+            case .orderItem:
+                return ProductDetailsTableViewCell.self
+            case .customerNote:
+                return CustomerNoteTableViewCell.self
+            case .shippingAddress:
+                return CustomerInfoTableViewCell.self
+            case .shippingMethod:
+                return CustomerNoteTableViewCell.self
+            case .billingDetail:
+                return WooBasicTableViewCell.self
+            case .tracking:
+                return OrderTrackingTableViewCell.self
+            case .trackingAdd:
+                return LeftImageTableViewCell.self
+            }
+        }
+    }
+}
+
 // MARK: - Localization
 //
 private extension ReviewOrderViewModel {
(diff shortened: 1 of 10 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Review Order/ReviewOrderViewModel.swift
```diff
@@ -3,9 +3,27 @@ import Yosemite
 import protocol Storage.StorageManagerType
 
 final class ReviewOrderViewModel {
+    /// Quick access to header types for table view registration
+    ///
+    let allHeaderTypes: [UITableViewHeaderFooterView.Type] = {
+        [PrimarySectionHeaderView.self,
+         TwoColumnSectionHeaderView.self]
+    }()
```
rachelmcr: nit: To avoid having to go back and forth between the VC and VM to see how `allHeaderTypes` and `allCellTypes` are defined and used, how about calling a method that defines and registers them together? (For example, [here's how that's done in OrderDetailsViewModel](https://github.com/woocommerce/woocommerce-ios/blob/c6a1a399eea1cae2035264d46037d6db5e361f2a/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsViewModel.swift#L197-L208).)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Review Order/ReviewOrderViewModel.swift
```diff
@@ -33,15 +71,117 @@ final class ReviewOrderViewModel {
...
-// MARK: - Localization
+// MARK: - Sections configuration
 //
 private extension ReviewOrderViewModel {
-    enum Localization {
-        static let screenTitle = NSLocalizedString("Review Order", comment: "Title of Review Order screen")
+    /// Product section setup
+    ///
+    var productSection: Section {
+        let rows = order.items.map { Row.orderItem(item: $0) }
```
rachelmcr: I noticed that if I refund a product in an order, that refunded product still shows up on the Review Order screen. I think that happens because `order.items` includes all items in the original order, including refunded items. We'll have to filter out the refunded items so they aren't included in the product list.

rachelmcr (commented):
Looks good! I tested and confirmed it works as expected for orders with products that do or don't have add-ons, and it respects the experimental features beta switch for add-ons.

However, I noticed that the Review Order screen still included refunded products on the order. I tested with an order that had 2 products, and refunded one product on the order. The Review Order screen still showed both products:

Order Detail after refund | Review Order after refund
-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-07-07 at 17 47 45|Simulator Screen Shot - iPhone 11 Pro - 2021-07-07 at 17 47 50

itsmeichigo:
Thank you @rachelmcr for reviewing! I have updated to use AggregateOrderItem to take into account refunds; and added a unit test for that case. Which causes the file changes to 682 lines 🤦‍♀️ Let me know what you think about the changes!

WooCommerce/WooCommerceTests/ViewRelated/Orders/Review Order/ReviewOrderViewModelTests.swift
```diff
@@ -33,46 +34,97 @@ class ReviewOrderViewModelTests: XCTestCase {
...
+        let productSection = viewModel.sections.first(where: { $0.category == .products })
+        XCTAssertNotNil(productSection)
+        let productRow = productSection?.rows.first(where: {
+            if case .orderItem(let item) = $0,
+               item.productID == productID {
+                return true
+            }
+            return false
+        })
+        XCTAssertNotNil(productRow)
```
rachelmcr: Thank you for adding this test! 🎉 

I might be mistaken, but I think this only checks that the non-refunded item is in the Products section, right? It doesn't check to make sure the refunded item **isn't** there.

What do you think of replacing this check (or adding another check) like this:

```
        let refundedProductRow = productSection?.rows.first(where: {
            if case .orderItem(let item) = $0,
               item.productID == productID2 {
                return true
            }
            return false
        })
        XCTAssertNil(refundedProductRow)
```

rachelmcr (approved):
These changes look great, and test well!

I just have a small comment about the assertions in the new unit test, but otherwise this looks good to go. 👍

WooCommerce/WooCommerceTests/ViewRelated/Orders/Review Order/ReviewOrderViewModelTests.swift
itsmeichigo: That makes better sense! Let me update that 😄

PR #4536, opened 2021-07-05
Shipping Labels: fixed navigation title showed in tab bar in Package Details and Package List screens

### LINKED ISSUE #3995: Shipping Labels: fix navigation title showed in TabBar in Package List

pmusolino:
Fixes #3995 

## Description
In this PR, I fixed the issue that we have in the Package Details and Package List screens, implemented both in SwiftUI, where the navigation title was showed also in the tab bar as UITabBarButtonLabel. 
We didn't found any working solution, and we spent a lot of time looking for a solution that wasn't a workaround. I also contacted Apple, and I opened a radar about this issue that seems a bug on the Apple side, previously solved. Unfortunately, there was a regression, so the issue is still present, and there are some users that complain about it. You can read more on the issue I opened originally: https://github.com/woocommerce/woocommerce-ios/issues/3995

The best solution I have found, other than a real alternative fix, is to present the Package List screen as a modal screen. This doesn't compromise the functionality of the screen, it works fine, and the problem does not arise because it has nothing to do with the current navigation flow.
The majority of the changes are because of the indentation since I needed to embed the content of `PackageList` inside a `NavigationView`.

## Testing
5. Fill out all sections of the Create Shipping Label form, until you get to "Package Details".
6. Open "Package Details" and select a package.
7. You should not see a "Label" of the title of the view in lower-left corner.

## Screenshots
Simulator Screen Shot - iPhone 12 Pro - 2021-07-05 at 10 39 58 | Simulator Screen Shot - iPhone 12 Pro - 2021-07-05 at 10 25 03

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/ShippingLabelPackageList.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/ShippingLabelPackageList.swift
@@ -6,48 +6,53 @@ struct ShippingLabelPackageList: View {
     @Environment(\.presentationMode) var presentation
 
     var body: some View {
-        ScrollView {
-            LazyVStack(spacing: 0) {
+        NavigationView {
+            ScrollView {
+                LazyVStack(spacing: 0) {
 
-                /// Custom Packages
-                ///
-                if viewModel.showCustomPackagesHeader {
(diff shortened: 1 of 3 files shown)


ctarda (approved):
This is how it looks for me:



:shipit:

PR #4542, opened 2021-07-05
[Mobile Payments] Fix crash when attempting to collect payment while retrying (7.0 Backport)

### LINKED ISSUE #4537: [Mobile Payments] Crash when attempting to collect payment while retrying

koke:
Backport of #4540 to 7.0. I tried applying some of the simpler commits only, but while that prevented the crash, it still left the app in an inconsistent state where it wasn't possible to capture payments anymore.

Fixes #4537 

There were a few things that went wrong in this crash, and I've tried to address all of them to be extra safe. Because the first modal was dismissed when tapping on retry, we were able to tap on collect payment again. Then we got a modal alert immediately, warning us that there was an operation in progress (the retry). When the VC tried to present the `readerIsReady` alert, that modal couldn't be presented because there was another modal being presented. When the reader sent a message for the user to tap or insert their card, it tried to replace the alert's view model and `populateContent`, but because the modal was never presented, the view hadn't loaded yet, and it's `IBOutlets` were `nil`.

So there are three separate fixes:

- With 2ce42d3, we don't dismiss the modal when retry is tapped, as it should present another one right away.
- With 241fd85, `CardPresentPaymentsModalViewController` will only try to call `populateContent` if the view has been loaded. It will always call `populateContent` from `viewDidLoad` anyway, so it's harmless to skip it.
- 01d9188 is a bit more complicated, but it will let go of the assumption that the modal is always presented from the `CardPresentModalReaderIsReady` model, and will handle presentation automatically whenever a new view model is set.

One side effect is that now a modal can be replaced with a different one. For instance, when a payment fails, if you quickly double tap the retry button, you'll get a brief alert saying that the system is busy, then it will show the "reader is ready" alert. I think this is probably OK, but I'd love a second opinion on whether that could have unintended consequences.

## To test

I tested the steps in #4537, but I also encourage trying many possible states of the payment flow (canceling from modals, double tapping things,...)

1. Go to an order with a total amount ending in `.75` (or any other amount that will cause payment to fail)
2. Tap on Collect payment.
3. Connect to a reader and tap the card.
4. The app will show an error saying it couldn't collect payment.
5. Tap retry, the modal will dismiss.
6. Quickly tap on Collect Payment on the order screen before a new modal pops up.
7. The app should not crash

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalError.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalError.swift
@@ -42,9 +42,7 @@ final class CardPresentModalError: CardPresentPaymentsModalViewModel {
     }
 
     func didTapPrimaryButton(in viewController: UIViewController?) {
-        viewController?.dismiss(animated: true, completion: {[weak self] in
-            self?.primaryAction()
-        })
+        primaryAction()
     }
 
     func didTapSecondaryButton(in viewController: UIViewController?) {
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsPaymentAlerts.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsPaymentAlerts.swift
@@ -6,70 +6,78 @@ import WordPressUI
 /// presented to provide user-facing feedback about the progress
 /// of the payment collection process
 final class OrderDetailsPaymentAlerts {
-    private var modalController: CardPresentPaymentsModalViewController?
+    private weak var presentingController: UIViewController?
+
+    // Storing this as a weak variable means that iOS should automatically set this to nil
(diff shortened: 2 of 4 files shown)


ctarda (approved):
Checked the diff and it's what's already been reviewed in #4540 

I run another round of tests, and failed to make it crash.

:shipit:

PR #4539, opened 2021-07-05
[Order Fulfillment][Part 3] Add customer section to Review Order screen

itsmeichigo:
Part 3 of #4136 
For more context there's [part 1](https://github.com/woocommerce/woocommerce-ios/pull/4534), [part 2](https://github.com/woocommerce/woocommerce-ios/pull/4535), [part 4](https://github.com/woocommerce/woocommerce-ios/pull/4549), [part 5](https://github.com/woocommerce/woocommerce-ios/pull/4567).

# Description
This PR follows up #4535 to add new section Customer Information to the Review Order screen.

# Solution
- ReviewOrderViewModel was updated to config rows for Customer section. This section should contains the following rows:
  - Customer note if order has additional note
  - Shipping address if order has at least one non-virtual product
  - Shipping method if order has shipping info
  - View billing info
- ReviewOrderViewController was updated to config styling and content for cells based on row information.

# Screenshots
| Customer section with note only | Customer section with Shipping Details |
|----|----|
|||

# Testing
- Navigate to Orders tab
- Select an order with In Progress status
- Select Mark Order Complete
- You should then be navigated to Review Order screen with two sections Product and Customer Information.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4549, opened 2021-07-06
[Order Fulfillment][Part 4] Add Tracking Section to Review Order screen

itsmeichigo:
Part 4 of #4136. 
For more context, please take a look at [Part 1](https://github.com/woocommerce/woocommerce-ios/pull/4534), [Part 2](https://github.com/woocommerce/woocommerce-ios/pull/4535), [Part 3](https://github.com/woocommerce/woocommerce-ios/pull/4539), [Part 5](https://github.com/woocommerce/woocommerce-ios/pull/4567).

# Description
This PR follows up #4539 to add another section for Shipment Tracking on Review Order screen. There are nearly 600 lines of code changes and I blame my generated `Copiable` and unit tests for that.

# Solution
- The view model was updated to config tracking section for the review order table view with conditions:
  - There should be no non-refunded shipping labels, otherwise the whole Tracking section is hidden.
  - Shipment & Tracking section should be installed to the store, otherwise Add Tracking row is hidden.
  - There should be at least one shipment tracking added for the order, otherwise no tracking row is shown.
- The view controller was updated to config UI styling of the tracking and add tracking row, as well as handle the actions on these rows. Tracking rows allow 3 actions:
  - View tracking details on a web view
  - Copy tracking number
  - Delete tracking.
- The view controller and view model were updated to reload the table view when there's change in either add-ons, shipping labels or shipment tracking.

# Demo

# Testing
- Navigate to Orders tab
- Select Mark Order Complete
- You should then be navigated to Review Order screen
- You should find Tracking section if conditions in Solution are met.
- When Add Tracking button is tapped, Add Tracking screen is presented. When new tracking is added and Add Tracking screen is dismissed, Review Order screen should show the new tracking immediately.
- Copying tracking code should work properly and include haptic feedback
- Viewing tracking detail should navigate to a web view of the specified tracking company with the tracking code.
- Delete tracking should remove the tracking properly.

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Review Order/ReviewOrderViewController.swift
```diff
@@ -284,6 +343,100 @@ private extension ReviewOrderViewController {
...
+
+        actionSheet.addCancelActionWithTitle(TrackingAction.dismiss)
+
+        actionSheet.addDefaultActionWithTitle(TrackingAction.copyTrackingNumber) { [weak self] _ in
+            self?.sendToPasteboard(tracking.trackingNumber, includeTrailingNewline: false)
+        }
+
+        if tracking.trackingURL?.isEmpty == false {
+            actionSheet.addDefaultActionWithTitle(TrackingAction.trackShipment) { [weak self] _ in
+                self?.openTrackingDetails(tracking)
```
pmusolino, line 378: Q: Do we not track any analytics events here?

pmusolino (approved):
LGTM 🚀 

Code ✅ 
Testing ✅

pmusolino:
A thing that I noticed immediately after the review. When in landscape mode, the left, and right margins are wrong. They should be like in the Order List ;) feel free to fix it in another PR 😉 

Simulator Screen Shot - iPhone 12 Pro - 2021-07-09 at 12 04 15

WooCommerce/Classes/ViewRelated/Orders/Order Details/Review Order/ReviewOrderViewController.swift
itsmeichigo: I have taken a look at the Analytics sheet and saw that the analytics events for order fulfillment were deprecated since 5.6. I will post a question to ask the team about updated events for this new screen, but for now I guess we will not have analytics (I found [it's the same](https://github.com/woocommerce/woocommerce-android/pull/4086) on Android too).

itsmeichigo:
Thank you @pmusolino for helping me review this! As for the left and right margins of the table view, I think it's better that we keep the leading and trailing constraints to the super view instead of safe area since table views can now adjust its contents to not be clipped off by the edges (you can see from your screenshot 😄). 

I think we should migrate other screens to have the same look to0 because it'll be more intuitive that way (@Ecarrion suggested the same thing during my trial!)

PR #4552, opened 2021-07-06
[Mobile Payments] Use "cancel style" for secondary modal buttons

### LINKED ISSUE #4507: [Mobile Payments] Wrong colors for auxiliary button
I just noticed that the auxiliary button in `CardPresentPaymentsModalViewController` has the wrong style while being pressed.

Screen Shot 2021-06-30 at 13 47 29

**To Reproduce**

This is shown is several screens, but to show an example:

1. Go to Settings > Manage Card Readers
2. Tap "Connect Card Reader"
3. Turn on your reader
4. When the reader is found press and hold the "Keep searching" button

koke:
Fixes #4507 

The secondary button in the payment modals was showing the wrong colors while pressed:

Normal | Highlighted
-|-
IMG_DBF20042A5B5-1|IMG_6391

I started trying to fix the background, but then I realized that the designs didn't have this kind of borderless button, so I switched it to the alternate "ModalCancelButtonStyle".

## Screenshots (when pressed)

Before:

Light|Dark
-|-
IMG_6391|IMG_6392

After:

Light|Dark
-|-
IMG_0004|IMG_0005

## To test

This is used in several modals, but the easiest one to test:

1. Go to Settings > Manage Card Reader
2. Tap on Connect Card Reader
3. Turn on the reader and wait for the alert to show it's avaiable
4. Press and old Keep Searching

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIButton+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/UIButton+Helpers.swift
@@ -68,37 +68,6 @@ extension UIButton {
         setBackgroundImage(disabledBackgroundImage, for: .disabled)
     }
 
-    /// Applies the Secondary Button Style: Solid BG, always light!
-    ///
-    func applySecondaryLightButtonStyle() {
-        contentEdgeInsets = Style.defaultEdgeInsets
-        layer.borderColor = UIColor.primaryButtonBorder.cgColor
-        layer.borderWidth = Style.defaultBorderWidth
-        layer.cornerRadius = Style.defaultCornerRadius
-        titleLabel?.applyHeadlineStyle()
-        enableMultipleLines()
-        titleLabel?.textAlignment = .center
-
-        setTitleColor(.black, for: .normal)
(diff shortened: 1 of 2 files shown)


ctarda (approved):
Some screenshots while the button is being tapped:





:shipit:

ghost (changes requested):
@koke @ctarda all of these look great thanks. However, for the text buttons, I think we should just lower the opacity when it's pressed. Could we try 50%, please

Screen Shot 2021-07-07 at 1 16 35 pm

koke:
> for the text buttons

@adamzelinski I didn't see the text buttons in the Figma file, so I changed them to have a border. Is there a newer design that I missed? Can you confirm if we want those to be borderless?

Screen Shot 2021-07-07 at 08 52 32

ghost:
> I didn't see the text buttons in the Figma file, so I changed them to have a border. Is there a newer design that I missed? Can you confirm if we want those to be borderless?

Let's keep them with the border thanks @koke

ghost (approved):

PR #4554, opened 2021-07-06
[Mobile Payments] Show Collect Payment button only for eligible orders with a total amount greater than zero

### LINKED ISSUE #4254: [Mobile Payments] Collect Payment button is available on free  and paid orders
I have an order with a $0 amount. It is a virtual free product, but I imagine a regular product with a coupon for the full price will have the same effect.

When I view the order from the app, I have the option to "Collect Payment", even though there's nothing to pay. If I try to pay, it fails immediately

Screen Shot 2021-05-21 at 15 42 09 | Screen Shot 2021-05-21 at 15 42 50
-|-

Similarly, I had an order that already seems to be paid for and the Collect Payment button is available. These have been generated with [wc-smooth-generator](https://github.com/woocommerce/wc-smooth-generator) using `wp wc generate orders --status=processing`

Screen Shot 2021-05-27 at 10 35 29

ctarda:
Closes #4254 partially

As noted in #4254 the "Collect Payment" button is available for orders with a total amount of zero. On example might be an order containing only free products, or an order with a coupon for a 100% discount applied.

| Before | After |
| ---- | ---- |
|  |  |

## Changes
* Add logic to check that the total order amount is greater than zero, and hide the button if it isn't.
* Check also that the order has not been paid. 

## How to test
* Create an order. Make it be a total amount of zero (for example, by applying a 100% coupon code)
* Navigate to the order details screen. The Collect Payment button should not be visible.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
+++ b/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
@@ -67,7 +67,7 @@ extension Order {
         let customerNote = customerNote ?? self.customerNote
         let dateCreated = dateCreated ?? self.dateCreated
         let dateModified = dateModified ?? self.dateModified
-        let datePaid = datePaid ?? self.datePaid
+        let datePaid = datePaid ?? nil
         let discountTotal = discountTotal ?? self.discountTotal
         let discountTax = discountTax ?? self.discountTax
         let shippingTotal = shippingTotal ?? self.shippingTotal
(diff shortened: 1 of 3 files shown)


koke (commented):
Not sure what's going on yet, but I'm not seeing the Collect Payment button on any order. It seems like `order.datePaid` is set for all my orders 🤔

koke (approved):
My bad, it seems that all my orders generated with wc-smooth-generator were actually flagged as paid, so I guess I've been collecting payment on paid orders all this time 😅 

I created a new manual order and the button was present.

ctarda:
😅 Thanks for the review @koke 

I'll merge for now, give this one goes into develop we have time to pick up any potential breakages.

PR #4555, opened 2021-07-06
[Mobile Payments] Fix "Keep Searching" for readers

### LINKED ISSUE #4553: [Mobile Payments] Keep searching for readers fails every time
**Describe the bug**
When you are searching for readers, if you don't connect to the first one that appears and want to keep searching, it will fail every time

**To Reproduce**
1. Go to Settings > Manage Card Reader
2. Tap on Connect Card Reader
3. Turn on your reader and wait it to be detected
4. Tap Keep Searching
5. Wait a few seconds and an error will appear

**Screenshots**

https://user-images.githubusercontent.com/8739/124582016-eaec3d00-de51-11eb-8e0f-17cf7f222699.mp4

koke:
Fixes #4553

In #4339, we implemented a system to restart the discovery process when the Keep Searching button was tapped. Then, in #4351 we introduced a locking mechanism to avoid starting a discovery process if one was already in progress.

I'm not sure if I didn't test that one properly or something else changed since then, but the issue here is that I relied on a `defer` block to release the lock in the `asyncAfter` block so I didn't have to put the `unlock` on each specific case (success/failure). However, using `defer` means that the lock will be released when the current closure is finished, but we are calling the `promise` before that, which will eventually trigger a new discovery process, which will try to acquire the lock and deadlock until it times out a second later, failing with a `busy` error.

Instead of using `defer`, we now manually release the lock right before calling `promise` on each scenario, which seems to work nicely.

## To test

1. Go to Settings > Manage Card Reader
2. Tap on Connect Card Reader
3. Turn on your reader and wait it to be detected
4. Tap Keep Searching
5. Wait a few seconds. The reader should be detected

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -153,14 +153,14 @@ extension StripeCardReaderService: CardReaderService {
                 // Horrible, terrible workaround.
                 // And yet, it is the classic "dispatch to the next run cycle".
                 DispatchQueue.main.asyncAfter(deadline: .now() + 1) { [discoveryLock] in
-                    // When we're done with the completion, we let go of the discovery lock
-                    defer { discoveryLock.unlock() }
                     guard let error = error else {
                         self?.switchStatusToIdle()
+                        discoveryLock.unlock()
                         return promise(.success(()))
                     }
 
                     self?.internalError(error)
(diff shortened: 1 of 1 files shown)


ctarda (approved):
Ouch.

The funny thing is that the old code seems to work well on the simulator, at last on my machine.

Maybe I tested with the simulator at the time and that's why we didn't notice?

Anyway, tested again, on device and on simulator, and it seems to work perfect.

:shipit:

PR #4561, opened 2021-07-06
Shipping Label: Refactor ReprintShippingLabel for use printing new shipping labels

rachelmcr:
Part of #4089

## Description

This refactors the `Print Shipping Label` UI (view controller, view model, and coordinator), renaming the classes and methods to make it clearer to use in the shipping label purchase flow (not just for reprinting existing shipping labels).

A followup PR will include the changes to customize that screen for use in the shipping label purchase flow.

## Changes

* Renamed classes, methods, etc. to use "print" instead of "reprint" everywhere, unless it's specifically about reprinting.
* No visual changes.

## Testing

1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more orders with a purchased shipping label.
2. Open an order detail that has an existing shipping label.
3. Tap the "Print Shipping Label" button and confirm the "Print Shipping Label" screen appears.
4. Confirm you can still interact with this screen as expected (change the paper size, tap the help text, print the label).

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
@@ -497,8 +497,8 @@ private extension OrderDetailsViewController {
                 assertionFailure("Cannot reprint a shipping label because `navigationController` is nil")
                 return
             }
-            let coordinator = ReprintShippingLabelCoordinator(shippingLabel: shippingLabel, sourceViewController: navigationController)
-            coordinator.showReprintUI()
+            let coordinator = PrintShippingLabelCoordinator(shippingLabel: shippingLabel, sourceViewController: navigationController)
+            coordinator.showPrintUI()
         case .createShippingLabel:
             let shippingLabelFormVC = ShippingLabelFormViewController(order: viewModel.order)
             navigationController?.show(shippingLabelFormVC, sender: self)
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -430,8 +430,8 @@ private extension ShippingLabelFormViewController {
         }
 
         // TODO: Customize the reprint shipping label VC
-        let printCoordinator = ReprintShippingLabelCoordinator(shippingLabel: purchasedShippingLabel, sourceViewController: navigationController)
-        printCoordinator.showReprintUI()
+        let printCoordinator = PrintShippingLabelCoordinator(shippingLabel: purchasedShippingLabel, sourceViewController: navigationController)
+        printCoordinator.showPrintUI()
     }
 }
 
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
@@ -2,13 +2,13 @@ import Combine
 import Yosemite
 import UIKit
 
-/// Allows the user to select a paper size and reprint a shipping label given the selected paper size.
+/// Allows the user to select a paper size and print a shipping label given the selected paper size.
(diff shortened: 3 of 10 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -430,8 +430,8 @@ private extension ShippingLabelFormViewController {
         }
 
         // TODO: Customize the reprint shipping label VC
-        let printCoordinator = ReprintShippingLabelCoordinator(shippingLabel: purchasedShippingLabel, sourceViewController: navigationController)
-        printCoordinator.showReprintUI()
+        let printCoordinator = PrintShippingLabelCoordinator(shippingLabel: purchasedShippingLabel, sourceViewController: navigationController)
```
allendav, line 433: Minor. We could update the TODO comment (reprint->print) above this line too

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
```diff
@@ -259,9 +259,9 @@ private extension ReprintShippingLabelViewController {
     enum Localization {
         static let navigationBarTitle = NSLocalizedString("Print Shipping Label",
                                                           comment: "Navigation bar title to print a shipping label")
-        static let reprintButtonTitle = NSLocalizedString("Print Shipping Label",
+        static let printButtonTitle = NSLocalizedString("Print Shipping Label",
                                                           comment: "Button title to generate a shipping label document for printing")
-        static let paperSizeSelectorTitle = NSLocalizedString("Paper Size", comment: "Title of the paper size selector row for reprinting a shipping label")
+        static let paperSizeSelectorTitle = NSLocalizedString("Paper Size", comment: "Title of the paper size selector row for printing a shipping label")
```
allendav, line 264: Should the header text be updated too (next line)? It says "print it again" and "reprinting"

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -497,8 +497,8 @@ private extension OrderDetailsViewController {
                 assertionFailure("Cannot reprint a shipping label because `navigationController` is nil")
```
allendav, line 497: How about this case (`.reprintShippingLabel`) and the related reprint symbols in `OrderDetailsDataSource`?

allendav (approved):
I think we should update the header (see below), and I was surprised where I was left after printing (see 2nd screengrab below) but otherwise this is good to go. A few other comments to consider here and there. :shipit: 

Header


After Printing

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
rachelmcr: This TODO is resolved in the followup PR (#4562) 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
rachelmcr: I left that as-is because that header text will only be used when we're reprinting an existing shipping label. I have customizations for this screen in the followup PR that will swap out the header text for use in the purchase flow (when the label is being printed for the first time).

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
rachelmcr: This case is specifically for reprinting an existing shipping label, so I think it makes sense to leave it as-is. (Order Details displays existing/already purchased shipping labels, which can then be reprinted, whereas the shipping label purchase flow allows you to print labels for the first time after purchase.)

rachelmcr:
Hey @allendav, thanks for the review! Sorry, I think my description was a little confusing — the actual customization of this screen for use in the purchase flow is in #4562, and addresses the concerns you had about the header text and navigation.

PR #4562, opened 2021-07-06
Shipping Labels: Customize print label screen for purchase flow

rachelmcr:
Part of #4089
Relies on #4561

## Description

This customizes the `Print Shipping Label` screen so it can be used in the shipping label purchase flow (in addition to reprinting existing shipping labels).

This PR also hooks up the "Mark this order complete and notify the customer" toggle in the Shipping Label Form, so the order is marked complete when that toggle is enabled.

## Changes

* Adds a `PrintType` to `PrintShippingLabelCoordinator`, to handle whether it is printing a new label or reprinting an existing label.
* Customizes `PrintShippingLabelViewController` based on the `PrintType`. The screen doesn't change when reprinting a label, but when printing a label it shows the relevant header text, header image, and "Save for Later" button in addition to the print options.
* Removes the `Shipping Label Form` screen from the navigation stack, so when you tap the back button on the `Print Shipping Label` screen you go back to the order details.
* Hooks up the "Mark this order complete and notify the customer" toggle in the `Shipping Label Form` screen. When the label purchase is successful, if this toggle is enabled the order is marked as complete.

Reprint shipping label (existing labels) | Print shipping label (new purchase) | Print shipping label (after selecting paper size)
-|-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-07-06 at 15 07 25|Simulator Screen Shot - iPhone 11 Pro - 2021-07-06 at 15 08 16|Simulator Screen Shot - iPhone 11 Pro - 2021-07-06 at 15 08 23

## Testing

5. Complete all the fields in the form.
6. In the Order Summary section, enable the "Mark order complete" toggle and tap the button "Purchase Label".
7. On the Print Shipping Label screen, select the paper size and confirm the "Print Shipping Label" button is enabled and you can tap it to print your label.
8. Confirm you can tap the help text ("See layout ..." and "Don't know how ...") for more information.
9. Confirm you can tap the "Save for Later" button and you are taken back to the order details, which updates to display your purchased shipping label and shows the order with a `Completed` status.

Additional testing:

* Repeat the steps above, but don't enable the "Mark order complete" toggle and confirm your order is not marked as complete.
* Repeat the steps above, but tap the Back button on the Print Shipping Label screen and confirm you are taken back to the order details (not the Shipping Label Form).

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -290,11 +290,11 @@ private extension ShippingLabelFormViewController {
             let discountInfoVC = ShippingLabelDiscountInfoViewController()
             let bottomSheet = BottomSheetViewController(childViewController: discountInfoVC)
             bottomSheet.show(from: self, sourceView: cell)
-        } onSwitchChange: { (switchIsOn) in
-            self.shouldMarkOrderComplete = switchIsOn
-        } onButtonTouchUp: {
-            self.displayPurchaseProgressView()
-            self.viewModel.purchaseLabel { [weak self] result in
+        } onSwitchChange: { [weak self] (switchIsOn) in
+            self?.shouldMarkOrderComplete = switchIsOn
+        } onButtonTouchUp: { [weak self] in
+            self?.displayPurchaseProgressView()
+            self?.viewModel.purchaseLabel { [weak self] result in
                 guard let self = self else { return }
                 switch result {
                 case .success:
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -430,7 +430,9 @@ private extension ShippingLabelFormViewController {
         }
 
         // TODO: Customize the reprint shipping label VC
-        let printCoordinator = PrintShippingLabelCoordinator(shippingLabel: purchasedShippingLabel, sourceViewController: navigationController)
+        let printCoordinator = PrintShippingLabelCoordinator(shippingLabel: purchasedShippingLabel,
+                                                             printType: .print,
+                                                             sourceViewController: navigationController)
         printCoordinator.showPrintUI()
     }
 }
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -299,12 +299,13 @@ private extension ShippingLabelFormViewController {
                 switch result {
                 case .success:
                     self.onLabelPurchase?(self.shouldMarkOrderComplete)
+                    self.dismiss(animated: true)
                     self.displayPrintShippingLabelVC()
                 case .failure:
                     // TODO: Implement and display error screen for purchase failures
+                    self.dismiss(animated: true)
                     break
                 }
-                self.dismiss(animated: true)
             }
         }
         cell.isOn = false
@@ -437,12 +438,19 @@ private extension ShippingLabelFormViewController {
         present(inProgressViewController, animated: true)
     }
 
+    /// Removes the Shipping Label Form from the navigation stack and displays the Print Shipping Label screen.
+    /// This prevents navigating back to the purchase form after successfully purchasing the label.
+    ///
     func displayPrintShippingLabelVC() {
         guard let purchasedShippingLabel = viewModel.purchasedShippingLabel,
               let navigationController = navigationController else {
             return
         }
 
+        if let indexOfSelf = navigationController.viewControllers.firstIndex(of: self) {
+            let viewControllersExcludingSelf = Array(navigationController.viewControllers[0..<indexOfSelf])
+            navigationController.setViewControllers(viewControllersExcludingSelf, animated: false)
+        }
         let printCoordinator = PrintShippingLabelCoordinator(shippingLabel: purchasedShippingLabel,
                                                              printType: .print,
                                                              sourceViewController: navigationController,
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -9,6 +9,18 @@ final class ShippingLabelFormViewController: UIViewController {
 
     private let viewModel: ShippingLabelFormViewModel
 
+    /// Can be set to true to mark the order as complete when the label is purchased
+    ///
+    private var shouldMarkOrderComplete = false
+
+    /// Assign this closure to be notified after a shipping label is successfully purchased
+    ///
+    var onLabelPurchase: ((_ isOrderComplete: Bool) -> Void)?
+
+    /// Assign this closure to be notified after a shipping label is saved for later
+    ///
+    var onLabelSave: (() -> Void)?
+
     /// Init
     ///
     init(order: Order) {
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
```diff
@@ -6,25 +6,27 @@ import UIKit
 /// Informational links are displayed for printing instructions and paper size options.
 final class PrintShippingLabelViewController: UIViewController {
     @IBOutlet private weak var tableView: UITableView!
-    @IBOutlet private weak var printButton: UIButton!
+    @IBOutlet weak var reprintButton: UIButton!
```
pmusolino, line 12: Should we continue to use the generic `printButton` name, which is probably more appropriate?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
```diff
@@ -110,10 +118,14 @@ private extension PrintShippingLabelViewController {
...
+        guard printType == .reprint else {
+            reprintButton.isHidden = true
+            return
+        }
+        reprintButton.applyPrimaryButtonStyle()
+        reprintButton.setTitle(Localization.printButtonTitle, for: .normal)
+        reprintButton.on(.touchUpInside) { [weak self] _ in
             self?.printShippingLabel()
         }
     }
```
pmusolino, line 134: It's not clear to me why we have a `reprintButton`, since we already have a row for it.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
```diff
@@ -185,21 +201,69 @@ private extension PrintShippingLabelViewController {
...
+                .headerText,
+                .infoText,
+                .spacerBetweenInfoTextAndPaperSizeSelector,
+                .paperSize,
+                .spacerBetweenPaperSizeSelectorAndInfoLinks,
+                .paperSizeOptions,
+                .printingInstructions
+            ]
+        }
+        return rows.compactMap { $0 }
```
pmusolino: nit: it seems that we can't have `nil` values, so the `compactMap` should be superfluous here.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
```diff
@@ -185,21 +201,69 @@ private extension PrintShippingLabelViewController {
...
-        cell.textLabel?.text = Localization.headerText
+        switch printType {
+        case .print:
+            cell.textLabel?.text = Localization.printHeaderText
+            cell.textLabel?.applyHeadlineStyle()
+            cell.textLabel?.textAlignment = .center
+        case .reprint:
+            cell.textLabel?.text = Localization.reprintHeaderText
+            cell.textLabel?.applyBodyStyle()
+        }
```
pmusolino: nit: it would be good to have the `textAlignment` also inside the `reprint` case. The main reason is that since we are using reusable cells, the `textAlignment` can remain `.center` in the case for some reason, there is a switch between `print` and `reprint`.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
```diff
@@ -248,24 +316,50 @@ private extension PrintShippingLabelViewController {
...
         cell.selectionStyle = .default
     }
+
+    func configurePrintButtonRow(cell: ButtonTableViewCell) {
+        cell.configure(style: .primary,
+                       title: Localization.printButtonTitle,
+                       topSpacing: Constants.buttonVerticalSpacing,
+                       bottomSpacing: Constants.buttonVerticalSpacing) {
+            self.printShippingLabel()
+        }
```
pmusolino: nit: It's important here to add `[weak self]` because we don't want a strong reference that can cause a retain cycle 😉 
```suggestion
        cell.configure(style: .primary,
                       title: Localization.printButtonTitle,
                       topSpacing: Constants.buttonVerticalSpacing,
                       bottomSpacing: Constants.buttonVerticalSpacing) { [weak self] in
            self?.printShippingLabel()
        }
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
```diff
@@ -248,24 +316,50 @@ private extension PrintShippingLabelViewController {
...
+        cell.enableButton(selectedPaperSize != nil)
+    }
+
+    func configureSaveButton(cell: ButtonTableViewCell) {
+        cell.configure(style: .secondary,
+                       title: Localization.saveButtonTitle,
+                       topSpacing: Constants.buttonVerticalSpacing,
+                       bottomSpacing: Constants.buttonVerticalSpacing) {
+            self.saveLabelForLater()
+        }
```
pmusolino: nit: same here about the `[weak self]`

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -279,18 +291,21 @@ private extension ShippingLabelFormViewController {
             let bottomSheet = BottomSheetViewController(childViewController: discountInfoVC)
             bottomSheet.show(from: self, sourceView: cell)
         } onSwitchChange: { (switchIsOn) in
```
pmusolino: nit: `[weak self]` to avoid a retain cycle

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -279,18 +291,21 @@ private extension ShippingLabelFormViewController {
...
-            // TODO: Handle order completion
+            self.shouldMarkOrderComplete = switchIsOn
         } onButtonTouchUp: {
```
pmusolino: nit: same here, we should use `[weak self]`

pmusolino (commented):
Thank you 💯 for working on this Rachel.

I reviewed it and I've mainly a question about the `reprintButton`, it's not totally clear to me why it's in a separate view and we don't use a `Row` for it.

Another important thing, not related to this PR: while I was testing this, I discovered a new decoding error, that we should track and solve, ideally in another PR.
I'm not sure why this is happening, because initially I thought it was a problem related to choosing a custom package, but it also happens with the predefined ones on this specific order of my testing website. Basically, it seems that `service_name` can't be `nil`.
The creation of the shipping label is done correctly, but then it is not possible to download it and see it in the order. On the contrary, on the web, I can now see many shipping labels bought for the same order (this means that I can mistakenly buy multiple shipping labels, and there is no server-side control on this).

```
Error synchronizing shipping labels: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labelsData", intValue: nil), _JSONKey(stringValue: "Index 3", intValue: 3), CodingKeys(stringValue: "service_name", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
```

This is a video of what happens:
https://user-images.githubusercontent.com/495617/124739608-500b6580-df1a-11eb-9245-40299894c534.mp4

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
rachelmcr: The `reprintButton` is from the original implementation of the `ReprintShippingLabelViewController`. The design for that screen has the print button at the bottom of the screen, and as I understand it, there's no way to do that with a table row.

So I decided to leave that `reprintButton` so it could be pinned to the bottom of the screen for reprinting a label, and introduced a row for the print button for printing a new label. Is there a better way to handle those differences in the design?

Reprint label (print button at bottom) | Print new label (print button in table)
-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-07-06 at 15 07 25|Simulator Screen Shot - iPhone 11 Pro - 2021-07-06 at 15 08 23

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
rachelmcr: I decided to use `reprintButton` because this button is only used for the design where the button is at the bottom of the screen (when reprinting a label). If that's confusing I could definitely change it.

pmusolino: Oh got it, now it's more clear. It would be enough to add a comment on why there is this extra button, to avoid confusion. 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewController.swift
rachelmcr: Agh, thanks for catching all of these places where I forgot to add `[weak self]`!

rachelmcr:
Thank you for the review, @pmusolino! I addressed your feedback so this is ready for another look.

And thanks for noting that decoding error, how weird. I filed that in https://github.com/woocommerce/woocommerce-ios/issues/4568 and can look into it after finishing #4089 (almost done!).

pmusolino (approved):
Looks good to me :shipit:

PR #4563, opened 2021-07-06
Update "sort and filter" product toolbar behaviour

### LINKED ISSUE #4516: Products List → Sort and Filter buttons should be hidden if there are no products
**Describe the bug**

The Sort and Filter buttons are visible even if there are no products.

**To Reproduce**

Use a site with no products and navigate to the Products list.

**Expected behavior**

They should be hidden.

fernandofgfer:
Fixes #4516 
# Description
When a store doesn't have products, the app should hide the sort and filter buttons. 
If the user adds a product as a first one, the app should show again the buttons.

# Testing
Case 1: Store without products.
- Login with a store without products.
- Check that buttons don't appear.

Case 2: Add first product.
- Login with a store without products.
- Check that buttons don't appear.
- Add one product.
- Check that buttons appear.

Case 3: Remove last product.
- Get a Store with only one product.
- Remove this product.
- Check that buttons don't appear.

# Videos
start with one product
https://user-images.githubusercontent.com/11411847/124625749-5ea43f00-de7e-11eb-93ed-0100542ca815.mp4

start with 0 products
https://user-images.githubusercontent.com/11411847/124633692-d4f86f80-de85-11eb-84c8-530908cf7e0b.mp4

delete last product
https://user-images.githubusercontent.com/11411847/124625886-7d0a3a80-de7e-11eb-993b-700705d0c9b5.mp4

add first product
https://user-images.githubusercontent.com/11411847/124625970-8eebdd80-de7e-11eb-9a95-5de756e58fb6.mp4


WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -404,6 +404,14 @@ private extension ProductsViewController {
     func registerTableViewCells() {
         tableView.register(ProductsTabProductTableViewCell.self)
     }
+
+    func showOrHideToolBar(resultController: ResultsController<StorageProduct>) {
```
Ecarrion: A couple of comments here!

-  Since `ResultsController` is part of `self`, I don't think we need to pass it as a parameter, or am I missing something?

- We try to maintain all of our methods documented, could you add some comments here, please?

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -404,6 +404,14 @@ private extension ProductsViewController {
...
+        if resultsController.isEmpty {
+            toolbar.isHidden = true
+        } else {
+            toolbar.isHidden = false
+        }
```
Ecarrion: I think this could be 
```suggestion
        toolbar.isHidden = resultsController.isEmpty
```

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -694,6 +704,8 @@ private extension ProductsViewController {
         }
         displayEmptyStateViewController(emptyStateViewController)
         emptyStateViewController.configure(config)
+        // Hide toolbar when user doesn't have products
+        toolbar.isHidden = true
```
Ecarrion: Could you explain to me why is this needed? I would think that the change introduced above was enough! Maybe I'm missing something. 😅

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -500,11 +508,13 @@ private extension ProductsViewController {
...
-        resultsController.onDidChangeContent = {
+        resultsController.onDidChangeContent = { [weak self] in
+            self?.showOrHideToolBar(resultController: resultsController)
             onReload()
         }
 
-        resultsController.onDidResetContent = {
+        resultsController.onDidResetContent = { [weak self] in
+            self?.showOrHideToolBar(resultController: resultsController)
             onReload()
```
Ecarrion: I think you could move `showOrHideToolBar` to the `onReload` declaration and avoid a little bit of repetition, right?

Ecarrion (changes requested):
Thanks for working on this @fernandofgfer!

I did found a problem with your proposed solution. If you add a filter where 0 products are returned, then you are stuck and you can't go back to your product list.


https://user-images.githubusercontent.com/562080/124665629-f2740c00-de7a-11eb-9144-9f77c11d73bf.mov

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: - First comment: I wanted to do a pure function, but in the end is better like you said. I'm going to use isEmpty properly.
- Second comment: No problem, comment added

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: Good catch! Added!

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: Yep, it's needed because we're using the functions `onDidChangeContent` and `onDidResetContent`, but if a user doesn't have products since the first time that opened the view, we need this to remove the toolbar

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: Totally agree, moved to the declaration! thanks!

fernandofgfer:
I've fixed a small bug that didn't use the `onDidResetContent` and `onDidChangeContent` of the ProductsViewController, removing `resultsController.stopForwardingEvents()` and `resultsController.startForwardingEvents(to: tableView)`, because this method were changing the `onDidResetContent` and `onDidChangeContent` functions of the ProductController.

fernandofgfer:
> Thanks for working on this @fernandofgfer!
> 
> I did found a problem with your proposed solution. If you add a filter where 0 products are returned, then you are stuck and you can't go back to your product list.
> 
>  Screen.Recording.2021-07-06.at.4.42.59.PM.mov

Hi! I've pushed a solution. Now if you have any filter applied, the bar will be visible always. What do you think?

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -404,6 +405,15 @@ private extension ProductsViewController {
...
         tableView.register(ProductsTabProductTableViewCell.self)
     }
+
+    /// Show or hide the toolbar based on number of products
+    /// if there is any filter applied, toolbar must be always visible
+    /// If there is 0 products, toolbar will be hidden
+    /// if there is 1 or more products, toolbar will be visible
+    ///
+    func showOrHideToolBar() {
+        toolbar.isHidden = filters.numberOfActiveFilters == 0 ? self.isEmpty : false
```
Ecarrion:
```suggestion
        toolbar.isHidden = filters.numberOfActiveFilters == 0 ? isEmpty : false
```

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -663,16 +673,12 @@ private extension ProductsViewController {
...
-
-        resultsController.stopForwardingEvents()
     }
 
     /// Removes the Placeholder Products (and restores the ResultsController <> UITableView link).
     ///
     func removePlaceholderProducts() {
         tableView.removeGhostContent()
-        resultsController.startForwardingEvents(to: tableView)
-        tableView.reloadData()
```
Ecarrion: I'm a bit concerned that these are being removed. AFAIK these are there because they can cause crashes to get updates while the ghost table view is being shown. 

One way to solve it is to use a separate table view for the ghost cells. Could you explain more deeply why these need to be removed? and possibly consider another alternative to avoid crashes?

Ecarrion (changes requested):
I found a bug that precedes your changes but since you are working on this it's probably worth it to fix it right away.

If you add a product from the "empty state screen" the screen is not updating with the new product. We should remove the "empty state" screen when we detect that there are products available to show.


https://user-images.githubusercontent.com/562080/124834134-3afbfa00-df4d-11eb-84df-89e3cdebc273.mp4

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: The problem here is where `startForwardingEvents` is called, `onDidChangeContent` and `onDidResetContent` change their assignation, so `onReload` is not called anymore, and that the reason why I removed it.
I've done a small update, where these methods have come back again, and I assign again `onDidChangeContent` and `onDidResetContent` to  `onReload` after `stopForwardingEvents`. Could you check it, please?

fernandofgfer: I've recovered the removed code and I've change `resultsController.startForwardingEvents(to: tableView)` by `setClosuresToResultController`. Let me know your thoughts 😄

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: Removed, thanks!

fernandofgfer:
> I found a bug that precedes your changes but since you are working on this it's probably worth it to fix it right away.
> 
> If you add a product from the "empty state screen" the screen is not updating with the new product. We should remove the "empty state" screen when we detect that there are products available to show.
> 
>  Simulator.Screen.Recording.-.iPhone.12.-.2021-07-07.at.17.50.22.mp4

I've added a function to remove the overlay in case that resultsController has one or more products.

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -65,9 +65,7 @@ final class ProductsViewController: UIViewController {
     ///
     private lazy var resultsController: ResultsController<StorageProduct> = {
         let resultsController = createResultsController(siteID: siteID)
-        configureResultsController(resultsController) { [weak self] in
-            self?.tableView.reloadData()
-        }
+        configureResultsController(resultsController, onReload: reloadTableAndView)
```
Ecarrion, line 68: Nice!

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -499,23 +506,49 @@ private extension ProductsViewController {
...
-            try resultsController.performFetch()
-        } catch {
-            ServiceLocator.crashLogging.logError(error)
-        }
-
+    /// Manages view components and reload tableview
+    ///
+    func reloadTableAndView() {
+        showOrHideToolBar()
+        removeOverlayIfControllerHasResults()
```
Ecarrion: I think we could update this function to something along the lines of `updateOverlays` and decide if we need to show or hide the overlay depending on the content of our results controller, because at the moment, we have a case where the "no products" overlay is not shown after we remove the last product. And we are forced to pull to refresh.


https://user-images.githubusercontent.com/562080/124985118-0c8f2500-e008-11eb-8e43-0d95e663ec47.mp4

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -694,6 +727,8 @@ private extension ProductsViewController {
         }
         displayEmptyStateViewController(emptyStateViewController)
         emptyStateViewController.configure(config)
+        // Hide toolbar when user doesn't have products
+        showOrHideToolBar()
```
Ecarrion: nit: I this call belongs on the `didEnter` method in the `.noResultsPlaceholder` case. As this method is only for showing the no products overlay and nothing more!

Ecarrion (changes requested):
This looks almost complete! Some little details to address and we should be good to go!

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: Added! I think this bug happened before this task, so we fixed two different bugs doing this PR, that's awesome! 💪

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
fernandofgfer: You're right. I've moved to a new function that is called in init method.

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -499,23 +513,50 @@ private extension ProductsViewController {
...
+    ///
+    func reloadTableAndView() {
+        showOrHideToolBar()
+        removeOverlayIfControllerHasResults()
         tableView.reloadData()
     }
+
+    /// Try to remove Overlay if resultController has products
+    ///
+    func removeOverlayIfControllerHasResults() {
```
Ecarrion: I think we should also update the method name, as right now it solely implies that the overlay will be removed but not displayed.

fernandofgfer: Totally forgot! Refactored!

jkmassel:
We're cutting `7.1` today, so this is being bumped to `7.2`

Ecarrion (approved):

PR #4565, opened 2021-07-06
[Mobile Payments] Add CardReaderSettingsKnown(Readers)ViewModel

allendav:

This PR adds a new view model CardReaderSettingsKnownViewModel that will return shouldShow true when we are searching for readers AND we have one or more remembered readers in app settings.

Although very similar to CardReaderSettingsUnknownViewModel, a separate view model is desired to keep that view model (and view) single-concern, i.e. because
- the view controller this view model pairs with needs to initiate search automatically
- this view model needs to connect automatically if discovery yields a known reader ID
- the view controller this view model will be paired with may soon need different content (than the CardReaderSettingsUnknownViewController has today) since the merchant will have connected successfully to a reader already

The next PR will add that new view controller.

Note that we have not added this view model to CardReaderSettingsViewModelsOrderedList. That comes in a later PR.

Note that removing the older known reader logic from places like the CardReaderSettingsUnknownViewModel comes in a later PR in this sequence. The sequence of planned PRs can be found in the last comment on #3559 

To test:

No user facing changes for this viewmodel
Ensure linting passes
Ensure unit tests pass

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownViewModel.swift
```diff
@@ -0,0 +1,228 @@
...
+final class CardReaderSettingsKnownViewModel: CardReaderSettingsPresentedViewModel {
+
+    private(set) var shouldShow: CardReaderSettingsTriState = .isUnknown
+    var didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?
+    var didUpdate: (() -> Void)?
+
+    private var noConnectedReaders: CardReaderSettingsTriState = .isUnknown
+    private var noKnownReaders: CardReaderSettingsTriState = .isUnknown
+    private var knownReadersProvider: CardReaderSettingsKnownReadersProvider?
+    private var siteID: Int64 = Int64.min
```
ctarda: It seems like the siteID is assigned in the initializer and never modified? If that's the case, it could be a let instead of a var

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownViewModel.swift
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    /// Updates whether the view this viewModel is associated with should be shown or not
+    /// Notifes the viewModel owner if a change occurs via didChangeShouldShow
+    ///
+    private func reevaluateShouldShow() {
+        var newShouldShow: CardReaderSettingsTriState = .isUnknown
+
+        if ( noKnownReaders == .isUnknown ) || ( noConnectedReaders == .isUnknown ) {
```
ctarda: I'm surprised the linter didn't yell at you because of the spaces after the parenthesis 🤷🏼

ctarda (approved):
:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownViewModel.swift
allendav: Huh! Me too. Other coding standards die hard.

allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4565/commits/82ca59bdac562dc1a7f3408ff317d37f20fbabfc

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsKnownViewModel.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4565/commits/82ca59bdac562dc1a7f3408ff317d37f20fbabfc

PR #4570, opened 2021-07-07
Shipping Label: Handle missing service_name in shipping label

### LINKED ISSUE #4568: Shipping Labels: Decoding error when service_name is nil
**Describe the bug**

As reported by @pmusolino in https://github.com/woocommerce/woocommerce-ios/pull/4562#pullrequestreview-700818832:

I'm not sure why this is happening, because initially I thought it was a problem related to choosing a custom package, but it also happens with the predefined ones on this specific order of my testing website. Basically, it seems that `service_name` can't be `nil`.
The creation of the shipping label is done correctly, but then it is not possible to download it and see it in the order. On the contrary, on the web, I can now see many shipping labels bought for the same order (this means that I can mistakenly buy multiple shipping labels, and there is no server-side control on this).

```
Error synchronizing shipping labels: valueNotFound(Swift.String, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), CodingKeys(stringValue: "labelsData", intValue: nil), _JSONKey(stringValue: "Index 3", intValue: 3), CodingKeys(stringValue: "service_name", intValue: nil)], debugDescription: "Expected String value but found null instead.", underlyingError: nil))
```

This is a video of what happens:
https://user-images.githubusercontent.com/495617/124739608-500b6580-df1a-11eb-9245-40299894c534.mp4

rachelmcr:
Fixes: #4568

## Description

In rare cases, a shipping label can be missing its `service_name`. This was causing the Order Details screen to not show any shipping labels for an order, even if the order had one or more labels. This PR updates the `ShippingLabel` model to handle when `service_name` is `nil`, defaulting to an empty string.

Usual behavior (with `service_name`) | Missing `service_name` (order details) | Missing `service_name` (shipment details)
-|-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-07-07 at 16 14 23|Simulator Screen Shot - iPhone 11 Pro - 2021-07-07 at 16 46 45|Simulator Screen Shot - iPhone 11 Pro - 2021-07-07 at 16 46 55

@pmusolino I'm requesting a review from you because I'm not sure how to reproduce the original issue and I know you can see this issue on your store.

## Testing

1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more orders with a purchased shipping label with a missing `service_name`.
2. Open an order detail for an order with such a label (with a missing `service_name`).
3. Confirm all the labels appear in the order detail.
4. Tap "View Shipment Details" for the label with a missing `service_name` and confirm the details appear.

## Submitter Checklist

Update release notes:

### DIFF
--- a/Networking/Networking/Model/ShippingLabel/ShippingLabel.swift
+++ b/Networking/Networking/Model/ShippingLabel/ShippingLabel.swift
@@ -32,6 +32,7 @@ public struct ShippingLabel: Equatable, GeneratedCopiable, GeneratedFakeable {
     public let trackingNumber: String
 
     /// The name of service for the shipping label (e.g. "USPS - Media Mail").
+    /// Can sometimes be empty, see: https://github.com/woocommerce/woocommerce-ios/issues/4568
     public let serviceName: String
 
     /// The amount that is refundable.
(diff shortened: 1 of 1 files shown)


pmusolino (approved):
It works like a charm 🥳

PR #4572, opened 2021-07-07
[Mobile Payments] Add view controller for when we know at least one card reader to search for

allendav:
This is the second to last PR in a series of PRs for the "automatically reconnect to known card readers" feature

This PR takes the Known Readers View Model introduced in #4565 and adds its associated the Known Readers View Controller (and Storyboard content).

This view controller is very similar to the existing CardReaderSettingsUnknownViewController but automatically initiates search on loading the view. In time, we may want to tailor the messaging on this view, since the merchant will have connected successfully already once, but for now the messaging is the same.

The **next** (and final) PR adds the VM and VC to the CardReaderSettingsViewModelsOrderedList, removes legacy Known Readers logic from the VMs and VCs, and retires the feature flag. The entire known readers flow will be testable then - i.e. reviewers will be able to exercise the VC in the next PR.
 
The sequence of planned PRs can be found here: #3559 (comment)

To test:

Ensure lint-ing passes
Ensure unit tests pass

Update release notes:


koke (commented):
> This view controller is very similar to the existing CardReaderSettingsUnknownViewController but automatically initiates search on loading the view

I wonder if it makes sense to make a distinction in the UI layer between having a known reader or not. These VC/VM pairs are almost identical except for the initial state. `CardReaderSettingsConnectedView*` is more clearly a different view, but I don't think these two deserve separate VCs.

I'd suggest splitting the two possible child views into connected/disconnected instead.

Thinking about decoupling the auto-connection logic from the UI, I'm also now wondering if it makes as much sense to auto-connect when you specifically go into Settings > Manage Card Readers, as it does when you tap on Collect Payment. This goes beyond this PR, but if nothing has auto-connected yet, going into settings I think it would feel a bit off if a modal was presented as soon as you navigate to that screen. At the same time, if there is a known reader when you tap collect payment, having the full modal in the background seems odd if the "Scanning for reader" alert shows automatically and the modal will be dismissed when a reader is connected.

allendav:
Thanks @koke - I've been carefully considering your feedback and I think you are onto something...

> I wonder if it makes sense to make a distinction in the UI layer between having a known reader or not.

I still think it _can_ since the "unknown" state is primarily intended for merchants that have never connected to a reader before and the "known" state is primarily intended for merchants that have. In time I expect the settings views for those two states will diverge.

> Thinking about decoupling the auto-connection logic from the UI, I'm also now wondering if it makes as much sense to auto-connect when you specifically go into Settings > Manage Card Readers, as it does when you tap on Collect Payment.  This goes beyond this PR, but if nothing has auto-connected yet, going into settings I think it would feel a bit off if a modal was presented as soon as you navigate to that screen.

I still think autoreconnect in settings makes sense, but I agree the modal is surprising. This is where we could perhaps present a different known full screen VC that says it is attempting to auto reconnect to a specific known reader, prompts the merchant to turn on the reader if they haven't already, and gives a button to explicitly search for a different reader. Let's see how the merchant feed back goes during the initial phases of the beta before we make changes like this though?

> At the same time, if there is a known reader when you tap collect payment, having the full modal in the background seems odd if the "Scanning for reader" alert shows automatically and the modal will be dismissed when a reader is connected.

I agree. I've chatted with @ctarda about this as well. I have some ideas on how we could have the discovery modals in the collect payment flow NOT fire up the "unknown" (or "known") VC too and just use the order detail view as its background. I'll write those up in a P2 post soon.

allendav:
@koke @ctarda - I think this is ready for a final round? I'm going to turn on auto-merge

koke (approved):
Not sure how to test these yet, but code looks good

PR #4574, opened 2021-07-07
Variations: Show notice after creating a variation

### LINKED ISSUE #4452: Variations: Updates after generate variations.
When we select “Generate New Variation” on the Variation list, can we:
1. show a “Variation created” snackbar too?
2. display the newly created variation first on the list? (right now it gets added at the bottom of the list)

reported by @Garance91540

Ecarrion:
fixes #4452 

# Why

It was reported during testing 7.0 that it will be nice to show a success notice after creating variations. Previously we displayed the notice but only after creating the first variation.

# How

- Show a notice using `noticePresenter` after creating a variation in `ProductVariationsViewController`.

- Added a local `DDLOG` if an error happened.

# Demo

Simulator Screen Recording - iPhone 12 - 2021-07-07 at 16 57 43

# Testing Steps

- Launch the app and navigate to a variable product with variations
- Tap on the variations row
- Tap on the "generate variation" button
- See that after a successful generation a "Success Notice" is shown.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
@@ -614,11 +614,14 @@ extension ProductVariationsViewController: SyncingCoordinatorDelegate {
             progressViewController.dismiss(animated: true)
 
             guard let self = self else { return }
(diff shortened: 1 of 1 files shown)


itsmeichigo (approved):
✅ Code OK
✅ Works correctly as described:


LGTM :shipit:

PR #4576, opened 2021-07-08
Move Fastlane secrets outside of repository

mokagio:
Moves the Fastlane-related secrets outside of the repository into `~/.configure/woocommerce-ios/secrets/`

To test locally, checkout this branch, run `bundle exec run configure_apply` and then inspect the content of `~/.configure/woocommerce-ios/secrets/`:

```
$ ls ~/.configure/woocommerce-ios/secrets
app_store_connect_fastlane_api_key.json
google_cloud_keys.json
project.env
```

I also opened #4610 to test these changes. It exercises each secret by:

1. Loading Fastlane, which checks the `project.env` location
2. Calling `match` for each build configuration, to test the `google_cloud_keys.json` location
3. Making a build and uploading it to TestFlight, to test the ASC API key JSON location. _This is expected to [fail](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/12963/workflows/7167d4bd-8e36-4f02-b63f-f64cdfd7a155/jobs/23871) because a build with the same version and build number already exists in on App Store Connect, but the fact that we get that error is validation that the API key file can be found._

---

Update release notes:

### DIFF
--- a/.configure
+++ b/.configure
@@ -5,7 +5,7 @@
   "files_to_copy": [
     {
       "file": "shared/google_cloud_keys.json",
-      "destination": ".configure-files/google_cloud_keys.json",
+      "destination": "~/.configure/woocommerce-ios/secrets/google_cloud_keys.json",
       "encrypt": true
     },
     {
--- a/.configure
+++ b/.configure
@@ -15,7 +15,7 @@
     },
     {
       "file": "iOS/WCiOS/project.env",
-      "destination": ".configure-files/project.env",
+      "destination": "~/.configure/woocommerce-ios/secrets/project.env",
       "encrypt": true
     },
     {
--- a/.configure
+++ b/.configure
@@ -20,7 +20,7 @@
     },
     {
       "file": "iOS/app_store_connect_fastlane_api_key.json",
-      "destination": ".configure-files/app_store_connect_fastlane_api_key.json",
+      "destination": "~/.configure/woocommerce-ios/secrets/app_store_connect_fastlane_api_key.json",
       "encrypt": true
     }
   ],
--- a/fastlane/Matchfile
+++ b/fastlane/Matchfile
@@ -1,9 +1,9 @@
 # This Matchfile has the shared properties used for all signing types
 
 # Store certs/profiles encrypted in Google Cloud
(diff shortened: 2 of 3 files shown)


fastlane/Matchfile
```diff
@@ -1,9 +1,10 @@
 # This Matchfile has the shared properties used for all signing types
 
 # Store certs/profiles encrypted in Google Cloud
-storage_mode("google_cloud")
-google_cloud_bucket_name("a8c-fastlane-match")
-google_cloud_keys_file(".configure-files/google_cloud_keys.json")
+storage_mode('google_cloud')
+google_cloud_bucket_name('a8c-fastlane-match')
```
mokagio, line 5: Sorry for the noise in the diff. While I was on this file, I converted Ruby `""` into `''`, which is the recommended syntax for strings that don't use interpolation.

.configure
```diff
@@ -15,16 +15,16 @@
...
+      "destination": "~/.configure/woocommerce-ios/secrets/app_store_connect_fastlane_api_key.json",
       "encrypt": true
     }
   ],
   "file_dependencies": [
 
   ]
-}
\ No newline at end of file
+}
```
AliSoftware: It's the little things ❤️

.configure
```diff
@@ -5,7 +5,7 @@
   "files_to_copy": [
```
AliSoftware, line 5: (We definitively need to make our `configure` tool support a `files_to_remove` entry there at some point, to be able to nicely deprecate old locations…)

AliSoftware (approved):
The PR looks good to me code-wise, and I saw you tested it in the other PR, so it's a go for me. ✅ 

(Approving, but not merging myself because I'm thinking maybe Jeremy, as WC RM, might want to do a review too? 🤷 )

.configure
jkmassel: This is a super good idea

jkmassel (approved):

jkmassel:
Landing this now for testing with the next beta

mokagio:
Thanks @jkmassel! 

Now I wonder why I didn't merge it when Olivier approved it 🤔

PR #4577, opened 2021-07-08
Tracks events for Shipping Labels M2 and M3

pmusolino:
Closes #4509 

## Description
In this PR I added all the analytics events for Shipping Labels M2 and M3, which will be released to our users from the next release.

## Events included
- [x] `shipping_label_purchase_flow`: Action events associated with the shipping label purchase flow.
- [x] `shipping_label_order_fulfill_succeeded`: The request to mark an order as fulfilled after a shipping label purchase succeeded.
- [x] `shipping_label_order_fulfill_failed`: The request to mark an order as fulfilled after a shipping label purchase failed
- [x] `shipping_label_discount_info_button_tapped`: The user tapped the discount button in shipping labels form
- [x] `shipping_label_edit_address_open_map_button_tapped`: The user tapped on the open map button in the address editor
- [x] `shipping_label_edit_address_done_button_tapped`: The user tapped the done button to validate the address
- [x] `shipping_label_edit_address_use_address_as_is_button_tapped`: The user used the button "Use address as is" to save the address without validation
- [x] `shipping_label_edit_address_contact_customer_button_tapped`: The user tapped on contact customer button
- [x] `shipping_label_address_suggestions_use_selected_address_button_tapped`: The user selected one of the suggested addresses
- [x] `shipping_label_address_suggestions_edit_selected_address_button_tapped`: The user tapped on edit selected address button
- [x] `shipping_label_address_validation_failed`: The address validation failed
- [x] `shipping_label_address_validation_succeeded`: The address validation succeeded

## Testing
Since there are a lot of events, the easiest thing to do is to read the list of events above and perform the indicated action to make sure it appears in the console.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4578, opened 2021-07-08
Enabled the Shipping Label creation for all the US store owners

### LINKED ISSUE #4541: Shipping Labels M3: enable the feature flag for all the end-users
Before release the Shipping Labels M2 and M3 together, we should enable the feature flag for all the end-users.

We should also update the `RELEASE-NOTES` about the new feature.

pmusolino:
Closes #4541 

## Description
I enabled the `shippingLabelsM2M3` feature flag for all the end-users.
From release 7.1, merchants from US can create shipping labels for physical orders from the app, the feature supports for now only orders where the shipping address is in the US.

## Testing
The app should run in release mode to test this. It will be tested in the TestFlight build, so, just check the code and make sure the CI is ✅

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,7 @@
 
 7.1
 -----
-- [***] Merchants from US can create shipping labels for physical orders from the app. The feature supports for now only orders where the shipping address is in the US.
+- [***] Merchants from US can create shipping labels for physical orders from the app. The feature supports for now only orders where the shipping address is in the US. [https://github.com/woocommerce/woocommerce-ios/pull/4578]
 - [*] Fix: Interactive pop gesture on Order Details and Settings screen. [https://github.com/woocommerce/woocommerce-ios/pull/4504]
 - [*] Fix: Frozen refresh control and placeholder when switching tabs [https://github.com/woocommerce/woocommerce-ios/pull/4505]
 - [internal] Stats tab: added network sync throttling [https://github.com/woocommerce/woocommerce-ios/pull/4494]
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,7 @@
 
 7.1
 -----
+- [***] Merchants from US can create shipping labels for physical orders from the app. The feature supports for now only orders where the shipping address is in the US.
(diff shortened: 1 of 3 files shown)


rachelmcr:
> It will be tested in the TestFlight build

That's true, but I do think this is a good time for us to step back and confirm that we're releasing what we expected and the full flow works from beginning to end at least in happy paths. (That way the testing that happens in the TestFlight build can focus on things like diverse store configurations, unexpected paths, etc.)

I'll do a quick run-through of the use cases identified in our M3 project board:

- [ ] I want to see an error banner if I don’t have any packages (custom or default) created for the store.
- [x] I want to see what carriers and rates are available for a shipment when creating a label.
- [x] I want to be reminded of what the customer paid for shipping on an order when viewing carrier rates.
- [x] I want to see optional services for a carrier rate after selecting that rate.
- [x] I want to see the payment details for a shipment when creating a label.
- [x] After entering all details, I will see the rate with Woo discount on shipping labels screen.
- [x] I want to learn more about what the WooCommerce discount is.
- [x] I want to optionally not mark the order as complete when purchasing a shipping label.
- [x] I want to be able to purchase a shipment label after all the options are entered.
- [x] I want to know if there was an error during purchasing of a shipment label.
- [x] I want to print a shipping label that has been purchased.
- [x] I want to be able to change the paper size for a label.

pmusolino:
This is a great idea @rachelmcr 🙌

rachelmcr (approved):
Code ✅ 
Testing:

All of our use cases are working except this:

> I want to see an error banner if I don’t have any packages (custom or default) created for the store.

The ability to create packages in the client is part of the remote eligibility check. If the store doesn't have any packages, then in the M3 release they won't be eligible to create shipping labels and the option to create a shipping label just doesn't appear.

Did we decide not to implement this error banner, or is this an oversight? (I don't see any designs for it.) Either way, I've approved the PR because everything else is working and I don't think this is significant enough to stop us from releasing this milestone.

pmusolino:
@rachelmcr It will be part of M4 or M5, since in one of these milestones we will allow the store owner to create custom packages.

pmusolino:
I'll merge this just after this one https://github.com/woocommerce/woocommerce-ios/pull/4577

PR #4580, opened 2021-07-08
Shipping Labels: Smoother navigation bar transition for UIHostingViewController

### LINKED ISSUE #4336: Shipping Labels: bar button items & Navigation titles appear after the view is transitioned
@Ecarrion discovered [here](https://github.com/woocommerce/woocommerce-ios/pull/4327) that the bar button items & navigation titles appear after the view is transitioned in Package Details and Shipping Carrier and Rates. Those, for the moment, are the only two flows implemented in SwiftUI under the Shipping Labels, so we should investigate.

https://user-images.githubusercontent.com/562080/120236202-5cb4e380-c221-11eb-98c3-7bc073896758.mov

itsmeichigo:
Fixes #4336 

# Description
This PR aims to make the transition in navigation bar of UIHostingViewController smoother.

# Solution
As discussed in #4336, it seems that Apple is not supporting the transition animation of pushing SwiftUI view into an existing navigation controller stack that was instantiated with UIKit. This causes an abrupt appearance of the title and navigation bar items of the SwiftUI view, as they take some time to be rendered after its hosting view controller is pushed into the stack. So I had to do 2 steps before pushing the hosting view controller:
- Set title for the embedding hosting view controller.
- Set a fake navigation bar button without any action.

The title and the bar button will be replaced by the actual one rendered by the SwiftUI view. Therefore we have our desired views in place when the controller is in transition, and still have proper feature after the view is rendered.

# Demo

# Problem
If you look closely at the demo, you'll see that the color of the bar button in disabled state is different between UIKit and SwiftUI. This is a facepalm to me - I can't seem to get around that but I think this is still better than the old transition issue.

# Testing
- Navigate to Shipping Labels Creation form
- Navigate to SwiftUI views: package details, carrier & rates, payment methods
- Notice that the title and bar buttons of these views are visible during the transition.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -395,6 +395,11 @@ private extension ShippingLabelFormViewController {
         }
 
         let hostingVC = UIHostingController(rootView: packageDetails)
+        // Hack to make transition of navigation bar smooth
+        hostingVC.title = Localization.packageDetailsNavigationBarTitle
+        let doneButton = UIBarButtonItem(title: Localization.packageDetailsDoneButton, style: .done, target: nil, action: nil)
+        doneButton.isEnabled = false
+        hostingVC.navigationItem.rightBarButtonItem = doneButton
         navigationController?.show(hostingVC, sender: nil)
     }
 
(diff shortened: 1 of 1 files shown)


Ecarrion:
Hi @itsmeichigo thanks for caring about this! 

But as I expressed in https://github.com/woocommerce/woocommerce-ios/issues/4336#issuecomment-876455122 I don't think this approach is worth it for the minimum UX improvement that it brings to the table.

Looking back, I created that issue because I saw something off but no one else has even noticed it 😅.

Given how manual, how much duplication, and that it does not scale well, I think it's better to wait for a possible apple fix than to carry that burden ourselves. 

I think we can revisit this if/when people start noticing it or if we came up with another alternative.

What do you think?

itsmeichigo:
Hi Ernesto, I think that's fair! It's fun to notice the issue and find the workaround, although not a very nice one. I can close this for now 👍

PR #4585, opened 2021-07-09
Inform translators of the constraints of the "99+" string

### LINKED ISSUE #4502: Badge counts should be 99+ for values greater than 100
Steps to reproduce:

1. Have a site with 99+ unread orders, products, and reviews.
1. View the badge count for each area.

**Result:** the badge count display should be consistent for each area and should show 99+ for numbers greater than 100.

> When there are more than 99 orders the text doesn’t quite fit the label, I think we should be using 99+ since that would work nicely on all languages:

image

<sup>(internal reference: p5T066-2rI-p2#comment-8963 /hat tip @elibud)</sup>

**Note:** possibly a regression of https://github.com/woocommerce/woocommerce-ios/pull/1764.

shiki:
Closes #4502. 

Eli reported this unsightly Spanish translation of "99+". 

## Findings

We originally thought that it's a problem in the logic to determine whether to display 99+ or not. But it looks like it's just a problem in the translation. 

https://github.com/woocommerce/woocommerce-ios/blob/fc83fa9638f013457ac653774c67581471deeadc/WooCommerce/Resources/es.lproj/Localizable.strings#L149-L150

## Solution

I wanted to solve this in a future-proof way. I asked Platform (p1625777265302800-slack-CC7L49W13) and they said that changing the comment would make GlotPress fetch this again and request new translations for "99+". So I did that and changed the comment and provided a (hopefully) helpful instruction for our dear translators. 

## Testing

You can verify that the new `comment` is used by running this in the root of the project (make sure your Python version is less than 3.0):

```
./Scripts/localize.py
```

And then verify that the comment was replaced:

```
git diff WooCommerce/Resources/en.lproj/Localizable.strings
```

I'll also make sure that this is sent to GlotPress when the time comes to make a release. 

## Author Checklist

### DIFF
--- a/WooCommerce/Classes/Extensions/NumberFormatter+LocalizedOrNinetyNinePlus.swift
+++ b/WooCommerce/Classes/Extensions/NumberFormatter+LocalizedOrNinetyNinePlus.swift
@@ -15,7 +15,8 @@ extension NumberFormatter {
     private enum Constants {
         static let ninetyNinePlus = NSLocalizedString(
             "99+",
-            comment: "Shown when there are more than 99 items of something (e.g. Processing Orders)."
+            comment: "Please limit to 3 characters if possible. This is used if " +
+                "there are more than 99 items like in the Orders tab."
         )
     }
 }
--- a/WooCommerce/Classes/Extensions/NumberFormatter+LocalizedOrNinetyNinePlus.swift
+++ b/WooCommerce/Classes/Extensions/NumberFormatter+LocalizedOrNinetyNinePlus.swift
@@ -16,7 +16,7 @@ extension NumberFormatter {
         static let ninetyNinePlus = NSLocalizedString(
             "99+",
             comment: "Please limit to 3 characters if possible. This is used if " +
-                "there are more than 99 items like in the Orders tab."
+                "there are more than 99 items in a tab, like Orders."
         )
     }
 }


WooCommerce/Classes/Extensions/NumberFormatter+LocalizedOrNinetyNinePlus.swift
```diff
@@ -15,7 +15,8 @@ extension NumberFormatter {
     private enum Constants {
         static let ninetyNinePlus = NSLocalizedString(
             "99+",
-            comment: "Shown when there are more than 99 items of something (e.g. Processing Orders)."
+            comment: "Please limit to 3 characters if possible. This is used if " +
+                "there are more than 99 items like in the Orders tab."
```
koke: While I think it's correct, that "like" made the sentence a bit hard to parse for me. Maybe we can drop it if this is only used in orders, or switch it around: "...if there are more than 99 items in a tab, like Orders"

koke (approved):

WooCommerce/Classes/Extensions/NumberFormatter+LocalizedOrNinetyNinePlus.swift
shiki: Thank you for being thorough, @koke. 🙇 I used your second suggestion. 404547878772e8e4416197447867927537ac6282

PR #4592, opened 2021-07-12
[Mobile Payments] Hide the collect payment button when the net amount pending does not match the total order amount

### LINKED ISSUE #4583: [Mobile Payments] Order with partial refund is still charged full order amount instead of net amount
**Describe the bug**

If I have an order that is already partially refunded, and I try to collect payment for the remaining amount in the app, the card is charged for the **Order Total** amount (the full order amount before the refund) instead of the **Net Payment** amount (the remaining amount due after the refund).

In addition, all of the products are shown on the order receipt, including any refunded products.

**To Reproduce**
1. Make sure your store is set up to accept in-person payments.
2. Create a new order on your store with more than one product.
3. Partially refund the order, so there is still at least one product in the order.
4. In the app, open the order detail for that order.
5. Tap the "Collect payment" button.
6. Notice that the modal that appears shows more than the customer should have to pay for the remaining order.
7. Collect the payment, and notice that the app shows the **Net Payment** as the remaining amount on the order, but wp-admin shows the full amount was charged to the card.

**Screenshots**

Order with refund ($14 left to pay) | Collect payment ($32 charge) | Receipt (both products appear) | Order details after payment (Net Payment is $14)
-|-|-|-
IMG_6800|IMG_6801|IMG_6802|IMG_6803

Charge in wp-admin:
Screen Shot 2021-07-08 at 7 59 01 PM

ctarda:
Closes #4583 , at least until we have a better long term solution for it.

At this point in time, we can only collect payment for the full total order amount. Customizing the actual dollar amount that we charge using a credit card is not yet possible. That is a functionality that requires careful consideration and that we will attack as a full project.

In the meantime, a way to avoid overcharging customers that have already paid partially for their orders, or that want to pay for an order that has partially refunded is to hide the "Collect Payment " button for all orders where the total amount to charge does not match the total order amount.

|  | Before | After |
| ---- | ---- | ---- |
| Total amount pending |  |  |
| Partial refund |  |  |

## Changes
* Tweak the logic in OrderDetailsDataSource to mark the order as not eligible when the order has been partially charged
* Added a property to OrderPaymentDetailsViewModel to encapsulate what "being partially charged" means

## How to test
* Create a few orders. Leave at least one of them as it was created originally, and edit the others on the store admin to add partial refunds, total refunds, and whatever else you can come up with that makes the pending amount differ from the total order amount.
* Check that the "Collect Payment" button in the order details screen is available only or orders where the pending amount matches the total order amount

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
@@ -173,6 +173,12 @@ final class OrderPaymentDetailsViewModel {
         return order.coupons
     }
 
+    /// Signals whether the net amount for the order matches the total order amount
+    ///
+    var hasBeenPartiallyCharged: Bool {
+        return totalValue != netAmount
+    }
+
     init(order: Order, refund: Refund? = nil, currencySettings: CurrencySettings = ServiceLocator.currencySettings) {
         self.order = order
         self.refund = refund
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -1429,6 +1429,10 @@ private extension OrderDetailsDataSource {
             return false
         }
 
+        guard let totalAmount = currencyFormatter.convertToDecimal(from: order.total), totalAmount.decimalValue > 0 else {
+            return false
+        }
+
         // If there is a discrepancy between the orderTotal and the remaining amount to collect, it is not eligible
         // This is a temporary solution that will exclude, for example:
         // * orders that have been partially refunded.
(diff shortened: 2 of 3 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderPaymentDetailsViewModel.swift
```diff
@@ -173,6 +173,12 @@ final class OrderPaymentDetailsViewModel {
         return order.coupons
     }
 
+    /// Signals whether the net amount for the order matches the total order amount
+    ///
+    var hasBeenPartiallyCharged: Bool {
```
Ecarrion, line 178: Should we add a quick unit test to this method?

Ecarrion (approved):
Code and tests looked good for me! 

I may be very outdated but just o confirm, is it expected for the "Issue Refund" button to be hidden when the "Collect Payment" button is visible?

ctarda:
Thanks for the review @Ecarrion 

>  is it expected for the "Issue Refund" button to be hidden when the "Collect Payment" button is visible?

I would say that at this point, it is. The "Collect Payment" button should only be visible on orders that are "Cash on delivery" and that are awaiting payment in total (therefore can not be refunded). That will change in the future though.

Issue #3716, opened 2021-02-22
Landscape backgrounds should extend edge-to-edge (not limited to safe area)


rachelmcr:
**Expected behavior**

In landscape orientation, all backgrounds should extend edge-to-edge (while still containing the contents within the safe area). Internal ref: p91TBi-4q2-p2

**Actual**

On some screens, the backgrounds stay within the safe area (leaving margins on the left/right). These should be fixed to have edge-to-edge backgrounds:

My store:
- [x] My store
- [x] Settings subsections

Orders:
- [x] Orders list
- [x] Order details
   - [x] Order details > Billing information
   - [x] Order details > Add tracking
   - [x] Order details > Add tracking > Shipping carriers
   - [x] Add order note
   - [ ] Add-on list

Reviews:
- [x] Reviews list
- [x] Review Details

Products:
- [x] Products list header
- [x] Products search
- [x] Products filter

Shipping Labels:
- [x] Shipping Labels creation flow

Product details:
- [x] Variable product details > Variations list
- [x] Price
   - [x] Price > Tax classes
- [x] Shipping
   - [x] Shipping > Shipping classes
- [x] Inventory
- [x] Categories
- [x] Linked Products
- [x] Downloadable Files (virtual, downloadable product)
- [x] Product Link (external product)
- [x] Product Settings
   - [x] Product Settings > Visibility
   - [x] Product Settings > Catalog Visibility
   - [x] Product Settings > Purchase Note

**Screenshots**

My store
![Simulator Screen Shot - iPhone 12 - 2021-02-09 at 17 37 27](https://user-images.githubusercontent.com/8658164/108706782-0429c800-7507-11eb-91a0-0239221f451c.png)

Orders list
![Simulator Screen Shot - iPhone 12 - 2021-02-09 at 17 37 55](https://user-images.githubusercontent.com/8658164/108706795-08ee7c00-7507-11eb-8109-df5a9eb2d055.png)

Reviews list
![Simulator Screen Shot - iPhone 12 - 2021-02-09 at 17 38 01](https://user-images.githubusercontent.com/8658164/108706806-0db33000-7507-11eb-820e-e7b3439d17bc.png)

Variable product details > Variations list
![Simulator Screen Shot - iPhone 12 - 2021-02-09 at 17 40 16](https://user-images.githubusercontent.com/8658164/108706818-1277e400-7507-11eb-8c95-a2e28222d2ab.png)

Product details > Price
[...]

PR #4593, opened 2021-07-12
Fix: leading and trailing padding in Product List

### LINKED ISSUE #4511: Product list in landscape has a wrong leading and trailing padding
Looking at the product list when your iPhone is in landscape mode, has a wrong leading and trailing padding.
In fact, the top banner and the header have the correct padding, but every cell is wrong since took the entire width of the screen. Instead, it should be like in the Order list.

| Product List (wrong padding)           |  Order List (correct padding) |
Simulator Screen Shot - iPhone 12 Pro - 2021-06-30 at 17 58 15 | Simulator Screen Shot - iPhone 12 Pro - 2021-06-30 at 18 00 46

pmusolino:
Fixes #4511 

## Description
Fixed the leading and trailing padding in Product List, when the screen is in landscape mode.

## Testing
Make sure the product list in landscape now has the correct leading/trailing padding.

## Screenshots
Simulator Screen Shot - iPhone 12 Pro - 2021-07-12 at 15 34 27 | Simulator Screen Shot - iPhone 12 Pro - 2021-07-12 at 15 47 48

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,9 @@
 *** PLEASE FOLLOW THIS FORMAT: [<priority indicator, more stars = higher priority>] <description> [<PR URL>]
 
+7.2
+-----
+- [*] Fix: leading and trailing padding in Product List, when the screen is in landscape mode. [https://github.com/woocommerce/woocommerce-ios/pull/4593]
+
 7.1
 -----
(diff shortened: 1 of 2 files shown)


Ecarrion (changes requested):
Hi @pmusolino, thanks for working on this.

I'm afraid we want to go in the opposite direction and [make all backgrounds go edge-to-edge.](https://github.com/woocommerce/woocommerce-ios/issues/3716) 

For this case, I think we need to update `TopBannerView` to have its background edge-to-edge!

Ref: p91TBi-4q2

pmusolino:
I totally missed that post @Ecarrion. Good to know that we are going in the opposite direction 👍

PR #4602, opened 2021-07-13
Login: Make sure account settings and site plan are synced immediately on login

### LINKED ISSUE #4600: Login: Account settings and site plan are not synchronized on login
**Describe the bug**

During login, [we call `ServiceLocators.stores.synchronizeEntities`](https://github.com/woocommerce/woocommerce-ios/blob/f9ce36df151aa7431b95d92784a0d1d2fbbadb8f/WooCommerce/Classes/Authentication/AuthenticationManager.swift#L355) which we expect will [synchronize the account, account settings, sites, and site plan](https://github.com/woocommerce/woocommerce-ios/blob/f9ce36df151aa7431b95d92784a0d1d2fbbadb8f/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift#L116-L147).

However, account settings relies on other synchronized data (the [userID](https://github.com/woocommerce/woocommerce-ios/blob/f9ce36df151aa7431b95d92784a0d1d2fbbadb8f/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift#L238), which is synced with the account). We don't wait for the account to be synced before attempting to sync the account settings, so it fails to sync. Those settings will sync the next time the app is launched, because the account has already been synced at that point.

The site plan also fails to sync. In this case, it's because it depends on the [defaultSite](https://github.com/woocommerce/woocommerce-ios/blob/f9ce36df151aa7431b95d92784a0d1d2fbbadb8f/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift#L283), which isn't set until the store is selected on the login epilogue screen (after the sync is performed).

This results in issues like #4480 where we expect that account settings are already in storage, but they aren't there until the app is relaunched.

rachelmcr:
Fixes: #4600

## Description

During login, we expect to synchronize the account, account settings, sites, and site plan. However, the account settings and site plan were never being synced immediately on login (only later, e.g. after the app was relaunched). They were being synced too soon — they depend on the default account user ID and default site ID, which weren't yet set.

This PR changes the timing of when those are synced:

* Account settings are synced only after the account is synced, to ensure the default account is set.
* Site plan is synced along with other site settings. This ensures the site ID is available before starting that sync (which, during the login flow, happens after the login epilogue is shown).

## Changes

In `DefaultStoresManager`:

* `synchronizeAccountSettings` is moved to the `synchronizeAccount` closure (still within `synchronizeEntities`) to ensure it is called after the account sync is complete.
* `synchronizeSitePlan` is moved from `synchronizeEntities` (which is called before the default site is selected in the login flow) to `synchronizeSettings` (which is called after the default site is selected in the login flow).

## Testing

* Launch the app and make sure you're logged out.
* Log in to your account.
* Confirm your account settings are synced immediately upon login. The easiest way to confirm this is to check that your Tracks preference is set with either `🔵 Tracking started.` or `🔴 Tracking opt-out complete.` logged in the console.
* Confirm your site plan is synced immediately upon confirming your selected site in the login epilogue screen. I believe the site plan is only used when opening a support ticket (to add a tag to the support ticket with your site plan), so the easiest way to confirm this is probably to set a breakpoint on `synchronizeSitePlan` to be sure it's being called.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
+++ b/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
@@ -121,6 +121,10 @@ class DefaultStoresManager: StoresManager {
 
         group.enter()
         synchronizeAccount { _ in
+            group.enter()
+            self.synchronizeAccountSettings { _ in
+                group.leave()
+            }
             group.leave()
         }
 
@@ -130,10 +134,7 @@ class DefaultStoresManager: StoresManager {
         }
 
         group.notify(queue: .main) {
-            // Wait to synchronize account settings because it depends on the userID from synchronizeAccount
-            self.synchronizeAccountSettings { _ in
-                onCompletion?()
-            }
+            onCompletion?()
         }
 
         return self
--- a/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
+++ b/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
@@ -129,11 +129,6 @@ class DefaultStoresManager: StoresManager {
             group.leave()
         }
 
-        group.enter()
-        synchronizeSitePlan { _ in
-            group.leave()
-        }
-
         group.notify(queue: .main) {
             // Wait to synchronize account settings because it depends on the userID from synchronizeAccount
             self.synchronizeAccountSettings { _ in
@@ -315,6 +310,14 @@ private extension DefaultStoresManager {
             group.leave()
         }
         dispatch(productSettingsAction)
+
+        group.enter()
+        synchronizeSitePlan { result in
+            if case let .failure(error) = result {
+                errors.append(error)
+            }
+            group.leave()
+        }
 
         group.notify(queue: .main) {
             if errors.isEmpty {
--- a/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
+++ b/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
@@ -271,18 +271,6 @@ private extension DefaultStoresManager {
         dispatch(action)
     }
 
-    /// Synchronizes the WordPress.com Site Plan.
-    ///
-    func synchronizeSitePlan(onCompletion: @escaping (Result<Void, Error>) -> Void) {
-        guard let siteID = sessionManager.defaultSite?.siteID else {
-            onCompletion(.failure(StoresManagerError.missingDefaultSite))
-            return
-        }
-
-        let action = AccountAction.synchronizeSitePlan(siteID: siteID, onCompletion: onCompletion)
-        dispatch(action)
-    }
-
     /// Synchronizes the settings for the specified site, if possible.
     ///
     func synchronizeSettings(with siteID: Int64, onCompletion: @escaping () -> Void) {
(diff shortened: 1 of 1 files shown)


itsmeichigo (commented):
Great job investigating the issue! 🎉 This seems to be an old issue that some how slipped through the crack, so it's nice that you got it fixed! I have only a few comments below though, mostly just nitpicking 😄

Also - do you think it's possible to write some tests for this as a way to document our syncing mechanism? We can maybe move this to a discussion later for future improvement, but it's just something that I think would be desirable.

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
```diff
@@ -121,11 +121,10 @@ class DefaultStoresManager: StoresManager {
 
         group.enter()
         synchronizeAccount { _ in
-            group.leave()
-        }
-
-        group.enter()
-        synchronizeAccountSettings { _ in
+            group.enter()
+            self.synchronizeAccountSettings { _ in
```
itsmeichigo: I think it'd be great if you can explicitly state your intention with a capture list - whether to keep `self` weak, strong or unowned. I know that `DefaultStoresManager` is kept alive all through the app's life, so in this case it should not cause trouble with a strong self. But do you think we should keep a weak or unowned self instead?

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
```diff
@@ -318,6 +312,14 @@ private extension DefaultStoresManager {
         }
         dispatch(productSettingsAction)
 
+        group.enter()
+        synchronizeSitePlan { result in
```
itsmeichigo: I see that the method `synchronizeSitePlan` is checking for the existence of `sessionManager.defaultSite?.siteID`. In this case we already have a `siteID` to work with, so I think that check is redundant - how about moving the content of the method out here like so:
```Swift
        group.enter()
        let sitePlanAction = AccountAction.synchronizeSitePlan(siteID: siteID) { result in
            if case let .failure(error) = result {
                errors.append(error)
            }
            group.leave()
        }
        dispatch(sitePlanAction)
```
We can then remove the method and the code will be easier to read I think.

WooCommerce/Classes/Yosemite/DefaultStoresManager.swift
rachelmcr: Thanks for pointing this out! I am learning more about memory management now because I keep forgetting to be explicit and consider when to keep `self` weak. I think it makes sense to keep a weak self here, instead. 👍

rachelmcr:
Thank you for the review, @itsmeichigo! I appreciate all nitpicks because I'd like learn from having them pointed out. :) This is ready for another look now.

> Also - do you think it's possible to write some tests for this as a way to document our syncing mechanism? We can maybe move this to a discussion later for future improvement, but it's just something that I think would be desirable.

As discussed in Slack, I definitely agree that it would be good to have test coverage for this sync code and we can look together at adding that as a followup to this PR.

itsmeichigo (approved):
Thank you for the updates @rachelmcr 👍 Code looks good and works as expected now! :shipit:

Issue #4611, opened 2021-07-14
CPP Onboarding: Epic Phase 1


koke:
Provide merchants with information about the next necessary step they need to take in order to enable in-person payments on their stores.

Designs are available on vaw3DvewbUwuQnPXpdNGGH-fi-370%3A468 - **we should consider implementing any deviations from the scope (p91TBi-5uD-p2) in the next phase of this project.**

Eligibility Checks
- [x] Introduce a feature flag (PR https://github.com/woocommerce/woocommerce-ios/pull/4612)
- [x] Create initial generic error page (PR https://github.com/woocommerce/woocommerce-ios/pull/4613)
- [x] Create initial settings page with success state (PR https://github.com/woocommerce/woocommerce-ios/pull/4615)
- [x] Loading state UI (#4759)
- [x] Unsupported country state UI (#4750)
- [x] Unsupported country state business logic (#4724, #4730)
- [x] WCPay not installed state UI (#4738)
- [x] WCPay not installed state business logic (#4729)
- [x] WCPay not activated state UI (PR https://github.com/woocommerce/woocommerce-ios/pull/4751)
- [x] WCPay not activated state business logic (#4729)
- [x] Unsupported WCPay version state UI (PR https://github.com/woocommerce/woocommerce-ios/pull/4752)
- [x] Unsupported WCPay version state business logic (#4729)
- [x] WCPay not setup state UI (#4761)
- [x] WCPay not setup state business logic (#4731)
- [x] Stripe Account Restricted – Under Review state UI (PR https://github.com/woocommerce/woocommerce-ios/pull/4760)
- [x] Stripe Account Restricted – Under Review state business logic (#4731)
- [x] Stripe Account Restricted – Pending Requirements state UI (PR https://github.com/woocommerce/woocommerce-ios/pull/4762)
- [x] Stripe Account Restricted – Pending Requirements state business logic (#4731)
- [x] Stripe Account Restricted – Overdue Requirements state UI (PR https://github.com/woocommerce/woocommerce-ios/pull/4763)
- [x] Stripe Account Restricted – Overdue Requirements state business logic (#4731)
- [x] Stripe Account Rejected state UI (PR https://github.com/woocommerce/woocommerce-ios/pull/4764)
- [x] Stripe Account Rejected state business logic (#4731)
- [x] ~Live Account in Test Mode state UI~ (moved to #4774)
- [x] ~Live Account in Test Mode state business logic~ (moved to #4774)
[...]

PR #4613, opened 2021-07-14
[CPP Onboarding] Generic error page

koke:
Part of #4611

This PR adds a temporary screen to show when In-Person Payments is unavailable. In this first PR, it will always show unavailable, and as the rest of the checks are progressively implemented, this screen should appear less and less, until it's eventually replaced with the definitive design for generic errors. For now, I've taking a few elements from the designs that would make this generic enough.

## To test

1. Go to settings
2. Tap on In-Person Payments
3. You should see this screen, regardless of whether your store supports payments

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsViewController.swift
@@ -1,0 +1,43 @@
+import SwiftUI
+
+final class InPersonPaymentsViewController: UIHostingController<InPersonPaymentsView> {
+    init() {
+        super.init(rootView: InPersonPaymentsView())
+    }
+
+    @objc required dynamic init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+struct InPersonPaymentsView: View {
+    var body: some View {
+        InPersonPaymentsUnavailableView()
+            .navigationTitle(Localization.title)
+    }
+}
+
+private enum Localization {
+    static let title = NSLocalizedString(
+        "In-Person Payments",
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsViewController.swift
```diff
@@ -0,0 +1,43 @@
...
+    }
+}
+
+private enum Localization {
+    static let title = NSLocalizedString(
+        "In-Person Payments",
+        comment: "Title for the In-Person Payments settings screen"
+    )
+
+    static let unavailable = NSLocalizedString(
```
ctarda: Is this really necessary? It seems like it is not used

ctarda (approved):
Code is looking good. 



It would be good to get input on the end result from @woocommerce/mobile-designers , and I think we can attack their feedback, if any, in a follow-up PR. But from the code point of view it's a :shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsViewController.swift
koke: Good catch, these used to be here but were moved to `InPersonPaymentsUnavailableView`

Garance91540:
Hi there! It's looking good except I think there's some padding issues for the text in the screenshot cesar shared: https://d.pr/i/jdbk4W

Let's try to always have a minimum of 24px left and right padding for those error screens.

koke:
> Let's try to always have a minimum of 24px left and right padding for those error screens.

Thanks, this screen is meant to be temporary, but since we'll likely use it as a base for all the other screens, I put up a PR correcting this #4624

PR #4615, opened 2021-07-14
[CPP Onboarding] Initial structure

koke:
Part of #4611
This should be merged after #4613, and includes the changes in that PR. ⚠️ This needs to retarget develop once #4613 is merged

This PR adds the basic structure to perform the necessary checks for In-Person Payments. There is a new `enum` with all the possible states, and a new `checkOnboardingState` action to evaluate the current state. Currently, it only performs the same checks that `SettingsViewController` was already doing to decide if the Manage Card Readers button was visible. That is, if there is a WCPay account that's set up and eligible for card-present payments.

If that's the case, the In-Person Payments screen will embed the existing screen to connect a card reader. Otherwise, it will show the generic error introduced in #4613.

## To test

Go to Settings > In-Person Payments:
1. With a store correctly set up for payments, you should see the option to connect to a reader
2. With a store that's not set up for payments, you should see a generic error

Available | Unavailable
-|-
Screen Shot 2021-07-14 at 13 23 58|Screen Shot 2021-07-14 at 13 24 11

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Model/Enums/CardPresentPaymentsOnboardingState.swift
+++ b/Yosemite/Yosemite/Model/Enums/CardPresentPaymentsOnboardingState.swift
@@ -42,7 +42,8 @@ public enum CardPresentPaymentOnboardingState {
     ///
     case stripeAccountOverdueRequirement
 
-    /// The Stripe account was rejected by Stripe. This can happen for example when the account is flagged as fraudulent or the merchant violates the terms of service.
+    /// The Stripe account was rejected by Stripe.
+    /// This can happen for example when the account is flagged as fraudulent or the merchant violates the terms of service.
     ///
     case stripeAccountRejected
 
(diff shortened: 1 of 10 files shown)


Yosemite/Yosemite/Model/Enums/CardPresentPaymentsOnboardingState.swift
```diff
@@ -0,0 +1,57 @@
+/// Represents the possible states for onboarding to In-Person payments
+public enum CardPresentPaymentOnboardingState {
```
ctarda, line 2: My goodness, that's a lot off states!

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -75,6 +77,19 @@ public final class CardPresentPaymentStore: Store {
 // MARK: - Services
 //
 private extension CardPresentPaymentStore {
+    func checkOnboardingState(siteID: Int64, onCompletion: (CardPresentPaymentOnboardingState) -> Void) {
```
ctarda, line 80: Would it be possible to add some tests for this method?

ctarda (approved):
It works perfect. 

This is a site enrolled into WCPay and properly configured:


And this is what happens on a store that does not have WCPay installed:


The code looks good, with one question about tests only.

:shipit:

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
koke: Created an issue for that in #4630

Issue #4604, opened 2021-07-13
Plugins: Storage layer


Ecarrion:
Parent: https://github.com/woocommerce/woocommerce-ios/issues/4278

## Tasks
- [ ] Add a new model
- [ ] Add migration test.
- [ ] Update Storage readme
- [ ] Add model to Network Layer

fernandofgfer:
**Quick update:**

- [x]  Add a new model
- [x]  Add migration test
- [x]  Update Storage readme
- [x]  Add model to Network Layer

Issue #4278, opened 2021-05-25
Plugins: Sync plugins from system_status


Ecarrion:
## Context

Plugin list synchronization was introduced in https://github.com/woocommerce/woocommerce-ios/issues/3921. However, that is only accessible for admin users.

## Requirement

We have the need of fetching the store plugin list from an endpoint that is accessible for all roles. 
`wc/v3/system_status` so we can determine which plugins are installed before presenting certain features, like "Present Card Payments"

## Extra

-  Provide a way to check if a plugin is installed or not.

## Considerations

- Don't delete the current plugin work made on https://github.com/woocommerce/woocommerce-ios/issues/3921 as that is still required for managing plugins as an admin.

## Breakdown

- [x] https://github.com/woocommerce/woocommerce-ios/issues/4604
- [x] https://github.com/woocommerce/woocommerce-ios/issues/4605
- [x] https://github.com/woocommerce/woocommerce-ios/issues/4606
- [x] https://github.com/woocommerce/woocommerce-ios/issues/4607

fernandofgfer:
Thanks, @Ecarrion solving my doubts by Slack.
I think we can split this issue into these smaller:

1. Storage layer:
- Add a new model
- Add migration test.
- Update Storage readme
- Add model to Network Layer
2. Yosemite Layer:
- Add Action
- Add Remote
- Add Mapper
- Add Storage
3. Recover plugins in plugin screen:
- Move sync site plugin to plugin screen
- Recover system plugins in plugin screen
4. Extra - Check plugins to enable features:
- Create component.
- Make an example with "Present Card Payments" feature.

Could you help me to create the issues, please? 
Thanks 😄

Ecarrion:
@fernandofgfer I have created the new issues and have linked them into the main issue description.
If you add a comment on them I could set you as the assignee!

PR #4617, opened 2021-07-14
System plugins - Adding new model into Storage layer

fernandofgfer:
#4604 - PR 1/2 - part from #4278 
## Description
This issue consists in add the new System plugin model to the Storage layer

## How
- Create new xcdatamodel version -> 54
- Add new entity to this datamodel
- Create NSManagedObject
- Add load and delete methods
- Update MIGRATIONS.md
- Add unit test for load an delete methods
- Add unit test for migration 53 -> 54
 
## How to test:
Launch unit tests

Update release notes:

### DIFF
--- a/Storage/Storage.xcodeproj/project.pbxproj
+++ b/Storage/Storage.xcodeproj/project.pbxproj
@@ -34,6 +34,8 @@
 		02DA64172313C26400284168 /* StatsVersionBySite.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02DA64162313C26400284168 /* StatsVersionBySite.swift */; };
 		02DA64192313C2AA00284168 /* StatsVersion.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02DA64182313C2AA00284168 /* StatsVersion.swift */; };
 		02EAB6D72480A86D00FD873C /* CrashLogger.swift in Sources */ = {isa = PBXBuildFile; fileRef = 02EAB6D62480A86D00FD873C /* CrashLogger.swift */; };
+		077F39C4269F1F4600ABEADC /* SystemPlugin+CoreDataClass.swift in Sources */ = {isa = PBXBuildFile; fileRef = 077F39C3269F1F4600ABEADC /* SystemPlugin+CoreDataClass.swift */; };
+		077F39C6269F1F7C00ABEADC /* SystemPlugin+CoreDataProperties.swift in Sources */ = {isa = PBXBuildFile; fileRef = 077F39C5269F1F7C00ABEADC /* SystemPlugin+CoreDataProperties.swift */; };
 		2618707325409C65006522A1 /* ShippingLineTax+CoreDataClass.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2618707125409C65006522A1 /* ShippingLineTax+CoreDataClass.swift */; };
 		2618707425409C65006522A1 /* ShippingLineTax+CoreDataProperties.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2618707225409C65006522A1 /* ShippingLineTax+CoreDataProperties.swift */; };
 		2619F71925AF95030006DAFF /* StorageTypeExtensionsTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2619F71825AF95020006DAFF /* StorageTypeExtensionsTests.swift */; };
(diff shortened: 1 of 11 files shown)


Storage/Storage.xcodeproj/project.pbxproj
```diff
@@ -45,7 +47,7 @@
 		262B270E2621412000A421CF /* ProductAddOn+CoreDataProperties.swift in Sources */ = {isa = PBXBuildFile; fileRef = 262B26752621401A00A421CF /* ProductAddOn+CoreDataProperties.swift */; };
 		262B27132621412700A421CF /* ProductAddOnOption+CoreDataClass.swift in Sources */ = {isa = PBXBuildFile; fileRef = 262B26762621401A00A421CF /* ProductAddOnOption+CoreDataClass.swift */; };
 		262B27182621412D00A421CF /* ProductAddOnOption+CoreDataProperties.swift in Sources */ = {isa = PBXBuildFile; fileRef = 262B26772621401A00A421CF /* ProductAddOnOption+CoreDataProperties.swift */; };
-		263E37D72641AD5100260D3B /* BuildFile in Frameworks */ = {isa = PBXBuildFile; productRef = 263E37D62641AD5100260D3B /* SwiftPackageProductDependency */; };
+		263E37D72641AD5100260D3B /* Codegen in Frameworks */ = {isa = PBXBuildFile; productRef = 263E37D62641AD5100260D3B /* Codegen */; };
```
Ecarrion: ⚠️  Are these changes necessary? 
Maybe we need to revert them.

Storage/Storage/Tools/StorageType+Extensions.swift
```diff
@@ -537,4 +537,22 @@ public extension StorageType {
...
         return firstObject(ofType: PaymentGatewayAccount.self, matching: predicate)
-    }}
+    }
+
+    // MARK: - System plugins
+
+    /// Returns all stored system plugins for a provided `siteID`.
+    ///
+    func loadSystemPlugins(siteID: Int64) -> [SystemPlugin] {
+        let predicate = \SystemPlugin.siteID == siteID
```
Ecarrion, line 555: Do you have any opinion on building keypaths like this?

Storage/Storage/Tools/StorageType+Extensions.swift
```diff
@@ -537,4 +537,22 @@ public extension StorageType {
...
+        let descriptor = NSSortDescriptor(keyPath: \SystemPlugin.name, ascending: true)
```
Ecarrion, line 556: [Q] Is there any reason to sort them by name?

Storage/Storage/Tools/StorageType+Deletions.swift
```diff
@@ -129,4 +129,15 @@ public extension StorageType {
             deleteObject($0)
         }
     }
+
+    // MARK: - System plugins
```
Ecarrion: Nit, I don't think it's necessary to add this MARKS, as we haven't been doing them in this file

Storage/Storage/Tools/StorageType+Deletions.swift
```diff
@@ -129,4 +129,15 @@ public extension StorageType {
...
+
+    /// Deletes all of the stored SystemPlugins for the provided siteID.
+    ///
+    func deleteSystemPlugins(siteID: Int64) {
```
Ecarrion: I think this function will be insufficient when you get to use it.
We try to only delete objects that are no longer necessary, if you look at the function above, you can see how we only delete stale entities.

⚠️ Do you think we can update this to follow that pattern?

Ecarrion (changes requested):
Hi @fernandofgfer. This PR looks solid! I have a couple of requests marked with ⚠️ and questions.

Also, could you update the data model so properties are not marked as optional?

change


PS: I think you are missing some hound warnings too.

Storage/Storage/Tools/StorageType+Extensions.swift
fernandofgfer: I think is a well-readable way to do it! I like it

Storage/Storage/Tools/StorageType+Extensions.swift
fernandofgfer: Nope. I did it because maybe in the future we need to list it in the correct order, but I can remove it since we don't need it yet.

Storage/Storage.xcodeproj/project.pbxproj
fernandofgfer: I don't know why appeared these modifications, but I've reverted them.

Storage/Storage/Tools/StorageType+Deletions.swift
fernandofgfer: Ok, it's a personal way to separate concerts, but if you won't use it here, I will remove it.

Storage/Storage/Tools/StorageType+Deletions.swift
fernandofgfer: Sure! Modified!

fernandofgfer:
Hi @Ecarrion!
I've updated the changes requested!

Storage/Storage/Tools/StorageType+Extensions.swift
Ecarrion: We can leave it, I was just curious about the intention. 
FYI we don't have to show this list anywhere for now...

Ecarrion (approved):
This looks great! 🎉 
I think you may need to merge develop to get this merged tho!

PR #4618, opened 2021-07-14
System plugins - Adding new model into Networking layer

fernandofgfer:
#4604 - PR 2/2 - part from #4278 
## Description
This issue consists in add the new System plugin model to the Network layer.

## How
Tasks done in Network layer:
- Add new model
- Generate copiable

I've added also a Fake extension.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
+++ b/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
@@ -67,7 +67,7 @@ extension Order {
         let customerNote = customerNote ?? self.customerNote
         let dateCreated = dateCreated ?? self.dateCreated
         let dateModified = dateModified ?? self.dateModified
-        let datePaid = datePaid ?? self.datePaid
+        let datePaid = datePaid ?? nil
         let discountTotal = discountTotal ?? self.discountTotal
         let discountTax = discountTax ?? self.discountTax
         let shippingTotal = shippingTotal ?? self.shippingTotal
--- a/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
+++ b/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
@@ -67,7 +67,7 @@ extension Order {
         let customerNote = customerNote ?? self.customerNote
         let dateCreated = dateCreated ?? self.dateCreated
         let dateModified = dateModified ?? self.dateModified
-        let datePaid = datePaid ?? nil
+        let datePaid = datePaid ?? self.datePaid
         let discountTotal = discountTotal ?? self.discountTotal
         let discountTax = discountTax ?? self.discountTax
         let shippingTotal = shippingTotal ?? self.shippingTotal
@@ -682,36 +682,6 @@ extension ProductVariation {
     }
 }
 
-extension ShipmentTracking {
-    public func copy(
-        siteID: CopiableProp<Int64> = .copy,
-        orderID: CopiableProp<Int64> = .copy,
-        trackingID: CopiableProp<String> = .copy,
-        trackingNumber: CopiableProp<String> = .copy,
-        trackingProvider: NullableCopiableProp<String> = .copy,
-        trackingURL: NullableCopiableProp<String> = .copy,
-        dateShipped: NullableCopiableProp<Date> = .copy
-    ) -> ShipmentTracking {
-        let siteID = siteID ?? self.siteID
-        let orderID = orderID ?? self.orderID
-        let trackingID = trackingID ?? self.trackingID
-        let trackingNumber = trackingNumber ?? self.trackingNumber
-        let trackingProvider = trackingProvider ?? self.trackingProvider
-        let trackingURL = trackingURL ?? self.trackingURL
-        let dateShipped = dateShipped ?? self.dateShipped
-
-        return ShipmentTracking(
-            siteID: siteID,
-            orderID: orderID,
-            trackingID: trackingID,
-            trackingNumber: trackingNumber,
-            trackingProvider: trackingProvider,
-            trackingURL: trackingURL,
-            dateShipped: dateShipped
-        )
-    }
-}
-
 extension Refund {
     public func copy(
         refundID: CopiableProp<Int64> = .copy,
@@ -750,6 +720,36 @@ extension Refund {
             createAutomated: createAutomated,
             items: items,
             shippingLines: shippingLines
+        )
+    }
+}
+
+extension ShipmentTracking {
+    public func copy(
+        siteID: CopiableProp<Int64> = .copy,
+        orderID: CopiableProp<Int64> = .copy,
+        trackingID: CopiableProp<String> = .copy,
+        trackingNumber: CopiableProp<String> = .copy,
+        trackingProvider: NullableCopiableProp<String> = .copy,
+        trackingURL: NullableCopiableProp<String> = .copy,
+        dateShipped: NullableCopiableProp<Date> = .copy
+    ) -> ShipmentTracking {
+        let siteID = siteID ?? self.siteID
+        let orderID = orderID ?? self.orderID
+        let trackingID = trackingID ?? self.trackingID
+        let trackingNumber = trackingNumber ?? self.trackingNumber
+        let trackingProvider = trackingProvider ?? self.trackingProvider
+        let trackingURL = trackingURL ?? self.trackingURL
+        let dateShipped = dateShipped ?? self.dateShipped
+
+        return ShipmentTracking(
+            siteID: siteID,
+            orderID: orderID,
+            trackingID: trackingID,
+            trackingNumber: trackingNumber,
+            trackingProvider: trackingProvider,
+            trackingURL: trackingURL,
+            dateShipped: dateShipped
         )
     }
 }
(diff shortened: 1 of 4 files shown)


Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
```diff
@@ -67,7 +67,7 @@ extension Order {
         let customerNote = customerNote ?? self.customerNote
         let dateCreated = dateCreated ?? self.dateCreated
         let dateModified = dateModified ?? self.dateModified
-        let datePaid = datePaid ?? nil
+        let datePaid = datePaid ?? self.datePaid
```
Ecarrion: Hi @ctarda I see that this was changed manually in https://github.com/woocommerce/woocommerce-ios/commit/09bfa87754ec33eb83d78cfdf39396342d085e71

However, it will get updated every time someone runs `rake generate` 
Will this break something in CPP?

Networking/Networking/Model/SystemPlugin.swift
```diff
@@ -0,0 +1,91 @@
...
+public struct SystemPlugin: Decodable, GeneratedFakeable, GeneratedCopiable {
+    public let siteID: Int64
+    public let plugin: String           // e.g. woocommerce/woocommerce.php
+    public let name: String             // e.g. WooCommerce
+    public let version: String          // e.g. 3.0.0-rc.1
+    public let versionLatest: String    // e.g. 2.6.14
+    public let url: String              // e.g. https://woocommerce.com/
+    public let authorName: String       // e.g. Automattic
+    public let authorUrl: String        // e.g. https://woocommerce.com
+    public let networkActivated: Bool   // i.g false | true
```
Ecarrion: I think it will be better if we add these comments as proper documentation so it is shown when using the autocompletion tool.

Networking/Networking/Model/SystemPlugin.swift
```diff
@@ -0,0 +1,91 @@
+import Foundation
+import Codegen
+
+public struct SystemPlugin: Decodable, GeneratedFakeable, GeneratedCopiable {
```
Ecarrion, line 4: What are your opinions on `GeneratedFakeable` and `GeneratedCopiable`?

Networking/Networking/Model/SystemPlugin.swift
```diff
@@ -0,0 +1,91 @@
...
+        case url                 = "url"
+        case authorName          = "author_name"
+        case authorUrl           = "author_url"
+        case networkActivated    = "network_activated"
+    }
+
+    enum DescriptionCodingKeys: String, CodingKey {
+        case raw      = "raw"
+        case rendered = "rendered"
+    }
```
Ecarrion: What are these? I don't see them referenced anywhere.

Networking/Networking/Model/SystemPlugin.swift
```diff
@@ -0,0 +1,91 @@
...
+
+    enum CodingKeys: String, CodingKey {
+        case plugin              = "plugin"
+        case name                = "name"
+        case version             = "version"
+        case versionLatest       = "version_latest"
+        case url                 = "url"
+        case authorName          = "author_name"
+        case authorUrl           = "author_url"
+        case networkActivated    = "network_activated"
```
Ecarrion: I don't think we need to have explicit values when the case represents the exact string. We should only map the camel-case ones.

Also, let's use Xcode default indentation for those!

Networking/Networking/Model/SystemPlugin.swift
```diff
@@ -0,0 +1,91 @@
...
+                  plugin: plugin,
+                  name: name,
+                  version: version,
+                  versionLatest: versionLatest,
+                  url: url,
+                  authorName: authorName,
+                  authorUrl: authorUrl,
+                  networkActivated: networkActivated)
+    }
+
```
Ecarrion: nit: Extra white space

Networking/Networking/Model/SystemPlugin.swift
```diff
@@ -0,0 +1,91 @@
...
+        }
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let plugin = try container.decode(String.self, forKey: .plugin)
+        let name = try container.decodeIfPresent(String.self, forKey: .name) ?? ""
+        let version = try container.decodeIfPresent(String.self, forKey: .version) ?? ""
+        let versionLatest = try container.decodeIfPresent(String.self, forKey: .versionLatest) ?? ""
+        let url = try container.decodeIfPresent(String.self, forKey: .url) ?? ""
+        let authorName = try container.decodeIfPresent(String.self, forKey: .authorName) ?? ""
+        let authorUrl = try container.decodeIfPresent(String.self, forKey: .authorUrl) ?? ""
+        let networkActivated = try container.decodeIfPresent(Bool.self, forKey: .networkActivated) ?? false
```
Ecarrion: Why are you using `decodeIfPressent` here? Have you found scenarios where those fields are optional?

Ecarrion (changes requested):
Thanks, @fernandofgfer for working on this.

I've left you some small suggestions and let's wait for @ctarda's response to make sure we are not breaking anything.

Networking/Networking/Model/SystemPlugin.swift
fernandofgfer: Good catch! This has more sense.

Networking/Networking/Model/SystemPlugin.swift
fernandofgfer: I think both are useful! With Fakeable we win some time doing mocks and with Copiable you can generate a copy of your model avoiding modifications, so it's very secure! I like both of them

Networking/Networking/Model/SystemPlugin.swift
fernandofgfer: Sorry, I forgot to remove it!

Networking/Networking/Model/SystemPlugin.swift
fernandofgfer: Sure, I will remove unnecessary values!

Networking/Networking/Model/SystemPlugin.swift
fernandofgfer: 👍

Networking/Networking/Model/SystemPlugin.swift
fernandofgfer: Well, since I don't know all the scenarios, I decided to use `decodeIfPressent` to avoid any kind of crash here. In the end, I don't know if the service could return a semi-empty object.
If you know that we won't have this scenario, I will change it!

Ecarrion: Using `decode` won't crash because this function `throws`, so the error should be captured and logged in the consumer layer.

As those fields should always exist, I think it's best to guarantee the entity integrity and provide only the ones that we consider complete and valid.

Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
Ecarrion: > Will this break something in CPP?

I guess so because some tests fail now. 😓 

```swift

test_collect_payment_button_is_visible_and_primary_style_if_order_is_eligible_for_cash_on_delivery()

test_collect_payment_button_is_visible_if_order_is_eligible_for_cash_on_delivery_and_total_amount_is_greater_than_zero()
```

fernandofgfer: Maybe we can put `nil` there and remove the protocol `GeneratedCopiable` from the `Order` model, as a first solution. What do you think?

Networking/Networking/Model/SystemPlugin.swift
fernandofgfer: You are right. I've changed it and now looks better.

Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
ctarda: 👋🏼  I completely missed the fact that it would be overwritten. 

The reason to update it was to be able to apply the lense to the datePaid property of a mock order. If the default value is nil, this would not set the datePaid to an actual nil value:

```
        let order = makeOrder().copy(datePaid: nil)
```

I think the long term solution would be to modify the way that `Models+Copiable.generated.swift` is generated so that the datePaid can actually be nullified, but in the meantime, it is okay to disable those two tests. 

If this seems like a reasonable approach, would you mind logging a ticket for this in the [payments board](https://github.com/woocommerce/woocommerce-ios/projects/30) please, so that we can get to it next week?

Ecarrion: Oh, I understand the use case now!

To nullify a field we wan use 
```swift
 let order = makeOrder().copy(datePaid: .some(nil))
```

The regular `.copy(datePaid: .some(nil))` is used to leave a value unmodified.

@fernandofgfer would you mind creating a separate PR fixing those tests, so we can merge this PR?

fernandofgfer: Hi! I've reverted this change in this PR, in order to be merged. I will create a new PR to try to fix this issue!

fernandofgfer: Hi again! You have the PR that fixes these tests here: https://github.com/woocommerce/woocommerce-ios/pull/4634!

Ecarrion (approved):

PR #4493, opened 2021-06-29
Auth Role Checks: Part 4 - Implement user-facing flow changes

dvdchr:
Resolves #3963 
Depends on #4492 and #4665

## Summary

This is the **final** part out of 4 parts that make up the changes related to role authentication (for more details, refer to #3963). This PR introduces **user-facing changes** as it modifies the application flow. Here's a summary of this change: 

### 1. Check role eligibility after selecting a store

With this change, a blocking role eligibility check is performed after user taps "Continue" in Store Picker. If the role eligibility check fails, the user will be presented with a role error page. To see how the error page looks, refer to #4478.

At this moment, there are 3 ways to present a Store Picker: from the Authentication flow, from app launch, or from the Settings page. Here's how the role error page behaves on each of the behaviors:
- **From the authentication flow** 🔒 : The role error page will be directly pushed onto the navigation stack. This is possible because the authentication flow directly pushes Store Picker onto the navigation controller.
- **From app launch** 🚀 : The role error page will be directly pushed onto the _top-most_ navigation controller. In this flow, the Store Picker is wrapped in a navigation controller and then modally presented from a navigation controller. This is why we need to find the top-most navigation controller, and present it from there.
- **From the Settings page** ⚙️ : Same as previous, the role error page will be directly pushed onto the top-most navigation controller. The Store Picker is also presented modally, but the difference is that the Settings page presents it using default presentation style (i.e. it can be dismissed).

In all of the cases, the role eligibility check is performed _before_ the selected `storeID` is updated, therefore user can freely go back from the error page to the Store Picker page to (for example) try accessing a different store. 

Additionally, in the case where the user selected a store and were presented with the role error page, if they restart the app, they would enter the Store Picker again (through the app launch 🚀  flow) – giving them a chance to select a different store.

### 2. Check role eligibility on every app launch

This change also introduces a non-blocking role eligibility check during app launch. The result of the check is stored in `GeneralAppSettings`, and will be consumed by `AppCoordinator` on the next launch, should the result differ from the current state.

The role check is only performed for users that have **authenticated AND selected a store**. The purpose of doing this on every app launch is to detect when the user's roles are updated _after_ they gained access to the store (in the WooCommerce app).

Different from the Store Picker behavior, the error page presented from the app launch flow ⚠️ **may block the user** ⚠️. This can only happen under these conditions:
- User is authenticated,
- Has selected a store,
- User was previously eligible, but the latest role eligibility check failed.

[...]

PR #4619, opened 2021-07-14
Auth Role Checks: Part 2.5 - Add accessibility support for role error screen

dvdchr:
Refs #4478, #3963 
Depended by #4492, #4493

## Summary
Based on this [code review](https://github.com/woocommerce/woocommerce-ios/pull/4493#pullrequestreview-697643887) in #4493, there were problems with large text sizes. This PR provides accessibility support for large text. In a gist:

- The content view is nested in a scroll view, so the contents are vertically scrollable when using large text sizes.
- The link button does not have multi-line support by default, causing the UIButton to not respect the adjusted intrinsic content size. This PR addresses that.
- Since the action buttons are vertically stacked _and_ the secondary button title is rather long, I've applied a maximum font size for the buttons to allow for readable space for the content. Note that **this is only applied in iPhone interface idioms**.
- The `class` to `AnyObject` change in `RoleErrorOutput` is to get rid of an Xcode warning.

## 👀 Screenshots

• | Portrait | Landscape
--|--|--
iPhone, normal size | 4478_a11y_reg_iphone_portrait | 4478_a11y_reg_iphone_landscape
iPhone, large size | 4478_a11y_xl_iphone_portrait | 4478_a11y_xl_iphone_landscape
iPad, normal size | 4478_a11y_reg_ipad_portrait | 4478_a11y_reg_ipad_landscape
iPad, large size | 4478_a11y_ipad_portrait | 4478_a11y_ipad_landscape

Note: for dark mode looks, please refer to #4478.

## To Test
Currently, there's no path present to reach the new role error page. For the sake of just testing the layout, you'd have to intercept with the navigation flow temporarily. Modify `AppCoordinator` like so:

```swift
// AppCoordinator.swift:L38

// add these lines to always show the role error page.
let viewModel = RoleErrorViewModel(displayName: "John Doe", roles: ["editor"])
let viewController = RoleErrorViewController(viewModel: viewModel)
let navigationController = WooNavigationController(rootViewController: viewController)
self.setWindowRootViewControllerAndAnimateIfNeeded(navigationController)
self.isLoggedIn = isLoggedIn
return
```
When running the app again, you should see the role error page pop up first thing.

## Author Checklist
Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
+++ b/WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
@@ -80,6 +80,7 @@ class RoleErrorViewController: UIViewController {
 
     func configureLinkButton() {
         linkButton.applyLinkButtonStyle(enableMultipleLines: true)
+        linkButton.titleLabel?.textAlignment = .center
         linkButton.setTitle(viewModel.auxiliaryButtonTitle, for: .normal)
         linkButton.on(.touchUpInside) { [weak self] _ in
             self?.viewModel.didTapAuxiliaryButton()
@@ -88,14 +89,6 @@ class RoleErrorViewController: UIViewController {
 
     func configurePrimaryActionButton() {
         primaryActionButton.applyPrimaryButtonStyle()
-
-        // Since the buttons are vertically stacked, override the font
-        // on iPhone interface idioms so we get a maximum font size to give
-        // space for the content above the button container.
-        if WPDeviceIdentification.isiPhone() {
-            primaryActionButton.titleLabel?.font = StyleManager.headlineSemiBold
-        }
-
         primaryActionButton.setTitle(viewModel.primaryButtonTitle, for: .normal)
         primaryActionButton.on(.touchUpInside) { [weak self] _ in
             self?.viewModel.didTapPrimaryButton()
@@ -104,14 +97,6 @@ class RoleErrorViewController: UIViewController {
 
     func configureSecondaryActionButton() {
         secondaryActionButton.applySecondaryButtonStyle()
-
-        // Since the buttons are vertically stacked, override the font
-        // on iPhone interface idioms so we get a maximum font size to give
-        // space for the content above the button container.
-        if WPDeviceIdentification.isiPhone() {
-            secondaryActionButton.titleLabel?.font = StyleManager.headlineSemiBold
-        }
-
         secondaryActionButton.setTitle(viewModel.secondaryButtonTitle, for: .normal)
         secondaryActionButton.on(.touchUpInside) { [weak self] _ in
             self?.viewModel.didTapSecondaryButton()
--- a/WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
+++ b/WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
@@ -79,16 +79,7 @@ class RoleErrorViewController: UIViewController {
     }
 
     func configureLinkButton() {
-        linkButton.applyLinkButtonStyle()
-
-        // enable multi line for accessibility support
-        linkButton.titleLabel?.textAlignment = .center
-        linkButton.titleLabel?.adjustsFontForContentSizeCategory = true
-        linkButton.titleLabel?.lineBreakMode = .byWordWrapping
-        if let linkTitleLabel = linkButton.titleLabel {
-            linkButton.pinSubviewToAllEdgeMargins(linkTitleLabel)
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
```diff
@@ -80,6 +80,15 @@ class RoleErrorViewController: UIViewController {
...
     func configureLinkButton() {
         linkButton.applyLinkButtonStyle()
+
+        // enable multi line for accessibility support
+        linkButton.titleLabel?.textAlignment = .center
+        linkButton.titleLabel?.adjustsFontForContentSizeCategory = true
+        linkButton.titleLabel?.lineBreakMode = .byWordWrapping
+        if let linkTitleLabel = linkButton.titleLabel {
+            linkButton.pinSubviewToAllEdgeMargins(linkTitleLabel)
+        }
```
Ecarrion: We have a private method `enableMultipleLines()` method on `UIButton+Helper` to help with these situations.

Maybe it's worth mixing it with the `applyLinkButtonStyle` to avoid duplicating that logic.

Something like
```swift
private func applyLinkButtonStyle(enableMultipleLine: Bool = false) {
    // code
}
```

WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
```diff
@@ -88,6 +97,14 @@ class RoleErrorViewController: UIViewController {
 
     func configurePrimaryActionButton() {
         primaryActionButton.applyPrimaryButtonStyle()
+
+        // Since the buttons are vertically stacked, override the font
+        // on iPhone interface idioms so we get a maximum font size to give
+        // space for the content above the button container.
+        if WPDeviceIdentification.isiPhone() {
+            primaryActionButton.titleLabel?.font = StyleManager.headlineSemiBold
+        }
```
Ecarrion: I'm not sure if this is a good alternative since if users use a gigantic font size is because they need it 😅 

Look the feedback completion screen is handling this scenario.(I modified it to have a multiple line button)

Screen Shot 2021-07-15 at 4 23 02 PM

Ecarrion (commented):
Hi @dvdchr, thanks for tackling this! I have a few suggestions, let me know what you think?

WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
dvdchr: @Ecarrion 

Ah, I see. Thanks for the input! I'd like to clarify some assumptions that I've had, that led me to design it this way:
1. With increased text sizes, the Help menu is always visible (since it's located on the navigation bar), while the action buttons would be hidden below-the-fold. From the user experience perspective, **I assumed that preserving the action buttons above-the-fold would help users be more aware of what actions they could take**. However, in iPhone landscape mode, the buttons are taking too much of the viewport when the text size are allowed to be increased naturally.
2. The font style with maximum font size is available to use in `StyleManager`. I assumed that everything there is a11y-friendly (at least to some acceptable degree). This raises a question though: does this mean that `StyleManager.headlineSemiBold` (or any font presets in `StyleManager` that have max font size) should not be used? 🤔 
3. Taking into account the interfaces around iOS when maximum a11y text size is selected, I noticed that not every interface component is inflated to a gigantic font size. For example: navigation bar buttons. This led to think that it's _okay_ to use a smaller size, as long as it's not the smallest font size displayed within that accessibility setting.

Also, thank you for sharing your reference! 👍 With the latest commit (https://github.com/woocommerce/woocommerce-ios/pull/4619/commits/132cb06b291d836019b0792f226cc4d1e89b668f), I've decided to follow your reference and modified the layout to be entirely scrollable, and the max font size on the action buttons are removed. Here's how it looks now:

Portrait | Landscape
--|--
Simulator Screen Shot - iPhone 8 - 2021-07-17 at 22 38 17 | Simulator Screen Shot - iPhone 8 - 2021-07-17 at 22 20 10

Let me know what you think!

WooCommerce/Classes/Authentication/Navigation Exceptions/RoleErrorViewController.swift
dvdchr: Sure! Done in https://github.com/woocommerce/woocommerce-ios/pull/4619/commits/7a58d655a700491267f3b2309cd596ed0bb62dbd 👍

Ecarrion (approved):
This looks great @dvdchr.
Thanks for the changes!

PR #4624, opened 2021-07-15
[CPP Onboarding] Set padding for generic error to 24 pixels

koke:
Part of #4611 
As noted in https://github.com/woocommerce/woocommerce-ios/pull/4613#issuecomment-880204405, we should use 24px of horizontal padding instead of the default SwiftUI padding.

## To test

Go to Settings > In-Person Payments on a store without payments set up

Before|After
-|-
before|after

Update release notes:


Garance91540 (approved):
👏

Issue #4449, opened 2021-06-18
[Mobile Payments] Replace table view in Connect to Card Reader screen with a simpler solution


koke:
The remaining bits in #4361 seem hard to achieve while using a table view for layout:
- The height of the button is controlled by cell size, which is calculated automatically
- Padding between the button and the text below is hard to specify explicitly
- I don't see how to bottom-align the button and text while they are table view cells

Replacing the view with either SwiftUI or UIStackView will make these improvements easier, and will likely simplify the code overall.

joshheald:
Relates to #4548 
Should probably be done together...

ctarda:
I am going to close this one for now. We can reopen later if needed

Issue #4546, opened 2021-07-05
Payments: There is a Brief Flash of the Disconnect Button When Connecting a Reader


shiki:
When connecting a reader from the Order Details screen, I notice a brief flash of the _Disconnect_ screen. You can see it between the 5 to 7 second mark of this video:

https://user-images.githubusercontent.com/198826/124522514-c429fa00-ddb0-11eb-96e4-b2bc6197b86f.mp4

koke:
Confirmed. I've been seeing this one, and while it's not a critical issue, it looks poorly.

allendav:
@ctarda @Garance91540  @adamzelinski - I just moved this to the Now column in the project

Right now, if the merchant begins the collect payment flow from within the order and they are not yet connected to their reader, we push the "Connect your card reader" view onto the stack and _then_ add the "Scanning for readers" modal on top of that. As the merchant completes connecting to their reader,  the "Connect your card reader" view is briefly replaced with the Connected Reader view (with the disconnect button). This is very busy and distracting:

![disco3](https://user-images.githubusercontent.com/1595739/126568079-9467bbe4-f2d5-4b10-b521-995fc395097e.gif)

To fix this, I'm going to remove the dependency on that intermediate view and allow the connection modals to appear directly on top of the order details view.  This will eliminate the "Connect your card reader" view and the "Connected reader/disconnect" views appearing in the context of the order.  It will look more like this:

<img width="1561" alt="new" src="https://user-images.githubusercontent.com/1595739/126568844-b9e4cf22-697e-4345-a65f-4f49776c540b.png">

@ctarda @koke - here's the changes needed - I see a short series of PRs

- PR 1 - Add CardReaderConnectionController, repurposing reader discovery state and code from our existing settings view models and change OrderDetailsViewController::connectToCardReader to use it if needs to connect to a reader
- PR 2 - Change CardReaderSettingsUnknownViewController to use CardReaderConnectionController to connect to a reader and get rid of updateModal and delete reader discovery state and code from CardReaderSettingsUnknownViewModel
- PR 3 - Change CardReaderSettingsKnownViewController to use CardReaderConnectionController to connect to a reader and get rid of updateModal and delete reader discovery state and code from CardReaderSettingsKnownViewModel

CardReaderConnectionController will be invoked by ViewControllers that need it using something like the following:

```
let readerController = CardReaderConnectionController(self)
connectionController.discoverAndConnect() {
   ...
}
```

where

```
[...]

PR #4626, opened 2021-07-15
[Proof-of-concept] Card Reader Settings in SwiftUI

koke:
Part of #4449

This PR isn't finished and will likely need to be split up into smaller PRs before merging, but I'm sharing the progress so far in rewriting Card Reader Settings in SwiftUI

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
```diff
@@ -40,7 +40,11 @@ final class CardReaderSettingsPresentingViewController: UIViewController {
             return
         }
 
-        childViewController = self.storyboard!.instantiateViewController(withIdentifier: viewModelAndView.viewIdentifier)
+        if viewModelAndView.viewIdentifier == "CardReaderSettingsUnknownViewController" {
```
allendav, line 43: If we could avoid having to change CardReaderSettingsPresentingViewController each time we add a new view to the stack that would be awesome. I assume you needed to do this just for the sake of this POC?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownView.swift
```diff
@@ -0,0 +1,157 @@
...
+    @objc required dynamic init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+struct CardReaderSettingsUnknownView: View {
+    @State var viewModel: CardReaderSettingsUnknownViewModel
+    @State var presentedURL: URL? = nil
+
+    var body: some View {
```
allendav, line 26: I'd welcome programmatic definition of views (over storyboard). So much easier to understand diffs. I am curious if @ctarda concurs. I've used both approaches professionally. The big advantage of storyboards (IMHO) is designers can develop components, although that is not our approach here.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AttributedText.swift
```diff
@@ -0,0 +1,150 @@
+import SwiftUI
```
allendav, line 1: This reminded me - attributed text is coming soon - we could use this approach until we drop iOS14 support sometime in 2021/2022 - https://betterprogramming.pub/ios-15-attributed-strings-in-swiftui-markdown-271204bec5c1 https://developer.apple.com/documentation/foundation/attributescopes/swiftuiattributes

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownView.swift
```diff
@@ -0,0 +1,157 @@
...
+                    Text(Localization.connectButton)
+                        .frame(maxWidth: .infinity)
+                }
+                .buttonStyle(PrimaryButtonStyle())
+
+                HStack(alignment: .center, spacing: 20) {
+                    Image(uiImage: .infoOutlineImage)
+                        .accentColor(Color(.lightGray))
+                        .frame(width: 20, height: 20)
+                    AttributedText(Localization.learnMore)
```
allendav, line 51: See comment below about support coming in iOS15. Maybe we could negotiate with design to avoid having to do this (i.e. by making the entire text a link)

allendav:
Thanks for putting this together! This was fun to review and test.




I'm game to switch to SwiftUI but I'm not sure rewriting existing views should be the highest priority right now, especially since I am probably about to do this: https://github.com/woocommerce/woocommerce-ios/issues/4546#issuecomment-884548812 which will simplify these views greatly.

So... perhaps instead of rewriting those existing views right now, if new views could be added using SwiftUI (using this PR as a guide perhaps) (e.g. new views for "Your country isn't supported" or "You need to activate the plugin" or ...) that might be a good approach, provided we can avoid changing CardReaderSettingsPresentingViewController each time we add a new view to the stack

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsPresentingViewController.swift
koke: If the whole view hierarchy was converted to SwiftUI, it would be a lot simpler to switch between child views for the different states. For this PoC though, I wanted to focus on replacing this one view while minimizing the changes required elsewhere (the PR already got big enough with that limitation).

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsUnknownView.swift
koke: > The big advantage of storyboards (IMHO) is designers can develop components

In my experience, it's been very rare that we were able to build a component using just what Storyboards offer. It might allow for easy tweaking of the visual properties once its build, but Xcode also offers some design tools to create and modify SwiftUI hierarchies without writing code.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AttributedText.swift
koke: There was some discussion about this in paNNhX-fz-p2. That would be ideal, but it will take a while for us to be able to use that solution. I considered making the whole text a link, but there are some upcoming designs for onboarding with longer paragraphs where I don't think that behavior would feel right:

Screen Shot 2021-07-27 at 13 47 39

koke:
> if new views could be added using SwiftUI

I don't think rewriting these is strictly necessary, but it makes navigation between screens a bit more complicated, at least temporarily. The final designs have reader connection in a sub-menu alongside ordering a reader, so there's no need to switch child views, but until we implement that, I was going to skip the extra menu and present the card reader connection screen directly when onboarding is complete.

However, I can start there and see how it goes. And if the friction becomes too much of a problem we can re-prioritize. I'm going to close this for now, and add the necessary bits from here alongside the onboarding PRs

PR #4628, opened 2021-07-15
Fix: updated the top banner in Order Detail screen informing the user of the shipping label creation

### LINKED ISSUE #4625: Shipping Labels: update the top banner in order detail informing the user of the shipping label creation
@anitaa1990 tested the Shipping Labels flow.

The Shipping Labels banner is only displayed for orders that have shipping labels. But since we support purchasing labels now, we should display the banner for all orders that are eligible to create shipping labels (So if the Create Shipping Label button is displayed, the banner should also be displayed.

The text of the banner should be changed to reflect that we can now purchase shipping labels.

| iOS             |  Android |
img_7944 | screenshot_20210715-095832

pmusolino:
Fixes #4625 

## Description
Updated the top banner in Order Detail screen, informing the user of the Shipping Label Creation.
This PR will be merged in release 7.1, since we need to inform users about this new beta feature.

#### Case 1
1. Open an order that is eligible for Shipping Label creation.
2. You should see the new top banner appearing with the new text.

#### Case 2
1. Open an order with existing shipping labels.
2. You shouldn't see the top banner.

## Screenshots
| Before (Order with existing Shipping Labels)            |  After (Order eligible for Shipping Label creation) |
Simulator Screen Shot - iPhone 12 Pro - 2021-07-15 at 11 15 27 | Simulator Screen Shot - iPhone 12 Pro - 2021-07-15 at 11 44 43

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -83,12 +83,12 @@ extension WooConstants {
         ///
         case productsFeedback = "https://automattic.survey.fm/woo-app-feature-feedback-products"
 
-        /// URL for the shipping labels M1 feedback survey
+        /// URL for the shipping labels M3 feedback survey
         ///
 #if DEBUG
-        case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-testing-feature-feedback-shipping-labels"
+        case shippingLabelsRelease3Feedback = "https://automattic.survey.fm/woo-app-testing-feature-feedback-shipping-labels"
 #else
-        case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-shipping-labels"
+        case shippingLabelsRelease3Feedback = "https://automattic.survey.fm/woo-app-feature-feedback-shipping-labels"
 #endif
 
         /// URL for the order add-on i1 feedback survey
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelsTopBannerFactory.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelsTopBannerFactory.swift
@@ -4,13 +4,16 @@ import Yosemite
 /// The top banner has two actions: "give feedback" and "dismiss".
 /// This class is not meant to be retained since it uses strong `self` in action handling, so that it has the same life cycle as the top banner.
 final class ShippingLabelsTopBannerFactory {
+    private var isEligibleForShippingLabelCreation: Bool
     private let shippingLabels: [ShippingLabel]
     private let stores: StoresManager
     private let analytics: Analytics
 
-    init(shippingLabels: [ShippingLabel],
+    init(isEligibleForShippingLabelCreation: Bool,
+         shippingLabels: [ShippingLabel],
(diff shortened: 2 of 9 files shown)


pmusolino:
I'm not sure how handled the release 7.1, so i'm tagging you both 🙇 @AliSoftware @jkmassel

WooCommerce/Classes/System/WooConstants.swift
```diff
@@ -83,12 +83,12 @@ extension WooConstants {
         ///
         case productsFeedback = "https://automattic.survey.fm/woo-app-feature-feedback-products"
 
-        /// URL for the shipping labels M1 feedback survey
+        /// URL for the shipping labels M3 feedback survey
         ///
 #if DEBUG
-        case shippingLabelsRelease1Feedback = "https://automattic.survey.fm/woo-app-testing-feature-feedback-shipping-labels"
+        case shippingLabelsRelease3Feedback = "https://automattic.survey.fm/woo-app-testing-feature-feedback-shipping-labels"
```
rachelmcr, line 89: Just checking, is it ok that we are linking to the same survey that we used for M1 features? Or do we need to be able to distinguish between people on older app versions sending in feedback about M1 features vs. M3 features? (I figure it's probably ok but wanted to point out that all the feedback will go to the same place.)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/ShippingLabelsTopBannerFactory.swift
```diff
@@ -97,13 +100,12 @@ private extension ShippingLabelsTopBannerFactory {
...
-            NSLocalizedString("Print shipping labels from your device!",
+            NSLocalizedString("Create shipping labels from your device!",
                               comment: "The title of the shipping labels top banner in order details")
         static let info =
             NSLocalizedString(
-                "We are working on making it easier for you to print shipping labels directly from your device! "
-                    + "For now, if you have created labels for an order in your store admin with WooCommerce Shipping, "
-                    + "you can print them in your Order details here.",
+                "You can now create shipping labels for all physical orders directly from your device with the free WooCommerce Shipping plugin. "
+                    + "Tap on Create shipping label to try our beta feature!",
```
rachelmcr: nit: Maybe capitalize `Create Shipping Label` here to match the button.

rachelmcr (approved):
Code ✅  
Testing ✅ 

I tested an eligible order without shipping labels (banner appeared), an ineligible order without shipping labels (no banner), and an order with shipping labels (no banner) and it worked as expected. I just have a small nitpick about the capitalization in the banner text.

I did discover that if an order has a refunded shipping label, even if the remote eligibility endpoint says it's eligible, neither the banner nor the "Create Shipping Label" button appears. However, that's not limited to this change so it can be addressed separately; I opened an issue about it: https://github.com/woocommerce/woocommerce-ios/issues/4629

WooCommerce/Classes/System/WooConstants.swift
pmusolino: It's ok, because we send some query items in the URL, like the milestone https://github.com/woocommerce/woocommerce-ios/pull/4628/files#diff-5ae6bdf31bfc7ab2a1149875d68fb54156193b998854de8c4afb472b1bcfef85R87

pmusolino:
Thank you Rachel for the review! I fixed the capitalized text 👍

Issue #4629, opened 2021-07-15
Shipping Labels: Option to create a shipping label is missing from orders with refunded labels


rachelmcr:
**Describe the bug**

If an order has a refunded shipping label, it is still eligible for a new shipping label to be created. (This is the behavior seen in wp-admin.)

However, in the app, even though the remote eligibility endpoint says the order is eligible for shipping label creation, we don't show the "Create Shipping Label" button. I believe this is because we check for `shippingLabels.isEmpty` instead of `shippingLabels.nonRefunded.isEmpty` before displaying that button.

**To Reproduce**
1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more payment methods set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Make sure you have an order with a refunded shipping label. (You can purchase a shipping label and then immediately refund it.)
3. Launch the app.
4. Open the order details for the order with a refunded shipping label.

Notice that in wp-admin you see an option to create a new label, but that option is not available in the app.

**Screenshots**

Order in wp-admin | Order in app
-|-
![Screen Shot 2021-07-15 at 2 34 34 PM](https://user-images.githubusercontent.com/8658164/125797026-c23b4fbd-a0ab-4f64-ad07-3d391abaefb3.png)|![Simulator Screen Shot - iPhone 11 Pro - 2021-07-15 at 14 15 54](https://user-images.githubusercontent.com/8658164/125797027-cea961b5-8e95-461c-80a6-698e179b98bc.png)

rachelmcr:
Fix for the 7.1 release in https://github.com/woocommerce/woocommerce-ios/pull/4632.

PR #4632, opened 2021-07-16
Shipping Labels: Enable Create Shipping Label button on orders with only refunded labels

rachelmcr:
Fixes: #4629

## Description

If an order has a refunded shipping label, it is still eligible for a new shipping label to be created. (This is the behavior seen in wp-admin.)

However, in the app, even though the remote eligibility endpoint says the order is eligible for shipping label creation, we didn't show the "Create Shipping Label" button. This PR fixes that behavior so the "Create Shipping Label" button is shown if an order only has refunded shipping labels; it also makes sure the shipping label banner is displayed on those orders.

## Changes

* Updates the checks for displaying the "Create Shipping Label" button and shipping label banner to check `shippingLabels.nonRefunded.isEmpty` instead of `shippingLabels.isEmpty`.
* Updates the shipping label banner unit test for orders with a refunded label.
* Adds unit tests for the "Create Shipping Label" button visibility.

Before|After
-|-
Shipping Label before|Simulator Screen Shot - iPhone 11 Pro - 2021-07-15 at 16 33 16

## Testing

2. Make sure you have an order with a refunded shipping label. (You can purchase a shipping label and then immediately refund it.)
3. Launch the app.
4. Open the order details for the order with a refunded shipping label.
5. Confirm the shipping label banner appears at the top of the order detail screen, and the "Create Shipping Label" button is visible.
6. Tap the "Create Shipping Label" button and confirm you can successfully purchase a shipping label for that order.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -868,19 +868,19 @@ extension OrderDetailsDataSource {
 
             var rows: [Row] = Array(repeating: .aggregateOrderItem, count: aggregateOrderItemCount)
 
-            if isEligibleForShippingLabelCreation && shippingLabels.isEmpty {
+            if isEligibleForShippingLabelCreation && shippingLabels.nonRefunded.isEmpty {
                 rows.append(.shippingLabelCreateButton)
             }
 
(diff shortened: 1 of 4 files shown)


pmusolino (approved):
:shipit:  Looks good to me! 

Code ✅ 
Testing ✅

PR #4638, opened 2021-07-19
Edge-to-edge table views on Orders tab

itsmeichigo:
Part of #3716 

# Description
This PR aims to keep consistent edge-to-edge look on table views of Orders tab. 

Note: This PR only covers Orders and Order Details screens - Shipping Labels screens will be handled in a separate PR.

# Changes
- Updated tab view on Orders screen to be edge-to-edge - the tab scroll view is pinned to its super view's leading and trailing edges.
- Both Root Orders controller and Order List table view were also updated to constrained to super view's edges.
- Order Details table view and header view were both pinned to their super view's edges as well.

# Screenshots
| Before | After |
| ------ | ------ |
|  |  |

# Testing
1. Navigate to Orders tab
2. Switch to landscape mode
3. Notice that there's no leading or trailing margins on the Orders table view or tab view.
4. Select an order to see details.
5. Notice that there's no leading or trailing margins on the details table view order header view.

Update release notes:


ealeksandrov (approved):
LGTM :shipit:

PR #4601, opened 2021-07-13
Shipping Labels: Validate that address name field is not empty

rachelmcr:
Fixes: #4480

## Description

When you start to create a shipping label, we auto-fill the origin address name field with the first and last name from the user's account settings. If the account settings are not in storage, or don't include a first/last name, the name is left blank. This causes problems later on, because the name is required (e.g. to load carriers and rates for the package).

This PR adds local validation to check that the name field is not empty. If it is empty, it opens the Edit Address screen to prompt the user to enter a name. (We already have validation on the Edit Address screen to ensure the name field is filled out there.)

Note that we have related issues for remote name validation (https://github.com/Automattic/woocommerce-services/issues/2457) and making sure account settings are synced in storage after login (https://github.com/woocommerce/woocommerce-ios/issues/4600), which aren't part of this PR.

## Changes

* Adds a check for `address.name.isEmpty` before doing the remote address validation in `ShippingLabelFormViewModel`.
* Adds a unit test.

Before|After
-|-
validate-name-before|validate-name-after

## Testing

1. Make sure your `First name` and `Last name` are empty in your [WordPress.com account settings](https://wordpress.com/me).
2. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more payment methods set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
3. Launch the app.
4. If you are already logged in, log out and log in again (to make sure you don't have old account settings stored in the app).
5. Open an order detail that is eligible for the shipping label.
6. Tap the button "Create Shipping Label".
7. In the `Ship from` section, tap the "Continue" button.
8. Confirm that the Edit Address screen opens (with the "Name missing" validation error prompting you to enter a name).

## Submitter Checklist

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4644, opened 2021-07-19
Shipping Labels: Validate that address name field is not empty

rachelmcr:
## Description

This PR adds local validation to check that the shipping address name field is not empty when creating shipping labels. If it is empty, it opens the Edit Address screen to prompt the user to enter a name.

This was [fixed on `develop`](https://github.com/woocommerce/woocommerce-ios/pull/4601) and the fix is being introduced to `release/7.1` based on beta testing feedback. (Internal ref: p5T066-2uU-p2#comment-9242)

## Changes

* Adds a check for `address.name.isEmpty` before doing the remote address validation in `ShippingLabelFormViewModel`.
* Adds a unit test.

Before|After
-|-
validate-name-before|validate-name-after

## Testing

1. Make sure your `First name` and `Last name` are empty in your [WordPress.com account settings](https://wordpress.com/me).
2. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more payment methods set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
3. Launch the app.
4. If you are already logged in, log out and log in again (to make sure you don't have old account settings stored in the app).
5. Open an order detail that is eligible for the shipping label.
6. Tap the button "Create Shipping Label".
7. In the `Ship from` section, tap the "Continue" button.
8. Confirm that the Edit Address screen opens (with the "Name missing" validation error prompting you to enter a name).

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -498,6 +498,14 @@ extension ShippingLabelFormViewModel {
 
         let addressToBeVerified = ShippingLabelAddressVerification(address: address, type: type)
 
+        // Validate name field locally before validating the address remotely.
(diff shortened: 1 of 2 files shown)


pmusolino (approved):
:shipit: thanks for moving the name validation on release 7.1

Issue #4640, opened 2021-07-19
Shipping Labels: manage decimal quantities in Package Details


pmusolino:
As stated [here](https://a8c.slack.com/archives/C6H8C3G23/p1624620001117600), when the order has decimal quantities, we should `ceil` the quantity to the nearest bigger integer. Instead, we are `flooring` it.

As said by @hichamboushaba, it's about quantity, and it will affect the prefilled weight, but it's a side effect.
If an order has 0.5 as quantity, the list of products in the package should show 1 product, and if it's 3.5, it should show 4, which is done by ceiling the quantity to the nearest bigger integer.

<img width="464" alt="Screen Shot 2021-06-25 at 12 19 33" src="https://user-images.githubusercontent.com/495617/126131259-09382578-90d4-488a-bcc7-a1490ab0f579.png">

pmusolino:
Ref: p5T066-2uU#comment-9242

rachelmcr:
This is fixed in #4656 for the 7.1 release.

PR #4645, opened 2021-07-19
Shipping Labels: fixed total weight in Package Details

### LINKED ISSUE #4643: Shipping Labels: total package weight is wrong
Currently, the total package weight is the sum of just the products, and it does not consider the number of items of every product, so the total package weight will be wrong.

pmusolino:
Fixes #4643 

## Description
Previously, the total package weight in the Package Details screen under the Shipping Label Creation flow was the sum of the weight of the products, without considering the quantities. Now, the total weight is calculated based on items quantity.

## Testing
2. Launch the app.
6. In the package details section, make sure that the "Total package weight" displayed is now correct, and it's the sum of the weight of the item shown on the screen.

before | after

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -174,7 +174,9 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
                 product = products.first { $0.productID == item.productID }
             }
             if product?.virtual == false || productVariation?.virtual == false {
-                tempTotalWeight += Double(productVariation?.weight ?? product?.weight ?? "0") ?? 0
+                for _ in 0..<item.quantity.intValue {
+                    tempTotalWeight += Double(productVariation?.weight ?? product?.weight ?? "0") ?? 0
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -174,7 +174,9 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
                 product = products.first { $0.productID == item.productID }
             }
             if product?.virtual == false || productVariation?.virtual == false {
-                tempTotalWeight += Double(productVariation?.weight ?? product?.weight ?? "0") ?? 0
+                for _ in 0..<item.quantity.intValue {
```
rachelmcr, line 177: Clever! Did you use a `for` loop here to avoid having to convert `item.quantity.intValue` to a `Double` to multiple it with the product weight? Or for readability? (Just asking for my own curiosity / learning.)

rachelmcr (approved):
Code ✅ 
Testing ✅ 
:shipit: 

The only scenario I found where the weight isn't correct is when the product has a decimal quantity in the order, but that's a known issue due to how we're currently handling decimal quantities.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
pmusolino: I used this approach because we are using the `item.quantity.intValue` also for calculating the number of items rows, so this means that we will expect to have the same identical result here (which may not happen by multiplying a decimal number). Plus, we have in plan to change it in the near future because of this new issue: https://github.com/woocommerce/woocommerce-ios/issues/4640

rachelmcr: Ah, that makes sense! Thanks for explaining.

PR #4646, opened 2021-07-19
Issue/4605 system plugins update yosemite

fernandofgfer:
#4605 
## Summary
This PR contains the Yosemite updates of the System plugins project. 
In this branch, I've included the next classes
- SystemPluginRemote: To retrieve data from Network layer.
- SystemPluginMapper: To map this data.
- SystemPluginAction: To define both actions.
- SystemPluginStore: To execute these actions.

## Actions
I've added two actions:
- `synchronizeSystemPlugins`: This action retrieves all plugins and saves them into specific storage.
- `fetchSystemPlugins`: This action returns all plugins saved in storage.

## How to test it
I've added unit tests for SystemPluginRemote, SystemPluginMapper, and SystemPluginStore.
Execute all unit tests

Update release notes:
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4647, opened 2021-07-19
System Plugins - Update Yosemite layer - Add Remote and Mapper - 1/2

fernandofgfer:
#4646 1/2
## Summary
This PR has been splitter into two PRs.
This PR contains the Yosemite updates of the System plugins project.
In this branch, I've included the next classes
- SystemPluginRemote: To retrieve data from the Network layer.
- SystemPluginMapper: To map this data.

## How to test it
I've added unit tests for SystemPluginRemote and SystemPluginMapper
Execute all unit tests

Update release notes:


Networking/Networking/Remote/SystemPluginsRemote.swift
```diff
@@ -0,0 +1,30 @@
...
+    /// Retrieves all of the `SystemPlugin`s for a given site.
+    ///
+    /// - Parameters:
+    ///   - siteID: Site for which we'll fetch the system plugins.
+    ///   - completion: Closure to be executed upon completion.
+    ///
+    public func loadSystemPlugins(for siteID: Int64,
+                            completion: @escaping (Result<[SystemPlugin], Error>) -> Void) {
+        let path = Constants.systemPluginsPath
+        let request = JetpackRequest(wooApiVersion: .none, method: .get, siteID: siteID, path: path, parameters: nil)
```
Ecarrion: I think here we should use `.mark3` instead of `.none`.

Networking/Networking/Remote/SystemPluginsRemote.swift
```diff
@@ -0,0 +1,30 @@
...
+        enqueue(request, mapper: mapper, completion: completion)
+    }
+}
+
+
+// MARK: - Constants!
+//
+private extension SystemPluginsRemote {
+    enum Constants {
+        static let systemPluginsPath: String = "wc/v3/system_status"
```
Ecarrion: And here, we should drop the API version!

Networking/NetworkingTests/Mapper/SystemPluginsMapperTests.swift
```diff
@@ -0,0 +1,65 @@
...
+private extension SystemPluginsMapperTests {
+
+    /// Returns the SystemPluginsMapper output upon receiving `filename` (Data Encoded)
+    ///
+    func mapPlugins(from filename: String) -> [SystemPlugin] {
+        guard let response = Loader.contentsOf(filename) else {
+            return []
+        }
+
+        return try! SystemPluginsMapper(siteID: dummySiteID).map(response: response)
```
Ecarrion: Not super urgent, but probably a good idea to avoid `try!` and make the function throw!
This will prevent the test_suite to stop in case of any failure.

Networking/Networking/Remote/SystemPluginsRemote.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Retrieves all of the `SystemPlugin`s for a given site.
+    ///
+    /// - Parameters:
+    ///   - siteID: Site for which we'll fetch the system plugins.
+    ///   - completion: Closure to be executed upon completion.
+    ///
+    public func loadSystemPlugins(for siteID: Int64,
+                            completion: @escaping (Result<[SystemPlugin], Error>) -> Void) {
+        let path = Constants.systemPluginsPath
```
Ecarrion, line 15: Since the `system_status` resource contains a lot of information we don't need. I think it's worth to limit the endpoint to only return the "active_plugins" field.

You can take a look at how this is done here https://github.com/woocommerce/woocommerce-ios/blob/develop/Networking/Networking/Remote/ProductsRemote.swift#L232

Ecarrion (changes requested):
Good work @fernandofgfer, I left you some comments to further improve it.

Networking/Networking/Remote/SystemPluginsRemote.swift
fernandofgfer: Thanks for the tip!

Networking/Networking/Remote/SystemPluginsRemote.swift
fernandofgfer: Removed!

Networking/NetworkingTests/Mapper/SystemPluginsMapperTests.swift
fernandofgfer: Good catch. Refactored!

Networking/Networking/Remote/SystemPluginsRemote.swift
fernandofgfer: Wow, I didn't know that. Added!

fernandofgfer:
Thanks for reviewing it @Ecarrion! Requests added! 😄

Ecarrion (changes requested):
Hi @fernandofgfer, I tested the new remote against the real API. And I think the JSON you are testing against is not representative of the real JSON.

```swift
Mapping Error: keyNotFound(CodingKeys(stringValue: "active_plugins", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"active_plugins\", intValue: nil) (\"active_plugins\").", underlyingError: nil))
```


You can use WormHolly, to see how the JSON responses are retrieved in the app.

Screen Shot 2021-07-20 at 1 03 17 PM

fernandofgfer:
> Hi @fernandofgfer, I tested the new remote against the real API. And I think the JSON you are testing against is not representative of the real JSON.
> 
> ```swift
> Mapping Error: keyNotFound(CodingKeys(stringValue: "active_plugins", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"active_plugins\", intValue: nil) (\"active_plugins\").", underlyingError: nil))
> ```
> 
> You can use WormHolly, to see how the JSON responses are retrieved in the app.
> 
> Screen Shot 2021-07-20 at 1 03 17 PM

Hi @Ecarrion! You're right. I was following the example that appears here [WooCommerce Doc - System status](http://woocommerce.github.io/woocommerce-rest-api-docs/#list-all-system-status-items) and the "data" object doesn't appear. I've added and test it. It's working now.
Thanks for realizing it!

Ecarrion (approved):
👍👍

PR #4641, opened 2021-07-19
Drop custom equatable conformance

ealeksandrov:
Resolves https://github.com/woocommerce/woocommerce-ios/issues/3150.

## Description

Several of our Networking entities have a custom `Equatable` conformance that was implemented before the Swift introduced [synthesized conformance](https://developer.apple.com/documentation/swift/equatable).

The problem with the custom conformance is that is easy to forget to update it when adding new properties. To avoid that situation we should migrate to synthesized `Equatable` conformance when possible.

This PR reduces amount of custom `Equatable` implementations in the project from 61 to 22.

## Side notes

I found that many `Comparable` conformances are unnecessary and will submit separate PR removing some of them.

## Test

Just check CI status.

---
Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4650, opened 2021-07-19
Drop unnecessary Comparable conformance

ealeksandrov:
## Description

While working on https://github.com/woocommerce/woocommerce-ios/pull/4641 I've noticed that there are many unused `Comparable` implementations. This PR removes most of unused stuff.

Also there is a line in the docs that feels too strict to me and is removed in this PR too:
```
Model objects should implement `Comparable`.
```

Generally `Comparable` conformance in WC-iOS is used so we can ignore order for arrays of objects in tests. There is no real need for boilerplate in every model unless we want to sort it.

Related discussion ref: p91TBi-5Ea-p2

## Test

Check CI status.

---
Update release notes:

### DIFF
--- a/docs/NETWORKING.md
+++ b/docs/NETWORKING.md
@@ -52,8 +52,6 @@ Mappers receive an instance of `Data` and return  the result of parsing that dat
 ## Model objects
 Model objects declared in `Networking` are immutable, and modelled as value types (structs) that typically implement `Decodable`.
 
-Model objects should implement `Comparable`.
-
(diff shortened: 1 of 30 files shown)


docs/NETWORKING.md
```diff
@@ -52,8 +52,6 @@ Mappers receive an instance of `Data` and return  the result of parsing that dat
 ## Model objects
 Model objects declared in `Networking` are immutable, and modelled as value types (structs) that typically implement `Decodable`.
 
-Model objects should implement `Comparable`.
```
shiki, line 55: Thank you for being thorough!

shiki (approved):
Thanks for cleaning up! I did a quick smoke test. :shipit:

PR #4656, opened 2021-07-20
Shipping Labels: Handle decimal quantities in package details

rachelmcr:
Fixes: #4640

## Description

When an order has decimal quantities, we should round up the quantity to the nearest bigger integer. If an order has 0.5 as quantity, the list of products in the package should show 1 product, and if it's 3.5, it should show 4, which is done by rounding the quantity to the nearest bigger integer.

This PR updates our Decimal `intValue` to round up by default. As far as I can see, we only use this when handling order item quantities, so it shouldn't affect other parts of the app.

This also affects the order item weight and total package weight, so this PR updates those weight calculations to provide an accurate weight for non-integer item quantities. For example, if an order has 0.5 as the quantity for an item that weighs 100g, the list of products in the package shows 1 product that weighs 50g, and the total package weight is 50g.

## Changes

* Updates `Decimal+Helpers` to default to rounding up instead of down for positive numbers (and down instead of up for negative numbers).
* Updates `ShippingLabelPackageDetailsViewModel`:
   * For each item in the order, the item weight is displayed as the product weight unless the quantity is less than 1. In that case, the weight is calculated by multiplying the product weight by the quantity to get an accurate weight. (I used `tempItemQuantity` to keep track of the quantity for calculating the weight for each item in the order, in particular in cases where there is a decimal quantity greater than 1, e.g. 2.5 items in the order. I'm very open to feedback about other ways to do this, if you have ideas.)
   * For the total package weight, the item weight is multiplied by the actual item quantity (not rounded) for an accurate package weight.

/|Order details|Package details - Before|Package details - After
-|-|-|-
Order with 0.5 item quantity|Simulator Screen Shot - iPhone 11 Pro - 2021-07-20 at 12 25 29|Simulator Screen Shot - iPhone 11 Pro - 2021-07-20 at 13 02 00|Simulator Screen Shot - iPhone 11 Pro - 2021-07-20 at 12 25 40
Order with 2.5 item quantity|Simulator Screen Shot - iPhone 11 Pro - 2021-07-20 at 12 17 50|Simulator Screen Shot - iPhone 11 Pro - 2021-07-20 at 13 01 38|Simulator Screen Shot - iPhone 11 Pro - 2021-07-20 at 12 18 00

## Testing

2. Launch the app.
3. Open an order detail that is eligible for the shipping label and has a decimal item quantity in the order.
6. In the package details section, under "Items to fulfill" make sure the number of items is rounded up and the weight is accurate for the actual item quantity. (So an order with 0.5 as the item quantity shows 1 item and the weight is half the item weight.)
7. Under "Package details," make sure that the "Total package weight" displayed is now correct considering the decimal item quantity, and it's the sum of the weight of the item shown on the screen.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/Decimal+Helpers.swift
+++ b/WooCommerce/Classes/Extensions/Decimal+Helpers.swift
@@ -2,20 +2,20 @@ import Foundation
 
 extension Decimal {
 
-    /// Returns the int value of a decimal. We ensure we round down our Decimal before converting it to an Int, using NSDecimalRound.
+    /// Returns the int value of a decimal. We ensure we round up our Decimal before converting it to an Int, using NSDecimalRound.
     ///
     var intValue: Int {
         NSDecimalNumber(decimal: whole).intValue
     }
 
-    private func rounded(_ roundingMode: NSDecimalNumber.RoundingMode = .down, scale: Int = 0) -> Self {
+    private func rounded(_ roundingMode: NSDecimalNumber.RoundingMode = .up, scale: Int = 0) -> Self {
         var result = Self()
         var number = self
         NSDecimalRound(&result, &number, scale, roundingMode)
         return result
     }
 
-    private var whole: Self { rounded( self < 0 ? .up : .down) }
+    private var whole: Self { rounded( self < 0 ? .down : .up) }
 
     private var fraction: Self { self - whole }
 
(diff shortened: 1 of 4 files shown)


pmusolino (approved):
:shipit: thank you Rachel for working on this!

> This PR updates our Decimal intValue to round up by default. As far as I can see, we only use this when handling order item quantities, so it shouldn't affect other parts of the app.

Correct. I double-checked it just to be sure to not introduce a regression elsewhere.

Issue #4648, opened 2021-07-19
Shipping Labels: Total package weight should include weight of selected package


rachelmcr:
**Describe the bug**

Currently, the total package weight is the sum of the weight of all the products in the order, and it does not consider the weight of the selected package. (Right now, predefined packages all have `box_weight` of 0, but custom packages can have an additional weight that should be included in the total package weight.)

<img src="https://user-images.githubusercontent.com/495617/126153764-ddf0e75c-c201-47fb-9a2e-811332d5516f.png" width="300px">

rachelmcr:
Fixed in #4658 for the 7.1 release.

PR #4658, opened 2021-07-20
Shipping Labels: Include selected package weight in total package weight

rachelmcr:
Fixes: #4648

## Description

When creating a shipping label, the Package Details screen has an editable field for the `total package weight`. The total package weight should include the weight of all the order items _and_ the weight of the selected package. We weren't including the selected package weight, so this PR adds that to the total package weight.

The total package weight is calculated when you first open the Package Details screen and any time the selected package is changed. (Note that only custom packages have a set weight, so if you choose a predefined package the total weight will be the total weight of the order items.) However, if that total weight is edited manually or saved, we _stop_ automatically updating the weight, even if the selected package changes. (This is consistent with web behavior and avoids overwriting the weight if the user has set it themselves.)

## Changes

* Adds an optional `onEditingChanged` closure to `TitleAndTextFieldRow`, so we can track when the text field has been edited. (I used `onEditingChanged` and not `onCommit` because we're using this with a decimal keyboard, which doesn't include a Return key, so we can't track committed changes to the text field.)
* In `ShippingLabelPackageDetailsViewModel`:
   * Adds `isPackageWeightEdited` to track when the package weight has been edited or saved.
   * In `setTotalWeight()`, check for `isPackageWeightEdited` instead of `totalWeight.isEmpty` before setting the total package weight, and add the selected package weight to the total. This makes it possible to update the total weight after it is set initially (e.g. to update it when the selected package changes) but still prevents the total weight from changing if the weight has been updated manually.
   * Set the total weight when a selected package is confirmed (in `confirmPackageSelection()`) to make sure the total weight is updated when the selected package changes.
* In `ShippingLabelPackageDetails`, sets `isPackageWeightEdited` to `true` (to track that the weight has been manually edited) if the text field is edited.
* Updates unit tests.

## Testing

1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more payment methods and one or more custom packages (with a set package weight) set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Launch the app.
6. In the package details section, confirm that the "Total package weight" include both the weight of the order items and the weight of the selected package (if any package is selected).
7. Select a custom package and confirm the "Total package weight" is updated with the custom package's weight.
8. Edit the "Total package weight" and confirm that it doesn't change automatically after that, even if you select a new package, unless you tap the "Back" button to exit without saving your changes.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -120,6 +120,7 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         products = resultsControllers?.products ?? []
         productVariations = resultsControllers?.productVariations ?? []
         itemsRows = generateItemsRows()
+        setTotalWeight()
     }
 
     /// Generate the items rows, creating an element in the array for every item (eg. if there is an item with quantity 3,
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -66,6 +66,10 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
     @Published private(set) var selectedPredefinedPackage: ShippingLabelPredefinedPackage?
     @Published var totalWeight: String
 
+    /// Whether the user has edited the total package weight. If true, we won't make any automatic changes to the total weight.
+    ///
+    @Published var isPackageWeightEdited: Bool = false
+
     /// Returns if the custom packages header should be shown in Package List
     ///
     var showCustomPackagesHeader: Bool {
@@ -90,6 +94,9 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.packagesResponse = packagesResponse
         self.selectedPackageID = selectedPackageID
         self.totalWeight = totalWeight ?? ""
+        if totalWeight != nil {
+            self.isPackageWeightEdited = true
+        }
         configureResultsControllers()
         syncProducts()
         syncProductVariations()
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -64,7 +64,7 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
     }
     @Published private(set) var selectedCustomPackage: ShippingLabelCustomPackage?
     @Published private(set) var selectedPredefinedPackage: ShippingLabelPredefinedPackage?
-    @Published var totalWeight: String
+    @Published var totalWeight: String = ""
 
     /// Whether the user has edited the total package weight. If true, we won't make any automatic changes to the total weight.
     ///
@@ -93,14 +93,11 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.weightUnit = weightUnit
         self.packagesResponse = packagesResponse
         self.selectedPackageID = selectedPackageID
-        self.totalWeight = totalWeight ?? ""
-        if totalWeight != nil {
-            self.isPackageWeightEdited = true
-        }
         configureResultsControllers()
         syncProducts()
         syncProductVariations()
         setDefaultPackage()
+        setTotalWeight(using: totalWeight)
     }
 
     private func configureResultsControllers() {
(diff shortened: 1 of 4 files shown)


itsmeichigo (commented):
Thank you or fixing this bug! I have a few comments and questions about your changes below:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -66,6 +66,10 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
     @Published private(set) var selectedPredefinedPackage: ShippingLabelPredefinedPackage?
     @Published var totalWeight: String
 
+    /// Whether the user has edited the total package weight. If true, we won't make any automatic changes to the total weight.
+    ///
+    @Published var isPackageWeightEdited: Bool = false
```
itsmeichigo: Since we're not observing `isPackageWeightEdited`, I think it's safe to remove `@Published` here and keep it a plain variable. What do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -90,6 +94,9 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.packagesResponse = packagesResponse
         self.selectedPackageID = selectedPackageID
         self.totalWeight = totalWeight ?? ""
+        if totalWeight != nil {
+            self.isPackageWeightEdited = true
+        }
```
itsmeichigo: I'm not sure this is the expected behavior. I tested with the following steps:
- I select one of the boxes (custom or not) and tapped Done button
- After navigating back to the Create Shipping Label form - I get back and change to another box.
- Now my previous choice is deemed custom input (but it isn't), therefore the weight of my box isn't updated automatically when I choose any other box.

I tested the same steps on the web app and saw that as long as I don't manually input a weight, I can still get automatic weight updates when I get back.

I would suggest that you check for the initial `isPackageWeightEdited` by comparing the initial `totalWeight` with a calculated weight from the products, variations and initial package. If the weight is the same as calculated - we can safely decide that the initial weight was not input manually.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -104,12 +111,10 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
             guard let self = self else { return }
             self.products = products
             self.itemsRows = self.generateItemsRows()
-            self.setTotalWeight()
```
itsmeichigo: Do you think it is safe to not calculate the total weight again when product list change? I know the chance that product list changes in this case is low - but maybe some extra safety is necessary?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -104,12 +111,10 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
...
         }, onProductVariationsReload: { [weak self] (productVariations) in
             guard let self = self else { return }
             self.productVariations = productVariations
             self.itemsRows = self.generateItemsRows()
-            self.setTotalWeight()
```
itsmeichigo: Same question as above - but about product variations changes.

WooCommerce/WooCommerceTests/ViewRelated/Shipping Label/Create Shipping Label/ShippingLabelPackageDetailsViewModelTests.swift
```diff
@@ -336,6 +336,34 @@ final class ShippingLabelPackageDetailsViewModelTests: XCTestCase {
...
+        // Then
+        XCTAssertEqual(viewModel.totalWeight, "120.0")
+
+        // When
+        viewModel.didSelectPackage("Box")
+        viewModel.confirmPackageSelection()
+
+        // Then
+        XCTAssertEqual(viewModel.totalWeight, "130.0")
+    }
```
itsmeichigo, line 366: It would be great if we have more tests for cases when total weight should not be updated automatically:
- when initial total weight is an arbitrary weight (not equal to automatically calculated weight)
- when user enters a new total weight manually.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
rachelmcr: Thanks for catching this! I removed these from products/variations at one point and then meant to put them back, but somehow missed them — I agree it's better to have them there.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
rachelmcr: You're right, thanks! I must have input the weight manually while testing the web behavior, and didn't think about the different expectation when the weight was saved but hadn't been input manually. I have a fix for that using your suggested approach in 2689621.

WooCommerce/WooCommerceTests/ViewRelated/Shipping Label/Create Shipping Label/ShippingLabelPackageDetailsViewModelTests.swift
rachelmcr: Great ideas! Thanks for keeping me on my toes with test coverage. :) I added those tests in 3dcd3c2.

rachelmcr:
Thank you for the careful review, @itsmeichigo! This is ready for another look.

itsmeichigo (approved):
Thank you Rachel for the updates! 🎉 
Code looks good and works as expected now :shipit:

Issue #3745, opened 2021-02-25
[Mobile Payments] Connect Your Card Reader View: Hide Footer Bar, Tweak Styles


allendav:
Feedback on https://github.com/woocommerce/woocommerce-ios/pull/3699 https://github.com/woocommerce/woocommerce-ios/pull/3699#issuecomment-785791371

- [x] Change screen title to “Manage Card Reader”
- [x] Use Headline font for “Connect your card reader”
- [ ] Image too big - should be 206 x 180px
- [x] Title case on Button (should be “Connect Card Reader”)
- [x] Pink Learn more link
- [x] Hide Bottom Nav (Tab) Bar

allendav:
@ctarda @Garance91540 - I've ticked off the things that have been accomplished. All that remains is tweaking the image size (it comes in at about 310x263 (x2) on my iPhone 12 and the bottom nav is not (yet) suppressed:

<img width="533" alt="Screen Shot 2021-07-19 at 11 41 18 AM" src="https://user-images.githubusercontent.com/1595739/126210695-7f9d35c0-5ff9-4d89-9076-92a4211e6f40.png">

Garance91540:
Looking great @allendav , thank you for working on this!

allendav:
Thank you! You are most welcome! :)

PR #4660, opened 2021-07-20
Hide bottom tab bar for card reader settings presenting view controller

allendav:
Addresses "hide tab bar in card reader settings" from #3745 

To test:
- Enter settings > manage card readers
- Ensure the tab bar is hidden and the view occupies the space where the tab bar was
- Exit back to settings
- Ensure the tab bar reappears
- Repeat for both light and dark modes and ensure the view background color is used where the tab bar was

fyi @Garance91540 @adamzelinski 

hide-tab-bar

Note:
- This change also hides the tab bar if the reader is not already connected when collecting a payment for an order. An upcoming change will remove the CardReaderSettingsPresentingViewController from that flow, but for now you may also verify that the tab bar is restored when returning to collect payment flow after connecting a reader in that context

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Dashboard.storyboard
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Dashboard.storyboard
@@ -177,7 +177,7 @@
         <!--Card Reader Settings Presenting View Controller-->
         <scene sceneID="NEs-n8-mBS">
             <objects>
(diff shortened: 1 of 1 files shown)


ctarda (approved):
I tested going in and out of this screen connecting and disconnecting, and all the combinations of the above I could think of, and it seems to work well.

:shipit:

Garance91540:
Looking good!

PR #4662, opened 2021-07-20
System Plugins - Update Yosemite layer - Add Action and Store - 2/2

fernandofgfer:
#4646 2/2

## Summary
This PR has been splitter into two PRs.  
This PR contains the Yosemite updates of the System plugins project.  
In this branch, I've included the next classes
- SystemPluginAction: To define both actions.
- SystemPluginStore: To execute these actions.

## Actions
I've added two actions:
- synchronizeSystemPlugins: This action retrieves all plugins and saves them into specific storage.
- fetchSystemPlugins: This action returns all plugins saved in storage.

## How to test it
I've added unit tests for SystemPluginStore.
Execute all unit tests

Update release notes:


Yosemite/Yosemite/Actions/SystemPluginAction.swift
```diff
@@ -0,0 +1,14 @@
...
+///
+public enum SystemPluginAction: Action {
+
+    /// Synchronize all system plugins for a site given its ID
+    ///
+    case synchronizeSystemPlugins(siteID: Int64, onCompletion: (Result<Void, Error>) -> Void)
+
+    /// Fetch all system plugins for a site given its ID
+    ///
+    case fetchSystemPlugins(siteID: Int64, onCompletion: ([SystemPlugin]?) -> Void)
```
Ecarrion: Could you explain the use case for this action? I don't think we need it 🤔

Yosemite/Yosemite/Stores/SystemPluginStore.swift
```diff
@@ -0,0 +1,102 @@
...
+        // remove stale system plugins
+        let currentSystemPlugins = readonlySystemPlugins.map(\.name)
+        storage.deleteStaleSystemPlugins(siteID: siteID, currentSystemPlugins: currentSystemPlugins)
+    }
+
+    /// Retrieve `SystemPlugin` entities of a specified storage by siteID
+    ///
+    func fetchSystemPlugins(siteID: Int64, completionHandler: @escaping ([SystemPlugin]?) -> Void) {
+        let writerStorage = storageManager.writerDerivedStorage
+        let systemPlugins = writerStorage.loadSystemPlugins(siteID: siteID).map { $0.toReadOnly() }
```
Ecarrion: ⚠️ We should not be using a writer-derived storage to read data. It's probably better to use `storageManager.viewStorage`

Ecarrion (changes requested):
Hi @fernandofgfer, This looks good! I only left a small concern to address!

Suggestion: What do you think about naming these `Actions` and `Store` something like `SystemStatusAction` and `SystemStatusStore`. The reason is to not having to create multiple new stores when in the future we need to fetch more information from that endpoint!

Yosemite/Yosemite/Actions/SystemPluginAction.swift
fernandofgfer: Hi! As we talked about, in the tool that we will build to know if a feature could be enabled by his plugins, we will need something to get all plugins. So I think we will need a method to retrieve the plugins saved in the storage layer.

Ecarrion: Sure, but why we would need to fetch all of them? I think fetching just one should be enough, don't you think?

fernandofgfer: Yes, If we do the tools in that way, of course. In the end, I think it depends on if you want to load all plugins and check if a feature is enabled with that list, or if you want to search a specific plugin to know if a feature should be enabled or not. I will change with the second option.

fernandofgfer:
> Hi @fernandofgfer, This looks good! I only left a small concern to address!
> 
> Suggestion: What do you think about naming these `Actions` and `Store` something like `SystemStatusAction` and `SystemStatusStore`. The reason is to not having to create multiple new stores when in the future we need to fetch more information from that endpoint!

Nice catch!! It's a good idea to reuse the store and the action in the future. Thanks!

fernandofgfer:
Hi @Ecarrion. Changes added!

Yosemite/Yosemite/Actions/SystemStatusAction.swift
```diff
@@ -0,0 +1,14 @@
...
+///
+public enum SystemStatusAction: Action {
+
+    /// Synchronize all system plugins for a site given its ID
+    ///
+    case synchronizeSystemPlugins(siteID: Int64, onCompletion: (Result<Void, Error>) -> Void)
+
+    /// Fetch an specific systemPlugin by siteID and name
+    ///
+    case fetchSystemPlugins(siteID: Int64, systemPluginName: String, onCompletion: (SystemPlugin?) -> Void)
```
Ecarrion:
```suggestion
    case fetchSystemPlugin(siteID: Int64, systemPluginName: String, onCompletion: (SystemPlugin?) -> Void)
```

Ecarrion (approved):
Good Job, the only thing I see missing is that this `Store` is not yet registered in our `AuthenticatedState`. And you won't be able to use it unless you register it.

fernandofgfer:
> Good Job, the only thing I see missing is that this `Store` is not yet registered in our `AuthenticatedState`. And you won't be able to use it unless you register it.

Thanks. I will add it in the next PR!

Ecarrion (approved):

PR #4663, opened 2021-07-21
[Shipping Labels] Hide contact button when phone number is missing

### LINKED ISSUE #4659: Shipping Labels: Don't show contact customer button if phone number is empty
**Describe the bug**

h/t @hichamboushaba for the report:

The contact customer button shouldn’t be displayed if the phone is empty, as it caused showing an empty dial dialog.

**To Reproduce**

2. Create a new order without entering a customer phone number.
3. Open the order in the app.
3. Tap the button "Create Shipping Label".
4. Confirm the "Ship to" address.
5. Select "Continue" on the "Ship from" address and notice the "Contact Customer" button appears even though there is no phone number to contact the customer.

**Screenshots**

itsmeichigo:
Closes #4659 

# Description
This PR aims to hide contact button on destination address form top banner if when customer doesn't provide phone number.

# Changes
- Added new parameter phone number to `ShippingLabelAddressTopBannerFactory`
- In the factory, hide contact button if phone number is missing.
- Updated title of banner if phone number is missing.
- Made `topBannerView` calculated variable instead of lazy to make sure that contact button shows up if phone number is entered in the form.

# Screenshots
| Phone number present | Phone number missing |
| ----- | ----- |
|  |  |

# Testing
0. Make sure your test store is eligible to create shipping labels.
1. On WooCommerce web app, make phone number optional in Settings > Checkout > Billing.
2. Create an order without phone number.
3. On WCiOS app, open the new order and select Create Shipping Label.
4. On Ship To, update the address to a fake one (e.g 123 Test Street) and tap Done.
5. Notice that there's a warning banner appearing on the top without Contact Customer button.
6. Enter phone number manually in Phone field and tap Done.
7. Notice that the warning banner on the top shows Contact Customer button.

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressTopBannerFactory.swift
```diff
@@ -5,17 +5,37 @@ import Yosemite
 ///
 final class ShippingLabelAddressTopBannerFactory {
     static func addressErrorTopBannerView(shipType: ShipType,
+                                          phoneNumber: String?,
                                           openMapPressed: @escaping () -> Void,
                                           contactCustomerPressed: @escaping () -> Void) -> TopBannerView {
         // Set banner text and action buttons based on shipping address type (origin or destination).
```
rachelmcr: nit: It could be helpful to update this comment since the text and action buttons now also depend on the shipping phone number.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressTopBannerFactory.swift
```diff
@@ -5,17 +5,37 @@ import Yosemite
...
-        let infoText = shipType == .destination ? Localization.infoShipTo : Localization.infoShipFrom
+        let missingPhoneNumber = phoneNumber == nil || phoneNumber?.isEmpty == true
```
rachelmcr, line 13: nit: We can simplify `phoneNumber?.isEmpty == true` to just check `phoneNumber?.isEmpty` since that returns a `Bool`, right?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -17,8 +17,11 @@ final class ShippingLabelAddressFormViewController: UIViewController {
 
     /// Top banner that shows a warning in case there is an error in the address validation.
     ///
-    private lazy var topBannerView: TopBannerView = {
-        let topBanner = ShippingLabelAddressTopBannerFactory.addressErrorTopBannerView(shipType: viewModel.type) { [weak self] in
+    private var topBannerView: TopBannerView {
```
rachelmcr, line 20: I really appreciate that you thought of the case where a phone number is added to the form. 🎉   However, an unfortunate side effect of making this a computed property instead of lazy is that multiple instances of the banner can appear on the screen.

For example, if you start with an invalid address on the order and no phone number, you'll see the banner with just the "Open Map" button. If you then add a phone number and tap "Done" you'll see a second banner with the "Contact Customer" button. Each time you tap "Done" another banner appears:

Banner is initially shown | After adding phone and tapping "Done"
-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-07-21 at 18 36 27|Simulator Screen Shot - iPhone 11 Pro - 2021-07-21 at 18 36 38

rachelmcr (commented):
Thanks for working on this fix! I just have a concern (described below) about multiple banners appearing at the same time.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressTopBannerFactory.swift
itsmeichigo: I'd want to consider a nil `phoneNumber` as missing too (this happens when address is not found). So I think this check covers more edge cases - that makes sense right?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
itsmeichigo: Thank you for catching that bug! I just pushed a fix for this in db7ad1a4028024ddf8620de238c7f349b18c9d38 by removing the old banner from the stack view when displaying the banner again.

itsmeichigo:
Thanks @rachelmcr for your comments! I have pushed some updates and responded to one comment above, please take another look 🙇‍♀️

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressTopBannerFactory.swift
rachelmcr: Oh, yes! I just meant to ask whether the `isEmpty` check can be simplified. So instead of this:

`let missingPhoneNumber = phoneNumber == nil || phoneNumber?.isEmpty == true`

Can that line be this:

`let missingPhoneNumber = phoneNumber == nil || phoneNumber?.isEmpty`

(It isn't a big deal, I mostly wanted to double check my own understanding.)

rachelmcr (approved):
Code & testing looks great now! :shipit:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressTopBannerFactory.swift
itsmeichigo: @rachelmcr since `phoneNumber?.isEmpty` is an optional boolean, I have to explicitly check with ` == true`, otherwise the compiler will be confused and complain something like "Optional type 'Bool?' cannot be used as a boolean; test for '!= nil' instead". So that check is necessary 😄

rachelmcr: Ah, right! Now I get it; thanks for the explanation :)

PR #4669, opened 2021-07-22
System plugins - Sync System plugins in DefaultStoreManager and Site plugins in PluginList

### LINKED ISSUE #4606: Plugins: Recover plugins in plugin screen
Parent: #4278

- [ ] Move sync site plugin to plugin screen
- [ ] Sync System plugins in `DefaultStoresManager`

fernandofgfer:
Closes #4606 
## Description
This PR aims to sync the new System Plugins in DefaultStoreManager and sync Site Plugins in PluginListViewController

## Changes
- Remove sync Site Plugins of DefaultStoreManager, it's not needed.
- Add sync System Plugins in DefaultStoreManager.
- Add sync Site Plugins in PluginList screen before `resultController` init.

## Testing
Test System plugins:
- Use "Launch Wormholy Debug" feature to see that `system_status` call is launched.

Test Site plugins:
- Go to the Plugins screen and check that the ghost table appears and the data is loaded.
- Add a new plugin to your store and scroll the Plugins screen, it should be updated with the new plugin added.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
@@ -43,7 +43,7 @@ final class PluginListViewController: UIViewController {
         configureNavigation()
         configureTableView()
         configureGhostTableView()
-        configureViewModel()
+        syncPlugins(onCompletion: configureViewModel)
     }
 }
 
@@ -97,16 +97,27 @@ private extension PluginListViewController {
 // MARK: - Actions
 //
 private extension PluginListViewController {
+    
+    /// resyncPlugins is used when the view need to resynchronize data
+    ///
     @objc func resyncPlugins() {
+        syncPlugins(onCompletion: nil)
+    }
+
+    /// syncPlugins synchronizes all plugins and execute onCompletion when is finished
+    ///
+    @objc func syncPlugins(onCompletion: (() -> Void)?) {
         removeErrorStateView()
         startGhostAnimation()
-        viewModel.resyncPlugins { [weak self] result in
+        viewModel.syncPlugins { [weak self] result in
             guard let self = self else { return }
             self.refreshControl.endRefreshing()
             self.stopGhostAnimation()
             if result.isFailure {
                 self.displayErrorStateView()
             }
+            guard let onCompletion = onCompletion else { return }
+            onCompletion()
         }
     }
 }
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -97,16 +97,27 @@ private extension PluginListViewController {
...
-        viewModel.resyncPlugins { [weak self] result in
+        viewModel.syncPlugins { [weak self] result in
             guard let self = self else { return }
             self.refreshControl.endRefreshing()
             self.stopGhostAnimation()
             if result.isFailure {
                 self.displayErrorStateView()
             }
+            guard let onCompletion = onCompletion else { return }
+            onCompletion()
```
Ecarrion: I think this is simpler and more common in our code base :-)
```suggestion
            onCompletion?()
```

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
```diff
@@ -43,7 +43,7 @@ final class PluginListViewController: UIViewController {
         configureNavigation()
         configureTableView()
         configureGhostTableView()
-        configureViewModel()
+        syncPlugins(onCompletion: configureViewModel)
```
Ecarrion: Could you explain a little bit why do we need to configure the view model after plugins are synced? 

The `VM` is in charge of fetching the plugins, but this call stack implies that something else syncs the plugin and then we configure the view model.

I think we should aim to sync the plugins inside the `configureViewModel` method.

Ecarrion: Something like
```swift
func configureViewModel() {
    viewModel.observePlugins { [weak self] in
        self?.tableView.reloadData()
    }

    syncPlugins(onCompletion: nil)
}
```

Ecarrion (changes requested):
Hi @fernandofgfer, I have only one concern regarding how are we invoking view model methods in the view controller!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
fernandofgfer: Hi! I was using the guard to avoid the use of the optional. I will change it.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Plugins/PluginListViewController.swift
fernandofgfer: The only reason why I did it in that way is to first sync the data and second initialize the resultController. But in the end the result is exactly the same. I'm going to change it.

Ecarrion (approved):
Looks and works well to me!

PR #4670, opened 2021-07-23
Fix: Add network mock for users/me v2 endpoint

dvdchr:
The UI test were broken because there's an additional API endpoint being depended by the Store Picker flow (`wp/v2/sites/{siteID}/users/me`) that wasn't mocked. This caused the UI Test to be stalled in Store Picker screen, causing the tests to fail.

This PR fixes that by adding the proper mocks.

## To Test

Ensure the WooCommerceUITests passes successfully.

## Author Checklist
- [x] I have ensured that the UI tests pass locally.


ealeksandrov (approved):
Looks good. Thanks for quick fix! :shipit:

dvdchr:
Thank you @ealeksandrov !

PR #4672, opened 2021-07-23
Update MediaType to use modern UTType API

ealeksandrov:
## Description

Initially I stumbled upon `MediaType` because it was failing for me in all tests. I thought it's because of old API + M1 + Rosetta and decided to update it since our deployment target is already iOS 14 (required for `UniformTypeIdentifiers` module).

In the end, tests on new API failed as well 😅
I found Rosetta layer to be the issue - since old and new types API returns `nil` for some reason. On native ARM Xcode and Intel it works well in both cases.

I'm pushing the change since it's an improvement anyway, removing all CF bridging from codebase (`takeRetainedValue`).

## Further improvements

We can replace all usage of `kUTType..`  strings (from `MobileCoreServices`) with [UTType](https://developer.apple.com/documentation/uniformtypeidentifiers/uttype) definitions (from `UniformTypeIdentifiers`, iOS 14+), but the immediate benefit is unclear.

## Tests

All changes already well covered by unit tests, but please upload couple of different media files to smoke test data flow.

---
Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Tools/Media/MediaAssetExporter.swift
+++ b/Yosemite/Yosemite/Tools/Media/MediaAssetExporter.swift
@@ -2,6 +2,7 @@ import Foundation
 import MobileCoreServices
 import AVFoundation
 import Photos
+import UniformTypeIdentifiers
 
 /// Exports a media item of `PHAsset` type to be uploadable.
 ///
@@ -108,8 +109,7 @@ private extension MediaAssetExporter {
         }
 
         guard allowableFileExtensions.isEmpty == false,
-            let extensionType = UTTypeCopyPreferredTagWithClass(uti as CFString, kUTTagClassFilenameExtension)?.takeRetainedValue() as String?
-            else {
+              let extensionType = UTType(uti)?.preferredFilenameExtension else {
                 return nil
         }
         guard allowableFileExtensions.contains(extensionType) else {
--- a/Yosemite/Yosemite/Tools/Media/MediaAssetExporter.swift
+++ b/Yosemite/Yosemite/Tools/Media/MediaAssetExporter.swift
@@ -2,7 +2,6 @@ import Foundation
 import MobileCoreServices
 import AVFoundation
 import Photos
-import UniformTypeIdentifiers
 
 /// Exports a media item of `PHAsset` type to be uploadable.
 ///
@@ -109,10 +108,10 @@ private extension MediaAssetExporter {
         }
 
         guard allowableFileExtensions.isEmpty == false,
-              let extensionType = UTType(uti)?.preferredFilenameExtension else {
+              let fileExtensionForType = URL.fileExtensionForUTType(uti) else {
                 return nil
         }
-        guard allowableFileExtensions.contains(extensionType) else {
+        guard allowableFileExtensions.contains(fileExtensionForType) else {
             return kUTTypeJPEG as String
         }
         return uti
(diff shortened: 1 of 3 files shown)


Yosemite/Yosemite/Tools/Media/MediaType.swift
```diff
@@ -36,22 +34,18 @@ public enum MediaType {
...
-        if UTTypeConformsTo(fileUTI, kUTTypeImage) {
+        if dataType.conforms(to: UTType.image) {
             self = .image
-        } else if UTTypeConformsTo(fileUTI, kUTTypeVideo) {
+        } else if dataType.conforms(to: UTType.movie) {
             self = .video
-        } else if UTTypeConformsTo(fileUTI, kUTTypeMovie) {
-            self = .video
-        } else if UTTypeConformsTo(fileUTI, kUTTypeMPEG4) {
-            self = .video
```
ealeksandrov, line 51: Video and MPEG4 types removed, because:
Video types conform to `video`, `video` [conforms](https://developer.apple.com/documentation/uniformtypeidentifiers/uttypevideo) to `movie`, `movie` is already in the list.

Yosemite/Yosemite/Tools/Media/MediaAssetExporter.swift
```diff
@@ -108,8 +109,7 @@ private extension MediaAssetExporter {
         }
 
         guard allowableFileExtensions.isEmpty == false,
-            let extensionType = UTTypeCopyPreferredTagWithClass(uti as CFString, kUTTagClassFilenameExtension)?.takeRetainedValue() as String?
```
shiki, line 111: The PR description mentions that these are covered by unit tests. But this line doesn't look like it's covered. I placed a breakpoint here and ran all the YosemiteTests but this line is never executed. Code coverage says it's also zero: 

image

Nevertheless, this seems to be working the same as the previous implementation and the cost of adding a unit test for this class seems too large. So I guess it's fine. 🤷

Yosemite/Yosemite/Tools/Media/MediaType.swift
shiki: Thanks for the explanation!

shiki (approved):
LGTM! I did a quick uploading test using a real device and didn't notice any problems. :shipit:

Yosemite/Yosemite/Tools/Media/MediaAssetExporter.swift
ealeksandrov: Good catch! This was a last-minute find and it forgot about it while writing PR.
I also don't see a reasonable way to test this code. But on a second look, UTType operation here can be replaced with URL extension [already used](https://github.com/woocommerce/woocommerce-ios/blob/4fe952211777fad134084cbc19d4db229a37c8c9/Yosemite/Yosemite/Tools/Media/MediaImageExporter.swift#L112) in `MediaImageExporter`. And URL extension on itself is [covered](https://github.com/woocommerce/woocommerce-ios/blob/4fe952211777fad134084cbc19d4db229a37c8c9/Yosemite/YosemiteTests/Tools/Media/URL%2BMediaTests.swift#L33) by unit tests, so it'll be a slightly better trade-off 😄
The change pushed in 776397d.

ealeksandrov:
Thank you for review @shiki! I've added some minor improvements/refactoring in 776397d, but there are no functional changes, so I'll merge it right away.

PR #4673, opened 2021-07-24
Merge 7.1 final into develop

jkmassel:
Merges changes from the last `7.1` beta and the release finalization into `develop`.

@rachelmcr or @pmusolino – I've assigned both of you, but I just need one person to give a 👍 on the merge conflict resolution please – I had to do a merge in `WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetails.swift`, which shouldn't have any issues, but I'd like a second set of eyes on that just in case 🙂


mokagio (approved):
- [x] Fastlane  localized `release_notes.txt` updates
- [x] Localized `.strings` updates
- [x] Build number bump in version `.xcconfig` file
- [x] Diffs from the PRs that landed on the release branch since the last beta

I'm approving but holding off merging because Jeremy explicitly asked for a merge conflict resolution review which I'm in no position to give. I checked and saw that there's no auto-merge on this PR, so it's should be okay. That is, the PR won't end up accidentally merged before the resolution is validated.

pmusolino:
I quickly tested the PR, but it seems that something goes wrong during the merge because I noticed that the background of some cells in the Shipping Label Package Details is wrong (should be white).
I believe this has to do with @itsmeichigo's work on fixing the edge-to-edge screen, and that something blew up during the merge.

itsmeichigo:
@pmusolino  I have pushed 80528bcacf76751291781ebef223fbfb5d9ddc3f to fix the incorrect conflict resolution and run a quick smoke test on shipping labels, looks like it works correctly now (removed the duplicated section and added @rachelmcr's fix for package weight update).

pmusolino (approved):
I re-tested it, and now it looks good to me. Thanks 💯 @itsmeichigo for fixing it super quickly 🙇

jkmassel:
Thanks y'all!!! 💜

PR #4679, opened 2021-07-27
Shipping Labels: Refactor package details view model with Combine

### LINKED ISSUE #4666: Shipping Labels: Use Combine to reduce duplicated code in ShippingLabelPackageDetailsViewModel
# Description
Currently in `ShippingLabelPackageDetailsViewModel`, I see there are repeated calls to `generateItemsRows` here and there. This can be a good case to use Combine's `combineLatest` to observe changes in 2 or more publishers and update variables.

# Suggested solution
If `products` and `productVariations` were updated to be `@Published` variables, they can be used to observe changes in their values and then  Combine's`combineLatest` can be used to update `itemRows` directly.

The same thing can be updated for calculating and updating package weight based on changes of product, variations and selected package ID.

itsmeichigo:
Closes #4666 

# Description
Currently the view model of Package Details has duplicated calls to `setTotalWeight` and `setItemRows`. Since we're already using publishers, I figure this is a nice opportunity to make use of Combine. This PR aims to refactor these functions by observing changes in relevant variables and re-calculate data when appropriate.

# Changes
## Configuring item rows
- Updated `generateItemsRows` method to accept products and product variations as input instead of reading from class properties directly.  This is to make sure of the single source of truth for the method.
- Create new method `configureItemsRows` to observe changes in products and variations and get results from `generateItemsRows` to assign to `itemsRows` directly. This method is called directly from the initializer of the class.
- Remove duplicated calls to `generateItemsRows`.

## Configuring total weight
- Renamed `setTotalWeight` to `calculateTotalWeight`, and updated it to accept products, product variations and an optional custom package as input. The result weight will be calculated solely on these input.
- Added an optional variable `initialTotalWeight` as a new private property for the view model.
- Added a new method `configureTotalWeight` to observe changes in selected custom package, products and variations to calculate total weight accordingly. This observing is only handled when the package weight is not manually input, and there's also a check for initial total weight to update `isPackageWeightEdited` appropriately.
- Updated unit tests for `isPackageDetailsDoneButtonEnabled` to reflects the function correctly.

# Testing
Unit tests should still pass after the refactoring. When running the feature, everything should still work as expected:
1. Navigate to Orders tab.
2. Select an order with status Processing.
4. Configure package details: notice that order items are correct, and package weight is updated correctly when switching between packages.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -95,8 +95,8 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.packagesResponse = packagesResponse
         self.selectedPackageID = selectedPackageID
 
-        setDefaultPackage()
         configureResultsControllers()
+        setDefaultPackage()
         syncProducts()
         syncProductVariations()
         configureItemRows()
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -69,7 +69,7 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
 
     /// Whether the user has edited the total package weight. If true, we won't make any automatic changes to the total weight.
     ///
-    var isPackageWeightEdited: Bool = false
+    @Published var isPackageWeightEdited: Bool = false
 
     /// Returns if the custom packages header should be shown in Package List
     ///
@@ -124,7 +124,7 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
             }
             .share()
 
-        // grab the first calculated weight, check with initialTotalWeight
+        // Grab the first calculated weight, check with initialTotalWeight
         // to determine if the weight was manually updated
         calculatedWeight
             .first()
@@ -142,12 +142,16 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
             }
             .assign(to: &$totalWeight)
 
-        // From 2nd element onward, send the calculated weight to total weight
-        // if user has not updated the weight manually
-        calculatedWeight
-            .dropFirst()
-            .filter { [weak self] _ in self?.isPackageWeightEdited == false }
-            .map { String($0) }
+        // Send the calculated weight to total weight; or raise error to cancel the stream
+        // if user has updated the weight manually.
+        calculatedWeight.combineLatest($isPackageWeightEdited)
+            .tryMap { (weight, isEdited) in
+                guard !isEdited else {
+                    throw Errors.overridenTotalWeight
+                }
+                return String(weight)
+            }
+            .catch { _ in Empty() }
             .assign(to: &$totalWeight)
     }
 
@@ -165,6 +169,13 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
 
         products = resultsControllers?.products ?? []
         productVariations = resultsControllers?.productVariations ?? []
+    }
+}
+
+// MARK: - Subtypes
+private extension ShippingLabelPackageDetailsViewModel {
+    enum Errors: Error {
+        case overridenTotalWeight
     }
 }
 
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -200,6 +200,8 @@ private extension ShippingLabelPackageDetailsViewModel {
     /// Calculate total weight based on the weight of the selected package if it's a custom package;
     /// And the products and products variation inside the order items, only if they are not virtual products.
     ///
+    /// Note: Only custom package is needed for input because only custom packages have weight to be included in the total weight.
+    ///
     func calculateTotalWeight(products: [Product], productVariations: [ProductVariation], customPackage: ShippingLabelCustomPackage?) -> Double {
         var tempTotalWeight: Double = 0
 
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -77,6 +77,10 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         return customPackages.count > 0
     }
 
+    /// Initially input package weight if any
+    ///
+    private var initialTotalWeight: String?
+
     init(order: Order,
          packagesResponse: ShippingLabelPackagesResponse?,
          selectedPackageID: String?,
@@ -94,13 +98,14 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.weightUnit = weightUnit
         self.packagesResponse = packagesResponse
         self.selectedPackageID = selectedPackageID
+        self.initialTotalWeight = totalWeight
 
         configureItemRows()
+        configureTotalWeights()
         configureResultsControllers()
         syncProducts()
         syncProductVariations()
         setDefaultPackage()
-        setTotalWeight(using: totalWeight)
     }
 
     /// Observe changes in products and variations to update item rows.
@@ -113,6 +118,30 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         .assign(to: &$itemsRows)
     }
 
+    /// Observe changes in selected package ID, products and variations to update total package weight.
+    ///
+    private func configureTotalWeights() {
+        $selectedCustomPackage.combineLatest($products, $productVariations) { (customPackage, products, variations) in
+            (customPackage, products, variations)
+        }
+        .filter { [weak self] _ in self?.isPackageWeightEdited == false }
+        .map { [weak self] (customPackage, products, variations) -> Double in
+            self?.calculateTotalWeight(products: products, productVariations: variations, customPackage: customPackage) ?? 0
+        }
+        .map { [weak self] in
+            let calculatedWeight = String($0)
+            // Set the total weight as the initial value if it was input manually; otherwise use the calculated weight.
+            if let initialTotalWeight = self?.initialTotalWeight,
+               calculatedWeight != initialTotalWeight {
+                self?.initialTotalWeight = nil
+                self?.isPackageWeightEdited = true
+                return initialTotalWeight
+            }
+            return calculatedWeight
+        }
+        .assign(to: &$totalWeight)
+    }
+
     private func configureResultsControllers() {
         resultsControllers = ShippingLabelPackageDetailsResultsControllers(siteID: order.siteID,
                                                                            orderItems: order.items,
@@ -120,16 +149,13 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
            onProductReload: { [weak self] (products) in
             guard let self = self else { return }
             self.products = products
-            self.setTotalWeight()
         }, onProductVariationsReload: { [weak self] (productVariations) in
             guard let self = self else { return }
             self.productVariations = productVariations
-            self.setTotalWeight()
         })
 
         products = resultsControllers?.products ?? []
         productVariations = resultsControllers?.productVariations ?? []
-        setTotalWeight()
     }
 }
 
@@ -173,16 +199,10 @@ private extension ShippingLabelPackageDetailsViewModel {
         return itemsToFulfill
     }
 
-    /// Set the total weight based on the weight of the selected package and
-    /// the products and products variation inside the order items, only if they are not virtual products.
+    /// Calculate total weight based on the weight of the selected package if it's a custom package;
+    /// And the products and products variation inside the order items, only if they are not virtual products.
     ///
-    /// - Parameter initialWeight: An initial value used to set the total package weight if it was input manually.
-    ///
-    func setTotalWeight(using initialWeight: String? = nil) {
-        guard !isPackageWeightEdited else {
-            return
-        }
-
+    func calculateTotalWeight(products: [Product], productVariations: [ProductVariation], customPackage: ShippingLabelCustomPackage?) -> Double {
         var tempTotalWeight: Double = 0
 
         // Add each order item's weight to the total weight.
@@ -205,17 +225,10 @@ private extension ShippingLabelPackageDetailsViewModel {
 
         // Add selected package weight to the total weight.
         // Only custom packages have a defined weight, so we only do this if a custom package is selected.
-        if let selectedPackage = selectedCustomPackage {
+        if let selectedPackage = customPackage {
             tempTotalWeight += selectedPackage.boxWeight
         }
-
-        // Set the total weight as the initial value if it was input manually; otherwise use the calculated weight.
-        if let initialWeight = initialWeight, initialWeight != String(tempTotalWeight) {
-            isPackageWeightEdited = true
-            totalWeight = initialWeight
-        } else {
-            totalWeight = String(tempTotalWeight)
-        }
+        return tempTotalWeight
     }
 }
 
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -17,11 +18,11 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
 
     /// Products contained inside the Order and fetched from Core Data
     ///
-    private var products: [Product] = []
+    @Published private var products: [Product] = []
 
     /// ProductVariations contained inside the Order and fetched from Core Data
     ///
-    private var productVariations: [ProductVariation] = []
+    @Published private var productVariations: [ProductVariation] = []
```
pmusolino, line 25: A thing that we should keep in mind when working with Combine, is that not always declaring a variable as `@Published` is a benefit since the variables will be observed also by SwiftUI (because the view model is declared as `ObservableObject`), so every change will affect the UI performances. In this specific case, there shouldn't be any performance issues, but that's something to keep in mind for the future.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -93,11 +98,46 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         self.weightUnit = weightUnit
         self.packagesResponse = packagesResponse
         self.selectedPackageID = selectedPackageID
+        self.initialTotalWeight = totalWeight
+
+        configureItemRows()
+        configureTotalWeights()
```
pmusolino: Isn't possible here to pass the `totalWeight` directly to the method, without declaring a new variable `initialTotalWeight`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -160,16 +197,10 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         return itemsToFulfill
     }
 
-    /// Set the total weight based on the weight of the selected package and
-    /// the products and products variation inside the order items, only if they are not virtual products.
-    ///
-    /// - Parameter initialWeight: An initial value used to set the total package weight if it was input manually.
+    /// Calculate total weight based on the weight of the selected package if it's a custom package;
```
pmusolino, line 217: nit: could we add here, for the future, that we are using just the custom package because they are the only type with a weight?

pmusolino (approved):
I still have to enter the reactive programming paradigm, so probably I didn't understand totally all the benefits from it.
Probably, the use of the Combine framework here, for this specific case seems a little bit overkill (more verbose, it seems complex to read, to understand, and probably hard to debug because of all the async callbacks) but at the same time, can be difficult for those how didn't implement the code also to understand the logic behind the old code (which was placed here and there along with the class), so I'm ok with the migration, and it will probably come in handy when we will integrate the multi-package support.
:shipit: 

I left just some minor suggestions.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
itsmeichigo: That makes sense! I'll update the comment to make it clearer.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
itsmeichigo: This variable `initialTotalWeight` honestly seems like a hack since we only use it once to check if the weight was input manually initially, and then it is set to nil again (please look at line 129). 

I can't use `totalWeight` because my intention is to only check the initial weight if any; and if the initial weight is different from calculated weight, we can safely say that the weight was input manually and switching packages should not affect the weight.

Since you're confused with this, I'm afraid it's not the best way to handle it. I'm not sure of a better way though - any suggestion?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
itsmeichigo: I agree! We should totally make a property a publisher only when needed. In this case, I need these to be publishers in order to observe changes in their values.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -93,11 +98,46 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
...
+            return self.generateItemsRows(products: products, productVariations: variations)
+        }
+        .assign(to: &$itemsRows)
+    }
+
+    /// Observe changes in selected custom package, products and variations to update total package weight.
+    ///
+    private func configureTotalWeights() {
+        $selectedCustomPackage.combineLatest($products, $productVariations)
+            .filter { [weak self] _ in self?.isPackageWeightEdited == false }
```
koke: It felt a bit strange to use an "external" variable as the filter. If I'm understanding the code correctly, when `initialTotalWeight` is set, that's the weight that will be used regardless of any UI updates (I don't see this getting any new values when `isPackageWeightEdited` is `true`, and I see no code setting it back to `false`).

If that's the case, maybe we skip the combine logic when there is an initial value. I haven't tested it, but something like:
https://gist.github.com/koke/375587ea5c3a80c2ea756bbc51265580

itsmeichigo: Thank you @koke for taking a look! 

The reason why I keep `initialTotalWeight` inside the data stream is because we have to calculate the total weight before comparing it with the initial value. The calculation depends on values selected package, product and variation list, so comparing the initial value with the total weight directly is not what I intend to do.

`isPackageWeightEdited` is set to `true` in 2 cases:
- when user enters any arbitrary value for the weight (this is updated when text field is edited)
- when the initial value is different from the calculated weight.

Since we intend to not calculate package weight after we detect that user has updated the weight manually, `isPackageWeightEdited` is never set to false. This is also the reason why I use this variable to filter the stream. It would not make sense to make it a publisher because I only want to use it for the condition check. There has to be a better way for this but I'm not sure how.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
pmusolino: Instead of using `initialTotalWeight`, we can check if `isPackageWeightEdited` is `false`, or I'm wrong? In this case, we are sure that the initial total weight never changed, on the contrary, if there is a change, we will save the new total weight and we will set `isPackageWeightEdited` to `true` (like the current logic).

itsmeichigo: The reason why I think `initialTotalWeight` is necessary is in [this discussion](https://github.com/woocommerce/woocommerce-ios/pull/4658#discussion_r673653542). Basically we have to compare the input weight with the calculated one to see if user has manually edited the weight - does that make sense?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
itsmeichigo: @koke  I have updated `configureTotalWeight` with `initialTotalWeight` as input and use it in a separate stream in 30b50093a90dfa27b3f515ef8cb387a3dd1f65db - please let me know what you think about this. (I was distracted so the commit message doesn't convey all the idea)

koke: If I'm understanding this correctly, you want to peek at the first value emitted, and if it doesn't match an existing initial value, then we don't want to update the values anymore. I think this implementation should work, but using `share` to handle the two cases and the `isPackageWeightEdited` to flag the case where we stop processing values seems a bit hard to follow.

I'd like to suggest leveraging publisher completion for that case, which might make for something that's a bit easier to read. I don't think there's a `completeIf` operator, but we can fake that with `tryMap`/`catch`. 

I'm starting to feel sick from the vaccine so I might have missed something obvious, but here's a totally untested idea:

```swift
private func configureTotalWeights(initialTotalWeight: String?) {
    self.totalWeight = initialTotalWeight ?? ""

    $selectedCustomPackage.combineLatest($products, $productVariations)
        .map { [weak self] (customPackage, products, variations) -> Double in
            self?.calculateTotalWeight(products: products, productVariations: variations, customPackage: customPackage) ?? 0
        }
        .tryMap { numericWeight in
            let calculatedWeight = String(numericWeight)

            // Set the total weight as the initial value if it was input manually;
            // otherwise use the calculated weight.
            if let initialTotalWeight = initialTotalWeight,
               calculatedWeight != initialTotalWeight {
                throw Errors.overridenTotalWeight
            }
            return calculatedWeight
        }
        .catch({ _ in
            Empty()
        })
        .assign(to: &$totalWeight)
}
```

With the error defined:

```swift
private extension ShippingLabelPackageDetailsViewModel {
    enum Errors: Error {
        case overridenTotalWeight
    }
}
```

If I understand correctly how Combine works, sending a completion should prevent `assign` from storing any new values. According to the docs:

> The Subscribers.Assign instance created by this operator maintains a strong reference to object, and sets it to nil when the upstream publisher completes (either normally or with an error).

itsmeichigo: Hi @koke - I like the idea of using error to stop the stream! However my intention is to check with `initialTotalWeight` only after the first calculation. Keeping that check in the middle of the stream causes failure when we have a new calculated weight that is different from the initial value. Separating into 2 streams seem to be safer I think.

As for `isPackageWeightEdited`, since this value is also updated from the view (when user start entering the weight manually), I don't think we should omit it and rely only on `initialTotalWeight`. Instead, in 627e49fc9bf5619964fe78bd6c89f9d8064feaf6 I have made `isPackageWeightEdited` a publisher and observe it to throw an error when needed. WDYT?

koke: Sorry for the delay, I'm not sure why but I didn't get a notification for this comment.

> However my intention is to check with initialTotalWeight only after the first calculation

I looked into this more closely, and I think I finally understand the intention now. I hadn't realized the total package weight was editable in the app (the field didn't seem editable to me).

It seems to me that we don't save explicitly whether the user manually edited the weight, so we have to rely on inspecting whether the saved weight matches the selected package. I understand you only want to check the first time it calculates because selecting a different package could give a different weight result.

Now that I think I understand this better, I think separating it in two streams works and it's the right logic, although I'd go back to my first comment. While testing this, I see we configure the results controllers before calling `configureTotalWeights`, so all the values we need for the first weight calculation are already present on initialization, which means we don't have to wait asynchronously for that to be published. It seems `assign(to:)` tries to immediately get a value, and if you set a breakpoint on the `map` that compares the initial value, you can see that the  `init` method is still in the call stack.

Maybe this initial check doesn't really need combine and can happen synchronously, while we still use combine to update the value (as long as `isPackageWeightEdited` is `false`). Maybe this reads a bit more easily, since it's clearer that the initial value comparison isn't really async:

```swift
    private func configureTotalWeights(initialTotalWeight: String?) {
        let calculatedWeight = String(calculateTotalWeight(products: products, productVariations: productVariations, customPackage: selectedCustomPackage))
        if let initialTotalWeight = initialTotalWeight,
           initialTotalWeight != calculatedWeight {
            isPackageWeightEdited = true
            // No need to observe changes if weight was edited
            return
        }

        $selectedCustomPackage.combineLatest($products, $productVariations)
            .map { [weak self] (customPackage, products, variations) -> Double in
                self?.calculateTotalWeight(products: products, productVariations: variations, customPackage: customPackage) ?? 0
            }
            .combineLatest($isPackageWeightEdited)
            .tryMap { (weight, isEdited) in
                guard !isEdited else {
                    throw Errors.overridenTotalWeight
                }
                return String(weight)
            }
            .catch { _ in Empty() }
            .assign(to: &$totalWeight)
    }
```

On top of that, I thought of a slightly nicer approach to completing the stream when `isPackageWeightEdited` changes, although I haven't tested it. Instead of `tryMap`, we can use `prefix(while:)` on the combined stream, and stop when `isPackageWeightEdited` is true:

```swift
    private func configureTotalWeights(initialTotalWeight: String?) {
       // ...Initial checks...

        $selectedCustomPackage.combineLatest($products, $productVariations)
            .map { [weak self] (customPackage, products, variations) -> Double in
                self?.calculateTotalWeight(products: products, productVariations: variations, customPackage: customPackage) ?? 0
            }
            .map { String($0) }
            .combineLatest($isPackageWeightEdited)
            .prefix(while: { _, isPackageWeightEdited in !isPackageWeightEdited })
            .map { weight, _ in weight }
            .assign(to: &$totalWeight)
    }
```

I also found a new issue while testing this. It seems like, according to the docs, the `onEditingChanged` of `TextField` will be called _"when the user begins editing text and after the user finishes editing text"_, which is when the UI sets `isPackageWeightEdited` as `true`, even if the value hasn't changed. So as soon as you tap on the editable field, it's flagged as manually edited

koke: Also, for reference, it looks like the web actually [persists in its state](https://github.com/Automattic/woocommerce-services/blob/dbeed610f336f8db7995743765dbcd0ff2e18474/client/extensions/woocommerce/woocommerce-services/state/shipping-label/reducer.js#L355) when the weight has been manually changed, so that might be an even simpler alternative: persist `isPackageWeightEdited` in the selected package entity.

itsmeichigo:
Thank you @koke for the thorough review! I figure I should post a separate comment to make it easier to follow.

> Maybe this initial check doesn't really need combine and can happen synchronously

This is a great suggestion! I have updated my version to set `totalWeight` to be the same as `initialTotalWeight` when a manual input is detected in 5961d7bd16398a3518ea1b77003182c0988fac06.

> Instead of tryMap, we can use prefix(while:) on the combined stream, and stop when isPackageWeightEdited is true

This is definitely a better approach since we can get rid of the new error type. Thanks for pointing me to this!

> So as soon as you tap on the editable field, it's flagged as manually edited

Great catch! `onEditingChanged` shouldn't be the correct way to check for manual input then, so I used a new approach in b722d2e1f162c0597d55491a5caf25b7af27e8bb. Essentially my idea is:
1. Create a separate stream for calculated weight. This stream should be completed when a manual input is detected so that we can complete all of its subscriptions whenever possible.
2. Any calculated weight can then be used to update `totalWeight`.
3. Observe changes of `totalWeight` considering the latest calculated weight. If the total weight  is different from the calculated weight, we can safely assume that it was manually input.

To implement the 3rd point, I had to add a new operator `withLatestFrom`, which is a [Combine equivalent of the RxSwift operator](https://medium.com/@sergebouts/combine-withlatestfrom-operator-8c529e809fd3). This operator combines every elements emitted by a publisher with the last element emitted by the other:


I'm not sure if this is over-engineering (since it can be hard to follow for folks not familiar with reactive programming), but I think I've tried my best to present my intention. Please let me know what you think about this.

koke (approved):
I think this should work fine, although I had to admit that it took me a while (and a couple drawings) to fully understand all things in motion. Just to make sure I understood correctly, `withLatestFrom` is required because when `calculatedWeight` emits a value the result hasn't been assigned to `totalWeight` yet, and so they will briefly differ. Otherwise we could use `combineLatest` right?

One thing I'd suggest if we are introducing a new operator is adding unit tests and some extra documentation/examples, even if it's linking to something like the post you shared, or another Rx implementation. The marble diagram was really helpful understanding this one.

PR #4680, opened 2021-07-27
Order Detail: "email note to customer" option is now hidden if no email is available

### LINKED ISSUE #3840: Order Details: do not offer the "email note to customer" option if no email is available
Steps to reproduce:

1. Create an order that does not have an email included.
1. Open the oder in the app.
1. Tap "Add a note".
1. Observe you are able to toggle the "Email note to customer" option even if there is no email on file.

**Result:** if you try using the "Email note to customer" option if there is no email saved as part of the order details then the email will fail silently—we should not show the email toggle if there is no email on file to send the message to.

image
<sup>Tested with WCiOS 6.2.0.0 TestFlight beta on iPhone 6S iOS 13.7.</sup>

pmusolino:
Fixes #3840 

## Description
if you try using the "Email note to customer" option if there is no email saved as part of the order details then the email will fail silently—we should not show the email toggle if there is no email on file to send the message to.

## Testing
1. Create an order that does not have an email included.
2. Open the order in the app.
3. Tap "Add a note".
4. Observe that the section "Email note to customer" option is hidden.
5. Try the same procedure with an order with email included. The section should be visible.

## Screenshots
| Order with Customer's email             |  Order without Customer's email |
Simulator Screen Shot - iPhone 12 Pro - 2021-07-27 at 12 16 30 | Simulator Screen Shot - iPhone 12 Pro - 2021-07-27 at 12 24 54

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
@@ -105,10 +105,12 @@ private extension NewNoteViewController {
     private func loadSections() {
         let writeNoteSectionTitle = NSLocalizedString("WRITE NOTE", comment: "Add a note screen - Write Note section title")
         let writeNoteSection = Section(title: writeNoteSectionTitle, rows: [.writeNote])
-        var emailCustomerSection: Section? = nil
-        if viewModel.order.billingAddress?.hasEmailAddress == true {
-            emailCustomerSection = Section(title: nil, rows: [.emailCustomer])
-        }
+        let emailCustomerSection: Section? = {
+            if viewModel.order.billingAddress?.hasEmailAddress == true {
+                return Section(title: nil, rows: [.emailCustomer])
+            }
+            return nil
+        }()
 
         sections = [writeNoteSection, emailCustomerSection].compactMap { $0 }
     }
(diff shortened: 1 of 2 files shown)


itsmeichigo (approved):
Thanks for fixing this! It works as expected, I just have a small comment below 👇

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Add Order Note/NewNoteViewController.swift
```diff
@@ -105,9 +105,12 @@ private extension NewNoteViewController {
     private func loadSections() {
         let writeNoteSectionTitle = NSLocalizedString("WRITE NOTE", comment: "Add a note screen - Write Note section title")
         let writeNoteSection = Section(title: writeNoteSectionTitle, rows: [.writeNote])
-        let emailCustomerSection = Section(title: nil, rows: [.emailCustomer])
+        var emailCustomerSection: Section? = nil
+        if viewModel.order.billingAddress?.hasEmailAddress == true {
+            emailCustomerSection = Section(title: nil, rows: [.emailCustomer])
+        }
```
itsmeichigo: So instead of using a variable, how about using a computed constant instead? Like so: 
```Swift
let emailCustomerSection: Section? = {
    if viewModel.order.billingAddress?.hasEmailAddress == true {
        return Section(title: nil, rows: [.emailCustomer])
    }
    return nil
}()
```
I think it would read better that way 🤔 WDYT?

pmusolino: Definitely better! 🙌

PR #4682, opened 2021-07-27
[Mobile Payments] Remove direct access to cell views

### LINKED ISSUE #3771: [Mobile Payments] Change table row view outlets to private
and pass values via a `configure` method

Affected classes
- ConnectedReaderTableViewCell
- ImageTableViewCell
- LearnMoreTableViewCell
- NumberedListItemTableViewCell
- TitleTableViewCell

Props @ctarda  https://github.com/woocommerce/woocommerce-ios/pull/3761#discussion_r585587025

allendav:
Resolves #3771 

Changes:
- Added `configure` methods to the table cell views accepting either a single value, or a viewmodel for mutliple values - to bring the card reader table cell views we added to harmony with how other table cell views were written
- Removed `TitleTableViewCell` entirely as it wasn't being used anywhere
- Added `UIImage.cardReaderConnect` and a test for it to simplify setting up ImageTableViewCells

To test:
- No visual changes should result from these changes. Connect and disconnect a card reader and ensure all text and images are as expected. Ensure the "learn more" link continues to open the URL as before.
- Ensure the newly added "icon" test for cardReaderConnect image passes

Update release notes:


itsmeichigo (commented):
Thanks for migrating the cell configuration, the code looks much neater now 🎉
I have a small comment below as I think I spotted something 👀 . But I don't have a card reader yet so I can't really test the UI, maybe we'll need an extra pair of eyes to help with testing?

WooCommerce/Classes/ViewRelated/ReusableViews/LearnMoreTableViewCell.swift
```diff
@@ -1,12 +1,37 @@
...
+        configureCell()
+    }
+
+    private func configureCell() {
         learnMoreButton.setImage(.infoOutlineImage, for: .normal)
+        learnMoreTextView.tintColor = .textLink
+        learnMoreTextView.linkTextAttributes = [
+            .foregroundColor: UIColor.textLink,
+            .underlineColor: UIColor.clear
+        ]
```
itsmeichigo, line 22: I think the delegate for `learnMoreTextView` is missing here, WDYT? 🤔

allendav: I don't think so? `configureCell` is just about setting things up from the storyboard. Setting the text and delegate callback is supported through the public `configure` method and protocol conformance below

allendav: Oh! 🤦 I see now. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4682/commits/34b050807c24a9aa88bf8df35885ea1e20921860 - I also noticed that with this fix we are getting the sheet style presentation (the un-delegated behavior uses Safari - which I didn't appreciate was due to the un-delegated behavior)

itsmeichigo:
Thanks @allendav for the update, code looks good to me now 👍 
However I can't approve this since I don't have the necessary hardware to test so I will ask someone in the team to double-check.

ealeksandrov (approved):
Tested with reader - all works well :shipit:

I've found a minor concern about "learn more", unrelated to this PR, filed in https://github.com/woocommerce/woocommerce-ios/issues/4695.

PR #4683, opened 2021-07-28
Fix-4579: Wrong product price in Order Detail if the price of the product is negative

### LINKED ISSUE #4579: Wrong product price in Order Detail if the price of the product is negative
From the web but also from mobile, we allow store owners to set a negative price for their products.

Unfortunately, the behavior within the detail of an order that contains a product with a negative price is not aligned.
The price of a product is always shown as positive, and this makes the order total seemingly incorrect.
The payment summary is correct and aligned to the web.

Related discussion on the web: https://github.com/woocommerce/woocommerce/issues/25037

iamgabrielma:
Fixes #4579 

## Description

Currently the `Order View`, `Order Details View`, and `Refund View` show a positive price number when a negative price is set for the product.

As per this [open issue](https://github.com/woocommerce/woocommerce/issues/25037) we want to allow negative numbers, mainly due to plugins that rely on this in order to function properly, meaning that showing a positive or a zero price in the UI when this is actually negative is incorrect.

## Changes

The problem comes from the Order initializer in `ProductDetailsCellViewModel.swift`, where we're setting `positiveTotal`, and `positivePrice` parameters to be absolute-numbers-only via the `abs()` method:

- Removed the `abs()` method from these parameters.
- Renamed the parameters from `positiveTotal` to `total`, and from `positivePrice` to `price`

The price shows as expected now:

fixed-4579 Order details view

## Testing steps
1. A negative-priced product can be created via the website, the app currently does not allow to create negative-priced products.
2. Fill an order with negative-priced product/s
3. From the app: Tap on Orders > Order > See Order details

## Work in progress / Questions:

- [x] At the moment is not possible to add a product with a negative price via the app, only the web. Created new bug report here: https://github.com/woocommerce/woocommerce-ios/issues/4692

Update release notes:


allendav:
What plugins do this? I cannot fathom the use case.

iamgabrielma:
> What plugins do this? I cannot fathom the use case.

I haven't found many specific cases, but the `Quickbooks for WooCommerce` integration is an example:

> My client is offering a free one-year membership, but they need this to show up in QuickBooks as the sale of a positively-priced membership product AND a negatively-priced membership gift product for accounting purposes. 

As per [this comment](https://github.com/woocommerce/woocommerce/issues/25037#issuecomment-554072804) it seems that's the direction WooCommerce core intends to go moving forward and allow negative prices for all flows, unfortunately I couldn't find the link to the original conversation.

allendav:
> At the moment is not possible to add a product with a negative price via the app, only the web. Should this be part of this PR as well?

I'd open a separate issue or two. We'll need to cover updating the price of products with a negative price as well as creating products on the app with a negative price

allendav:
Tapping into refund detail shows one more spot you need to consider.

This is for an order with coupon and tax (and negative and positive cost items). I refunded 2 of the positive price products and the only negative price product:

neg-core

Admittedly core isn't blameless here (they show a double negative sign)

With this PR, when I tap into the refund detail on this order I see



I think we need a minus sign on one of the products for that subtotal to make mathematical sense

iamgabrielma:
Thanks! Opened new issue here: https://github.com/woocommerce/woocommerce-ios/issues/4692 to deal with the ability to add negative-priced products via the app.

> With this PR, when I tap into the refund detail on this order I see. I think we need a minus sign on one of the products for that subtotal to make mathematical sense

~~I'm seeing in your screenshot that the `A simple product` is not showing the minus sign in `QTY`, neither the `A product with negative price` is showing the negative price within the `subtitle`,  maybe I missed the steps you followed to land on that screen? I'm checking the Refund details view after an order has already been refunded.~~

~~Here's how the refund screen shows on my end when there's both a negative-priced and a positive-priced in the order and both are refunded, these will show with the opposite sign under `QTY` and the `+/- price` will appear within the `subtitle`~~

Screenshot 2021-07-29 at 12 36 21 - refunded order with negative products

Edit: My mistake, I was checking it with the changes on my [first commit](2806049df617c1874661f48fedc23836dd1e2a67), not on the [second commit](2f388d19449b1ea61b227d557b63b522a4516753).

As you'll see in the first commit initially I contemplated to remove the `abs()` method from the `OrderItemRefund` initializer as well, precisely to show on-screen what you're pointing to, but this opened a new range of new problems with negative prices and double-negative-sign refunds, current unit tests would fails as well as this case is not expected, for example:

`test_initializing_with_OrderItemRefund_with_empty_sku() - ProductDetailsCellViewModelTests -> XCTAssertEqual failed: ("-$18.00") is not equal to ("$18.00") `

Tests here will need some changes as well as we're contemplating that this will always be negative ( up until now! )

Then what happens at this point if we trigger the refund depends on 2 cases:

* **Case 1**: The total value of the refund is `> 0` despite having negative-priced products. In this case the refund will go through normally.
* **Case 2**: The total value of the refund is `< 0` . In this case the refund will not work and will throw an error:

An example of case 2 both In the web:

Screenshot 2021-07-29 at 12 30 19- invalid refund amount web


And in the app:

Screenshot 2021-07-29 at 12 31 08

I'll keep digging on different cases taking into account coupons and taxes as well and update the PR shortly.

iamgabrielma:
Some case scenarios showing the changes for different setups:

* Case of a refund of negative-priced products, when we add shipping fees only the fees are refundable if these are higher than the order total :

case 1 pre-refund

* Case of an order with both positive-priced, negative-priced products, and coupons:

case 2 - negative price and positive prices, total positive

* Case of a refund for negative-priced products where the order total is < 0, this creates a positive refund ( will refund 0 ) :

case 1 post-refund, confusing but mathematically correct

* Case of an order with both positive-priced, and negative-priced products, where the order total is > 0, this creates a negative refund ( a normal refund, will refund the difference between the negative-priced-products and the positive-priced-products ) :

case 2 - post-refund 2

Changes in latest new commit:

* Removed `.abs()` method from the refunded Order Item initializer in `ProductDetailsCellViewModel.swift` , this way the Order can return a positive or negative number.
* Following the rationale in the previous point, changed the `OrderItemRefund` existing test as the expected value  for `viewModel.total` and `viewModel.subtitle` will not only return an absolute number anymore.
* Added 2 new tests to check if OrderItemRefund can go either positive or negative.

allendav:
Thanks @iamgabrielma - do you know if these latest changes match core's behavior for these cases?

iamgabrielma:
> Thanks @iamgabrielma - do you know if these latest changes match core's behavior for these cases?

Yes, they do. All of them fit the what I can see in the order details via the website, except the part where WooCommerce core  uses a double negative sign `- -` to express the positive refunds, which I reported here: https://github.com/woocommerce/woocommerce/issues/30400 as a new bug.

Examples of these orders from wp-admin:

Screenshot 2021-07-31 at 09 05 23

Screenshot 2021-07-31 at 09 14 58

Screenshot 2021-07-31 at 09 14 23

Ecarrion:
Hi @iamgabrielma and @allendav! Do we need any help on something to move this forward?

iamgabrielma:
> Hi @iamgabrielma and @allendav! Do we need any help on something to move this forward?

All is good from my end, just pushed the RELEASE-NOTES.txt addition.

allendav:
I'll test today

allendav:
Tested this just now. Here's some concerns:

When a coupon is applied to a cart with both positive price and a negative price items, the product total doesn't reflect the total of the product line items above in the UI, since a portion of the coupon is applied to each line item up there. (If you take the discount line into account, it all adds up correctly, but it does seem confusing)

In this example, quantity 3 of a $1 product and quantity 1 of a $-2 product with a $1 off coupon:



That said, feel free to open a separate issue for design for this.

allendav:
The next concern was that I can attempt to refund a negative amount. This ultimately fails:

(FWIW, the taxes on my store are set at 100%)









Feel free to open a separate issue for this as well

allendav (approved):
Product prices are displayed correctly for products with both negative and positive prices, including refund flows. I did note a few concerns in the conversation that could be worthy of separate issues/fixes. :shipit:

rachelmcr:
I resolved a merge conflict in the Release Notes and enabled auto-merge on the PR (since @iamgabrielma doesn't have permissions to merge in this repo).

PR #4689, opened 2021-07-28
[CPP Onboarding] Add initial unit tests for onboarding state calculation

### LINKED ISSUE #4630: [CPP Onboarding] Add unit tests for state calculation
`checkOnboardingState` is going to contain most of the logic for calculating the onboarding state. We should have unit tests covering that

_Originally posted by @ctarda in https://github.com/woocommerce/woocommerce-ios/pull/4615#discussion_r670483409_

koke:
Fixes #4630 

So far there's not a lot of logic to test, as we are only checking if everything is set up correctly, and showing a generic error otherwise (see #4613).

This PR sets up some initial tests, and adds the necessary helpers to mock payment accounts, so it's easier to add more tests as we implement more of the functionality

## To test

Make sure unit tests pass

Update release notes:

### DIFF
--- a/Fakes/Fakes/Fakes.generated.swift
+++ b/Fakes/Fakes/Fakes.generated.swift
@@ -471,6 +471,25 @@ extension PaymentGateway {
         )
     }
 }
+extension PaymentGatewayAccount {
+    /// Returns a "ready to use" type filled with fake values.
+    ///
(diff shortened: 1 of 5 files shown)


allendav (approved):
Tests pass. Changes make sense :shipit:

PR #4691, opened 2021-07-28
[Mobile Payments] Add card reader ID to payment intent

### LINKED ISSUE #4621: [Mobile Payments] Add the card reader serial number (ID) to the payment intent meta data
We are charged for each reader that has processed 3 or more authorizations in a calendar month. To help us validate that assessment, we should add the card reader serial number (e.g. "CBH204909005931") to the payment intent meta data.

This will allow our back end to identify which readers have processed authorizations for a given calendar month.

cc @ctarda @bmccotter @laurendavissmith

allendav:
Closes #4621 

Changes:
- Adds the connected card reader ID to the payment intent metadata
- This data will be used by our backend to correctly assess the number of readers actually used in a given month
- For simplicity's sake, and to have the insertion of the data as close to the consumer of the data, I elected to implement this wholly in StripeCardReaderService by appending the reader_ID to the metadata passed in instead of a more complex solution that would require modifications to have PaymentCaptureOrchestrator pass the reader ID into PaymentIntent initMetadata

To test:
- Complete a charge and then open it in our dashboard. Confirm the expected reader ID is present in the metadata, i.e.:

readerid

See also:
- https://github.com/woocommerce/woocommerce-android/pull/4470
- fyi @kidinov 

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -383,7 +383,7 @@ private extension StripeCardReaderService {
             return nil
         }
 
-        return connectedReaders[0].id
+        return connectedReaders.first?.id
     }
 
     func createPaymentIntent(_ parameters: PaymentIntentParameters) -> Future<StripeTerminal.PaymentIntent, Error> {
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -375,6 +375,17 @@ extension StripeCardReaderService: CardReaderService {
 
 // MARK: - Payment collection
 private extension StripeCardReaderService {
+    /// Returns the id of the connected reader, if any
+    ///
+    func readerIDForIntent() -> String? {
+        let connectedReaders = connectedReadersSubject.value
+        guard connectedReaders.count == 1 else {
+            return nil
+        }
+
+        return connectedReaders[0].id
+    }
+
     func createPaymentIntent(_ parameters: PaymentIntentParameters) -> Future<StripeTerminal.PaymentIntent, Error> {
         return Future() { [weak self] promise in
             // Shortcircuit if we have an inconsistent set of parameters
@@ -382,6 +393,10 @@ private extension StripeCardReaderService {
                 promise(.failure(CardReaderServiceError.intentCreation()))
                 return
             }
+
+            /// Add the reader_ID to the request metadata so we can attribute this intent to the connected reader
+            ///
+            parameters.metadata?[Constants.readerIDMetadataKey] = self?.readerIDForIntent()
 
             Terminal.shared.createPaymentIntent(parameters) { (intent, error) in
                 if let error = error {
@@ -547,6 +562,18 @@ private extension StripeCardReaderService {
 private extension StripeCardReaderService {
     func internalError(_ error: Error) {
         // Empty for now. Will be implemented later
+    }
+}
+
+// MARK: - Constants
+//
+private extension StripeCardReaderService {
+    enum Constants {
+        /// Used to decorate the payment intent with the reader ID so that we can correctly count
+        /// the number of active readers for a store for a given time period. This key is also used
+        /// by the Android app.
+        ///
+        static let readerIDMetadataKey = "reader_ID"
     }
 }
 


ealeksandrov:
@allendav how do I get to Stripe dashboard for my test store? I don't see link and UI like in [existing public doc](https://docs.woocommerce.com/document/payments/faq/access-stripe-account/).
(WP 5.8 + WC 5.5.2 + WCPay 2.7.1)

allendav:
@ealeksandrov wrote:

> @allendav how do I get to Stripe dashboard for my test store?

Sent instructions in slack

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -375,6 +375,17 @@ extension StripeCardReaderService: CardReaderService {
...
 private extension StripeCardReaderService {
+    /// Returns the id of the connected reader, if any
+    ///
+    func readerIDForIntent() -> String? {
+        let connectedReaders = connectedReadersSubject.value
+        guard connectedReaders.count == 1 else {
+            return nil
+        }
+
+        return connectedReaders[0].id
```
ealeksandrov: nit: I prefer safer access where it's possible, just as a habit (this case has already safety check in guard)
```suggestion
        return connectedReaders.first?.id
```

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -375,6 +375,17 @@ extension StripeCardReaderService: CardReaderService {
 
 // MARK: - Payment collection
 private extension StripeCardReaderService {
+    /// Returns the id of the connected reader, if any
+    ///
+    func readerIDForIntent() -> String? {
+        let connectedReaders = connectedReadersSubject.value
+        guard connectedReaders.count == 1 else {
```
ealeksandrov, line 382: Q: So is there a case of many readers where we shouldn't pass any IDs? Does it mean that we're not sure which one of them just used, so we prefer "no data" instead of "possibly wrong data"?

ealeksandrov (approved):
LGTM, I can see reader ID in Stripe dashboard :shipit:
Left a nitpick and question about case for many readers, but that's just lack of my knowledge for readers flow.

image

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
koke: We probably should convert `connectedReader` from an array to an optional at some point. From what I understand, only one reader can be connected at a time

allendav: > We probably should convert connectedReader from an array to an optional at some point. From what I understand, only one reader can be connected at a time

We should. I can't remember why we made it an array.

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
allendav: Good idea. Will do

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
allendav: > So is there a case of many readers where we shouldn't pass any IDs? Does it mean that we're not sure which one of them just used, so we prefer "no data" instead of "possibly wrong data"?

The latter. I didn't want to abort the payment intent in the unlikely event that we think we have no connectedReader. This is just defensive coding that is unlikely to be triggered by real world use.

Only one reader can be connected at a time.

allendav: > So is there a case of many readers where we shouldn't pass any IDs? Does it mean that we're not sure which one of them just used, so we prefer "no data" instead of "possibly wrong data"?

The latter. I didn't want to abort the payment intent in the unlikely event that we think we have no connectedReader. This is just defensive coding that is unlikely to be triggered by real world use.

Only one reader can be connected at a time.

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4691/commits/8264301722f71d678d90b84913e368c0a1c375e2

PR #4698, opened 2021-07-30
Shipping Labels: Make it possible to manually input state for countries with no states

### LINKED ISSUE #4525: Shipping Labels: Empty Table View for States for Orders with Empty "Ship to" Address
Observed while testing #4523 

If you create an order in wp-admin and skip providing Bill To and Ship To addresses, i.e.:

Screen Shot 2021-07-01 at 3 37 26 PM

When you enter the shipping label flow in the mobile app and attempt to set the state, the table view is empty:

stateless

itsmeichigo:
Closes #4525 

# Description
Currently on address validation form, tapping on State navigates to an empty list if no state is found for specified country. This PR fixes that by letting user enter state manually as needed.

# Changes
- Updated State cell to enable text field if no state for the specified country is found.
- Updated tap action of State cell to only navigate to state list if the list is not empty.

# Demo
country-state

# Testing
1. Navigate to Orders tab, select an order with Processing status.
2. Select Create Shipping Label and skip to configure Ship To address.
3. Select Australia / US as country, then tap State and notice that you're navigated to state list to pick one.
4. Select Vietnam, then tap State and notice that you'll have to enter name of the state (or keep it empty).

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -266,7 +266,7 @@ extension ShippingLabelAddressFormViewController: UITableViewDelegate {
         switch row {
         case .state:
             let states = viewModel.statesOfSelectedCountry
-            guard !states.isEmpty else {
+            guard states.isNotEmpty else {
                 return
             }
             let selectedState = states.first { $0.code == viewModel.address?.state }
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -527,8 +527,10 @@ private extension ShippingLabelAddressFormViewController {
         static let postcodeFieldPlaceholder = NSLocalizedString("Required", comment: "Text field placeholder in Shipping Label Address Validation")
         static let stateField = NSLocalizedString("State", comment: "Text field state in Shipping Label Address Validation")
         static let stateFieldPlaceholder = NSLocalizedString("Required", comment: "Text field placeholder in Shipping Label Address Validation")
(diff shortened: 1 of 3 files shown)


koke:
Looking at the web implementation, the State field only seems to be required in countries that have a known state list. Should this PR include this logic as well?

pmusolino (commented):
As said by @koke, it seems that the State is not mandatory but becomes optional if the Country doesn't have a State list. I think that we should update the behavior also on our side

itsmeichigo:
Thank you @koke and @pmusolino for pointing that out! I have pushed some updates to check whether state is required:
country-state-updated

Please take another look 🙇‍♀️

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -266,6 +266,9 @@ extension ShippingLabelAddressFormViewController: UITableViewDelegate {
         switch row {
         case .state:
             let states = viewModel.statesOfSelectedCountry
+            guard !states.isEmpty else {
```
pmusolino: nit: here you can use directly `isNotEmpty` 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
```diff
@@ -52,6 +52,10 @@ final class ShippingLabelAddressFormViewModel {
         countries.first { $0.code == address?.country }?.states.sorted { $0.name < $1.name } ?? []
     }
 
+    var stateOfCountryRequired: Bool {
+        !statesOfSelectedCountry.isEmpty
```
pmusolino:
```suggestion
        statesOfSelectedCountry.isNotEmpty
```

pmusolino (approved):
LGTM 🚢 


Code ✅ just some minor nits
Testing ✅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
itsmeichigo: Thanks for the suggestion! I didn't know about this before 👍

PR #4702, opened 2021-07-30
[CPP Onboarding] Extract onboarding logic to separate file for better testability

koke:
Part of #4611 

This refactors the logic in #4689 and the tests from #4689 so that all the logic is in a separate `CardPresentPaymentsOnboardingUseCase` file instead of `CardPresentPaymentStore`, so it's easier to test.

## To test

Go to Settings > In-Person Payments:

1. With a store correctly set up for payments, you should see the option to connect to a reader
2. With a store that's not set up for payments, you should see a generic error

Note that this new implementation checks that the WCPay plugin is installed, but it doesn't manage refreshing the site plugins yet, so it might not work correctly if nothing else in the app has refreshed that data yet. I will look into that next.

Update release notes:


rachelmcr (approved):
Thanks for this note:

> Note that this new implementation checks that the WCPay plugin is installed, but it doesn't manage refreshing the site plugins yet, so it might not work correctly if nothing else in the app has refreshed that data yet.

I had to visit Settings > Plugins first to make sure my site's plugins were saved in storage; after that, this worked as expected in both scenarios (a store set up for payments and a store that isn't).

Code changes look good and it's great to see improvements for testability! :shipit:

koke:
Thanks for the review and the extra testing 😁

PR #4703, opened 2021-07-30
[Tooling] Fix hotfix lanes

AliSoftware:
Part of the paaHJt-1WQ-p2 project.

_Note: In the case of WCiOS, not much has been changed on the hotfix lanes as part of this PR (except the lane's description to make them more descriptive), but as part of this project, it's still worth going through the test scenario below to ensure that hotfix were already working as expected (and didn't break in the past but got unnoticed)_

## To test

 - Checkout this branch, run `bundle install`
 - Run `bundle exec fastlane new_hotfix_release version:7.0.1`
 - Verify that it cut a `release/7.0.1` branch from the `7.0` tag
 - Verify that it bumped the version in `Version.internal.xcconfig`, `Version.public.xcconfig` and `fastlane/Deliverfile` (*)
 - cherry-pick all the commits from this PR on top of the `release/7.0.1` branch. This is because we are gonna test the updated finalize lanes, and need to be on the hotfix release branch to test them. (You might get conflicts during the cherry-pick)
 - [Open the CircleCI page for WCiOS](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios) in advance, to be ready to cancel the release build
 - Run `bundle exec fastlane finalize_release`. You should get an error telling you that you should use `finalize_hotfix_release` for hotfix branches instead
 - Run `bundle exec fastlane finalize_hotfix_release` and verify that it triggered a release build, and that is is building it from the release/7.0.1 hotfix branch. Cancel the CI build immediately.
 - Delete the `release/7.0.1` branch (from both local and remote)

## A note on Deliverfile bump

@mokagio is working on removing the `Deliverfile` – see https://github.com/wordpress-mobile/release-toolkit/pull/287 – in our repos, to run `deliver` as part of a lane to which we'll explicitly pass the right parameters, including the version, hence the bump not needing to change the app version in `Deliverfile` anymore in the future.

This means that once @mokagio's work lands, the hotfix lanes will need to be amended again, to specify that the `Deliverfile` should not be updated as part of any version bump… including hotfix ones. This work will be done separately to the hotfix alignment work from this PR, to avoid postponing things too much and blocking the project.

### DIFF
--- a/.bundle/config
+++ b/.bundle/config
@@ -1,5 +1,5 @@
 ---
 BUNDLE_RETRY: "3"
 BUNDLE_PATH: "vendor/bundle"
-BUNDLE_JOBS: "3"
+BUNDLE_JOBS: "16"
 BUNDLE_WITHOUT: "screenshots"
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -197,7 +197,7 @@ platform :ios do
   # bundle exec fastlane new_hotfix_release version:10.6.1
   # bundle exec fastlane new_hotfix_release skip_confirm:true version:10.6.1
   #####################################################################################
-  desc 'Creates a new hotfix branch from the given tag'
+  desc 'Creates a new hotfix branch for the given version:x.y.z. The branch will be cut from the tag x.y of the previous release'
   lane :new_hotfix_release do |options|
     prev_ver = ios_hotfix_prechecks(options)
     ios_bump_version_hotfix(previous_version: prev_ver, version: options[:version])
@@ -214,7 +214,7 @@ platform :ios do
   # Example:
   # bundle exec fastlane finalize_hotfix_release skip_confirm:true
   #####################################################################################
-  desc 'Creates a new hotfix branch from the given tag'
+  desc 'Performs the final checks and triggers a release build for the hotfix in the current branch'
   lane :finalize_hotfix_release do |options|
     ios_finalize_prechecks(options)
     version = ios_get_app_version
@@ -236,24 +236,25 @@ platform :ios do
   #####################################################################################
   desc 'Trigger the final release build on CI'
   lane :finalize_release do |options|
+    UI.user_error!('To finalize a hotfix, please use the finalize_hotfix_release lane instead') if ios_current_branch_is_hotfix
+
     ios_finalize_prechecks(options)
-    unless ios_current_branch_is_hotfix
-      UI.message('Checking app strings translation status...')
-      check_translation_progress(
-        glotpress_url: 'https://translate.wordpress.com/projects/woocommerce/woocommerce-ios/',
-        abort_on_violations: false
-      )
 
-      UI.message("Checking release notes strings translation status...")
-      check_translation_progress(
-        glotpress_url: 'https://translate.wordpress.com/projects/woocommerce/woocommerce-ios/release-notes/',
-        abort_on_violations: false
-      )
+    UI.message('Checking app strings translation status...')
+    check_translation_progress(
+      glotpress_url: 'https://translate.wordpress.com/projects/woocommerce/woocommerce-ios/',
+      abort_on_violations: false
+    )
 
-      ios_update_metadata(options)
-      ios_lint_localizations(input_dir: 'WooCommerce/Resources', allow_retry: true)
-      ios_bump_version_beta
-    end
+    UI.message("Checking release notes strings translation status...")
+    check_translation_progress(
+      glotpress_url: 'https://translate.wordpress.com/projects/woocommerce/woocommerce-ios/release-notes/',
+      abort_on_violations: false
+    )
+
+    ios_update_metadata(options)
+    ios_lint_localizations(input_dir: 'WooCommerce/Resources', allow_retry: true)
+    ios_bump_version_beta
 
     # Wrap up
     version = ios_get_app_version
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -89,10 +89,11 @@ platform :ios do
 
     ios_bump_version_release
     new_version = ios_get_app_version
-    extract_release_notes_for_version(version: new_version,
-                                      release_notes_file_path: 'RELEASE-NOTES.txt',
-                                      extracted_notes_file_path: File.join('WooCommerce',
-                                                                           'Resources', 'release_notes.txt'))
+    extract_release_notes_for_version(
+      version: new_version,
+      release_notes_file_path: 'RELEASE-NOTES.txt',
+      extracted_notes_file_path: File.join('WooCommerce', 'Resources', 'release_notes.txt')
+    )
     ios_update_release_notes(new_version: new_version)
     get_prs_list(repository: GHHELPER_REPO, milestone: new_version,
                  report_path: File.join(File.expand_path('~'), "wcios_prs_list_#{old_version}_#{new_version}.txt"))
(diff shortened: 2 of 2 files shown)


.bundle/config
```diff
@@ -1,5 +1,5 @@
 ---
 BUNDLE_RETRY: "3"
 BUNDLE_PATH: "vendor/bundle"
-BUNDLE_JOBS: "3"
+BUNDLE_JOBS: "16"
```
mokagio, line 4: That's quite the bump!

Just curious: where does 16 come from? Why not a different value?

fastlane/Fastfile
```diff
@@ -89,10 +89,11 @@ platform :ios do
...
     new_version = ios_get_app_version
-    extract_release_notes_for_version(version: new_version,
-                                      release_notes_file_path: 'RELEASE-NOTES.txt',
-                                      extracted_notes_file_path: File.join('WooCommerce',
-                                                                           'Resources', 'release_notes.txt'))
+    extract_release_notes_for_version(
+      version: new_version,
+      release_notes_file_path: 'RELEASE-NOTES.txt',
+      extracted_notes_file_path: File.join('WooCommerce', 'Resources', 'release_notes.txt')
+    )
```
mokagio, line 96: Love a good alignment fix

fastlane/Fastfile
```diff
@@ -135,11 +136,11 @@ platform :ios do
   #####################################################################################
   desc 'Updates the AppStoreStrings.pot file with the latest data'
   lane :update_appstore_strings do |options|
-    prj_folder = Pathname.new(File.join(Dir.pwd, '..')).expand_path.to_s
-    source_metadata_folder = File.join(prj_folder, 'fastlane/appstoreres/metadata/source')
+    project_root = File.dirname(File.expand_path(__dir__))
```
mokagio, line 139: I'm guessing you went for `File` instead of `Pathname` because we don't need to do any path operation here and this saves the awkward `.to_s` call?

fastlane/Fastfile
```diff
@@ -194,7 +197,7 @@ platform :ios do
   # bundle exec fastlane new_hotfix_release version:10.6.1
   # bundle exec fastlane new_hotfix_release skip_confirm:true version:10.6.1
   #####################################################################################
-  desc 'Creates a new hotfix branch from the given tag'
+  desc 'Creates a new hotfix branch for the given version:x.y.z. The branch will be cut from the tag x.y of the previous release'
```
mokagio, line 200: Not sure if it's worth clarifying here, but if we were to do an hotfix on top of another hotfix, e.g. what I did to test https://github.com/wordpress-mobile/release-toolkit/pull/287, the branch would be cut from the tag of the previous hotfix release, which is `x.y.n` instead of `x.y`.

🤷‍♂️

fastlane/Fastfile
```diff
@@ -233,24 +236,25 @@ platform :ios do
...
+    )
+
+    UI.message("Checking release notes strings translation status...")
+    check_translation_progress(
+      glotpress_url: 'https://translate.wordpress.com/projects/woocommerce/woocommerce-ios/release-notes/',
+      abort_on_violations: false
+    )
+
+    ios_update_metadata(options)
+    ios_lint_localizations(input_dir: 'WooCommerce/Resources', allow_retry: true)
```
mokagio, line 256: This PR copied the code from the previous version, which I think is fine.

I wonder if it would be worth following up with using `File.join(PROJECT_ROOT, 'WooCommerce', 'Resources')` (where `PROJECT_ROOT` is the `project_root` defined above, but extracted as a constant).

mokagio (approved):
- [x] Verify that it cut a `release/7.0.1` branch from the `7.0` tag
- [x] Verify that it bumped the version in ~~`Version.internal.xcconfig`,~~ [there's no internal version in Woo] `Version.public.xcconfig` and `fastlane/Deliverfile`
- [x] Run `bundle exec fastlane finalize_release`. You should get an error telling you that you should use `finalize_hotfix_release` for hotfix branches instead
- [x] Run `bundle exec fastlane finalize_hotfix_release` and verify that it triggered a release build, and that is is building it from the release/7.0.1 hotfix branch. Cancel the CI build immediately.

All working as described 👍

.bundle/config
AliSoftware: tbh I have no idea. It suddenly started to change that value on its own when I did a `bundle install`, I didn't change it myself, so I think it's just the new default value for more recent `bundler` versions (I'm on `bundler 2.2.19`)

fastlane/Fastfile
AliSoftware: Yeah I thought about that after the fact. Not sure how to express this in a wording that doesn't end up being too complicated trying to explain all the subtle cases 🤔

fastlane/Fastfile
AliSoftware: Yeah and also because tbh this changed twice while I was working on this:
 - When I started that branch a while ago, there was the `PROJECT_ROOT` top-level constant (around [those lines](https://github.com/woocommerce/woocommerce-ios/blob/218cb5ea100d79a98a031fc273b817e4f676c235/fastlane/Fastfile#L6-L8)), so at that time I just removed the `prj_folder` and then directly used `source_metadata_folder = File.join(PROJECT_ROOT, …)` instead of where `prj_folder` was used.
 - But then your PR to move secrets outside the project got rid of that `PROJECT_ROOT` top-level constant, so when I came back to fix the conflicts on this PR and rebased it, it didn't exist anymore. Instead of re-introducing the top-level constant, I decided to just define it as a local variable as before, just borrowing how it was previously defined as `PROJECT_ROOT`, but keeping it in the scope of this lane only because that's the only place it was used.

I didn't really think twice about the fact that those 2 steps ended up changing a call using `Pathname` into a call just using `File`, but otoh, since we indeed don't really need complex path operations here, I end up also feeling that using `File` methods is nice here anyway 😛 

---

To be honest I am not really sure we need to specify `project_root` in those paths (and provide the toolkit with full, absolute paths) or if the toolkit would work just as well if we just passed `fastlane/appstoreres/metadata/source` as a relative path, just like we pass `WooCommerce/Resources` to `ios_lint_localizations` below. Which, if that works, could mean we'd be able to totally get rid of the project_root variable (or constant) – which would be a nice step in the direction of https://github.com/wordpress-mobile/release-toolkit/issues/205 too btw. But I figured this would be a problem for another PR 😅

fastlane/Fastfile
AliSoftware: See [comment above](https://github.com/woocommerce/woocommerce-ios/pull/4703#discussion_r680886239).

.bundle/config
mokagio: Interesting 🤔 

From the [docs](https://github.com/rubygems/rubygems/blob/faa8cf6cccd2ae18eba7191618ff639cc0ddfd7c/bundler/lib/bundler/man/bundle-config.1.ronn):

> jobs (`BUNDLE_JOBS`): The number of gems Bundler can install in parallel. Defaults to 1 on Windows, and to the the number of processors on other platforms.

There must be some bug, unless you are on some [new secret hardware](https://arstechnica.com/gadgets/2021/05/apple-planning-macbook-pro-with-16-core-gpu-mac-pro-with-128-gpu-cores/)

fastlane/Fastfile
mokagio: >  But I figured this would be a problem for another PR 😅

Sounds good to me 😄 🙌

fastlane/Fastfile
mokagio: Yeah. It's a minutia. Maybe we should spend all this mental energy in making the code and tests so good that we will never need to ship more than one hotfix 🤣 

---

As I mentioned in another PR where I made this comment, this is good for me. 

ship it

.bundle/config
AliSoftware: 😄 if only 😂

Fwiw this change made by `bundle install` when I run it on my machine happens apparently everywhere (ie on all repositories, not just this one). At least that's far from the only time I see it changing that value (and why I started committing it otherwise it kept changing it)

PR #4704, opened 2021-07-30
Errors: Banner message for error loading data on My Store dashboard

### LINKED ISSUE #4382: Errors: Banner message for error loading data on Home screen, Product List, Order List, Review List

**Error:** Plugin conflict, site performance issue (timeout errors due to overloaded servers or outdated versions of PHP)

**Design:** Let’s use our existing banner component to display that message. The banner is collapsed by default and on select, users can read our troubleshooting tips or contact support directly. (If this error is coming up, it should be the only banner displayed. Other banners should not be displayed, e.g. on the Products screen.)

Whenever users select “Troubleshoot” in the error message banner, they’ll load a page on WC docs with a list of things they could try out before reaching out to support.

Screens to update:

- [x] My Store
- [x] Orders
- [x] Products
- [x] Reviews
- [x] Empty state screens (when we have no cached data)

image

rachelmcr:
Resolves: #4382

## Description

To improve our error states, we are introducing a top banner that can be used whenever there is a problem loading data. The banner is collapsed by default and on select, users can read our troubleshooting tips or contact support directly.

This PR sets up the banner and adds it to the top of the My Store screen. (This banner replaces the "Unable to load content" notice we displayed before.)

## Changes

* Adds the error banner to `DashboardViewController`. A quick overview of the changes there:
   * Currently we display the store name in a stack view at the top of the screen, as long as `shouldShowStoreNameAsSubtitle` is `true`. The store name is now embedded inside another stack view, to retain its leading margin. (That margin is now defined with a constant equivalent to the margins it had before.)
   * If `shouldShowStoreNameAsSubtitle` is `false`, the stack view is added when the banner is displayed, with all necessary constraints. (Those constraints are otherwise set when the subtitle is configured and the dashboard UI is added below it).
* The error banner is displayed when there is a syncing error, and hidden when the screen refreshes.

My Store - no errors | My Store - error banner collapsed | My Store - error banner expanded
-|-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-07-30 at 13 14 40|Simulator Screen Shot - iPhone 11 Pro - 2021-07-30 at 13 14 49|Simulator Screen Shot - iPhone 11 Pro - 2021-07-30 at 13 14 51

## Testing

1. Build and run the app.
2. Disconnect your internet connection. (This is so you will trigger a sync error; you can also disrupt your store's Jetpack connection for a more realistic scenario.)
3. Pull to refresh the My Store screen.
4. Notice the error banner appears at the top of the screen, below your store name, and remains on screen when you scroll.
5. Reconnect to the internet.
6. Expand the banner and confirm the "Troubleshoot" button opens a troubleshooting doc, and "Contact support" opens a new support ticket in the app.
7. Pull to refresh or navigate to another tab and back to refresh the My Store screen. Notice the error banner disappears on refresh.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
@@ -53,7 +53,7 @@ final class DashboardViewController: UIViewController {
 
     /// A stack view for views displayed between the navigation bar and content (e.g. store name subtitle, top banner)
     ///
-    private lazy var stackView: UIStackView = {
+    private lazy var headerStackView: UIStackView = {
         let view = UIStackView()
         view.translatesAutoresizingMaskIntoConstraints = false
         view.backgroundColor = .listForeground
@@ -79,6 +79,8 @@ final class DashboardViewController: UIViewController {
                                               })
     }()
 
+    /// A spacer view to add a margin below the top banner (between the banner and dashboard UI)
+    ///
     private lazy var spacerView: UIView = {
         let view = UIView()
         view.heightAnchor.constraint(equalToConstant: Constants.bannerBottomMargin).isActive = true
@@ -149,12 +151,12 @@ private extension DashboardViewController {
         }
         storeNameLabel.text = ServiceLocator.stores.sessionManager.defaultSite?.name ?? Localization.title
         innerStackView.addArrangedSubview(storeNameLabel)
-        stackView.addArrangedSubview(innerStackView)
-        containerView.addSubview(stackView)
+        headerStackView.addArrangedSubview(innerStackView)
+        containerView.addSubview(headerStackView)
         NSLayoutConstraint.activate([
-            stackView.topAnchor.constraint(equalTo: containerView.topAnchor),
-            stackView.leadingAnchor.constraint(equalTo: containerView.leadingAnchor),
-            stackView.trailingAnchor.constraint(equalTo: containerView.trailingAnchor)
+            headerStackView.topAnchor.constraint(equalTo: containerView.topAnchor),
+            headerStackView.leadingAnchor.constraint(equalTo: containerView.leadingAnchor),
+            headerStackView.trailingAnchor.constraint(equalTo: containerView.trailingAnchor)
         ])
     }
 
@@ -164,7 +166,7 @@ private extension DashboardViewController {
         }
         contentView.translatesAutoresizingMaskIntoConstraints = false
         NSLayoutConstraint.activate([
-            contentView.topAnchor.constraint(equalTo: stackView.bottomAnchor),
+            contentView.topAnchor.constraint(equalTo: headerStackView.bottomAnchor),
             contentView.leadingAnchor.constraint(equalTo: containerView.leadingAnchor),
             contentView.trailingAnchor.constraint(equalTo: containerView.trailingAnchor),
             contentView.bottomAnchor.constraint(equalTo: containerView.bottomAnchor)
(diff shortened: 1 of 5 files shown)


itsmeichigo (commented):
Thank you for handling this issue! Overall everything works as expected for happy cases, however I have a few nits and comments below.

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -41,18 +42,50 @@ final class DashboardViewController: UIViewController {
         return label
     }()
 
+    /// A stack view to display `storeNameLabel` with additional margins
+    ///
+    private lazy var innerStackView: UIStackView = {
+        let view = UIStackView()
+        view.layoutMargins = UIEdgeInsets(top: 0, left: Constants.leadingMargin, bottom: 0, right: 0)
```
itsmeichigo: I think it'd better if there's margin for the right inset too - this could cause missing margin for users using right-to-left languages like Arabic:

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -41,18 +42,50 @@ final class DashboardViewController: UIViewController {
...
+        view.isLayoutMarginsRelativeArrangement = true
+        return view
+    }()
+
+    /// A stack view for views displayed between the navigation bar and content (e.g. store name subtitle, top banner)
+    ///
     private lazy var stackView: UIStackView = {
```
itsmeichigo: Do you think it might be necessary to update the name of this property to `mainStackView` or `headerStackView` to make it clearer?

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -41,18 +42,50 @@ final class DashboardViewController: UIViewController {
...
+                                                let safariViewController = SFSafariViewController(url: WooConstants.URLs.troubleshootErrorLoadingData.asURL())
+                                                self?.present(safariViewController, animated: true, completion: nil)
+                                              },
+                                              onContactSupportButtonPressed: { [weak self] in
+                                                guard let self = self else { return }
+                                                ZendeskManager.shared.showNewRequestIfPossible(from: self, with: nil)
+                                              })
+    }()
+
+    private lazy var spacerView: UIView = {
```
itsmeichigo, line 84: I think a comment for this property can be helpful to understand its purpose - WDYT?

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -41,18 +42,50 @@ final class DashboardViewController: UIViewController {
...
+        return view
+    }()
+
+    /// A stack view for views displayed between the navigation bar and content (e.g. store name subtitle, top banner)
+    ///
     private lazy var stackView: UIStackView = {
         let view = UIStackView()
         view.translatesAutoresizingMaskIntoConstraints = false
         view.backgroundColor = .listForeground
-        view.layoutMargins = UIEdgeInsets(top: 0, left: navigationController?.navigationBar.directionalLayoutMargins.leading ?? 0, bottom: 0, right: 0)
```
itsmeichigo, line 48: This line seems to have caused an issue with the margin when app is launched in landscape mode - removing this seems to have fixed the issue. Bravo 🎉

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -178,6 +211,36 @@ private extension DashboardViewController {
...
+                dashboardUI.view.bottomAnchor.constraint(equalTo: containerView.bottomAnchor)
+            ])
+        }
+    }
+
+    /// Hide the error banner
+    ///
+    func hideTopBannerView() {
+        topBannerView.removeFromSuperview()
+        spacerView.removeFromSuperview()
```
itsmeichigo: I think there's no need for adding and removing `topBannerView` and `spacerView` when showing / hiding the banner view. The great thing about stack view is that it can collapse or expand when showing / hiding its subviews. 

Therefore, I'd suggest that you add these as arranged subviews  right after adding the `innerStackView`, and hide them by default. When showing or hiding the banner view, all you need to do is show / hide these subviews and it should still work as expected.

WooCommerce/Classes/ViewRelated/Dashboard/Factories/DashboardUIFactory.swift
```diff
@@ -9,7 +9,7 @@ protocol DashboardUI: UIViewController {
     var scrollDelegate: DashboardUIScrollDelegate? { get set }
 
     /// Called when the Dashboard should display syncing error
-    var displaySyncingErrorNotice: () -> Void { get set }
+    var displaySyncingError: () -> Void { get set }
```
itsmeichigo, line 12: Thanks for renaming this too! 💯

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
rachelmcr: Good catch, thanks! I added a margin for the right inset in cd48e49. I thought about using `directionalLayoutMargins` to add the inset to the leading edge specifically (not left/right) but I realized it would be good to have insets on both sides anyway in case of very long store names.

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
rachelmcr: This is a great approach that I didn't think of, thanks! In 92e6c6c I changed how these views are configured, and it simplifies the way the constraints are set, too. Now the header stack view is configured up front with all the arranged subviews added, and as you suggested the methods to show or hide the banner just show/hide those subviews.

rachelmcr:
Thanks for the review and suggestions, @itsmeichigo! This is ready for another look when you have a chance.

itsmeichigo (approved):
This looks great, thanks for updating! :shipit:

PR #4711, opened 2021-08-02
Shipping Labels: Require phone number for international shipping

### LINKED ISSUE #4686: Shipping Labels: Require a valid phone number for international shipping.
This is a subtask of #4596.

Currently phone number not required because we're only supporting US shipping.

When detecting international shipping, phone number should not be empty for origin address. If this field is empty, we should show an error message for that when tapping either "Done" or "Use address as entered" button.

itsmeichigo:
Closes #4686 

# Description
The PR adds validation for phone number of origin address when international shipping is detected.

# Changes
- Updated `ShippingLabelAddressFormViewModel` with a check for `isInternationalShipping`. This is based on whether destination address is a country outside of the US (country code other than "US").
- Added new param `phoneNumberRequired` for the initializer of `ShippingLabelAddressFormViewController` and its view model.
- Replaced phone error validation case with 2 separate cases for missing and invalid phone number. Each case is handled with a different error message.
- Update local validation of `ShippingLabelAddressFormViewModel` to check whether phone number is required and invalid.
- Added unit tests for `isInternationalShipping` and validation of `ShippingLabelAddressFormViewModel`.

# Demo
phone

# Testing
1. Navigate to Orders tab and select an order with status Processing and destination address outside of the US.
2. Select Create Shipping Labels and configure origin address.
3. Notice that if phone number is empty, you cannot proceed and there's a validation error message about a phone number is required.
4. Notice that if you try to proceed with a phone number with length other than 10 (or 11 if it starts with "1" - area code of US), there's a validation error message for requirement of 10-digit phone number.
5. Head back to Orders tab and select an order with destination address in the US. Repeat the same step and notice that phone number is optional for origin address.

# Discussion
I noticed that the phone number length is not validated when shipping within the US (this behavior is confirmed on both the web app and Android app). I figure I should keep the same behavior for the iOS app because general phone validation in the address form will require checking for country code too which can become messy.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -415,8 +415,10 @@ private extension ShippingLabelAddressFormViewController {
             errorMessage = Localization.missingState
         case .country:
             errorMessage = Localization.missingCountry
-        case .phoneNumber:
+        case .missingPhoneNumber:
             errorMessage = Localization.missingPhoneNumber
+        case .invalidPhoneNumber:
+            errorMessage = Localization.invalidPhoneNumber
         }
 
         cell.textLabel?.text = errorMessage
@@ -562,8 +564,11 @@ private extension ShippingLabelAddressFormViewController {
         static let missingCountry = NSLocalizedString("Country missing",
                                                       comment: "Error showed in Shipping Label Address Validation for the country field")
         static let missingPhoneNumber = NSLocalizedString("A phone number is required because this shipment requires a customs form",
-                                                          comment: "Error showed in Shipping Label Origin Address Validation for the " +
+                                                          comment: "Error shown in Shipping Label Origin Address validation for " +
                                                           "phone number field for international shipment")
+        static let invalidPhoneNumber = NSLocalizedString("Custom forms require a 10-digit phone number",
+                                                          comment: "Error shown in Shipping Label Origin Address validation for " +
+                                                            "phone number when the it doesn't have expected length for international shipment.")
         static let appleMapsErrorNotice = NSLocalizedString("Error in finding the address in Apple Maps",
                                                             comment: "Error in finding the address in the Shipping Label Address Validation in Apple Maps")
         static let phoneNumberErrorNotice = NSLocalizedString("The phone number is not valid or you can't call the customer from this device.",
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -60,10 +60,16 @@ final class ShippingLabelAddressFormViewController: UIViewController {
     init(siteID: Int64,
          type: ShipType,
          address: ShippingLabelAddress?,
+         phoneNumberRequired: Bool = false,
          validationError: ShippingLabelAddressValidationError?,
          countries: [Country],
          completion: @escaping Completion ) {
-        viewModel = ShippingLabelAddressFormViewModel(siteID: siteID, type: type, address: address, validationError: validationError, countries: countries)
+        viewModel = ShippingLabelAddressFormViewModel(siteID: siteID,
+                                                      type: type,
+                                                      address: address,
+                                                      phoneNumberRequired: phoneNumberRequired,
+                                                      validationError: validationError,
+                                                      countries: countries)
         onCompletion = completion
         super.init(nibName: nil, bundle: nil)
         switch type {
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
```diff
@@ -165,6 +177,25 @@ extension ShippingLabelAddressFormViewModel {
...
+    /// Note: This logic may need to be updated if there is a need for validating other cases.
+    ///
+    private var isPhoneNumberValid: Bool {
+        guard let phone = address?.phone, phone.isNotEmpty else {
+            return false
+        }
+        if phone.hasPrefix("1") {
+            return phone.count == 11
+        } else {
+            return phone.count == 10
```
koke, line 197: I don't know how important this is, but [the web implementation](https://github.com/Automattic/woocommerce-services/blob/f772b6ed36fad5576582c9b8c5c044a6f0b7d2c2/client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js#L246) seems to discard anything that's not a digit from the count, so you can input the phone number with spaces, dashes, or parentheses. This might seem unnecessary because the keyboard type will only allow numeric inputs, but phone numbers formatted differently can still be pasted:

**Web:**

Screen Shot 2021-08-03 at 08 09 15

**App:**
Screen Shot 2021-08-03 at 08 13 11

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -89,6 +89,15 @@ final class ShippingLabelFormViewModel {
         resultsController.fetchedObjects
     }
 
+    /// Check for international shipping based on whether destination country is other than US.
+    ///
+    var isInternationalShipping: Bool {
```
koke: Looking at the [web implementation](https://github.com/Automattic/woocommerce-services/blob/f772b6ed36fad5576582c9b8c5c044a6f0b7d2c2/client/extensions/woocommerce/woocommerce-services/state/shipping-label/selectors.js#L191), it seems like we should require a phone number for customs if the origin or destination is one of the US Military States.

Other than that, it's checking that the origin and destination country is different. I understand right now we only support shipping from the US, so I don't think there's a practical difference, but it might be more future-proof to have the equality check rather than `destinationCountriesWithoutPhoneRequirement`

koke (approved):
Code looks good and it seems to work fine. I found a couple issues, but I don't know how important they are in the context of the project, so I'll approve and let you decide if we want to solve those in this PR or as a separate issue

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
itsmeichigo: Great catch! I think it's safer to filter out special characters so I've done that in 81cbf65488437efdd2f5f38a352ea1a01786cecf.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
itsmeichigo: Thank you so much for pointing me to the web implementation! I really should look at that to double check my logic - I have duplicated the implementation for extra safety in fbb903f5ff31bcc4c5115e9eba9af6db3818a1eb.

itsmeichigo:
Thank you @koke for the suggestions! I know you already approved this but can you please take another look at my changes to make sure that they're good to go?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -370,7 +370,8 @@ private extension ShippingLabelAddressFormViewController {
                                                                      state: .normal,
                                                                      keyboardType: .phonePad,
                                                                      textFieldAlignment: .leading) { [weak self] (newText) in
-            self?.viewModel.handleAddressValueChanges(row: row, newValue: newText)
+            let phone = newText?.filter { $0.isNumber }
```
koke: I'm getting picky here, but I checked the docs for `isNumber` and it seems to match many kinds of numeric characters, like `⅚` or `㊈`. This seems very unlikely to cause issues, so maybe it's ok to leave it like that, but I wanted to point that out.

koke (approved):
Looks good to me 🚀

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
itsmeichigo: That's a valid point! I have updated the check to use a close range in 671efa984dfceeb3782992942bce42d19d3f6fea for safety.

PR #4712, opened 2021-08-03
Remove unused Buddybuild prebuild script

mokagio:
I noticed this while working on the secrets. 

I'm not sure when we stopped using Buddybuild, but it certainly unused at the moment. On top of that, Apple shut it down, so, even if we wanted to keep this "just in case, for the future" there would be no point.

I also checked with `git grep buddy` and found no match.


jkmassel (approved):
We never used buddy build for this (AFAIK), so I'm surprised it's here at all?

Nice catch!

PR #4713, opened 2021-08-03
Move JSON credentials secret outside of the repository

mokagio:
Move the remaining secret, `woo_app_credentials.json`, outside of the repository. The work was straightforward because everything was already mostly centralized. Thank you very much 🙇‍♂️ .

## How to test

1. Checkout this branch
1. Build the app, you should see a warning saying about the missing secret

Screen Shot 2021-08-04 at 9 33 46 pm

1. Run `bundle exec fastlane run configure_apply`
1. Run `ls ~/.configure/woocommerce-ios/secrets/` to ensure `woo_app_credentials.json` is there
1. Build, the warning should be gone. Compare `ApiCredentials.swift` with `woo_app_credentials.json` to make sure they have the same values.

---

Update release notes:

### DIFF
--- a/Scripts/build-phases/generate-credentials.sh
+++ b/Scripts/build-phases/generate-credentials.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env bash -euo pipefail
 
 DERIVED_PATH=${SOURCE_ROOT}/DerivedSources
 SCRIPT_PATH=${SOURCE_ROOT}/Credentials/replace_secrets.rb
--- a/Scripts/build-phases/generate-credentials.sh
+++ b/Scripts/build-phases/generate-credentials.sh
@@ -16,6 +16,8 @@ PLIST_TEMPLATE_PATH=${SOURCE_ROOT}/Credentials/Templates/InfoPlist-Template.h
 BASH_INPUT_PATH=${SOURCE_ROOT}/Credentials/bash_secrets.tpl
 BASH_OUTPUT_PATH=${DERIVED_PATH}/bash_secrets
 
+SECRETS_PATH="${SRCROOT}/../.configure-files/woo_app_credentials.json"
+
 ## Validate Secrets!
 ##
 if [ ! -f $SECRETS_PATH ]; then
--- a/Scripts/build-phases/generate-credentials.sh
+++ b/Scripts/build-phases/generate-credentials.sh
@@ -16,7 +16,7 @@ PLIST_TEMPLATE_PATH=${SOURCE_ROOT}/Credentials/Templates/InfoPlist-Template.h
 BASH_INPUT_PATH=${SOURCE_ROOT}/Credentials/bash_secrets.tpl
 BASH_OUTPUT_PATH=${DERIVED_PATH}/bash_secrets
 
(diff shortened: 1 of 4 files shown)


Scripts/build-phases/generate-credentials.sh
```diff
@@ -16,10 +16,14 @@ PLIST_TEMPLATE_PATH=${SOURCE_ROOT}/Credentials/Templates/InfoPlist-Template.h
 BASH_INPUT_PATH=${SOURCE_ROOT}/Credentials/bash_secrets.tpl
 BASH_OUTPUT_PATH=${DERIVED_PATH}/bash_secrets
 
+SECRETS_PATH="${HOME}/.configure/woocommerce-ios/secrets/woo_app_credentials.json"
```
mokagio, line 19: `SECRETS_PATH` used to be a user defined build setting but used only in this script.

By moving the definition here, we save `n - 1` lines in the diff, where `n` is the number of build configurations. You can see that below in the project file (they are six, because the setting was defined in two targets and the project has three build configurations).

More importantly, by moving _stuff_ away from the project file, any diff related to that _stuff_ will be easier to read.

Scripts/build-phases/generate-credentials.xcfilelist
```diff
@@ -0,0 +1,12 @@
+# Input files for the script generating the app's credentials.
+# The script is currently run via a build phase in the GenerateCredentials
+# aggregate target.
+$(SRCROOT)/Credentials/replace_secrets.rb
+$(SRCROOT)/Credentials/ApiCredentials.tpl
+$(SRCROOT)/Credentials/InfoPlist.tpl
+$(SRCROOT)/Credentials/Templates/APICredentials-Template.swift
+$(SRCROOT)/Credentials/Templates/InfoPlist-Template.h
+$(HOME)/.configure/woocommerce-ios/secrets/woo_app_credentials.json
```
mokagio, line 9: The thought occurred to me to move `SECRETS_PATH` in an `.xcconfig` and use it here, so it's fully centralized. Or, at least, as centralized as it can be between the `.configure` JSON and a single place in the project.

mokagio:
~~@jkmassel @AliSoftware I asked for your review even though the tests are failing in CI. I run the same command locally and it passed 🤔 Maybe your fresh eye can find what I'm missing? Thanks.~~

Fixed it. Or, at least, fixed the issue I experienced earlier. The build is still going, so there's still time for surprises 🙃 

The problem was that I deleted the `.enc` file, forgetting how it's required for the secrets decryption in machines that don't have access to the secrets repo. 🤦

Scripts/build-phases/generate-credentials.xcfilelist
jkmassel: That seems like a nice improvement for sure 🙂

jkmassel (approved):
LGTM, though same question as in https://github.com/Automattic/simplenote-macos/pull/1012#pullrequestreview-723555688

mokagio:
> LGTM, though same question as in https://github.com/Automattic/simplenote-macos/pull/1012#pullrequestreview-723555688

For the record, I (sort of) addressed that [here](https://github.com/Automattic/simplenote-macos/pull/1012#issuecomment-893816747)

PR #4720, opened 2021-08-03
[Mobile Payments] Hide Shipping label button on orders eligible for In Person Payments

### LINKED ISSUE #4101: [Mobile Payments] Do not display Create Shipping Label for cash-on-delivery orders
When looking at an order that was placed using the cash-on-delivery payment method, don't display the Create Shipping Label button nor the "Learn more" link below it. This might help the payment section float a bit closer to the top of the device viewport

I am assuming we don't support [USPS COD](https://faq.usps.com/s/article/Collect-on-Delivery) labels nor [FedEx COD](https://www.easypost.com/docs/api) - @c-shultz  can you confirm that?

cc @ctarda  @malinajirka @hichamboushaba

ctarda:
Closes #4101 

Hide the Create Shipping Label button on cash on delivery orders.

For cash on delivery orders:

| Before | After |
| ---- | ---- |
|  |  |

A couple more "afters":

* order paid by credit card:

* order paid by bank transfer:
 

## Changes
* Encapsulate the logic that decides if the "Create Shipping Label" button is visible in a private var, add logic to the previous check (order eligibility and shipping labels not refunded empty) to validate tat the order is not eligible for In-Person Payments
* Add a test for this use case.

## How to test
* Create an order eligible for shipping labels, but not eligible for In-Person Payments. Check that the "Create Shipping label" button is visible
* Create an order eligible for shipping labels and eligible for In-Person Payments. Check that the "Create Shipping label" button is not visible, while the "Collect Payment" button is visible.

Requesting review from @allendav (as the one logging the issue) and/or @pmusolino and/or @rachelmcr  (as the persons more knowledgeable of shipping labels)

Update release notes:

### DIFF
--- a/WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/OrderDetailsDataSourceTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/OrderDetailsDataSourceTests.swift
@@ -329,11 +329,14 @@ final class OrderDetailsDataSourceTests: XCTestCase {
 
     func test_create_shipping_label_button_is_not_visible_for_cash_on_delivery_order() throws {
         // Given
-        let order = makeOrder().copy(status: .processing, datePaid: .some(nil), paymentMethodID: "cod")
+        let order = makeOrder().copy(status: .processing, datePaid: .some(nil), total: "100", paymentMethodID: "cod")
+        storageManager.insertCardPresentEligibleAccount()
+        storageManager.viewStorage.saveIfNeeded()
         let dataSource = OrderDetailsDataSource(order: order, storageManager: storageManager)
-        dataSource.isEligibleForShippingLabelCreation = false
+        dataSource.isEligibleForShippingLabelCreation = true
 
         // When
+        dataSource.configureResultsControllers { }
         dataSource.reloadSections()
 
         // Then
--- a/WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/OrderDetailsDataSourceTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/OrderDetailsDataSourceTests.swift
@@ -327,6 +327,20 @@ final class OrderDetailsDataSourceTests: XCTestCase {
         XCTAssertNil(createShippingLabelRow)
     }
 
+    func test_create_shipping_label_button_is_not_visible_for_cash_on_delivery_order() throws {
+        // Given
+        let order = makeOrder().copy(status: .processing, datePaid: .some(nil), paymentMethodID: "cod")
+        let dataSource = OrderDetailsDataSource(order: order, storageManager: storageManager)
+        dataSource.isEligibleForShippingLabelCreation = false
+
+        // When
+        dataSource.reloadSections()
+
+        // Then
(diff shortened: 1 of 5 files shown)


WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/OrderDetailsDataSourceTests.swift
```diff
@@ -327,6 +327,20 @@ final class OrderDetailsDataSourceTests: XCTestCase {
         XCTAssertNil(createShippingLabelRow)
     }
 
+    func test_create_shipping_label_button_is_not_visible_for_cash_on_delivery_order() throws {
+        // Given
+        let order = makeOrder().copy(status: .processing, datePaid: .some(nil), paymentMethodID: "cod")
+        let dataSource = OrderDetailsDataSource(order: order, storageManager: storageManager)
+        dataSource.isEligibleForShippingLabelCreation = false
```
rachelmcr: For this test to best reflect the COD order scenario, I'd suggest testing these conditions:

```
dataSource.isEligibleForShippingLabelCreation = true
dataSource.isEligibleForCardPresentPayment = true
```

That way we're testing that even if the order is otherwise eligible for shipping label creation, the button doesn't appear because it's eligible for card-present payment.

rachelmcr (commented):
These changes look great! I just have one more suggested change based on shipping label behavior:

We show a banner about shipping label creation at the top of the Order Details screen when the order is eligible for shipping label creation. Right now the banner's appearance depends on `isEligibleForShippingLabelCreation` so the banner still shows up even when `shouldShowShippingLabelCreation` is false:



If we make `shouldShowShippingLabelCreation` a public var, we can use it [to determine if the banner should be displayed](https://github.com/woocommerce/woocommerce-ios/blob/fd1edbbf150834a93d41a68ec507d06e23e5d2f9/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/OrderDetailsViewController.swift#L231) so it's hidden as well on orders eligible for in person payments.

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/OrderDetailsDataSourceTests.swift
ctarda: Thanks for taking a look @rachelmcr Great catch!

I pushed an update to the test that:
* sets dataSource.isEligibleForShippingLabelCreation to true
* ensures the order is eligible for In-Person Payments by setting its total amount, and by inserting a valid payment gateway into the moved coredata context (dataSource.isEligibleForCardPresentPayment is read-only so it can not be set directly)

ctarda:
Oops, I missed the comment about the banner. Another update incoming!

ctarda:
Thanks for the review @rachelmcr 

I pushed a couple of commits to:
* improve the test
* Hide the banner on orders eligible for In-Person Payments.

This is an order eligible for shipping labels:


And this would be an order eligible for In-person Payments:


I think this is ready for another round of review! 🙇🏼

rachelmcr (approved):
Looks great! Thanks for this, @ctarda! :shipit:

ctarda:
Thank you for the review @rachelmcr . Merging now!

PR #4721, opened 2021-08-03
[Mobile Payments] Display reader connection modals on top of Order Details view

allendav:
Ready for review

Partially addresses #4546 - avoids the "connect your card reader" and "disconnect" views from appearing when connecting a card reader in the context of order details

NOTE:
- Unit tests will be added in a subsequent PR
- At least one modal briefly flashes after connecting.  If I can't resolve that shortly, I'll open an issue for that as well.

To test:
- Create a test order with cash on delivery as the payment method
- Make sure the card reader is off and NOT known to the mobile device
- Navigate to the order and tap collect payment
- Ensure the Scanning for reader modal appears in front of the Order details
- Tap cancel
- Ensure you are returned to Order details
- Navigate to the order and tap collect payment
- Ensure the Scanning for reader modal appears in front of the Order details
- Power on the reader
- When prompted “Do you want to connect to reader XXX”, tap Connect to Reader
- Ensure you get the “Collect payment from XXX” modal next
- Tap cancel to abort the payment collection
- Tap collect payment again and ensure you get the “Collect payment from XXX” modal again immediately
- Tap cancel to abort the payment collection
- Power off the reader
- Tap collect payment
- Ensure the Scanning for reader modal appears in front of the Order details
- Power on the reader
- Ensure you are immediately sent to the get the “Collect payment from XXX” modal since the reader is now known

Screengrabs:

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
@@ -23,23 +23,25 @@ final class CardReaderConnectionController {
         ///
         case searching
 
-        /// Ending search for card readers
-        ///
-        case endSearch
-
         /// Found a card reader
         ///
         case foundReader
 
-        /// Attempting to connect to a card reader
+        /// Attempting to connect to a card reader. The completion passed to `searchAndConnect`
+        /// will be called with a `success` `Bool` `True` result if successful, after which the view controller
+        /// passed to `searchAndConnect` will be dereferenced and the state set to `idle`
         ///
         case connectToReader
 
-        /// Connected successfully to a card reader
+        /// User cancelled search/connecting to a card reader. The completion passed to `searchAndConnect`
+        /// will be called with a `success` `Bool` `False` result. The view controller passed to `searchAndConnect` will be
+        /// dereferenced and the state set to `idle`
         ///
-        case connectedToReader
+        case cancel
 
-        /// A failure occurred during search or connection
+        /// A failure occurred. The completion passed to `searchAndConnect`
+        /// will be called with a `failure` result. The view controller passed to `searchAndConnect` will be
+        /// dereferenced and the state set to `idle`
         ///
         case failed(Error)
     }
@@ -49,7 +51,7 @@ final class CardReaderConnectionController {
             didSetState()
         }
     }
-    private var fromController: UIViewController
+    private var fromController: UIViewController?
     private var siteID: Int64
     private var knownCardReadersProvider: CardReaderSettingsKnownReadersProvider
 
@@ -57,17 +59,17 @@ final class CardReaderConnectionController {
     private var knownReaderIDs: [String]
     private var subscriptions = Set<AnyCancellable>()
 
+    private var onCompletion: ((Result<Bool, Error>) -> Void)?
+
     private lazy var alerts: CardReaderSettingsAlerts = {
         CardReaderSettingsAlerts()
     }()
 
     init(
-        from: UIViewController,
         forSiteID: Int64,
         knownReadersProvider: CardReaderSettingsKnownReadersProvider
     ) {
         state = .idle
-        fromController = from
         siteID = forSiteID
         knownCardReadersProvider = knownReadersProvider
         knownReaderIDs = []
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -66,6 +66,17 @@ final class OrderDetailsViewController: UIViewController {
     ///
     private var cardReaderAvailableSubscription: Combine.Cancellable? = nil
 
+    /// Controller that facilitates connecting to card readers
+    private lazy var cardReaderConnectionController: CardReaderConnectionController = {
```
ctarda: It seems to me like there is potential for a leak here. CardReaderConnectionController holds a reference to OrderDetailsViewController and the other way around.

Is this declared as a property because it needs to be a property, or to encapsulate the initial configuration of the view controller that is being instantiated? Because if we don't need to retain the new view controller (CardReaderConnectionController) as a property, and what we want is to tidy up the way it is initialled, it could be initialized in a private method.

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /// Begin the search for a card reader
+    ///
+    func onBeginSearch() {
+        self.state = .searching
+        let action = CardPresentPaymentAction.startCardReaderDiscovery(
+            siteID: siteID,
+            onReaderDiscovered: { [weak self] cardReaders in
+                /// Surprisingly, onReaderDiscovered may be called with an empty array
```
ctarda, line 146: I would suggest guarding here against self being nil. That would remove all the optional self in the rest of the inplementation

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -0,0 +1,237 @@
...
+        let action = CardPresentPaymentAction.cancelCardReaderDiscovery() {_ in
+            self.state = .idle
+        }
+        ServiceLocator.stores.dispatch(action)
+    }
+
+    /// A reader has been found
+    ///
+    func onFoundReader() {
+        guard foundReader != nil else {
```
ctarda: What about something like:
```
guard let foundReaderId = foundReader.id else {
}
```

That would allow packing the two guards into one

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -0,0 +1,237 @@
...
+            self.state = .endSearch
+            return
+        }
+
+        let action = CardPresentPaymentAction.connect(reader: reader) { [weak self] result in
+            switch result {
+            case .success(let reader):
+                self?.knownCardReadersProvider.rememberCardReader(cardReaderID: reader.id)
+                // If the reader does not have a battery, or the battery level is unknown, it will be nil
+                let properties = reader.batteryLevel
```
ctarda, line 244: I'm just curious here. Could this be `let properties = ["battery_level": reader.batteryLevel`? What is the benefit of passing through map?

ctarda (approved):
This is definitely an improvement. There is one thing about a potential memory leak for you to consider.

:shipit:

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
koke: That would not give the same result. Given that the tracks implementation is done in Objective-C, I'm not even sure what would happen if we pass a dictionary with a `nil` value there 😰 

Screen Shot 2021-08-04 at 12 48 18

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
allendav: I agree. I changed the approach in https://github.com/woocommerce/woocommerce-ios/pull/4721/commits/af4232cacb613e4b77f3cf66ea0d88a4f88aa101 such that the reference to the "from" view controller is discarded at the end of the flow - i.e. after a reader is connected, the user cancels, or an error occurs.

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4721/commits/af4232cacb613e4b77f3cf66ea0d88a4f88aa101

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4721/commits/8d8e83d98fb675e5c59ae16a3420d6ce0e104236

PR #4722, opened 2021-08-04
[CPP Onboarding] Synchronize required data when loading CPP Onboarding screens

koke:
Part of #4611

While working on #4702 I realized that, if we need to check data about plugins and settings, we need to ensure that the information is relatively fresh. Specially in the case of plugins, it doesn't seem like they are synced anywhere other than the plugins list, so if the user hasn't visited that, the app won't know about the installed plugins.

This PR ensures that all the required data is refreshed when In-Person Payments is loaded, and that the UI updates if the underlying data has changed.

## To test

I didn't fully test this in real world conditions because this is still mostly a skeleton for the future and most of the implementation is mocked. What I tested was:

1. Have a store correctly set up for payments.
2. In `InPersonPaymentsViewModel.init(siteID:)`, I replaced `state = useCase.checkOnboardingState()` with `state = .genericError`. This makes the screen show a generic error when you first visit.
3. Go to Settings > In-Person Payments. The screen should show the generic error we just forced.
4. After a moment, when the data is loaded from the API, the screen should change to the "Connect to Reader" screen.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsViewController.swift
@@ -36,7 +36,7 @@ private enum Localization {
 struct InPersonPaymentsView_Previews: PreviewProvider {
     static var previews: some View {
         NavigationView {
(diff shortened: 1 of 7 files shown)


ctarda (approved):
If I force the initial state to be an error, as suggested in the testing instructions, I can see the error and the UI change to the connection screen after about a second.

:shipit:

PR #4723, opened 2021-08-04
Shipping Label: Add customs row for international shipping label form

### LINKED ISSUE #4688: Shipping Labels: Customs field for international shipping
This is a subtask of #4596.

When detecting international shipping, we should display a field on the Create Shipping Label form for filling out Customs form:

Customs field on Create Shipping Label form

itsmeichigo:
Closes #4688 

# Description
This PR updates the shipping label creation form to add a new row for customs form when international shipping is detected. There are a few things to be considered:
- Customs row should be inserted / removed when requirement for customs form changes. E.g: when destination address is updated so that shipping is within the US only, the row should be removed.
- When either address is updated so that customs form is required, there should be a check to make sure that phone number of origin address is present; if not - origin address row should be updated to pending and the subsequent rows should be updated to make sure that only 1 pending row is editable at a time.

# Changes
- Added new Row case `.customs` and configured table view cell for this type.
- Updated UIImage with new globe icon for use in the customs cell
- Updated ShippingLabelFormViewModel:
  - Made `generateInitialSections` method non-static, let it check for `customsFormRequired` to insert customs cell if needed
  - Added new private method `updateRowsForCustomsIfNeeded` to update origin address row to pending state if customs form is required and phone number for origin address is missing. This method is called whenever there's an update to either origin or destination address, since both these addresses can affect value of `customsFormRequired`.
  - Updated method `updateRowState` to insert / remove customs row when necessary. Also added a check to make sure that there's only 1 pending row editable at once.
- ShippingLabelFormViewController was also updated to make sure that it navigates to edit address form if phone number is missing (disregard of whether the address is validated or not).

# Demo
customs-row

# Testing steps
2. Navigate to Orders tab and select an order with status Processing.
4. If the order is international (origin country different from destination country, or one of the addresses is one of [US military states](https://git.io/JBuja)), notice that customs row is present. Otherwise, customs row should not be present.
5. Suppose the order is international, update the destination country to US and state being non-military state, proceed and notice that customs rows has been removed.
6. Update destination row to an international address, proceed and notice that customs row is added to the form.
7. If the origin address doesn't has a phone number, notice that the Ship From row is pending and editable, while other rows are not. Tapping on the Ship From row should navigate to edit address screen with a validation error about missing phone number.

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -413,6 +458,16 @@ private extension ShippingLabelFormViewModel {
 
         state.sections = [Section(title: nil, rows: rows), summarySection].compactMap { $0 }
     }
+
+    func updateRowsForCustomsIfNeeded() {
```
pmusolino, line 448: nit: it could be good to implement a test, making sure that the row state is updated under certain conditions.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -145,10 +153,10 @@ final class ShippingLabelFormViewModel {
                                                                account: defaultAccount)
         self.destinationAddress = ShippingLabelFormViewModel.fromAddressToShippingLabelAddress(address: destinationAddress)
 
-        state.sections = ShippingLabelFormViewModel.generateInitialSections()
         self.stores = stores
         self.storageManager = storageManager
 
+        state.sections = generateInitialSections()
```
pmusolino, line 165: Nonblocking question (just curiosity): is there a reason why you moved it after the `stores` and `storeManager` allocation?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -210,8 +212,14 @@ private extension ShippingLabelFormViewController {
...
-                    self.viewModel.handleOriginAddressValueChanges(address: response?.address,
-                                                                   validated: true)
+                    // Navigate to edit address if phone number is required but empty
+                    if self.viewModel.customsFormRequired &&
+                        shippingLabelAddress?.phone.isEmpty == true {
+                        self.displayEditAddressFormVC(address: response?.address, validationError: nil, type: .origin)
+                    } else {
+                        self.viewModel.handleOriginAddressValueChanges(address: response?.address,
+                                                                       validated: true)
+                    }
```
pmusolino: It would be ideal here, to not modify this logic, because for us it means that the origin address is still validated, but it is not in reality (because of the missing phone). So, the ideal solution would be to modify the origin address and to consider its data state as `pending` instead of `validated`. Do it makes sense?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -403,6 +417,37 @@ private extension ShippingLabelFormViewModel {
...
+            let customsRowIndex = rows.index(after: packageDetailsRowIndex)
+            let customsRow = Row(type: .customs, dataState: .pending, displayMode: customsRowState)
+            rows.insert(customsRow, at: customsRowIndex)
+        }
+
+        // Remove customs row if necessary
+        if !customsFormRequired,
+           let customsRowIndex = rows.firstIndex(where: { $0.type == .customs }) {
+            rows.remove(at: customsRowIndex)
+        }
```
pmusolino: It seems to me that we shouldn't use the `updateRowState` to insert/remove a row because they are two different actions, in this case, the method would become too generic.
I would suggest in this case to separate the concept of insertion from that of updating.

pmusolino (commented):
The testing is ✅ I just left some comments with some code improvements 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
itsmeichigo: The reason is that I need to instantiate all stored properties before I can call the instance method 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
itsmeichigo: I have updated the tests in 9a525ba4ffd0fb9820cfdaa5062f332e26144762 to make them clearer.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
itsmeichigo: Good point! I have moved this part to a separate method that gets called from `updateRowsForCustomsIfNeeded` in f570f8131843ac5e204eac54cb5bba88a4464ba6, since it makes more sense that they belong there.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
itsmeichigo: I have reverted this change and move the redirect outside of the remote validation callback in 646fe47f9fbab5852e2abb258ce45fed148f0b15 👍

itsmeichigo:
Thanks @pmusolino for the feedback! I have made some changes and this PR is ready for another look! 

**EDIT**: I have also pushed some changes with regards to the missing `customsFormRequired` logic, please also take a look at that 🙇‍♀️

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
pmusolino: It looks good 💯

pmusolino (approved):
I took another look at your changes, and they look good :shipit:

itsmeichigo:
Thanks for the review @pmusolino!

> since here we are stopping the flow, it would be more clear for this specific case to use a guard statement instead of the if

I agree that using `guard` clarifies my intention much more, but that would require inverting the logic check and it doesn't look neat to me 😅. I'd prefer to keep the `if` for now.

PR #4724, opened 2021-08-04
[CPP Onboarding] Implement country checks

koke:
Part of #4611 

This PR checks the store address and returns a `wcpayUnsupportedVersion` state if the country is not supported (we currently only support the US). 

The implementation is relatively simple but I had to move around and refactor quite a few things. Before this PR, I had added `CardPresentPaymentsOnboardingUseCase` to Yosemite instead of WooCommerce, but now I realized it would make things a lot easier if we can access the existing logic for figuring out a site address via `SiteAddress` and `SelectedSiteSettings`, so I moved the existing code to `WooCommerce`.

Because the existing code relies more heavily on `ServiceLocator`, I changed the API so it doesn't take a `siteID` anymore, and takes it from the session instead. I also had to update `SelectedSiteSettings` to make it more testable.

Finally, I made `SiteSettings` copiable, but there was a mismatch where one of the initializer arguments didn't match the property name, so the generated `copy()` had a syntax error. This was an easy find and replace, but it makes the PR a bit noisier.

Because of all these things, I feel this PR is getting large enough, so I'll handle the error screen in a separate PR.

## To test

Unit tests should pass

Update release notes:

### DIFF
--- a/Fakes/Fakes/Fakes.generated.swift
+++ b/Fakes/Fakes/Fakes.generated.swift
@@ -1314,7 +1314,7 @@ extension SiteSetting {
             siteID: .fake(),
             settingID: .fake(),
             label: .fake(),
-            description: .fake(),
+            settingDescription: .fake(),
             value: .fake(),
             settingGroupKey: .fake()
         )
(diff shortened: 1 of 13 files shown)


Ecarrion (approved):
I had a bit of trouble understanding why all of these changes were needed and how they play together but it looks good to me!

PR #4728, opened 2021-08-05
Add background for button bar of order list for edge-to-edge look

itsmeichigo:
Part of #3716 

# Description
@designsimply found while beta testing v7.2 that the order list is still not extending edge-to-edge in dark mode. This is due to the inconsistent background color in dark mode.

# Changes
- Added new background view underneath the button bar view. This view's leading and trailing edges are constrained to its super view and not safe area.
- Set background color for the background view the same color as the button bar view's. This gives the illusion that the button bar view extends edge-to-edge.

# Screenshots
| Light Mode | Dark Mode |
| ------ | ------ |
| Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 09 48 17 | Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 09 48 27 |

# Testing
1. Select Orders tab.
2. Switch to landscape mode. 
3. Notice that the screen expands edge-to-edge in both light and dark mode.

Update release notes:


pmusolino (approved):
LGTM :shipit:

PR #4729, opened 2021-08-05
[CPP Onboarding] Implement onboarding plugin checks

koke:
Part of #4611 

This PR adds all the onboarding checks related to plugins:
- Is WooCommerce Payments installed?
- Is WooCommerce Payments active?
- Is the WooCommerce Payments version recent enough (2.5+)?

## To test

I've tested on some actual stores that Jirka set up with every possible state. For this one, it might still be easy enough to test, but otherwise I'd say it might be OK to rely on unit tests for now, and do more thorough testing once all the checks are implemented.

If you want to test this, you'll need a store with a US address (only the country is checked, so the address might be fake). There is no UI to show this states so, to see any changes, you'll need to inspect the `viewModel.state` in `InPersonPaymentsView`. You can set a breakpoint, log to the console, or apply [this patch](https://gist.github.com/koke/c896431264a5284903aac84a0f9cfae5) so that the generic error screen includes the state.

Then go to Settings > In-Person payments, and you should see different states depending on the state of WCPay plugin installation on the store:

Before installing plugin | Installed but not active | Installed an old (<2.5) version | Installed and active
-|-|-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 10 37 52|Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 10 38 09|Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 10 38 30|Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 10 38 53

Update release notes:

### DIFF
--- a/Networking/Networking/Model/SitePluginStatusEnum.swift
+++ b/Networking/Networking/Model/SitePluginStatusEnum.swift
@@ -39,6 +39,15 @@ extension SitePluginStatusEnum: RawRepresentable {
         case .unknown:       return Keys.unknown
         }
     }
+
+    /// Returns true if the plugin is active, either individually or network-wide
+    ///
+    public var isActive: Bool {
+        switch self {
+        case .active, .networkActive:   return true
+        case .inactive, .unknown:       return false
+        }
+    }
 }
 
 /// Enum containing all possible plugin status keys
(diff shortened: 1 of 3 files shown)


ctarda (approved):
Tested on my own stores, I can see the error messages for plugin not installed, installed an not active, and installed but unsupported.

:shipit:

PR #4730, opened 2021-08-05
[CPP Onboarding] Refactor country setup in tests

koke:
Part of #4611 

**Note:** This targets #4729, so that should be merged first, although it can be reviewed independently.

This PR extracts the logic to set up a store's address country to a helper method. Since we have to do that for almost every onboarding test (and there will be a lot of them), having a method makes the test intention clearer.

In the end, the `Given` part of each unit test will look like:

```swift
setupCountry(country: .us)
setupPlugin(status: .active, version: .supported)
setupPaymentGatewayAccount(status: .restricted, hasOverdueRequirements: true)
```

The `setupPlugin` part comes in #4729, and the `setupPaymentGatewayAccount` will come in a future PR alongside the actual account checks and tests. 

## To test

Unit tests should pass, the logic hasn't changed.

Update release notes:


ctarda (approved):
Nice!

:shipit:

PR #4731, opened 2021-08-05
[CPP Onboarding] Implement onboarding payment account checks

### LINKED ISSUE #4707: [Mobile Payments] Add missing account state "restricted soon" to WCPayAccountStatusEnum
We missed one when we set up the enum. Restricted soon is similar to restricted, but without overdue requirements.

See backend server class-stripe-account.php

cc @malinajirka

koke:
Part of #4611 
Fixes #4707

This PR adds the final set of onboarding checks related to the WCPay account:
- There is an account set up
- (Temporarily) It is eligible for In-Person payments
- Setup is complete
- The account is not under review
- The account doesn't have pending/overdue requirements
- The account hasn't been rejected

All these states come from the backend, but we want to show different error screens with different actions for each state. 

Note that this is heavily inspired on the existing android implementation:
- [CardReaderOnboardingChecker](https://github.com/woocommerce/woocommerce-android/blob/develop/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/cardreader/onboarding/CardReaderOnboardingChecker.kt)
- [CardReaderOnboardingCheckerTest](https://github.com/woocommerce/woocommerce-android/blob/develop/WooCommerce/src/test/kotlin/com/woocommerce/android/ui/prefs/cardreader/onboarding/CardReaderOnboardingCheckerTest.kt)

## To test

I've tested on some actual stores that Jirka set up with some of the possible states. I'd say it might be OK to rely on unit tests for now as it will be time consuming to test every possibility now. We can do more thorough testing once all the feature implementation is complete. I'm not sure if we'll be able to manually test every case anyway (e.g. the account was rejected for fraud).

If you want to test this, you'll need a store with a US address (only the country is checked, so the address might be fake). There is no UI to show this states so, to see any changes, you'll need to inspect the viewModel.state in InPersonPaymentsView. You can set a breakpoint, log to the console, or apply this patch so that the generic error screen includes the state.

Then go to Settings > In-Person payments, and you should see different states depending on the state of WCPay setup:

Account with overdue requirements | Account with pending requirements | Account rejected
-|-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 11 08 18|Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 11 08 30|Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 11 08 56

Update release notes:

### DIFF
--- a/Networking/Networking/Model/WCPayAccountStatusEnum.swift
+++ b/Networking/Networking/Model/WCPayAccountStatusEnum.swift
@@ -18,6 +18,11 @@ public enum WCPayAccountStatusEnum: Decodable, Hashable, GeneratedFakeable {
     /// review by Stripe and the merchant will probably not be able to collect card present payments.
     case restricted
 
+    /// This state occurs when there is required business information missing from the account but the
+    /// currentDeadline hasn't passed yet (aka there are no overdueRequirements). The merchant will probably be able
+    /// to collect card present payments.
+    case restrictedSoon
+
     /// This state occurs when our payment processor rejects the merchant account due to suspected fraudulent
     /// activity. The merchant will NOT be able to collect card present payments.
     case rejectedFraud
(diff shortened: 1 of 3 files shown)


ctarda (approved):
:shipit:

PR #4733, opened 2021-08-05
Add primary and secondary button styles for SwiftUI

koke:
The upcoming UI for Payments Onboarding (see #4611) will use SwiftUI for its screens. Currently we don't have a way to apply our primary/secondary button styles to SwiftUI buttons, so I'm adding those in this PR, since they're not specific to onboarding.

The styles have been copied from `UIButton+Helpers.swift`, and adapted to SwiftUI. An example of usage can be seen in the SwiftUI preview code:

```swift
Button("Primary button") {}
    .buttonStyle(PrimaryButtonStyle())

Button("Secondary button") {}
    .buttonStyle(SecondaryButtonStyle())
```

button-styles

## To test

These aren't used anywhere in the app yet, but it can be tested through the SwiftUI preview

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ButtonStyles.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ButtonStyles.swift
@@ -30,22 +30,46 @@ private struct PrimaryButton: View {
 
     var body: some View {
         BaseButton(configuration: configuration)
-            .foregroundColor(Color(isEnabled ? .primaryButtonTitle : .buttonDisabledTitle))
+            .foregroundColor(Color(foregroundColor))
             .background(
                 RoundedRectangle(cornerRadius: Style.defaultCornerRadius)
-                    .fill(Color(isEnabled
-                                        ? configuration.isPressed ? .primaryButtonDownBackground : .primaryButtonBackground
-                                        : .buttonDisabledBackground))
+                    .fill(Color(backgroundColor))
             )
             .overlay(
                 RoundedRectangle(cornerRadius: Style.defaultCornerRadius)
                     .strokeBorder(
-                        Color(isEnabled
-                                    ? configuration.isPressed ? .primaryButtonDownBorder : .primaryButtonBorder
-                                    : .buttonDisabledBorder),
+                        Color(borderColor),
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ButtonStyles.swift
```diff
@@ -0,0 +1,103 @@
...
+    let configuration: ButtonStyleConfiguration
+
+    var body: some View {
+        BaseButton(configuration: configuration)
+            .foregroundColor(Color(isEnabled ? .primaryButtonTitle : .buttonDisabledTitle))
+            .background(
+                RoundedRectangle(cornerRadius: Style.defaultCornerRadius)
+                    .fill(Color(isEnabled
+                                        ? configuration.isPressed ? .primaryButtonDownBackground : .primaryButtonBackground
+                                        : .buttonDisabledBackground))
```
Ecarrion: For these color configurations do you think it's worth it to have something like
```swift
private var enabledColor: UIColor {
    configuration.isPressed ? .primaryButtonDownBackground : .primaryButtonBackground
}

private var disabledColor: UIColor {
    .buttonDisabledBackground
}
```

For then being able to configure it like:
```swift
.background(
    RoundedRectangle(cornerRadius: Style.defaultCornerRadius)
        .fill(Color(isEnabled ? enabledColor : disabledColor))
)
```

I mention this because I found really hard to know what color was being set when I was reading the code.

Ecarrion (approved):
Looks good! Only a small suggestion that may help with readability!

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ButtonStyles.swift
koke: That makes sense, the nested ternaries are not very readable 😞

koke: I've extracted those to computed properties, and used if/else instead of ternaries in d0e750b5c35ecc37d5318c89c6a8e60675807d78, which I think reads a lot clearer now

PR #4734, opened 2021-08-05
[CPP Onboarding] Add learn more link to InPersonPaymentsUnavailableView

koke:
Part of #4611 

This PR adds the "Learn more" link at the bottom of `InPersonPaymentsUnavailableView`. This is mostly an opportunity to  bring in `AttributedText` and `SafariSheet` from #4626 in a standalone PR so we can use it in all the future onboarding views.

The [AttributedText](https://github.com/gonzalezreal/AttributedText) file is a fork of a library necessary so we can customize how links are opened, instead of the default behavior of leaving the app to open in Safari (See paNNhX-fz-p2 for some extra discussion).

`SafariSheet` wraps a `SFSafariViewController` into a more typical SwiftUI sheet API. The method that uses a `Binding<URL?>` argument ended up a bit weird because I turn it into a `Binding<Bool>` to indicate whether the sheet should be presented. Technically, this allows an odd case where the binding could go from `false` to `true` but that couldn't translate into a URL to feed back to the view, so it does nothing. However, the `sheet` implementation will only ever set that from `true` to `false` on dismiss, and we only ever make it true when setting a URL so it's not a real problem. Still, the API is a bit uncomfortable to look at, but it's the best I got considering the URL to be presented is somewhat dynamic (it's not really, but it comes from an attributed string).

## To test

1. Go to Settings > In-Person Payments on a store that isn't ready to take payments so you can see the error screen.
2. Tap on the learn more link and make sure the Safari sheet is presented

Error screen | After tapping on Learn more
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 15 35 27|Simulator Screen Shot - iPhone 12 Pro - 2021-08-05 at 15 35 31

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AttributedText.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AttributedText.swift
@@ -1,0 +1,150 @@
+import SwiftUI
+
+/*
+ This class was based on the [AttributedText](https://github.com/gonzalezreal/AttributedText) library
+ by [Guille Gonzalez](https://github.com/gonzalezreal). It has then been adapted to fit our requirements.
+
+ Copyright (c) 2020 Guille Gonzalez
+
+ Permission is hereby granted, free of charge, to any person obtaining a copy
+ of this software and associated documentation files (the "Software"), to deal
+ in the Software without restriction, including without limitation the rights
+ to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ copies of the Software, and to permit persons to whom the Software is
+ furnished to do so, subject to the following conditions:
+
+ The above copyright notice and this permission notice shall be included in all
+ copies or substantial portions of the Software.
+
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AttributedText.swift
```diff
@@ -0,0 +1,150 @@
+import SwiftUI
+
+/*
```
ctarda, line 3: I think we have an "Open Source licenses" section in Settings. Should this be in there?

ctarda (approved):
It works great. This is what I see in a store that is not payments-ready:



And when tapping Learn More:


I left a question about the license. I am not sure if that's a valid concern or not. But leaving that aside, 

:shipit:

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AttributedText.swift
koke: Good call, I didn't know about that. I've added a line in 80710432e

PR #4735, opened 2021-08-05
Plugins: Hide plugins from settings from non-admin roles

### LINKED ISSUE #4276: Plugins: Hide feature for non admin roles
The plugin list endpoint `wp/v2/plugins` is only accessible for `admin` users. 
To release the feature we need to hide its accessibility from the setting screen for non-admin roles.

Tasks:
- [x] Corroborate that we can access the user's role.
- [x] If the role is `admin` show the "plugin" section in the settings screen.
- [x] Refactor code to move the plugin list synchronization out of the session manager, as we don't need to sync it at startup anymore.
- [x] Remove feature flag.

Ecarrion:
closes #4276

# Why 

Managing plugins is only available for admin roles, when @itsmeichigo was working on listing plugins for her trial, we had the idea to restrict the app only to admin roles, however, that decision changed and this PR **hides the plugin list from non-admin roles and removes the feature flag the feature was behind.**

# How

To accomplish this I'm taking advantage of the work recently introduced in #4492, which already fetches the role of the user at start-up or after a store switch and saving those roles in the `SessionManager`. 

I spent a while thinking if I should store those roles in `CoreData` using the `User` entity but I finally decided against it because most(or all?) "default store" information is already persisted in the session manager.

Finally, `SettingsViewControllers` fetches the session manager information to know if it should display the plugins section or not.

# Screenshots

Shop Manager | Admin
---- | ----
no-plugins | plugins

# Testing Steps

- Launch the app with an admin user(make sure you are able to change the user's role).
- Navigate to the settings screen and check that you see the plugins section.
- In Core change the user role to "shop manager"
- Relaunch the app
- Navigate to the settings screen and check that you don't see the plugins section.

Update release notes:


WooCommerce/Classes/System/SessionManager.swift
```diff
@@ -95,6 +95,17 @@ final class SessionManager: SessionManagerProtocol {
         }
     }
 
+    /// Roles for the default Store Site.
+    ///
+    var defaultRoles: [String] {
```
koke: Maybe we can have this be `[User.Role]`? I imagine if we want to use it in more places, the `enum` will almost always be more useful than the raw strings. It doesn't seem like a big deal right now, so I'm OK with it either way.

koke (approved):
Code looks good, and tested with both admin and shop manager and it works as expected 🚢

WooCommerce/Classes/System/SessionManager.swift
Ecarrion: That was my initial approach, but sadly, `SessionManager` stores its information using `UserDefaults` so we are limited to use scalar? types.

Although, now that I think about it I could do that conversion in the session manager 🤔

Ecarrion: Updated in https://github.com/woocommerce/woocommerce-ios/pull/4735/commits/b2d0622102a4701a1d8e4c965bec6ca7a68aa1e1 

Thanks for making me double-check!

PR #4657, opened 2021-07-20
[Shipping Labels] Edge-to-edge table views

itsmeichigo:
Part of #3716 

# Description
This PR aims to update table views in Shipping Labels forms to extend edge-to-edge.

# Changes
## UIKit views: Address Form and Suggested Address 
- Updated leading and trailing edges for table views to constrain to their super views' edges.
- Pinned top banner view to all edges of the table views' headers.
- Pinned content of top banner view to safe areas.

## SwiftUI views: Package details, package list, Carriers and Payment methods
- Added a new view modifier to add paddings to a view based on input insets and a set of edges.
- Ignore safe areas on the scroll views.
- Retrieved safe area insets from GeometryReader and use them to add paddings to table contents.
- Updated top banner wrapper view to accept edge insets to use as paddings for the top banner view.
- Sent safe area insets to banner views of Carriers and Payment Methods screens to keep contents within safe areas.

# Screenshots
| Before | After |
| ------ | ------ |
|  |  |
|  |  |
|  |  | 
|  |  | 
|  |  | 
|  |  | 
|  |  |

# Testing
1. Switch to a store that's eligible for creating shipping labels.
2. Navigate to Orders tab.
3. Select an order with Processing status.
4. Select Create Shipping Label.
5. Navigate through each step of the form in portrait mode - notice that layout is not broken.
6. Switch to landscape mode and navigate through each step again - notice that the table views and banner views all stretch to edges of the screen while contents stay in safe areas (not clipped by iPhone notch).

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4736, opened 2021-08-05
HF: Fix TopBannerView on SwiftUI

Ecarrion:
# Why

It was reported by @designsimply that the Top Banner in the Add-Ons view is broken on Release 7.2.
After some investigation, the issue is a regression introduced on https://github.com/woocommerce/woocommerce-ios/pull/4657.

As the UI bug causes such a bad first impression(even if the feature is experimental), we have decided to ship the fix in `7.2` and not in `7.3`.

@jkmassel would you be so kind to include this in the `7.2 Release` 🙏 

# How

The issue was that the `TopBannerView` was not being pinned to its parent because the pinning happens when the edge insets are set

```swift
   var edgeInsets: EdgeInsets = .zero {
        didSet {
            removeConstraints(constraints)
            bannerView?.pinSubviewToAllEdges(self, insets: contentInsets)
        }
    }
```

But the banner was being set after the insets are set, so `bannerView` was nill at that time.

Assigning the banner before assigning the insets fixes the issue.

# Screenshots

Before | After
bad | good

# Testing Steps

- Enable add-ons experimental feature
- Navigate to an order with add-ons and tap the view add-ons button
- See that the banner has the correct layout.

PS @itsmeichigo could you check that this does not break any other change that you meant with the original PR, please?

 
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerWrapperView.swift
+++ b/WooCommerce/Classes/ViewRelated/Top Banner/TopBannerWrapperView.swift
@@ -54,8 +54,8 @@ final class TopBannerWrapperView: UIView {
 
         bannerView.translatesAutoresizingMaskIntoConstraints = false
         addSubview(bannerView)
-        self.edgeInsets = edgeInsets
         self.bannerView = bannerView
+        self.edgeInsets = edgeInsets
         self.backgroundColor = bannerView.backgroundColor
     }
 


itsmeichigo (approved):
Thanks @designsimply for catching this bug, and my apologies for letting it slip when updating the top banner view wrapper. And thanks @Ecarrion for the speedy fix!

Everything works great, including the shipping labels screens 👍 I notice that the add-ons list is not extending edge-to-edge in landscape mode, but we can work on this later as part of #3716.

astralbodies:
Nice catch, @designsimply and thanks @itsmeichigo for a helpful quick review for @Ecarrion!

PR #4738, opened 2021-08-06
[CPP Onboarding] Implement view for plugin not installed

koke:
Part of #4611 

This adds a specific onboarding error screen for the case when the store is in a supported country but the WooCommerce Payments plugin is not installed.

## To test

You should have a store with an address in the US and without the WooCommerce Payments installed

1. Go to Settings > In-Person Payments. The app should show a screen prompting you to install the extension.
2. From wp-admin, install WooCommerce Payments from WooCommerce > Extensions > Payments.
3. Tap on Refresh After Installing.

Extension not installed | After tapping Refresh
-|-
Screen Shot 2021-08-06 at 11 50 16|Screen Shot 2021-08-06 at 11 11 42

Update release notes:


koke:
Noticed some issues with the style for the learn more link, but I'll handle those separately in #4739

WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -455,6 +455,12 @@ extension UIImage {
         return UIImage(named: "woo-payments-error")!
     }
 
+    /// Payments plugin
+    ///
+    static var paymentsPlugin: UIImage {
```
ctarda, line 460: It would be a good idea to add a test in IconTests, to cover against the possibility of the asset being missing

ctarda (approved):
Works as advertised 🥳 

One comment about adding a test, but other than that :shipit:

allendav:
@Garance91540 - it isn't clear what we'd like the merchant to do here? Can we add some more actionable text (i.e. "activate the plugin")?

Garance91540:
Hi @allendav , since we are not able to activate the plugin yet on from the app, we are telling merchants to "install the free WooCommerce Payments extension on [their] store to accept payments". The CTA is then "refresh after installing" [the plugin]. We could link to the store admin here, but honestly I don't think installing a plugin from the mobile web has a big chance of success so that's why I didn't included here.

 I personally think the current text is pretty clear but what would you suggest?

allendav:
I should have been clearer. The CTA is unclear for the "After tapping refresh" (right hand side above) view

Garance91540:
I guess I'm confused by this use case.

When will merchants see [this screen](https://d.pr/i/TCKuTa) after installing WC Pay on their site? Only when they have installed it and not activated it? In that case we should show [the same screen as before](https://d.pr/i/n3rN4Y), but updating the text with "Activate the WooCommerce Payments"/ " You'll need to activate the plugin...."

allendav:
> When will merchants see this screen after installing WC Pay on their site? Only when they have installed it and not activated it?

I think that is correct.

> In that case we should show the same screen as before, but updating the text with "Activate the WooCommerce Payments"/ " You'll need to activate the plugin...."

Yes, I think that captures my thoughts as well

Issue #4687, opened 2021-07-28
Shipping Labels: Customs form for international shipping


itsmeichigo:
This is a subtask of #4596.

When detecting international shipping, we should allow user to fill out a form for customs. This screen should be navigated from the Customs field of the Create Shipping Label form:

<img src="https://user-images.githubusercontent.com/495617/125462081-0cf9b402-7fb5-4579-9de7-42249d39622c.png" />

itsmeichigo:
To do:
- [x] Create new model for customs form.
- [x] Update `ShippingLabelPackagePurchase` with the addition of customs forms.
- [x] Update `ShippingLabelPackageSelected` with the addition of customs forms.
- [x] Create `ShippingLabelCustomsFormListViewModel` with input: order, custom forms, stores, storage manager.
- [x] Create `ShippingLabelCustomsForm` (SwiftUI view)
- [x] Update `ShippingLabelFormViewModel` with customs forms and validation of customs.
- [x] Navigate from shipping label form to customs form.
- [x] Update customs form UI with input info.
- [x] Handle validation of customs info.
- [x] Handle submitting customs form.

PR #4740, opened 2021-08-06
Shipping Labels: Navigate from creation form to customs form list

itsmeichigo:
Part of #4687 

# Description
This PR aims to handle navigation between Shipping Label form and customs form list. This navigation is handled when tapping "Done" button on the customs cell added in #4723, or tapping on that cell when customs forms are input and validated. The UI for the customs form list is still a placeholder and will be implemented in the next PR.

# Changes
- Added new view and view model for customs form list. This is called a list since I want to prepare for multi-package support, and the view model should accept a list of customs forms instead of just one.
- Added a placeholder view for customs form input, which is also empty for now. The customs form list should include of several input views for each package associated with the shipping label.
- Added new property `packageID` to associate a customs form model with a package. This is for use in development only and not to encode and send to remote.
- Updated `ShippingLabelCustomsForm` model with a convenient intializer to accept package ID and product ID list as input. This is helpful for creating default customs forms for each package with associating order items.
- Updated `ShippingLabelFormViewModel` to setup default customs forms based on currently selected package ID. These forms will be used as initial input for the customs form list.
- Added new property `customsForms` and updated how customs forms are set to selected package and purchase package.
- Updated handling changes of data for shipping label form rows, to make sure that both customs and carrier cells are reset when updating package details. Carrier cells should also be updated when updating customs forms.
- Implement navigation between the shipping label form and customs form list. When initially configuring customs form, the default form list is used as input; otherwise the configured form list will be used.

# Testing
1. Create an order on your test store that's eligible for international shipping (different countries for origin and destination addresses; or shipping between US and US military states).
2. Navigate to Orders tab, select the new order.
3. Select Create Shipping Label, skip through ship from and to and package details.
4. Tap on Continue button on Customs row, you should be navigated to a new screen with placeholder text "Hello World". The content of this view will be implemented in the next PR.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
+++ b/Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
@@ -47,38 +47,7 @@ public struct ShippingLabelPackageSelected: Equatable, GeneratedFakeable {
 }
 
 // MARK: Codable
-extension ShippingLabelPackageSelected: Codable {
-
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-
-        let boxID = try container.decode(String.self, forKey: .boxID)
-        let length = try container.decode(Double.self, forKey: .length)
-        let width = try container.decode(Double.self, forKey: .width)
-        let height = try container.decode(Double.self, forKey: .height)
-        let weight = try container.decode(Double.self, forKey: .weight)
-        let isLetter = try container.decode(Bool.self, forKey: .isLetter)
-
-        let customsForm: ShippingLabelCustomsForm? = try? {
-            let contentsType = try container.decode(ShippingLabelCustomsForm.ContentsType.self, forKey: .contentsType)
-            let contentsExplanation = (try? container.decode(String.self, forKey: .contentsExplanation)) ?? ""
-            let restrictionType = try container.decode(ShippingLabelCustomsForm.RestrictionType.self, forKey: .restrictionType)
-            let restrictionComments = (try? container.decode(String.self, forKey: .restrictionComments)) ?? ""
-            let nonDeliveryOption = try container.decode(ShippingLabelCustomsForm.NonDeliveryOption.self, forKey: .nonDeliveryOption)
-            let itn = (try? container.decode(String.self, forKey: .itn)) ?? ""
-            let items = try container.decode([ShippingLabelCustomsForm.Item].self, forKey: .items)
-
-            return ShippingLabelCustomsForm(contentsType: contentsType,
-                                            contentExplanation: contentsExplanation,
-                                            restrictionType: restrictionType,
-                                            restrictionComments: restrictionComments,
-                                            nonDeliveryOption: nonDeliveryOption,
-                                            itn: itn,
-                                            items: items)
-        }()
-
-        self.init(boxID: boxID, length: length, width: width, height: height, weight: weight, isLetter: isLetter, customsForm: customsForm)
-    }
+extension ShippingLabelPackageSelected: Encodable {
 
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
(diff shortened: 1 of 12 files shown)


Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
```diff
@@ -47,38 +47,7 @@ public struct ShippingLabelPackageSelected: Equatable, GeneratedFakeable {
...
-                                            restrictionType: restrictionType,
-                                            restrictionComments: restrictionComments,
-                                            nonDeliveryOption: nonDeliveryOption,
-                                            itn: itn,
-                                            items: items)
-        }()
-
-        self.init(boxID: boxID, length: length, width: width, height: height, weight: weight, isLetter: isLetter, customsForm: customsForm)
-    }
+extension ShippingLabelPackageSelected: Encodable {
```
pmusolino, line 50: I love the smell of code removed 💯

Networking/Networking/Model/ShippingLabel/Packages/ShippingLabelCustomsForm.swift
```diff
@@ -4,6 +4,13 @@ import Codegen
 /// Represents customs info for a shipping label package
 ///
 public struct ShippingLabelCustomsForm: Equatable, GeneratedFakeable {
+    /// ID of the associated package.
+    ///
+    /// This is for identifying the package when inputing customs form only,
+    /// no need for encoding and sending to remote.
+    ///
+    public let packageID: String
```
pmusolino, line 12: It makes sense to me. Not sure what will represent a `packageID` at the moment, because it should be if I'm right, an internal implementation.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormList.swift
```diff
@@ -0,0 +1,29 @@
+import SwiftUI
+import Yosemite
+
+struct ShippingLabelCustomsFormList: View {
+    @ObservedObject private var viewModel: ShippingLabelCustomsFormListViewModel
```
pmusolino, line 5: It makes a lot of sense to me 👍 I would have used the same approach to manage the customs list

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -42,6 +42,21 @@ final class ShippingLabelFormViewModel {
...
+    /// Temporary solution for creating default customs forms.
+    /// When multi-package support is available, we should create separate form for each package ID.
+    ///
+    var defaultCustomsForms: [ShippingLabelCustomsForm] {
+        guard let packageID = selectedPackageID else {
+            return []
+        }
+        let productIDs = order.items.map { $0.productOrVariationID }
+        return [ShippingLabelCustomsForm(packageID: packageID, productIDs: productIDs)]
+    }
```
pmusolino: Why not using directly the property `customsForms`?

pmusolino (approved):
Looks good to me @itsmeichigo. Your solution to support custom forms for multi-package seems ok to me, I would have used the same approach. Good job 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
itsmeichigo: This makes sense - I have updated in 12bc6847ee78f6c1b2e5e8eca6e20392906a732d to reset customs forms to default values when a new package is updated.

Networking/Networking/Model/ShippingLabel/Packages/ShippingLabelCustomsForm.swift
itsmeichigo: That's correct - it's an internal implementation. I'm not quite happy with this, but I have yet to come up with a better solution.

PR #4741, opened 2021-08-06
Adds "Draft" label below the image on the product editing screen

### LINKED ISSUE #4544: Add “Draft” Label When Editing a Draft Product
Ref p91TBi-5oi-p2#comment-5224.

When editing a draft product, Android shows a “Draft” label below the product image. We should do the same for iOS. 

image

iamgabrielma:
Fixes #4544

## Description

When editing a draft product we want to show a “Draft” label below the product image for consistency with the Android app:

completed

## Changes

| Draft products status  | Other product status |
|---------------------|-----------------|
| Screenshot 2021-08-09 at 14 50 05 | Screenshot 2021-08-07 at 16 53 25 |

| Landscape View |
|---------------------|
| Screenshot 2021-08-09 at 14 50 15 |

| Example of LTR screen & Localized strings |
|---------------------|
| Screenshot 2021-08-09 at 15 05 49 |

I've added a new `UILabel` to `ProductImagesHeaderTableViewCell.xib` via the Interface Builder . Then within the `ProductFormTableViewDataSource` I'm performing a check to see if the product status is a draft, so this label will only be displayed if the product status is set to `draft`, no label will be shown otherwise. 

I used existing styles that we use elsewhere in order to apply some sort of design, in this case using the same style as completed orders via `applyStyle(for: .completed)` seemed appropriate as fits the the style shown in the Android version of the app.

## Work in progress:
- [x] In order to fit as close as possible the Android one, I need to style it further by adding inner padding and changing the text colour.
- [x] Move the new label cell configuration outside the Data Source.
- [x] Fix UI constraints
- [x] Test for localized strings
- [x] Test for dark mode
- [x] Move the `.cell` appearance-only config to be called from `awakeFromNib()` instead, as these do not depend on data updates.

## Testing steps
- Create a new product, or edit any existing product.
- If the product is currently a `draft` , the `draft` label will appear below the image cell.
- In the top right corner, tap on "..." > Product Settings > Status
- Switching between different status will make the label appear only when `draft` is selected, and disappear on other statuses.

Update release notes:


iamgabrielma:
With the [latest commit](007982adf1f698440740ea0bbd3b1ca527ac406a) the label is now more similar to its Android version (white text & grey background):

Screenshot 2021-08-09 at 18 50 27
Screenshot 2021-08-09 at 18 50 39

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
```diff
@@ -33,7 +33,7 @@ private extension DefaultProductFormTableViewModel {
         return actions.map { action in
             switch action {
             case .images(let editable):
-                return .images(isEditable: editable, allowsMultiple: product.allowsMultipleImages())
+                return .images(isEditable: editable, allowsMultiple: product.allowsMultipleImages(), productLabel: product.status.rawValue)
```
koke: One thing I wonder here is if this makes sense only for drafts, or also for any status other than publish.

In any case, passing the `ProductStatus` from `product.status` might be more convenient than using the raw value, since you can compare agains `.draft` to check if it's a draft, or use the `description` method to get the label text instead of hardcoding the "Draft" string in the view.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewDataSource.swift
```diff
@@ -121,7 +119,17 @@ private extension ProductFormTableViewDataSource {
...
-
+        if productLabel.contains("draft") {
+            cell.productStatusLabelHolder.backgroundColor = UIColor.gray
+            cell.productStatusLabelHolder.layer.cornerRadius = CGFloat(2.0)
+            cell.productStatusLabelHolder.layer.borderWidth = CGFloat(0.5)
+            let statusLabel = NSLocalizedString("Draft", comment: "Display label for the product's draft status")
+            cell.productStatusLabel.text? = statusLabel
+            cell.productStatusLabel.font = UIFont.preferredFont(forTextStyle: .body)
+            cell.productStatusLabel.textColor = UIColor.white
+        } else {
```
koke: A "data source" object should probably not be configuring the internals of the cell. A better approach would be to extend the `ProductImagesHeaderTableViewCell` `configure` method to take the product status and handle the label logic internally.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewDataSource.swift
```diff
@@ -121,7 +119,17 @@ private extension ProductFormTableViewDataSource {
...
+            cell.productStatusLabel.text = ""
```
koke: This isn't enough to clear the badge. When you are viewing a draft product and hit publish, the badge will clear the text but leave the gray box. If you leave the screen and come back (or visit a published product), there will be a white box on top of the first image:

Screen Shot 2021-08-09 at 17 30 35
Screen Shot 2021-08-09 at 17 30 38

koke (commented):
Commenting inline on XIB files is awkward so I'll comment here. It seems the holder is using autoresizing masks instead of layout constraints, which seems to work initially, but it won't adapt to a different size when using accessibility font sizes or when the text is localized:

Screen Shot 2021-08-09 at 17 38 48

It also doesn't look right on dark mode, but I'm not sure what the right colors should be for that, so I'd ping a designer for a UI review as well.

For the layout, you probably want to set constraints to pin the view to the bottom leading edge (perhaps with some offset), but then let the holder grow as the text gets larger. For that, the label should use one of the dynamic fonts. Again, not sure of which one since I can't see this label in the design files, so maybe @woocommerce/mobile-designers can help with the exact font/color/dimensions here.

WooCommerce/Classes/ViewRelated/Products/Edit Product/DefaultProductFormTableViewModel.swift
iamgabrielma: > One thing I wonder here is if this makes sense only for drafts, or also for any status other than publish.

That's something that crossed my mind as well. At the time I only added the option for "Draft" in order to replicate how the Android app works, but it would definitely make sense for other statuses and I'm happy to make the changes. I'll be opening a new enhancement request in the Android repo as well to get feedback on this.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewDataSource.swift
iamgabrielma: Gotcha! I missed this when testing 😓

iamgabrielma:
> It seems the holder is using autoresizing masks instead of layout constraints, which seems to work initially, but it won't adapt to a different size when using accessibility font sizes or when the text is localized:
> It also doesn't look right on dark mode, but I'm not sure what the right colors should be for that, so I'd ping a designer for a UI review as well.
> For the layout, you probably want to set constraints to pin the view to the bottom leading edge (perhaps with some offset), but then let the holder grow as the text gets larger. For that, the label should use one of the dynamic fonts. Again, not sure of which one since I can't see this label in the design files, so maybe @woocommerce/mobile-designers can help with the exact font/color/dimensions here.

Thanks for the pointers! With the latest commit I believe to have resolved this by setting layout constraints indeed, rather than the existing autoresizing mask, for now I haven't modified the colours in dark mode as seems we're not modifying them neither in other places like Order statuses, however is true that the current grey/white combination is not great with a black background.

I'll give it a few more tests and wait for @woocommerce/mobile-designers feedback on styling the labels before requesting a new review

| English label | Localized Label  | Dark mode |
|---------------------|---|-----------------|
| autolayout-english | autolayout-spanish | Screenshot 2021-08-11 at 15 48 05 |

Garance91540:
Hi! @adamzelinski do you have some time to provide some designs for this issue this week?

adamzelinski:
Hi @iamgabrielma, thanks for working on this. I think we should just use our current status pills for the draft status. I also think we should add the pill under the image instead of over the top of it. I've put together a design for this. Thanks :)

Screen Shot 2021-08-12 at 12 14 20 pm

iamgabrielma:
| Light mode | Dark mode  | Dark mode with no images |
|---------------------|---|-----------------|
| update_20210813_Screenshot 2021-08-13 at 12 03 45 | update_20210813_Screenshot 2021-08-13 at 12 03 24 | update_20210813_Screenshot 2021-08-13 at 12 02 28 |


Latest changes:

- Moved the cell configs that do not deal directly with data changes to its own function to be called from `awakeFromNib()`
- Fixed label and text constraints to fit localized strings and to look good in dark mode
- Applied design recommendations and now follows the "Order Status" pill design except for the positioning: While I adjusted it to the bottom of the view, is not fully under the image for various reasons:
	- It would involve modify the whole style of `ProductImagesHeaderTableViewCell` as well as push both the images to the top and the rest of the Order details to the bottom. This causes visibility problems on smaller devices and is inconsistent when the product is not a Draft, as the space reserved for this label is not used.
	- For consistency with the Android version, where the pill covers part of the image on purpose.

Happy to make any further modifications!

ghost:
Hi, @iamgabrielma here is the design for adding the draft tag. Can we please just use the current status pills we use next to the Product name.

Thanks!

Screen Shot 2021-08-13 at 4 05 27 pm

iamgabrielma:
Thanks for the pointers @adamzelinski ! I have a question before proceeding: Would we consider to use an icon rather than a label with text? 

While checking the current textfield that is being used for the product text I realized that there's already some code in place that is not being used and its purpose is precisely [to show an icon when necessary](https://github.com/woocommerce/woocommerce-ios/pull/176#discussion_r205552651).

This is the current structure of the field:

tttvcell-before

If we add the label with a text, this is how it would be:

tttvcell-after

By using one of [Apple's Human Interface Guideline icons](https://developer.apple.com/design/human-interface-guidelines/ios/icons-and-images/system-icons/) here rather than label and text we could avoid the clutter as well as potential problems like the length of a localized text overlapping or pushing the product name.

This is roughly how it would look like ( I used the `.square.and.pencil` icon, and applied a generic grey just for testing ):

Screenshot 2021-08-15 at 19 44 52

Let me know if that would make sense, or if we should stick to the original idea of using a label & text.

ghost:
Hey @iamgabrielma, I think we should just use the label instead of an icon for now as it's an existing component we use to display the status of something. Definitely, something we will keep in mind though so thanks for looking into this :)

iamgabrielma:
I'm moving this PR back to `draft` for the moment, as the latest changes require to redo it from zero 😅  

At this stage I had to choices for the new implementation: To create a new cell, or reuse an existing one. `TitleAndTextFieldTableViewCell` seems to adapt well however this cell is already being used across the app and I can't modify it without affecting other places, so I opted to create a new cell for this: `LabeledTextViewTableViewCell` based on this one.

TODO:
- [x] Revert changes at `ProductImagesHeaderTableViewCell.swift` , `ProductImagesHeaderTableViewCell.xib`, `TitleAndTextFieldTableViewCell.xib`
- [x] Editing does not trigger `onTextChange` 
- [x] Fix first responder when the new cell is tapped.
- [x] Layout should adapt to the screen when there's no label.
- [x] Fix cell styling consistency, and label autolayout with previous implementation.

iamgabrielma:
Alright, I believe this should be good (or pretty close) now:

-  In `DefaultProductFormTableViewModel` now we're passing the product status inside the `primaryFieldRows` when generating the  `.name` row.
- This returns a new prototype cell `LabeledTextViewTableViewCell` rather than the previous `TextViewTableViewCell` 
- This new cell contains a `UILabel` and a textfield ( using `EnhancedTextView` ). The label will display our product status if is set to draft, and nothing otherwise.

| Draft Product | Draft Product (Dark Mode)  | Localized | No label |
|---------------------|-----------------|-----------------|-----------------|
|Simulator Screen Shot - iPhone 8 - 2021-08-29 at 15 52 14 | Simulator Screen Shot - iPhone 8 - 2021-08-29 at 16 03 06 | Simulator Screen Shot - iPhone 8 - 2021-08-29 at 16 04 59 | Simulator Screen Shot - iPhone 8 - 2021-08-29 at 15 52 57|

Let me know if needs further iterations 👍

ghost:
@iamgabrielma looking good so far thanks! Just noticed a few little things to tweak.

Screen Shot 2021-08-30 at 2 06 32 pm

WooCommerce/Classes/ViewRelated/Products/Cells/Product Images/ProductImagesHeaderTableViewCell.xib
```diff
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="15505" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
-    <device id="retina6_1" orientation="portrait" appearance="light"/>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="18122" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
```
koke: If I understand correctly, `ProductImagesHeaderTableViewCell` isn't being modified anymore, so maybe this change can be reverted? The diff looks like Xcode just messing around with XIB files with no meaningful changes

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.xib
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="16096" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="18122" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
```
koke: Similar to `ProductImagesHeaderTableViewCell`, this looks like Xcode messing around with XIB files with no meaningful changes, so maybe we can revert this.

WooCommerce/Classes/ViewRelated/ReusableViews/TitleAndTextFieldTableViewCell.xib
```diff
@@ -1,9 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="17701" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.XIB" version="3.0" toolsVersion="18122" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES">
```
koke: Another noisy Xcode diff with no changes 😬

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormTableViewDataSource.swift
```diff
@@ -110,53 +109,49 @@ private extension ProductFormTableViewDataSource {
...
-            configureEditableName(cell: cell, name: name)
+            configureEditableName(cell: cell, name: name, productLabel: productLabel)
         } else {
             configureReadonlyName(cell: cell, name: name ?? "")
         }
     }
 
-    func configureEditableName(cell: UITableViewCell, name: String?) {
-        guard let cell = cell as? TextViewTableViewCell else {
+    func configureEditableName(cell: UITableViewCell, name: String?, productLabel: ProductStatus) {
```
koke: Minor nitpick, but if we're using "status" both for the data source and the cell view model, maybe we should use `productStatus` here as well.

WooCommerce/Classes/ViewRelated/ReusableViews/LabeledTextViewTableViewCell.swift
```diff
@@ -0,0 +1,88 @@
...
+        productStatusLabel.font = UIFont.preferredFont(forTextStyle: .body)
+        productStatusLabel.textAlignment = .center
+        productStatusLabel.textColor = UIColor.black
+        productLabelHolder.backgroundColor = .gray(.shade5)
+        productLabelHolder.layer.cornerRadius = CGFloat(4.0)
+    }
+
+    func configureProductStatusLabel(productStatus: ProductStatus) {
+            if productStatus == ProductStatus.draft {
+                let statusLabel = NSLocalizedString(productStatus.description, comment: "Display label for the product's draft status")
```
koke: This won't work for localization, only literal strings are picked up for translation. In this case, since we only show the label for `ProductStatus.draft`, you could go ahead and use `"Draft"` directly.

That said, if the cell can only handle drafts, maybe a `isDraft` flag is more appropriate than passing the full status.

WooCommerce/Classes/ViewRelated/ReusableViews/LabeledTextViewTableViewCell.swift
```diff
@@ -0,0 +1,88 @@
...
+
+                productStatusLabel.isHidden = false
+                productStatusLabel.text? = statusLabel
+            } else {
+                productLabelHolder.isHidden = true
```
koke, line 64: The previous block in the if/else is setting `isHidden` on `productStatusLabel`, but this one is changing `productLabelHolder`. I think both should change the holder?

Right now, if I product is a draft and you change the status the badge disappears, but if a product is published, and you revert to draft, the badge won't show up until you close the product view and open it again.

koke (commented):
The code looks so much simpler now than when it was on top of the images 😅 
I left mostly minor cleanup comments, except for the one about the hiding logic. I'm holding on reviewing the label in more detail, since there are more design changes requested, but the general approach looks fine.

WooCommerce/Classes/ViewRelated/Products/Cells/Product Images/ProductImagesHeaderTableViewCell.xib
iamgabrielma: Thanks! Yes for all 3 `ProductFormViewController.xib`, `ProductImagesHeaderTableViewCell.xib`, and `TitleAndTextFieldTableViewCell.xib` should return as they were before the initial changes on the top of the images. I wasn't 100% sure how to do it without polluting further the commit history, just reverting the files separately should do, right ?

koke:
Also, I'm seeing some auto layout errors:

```
2021-08-31 09:27:22.362020+0200 WooCommerce[1094:716536] [LayoutConstraints] Unable to simultaneously satisfy constraints.
	Probably at least one of the constraints in the following list is one you don't want. 
	Try this: 
		(1) look at each constraint and try to figure out which you don't expect; 
		(2) find the code that added the unwanted constraint or constraints and fix it. 
(
    "<NSLayoutConstraint:0x282189180 H:[UIView:0x11e5b5a90]-(5)-[WooCommerce.EnhancedTextView:0x10c93c000'Wapuu Plush Toy']   (active)>",
    "<NSLayoutConstraint:0x28218a080 'UISV-spacing' H:[UIView:0x11e5b5a90]-(42)-[WooCommerce.EnhancedTextView:0x10c93c000'Wapuu Plush Toy']   (active)>"
)
```

It looks like there is a manual constraint for the spacing between the status holder and the product name (5), but the stack view that contains them would already manage the spacing between children, so there's a conflict there. You can remove that constraint and adjust the stack view's spacing to 5 to achieve the same effect.

WooCommerce/Classes/ViewRelated/Products/Cells/Product Images/ProductImagesHeaderTableViewCell.xib
koke: > just reverting the files separately should do, right?

If that's possible then it sounds like a good option. Personally I don't worry too much about commit history these days, as I see the PR as the unit of work. To be honest, I thought our repo was set up to squash and merge by default, but it seems it's not 😬 I would probably squash/merge this PR since the final approach is completely different than the initial one, and it's a pretty self-contained change, but I have no strong preference either way.

iamgabrielma:
Changes: 
- Fixed the missing background colour in dark mode
- Reduced the label size to make it similar to the order status labels
- Removed the top divider. This is actually the bottom divider from the `ProductImagesHeaderTableViewCell`, I also added more space to the bottom of the new cell so looks it looks centered and aligned correctly after removing the divider:

Screenshot 2021-09-04 at 18 41 14


- Reverted noise for `ProductImagesHeaderTableViewCell.xib`, `ProductFormViewController.xib`, and `TitleAndTextFieldTableViewCell.xib` to their own respective previous commits.
- Refactored `productLabel` to `productStatus`
- Changed `NSLocalizedString` to accept the String literal 
- Fixed auto-layout errors by removing constraints from the child elements and just setting them on the StackView.
- Fixed label holder appearance logic. I have not used a `isHidden` check here as there's an open issue to potentially add other states to these labels: https://github.com/woocommerce/woocommerce-android/issues/4587

WooCommerce/Classes/ViewRelated/ReusableViews/LabeledTextViewTableViewCell.swift
```diff
@@ -0,0 +1,89 @@
...
+
+    func configure(with viewModel: ViewModel) {
+        productTextField.text = viewModel.text
+        productTextField.placeholder = viewModel.placeholder
+        productStatusLabel.text = viewModel.productStatus.description
+        productTextField.isScrollEnabled = viewModel.isScrollEnabled
+        productTextField.onTextChange = viewModel.onNameChange
+        productTextField.onTextDidBeginEditing = viewModel.onTextDidBeginEditing
+        productTextField.keyboardType = viewModel.keyboardType
+        configureProductStatusLabel(productStatus: viewModel.productStatus)
```
koke, line 36:
```suggestion
        configureProductStatusLabel(productStatus: viewModel.productStatus)
        applyStyle(style: viewModel.style)
```

It looks like this method is missing a call to `applyStyle`, so the text field isn't properly configured, and the font won't respond to dynamic text changes:

Screen Shot 2021-09-06 at 11 11 51

koke (commented):
I found a couple issues besides the inline comment. The first is that the row is still selectable, so when you tap outside the product title field, the whole row highlights, but the product background is white instead of clear, so it looks weird. I think the row shouldn't be selectable, but if it is, the backgrounds should match:

Screen Shot 2021-09-06 at 11 16 11

The second is trickier, and it's that the draft label and name don't look visually aligned:

Screen Shot 2021-09-06 at 11 18 36

Both views have the same height, but since they use different fonts, the baselines are not aligned, and having different backgrounds makes it more jarring. A stack view supports aligning to baselines, but it doesn't work in this case. At first I thought it was because the draft label was wrapped within another view, but removing that didn't work. I think the reason is that the product name is a UITextView instead of a UITextField, and so it can't align on baseline because it can have multiple lines. I'm not sure why we are using a text view there in the first place, but we might have to replace it if we want these to be vertically aligned.

WooCommerce/Classes/ViewRelated/ReusableViews/LabeledTextViewTableViewCell.swift
iamgabrielma: Thanks for pointing this out! I may have missed this bit while trying to adapt the new cell to the old implementation. By your screenshot, applying the current style doesn't seem to fit the overall design and font size ( is too small ), so I'll iterate over it and find a more suitable font and size for this.

iamgabrielma:
Latest changes:
- I set to `none` the cell's selection style, so the row is no longer highlighted when we tap outside the product field.
- I added `applyStyle(style: viewModel.style)` so the `.headline` style for product title is applied properly:

Screenshot 2021-09-24 at 14 25 36

- For the alignment between the status label and product text I'll be opening a new issue that addresses this specifically. As discussed in Slack we're using a UITextView here because of the multi-line support, but this also causes the miss-alignment and may require manual alignment if we want to keep this implementation.
- Resolved merge conflict.

koke (approved):
One thing I noticed is that the text doesn't left-align with the other cells when there's no badge, but we can handle that as we solve the other alignment issue:

Simulator Screen Shot - iPhone 12 mini - 2021-09-24 at 16 36 17

This is looking pretty good already, thanks for putting up with all the change requests 😅

iamgabrielma:
> One thing I noticed is that the text doesn't left-align with the other cells when there's no badge, but we can handle that as we solve the other alignment issue.

Thanks for the review! I'll be sure to add this bit to the new issue as well :D

PR #4750, opened 2021-08-09
[CPP Onboarding] Implement unsupported country UI

koke:
Part of #4611 

This adds a specific error screen when the store's address is in a country that doesn't support In-Person Payments yet.

A big part of this PR is that it's the first screen implemented that includes a link to contact support from a SwiftUI view, so I had to add a way to handle that. I'm still using `AttributedText` with a custom URL, but I've found a better way to handle the callbacks.

In #4734, I introduced AttributedText with SafariSheet as a native way to present a web view from SwiftUI. However, that required some quirky tricks with bindings. In this case, for contacting support, everywhere else in the app we push a new view to the navigation stack instead of presenting a modal, so a sheet wouldn't work. I can't see a nice way to bridge that gap and push a new UIViewController from a SwiftUI view, so I'm delegating the responsibility to the container VC. Because `customOpenURL` is an environment value, it's ok to define up the stack, so now `InPersonPaymentsView` does that since it can communicate directly with the container VC. I've also replaced the safari sheet usage with this shared URL handling.

## To test

1. Set up a store in a country other than US
2. Go to Settings > In-Person Payments
3. The app should show the country not supported screen
4. Tapping learn more should show a web view with more information about WooCommerce Payments
5. Tapping Contact Support should open a new Zendesk request

Country not supported | Tapped contact support
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-08-09 at 13 10 19|Simulator Screen Shot - iPhone 12 Pro - 2021-08-09 at 13 10 22

Update release notes:


ctarda (approved):
I works well, Learn More and Contact Us do what they are supposed to do. 

The "Need some help" label is not visible in Dark Appearance though. 



Other than that, :shipit:

koke:
> The "Need some help" label is not visible in Dark Appearance though.

Fixed in ade4c1be0

Simulator Screen Shot - iPhone 12 Pro - 2021-08-09 at 15 56 32

PR #4751, opened 2021-08-09
[CPP Onboarding] Implement view for plugin not activated

ctarda:
Part of #4611 

This adds a specific onboarding error screen for the case when WCPay is installed but not activated.

## To test

You should have a store with an address in the US and with the WooCommerce Payments installed.

1. From wp-admin, deactivate WooCommerce Payments from Plugins > Installed Plugins.
2. Go to Settings > In-Person Payments. The app should show a screen prompting you to activate the extension.
3. From wp-admin, activate WooCommerce Payments from Plugins > Installed Plugins.
3. Tap on Refresh After Activating.

| Extension not activated | After tapping Refresh |
| ---- | ---- |
|  |  |

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsPluginNotActivatedView.swift
```diff
@@ -0,0 +1,55 @@
...
+                Text(Localization.title)
+                    .font(.headline)
+                    .multilineTextAlignment(.center)
+                Image(uiImage: .paymentsPlugin)
+                    .resizable()
+                    .scaledToFit()
+                    .frame(height: 108.0)
+                Text(Localization.message)
+                    .font(.callout)
+                    .multilineTextAlignment(.center)
```
koke: We can consolidate this at the end, but in #4750 I realized we can put the `multilineTextAlignment` once in the `VStack` and all the Text views inside it will inherit that.

koke (approved):

WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsPluginNotActivatedView.swift
ctarda: I've just pushed an update to apply the style to the Stack. It takes just one minute, and it's one less thing to worry about.



I'll merge as soon as CI is ✅

PR #4752, opened 2021-08-09
[CPP Onboarding] Implement view for plugin version not supported

ctarda:
Part of #4611 

This adds a specific onboarding error screen for the case when the version of WooCommerce Payments installed is lower than the minimum supported version.

I made up the error messages completely.

Also, this PR removes a warning in SettingsViewController. I am not sure if this is really necessary, but I noticed it so 🤷🏼 

## To test

There is a quick way to test it and a long and painful way to test this. Your choice! 

The quick way:
1. Edit CardPresentPaymentsOnboardingUseCase. In line 201, change:
`static let supportedWCPayVersion = "2.5"` 
to 
`static let supportedWCPayVersion = "5.0"`

2. Login to an store with WCPay installed (ideally, the store would be fully configured and gated into In-Person Payments
3. Notice the error message:

The long and painful way:
1. Install a version of WCPay lower than 2.5
1. Go to Settings > In-Person Payments. The app should show a screen prompting you to update the extension.
2. From wp-admin, update WooCommerce Payments from WooCommerce > Extensions > Payments.
3. Tap on Refresh After Updating.

Update release notes:


koke:
> I made up the error messages completely.

I see [Android is using](https://github.com/woocommerce/woocommerce-android/blob/3c02b1d77ef994554d5c8f1a0d6295485b856683/WooCommerce/src/main/res/values/strings.xml#L859):

> Outdated version of the WooCommerce Payments extension is installed on your store. Please update it to accept In-Person Payments.

That sounds a bit odd to me, we should figure out the right copy and consolidate both platforms

koke (approved):
Looks good. We can update the copy separately once we agree on the right one 🚢

PR #4753, opened 2021-08-09
Show extension metadata in order item details

### LINKED ISSUE #4708: Measurement Price Calculator: Metadata that are not Variation Attributes are not shown in Order Details and Fulfillment screens
## Describe the bug

Reported by a merchant:  4182704-zd-woothemes
Discussed in: p1627660879010800-slack-C6H8C3G23

I'm only able to reproduce this using the [Measurement Price Calculator](https://woocommerce.com/products/measurement-price-calculator/) plugin. If using the plugin, the metadata is displayed in Order Details.

Android displays the metadata from the plugin as shown here:

But iOS does not display it:

## To Reproduce

1. Install the Measure Price Calculator plugin. 
2. Set up a product so it has the "Calculated Price" and "Calculated Weight" checked. I'm not sure if having one field is enough but this is what I did when I tried to reproduce the problem. 
    
3. Create an order on the web. Enter a value in the Required Weight:
    
4. Submit the order.
5. View the order in the app. 

## Expected behavior

Since Android is able to display this, there might be some differences in how we parse the metadata. We can try and follow how Android does the metadata parsing.

Ecarrion:
fix #4708

# Why

It was reported(4182704-zd-woothemes) that merchants were able to see some order metadata in Android but not in iOS.

After some investigation, this happens because some extensions(like [Measurement Price Calculator](https://woocommerce.com/products/measurement-price-calculator/)) adds some extra metadata that does not have the format we expect, hence making the whole metadata/attributes decoding to fail.

# How

This PR solves the problem by making the `value` field from `OrderItemAttribute` to be decoded safely when an unexpected format is found.

# Screenshots 

Core | Before | After
--- | --- | ---
core | bad | good

# Testing Steps

Follow testings steps on https://github.com/woocommerce/woocommerce-ios/issues/4708

Update release notes:

### DIFF
--- a/Networking/Networking/Model/OrderItemAttribute.swift
+++ b/Networking/Networking/Model/OrderItemAttribute.swift
@@ -22,7 +22,7 @@ public struct OrderItemAttribute: Decodable, Hashable, Equatable, GeneratedFakea
         let metaID = try container.decode(Int64.self, forKey: .metaID)
         let name = try container.decode(String.self, forKey: .name)
 
-        /// Some extensions sent the `value` field with non-string format.
+        /// Some extensions send the `value` field with non-string format.
         /// We don't support those, but we also don't want the whole decoding to fail.
         let value = container.failsafeDecodeIfPresent(String.self, forKey: .value) ?? ""
 
--- a/Networking/Networking/Model/OrderItemAttribute.swift
+++ b/Networking/Networking/Model/OrderItemAttribute.swift
@@ -21,7 +21,10 @@ public struct OrderItemAttribute: Decodable, Hashable, Equatable, GeneratedFakea
         let container = try decoder.container(keyedBy: CodingKeys.self)
         let metaID = try container.decode(Int64.self, forKey: .metaID)
         let name = try container.decode(String.self, forKey: .name)
-        let value = try container.decode(String.self, forKey: .value)
+
+        /// Some extensions sent the `value` field with non-string format.
+        /// We don't support those, but we also don't want the whole decoding to fail.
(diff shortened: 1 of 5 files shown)


Networking/Networking/Model/OrderItemAttribute.swift
```diff
@@ -21,7 +21,10 @@ public struct OrderItemAttribute: Decodable, Hashable, Equatable, GeneratedFakea
         let container = try decoder.container(keyedBy: CodingKeys.self)
         let metaID = try container.decode(Int64.self, forKey: .metaID)
         let name = try container.decode(String.self, forKey: .name)
-        let value = try container.decode(String.self, forKey: .value)
+
+        /// Some extensions sent the `value` field with non-string format.
```
allendav: sent -> send

Networking/NetworkingTests/Mapper/OrderMapperTests.swift
```diff
@@ -325,6 +337,12 @@ private extension OrderMapperTests {
         return mapOrder(from: "order-with-line-item-attributes")
     }
 
+    /// Returns the OrderMapper output upon receiving `order-with-faulty-attributes`
+    ///
+    func mapLoadOrderWithFaultyAttributesResponse() -> Order? {
```
allendav, line 343: Please consider mentioning in a comment that the `_measurement_data` value having object format (and not a string) is what is "faulty" here. I had to fire up the debugger to understand what in the order was the problem.

allendav (approved):
Tests fine. Two suggestions to consider. :shipit:

Ecarrion:
Thanks for the review and suggestions @allendav!

PR #4754, opened 2021-08-09
Shipping Labels: Add the button "Create new package" in the package selected screen

### LINKED ISSUE #4742: Shipping Labels: Add the button "Create new package" in the package selected screen.
This is a subtask of #3909.

Add the button "Create new package" in the package selected screen. This button will enter the package creation flow.

Be sure to use the `shippingLabelsAddCustomPackages` feature flag when adding this button, so it's only visible if this feature is enabled.

iOS-shipping-labels-create-package

rachelmcr:
Resolves: #4742

## Description

Adds the button "Create new package" when creating a new shipping label, at the bottom of the Package Selected screen.

This button will allow users to create a new package to use with their shipping label (to be implemented in a future PR; for now the button doesn't do anything).

## Changes

* Adds a LinkButtonStyle to the SwiftUI `ButtonStyles`. This is the style used for the "Create new package" button.
* Adds a new reusable SwiftUI component `BottomButtonView`, which acts like our reusable UIKit `BottomButtonContainerView` (a view that can be used to add a button with the specified styles to the bottom of a SwiftUI view).
* Adds the "Create new package" button to the bottom of the Package Selected screen (`ShippingLabelPackageList`), by adding a new `VStack` that includes the existing `ScrollView` and a new `BottomButtonView`. This button is only displayed when the `shippingLabelsAddCustomPackages` feature flag is enabled.

Before|After (light)|After (dark)|After (short list)
-|-|-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-08-09 at 17 22 11|Simulator Screen Shot - iPhone 11 Pro - 2021-08-09 at 16 37 44|Simulator Screen Shot - iPhone 11 Pro - 2021-08-09 at 16 37 28|Simulator Screen Shot - iPhone 11 Pro - 2021-08-09 at 15 53 02

## Testing

1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more payment methods and one or more packages set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
7. On the Package Selected screen, confirm the “Create new package” button appears at the bottom of the screen.

You can also test with the `shippingLabelsAddCustomPackages` feature flag disabled (in production mode or by manually disabling that flag) to confirm the screen looks the same as before when the button is not on screen.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/BottomButtonView.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/BottomButtonView.swift
@@ -1,0 +1,50 @@
+import SwiftUI
+
+struct BottomButtonView<Style>: View where Style: ButtonStyle {
+    let style: Style
+    let title: String
+    let image: UIImage?
+    let onButtonTapped: () -> Void
+
+    var body: some View {
+        VStack(spacing: 0) {
+            Divider()
+            Button(action: { onButtonTapped() }) {
+                HStack {
+                    if let image = image {
+                        Image(uiImage: image)
+                    }
+                    Text(title)
+                    Spacer()
+                }
+            }
+            .buttonStyle(style)
+        }.edgesIgnoringSafeArea([.bottom])
+    }
+}
+
+struct BottomButtonView_Previews: PreviewProvider {
+    static var previews: some View {
+        BottomButtonView(style: LinkButtonStyle(),
+                         title: "Bottom Button",
+                         image: .plusImage,
+                         onButtonTapped: {})
+            .previewLayout(.sizeThatFits)
+
+        BottomButtonView(style: LinkButtonStyle(),
+                         title: "Bottom Button",
+                         image: .plusImage,
+                         onButtonTapped: {})
+            .previewLayout(.sizeThatFits)
+            .environment(\.sizeCategory, .accessibilityExtraExtraExtraLarge)
+            .previewDisplayName("Large font")
+
+        BottomButtonView(style: LinkButtonStyle(),
+                         title: "Bottom Button",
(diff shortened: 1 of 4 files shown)


itsmeichigo (approved):
✅ Works as expected
I have a non-blocking comment though 👇

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/BottomButtonView.swift
```diff
@@ -0,0 +1,50 @@
...
+                HStack {
+                    if let image = image {
+                        Image(uiImage: image)
+                    }
+                    Text(title)
+                    Spacer()
+                }
+            }
+            .buttonStyle(style)
+        }.edgesIgnoringSafeArea([.bottom])
```
itsmeichigo: I think `edgesIgnoringSafeArea` is a bit case-specific, i.e in some other screens we may not want to show the button in non-safe area. How about moving it to the package list screen instead?

Also, I know that I've been using  `edgesIgnoringSafeArea` here and there but maybe we should use the newer variant [`ignoresSafeArea(_:edges:)`](https://developer.apple.com/documentation/swiftui/view/ignoressafearea(_:edges:)) since the old one may be deprecated soon 🤔

rachelmcr: Good point. I just played around with this again and realized that we don't need it even in this case, so I'll remove it entirely. (I thought I'd need it to extend the background to the edge, but I just tried again and the background extends even without it.)

Thanks for the tip about `ignoresSafeArea(_:edges:)` — I missed that it's the newer way of doing this!

Issue #97, opened 2018-06-14
Order List: Order Persistence


bummytime:
Add everything needed the support the `Order List: Wire Up to Yosemite` epic #25 in the storage framework.
* Data model
* MO's
* Tests
* etc

PR #4759, opened 2021-08-10
[CPP Onboarding] Implement onboarding loading UI

koke:
Part of #4611 

Implements a loading screen to be shown while the app is refreshing the onboarding state. Because the state is unlikely to change after a successful setup, I decided to only show a loading screen if the initial state was not `.completed`.

## To test

1. Go to Settings > In-Person Payments
2. If the store is correctly onboarded into payments, the loading screen shouldn't be visible, otherwise there should be a loading screen while the app refreshes the data.

Another good way to test this is to try it in a store that doesn't have WCPay installed, as that screen will have a "Refresh" button, which should show the loading screen again.

Light mode | Dark mode
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-08-10 at 10 56 28|Simulator Screen Shot - iPhone 12 Pro - 2021-08-10 at 11 12 17

Update release notes:


ctarda:
@koke  I am testing this with a store that does not have anything related to payments installed and the loading state is visible, but the app is crashing on me. The crash is on `#97	0x0000000106a93b44 in String.htmlToAttributedString.getter at /Users/ctarda/Development/woocommerce-ios/WooCommerce/Classes/Extensions/String+HTML.swift:21`

Screen Shot 2021-08-10 at 05 29 28

ctarda:
Wait, on retry it does not crash. Let me check again

ctarda:
@koke It does not happen all the time, but the crash can be reproduced by navigating in and out of In-Person Payments in a store that does not have WCPay installed repeatedly. Eventually, it will crash in `htmlToAttributedString`

koke:
😬 we had some issues like that on #4527 and so far I thought it was enough to ensure the attributed string was created on the main thread, but maybe not? There's not an easy fix for this, unless we can find a different way to easily parse HTML 😞

ctarda:
I see. Then let's get it in for now, we should probably start looking for alternative solutions because the crash is quite frequent.

koke:
> @koke It does not happen all the time, but the crash can be reproduced by navigating in and out of In-Person Payments in a store that does not have WCPay installed repeatedly. Eventually, it will crash in `htmlToAttributedString`

I've tried that though, and can't seem to be able to reproduce. I'd like to understand better exactly what is crashing to be sure if it's the same issue

ctarda (approved):
The layout looks good and it shows where it should. There are concerns about a possible crash

:shipit:

koke:
I applied the same in 00e466974 fix as I did in #4527. This might still crash the first time the link is parsed, but would be cached afterwards and not crash. I'll open up a discussion later to figure out a better long term solution

PR #4761, opened 2021-08-10
[CPP Onboarding] Implement screen for Payments installed but not set up

koke:
Part of #4611 

This adds a specific error when WooCommerce Payments is installed, but the merchant hasn't set up an account yet.

I choose to use an in-app Safari view controller instead of taking the user outside the app, but I'm not sure if we want the opposite. I figured that using this method, it's easier to see when the web view was dismissed, which can trigger a refresh of the data.

## To test

Testing this gets trickier because right now we are checking `isCardPresentEligible` first and showing a generic error if the account is not eligible. To test this, I modified `WCPayAccount.noAccount` to ensure the fake account object has `isCardPresentEligible = true`. The alternative would be to test this in a store that somehow had been enrolled in the program but not set up WCPay yet.

1. Use a store that's not set up for payments, then install the WooCommerce Payments but don't set it up yet
2. In the app, go to Settings > In-Person payments
3. The specific screen should show
4. Tapping "Finish Setup in Store Admin" should open an in-app browser that takes you to the right wp-admin page (you might need to log in first)
5. Closing the web browser, triggers a refresh of the state. This should briefly show a loading screen.

Error screen | Web browser
-|-
Screen Shot 2021-08-10 at 16 50 39|Screen Shot 2021-08-10 at 16 52 14

Update release notes:


ctarda (approved):
I took the liberty of merging develop into this branch and pushing the merge. 

The PR tests well:


:shipit:

PR #4763, opened 2021-08-10
[CPP Onboarding] Implement Stripe account restricted due to overdue requirements

ctarda:
Part of #4611 

This adds an specific error when the Stripe account is restricted due to overdue requirements.

## How to test
In order to test this in real conditions, it is necessary to have a store associated to an Stripe account that is has overdue requirements, or...
make `CardPresentPaymentsOnboardingUseCase.checkOnboardingState` return `stripeAccountOverdueRequirement`

1. Navigate to Settings > In-Person Payments
2. Notice the error screen.

Update release notes:


koke (approved):
Tested by hardcoding state in checkOnboardingState and works as expected.

I didn't realize we weren't implementing the CTAs on this phase so I now wonder if I should remove the one in the WCPay not set up screen 🤔

ctarda:
Thanks for the review! 

>  I now wonder if I should remove the one in the WCPay not set up screen 🤔

We can do that in one big sweep at the end if needed, right?

PR #4769, opened 2021-08-11
Link XCUITestHelpers Swift package to UI and screenshot tests targets

mokagio:
To verify it's linked properly, I removed the custom `pasteText(_:)` method in favor of the one offered by the package.

I run the UI tests in CI and they passed 🎉 

### DIFF
--- a/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
+++ b/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
@@ -494,6 +494,8 @@
 		31F92DE125E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31F92DE025E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift */; };
 		31FE28C225E6D338003519F2 /* LearnMoreTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31FE28C125E6D338003519F2 /* LearnMoreTableViewCell.swift */; };
 		31FE28C825E6D384003519F2 /* LearnMoreTableViewCell.xib in Resources */ = {isa = PBXBuildFile; fileRef = 31FE28C725E6D384003519F2 /* LearnMoreTableViewCell.xib */; };
+		3F1CA81D26C3542600228BF2 /* XCUITestHelpers in Frameworks */ = {isa = PBXBuildFile; productRef = 3F1CA81C26C3542600228BF2 /* XCUITestHelpers */; };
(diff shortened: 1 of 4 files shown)


WooCommerce/WooCommerceUITests/Screens/Login/LoginPasswordScreen.swift
```diff
@@ -28,7 +28,7 @@ final class LoginPasswordScreen: BaseScreen {
     func tryProceed(password: String) -> LoginPasswordScreen {
         XCTAssert(passwordTextField.waitForHittability(timeout: 3))
 
-        passwordTextField.pasteText(text: password)
+        passwordTextField.paste(text: password)
```
mokagio, line 31: This logic was copy pasted across Simplenote and Woo. I moved it from Simplenote to XCUITestHelpers initially, and now we can remove the custom implementation from Woo, too. 

🧹 💨 💨 💨

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
```diff
@@ -8453,6 +8468,16 @@
...
+		3F1CA81C26C3542600228BF2 /* XCUITestHelpers */ = {
+			isa = XCSwiftPackageProductDependency;
+			package = 3F1CA81B26C3542600228BF2 /* XCRemoteSwiftPackageReference "XCUITestHelpers" */;
+			productName = XCUITestHelpers;
+		};
+		3F1CA81E26C3543C00228BF2 /* XCUITestHelpers */ = {
+			isa = XCSwiftPackageProductDependency;
+			package = 3F1CA81B26C3542600228BF2 /* XCRemoteSwiftPackageReference "XCUITestHelpers" */;
+			productName = XCUITestHelpers;
+		};
```
mokagio, line 8480: There's two of these, like in lines 497 and 498 above, because both the UI and screenshot tests link to the package.

pachlava (approved):
It looks like the method with adjusted `paste` call is not used in the tests (the tests are using `tryProceed(password: String)` from `PasswordScreen`, not from `LoginPasswordScreen`), I locally edited the method from `PasswordScreen` to use pasting instead of text input, and it worked per se. 

However, there was some flakiness with pasting and I had troubles to have all tests pass in one run.

The flakiness is not related to the PR, since the main goal of the PR is linking the helper package and proving that helper methods are available to the code - which is definitely so 👍

mokagio:
> It looks like the method with adjusted paste call is not used in the tests

Good catch! I'll look into it and see whether it's appropriate to open a followup PR to remove that unused code, or if it's okay to keep that code there for future use. (As I write that, a voice in my head screams YAGNI! but I do wonder how much that applies in the context of screen object that are meant to make it easy for folks to write tests 🤔 )

> However, there was some flakiness with pasting and I had troubles to have all tests pass in one run.

Maybe that's something we can address with this new infrastructure 🤞

PR #4770, opened 2021-08-11
Remove unused and unnecessary code from `PrologueScreen`

mokagio:
Hat tip @jkmassel, see [here](https://github.com/woocommerce/woocommerce-ios/pull/4392#discussion_r686201816).

It's been a while since I wrote that code... My guess was that past me was experimenting with closures and runtime evaluation of the objects and didn't cleanup properly.

I think there might be a use case for defining getter closures within a `ScreenObject` subclass to expose them statically then reuse them in instance properties and method, but that's not the case in this particular case.


jkmassel (approved):
Nice!! 🙂

Issue #4743, opened 2021-08-06
Shipping Labels: Custom package creation screens


rachelmcr:
This is a subtask of #3909.

- [x] Create a new screen (preferably in SwiftUI, since this flow is entirely in SwiftUI), which contains the two tab selectors "Custom Package" and "Service Packages". Ideally, they are two separates SwiftUI subviews.
- [x] Create the screen "Custom Package", and check that all the fields in the Custom Package form are mandatory.
- [x] Create the Package Type screen with the list selector containing just the two options Box and Envelope.

Things to be aware of:

- [x] Core allows empty package weight to be zero, but length/width/height can't be zero.
- [x] Once a custom package is created, make sure it is then listed in the package selector screen (this will be the screen before the package creation screen if there's already existing packages). -> Part of #4745
- [x] If the store has no saved packages, selecting Package Details > Package Selected opens the "Custom Package" screen.

<img src="https://user-images.githubusercontent.com/495617/113698539-dd99a800-96d4-11eb-9018-d28ec451e0ff.png">

PR #4781, opened 2021-08-11
Shipping Labels: Implement "Add New Package" screen

rachelmcr:
Part of: #4743

## Description

This introduces the initial UI for adding a new package when creating a shipping label. It has the following behavior in the Shipping Label creation flow:

* If your store doesn't have any packages set up for shipping labels, when you get to the Package Details screen and tap on Package Selected it opens the "Add New Package" screen with the "Custom Package" tab selected.
* If your store does have packages set up already, tapping Package Selected opens the package list with a "Create new package" button at the bottom of the screen. Tapping that button opens the "Add New Package" screen.

## Changes

* Implements an "Add New Package" view in `ShippingLabelAddNewPackage`. For now, this view only contains "Custom Package" and "Service Package" tabs. (The tab subviews will be added in a later PR.)
* Implements a `ShippingLabelPackageSelection` view with a `NavigationView` that conditionally shows either the package list or add new package screen. (This ensures the navigation hierarchy is set up correctly regardless of which view is shown first.)
* Removes the `NavigationView` from `ShippingLabelPackageList` (since that's now handled in `ShippingLabelPackageSelection`) and implements the navigation link from the "Create new package" button to the "add new package" view.
* Adds a "Select a package" placeholder to the Package Details > Package Selected row if the store doesn't have any packages set up yet.

### Store with packages already set up

Package Details | Package List | Add New Package
-|-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-08-11 at 16 00 16|Simulator Screen Shot - iPhone 11 Pro - 2021-08-11 at 16 00 19|Simulator Screen Shot - iPhone 11 Pro - 2021-08-11 at 16 00 23

### Store without any packages

Package Details | Add New Package
-|-
Simulator Screen Shot - iPhone 11 Pro - 2021-08-11 at 16 01 08|Simulator Screen Shot - iPhone 11 Pro - 2021-08-11 at 16 01 11

### Store with packages

8. Confirm you are taken to the "Add New Package" screen and with selectable tabs for "Custom Package" and "Service Package." 

### Store without any packages

1. Make sure you have configured the WooCommerce WCShip plugin, and **no** packages set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
6. In the package details section, confirm that "Package Selected" shows the placeholder "Select a package." Tap on the "Package Selected" row.
7. Confirm you are taken to the "Add New Package" screen and with selectable tabs for "Custom Package" and "Service Package."

## Submitter Checklist

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
```diff
@@ -0,0 +1,46 @@
...
+        .toolbar {
+            ToolbarItem(placement: .navigationBarLeading) {
+                Button {
+                    presentation.wrappedValue.dismiss()
+                } label: {
+                    Image(uiImage: .chevronImage.withHorizontallyFlippedOrientation().imageFlippedForRightToLeftLayoutDirection())
+                }
+
+            }
+        }
```
rachelmcr, line 29: SwiftUI doesn't seem to have an equivalent to UIKit's `navigationItem.backButtonDisplayMode = .minimal` for removing text from the navigation bar back button. Instead, we have to remove the default back button and then display our own custom back button.

We should make this reusable for any SwiftUI views that need it. I decided to wait on that for now, since I'm not sure if it would be more useful to have a reusable `ToolbarItem` for the minimal back button or a custom modifier that does all the toolbar setup for this.

itsmeichigo: I'd say that a custom modifier would be great for reusability here! But I think we can use `.navigationBarItems(leading:)` here instead of `toolbar` as it's more straightforward that we're dealing with navigation bar here - plus it's shorter! What do you think?

Also - thank you for adding `imageFlippedForRightToLeftLayoutDirection()` for RTL languages here 🎉. Do you think it's necessary to add `.chevronImage.withHorizontallyFlippedOrientation()` to the `UIImage+Woo` extension as something like `chevronLeftImage`? Just in case it may be necessary in other places as well.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
```diff
@@ -0,0 +1,17 @@
...
+final class ShippingLabelAddNewPackageViewModel: ObservableObject {
+    @Published var selectedIndex: Int
+
+    init(_ selectedIndex: Int = SelectedIndex.customPackage.rawValue) {
+        self.selectedIndex = selectedIndex
+    }
+
+    enum SelectedIndex: Int {
+        case customPackage = 0
+        case servicePackage = 1
```
itsmeichigo, line 15: Quick tip: For enums with Int as raw value, the values starts from 0 by default and increase 1 by 1 for each of the next cases. But I appreciate the clarity here, so I think there's no need for removing the explicit values 👍

itsmeichigo (approved):
Code looks good and works as expected 👍 :shipit:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
rachelmcr: > I think we can use .navigationBarItems(leading:) here

I used that originally, but then I [looked at the docs](https://developer.apple.com/documentation/swiftui/view/navigationbaritems(leading:trailing:)) and realized it's deprecated. 😬 The recommended approach is `toolbar(content:)` now, which is why I used it here.

> Do you think it's necessary to add .chevronImage.withHorizontallyFlippedOrientation() to the UIImage+Woo extension as something like chevronLeftImage?

Good point, I'll go ahead and add that! I just realized there's a Gridicon for the left chevron so we don't have to flip the regular chevron for that, either.

rachelmcr: I'm going to go ahead and merge this PR as-is, and I'll make both changes (the custom modifier for reusability, and the left-facing chevron image) in a followup PR.

PR #4782, opened 2021-08-11
[CPP Onboarding] Ordering card readers

koke:
Part of #4611 

This adds a new menu when WCPay is set up correctly to show the option to either order or manage a card reader, instead of showing the card reader settings directly.

## Why UIKit?

There is an unresolved SwiftUI issue (see #3995) that will show an extra tab bar item when `NavigationLink` is used within a `UITabBarController`. Since we don't have a solution yet, and the card reader settings are still implemented in UIKit, one solution was to implement this new menu as a UITableView.

This is fine, and we use this approach for many other views in the app, but it's still painful that, without that bug, the implementation would have been so much simpler 😭 

```swift
struct InPersonPaymentsMenu: View {
    var body: some View {
        List {
            NavigationLink("Order card reader", destination: OrderCardReader())
            NavigationLink("Manage card reader", destination: CardReaderSettings())
        }
    }
}
```

## To Test

Use a store that was WCPay completely set up and is eligible for In-person payments

1. Go to Settings > In-Person Payments
2. See the new menu
3. Tapping on Order card reader should open a web browser. This is yet to be implemented in the backend, so the URL returns a 404 for now
4. Tapping on Manage card reader should push the card reader settings

In-Person Payments | Order card reader | Manage card reader
-|-|-
Screen Shot 2021-08-11 at 16 54 23|Screen Shot 2021-08-11 at 16 54 30|Screen Shot 2021-08-11 at 16 55 08

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsMenuViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsMenuViewController.swift
@@ -83,7 +83,7 @@ private extension InPersonPaymentsMenuViewController {
 //
 extension InPersonPaymentsMenuViewController {
     func orderCardReaderWasPressed() {
-        // TODO
+        WebviewHelper.launch(Constants.woocommercePurchaseCardReaderURL, with: self)
     }
 
     func manageCardReaderWasPressed() {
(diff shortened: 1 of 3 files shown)


ctarda (approved):
Switching to UIKit must have been painful... yet pragmatic 😊 

It works perfect:

https://user-images.githubusercontent.com/2722505/129071830-fcc01717-6eee-46d0-a056-e454a251b8df.mov

:shipit:

PR #4765, opened 2021-08-10
[Mobile Payments] [Customers] Add a remote to create a wcpay customer on an order

allendav:
Partially addresses #4298 

This PR adds customer creation to the WCPay Remote, taking advantage of the new create_customer endpoint added in plugin verison 2.8.3. A subsequent PR will integrate this into CardPresentPaymentStore to allow us to pass the customer ID into payment intent creation

To test:
- Run the new unit tests `test_fetchOrderCustomer_properly_returns_customer` and `test_fetchOrderCustomer_properly_handles_error_response` and ensure they pass

See also:
- https://github.com/Automattic/woocommerce-payments/pull/2501

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4788, opened 2021-08-11
[Mobile Payments] [Customers] Fetch and pass customer when creating payment intent

### LINKED ISSUE #4298: [Mobile Payments] Pass registered user customer ID when creating a payment intent
Update: The plugin will manage the customer IDs for us through a [new endpoint](https://href.li/?https://github.com/Automattic/woocommerce-payments/pull/2501). We no longer need to concern ourselves with registered or not-registered users. We just need to:

- [ ] Call the new endpoint passing the order ID for which we are about to create the payment intent for
- [ ] Accept the customer ID returned by it and include it in the payment intent

To wit, I am closing #4299 and #4300 as this is all now much simpler and can be worked under this issue alone.

<del>For an order created by a registered user, we need to query the user meta data for a WCPay (Stripe) Customer ID and include the ID in the payment intent parameters during intent creation. </del>

<del>If we don't have the registered buyer WordPress user ID in the order already, we'll need a ticket to add it to the order remote and mapper.</del>

<del>If we don't have a means of fetching meta data for registered buyers, we'll need to add a ticket for that as well.</del>

Note: Customers created when WCPay is in test mode may have a different customer ID than customers created when WCPay is in live mode (see below)

Related issues: #3863 #4299 #4300 https://github.com/Automattic/woocommerce-payments/issues/1918

cc @ctarda @bmccotter

allendav:
Closes #4298 

Summary of changes:
- Builds on the remote added in #4765
- Adds an action to CardPresentPaymentAction, CardPresentPaymentStore for fetching a customer through that remote
- Adds customer ID to the payment intent parameters
- Updates `PaymentCaptureOrchestrator` to request the customer from the site for the order - also a bit of whitespace reformatting

To test:
- Create at least two COD orders as a registered user and one as a guest user
- Set a breakpoint in PaymentCaptureOrchestrator collectPayment after the completion switch for fetchOrderCustomer
- Collect payment for all three orders
- When the breakpoint hits, po parameters.customerID and then resume
- Ensure the two registered user orders are assigned the same Stripe customer ID. Ensure the guest order gets a different Stripe customer ID.
- Or, if you prefer, instead of breakpointing, you can view the payment intent details in the private dashboard, e.g.

pic

- Backwards compatibility is also important, since this customer-returning endpoint was only added to recently (plugin version 2.8.3) - repeat the test with a 2.7.x release, or edit the plugin to remove the endpoint, i.e.:

Screen Shot 2021-08-11 at 2 37 13 PM

- Make sure you can still collect payment. When you look at the transaction in the dashboard, you should see "null" for the customer on the payment intent.

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Actions/CardPresentPaymentAction.swift
+++ b/Yosemite/Yosemite/Actions/CardPresentPaymentAction.swift
@@ -26,6 +26,10 @@ public enum CardPresentPaymentAction: Action {
     ///
     case observeConnectedReaders(onCompletion: ([CardReader]) -> Void)
 
+    /// Get a Stripe Customer for an order.
+    ///
+    case fetchOrderCustomer(siteID: Int64, orderID: Int64, onCompletion: (Result<WCPayCustomer, Error>) -> Void)
+
     /// Collected payment for an order.
     ///
     case collectPayment(siteID: Int64,
--- a/Hardware/Hardware/CardReader/PaymentIntentParameters.swift
+++ b/Hardware/Hardware/CardReader/PaymentIntentParameters.swift
@@ -32,17 +32,24 @@ public struct PaymentIntentParameters {
     /// This can be useful for storing additional information about the object in a structured format.
     public let metadata: [AnyHashable: Any]?
 
+    /// A Stripe issued customer ID
+    /// See https://stripe.com/docs/api/customers
+    ///
+    public let customerID: String?
+
     public init(amount: Decimal,
                 currency: String,
                 receiptDescription: String? = nil,
                 statementDescription: String? = nil,
                 receiptEmail: String? = nil,
-                metadata: [AnyHashable: Any]? = nil) {
+                metadata: [AnyHashable: Any]? = nil,
+                customerID: String? = nil) {
         self.amount = amount
         self.currency = currency
         self.receiptDescription = receiptDescription
         self.statementDescription = statementDescription
         self.receiptEmail = receiptEmail
         self.metadata = metadata
+        self.customerID = customerID
     }
 }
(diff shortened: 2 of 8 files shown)


Yosemite/Yosemite/Actions/CardPresentPaymentAction.swift
```diff
@@ -26,6 +26,10 @@ public enum CardPresentPaymentAction: Action {
     ///
     case observeConnectedReaders(onCompletion: ([CardReader]) -> Void)
 
+    /// Get a Stripe Customer for an order.
+    ///
+    case fetchOrderCustomer(siteID: Int64, orderID: Int64, onCompletion: (Result<WCPayCustomer, Error>) -> Void)
```
ctarda, line 31: I am not sure about the placement of this method in this store / action. So far, the concern of the store has only been communicating with readers. Creating a customer seems like a different concern. I think we should reconsider this, but for now, let's move forward.

ctarda (approved):
It works well. I think the action could be extracted to a different store, but I am not sure yet where exactly. Anyway, that's something we can discuss at a later stage. 

@allendav is afk, so I took the liberty of pushing a couple of tests.

:shipit: 

I will merge on behalf of the author when CI is ✅

Yosemite/Yosemite/Actions/CardPresentPaymentAction.swift
allendav: > I am not sure about the placement of this method in this store / action. So far, the concern of the store has only been communicating with readers.

Your point is well taken. The thought crossed my mind as well, and I agree that the store has been, until this, reader I/O focused in its actions.

allendav:
@ctarda wrote:

> so I took the liberty of pushing a couple of tests

Thank you :)

PR #4789, opened 2021-08-12
Hide refund button if order is not eligible for refund

### LINKED ISSUE #3845: Refunds: consider hiding the option to refund if there is nothing to refund
Steps to reproduce:

1. Create an order where the price of the product(s) is equal to zero.<br><sup>Note: one way to do this would be to add a sale price of zero for a product.</sup>
1. Open the order in the app.
1. Attempt to issue a refund for the order. 

**Result:** it's possible to submit a refund for a $0 order. 

Issue Refund button|Refund selection|Refund details|Refund confirm|Refund successful
---|---|---|---|---
IMG_4075|IMG_4077|IMG_4078|IMG_4080|IMG_4082

<sup>Tested with WCiOS 6.2.0.0 TestFlight beta on iPhone 6S iOS 13.7.</sup>

In comparison, the refund option is not available in the Android app for the same order where there is nothing to refund:

Screenshot_20210317-171502
<sup>Tested with WCAndroid 6.2-rc-1 on Pixel 3 Android 11.</sup>

<sup>(internal reference: p5T066-21o-p2#comment-7522 /hat tip @pmusolino)</sup>

iamgabrielma:
Fixes #3845 

## Description

This PR removes the `Issue Refund` button from the Order Details screen if there's actually no refund to be done. This also fits with how the Android version works.

## Changes

I have added an additional `isEligibleForRefund` check to the Order Details data source. This checks if the current `order.total` is positive. If the `order.total` is positive then the check will pass as there's something to be refunded, and we can load the cell for the refund button on this view.

Why use `order.total` for this check? This contains what a customer has effectively paid and sets the maximum that can be refunded, with coupons, shipping, fees, and taxes taken into consideration.

| Hide button: Order details when refund is `0` |  Show button: Order details when refund is `>0` | Hide button: Order details when refund is `0` because negative prices | Show button: Order details when partial refund | 
|---------------------|---|-----------------|----------------|
| refund is 0, no button | refund is positive, button | refund is negative, no button| partial refund, shows button |

**Edge Case: Refunding payment gateway fees:**

At the moment I left the option to see the Refund Button if there's fees in the order, so when we have the case `order.total = 0 && order.fees > 0` it will show the refund button, this is in consonance with how works in core: The button appears but is only possible to refund these fees directly through the payment gateway:

refund only fees

I believe we have 2 options here:

- Option 1) Show the refund button when there's remaining fees, once the user taps on it we can edit the view to show "You need to manually issue the refund directly through your payment gateway" message.
- Option 2) Hide the button, I'm not personally on board with this option as there's actually something that could still be refunded, just not through the app.

## Work in progress
- [x] Is `order.total` updated after a partial refund?
- [x] Payment processor fees are added after the total. Should these be refundable as well? How does it work in WooCommerce core?

## Testing steps
1 - Create an order where the price of the product(s) is equal to zero. One way to do this would be to add a sale price of zero for a product.
2 - Open the order in the app.
3 - Attempt to issue a refund for the order. The button is not there.
4 - Now try the same with an order where the price of the product is positive. 
5 - Attempt to issue a refund for the order. The button will appear.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -33,9 +33,6 @@ final class OrderDetailsDataSource: NSObject {
     private var isRefundedStatus: Bool {
         return order.status == OrderStatusEnum.refunded
     }
-    /// Is this order eligible for a refund?
-    ///
-    var isEligibleForRefund: Bool = true
 
     /// Is the shipment tracking plugin available?
     ///
@@ -975,12 +972,8 @@ extension OrderDetailsDataSource {
             if shouldShowReceipts {
                 rows.append(.seeReceipt)
             }
-
-            if let orderTotal = Double(self.order.total) {
-                if orderTotal <= 0.0 {
-                    isEligibleForRefund = false
-                }
-            }
+            let orderTotal = Double(self.order.total) ?? 0
+            let isEligibleForRefund = orderTotal > 0
             if !isRefundedStatus && !isEligibleForCardPresentPayment && isEligibleForRefund {
                 rows.append(.issueRefundButton)
             }
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -35,7 +35,7 @@ final class OrderDetailsDataSource: NSObject {
     }
     /// Is this order eligible for a refund?
     ///
-    var isEligibleForRefund : Bool = true
+    var isEligibleForRefund: Bool = true
 
     /// Is the shipment tracking plugin available?
     ///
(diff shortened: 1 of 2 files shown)


itsmeichigo (approved):
Thank you for fixing this - I really appreciate your informative description of the PR and all the cases taken into consideration for the fix. This works as expected ✅ - would you mind updating the `RELEASE-NOTES.txt` file with your fix? Thanks!

I have a non-blocking comment though - just a little nit-picking really 😅

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -972,8 +972,9 @@ extension OrderDetailsDataSource {
             if shouldShowReceipts {
                 rows.append(.seeReceipt)
             }
-
-            if !isRefundedStatus && !isEligibleForCardPresentPayment {
+            let orderTotal = Double(self.order.total) ?? 0
```
itsmeichigo: Great work making this a separate check - this clarifies your intention nicely 👍 
However I think `self` isn't really necessary here, and our overall coding convention (I can't really find the reference anywhere though 😓) that we omit `self` whenever possible.

iamgabrielma: Gotcha, fixed!

PR #4794, opened 2021-08-12
[CPP Onboarding] No connection error screen

koke:
Part of #4611 

This implements a specific error screen when the connection failed and the onboarding requirements couldn't be checked.

## Decisions to be questioned

1. **What kind of errors should we include here?** I did the easy thing and triggered this as long as any of the requests returns an error with `NSURLErrorDomain`. There are some possible errors where this screen is a clearer match (e.g. `NSURLErrorNotConnectedToInternet`, `NSURLErrorNetworkConnectionLost`,...), but some others might indicate a server issue, which doesn't fit the "No connection" title.
2. **Should we auto-retry when the connection is back?** I implemented this on 1eca1f0, so we can revert it if we think it's overcomplicating things, but since we have Reachability, I thought it'd be nicer to auto-retry when we detect the connection is back. Given the previous point though, I'm not sure we want to trigger this behaviour unless we're certain that the error is caused by the network being unreachable, or we might start an endless retry loop 🤔. I'm kind of answering myself, and I think I'll revert the commit until it's more tested, then we can decide if we want to implement it separately.
3. **What should be the UI feedback on retry?**. When I initially implemented this and tested on Airplane Mode, tapping on Retry would set the `loading` state, then immediately fail again, which would flash the loading screen for a moment. I avoided this by adding a 100ms debounce, but then the Retry button might seem like it's not working. Should I replace the debounce with a delay so that the loading screen always shows for at least 500ms? Note that I can still see the loading screen briefly when I navigate to In-Person Payments initially, should we get rid of that as well? cc @woocommerce/mobile-designers 
4. **Should we show this error when we have previously seen a `completed` status?** I already hide the loading screen in this case, as the state is unlikely to change after the initial onboarding. It seems unlikely that this would be a problem, since most of the features gated by this require a connection. The only potential annoyance is a user wanting to disconnect a reader while offline, but there's an easy way out of that by turning off the reader.

## To test

1. Put your phone in airplane mode
2. Go to Settings > In-Person Payments
3. See the new error screen
4. Tap Retry, the loading screen doesn't show
5. Disable airplane mode
6. Tap Retry, the loading screen shows, and the correct state shows after that (it might take a few attempts until the device restores the connection)

Screen Shot 2021-08-12 at 10 35 06

Update release notes:


ctarda (approved):
> I'm kind of answering myself, and I think I'll revert the commit until it's more tested, then we can decide if we want to implement it separately.

😄 

I think it's better to avoid retrying. As you said, it's a bit of a can of worms, and as far as I can tell, it's not something we do anywhere in the app.

Anyway, it works great. Tested on simulator and on device. Here is a recording of the simulator, switching the laptop's wifi on and off:


https://user-images.githubusercontent.com/2722505/129188223-ce86e0bc-b073-4171-b03e-7094ccd3f859.mov

:shipit:

PR #4797, opened 2021-08-12
SwiftUI: Add custom view modifier for adding a minimal back button to the navigation bar

rachelmcr:
## Description

In https://github.com/woocommerce/woocommerce-ios/pull/4781 I introduced a minimal back button in a SwiftUI view. As discussed in https://github.com/woocommerce/woocommerce-ios/pull/4781#pullrequestreview-727681638, SwiftUI doesn't seem to have an equivalent to UIKit's `navigationItem.backButtonDisplayMode = .minimal` for removing text from the navigation bar back button. Instead, we have to remove the default back button and then display our own custom back button.

This PR moves that implementation into a custom view modifier so we can reuse it across SwiftUI views.

## Changes

* Adds the `minimalNavigationBarBackButton()` view modifier, which removes the default back button and adds a minimal (left chevron) button that navigates back.
* Introduces a left chevron icon and unit test.

View using minimal back button:

## Testing

8. On the "Add New Package" screen, confirm a minimal back button (chevron) appears and you can tap the back button to go back.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIImage+Woo.swift
+++ b/WooCommerce/Classes/Extensions/UIImage+Woo.swift
@@ -103,6 +103,12 @@ extension UIImage {
         return UIImage.gridicon(.chevronRight)
     }
 
+    /// Chevron Pointing Left
+    ///
+    static var chevronLeftImage: UIImage {
+        return UIImage.gridicon(.chevronLeft)
+    }
+
     /// Chevron Pointing Down
     ///
     static var chevronDownImage: UIImage {
--- a/WooCommerce/Classes/View Modifiers/View+Toolbars.swift
+++ b/WooCommerce/Classes/View Modifiers/View+Toolbars.swift
@@ -1,0 +1,27 @@
+import SwiftUI
+
+/// Custom view modifier for displaying a minimal back button in the navigation bar
+struct MinimalNavigationBarBackButton: ViewModifier {
+    @Environment(\.presentationMode) var presentation
+
+    func body(content: Content) -> some View {
+        content
+            .navigationBarBackButtonHidden(true)
+            .toolbar {
+                ToolbarItem(placement: .navigationBarLeading) {
+                    Button {
+                        presentation.wrappedValue.dismiss()
+                    } label: {
+                        Image(uiImage: .chevronLeftImage.imageFlippedForRightToLeftLayoutDirection())
+                    }
+                }
+            }
+    }
+}
+
+extension View {
+    /// Displays a minimal back button in the navigation bar
+    func minimalNavigationBarBackButton() -> some View {
+        self.modifier(MinimalNavigationBarBackButton())
+    }
+}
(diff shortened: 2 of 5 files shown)


itsmeichigo (approved):
Thank you for adding this modifier! The back button works as expected and looks good in both LTR and RTL languages.

However - I realize just now that this fix disables the native interactive pop gesture on the navigation controller, i.e we cannot use swipe gesture to get back to the previous screen. I'm thinking whether we should apply a [global solution](https://stackoverflow.com/a/67746167) instead, but it may be frowned upon. 

So I'm approving this PR and we can leave the issue as a tech debt and wait until `backButtonDisplayMode` is supported on SwiftUI.

koke:
> I'm thinking whether we should apply a [global solution](https://stackoverflow.com/a/67746167) instead

That global solution got me inspired and I wanted to try a different approach. I haven't been able to make it work and I want to finish some other things today, but in theory this should work, although it requires using a custom UIHostingController subclass.

We can use SwiftUI preferences to implement our custom `navigationBackButtonDisplayMode`, then read that value from the root view and pass it to the View Controller.

I tried this, but for some reason, the callback in `onPreferenceChange` is not being called when the mode is changed. I'll leave the code here in case someone else wants to investigate further.

```swift
import SwiftUI
import UIKit

final class WooHostingController<Content: View>: UIHostingController<WooHostingControllerRootView<Content>> {
    init(rootView: Content) {
        weak var futureSelf: WooHostingController<Content>? = nil
        let navigationBackButtonDisplayModeChanged = { backButtonDisplayMode -> Void in
            futureSelf?.navigationItem.backButtonDisplayMode = backButtonDisplayMode
        }
        super.init(rootView: WooHostingControllerRootView(navigationBackButtonDisplayModeChanged: navigationBackButtonDisplayModeChanged) {
            rootView
        })
        futureSelf = self
    }

    @objc required dynamic init?(coder aDecoder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }
}

struct WooHostingControllerRootView<Content: View>: View {
    let navigationBackButtonDisplayModeChanged: (UINavigationItem.BackButtonDisplayMode) -> Void
    let content: () -> Content

    var body: some View {
        content()
            .onPreferenceChange(NavigationBackButtonDisplayModePreferenceKey.self) { backButtonDisplayMode in
                navigationBackButtonDisplayModeChanged(backButtonDisplayMode)
            }
    }
}

struct NavigationBackButtonDisplayModePreferenceKey: PreferenceKey {
    static var defaultValue: UINavigationItem.BackButtonDisplayMode = .default

    static func reduce(value: inout UINavigationItem.BackButtonDisplayMode, nextValue: () -> UINavigationItem.BackButtonDisplayMode) {
        value = nextValue()
    }
}

extension View {
    func navigationBackButtonDisplayMode(_ mode: UINavigationItem.BackButtonDisplayMode) -> some View {
        self.preference(key: NavigationBackButtonDisplayModePreferenceKey.self, value: mode)
    }
}
```

rachelmcr:
Thank you both! I missed that this solution disables the swipe gesture, which is unfortunate. I really like the idea of a more global solution, and I tinkered a bit with @koke's suggestion but couldn't figure out why it isn't working.

I went ahead and opened #4803 to keep track of this, and I'll bring it up on a p2 for discussion. For now I'll go ahead and merge this PR as a compromise/interim solution.

PR #4806, opened 2021-08-13
Shipping Labels: Update default item details for customs form

itsmeichigo:
Part of #4687

# Description
This PR update default item details form custom form based on fetched products and variations. Item details should now display correct data.

# Changes
- Updated ShippingLabelsCustomsForm quantity property type to Decimal to match with order item's for safety.
- Customs form list view model: 
   - Fetch currency from order details, then send the currency all the way down to item detail form view models to display on the UI.
   - Reuse `ShippingLabelPackageDetailsResultsControllers` to fetch products and variations.
   - Observe changes of products and variations to update default item details. 
- Updated item details view model to fetch weight unit from `shippingSettingsService` to display on the UI.

# Demo
item-details

# Testing steps
1. Make sure that your test store has WCShip plugin installed and package details set up.
2. Navigate to Orders tab and select an order with international shipping (origin country different from destination).
4. Skip through Ship To, Ship From, Package Details to configure Customs.
5. Notice that in Package Details section, item details are displayed correctly: description, weight, value, weight unit, currency. Origin country is set the same as your site address country by default.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormListViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormListViewModel.swift
@@ -116,7 +116,6 @@ private extension ShippingLabelCustomsFormListViewModel {
                     } else if let product = products.first(where: { $0.productID == orderItem.productID }) {
                         return Double(product.weight ?? "0") ?? 0
                     }
-
                     return 0
                 }()
 
@@ -125,7 +124,7 @@ private extension ShippingLabelCustomsFormListViewModel {
                              value: orderItem.price.doubleValue,
                              weight: weight,
                              hsTariffNumber: "",
-                             originCountry: "",
+                             originCountry: SiteAddress().countryCode, // Default value
                              productID: item.productID)
             }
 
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormListViewModel.swift
```diff
@@ -49,6 +61,101 @@ final class ShippingLabelCustomsFormListViewModel: ObservableObject {
...
+            self?.productVariations = productVariations
+        })
+
+        products = resultsControllers?.products ?? []
+        productVariations = resultsControllers?.productVariations ?? []
+    }
+
+    /// Return copy of customs forms with updated item details based on fetched products and variations.
+    ///
+    func updateCustomsForms(products: [Product], productVariations: [ProductVariation]) -> [ShippingLabelCustomsForm] {
```
rachelmcr, line 113: At some point here we should be sure to filter out any order items that are virtual (virtual products or virtual product variations), since those aren't shipped. Right now, those are showing up in the Package Content list for the customs form even though they aren't shown elsewhere in the shipping label flow.

rachelmcr (commented):
This looks good and works great in most cases! The only issue I came across is with virtual products, which should be excluded (since they aren't shipped) but are showing up in the Package Content list on the customs form screen.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormListViewModel.swift
itsmeichigo: Thank you Rachel for pointing out this - I totally missed it earlier. I have fixed it in 6439bd453cf2249f46a72c78ca368cc5a675a656.

itsmeichigo:
Thank you @rachelmcr for pointing out the issue with virtual products - I have fixed it in 6439bd453cf2249f46a72c78ca368cc5a675a656 and added a unit test to check it. Please take another look and let me know what you think 🙇‍♀️

rachelmcr (approved):
Code & testing looks great, thank you! ✅

PR #4809, opened 2021-08-15
Edited NSLocalized strings

### LINKED ISSUE #4746: [Shipping Labels] "...from your phone" - should probably say "mobile device" since iPad is also an option, no?
In several spots we have localized strings that reference "phone" - maybe that should say "mobile device" instead since some of our merchants will prefer to manage their stores, print labels, etc from an iPad

https://github.com/woocommerce/woocommerce-ios/blob/9756551209f392afae5cb3b2e8092f78965aca0b/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsDataSource.swift#L1211

https://github.com/woocommerce/woocommerce-ios/blob/4fe952211777fad134084cbc19d4db229a37c8c9/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Shipping%20Labels/Create%20Shipping%20Label%20Info/ShippingLabelCreationInfoViewController.swift#L53

Possibly other places as well

iamgabrielma:
Fixes #4746 

# Description

This PR edits some instances of `NSLocalizedString` where we use "phone", but using "mobile device" instead would be more adequate.

# Changes 

I've changed several of the `NSLocalizedString` where it made sense to refer to a mobile device.

I skipped others where it makes more sense to leave it as `phone` instead, like:

```
NSLocalizedString("A phone number is required because this shipment requires a customs form")
NSLocalizedString("Custom forms require a 10-digit phone number")
NSLocalizedString("The phone number is not valid or you can't call the customer from this device")

```
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -1209,7 +1209,7 @@ extension OrderDetailsDataSource {
         static let payment = NSLocalizedString("Payment", comment: "Payment section title")
         static let notes = NSLocalizedString("Order Notes", comment: "Order notes section title")
         static let shippingLabelCreationInfoAction =
-            NSLocalizedString("Learn more about creating labels with your phone",
+            NSLocalizedString("Learn more about creating labels with your mobile device",
                               comment: "Title of button in order details > info link for creating a shipping label on the mobile device.")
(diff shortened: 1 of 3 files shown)


shiki (approved):
Thank you for this, @iamgabrielma. I confirmed and tested the affected screens and this LGTM! :shipit:

shiki:
Someone else who is not Gabriel: Please don't merge this yet since we're testing something. 🤪

PR #4810, opened 2021-08-16
Shipping Labels: Customs form validation for item details

itsmeichigo:
Part of #4687
This PR is based on #4806, so please make sure that PR is reviewed first.

# Description
This PR adds validation for item details in customs form of Shipping Labels flow.

# Changes
1. Updated reusable view / view modifier:
- New modifier `ErrorStyle` for text with `.body` font and red color for errors.
- New view `ValidationErrorRow` to display validation error.
2. Updated view model for item details to add validations for item description, value, weight, HS tariff number, origin country.
3. Updated item details input view to add validation error rows when validation fails.

# Demo
item-details-validation

# Testing steps
2. Open Orders tab and select an order with international shipping (origin country different from destination country).
4. Skip through Ship From, Ship To, Package Detail to configure Customs.
5. Notice that a validation error appears when either description / value / weight is empty.
6. Notice that a validation error appears when tariff number is entered but the length is less than 6. You can also try pasting some random text in this field and see that the error says the number is invalid.
7. Notice that when a validation fails, there's a red notice icon shown in the header of the item detail input form.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetails.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetails.swift
@@ -21,6 +21,10 @@ struct ShippingLabelCustomsFormItemDetails: View {
                 Image(uiImage: .inventoryImage)
                 Text(String(format: Localization.customLineTitle, itemNumber))
                     .bodyStyle()
+                Spacer()
+                Image(uiImage: .noticeImage)
+                    .foregroundColor(Color(.error))
+                    .renderedIf(viewModel.validatedItem == nil)
             }
         }, content: {
             // Item Description
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetails.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetails.swift
@@ -20,7 +20,7 @@ struct ShippingLabelCustomsFormItemDetails: View {
             HStack(spacing: Constants.horizontalSpacing) {
                 Image(uiImage: .inventoryImage)
                 Text(String(format: Localization.customLineTitle, itemNumber))
-                    .font(.body)
+                    .bodyStyle()
             }
         }, content: {
             VStack(spacing: 0) {
@@ -36,7 +36,8 @@ struct ShippingLabelCustomsFormItemDetails: View {
             VStack(spacing: 0) {
                 TitleAndTextFieldRow(title: Localization.hsTariffNumberTitle,
                                      placeholder: Localization.hsTariffNumberPlaceholder,
-                                     text: $viewModel.hsTariffNumber)
+                                     text: $viewModel.hsTariffNumber,
+                                     keyboardType: .numberPad)
                 Divider()
                     .padding(.leading, Constants.horizontalSpacing)
             }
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetailsViewModel.swift
```diff
@@ -54,3 +74,38 @@ final class ShippingLabelCustomsFormItemDetailsViewModel: ObservableObject {
...
+    }
+
+    var hasValidOriginCountry: Bool {
+        originCountry.code.isNotEmpty
+    }
+
+    var hasValidHSTariffNumber: Bool {
+        if hsTariffNumber.isNotEmpty,
+           (hsTariffNumber.count != 6 ||
+                hsTariffNumber.filter({ "0"..."9" ~= $0 }).count != 6) {
```
rachelmcr: I noticed that I can enter an HS tariff number over 6 digits long, and I get the validation error. I found this a little confusing because our docs say this (which made me think that at least 6 digits were needed but I wasn't sure why longer numbers would be invalid):

> Though some codes may be longer, you only need to include the first six digits.

On Android and web, we limit this field to 6 digits so it isn't possible to enter anything longer. What do you think of matching this behavior (limit the text field entry to 6 digits)?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetails.swift
```diff
@@ -20,81 +20,136 @@ struct ShippingLabelCustomsFormItemDetails: View {
...
-                    .padding(.leading, Constants.horizontalSpacing)
+                VStack(spacing: 0) {
+                    TitleAndTextFieldRow(title: Localization.hsTariffNumberTitle,
+                                         placeholder: Localization.hsTariffNumberPlaceholder,
+                                         text: $viewModel.hsTariffNumber,
+                                         keyboardType: .numberPad)
+                    Divider()
+                        .padding(.leading, Constants.horizontalSpacing)
+                }
+                .padding(.horizontal, insets: safeAreaInsets)
```
rachelmcr, line 61: I just noticed that the HS Tariff Number row is indented farther than the other rows in landscape mode, I think because this padding isn't needed:

Simulator Screen Shot - iPhone 11 Pro - 2021-08-17 at 12 41 34

rachelmcr: And now I'm noticing that the background for the item details isn't edge-to-edge, so we should fix that at some point (doesn't have to be right now). Sorry for missing that in my earlier review!

rachelmcr (approved):
The validation works great! Thanks for adding the reusable validation row 🤩 

I just had a couple small comments, but nothing that has to be blocking for this PR.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetailsViewModel.swift
itsmeichigo: Thank you for pointing that out! I have implemented the behavior in d6b94761968b6de4dae5d33a587b4673e0aae996.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetails.swift
itsmeichigo: These seem to be regression issues after I restructured the stacks on the view 😅 I have fixed them in 859d4ad14116f0b7eb51dfae7be2b1b6a7090a85.

itsmeichigo:
Thank so much @rachelmcr for helping me find the issues! I have pushed some changes and this PR is ready for another look!

EDIT: I didn't notice that you have already approved the PR - so I think I'll merge right away 😅 Do let me know if you think there's some problem with the new changes, I'll update it in the next PR!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetails.swift
rachelmcr: Yeah, I was thinking it might be worth doing an overall visual check (for things like landscape, dynamic type, etc.) before we release each feature in this milestone. I've noticed regressions like this sneak in to some of my UI work and it might be easier to do a targeted review for anything like this once at the end.

PR #4812, opened 2021-08-17
Added dismiss upon selection to product 'Order By'

### LINKED ISSUE #4496: Products Bottom Sheet Should be Dismissed After Choosing an Option
### Expectations

The Products List' Sort and Filter bottom sheets should be dismissed right away once you pick an option. 

### Actual

The Products List' Sort and Filter bottom sheets are not dismissed once you pick an option. 

https://user-images.githubusercontent.com/198826/123838807-a26cd680-d8c9-11eb-9455-cb6e7effc6c7.MP4

### To Reproduce

1. Navigate to Products List
2. Tap on Sort or Filter
3. Pick a different option. 

The bottom sheet should be dismissed.

iamgabrielma:
Fixes #4496 

## Description

This PR fixes the Products List "Sort by" modal that appears when the button is tapped. Now the modal will disappear after the user has selected one of the available choices.

4496-fix

## Changes

I added a `dismiss()` call at the end of the current `tableView(_:didSelectRowAt:)` delegate method for `BottomSheetListSelectorViewController`, so is triggered after the user selects one of the options.

The original issue mentions the Filter button as well, but this one does not display a modal when tapped, possibly has been changed since the original issue was reported, so this change only affects the "Order by" button.

## Testing steps

1 - Navigate to Products List
2 - Tap on `Sort by` 
3 - Pick a different option.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/BottomSheet/ListSelector/BottomSheetListSelectorViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/BottomSheet/ListSelector/BottomSheetListSelectorViewController.swift
@@ -78,7 +78,6 @@ UIViewController, UITableViewDataSource, UITableViewDelegate where Command.Model
             command.handleSelectedChange(selected: selected)
             tableView.reloadData()
         }
-        dismiss(animated: true, completion: nil)
     }
 
     func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
--- a/WooCommerce/Classes/ViewRelated/BottomSheet/ListSelector/BottomSheetListSelectorViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/BottomSheet/ListSelector/BottomSheetListSelectorViewController.swift
@@ -78,6 +78,7 @@ UIViewController, UITableViewDataSource, UITableViewDelegate where Command.Model
             command.handleSelectedChange(selected: selected)
             tableView.reloadData()
         }
+        dismiss(animated: true, completion: nil)
     }
 
     func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
--- a/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
@@ -3,7 +3,7 @@ import XCTest
 @testable import Yosemite
 
 final class ProductsSortOrderBottomSheetListSelectorCommandTests: XCTestCase {
-    func testInitialSelectedValue() {
+    func test_initial_selected_value_is_stored_in_command() {
         // Arrange
         let selected = ProductsSortOrder.nameAscending
         var selectedActions = [ProductsSortOrder]()
@@ -16,44 +16,35 @@ final class ProductsSortOrderBottomSheetListSelectorCommandTests: XCTestCase {
         XCTAssertEqual(command.selected, selected)
     }
 
-    func testSelectedValueAfterChange() {
+    func test_selected_value_after_change() {
         // Arrange
         let selected = ProductsSortOrder.nameAscending
-        var selectedActions = [ProductsSortOrder]()
         let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected) { (selected) in
-            selectedActions.append(selected)
+            // noop
         }
-        // Action
-        let newSelected = ProductsSortOrder.dateDescending
-        command.handleSelectedChange(selected: newSelected)
-
         // Assert
-        XCTAssertEqual(command.selected, newSelected)
+        XCTAssertEqual(command.selected, selected)
     }
 
-    func testIsSelectedValueAfterChange() {
+    func test_isSelected_returns_true_if_given_the_initial_value() {
         // Arrange
-        let selected = ProductsSortOrder.nameAscending
-        var selectedActions = [ProductsSortOrder]()
-        let notSelected: [ProductsSortOrder] = [.nameDescending, .dateAscending, .dateDescending]
-        //
-        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected) { (selected) in
-            selectedActions.append(selected)
+        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: .nameAscending) { (selected) in
+            // noop
         }
-        XCTAssertTrue(command.isSelected(model: selected))
-        notSelected.forEach { notSelectedSortOrder in
-            XCTAssertFalse(command.isSelected(model: notSelectedSortOrder))
-        }
-
         // Action
-        let newSelected = ProductsSortOrder.dateDescending
-        let newNotSelected: [ProductsSortOrder] = [.nameDescending, .nameAscending, .dateAscending]
-        command.handleSelectedChange(selected: newSelected)
-
+        let isSelected = command.isSelected(model: .nameAscending)
         // Assert
-        XCTAssertTrue(command.isSelected(model: newSelected))
-        newNotSelected.forEach { notSelectedSortOrder in
-            XCTAssertFalse(command.isSelected(model: notSelectedSortOrder))
-        }
+        XCTAssertTrue(isSelected)
     }
+    func test_isSelected_returns_false_if_given_a_different_value() {
+        // Arrange
+        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: .nameAscending) { (selected) in
+            // noop
+        }
+        // Action
+        let isSelected = command.isSelected(model: .nameDescending)
+        // Assert
+        XCTAssertFalse(isSelected)
+    }
+
 }
--- a/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
@@ -4,16 +4,25 @@ import XCTest
 
 final class ProductsSortOrderBottomSheetListSelectorCommandTests: XCTestCase {
     func testInitialSelectedValue() {
+        // Arrange
         let selected = ProductsSortOrder.nameAscending
-        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected)
+        var selectedActions = [ProductsSortOrder]()
+        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected) { (selected) in
+            selectedActions.append(selected)
+        }
+        // Action
+        command.handleSelectedChange(selected: .nameAscending)
+        // Assert
         XCTAssertEqual(command.selected, selected)
     }
 
     func testSelectedValueAfterChange() {
         // Arrange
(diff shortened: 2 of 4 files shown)


WooCommerce/Classes/ViewRelated/BottomSheet/ListSelector/BottomSheetListSelectorViewController.swift
```diff
@@ -78,6 +78,7 @@ UIViewController, UITableViewDataSource, UITableViewDelegate where Command.Model
             command.handleSelectedChange(selected: selected)
             tableView.reloadData()
         }
+        dismiss(animated: true, completion: nil)
```
shiki: I looked at other places where this ViewController is used. It looks like the original expectation is to have the _presenting view controller_ dismiss the `BottomSheetListSelectorViewController`. Examples:

For the "More details" in a product:

https://github.com/woocommerce/woocommerce-ios/blob/c57b3472b0569f732a8e70728fcf42955cc746c5/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/ProductFormViewController.swift#L565-L591

For the product type selection when adding a product: 

https://github.com/woocommerce/woocommerce-ios/blob/c57b3472b0569f732a8e70728fcf42955cc746c5/WooCommerce/Classes/ViewRelated/Products/Add%20Product/AddProductCoordinator.swift#L46-L53

Notice the `dismiss()` calls in those places. And there are a couple more that I have not listed.

Adding this dismissal here would negate those implementations. Or I don't know what else would happen if `dismiss()` is called twice. While we could remove those places where `dismiss()` is called, I think it's safer to follow the convention instead and manually call `dismiss()` where the sorting dialog is presented. 

To do that, it looks like we can change the implementation of [`ProductsSortOrderBottomSheetListSelectorCommand`](https://github.com/woocommerce/woocommerce-ios/blob/c57b3472b0569f732a8e70728fcf42955cc746c5/WooCommerce/Classes/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommand.swift#L22) so it's more like [`ProductTypeBottomSheetListSelectorCommand`](https://github.com/woocommerce/woocommerce-ios/blob/c57b3472b0569f732a8e70728fcf42955cc746c5/WooCommerce/Classes/ViewRelated/Products/Edit%20Product/BottomSheetListSelector/ProductTypeBottomSheetListSelectorCommand.swift#L132). 

Notice that `ProductTypeBottomSheetListSelectorCommand` propagates the `onSelection` event during the `handleSelectedChange` event. That `onSelection` block dismisses the view. We can follow the same strategy with sorting. 

What do you think?

shiki (changes requested):
Hi, @iamgabrielma! Thanks for fixing this. 

This works but I'm worried that the implementation bypasses the current conventions of using the `*Command` classes to dismiss the dialog. I fear that this would cause some hidden regressions in the long run and it might be better to copy the other `*Command` implementations in the code. 

More info inline. 

Let me know what you think.

WooCommerce/Classes/ViewRelated/BottomSheet/ListSelector/BottomSheetListSelectorViewController.swift
iamgabrielma: That make a lot of sense, thanks for the pointers and the detailed walkthrough!

iamgabrielma:
The latest commit break some existing unit tests that don't account for the new implementation, I'll review them shortly 👍 

```
/Users/distiller/project/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift:8:89: missing argument for parameter 'onSelection' in call

let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected)
        /Users/distiller/project/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift:15:89: missing argument for parameter 'onSelection' in call

let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected)
    /Users/distiller/project/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift:29:89: missing argument for parameter 'onSelection' in call

let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected)
```

shiki:
No worries! Please press the re-review button if you want me to take a look at this again. 🙂 I looked at the changes and it seems like what we expected! Those darn unit tests!

iamgabrielma:
> No worries! Please press the re-review button if you want me to take a look at this again. 🙂 I looked at the changes and it seems like what we expected! Those darn unit tests!

Great! In the latest changes I changed the implementation of `ProductsSortOrderBottomSheetListSelectorCommand` so it follows the same convention that other bottom sheets behaviour, and dismissal is handled by `onSelection` and `handledSelectedChange` rather than hardcoding the dismissal after the closure.

Regarding the Unit Tests I may need some pointer on how to handle these, as I'm not entirely sure how to test this part of the new implementation for the asynchronous code:

I fixed `testInitialSelectedValue()` as the async code does not affect its initial selected value, but I'm having some troubles with `testSelectedValueAfterChange()` and `testIsSelectedValueAfterChange()`, and I'm not sure if I should change these two to test if the callback is handled correctly instead, as for example [ProductTypeBottomSheetListSelectorCommandTests.swift](https://github.com/woocommerce/woocommerce-ios/blob/fa2bd01933ceb2c1a4e36d434048d134125fad2c/WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductTypeBottomSheetListSelectorCommandTests.swift) does.

WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
```diff
@@ -4,16 +4,25 @@ import XCTest
 
 final class ProductsSortOrderBottomSheetListSelectorCommandTests: XCTestCase {
     func testInitialSelectedValue() {
```
shiki: We can rename this to clarify it further:

```suggestion
    func test_initial_selected_value_is_stored_in_command() {
```

This is using a very old test method convention. Our new conventions use snake case: https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/naming-conventions.md#test-methods. 

Also, the `testInitialSelectedValue` does not really explain what is _expected_. My suggestion isn't that great but it's probably at least a little bit better than what we have here. 🤪 

I don't think this test is very helpful but it doesn't harm anyone to keep it.

WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
```diff
@@ -4,16 +4,25 @@ import XCTest
...
+        // Arrange
         let selected = ProductsSortOrder.nameAscending
-        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected)
+        var selectedActions = [ProductsSortOrder]()
+        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected) { (selected) in
+            selectedActions.append(selected)
+        }
+        // Action
+        command.handleSelectedChange(selected: .nameAscending)
+        // Assert
```
shiki, line 15: Since we're only testing that the initial value is set and **we never modify it after that**, a lot of these can be removed. 

```suggestion
        // Arrange
        let selected = ProductsSortOrder.nameAscending

        // Action
        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected) { (selected) in
            // noop
        }

        // Assert
        XCTAssertEqual(command.selected, selected)
```

- The `selectedActions` is not tested so we don't need to keep it
- The `command.handleSelectedChange(selected: .nameAscending)` has [no effect aside from calling the callback](https://github.com/woocommerce/woocommerce-ios/blob/871b9910e5178d39800a11c23af4168f6c55ad8e/WooCommerce/Classes/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommand.swift#L46-L48) so it doesn't really affect the value of `command.selected`. 

The shorter the test, the easier it is to understand. 😊

WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
```diff
@@ -25,8 +34,12 @@ final class ProductsSortOrderBottomSheetListSelectorCommandTests: XCTestCase {
     func testIsSelectedValueAfterChange() {
```
shiki: It looks like `testIsSelectedValueAfterChange` was originally testing the that `command.isSelected` was returns the correct boolean value for the sort value passed to `command.handleSelectedChange()`. However, `handleSelectedChange` doesn't do anything that affects that since it now just calls a callback function.

We can repurpose this to only test the initial value. Something like this: 

```swift
    func test_isSelected_returns_true_if_given_the_initial_value() {
        // Arrange
        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: .nameAscending) { (selected) in
            // noop
        }

        // Action
        let isSelected = command.isSelected(model: .nameAscending)

        // Assert
        XCTAssertTrue(isSelected)
    }
```

We can also add another unit test that returns `false` if the incorrect sort value is passed but I leave that up to you if you want to do that. It's not that critical.

WooCommerce/WooCommerceTests/ViewRelated/Products/BottomSheetListSelector/ProductsSortOrderBottomSheetListSelectorCommandTests.swift
```diff
@@ -4,16 +4,25 @@ import XCTest
...
+        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: selected) { (selected) in
+            selectedActions.append(selected)
+        }
+        // Action
+        command.handleSelectedChange(selected: .nameAscending)
+        // Assert
         XCTAssertEqual(command.selected, selected)
     }
 
     func testSelectedValueAfterChange() {
```
shiki: It looks like the original code is testing that `command.selected` changes whenever `command.handleSelectedChange` is called with a different value. Since `handleSelectedChange`'s behavior changed, this is no longer valid. 

You're right that we can replace this with a unit test that verifies that the callback was called. We can replace this test function with something like this: 

```swift
    func test_handleSelectedChange_notifies_the_callback_the_new_selected_value() {
        // Arrange
        var receivedSelectedSortOrder: ProductsSortOrder?
        let command = ProductsSortOrderBottomSheetListSelectorCommand(selected: .nameAscending) { (selected) in
            receivedSelectedSortOrder = selected
        }

        // Action
        let newSelected = ProductsSortOrder.dateDescending
        command.handleSelectedChange(selected: newSelected)

        // Assert
        XCTAssertEqual(receivedSelectedSortOrder, newSelected)
    }
```

We can have a local variable, `receivedSelectedSortOrder`, that is updated whenever the callback is called. And test that it's updated to what we expect.

shiki (approved):
Hi, @iamgabrielma! You're right about those tests. They need to be repurposed. I have left suggestions for how we can handle this. Feel free to apply, modify, add until you're satisfied. If you have any questions, let me know. 🙂 

Once the tests pass, feel free to merge. :shipit: But if you'd like me to take a look before merging, that's cool too.

Issue #4717, opened 2021-08-03
Shipping Labels: allow moving items to new packages or shipping them in their original packaging.


pmusolino:
We should allow moving items to new packages or shipping them in their original packaging. Use Figma as the source of truth.

<img width="682" alt="Schermata 2021-08-03 alle 17 36 46" src="https://user-images.githubusercontent.com/495617/128044299-0ed9b982-74e3-41c8-88c2-f8139811816c.png">

PR #4814, opened 2021-08-17
Fix unbound bash variables in GenerateCredentials script

AliSoftware:
Some variables in our `generate-credentials.sh` script were unbound (aka non-existing variables, due to typo). This small PR fixes this so that external contributors (and trialees) can work with the project.

See discussion in p1629212320273000-slack-C6H8C3G23

### DIFF
--- a/Scripts/build-phases/generate-credentials.sh
+++ b/Scripts/build-phases/generate-credentials.sh
@@ -34,7 +34,7 @@ if [ ! -f $SECRETS_PATH ]; then
     ## then copy it into place for the build.
     ##
     if [ ! -f $CREDS_OUTPUT_PATH ]; then
-        echo ">> Creating Credentials File from Template: ${CREDS_FILE_PATH}"
+        echo ">> Creating Credentials File from Template: ${CREDS_TEMPLATE_PATH}"
         cp ${CREDS_TEMPLATE_PATH} ${CREDS_OUTPUT_PATH}
     fi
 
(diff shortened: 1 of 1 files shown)


ealeksandrov (approved):

caiobzen (approved):

AliSoftware:
@mokagio FYI, I didn't check if the other repositories on which you also migrated the secrets outside the repo might have a similar typo, could be worth a quick check?

I don't think we'll find the same issue in other repos tbh, since to my knowledge this particular script is pretty specific to WCiOS, and the unbound variables we fixed here don't seem to be directly related to your changes in #4717… _but_ the error started to appear because your changes [(rightfully) introduced `-euo pipefile`](https://github.com/woocommerce/woocommerce-ios/pull/4713/files#diff-d49209d2dbe20001133152e9d805592a64cfe43ffc952caa5e35f89e90d02224R1), so the unbound variables were already there before but the script only failed _silently_ without reporting the error, while now it does.

Which is a good thing ultimately (that allowed us to catch the issue 👍 ), but makes me think it might be worth double-checking if such previously-uncaught unbound variables existed as well in _other_ scripts in _other_ repos – and we might have failed to test the scenario of new contributor / trialee exhaustively for those during our PR reviews (especially worth trying to do a clean checkout, but most importantly temporarily removing mobile secrets and `~/.configure` in `$HOME` as well to simulate a clean machine, etc) 🤷‍♂️

mokagio:
Thanks @AliSoftware 🙌 

I tested WordPress just now and the external contributor workflow works:

image

I'll make another note here when I'll have checked Simplenote.

mokagio:
Closing the loop:

### Simplenote macOS ✅

image

### Simplenote iOS 🙅‍♂️

image

But... that hasn't been the case for a while and is [a known "issue"](https://github.com/Automattic/simplenote-ios/pull/1284#discussion_r635489960) and is [documented in the `README`](https://github.com/Automattic/simplenote-ios/tree/c771fcd43949c32d23f56e2c281a9d932aa5719b#setup-credentials).

I've logged a request internally on P2 to avoid forgetting about this.

PR #4799, opened 2021-08-12
Shipping Labels: Add custom package creation form

rachelmcr:
Part of: #4743 
Based on: #4797 **Do not merge until this PR is merged**

## Description

In the shipping label creation flow, this introduces the UI to enter the details for a new custom package to use for your shipping label. For now, the form is visual only — the "Package Type" row does nothing (it will eventually open another screen, to be implemented) and data cannot be saved (data handling will be implemented later).

## Changes

* Adds a `ShippingLabelCustomPackageForm` and view model with all of the rows for creating a new custom package.
* Changes to `ShippingLabelAddNewPackage`:
   * Adds a `selectedView` property to the `ShippingLabelAddNewPackageViewModel`, and uses that property to display the selected subview on `ShippingLabelAddNewPackage`.
   * Updates the view with a `GeometryReader` to allow the custom form backgrounds to extend edge-to-edge while keeping the content in the safe area.

/|Portrait|Landscape
-|-|-
Light|Simulator Screen Shot - iPhone 11 Pro - 2021-08-12 at 20 45 07|Simulator Screen Shot - iPhone 11 Pro - 2021-08-12 at 20 45 29
Dark|Simulator Screen Shot - iPhone 11 Pro - 2021-08-12 at 20 45 16|Simulator Screen Shot - iPhone 11 Pro - 2021-08-12 at 20 50 53

## Testing

1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more packages set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Launch the app in debug mode (so the feature flag is enabled).
3. Open an order detail that is eligible for the shipping label.
4. Tap the button "Create Shipping Label.”
5. Confirm "Ship From" and "Ship To.”
6. In the package details section, tap on “Package Selected.”
7. On the Package Selected screen, tap the “Create new package” button at the bottom of the screen.
8. On the "Add New Package" screen, confirm that when "Custom Package" is selected you see the expected form to enter the details for your custom package.

## Submitter Checklist

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4816, opened 2021-08-17
Shipping Labels: Package type selection and default values for custom package creation UI

rachelmcr:
Part of: #4743

## Description

In the shipping label creation flow, this adds to the UI (started in #4799) to enter the details for a new custom package to use for your shipping label. In this PR:

* The "Package Type" field is now active, so you can select the package type from a list.
* The default values for the form are set:
   * The length and weight units are set to the units in the store's settings.
   * The package type is set to "Box" by default.
   * Other fields (package dimensions and weight) are empty.

## Changes

* A new `PackageType` enum is added as an extension to `ShippingLabelCustomPackageFormViewModel` to represent the available package types. (In the networking model `ShippingLabelCustomPackage` this is represented simply by the `bool` property `isLetter`.)
* A `SelectionList` is added to `ShippingLabelCustomPackageForm` to select the package type.
* The expected default values are set in the `ShippingLabelCustomPackageFormViewModel` initializer.
* The view models for `ShippingLabelAddNewPackage` and `ShippingLabelCustomPackageForm` are now designated as `StateObject` to prevent them from losing state when the view is redrawn (e.g. after closing the selection list).
* A basic unit test is added for `ShippingLabelCustomPackageFormViewModel`, along with a `MockShippingSettingsService` to facilitate testing.

Before|After|After (package type)
-|-|-
Add New Package - Custom Package - Before|Simulator Screen Shot - iPhone 11 Pro - 2021-08-18 at 00 10 16|Simulator Screen Shot - iPhone 11 Pro - 2021-08-18 at 00 10 19

## Testing

8. On the "Add New Package" screen, confirm that "Package Type" defaults to "Box" and you can tap that field to select the package type.
9. Confirm that your store's length unit (e.g. "in" or "cm") appears next to the number field in the "Length," "Width," and "Height" rows.
10. Confirm that your store's weight unit (e.g. "oz" or "kg") appears next to the number field in the "Empty Package Weight" row.
11. Confirm all of the fields aside from package type default to empty and the expected placeholders appear.

## Submitter Checklist

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageFormViewModel.swift
```diff
@@ -13,26 +13,65 @@ final class ShippingLabelCustomPackageFormViewModel: ObservableObject {
...
 
     /// The weight of the custom package when empty
     ///
-    @Published var emptyPackageWeight: String = "0"
+    @Published var emptyPackageWeight: String
+
+    init(lengthUnit: String? = ServiceLocator.shippingSettingsService.dimensionUnit,
+         weightUnit: String? = ServiceLocator.shippingSettingsService.weightUnit,
+         packageName: String = "",
+         packageType: PackageType = .box,
```
rachelmcr, line 41: Just a note that although the designs show a "Select Type" placeholder (rather than defaulting to "Box") for the package type, both web and Android behavior defaults to the "Box" type, so I follow that behavior here.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageFormViewModel.swift
```diff
@@ -13,26 +13,65 @@ final class ShippingLabelCustomPackageFormViewModel: ObservableObject {
...
+         packageLength: String = "",
+         packageWidth: String = "",
+         packageHeight: String = "",
+         emptyPackageWeight: String = "") {
```
itsmeichigo, line 45: It's not apparent to me why we need to keep all these in the intializer. Maybe I'm missing out on some special cases - can you help me understand here?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
```diff
@@ -2,7 +2,7 @@ import SwiftUI
 
 struct ShippingLabelAddNewPackage: View {
     @Environment(\.presentationMode) var presentation
-    @ObservedObject private var viewModel = ShippingLabelAddNewPackageViewModel()
+    @StateObject private var viewModel = ShippingLabelAddNewPackageViewModel()
```
itsmeichigo, line 5: Kudos 👏 This is new to me too #TIL!

WooCommerce/WooCommerceTests/ViewRelated/Shipping Label/Create Shipping Label/ShippingLabelCustomPackageFormViewModelTests.swift
```diff
@@ -0,0 +1,25 @@
...
+@testable import Yosemite
+@testable import Storage
+
+class ShippingLabelCustomPackageFormViewModelTests: XCTestCase {
+
+    func test_properties_return_expected_values_for_empty_form() {
+        // Given
+        let shippingSettingsService = MockShippingSettingsService(dimensionUnit: "in", weightUnit: "oz")
+        ServiceLocator.setShippingSettingsService(shippingSettingsService)
+        let viewModel = ShippingLabelCustomPackageFormViewModel()
```
itsmeichigo: Since we already have `lengthUnit` and `weightUnit` in the intializer of `ShippingLabelCustomPackageFormViewModel` I think we can easily inject those without the need for a mock service?

itsmeichigo (approved):
I've tested and everything works great! I left a few comments which shouldn't block this PR. :shipit:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageFormViewModel.swift
rachelmcr: My thinking here was to keep them in the initializer because it makes it easy to initialize an empty form, but it also would make the form more flexible/reusable later on, for example if we want to support editing existing custom packages in the future. Do you think that's overkill here? (I'm still trying to figure out the right balance between implementations that make sense just for what we're doing now vs. thinking ahead to how we might want to use them later on.)

WooCommerce/WooCommerceTests/ViewRelated/Shipping Label/Create Shipping Label/ShippingLabelCustomPackageFormViewModelTests.swift
rachelmcr: That's true. I was thinking that this way the test more closely mimics the expected behavior (getting those values from the service by default) but you're right that it isn't really needed here. Thanks!

rachelmcr:
I'll go ahead and merge this PR, but @itsmeichigo I'm curious about your thoughts on the initializer and can definitely update it in one of my upcoming PRs.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageFormViewModel.swift
itsmeichigo: > for example if we want to support editing existing custom packages in the future

Oh I get it now! I see that we have that edit feature in the web also so that makes sense to think a bit ahead! Although I'd prefer to keep the initializer "neat" with just an optional package (which defaults to `nil`) and we can use it to fill out the fields. That also matches with what we're doing with other view models so far.

But if we allow editing then we'll need to create the view model from outside and inject it into the view instead of letting the view create it like what we're having now.

rachelmcr: > Although I'd prefer to keep the initializer "neat" with just an optional package (which defaults to nil) and we can use it to fill out the fields. That also matches with what we're doing with other view models so far.

Ah, that's definitely a simpler approach. Thanks for the suggestion! I'll make that change in my next PR. (It's true that we'll still need to make changes to how the view model is created, to support that scenario, but at least the view model will be set up for it already.)

PR #4824, opened 2021-08-18
[CPP Onboarding] UI review

koke:
Using this PR to do a full UI review for CPP Onboarding. This is not meant to be merged, we'll handle the necessary fixes in separate PRs.

I replaced the link to In-Person Payments with a temporary selector so it's easier to inspect all the possible states:

CPP-selector

I've recorded a video showcasing light vs dark mode and portrait vs landscape, but I'd recommend reviewing by running this branch directly on a device, and going to Settings > In-Person Payments:

https://cloudup.com/chFJNF2YTxg

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsLoadingView.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsLoadingView.swift
@@ -2,7 +2,7 @@ import SwiftUI
 
 struct InPersonPaymentsLoading: View {
     var body: some View {
-        VStack {
+        ScrollableVStack {
             Spacer()
 
             VStack(alignment: .center, spacing: 42) {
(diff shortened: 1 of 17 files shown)


koke:
The screens that have a support link might benefit from some padding between that and the learn more link when the content doesn't fit the screen

On landscape:
IMG_BEA69A928AE1-1

Normal|Larger text
-|-
IMG_B02E04CA60B4-1|IMG_669F7F3700E3-1

koke:
I think the "learn more" icon would look better top-aligned with the larger font sizes, but I also think it looks better vertically centered when the text is smaller, so maybe it's fine as it is.

koke:
Loading screen is left-aligned on iPad (fixed in e6d23fa6323f76151f6b4e4f3d15ad7407738c2d)

Screen Shot 2021-08-18 at 16 06 19

Garance91540 (approved):
This is looking good! Thank you for including a video. As you said, that might be easier for me to review once I can test it on my device, so you might see more comments during beta testing. 

The only thing I could pick up from the recording would be to add "..." after "Please wait" to convey the sense of progress (so "Please wait...)

PR #4825, opened 2021-08-18
Products: Shouldn't be able to schedule a sale without sale price

### LINKED ISSUE #2496: Products: Shouldn't be able to schedule a sale without sale price
**Describe the bug**

@Garance91540 noticed it was possible to schedule a sale without entering a sale price. In that case it seems like the sale price is set to 0$. Can we display the following error message if users try to schedule a sale without entering a sale price?

ios-add-price-1

##### Tested in Woo iOS 4.6.0.0 beta (internal ref: p5T066-1o1-p2#comment-5289)

To be worked by @caiobzen

caiobzen:
Closes #2496 

# Description

This PR fixes a behavior when saving a new sale without specifying a sale price.

# Changes

- Added a new error case `.newSaleWithEmptySalePrice` to be handled by `ProductPriceSettingsViewController`  
- Updated `ProductPriceSettingsViewModel` to trigger the new error in case of a sale without a salePrice
- Added tests + minor cleanups on `ProductPriceSettingsViewModelTests `

# Videos

| Before  |  After  |
|   |    |

# Testing

1. Make sure you are in a store with at least one product
1. In Products tab of the app, select any product
1. Select the **Price** cell
1. Toggle on the **Schedule sale** toggle
1. Press **Done** on the top right navigation bar button
1. **It should display an alert to warn users about the empty sale price** 
---

## Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
@@ -238,7 +238,7 @@ extension ProductPriceSettingsViewModel: ProductPriceSettingsActionHandler {
 
     func completeUpdating(onCompletion: ProductPriceSettingsViewController.Completion, onError: (ProductPriceSetingsError) -> Void) {
         // Check if sale price was set if there is a sale to be created
-        if dateOnSaleStart != nil && dateOnSaleEnd != nil && salePrice == nil {
+        if dateOnSaleStart != nil && dateOnSaleEnd != nil && getDecimalPrice(salePrice) == nil {
             onError(.newSaleWithEmptySalePrice)
             return
         }
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
@@ -44,6 +44,7 @@ protocol ProductPriceSettingsActionHandler {
 enum ProductPriceSetingsError: Error {
     case salePriceWithoutRegularPrice
     case salePriceHigherThanRegularPrice
+    case newSaleWithEmptySalePrice
 }
 
 /// Provides view data for price settings, and handles init/UI/navigation actions needed in product price settings.
@@ -236,6 +237,11 @@ extension ProductPriceSettingsViewModel: ProductPriceSettingsActionHandler {
     // MARK: - Navigation actions
 
     func completeUpdating(onCompletion: ProductPriceSettingsViewController.Completion, onError: (ProductPriceSetingsError) -> Void) {
+        // Check if sale price was set if there is a sale to be created
+        if dateOnSaleStart != nil && dateOnSaleEnd != nil && salePrice == nil {
+            onError(.newSaleWithEmptySalePrice)
+            return
+        }
 
         // Check if the sale price is populated, and the regular price is not.
         if getDecimalPrice(salePrice) != nil, getDecimalPrice(regularPrice) == nil {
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
```diff
@@ -236,6 +237,11 @@ extension ProductPriceSettingsViewModel: ProductPriceSettingsActionHandler {
     // MARK: - Navigation actions
 
     func completeUpdating(onCompletion: ProductPriceSettingsViewController.Completion, onError: (ProductPriceSetingsError) -> Void) {
+        // Check if sale price was set if there is a sale to be created
+        if dateOnSaleStart != nil && dateOnSaleEnd != nil && getDecimalPrice(salePrice) == nil {
+            onError(.newSaleWithEmptySalePrice)
+            return
+        }
```
Ecarrion: I think this can be rephrased to something like 
```swift
guard doesScheduleDateHasPrice() {
     return onError(.newSaleWithEmptySalePrice)
}
```

Which is a tad bit easier to read, what do you think?

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModelTests.swift
```diff
@@ -512,6 +512,26 @@ final class ProductPriceSettingsViewModelTests: XCTestCase {
...
+        // Act
+        let expectation = self.expectation(description: "Wait for error")
+        viewModel.completeUpdating(onCompletion: { (_, _, _, _, _, _, _) in
+            XCTFail("Completion block should not be called")
+        }, onError: { error in
+            // Assert
+            XCTAssertEqual(error, .newSaleWithEmptySalePrice)
+            expectation.fulfill()
+        })
+        waitForExpectations(timeout: Constants.expectationTimeout, handler: nil)
```
Ecarrion: We have some `waitFor` functions defined in https://github.com/woocommerce/woocommerce-ios/blob/develop/TestKit/Sources/TestKit/XCTestCase+Wait.swift that can help this test to be read in a more synchronous way.

EG:
https://github.com/woocommerce/woocommerce-ios/blob/4fe952211777fad134084cbc19d4db229a37c8c9/Networking/NetworkingTests/Remote/ProductAttributesRemoteTests.swift#L54-L68

Ecarrion (changes requested):
Congrats on submitting your first PR @caiobzen!

I left you some suggestions & I'm also worried that this toast is not visible when the product form is presented modally.


https://user-images.githubusercontent.com/562080/129973160-3445969f-3946-4f2a-9797-8c2f81677966.mov



Can we fix that, please?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModel.swift
caiobzen: Sure thing! I'll be changing this and I will also remove that comment 😉

Ecarrion:
Also, since this is a user-facing change, I think you should update the Release Notes file too!

WooCommerce/WooCommerceTests/ViewRelated/Products/Edit Product/Edit Price/ProductPriceSettingsViewModelTests.swift
caiobzen: That's great! Thanks for sharing 🙏 The test looks super clean now!

caiobzen:
@Ecarrion  Thanks for the suggestions and also for the finding that issue with the alert + modal page sheet!

Ecarrion (approved):
Looks and works great! 🚀

PR #4829, opened 2021-08-19
Shipping Labels: Complete validation for customs form

### LINKED ISSUE #4687: Shipping Labels: Customs form for international shipping

itsmeichigo:
Closes #4687 

# Description
This PR aims to complete validation of customs form with two parts:
### Customs form list Done button. 
To prepare for multi-package support of Shipping Labels, the customs screen was structured like so: Customs Form List -> Customs Form -> Item Details. 

For validation, we want to enable / disable Done button on the Customs Form List immediately when any field is updated. So the validation states have to be sent upward from a child item to its parent form, all the way to the containing form list.

### ITN field validation
It is required that that if total value of items that share the same HS tariff number exceeds $2,500, ITN field in the customs form is required.

To do this, each item has to let its parent know of the latest validation states of the item's value and tariff number. The parent will collect these states to calculate whether any tariff number valued more than $2,500, to decide whether ITN field should be required.

# Changes
### Item details view model
- Added new `@Published` variables for validated tariff number and total value of the item. These will be observed by the parent form to check which tariff number's value exceed $2,500 and requires ITN.
- Added ~~property observers~~ mappers for `value` and `hsTariffNumber` to update the validated total value and tariff number.
- Added new `@Published` variable `validItem` to let parent form know if all the fields in the item details are valid or not.
- Added private method helpers for validation, which are reused between calculated variables and overall validation check.

### Customs form input view model
- Added new `@Published` variable for `validForm` to let parent list know if all the fields as well as items in the form are valid.
- Observe all item view models' `validItem` to save the validation states for all items. These states are used to validate the form.
- Observe all item view models' `validatedHSTariffNumber` and `validatedTotalValue` and save them to configure a list of tariff numbers that require ITN.
- Added private method helpers for validation, which are reused between calculated variables and overall validation check.

### Customs form list view model
- Observe validation states of all child forms and save them locally.
- Configured Done button to enable only when all child forms are valid.
- Configured validated form list by fetching validated form from all input view models.

### Customs form list UI
- Return the validated form list when Done button is tapped.
- Disable Done button when it should not be enabled

### Update unit tests for view models.

# Demo
customs-validation

# Testing steps
2. Create an order that requires international shipping (origin country different from destination country).
3. Open app, navigate to Orders tab and select the created order.
5. Skip through Ship From, Ship To, Package Details to configure Customs.
6. Edit an item in the customs form to have a 6-digit tariff number, and a value of more than $2500 (or if there are more than 1 item, update all item to have same tariff number and total value more than $2500). Notice that there's an validation error in ITN field showing up asking for input.
7. Edit fields in the form to invalid values, notice that Done button is disabled. Update the fields to be valid, Done button should be enabled.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetailsViewModel.swift
@@ -48,7 +48,7 @@ final class ShippingLabelCustomsFormItemDetailsViewModel: ObservableObject {
 
     /// Whether all fields are validated.
     ///
-    @Published private(set) var isItemValidated: Bool = false
+    @Published private(set) var validItem: Bool = false
 
     /// The validated tariff number to be observed by `ShippingLabelCustomsFormInputViewModel`
     /// to calculate total value for each tariff number.
@@ -161,6 +161,8 @@ private extension ShippingLabelCustomsFormItemDetailsViewModel {
         return hsTariffNumber
     }
 
+    /// Observe changes in all fields to check if everything is validated.
+    ///
     func configureValidationCheck() {
         let groupOne = $description.combineLatest($value, $weight)
         let groupTwo = $hsTariffNumber.combineLatest($originCountry)
@@ -178,13 +180,9 @@ private extension ShippingLabelCustomsFormItemDetailsViewModel {
                     self.getValidatedWeight(from: weight) != nil
             }
             .removeDuplicates()
-            .assign(to: &$isItemValidated)
+            .assign(to: &$validItem)
     }
-}
 
-// MARK: - Helpers
-//
-private extension ShippingLabelCustomsFormItemDetailsViewModel {
     /// Limit length HS Tariff Number to only 6 characters max.
     ///
     func limitHSTariffNumberLength() {
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ItemDetails/ShippingLabelCustomsFormItemDetailsViewModel.swift
@@ -20,14 +20,7 @@ final class ShippingLabelCustomsFormItemDetailsViewModel: ObservableObject {
 
     /// Price of item per unit.
     ///
-    @Published var value: String {
-        didSet {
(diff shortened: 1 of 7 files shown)


rachelmcr:
I haven't finished reviewing the code yet, but I noticed a scenario where the validation isn't working as expected:

**Editing an item's tariff number without editing the value**
1. Start with an international order where the order includes more than $2,500 value of an item. (For example, I tested with an order of 30 sunglasses valued at $90 per item, so the value of the sunglasses in the order was $2,700.)
2. In the customs form, open the item details and notice that all of the information is correct except the item doesn't have an HS Tariff Number.
3. Enter a tariff number for the item. At this point, I expect the ITN validation error to appear, but it doesn't.
4. Now, edit the item value and notice the ITN validation error appears.


https://user-images.githubusercontent.com/8658164/130145224-6581d015-6a8f-4349-af88-fcf7dc7e52e6.mp4

Editing to add that this also happens in the opposite scenario:

**Editing an item's value without editing the HS Tariff Number**
1. Start with an international order.
2. In the customs form, open the item details and enter an HS Tariff Number. Ensure the total value for the item doesn't exceed $2,500 (ITN isn't required).
3. Tap the Done button to save the form.
4. Open the customs form again. Confirm the HS Tariff Number is still entered, and edit only the item value per unit so the total value exceeds $2,500. At this point, I expect the ITN validation error to appear, but it doesn't.
5. Now, edit the tariff number and notice the ITN validation error appears.

rachelmcr (changes requested):
Apologies for not having specific comments about the code in this review! I found a couple issues (reported in my comment above) in the validation, but perhaps because I'm new to Combine I can't work out where the problem stems from.

It's a little tricky to follow how everything is connected, especially between the two view models, but for the most part the validation is working for each part of the form and each piece of the validation is documented clearly. I also confirmed that (aside from the two scenarios outlined above) the validation used for enabling the Done button also works great. 👍 

It might be worth having a second reviewer who is more familiar with Combine take a look, too, but I can certainly review/test again when you're ready.

itsmeichigo:
@rachelmcr Thank you for catching that issue about the ITN validation!

The issue was that I used property observer `didSet`, which is only triggered on change of these properties and that's why their initial values weren't taken into account. I fixed that in 891e5e8e0a3d42967764831cb19ac8f917f3b830 by attempting to validate each value of `value` and `hsTariffNumber` instead (using Combine). 

Another way (simpler way) to fix it is to keep the property observers and trigger the validation in the initializer as well, to make sure that initial values are also validated. But I figure I can keep the similar checks in one place with Combine so I chose that route instead 😅.

I understand there are so much complicated logic in this PR - I struggled to put them together too since there are so many states to take care of. I agree that it'd be nice to have an extra pair of eyes to review this, so I'll ask the team for help!

rachelmcr (approved):
Thanks for the explanation, makes sense and the changes look good. 👍  I tested and confirmed that it works as expected now.

I'm approving the PR and will leave it up to you whether you want to wait for another set of eyes on this.

PR #4832, opened 2021-08-19
Shipping Labels: Fix parsing failure for carriers when insurance value is a string instead of double

### LINKED ISSUE #4830: Shipping Labels: Failure to parse carriers & rates response when the returned insurance is a string.
# Description
When fetching carriers and rates for an international order, I got an empty list. Looking at the debug messages I found that there was a failure in mapping `ShippingLabelCarrierRate` as this model is expecting a double value for `insurance` while the response got a string ("limited").

Also - the empty view showed a clipped message (tested on iPhone 12 Pro).

# Screenshot
| iOS | Android |
| ----- | ----- |
|  |  |

# Expectation
Carrier and rates should be parsed successfully whether insurance is a string or number. If insurance is a string, it should be displayed similar to how it is on Android.

Also, if there really is no carriers and rates available, the message on the empty screen should not be clipped off.

itsmeichigo:
Fixes #4830 
⚠️ Testing for this PR requires that customs form validation is complete. So please make sure that #4829  is merged before this! ⚠️

# Description
Currently `ShippingLabelCarrierRate` view model is checking for insurance as a Double. However there are cases that this value can be returned as a string (e.g "limited") - making the parsing to fail and therefore unable to load the carrier and rate list.

This PR updates the parsing to check for both double and string values, and also fix clipped message on the empty state screen as found when the carrier and rate list is empty.

# Changes
- Made `insurance` property of `ShippingLabelCarrierRate` a string, and check for value of both types Double and String. If a double value is found, the value is converted to string.
- Updated view model for Carriers and Rates view to display insurance as before if possible, otherwise render the string value as is.
- Updated unit tests related to carriers and rates in WooCommerce and Networking frameworks.
- Fixed `EmptyState` view to display full length of message.

# Screenshots
| Empty list | List with data |
| ----- | ----- |
|  |  |

# Testing steps
2. Create an order that requires international shipping (origin country different from destination country).
3. Open app, navigate to Orders tab and select the created order.
5. Skip through Ship From and Ship To.
6. In Package Details, select a custom package or DHL package.
7. Skip through Customs (using default configs).
8. In Carriers and Rates screen, notice that the list is loaded properly. The row for DHL should display insurance as "(limited)".
9. For the fix of the empty state screen, turn off wifi before loading carriers and rates. After around 1 minute the empty state screen should display - notice that the text is not clipped off. Please test with iPhone 12 (or Pro) in simulator or device to make sure since this is where the error was found.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelCarrierRate.swift
+++ b/Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelCarrierRate.swift
@@ -6,7 +6,7 @@ import Codegen
 public struct ShippingLabelCarrierRate: Equatable, GeneratedFakeable {
 
     public let title: String
-    public let insurance: Double
+    public let insurance: String
     public let retailRate: Double
     public let rate: Double
     public let rateID: String
@@ -20,7 +20,7 @@ public struct ShippingLabelCarrierRate: Equatable, GeneratedFakeable {
     public let deliveryDateGuaranteed: Bool
 
     public init(title: String,
-                insurance: Double,
+                insurance: String,
                 retailRate: Double,
                 rate: Double,
                 rateID: String,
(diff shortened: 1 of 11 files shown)


rachelmcr (approved):
Code ✅ 
Testing (with iPhone 12 Pro) ✅ 

Thanks for fixing the empty screen issue, too! :shipit:

Issue #4597, opened 2021-07-13
Commercial invoice (customs form) printing in Shipping Labels


pmusolino:
Make sure to follow the release order: #4595

While purchasing an international label, we should check if the label requires a separate customs form or not.

Tasks:
- [x] After purchasing the label, check commercial_invoice_url to see if a label requires a separate customs form or not.
- [x] Add a “print customs form” screen, that fetches the form from commercial_invoice_url and prints it.
- [x] Link to the above screen from “print label” screen.
- [x] Add a button “print customs form” to the label’s UI in order details that open this screen.

PR #4834, opened 2021-08-19
Shipping Labels: Update Storage with new attribute commercialInvoiceURL for ShippingLabel model

itsmeichigo:
Part of #4597 

# Description
Some carriers require that customs form details are printed separately as a commercial invoice instead of including them in shipping labels. To check that invoice is available for printing, we need to fetch URL of the invoice and save it along with the shipping label.

This PR prepares for the saving of the commercial invoice by updating Storage layer with a new attribute for the invoice URL in ShippingLabel model.

# Changes
- Added new Core Data model version 55.
- Added new attribute `commercialInvoiceURL` to `ShippingLabel` model.
- Added unit test for migration from model version 54 to 55.
- Updated migration log.

Update release notes:


pmusolino (approved):
LGTM :shipit:

PR #4836, opened 2021-08-19
[Mobile Payments] [Reader Updates] Add feature flag

allendav:
Partially addresses #4059 

Changes:
- This is the very first PR in a sequence of PRs to hook up a software update UI to card reader settings
- This builds on the changes introduced by @ctarda in https://github.com/woocommerce/woocommerce-ios/pull/3947
- This change only adds the feature flag under which the rest of the development of this feature will proceed
- The next PR will update the `CardReaderSettingsConnectedViewModel` to dispatch a `checkForCardReaderUpdate`

To test:
- Nothing to test

Update release notes:
- [ ] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #4059, opened 2021-04-27
[Mobile Payments] Card Reader Settings v2-13: Card Reader Software Updates


allendav:
Add a new enum enumerating the possible update states:

	.checkingForUpdate
	.noUpdateAvailable
	.updateAvailable
	.updateInProgress
	.updateCompleted
	.updateFailed

Update CardReaderSettingsConnectedViewModel
- to dispatch checkForUpdate when connected to a reader
- expose the readerUpdateStatus Enum and readerUpdateProgress Float to its ViewController
- expose a beginUpdate func for the ViewController to be able to call
- add a didUpdate (similar to how the CardReaderSettingsUnknownViewModel) to inform the VC when changes to the reader update status (or progress) have changed

Update CardReaderSettingsConnectedViewController
- hook into the VM didUpdate for notifications
- in that, monitor for changes to the status or progress and display notice, button, or progress modal as appropriate

NOTE: According to Stripe, we can expect that when the reader update completes, the reader will disconnect and need to be discovered again.

NOTE: Recommend doing this in at least two PRs - i.e. one to show that an update is available and a second PR to actually allow the user to do something about it

ctarda:
@allendav may I suggest implementing this against a feature branch instead of develop?

We still don't know how much stabilizing we are going to have to do for beta 1. If there is not enough bugfixing to do, and we have time to implement this on time, we can merge before the release. But if we get pulled away in order to stabilize beta 1, I think it would be safer to avoid adding any other potential instability. Have in mind that we release every two weeks, so worst case scenario it wouldn't be much of a delay 😄

ctarda:
> may I suggest implementing this against a feature branch instead of develop?

Probably a feature flag specific for this would be simpler?

allendav:
Hey @ctarda - I can name this tune in four small PRs - what do you think?

- [ ] First
    - [ ] Add a feature flag for reader-updates
    - [ ] Feature flag should auto-use simulated reader
    - [ ] Add a onViewModelDidUpdate to the CardReaderSettingsConnectedViewController similar to what is already being done for CardReaderSettingsUnknownViewController, and set the VM didUpdate to it
    - [ ] Dispatch the checkForCardReaderUpdate action from the viewmodel updateProperties method 
    - [ ] Have the “please update your reader” row and button appear if a reader update is available (hook up didUpdate in the VM and fire it in the completion for checkForCardReaderUpdate when an update is available)
[...]

PR #4837, opened 2021-08-19
[Mobile Payments] [Reader Updates] Add checkForCardReaderUpdate to CardReaderSettingsConnectedViewModel

allendav:
Partially addresses #4059 

Changes:
- This is the next PR to bring the card reader software updates to life
- It adds checkForCardReaderUpdate to CardReaderSettingsConnectedViewModel but does not connect it to anything user facing. That will come in an upcoming PR
- Also adds readerUpdateAvailable which the VC will be able to use to know when to show the update button
- Because nothing user facing is added by this PR ( and in fact outside of the unit tests, checkForCardReaderUpdate is not called ) there is no need to gate this particular change behind the feature flag being added in #4836 

To test:
- Ensure all the new unit tests added pass

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -10,6 +10,7 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     private var didGetConnectedReaders: Bool = false
     private var connectedReaders = [CardReader]()
     private let knownReadersProvider: CardReaderSettingsKnownReadersProvider?
+    private(set) var readerUpdateAvailable: CardReaderSettingsTriState = .isUnknown
 
     var connectedReaderID: String?
     var connectedReaderBatteryLevel: String?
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -56,6 +57,22 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
         connectedReaderBatteryLevel = String.localizedStringWithFormat(Localization.batteryLabelFormat, batteryLevelString)
     }
 
+    /// Dispatch a request to check for reader updates
+    ///
+    func checkForCardReaderUpdate() {
+        let action = CardPresentPaymentAction.checkForCardReaderUpdate() { [weak self] result in
+            switch result {
+            case .success(let update):
+                self?.readerUpdateAvailable = update != nil ? .isTrue : .isFalse
```
ctarda: Your choice, but guarding against update being nil would read a bit "swiftier".

ctarda (approved):
Looking good! 

:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: I had it that way first, but it was so much more verbose, i.e.:

```
guard update != nil else {
    self?.readerUpdateAvailable = .isFalse
    self?.didUpdate?()
    return
}

self?.readerUpdateAvailable = .isTrue
self?.didUpdate?()
```

I'd like to leave the ternary. But I do see the opportunity to guard early and return for nil self and I will do that

PR #4839, opened 2021-08-19
[CPP Onboarding] Remove feature flag for In-Person Payments Onboarding

### LINKED ISSUE #4838: [CPP Onboarding] Remove feature flag
Remove the cardPresentOnboarding feature flag

ctarda:
Closes #4838 

This PR removes the feature flag for onboarding into In-Person Payments. The feature is still available only for stores gated into the beta.

| Before | After |
| ---- | ---- |
|  |  |

## Changes
* Remove the feature flag.Bye!

## How to test
* On a store that is gated into In-Person Payments, navigate to Settings. Notice the previous "Manage Card Reader" entry is gone, and now it should read "In-Person Payments"
* Test that it is still possible to connect to a card reader
* Switch to a Store that is not gated into In-Person Payments. Navigate to Settings Check that "In-Person Payments" is not visible

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -209,9 +209,6 @@ private extension SettingsViewController {
     private var storeSettingsRows: [Row] {
         var result = [Row]()
         if canCollectPayments {
-            result.append(.cardReadersV2)
-        }
-        if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentOnboarding) {
             result.append(.inPersonPayments)
         }
         return result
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -209,9 +209,6 @@ private extension SettingsViewController {
     private var storeSettingsRows: [Row] {
         var result = [Row]()
         if canCollectPayments {
-            result.append(.cardReadersV2)
```
allendav, line 212: We don't want to delete this

ctarda: Yes, we do 😄 . This is part of the new "In Person Payments menu item"

allendav: Ah, so the connection flow will still be served by that. I see.

I withdraw my complaint. Carry on.

allendav (approved):
Tests well in light mode and dark mode on iPhone 12 Pro Max :shipit:

allendav (approved):
> Remove other traces of the cardReaderV2 section

Continues to test well in light mode and dark mode on iPhone 12 Pro Max :shipit:
Was able to connect to reader.
Noted that tracks event made it over too.

ctarda:
FYI @koke

PR #4843, opened 2021-08-19
Remove pass supression from alpha builds

jkmassel:
Update release notes:


WooCommerce/Resources/Woo-Alpha.entitlements
```diff
@@ -2,8 +2,6 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>com.apple.developer.passkit.pass-presentation-suppression</key>
-	<true/>
```
AliSoftware, line 6: What happens then if you use the API in the code from an alpha build? (I couldn't test, because I don't have a card reader and the appropriate hardware to test that flow)

If it just does nothing then that's ok, but if it crashes when the entitlement isn't present, could be worth adding either guards in the code for alpha builds, or document somewhere that this is expected in alpha builds? (btw not sure if we have documentation somewhere about what the limitations of our alpha builds like Google Signin –or am I confusing this limitation with Android?– and now this)

AliSoftware (approved):
Took some perseverance but finally managed to test this PR by:

 - ✅  Trying to build the Alpha scheme in Xcode
    - I encountered some ruby/rbenv trouble during the "Generate Credentials" script while trying to build, and that took me some time to find a right fix for it.
    - I finally found a fix, which I pushed on this PR via 41467040a, would appreciate if others test this fix on their machine

 - ✅  Installed the Installable Build from this PR on my device and was able to launch it
    - Apparently I wasn't in the right group in AppCenter and didn't see the build at first…  so had to go get a bit lost in the AppCenter settings rabbit hole to find the right place to add myself, but finally got it 😅 


Approving, but ⚠️  @jkmassel @koke I'd love for you to test the change I made in 41467040a to ensure it still works in other's machines with potentially different ruby/rbenv/bashrc setups before merging 😉

Scripts/build-phases/generate-credentials.sh
```diff
@@ -62,6 +62,16 @@ else
...
+    if which rbenv; then
+      # Fix an issue where, depending on the shell you are using on your machine and your rbenv setup,
+      #   running `ruby` in a bash script from Xcode script build phase might not use the right ruby
+      #   (and thus not find the appropriate gems installed by bundle & Gemfile.lock and crash).
+      # So if rbenv is installed, make sure the shims for `ruby` are too in the context of this bash script,
+      #   so that it uses the right ruby version defined in `.ruby-version` instead of risking to use the system one.
+      eval "$(rbenv init -)"
+      rbenv rehash
+    fi
+
```
AliSoftware, line 74: I added this patch in order to be able to compile, otherwise I had a failure on the build phase when running it on my machine. Not sure why it worked in the past back when I was still working on (and compiling) WCiOS once in a while as RN and failed today, but I figured it probably doesn't hurt anyway?

Would love to find someone who might not have rbenv setup (at all, or in a different way that I did with my `.zshrc` – which this `bash` script doesn't use anyway) to check if this doesn't break _their_ setup

allendav:
Is there an issue I can read about why we needed this change? I'm not contesting it. I am just curious.

koke:
> Approving, but ⚠️ @jkmassel @koke I'd love for you to test the change I made in [4146704](https://github.com/woocommerce/woocommerce-ios/commit/41467040ae6b13bc95594306987ad7e0c6836c0e) to ensure it still works in other's machines with potentially different ruby/rbenv/bashrc setups before merging

I was having trouble with the generate credentials step and this fixes it for me 👏🏽👏🏽👏🏽

AliSoftware:
> Is there an issue I can read about why we needed this change? I'm not contesting it. I am just curious.

@allendav If I'm not mistaken, this is needed because the alpha / internal builds use a different (Enterprise) Apple account, which doesn't have that entitlement provisionned by Apple in our dev portal (they only provisionined it on our regular, AppStore developer account). So when code-signing the Alpha builds with our Enterprise account and profiles, we can't use that entitlement, unlike when using profiles from our regular developer account.

jkmassel:
I've enabled auto-merge to land this PR in `7.5`. If we run into issues with crashes as suggested by @AliSoftware, we can always open another PR at that time (because it would only affect alpha builds it's IMHO fine if we address it later).

PR #4844, opened 2021-08-20
[Mobile Payments] [Reader Updates] Connect VC to VM, add prompt and button if update available

allendav:
Ready for Review

Partially addresses #4059 

Note:
- Don't worry - although this PR detects whether your reader needs an update, it DOES NOT DO THE UPDATE

Changes:
- Kick off checkForCardReaderUpdate when view loads if feature flag is set
- Adds the top row and the button when viewModel readerUpdateAvailable is true
- Adds cell with prompt
- Adds button - doesn’t do anything yet
- Tweaks the CardReaderService protocol checkForUpdate to make CardReaderSoftwareUpdate an optional... I did this because the Stripe Terminal SDK will return a nil if there is no error and no update available and I had assumed this was also how our service worked, which was incorrect
- Tweaks the StripeCardReaderService to properly handle that success nil case
- Tweaks the CardPresentPaymentStore checkForUpdate publisher - it was firing twice because of the use of finished, and that finished was passing success(nil) which should indicate no update available

To test:
- In light mode and in dark mode...
- Tap on Settings (Gear) > In-Person Payments > Manage card reader
- Turn on a reader that needs a software update and connect to it
- A few seconds after connecting, the call to Stripe's servers will complete and display that a reader update is available and give a button to kick it off
- Pressing the button has no effect - that will come in a later PR

update

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
@@ -123,7 +123,10 @@ private extension CardReaderSettingsConnectedViewController {
     private func configureUpdatePrompt(cell: LeftImageTableViewCell) {
         cell.configure(image: .infoOutlineImage, text: Localization.updatePromptText)
         cell.selectionStyle = .none
-        cell.backgroundColor = .warning
+        cell.backgroundColor = .warningBackground
+        cell.imageView?.tintColor = .warning
+        cell.textLabel?.numberOfLines = 0
+        cell.textLabel?.textColor = .text
     }
 
     private func configureConnectedReader(cell: ConnectedReaderTableViewCell) {
@@ -140,6 +143,7 @@ private extension CardReaderSettingsConnectedViewController {
             // TODO in a following PR
         }
         cell.selectionStyle = .none
+        cell.backgroundColor = .clear
     }
 
     private func configureDisconnectButton(cell: ButtonTableViewCell) {
@@ -148,6 +152,7 @@ private extension CardReaderSettingsConnectedViewController {
             self?.viewModel?.disconnectReader()
         }
         cell.selectionStyle = .none
+        cell.backgroundColor = .clear
     }
 }
 
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
@@ -14,6 +14,9 @@ final class CardReaderSettingsConnectedViewController: UIViewController, CardRea
     ///
     private var viewModel: CardReaderSettingsConnectedViewModel?
 
+    // TEMPORARY UNTIL PREVIOUS PR LANDS
+    private var showUpdateControls: Bool = true // viewModel.readerUpdateAvailable == .isTrue
+
     /// Table Sections to be rendered
     ///
     private var sections = [Section]()
@@ -54,11 +57,35 @@ private extension CardReaderSettingsConnectedViewController {
     /// Setup the sections in this table view
     ///
     func configureSections() {
-        sections = [Section(title: Localization.sectionHeaderTitle.uppercased(),
-                            rows: [
-                                .connectedReader,
-                                .disconnectButton
-                            ])]
+        sections = []
+
+        /// This section, if present, displays a prompt to update a reader running old software
+        ///
+        if showUpdateControls {
+            sections.append(
+                Section(title: nil,
+                        rows: [
+                            .updatePrompt
+                        ]
+                )
+            )
+        }
+
+        /// This section displays details about the connected reader
+        ///
+        var rows: [Row] = [.connectedReader]
+
+        if showUpdateControls {
+            rows.append(.updateButton)
+        }
+
+        rows.append(.disconnectButton)
+
+        sections.append(
+            Section(title: Localization.sectionHeaderTitle.uppercased(),
+                    rows: rows
+            )
+        )
     }
 
     func configureTable() {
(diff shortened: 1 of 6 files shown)


Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -196,11 +196,8 @@ private extension CardPresentPaymentStore {
     func checkForCardReaderUpdate(onCompletion: @escaping (Result<CardReaderSoftwareUpdate?, Error>) -> Void) {
         cardReaderService.checkForUpdate()
             .subscribe(Subscribers.Sink { value in
-                switch value {
-                case .failure(let error):
+                if case .failure(let error) = value {
                     onCompletion(.failure(error))
-                case .finished:
```
allendav, line 202: @ctarda - I need to understand this finished better - it causes the publisher to fire twice when an update is available, once with the update and once with nil

for now, I've removed it, but I'll need to add back something in this PR to properly handle when no error and no update is available - see https://github.com/stripe/stripe-terminal-ios/blob/8462926231471e283eb745ac2362aacd32148b5a/docs/1.4.0/Classes/SCPTerminal.html#L1697-L1699

Were you hoping to have finished cover the no-error-no-update situation basically (with the unexpected double publishing I presume)? Or was there another reason?

ctarda: It's good to know when the subscriber has finished / the promise offered by the service has been fulfilled? I would say that this is the expected behaviour.

What would be the problem if the completion block is fired twice, considering the second time it would be fired it would pass a nil as a parameter?

allendav: > What would be the problem if the completion block is fired twice, considering the second time it would be fired it would pass a nil as a parameter?

We have three states to watch for - 1) success-update when there is an update, 2) success-nil if there is no update, 3) failure - an error occurred while trying to check for an update

Calling the completion with the update object and then again with nil when an update is available will make the consuming logic weird.

I'll propose a fix in this PR to make the completion fire once for each of the three cases.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -27,23 +27,58 @@ final class CardReaderSettingsConnectedViewController: UIViewController, CardRea
...
 }
 
 // MARK: - View Configuration
 //
 private extension CardReaderSettingsConnectedViewController {
+    func checkForCardReaderUpdate() {
+        guard ServiceLocator.featureFlagService.isFeatureFlagEnabled(.cardPresentSoftwareUpdates) else {
+            return
+        }
+        guard let viewModel = viewModel else {
```
ctarda, line 62: I was going to say that both guards could be collapsed into one, but leaving the check for the feature flag separated would make removing it easier so 🤷🏼

ctarda:
I see the expected behaviour when connecting to a reader and entering the view that displayed the connected reader for the first time:



However, if I navigate out of that view and then back into it, the banner and the buttons to update the reader are not visible anymore, and I need to wait to a few seconds for them to be visible again. Should we try to cache the state that presents the banner in that view? (this is an open question, not a leading one 😉 )


https://user-images.githubusercontent.com/2722505/130250646-1067f226-007b-47b9-9be4-2bb251ad8ff8.mov

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
ctarda: > Calling the completion with the update object and then again with nil when an update is available will make the consuming logic weird.

I do not see it that way. The client can guard against a value of the software update object and exit early if it is nil.

allendav:
> Should we try to cache the state that presents the banner in that view? (this is an open question, not a leading one 😉 )

I'd like that, and if/when we do that (later PR I suggest), we'll also have to then consider what to do about that if they press the button and it turns out an update is no longer available for whatever reason.

Edit: I'll open an issue, but I suggest we prioritize it low since updates only happen a couple times a year.

ctarda:
> I'd like that, and if/when we do that (later PR I suggest), we'll also have to then consider what to do about that if they press the button and it turns out an update is no longer available for whatever reason.

I think it make sense to delay it to a later PR, but the update disappearing seems like too much of a corner case to me. I do expect it will be more likely that merchants see the prompt for an update when connecting, collect the payment, go back to settings to update the reader, and the UI has changed.

allendav:
Opened https://github.com/woocommerce/woocommerce-ios/issues/4856 for caching update state to avoid re-layout on re-entering

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -98,11 +170,21 @@ private extension CardReaderSettingsConnectedViewController {
...
+        cell.backgroundColor = .clear
+    }
+
+    private func configureDisconnectButton(cell: ButtonTableViewCell) {
+        let style: ButtonTableViewCell.Style = shouldShowUpdateControls() ? .secondary : .primary
+        cell.configure(style: style, title: Localization.disconnectButtonTitle) { [weak self] in
             self?.viewModel?.disconnectReader()
         }
         cell.selectionStyle = .none
+        cell.backgroundColor = .clear
```
allendav, line 187: This was missed in previous PRs. This updates the button background to match the designs in Figma.

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
allendav: Take a look at the changes now that unit tests have been added. It means that when the completion is called (once) it will have the correct state: has-an-update-available, doesn't-have-an-update-available or error-checking-for-an-update

ctarda (approved):
Thanks for clarifying the behaviour of the Future 😅 

:shipit:

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
allendav: Summarizing our offline discussion - I updated the store's handling of the update future to avoid the double completion following the Future example from raywenderlich's Combine textbook in https://github.com/woocommerce/woocommerce-ios/pull/4844/commits/7a2599d601b61a8eda1a214d5c77b26eb00b4d38

I also added a note explaining why an explicit call to `cancel()` is not needed in https://github.com/woocommerce/woocommerce-ios/pull/4844/commits/c8450c9505979ed7ed500c81613d9812206b07a9

PR #4845, opened 2021-08-20
Shipping Labels: Fix issue Shipping Adress screen being pushed twice when address name is missing

### LINKED ISSUE #4701: Shipping Label creation: bug when the destination address is empty
When an order doesn't have a shipping address, when we try to create a shipping label the `Ship To` address will be empty (correct), but the screen will be pushed two times.

https://user-images.githubusercontent.com/495617/127638352-575c19c8-59b3-4569-a22b-c0d5d6f4a1ee.mp4

itsmeichigo:
Closes #4701 

# Description
Currently when either origin or destination address doesn't have a name, tapping on Continue button of Ship From / Ship To row in Shipping Label form will push to Edit Address screen twice. This PR aims to fix that issue.

# Changes
- In `validatedAddress` method we're sending a callback with error if address name is missing. This local check was updated to a `guard` instead of `if`, and return early after sending callback to avoid triggering the remote validation action.
- Updated unit tests to check for triggering of the validation action.

# Demo
edit-address

# Testing
2. Create an order with at least one physical item and no name on the shipping address.
3. Navigate to Orders tab on the app and select the new order.
4. Select Create Shipping Label and skip through Ship From.
5. Tap Continue on Ship To, notice that the Edit Address screen is pushed immediately, with no loading state before that. Wait for around 1 second to see that no extra screen is pushed after that.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -6,6 +6,7 @@
 - [*] Fix: The refund button within Order Details will be hidden if the refund is zero. [https://github.com/woocommerce/woocommerce-ios/pull/4789]
 - [*] Fix: Incorrect arrow direction for right-to-left languages on Shipping Label flow. [https://github.com/woocommerce/woocommerce-ios/pull/4796]
 - [*] Fix: Shouldn't be able to schedule a sale without sale price. [https://github.com/woocommerce/woocommerce-ios/pull/4825]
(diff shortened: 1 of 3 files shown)


rachelmcr (approved):
Nice fix! Thanks for adding the unit tests, too.
Code ✅ 
Testing ✅ 

I'll go ahead and merge this to make sure it makes it into the code freeze.

PR #4855, opened 2021-08-20
Show percentage options with percentage sign

### LINKED ISSUE #4842: [Product Add-ons] Percentage is being shown as dollar value
I had my add-ons configured liked this: 

And when purchasing, it shows like this: 

Notice the +5.13%. 

However, navigating to the Product → Product Add-ons, we don't show the percentage sign. We incorrectly show a dollar sign instead.

## Expectations

The app should show "+5.13%".

Ecarrion:
fix #4842 

# Why

It was reported that when having a multiple option add-ons, if one of those options was percentage-based, the app displays that percentage formatted as a currency.

Core | App
--- | ---
core-2 | before

# How

This PR fixes that issue by applying a percentage symbol, instead of the currency symbol, when the add-on has a `.percentageBased` price type.

# Screenshots

after

# Testing Steps

- Make sure to set a product with add-ons when one of them has percentage based options
- Launch the app and navigate to that product (make sure the exp feature is enabled)
- Tap on the add-on button
- See the percentage-based option rendered with a percentage symbol.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Edit Product/Addons/ProductAddOnViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Edit Product/Addons/ProductAddOnViewModel.swift
@@ -92,8 +92,19 @@ extension ProductAddOnViewModel {
                 return nil
             }
             return Option(name: option.label ?? "",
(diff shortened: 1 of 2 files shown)


shiki (approved):
It works! Negative percentages are also displayed correctly.

PR #4857, opened 2021-08-20
Shipping Labels: Custom package form validation (part 1)

rachelmcr:
Part of: #4743

## Description

This PR introduces validation for the custom package creation form, when creating a new custom package in the Shipping Label flow. Some details about the validation:

* All fields are required, and each should display a message about that if you enter invalid input / erase the input
* The name field must have content (besides whitespace/newlines)
* The dimensions fields (length/width/height) must have values greater than 0
* The weight field must have a value 0 or greater
* Once all fields have valid content, the Done button is enabled

I stopped at this point because I didn't want this PR to get too big, and I wanted to get feedback on the approach I used here. What isn't done yet:

* Right now, the only validation error message is "This field is required" (to match the web and Android clients). However, this doesn't account for invalid input. We'll eventually need to distinguish between empty fields and invalid input.
* It would be nice to sanitize the input, e.g. to not allow entering letters in the number fields.
* Tapping the Done button only navigates back; it doesn't yet create/use the custom package.

## Changes

I used a different approach here than our validation on other screens, because we're starting with invalid values in all the fields (since they're all empty). To avoid having the validation errors appear immediately, I decided to use SwiftUI's `onReceive` to trigger the validation for each field once when the field is edited. I relied heavily on [this suggested approach](https://newcombe.io/2020/03/05/validation-with-swiftui-combine-part-1/) for inspiration and am open to suggestions for doing this differently if there are other ideas for this.

* Introduces a set of reusable publishers in `Publisher+Validation.swift` to validate if a string has content, has a double value greater than a given number, or has a double value greater than or equal to a given number.
* In `ShippingLabelCustomPackageFormViewModel`:
   * A set of validation publishers are defined, based on the reusable publishers, related to each of the fields in the form.
   * A validation publisher `packageValidation` is also defined that combines all of the validators for the individual fields. This is used to determine if the Done button should be enabled.
   * These publishers are unit tested to confirm that the validation works as expected for the input in each field in the form.
* In `ShippingLabelCustomPackageForm`:
   * Each of the text fields uses `onReceive` to trigger the appropriate validation publisher and update a property (e.g. `isNameValidated`) that tracks whether the field has valid input.
   * There is a `ValidationErrorRow` below each text field that is displayed when the input for that field is validated.
   * Does the navigation bar setup, and on any change to the entire form validates the package as a whole. This is used to determine if the Done button should be enabled.

## Testing

8. Confirm the Add New Package screen appears and there are no validation errors yet and the Done button is disabled.
9. Start entering text in the fields, and confirm that if you remove the text you entered or enter invalid input, a validation error appears for that field.
10. Enter valid input for each field in the form, and confirm the Done button is enabled.

## Submitter Checklist

Update release notes:


itsmeichigo (changes requested):
Thank you for the work! I have tested and overall everything works as expected. However, I have some concerns:

I noticed on both the web app and Android that the Done button is enabled regardless of whether all fields are validated. I think this makes sense since we're not showing validation error initially - if the Done button is disabled in this case, it can be confusing to users. How about keeping the Done button enabled, and only handling the action when `validatedCustomPackage` is not nil?

Besides, I also found that the validation error rows are not extending edge-to-edge in landscape mode, and the some rows seem to be missing a `Divider` underneath:

Simulator Screen Shot - iPhone 12 Pro - 2021-08-23 at 09 56 32

I have some other comments in the code too 👇

WooCommerce/Classes/Extensions/Publisher+Validation.swift
```diff
@@ -0,0 +1,57 @@
...
+class ValidationPublishers {
+    /// Validates whether a string has content other than whitespace and newlines
+    ///
+    static func contentValidation(for publisher: Published<String>.Publisher) -> ValidationPublisher {
+        return publisher.map { value in
+            return value.trimmingCharacters(in: .whitespacesAndNewlines).isNotEmpty
+        }
+        .dropFirst()
+        .eraseToAnyPublisher()
+    }
```
itsmeichigo: I'm not sure I like the idea of hiding away `dropFirst()` in the encapsulation of `ValidationPublishers`. If there are cases where we want to reuse this, we'd need to make sure that we want the same behavior: to ignore the first element of the publisher. I think we can omit this implementation and replace it with something simpler, more information in the next comment.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageFormViewModel.swift
```diff
@@ -35,25 +37,92 @@ final class ShippingLabelCustomPackageFormViewModel: ObservableObject {
...
+        $packageWidth.greaterThan(0)
+    }()
+
+    lazy var packageHeightValidation: ValidationPublisher = {
+        $packageHeight.greaterThan(0)
+    }()
+
+    lazy var packageWeightValidation: ValidationPublisher = {
+        $emptyPackageWeight.greaterThanOrEqualTo(0)
+    }()
```
itsmeichigo: I think this implementation along with the use of `onChange` are more complicated than necessary. We can simplify the logic like so:
- For package name validation, we can make `isNameValidated` a  publisher. We keep the value `true` by default to not show any validation error initially.
```Swift
@Published private(set) var isNameValidated: Bool = true
```
- Then after intializing all stored properties, we can configure the value of this publisher by observing the value of package name starting from the 2nd value (using `dropFirst`):
```Swift
$packageName
        .dropFirst()
        .map { $0.trimmingCharacters(in: .whitespacesAndNewlines).isNotEmpty }
        .assign(to: &$isNameValidated)
```
This way we can omit both `packageNameValidation` and `onChange`, while keeping the same behavior - starting validation only from the 2nd value. We can apply the same thing to other fields as well.

What makes me prefer this solution: 
- It's simpler and more straightforward, and no logic is hidden in other encapsulation like in `ValidationPublishers`.
- Writing unit tests for `@Published` variables is so much easier since we can access their values directly without having to use `.sink` for observing.

WooCommerce/WooCommerceTests/ViewRelated/Shipping Label/Create Shipping Label/ShippingLabelCustomPackageFormViewModelTests.swift
```diff
@@ -18,4 +19,239 @@ class ShippingLabelCustomPackageFormViewModelTests: XCTestCase {
...
+        // Given
+        let viewModel = ShippingLabelCustomPackageFormViewModel()
+
+        // When
+        viewModel.packageName = "Test"
+
+        // Then
+        let _ = viewModel.packageNameValidation.sink { validated in
+            XCTAssertTrue(validated)
+        }
```
itsmeichigo: Although this works, the test can fail silently if `packageNameValidation` doesn't emit any event. Which means the block within `.sink` will not be triggered and there will be no assertion at all.

When testing with asynchronous code, it is recommended that we use `expectation` and wait for timeout to make sure that we catch the error if a publisher doesn't emit the expected event. I'd recommend this [nice article](https://www.swiftbysundell.com/articles/unit-testing-combine-based-swift-code/) as a guide for testing Combine code.

However, if we use `@Published` variable for validation instead, we can assert the value of the variable directly - and therefore there's no need for using `.sink` and `expectation`.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageFormViewModel.swift
rachelmcr: Thank you so much! This simpler solution makes a lot of sense to me. I was trying to come up with something like this but couldn't figure out how to make it work, and I think I was just missing the fact that I could use `.dropFirst()` directly when observing the value to get what I wanted (rather than using it in the abstracted publisher). I'll work on making the changes to use this approach. 👍

rachelmcr:
@itsmeichigo This is ready for another look! I implemented the simpler validation approach you recommended and fixed the UI issues you caught (those always seem to sneak in!).

I also updated the Done button so it's now always enabled. Tapping the "Done" button validates the form — if the input isn't valid `validatePackage()` makes sure the appropriate validation errors appear in the form, and if it's valid we handle the action.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageFormViewModel.swift
```diff
@@ -35,25 +37,114 @@ final class ShippingLabelCustomPackageFormViewModel: ObservableObject {
...
+            .assign(to: &$isWidthValidated)
+
+        $packageHeight
+            .dropFirst()
+            .map { self.validatePackageDimension($0) }
+            .assign(to: &$isHeightValidated)
+
+        $emptyPackageWeight
+            .dropFirst()
+            .map { self.validatePackageWeight($0) }
```
itsmeichigo: Since we mostly work with asynchronous code in Combine (most of the closures are `@escaping`), I'd recommend not capturing `self` strongly to avoid retain cycles. You can read more about that [here](https://trycombine.com/posts/self-weak-unowned/).

itsmeichigo (changes requested):
The UI looks great now! I've left a comment on self capturing in Combine closures, please check that out.

rachelmcr:
Thanks @itsmeichigo! I made a change in 18193a9 to capture `self` weakly in those closures, so this is ready for another look. That post was really helpful to clarify for me the various scenarios including when it's ok to capture `self` strongly in Combine. 👍

itsmeichigo (approved):
Everything looks great! Thanks for updating :shipit:

Issue #4856, opened 2021-08-20
[Mobile Payments] [Reader Updates] If a software update is available, cache that to reduce screen re-layout on re-entering


allendav:
Based on feedback from here: https://github.com/woocommerce/woocommerce-ios/pull/4844#issuecomment-902797826

On re-entering settings, if an update is available, it takes a few seconds to close the loop with Stripe and show the update notice and button. It's a little odd. We should cache unapplied update state somehow to avoid re-laying out the connected reader screen on re-entering the screen

cc @ctarda

allendav:
This can be closed now thanks to #5008

PR #4867, opened 2021-08-23
[Mobile Payments] [Reader Updates] Connect startCardReaderUpdate and UpdateViewController

allendav:
Partially addresses #4059 

Changes:
- Adds dispatching of startCardReaderUpdate to CardReaderSettingsConnectedViewModel
- Allows CardReaderSettingsConnectedViewController to request that get kicked off
- Adds new flags for the viewmodel to prompt CardReaderSettingsConnectedViewController to display the UpdateViewController or success snackbar
- Adds UpdateViewController

To test:
- **IMPORTANT: To avoid updating an actual card reader, first enable `-simulate-stripe-card-reader` for the scheme before running**
- Tap on settings (gear), then In-Person Payments, then Manage card reader
- Tap connect and wait for the reader to connect
- When notified of the update, tap on update
- Wait for the update to complete
- Verify a success notice is displayed
- Note that the "update available" is no longer present

Unfinished business for a subsequent PRs:
- [ ] Displaying a failure message
- [ ] Expand unit tests to cover new viewmodel additions

Screenshots:

iPad (10.2") Portrait Dark Updating

iPad (10.2") Portrait Dark Updated

iPad (10.2") Landscape Dark Updating

iPhone (12 Pro Max) Portrait Light Updating

iPhone (12 Pro Max) Portrait Light Updated

iPad Movie

https://user-images.githubusercontent.com/1595739/130701185-c833d077-b631-4fe0-8d34-2d80c44d61c7.MP4

Update release notes:


WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce.xcscheme
```diff
@@ -129,7 +129,7 @@
          </CommandLineArgument>
          <CommandLineArgument
             argument = "-simulate-stripe-card-reader"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
allendav: Will omit this change when I publish the PR (i.e. as not draft)

allendav: Removed in https://github.com/woocommerce/woocommerce-ios/pull/4867/commits/76fc831930b7646dbad863ba0a93ccea14efd7b2

allendav:
@ctarda @Garance91540 this is ready for review - two things to bear in mind

- please use `-simulate-stripe-card-reader` before running unless you really want to update your reader. Once updated, a reader cannot be updated again until the next update, and those only happen once or twice a year.
- the alpha seems too transparent to me? I matched the value in Figma, but I suspect it is also combining with the `overFullScreen` presentation style to come out lighter than it should - what do you think?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -127,6 +134,43 @@ private extension CardReaderSettingsConnectedViewController {
...
+                    return
+                }
+                guard let readerUpdateCompletedSuccessfully = self.viewModel?.readerUpdateCompletedSuccessfully else {
+                    return
+                }
+                if readerUpdateCompletedSuccessfully {
+                    self.displayReaderUpdateSuccessNotice()
+                }
+
+                self.updateViewController = nil
```
allendav: @ctarda - thoughts? I'd like to completely nuke the view controller when I'm done with it. Although this works, is it a good idea?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -127,6 +134,43 @@ private extension CardReaderSettingsConnectedViewController {
...
+            })
+            return
+        }
+
+        // Otherwise, instantiate and present an updateViewController
+        updateViewController = UpdateViewController()
+        guard updateViewController != nil else {
+            return
+        }
+        updateViewController!.loadViewIfNeeded()
```
allendav, line 166: @ctarda Curiously, my outlets fail to connect unless I do this. Did I create and connect the XIB to the VC incorrectly?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -127,6 +134,43 @@ private extension CardReaderSettingsConnectedViewController {
         tableView.reloadData()
     }
 
+    func configureUpdateView() {
+        guard viewModel != nil, let viewModelUpdateInProgress = viewModel?.readerUpdateInProgress else {
```
ctarda: I would suggest this:
`guard let viewModel = viewModel, let viewModelUpdateInProgress = viewModel.readerUpdateInProgress, updateInProgress != viewModelUpdateInProgress else `
What seems a bit odd to me is not having two different guard clauses, which is okay, but the negative check in the first guard, that does not read very familiar in Swift.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
ctarda: The [documentation about dismiss](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621505-dismiss) says:

> If you want to retain a reference to the view controller's presented view controller, get the value in the presentedViewController property before calling this method.

Which I interpret as "the reference to the view controller will become nil eventually". I don't think it's strictly necessary to nullify this reference, but it might not hurt do to so?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
ctarda: Yeah, that's not a good sign, and should not be necessary. Do you have an initializer that calls super.init(nibName:bundle:)?

ctarda (approved):
It works well in the simulator (I won't test on my only reader).

I left a couple of suggestions and hopefully answered your questions 😊 

:shipit:

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: Thanks! Simplified in https://github.com/woocommerce/woocommerce-ios/pull/4867/commits/a117c90bb6b68e167a4f2f15dcc7027b7049af7b

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: After careful consideration, it isn't needed. The situation I was trying to avoid is calling dismiss on an already dismissed VC and that's not possible, so I removed it and simplified the guarding in https://github.com/woocommerce/woocommerce-ios/pull/4867/commits/a18b730b73ecf868658176ab6a7a9e46a231c2a8

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: I've tried

```
updateViewController = UpdateViewController(nibName: "UpdateViewController", bundle: nil)
```

and I've also tried

```
    init() {
        super.init(nibName: "UpdateViewController", bundle: nil)
    }

```

but only `loadViewIfNeeded` sets up the outlets so far

allendav: Hmmm. I've tried six ways to Sunday to get this to work without `loadViewIfNeeded`

I'm going to proceed with this unless there are objections.

Garance91540:
Hi @allendav , for the next issues is it possible to add @woocommerce/mobile-designers as reviewers? 

Except the UI issues with the size of the modals during the connexion flow that I already reported, it's looking good! 

One thing: once we land the merchants on the "Manage Card Reader" screen, there's a 1-2 delay before we display the "Please update reader software...banner". Is it possible to make it appear instantly? If not I worry this screen will be skipped quickly. Thanks!

allendav:
Thanks @Garance91540 ! Will do!

> One thing: once we land the merchants on the "Manage Card Reader" screen, there's a 1-2 delay before we display the "Please update reader software...banner"

Yeah, we wrote https://github.com/woocommerce/woocommerce-ios/issues/4856 for that and that will also need careful consideration in light of how reader updates will be changing with the new Stripe SDK and new M2 reader: pdfdoF-4x-p2

Issue #4863, opened 2021-08-23
[What's New Component] Present Latest Features on App Launch and Settings


caiobzen:
# Description 📜

In order to show the latest features available on WooCommerce app, now that we've fetched and stored the latest announcements, next step is to build up the screen.

# Screenshot 📱
<img width="552" alt="image" src="https://user-images.githubusercontent.com/997521/130467303-a4f19440-85b4-486e-b3b1-1a7582681437.png">

# Expectation 🔍

1. When user opens the app, we must show the What's New Screen.
1. When user opens What's new on WooCommerce" under the app Settings, we must show the What's New Screen.

# Todo ✓ 
- [x] Create SwiftUI Views (Check existing ones on SwiftUI Components Folder)
- [x] iPad!
- [x] RTL!
- [x] Create Settings cell and prepare the navigation to the What's New Screen 
- [ ] Handle Dark Mode! 🌚
- [x] Create ViewController populated with ViewModel data
- [x] Handle Modal dismissal
- [x] Present What's New screen on App launch
- [x] Make sure we present What's New Component only once on app launch (toggle displayed property)

Issue #4862, opened 2021-08-23
[What's New Component] Fetch and Store Feature Announcements on App Launch


caiobzen:
# Description 📜

In order to show the latest features available on WooCommerce app, we must fetch this data from a public WordPress API and persist it to show this data later on under _Settings_.

# Screenshot 📱
<img width="552" alt="image" src="https://user-images.githubusercontent.com/997521/130467303-a4f19440-85b4-486e-b3b1-1a7582681437.png">

# Expectation 🔍

1. When user opens the app, if not fetched yet, we must fetch the latest announcements and persist it on a plist file.
2. For now, just by looking at the network traffic and seeing that announcements was fetched only once is ok to validate the caching mechanism

# Todo ✓ 
- [x] Create Store
- [x] Invoke Announcements fetching with Yosemite Action
- [x] Implement system to prevent fetching features of the same version
- [x] Persist Announcements in Store
- [x] Handle Errors!
- [x] Implement request to WordPress announcements
- [x] Invoke Store from app coordination start
- [x] Remove WordPressKit dependency from Yosemite

PR #4868, opened 2021-08-23
[What's New Component] Fetch and Store Feature Announcements on App Launch

caiobzen:
Part of: #4862

# Description 📜

This PR introduces the main mechanism of fetching new announcements to warn users about the newest released features of the current app version.

For now, when user opens the app, an action for fetching the announcements is being dispatched but doing nothing with the response (as this will be implemented on #4863).

More information about the endpoint that is being used can be found [here](https://fieldguide.automattic.com/mobile-feature-announcements/mobile-feature-announcement-endpoint/)

# Main Changes ⚙️
- Added a new _Store_ named `AnnouncementsStore`, responsible for fetching _Announcements_ and persisting its _Features_  on disk (.plist file). 
- Dispatch a new action (`AnnouncementsAction.synchronizeFeatures`) on app coordination start.
- Inject _AnnouncementServiceRemote_ into Yosemite (as we are going to user a pre-existing this WordPressKit service remote to fetch the latest announcements). For this, a new protocol to create the boundary was created with the name `AnnouncementsRemoteProtocol`.
- The Features are being mapped into _WooCommerceFeature_, a model that will be displayed on the Whats New Component screen.

# Screenshots 📱
image

# Testing 🧑‍🔬🧪

1. Launch the app
2. Go to Settings (top-right corner of the NavigationBar)
3. Tap on `Launch Wormholy Debug`
4. Filter for "announcements"
5. There should be a request for the announcements endpoint

### If you want to test it with real data, please do the following change:
image

⚠️ As of today, there are no versions with Announcements, so the caching mechanism cannot be tested (unless you use a network app to stub the network responses, like [MockingBird](https://github.com/Farfetch/mockingbird) or [Charles proxy](https://www.charlesproxy.com))

## Update release notes:


Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,114 @@
...
+import WordPressKit
+
+/// Protocol for `AnnouncementsRemote` mainly used for mocking.
+///
+public protocol AnnouncementsRemoteProtocol {
+
+    func getAnnouncements(appId: String,
+                          appVersion: String,
+                          locale: String,
+                          completion: @escaping (Result<[Announcement], Error>) -> Void)
```
caiobzen, line 11: I will probably stop using `Announcement` and `Feature` from WordPressKit in protocols, etc.
I might create a `WooCommerceFeature` or something like this to stop spreading the `import WordPressKit` everywhere!

caiobzen: In fact, I think I can even **remove** WordPresskit from **Yosemite** and use a protocol to create this boundary 👍

Storage/Storage/Model/WooCommerceFeature.swift
```diff
@@ -0,0 +1,12 @@
+public struct WooCommerceFeature: Codable, Identifiable {
```
Ecarrion: I think this should be just called `Feature`, why did you choose to prepend `Woocommerce` to it?

Storage/Storage/Model/WooCommerceFeature.swift
```diff
@@ -0,0 +1,12 @@
...
+    public var id = UUID()
```
Ecarrion: Why is this random `id` needed? is it for `SwiftUI`?

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
```diff
@@ -0,0 +1,13 @@
...
+/// AnnouncementsAction: Defines all of the Actions supported by the AnnouncementsStore.
+///
+public enum AnnouncementsAction: Action {
+
+    public typealias IsCached = Bool
+
+    /// Synchronizes the latest Announcements.
+    /// If information was already fetched, the list of features will be retrieved along with the boolean indicating that the data was already fetched
+    ///
+    case synchronizeFeatures(onCompletion: ([WooCommerceFeature], IsCached) -> Void)
```
Ecarrion: Some comments here:

1. Why synching only features and not the full announcement? I think we could use some announcement information(like the app version) to make sure we don't show the wrong data

2. Why `IsCached` is needed? I think `synchronizeFeatures` should always sync/request the latest announcement from the network.

3. We may also be missing an action to load them without downloading them. This can be in another PR tho.

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,108 @@
+import Networking
+import Storage
+
+/// Protocol for `AnnouncementsRemote` mainly used for mocking.
+///
+public protocol AnnouncementsRemoteProtocol {
+
+    func getFeatures(appId: String,
```
Ecarrion: Thanks for thinking about unit testing. 
I think we should make this protocol to include `getAnnouncements...` so it matches 1-1 what is trying to hide(AnnouncementsRemote)

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,108 @@
...
+                storageManager: StorageManagerType,
+                network: Network,
+                remote: AnnouncementsRemoteProtocol,
+                fileStorage: FileStorage) {
+        self.remote = remote
+        self.fileStorage = fileStorage
+        super.init(dispatcher: dispatcher, storageManager: storageManager, network: network)
+    }
+
+    private var appVersion: AppVersion { UserAgent.bundleShortVersion }
```
Ecarrion: Is there a specific reason to make this a computed variable?

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,108 @@
...
+
+    private lazy var featureAnnouncementsFileURL: URL! = {
+        let documents = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first
+        return documents!.appendingPathComponent(Constants.featureAnnouncementsFileName)
+    }()
```
Ecarrion: We try to avoid using implicitly unwrapped optional.
Do you think you can model these without them?

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,108 @@
...
+        guard let savedFeatures: [AppVersion: [WooCommerceFeature]] = try? fileStorage.data(for: featureAnnouncementsFileURL) else {
+            return []
+        }
+
+        return savedFeatures[appVersion] ?? []
+    }
+
+    /// Save the `WooCommerceFeature`(s) to the appropriate file.
+    func saveFeatures(_ features: [WooCommerceFeature]) throws {
+        try fileStorage.write([appVersion: features], to: featureAnnouncementsFileURL)
```
Ecarrion: I'm not sure that saving features for different app versions is what we want for this project.
To avoid having to deal with deletion and stale data, I think saving and replacing an `Announcement` type that includes features is more straightforward, and will only have the latest content, which is what we need.

What do you think?

Ecarrion (changes requested):
Congrats on submitting your first PR. 
I have added some comments so we can analyze if there is a simpler approach to take!

Storage/Storage/Model/WooCommerceFeature.swift
caiobzen: I named it feature not to conflict with WordPressKit Feature model, but since It is not part of the Yosemite framework as a dependency, I think I can revert this name back to Feature indeed!

Storage/Storage/Model/WooCommerceFeature.swift
caiobzen: Yes indeed (I may have gone too far on this PR). Please let me know if this breaking any convention 🙏

Storage/Storage/Model/WooCommerceFeature.swift
caiobzen: Done! The struct name is changed to `Feature` 👍

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
caiobzen: Good points!

1. The way I structured the storage is by taking the current app version and using it as a key for the features that belong to the Announcement.For example:

```swift
[7.3: [Feature(...), Feature(...) ...]]

```
 This way, we would be free from having to use WordPressKit models in the Store layer. If we would save all the data from the Announcements model, I'd suggest creating a version of it (or a protocol and make the WordpressKit.Announcement conform to it to keep the boundary between the app and the other app layers.
Another way to solve this and just persist whatever comes from the endpoint, is to save whenever the app was opened with a different app version 🤷. Any thoughts on that?

2. IsCached is a way to know if the data persisted when the app is opened was already fetched before. This will allow us to know when the What's New component must be presented or not.

3. For sure! I intend to create this action when developing the What's New Component. Is that ok?

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: There isn't! good catch! I've made it a `private let` instead 👍

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: You are totally right. I blindly followed the same pattern used in the app settings storage.
I removed the force unwrap from it 👍

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: I agree! I have answered it back in [this comment](https://github.com/woocommerce/woocommerce-ios/pull/4868#discussion_r696093343) by the way 👍

Storage/Storage/Model/WooCommerceFeature.swift
Ecarrion: Well, this struct resides on Storage and should be used to store information.
I don't think adding View dependencies is a good idea to keep things modular.

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
Ecarrion: > This way, we would be free from having to use WordPressKit models in the Store layer. If we would save all the data from the Announcements model, I'd suggest creating a version of it (or a protocol and make the WordpressKit.Announcement conform to it to keep the boundary between the app and the other app layers.

If you see other entities on the project, you can see how the model entities and storage models are almost identical but replicated.

Storing/Creating a new model for this layer almost has no costs but makes the boundaries clear and are fairly easy to maintain.

Feel free to create a smaller version of `Announcement` if you feel like you don't need all the data and we can expand it later.

> , is to save whenever the app was opened with a different app version 🤷. Any thoughts on that?

I don't think this is a good idea, we don't really know how the announcement will get updated. For example, it went live with wrong information and is corrected after the app has fetched, we should be able to update our internal storage and reflect the correct information.

> IsCached is a way to know if the data persisted when the app is opened was already fetched before. This will allow us to know when the What's New component must be presented or not.

I think this can bring some inconsistencies and undefined behavior. What would happen if the app crashes just before showing the view, the announcement will be marked as cached but it was never shown.

I think we should be explicit when marking an announcement as viewed.

EG:
```swift
struct Announcement {
    let appVersion: String
    let displayed: String
    let features: [Features]
}
```

What do you think?

Storage/Storage/Model/WooCommerceFeature.swift
caiobzen: I do agree! I will remove this and re-think how the modeling will happen when designing the screen 👍

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
caiobzen: Good idea! I will plan to do this changes in a few 👍

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: I've renamed this to getAnnouncement 👍

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
caiobzen: Thanks for all the ideas you've shared! I was able to do most of it but I'm not totally convinced about having to explicitly save the displayed information. 
Mostly because what I'm thinking in terms of flow is:

### Flow 1 (user without persisted announcement)
1. User opens the app
2. We Look at the persisted announcement version
3. If there is no announcement saved, do nothing
4. synchronize announcements
5. If there is an announcement saved and the AppVersion is the same, show What's New Screen


### Flow 2 (user with saved announcement)
1. User opens the app
2. We look at the persisted announcement version
3. If there is an announcement saved, and it has the same AppVersion, do nothing
4. synchronize announcements
5. If there is an announcement saved, and it has a different appVersion, show What's New Screen

Do you think we are still going to need the `displayed` property? If so, could you please help me to understand how to take the best usage out of this property? 🙏

Ecarrion: The problem with that is that you are assuming that an stored announcement is a displayed announcement and that could not necessarily be the case. We can easily avoid all kinds of edge cases by just setting a `displayed` flag when the view actually appears or when the close button is pressed. 

I think a more simple flow would be:

- User opens the app
- Sync announcements 
- If the announcement has not been displayed/dismissed, display it and mark the announcement as displayed/dismissed.

What do you think?

Storage/Storage/Model/Announcement.swift
```diff
@@ -0,0 +1,9 @@
+public struct Announcement: Codable {
```
Ecarrion, line 6: We normally add some documentation to what these models are, can you add that please?

Storage/Storage/Model/Feature.swift
```diff
@@ -0,0 +1,11 @@
+public struct Feature: Codable {
```
Ecarrion, line 4: Same as the above.

Storage/Storage/Model/Feature.swift
```diff
@@ -0,0 +1,11 @@
...
+    public let title: String
+    public let subtitle: String
+    public let iconUrl: String
```
Ecarrion, line 7: I think I remember that in your trial post, we were alerted to also support `base64` icon files, are you planning to add it later?

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
```diff
@@ -0,0 +1,46 @@
...
+    /// Fetch announcements from WordPressKit announcements public API
+    ///
+    /// - Parameter appId: the application identifier. 4 stands for WooCommerce
+    /// - Parameter appVersion: the application version that we are looking forbthe translations
+    /// - Parameter completion: A result that may contain an Announcement or an Error
+    ///
+    public func getAnnouncement(appId: String,
+                                appVersion: String,
+                                locale: String,
+                                completion: @escaping (Result<Storage.Announcement?, Error>) -> Void) {
```
Ecarrion: We shouldn't reach the storage layer directly here. 
Take a look at how we alias them [here](https://github.com/woocommerce/woocommerce-ios/blob/develop/Yosemite/Yosemite/Model/Model.swift)

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        getAnnouncements(appId: appId, appVersion: appVersion, locale: locale) { result in
+            switch result {
+            case .success(let announcements):
+                guard let firstAnnouncement = announcements.first else {
+                    completion(.success(nil))
+                    return
```
Ecarrion: nit: I think this can be made as a one-liner
```swift
return completion(.success(nil))
```

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
```diff
@@ -0,0 +1,46 @@
...
+                }
+                let announcement = self.mapAnnouncementToStorageModel(firstAnnouncement)
```
Ecarrion: I think we should guard against a deallocated self. don't you?

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
```diff
@@ -0,0 +1,46 @@
...
+                completion(.success(announcement))
+            case .failure(let error):
+                completion(.failure(error))
+            }
+        }
+    }
+
+    private func mapAnnouncementToStorageModel(_ announcement: WordPressKit.Announcement) -> Storage.Announcement {
+        let mappedFeatures = announcement.features.compactMap {
```
Ecarrion: Any reason to use `CompactMap` instead of a regular map?

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
```diff
@@ -0,0 +1,46 @@
...
+            Feature(title: $0.title, subtitle: $0.subtitle, iconUrl: $0.iconUrl)
+        }
+
+        return Storage.Announcement(appVersion: announcement.appVersionName, features: mappedFeatures)
```
Ecarrion: Related to the comment where these are aliased.

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
```diff
@@ -0,0 +1,10 @@
+import Storage
+
+/// AnnouncementsAction: Defines all of the Actions supported by the AnnouncementsStore.
+///
+public enum AnnouncementsAction: Action {
+
+    /// Synchronizes the latest Announcements and save it on disk
+    ///
+    case synchronizeAnnouncements(onCompletion: (Announcement?) -> Void)
```
Ecarrion: I think here, this should provide a `result` in the completion block to log any possible error on the network request.

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,115 @@
...
+        remote.getAnnouncement(appId: Constants.WooCommerceAppId,
+                               appVersion: appVersion,
+                               locale: Locale.current.identifier) { [weak self] result in
+            switch result {
+            case .success(let announcement):
+                guard let announcement = announcement else {
+                    onCompletion(nil)
+                    return
+                }
+                try? self?.saveAnnouncement(announcement)
```
Ecarrion: I think here we should also handle errors to properly log later if there was a problem saving the announcement.

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,115 @@
...
+
+    /// Load the latest saved `Announcement`. Returns nil if there is no saved announcement.
+    func loadSavedAnnouncement() -> Announcement? {
+        guard let fileURL = featureAnnouncementsFileURL,
+              let savedAnnouncement: Announcement = try? fileStorage.data(for: fileURL) else {
+            return nil
+        }
+
+        return savedAnnouncement
+    }
```
Ecarrion: It seems like you are not using this, right?

Ecarrion (changes requested):
Hi Carlos, thanks for working on this.

I have some other questions and requests, let me know what you think!

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
caiobzen: Yep! That sounds reasonable! I will edit the model to include the `displayed` property 👌

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: 🥇

Storage/Storage/Model/Announcement.swift
caiobzen: Sure! Just added some docs for this model!

Storage/Storage/Model/Feature.swift
caiobzen: Roger that! Docs were added 👍

Storage/Storage/Model/Feature.swift
caiobzen: Yep! I just added it 👍

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Totally! I missed that one! Thanks for spotting it!

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Not at all! There's nothing nil going on here to get rid of. Thanks!

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Yes. I intend to use it in the next PR where we are going to present the saved announcement 👍

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Done! I've used the same approach as the ones you've send! Thanks

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Done!

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Done!

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Done!

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Done! 😄

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
caiobzen: Sounds good. I have initially thought about silently failing this request since it's not a deal breaker, but it makes sense to at lease inform developers that something bad happened! 
I will do that in a few 👍

caiobzen: Done!

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Sounds good. I raised the error thrown in the saveAnnouncement! 👍

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
caiobzen: Done! ✅

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
```diff
@@ -0,0 +1,52 @@
...
+extension AnnouncementServiceRemote: AnnouncementsRemoteProtocol {
+
+    override convenience init() {
+        self.init(wordPressComRestApi: WordPressComRestApi(baseUrlString: Settings.wordpressApiBaseURL))
+    }
+
+    /// Fetch announcements from WordPressKit announcements public API
+    ///
+    /// - Parameter appId: the application identifier. 4 stands for WooCommerce
+    /// - Parameter appVersion: the application version that we are looking forbthe translations
```
Ecarrion:
```suggestion
    /// - Parameter appVersion: the application version that we are looking for the translations
```

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
```diff
@@ -0,0 +1,52 @@
+import WordPressKit
+import Storage
+import Networking
+import Yosemite
+
+/// Makes AnnouncementServiceRemote from WordPressKit conform with AnnouncementsRemoteProtocol so we can inject other remotes. (For testing purposes)
+extension AnnouncementServiceRemote: AnnouncementsRemoteProtocol {
```
Ecarrion: I have been thinking a bit about this extension and I wonder why is this on the `WooCommerce` target? as this is used on `Yosemite`, I think we should have this extension living there. Ideally, networking objects should not leak to our outer layer (WooCommerce)

Ecarrion: I just tried to remove the `WordpressKit` dependency and imports from `WooCommerce` and just leave it in `Yosemite` and everything seems to compile well!

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,123 @@
...
+
+/// Protocol for `AnnouncementsRemote` mainly used for mocking.
+///
+public protocol AnnouncementsRemoteProtocol {
+
+    func getAnnouncement(appId: String,
+                         appVersion: String,
+                         locale: String,
+                         completion: @escaping (Result<Announcement?, Error>) -> Void)
+}
```
Ecarrion: I have been thinking about this protocol and how the WooCommerce extension that implements it mixes some responsibilities like getting a networking entity and mapping it into a storage model. I think all of that should happen here like in all of our other stores.

If we move the `WordPressKit` dependency to `Yosemite` this piece of code can become

```swift
public protocol AnnouncementsRemoteProtocol {
    func getAnnouncements(appId: String, appVersion: String, locale: String, completion: @escaping (Result<[Announcement], Error>) -> Void)
}

extension AnnouncementServiceRemote: AnnouncementsRemoteProtocol { }
```

Then the store can
- Fetch the Networking/WordPressKit entity
- Map it into a storage object, see `ReadOnlyConvertible` protocol.
- Store the storage object.

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
Ecarrion: Next time, maybe it's better to not include code that is not being used in the PR, as it can distract the reviewer, AKA me 😛

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,123 @@
...
+                guard let announcement = announcement else {
+                    onCompletion(.success(nil))
+                    return
+                }
+                do {
+                    try self?.saveAnnouncement(announcement)
+                } catch {
+                    onCompletion(.failure(error))
+                }
+                onCompletion(.success(announcement))
```
Ecarrion: If the announcement fails to save, won't this call `onCompletion` twice?
I don't think that is expected or good, I think it should only return the error.

Ecarrion (changes requested):
Hey @caiobzen, I have some more thoughts and comments, let me know what you think!

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: I do agree! I will remove this for now 👍

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Thanks! I also managed to add the `locale` in the params list!

WooCommerce/Classes/Extensions/AnnouncementServiceRemote+Woo.swift
caiobzen: Cool! That was my main idea, to have everything inside Yosemite. But as I saw the WordPressKit dependency in the WooCommerce target, I thought it was better to leave it there and inject into Yosemite. But yea I do agree this is better to stay inside Yosemite!
I'll make this change in a few 😉

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Nice catch! I've moved this success completion to the `do try` block 👍

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Nice observation. The only thing I don't think that fits for StorageAnnouncement is the ReadOnlyConvertible protocol, as it is uses mostly with NSManagedObject models.
I've updated the Store to Fetch, Map and Store the Announcement Storage object 👍

Yosemite/YosemiteTests/Stores/AnnouncementsStoreTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+        // Assert
+        XCTAssertEqual(resultError, .unableToGetAnnouncement)
+    }
+}
+
+// MARK: - Mocks
+//
+private extension AnnouncementsStoreTests {
+    func makeWordPressAnnouncement() throws -> WordPressKit.Announcement {
```
caiobzen, line 106: @Ecarrion I haven't seen any Dictionary / JSON to Model in Yosemite, but with json files being loaded in the Network Layer. I'm not 100% convinced it is ok doing it like this. Please let me know if it's ok 👍

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
```diff
@@ -0,0 +1,8 @@
+/// AnnouncementsAction: Defines all of the Actions supported by the AnnouncementsStore.
+///
+public enum AnnouncementsAction: Action {
+
+    /// Synchronizes the latest Announcements and save it on disk
+    ///
+    case synchronizeAnnouncements(onCompletion: (Result<StorageAnnouncement, Error>) -> Void)
```
Ecarrion: I think we still have a small misunderstanding here.
We shouldn't leak the storage object or knowledge to the outer layers. If you see how other actions work in the project, we alias them like 

```swift
typealias Announcement = WordpressKit.Announcement
```

And then the only types we expose to those layers are those `Yosemite` types.

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,134 @@
...
+            Feature(title: $0.title,
+                    subtitle: $0.subtitle,
+                    iconUrl: $0.iconUrl,
+                    iconBase64: $0.iconBase64)
+        }
+
+        return StorageAnnouncement(appVersion: announcement.appVersionName,
+                                   features: mappedFeatures,
+                                   announcementVersion: announcement.announcementVersion,
+                                   displayed: false)
```
Ecarrion, line 102: This won't always be false right, what is your plan to keep this with the correct value towards the future?

Yosemite/YosemiteTests/Stores/AnnouncementsStoreTests.swift
Ecarrion: We normally have json files, but I think this is fine for now.

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -0,0 +1,134 @@
...
+    /// Get Announcements from Announcements API and persist this information on disk.
+    func synchronizeAnnouncements(onCompletion: @escaping (Result<StorageAnnouncement, Error>) -> Void) {
+
+        remote.getAnnouncements(appId: Constants.WooCommerceAppId,
+                                appVersion: appVersion,
+                                locale: Locale.current.identifier) { [weak self] result in
+            switch result {
+            case .success(let announcements):
+                guard let self = self, let announcement = announcements.first else {
+                    return onCompletion(.failure(AnnouncementsError.unableToGetAnnouncement))
```
Ecarrion: I think renaming this to `.announcementNotFound` would express the intention better, right?

Ecarrion (changes requested):
Hi @caiobzen, this is looking almost done, but I think we still have some misalignments regarding the different layers and frameworks of the projects.

Also, I created an announcement v `107.4` in order to test this PR, feel free to use that in your tests. Also let me know when you need updates to it or a new one.

Yosemite/Yosemite/Actions/AnnouncementsAction.swift
caiobzen: Yep! It makes total sense! I've just made this change 👍

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Good point. I didn't bother doing this logic yet (I intend to do this in the next PR) but I'm thinking of using the `announcementVersion`, which I'm assuming will change every time we create a new version of it and check if it has a new version when the sync happens.
Also, I'll update this value every time we open or dismiss the Whats New component screen.
Sounds good?

Ecarrion: Sure, feel free to do it in a different PR!

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Totally! Changed ✅

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
```diff
@@ -52,11 +52,20 @@ final class AppCoordinator {
                     }
                 }
                 self.isLoggedIn = isLoggedIn
+                self.showWhatsNewIfNeeded()
```
Ecarrion, line 55: Something good could be to add this call under a feature flag, so we avoid firing in production apps once it's merged.

You can do this in a separate PR tho.

Ecarrion (approved):
This looks good now @caiobzen 🎉 
Thanks for taking the time!

PR #4870, opened 2021-08-24
Shipping Labels: Custom package form validation (part 2)

rachelmcr:
Part of: #4743

## Description

This PR finishes the validation for the custom package creation form started in #4857, when creating a new custom package in the Shipping Label flow. As of this PR:

* There are two validation error messages that can be displayed: "This field is required" if the field is empty, or "Invalid value" if there is input but it's not valid.
* The numeric input (package dimensions and weight) is now sanitized, so only `Double` values can be entered.

## Changes

* Introduces `sanitizeNumericInput(_:)` to the view model. This is used with `didSet` on the package dimensions and weight properties to sanitize the input as those properties are changed.
* Introduces `getErrorMessage(for:)` to the `Localization` enum in `ShippingLabelCustomPackageForm`. It checks if the field input is empty and returns the appropriate error message. (To use this method, the reusable `validationErrorRow` property was removed and a `ValidationErrorRow` is added separately for each field, so the appropriate error message can be fetched.)

Missing input errors | Invalid input errors
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-08-24 at 21 05 28|Simulator Screen Shot - iPhone 12 Pro - 2021-08-24 at 21 05 14

## Testing

6. Open the package details section and tap on “Package Selected.”
8. On the Add New Package screen:
   * Enter invalid input in one or more fields (e.g. only spaces for the package name, 0 for the package dimensions, or a negative number for the package weight) and confirm the error "Invalid value" appears.
   * Remove the input from one or more fields and confirm the error "This field is required" appears.
   * Try entering non-numeric input in the package dimensions or weight fields and confirm it's sanitized as expected.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelCustomPackageForm.swift
@@ -65,12 +65,6 @@ struct ShippingLabelCustomPackageForm: View {
                                              text: $viewModel.packageLength,
                                              symbol: viewModel.lengthUnit,
                                              keyboardType: .decimalPad)
-                            .onReceive(Just(viewModel.packageLength), perform: { newValue in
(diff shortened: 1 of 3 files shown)


itsmeichigo (approved):
Code looks great and works as expected! 
I like your sanitization logic! Very simple and works great 💯

Issue #4776, opened 2021-08-11
Order Editing: Edit Customer Note UI


Ecarrion:
Discovery task with the following alternatives:

- Refactor and/or reuse `NewNoteViewController` UI
**OR**
- Build UI from scratch

<img width="341" alt="Screen Shot 2021-08-11 at 8 48 26 AM" src="https://user-images.githubusercontent.com/562080/129044024-c1ab2bf3-608c-46c0-9a73-c2c3bf38c884.png">

Considerations:
- [  ] Connect this screen to the edit button from   #4475

Ecarrion:
I have decided to not refactor `NewNoteViewController` as the UI and network actions are different. If we try to reuse this, basically we will just reuse a text view.

<img width="736" alt="Screen Shot 2021-08-19 at 9 10 34 AM" src="https://user-images.githubusercontent.com/562080/130083875-52f372c1-05e6-4fe1-9880-114344641ae2.png">

Issue #4777, opened 2021-08-11
Order Editing: Submit new customer provided note


Ecarrion:
After tapping the done button the app should:

- [x] Perform network request to update order
- [x] Show a loading indicator instead of the "Done" button
- [x] Hide/disable the cancel button
- [ ] Disable input while the order is being updated.
- [x] Show success/error notice after completion
- [x] Dismiss ViewContoller if successful

PR #4871, opened 2021-08-24
[Order Editing] Fill customer note UI

### LINKED ISSUE #4800: Order Editing: Fill order note UI
With #4776 completed, let's fill the UI with the order's customer-provided note component.

Note: 
- [x] Make sure to enable/disable the "Done" button when there are changes to content.

Ecarrion:
closes #4800 

# Why

As the edit customer provided note UI is now in place #4776, this PR makes sure to:
- Render the existing customer provider note
- Allow text editing of the note
- Enable/Disable the done button when the note is edited
- Provide mechanisms to show a loading spinner while the note is being updated remotely(network operation to be done in a separate PR)

# How

- Added `EditCustomerNoteViewModel` that handles:
  - Storing the original & new note content
  - Making sure the done & loading bar items are swapped when the view model state changes

- Update `EditCustomerNote` to render content based on the UI.

# Screenshot
https://user-images.githubusercontent.com/562080/130699873-2aad6acb-270c-4c48-a3bc-83ee673c4c3d.mov

# Testing Steps

- Navigate to an order
- Tap the edit customer note button
- If the order had a note, see it populated
- Edit the note and see the done button become enabled
- Tap on done and see a loading spinner for 3 seconds, then see the view get dismissed

# Todo
- Perform network operations
- Add error & success notices

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Customer Note/EditCustomerNoteViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Customer Note/EditCustomerNoteViewModel.swift
@@ -37,13 +37,14 @@ final class EditCustomerNoteViewModel: ObservableObject {
         bindNavigationTrailingItemPublisher()
     }
 
-    /// Update the note remotely and fire and try to dismiss the view.
+    /// Update the note remotely and invoke a completion block when finished
     ///
-    func updateNote() {
-        // TODO: Fire network request & dismiss the view
+    func updateNote(onFinish: @escaping () -> Void) {
+        // TODO: Fire network request
         performingNetworkRequest.send(true)
         DispatchQueue.main.asyncAfter(deadline: .now() + 3) { [weak self] in
             self?.performingNetworkRequest.send(false)
+            onFinish()
         }
     }
 }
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Notes Section/Customer Note/EditCustomerNoteViewModel.swift
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /// Update the note remotely and invoke a completion block when finished
+    ///
+    func updateNote(onFinish: @escaping () -> Void) {
+        // TODO: Fire network request
+        performingNetworkRequest.send(true)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 3) { [weak self] in
+            self?.performingNetworkRequest.send(false)
+            onFinish()
```
Ecarrion, line 47: I wasn't very happy to add a completion block and break the reactive chain but sending this `onFinish` signal via a `@Published` variable seemed like an over-complication 😅

ealeksandrov (approved):
Works well and I like the implementation, especially state management for right navbar button 👍 :shipit:

Couple of minor considerations:
1. Can we activate keyboard on a textfield upon presenting the screen? I remember it was impossible/hard in SwiftUI previously.
2. Maybe disable input after a click on "done" button? So user won't be able to mess with text while waiting for network.

Ecarrion:

> Can we activate keyboard on a textfield upon presenting the screen? I remember it was impossible/hard in SwiftUI previously.

I tried to do that but it was not possible without wrapping an `UITextView` which is not ideal. Apparently, there will be new things on iOS 15 that could help with it. Here is a card to remember https://github.com/woocommerce/woocommerce-ios/projects/39#card-67495443

> Maybe disable input after a click on "done" button? So user won't be able to mess with text while waiting for network.

Good point, will do it in the next PR 

- https://github.com/woocommerce/woocommerce-ios/issues/4777

PR #4875, opened 2021-08-25
[Mobile Payments] [Card Reader Updates] Add retryable error modal, connect to reader update flow

allendav:
Partially addresses #4059 

Changes:
- Adds a `CardPresentRetryableError` modal to `OrderDetailsPaymentAlerts`
- Adds it to `CardReaderSettingsConnectedViewController` for when update fails
- Updates the UpdateViewController guarding to be more inline with our style
- Adds Unit Tests
- Updates UpdateViewController initialization to not require loadViewIfNeeded props @ctarda 
- Increases alpha of UpdateViewController background slightly for better legibility

To test:
- **IMPORTANT Set `-simulate-stripe-card-reader` in your Run Scheme to use the simulator**
- Edit `StripeCardReaderService.swift` to immediately fail on update, e.g.:

```
let installFuture = Future<Bool, Error> { promise in
            // Just immediately fail
            promise(.failure(CardReaderServiceError.bluetoothDenied))
```

- Kick off an update. Note that it fails immediately and presents the modal:

IMG

- Tap try again and verify it tries again
- Tap cancel and verify it doesn't

What's coming in the next PR?
- Test with an actual reader
- Remove the feature flag

@Garance91540 Screenshots of opacity change:

Before

IMG_0115

After

IMG_0114

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentRetryableError.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentRetryableError.swift
@@ -1,0 +1,60 @@
+import UIKit
+
+/// Modal presented on error. Provides a retry action.
+final class CardPresentModalRetryableError: CardPresentPaymentsModalViewModel {
+    private let primaryAction: () -> Void
+
+    let textMode: PaymentsModalTextMode = .reducedBottomInfo
+    let actionsMode: PaymentsModalActionsMode = .twoAction
+
+    let topTitle: String = Localization.title
+
+    var topSubtitle: String?
+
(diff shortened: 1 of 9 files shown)


ctarda (approved):
LGTM but I am a bit surprised we don't have this alert already implemented. I would swear I did implement something like this months ago. 🤷🏼 

:shipit:

allendav:
> LGTM but I am a bit surprised we don't have this alert already implemented

You almost did. You implemented `CardPresent**Non**RetryableError`

Garance91540 (commented):
Definitely more readable with the latest changes (more opaque screen) !

PR #4879, opened 2021-08-26
Shipping Labels: Require phone number for destination address for international shipments

### LINKED ISSUE #4874: Shipping Labels: Require phone number for destination address if DHL is selected as carrier for international shipping
# Description
DHL requires a phone number for destination address to create a shipping label successfully. Currently without the phone number, the request to purchase fails.

This was an issue on Android: https://github.com/woocommerce/woocommerce-android/issues/4195 and on the web plugin too: https://github.com/Automattic/woocommerce-services/issues/2276.

# Suggested solution
We can follow Android's solution: check if the shipment is international and carrier ID contains "DHL".

Then user should be redirected back to update destination address if the phone number is missing.

Another approach: require phone number for destination address for all international shipments like how the web plugin did: https://github.com/Automattic/woocommerce-services/pull/2470.

itsmeichigo:
Closes #4874 

# Description
There's a known issue that DHL is currently requiring phone number on destination address for the shipping label purchase to succeed. This PR updates the logic in shipping label creation form to require phone number on destination address (regardless of carrier).

# Changes
## ShippingLabelsFormViewController
- Updated to force navigating user to Edit Address form when detecting international shipping.

## ShippingLabelsFormViewModel
- Updated `updateRowsForCustomsIfNeeded` to update row state of Ship To row when phone number is missing and needs updating.
- Trigger `updateRowsForCustomsIfNeeded` near the end of `handleOriginAddressValueChanges` and `handleDestinationAddressValueChanges` instead of in property observers of `originAddress` and `destinationAddress`. This is to make sure that all other row updates (e.g for Customs or Carrier rows) are done before the final update of the address rows.

## Unit tests updates

# Demo
destination-phone

# Testing
2. Create an order that requires international shipping (origin country different from destination country). Leave the phone number in the shipping & billing address empty.
3. Open app, navigate to Orders tab, select the new order and select Create Shipping Label.
4. Notice that phone number is required in Ship From. Enter 10-digit number and proceed.
5. Notice that phone number is required in Ship To. You should be able to proceed only after entering 10-digit number for it too.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -369,12 +369,12 @@ private extension ShippingLabelFormViewController {
             ServiceLocator.noticePresenter.enqueue(notice: notice)
             return
         }
-        let phoneNumberRequired = viewModel.customsFormRequired
+        let isPhoneNumberRequired = viewModel.customsFormRequired
         let shippingAddressVC = ShippingLabelAddressFormViewController(
             siteID: viewModel.siteID,
             type: type,
             address: address,
-            phoneNumberRequired: phoneNumberRequired,
+            phoneNumberRequired: isPhoneNumberRequired,
             validationError: validationError,
             countries: viewModel.filteredCountries(for: type),
             completion: { [weak self] (newShippingLabelAddress) in
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
@@ -369,7 +369,7 @@ private extension ShippingLabelFormViewController {
             ServiceLocator.noticePresenter.enqueue(notice: notice)
             return
         }
-        let phoneNumberRequired = type == .origin && viewModel.customsFormRequired
+        let phoneNumberRequired = viewModel.customsFormRequired
         let shippingAddressVC = ShippingLabelAddressFormViewController(
             siteID: viewModel.siteID,
             type: type,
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -361,7 +369,7 @@ private extension ShippingLabelFormViewController {
             ServiceLocator.noticePresenter.enqueue(notice: notice)
             return
         }
-        let phoneNumberRequired = type == .origin && viewModel.customsFormRequired
+        let phoneNumberRequired = viewModel.customsFormRequired
```
pmusolino: minor nit (feel free to left as is if it makes more sense to you without the `is`): 
```suggestion
        let isPhoneNumberRequired = viewModel.customsFormRequired
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -465,11 +461,11 @@ private extension ShippingLabelFormViewModel {
...
 
-        guard let originAddress = originAddress else {
-            return
-        }
         // Require user to update phone address if customs form is required
-        if customsFormRequired && originAddress.phone.isEmpty {
+        if customsFormRequired && destinationAddress?.phone.isEmpty == true {
+            updateRowState(type: .shipTo, dataState: .pending, displayMode: .editable)
+        }
+        if customsFormRequired && originAddress?.phone.isEmpty == true {
```
pmusolino: Is there a reason why the destination address is updated first and then the source address? I would reverse the order, following the logical order and the UI ;) unless there is a more important reason to maintain this order (in that case, a comment would be useful).

pmusolino (approved):
LGTM 👍 

I just left some nits ;)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
itsmeichigo: The order does not matter 😄 I have updated it in 186647b24b3bcf66371207d97d7ec906f4a2981d to avoid confusion.

PR #4880, opened 2021-08-26
[Order Editing] Update customer provided note remotely

Ecarrion:
part of #4477

# Why

After filling the UI with the real information #4800, now it's time to send the network request to update the customer note remotely.

# How

- Added a new `updateOrder` action that will receive from the consumer which fields to update. 
This is in order to not send a huge payload and prevent extra work on the server. 
This also forced me to rename the existing `updateOrder` action to `updateOrderStatus` because it only updates the status. 

- Added a new method on `OrderRemote` to fire the network request to update the order by listening to the provided fields.
Here I've made the Order's `CodingKeys` internal to not duplicate the fields key value.

- Updated `ViewModel` to modify the order with the new note content and dispatch the correct action

# Demo
https://user-images.githubusercontent.com/562080/130980177-3df6238c-847e-46d0-b068-e64391369095.mov

# Testing Steps

- Navigate to an order
- Tap on the edit icon on the customer provided note
- Modify the content and tap done
- Wait for the network operation to complete and see that the customer note got updated

# TODO:

- Success and error notices will be done in the next PR as I have to make sure they work well with `SwiftUI`.

- Work for disabling the user input while the network operation is ongoing will be done in a next PR.

- There is a separate action to update the order status, I will be working on unifying that into the new `updateOrder` action in a separate PR. https://github.com/woocommerce/woocommerce-ios/projects/39#card-67530782

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/OrdersRemote.swift
+++ b/Networking/Networking/Remote/OrdersRemote.swift
@@ -125,6 +125,23 @@ public class OrdersRemote: Remote {
         enqueue(request, mapper: mapper, completion: completion)
     }
 
+    /// Updates the specified fields of a given order.
+    ///
+    /// - Parameters:
+    ///     - siteID: Site which hosts the Order.
+    ///     - order: Order to be updated.
+    ///     - fields: Fields from the order to be updated.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func updateOrder(from siteID: Int64, order: Order, fields: [String], completion: @escaping (Result<Order, Error>) -> Void) {
+        let path = "\(Constants.ordersPath)/\(order.orderID)"
+        let parameters: [String: AnyHashable] = [:]
+        let mapper = OrderMapper(siteID: siteID)
+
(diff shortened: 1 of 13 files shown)


Networking/Networking/Remote/OrdersRemote.swift
```diff
@@ -125,6 +125,30 @@ public class OrdersRemote: Remote {
...
+    ///     - fields: Fields from the order to be updated.
+    ///     - completion: Closure to be executed upon completion.
+    ///
+    public func updateOrder(from siteID: Int64, order: Order, fields: [UpdateOrderField], completion: @escaping (Result<Order, Error>) -> Void) {
+        let path = "\(Constants.ordersPath)/\(order.orderID)"
+        let mapper = OrderMapper(siteID: siteID)
+        let parameters: [String: AnyHashable] = {
+            fields.reduce(into: [:]) { params, field in
+                switch field {
+                case.customerNote:
```
ealeksandrov: nit: missing a space

```suggestion
                case .customerNote:
```

ealeksandrov (approved):
LGTM - note editing works 🚀

PR #4885, opened 2021-08-26
[Mobile Payments] [Several Readers Found] Switch to Bluetooth scan, handle multiple found readers

### LINKED ISSUE #4622: [Mobile Payments] Switch from proximity (nearest reader only) to scan (any reader)
Right now iOS will only discover the closest reader to the mobile device, aka bluetoothProximity. Android doesn't have this option in the Stripe Terminal SDK. They can only use bluetoothScan.

To simplify support, we should consider switching iOS to use bluetoothScan as well.

See https://stripe.com/docs/terminal/readers/connecting/bbpos-chipper2xbt

allendav:
Partially addresses #4333
Closes #4622 

Changes:
- Changes `StripeCardReaderService` to use scan (multiple reader support) instead of proximity
- Updates `CardReaderConnectionController` to handle multiple found readers and also properly handle "Continue Searching" UI which didn't really make sense for proximity
- Tweaks `CardPresentPaymentsModalViewController` to not cut off the reader name when present in the modal's headline by allowing the font to be downscaled further

To test:
- Power up two readers
- Enter Settings > In Person Payments > Manage card reader
- If either of your readers is a known reader, it will just automatically connect. This is expected. If there are multiple known readers found, the first known reader found will automatically connect.
- Go through the process to disconnect from any known reader and start search over again
- Ensure you are prompted whether you'd like to connect to each of your readers in turn, i.e.:

IMG_0119
IMG_0120

Discussion:
- This PR does **not** test the Several Readers Found feature flag. There are two reasons for this: 1) `StripeCardReaderService` doesn't know about ServiceLocator and I'm not sure I want to go through the trouble of passing in that flag (I don't think Hardware layer should know about WooCommerce layer ServiceLocator) 2) this actually works just fine with just one reader, not unlike how it used to work, so there is no change from the users perspective - thoughts, @ctarda ?

What comes next:
- Next, instead of these successive dialogs (one-per-reader, one-after-the-other) we'll be implementing a single dialog listing the found readers from which the user can pick which reader to connect to. That will start in the next PR

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -80,7 +80,7 @@ extension StripeCardReaderService: CardReaderService {
         Terminal.shared.logLevel = terminalLogLevel
 
         let config = DiscoveryConfiguration(
-            discoveryMethod: .bluetoothProximity,
+            discoveryMethod: .bluetoothScan,
             simulated: shouldUseSimulatedCardReader
         )
 
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.xib
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.xib
@@ -42,7 +42,7 @@
                                 <stackView opaque="NO" contentMode="scaleToFill" axis="vertical" spacing="8" translatesAutoresizingMaskIntoConstraints="NO" id="Q0r-qo-iZS">
                                     <rect key="frame" x="135.5" y="0.0" width="41.5" height="49"/>
                                     <subviews>
-                                        <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" text="Label" textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="oDv-zz-CTp">
+                                        <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" text="Label" textAlignment="center" lineBreakMode="tailTruncation" numberOfLines="2" baselineAdjustment="alignBaselines" minimumFontSize="5" translatesAutoresizingMaskIntoConstraints="NO" id="oDv-zz-CTp">
                                             <rect key="frame" x="0.0" y="0.0" width="41.5" height="20.5"/>
                                             <fontDescription key="fontDescription" type="system" pointSize="17"/>
                                             <nil key="textColor"/>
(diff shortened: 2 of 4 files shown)


ctarda (approved):
I agree with your assessment about the card reader service not needing to know certain things.

However, this seems to break when testing with the simulator.

The first time I run the connection flow on the simulator, the app connected automatically to a known reader, as expected. Then I disconnected from the reader, triggered the connection again, and I got the following error:

```
[StripeTerminal] ⚠️  An integration error was detected in your app. This device type is not supported.CoreSimulator 757.5 - Device: iPhone 12 (178F766A-AFB5-4232-A42C-8431409D8F8C) - Runtime: iOS 14.5 (18E182) - DeviceType: iPhone 12
```

The actual reader the app is trying to connect to is: 
```
<SCPReader: 0x600003bfccb0; deviceSoftwareVersion = (null); deviceType = bbpos_wisepad3; label = (null); serialNumber = WPC32SIMULATOR1; stripeId = (null); hardwareInfo = (null); batteryLevel = (null)>
```

Notice the deviceType. Surprise! It's our old friend the wisepad.

This is probably only surfacing an existing flaw in our connection logic, but we should filter out readers that are not supported. I would do that in CardPresentPaymentStore, as deciding which reader types are supported by the app seems like business logic to me 😉 

I am giving a :shipit: but I really think we should fix this before merging, and not defer it to a separate PR, otherwise we would break testing in the simulator for everyone.

On the bright side, I tested on device with my only reader, a Chipper, and everything worked fine.

allendav:
I agree. I'll try to repro this ASAP and figure out what is happening and not merge this yet. Thanks!

allendav:
> The first time I run the connection flow on the simulator, the app connected automatically to a known reader, as expected. Then I disconnected from the reader, triggered the connection again, and I got the following error

I get it too. I had to connect and disconnect a few times to get it though.

> we should filter out readers that are not supported. I would do that in CardPresentPaymentStore, as deciding which reader types are supported by the app seems like business logic to me

Agreed. Will add a commit to filter down to supported readers.

allendav:
This was neat. There are at least three simulated readers being discovered. Only the CHB20* avoid the error.  They have very similar names:

`CHB20SIMULATOR1`
- readerType: Hardware.CardReaderType.mobile
- vendorIdentifier: nil
- This reader can be successfully connected to

`CHB20SIMULATOR2`
- readerType: Hardware.CardReaderType.mobile
- vendorIdentifier: nil
- This reader can be successfully connected to

`WPC32SIMULATOR1`
- readerType: Hardware.CardReaderType.notSupported
- vendorIdentifier: nil
- Connecting to this reader fails with `[StripeTerminal] ⚠️  An integration error was detected in your app. This device type is not supported.`

I think we can safely filter on readerType as a result

allendav:
@ctarda - in https://github.com/woocommerce/woocommerce-ios/pull/4885/commits/74edbfc00ce1a1a38293ff728f337decae969c71 I filtered out unsupported readers (like the `WPC32SIMULATOR1`) and I also fixed a flaw in my design that was causing the selected reader to not always be the reader connected to - would you mind taking a second look? I've tested it successfully with simulated and real readers

ctarda:
@allendav  I just left a question in the new commit, but you make the call. I maintain my :shipit:

allendav:
> This would allow the verifoneP400 if I understand correctly?

Kinda? The P400 needs to be registered before it can be discovered, but I don't know if that applies to simulations. I've asked our Stripe friends.

> Would you consider filtering by .readerType == .chipper2x? That would also let us add || .m2 whenever we ad support for the new reader

As you later noted, we actually map .chipper2x to .mobile, and since we haven't yet seen .counterTop (the P400), let's keep filtering out .notSupported instead

PR #4886, opened 2021-08-27
Shipping Labels: Fix error parsing purchased labels due to null tracking number

itsmeichigo:
Closes #4877 

# Description
This PR fixes issue fetching shipping labels on Order Details screen due to failure to parse the label list when one of the items is a failed label.

The solution is to not force shipping labels to have a non-nil tracking number; and to not return failed purchases in the list of labels.

# Changes
- Updated decoder of `ShippingLabel` to return `nil` tracking number as an empty string instead of throwing parsing error.
- Updated `OrderShippingLabelListMapper` to filter only the labels with non-empty tracking number and status `.purchased`.
- Updated test JSON response for the label list to test parsing of the label list.

# Testing
With the updated JSON response, unit tests should still pass for `OrderShippingLabelListMapper`.
This can also be tested by purchasing an international shipping label with DHL, however it's not 100% reproducible to get failed purchases for the edge case. These are the steps to test:

1. Make sure that your test store has installed WCShip plugin and configured packages including DHL packages in WPAdmin > WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Create an international order (origin and destination countries are different).
3. Navigate to Orders tab on iOS app, select the new order and select Create Shipping Label.
4. Configure Ship From, Ship To, Package Details and Customs. Please make sure to enter phone number for both origin and destination addresses.
5. In Carriers and Rates, select DHL as the carrier.
6. Proceed to purchase the label.
7. After the purchase succeeds, dismiss the print view and get back to the Order Details screen.
8. Notice that Create Shipping Label button is no longer present, and instead you should see Shipping Label section with the label that was just purchased.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4887, opened 2021-08-27
Shipping Labels: Print customs forms

### LINKED ISSUE #4597: Commercial invoice (customs form) printing in Shipping Labels

itsmeichigo:
Closes #4597

# Description
This PR creates new view for printing customs form, and redirect to this view from order details and shipping label printing view if the label has a separate customs invoice.

# Changes
- Add new view `PrintCustomsFormsView`. This view was set up to accept more than one invoice to support multi-package later.
- Updated Order Details screen to show option to print customs form from a shipping label section if the label has `commercialInvoiceURL`.
- Updated `PrintShippingLabelCoordinator` to show the print view for customs form in the print flow if the current label has an invoice.

# Demo
| From Order Details | From Print flow |
| ----- | ----- |
| print-customs-1 | print-customs-2 |

# Testing
2. Create an international order (origin and destination countries are different).
3. Navigate to Orders tab on iOS app, select the new order and select Create Shipping Label.
4. Configure Ship From, Ship To, Package Details and Customs. Please make sure to enter phone number for both origin and destination addresses.
5. In Carriers and Rates, select DHL as the carrier.
6. Proceed to purchase the label.
7. After the purchase succeeds, select a paper size and select Print Shipping Label.
8. Close the print view, notice that customs forms printing view is displayed. Tapping on Print Customs Form should navigate to a print view, tapping on Save for Later should dismiss the view.
9. Head back to Order Details screen, select the overflow menu in the Shipping Label section. Notice that there's an option for Print Customs Form in the presented action sheet.
10. Select the option, notice that the same customs form printing view is displayed.

Update release notes:


pmusolino (changes requested):
I tested it, but I found some issues.

1) I found a small bug about the image of the FedEx courier. I created an issue here https://github.com/woocommerce/woocommerce-ios/issues/4888 since it's not related to this PR.
2) Tapping on Print Customs Form navigate to the print view, but during the loading, the screen is not attached to the borders of the view, but to the safe area.

3) I wasn't able to test step n°9. `Head back to Order Details screen, select the overflow menu in the Shipping Label section. Notice that there's an option for Print Customs Form in the presented action sheet.`. Basically, I'm not able to see the shipping label purchased. 🤔

itsmeichigo:
Thanks @pmusolino for spotting the issues! I have fixed the safe area issue with the progress view in 12d05eb0a5b23d48434ab60cacea1c2c237996ba.

As for step 9, I'm fairly positive this was due to the failure in parsing purchased labels, which I fixed in #4886. I have merged the PR and merged develop back into this branch so can you try again with the order that you could not find the purchased labels?

itsmeichigo:
**Update** - I've pushed quite a few changes after testing with Android app:
- Save for Later button is only present in print flow and not reprint flow.
- For print flow: navigation stack is updated to remove Shipping Label print view before pushing Customs Invoice print view. This is to make sure that when tapping back or Save for Later button, user is redirected to Order Details screen immediately. Internal reference: p1622636091163700/1622543318.153700-slack-CGPNUU63E.
- Only show Customs Invoice printing in print flow after completing printing label. Android is showing the screen regardless since they cannot detect if printing is completed; but since we can get this from the completion block of `UIPrintInteractionController`, we should use it. Internal reference: p1622636072163500/1622543318.153700-slack-CGPNUU63E.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Customs Form/PrintCustomsFormsView.swift
```diff
@@ -0,0 +1,165 @@
+import SwiftUI
+
+struct PrintCustomsFormsView: View {
+    let invoiceURLs: [String]
```
pmusolino: nit 
```suggestion
    private let invoiceURLs: [String]
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Customs Form/PrintCustomsFormsView.swift
```diff
@@ -0,0 +1,165 @@
...
+    let showsSaveForLater: Bool
```
pmusolino: nit 
```suggestion
    private let showsSaveForLater: Bool
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Customs Form/PrintCustomsFormsView.swift
```diff
@@ -0,0 +1,165 @@
...
+
+    private var saveForLaterButton: some View {
+        Button(action: {
+            presentationMode.wrappedValue.dismiss()
+        }, label: {
+            Text(Localization.saveForLaterButton)
+        })
+        .buttonStyle(SecondaryButtonStyle())
+        .renderedIf(showsSaveForLater)
+    }
```
pmusolino, line 95: I like this approach to separate small views into private sub-components. The view becomes more readable, and with less nesting. 💯

pmusolino (approved):
Thank you @itsmeichigo for fixing all the things, now it works perfectly :shipit:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Customs Form/PrintCustomsFormsView.swift
itsmeichigo: Actually I extract this view into a separate component since it's used in 2 places and I don't want to repeat myself 😅. But yeah, I'm trying to keep the balance between keeping `body` flat while avoiding scattering too many small components since it can be hard to follow.

PR #4407, opened 2021-06-11
Errors: Banner message for error loading data on Orders and Reviews lists

rachelmcr:
Part of: #4382 
Relies on: #4400

## Description

To improve our error states, we are introducing a top banner that can be used whenever there is a problem loading data. The banner is collapsed by default and on select, users can read our troubleshooting tips or contact support directly.

This PR adds the banner to the Orders and Reviews lists. It also adds the banner to the Empty State screen, so it can be shown/hidden as needed when that screen is used.

## Changes

* Adds the error banner to `EmptyStateViewController`. The banner appears above the stack view containing the empty state content (title, image, message, and button).
* Updates `EmptyStateViewController.xib` to add the `stackView` outlet (for the banner's layout constraints). Other changes to that file were made automatically when the outlet was added.
* Adds the error banner to `OrderListViewController`. Any sync errors will set `hasErrorLoadingData` to `true`, which determines whether the error banner is displayed on the order list or the order empty state screen.
* Adds the error banner to `ReviewsViewController`. Any sync errors will set `hasErrorLoadingData` to `true`, which determines whether the error banner is displayed on the order list or the order empty state screen.

/|Collapsed (empty state)|Expanded (empty state)|Expanded (cached data)
-|-|-|-
Orders (Processing)|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 01 24|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 01 51|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 08 26
Orders (All Orders)|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 02 50|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 03 02|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 08 29
Reviews|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 01 34|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 01 39|Simulator Screen Shot - iPhone 11 Pro - 2021-06-11 at 22 26 31

## Testing

1. Build and run a fresh install of the app (to begin without any cached data).
2. Log in to the app.
3. Disconnect your internet connection. (This is so you will trigger a sync error; you can also disrupt your store's Jetpack connection for a more realistic scenario.)
4. Select Orders to open the product list.
[...]

PR #4890, opened 2021-08-27
Fix Vertical Alignment of Empty State

### LINKED ISSUE #2252: Orders Empty States are not vertically aligned
## To Reproduce

1. Use the #2245 branch
2. Navigate to Orders → Search
3. Make sure the keyboard is shown if you're using the simulator. 
4. Pick a filter that is empty.

This seems to be related to the `KeyboardFrameObserver` usage in `EmptyStateViewController`. 

---

More reproducible steps from Garance at p5T066-2p9-p2#comment-8783: 

1. Navigate to Products 
2. Tap on plus button. 
3. Pick Variable product. 
4. Enter a text in the title. Make sure the keyboard is present if you're using the simulator.
5. Abort everything and go back to the Products list. 
6. Tap on plus button again. 
7. Pick Variable product.
4. Tap on Add variations row. This is the result:
    
 
## Expectations

The content should be vertically centered.

shiki:
Fixes #2252. The incorrectly aligned empty state content has been reported almost on every release. The most recent is p5T066-2yX-p2#comment-9591. I started fixing this [more than a year ago](https://github.com/woocommerce/woocommerce-ios/issues/2252#issuecomment-625330625) and didn't get around to it until now. I'm glad @designsimply never stopped reporting these. 😅 

## UI Changes

Before | After 
--------|-------
        |       
  | 
 | 

Large Font Size Scrolling (Before) | After 
--------|-------
|

## Trials and Tribulations

There were two problems here. 

### Non-Zero Keyboard Frame

First, `EmptyStateViewController` [relies on `KeyboardFrameObserver`](https://github.com/woocommerce/woocommerce-ios/blob/0d869cadc0016194e459dd366655a6173edc03f3/WooCommerce/Classes/ViewRelated/ReusableViews/EmptyStateViewController/EmptyStateViewController.swift#L90-L93) to correctly align the content. However, `KeyboardFrameObserver` [can sometimes (or often) return a non-zero `frame`](https://github.com/woocommerce/woocommerce-ios/blob/0d869cadc0016194e459dd366655a6173edc03f3/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardFrameObserver.swift#L50-L52) even if the keyboard is not visible. This causes the [layout calculations in `EmptyStateViewController` ](https://github.com/woocommerce/woocommerce-ios/blob/0d869cadc0016194e459dd366655a6173edc03f3/WooCommerce/Classes/ViewRelated/ReusableViews/EmptyStateViewController/EmptyStateViewController.swift#L208-L229) to incorrectly assume that there is a keyboard present even though there's not. That's why you'll see that the content is aligned at the top. The calculation expected a keyboard to be at the bottom. 🤦 

I fixed that by adding a check in `KeyboardFrameObserver` to always make sure that the `frame` is zero if the keyboard is not supposed to be visible (f94eab89bd728f391cf7fdd0bcca1fa155658cb0).

### Delayed Keyboard Visibility State

I expected the above to be enough to fix the problems. Of course, I have forgotten that life hates me. 🍋 😄 When navigating from Orders Search to an empty filter page, the `KeyboardFrameObserver` still [assumed that the keyboard was present](https://user-images.githubusercontent.com/198826/131189312-34b8ed2f-82a3-4aef-81a0-c58d19e10b0a.png). (╯°□°)╯︵ ┻━┻

What I found was that [`currentState.isVisible` was returning `true`](https://github.com/woocommerce/woocommerce-ios/blob/issue/2252-fix-empty-state-alignment/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardFrameObserver.swift#L54) even though the keyboard was being dismissed by the time the empty state was shown. And the culprit was `KeyboardStateProvider`, the object that we rely on to give us the keyboard state. It has a delay when [updating the `KeyboardState`](https://github.com/woocommerce/woocommerce-ios/blob/0d869cadc0016194e459dd366655a6173edc03f3/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift#L55-L60). The layout finishes first before `KeyboardStateProvider` gets a chance to change the `isVisible` state to `false`. 

The reason for the delay was `KeyboardFrameObserver` observes [`willShow` and `willHide` notifications](https://github.com/woocommerce/woocommerce-ios/blob/0d869cadc0016194e459dd366655a6173edc03f3/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardFrameObserver.swift#L40-L48) while `KeyboardStateProvider` observes [`didShow` and `didHide` notifications](https://github.com/woocommerce/woocommerce-ios/blob/0d869cadc0016194e459dd366655a6173edc03f3/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift#L46-L52). Naturally, `KeyboardFrameObserver`'s notification handlers are executed first with incorrect `isVisible` values. 🤦 

I fixed this by changing `KeyboardStateProvider` to use the `willShow` and `willHide` notifications (5ce160fc7d7e0863eb703301490be370bac38fbd). Since `KeyboardStateProvider` [is initialized in the `ServiceLocator`](https://github.com/woocommerce/woocommerce-ios/blob/0d869cadc0016194e459dd366655a6173edc03f3/WooCommerce/Classes/ServiceLocator/ServiceLocator.swift#L162-L167), we are guaranteed that it will update the `isVisible` value first before `KeyboardFrameObserver` instances.

All of these were obviously my mistakes. 😆 

## Known Problems

There's an issue with how the error banners (https://github.com/woocommerce/woocommerce-ios/pull/4407) interact with the rest of the empty state content. For example, the content does not automatically resize or expand if the banner content is expanded. 

Default (Looks Fine) | Expanded ([Sad Trombone](https://www.youtube.com/watch?v=yJxCdh1Ps48))
--------|-------
 | 

The fix for this can be done separately. I have no idea where to start. ¯\\\_(ツ)\_/¯

## Testing

Please consider running these steps in `develop` first so you'll get a feel of what's being fixed. 🙂 

1. Navigate to Orders → Search. If you're using the simulator, make sure that the keyboard is visible (CMD+K).
2. Tap on an Order Status that has zero orders. 
3. Confirm that the “We're sorry, we couldn't find any .... orders” content is vertically centered. 
4. Navigate to Products. 
5. Tap + to add a product. Pick the Variable product option.
6. Tap on the Add Variations cell. 
7. Confirm that the “Create your first variation...” content is vertically centered.

### Regression Tests

If you can, please do a regression test of the banners by following the steps in #4407.  

## Reviewing 

## Author Checklist

- [x] Create an issue for the error banner layout problem: https://github.com/woocommerce/woocommerce-ios/issues/4900

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
+++ b/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
@@ -8,7 +8,17 @@ struct KeyboardState: Equatable {
     let isVisible: Bool
     /// The frame of the keyboard when it is fully shown or hidden.
     ///
-    /// The value is usually from `UIResponder.keyboardFrameEndUserInfoKey`.
+    /// The value is from `UIResponder.keyboardFrameEndUserInfoKey`.
+    ///
+    /// Note that even if `isVisible` is `false`, this can still have a **non-zero** value. This
+    /// can happen in this scenario:
+    ///
+    /// 1. View-A is shown and the keyboard is shown.
+    /// 2. User taps on something which presents View-B **while** the keyboard is present.
+    ///    View-B does not have a user responder (text field) so the keyboard is not visible.
+    /// 3. NSNotificationCenter emits a `keyboardDidHideNotification` but with a
+    ///    `keyboardFrameEndUserInfoKey` value set to the **previously shown keyboard's frame**.
+    ///
     let frameEnd: CGRect
 }
 
--- a/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
+++ b/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
@@ -6,6 +6,7 @@ import UIKit
 struct KeyboardState: Equatable {
     /// True if the keyboard is visible.
     let isVisible: Bool
+
     /// The frame of the keyboard when it is fully shown or hidden.
     ///
     /// The value is from `UIResponder.keyboardFrameEndUserInfoKey`.
--- a/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
+++ b/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
@@ -16,7 +16,7 @@ struct KeyboardState: Equatable {
     /// 1. View-A is shown and the keyboard is shown.
     /// 2. User taps on something which presents View-B **while** the keyboard is present.
     ///    View-B does not have a user responder (text field) so the keyboard is not visible.
-    /// 3. NSNotificationCenter emits a `keyboardDidHideNotification` but with a
+    /// 3. NSNotificationCenter emits a `keyboardWillHideNotification` but with a
     ///    `keyboardFrameEndUserInfoKey` value set to the **previously shown keyboard's frame**.
     ///
     let frameEnd: CGRect
--- a/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
+++ b/WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
@@ -53,7 +53,7 @@ final class KeyboardStateProvider: KeyboardStateProviding {
     init(notificationCenter: NotificationCenter = NotificationCenter.default) {
         self.notificationCenter = notificationCenter
 
-        let notificationNames = [UIResponder.keyboardDidShowNotification, UIResponder.keyboardDidHideNotification]
+        let notificationNames = [UIResponder.keyboardWillShowNotification, UIResponder.keyboardWillHideNotification]
 
         observations.append(contentsOf: notificationNames.map { notificationName in
             notificationCenter.addObserver(forName: notificationName, object: nil, queue: nil) { [weak self] notification in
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
```diff
@@ -8,7 +8,17 @@ struct KeyboardState: Equatable {
     let isVisible: Bool
     /// The frame of the keyboard when it is fully shown or hidden.
```
ealeksandrov, line 10: nit: a single newline before comment block can improve readability

ealeksandrov (approved):
Empty orders view looks good ✅
Empty variations view looks good ✅
Expanded error banner on orders list doesn't interfere with empty state ✅ (but empty state alignment is not perfect as described)

Code LGTM :shipit:
Great work! Special thanks for comments and detailed PR description 👏

WooCommerce/Classes/ViewRelated/Keyboard/KeyboardStateProvider.swift
shiki: Thanks. I'm never sure about the spacing of these properties. 😅 Added a newline in 9a01ed36bcb8147a9dd066ccfa0f980b7e6286c0.

shiki (commented):
Thank you very much for reviewing and testing this thoroughly, @ealeksandrov! 🙏

Issue #4865, opened 2021-08-23
Order Editing: Tracks events


ealeksandrov:
Following events should be added:

- order_detail_edit_flow_started
- order_detail_edit_flow_completed
- order_detail_edit_flow_canceled
- order_detail_edit_flow_failed

All events should have single parameter "subject" which describes data section. Values for this iteration will be:
- customer_note
- shipping_address
- billing_address

This means 4 events x 3 sections = 12 new unique events total.

Discussion ref: p1629219108064100-slack-CGPNUU63E

Note: Make sure the speadsheet is updated & events are validated

PR #4893, opened 2021-08-30
Order Editing: Add analytics & release customer note edit

Ecarrion:
part of #4865 

# Why 

Now that the customer provided now is ready, this PR adds the necessary tracking, releases notes, and removes it from being behind a feature flag.

# How 

- Define events
- Adds tracking in the view model & tests
- Remove the feature flag conditionals
- Add release notes

# Testing Stets
- Play with the feature and see the proper events fired in the console.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,7 @@
 
 7.5
 -----
-
+- [**] Merchants can now edit customer provided notes from orders. [https://github.com/woocommerce/woocommerce-ios/pull/4893]
 
 7.4
 -----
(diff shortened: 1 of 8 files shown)


ealeksandrov (commented):
Looks good! Verified all events, only one concern:

`order_detail_edit_flow_canceled` is not fired when I close modal view by swipe to bottom. Maybe we can add this event under conditional check to `.onDisappear`?

Ecarrion:

> order_detail_edit_flow_canceled is not fired when I close modal view by swipe to bottom. Maybe we can add this event under conditional check to .onDisappear?

I don't think `.onDissapear` is the best idea because that can be triggered in multiple situations. for example, if another modal is presented in the future or if we navigate to another view.

I have decided to use the UIKít `presentationControllerDidDismiss` method to handle this scenario.

ealeksandrov (approved):
LGTM :shipit:

PR #4899, opened 2021-08-31
[What's New Component] Add What's New SwiftUI components along with WhatsNewViewModel

caiobzen:
Part of: #4863 

# Description 📜

This PR introduces the User Interface designed to show to users what's new in WooCommerce!

# Main Changes ⚙️
- Added a few SwiftUI components **_(Most of the naming for every SwiftUI component was considering its reusability, as this screen may be used to present different content like reports, etc.)_**
  -  LargeTitleView: For displaying Large texts at the top of the screen
  - IconListItem: A view with a thumbnail, title and subtitle. 
  - ReportListView: A view with a large title, a list of iconListItems and a CTA button to dismiss the screen.  
- Bump Kingfisher to version 6.0 to have support for swiftUI Async Images

# Screenshots 📱

| iPhone 12 Pro Max  |  Phone 12 Pro Max (Dark mode)  | Phone 12 Pro Max (Dark mode) + RTL  | Phone SE |
| ------------------- | ------------------- | ------------------- | ------------------- |
|  image |  Captura de Tela 2021-08-30 às 11 13 15 PM | Captura de Tela 2021-08-30 às 7 32 02 PM | image  |

| iPad Pro | iPad Pro  (Dark mode)  |
|  image | image | 

# Testing 🧑‍🔬🧪

For now, use The component Previews on Xcode Canvas

## Update release notes:


WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+import Yosemite
+
+/// View model used to fill up the What's New Component, map from Features to ReportItems and dispatch analytics events
+final class WhatsNewViewModel: ReportListPresentable {
+    let items: [ReportItem]
+    let onDismiss: () -> Void
```
Ecarrion: nit: Could you add comments and documentation to these variables please?

WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
```diff
@@ -0,0 +1,15 @@
...
+    let title = NSLocalizedString("What’s New in WooCommerce", comment: "Title of Whats New Component")
+    let ctaTitle = NSLocalizedString("Continue", comment: "Title of continue button")
```
Ecarrion: We normally have this as a separate extension.
```swift
private extension WhatsNewViewModel {
    enum Localization {
        ...
    }
}
```

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
```diff
@@ -0,0 +1,58 @@
+import Kingfisher
+import SwiftUI
+
+/// Represent a row with a title, subtitle and icon. Used to fill up reports such as What's New in WooCommerce
+///
+struct IconListItem: View {
+    let title: String
+    let subtitle: String
+    let iconUrl: String
+    let icon: UIImage?
```
Ecarrion: Can we make this an enum, because it will be either one or the other right?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
```diff
@@ -0,0 +1,58 @@
...
+                    .resizable()
+                    .frame(width: Constants.iconSize.width, height: Constants.iconSize.height)
+            } else if let url = URL(string: iconUrl) {
+                KFImage(url)
+                    .resizable()
+                    .frame(width: Constants.iconSize.width, height: Constants.iconSize.height)
+            }
+            VStack(alignment: .leading, spacing: 2) {
+                Text(title)
+                    .font(.system(size: 17, weight: .semibold, design: .default))
```
Ecarrion: We have some custom styles modifiers to prevent styles from getting inconsistent.
eg:
```swift
Text(title)
    .bodyStyle()
```

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
```diff
@@ -0,0 +1,58 @@
...
+        .padding(.vertical, 16)
+    }
+
+    private var subtitleColor: Color {
+        Color(UIColor { $0.userInterfaceStyle == .dark
+            ? UIColor(red: 1, green: 1, blue: 1, alpha: 0.6)
+            : UIColor(red: 0, green: 0, blue: 0, alpha: 0.6)
+        })
+    }
+}
```
Ecarrion: I think this should not be needed, have you taken a look at `UIColor+SemanticColors`?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
```diff
@@ -0,0 +1,58 @@
...
+/// Represent a row with a title, subtitle and icon. Used to fill up reports such as What's New in WooCommerce
+///
+struct IconListItem: View {
+    let title: String
+    let subtitle: String
+    let iconUrl: String
+    let icon: UIImage?
+
+    var body: some View {
+        HStack(alignment: .center, spacing: 16) {
```
Ecarrion: Could you please extract all of these spacings, margins, etc into the `Constans` enum.
That way, it will be easier to maintain them in the future.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitleView.swift
```diff
@@ -0,0 +1,19 @@
+import SwiftUI
+
+/// Represents a large title, mostly to be used at the top of view controllers like What's New Component
+struct LargeTitleView: View {
+    let text: String
+
+    var body: some View {
+        Text(text)
+            .multilineTextAlignment(.center)
+            .font(.system(size: 34, weight: .bold, design: .default))
```
Ecarrion: Same comment about our custom styles. 😅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
```diff
@@ -0,0 +1,67 @@
+import SwiftUI
+import Yosemite
+
+protocol ReportListPresentable {
+    var items: [ReportItem] { get }
+    var title: String { get }
+    var ctaTitle: String { get }
+    var onDismiss: () -> Void { get }
+}
```
Ecarrion: Can you explain a bit what is the use case for this protocol?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitleView.swift
```diff
@@ -0,0 +1,19 @@
+import SwiftUI
+
+/// Represents a large title, mostly to be used at the top of view controllers like What's New Component
+struct LargeTitleView: View {
```
Ecarrion: I don't think we should use the postfix `View` right? to be in-line with apple naming.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
```diff
@@ -0,0 +1,67 @@
...
+
+struct ReportItem: Identifiable {
+    let id = UUID()
+    let title: String
+    let subtitle: String
+    let iconUrl: String
+    let iconBase64: String?
```
Ecarrion: I think we can use an enum here too.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    var icon: UIImage? {
+        guard let base64String = iconBase64 else { return nil }
+        return Data(base64Encoded: base64String).flatMap { UIImage(data: $0) }
+    }
+}
+
+/// Represent a screen with a list of IconListItems. Mainly used to present reports such as What's New in WooCommerce.
+///
+struct ReportListView: View {
```
Ecarrion: nit: Drop the view postfix.
Also, why Report instead of `FeatureList` or `FeaturesAnnouncement`

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
```diff
@@ -0,0 +1,67 @@
...
+    let viewModel: ReportListPresentable
+    private var isPad: Bool { UIDevice.isPad() }
```
Ecarrion: I don't think this is wise, especially since iPad apps can be displayed in different size classes.

Probably it's better to use the size class environment value.

https://www.hackingwithswift.com/quick-start/swiftui/how-to-create-different-layouts-using-size-classes

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
```diff
@@ -0,0 +1,67 @@
...
+                                 subtitle: $0.subtitle,
+                                 iconUrl: $0.iconUrl,
+                                 icon: $0.icon)
+                }
+            }
+            Spacer()
+            Button(viewModel.ctaTitle, action: viewModel.onDismiss)
+                .buttonStyle(PrimaryButtonStyle())
+                .padding(.horizontal, isPad ? 40 : 24)
+                .padding(.bottom, isPad ? 40 : 60)
```
Ecarrion: I think it would be easier to read and maintain to expose this as
```swift
private extension ReportListView {
    enum Layout {
        static func buttonPadding(sizeClass: UserInterfaceSizeClass) ->  {
            sizeClass == .compact ? 24 : 40
        }
    }
}

File: WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift [hunk 14]
```diff
@@ -0,0 +1,67 @@
...
+// MARK: - Preview
+struct ReportListView_Previews: PreviewProvider {
+    static var features: [Feature] {
+        let jsonData = try? JSONSerialization.data(withJSONObject: ["title": "foo",
+                                                                    "subtitle": "bar",
+                                                                    "iconBase64": "",
+                                                                    "iconUrl": "https://s0.wordpress.com/i/store/mobile/plans-premium.png"],
+                                                   options: .fragmentsAllowed)
+        let feature = try? JSONDecoder().decode(Feature.self, from: jsonData ?? Data())
+        return [feature, feature, feature].compactMap { $0 }
```
Ecarrion: I think it's safer, and for the sake of the project compile & preview times to expose a memberwise initializer on `Feature`, don't you think?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
```diff
@@ -0,0 +1,67 @@
...
+import Yosemite
+
+protocol ReportListPresentable {
+    var items: [ReportItem] { get }
+    var title: String { get }
+    var ctaTitle: String { get }
+    var onDismiss: () -> Void { get }
+}
+
+struct ReportItem: Identifiable {
```
Ecarrion: Why report item, Won't `FeatureItem` be more understandable?

Ecarrion (changes requested):
Thanks for splitting up into smaller PRs, much easier to review.
I have left some comments & requests. 
Also, I noticed that the items are not handling well bigger accessibility sizes.


Screen Shot 2021-08-31 at 10 58 35 AM

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
caiobzen: I thought about making these SwiftUI components less coupled to the WhatsNew context. As we could use them to represent other types of reports like the slow network one. Does it make sense?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
caiobzen: I'm not sure we are able to create a custom init for this model since it is Codable by default. This removed the ability to have memberwise inits. Another way around would be doing a bump on WordPressKit to make this change:

```swift
struct MyStruct {
    var name: String
}
extension MyStruct: Codable {} // preserves memberwise initializer

MyStruct(name: "Carlos")
```
Thoughts?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
caiobzen: Same as the comment I dropped [here](https://github.com/woocommerce/woocommerce-ios/pull/4899#discussion_r699497825)

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: Cool! This may fix the accessibility sizes issue! I'll do that 👍

WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
caiobzen: Sure thing! Done! ✅

WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
caiobzen: Super! Done ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: Yep! I've created an Icon enum with both cases with a function that returns an opaque type for the image view we are going to display: Regular image or Kingfisher async image 👍 

Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
caiobzen: Oh I didn't know about that one. I'll adopt it for sure! Thanks for sharing 🙏

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: image
🙌

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: Cool! I've used some of the colors from UIColor+SemanticColors! 👍 
Done ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
Ecarrion: The other, and probably better option now that I think about it, is to provide a memberwise initializer for your view model.

That way you can `init` it with that is needed.

```swift
init(items: [ReportItem], onDismiss: @escaping () -> Void) {
    self.items = items
    self.onDismiss = onDismiss
}
```

Then you can create it with `ReportItem` instances without having to mess with Json decoding in previews.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
Ecarrion: Oh right, that makes sense! 👍

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
Ecarrion: 👍 
Still worth to only name it it `ReportList` tho!

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
caiobzen: Yep! Sounds about right :+1:

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
caiobzen: Good idea! I've created an extension like this!
Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportListView.swift
caiobzen: Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
```diff
@@ -0,0 +1,67 @@
...
+        switch self {
+        case .base64(let image):
+            Image(uiImage: image)
+                .resizable()
+        case .remote(let url):
+            KFImage(url)
+                .resizable()
+        }
+    }
+}
```
Ecarrion: This looks nice!
I wonder if this should live under the `IconListItem` namespace? 🤔 

Anyway, I think it's fine like this for now.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
Ecarrion: Ping: Did you miss this comment? 😅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
```diff
@@ -0,0 +1,67 @@
...
+    var body: some View {
+        HStack(alignment: .center, spacing: 16) {
+            icon?.getImage()
+                .frame(width: Constants.iconSize.width, height: Constants.iconSize.height)
+            VStack(alignment: .leading, spacing: 2) {
+                Text(title)
+                    .headlineStyle()
+                Text(subtitle)
+                    .foregroundColor(Color(UIColor.gray(.shade30)))
+                    .bodyStyle()
```
Ecarrion: Can't this be `.secondaryBodyStyle()`?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitleView.swift
Ecarrion: Did you miss this comment?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitleView.swift
Ecarrion: Did you miss this comment?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportList.swift
```diff
@@ -0,0 +1,93 @@
+import SwiftUI
+import Yosemite
+
+protocol ReportListPresentable {
+    var items: [ReportItem] { get }
+    var title: String { get }
+    var ctaTitle: String { get }
+    var onDismiss: () -> Void { get }
+}
```
Ecarrion, line 9: I still do not understand the use-case for this protocol.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportList.swift
```diff
@@ -0,0 +1,93 @@
...
+    var icon: Icon? {
+        if let base64String = iconBase64,
+           let imageData = Data(base64Encoded: base64String),
+           let image = UIImage(data: imageData) {
+            return .base64(image)
+        } else if let url = URL(string: iconUrl) {
+            return .remote(url)
+        }
+        return nil
+    }
```
Ecarrion: Can't we just inject the correct enum? I don't see why we would want to have those tho properties if one won't be used.

Something like 
```swift
struct ReportItem: Identifiable {
    let id = UUID()
    let title: String
    let subtitle: String
    let icon: Icon
}
```

Ecarrion (changes requested):

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: Totally! I've changed it to secondaryBodyStyle right now 👍 
Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportList.swift
caiobzen: It was meant to have a way to create viewModels to conform to it. For example, we had the WhatsNewViewModel that received a list of features and transformed into ReportItem. Now that the viewModel doesn't transform this data anymore, we might want to remove the need of this protocol! 
I will do it in a few 👍

caiobzen: I've created this protocol so we can conform to it to have different title / ctaTitle, transform models into ReportItem, etc.
Now it is going to be used for WhatsNewViewModel only. Do you think we should remove this boundary and re-introduce something similar when we need to build different screens in the future? (for the other reports to come like the slow store one)

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: I think it makes sense! I'll change it on this PR if you don't mind :)
Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportList.swift
caiobzen: Yep, makes total sense!
Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
caiobzen: Yep! Thanks for reminding me! 
Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitleView.swift
caiobzen: Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitleView.swift
caiobzen: Yep! Thanks for the reminder!
Done! ✅

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitle.swift
```diff
@@ -0,0 +1,20 @@
+import SwiftUI
+
+/// Represents a large title, mostly to be used at the top of view controllers like What's New Component
+struct LargeTitle: View {
+    let text: String
+
+    var body: some View {
+        Text(text)
+            .font(.largeTitle)
+            .bold()
```
Ecarrion: We should probably create a modifier like `.titleStyle()`
Also, doesn't this needs color?

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ReportList.swift
Ecarrion: I would have probably renamed the view model to something like `ReportListViewModel` and let consumers create an instance. 

But this is also not wrong, so I'm fine if you choose to leave it like this!

Ecarrion (approved):
This is looking good! Only a minor requests and we should be good to go.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/LargeTitle.swift
caiobzen: Sure thing! I've created a new Style for the LargeTitleStyle 👍 
Done! ✅

PR #4902, opened 2021-09-01
Shipping Labels: Hide bottom bar on creation form

### LINKED ISSUE #4852: Shipping Labels: do not show bottom nav bar on secondary screens
> Can we hide the nav bar in the Shipping label flow? In general, secondary screens shouldn’t be displaying the nav bar.

image

itsmeichigo:
Closes #4852 

# Description
To preserve more real estate on shipping label form, this PR aims to hide bottom bar entirely from the main creation form and its subsequent screens: Edit Address, Package Details, Customs, Carriers, Payment Method. The screens should then have edge-to-edge look where the bottom bar used to be, but content should be displayed inside safe area.

# Changes
- Set `hidesBottomBarWhenPushed` to true for the shipping label form controller before pushing.
- Updated shipping label form table view to extend to bottom.
- Update bottom padding for SwiftUI screens: Package Details, Customs, Carriers, Payment Method to show contents within safe area.

# Demo
no-tab-bar

# Testing
2. Create an international order (origin and destination countries are different).
3. Navigate to Orders tab, select the new order and select Create Shipping Label.
4. Notice that the shipping label form no longer shows bottom bar, and the list is extended to bottom edge while its contents stay within safe area when scrolled all the way to the bottom.
5. Configure each field: address, package, customs, carrier, payment method and notice that all screens extend to bottom edge while contents stay within safe area when scrolled all the way to the bottom.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -3,6 +3,7 @@
 7.5
 -----
 - [*] Fix empty states sometimes not centered vertically [https://github.com/woocommerce/woocommerce-ios/pull/4890]
+- [*] Hide bottom bar on shipping label purchase form. [https://github.com/woocommerce/woocommerce-ios/pull/4902]
 
 7.4
 -----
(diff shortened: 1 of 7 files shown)


pmusolino (approved):
LGTM :shipit:

PR #4882, opened 2021-08-26
Shipping Labels: Initial UI for selecting/activating a service package screen

rachelmcr:
Part of: #4744

## Description

In the shipping label creation flow, this introduces the UI to select a new (not yet enabled) service package to use for your shipping label. This view is visible when you select the "Service Package" tab on the Add New Package screen.

For now, this screen shows a list of the enabled service packages on your store, for demo purposes. Getting the list of not yet enabled service packages requires changes in the networking model, which will be done in a future PR. This also doesn't yet handle the empty state (when the service package list is empty), which will come in a future PR.

## Changes

* Adds `ShippingLabelServicePackageList` and view model, which is based on the UI in `ShippingLabelPackageDetails` (but simplified to fit this simpler view).
* Updates `ShippingLabelAddNewPackage` to display `ShippingLabelServicePackageList` when the "Service Package" tab is selected.

Design:

Portrait|Landscape
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-08-26 at 17 35 51|Simulator Screen Shot - iPhone 12 Pro - 2021-08-26 at 17 36 10

## Testing

1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more packages set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Launch the app in debug mode (so the feature flag is enabled).
3. Open an order detail that is eligible for the shipping label.
4. Tap the button "Create Shipping Label.”
5. Confirm "Ship From" and "Ship To.”
6. In the package details section, tap on “Package Selected.”
7. On the Package Selected screen, tap the “Create new package” button at the bottom of the screen.
8. On the Add New Package screen, select "Service Package" and confirm a list of service (predefined) packages appears, organized by shipping provider. Confirm you can select a package from the list and a checkmark appears next to the selected package.

## Submitter Checklist

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #4744, opened 2021-08-06
Shipping Labels: Service package screens


rachelmcr:
This is a subtask of #3909.

- [x] Create the screen "Service Packages": we should do a GET on packages. It will return formSchema and formData. formSchema is the list of all available service packages. formData is the list of all created/activated custom/service packages. We had to compare the two to get the service packages that can still be enabled. We should allow the selection of just one predefined package, and this will be the only difference with wp-admin.
- [x] Empty view design is needed in case all packages are already activated. (We have an existing SwiftUI EmptyState reusable component we should be able to use for this.)

Things to be aware of:

- Activate only one service at a time.
- Once a service package is activated, make sure it is then listed in the package selector screen and removed from the service package activation screen (in case the user goes back to activate another service package).

![Service-Package-List](https://user-images.githubusercontent.com/8658164/128549046-71a0cb0a-a1dd-4004-811f-c7be9795098c.png)

PR #4905, opened 2021-09-01
Shipping Labels: Display list of unactivated service packages on Add New Package > Service Package tab

rachelmcr:
Part of: #4744

## Description

In the shipping label creation flow, https://github.com/woocommerce/woocommerce-ios/pull/4882 added the initial UI on the "Add New Package" screen to select a new (not yet activated) service package to use for your shipping label. This PR makes the changes to display the correct list of service packages on that screen.

In wp-admin, these packages are found under WooCommerce > Settings > Shipping > WooCommerce Shipping (`/wp-admin/admin.php?page=wc-settings&tab=shipping&section=woocommerce-services-settings`) by selecting Add Package > Service package. The activated packages have a checkmark there (and are shown on the "Package Selected" list in the app), while the unactivated packages are unchecked there (and are now shown on Add New Package > Service Package in the app).

There are a couple limitations that will be addressed in a future PR:

* If you switch tabs on the "Add New Package" screen, your selection is lost.
* There is not yet an empty state screen, so if all service packages are enabled on your site you will see a blank list.

## Changes

* The networking model `ShippingLabelPackagesResponse` now has a property `unactivatedPredefinedOptions` created from the shipping label packages API response. This property contains the predefined options that are not yet activated on the store. It's populated by getting a list of all the predefined packages on the store and filtering out the ones that are already activated, and then organizing it by provider.
* `ShippingLabelServicePackageListViewModel` (the view model for the Add New Package > Service Package tab view) is updated to use the new `unactivatedPredefinedOptions` property.
* Updated fakes and unit tests, including a check in `ShippingLabelPackagesMapperTests` to confirm that `unactivatedPredefinedOptions` contains all of the unactivated service packages and none of the activated service packages for a shipping provider (DHL Express).

**Add New Package > Service Package**

**Related screen in wp-admin**

Screen Shot 2021-09-01 at 2 58 14 PM

## Testing

8. On the Add New Package screen, select "Service Package" and confirm a list of new (not yet activated) service packages appears, organized by shipping provider. There should be no overlap between the list of service packages that appear on the "Package Selected" screen and on this screen.

## Submitter Checklist

Update release notes:

### DIFF
--- a/Fakes/Fakes/Fakes.generated.swift
+++ b/Fakes/Fakes/Fakes.generated.swift
@@ -1141,7 +1141,8 @@ extension ShippingLabelPackagesResponse {
         .init(
             storeOptions: .fake(),
             customPackages: .fake(),
-            predefinedOptions: .fake()
+            predefinedOptions: .fake(),
+            unactivatedPredefinedOptions: .fake()
         )
     }
 }
--- a/Networking/Networking/Model/ShippingLabel/Packages/ShippingLabelPackagesResponse.swift
+++ b/Networking/Networking/Model/ShippingLabel/Packages/ShippingLabelPackagesResponse.swift
@@ -10,14 +10,20 @@ public struct ShippingLabelPackagesResponse: Equatable, GeneratedFakeable {
 
     public let customPackages: [ShippingLabelCustomPackage]
 
+    /// Activated predefined options
     public let predefinedOptions: [ShippingLabelPredefinedOption]
+
+    /// Unactivated predefined options
+    public let unactivatedPredefinedOptions: [ShippingLabelPredefinedOption]
 
     public init(storeOptions: ShippingLabelStoreOptions,
                 customPackages: [ShippingLabelCustomPackage],
-                predefinedOptions: [ShippingLabelPredefinedOption]) {
+                predefinedOptions: [ShippingLabelPredefinedOption],
+                unactivatedPredefinedOptions: [ShippingLabelPredefinedOption]) {
         self.storeOptions = storeOptions
         self.customPackages = customPackages
         self.predefinedOptions = predefinedOptions
+        self.unactivatedPredefinedOptions = unactivatedPredefinedOptions
     }
 }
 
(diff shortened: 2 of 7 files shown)


pmusolino (approved):
LGTM :shipit: 
Just a question: I'm not sure if it's a regression or if it's expected behavior, but I noticed that If I enter this new screen in portrait mode, and I rotate the device, the view will be dismissed. Is it expected or it's normal behavior as far as you know?

rachelmcr:
> Just a question: I'm not sure if it's a regression or if it's expected behavior, but I noticed that If I enter this new screen in portrait mode, and I rotate the device, the view will be dismissed. Is it expected or it's normal behavior as far as you know?

Oh weird, that must be a regression — I thought I had device rotation working ok as of 139569f (part of https://github.com/woocommerce/woocommerce-ios/pull/4882). I'm working now on some changes to better retain state, and I'll double check the behavior when rotating the device as part of those changes. Thanks for pointing it out!

rachelmcr:
I did some testing with the device rotation and realized that it's only happening to me on device, not in a simulator. (Probably why I didn't catch it before.) It's happening in the 7.4 beta after opening the Package Selected view, so it isn't specific to the Add New Package view but I wonder if it's related to the `ShippingLabelPackageSelection` view I introduced as part of this flow.

In any case, I opened https://github.com/woocommerce/woocommerce-ios/issues/4913 to keep track of this and will take a closer look, but probably separately from the other state behavior I'm working on since it seems to be a different issue.

pmusolino:
> I did some testing with the device rotation and realized that it's only happening to me on device, not in a simulator. (Probably why I didn't catch it before.) It's happening in the 7.4 beta after opening the Package Selected view, so it isn't specific to the Add New Package view but I wonder if it's related to the ShippingLabelPackageSelection view I introduced as part of this flow.

> In any case, I opened #4913 to keep track of this and will take a closer look, but probably separately from the other state behavior I'm working on since it seems to be a different issue.

I'm able to replicate it also on the simulator (iPhone 12 Pro, iOS 14.5)

Issue #4778, opened 2021-08-11
Order Editing: Address form UI


Ecarrion:
<img width="283" alt="Screen Shot 2021-08-11 at 9 54 50 AM" src="https://user-images.githubusercontent.com/562080/129054686-04a7f429-db71-4fc3-bb3d-eae329e9b32f.png">

Tasks:
- [x] Build UI extensible to be used for shipping/billing address.
- [x] Connect screen from indicator buttons #4768 

Considerations:
- Try to make the screen extensible enough to be used in other places where an address form may be needed.

Ecarrion:
Hi @ealeksandrov, 

I will take this tomorrow, do you have in-progress work?

ealeksandrov:
Hi @Ecarrion! I've added all billing address stuff in https://github.com/woocommerce/woocommerce-ios/pull/5054, sorry for delay with PR.

PR #4906, opened 2021-09-01
Order Editing: Add basic edit address form

ealeksandrov:
Part of https://github.com/woocommerce/woocommerce-ios/issues/4778.

## Description

This PR adds simple SwiftUI implementation of Address Form that'll be used on shipping and billing address screens.
It doesn't include data flow, only UI with labels, buttons and textfields.

### Considerations

- In design we have this screen as part of navigation stack. But note editing is implemented as a modal. Should we unify this?
- Alignment between textfields is slightly different then design (right alignment). We can improve it later, but I consider it unnecessary complex for now. For comparison, address form in shipping labels (UIKit) is not aligned between cells too.

## Screenshots

portrait | landscape
--|--
Simulator Screen Shot - iPhone 12|Simulator Screen Shot - iPhone 12

## Test

1. Open existing order details in development build (with feature flag enabled)
2. Tap edit button on shipping address cell

---
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
@@ -1,0 +1,165 @@
+import Foundation
+import Combine
+import SwiftUI
+
+/// Hosting controller that wraps an `EditAddressView`.
+///
+final class EditAddressHostingController: UIHostingController<EditAddressView> {
+
+    init() {
+        super.init(rootView: EditAddressView())
+    }
+
+    required dynamic init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+/// Allows merchant to edit the customer provided address of an order.
+///
+struct EditAddressView: View {
+    var body: some View {
+        GeometryReader { geometry in
+            ScrollView {
+                ListHeaderView(text: Localization.detailsSection, alignment: .left)
+                    .padding(.horizontal, insets: geometry.safeAreaInsets)
+                VStack(spacing: 0) {
+                    TitleAndTextFieldRow(title: Localization.firstNameField,
+                                         placeholder: "",
+                                         text: .constant(""),
+                                         symbol: nil,
+                                         keyboardType: .default)
+                    Divider()
+                        .padding(.leading, Constants.dividerPadding)
+                    TitleAndTextFieldRow(title: Localization.lastNameField,
+                                         placeholder: "",
+                                         text: .constant(""),
+                                         symbol: nil,
+                                         keyboardType: .default)
+                    Divider()
+                        .padding(.leading, Constants.dividerPadding)
+                    TitleAndTextFieldRow(title: Localization.emailField,
+                                         placeholder: "",
+                                         text: .constant(""),
+                                         symbol: nil,
+                                         keyboardType: .emailAddress)
+                    Divider()
+                        .padding(.leading, Constants.dividerPadding)
+                    TitleAndTextFieldRow(title: Localization.phoneField,
+                                         placeholder: "",
+                                         text: .constant(""),
+                                         symbol: nil,
+                                         keyboardType: .phonePad)
+                }
+                .padding(.horizontal, insets: geometry.safeAreaInsets)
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    required dynamic init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+/// Allows merchant to edit the customer provided address of an order.
+///
+struct EditAddressView: View {
```
Ecarrion: nit: Should this be called `EditAddress`? I have noticed that none of the apple components appends the `View` postfix and I think we should probably follow that naming style.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
```diff
@@ -0,0 +1,166 @@
...
+                VStack(spacing: 0) {
+                    Group {
+                        TitleAndTextFieldRow(title: Localization.companyField,
+                                             placeholder: Localization.placeholderOptional,
+                                             text: .constant(""),
+                                             symbol: nil,
+                                             keyboardType: .default)
+                        Divider()
+                            .padding(.leading, insets: geometry.safeAreaInsets)
+                            .padding(.leading, Constants.dividerPadding)
```
Ecarrion: Why do we need a double padding here? Was it a mistake or is it a weird `SwiftUI` view?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
```diff
@@ -0,0 +1,166 @@
...
+                                         symbol: nil,
+                                         keyboardType: .phonePad)
+                }
+                .padding(.horizontal, insets: geometry.safeAreaInsets)
+                .background(Color(.systemBackground))
+
+                ListHeaderView(text: Localization.shippingAddressSection, alignment: .left)
+                    .padding(.horizontal, insets: geometry.safeAreaInsets)
+                VStack(spacing: 0) {
+                    Group {
```
Ecarrion: I did not know there was a 10-child view limit on `SwiftUI` view builders! 🤯

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
Ecarrion: I just removed them and it all looks correct!
Screen Shot 2021-09-01 at 3 25 18 PM

Ecarrion (approved):
It all looks good to me! just some minor comments.

Regarding to:

> In design we have this screen as part of navigation stack. But note editing is implemented as a modal. Should we unify this?

and 

> Alignment between textfields is slightly different then design (right alignment). We can improve it later, but I consider it unnecessary complex for now. For comparison, address form in shipping labels (UIKit) is not aligned between cells too.

What are your thoughts @adamzelinski ?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
ealeksandrov: Looks like I forgot to remove internal padding on second section, while it's correct on first. Thanks for noticing!
Fixed in f25ff68f8fb5841dc2e0734963a5294011460d6a

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressView.swift
ealeksandrov: There are some examples on Apple's side like `ContentView`, `MapView`, but I agree with you and don't like overuse of `View` and `ViewController` naming.
Renamed to `EditAddressForm` since it describes function better in 506b12a7cfe4085c3788b325f8c321f3bb3cd66a

ealeksandrov:
Thank you for review @Ecarrion! I'll merge it for now and we can update navigation flow separately later.

PR #4908, opened 2021-09-01
[Whats New Component] Show What's New on app launch

caiobzen:
Part of: #4863 

# Description 📜

This PR makes the integration between the existing logic to fetch announcements (and now mark them as displayed as well) and the SwiftUI work done in the [previous Pull Request](https://github.com/woocommerce/woocommerce-ios/pull/4899).

# Main Changes ⚙️
- Add all the fields from WordPressKit to Yosemite.Announcement so we can encode/decode as we wish
- AppCoordinator now fetches the latest announcement and displays the WhatsNewComponent on app launch
- Introduces WhatsNewFactory to wrap app the viewController/viewModel creation and UI setup for What's new component
- Add new AnnouncementAction: `markSavedAnnouncementAsDisplayed `: responsible for marking the saved announcement as `displayed`.

### Flow Diagram
This is the flow for fetching and persisting an announcement on app launch
image

### Recordings:
| iPhone 12 Pro Max  |  iPad Pro                  |
| <video src="https://user-images.githubusercontent.com/997521/131748713-cde5ec9f-5599-4f1b-9ba9-4ac772724d23.mp4"> |     <video src="https://user-images.githubusercontent.com/997521/131748755-a51afdbd-6a72-407b-b0c6-c53a9a91cb93.mov">  |

# Testing 🧑‍🔬🧪

### ⚠️ Pre-requisite ⚠️
_Please do the following change_
image

## Scenario 1
1. Run the app
2. What's New component should be displayed upon app launch
3. Tap Continue to dismiss the modal

## Scenario 2
1. Run the app again (after seeing the What's New component once)
2. What's New component should **NOT** be displayed

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,7 +5,6 @@
 - [**] Merchants can now edit customer provided notes from orders. [https://github.com/woocommerce/woocommerce-ios/pull/4893]
 - [*] Fix empty states sometimes not centered vertically [https://github.com/woocommerce/woocommerce-ios/pull/4890]
 - [*] Hide bottom bar on shipping label purchase form. [https://github.com/woocommerce/woocommerce-ios/pull/4902]
-- [*] Show "What's New on WooCommerce" on app launch [https://github.com/woocommerce/woocommerce-ios/pull/4908]
 
 7.4
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,7 +5,7 @@
 - [**] Merchants can now edit customer provided notes from orders. [https://github.com/woocommerce/woocommerce-ios/pull/4893]
 - [*] Fix empty states sometimes not centered vertically [https://github.com/woocommerce/woocommerce-ios/pull/4890]
 - [*] Hide bottom bar on shipping label purchase form. [https://github.com/woocommerce/woocommerce-ios/pull/4902]
-- [*] Display Announcements on app launch [https://github.com/woocommerce/woocommerce-ios/pull/4908]
+- [*] Show "What's New on WooCommerce" on app launch [https://github.com/woocommerce/woocommerce-ios/pull/4908]
 
 7.4
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 - [**] Merchants can now edit customer provided notes from orders. [https://github.com/woocommerce/woocommerce-ios/pull/4893]
 - [*] Fix empty states sometimes not centered vertically [https://github.com/woocommerce/woocommerce-ios/pull/4890]
 - [*] Hide bottom bar on shipping label purchase form. [https://github.com/woocommerce/woocommerce-ios/pull/4902]
+- [*] Display Announcements on app launch [https://github.com/woocommerce/woocommerce-ios/pull/4908]
 
 7.4
 -----
--- a/Storage/Storage/Model/Announcement.swift
+++ b/Storage/Storage/Model/Announcement.swift
@@ -4,18 +4,36 @@
 /// These entities will be serialised to a plist file
 
 public struct Announcement: Codable {
-    public let appVersion: String
-    public let features: [Feature]
+    public let appVersionName: String
+    public let minimumAppVersion: String
+    public let maximumAppVersion: String
+    public let appVersionTargets: [String]
+    public let detailsUrl: String
     public let announcementVersion: String
+    public let isLocalized: Bool
+    public let responseLocale: String
+    public let features: [Feature]
     public let displayed: Bool
 
-    public init(appVersion: String,
-                features: [Feature],
+    public init(appVersionName: String,
+                minimumAppVersion: String,
+                maximumAppVersion: String,
+                appVersionTargets: [String],
+                detailsUrl: String,
                 announcementVersion: String,
+                isLocalized: Bool,
+                responseLocale: String,
+                features: [Feature],
                 displayed: Bool) {
-        self.appVersion = appVersion
-        self.features = features
+        self.appVersionName = appVersionName
+        self.minimumAppVersion = minimumAppVersion
+        self.maximumAppVersion = maximumAppVersion
+        self.appVersionTargets = appVersionTargets
+        self.detailsUrl = detailsUrl
         self.announcementVersion = announcementVersion
+        self.isLocalized = isLocalized
+        self.responseLocale = responseLocale
+        self.features = features
         self.displayed = displayed
     }
 }
--- a/WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
@@ -28,9 +28,9 @@ final class WhatsNewViewModel: ReportListPresentable {
         stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { result in
             switch result {
             case .success:
-                return DDLogInfo("📣 Announcement was marked as displayed! ✅")
+                DDLogInfo("📣 Announcement was marked as displayed! ✅")
             case .failure(let error):
-                return DDLogInfo("📣 Failed to mark announcement as displayed: \(error.localizedDescription)")
+                DDLogInfo("⛔️ Failed to mark announcement as displayed: \(error.localizedDescription)")
             }
         }))
     }
--- a/WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
@@ -15,9 +15,24 @@ final class WhatsNewViewModel: ReportListPresentable {
     /// Title of the Call to action button
     let ctaTitle = Localization.ctaTitle
 
-    init(items: [ReportItem], onDismiss: @escaping () -> Void) {
+    /// StoresManager that will be handling actions
+    private let stores: StoresManager
+
+    init(items: [ReportItem], stores: StoresManager = ServiceLocator.stores, onDismiss: @escaping () -> Void) {
         self.items = items
+        self.stores = stores
         self.onDismiss = onDismiss
+    }
+
+    func onAppear() {
+        stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { result in
+            switch result {
+            case .success:
+                return DDLogInfo("📣 Announcement was marked as displayed! ✅")
+            case .failure(let error):
+                return DDLogInfo("📣 Failed to mark announcement as displayed: \(error.localizedDescription)")
+            }
+        }))
     }
 }
 
--- a/WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
+++ b/WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
@@ -30,7 +30,7 @@ final class WhatsNewViewModel: ReportListPresentable {
             case .success:
                 DDLogInfo("📣 Announcement was marked as displayed! ✅")
             case .failure(let error):
-                DDLogInfo("⛔️ Failed to mark announcement as displayed: \(error.localizedDescription)")
+                DDLogError("⛔️ Failed to mark announcement as displayed: \(error.localizedDescription)")
             }
         }))
     }
--- a/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
@@ -74,6 +74,8 @@ private extension AppCoordinator {
         }))
     }
 
+    /// Load saved announcement and display it on What's New component if it was not displayed yet
+    ///
     func showWhatsNewIfNeeded() {
         stores.dispatch(AnnouncementsAction.loadSavedAnnouncement(onCompletion: { [weak self] result in
             guard let self = self, let (announcement, displayed) = try? result.get(), !displayed else {
@@ -83,8 +85,14 @@ private extension AppCoordinator {
                 self?.tabBarController.dismiss(animated: true)
             }
             self.tabBarController.present(whatsNewViewController, animated: true, completion: nil)
-            self.stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { error in
-                error.flatMap { return DDLogInfo("💬 \($0.localizedDescription)") }
+
+            self.stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { result in
+                switch result {
+                case .success:
+                    return DDLogInfo("📣 Announcement of version \(announcement.appVersionName) was marked as read! ✅")
+                case .failure(let error):
+                    return DDLogInfo("📣 Failed to mark announcement as displayed: \(error.localizedDescription)")
+                }
             }))
         }))
     }
--- a/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
@@ -86,15 +86,6 @@ private extension AppCoordinator {
                 self?.tabBarController.dismiss(animated: true)
             }
             self.tabBarController.present(whatsNewViewController, animated: true, completion: nil)
-
-            self.stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { result in
-                switch result {
-                case .success:
-                    return DDLogInfo("📣 Announcement of version \(announcement.appVersionName) was marked as read! ✅")
-                case .failure(let error):
-                    return DDLogInfo("📣 Failed to mark announcement as displayed: \(error.localizedDescription)")
-                }
-            }))
         }))
     }
 
--- a/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
@@ -62,6 +62,8 @@ private extension AppCoordinator {
     /// Synchronize announcements and present What's New Screen if needed
     ///
     func synchronizeAndShowWhatsNew() {
+        guard ServiceLocator.featureFlagService.isFeatureFlagEnabled(.whatsNewOnWooCommerce) else { return }
+
         stores.dispatch(AnnouncementsAction.synchronizeAnnouncements(onCompletion: { [weak self] result in
             guard let self = self else { return }
             switch result {
@@ -77,8 +79,6 @@ private extension AppCoordinator {
     /// Load saved announcement and display it on What's New component if it was not displayed yet
     ///
     func showWhatsNewIfNeeded() {
-        guard ServiceLocator.featureFlagService.isFeatureFlagEnabled(.whatsNewOnWooCommerce) else { return }
-
         stores.dispatch(AnnouncementsAction.loadSavedAnnouncement(onCompletion: { [weak self] result in
             guard let self = self else { return }
             guard let (announcement, displayed) = try? result.get(), !displayed else {
--- a/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/AppCoordinator.swift
@@ -68,7 +68,7 @@ private extension AppCoordinator {
             case .success(let announcement):
                 DDLogInfo("📣 Announcements Synced! AppVersion: \(announcement.appVersionName) | AnnouncementVersion: \(announcement.announcementVersion)")
             case.failure(let error):
-                DDLogInfo("⛔️ Failed to synchronize announcements: \(error.localizedDescription)")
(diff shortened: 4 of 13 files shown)


WooCommerce/Classes/ViewRelated/AppCoordinator.swift
```diff
@@ -49,21 +49,52 @@ final class AppCoordinator {
...
+        stores.dispatch(AnnouncementsAction.loadSavedAnnouncement(onCompletion: { [weak self] result in
+            guard let self = self, let (announcement, displayed) = try? result.get(), !displayed else {
+                return DDLogInfo("📣 There are no announcements to show!")
+            }
+            let whatsNewViewController = WhatsNewFactory.whatsNew(announcement) { [weak self] in
+                self?.tabBarController.dismiss(animated: true)
+            }
+            self.tabBarController.present(whatsNewViewController, animated: true, completion: nil)
+
+            self.stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { result in
```
caiobzen: @Ecarrion Not sure if marking the announcement as displayed right here is the best idea. I'd like to hear from you what you think about doing it here versus moving it to the CTA action triggered by the WhatsNewViewModel.

As far as I can tell, there's only one benefit on leaving this code here:
-  In case user closes the app before pressing the CTA button on what's new component, we would display the announcement once again when the app is reopened. _Doesn't seem a big deal IMHO._

The con is:
- This function is doing several things already.. (bye bye SRP)

Storage/Storage/Model/Announcement.swift
```diff
@@ -4,18 +4,36 @@
...
-    public let features: [Feature]
+    public let appVersionName: String
+    public let minimumAppVersion: String
+    public let maximumAppVersion: String
+    public let appVersionTargets: [String]
+    public let detailsUrl: String
     public let announcementVersion: String
+    public let isLocalized: Bool
+    public let responseLocale: String
+    public let features: [Feature]
```
Ecarrion, line 15: [Q] Why did you decide to store all of these values now?
To be able to reconstruct the object from a saved state?

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
```diff
@@ -49,21 +49,52 @@ final class AppCoordinator {
...
+    /// Synchronize announcements and present What's New Screen if needed
+    ///
+    func synchronizeAndShowWhatsNew() {
+        stores.dispatch(AnnouncementsAction.synchronizeAnnouncements(onCompletion: { [weak self] result in
+            guard let self = self else { return }
+            switch result {
+            case .success(let announcement):
+                DDLogInfo("📣 Announcements Synced! AppVersion: \(announcement.appVersionName) | AnnouncementVersion: \(announcement.announcementVersion)")
+            case.failure(let error):
+                DDLogInfo("📣 Failed to synchronize announcements: \(error.localizedDescription)")
```
Ecarrion:
```suggestion
                DDLogError("⛔️ Failed to synchronize announcements: \(error.localizedDescription)")
```

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
```diff
@@ -49,21 +49,52 @@ final class AppCoordinator {
...
+
+    /// Load saved announcement and display it on What's New component if it was not displayed yet
     ///
     func showWhatsNewIfNeeded() {
-        // TODO: Check the saved Announcement App Version in order to display or not the what's new component
-        stores.dispatch(AnnouncementsAction.synchronizeAnnouncements(onCompletion: { _ in }))
+        stores.dispatch(AnnouncementsAction.loadSavedAnnouncement(onCompletion: { [weak self] result in
+            guard let self = self, let (announcement, displayed) = try? result.get(), !displayed else {
+                return DDLogInfo("📣 There are no announcements to show!")
+            }
```
Ecarrion: We should probably separate the `guard let self = self` from the other conditions as the error message does not correspond correctly.

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
Ecarrion: Yeah, this does not seem right here.
I think we should mark this as displayed on `WhatsNewViewModel` and call in on the `.onAppear {}` modifier/Action.

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewFactory.swift
```diff
@@ -0,0 +1,39 @@
...
+        let viewModel = WhatsNewViewModel(items: items, onDismiss: onDismiss)
+        let rootView = ReportList(viewModel: viewModel)
+        let hostingViewController = WhatsNewHostingController(rootView: rootView)
+        return hostingViewController
+    }
+
+    /// Transform Features into ReportItem models
+    private static func mapFeaturesToItems(_ features: [Feature]) -> [ReportItem] {
+        features.map {
+            return ReportItem(title: $0.title, subtitle: $0.subtitle, icon: icon(for: $0))
```
Ecarrion:
```suggestion
           ReportItem(title: $0.title, subtitle: $0.subtitle, icon: icon(for: $0))
```

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewFactory.swift
```diff
@@ -0,0 +1,39 @@
...
+        var icon: IconListItem.Icon?
+        if let base64String = feature.iconBase64,
+           let imageData = Data(base64Encoded: base64String),
+           let image = UIImage(data: imageData) {
+            icon = .base64(image)
+        } else if let url = URL(string: feature.iconUrl) {
+            icon = .remote(url)
+        }
+        return icon
+    }
```
Ecarrion, line 38: Nice!

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewHostingController.swift
```diff
@@ -0,0 +1,28 @@
...
+        super.init(rootView: rootView)
+        if UIDevice.isPad() {
+            preferredContentSize = Layout.iPadContentSize
+        }
+        modalPresentationStyle = .formSheet
+    }
+
+    override var traitCollection: UITraitCollection {
+        self.presentingViewController?.traitCollection ?? .current
+    }
```
Ecarrion, line 16: I don't think this is needed, is it?

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewHostingController.swift
```diff
@@ -0,0 +1,28 @@
...
+    required dynamic init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
+
+// MARK: - Constants
+//
+private extension WhatsNewHostingController {
+    enum Layout {
+        static let iPadContentSize = CGSize(width: 360, height: 574)
```
Ecarrion, line 27: I'm not a fan of hardcoding sizes, specifically because we don't know how those sizes will evolve over time. 

This, however does look a bit nicer 🤔 

What are your thoughts?

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -88,34 +99,87 @@ private extension AnnouncementsStore {
...
     }
 
-    /// Save the `Announcement` to the appropriate file.
-    func saveAnnouncement(_ announcement: Announcement) throws {
+    /// Save the `Announcement` on disk if it's a new or a different announcement comparing to the existing one
+    func saveAnnouncementIfNeeded(_ announcement: Announcement) throws {
+        if let storageAnnouncement = storageAnnouncement {
+            guard !announcement.isSameVersion(storageAnnouncement) else {
+                throw AnnouncementsStorageError.announcementAlreadyExists
+            }
```
Ecarrion: Do you think we should check if the version is greater and not just a different one?

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -41,6 +43,11 @@ public class AnnouncementsStore: Store {
             .appendingPathComponent(Constants.featureAnnouncementsFileName)
     }()
 
+    private lazy var storageAnnouncement: StorageAnnouncement? = {
```
Ecarrion: I think this variable will be very problematic as it will act as a shared state between all actions handled by this store.

As it can become a source of stale content, I think we should fetch the announcement everytime we need it.

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -88,34 +99,87 @@ private extension AnnouncementsStore {
...
+
+        guard let fileURL = featureAnnouncementsFileURL else { throw AnnouncementsStorageError.unableToFindFileURL }
         let mappedAnnouncement = self.mapAnnouncementToStorageModel(announcement)
+        try fileStorage.write(mappedAnnouncement, to: fileURL)
+    }
+
+    /// Load the latest saved `Announcement`. Returns an Error if there is no saved announcement.
+    func loadSavedAnnouncement(onCompletion: (Result<(Announcement, IsDisplayed), Error>) -> Void) {
+        do {
+            onCompletion(.success((try savedAnnouncement(), storageAnnouncement?.displayed == true)))
```
Ecarrion: I think it's quite strange to return an announcement but check it's displayed properly from another variable, why not use the announcement that we just loaded?

Ecarrion (changes requested):
Hey @caiobzen, testing looks good! I really liked the factory class to decouple that view creation.

I have left you some questions and concerns!

Also, do you think it's possible to prevent the list from scrolling if the content is not long enough? I think it looks a bit off. Feel free to create a separate issue and tackle that later.


https://user-images.githubusercontent.com/562080/131868176-9f9a255b-3605-4153-bff7-226459102395.mov

Storage/Storage/Model/Announcement.swift
caiobzen: Exactly! those fields are all mandatory on WordPressKit.Announcement

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewHostingController.swift
caiobzen: Yes it is! Once you define a `preferredContentSize` for that view (iPad only), the traits are now `.compact` due to the view size. This makes it impossible to distinguish what is iPhone and what is iPad to apply the right sizing, margins, etc.

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewHostingController.swift
caiobzen: My first idea was to use (screen size * a multiplier), but it didn't worked as expected. The fallback solution was to draw it in this static size fashion (WhichI'm not a big fan either).

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: I do agree! but that's mainly because the WordPress model does't have such property and we don't want to leak the storageModel to the app. Therefore, grabbing this information from the saved model was the best plan I have so far!

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
caiobzen: Good call!
Done! ✅

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
caiobzen: Nice catch!
Done! ✅

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewFactory.swift
caiobzen: Oops! I've forgotten that one! Thanks!
Done! ✅

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: It depends.. I was assuming that the announcements service is not error-prone, as the announcementVersion wasn't supposed to come different than an incremented version if the previous one. But I do agree that we could add this check either way. I will make this change in a few 👍

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: It makes sense! I will update the places where we use it to load from disk immediately 👍

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewHostingController.swift
Ecarrion: Oh! The I think it's worth to add a comment explain why it is necessary, or we risk that a future maintainer removes it because they think it is not needed.

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewHostingController.swift
Ecarrion: Let's leave it like that for the time being!

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
Ecarrion: Yes but you can easily fix that by dissecting your `savedAnnouncement` method into two.

1. Have a function `loadStoredAnnouncement() -> StorageAnnouncement?` to load only the saved storage announcement
2. Have a function `mapStoredAnnouncement() -> Announcement` which I believe you already have
3. Modify this function to something like
```swift
func loadSavedAnnouncement(completion: Result<Announcement, Error> -> Void) {
	let storage = loadStoredAnnouncement()
	let readOnly = mapStoredAnnouncement(storage)
   onCompletion(readOnly, storage.displayed)
}
```

Does that makes sense?

caiobzen: Looks interesting! I will manage to do that in a few 👍 
Thanks for the insight!

WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewHostingController.swift
caiobzen: Makes total sense. I've added the docs!
Done! ✅

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
caiobzen: Ok, cool. I like the idea. I'll be doing this in a few!

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Done! ✅

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Done! ✅

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Done! ✅

caiobzen:
@Ecarrion I've just created a new issue to deal with the scrolling behavior that you've mentioned!
https://github.com/woocommerce/woocommerce-ios/issues/4915

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
caiobzen: Done! ✅

RELEASE-NOTES.txt
```diff
@@ -5,6 +5,7 @@
 - [**] Merchants can now edit customer provided notes from orders. [https://github.com/woocommerce/woocommerce-ios/pull/4893]
 - [*] Fix empty states sometimes not centered vertically [https://github.com/woocommerce/woocommerce-ios/pull/4890]
 - [*] Hide bottom bar on shipping label purchase form. [https://github.com/woocommerce/woocommerce-ios/pull/4902]
+- [*] Show "What's New on WooCommerce" on app launch [https://github.com/woocommerce/woocommerce-ios/pull/4908]
```
Ecarrion: As this is not(and should not) going live yet, we should not include a release note.
I think we really need to hide this behind a feature flag to be safe. Could you do that in the next or please?

WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
```diff
@@ -15,10 +15,25 @@ final class WhatsNewViewModel: ReportListPresentable {
...
         self.onDismiss = onDismiss
     }
+
+    func onAppear() {
+        stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { result in
+            switch result {
+            case .success:
+                return DDLogInfo("📣 Announcement was marked as displayed! ✅")
+            case .failure(let error):
+                return DDLogInfo("📣 Failed to mark announcement as displayed: \(error.localizedDescription)")
```
Ecarrion:
```suggestion
               DDLogError("⛔️ Failed to mark announcement as displayed: \(error.localizedDescription)")
```

WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
```diff
@@ -15,10 +15,25 @@ final class WhatsNewViewModel: ReportListPresentable {
...
         self.items = items
+        self.stores = stores
         self.onDismiss = onDismiss
     }
+
+    func onAppear() {
+        stores.dispatch(AnnouncementsAction.markSavedAnnouncementAsDisplayed(onCompletion: { result in
+            switch result {
+            case .success:
+                return DDLogInfo("📣 Announcement was marked as displayed! ✅")
```
Ecarrion: I don't think return is needed here, right?

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
```diff
@@ -49,21 +49,44 @@ final class AppCoordinator {
...
+    /// Synchronize announcements and present What's New Screen if needed
+    ///
+    func synchronizeAndShowWhatsNew() {
+        stores.dispatch(AnnouncementsAction.synchronizeAnnouncements(onCompletion: { [weak self] result in
+            guard let self = self else { return }
+            switch result {
+            case .success(let announcement):
+                DDLogInfo("📣 Announcements Synced! AppVersion: \(announcement.appVersionName) | AnnouncementVersion: \(announcement.announcementVersion)")
+            case.failure(let error):
+                DDLogInfo("⛔️ Failed to synchronize announcements: \(error.localizedDescription)")
```
Ecarrion:
```suggestion
                DDLogError("⛔️ Failed to synchronize announcements: \(error.localizedDescription)")
```

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
```diff
@@ -96,26 +130,80 @@ private extension AnnouncementsStore {
...
 
-    /// Save the `Announcement` to the appropriate file.
-    func saveAnnouncement(_ announcement: Announcement) throws {
+    /// Map `StorageAnnouncement` to `WordPressKit.Announcement` model
+    func mapStoredAnnouncement(_ storedAnnouncement: StorageAnnouncement) throws -> Announcement {
+        do {
+            let encodedObject = try JSONEncoder().encode(storedAnnouncement)
+            return try JSONDecoder().decode(Announcement.self, from: encodedObject)
+        } catch {
+            throw AnnouncementsStorageError.noAnnouncementSaved
```
Ecarrion: Shouldn't this be something like `AnnouncementsStorageError.invalidAnnouncement`?

Ecarrion (changes requested):
Almost there, just a few minor requests!

RELEASE-NOTES.txt
caiobzen: Yep, sure thing!
I just removed It from the release-notes!
Done! ✅

WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
caiobzen: Done! ✅

WooCommerce/Classes/ViewModels/WhatsNew/WhatsNewViewModel.swift
caiobzen: Not at all! Good catch! 🦅 👁️ 
Done! ✅

WooCommerce/Classes/ViewRelated/AppCoordinator.swift
caiobzen: Done! ✅

Yosemite/Yosemite/Stores/AnnouncementsStore.swift
caiobzen: Makes total sense. It must have lost context after the refactoring!
Thanks for catching this one too!

Done! ✅

Ecarrion (approved):

PR #4910, opened 2021-09-01
[Mobile Payments] [Several Readers Found] Add several readers found view controller

allendav:
Partially addresses #4333 

This large PR adds a new view controller to be used when multiple card readers have been discovered. It does NOT actually connect it to any flows. That will come in a subsequent PR. To test, see instructions below on how to temporarily add it to a convenient spot in the reader discovery flow.

Changes:
- Adds `SeveralReadersFoundViewController` with dummy data and no completion handling
- Handles light and dark mode
- Handles compact (and regular) vertical and horizontal size classes and orientation changes

Concerns:
- I focused on the light mode "several readers found" iOS (iPhone, portrait) designs in Figma. I made some guesses at what would be appropriate for landscape and for iPads in terms of the overall margins for the (modal) view. @Garance91540 @adamzelinski  - your feedback is needed
- Also, I noticed (late in the process) that the dark mode designs in Figma don't look like the light mode ones. They're unexpectedly full screen. I expect that is just an oversight. I hope the (modal) light mode designs that I worked from are the true design. That said, I could use some guidance especially on colors for dark mode. The modal and the background view all blend together as it is right now. See screenshots below.

To test:
- change your Run Scheme to enable simulated readers
- change `onFoundReader` in `CardReaderConnectionController` to something like this to see the new view controller
```
    func onFoundReader() {
        guard let candidateReader = candidateReader else {
            return
        }

        guard let from = fromController else {
            return
        }

        alerts.dismiss()

        let controller = SeveralReadersFoundViewController()
        from.present(controller, animated: true)

//        alerts.foundReader(
//            from: from,
//            name: candidateReader.id,
//            connect: {
//                self.state = .connectToReader
//            },
//            continueSearch: {
//                self.skippedReaderIDs.append(candidateReader.id)
//                self.candidateReader = nil
//                self.pruneSkippedReaders()
//                self.state = .searching
//            })
    }
```
- enter settings, in-person payments, manage card readers and begin a search if necessary
- rotate the device from portrait to landscape and vice versa and ensure the margins of the view are updated
- change from light to dark mode and vice versa and ensure the button colors update appropriately

Screenshots:

iPad Light Portrait

iPad Light Landscape

iPad Dark Portrait

iPad Dark Landscape

iPhone Light Portrait

iPhone Light Landscape

iPhone Dark Portrait

iPhone Dark Landscape

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
@@ -6,6 +6,12 @@ final class SeveralReadersFoundViewController: UIViewController, UITableViewDele
     @IBOutlet weak var tableView: UITableView!
     @IBOutlet weak var cancelButton: UIButton!
 
+
+    @IBOutlet weak var viewTrailingConstraint: NSLayoutConstraint!
+    @IBOutlet weak var viewBottomConstraint: NSLayoutConstraint!
+    @IBOutlet weak var viewTopConstraint: NSLayoutConstraint!
+    @IBOutlet weak var viewLeadingConstraint: NSLayoutConstraint!
+
     private var sections = [Section]()
 
     private var readerIDs = [String]()
@@ -36,6 +42,24 @@ final class SeveralReadersFoundViewController: UIViewController, UITableViewDele
         configureNavigation()
         configureSections()
         configureTable()
+        updateViewMargins()
+        updateViewAppearances()
+    }
+
+    /// Update constraints that vary by size class
+    ///
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        /// Handle size class and orientation
+        ///
+        updateViewMargins()
+
+        /// Handle changes to Light / Dark Appearance
+        ///
+        if let previousTraits = previousTraitCollection, previousTraits.hasDifferentColorAppearance(comparedTo: traitCollection) {
+            updateViewAppearances()
+        }
     }
 
     // TODO - accept updates to the list of CardReaderIDs to present and reloadData
@@ -75,6 +99,36 @@ private extension SeveralReadersFoundViewController {
         tableView.dataSource = self
         tableView.delegate = self
         tableView.reloadData()
+    }
+
+    /// Update the overall view's margins depending on the size classes of the device
+    /// for its present orientation
+    ///
+    private func updateViewMargins() {
+        /// Vertically constrained? Reduce top and bottom constraints
+        ///
+        if traitCollection.verticalSizeClass == .compact {
+            viewTopConstraint.constant = 47
+            viewBottomConstraint.constant = -47
+        } else {
+            viewTopConstraint.constant = 125
+            viewBottomConstraint.constant = -197
+        }
+
+        /// Horizontally unconstrained? Increase leading and trailing constraints
+        if traitCollection.horizontalSizeClass == .compact {
+            viewLeadingConstraint.constant = 47
+            viewTrailingConstraint.constant = -47
+        } else {
+            viewLeadingConstraint.constant = 160
+            viewTrailingConstraint.constant = -160
+        }
+    }
+
+    /// Update views that change appearance for light vs. dark mode
+    ///
+    private func updateViewAppearances() {
+        cancelButton.applySecondaryButtonStyle()
     }
 
     /// Register table cells.
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
@@ -67,8 +67,6 @@ final class SeveralReadersFoundViewController: UIViewController, UITableViewDele
 // MARK: - View Configuration
 //
 private extension SeveralReadersFoundViewController {
-    /// Set the title and back button.
-    ///
     func configureNavigation() {
         headlineLabel.text = Localization.headline
         cancelButton.setTitle(Localization.cancel, for: .normal)
@@ -86,6 +84,7 @@ private extension SeveralReadersFoundViewController {
             Section(rows: readerRows)
         )
 
+        // Prepare a row for our scanning indicator
         sections.append(
             Section(rows: [.scanning])
         )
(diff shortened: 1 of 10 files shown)


WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
```diff
@@ -0,0 +1,253 @@
+import UIKit
+
+final class SeveralReadersFoundViewController: UIViewController, UITableViewDelegate {
+
+    @IBOutlet weak var headlineLabel: UILabel!
```
ctarda: We tend to favour making the outlets private

WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    /// Update the overall view's margins depending on the size classes of the device
+    /// for its present orientation
+    ///
+    func updateViewMargins() {
+        /// Vertically constrained? Reduce top and bottom constraints
+        ///
+        if traitCollection.verticalSizeClass == .compact {
+            viewTopConstraint.constant = 47
```
ctarda: I would suggest declaring these numeric values in a private struct/enum

WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
```diff
@@ -0,0 +1,253 @@
...
+            viewTrailingConstraint.constant = -47
+        } else {
+            viewLeadingConstraint.constant = 160
+            viewTrailingConstraint.constant = -160
+        }
+    }
+
+    /// Update views that change appearance for light vs. dark mode
+    ///
+    func updateViewAppearances() {
```
ctarda, line 133: This should not be necessary. The styles should apply colours that should change automagically when changing the appearance.

WooCommerce/Classes/ViewRelated/ReusableViews/ActivitySpinnerAndLabelTableViewCell.swift
```diff
@@ -0,0 +1,7 @@
+import UIKit
+
+class ActivitySpinnerAndLabelTableViewCell: UITableViewCell {
```
ctarda: I know this is still WIP, but this could be final.

WooCommerce/Classes/ViewRelated/ReusableViews/LabelAndButtonTableViewCell.swift
```diff
@@ -0,0 +1,9 @@
+import UIKit
+
+class LabelAndButtonTableViewCell: UITableViewCell {
```
ctarda: This one could be final too

WooCommerce/Classes/ViewRelated/ReusableViews/LabelAndButtonTableViewCell.swift
```diff
@@ -0,0 +1,9 @@
...
+    @IBOutlet weak var label: UILabel!
+    @IBOutlet weak var button: UIButton!
+
+    // TODO: Private outlets, configure method
```
ctarda, line 7: 😛

ctarda (approved):
Code looks good. I left a few non-blocking suggestions.

It seems to work fine.

This is what I see:

ghost:
Layouts are looking really good thanks @allendav just noticed a couple of things:
- It looks like we're missing the drop shadows on the modals
- The dark mode modal looks a little dark. We use System background tertiary (#3A3A3C)

WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
allendav: Fixed in 2f51212b34e33b39cd1e7ee0b4a0c80d7372291c

WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
allendav: Fixed in 2d128a1e3ae3dc1164e9e2cafaa0e15663ce6c69

WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
allendav: Unfortunately not - for example, if I comment out the call to `updateViewAppearances` in the context of `traitCollectionDidChange` and then toggle the app from dark mode to light mode while the view is presented, the cancel button looks wrong:



So, I am not going to make a change here.

WooCommerce/Classes/ViewRelated/ReusableViews/ActivitySpinnerAndLabelTableViewCell.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4910/commits/8c373612c6e33caf4fd8227777a42d7959f0d41e

WooCommerce/Classes/ViewRelated/ReusableViews/LabelAndButtonTableViewCell.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4910/commits/8c373612c6e33caf4fd8227777a42d7959f0d41e

allendav:
@adamzelinski wrote:

> The dark mode modal looks a little dark. We use System background tertiary (#3A3A3C)

Thanks! It looks much better now:



Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4910/commits/6dfce381dc3137ca65545dccef8a73fcf98cedb4

allendav:
@adamzelinski wrote:

> It looks like we're missing the drop shadows on the modals

I looked and it looks like we have almost no drop shadows in the iOS app in general - a notable exception being the notice snackbar. I've opened a separate issue to add drop shadows to the card present payments modals:

https://github.com/woocommerce/woocommerce-ios/issues/4914

Garance91540:
Hi Allen, it's looking good, but the "Connect Reader" CTAs should be Pink 30.

allendav:
Thank you @Garance91540 ! Do you mean the "Connect" affordance in each cell? I see Pink 50 in Figma - I can change it to 30 

Screen Shot 2021-09-03 at 9 20 21 AM

Garance91540:
> Hi Allen, it's looking good, but the "Connect Reader" CTAs should be Pink 30.

Forgot to specify: it's Pink 30 in dark mode and Pink 50 in Light mode :)

allendav:
> it's Pink 30 in dark mode and Pink 50 in Light mode :)

Ah! Can do!

Issue #4913, opened 2021-09-02
Shipping Labels: Package Selected modal dismissed on device rotation


rachelmcr:
**Describe the bug**

As reported in https://github.com/woocommerce/woocommerce-ios/pull/4905#pullrequestreview-745152629:

In the shipping label package details screen, tapping Package Selected opens a modal view. Rotating the device from portrait to landscape dismisses the modal. (Rotating from landscape to portrait doesn't dismiss the modal.)

This also affects the feature-flagged Add New Package screen, which you can navigate to from the Package Selected screen if the app is run in debug mode.

Note: I can't reproduce this in a simulator, only on device (tested on iPhone 12 Mini, iOS 14.7.1 in WCiOS 7.4 beta).

**To Reproduce**

1. Make sure you have configured the WooCommerce WCShip plugin, and you have one or more packages set up in WP Admin under WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Launch the app in debug mode (so the feature flag is enabled).
3. Open an order detail that is eligible for the shipping label.
4. Tap the button "Create Shipping Label.”
5. Confirm "Ship From" and "Ship To.”
6. In the package details section, tap on “Package Selected.”
7. On the Package Selected screen, rotate the device and notice the modal closes so you're back on the Package Details screen.

**Screenshots**

<img src="https://user-images.githubusercontent.com/8658164/131887488-6013b942-9ba9-4029-a210-ac94c0fef33d.gif" width="300px">

itsmeichigo:
@rachelmcr @pmusolino  I can't seem to reproduce this issue on the new Package Details screen, can we close it then?

pmusolino:
If it's no more reproducible, yes, feel free to close it 🙌

rachelmcr:
Yes, I was testing this quite a bit at the end of last week and couldn't reproduce it anymore, either. I'll go ahead and close it now 👍

PR #4916, opened 2021-09-02
Shipping Labels: Ensure Add New Package view retains state

rachelmcr:
Part of: #4743 / #4744

## Description

In https://github.com/woocommerce/woocommerce-ios/pull/4882#discussion_r698176446 @itsmeichigo noted that switching between tabs on the Add New Package screen in the Shipping Label flow caused the tabs to lose state. (The Custom Package tab lost any data entered in the form, and the Service Package tab lost its package selection.)

This fixes the way those views and their view models are initialized so that they retain state. You should now be able to perform any action on those screens (e.g. switch tabs, rotate the device) and the changes on each tab will be retained.

~Note: I didn't intentionally address the issue reported in #4913, where the modal closes when rotating the device from portrait to landscape. However, in testing I found it is no longer an issue as of these changes.~ → As noted in the discussion below, this is still an issue.

## Changes

* Initializes the custom package and service package view models as `StateObject`s in the `ShippingLabelAddNewPackage` view. This ensures that each view model only has a single instance, even if the views are redrawn.
* Removes the use of `onAppear` to set the value of `packagesResponse` in the service package view model (since that is called every time the view is redrawn), and uses `StateObject(wrappedValue:)` to initialize that view model instead. Although the [Apple documentation](https://developer.apple.com/documentation/swiftui/stateobject/init(wrappedvalue:)) says not to call this initializer directly, the SwiftUI digital lounge at WWDC21 [clarified that this is an acceptable use](https://swiftui-lab.com/random-lessons/#data-10).

**State retained when switching tabs**

**Modal stays open when rotating device**

## Testing

8. On the Add New Package screen, switch between the "Custom Package" and "Service Package" tabs. Enter custom package details in the form and select a service package from the list, and confirm your changes are retained as you switch between tabs or rotate your device.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
@@ -4,6 +4,7 @@ import Yosemite
 struct ShippingLabelAddNewPackage: View {
     @StateObject private var viewModel = ShippingLabelAddNewPackageViewModel()
     @StateObject private var customPackageVM = ShippingLabelCustomPackageFormViewModel()
+    @StateObject private var servicePackageVM = ShippingLabelServicePackageListViewModel()
     let packagesResponse: ShippingLabelPackagesResponse?
 
     var body: some View {
@@ -21,7 +22,10 @@ struct ShippingLabelAddNewPackage: View {
                     case .customPackage:
                         ShippingLabelCustomPackageForm(viewModel: customPackageVM, safeAreaInsets: geometry.safeAreaInsets)
                     case .servicePackage:
-                        ShippingLabelServicePackageList(packagesResponse: packagesResponse, geometry: geometry)
+                        ShippingLabelServicePackageList(viewModel: servicePackageVM, geometry: geometry)
+                            .onAppear() {
+                                servicePackageVM.packagesResponse = packagesResponse
+                            }
                     }
                 }
                  .background(Color(.listBackground).ignoresSafeArea(.container, edges: .bottom))
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
@@ -3,6 +3,7 @@ import Yosemite
 
 struct ShippingLabelAddNewPackage: View {
     @StateObject private var viewModel = ShippingLabelAddNewPackageViewModel()
+    @StateObject private var customPackageVM = ShippingLabelCustomPackageFormViewModel()
     let packagesResponse: ShippingLabelPackagesResponse?
 
     var body: some View {
@@ -18,7 +19,7 @@ struct ShippingLabelAddNewPackage: View {
                 ScrollView {
                     switch viewModel.selectedView {
                     case .customPackage:
-                        ShippingLabelCustomPackageForm(safeAreaInsets: geometry.safeAreaInsets)
+                        ShippingLabelCustomPackageForm(viewModel: customPackageVM, safeAreaInsets: geometry.safeAreaInsets)
                     case .servicePackage:
                         ShippingLabelServicePackageList(packagesResponse: packagesResponse, geometry: geometry)
                     }
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
@@ -2,7 +2,7 @@ import SwiftUI
 import Yosemite
 
 struct ShippingLabelAddNewPackage: View {
-    @StateObject var viewModel: ShippingLabelAddNewPackageViewModel
+    @ObservedObject var viewModel: ShippingLabelAddNewPackageViewModel
 
     var body: some View {
         GeometryReader { geometry in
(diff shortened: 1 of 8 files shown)


itsmeichigo (commented):
Thank you @rachelmcr for fixing this! 👍

However I noticed when testing this PR that issue #4913 still happens, although not consistently. I think we should look at the issue more closely to find the root cause.

The states of the Custom Package and Service Package views are retained properly 👏 But I'm not sure I like the use of the `StateObject` initializer, since I still don't quite understand the explanation in the SwiftUI lounge that you mentioned.

If it were me in this situation, I'd make the main view model of `ShippingLabelAddNewPackage` an `ObservedObject`, which accepts `ShippingLabelPackagesResponse` in its initializer. The view model should then handle the creation of the child views' view models, and the view should then retrieve them from the view model for creating child views.
```swift
final class ShippingLabelAddNewPackageViewModel: ObservableObject {
    let customPackageVM = ShippingLabelCustomPackageFormViewModel()
    let servicePackageVM: ShippingLabelServicePackageListViewModel

    init(_ selectedIndex: Int = PackageViewType.customPackage.rawValue,
         packagesResponse: ShippingLabelPackagesResponse?) {
        self.selectedIndex = selectedIndex
        self.servicePackageVM = ShippingLabelServicePackageListViewModel(packagesResponse: packagesResponse)
    }
}
``` 
This way we can avoid the complications of `StateObject` 😅 But I'd love to hear @pmusolino's opinion on this also so I'll just leave my comment here.

rachelmcr:
> However I noticed when testing this PR that issue #4913 still happens, although not consistently. I think we should look at the issue more closely to find the root cause.

Thanks for testing that! I must be missing something when I test it; I've removed it from the PR description here.

> If it were me in this situation, I'd make the main view model of ShippingLabelAddNewPackage an ObservedObject, which accepts ShippingLabelPackagesResponse in its initializer. The view model should then handle the creation of the child views' view models, and the view should then retrieve them from the view model for creating child views.

Ah yeah, it makes sense to have the main view model of `ShippingLabelAddNewPackage` create the child views' view models. That avoids the question of whether we should use the `StateObject` initializer like this for those child views. We still need to the main view model to be a `StateObject` (to retain the state when the device is rotated) but I think that should be fine. I'll work on that change. 👍

jkmassel:
It's code freeze day, so I'm bumping this to `7.6`. Because it's a bug fix, we may want it to be in `7.6` after all – if so, please merge this PR into `release/7.6` and DM me once it's complete – I'll be happy to cut a new beta release!

itsmeichigo (commented):
@rachelmcr I don't see a new request for review, but I assume this PR is ready for another look?

I've been thinking a lot about when to use `StateObject` and `ObservedObject`, and although your new changes work as expected, I still have a bit concern, so I'm leaving a comment below.

Besides, I've also tested again for issue #4913 and it seems that I cannot reproduce it again. I've looked through the code and can't seem to find out how the issue can be caused, so _maybe_ it was me who got it wrong earlier? Feel free to disregard this message if you think it's best to keep the issue to observe it later.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
```diff
@@ -2,8 +2,7 @@ import SwiftUI
 import Yosemite
 
 struct ShippingLabelAddNewPackage: View {
-    @StateObject private var viewModel = ShippingLabelAddNewPackageViewModel()
-    let packagesResponse: ShippingLabelPackagesResponse?
+    @StateObject var viewModel: ShippingLabelAddNewPackageViewModel
```
itsmeichigo: I'm not quite comfortable with using `@StateObject` here, although it helps persisting the state of the the object. I've found [this article](https://www.donnywals.com/whats-the-difference-between-stateobject-and-observedobject/), and I'm interested in this sentence:

> If you'd have marked Counter.dataSource as @StateObject the new instance would be ignored and your app might now have a new hidden bug.

So the takeaway is - like `@State` - `@StateObject` should only be used if the object is created and owned by the view itself. `@State` and `@StateObject` should only be used to persist the state when the view is **redrawn**, not when it's **created**.

In our case, when rotating the device, `ShippingLabelPackageList` view is redrawn, which causes `ShippingLabelAddNewPackage` to be recreated. So we want to persist `ShippingLabelAddNewPackage`'s last state even after it's created, and we should make `ShippingLabelPackageDetailsViewModel` hold on to its state like so:
```Swift
final class ShippingLabelPackageDetailsViewModel: ObservableObject {
    lazy var addNewViewModel: ShippingLabelAddNewPackageViewModel = ShippingLabelAddNewPackageViewModel(packagesResponse: packagesResponse)
}
```

And we use this `addNewViewModel` when creating the `ShippingLabelAddNewPackage` view instead of creating a new view model every time:
```Swift
NavigationLink(destination: ShippingLabelAddNewPackage(viewModel: viewModel.addNewViewModel),
               isActive: $isShowingNewPackageCreation) {
    EmptyView()
}
```

We can then make the view model an `@ObservedObject` in `ShippingLabelAddNewPackage` and still keep its state when it's created again. That way we delegate the view's state persistance to its parent view, while avoiding incorrect use of `@StateObject`. What do you think - does that make sense?

rachelmcr: Thanks for this explanation, @itsmeichigo!

> So the takeaway is - like @State - @StateObject should only be used if the object is created and owned by the view itself. @State and @StateObject should only be used to persist the state when the view is redrawn, not when it's created.

I remember reading about this earlier but I had lost sight of it in the process of trying to get everything to work with these views. Thanks for the reminder! With that mind then definitely it makes sense to use `ShippingLabelPackageDetailsViewModel` to hold the view model for `ShippingLabelAddNewPackage`. I'll work on that and take a look through to make sure I'm not misusing `StateObject` anywhere else.

rachelmcr:
Ok @itsmeichigo this is ready for another look. Hopefully I've got it now! 🤞 😄 

> Besides, I've also tested again for issue #4913 and it seems that I cannot reproduce it again. I've looked through the code and can't seem to find out how the issue can be caused, so maybe it was me who got it wrong earlier? Feel free to disregard this message if you think it's best to keep the issue to observe it later.

I've tested #4913 so many different times in different conditions that I don't trust my results. :D Since the issue isn't really related to the changes here, my thought is to leave the issue open for now; we can always close it later if we still can't reproduce it when the rest of the flow is complete.

itsmeichigo (approved):
Thank you @rachelmcr for the update! I believe everything's good now! :shipit: 

On a side note, I was just be able to reproduce #4913 again in another order 😥, so it's best to keep it open for now!

Issue #4779, opened 2021-08-11
Order Editing: Country & State UI


Ecarrion:
<img width="246" alt="Screen Shot 2021-08-11 at 10 17 56 AM" src="https://user-images.githubusercontent.com/562080/129056645-ae3c9e5c-60ff-4088-9752-3d5afab18d33.png">

Tasks: 

- [ ] Build UI or reuse from Shipping Labels Feature
- [ ] Connect from country & state rows from #4778

PR #4917, opened 2021-09-03
Order Editing: Adds Country Selector UI

Ecarrion:
part of #4779 

# Why

While editing an address, merchants require the ability to edit the address country.

# How

- Added a `SwiftUI` wrapper around `ListSelectorViewController`

- Created `CountrySelector` which is a view that holds the `ListSelector` + a search header.

- Added a `CountrySelectorCommand` to populate the `ListSelector`. This command is currently populated with all available Regions/Countries, this is temporary and will be replaced in a subsequent PR.

- Added navigation from `EditAddressForm` to `CountrySelector`

# Screenshots

Normal | Dark | Big
---- | ---- | ----
normal | dark | big-size

# Testing Steps

- Launch the app and navigate to an order
- Tap on the edit icon from the shipping address row
- Tap on the country row
- See the list of countries.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
@@ -1,6 +1,7 @@
 import Foundation
 import Combine
 import SwiftUI
+import UIKit
 
 /// Hosting controller that wraps an `EditAddressForm`.
 ///
@@ -18,6 +19,9 @@ final class EditAddressHostingController: UIHostingController<EditAddressForm> {
 /// Allows merchant to edit the customer provided address of an order.
 ///
 struct EditAddressForm: View {
+
+    @State var showEditCountry = false
+
     var body: some View {
         GeometryReader { geometry in
             ScrollView {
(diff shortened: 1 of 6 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
```diff
@@ -18,6 +19,11 @@ final class EditAddressHostingController: UIHostingController<EditAddressForm> {
 /// Allows merchant to edit the customer provided address of an order.
 ///
 struct EditAddressForm: View {
+
+    /// Set it to `true` to present the country selector.
+    ///
+    @State var showCountrySelector = false
```
ealeksandrov, line 25: nit: can this be `private`?

ealeksandrov (approved):
LGTM 🚀

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
Ecarrion: For sure!

PR #4869, opened 2021-08-24
[Order Editing] Add edit buttons to shipping and billing addresses

ealeksandrov:
Closes https://github.com/woocommerce/woocommerce-ios/issues/4768.

## Description

This PR adds 2 edit buttons on shipping and billing addresses. They are hidden behind feature flag.

## How

Shipping Address:
- Reused `orderEditingEnabled` from data source to hide changes behind feature flag.
- Added action handler similar to "customer note" edit action.
- In `CustomerInfoTableViewCell` added `editButton` in a UIStackView (since it already used StackView).

Billing Address:
- Added `editingEnabled` init parameter on `BillingInformationViewController` to pass `orderEditingEnabled` flag in `OrderDetailsViewModel`. It won't be enabled if `BillingInformationViewController` initialized from `ReviewOrderViewController`.
- `orderEditingEnabled` property of `OrderDetailsDataSource` is internal to enable access from `OrderDetailsViewModel`.
- To keep edit button in line with other icons sizes on `BillingInformationViewController` - it's implemented with manual constraint manipulations, without StackView. The icon size is 20, but the button is still 44x44 with help of `contentEdgeInsets`. To align 44 button with 20 icons in other cells it has -12 trailing constraint.

## Screenshots

No shipping address|Existing shipping address|Billing address
--|--|--
Simulator Screen Shot|Simulator Screen Shot|Simulator Screen Shot

## Test

1. Open existing order
2. Check button on shipping address field, verify it prints to console on tap
3. Tap "View Billing Information"
4. Check button on billing address field, verify it prints to console on tap
5. Go back to order details, tap "Mark Order Complete".
6. Validate that shipping address and billing address (inside billing info subscreen) don't have edit buttons.
7. Disable feature flag and verify that above buttons don't appear.

---
Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4918, opened 2021-09-03
Order Edit: Fix shipping address row background

Ecarrion:
# Why

In a recent PR https://github.com/woocommerce/woocommerce-ios/pull/4869 we introduced edit indicators in the shipping address row and by accident we introduced a small glitch in dark mode.

error

# How

The fix is to set one of the subviews in from that cell to have transparent background, so it doesn't mess with the cell background color.

# Screenshots

good

Update release notes:


ealeksandrov (approved):
LGTM, thanks for fixing it :shipit:

PR #4921, opened 2021-09-03
[Mobile Payments] Don't allow mobile payments in orders with subscriptions

### LINKED ISSUE #4426: [Mobile Payments] Marking orders containing subscriptions or recurring services as non eligible for In-Person Payments
We want to avoid allowing accidentally collecting in person payments for orders that contain subscriptions or recurring payments

koke:
Fixes #4426 

To match what we do on Android, this prevents orders with any subscription to show the option to collect payment.

## To test

From the web:

1. Ensure your site has the Subscriptions extension added
2. Add a product with type subscription. Make sure you don't have a free trial, or the order will be marked as completed
3. Go to WooCommerce > Settings > Subscriptions and make sure "Accept Manual Renewals" is enabled. Otherwise, you won't be able to select Cash on Delivery as the payment method
4. Make an order for that product, using Cash on Delivery as the payment method

From the app:

1. Find the order in the orders tab, make sure the Collect Payment button doesn't show
2. Check that the Collect Payment button shows in other orders that don't have a subscription

Update release notes:

### DIFF
--- a/Networking/Networking/Model/Product/ProductType.swift
+++ b/Networking/Networking/Model/Product/ProductType.swift
@@ -8,6 +8,7 @@ public enum ProductType: Codable, Hashable, GeneratedFakeable {
     case grouped
     case affiliate
     case variable
+    case subscription
     case custom(String) // in case there are extensions modifying product types
 }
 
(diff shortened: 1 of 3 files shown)


ctarda (approved):
Tested on an order with a subscription product, and I didn't see the "collect payment" button.

:shipit:

PR #4925, opened 2021-09-03
[What's New Component] Add WhatsNewOnWooCommerce feature flag

caiobzen:
### Description 📜 
We must be able to turn the What's New on WooCommerce feature according to a feature flag until we finish its development.

### Main Changes ⚙️
- Add new FeatureFlag `.whatsNewOnWooCommerce` and check for it being enabled before fetching the service right away on app launch

Update release notes:


Ecarrion (approved):
👍

PR #4927, opened 2021-09-03
[Mobile Payments] Set eligibility for In-Person Payments based also on currency: USD

### LINKED ISSUE #4926: [Mobile Payments] Add check for USD
In-Person Payments are only supported in the US so far, so we should check that the order currency is USD.

For context:pdfdoF-6f-p2#comment-283

ctarda:
Closes: #4926 

As discussed in pdfdoF-6f, we want to make sure that we take currency into account for eligibility for In-Person Payments.

| Before | After |
| ---- | ---- |
|  |  |

## Changes
* Verify that the order currency is USD
* Add tests

## How to test
* Navigate to an order eligible for In-Person Payments. If the order currency is USD, the "Collect Payment" button should be visible. If the order currency is not USD, the "Collect Payment" button should not be visible.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -46,6 +46,7 @@ final class OrderDetailsDataSource: NSObject {
     ///
     var isEligibleForCardPresentPayment: Bool {
         return isOrderAmountEligibleForCardPayment() &&
+            isOrderCurrencyEligibleForCardPayment() &&
             isOrderStatusEligibleForCardPayment() &&
             isOrderPaymentMethodEligibleForCardPayment() &&
             hasCardPresentEligiblePaymentGatewayAccount() &&
@@ -1465,6 +1466,10 @@ private extension OrderDetailsDataSource {
         // * in general, all orders where we might want to capture a payment for less than the total order amount
         let paymentViewModel = OrderPaymentDetailsViewModel(order: order)
         return !paymentViewModel.hasBeenPartiallyCharged
+    }
+
+    func isOrderCurrencyEligibleForCardPayment() -> Bool {
+        CurrencySettings.CurrencyCode(rawValue: order.currency) == .USD
     }
 
     func isOrderStatusEligibleForCardPayment() -> Bool {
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -1467,6 +1468,10 @@ private extension OrderDetailsDataSource {
         return !paymentViewModel.hasBeenPartiallyCharged
     }
 
+    func isOrderCurrencyEligibleForCardPayment() -> Bool {
+        CurrencySettings.CurrencyCode(rawValue: order.currency) == .USD
```
allendav, line 1472: This will work for now, but we'll need the business folks to decide how this will work for non US merchants when we expand there. Please consider opening an issue for "how to determine supported currencies for in-person payments for non-US merchants".

Notably, supported currencies are passed in the account response. See https://github.com/woocommerce/woocommerce-ios/blob/beaf197f6a326020cf83a747d53e989f2d862ae4/Networking/Networking/Model/WCPayAccount.swift#L16

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/OrderDetailsDataSourceTests.swift
```diff
@@ -259,6 +259,50 @@ final class OrderDetailsDataSourceTests: XCTestCase {
...
+        storageManager.viewStorage.saveIfNeeded()
+    }
+
+    func test_collect_payment_button_is_not_visible_if_order_currency_is_not_usd() throws {
+        // Setup
+        let account = storageManager.insertCardPresentEligibleAccount()
+        storageManager.viewStorage.saveIfNeeded()
+
+        // Given
+        let order = makeOrder().copy(status: .processing, currency: "cad", datePaid: .some(nil), total: "100", paymentMethodID: "cod")
```
allendav, line 290: :poutine:

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
ctarda: Sure #4928

allendav (approved):
:shipit: Tested with a USD order and an EUR order. Only saw the collect payment button on the USD order.

PR #4929, opened 2021-09-03
[What's New Component] Add What's New on WooCommerce to app settings

caiobzen:
Part of: #4863 

## Description 📜
Add What's New on WooCommerce to the app settings.
This feature is currently under a feature flag! 🏁

## Main Changes ⚙️ 
- Add a new row in the "About the App" section (if the Feature Flag is on)
- Navigate to What's new on WooCommerce

## Evidences 📱

|           iPhone        |           iPad            |
|------------------|------------------|
| <video src="https://user-images.githubusercontent.com/997521/132043911-90607bc2-1759-4856-8b8a-7396a8c75fe1.mp4  ">            |      <video src="https://user-images.githubusercontent.com/997521/132044225-75e227c0-a6f9-421f-b4e0-e6290c6e2f0c.mp4"> |

## How to test 🧪🧑‍🔬
_Please do the following change_
image

### Scenario
1. Run the app
2. Go to settings
3. "Tap on What's New on WooCommerce"
4. What's New Component should be displayed

## Sidenote:
- Currently, if there are no announcements to show, the tap on "What's New in WooCommerce" will do nothing. cc @Ecarrion 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
@@ -303,7 +303,11 @@ private extension SettingsViewController {
         sections.append(Section(title: appSettingsTitle, rows: [.privacy], footerHeight: UITableView.automaticDimension))
 
         // About the App
-        sections.append(Section(title: aboutTheAppTitle, rows: [.about, .licenses], footerHeight: UITableView.automaticDimension))
+        if shouldShowWhatsNew() {
+            sections.append(Section(title: aboutTheAppTitle, rows: [.about, .whatsNew, .licenses], footerHeight: UITableView.automaticDimension))
+        } else {
+            sections.append(Section(title: aboutTheAppTitle, rows: [.about, .licenses], footerHeight: UITableView.automaticDimension))
+        }
 
         // Other
         #if DEBUG
@@ -351,6 +355,8 @@ private extension SettingsViewController {
             configureWormholy(cell: cell)
         case let cell as BasicTableViewCell where row == .logout:
             configureLogout(cell: cell)
+        case let cell as BasicTableViewCell where row == .whatsNew:
+            configureWhatsNew(cell: cell)
         default:
             fatalError()
         }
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -459,6 +471,10 @@ private extension SettingsViewController {
     func shouldShowPluginsSection() -> Bool {
         ServiceLocator.stores.sessionManager.defaultRoles.contains(.administrator)
     }
+
+    func shouldShowWhatsNew() -> Bool {
+        ServiceLocator.featureFlagService.isFeatureFlagEnabled(.whatsNewOnWooCommerce)
+    }
```
Ecarrion: I think we did decide to not show the row if there isn't any announcement to display right?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
```diff
@@ -574,6 +590,20 @@ private extension SettingsViewController {
...
+            guard let self = self else { return }
+            guard let (announcement, _) = try? result.get() else {
+                return DDLogInfo("📣 There are no announcements to show!")
+            }
+
+            let viewController = WhatsNewFactory.whatsNew(announcement) { [weak self] in
+                self?.dismiss(animated: true)
+            }
+            self.present(viewController, animated: true, completion: nil)
+        }))
```
Ecarrion: I think this could show an old announcement from a past app version right?
I'm sure some app versions won't have announcements because they don't have any significant user facing change.

Ecarrion (changes requested):

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
caiobzen: You're right! I forgot this requirement. I've added a check to verify if there's an announcement to show.
Done! ✅

WooCommerce/Classes/ViewRelated/Dashboard/Settings/SettingsViewController.swift
caiobzen: Yea, I thought for a second that we were going to always display the latest announcement regardless of the app version. But it may confuse users if for example we remove a feature and tell users that this is "new on WooCommerce". 
I've added a check for the app version! Thanks for the head's up!

Done! ✅

Ecarrion (approved):
Works and tests good! 👍

PR #4931, opened 2021-09-03
[Mobile Payments] [Several Readers Found] Integrate choose-a-reader list modal

allendav:
Partially addresses #4333 

Changes:
- Integrates the `SeveralReadersFoundViewController` into `CardReaderSettingsAlerts`
- Disables the inter-modal animations from the previous PR to be less jumpy when transitioning from single-reader-found to several - cc @Garance91540 @adamzelinski - thoughts? (see movie below)
- Adds a new state to `CardReaderConnectionController` for when multiple readers have been found
- Adds a new flag to `CardReaderConnectionController` to avoid switching out of the list view if the number of found readers drops below 2 during a particular searching flow
- Integrates the new `foundSeveralReaders` alert into `CardReaderConnectionController`
- Connects the connect buttons and cancel button in `SeveralReadersFoundViewController` to do their work
- Updates the `ActivitySpinnerAndLabelTableViewCell` to keep it spinning after re-use

What's next:
- The next (and last) PR for this issue will remove the feature flag and launch the feature for all users

To Test:

- First, set simulator to true in your Run Scheme and then run the app
- Tap on the gear icon, then In-Person Payments, then Manage card reader
- If you have a known reader, search will begin automatically. If not, tap the Connect Card Reader button
- A searching modal should display, possibly quite briefly
- You’ll then get a “Do you want to connect to reader CHB...” modal. Tap on Keep Searching.
- You’ll get the “Several readers found” list modal.
- Tap the cancel button and ensure it cancels search, and that you go back to the “Connect your card reader” screen and that nothing further happens (e.g. wait for at least 10 seconds)
- Tap the Connect Card Reader button to initiate another search
- A searching modal should display, possibly quite briefly
- You’ll then get a “Do you want to connect to reader CHB...” modal. Tap on Keep Searching.
- You’ll get the “Several readers found” list modal.
- With the simulator, you may initially see two entries (one for CHB20....1 and one for CHB20...2) but that quickly will just show the CHB20...2 reader
- Tap on Connect on the CHB20...2 reader’s row
- Ensure you are connected to the CHB20...2 reader
- Next, repeat the tests but with multiple physical readers
- Next, repeat the tests with the collect-payment flow (be sure to disconnect from each known reader in settings first if you want to avoid automatic known reader connection)
- Next, disable the feature flag, and retest ensuring that multiple found readers are presented one at a time during discovery (to disable the feature flag easily, add a return at the beginning of updateShowSeveralFoundReaders)
- Lastly, ensure all unit tests pass

Screengrabs / movies:

@Garance91540 : I have updated both of these movies today, Sep 10. In both cases I had to tap the "Keep Searching" button to continue on to the several-readers-found list view.

https://user-images.githubusercontent.com/1595739/132927754-9d7a87f2-9655-4b9d-a10d-a43699e9ac78.MP4

https://user-images.githubusercontent.com/1595739/132927778-072a6982-516d-404b-a89a-bf64f16ec0e8.MP4

Update release notes:


WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -191,11 +233,29 @@ private extension CardReaderConnectionController {
...
+
+                /// If the found-several-readers view is already presenting, update its list of found readers
+                ///
+                if case .foundSeveralReaders = self.state {
+                    self.alerts.updateSeveralReadersList(readerIDs: self.getFoundReaderIDs())
+                }
+
+                /// If we should switch from a single found reader prompt to several, do so now
+                ///
+                if case .foundReader = self.state {
```
ctarda: I know this is irrelevant, but I am never going to be able to parse this syntax

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -209,11 +269,19 @@ private extension CardReaderConnectionController {
                     return
                 }
 
+                /// If we have found multiple readers, advance to foundMultipleReaders
+                ///
+                if self.foundReaders.count > 1 {
```
ctarda: A bit inspired by Rails... I'm dead serious here: would it make sense to add a `containsMoreThanOne` property to Array+Helpers? 

I know it's forcing the metaphor a bit, but we are doing this `if count > 1` thing in several places and it could be a nice bit of syntactic sugar

WooCommerce/Classes/ViewRelated/CardPresentPayments/SeveralReadersFoundViewController.swift
```diff
@@ -156,16 +161,21 @@ private extension SeveralReadersFoundViewController {
...
+        )
         cell.selectionStyle = .none
     }
 
     func configureScanningRow(cell: UITableViewCell) {
         guard let cell = cell as? ActivitySpinnerAndLabelTableViewCell else {
             return
         }
-        cell.label.text = Localization.scanningLabel
+        cell.configure(labelText: Localization.scanningLabel)
```
ctarda, line 178: Better 😄

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsAlerts.swift
```diff
@@ -32,14 +33,62 @@ final class CardReaderSettingsAlerts: CardReaderSettingsAlertsProvider {
...
+        dismissCommon(animated: false)
+
+        severalFoundController = SeveralReadersFoundViewController()
+        severalFoundController?.configureController(
+            readerIDs: readerIDs,
+            connect: connect,
+            cancelSearch: cancelSearch
+        )
+
+        guard let severalFoundController = severalFoundController else {
```
ctarda: Very swift 😉

WooCommerce/Classes/ViewRelated/ReusableViews/ActivitySpinnerAndLabelTableViewCell.swift
```diff
@@ -1,7 +1,18 @@
 import UIKit
 
 final class ActivitySpinnerAndLabelTableViewCell: UITableViewCell {
-    @IBOutlet weak var label: UILabel!
+    @IBOutlet private weak var label: UILabel!
```
ctarda, line 4: Nice. Private all the things!

ctarda (approved):
It tests well and the code looks great as well.

:shipit:

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
allendav: I have to look it up myself

ghost:
Looking really good to me from a flow point of view @allendav thanks for all your hard work on this.

I did notice a couple of interactions that might need a little tweaking. Let me know if you agree @Garance91540 

I think the secondary copy here should go across two lines if possible just so it's not right to the modal's edge.
Screen Shot 2021-09-08 at 3 49 00 pm


When this screen dismisses it looks like the dark overlay and the modal itself moves down off the screen. Is this the correct interaction? I think it should just disappear from its initial location.
Screen Shot 2021-09-08 at 3 49 26 pm


This modal then looks like it comes up from the bottom. Maybe again it could appear in place. Also, may we please vertically center this information within the bounds. it looks a little bottom-heavy.
Screen Shot 2021-09-08 at 3 49 35 pm


Thanks

Garance91540:
Hi there, 

yes, all transitions should appear in place in the modals, like on Android:


https://user-images.githubusercontent.com/28450084/132529727-60f630d8-f0b2-4bd1-8608-83b5246436c6.mp4

Garance91540 (changes requested):
Regarding the flow, when several readers are detected, we shouldn't be displaying the modal "do you want to connect to reader x". Rather, after "scanning for reader", we should display right away "Several readers found"

allendav:
@Garance91540 wrote:

> Regarding the flow, when several readers are detected, we shouldn't be displaying the modal "do you want to connect to reader x". Rather, after "scanning for reader", we should display right away "Several readers found"

I wish we could, but the discovery of readers happens sequentially. We will get notified of one reader and then, a few seconds later, a second notification this time with both readers.

I see two ways to avoid this:

First, we stop using the "do you want to connect to reader x" single reader modal entirely and just always use the list based one. It can show a single reader as capably and actionably as several.

Second, and I do NOT recommend this, we could wait something like 5 seconds before we present our findings - this would give time for additional readers (if present) to be discovered. Unfortunately, this will negatively impact the experience for all of our users (especially since many will only have one reader powered on and nearby during discovery)

What do you think @Garance91540  ?

**Edit: Edit: Let's do Option 3: Let's merge this PR as is and we can open another issue to implement any change**

allendav:
@adamzelinski - thanks for the feedback! These comments largely fall outside of the code changes for this PR, so I've opened separate issues for them

The "two lines" prompt: 

https://github.com/woocommerce/woocommerce-ios/issues/4954

and a separate issue for our modals presentation animations:

https://github.com/woocommerce/woocommerce-ios/issues/4955

and an issue to better balance the modal content:

https://github.com/woocommerce/woocommerce-ios/issues/4956

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
allendav: Fixed in two places in https://github.com/woocommerce/woocommerce-ios/pull/4931/commits/436a42b73dec32e5ee558fe0a14e9a987806b454

Garance91540:
Hi @allendav , 

> I wish we could, but the discovery of readers happens sequentially. We will get notified of one reader and then, a few seconds later, a second notification this time with both readers.

How do we handle this on Android? -- cc @malinajirka ?

> First, we stop using the "do you want to connect to reader x" single reader modal entirely and just always use the list based one. It can show a single reader as capably and actionably as several.

I believe that would be better for merchants with multiple readers, but I think that's still would be the minority of our users. So I'd rather keep things as they are for a better experience for the majority of them. 

So that means:
1. scan for reader
2. do you want to connect to reader x? 
3. keep searching > open multiple readers found modal

thanks!

Garance91540 (approved):

Garance91540:
Hi @allendav 

> These comments largely fall outside of the code changes for this PR, so I've opened separate issues for them

To confirm, did you also include my comment above (making all transitions in modals like on Android) in the PRs you created?

allendav:
@Garance91540 wrote:

> So that means:
> scan for reader
> do you want to connect to reader x?
> keep searching > open multiple readers found modal
> thanks!

Sounds good

allendav:
> do you want to connect to reader x?
> keep searching > open multiple readers found modal

@Garance91540 : Done! I've made the changes to the logic to require the user to tap "keep searching" before they are advanced to the list of readers view.

I recorded these movies today, Sep 10 with those changes - let me know if this works (we have separate issues to address the backgrounds and presentation animation)

https://user-images.githubusercontent.com/1595739/132927754-9d7a87f2-9655-4b9d-a10d-a43699e9ac78.MP4


https://user-images.githubusercontent.com/1595739/132927778-072a6982-516d-404b-a89a-bf64f16ec0e8.MP4

Garance91540:
Looking good @allendav !!

malinajirka:
> > I wish we could, but the discovery of readers happens sequentially. We will get notified of one reader and then, a few seconds later, a second notification this time with both readers.

> How do we handle this on Android? -- cc @malinajirka ?

We discussed this during the implementation on Android and decided to take the easiest path since we are not even sure how common "multiple readers" scenario will be. The Android app shows "single reader found" view when the SDK returns a single reader, but when the SDK returns multiple readers after some delay we replace it with multiple-readers-found view (under user's hands). We couldn't test this scenario since we don't have multiple readers, but the single reader is usually found in < 1s after starting the discovery.

PR #4932, opened 2021-09-03
Order Editing: Add filter functionality to Country Selector UI

Ecarrion:
# Why

The previous PR added the country selector UI #4917, this PR wires that UI so that the search term entered by the user actually filters the content on the list.

# How

- Updated `ListSelector` view wrapper to reload the table view data each time SwiftUI is updating the underlying view.

- Update `CountrySelectorCommand` to allow filtering its data using an insensitive case contains.

- Added `CountrySelectorViewModel`. Right now it serves as a store for the search term and the command. Updating the search term also updates the command.

# Demo

https://user-images.githubusercontent.com/562080/132072251-ced378b8-f173-45f2-9840-bc8fface60cd.mov

# Testing Steps

- Launch the app and navigate to an order.
- Tap on the shipping address edit icon.
- Tap on the country row
- Search for countries/regions by entering a filter search term.

Update release notes:


ealeksandrov (approved):
Works well :shipit:

Issue #4599, opened 2021-07-13
Multi-package support in Shipping Labels


pmusolino:
We need to add the ability to move order’s items to different packages or ship them in their original packaging
Make sure to follow the release order: #4595

Subtasks (feel free to split them in more subtasks):
- [x] Update print screen to support multiple packages https://github.com/woocommerce/woocommerce-ios/issues/4714
- [x] Update Shipping Label order summary to support multiple packages https://github.com/woocommerce/woocommerce-ios/issues/4715
- [x] Update carriers step to accept multi packages https://github.com/woocommerce/woocommerce-ios/issues/4716
- [x] Update the customs form to accept multiple packages https://github.com/woocommerce/woocommerce-ios/issues/4718
- [x] Allow moving items to new packages or shipping them in their original packaging https://github.com/woocommerce/woocommerce-ios/issues/4717
- [x] If shipping in original packaging is selected, and the product dimensions are not defined, then show an error.
- [x] Update purchase form to support multi packages
- [x] Update Package Details screen to support multi packages
- [x] Tracks events https://github.com/woocommerce/woocommerce-ios/issues/5172
- [x] Enable the feature flag for all our merchants.

![Schermata 2021-07-13 alle 15 57 53](https://user-images.githubusercontent.com/495617/125464850-cb43a942-0e7d-4aad-8748-d53a4d5a3fe3.png)

![Schermata 2021-07-13 alle 15 58 28](https://user-images.githubusercontent.com/495617/125464926-8f71bec0-92f7-43e3-8607-b2e0f1eef304.png)

PR #4938, opened 2021-09-06
Shipping Labels: Update purchase form view model to support multi-package

itsmeichigo:
Part of #4599

# Description
To prepare for updates of the purchase form fields to support multi-package, I think it's best to first update the form view model. This PR aims to replace the single selected package ID and weight with a list of basic package details. This also leads to updates in the view controller and package details.

# Changes
## ShippingLabelFormViewModel
- Replaced `selectedPackageID` and `totalPackageWeight` with `selectedPackagesDetails`. This is a list with of a new model `ShippingLabelPackageInfo` with basic details: package ID, total weight and product ID list. Note: I'm not quite happy with the naming of the model and the variable so any suggestion for renaming is appreciated!
- Replaced `selectedPackage` with `selectedPackages` array and updated its calculated value to prepare appropriate model for each package ID.
- Updated other parts relating to selected packages: handling package selection, setting up default customs forms, purchasing label.

## ShippingLabelFormViewController & ShippingLabelPackageDetails
- Updated input and output of `ShippingLabelPackageDetails` to use the selected package list details.

## ShippingLabelPackageDetailsViewModel
- Updated initializer with new parameter `selectedPackages`. 
- Added temporary updates to support the new parameter. These should be fixed in the next PR to properly support multiple packages.

# Testing
With the changes in place, it is expected that configuring package details should still work properly as before. Fetching carriers and purchasing label should also work as before.

4. Configure Package Details: select any package and / or update package weight manually and tap Done. Notice that the details are displayed correctly on the purchase form.
5. Tap Package Details again, notice that correct package and weight are displayed.
6. Tap Carrier and Rates, notice that carrier list can still be fetched for the package if the details are appropriate.
7. Proceed through Payment Method and purchase the label. The purchase should still succeed.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Model/ShippingLabelPackageInfo.swift
+++ b/WooCommerce/Classes/Model/ShippingLabelPackageInfo.swift
@@ -3,7 +3,13 @@ import Foundation
 /// A structure to hold the basic details for a selected package on Package Details screen of Shipping Label purchase flow.
 ///
 struct ShippingLabelPackageInfo: Hashable {
+
+    /// ID of the selected package.
     let packageID: String
+
+    /// Total weight of the package in string value.
     let totalWeight: String
+
+    /// List of product or variation IDs for items in the package.
     let productIDs: [Int64]
 }
--- a/WooCommerce/Classes/Model/ShippingLabelPackageInfo.swift
+++ b/WooCommerce/Classes/Model/ShippingLabelPackageInfo.swift
@@ -1,0 +1,9 @@
+import Foundation
+
+/// A structure to hold the basic details for a selected package on Package Details screen of Shipping Label purchase flow.
+///
+struct ShippingLabelPackageInfo: Hashable {
+    let packageID: String
+    let totalWeight: String
+    let productIDs: [Int64]
+}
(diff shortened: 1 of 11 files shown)


WooCommerce/Classes/Model/ShippingLabelPackageInfo.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+/// A structure to hold the basic details for a selected package on Package Details screen of Shipping Label purchase flow.
+///
+struct ShippingLabelPackageInfo: Hashable, Equatable {
```
rachelmcr: I think this name works, but just thinking about possible alternate names:

What do you think of naming this `ShippingLabelPackageDetails`? Or would that get confusing because that's also the name of the Package Details screen? Another idea that comes to mind is `ShippingLabelPackageAttributes`.

rachelmcr (approved):
These changes all make sense in preparation for supporting multi-package, and everything worked as expected in testing. 👍 

I have a couple suggestions for naming the `ShippingLabelPackageInfo` model, feel free to take them or leave them!

WooCommerce/Classes/Model/ShippingLabelPackageInfo.swift
itsmeichigo: Thanks - I like the name `ShippingLabelPackageAttributes`! I don't think we can use `ShippingLabelPackageDetails` as we can't have two files of the same name in the same module.

PR #4919, opened 2021-09-03
[Mobile Payments] Modal UI fixes

koke:
Fixes #4550 

This the requests from #4550. I've been fighting auto layout for a couple days already and was about to give up on some of the things, so I decided to put up a PR with the things I managed to fix. As I was writing up descriptions of the remaining issues I kept finding new solutions 😅 

## What's fixed

- Modal shown when connecting to reader has "Connecting to reader" in the title instead of "Scanning for reader"
- Modal shown when connecting to reader is displayed when manually connecting as well as when we auto connect
- Modals are always 280px wide and at least 382px high
- The top title on the modal uses the `body` font style
- The bottom title on the modal uses the `subheadline` font style
- Modals have 16px padding on all edges except the top one, which has 32px
- Adjusted spacing between bottom labels
- Adjusted spacing between buttons
- Adjusted layout so top title aligns to top, buttons align to bottom, and the rest of the content is evenly distributed between top labels, image, bottom labels, and buttons. Also, when there are no buttons, I added a spacer to the bottom, so that all the content is aligned to top 

Note that the spacing between the bottom labels in the "Insert Card" screen seems wrong. On Figma, I see a 2px space between labels, which is the spacing I set on Xcode, but the result is visually different. On Figma, both text boxes are 20px high and include some whitespace, on iOS the boxes are 18px and 15.67px respectively with the default font, so I could account for the difference and add an extra 6px of spacing, but I'm not sure if that would be what's expected with dynamic text.

Figma|Xcode|Runtime
-|-|-
Screen Shot 2021-09-03 at 10 25 27|Screen Shot 2021-09-03 at 10 25 53|Screen Shot 2021-09-03 at 10 30 42

## How it's fixed

I thought I'd leave a few notes, because reviewing diffs of XIB files is an exercise in decryption (even I'm having trouble remembering what I've changed exactly) 😩 

The behavior we want is that the top item (top labels) stick to the top, and the bottom item (buttons) sticks to the bottom. A stack view with a `fill` distribution won't do this by default and distribute the space evenly among all children, and these will typically be vertically centered in the available space. After a lot of trying, I adjusted the hugging priority to required (1000) in all the items I did not want to expand (top labels and their container, buttons and their container). For the bottom labels I see now that I left the priority at high (750), but I guess the behavior would be the same, as the rest of the items have a lower priority. The image and container for the bottom labels have low (250) priority, so they're the ones which will grow to fill the available space.

[...]

PR #4939, opened 2021-09-06
[Mobile Payments] Improved error modal

### LINKED ISSUE #4551: Error modal - UI update
Screenshot | Design
------------ | -------------

IMG_1059 | Frame 636

 - With the error message, the modal was too crowded. Can we remove the price to make room for more useful content on error modals?
 - Can we replace the "No thanks" text button with a "Back to Order" secondary button? 
 - Can we update the copy of the Primary button to "Try Collecting Again"? (All the buttons on iOS are Title case) 
 - Can we update the modal sizes to match designs? 

For this specific error message I highlighted above, can we make sure to display the image. I also find the error message a bit scary and not super helpful: could we update with the error message for unknown reasons "Sorry, this payment couldn’t be processed" ?

koke:
Fixes #4551 

This addresses the UI requests in #4551:

- Updated copy for the generic error message
- Remove order amount
- Update copy on the buttons

Before|After
-|-
Screen Shot 2021-09-06 at 13 16 01|Screen Shot 2021-09-06 at 13 14 07

I've also removed the amount and update the button copy on the success screens, prompted by https://github.com/woocommerce/woocommerce-ios/pull/4919#issuecomment-913623633

Before|After
-|-
Screen Shot 2021-09-06 at 16 19 03|Screen Shot 2021-09-06 at 16 21 54

## To test

I've forced a generic error by replacing the implementation of `StripeCardReaderService.createPaymentIntent(_:)` with:

```swift
func createPaymentIntent(_ parameters: PaymentIntentParameters) -> Future<StripeTerminal.PaymentIntent, Error> {
    return Future() { [weak self] promise in
        return promise(.failure(CardReaderServiceError.intentCreation(underlyingError: .internalServiceError)))
    }
}
```

Then:

1. Find an order eligible for payment
2. Tap on Collect Payment
3. Connect to the reader if necessary
4. Tap the card to pay
5. See the error message

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardReaderServiceError.swift
+++ b/Hardware/Hardware/CardReader/CardReaderServiceError.swift
@@ -306,7 +306,7 @@ updating the application or using a different reader
             return NSLocalizedString("The card reader session has expired - please disconnect and reconnect the card reader and then try again",
                                      comment: "Error message when the card reader session has timed out.")
         case .internalServiceError:
-            return NSLocalizedString("The system experienced an unexpected internal error - please try again",
+            return NSLocalizedString("Sorry, this payment couldn’t be processed",
                                      comment: "Error message when the card reader service experiences an unexpected internal service error.")
         }
     }
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalError.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalError.swift
@@ -3,9 +3,6 @@ import UIKit
 /// Modal presented on error
 final class CardPresentModalError: CardPresentPaymentsModalViewModel {
 
-    /// Amount charged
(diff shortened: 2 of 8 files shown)


koke:
Switching back to draft as I address a couple more things noted in https://github.com/woocommerce/woocommerce-ios/pull/4919#issuecomment-913623633:

> All modals for phones should be 280 x 382px, but it seems like the "Payment successful" modal isn't following those specs. In the designs, I removed the price to make room for the extra CTAs. Also, this can probably be addressed in a different PR, but the CTA of the text button should be "Back to Order" instead of "No thanks"

ghost (approved):

ctarda (approved):
The code looks good and it tests well.

The error message:


And the success screen (when there is no email setup in the device):


:shipit:

Issue #4550, opened 2021-07-06
Connect and Payment flow: Modal UI updates


Garance91540:
Hi there, I know we discussed about the modal size issue before, but I thought it would be clearer to do an exhaustive audit anyway.

Here's a general comparison of screenshots and designs the whole connect reader + collect payment flow so we can get an overview, and I'll detailed the changes separately below:

<img width="719" alt="Screen Shot 2021-07-06 at 10 58 31 AM" src="https://user-images.githubusercontent.com/28450084/124572684-1fa7c680-de49-11eb-97ee-d4b6f4e8d2c3.png">

I made some changes to unify the connect and collect flow to make implementation a bit easier:
- the modal titles fonts are always "body" and have a fixed top vertical padding of 32px
- if there are buttons, they always have a bottom vertical padding of 16px, and left and right padding of 16px
- In general (there are some exceptions, as displayed on Figma) all the other modal elements are vertically distributed between the modal titles and the buttons

*Scanning Reader

Screenshots | Designs
------------ | -------------
![IMG_1057](https://user-images.githubusercontent.com/28450084/124571927-811b6580-de48-11eb-984d-3df4c7d9f4a5.png) | ![Frame 648](https://user-images.githubusercontent.com/28450084/124573583-fb001e80-de49-11eb-8393-22ab48c10b9e.png)

- Update modal size to 280 x 382 px
- Update text font to match design
- Update image size to match design

*Connect Reader screen

Screenshots | Designs
------------ | -------------
![IMG_1058](https://user-images.githubusercontent.com/28450084/124570982-a9569480-de47-11eb-87e4-e7be98a5563a.png) | ![Frame 649](https://user-images.githubusercontent.com/28450084/124574020-59c59800-de4a-11eb-9a21-3be0029ea761.png)

Here's the list of UI updates:
- Update modal size to 280 x 382 px
- Update text font to match design
- Update image size to match design
- Update button sizes and vertical padding to match designs

*Collect Payments - insert card

Screenshots | Designs
------------ | -------------
![IMG_1062](https://user-images.githubusercontent.com/28450084/124574217-8bd6fa00-de4a-11eb-8362-8aaf2ea0d90d.png)  | ![Frame 646](https://user-images.githubusercontent.com/28450084/124574259-942f3500-de4a-11eb-92b8-32f40107a3c8.png)

Here's the list of UI updates:
[...]

PR #4940, opened 2021-09-06
[Mobile Payments] Adjust modal dimensions for iPads

### LINKED ISSUE #4557: iPad modals - UI update
Hi there, 

I realized there were no tablet designs for the CPP projects. I added them in Figma for reference. Similarly to the mobile modals, is it possible to update the modal sizes so they are consistent across the flow?

Screen Shot 2021-07-06 at 2 37 40 PM

koke:
Fixes #4557 

This implements the changes requested in #4557 to make the modals a bit bigger on iPads. Looking at the screens in Figma, I wasn't able to see a consistent layout (the horizontal margin of the buttons was 96, 103, or 114.25 depending on the screen), so here's how I interpreted the rules:

- The modal size is 454x490
- The vertical padding is 42px
- Instead of setting a horizontal padding, I tried to keep the content at the same size it was on the iPhone modals (280-2x16 = 248px) and make sure it's centered. This would be equivalent to a 103px horizontal padding, but it made a bit more sense to define it that way
- All these new rules apply when both the horizontal and vertical [size class](https://useyourloaf.com/blog/size-classes/) are "regular"

## To test

1. Find an order that's eligible for payments
2. Tap Collect Payment
3. Connect to a reader and go through the payment flow

The modal view controller is generic to all the payment modals, so the changes should be the same for all screens, but feel free to test as many cases as you want.

You can also check that the modals haven't changed when running on an iPhone.

Screen Shot 2021-09-06 at 14 07 16|Screen Shot 2021-09-06 at 14 08 09|Screen Shot 2021-09-06 at 14 07 24
-|-|-
Screen Shot 2021-09-06 at 14 07 27|Screen Shot 2021-09-06 at 14 07 33|Screen Shot 2021-09-06 at 14 07 40

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardPresentPaymentsModalViewController.swift
@@ -11,6 +11,7 @@ final class CardPresentPaymentsModalViewController: UIViewController {
 
     @IBOutlet weak var containerView: UIView!
     @IBOutlet weak var mainStackView: UIStackView!
+    @IBOutlet weak var bottomPaddingRegular: NSLayoutConstraint!
     @IBOutlet weak var primaryActionButtonsStackView: UIStackView!
     @IBOutlet weak var buttonsSpacer: UIView!
     @IBOutlet private weak var topTitleLabel: UILabel!
@@ -82,6 +83,7 @@ final class CardPresentPaymentsModalViewController: UIViewController {
 
         heightConstraint.priority = .required
         widthConstraint.priority = .required
+        configureSpacer()
     }
 }
 
@@ -117,8 +119,8 @@ private extension CardPresentPaymentsModalViewController {
 
     func styleBottomLabels() {
         actionButtonsView.isHidden = true
-        buttonsSpacer.isHidden = false
         bottomLabels.isHidden = false
+        configureSpacer()
 
         styleBottomTitle()
         styleBottomSubtitle()
@@ -134,8 +136,8 @@ private extension CardPresentPaymentsModalViewController {
 
     func styleActionButtons() {
         actionButtonsView.isHidden = false
-        buttonsSpacer.isHidden = true
         bottomLabels.isHidden = true
+        configureSpacer()
 
         stylePrimaryButton()
         styleSecondaryButton()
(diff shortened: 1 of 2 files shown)


ghost:
These are looking good to me so far thanks @koke before I approve, @Garance91540 do you have any thoughts on the alignment of the modals that don't have a button, etc at the bottom? Should we just align vertically?

ctarda (approved):
Looks good to me, and tests well.

Some screenshots:






:shipit:

Garance91540 (commented):
> @Garance91540 do you have any thoughts on the alignment of the modals that don't have a button, etc at the bottom? Should we just align vertically?

Yes, I think we should keep the text top aligned and try centering the image in the modal for those ones. If not the white space in the bottom half part of the modal is a bit too prominent

koke:
> we should keep the text top aligned and try centering the image in the modal for those ones

To clarify this, I've been following this idea from #4550 that "all the other modal elements are vertically distributed between the modal titles and the buttons". In the case where there are no buttons, applying that behavior effectively distributes the content between the titles and the bottom of the modal, which I don't think it's what we want.

Screen Shot 2021-09-07 at 17 21 49

The latest thing I was trying (unsuccessfully so far, we'll see about that tomorrow) was to add a fixed size spacer with the height equivalent to the case when there are two buttons (95px), so that the content is distributed between the titles and the empty buttons view. It would look like this:

Screen Shot 2021-09-07 at 18 16 12

@Garance91540 does that look good to you?

Garance91540:
> @Garance91540 does that look good to you?

yes!

koke:
Added the missing centering on fa7e694. It took me a while to figure it out, and the final solution is a bit hacky but it seems to work fine. Instead of trying to adjust the spacer and all the content hugging priorities, I skip the spacer, and adjust the bottom padding to include all the extra space that the buttons would take.
Screen Shot 2021-09-08 at 16 25 40

I tested multiple sizes on iPad and iPhone:

Screen Shot 2021-09-08 at 16 17 26|Screen Shot 2021-09-08 at 16 17 01
-|-
Screen Shot 2021-09-08 at 16 17 04|Screen Shot 2021-09-08 at 16 26 53
Screen Shot 2021-09-08 at 16 19 15|Screen Shot 2021-09-08 at 16 19 19

@ctarda do you mind reviewing the code change in fa7e694?

ctarda (approved):
The new updates look good to me @koke Im sorry I missed the request for review. :shipit:

Garance91540 (approved):
👏

Issue #4798, opened 2021-08-12
Order Editing: Fill address form content


Ecarrion:
After #4788 is build, let's fill the address form with the shipping/billing address information.

- [x] shipping address prefill
- [x] billing address prefill

PR #4944, opened 2021-09-06
Order Editing: Add viewmodel to Edit Address Form 

ealeksandrov:
Part of https://github.com/woocommerce/woocommerce-ios/issues/4778, https://github.com/woocommerce/woocommerce-ios/issues/4798

## Description

This PR adds address data prefill on Edit Address Form + unsaved changes check.

### Technical Details

The main addition is `EditAddressFormViewModel` which encapsulates all data logic.
State and Country fields are not added yet, should be merged with `CountrySelectorViewModel`.

### Questions

- Do we need additional "Optional" placeholders for email and phone (they are optional in `Address`), in addition to "Company" and "Address 2"?
- Do we need any validation? Web allows empty data for any of address fields. Maybe remove rest of our "Optional" placeholders then, so user won't be confused that we require all fields except those 2?

(cc @adamzelinski)

## Testing

1. Open existing order details and tap edit button for shipping address
2. Observe correct address data in form (except State and Country)
3. Try updating any fields
4. Check that "Done" button in navbar is enabled if there are any unsaved changes

---
Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
```diff
@@ -103,7 +105,7 @@ struct EditAddressForm: View {
 
                     Group {
                         TitleAndValueRow(title: Localization.countryField, value: Localization.placeholderSelectOption, selectable: true) {
-                            showCountrySelector = true
+                            viewModel.showCountrySelector = true
```
Ecarrion: Why do you think this should go in the `ViewModel`? 
As this is just part of navigation don't you think it should reside in the `View`?

I have never done this before so I'm wondering what is the best place! 😅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+final class EditAddressFormViewModel: ObservableObject {
+
+    init(address: Address?) {
+        self.originalAddress = address
+        updateFieldsWithOriginalAddress()
+    }
+
+    /// Original `Address` model.
+    ///
+    private let originalAddress: Address?
```
Ecarrion: I wonder if this should be a non-null address with empty content? 🤔

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+        return !(firstName == originalAddress.firstName &&
+                    lastName == originalAddress.lastName &&
+                    email == originalAddress.email ?? "" &&
+                    phone == originalAddress.phone ?? "" &&
+
+                    company == originalAddress.company ?? "" &&
+                    address1 == originalAddress.address1 &&
+                    address2 == originalAddress.address2 ?? "" &&
+                    city == originalAddress.city &&
+                    postcode == originalAddress.postcode)
```
Ecarrion: Can't we rely on the synthetized `Equatable` method?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -0,0 +1,74 @@
...
+    /// Set it to `true` to present the country selector.
+    ///
+    @Published var showCountrySelector: Bool = false
+
+    /// Return `true` if the done button should be enabled.
+    ///
+    var isDoneButtonEnabled: Bool {
+        guard let originalAddress = originalAddress else {
+            return true
+        }
```
Ecarrion: If we pass nil as the `Original Address` then the button will be enabled by default, is that expected?

Ecarrion (commented):
This looks super clean, and tests are working ✅ 
I do have some small concerns that may be up to a little of debate 😅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
ealeksandrov: I'm also not sure 😄 But after a second look lets keep it in the `View`, since it's accessed only there.
Updated in bcce2f62d3d783dccb8d1aacca9bf4a102212239.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
ealeksandrov: Agree! Done in 6fa05527c04e3ccd49308b058a93279daf66930a + 56d9b90f8b3e0fc6409a17cfdaba9d2a96224b65.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
ealeksandrov: We can't rely on synthetized method, since we'll have false detection for `"" != nil`. But I moved it to custom `Equatable` implementation to handle this in 14a86bf3caf81790cbc708058db8c78b03ee1d83.
It looks verbose, but works well. Let me know If you see a way to simplify that custom comparison.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
ealeksandrov: Fixed by updated comparison flow + non-nil `originalAddress`.
Added tests in 0e3f8d79fe9072ad29dc9f8776fffb55ae6e5025.

ealeksandrov:
Thanks for review @Ecarrion! I've took another look and updated changes detection flow.

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/Addresses/EditAddressFormViewModelTests.swift
```diff
@@ -69,4 +86,18 @@ private extension EditAddressFormViewModelTests {
...
+                       company: "",
+                       address1: "234 70th Street",
+                       address2: "",
+                       city: "Niagara Falls",
+                       state: "NY",
+                       postcode: "14304",
+                       country: "US",
+                       phone: "",
+                       email: "")
+    }
```
ealeksandrov, line 102: 4 optional fields are empty strings here. `EditAddressFormViewModel` will serialize them into `nil` values in `addressFromFields` computed var.

Ecarrion (approved):
🚀 

Thanks for submitting such concrete PRs they made the review process so much easier!

Regarding your questions

> Do we need additional "Optional" placeholders for email and phone (they are optional in Address), in addition to "Company" and "Address 2"?

I'm not sure, I would say yes but probably up to @adamzelinski to decide.

> Do we need any validation? Web allows empty data for any of address fields. Maybe remove rest of our "Optional" placeholders then, so user won't be confused that we require all fields except those 2?

I think we should follow any web validation, and if they don't have any, shouldn't have either.

Issue #4780, opened 2021-08-11
Order Editing: Fill Country & State UI


Ecarrion:
After #4779 is ready, we can fill in the information using the `DataStore.swift` actions.

Tasks
- [x] Download and fill country & state data
- [x] After users select something, refresh the address form with the selection
- [ ] Make sure to preselect and scroll to current selection

PR #4946, opened 2021-09-06
Order Editing: State Selector UI

### LINKED ISSUE #4779: Order Editing: Country & State UI

Ecarrion:
closes #4779

# Why

This PR adds a navigation link from the `AddressForm` screen to the `StateSelector` screen to continue adding more functionality to the edit a shipping address.

# How

- Refactor `CountryListSelector` to receive a generic `ViewModel` protocol instead of a specific `CountrySelectorViewModel`

- Created `StateSelectorViewModel` and `StateSelectorCommand` 

- Added a navigation link from  `AddressForm` screen to the `ListSelector` view with the `StateSelectorViewModel` value.

# Demo

https://user-images.githubusercontent.com/562080/132261022-8ee00ddc-f323-4533-9e6a-b183907c44f5.mov

# Testing Steps

- Navigate to an order and tap on the shipping address edit icon.
- Tap in the "select country" row and see that everything works as expected
- Go back and tap in the "select state" row and see that everything works as expected.

Note: States, for now, are just a copy of countries content. real values will be fetched in #4780 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/FilterListSelector.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/FilterListSelector.swift
@@ -21,9 +21,9 @@ protocol FilterListSelectorViewModelable: ObservableObject {
     var filterPlaceholder: String { get }
 }
 
-/// Filterable List Selector View/
+/// Filterable List Selector View
 ///
-struct FilterListCountrySelector<ViewModel: FilterListSelectorViewModelable>: View {
+struct FilterListSelector<ViewModel: FilterListSelectorViewModelable>: View {
 
     /// View model to drive the view content
     ///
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/FilterListSelector.swift
```diff
@@ -1,21 +1,42 @@
 import SwiftUI
 
-/// Country Selector View
+protocol FilterListSelectorViewModelable: ObservableObject {
```
ealeksandrov, line 3: nit: I would call it just `FilterListSelectorViewModel` but I have no idea what is correct 😄

ealeksandrov (approved):
LGTM :shipit:

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/FilterListSelector.swift
Ecarrion: I named it like that because of this rule.

> Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).

[Source](https://swift.org/documentation/api-design-guidelines/)

Ecarrion: I always struggle with those suffixes! 😅

PR #4949, opened 2021-09-08
Order Editing: Display supported countries on  Country List Selector

Ecarrion:
part of #4780

# Why

Country Selector UI was introduced in #4779 by showing a list of hardcoded countries. This PR refactors the `ViewModel` in order to show the WC-supported countries that exist in our `CoreData` store.

Additionally, if the `ViewModel` detects that no country is stored, it will proceed to sync them from a remote source.

# How

- Update `CountrySelectorCommand` so it it's internal data(countries) can be modified. This also forced me to make the command an `ObservableObject` so the `ListSelector` view can render new content when the command data changes.

- Update `CountrySelectorViewModel` to query countries from a `ResultsController`. If we detect that there are no countries stored, then sync them via the `syncCountries` Data Action.

# Demo

https://user-images.githubusercontent.com/562080/132451857-5956a05b-b52f-4b26-82e3-7f6eb45f2070.mov

# Testing Steps

- Logout & login to guarantee a fresh store
- Navigate to an order and tap on the shipping address edit icon
- Tap on "select country" row.
- See that the correct countries are being displayed.

# Known issues & Next PRs

- Add a loading state while the VM is syncing countries
- Search term is persisted between navigations 
- Render currently selected country

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ListSelector/ListSelector.swift
+++ b/WooCommerce/Classes/ViewRelated/ListSelector/ListSelector.swift
@@ -1,12 +1,16 @@
 import SwiftUI
 
+/// Protocol required to re-render the view when the command updates any of it's content.
+///
+protocol ObservableCommand: ListSelectorCommand, ObservableObject {}
+
 /// `SwiftUI` wrapper for `ListSelectorViewController`
 ///
-struct ListSelector<Command: ListSelectorCommand>: UIViewControllerRepresentable {
+struct ListSelector<Command: ObservableCommand>: UIViewControllerRepresentable {
 
     /// Command that defines cell style and provide data.
     ///
-    let command: Command
+    @ObservedObject var command: Command
 
     /// Table view style.
     ///
--- a/WooCommerce/Classes/ViewRelated/ListSelector/ListSelector.swift
+++ b/WooCommerce/Classes/ViewRelated/ListSelector/ListSelector.swift
@@ -2,11 +2,11 @@ import SwiftUI
 
 /// Protocol required to re-render the view when the command updates any of it's content.
 ///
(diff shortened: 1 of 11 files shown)


WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/Addresses/CountrySelector/CountrySelectorViewModelTests.swift
```diff
@@ -5,57 +5,69 @@ import TestKit
...
+            "Congo - Kinshasa",
             "Cook Islands",
-            "Colombia",
             "Costa Rica",
-            "Comoros",
-            "Morocco",
-            "Monaco",
             "Mexico",
+            "Monaco",
+            "Morocco",
```
Ecarrion, line 37: The order of these changed because the `ResultsController` now filters them by name.

WooCommerce/Classes/ViewRelated/ListSelector/ListSelector.swift
```diff
@@ -1,12 +1,16 @@
 import SwiftUI
 
+/// Protocol required to re-render the view when the command updates any of it's content.
+///
+protocol ObservableCommand: ListSelectorCommand, ObservableObject {}
```
ealeksandrov: nit: this protocol name seems too generic, while it's a very specific mix for a single use case.

ealeksandrov (approved):
Works well and looks good 🚀

WooCommerce/Classes/ViewRelated/ListSelector/ListSelector.swift
Ecarrion: Right! I'll change it to `ObservableListSelectorCommand`

Issue #4745, opened 2021-08-06
Shipping Labels: Create custom packages and activate service packages (endpoint)


rachelmcr:
This is a subtask of #3909.

Use the [existing endpoint](https://github.com/woocommerce/woocommerce-ios/issues/3870) for creating a new Custom Package. It should be probably be edited on our side, for enabling also the predefined packages.

An example JSON request to create two custom packages and activate two service packages at once looks like this:

```
{
    "custom":[
        {
            "is_user_defined":true,
            "is_letter":false,
            "name":"My Own Package",
            "inner_dimensions":"10 x 10 x 10",
            "box_weight":1,
            "max_weight":0
        },
        {
            "is_user_defined":true,
            "name":"test 02",
            "inner_dimensions":"1 x 1 x 1",
            "box_weight":1,
            "max_weight":0
        }
    ],
    "predefined":{
        "usps":[
            "small_flat_box",
            "medium_flat_box_top"
        ],
        "dhlexpress":[
            "Box2Cube",
            "Box2Small",
            "Box2Medium",
            "Box3Large"
        ]
    }
}
```

To create just one custom package, we simply enter one object into the custom array and set predefined as an empty object.

To activate a predefined package, we add the carrier/package name pair in predefined and let the custom part be an empty array.

For custom package, as far as I can tell, is_user_defined is always set as true. max_weight do not seem to be used yet in Woo. In core there is no user-facing field to set max_weight value, and saving seems to always set that value to zero, so I’m replicating that behavior in the Android app.

**Things to keep in mind:**

- [x] Once a custom package is created or service package is activated, make sure it is then listed in the package selector screen (this will be the screen before the package creation screen if there's already existing packages).
- [x] If the package creation fails, display an error notice to the user.

PR #4950, opened 2021-09-08
Shipping Labels: Create/activate new package on Add New Package screen

rachelmcr:
Part of: #4745

## Description

This makes the Done button functional on the Add New Package screen in the shipping labels flow.

When the Done button is tapped, the new custom or service package is created/activated remotely and the screen is dismissed. The new package is selected for the shipping label and is listed in the Package Selected screen (the screen before the Add New Package screen if the store has existing packages).

Note: If the API calls fail, the Add New Package screen is not dismissed. However, this PR does not include any user-facing error messaging. We should show a notice if there is an error, but it may take some work to get the notice presenter working with SwiftUI so I haven't included that here.

## Changes

Networking:

* Updates the `createPackage` endpoint in `ShippingLabelRemote` to support activating a service package.
* Updates the models for `ShippingLabelPackagesResponse` and `ShippingLabelPredefinedOption` to save the provider ID (e.g. "usps" or "dhlexpress") for predefined options. This is used by the `createPackage` endpoint when activating a service package.
* Updates the Fakes and unit tests with those changes.

Yosemite:

* Updates the Shipping Label Store and Action to support the changes to the `createPackage` endpoint.

UI:

* Makes the Done button functional in `ShippingLabelAddNewPackage`:
   * Uses the state property `isSyncing` to display an activity indicator and disable the done button while the API calls are in progress.
   * If the API calls are successful, it dismisses the view.
* Adds helpers to `ShippingLabelAddNewPackageViewModel` for the Done button actions (`createCustomPackage` and `activateServicePackage`). It gets the new custom/service package from the child view, creates/activates the package remotely, syncs the package list from remote, and passes the new package and updated package list back to the Package List / Package Details.

## Testing

7. If you already have existing packages on your store, you will see the Package Selected screen. Tap on "Create new package" at the bottom of the screen.
8. On the Add New Package screen, fill in the Custom Package form or select a Service Package.
9. Tap the Done button and notice there is an activity indicator and the button is disabled, and when it finishes you are taken back to the previous screen where the new package is selected.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
@@ -8,7 +8,12 @@ final class ShippingLabelAddNewPackageViewModel: ObservableObject {
     private let stores: StoresManager
     private let siteID: Int64
 
+    /// Index of the selected tab
+    ///
     @Published var selectedIndex: Int
+
+    /// View for the selected tab
+    ///
     var selectedView: PackageViewType {
         PackageViewType(rawValue: selectedIndex) ?? .customPackage
     }
@@ -17,12 +22,20 @@ final class ShippingLabelAddNewPackageViewModel: ObservableObject {
     var customPackageVM = ShippingLabelCustomPackageFormViewModel()
     lazy var servicePackageVM = ShippingLabelServicePackageListViewModel(packagesResponse: packagesResponse)
 
+    /// Package selected on the Custom Package tab
+    ///
     private var validatedCustomPackage: ShippingLabelCustomPackage? {
         customPackageVM.validatedCustomPackage
     }
+
+    /// Package selected on the Service Package tab
+    ///
     private var selectedServicePackage: ShippingLabelPredefinedPackage? {
         servicePackageVM.selectedPackage
     }
+
+    /// Package details fetched from the API
+    ///
     private var packagesResponse: ShippingLabelPackagesResponse?
 
     /// Completion callback
@@ -63,6 +76,8 @@ extension ShippingLabelAddNewPackageViewModel {
 
         createPackage(customPackage: newCustomPackage) { [weak self] success in
             onCompletion(success)
+
+            // On success, reset tab state and save new package details
             guard success else { return }
             self?.customPackageVM = ShippingLabelCustomPackageFormViewModel()
             self?.onCompletion(newCustomPackage, nil, self?.packagesResponse)
@@ -84,6 +99,8 @@ extension ShippingLabelAddNewPackageViewModel {
 
         createPackage(predefinedOption: selectedOption) { [weak self] success in
             onCompletion(success)
+
+            // On success, reset tab state and save new package details
             guard success else { return }
             self?.customPackageVM = ShippingLabelCustomPackageFormViewModel()
             self?.servicePackageVM.packagesResponse = self?.packagesResponse
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
@@ -19,8 +19,8 @@ final class ShippingLabelAddNewPackageViewModel: ObservableObject {
     }
 
     // View models for child views (tabs)
-    var customPackageVM = ShippingLabelCustomPackageFormViewModel()
-    lazy var servicePackageVM = ShippingLabelServicePackageListViewModel(packagesResponse: packagesResponse)
+    private(set) var customPackageVM = ShippingLabelCustomPackageFormViewModel()
+    private(set) var servicePackageVM: ShippingLabelServicePackageListViewModel
 
     /// Package selected on the Custom Package tab
     ///
@@ -53,6 +53,7 @@ final class ShippingLabelAddNewPackageViewModel: ObservableObject {
         self.selectedIndex = selectedIndex
         self.stores = stores
         self.siteID = siteID
+        self.servicePackageVM = ShippingLabelServicePackageListViewModel(packagesResponse: packagesResponse)
         self.packagesResponse = packagesResponse
         self.onCompletion = onCompletion
     }
@@ -103,7 +104,7 @@ extension ShippingLabelAddNewPackageViewModel {
             // On success, reset tab state and save new package details
             guard let self = self, success else { return }
             self.customPackageVM = ShippingLabelCustomPackageFormViewModel()
-            self.servicePackageVM.packagesResponse = self.packagesResponse
+            self.servicePackageVM = ShippingLabelServicePackageListViewModel(packagesResponse: self.packagesResponse)
             self.onCompletion(nil, selectedServicePackage, self.packagesResponse)
         }
     }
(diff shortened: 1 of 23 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
```diff
@@ -5,23 +5,157 @@ import Yosemite
...
+    ///
     var selectedView: PackageViewType {
         PackageViewType(rawValue: selectedIndex) ?? .customPackage
     }
 
     // View models for child views (tabs)
-    let customPackageVM = ShippingLabelCustomPackageFormViewModel()
-    let servicePackageVM: ShippingLabelServicePackageListViewModel
+    var customPackageVM = ShippingLabelCustomPackageFormViewModel()
+    lazy var servicePackageVM = ShippingLabelServicePackageListViewModel(packagesResponse: packagesResponse)
```
itsmeichigo: Nit: I think these should be `private(set)` to make sure that the variables can only be updated within this view model.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
```diff
@@ -5,23 +5,157 @@ import Yosemite
...
+                                                           providerID: shippingProvider.providerID,
+                                                           predefinedPackages: [selectedServicePackage])
+
+        createPackage(predefinedOption: selectedOption) { [weak self] success in
+            onCompletion(success)
+
+            // On success, reset tab state and save new package details
+            guard let self = self, success else { return }
+            self.customPackageVM = ShippingLabelCustomPackageFormViewModel()
+            self.servicePackageVM.packagesResponse = self.packagesResponse
```
itsmeichigo: I'd recommend avoiding direct mutations of objects from outside. For instance, since the service package view model's package response is internal and mutable, there's a risk that it can be updated accidentally from outside. This can be updated with better access control as follows:
- Keep the service package view model's `packagesResponse` private and immutable.
- Make `servicePackageVM` a variable instead of a lazy one so that you can reset it with a new package response when needed.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -83,7 +83,11 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
         return customPackages.count > 0
     }
 
-    lazy var addNewPackageViewModel: ShippingLabelAddNewPackageViewModel = ShippingLabelAddNewPackageViewModel(packagesResponse: packagesResponse)
+    lazy var addNewPackageViewModel = ShippingLabelAddNewPackageViewModel(siteID: order.siteID,
+                                                                          packagesResponse: packagesResponse,
+                                                                          onCompletion: { (customPackage, predefinedOption, packagesResponse) in
+                                                                            self.handleNewPackage(customPackage, predefinedOption, packagesResponse)
```
itsmeichigo, line 100: Since `onCompletion` is an `@escaping` block, it's safer to use capture `self` weakly here.

itsmeichigo (changes requested):
I have done a quick test and both adding custom packages and predefined packages work as expected 💯

However it seems that the package response is not synced on the purchase form yet. I think the form should trigger syncing package response on `viewWillAppear`? Or we can use a callback if that's preferred.



Plus - do you think it's necessary to add unit tests to the add package view model to check whether view models are reset after successfully creating packages?

I have also left some comments regarding access control and memory management.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackageViewModel.swift
rachelmcr: Thank you! That's good advice for me to keep in mind, and makes a lot of sense here.

rachelmcr:
> However it seems that the package response is not synced on the purchase form yet. I think the form should trigger syncing package response on viewWillAppear? Or we can use a callback if that's preferred.

Oh, great catch, thank you! I decided to use a callback since we're already syncing the package response after the new package is added on the add new package screen, so we can pass it back to the purchase form and handle it along with other changes to package details. (Also, that way we don't sync the package response every time the purchase form appears.)

> Plus - do you think it's necessary to add unit tests to the add package view model to check whether view models are reset after successfully creating packages?

Oh yes! I should have mentioned that I was planning to follow up with more unit tests, but I just finished up tests for that so I went ahead and added them to this PR.

@itsmeichigo I think I've addressed all of your feedback and resolved the merge conflicts, so this is ready for another look!

WooCommerce/WooCommerceTests/ViewRelated/Shipping Label/Create Shipping Label/ShippingLabelAddNewPackageViewModelTests.swift
```diff
@@ -0,0 +1,89 @@
...
+                onCompletion(.success(ShippingLabelPackagesResponse.fake()))
+            default:
+                break
+            }
+        }
+
+        viewModel.createCustomPackage(onCompletion: { _ in })
+
+        // Then
+        XCTAssertNil(viewModel.customPackageVM.validatedCustomPackage)
```
itsmeichigo, line 40: Suggestion: it'd be nice to also assert that all fields are reset to default values?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetails.swift
```diff
@@ -8,7 +8,7 @@ struct ShippingLabelPackageDetails: View {
 
     /// Completion callback
     ///
-    typealias Completion = (_ selectedPackages: [ShippingLabelPackageAttributes]) -> Void
+    typealias Completion = (_ packagesResponse: ShippingLabelPackagesResponse?, _ selectedPackages: [ShippingLabelPackageAttributes]) -> Void
```
itsmeichigo: I think just updating package response on completion (after selection a package) is not sufficient. There's a case where a user adds new package but then navigates out of the Package Details screen without selecting any package - here the package response is not updated. 

I'd suggest creating a new callback block just for updating package response, inject into the view model and call it at the end of `handleNewPackage` method. That way we can be sure that the package list is updated whenever a new package is added.

itsmeichigo (changes requested):
Thanks for updating the unit tests! I found that there's still a case where package response is not updated on the purchase form. Once that's resolved I think we'll be good to go!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetails.swift
itsmeichigo: It's also nice if we can add both blocks to the view model's initializer instead of keeping one in the view. Tapping Done will then trigger a method in the view model that calls the completion block.

rachelmcr: Good catch! I didn't consider that case before. Thank you for the really clear suggestions; that makes a lot of sense. I made those suggested changes in 08a9e47, including putting both blocks in the view model.

rachelmcr:
Ok @itsmeichigo, this is ready for another look! In particular, now it should be able to handle updating the package response on the purchase form even when you don't save changes to the package details.

itsmeichigo (approved):
Everything looks great now! Thank you for your patience! :shipit:

PR #4952, opened 2021-09-08
Order Editing: Fix issue where country state is persisted across navigations

Ecarrion:
part of #4780

# Why

While implementing #4949 I noticed that we had a problem where the "filter search term" was persisted when navigating multiple times to the country selector.

After some investigation, it seems that the `NavigationLink` destination view is instantiated when the `NavigationLink` is defined and it is never deallocated but re-used in future navigations. 

ref: https://twitter.com/chriseidhof/status/1144242544680849410

# How

To solve this problem, I've introduced two utility views.

- `LazyView`: As a wrapper that makes use of the `@autoclosure` attribute to convert the view creation function into a closure to be executed when the body is rendered.

- `LazyNavigationLink`: As a `NavigationLink` wrapper that wraps the "Destination View" inside a `LazyView`

# Demo

### Before
https://user-images.githubusercontent.com/562080/132533608-3cf6ef3a-5998-43d6-ba25-63c03bac3704.mov

### After
https://user-images.githubusercontent.com/562080/132533622-3f7cb1ff-ee7b-49df-8a04-a5b7f364f0f7.mov

# Testing Steps

- Navigate to an order and tap on the shipping address edit icon
- Filter some countries
- Navigate back and tap on the "Select Country" row again.
- See that the search term is not persisted.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
@@ -136,13 +136,13 @@ struct EditAddressForm: View {
         .disabled(!viewModel.isDoneButtonEnabled))
 
         // Go to edit country
-        NavigationLink(destination: FilterListSelector(viewModel: viewModel.createCountryViewModel()), isActive: $showCountrySelector) {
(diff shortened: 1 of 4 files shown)


ealeksandrov (approved):
That's one of reasons why I didn't like navigation in SwiftUI.
Good solution 👍

Issue #4947, opened 2021-09-07
[Mobile Payments] Errors when connecting to a reader are not displayed


ctarda:
The `connect` method in StripeCardReaderService capture potential errors in the connection process. 

Those errors are supposed to be propagated via the Future returned by this method, but the errors are not presented on screen.

koke:
This seems somewhat related to #4923 and #4924

allendav:
Closing this. We'll track unfinished business in 

https://github.com/woocommerce/woocommerce-ios/issues/4991

and

#4924

PR #4957, opened 2021-09-08
[Mobile Payments] Display connecting-to-reader failures

### LINKED ISSUE #4923: [Mobile Payments] Inconsistent discovery state when attempting to connect to a powered off reader
**Describe the bug**
When you attempt to connect to a reader and the reader is gone, the app is left in an inconsistent state, and needs to force-quit before it can connect again.

**To Reproduce**
1. Go to Settings > In-Person Payments > Manage Card Reader
2. Tap on Connect Reader
3. When a reader is found, turn off the reader
4. Tap Connect

You end up back and the card reader settings screen, but from now on, every time you try to connect you get an error saying that the SDK is busy with another command

allendav:
Partially addresses #4947 
Closes #4923 

To test:
- Tap on Settings
- Tap on In-Person Payments, then Manage card reader
- When the Connect your card reader view appears, tap on Connect Card Reader
- The “Scanning for reader” modal should appear
- Power on a reader at this point
- (If the reader has previously connected, it will autoconnect. You’ll need to disconnect it and repeat the process)
- When the “Do you want to connect to reader xxx” modal appears, power off the reader by holding its power button down at least 5 seconds
- Tap on “Connect to Reader”
- The “Connecting to reader” modal will display
- After about 10 seconds, the “We couldn’t connect your reader” modal will display
- Tap “Cancel”
- You should be back on the Connect your card reader view now
- Tap on Connect Card Reader
- The Scaning for reader modal should display
- Power on the reader
- The Do you want to connect to reader xxx” modal should appear
- power off the reader by holding its power button down at least 5 seconds
- Tap on “Connect to Reader”
- The “Connecting to reader” modal will display
- After about 10 seconds, the “We couldn’t connect your reader” modal will display
- Tap on “Try Again”
- The “Scanning for reader” modal should return
- Power on your reader
- NOTE: There is a bug here. The reader won’t be found. It looks like it may be a problem with the Stripe SDK: https://github.com/stripe/stripe-terminal-ios/issues/104#issuecomment-916285167 - The workaround follows:

- Cancel the Scanning for reader modal
- Tap on Connect Card Reader
- The Do you want to connect to reader xxx” modal should appear
- Tap on “Connect to Reader”
- The reader should now connect successfully

Screenshots:

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
+++ b/WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
@@ -36,7 +36,7 @@ final class CardReaderConnectionController {
         /// A failure occurred while connecting. The search may continue or be canceled. At this time we
         /// do not present the detailed error from the service.
         ///
-        case connectingFailed
+        case connectingFailed(Error)
 
         /// User cancelled search/connecting to a card reader. The completion passed to `searchAndConnect`
         /// will be called with a `success` `Bool` `False` result. The view controller passed to `searchAndConnect` will be
@@ -127,8 +127,8 @@ private extension CardReaderConnectionController {
             onCancel()
         case .connectToReader:
             onConnectToReader()
-        case .connectingFailed:
-            onConnectingFailed()
+        case .connectingFailed(let error):
+            onConnectingFailed(error: error)
         case .discoveryFailed(let error):
             onDiscoveryFailed(error: error)
         }
@@ -319,7 +319,7 @@ private extension CardReaderConnectionController {
                 self.returnSuccess(connected: true)
             case .failure(let error):
                 ServiceLocator.analytics.track(.cardReaderConnectionFailed, withError: error)
-                self.state = .connectingFailed
+                self.state = .connectingFailed(error)
             }
         }
         ServiceLocator.stores.dispatch(action)
@@ -329,13 +329,14 @@ private extension CardReaderConnectionController {
 
     /// An error occurred while connecting
     ///
-    private func onConnectingFailed() {
+    private func onConnectingFailed(error: Error) {
         guard let from = fromController else {
             return
         }
 
-        /// Let's start over again - we don't want to try and connect to something obviously borked
-        /// right off the bat
+        /// Clear our copy of found readers to avoid connecting to a reader that isn't
+        /// there while we wait for `onReaderDiscovered` to receive an update.
+        /// See also https://github.com/stripe/stripe-terminal-ios/issues/104#issuecomment-916285167
         ///
         self.foundReaders = []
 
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -33,6 +33,11 @@ final class CardReaderConnectionController {
         ///
         case connectToReader
 
+        /// A failure occurred while connecting. The search may continue or be canceled. At this time we
+        /// do not present the detailed error from the service.
+        ///
+        case connectingFailed
```
ctarda: In preparation for the 2.0 SDK it might be a good idea to add either an extra case or make the view model title customizable, or pass the error as an associated value, for when the connection fails and we want to present a custom error message, instead of a standard one. "We couldn't connect, try again"

I have seen two error cases that might be worth treating with a custom error message:
1. The SDK detects a mandatory update, and the reader battery is below 50%. We should notify users that they need to charge their battery (pdfdoF-7i-p2#comment-327)
2. On the first connection to an unknown reader, the user needs to enter in the mobile app a 6 digit number located in the back of the reader. If the user enters the wrong number, the connection fails. I think this one might deserve an specific error as well

allendav: I agree with presenting more actionable dialogs, but I think extending and testing with these SDK 2 messages should wait for a later PR (it will be hard to test that we got-it-right with SDK 1). How about I open a separate issue for that?

allendav:
The bug where a powered-off-and-on-again reader is not easily re-discovered appears to be in the Stripe Terminal SDK. See this issue: https://github.com/stripe/stripe-terminal-ios/issues/104#issuecomment-916285167

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
allendav: On further discussion, I'll add an associated value we can leverage later.

allendav: Added associated value in https://github.com/woocommerce/woocommerce-ios/pull/4957/commits/0db9ef0b4bc11efbf1857839a089867dc84f323f

ctarda (approved):
Code looks great. :shipit:

ghost (approved):

koke:
I'll just note that, while this new modal is showing, the discovery process would still be active. So if you turn on a second reader, the app would present the option to connect to it, even without tapping Try again. Also, I would probably understand Try again as trying to connect to the same reader, not trying to scan for readers again.

PR #4958, opened 2021-09-08
Order Editing: Show a loading placeholders while countries are being synced.

Ecarrion:
part of #4780

# Why

As countries can be synced from a remote source, it's necessary to show some ghost/loading content while the sync finishes.

# How

- Update `FilterListSelector` to show a redacted + shimmer effect when the view model mandates it.

- Update `CountrySelectorViewModel` to provide a property `showPlaceholders` that indicates if we need to show gosht/loading content or not. 
This is achieved by streamlining two publishers into the variable, `syncCountriesTrigger` and `makeSyncCountriesFuture`
The idea is to maintain all state mutations within the same chain of publishers, so it's clear how the `showPlaceholders` property acquired their values.

# Demo
https://user-images.githubusercontent.com/562080/132597469-1ed59178-5bb3-4a0b-9f08-6631fb012542.mov

# Testing Steps

- Log out from the app to clear the storage
- Login and navigate to an order
- Tap on the shipping address edit button
- See some redacted views with a shimmering effect while the content is synced.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/CountrySelectorViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/CountrySelectorViewModel.swift
@@ -15,6 +15,10 @@ final class CountrySelectorViewModel: FilterListSelectorViewModelable, Observabl
         }
     }
 
+    /// Define if the view should show placeholders instead of the real elements.
+    ///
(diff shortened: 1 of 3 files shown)


ealeksandrov (approved):
LGTM :shipit:
Thank you for helpful comments on combine logic, very interesting solution!

Ecarrion:
Thanks for the quick review!

PR #4959, opened 2021-09-09
Shipping Labels: New packages form to support multiple packages

itsmeichigo:
Part of #4599 

# Description
This PR adds new view for the Package Details screen to support multiple packages. This screen shows a list of package instead of just one like before. For now the list only shows header of each package item; content of the items will be added in the next PR.

# Changes
- Added new view and view model for packages form. The form contains a list of package items.
- Added new view and view model for package items. Each item shows only a header for now.
- Updated navigation from shipping label purchase form to the packages form instead of the old package details when the multi-package feature flag is enabled.
- Added unit tests for the packages form view model.
- Updated `ShippingLabelPackageNumberRow` to remove padding and fix incorrect title for non-plural cases. Also updated padding for the row in the old Package Details screen so that it still looks good with feature flag multi-package off.

# Screenshot

# Testing
4. Tap Continue on Package Details row. Notice that you're navigated to a new Package Details screen with just title of one package.
5. Tap Done button, notice that you're navigated back to the purchase form. No package is configured yet, this will be updated in the next PR.

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackageItem.swift
```diff
@@ -0,0 +1,48 @@
+import SwiftUI
+
+struct ShippingLabelPackageItem: View {
```
rachelmcr, line 3: Could we give this a name that doesn't use the term "item"? I got a little confused here thinking that `ShippingLabelPackageItem` was for an item (like an order item) within a package.

A couple ideas:

* ShippingLabelPackageEntity
* ShippingLabelPackageUnit

I'm not crazy about those ideas, though, so I wouldn't object if you want to keep the name as it is. Once I started looking at it in context I figured out what the name meant.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackagesFormViewModel.swift
```diff
@@ -0,0 +1,298 @@
...
+
+    /// If no initial packages was input, set up default package from last selected package ID and all order items.
+    ///
+    private func configureDefaultPackage() {
+        guard selectedPackages.isEmpty,
+              let selectedPackageID = resultsControllers?.accountSettings?.lastSelectedPackageID else {
+            return
+        }
+        selectedPackages = [ShippingLabelPackageAttributes(packageID: selectedPackageID,
+                                                           totalWeight: "",
```
rachelmcr, line 66: Should the `totalWeight` here be set as the weight of the last selected package?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageNumberRow.swift
```diff
@@ -14,7 +14,7 @@ struct ShippingLabelPackageNumberRow: View {
         HStack {
             Text(String(format: Localization.package, packageNumber))
                 .font(.headline)
-            if numberOfItems == 0 {
+            if numberOfItems <= 1 {
```
rachelmcr, line 17: Thanks for fixing this! 🎉

rachelmcr (approved):
This looks great! I just have a couple minor questions below.

I confirmed that it works as expected with and without the feature flag enabled. 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackagesFormViewModel.swift
rachelmcr: After looking at your other PRs I think I understand this now. We configure just the default package ID here, and the package weight will be configured separately in `configureTotalWeight(initialTotalWeight:products:productVariations:)` (part of #4962).

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackageItem.swift
itsmeichigo: Thanks a lot for the suggestion! I agree that `item` does sound ambiguous, but I'm not sure about `entity` and `unit` either 😓 I'll keep this in mind and update in a later PR, since renaming here can cause issues for the subsequent PRs.

PR #4964, opened 2021-09-09
Shipping Labels: Display error when package creation fails

rachelmcr:
Part of: #4745 
⚠️ Merge after #4950 ⚠️ 

## Description

In the Shipping Label creation flow, if adding a new package fails we now display an alert with an error message to the user. This will show a specific error if there is a problem with the package itself (e.g. a custom package with a non-unique name) or a generic error for other problems (e.g. a network error).

For now, I used an alert because our other reusable components ( notice, error banner) aren't readily available to use with SwiftUI views. This feels like an acceptable compromise for now, to be able to release this feature sooner, and we can replace the alert with a different component later. @woocommerce/mobile-designers If you have a different perspective on that, please let me know!

## Changes

* Adds `PackageCreationError` to handle the expected errors when creating packages. In the UI layer, these errors are mapped to localized strings that match our UI (where only one package is created at a time).
* Adds an alert to `ShippingLabelAddNewPackage` that is displayed when adding a new package fails, with the errors from `PackageCreationError`.

Package name error|Generic error
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-09-09 at 18 40 55|Simulator Screen Shot - iPhone 12 Pro - 2021-09-09 at 18 41 06

## Testing

7. If you already have existing packages on your store, you will see the Package Selected screen. Tap on "Create new package" at the bottom of the screen.
8. On the Add New Package screen, fill in the Custom Package form — to make sure an error is triggered, give your package the same name as a custom package that already exists on your store.
9. Tap the Done button. (You can also trigger an error by disabling your network connection before tapping Done.)

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
@@ -77,8 +77,9 @@ struct ShippingLabelAddNewPackage: View {
                     })
                     .disabled(isSyncing)
                     .alert(isPresented: $showingAddPackageError, content: {
-                        Alert(title: Text(Localization.errorAlertTitle),
-                              message: Text(Localization.errorAlertMessage))
+                        let title = viewModel.error?.alertTitle ?? Localization.errorAlertTitle
+                        let message = viewModel.error?.errorDescription ?? Localization.errorAlertMessage
+                        return Alert(title: Text(title), message: Text(message))
                     })
                 })
             }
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Package Selection/Package Creation/ShippingLabelAddNewPackage.swift
```diff
@@ -4,7 +4,8 @@ import Yosemite
 struct ShippingLabelAddNewPackage: View {
     @ObservedObject var viewModel: ShippingLabelAddNewPackageViewModel
     @Environment(\.presentationMode) var presentation
-    @State var isSyncing = false
+    @State private var isSyncing = false
+    @State private var showingAddPackageError = false
```
itsmeichigo, line 8: Thanks for making these private 🎉

itsmeichigo (approved):
✅ Code looks good and works as advertised!

> because our other reusable components ( notice, error banner) aren't readily available to use with SwiftUI views.

I'd like to point out that we already have views like `ShippingLabelPaymentMethodsTopBanner` that is essentially a SwiftUI view that wraps `TopBannerView`. We certainly need to make it more generic for reuse in more views instead of creating specific wrappers like that.

rachelmcr:
> I'd like to point out that we already have views like ShippingLabelPaymentMethodsTopBanner that is essentially a SwiftUI view that wraps TopBannerView.

Oh, that's right — I forgot we'd done that. Thanks for pointing it out! I've made a note to look into making that reusable when I'm back from my AFK.

PR #4967, opened 2021-09-09
Order Editing: Sync countries on EditAddressForm

Ecarrion:
part of #4798

# Why 

In a previous PR, I have added the ability to sync countries in the country selector screen. 
To not replicate the same logic on the state selector screen this PR moves that sync logic from `CountrySelectorViewModel` to `EditAddressFormViewModel` so it can be reused when pushing the state selector screen.

# How

- Move all country syncing code from `CountrySelectorViewModel` to `EditAddressFormViewModel`.

- Remove code from `CountrySelectedCommand` that allowed consumers to modify the countries list, as these are now provided on the init and not asynchronously.

- In `EditAddressFormViewModel` I'm experimenting with a way to perform actions only one. In the view, we only have the `onAppear()` modifier but this is invoked every time the screen is presented, similar to `viewWillAppear`. To overcome this I'm using a publisher to act as a trigger for the actions that we need to be performed only once, like data fetching. This is needed in order to not perform any side effect on `previews` and to allow for easier testing.

# Demo

https://user-images.githubusercontent.com/562080/132762354-c4cd0418-b4a0-4925-88dc-b34e616879cb.mov

# Testing Steps

- Log out from the app to clear the storage
- Login and navigate to an order
- Tap on the shipping address edit button
- See the loading effect in the screen
- See the country selector screen

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/CountrySelectorCommand.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/CountrySelectorCommand.swift
@@ -9,7 +9,7 @@ final class CountrySelectorCommand: ObservableListSelectorCommand {
 
     /// Original array of countries.
     ///
-    private var countries: [Country]
+    private let countries: [Country]
 
     /// Data to display
     ///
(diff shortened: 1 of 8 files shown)


ealeksandrov (approved):
Works well :shipit: 
One nit that I've noticed - if countries list fetching fails we get empty picker UI without any explanations. It's a rare edge case, but maybe we can add simple empty state with "try again later" message?

Ecarrion:

Yes handling errors is on my TODO list! Will do that in a next PR!

PR #4962, opened 2021-09-09
Shipping Labels: New package details selection

itsmeichigo:
Part of #4599 
This PR depends on #4961 so please make sure that PR is reviewed and merged first.

# Description
This PR adds selection logic for the new package details screen. To reuse the logic between the old and the new screens, the selection logic is extracted into a separate view model.

# Changes
- Adds new view model for package list and copies the logic from `ShippingLabelPackageDetailsViewModel`.
- Updates the correct view model for `ShippingLabelPackageList` and `ShippingLabelPackageSelection` views.
- Removes package list and selection logic from `ShippingLablePackageDetailsViewModel` and adds package list view model.
- Adds package list view model to `ShippingLabelPackageItemViewModel` to handle package switching.
- Handles configuring package total weight in  `ShippingLabelPackageItemViewModel`.
- Updates `ShippingLabelPackageItemViewModel` initializer with new block for notifying the packages form of new package.
- Updates `ShippingLabelPackagesFormViewModel` to update selected package list.
- Updates unit tests for both package item and package details view models.

# Demo
package-selection

# Testing

1. Make sure that your test store has installed WCShip plugin and configured packages in WPAdmin > WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Open an order detail that is eligible for the shipping label.
3. Configure Ship To and Ship From.
4. Tap Continue on Package Details row. Switch between packages, notice that the correct title and weight of the package is displayed. Edit the weight to an arbitrary number and switch packages, notice that the weight is no longer updated.
5. Turn off multi-package feature flag, repeat steps 2, 3 and 4. Verify that everything still works correctly.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #4970, opened 2021-09-10
Shipping Labels: New package details validation

itsmeichigo:
Part of #4599 
⚠️ This PR depends on #4962 so please make sure that one is reviewed and merged first! ⚠️ 

# Description
This PR adds validation for total weight and handles confirm of package selection in the new Package Details screen (with multiple packages support). Validation error message is also added to make it clearer when the Done button is disabled - this is a new addition compared to the old Package Details screen.

# Changes
- Package item view model: 
   - Adds validation for total weight.
   - Returns validated package attributes if the total weight is valid.
- Package item view:
   - Shows validation error row when total weight is invalid.
- ShippingLabelPackageNumberRow:
   - Adds new error icon in case validation fails for the package. This is hidden by default to not show up in the old Package Details screen.
- Packages form view model:
   - Observes item view models for validation states and save them locally.
   - Observes validation states to check whether Done button should be enabled.
   - Adds new method to handle package selection confirm.
- Packages form: 
   - Update Done button state and trigger confirm when the button is tapped.
- Shipping Label purchase form: 
   - Updates selected packages when receiving completion from the packages form.
- Update unit tests for package item and packages form view models.

# Demo
package-validation

ℹ️  The footer of the package item and validation error look weird when stacking them together, so I'm showing either of them based on the validation state. @woocommerce/mobile-designers - let me know what you think about this.

# Testing 
4. Tap Continue on Package Details row. Edit the weight to an invalid value, notice that a validation error shows up and Done button is disabled.
5. Update the weight to a valid value and tap Done. Notice on the purchase form that the correct package name and weight is displayed. Tap on Package Details row again, notice that correct package details are displayed.
6. Turn off multi-package feature flag and repeat steps 2, 3, 4, 5. Verify that everything still works as expected.

Update release notes:


Garance91540 (approved):
👏

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackageItemViewModel.swift
```diff
@@ -183,6 +202,12 @@ private extension ShippingLabelPackageItemViewModel {
         }
         return tempTotalWeight
     }
+
+    /// Validate that total weight is a valid floating point number.
+    ///
+    private func validateTotalWeight(_ totalWeight: String) -> Bool {
+        totalWeight.isNotEmpty && Double(totalWeight) != 0 && Double(totalWeight) != nil
+    }
```
pmusolino: I believe that for validating the value, we should convert internally all the commas `,` to dots `.` or find another solution (it would be useful to check how we approach it in the currency formatter). The reason is that if you select a country (eg. Italy) where the separator for the floating-point numbers is the comma and not the dot, the validation will return false (and also the keyboard will show only the comma), so will be impossible to insert a floating-point number here. It would be also helpful to add a unit test for it.

pmusolino (changes requested):

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackageItemViewModel.swift
itsmeichigo: Nice catch! I have added an extension for `NumberFormatter` to use locale when converting between string and number in 10a3b3d, along with unit tests for it. Please take another look and let me know what you think - thanks!

pmusolino: Thanks for fixing it, it looks good 💯

WooCommerce/Classes/Extensions/NumberFormatter+Localized.swift
```diff
@@ -0,0 +1,26 @@
+import Foundation
+
+extension NumberFormatter {
```
pmusolino, line 3: Non-blocking question: since we already have a `NumberFormatter` extension, do you think it is worthwhile to aggregate everything under a single extension, just not to create too many extensions of the same class?

pmusolino (approved):
Looks good to me, and thanks @itsmeichigo for adding all the unit tests 🚢

WooCommerce/Classes/Extensions/NumberFormatter+Localized.swift
itsmeichigo: Personally I don't see having separate extensions of the same type is a problem - since each extension deals with a different problem (which is named in the files). So I'm keeping it as is and merging this PR.

PR #4973, opened 2021-09-10
Shipping Labels: Fix incorrect phone validation for non-US country and add instant local validation on value changes

### LINKED ISSUE #4971: Shipping Labels: Address validation issues
Currently address validation on Edit Address screen is having several issues:
- Phone validation is checking for US phone number requirement regardless of the selected country. Other countries should not be applied the same check with US.
- Validation errors require tapping on action buttons to be cleared. Local validation should be handled on value changes instead.
- When switching country, state field is empty but the address's state is not actually updated. So when switching from US to Australia, new state is required but tapping on "Use Address as Entered" allows the address to keep the US state.

Internal reference: p5T066-2En-p2#comment-9902

itsmeichigo:
Fixes #4971 

# Description
This PR fixes issues in Edit Address screen validation:
- Required 10-digit number is only required for US.
- Address state of country is cleared when switching country.
- Validation error messages are updated instantly on value changes. Text field focus states are kept even after validation rows are updated.

# Changes
- View model: 
   - Updates `onChange` closure to accepts optional index for focused row.
   - Updates `handleAddressValueChanges` to reset state of country when country is updated.
   - Updates `handleAddressValueChanges` to do local validation and update table view rows.
   - Checks for country code US before proceeding to check for 10-digit phone number.
- View controller:
   -  Update `observeViewModel` to focus on the specified row after reloading table view.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -142,16 +142,20 @@ private extension ShippingLabelAddressFormViewController {
     }
 
     func registerTableViewCells() {
-            tableView.registerNib(for: TitleAndTextFieldTableViewCell.self)
-            tableView.registerNib(for: BasicTableViewCell.self)
+        tableView.registerNib(for: TitleAndTextFieldTableViewCell.self)
+        tableView.registerNib(for: BasicTableViewCell.self)
     }
 
     func observeViewModel() {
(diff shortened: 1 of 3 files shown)


pmusolino (approved):
Looks good to me. Not sure if in this case, it's better to merge this change in the next release 7.6 (maybe I missed the discussion).

itsmeichigo:
@pmusolino It seems that this PR is a bit confusing since it solves more than one issue - do you think I should separate it into several ones and only merge the most critical part? 

The issue that requires 10-digit phone number for all countries is critical because it blocks user from creating shipping labels for destination countries that don't have 10-digit phone number.

pmusolino:
I'm totally ok with merging it in release 7.5. I just want to make sure it's ok for @jkmassel, since the date for the code freeze of the next release 7.6 has passed. Jeremy, can you give us a 👍 or 👎 ?

PR #4974, opened 2021-09-10
[Mobile Payments] [SDK 2] Firmware updates for StripeTerminal 2.0

koke:
This PR implements the new software update UI for required updates. Note that this targets the `feature/stripe-sdk-2` branch.
With Stripe Terminal 2.0, these are handled a bit differently than optional updates, so I'll handle those in a separate PR.

0%|53%|Complete
-|-|-
stripe2-required-update-0|stripe2-required-update-53|stripe2-required-update-complete

Some notes on implementation:

- `CardReaderSoftwareUpdateState`. Previously the service published a `Float` value with the installation progress, because the rest of the information came implicit from the user initiating the update. Now the UI layer can observe this publisher to have a more complete view of the state regarding software updates.
- `FallibleCancelable`. I went back and forth with this one a few times. I needed a way to pass the "cancellable" that Stripe gives us when an update starts so the UI can have a "Cancel" button. I didn't want to expose the type from Stripe directly, and I thought of using Combine's `Cancellable`, but this is different because the cancel operation might fail with an error. I added a new simple type to represent this, and named it `FallibleCancelable` to avoid confusion with Combine, and because it represents what it does more precisely.
- Progress animation. This might not be the most elegant way to solve this, but it works well enough for something that will be shown infrequently. I would have preferred to draw directly on the screen, but `CardPresentPaymentsModalViewModel` wants a `UIImage` and I didn't want to start refactoring all the modals for this. 

## To test

1. On Xcode, edit the WooCommerce scheme, and enable `-simulate-stripe-card-reader` in the Arguments tab
2. In `StripeCardReaderService.start(_:)`, after the SDK has been initialized (I add this right after `Terminal.shared.delegate = self`), add the following line: `Terminal.shared.simulatorConfiguration.availableReaderUpdate = .required`
3. Run the app
4. Go to Settings > In-Person Payments > Manage Card Reader and connect to a reader
5. After the reader connects, it should automatically start the update 
6. Tapping the cancel button should cancel the update

https://user-images.githubusercontent.com/8739/134147854-12f5e238-90d0-4f74-b4a1-1a0a71852dac.mp4

Update release notes:


WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -410,6 +441,23 @@ private extension CardReaderConnectionController {
             return
         }
 
+        ServiceLocator.cardReaderService
```
ctarda: 🤔 I don't think we should not attack the service directly, and we should expose an action in CardPresentPaymentStore that returns the publisher via a completion block, the way we do for example, to check if there are card readers connected.

I understand that adding a layer of business logic might seem unnecessary, but I think the tradeoff is worth it. Specially given that right below this code we rely on the same store to connect to a reader

WooCommerce/Classes/Extensions/UIImage+Woo.swift
```diff
@@ -372,6 +372,24 @@ extension UIImage {
         return UIImage(named: "card-reader-connecting")!
     }
 
+    /// Card Reader Update background
+    ///
+    static var cardReaderUpdateProgressBackground: UIImage {
```
ctarda, line 377: We usually add tests to cover these images, in case the actual asset is missing for who knows what reason

ctarda (commented):
It works well, and code looks pretty good to me. I think there is some logic in CardReaderConnectionController that should be moved to a store though.

Hardware/Hardware/CardReader/CardReaderSoftwareUpdateState.swift
```diff
@@ -0,0 +1,17 @@
...
+    /// A mandatory update has started
+    case started(cancelable: FallibleCancelable?)
+
+    /// The update is being installed
+    case installing(progress: Float)
+
+    /// The update has finished installing
+    case completed
+
+    /// No update available
```
allendav, line 15: What about update failures (i.e. `installationFailed`)? Please open an issue to add that case.

Hardware/Hardware/CardReader/CardReaderSoftwareUpdateState.swift
```diff
@@ -0,0 +1,17 @@
...
+    case none
```
allendav, line 16: Would a separate state for `unknown` be appropriate - i.e. we haven't yet checked if an update is available? Separate issue is fine for that too.

I bring it up because below we initialize with `none` and I would have expected something more like `unknown`

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -478,16 +484,19 @@ extension StripeCardReaderService: BluetoothReaderDelegate {
         sendReaderEvent(.softwareUpdateNeeded(softwareUpdate))
     }
 
-    public func reader(_ reader: Reader, didStartInstallingUpdate update: ReaderSoftwareUpdate, cancelable: Cancelable?) {
-        // No-OP
+    public func reader(_ reader: Reader, didStartInstallingUpdate update: ReaderSoftwareUpdate, cancelable: StripeTerminal.Cancelable?) {
+        print("==== started software update")
```
allendav, line 488: Still want this?

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    let bottomTitle: String?
+
+    var bottomSubtitle: String? = Localization.message
+
+    init(progress: Float, cancel: (() -> Void)?) {
+        self.progress = progress
+        self.cancelAction = cancel
+
+        topTitle = progress == 1 ? Localization.titleComplete : Localization.title
```
allendav: Comparing floats to ints makes me nervous. Could we do this differently? e.g. progress > 0.99 or something?

Edit: Or switching to integer 0..100

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
```diff
@@ -0,0 +1,82 @@
...
+        image = .softwareUpdateProgress(progress: CGFloat(progress))
+        bottomTitle = String(format: Localization.percentComplete, 100 * progress)
+
+        if cancel != nil {
```
allendav: A ternary could be used here to good effect

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -42,6 +42,10 @@ final class CardReaderConnectionController {
         ///
         case connectingFailed(Error)
 
+        /// A mandatory update is being installed
```
allendav, line 45: Can probably drop `mandatory` here, right?

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+
+        static let percentComplete = NSLocalizedString(
+            "%.0f%% complete",
+            comment: "Label that describes the completed progress of a software update being installed (e.g. 15% complete). Keep the %.0f%% exactly as is"
+        )
+
+        static let message = NSLocalizedString(
+            "Your card reader software needs to be updated to collect payments. Cancelling will block your reader connection.",
```
allendav, line 71: Will we have separate modals for mandatory and optional updates? Or will optional updates only occur in the settings flow using `WooCommerce/Classes/ViewRelated/ReusableViews/UpdateViewController/UpdateViewController.swift`?

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
```diff
@@ -418,7 +466,15 @@ private extension CardReaderConnectionController {
                 let properties = reader.batteryLevel
                     .map { ["battery_level": $0] }
                 ServiceLocator.analytics.track(.cardReaderConnectionSuccess, withProperties: properties)
-                self.returnSuccess(connected: true)
+                // If we were installing a software update, introduce a small delay so the user can
+                // actually see a success message showing the installation was complete
+                if case .updating(progress: 1) = self.state {
```
allendav, line 475: Building on comment above, as I read this conditional, would now be a good time to stop using `Float`? :) Isn't it possible this conditional will fail if progress isn't exactly equal to 1.00000... ?

WooCommerce/Classes/ViewRelated/ReusableViews/UpdateProgressImage.swift
```diff
@@ -0,0 +1,74 @@
...
+
+        let result = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+
+        return result
+    }
+
+    private static func softwareUpdateProgressFill(progress: CGFloat) -> UIImage? {
+        let rect = CGRect(x: 0, y: 0, width: Constants.size, height: Constants.size)
+        let clippingRect = CGRect(x: 0, y: (1 - progress) * Constants.size, width: Constants.size, height: Constants.size)
```
allendav, line 28: Please consider constraining progress to avoid unexpected values of `y`

WooCommerce/Classes/ViewRelated/ReusableViews/UpdateProgressImage.swift
```diff
@@ -0,0 +1,74 @@
...
+        UIGraphicsBeginImageContext(rect.size)
+        defer { UIGraphicsEndImageContext() }
```
allendav, line 30: TIL. Nice.

WooCommerce/Classes/ViewRelated/ReusableViews/UpdateProgressImage.swift
```diff
@@ -0,0 +1,74 @@
...
+            .softwareUpdateProgressFill(progress: progress),
+            symbol
+        ].compactMap { $0 }) ?? .init()
+    }
+}
+
+private enum Constants {
+    static let size: CGFloat = 91
+    static let borderWidth: CGFloat = 2
+    static let progressColor = UIColor(red: 0.498, green: 0.329, blue: 0.702, alpha: 1)
```
allendav: Could we move this to `UIColor+SemanticColors` ?

WooCommerce/Classes/ViewRelated/ReusableViews/UpdateProgressImage.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+struct UpdateProgressImage_Previews: PreviewProvider {
+    struct UpdateProgressImage: View {
+        @State var complete: CGFloat = 0.5
```
allendav, line 62: Why start at 0.5 and not 0?

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    let primaryButtonTitle: String? = nil
+
+    let secondaryButtonTitle: String? = Localization.cancel
+
+    let auxiliaryButtonTitle: String? = nil
+
+    let bottomTitle: String?
+
+    var bottomSubtitle: String? = Localization.message
```
allendav: The bottomSubtitle prompting them to update and not cancel remains on the screen when the update is complete. Perhaps, like topTitle, it should change (or go away entirely) when the update is complete?

Screen Shot 2021-09-21 at 1 56 24 PM

allendav (changes requested):
Please consider ditching float and also setting up updates through an action in CardPresentPaymentStore. A few other optional comments/questions to also consider.

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
ghost: Yes, I agree with this thanks @allendav @koke

ghost:
This is looking great thanks @koke.
As mentioned, could we please just hide the sub copy here when upload is at 100%
134246686-1fe27c29-addd-4aaa-b3c1-882fc6449088
 
 
The only other thing I noticed was that the up arrow when updating doesn't look centered within the circle. This may however just be from the video posted. Would you mind just double-checking though please?

Thanks!

koke:
> the up arrow when updating doesn't look centered within the circle

Good eye! it does look slightly off-center. I'll look into it

Screen Shot 2021-09-22 at 12 24 49

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
koke: I usually don't leave `print` calls, but I saw a bunch of them in different methods of `StripeCardReaderService` and kept that one for consistency

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
koke: It did feel a bit uncomfortable, but we are setting this value to `1` manually [here](https://github.com/woocommerce/woocommerce-ios/pull/4974/files/4129293b7734170dd21a41b748beb3653af84f38#diff-34012b275288124a9a1c606592b1c55684086c11bf413ab96e2f067fee680f67R454), when we receive a `.completed` event.

I think every time that value is 1 will be because we set a literal `1` in the code. And every time we check, we check against a literal `1`, there are no extra operations there. I'm not an expert in dealing with floating point issues, but it feels safe to me

WooCommerce/Classes/ViewRelated/ReusableViews/UpdateProgressImage.swift
koke: This was just used to preview the image code as I was writing it. 0.5 is because you can see the progress half way every time the preview reloads. In any case, this is only used for SwiftUI previews and it gets removed from the production build

WooCommerce/Classes/ViewRelated/ReusableViews/UpdateProgressImage.swift
koke: Good call, added it in 1ef2c2f2d. I thought of adding a `precondition` there, but I'd rather not crash during a software update, so clamping feels like a better idea in this case. I still added an `assert`, which would crash on debug builds so we can notice any bugs with that

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
koke: I was planning to reuse this and make the message configurable on that PR.

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
koke: Maybe, I'm not sure how useful it is to mention that is mandatory here, but optional updates won't happen on the connection controller, so this state will only ever handle mandatory updates. In other words, it might be superfluous but it's accurate 😁

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
koke: 👍🏽 Done in 9b8a0a094

Hardware/Hardware/CardReader/CardReaderSoftwareUpdateState.swift
koke: I'm not sure `unknown` is useful in this case, or that its usefulness outweighs the added complexity. Given the frequency of updates, it seems OK to assume that `none` is a good default. After all, if we haven't checked yet if there is an update, there is no update available. Maybe there is one in the server, but the same can be true if we just checked 10 seconds ago and Stripe just pushed an update now. 

More generally, I think an unknown state is useful if it's going to be reflected in the UI somehow, which I don't think it's the case here.

WooCommerce/Classes/Extensions/UIImage+Woo.swift
koke: Good catch, added in 063c9b9e1

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
ctarda: If there are prints in the pre-existing code, that's most likely my fault. That sounds like something I'd forget to remove 😊

Hardware/Hardware/CardReader/CardReaderSoftwareUpdateState.swift
koke: Good one: #5036. I missed this initially because the existing code was ignoring errors but was thinking about this as I was working on optional updates and those showed an error message. I'm not sure how we can test the failure case though 🤔

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
koke: Handled in b99dc9355.

WooCommerce/Classes/ViewRelated/ReusableViews/UpdateProgressImage.swift
koke: 👍🏽 Added in 374124528

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
koke: Changed in 4b3e505ec

koke:
I think this is ready for round 2, I hope I didn't miss anything:

https://user-images.githubusercontent.com/8739/134363286-d9f992ad-9a60-49ae-8972-940e999b58ed.mp4

WooCommerce/Classes/ViewRelated/CardPresentPayments/CardReaderConnectionController.swift
allendav: True... for the `1` we set we are probably safe, since we can very likely able to depend on the Swift compiler to encode the `1` the same way twice, and then isEqual should be comparing two identical IEEE-754 floats 🤞 

(Ref https://developer.apple.com/documentation/swift/float/1845934-isequal and https://ieeexplore.ieee.org/document/8766229)

WooCommerce/Classes/ViewModels/CardPresentPayments/CardPresentModalUpdateProgress.swift
allendav: (Resolved below, at least for now :) )

Hardware/Hardware/CardReader/CardReaderSoftwareUpdateState.swift
allendav: A check-in-progress for a (optional) update is in the current UI, but we can cross this bridge in a later PR at which time it might be useful to bind to `.unknown` to show an activity indicator while we are checking

Hardware/Hardware/CardReader/CardReaderSoftwareUpdateState.swift
allendav: > I'm not sure how we can test the failure case though

Could we mock the card reader service update and have it asplode at 0.1 complete?

allendav (approved):
Tests well on iPad. :shipit:

ghost (approved):

Issue #4595, opened 2021-07-13
Feature flag for Shipping Labels M4


pmusolino:
Since the Shipping Labels M4 will implement a lot of features that will be released in the future, we need to hide all the functionalities under different feature flags, that should be added in the codebase.

The M4 has four major parts. They are unrelated, and they will be released in this order:
1. International labels. https://github.com/woocommerce/woocommerce-ios/issues/4596
2. Add payment method. https://github.com/woocommerce/woocommerce-ios/issues/4598
3. Create packages. https://github.com/woocommerce/woocommerce-ios/issues/3909
4. Multiple packages. https://github.com/woocommerce/woocommerce-ios/issues/4599

So, we will follow the same approach used in Android, realizing every feature after they are done, without waiting for the whole milestone completion.

We need to start with international labels, which is the first feature that was released on Android.

More context here: p1624266823345400-slack-CGPNUU63E

designsimply:
### Shipping Labels M4 Test Ideas

I audited results from testing shipping labels in the Andriod app and came up with the following gotchas, scenarios, and ideas.

From Android beta & PR testing

* Check that all expected values are present to avoid strange crashes. (<a href="https://github.com/woocommerce/woocommerce-android/issues/3299">3299</a>)
* Ensure that selecting additional options for a label will update the price accordingly. (<a href="https://github.com/woocommerce/woocommerce-android/issues/4188">4188</a>)
* You should not be able to reprint or refund expired shipping labels. (<a href="https://github.com/woocommerce/woocommerce-android/issues/3845">3845</a>)
* Allow purchasing additional shipping labels. (<a href="https://github.com/woocommerce/woocommerce-android/issues/3856">3856</a>)
* Stores with an empty packages list should still work. (<a href="https://github.com/woocommerce/woocommerce-android/issues/4000">4000</a>)
* Notes can get truncated in languages other than English. (<a href="https://github.com/woocommerce/woocommerce-android/issues/4072">4072</a>)
* Customs forms can be hard to find later—consider combining the shipping label and customs form into one printout if possible.
* Printouts should match the selected formatting (i.e. the paper size you selected was actually used). (<a href="https://github.com/woocommerce/woocommerce-android/pull/3316">3316</a>)
* Keep an eye out for typos, a few (not many though!) made it through to beta testing in Android.

General scenarios to keep in mind

* Order eligibility.
* Accurate pricing & refund amounts.
* Correctly formatted printouts.
* Expired labels.
[...]

PR #4976, opened 2021-09-10
Shipping Labels: Enable feature flag for adding new packages

### LINKED ISSUE #3909: Shipping Labels: add custom package flow
As part of Shipping Labels M4, when creating a shipping label for an order, a store owner will get to a step where they need to set up the package used for the products to be shipped. During this step, we want to allow users to:
- Create a custom package, or
- Activate an existing service / predefined package (the API uses the term predefined, but Woo UI uses the term service)

A custom package has parameters that are completely set by the users, while a service/predefined package is a package with parameters that are already defined by the shipping carriers (USPS or DHL at the time of this writing).

Make sure to follow the release order: https://github.com/woocommerce/woocommerce-ios/issues/4595

@hafizrahman published a retrospective for Android here that explain basically everything important on this task p91TBi-5Jh

Subtasks (feel free to split them into more subtasks):
- [ ] Update the eligibility criteria to accept stores with no packages.
- [x] #4742
- [x] Implement an empty screen for all the cases when the Package Selected screen doesn't have any existing package.
- [x] Create a new screen (preferably in SwiftUI, since this flow is entirely in SwiftUI), which contains the two tab selectors "Custom Package" and "Service Packages". Ideally, they are two separates SwiftUI subviews.
- [x] Create the screen "Custom Package", and check that all the fields in the Custom Package form are mandatory.
- [x] Create the screen "Service Packages": we should do a `GET` on `packages`. It will return `formSchema` and `formData`. `formSchema` is the list of all available service packages. `formData` is the list of all created/activated custom/service packages. We had to compare the two to get the service packages that can still be enabled. We should allow the selection of just one predefined package, and this will be the only difference with `wp-admin`.
- [x] Create the Package Type screen with the list selector containing just the two options `Box` and `Envelope`.
- [x] Use the [existing endpoint](https://github.com/woocommerce/woocommerce-ios/issues/3870) for creating a new Custom Package. It should be probably be edited on our side, for enabling also the predefined packages.

Things to be aware of:
- Activate only one service at a time.
- Empty view design is needed in case all packages are already activated.
- Once a service package is activated, make sure it is then listed in the package selector screen and removed from the service package activation screen (in case the user goes back to activate another service package).

An example JSON request to create two custom packages and activate two service packages at once looks like this:
```
{
    "custom":[
        {
            "is_user_defined":true,
            "is_letter":false,
            "name":"My Own Package",
            "inner_dimensions":"10 x 10 x 10",
            "box_weight":1,
            "max_weight":0
        },
        {
            "is_user_defined":true,
            "name":"test 02",
            "inner_dimensions":"1 x 1 x 1",
            "box_weight":1,
            "max_weight":0
        }
    ],
    "predefined":{
        "usps":[
            "small_flat_box",
            "medium_flat_box_top"
        ],
        "dhlexpress":[
            "Box2Cube",
            "Box2Small",
            "Box2Medium",
            "Box3Large"
        ]
    }
}
```

To create just one custom package, we simply enter one object into the custom array and set predefined as an empty object.

To activate a predefined package, we add the carrier/package name pair in predefined and let the custom part be an empty array.

For custom package, as far as I can tell, `is_user_defined` is always set as `true`. `max_weight` do not seem to be used yet in Woo. In core there is no user-facing field to set max_weight value, and saving seems to always set that value to zero, so I’m replicating that behavior in the Android app.

The Figma file is the source of truth.

Schermata 2021-04-06 alle 12 37 31

rachelmcr:
Closes: #3909

## Description

This PR enables the feature flag for adding/creating new packages (custom and service packages) when creating a shipping label, to release this feature to the public.

**Question:** In https://github.com/woocommerce/woocommerce-ios/issues/4595 it says that we plan to follow that release order, which means releasing the Add Payment Method feature before this Create Package feature. These features aren't dependent on each other, so it should be ok to release this first. Do we want to wait to follow that release order anyway?

## Changes

Updated `shippingLabelsAddCustomPackages` to true for `DefaultFeatureFlagService`.

## Testing

Setup:

* Make sure that your test store has installed WCShip plugin and payment methods added in the "Shipping Labels" section in WPAdmin > WooCommerce > Settings > Shipping > WooCommerce Shipping.
* Remove any packages you have in the "Packaging" section in WPAdmin > WooCommerce > Settings > Shipping > WooCommerce Shipping, so you start testing without any existing packages on your store.

Testing:

4. In the package details section, tap on “Package Selected.”
5. On the Add New Package screen, create a new custom package (on the Custom Package tab) or select a service package (on the Service Package tab).
6. Tap the Done button and confirm your new package is selected on the "Package Selected" screen.
7. Tap the Done button and confirm your new package is listed in the package details section of the shipping label form.
8. Complete your shipping label purchase.
9. Repeat the steps above with another order. Now that you have existing packages on your store, after step 5 you should see a list of the packages already activated on your store with a "Create new package" button at the bottom of the screen. Tap that button to get to the Add New Package screen. After you add a package, confirm the new package is included (and selected) in the package list.

Scenarios to consider:

* Create a new custom package.
* Activate a service package.
* Add multiple packages in a row.
* Cancel adding a new package at any point in the process.
* Activate all of the available service packages and confirm you see an empty state screen on Add New Package > Service Package tab.

## Submitter Checklist

Update release notes:


pmusolino (commented):
I did a thorough test of the functionality, and everything works perfectly. Wonderful work @rachelmcr!
 I just found only one bug similar to the issue found here https://github.com/woocommerce/woocommerce-ios/pull/4970#discussion_r707307287 about decimal values in different countries where the separator for the floating-point numbers is the comma and not the dot (like Italy). In that case, it's impossible to edit the values in the custom package form. Since @itsmeichigo found a solution to this, I'll wait for her to merge the PR, and I'll open a fix for this before releasing this feature. 
Issue: https://github.com/woocommerce/woocommerce-ios/issues/4987

astralbodies:
@pmusolino It looks like the solution was merged, just want to make sure we merge this before code freeze if you're comfortable with it! 😄

pmusolino:
@astralbodies sure! The PR is almost ready, I'll merge it when this PR will be approved https://github.com/woocommerce/woocommerce-ios/pull/5002 😃

pmusolino (approved):
🚀

PR #4980, opened 2021-09-13
Order Editing: Select & update country when editing Shipping Address

Ecarrion:
part of #4780 

# Why

After moving the country synchronization to the EditAddress screen on #4967 now it's time to reflect on the `EditAddress` screen when a new country is selected.

# How

- Expose the `Binding` type to be used in the `WooCommerce` target without having to import `SwiftUI`. Similar to how [`Published` is exposed](https://developer.apple.com/documentation/foundation/published). 

- Expose a binding variable all the way from `CountrySelectorCommand` to the `EditAddressForm`. When this binding is assigned the fields property is repopulated and the trailing button recalculated. I tested several approaches to accomplish this(completion blocks, inout variables, subjects) but this binding approach was the simples and the one that looks cleaner!

- Create a `selectedCountry` published variable to track the latest selected country. This publisher is updated at startup with the initial value from the original address.

# Known Problems

- Everytime a country is selected the view is poped. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4981
- The "Select Country" row should render the country name in body style. https://github.com/woocommerce/woocommerce-ios/projects/39#card-68560797

# Demo

https://user-images.githubusercontent.com/562080/133095408-eeeacc39-635f-47b0-80df-dfb4aa59db22.mov

# Testing Steps

- Navigate to an order and edit its shipping address
- See that the the "select country" row now has the country name on it
- Select a new country
- See that the country name updates and the done button becomes enabled

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Tools/Binding.swift
+++ b/WooCommerce/Classes/Tools/Binding.swift
@@ -1,0 +1,6 @@
+import struct SwiftUI.Binding
+
+/// `Binding` resides inside `SwiftUI`.
+/// This alias allows us to use it inside view models to communicate with `SwiftUI` views without having the import the module on each file.
+///
+typealias Binding = SwiftUI.Binding
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
@@ -112,7 +112,7 @@ struct EditAddressForm: View {
                     }
 
                     Group {
-                        TitleAndValueRow(title: Localization.countryField, value: Localization.placeholderSelectOption, selectable: true) {
+                        TitleAndValueRow(title: Localization.countryField, value: viewModel.fields.country, selectable: true) {
                             showCountrySelector = true
                         }
                         Divider()
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/CountrySelectorCommand.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/CountrySelectorCommand.swift
@@ -1,6 +1,8 @@
 import Foundation
 import Yosemite
 
+import struct SwiftUI.Binding
+
 /// Command to be used to select a country when editing addresses.
 ///
 final class CountrySelectorCommand: ObservableListSelectorCommand {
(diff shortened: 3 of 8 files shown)


WooCommerce/Classes/Tools/Binding.swift
```diff
@@ -0,0 +1,6 @@
+import struct SwiftUI.Binding
+
+/// `Binding` resides inside `SwiftUI`.
+/// This alias allows us to use it inside view models to communicate with `SwiftUI` views without having the import the module on each file.
+///
+typealias Binding = SwiftUI.Binding
```
ealeksandrov, line 6: Nice!

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
```diff
@@ -112,7 +112,7 @@ struct EditAddressForm: View {
                     }
 
                     Group {
-                        TitleAndValueRow(title: Localization.countryField, value: Localization.placeholderSelectOption, selectable: true) {
+                        TitleAndValueRow(title: Localization.countryField, value: viewModel.fields.country, selectable: true) {
```
ealeksandrov, line 115: Shoould we add a placeholder here for empty country?

ealeksandrov (approved):
Works well and code looks good ✅
I've tried to remove duplication of `selectedCountry` that needs to be passed to both textfield and Address constructor, but your solution seems to be the easiest for now 👍

One edge case I've noticed is that wp-admin allows empty country. Therefore:
- Form is missing a placeholder when value is empty.
- We don't have ability to "deselect" a country. Should we have it anyway? I'm not sure. (cc @adamzelinski)

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
Ecarrion: Yes, I'm already tracking this [in this card](https://github.com/woocommerce/woocommerce-ios/projects/39#card-68670618) so I don't forget

Ecarrion:
> We don't have ability to "deselect" a country. Should we have it anyway? I'm not sure. (cc @adamzelinski)

What do you think @adamzelinski? Core kinda lets you select no country, but when you go back to the edit screen it pre-selects the US by default.


https://user-images.githubusercontent.com/562080/133260176-bc228ad0-93e4-417c-b6f0-99bf30ff6e11.mov

ghost:
@Ecarrion I could just add a new row called select a country / region to be consistent with core. Would that solve this issue?

Issue #4953, opened 2021-09-08
[Mobile Payments] Only "remember" the most recently connected reader (i.e. a max of 1 known reader)


allendav:
To harmonize with Android, simplify the implementation and simplify the user experience, let's limit the iOS app to remembering one reader (to which it will auto-connect) instead of many.

This will avoid us having to implement a "clear known readers cache" flow too.

See https://github.com/woocommerce/woocommerce-ios/issues/4934#issuecomment-915262935 for backstory

Issue #4922, opened 2021-09-03
[Mobile Payments] Disconnect Reader button fails silently


koke:
**Describe the bug**

When a reader is connected, tapping the Disconnect Reader button doesn't seem to work on the first attempt

**To Reproduce**
1. Go to Settings > In-Person Payments > Manage Card Reader
2. Tap on Connect Reader and connect to a reader
3. When the reader connects, tap on Disconnect Reader

I can see some error in Xcode's console:

```
2021-09-03 14:20:54.853302+0200 WooCommerce[3472:2018316] 🔵 Tracked card_reader_disconnect_tapped, properties: [AnyHashable("blog_id"): 191646963, AnyHashable("is_wpcom_store"): false]
2021-09-03 14:20:54.857371+0200 WooCommerce[3472:2018484] Unexpected error when disconnecting reader
```

koke:
After some debugging, what seems to be happening is that we start a check for updates as soon as we connect, and that prevents disconnection with an error:

> Error Domain=com.stripe-terminal Code=1000 "Could not execute disconnectReader because the SDK is busy with another command: checkForUpdate." UserInfo={NSLocalizedDescription=Could not execute disconnectReader because the SDK is busy with another command: checkForUpdate., com.stripe-terminal:Message=Could not execute disconnectReader because the SDK is busy with another command: checkForUpdate.}

One option would be to store the cancelable when we check for updates, and cancel it when we try to disconnect. However, I wonder if the checking for updates would also block other functionality, and it might be better to keep the "Connecting..." modal until the update check is finished.

allendav:
I think we could build on #4985 and address both this and #4856 by:

Disable the disconnect button until the update check comes back

and

The row at the top of the screen that appears when a software update is available should display something like "Checking for software updates" to help about re-layout re-juggling and explain why the disconnect button is disabled

and

Show the "update software button" but in a disabled state until the update check comes back (i.e. both the disconnect buttons and update software buttons will be initially visible and disabled)

and

Display a "Your reader is running the latest software" if the update check comes back and there is no update available. Leave the update software button on the screen and disabled in this case as well

Thoughts? It would 1) be more informative for the user 2) encourage updating out-of-date readers and 3) avoid the screen content jumping around

cc @adamzelinski  @Garance91540

ghost:
Those suggestions sound good to me @allendav

allendav:
[...]

PR #4985, opened 2021-09-13
[Mobile Payments] Disable disconnect button and show spinner while disconnecting from reader

### LINKED ISSUE #4984: [Mobile Payments] Disable and show activity indicator on disconnect button while disconnecting
Disconnecting from the card reader may take a few seconds. The disconnect button should be disabled and show an activity indicator while that is happening.

Related: p91TBi-5XR-p2

allendav:
Closes #4984 

Changes:
- Swaps out the UIButton in the ButtonTableViewCell for a ButtonActivityIndicator

Note:
- Does NOT address https://github.com/woocommerce/woocommerce-ios/issues/4953 - if you have multiple known readers you will get placed immediately in the search flow on disconnecting

To test:
- Connect to a card reader (or use the simulator)
- Disconnect from it
- Observe that the disconnect button is disabled (and an activity indicator replaces the title) during disconnect, which can take a second or two
- Note that similar logic has been added to disable the update button during an update or disconnect
- Try a few other buttons in table views in the app and ensure they continue to work properly (since ButtonTableViewCell is used in a few places, e.g. "Connect Card Reader" is one such button)

Movie:

https://user-images.githubusercontent.com/1595739/133163899-240ba125-1d9d-4cdf-9e41-2787027d4377.MP4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
@@ -213,6 +213,12 @@ private extension CardReaderSettingsConnectedViewController {
         cell.configure(style: .primary, title: Localization.updateButtonTitle, bottomSpacing: 0) {
             self.viewModel?.startCardReaderUpdate()
         }
+
+        let readerDisconnectInProgress = viewModel?.readerDisconnectInProgress ?? false
+        let readerUpdateInProgress = viewModel?.readerUpdateInProgress ?? false
+        cell.enableButton(!readerDisconnectInProgress && !readerUpdateInProgress)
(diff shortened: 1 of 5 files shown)


ghost (approved):

koke (commented):
I'll note that this doesn't solve the issue described in #4922, and the disconnect button still doesn't work right after connecting. This PR doesn't target that issue directly, but the reason that disconnect doesn't work is because the SDK is busy checking for updates. I think we should add "checking for updates" to the list of reasons that make the button disabled, since it happens every time you connect.

koke (approved):
Actually, I'll approve this since it's a good improvement, and my previous comment can be addressed in a separate PR

allendav:
> I think we should add "checking for updates" to the list of reasons that make the button disabled, since it happens every time you connect.

I was thinking that last night too. I'll make that change in a subsequent PR. Thanks for the review!

Issue #4384, opened 2021-06-07
Errors: Offline mode banner


rachelmcr:
**Context:** We should really reconsider our error states, providing more specific errors, an info button to help guide resolution on common errors, and provide easy direction to a support ticket

**Design:** The “Unable to load content”/”Unable to fetch list” error message could also be triggered by the fact that the app isn’t connected to internet. In that case, we should clearly display that the device is offline, like we do on Android with the “Offline – using cached data” banner:

![image](https://user-images.githubusercontent.com/8658164/121001245-039a0200-c783-11eb-8066-5efcfd077a2f.png)

Internal ref: p91TBi-3Ao-p2#comment-2101

itsmeichigo:
Task checklist:
- [ ] Add new service to check connectivity.
- [ ] Add new banner view for offline status.
- [ ] Extend UIViewController to reuse logic for connectivity check.
- [ ] Apply connectivity check for all screens that require network and contain cached data:
  - [ ] My Store
  - [ ] Order list
  - [ ] Order details
  - [ ] Shipping Label purchase flow
  - [ ] Product list
  - [ ] Product forms
  - [ ] Review list
  - [ ] Review details

PR #4986, opened 2021-09-14
Add new service for observing connectivity status

itsmeichigo:
Part of #4384 

# Description
This PR adds a new service for observing network connectivity status.

# Changes
- Adds new protocol `ConnectivityObserver` along with enumerators `ConnectivityStatus` and `ConnectionType`.
- Adds default implementation for observing connectivity with `DefaultConnectivityObserver`. This implementation wraps ~Alamofire's implementation of `NetworkReachabilityManager`~ Network framework's NWPathMonitor.
- Updates `ServiceLocator` with access point to `ConnectivityObserver` for network monitoring to works properly.

# Testing
Since `NWPathMonitor` is a final class and `NWPath` isn't opened for creating - mocking for unit tests is impossible. I'll keep this service untested - behavior will be tests along with UI updates in the next PR.

Update release notes:


pmusolino (approved):
Beautiful 💯 thanks @itsmeichigo for having used `NWPathMonitor` instead of the implementation of `NetworkReachabilityManager` under Alamofire.

Ecarrion:
> Since NWPathMonitor is a final class and NWPath isn't opened for creating - mocking for unit tests is impossible. I'll keep this service untested - behavior will be tests along with UI updates in the next PR.

You could test the logic in your service by creating a thin protocol around `NWPathMonitor` and using that type for the init instead of the concrete type.

That could also help in case we face problems during Unit Test or UI Test because of unwanted no-network situations.

itsmeichigo:
Thank you @Ecarrion for the suggestion! I thought of proxying `NWPathMonitor` but then got stuck with `NWPath` because there's no accessible initializer for the struct. But now that I think of it, I can proxy `NWPath` too! I'll create a separate PR for the unit tests.

PR #4988, opened 2021-09-14
Issue 4854: Order Details titles should be title case

### LINKED ISSUE #4854: Order Details: titles should be title case not sentence case
> The case on Shipment details is inconsistent with the View Shipment Details row in the shipping labels section. The d in details should be capitalized.

image

<sup>(internal reference: p5T066-2yX-p2#comment-9588 /hat tip @astralbodies)</sup>

joshheald:
Fixes #4854 

## Description
The title string "Shipping details" should be title case. This change fixes the casing, to make it consistent with the rest of the titles on the Order Details screen. 

Additionally, I've moved the relevant localizable strings to the Localization enum.

## Testing
On a store with an order, navigate to `Orders > Order #x` and scroll down to the `Shipping Details` section.
Shipping Details

Update release notes:

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -236,6 +236,7 @@ GEM
     memoist (0.16.2)
     mini_magick (4.11.0)
     mini_mime (1.1.1)
+    mini_portile2 (2.6.1)
     minitest (5.14.4)
     molinillo (0.6.6)
     multi_json (1.15.0)
@@ -244,6 +245,9 @@ GEM
     nap (1.1.0)
     naturally (2.2.1)
     netrc (0.11.0)
+    nokogiri (1.12.2)
+      mini_portile2 (~> 2.6.1)
+      racc (~> 1.4)
     nokogiri (1.12.2-x86_64-darwin)
       racc (~> 1.4)
     octokit (4.21.0)
@@ -328,6 +332,7 @@ GEM
     xctest_list (1.2.1)
 
 PLATFORMS
+  -darwin-20
   x86_64-darwin-19
   x86_64-darwin-20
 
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -346,4 +351,4 @@ DEPENDENCIES
   xcpretty-travis-formatter
 
 BUNDLED WITH
-   2.2.25
+   2.2.27
```
joshheald: @koke Just noticed that this had been changed... presumably because of doing this on an M1, though I am a bit surprised because I thought I got the right versions of Ruby and Bundler. I'll try to figure out what's going on here.

Gemfile.lock
```diff
@@ -244,7 +245,8 @@ GEM
     nap (1.1.0)
     naturally (2.2.1)
     netrc (0.11.0)
-    nokogiri (1.12.2-x86_64-darwin)
+    nokogiri (1.12.2)
```
joshheald: @koke I've got the bundler version exactly the same as previous change to the Gemfile.lock now. 
These changes are down to this not being `x86_64` when built on M1...

itsmeichigo: I'm also working on a M1 so I think I should share some of my thoughts. Since most of the team aren't using M1 machines, I tend to not commit the changes in that file to avoid causing troubles for peeps trying to build the project (it's a bit annoying I know). I think @ealeksandrov does the same (he uses a M1 too).

ealeksandrov: Yes, I have a set of lockfile changes that I always ignore.
I think when using 2.7.x Ruby it looks better (`arm64-darwin-20` instead of `-darwin-20`) - so my vote is to ignore any arm64 changes for now and update Gemfile.lock after planned Ruby version bump (ref: paaHJt-2pO-p2).

koke (approved):
Looks good to me. Thanks for reverting the bundler changes, and congrats on that first PR! 🎉

PR #4989, opened 2021-09-14
Add/Edit Products: product image stops loading on product image upload error

### LINKED ISSUE #3134: Add/Edit Products: product image keeps loading on product image upload error
**Describe the bug**

When product image upload fails, the image cell still stays in loading state which looks like something is going on in the background.

**To Reproduce**

- Go to the Products tab
- Add or edit a product
- Tap on the "+" cell or any image cell in the images header
- Tap "Add Photos"
- Choose a photo or take one with camera, and quickly switch to offline mode --> after a bit, an error alert is shown and the image cell is still in loading state after dismissing the alert

**Screenshots**

**Expected behavior**

The image cell that fails to upload should not be visible.

**Mobile Environment**
Please include:
- Device: iPhone 11
- iOS version: iOS 14.1
- WooCommerce iOS version: 5.5

pmusolino:
Fixes #3134 

Previously, when product image upload fails, the image cell still stays in the loading state.

## Testing
1. Go to the Products tab
2. Add or edit a product
3. Tap on the "+" cell or any image cell in the images header
4. Tap "Add Photos"
5. Choose a photo or take one with camera, and quickly switch to offline mode --> after a bit, an error alert is shown and the image cell should stop the loading state.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,7 @@
 
 7.6
 -----
-- [*] Fix: when product image upload fails, the image cell stop loading.
+- [*] Fix: when product image upload fails, the image cell stop loading. [https://github.com/woocommerce/woocommerce-ios/pull/4989]
 
 7.5
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,7 +2,7 @@
 
 7.6
 -----
-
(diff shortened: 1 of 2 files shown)


itsmeichigo (approved):
Code looks good and works as expected 👍 Thanks for fixing the issue!

(I confirmed the fix on simulator - I can't seem to get the timeout error on my device and I'm not sure why. Since the fix makes sense, I'm approving it anyway.)

PR #4992, opened 2021-09-15
[Mobile Payments] Add display of reader software version to connected view

### LINKED ISSUE #4990: [Mobile Payments] Display complete reader software version after connecting in settings
See https://github.com/woocommerce/woocommerce-ios/issues/4661#issuecomment-891166865

Note: We have decided to display the entire BBPOS software version string, e.g. "1.00.03.34-SZZZ_Generic_v45-300001" on a new line

pdfdoF-91-p2

allendav:
Closes #4990 

To test:
- Connect a reader and observe the connected software version. Test in landscape, portrait, light and dark modes.
- Ensure all unit tests continue to pass, and newly added tests pass as well

Changes:
- Added label for software version. Also added a missing unit test for battery level.

Screenshots:

Update release notes:


allendav:
@joshheald - if you don't have a reader yet, enable the simulator under Run > Scheme to connect to a simulated one

ghost (approved):

joshheald:
> @joshheald - if you don't have a reader yet, enable the simulator under Run > Scheme to connect to a simulated one

Let's chat about this at our 1:1 later as well... I've managed to get a simulator connecting to the simulated terminal and "taken payment" for an order, but that was _only_ via the `Orders` screen route, not the settings.

For some reason, on settings I only get a section `Store Settings` with an `In-Person Payments` cell that crashes the app because of a learn more link whenever I tap it. I'm sure that's not part of your changes but it does make it tricky to test!

joshheald:
So I hacked around the crash, and all looks good with the version string 🎉 ... but I did notice that when the reader needs a software update, and the `Disconnect Reader` becomes the secondary button, it doesn't behave well in Dark Mode. This is also the case on `develop`, so it looks like it's a pre-existing issue, but not recorded anywhere yet. I'll log an issue for it. Also, using develop I don't get the crash when going into the IPP screen any more 🥳 

Light Mode | Dark Mode
-------|-------
Light mode with updates needed | Dark mode with updates needed

Will take a look over the code too after lunch.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -178,5 +190,15 @@ private extension CardReaderSettingsConnectedViewModel {
...
             comment: "Card reader battery level as an integer percentage"
         )
+
+        static let unknownSoftwareVersion = NSLocalizedString(
+            "Unknown Software Version",
+            comment: "Displayed in the unlikely event a card reader has an indeterminate software version"
+        )
+
+        static let versionLabelFormat = NSLocalizedString(
+            "Version: %1$@%",
```
joshheald: I don't think the final % here is needed...
It doesn't print `%`, probably because it's being read as the start of a new (incomplete) placeholder. I'd worry a little that a future change to `string(with format:)` may not be so accommodating, and then we might get crashes or something.
It would need to be `%%` to print `%`, but I don't think we want that either.

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/CardReaderSettings/CardReaderSettingsConnectedViewModelTests.swift
```diff
@@ -39,6 +39,30 @@ final class CardReaderSettingsConnectedViewModelTests: XCTestCase {
...
+        let mockStoresManager = MockCardPresentPaymentsStoresManager(
+            connectedReaders: [MockCardReader.bbposChipper2XBT()],
+            discoveredReaders: [],
+            sessionManager: SessionManager.testingInstance
+        )
+        ServiceLocator.setStores(mockStoresManager)
+
+        let viewModel = CardReaderSettingsConnectedViewModel(didChangeShouldShow: nil)
+        XCTAssertEqual(viewModel.connectedReaderSoftwareVersion, "Version: 1.00.03.34-SZZZ_Generic_v45-300001")
+    }
```
joshheald: Perhaps there should also be tests for the unknown strings here? But I'm not sure on our approach to that, so if it's not the done thing, just say 😊

joshheald (approved):
Found [the issue you made for the crash](https://github.com/woocommerce/woocommerce-ios/issues/4995)! Hooray 😄 

The changes look good to me, but I did spot a small issue with the format string and make a suggestion about tests. Pre-emptively approved since we're in quite different time zones!

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: OMG - thanks for catching that! Copy pasta from `batteryLabelFormat` :) Fixed in https://github.com/woocommerce/woocommerce-ios/pull/4992/commits/61182b09e6bbdc6b804620dad32eb96ada85e874

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/CardReaderSettings/CardReaderSettingsConnectedViewModelTests.swift
allendav: Glad to add them. https://github.com/woocommerce/woocommerce-ios/pull/4992/commits/e0e0c81bf71502bf0de3b0746fa0505cbdc982e3

allendav:
Thank you @joshheald - I've made the corrections requested and I'll look into the dark mode formatting with #5008

PR #4993, opened 2021-09-15
Adds Store picker error modal UI

Ecarrion:
part of #4383 

# Why

For HACK Week, I'm working on improving the error shown when a merchant can't select a store.
Currently, it is just a notice and we will turn it into a modal popover.
This PR just adds the UI.

Note: Accessibility will be tackled later.

# How

- Added a `SwiftUI` view with the required components & Styles
- Added a hosting view controller so it can be presented later from the store picker view controller.

# Screenshots

Light | Dark
--- | ---
Screen Shot 2021-09-14 at 8 16 34 PM | Screen Shot 2021-09-15 at 8 12 57 AM

# Testing Steps
- In Xcode go to the `StorePickerError` file
- See the view rendered in the Preview Provider.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #4383, opened 2021-06-07
Errors: Modal message for errors on site picker screen


rachelmcr:
**Context:** We should really reconsider our error states, providing more specific errors, an info button to help guide resolution on common errors, and provide easy direction to a support ticket

**Design:** Instead of the existing notice, we can display it in a modal:

![image](https://user-images.githubusercontent.com/8658164/121000725-72c32680-c782-11eb-8888-f2f21f17e0f3.png)

Whenever users select “Troubleshoot” in the error message, they’ll load a page on WC docs with a list of things they could try out before reaching out to support.

Internal ref: p91TBi-3Ao-p2#comment-2101

Task:
- [x] Build UI https://github.com/woocommerce/woocommerce-ios/pull/4993
- [x] Connect Troubleshoot https://github.com/woocommerce/woocommerce-ios/pull/4994
- [x] Connect Contact support https://github.com/woocommerce/woocommerce-ios/pull/4994
- [x] Present & dismiss Modal #5006 
- [x] Handle Accessibility and rotation #5009

Ecarrion:
Hi @designsimply, do you know would be a good website to send the users once they press the "Troubleshooting Tips" button?

We currently have [this in the app](https://docs.woocommerce.com/document/android-ios-apps-troubleshooting-error-fetching-orders/) but it is related to fetching data once you have selected a store.

PR #4994, opened 2021-09-15
Adds closures to respond to action buttons

Ecarrion:
part of #4383

# Why

After #4993, this PR makes sure to add the necessary actions to make the UI useful.

- Launch the troubleshooting URL using a `SafaryViewController`
- Begin a `Zendesk` request 
- Dismiss the view controller.

Note: This VC is supposed to be presented as a modal with a custom presentation context but that will be done in the next PR, once I connect it to come from the real store picker.

# How

- Add closures on `StorePickerError` to be assigned from `StorePickerErrorHostingController`
- Added a factory function to create `StorePickerErrorHostingController` configured with propper actions.

# Demo

https://user-images.githubusercontent.com/562080/133471128-ae65fd55-1cca-44a7-a80c-22caeb2ff985.mov

# Testing Steps

In the `ViewDidLoad` of `DashboardViewController` add this code

```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
    let viewController = StorePickerErrorHostingController.createWithActions(presenting: self)
    self.present(viewController, animated: true)
}
```

- See that you can navigate to troubleshooting tips and to contact support.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Epilogue/StorePickerError.swift
+++ b/WooCommerce/Classes/Authentication/Epilogue/StorePickerError.swift
@@ -7,32 +7,39 @@ final class StorePickerErrorHostingController: UIHostingController<StorePickerEr
 
     /// Creates an `StorePickerErrorHostingController` with preconfigured button actions.
     ///
-    static func createWithActions(presentingViewController: UIViewController) -> StorePickerErrorHostingController {
-        .init(
-            troubleshootingAction: {
-                let safariViewController = SFSafariViewController(url: WooConstants.URLs.troubleshootErrorLoadingData.asURL())
-                presentingViewController.present(safariViewController, animated: true)
-            },
-            contactSupportAction: {
-                ZendeskManager.shared.showNewRequestIfPossible(from: presentingViewController)
-            },
-            dismissAction: {
-                presentingViewController.dismiss(animated: true)
+    static func createWithActions(presenting: UIViewController) -> StorePickerErrorHostingController {
+        let viewController = StorePickerErrorHostingController()
+        viewController.setActions(troubleshootingAction: {
+            let safariViewController = SFSafariViewController(url: WooConstants.URLs.troubleshootErrorLoadingData.asURL())
+            viewController.present(safariViewController, animated: true)
+        },
+        contactSupportAction: {
+            presenting.dismiss(animated: true) {
+                ZendeskManager.shared.showNewRequestIfPossible(from: presenting)
             }
-        )
+        },
+        dismissAction: {
+            presenting.dismiss(animated: true)
+        })
+        return viewController
     }
 
-    init(troubleshootingAction: @escaping () -> Void, contactSupportAction: @escaping () -> Void, dismissAction: @escaping () -> Void) {
-        super.init(rootView: StorePickerError(troubleshootingAction: troubleshootingAction,
-                                              contactSupportAction: contactSupportAction,
-                                              dismissAction: dismissAction))
+    init() {
+        super.init(rootView: StorePickerError())
     }
 
     required dynamic init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-}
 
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/Authentication/Epilogue/StorePickerError.swift
```diff
@@ -1,21 +1,62 @@
 import SwiftUI
+import SafariServices
 
 /// Hosting controller wrapper for `StorePickerError`
 ///
 final class StorePickerErrorHostingController: UIHostingController<StorePickerError> {
+
+    /// Creates an `StorePickerErrorHostingController` with preconfigured button actions.
+    ///
+    static func createWithActions(presenting: UIViewController) -> StorePickerErrorHostingController {
```
itsmeichigo, line 10: Nit: `presenting` sounds a bit ambiguous to me. Do you think `presentingViewController` may be clearer?

WooCommerce/Classes/Authentication/Epilogue/StorePickerError.swift
```diff
@@ -1,21 +1,62 @@
...
         fatalError("init(coder:) has not been implemented")
     }
-}
 
+    /// Actions are set in a separate function because most of the time, they will require to access `self` to be able to present new view controllers.
+    ///
+    func setActions(troubleshootingAction: @escaping () -> Void, contactSupportAction: @escaping () -> Void, dismissAction: @escaping () -> Void) {
+        self.rootView.troubleshootingAction = troubleshootingAction
+        self.rootView.contactSupportAction = contactSupportAction
+        self.rootView.dismissAction = dismissAction
```
itsmeichigo, line 40: Nit: I think `self` can be removed here?

WooCommerce/Classes/Authentication/Epilogue/StorePickerError.swift
```diff
@@ -1,21 +1,62 @@
...
+    ///
+    var troubleshootingAction: () -> Void = {}
+
+    /// Closure invoked when the "Contact Support" button is pressed
+    ///
+    var contactSupportAction: () -> Void = {}
+
+    /// Closure invoked when the "Back To Sites" button is pressed
+    ///
+    var dismissAction: () -> Void = {}
```
itsmeichigo, line 58: I'm not sure if it's a good practice to set default actions to empty, as normally we want the actions to work. This can cause trouble if one forgets to update the variables. Maybe it's safer to require these for initializing the view?

itsmeichigo (approved):
I have tested and this works as expected ✅ 

However there's a case that I couldn't catch in #4993 since we're only testing a real modal here - in landscape mode the view doesn't seem to display full content - maybe a scroll view is needed here:

Simulator Screen Shot - iPhone 12 Pro - 2021-09-16 at 16 29 40

I'm also leaving some non-blocking comments below. Feel free to merge this and address the issues in the next PR!

WooCommerce/Classes/Authentication/Epilogue/StorePickerError.swift
Ecarrion: Not sure, I took the name from the [UIViewControllerTransisioningDelegate protocol](https://developer.apple.com/documentation/uikit/uiviewcontrollertransitioningdelegate/1622037-animationcontroller) which call their parameters `presenting` and `presented`. So I guess it's fairly standard!

WooCommerce/Classes/Authentication/Epilogue/StorePickerError.swift
Ecarrion: We can't really pass them **only** upon initialization because most of the time those closures will need to access `self`. You can however still use the `initWithClosures` as they would be synthesized by Xcode.

To mitigate your worry there is a factory method that creates the `HostingViewController` with the appropriate closures set.

Ecarrion:
Thanks for the review @itsmeichigo and for checking the landscape problem. 
I have identified that as a task in the main issue https://github.com/woocommerce/woocommerce-ios/issues/4383 and will tackle accessibility + rotation in the last PR!

PR #4996, opened 2021-09-15
[Mobile Payments] Remove use of attributed string for learn more link

### LINKED ISSUE #4995: [Mobile Payments] Crash may occur on entering In-Person Payments
Highly intermittent. Props @jaclync 

> I can’t reproduce it consistently but it happened twice on launches about half an hour apart, the first one right after updating the app from TestFlight. It looks like the crash is on a calculated localizable attributed string for onboarding (the Learn More string)

https://github.com/woocommerce/woocommerce-ios/blob/c6f4f07ba4d7d39d6ba5362b0d6e9f4350f2d6b8/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person%20Payments/InPersonPaymentsLearnMore.swift#L39

Props @koke for analysis pointing the finger at the use of `NSAttributedString` which itself uses `htmlToAttributedString` which can evoke the crash if the asynchronous HTML parsing completes while the view itself is in the middle of updates. paNNhX-gK-p2

allendav:
Closes #4995 

Changes:
- Rolls back use to simply localized string and makes the entire "learn more" row tappable (to open the web view)

Videos:

https://user-images.githubusercontent.com/1595739/133498261-19908e7c-dce2-4958-bfd1-9b4295406add.MP4

https://user-images.githubusercontent.com/1595739/133498275-5219ecab-2ea5-4483-a0d0-3949cd52c322.MP4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsSearchingViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsSearchingViewController.swift
@@ -227,9 +227,7 @@ private extension CardReaderSettingsSearchingViewController {
     }
 
     private func configureLearnMore(cell: LearnMoreTableViewCell) {
-        cell.configure(text: Localization.learnMore) { [weak self] url in
-            self?.urlWasPressed(url: url)
-        }
+        cell.configure(text: Localization.learnMore)
         cell.backgroundColor = .systemBackground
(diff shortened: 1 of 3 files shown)


ghost (approved):

koke (approved):
I don't love the big link, but given how problematic this is turning out to be, I'm ok with the tradeoff for now. Just keep in mind that this is not the only place that uses `htmlToAttributedString`, although the crashes only seem to have started in the reader screens

PR #4997, opened 2021-09-15
Don't request line_items in the order list

joshbetz:
In the order list view, we don't show details about the individual
products in the order. Requesting full details about all the products
can have a significant impact on performance.

Here we differentiate between fields we need for single orders and a
list of orders. For now the only difference is that single orders do
load the line items.

### DIFF
--- a/Networking/Networking/Model/Order.swift
+++ b/Networking/Networking/Model/Order.swift
@@ -133,7 +133,7 @@ public struct Order: Decodable, GeneratedCopiable, GeneratedFakeable {
         let paymentMethodID = try container.decode(String.self, forKey: .paymentMethodID)
         let paymentMethodTitle = try container.decode(String.self, forKey: .paymentMethodTitle)
 
-        let items = try? container.decode([OrderItem].self, forKey: .items)
+        let items = try? container.decodeIfPresent([OrderItem].self, forKey: .items) ?? []
 
         let shippingAddress = try? container.decode(Address.self, forKey: .shippingAddress)
         let billingAddress = try? container.decode(Address.self, forKey: .billingAddress)
--- a/Networking/Networking/Model/Order.swift
+++ b/Networking/Networking/Model/Order.swift
@@ -60,7 +60,7 @@ public struct Order: Decodable, GeneratedCopiable, GeneratedFakeable {
                 totalTax: String,
                 paymentMethodID: String,
                 paymentMethodTitle: String,
-                items: [OrderItem],
+                items: [OrderItem]?,
                 billingAddress: Address?,
                 shippingAddress: Address?,
                 shippingLines: [ShippingLine],
@@ -91,7 +91,7 @@ public struct Order: Decodable, GeneratedCopiable, GeneratedFakeable {
         self.paymentMethodID = paymentMethodID
         self.paymentMethodTitle = paymentMethodTitle
 
-        self.items = items
+        self.items = items ?? []
         self.billingAddress = billingAddress
         self.shippingAddress = shippingAddress
         self.shippingLines = shippingLines
@@ -133,7 +133,7 @@ public struct Order: Decodable, GeneratedCopiable, GeneratedFakeable {
         let paymentMethodID = try container.decode(String.self, forKey: .paymentMethodID)
         let paymentMethodTitle = try container.decode(String.self, forKey: .paymentMethodTitle)
 
-        let items = try container.decode([OrderItem].self, forKey: .items)
+        let items = try? container.decode([OrderItem].self, forKey: .items)
 
         let shippingAddress = try? container.decode(Address.self, forKey: .shippingAddress)
         let billingAddress = try? container.decode(Address.self, forKey: .billingAddress)
--- a/Networking/Networking/Remote/OrdersRemote.swift
+++ b/Networking/Networking/Remote/OrdersRemote.swift
@@ -28,7 +28,7 @@ public class OrdersRemote: Remote {
                 ParameterKeys.page: String(pageNumber),
                 ParameterKeys.perPage: String(pageSize),
                 ParameterKeys.statusKey: statusKey ?? Defaults.statusAny,
-                ParameterKeys.fields: ParameterValues.fieldValues,
+                ParameterKeys.fields: ParameterValues.listFieldValues,
             ]
 
             if let before = before {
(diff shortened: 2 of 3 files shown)


Networking/Networking/Model/Order.swift
```diff
@@ -133,7 +133,7 @@ public struct Order: Decodable, GeneratedCopiable, GeneratedFakeable {
         let paymentMethodID = try container.decode(String.self, forKey: .paymentMethodID)
         let paymentMethodTitle = try container.decode(String.self, forKey: .paymentMethodTitle)
 
-        let items = try container.decode([OrderItem].self, forKey: .items)
+        let items = try? container.decode([OrderItem].self, forKey: .items)
```
Ecarrion: To avoid modifying the initializer and to not hide/shadow a possible parsing error, you could do 

```suggestion
        let items = try? container.decodeIfPresent([OrderItem].self, forKey: .items) ?? []
```

Networking/Networking/Remote/OrdersRemote.swift
```diff
@@ -199,11 +199,16 @@ public extension OrdersRemote {
     }
 
     enum ParameterValues {
-        static let fieldValues: String = """
+        static let listFieldValues: String = """
```
Ecarrion, line 203: We should probably add a comment telling the reader that the only field missing is `line_items` to make this easy to understand in the future!

Ecarrion (approved):
Congrats on your first? PR!

Changes look good, I only left some minor suggestions!

Testing worked well but I didn't find any improvements on the request loading times for my test store, both(this branch and develop) were in the `0.8s~1s` ranges, not sure if there is some cache going around tho.

The experience is a bit less ideal since now there is another section that needs to be filled and jumps tho.

https://user-images.githubusercontent.com/562080/133508244-686afcbd-cb10-4a63-8f02-acaf843e4e8e.mov

What loading times are you seeing @joshbetz?

joshbetz:
For each test (with and without line_items), I made two requests in a row. The second request was a bit faster each time, but regardless, there is a clear improvement (~33% in this case) by removing line_items when it's not needed on this test site.

With line_items:

```
$ curl -w 'ttfb: %{time_starttransfer}\ntotal: %{time_total}' -so /dev/null -u '<redacted>:<redacted>' 'https://teststore.betzster.com/wp-json/wc/v3/orders?_fields=id,parent_id,number,status,currency,customer_id,customer_note,date_created_gmt,date_modified_gmt,date_paid_gmt,discount_total,discount_tax,shipping_total,shipping_tax,total,total_tax,payment_method,payment_method_title,shipping,billing,coupon_lines,shipping_lines,refunds,fee_lines,line_items'
ttfb: 1.254984
total: 1.299505

$ curl -w 'ttfb: %{time_starttransfer}\ntotal: %{time_total}' -so /dev/null -u '<redacted>:<redacted>' 'https://teststore.betzster.com/wp-json/wc/v3/orders?_fields=id,parent_id,number,status,currency,customer_id,customer_note,date_created_gmt,date_modified_gmt,date_paid_gmt,discount_total,discount_tax,shipping_total,shipping_tax,total,total_tax,payment_method,payment_method_title,shipping,billing,coupon_lines,shipping_lines,refunds,fee_lines,line_items'
ttfb: 0.927129
total: 0.971201
```

Without line_items:

```
$ curl -w 'ttfb: %{time_starttransfer}\ntotal: %{time_total}' -so /dev/null -u '<redacted>:<redacted>' 'https://teststore.betzster.com/wp-json/wc/v3/orders?_fields=id,parent_id,number,status,currency,customer_id,customer_note,date_created_gmt,date_modified_gmt,date_paid_gmt,discount_total,discount_tax,shipping_total,shipping_tax,total,total_tax,payment_method,payment_method_title,shipping,billing,coupon_lines,shipping_lines,refunds,fee_lines'
ttfb: 0.874923
total: 0.875409

$ curl -w 'ttfb: %{time_starttransfer}\ntotal: %{time_total}' -so /dev/null -u '<redacted>:<redacted>' 'https://teststore.betzster.com/wp-json/wc/v3/orders?_fields=id,parent_id,number,status,currency,customer_id,customer_note,date_created_gmt,date_modified_gmt,date_paid_gmt,discount_total,discount_tax,shipping_total,shipping_tax,total,total_tax,payment_method,payment_method_title,shipping,billing,coupon_lines,shipping_lines,refunds,fee_lines'
ttfb: 0.614476
total: 0.617481
```

joshbetz:
One issue that should definitely get fixed though, is that we see the "Loading content" and jump even if you go back and forth between the list of orders and a given order multiple times. Should that get cached locally after we initially load an order?


https://user-images.githubusercontent.com/300429/133514615-b5c51a41-6905-46fd-86bc-c0cd6319ed59.mp4

joshbetz:
@Ecarrion 21bb361a6519e33e008b3a3fbd1a6b2618dec174 "fixes" the issue I mentioned above. I think it would be better to let `items` be `nil` to differentiate between an empty array and the field not existing. What are your thoughts?

Ecarrion:
@joshbetz 

>  I think it would be better to let items be nil to differentiate between an empty array and the field not existing. What are your thoughts?

I think that makes sense, then, we should probably do the same for `shipping` that I saw you are also removing!

joshbetz:
shipping address is already optional https://github.com/woocommerce/woocommerce-ios/blob/develop/Networking/Networking/Model/Order.swift#L36

joshbetz:
@Ecarrion Any suggestions on the best way to make that optional? It seems like there are a lot of places that use `items` that will need to be fixed.

Ecarrion:
> @Ecarrion Any suggestions on the best way to make that optional? It seems like there are a lot of places that use `items` that will need to be fixed.

Hey @joshbetz, I don't think there is an easy and clean way to handle that update. 🤔 

_Maybe,_ we could create a [property wrapper](https://www.swiftbysundell.com/articles/property-wrappers-in-swift/) that returns an empty array if the array is `nil`, but still allows us to know if the underlying value is `nil` or not.

Since property wrappers are a controversial topic, I think the best way to move forward is to leave the code with your change, and add a comment on why we need that `items.count > 0` check(btw we can also use `items.isEmpty`.)

Then create an issue where we explain the issue and the property wrapper solution so the team can debate it!

What do you think?

joshbetz:
Ok cool, that seems reasonable to me.

Ecarrion:
Hey @joshbetz do you need something to get this merged?

joshbetz:
@Ecarrion If it looks good to you, I'm happy to merge.

Ecarrion:
@joshbetz Yup, I approved it a while ago 😅 

Maybe just adding the comment we discussed here https://github.com/woocommerce/woocommerce-ios/pull/4997/files#diff-07ad979afe48bc7bcfd986f5c452866c93cc06133b6e13aa34adcec76ac2c31eR71 and then merge away!

PR #4998, opened 2021-09-15
Add Events for "Jetpack Required" Screen

shiki:
When I was analyzing the login data (p91TBi-5Yh-p2), I found out that some events for the "Jetpack required" screen are not tracked by iOS. These are tracked by Android. 

## Changes

I added tracking for the events to match Android's tracked events. 

Image 2021-09-15 16-51-49

Image 2021-09-15 16-56-42

## Testing

1. Log out.  😛 
2. In the login screen, tap Enter Your Store Address.
3. Enter a site URL that has no Jetpack. 
4. Confirm `login_jetpack_required_screen_viewed` is tracked. 
5. Tap See Instructions.
6. Confirm `login_jetpack_required_view_instructions_button_tapped` is tracked. 
7. Dismiss the window. Unless you want to read the page. Feel free! 🤪 
8. Tap What Is Jetpack?
9. Confirm `login_what_is_jetpack_help_screen_viewed` is tracked. 
10. Tap Learn more about Jetpack.
11. Confirm `login_what_is_jetpack_help_screen_learn_more_button_tapped` is tracked. 
12. Dismiss the browser window.
13. Tap Continue.
14. Confirm `login_what_is_jetpack_help_screen_ok_button_tapped` is tracked. 

If you can, please compare with Android's events too. 

## Reviewing

🏳️ Please nitpick me to the death. I am rusty. 😁 I'm not in a hurry to get this merged. 

Only 1 reviewer is needed but anyone can review.

## Author Checklist

- [x] If it's feasible, I have added unit tests. 
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.
- [x] Update Tracks events spreadsheet

PR #4999, opened 2021-09-15
Standardize Spacing in WooAnalyticsStat

shiki:
The [`WooAnalyticsStat`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/Analytics/WooAnalyticsStat.swift) event declarations are spaced inconsistently. Some blocks align the equal signs to each other. Some don't. 

As we can observe in https://github.com/woocommerce/woocommerce-ios/pull/4998, aligning the equal signs is not great for viewing diffs. It's hard to tell which lines have changed if we consistently update the spacing. 

image

## Changes

Here, I'm proposing that we just stick to one convention in this file. I used no alignment, like some of the already existing blocks there. Viewing diffs easily is more important IMO. 

## Testing

The build passing should be enough. 

## Reviewing

## Author Checklist

- [x] Merge https://github.com/woocommerce/woocommerce-ios/pull/4998 before merging this


pmusolino (approved):
I had never noticed that in some places the indentation was wrong! Thanks for fixing that.

PR #5001, opened 2021-09-16
4890 Center empty state in space below top banner

### LINKED ISSUE #4900: Empty State and Error Banners Vertical Alignment can be Better
## Describe the bug

Found in #4890. There's an issue with how the error banners (https://github.com/woocommerce/woocommerce-ios/pull/4407) interact with the rest of the empty state content. For example, the content does not automatically resize or expand if the banner content is expanded. 

Default (Looks Fine) | Expanded ([Sad Trombone](https://www.youtube.com/watch?v=yJxCdh1Ps48))
--------|-------
 | 

## To Reproduce

These steps were taken from #4407. 

1. Build and run a fresh install of the app (to begin without any cached data).
2. Log in to the app.
3. Disconnect your internet connection. (This is so you will trigger a sync error; you can also disrupt your store's Jetpack connection for a more realistic scenario.)
4. Select Orders tab.
5. Notice the error banner appears at the top of the Orders empty state screen on both the "Processing" and "All Orders" tabs.
6. Tap the expand button. You'll notice that the empty state will no longer be vertically centered. 

## Expected behavior

The remaining vertical space should be filled by the empty state. Both the error banner and the empty state can be aligned better together. 

## Notes

- Please make sure that the solution works with large font sizes too
- Avoid regressions from the changes in https://github.com/woocommerce/woocommerce-ios/pull/4407 and https://github.com/woocommerce/woocommerce-ios/pull/4407

joshheald:
Fixes #4900 

## Description
When the error view top banner showed on a screen with the EmptyStateViewController, the centering of the empty state stack view was still based on the total content view, rather than the visable space between the top banner and the bottom of the content view.

## Changes
This change provides a layout guide for the vertical center of the visible area, updated when the top banner is shown, expanded, or hidden. The empty state stack view has a centerY constraint equal to this guide's centerY anchor.

The >= 16 constraints for the top and bottom remain because of the keyboard centering adjustments, which come in to play on empty search results screens. Without these, a keyboard showing notification can push the empty state views outside the bounds of the content view and overlapping other views.

### Reproduction
These steps were taken from #4407 via #4900.

1. Build and run a fresh install of the app (to begin without any cached data).
2. Log in to the app.
3. Disconnect your internet connection. (This is so you will trigger a sync error; you can also disrupt your store's Jetpack connection for a more realistic scenario.)
4. Select Orders tab.
5. Notice the error banner appears at the top of the Orders empty state screen on both the "Processing" and "All Orders" tabs.
6. Tap the expand button. The empty state will now remain vertically centered.

Before | After
-------|-------
Before | After

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/EmptyStateViewController/EmptyStateViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/EmptyStateViewController/EmptyStateViewController.swift
@@ -98,6 +98,21 @@ final class EmptyStateViewController: UIViewController, KeyboardFrameAdjustmentP
     /// Used to present the Contact Support dialog if the `Config` is `.withSupportRequest`.
     private let zendeskManager: ZendeskManagerProtocol
 
+    /// Used to pin the stackView to the center of the visible area in the contact view,
+    /// taking account of the banner view when it's shown.
+    private lazy var visiblyCenteredLayoutGuide: UILayoutGuide = {
+        let centerGuide = UILayoutGuide()
+        contentView.addLayoutGuide(centerGuide)
+        let topConstraint = centerGuide.topAnchor.constraint(equalTo: contentView.topAnchor)
+        topConstraint.priority = UILayoutPriority(250)
+        NSLayoutConstraint.activate([
+            topConstraint,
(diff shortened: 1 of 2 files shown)


koke (approved):
Looks good to me, and I learned about UILayoutGuide 🎉

PR #5002, opened 2021-09-16
Shipping Labels: fix decimal values in different countries in custom package creation

pmusolino:
## Description
While I was testing https://github.com/woocommerce/woocommerce-ios/pull/4976 before enabling the feature flag to all our users, I discovered https://github.com/woocommerce/woocommerce-ios/pull/4976#pullrequestreview-753707349 a bug about decimal values in different countries where the separator for the floating-point numbers is the comma and not the dot (like in Italy). In that case, it's impossible to edit the values in the custom package form. 

In this PR, I fixed the issue using the new extension of `NumberFormatter` introduced by @itsmeichigo previously. Plus, I make sure to send always the dot to the backend instead of a comma when we send the dimensions property.

## Testing
1. Change your device location to Italy (or another country) where the decimal separator is the comma `,`.
2. Open an order detail that is eligible for a shipping label.
3. Tap the button "Create Shipping Label.”
4. Confirm "Ship From" and "Ship To.”
5. In the package details section, tap on “Package Selected.”
6. On the Add New Package screen, create a new custom package (on the Custom Package tab).
7. Add all the info needed, and make sure you are able to insert decimal values with a comma.
7. Tap the Done button and confirm your new package is selected on the "Package Selected" screen.

Repeat the testing procedure with a country where the decimal separator is the dot `.`.

Update release notes:


itsmeichigo (approved):
Everything looks and works great, thanks for spotting and fixing the issue! 👏

PR #5004, opened 2021-09-16
Accessibility labels on edit icons in Order Details

### LINKED ISSUE #5003: Edit buttons on Order Details don't all have Accessibility Labels
**Describe the bug**
Found during #4988 
Edit buttons in the cell accessory view on the Order Details screen, particularly `Shipping Address` and `Customer Provided Note` do not have any accessibility labels.
Accessibility on accessory labels

**To Reproduce**
Steps to reproduce the behavior on simulator:
1. Enable the Accessibility inspector and build on simulator
2. Go to 'Orders' tab
3. Tap on an order
4. Scroll down to 'Customer Provided Note' and 'Shipping Details'
5. Inspect pencil icon (button) on right hand side of cells
6. Element is only identified as "Button"

Steps to reproduce the behavior on device:
1. Go to 'Orders' tab
2. Tap on an order
3. Scroll down to 'Customer Provided Note' and 'Shipping Details'
4. Enable VoiceOver
5. Single-tap pencil icon (button) on right hand side of cells
6. Element is only identified as "Button"

**Screenshots**
Accessibility on accessory labels

**Expected behavior**
These buttons should have an Accessibility Label describing their purpose:
1. Update Note
2. Update Address

These labels will be read aloud by VoiceOver when in use.

**Mobile Environment**
Please include:
- Device: iPhone 12 mini
- iOS version: iOS 14.5
- WooCommerce iOS version: 7.5

joshheald:
Fixes #5003 

## Description
The two icon-only edit buttons (✎) for Shipping Address and Customer Note in the Order Details screen did not have accessibility labels, which meant that VoiceOver would read them as "Button" with no context or information about what the cell accessory button does.

This adds an accessibility label to these two buttons, briefly describing their function.

## Testing

Steps to reproduce the behavior on simulator:
1. Enable the Accessibility inspector and build on simulator
2. Go to 'Orders' tab
3. Tap on an order
4. Scroll down to 'Customer Provided Note' and 'Shipping Details'
5. Inspect pencil icon (button) on right hand side of cells
6. Element should be identified with specific label

Steps to reproduce the behavior on device:
1. Go to 'Orders' tab
2. Tap on an order
3. Scroll down to 'Customer Provided Note' and 'Shipping Details'
4. Enable VoiceOver
5. Single-tap pencil icon (button) on right hand side of cells
6. Element should be identified with specific label

Customer Note
Shipping Address

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -382,6 +382,10 @@ private extension OrderDetailsDataSource {
         cell.onEditTapped = { [weak self] in
             self?.onCellAction?(.editCustomerNote, nil)
         }
+
+        cell.editButtonAccessibilityLabel = NSLocalizedString(
+            "Update Note",
+            comment: "Accessibility Label for the edit button to change the Customer Provided Note in Order Details")
     }
 
     private func configureBillingDetail(cell: WooBasicTableViewCell) {
@@ -788,6 +792,10 @@ private extension OrderDetailsDataSource {
         cell.onEditTapped = orderEditingEnabled ? { [weak self] in
             self?.onCellAction?(.editShippingAddress, nil)
         } : nil
+
+        cell.editButtonAccessibilityLabel = NSLocalizedString(
+            "Update Address",
+            comment: "Accessibility Label for the edit button to change the Customer Shipping Address in Order Details")
     }
 
     private func configureShippingMethod(cell: CustomerNoteTableViewCell) {
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/CustomerInfoTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/CustomerInfoTableViewCell.swift
@@ -56,6 +56,17 @@ class CustomerInfoTableViewCell: UITableViewCell {
         }
     }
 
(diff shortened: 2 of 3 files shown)


koke (approved):
I tested it and worked great. A couple of potential improvements I noticed:

- Since the section says "Shipping Details", maybe the label should say "Update shipping details" instead of address
- If you go into "View Billing Information", there's another `Button` without a label there to edit the billing address

joshheald:
Thanks @koke  😊 

> * Since the section says "Shipping Details", maybe the label should say "Update shipping details" instead of address

Yeah I did this first, and then decided to shorten it, because I think it's good practice to keep the a11y labels terse, for speed of navigation while using VoiceOver. My reasoning for this being ok, is that it's immediately after the Shipping Details in the flow, so has the context from that.

I'm definitely not an expert in this stuff though, so I'm happy to look into it further if you're worried that it's not clear enough.

> * If you go into "View Billing Information", there's another `Button` without a label there to edit the billing address

Good spot, I'll fix that one too :)

koke:
> My reasoning for this being ok, is that it's immediately after the Shipping Details in the flow, so has the context from that.

That makes sense 👍🏽

joshheald:
Nice one. I'm going to do the billing address stuff on a new issue and PR because there are other a11y things to improve in there: the phone number and email buttons are using accessibilityHint rather that accessibilityLabel, which is slower to tell the user what they do and something to be used very sparingly, according to our contact with Apple on this pbArwn-2d9-p2-comment-3174

PR #5005, opened 2021-09-16
Add a lane to register new device and update certs and profiles

AliSoftware:
Add a `register_new_device` lane similar to the same lane we have in WPiOS, but this one would be updating the WC profiles instead.

Since the typical process is to call the `register_new_device` from the WPiOS repo first, I made it possible in this WC lane to skip the device registration if the device was already added (typically via the WPiOS lane), and directly jump to the actions updating the certificates and profiles.

I've also updated the instructions in our FieldGuide page (ref: PCYsg-B2Y-p2) accordingly.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -21,8 +21,8 @@ SCREENSHOT_DEVICES = [
   'iPad Pro (12.9-inch) (3rd generation)'
 ].freeze
 
-MAIN_BUNDLE_IDENTIFIERS = %[com.automattic.woocommerce]        # Registered in our main account, for development and AppStore
-ALPHA_BUNDLE_IDENTIFIERS = %[com.automattic.alpha.woocommerce] # Registered in our Enterprise account, for AppCenter / Installable Builds
+MAIN_BUNDLE_IDENTIFIERS = %w[com.automattic.woocommerce]        # Registered in our main account, for development and AppStore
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -21,6 +21,9 @@ SCREENSHOT_DEVICES = [
   'iPad Pro (12.9-inch) (3rd generation)'
 ].freeze
 
+MAIN_BUNDLE_IDENTIFIERS = %[com.automattic.woocommerce]        # Registered in our main account, for development and AppStore
+ALPHA_BUNDLE_IDENTIFIERS = %[com.automattic.alpha.woocommerce] # Registered in our Enterprise account, for AppCenter / Installable Builds
```
AliSoftware: Contrary to WPiOS which have many bundleIDs because of its various App Extensions (Widgets, ShareExt, etc), WooCommerce currently only has that single one (at least to my knowledge).

I think that's why we never bothered to add that `register_new_device` lane in WCiOS before, as this felt not as necessary and could be easily done manually compared to WPiOS which has so many profiles to be updated.
Still, I feel this is still very useful to have even for a single bundleID, both because it makes it easy to scale this if we end up adding App Extensions to WCiOS one day, but also because that makes it a no-brainer just having to run the same lane for WCiOS as WPiOS, instead of wondering which of the various profiles in ADC would need to be updated when doing it manually 🙂

AliSoftware:
Note: I tested the lane in practice while addressing pdnsEh-11-p2 (which highlighted an crash I fixed in `b905861` before re-running it), and after that fix, it worked as expected 🎉

jkmassel (approved):

PR #5006, opened 2021-09-16
Show StorePickerError as a modal.

Ecarrion:
part of #4383 

# Why

After actions were wired in #4994, now it's time to properly show the error screen as a modal and wire it in its final place.

# How

- Updated the view to add a `Group` as a parent view to allow us to visualize the padding at the side edges. This is in order to not clip the rounded corners.

- Show the screen as a modal using a custom transitioning delegate.

# Demo

### iPhone

https://user-images.githubusercontent.com/562080/133638463-d8a6843b-a249-4a42-a2e5-6f9197d5abea.mov

### iPad

https://user-images.githubusercontent.com/562080/133638431-260ed9b4-55f3-4e84-a68e-cc50046c9cd8.mov

# Testing Steps

- Navigate to the store picker screen
- Turn off wifi in order to simulate a network error
- Select another store
- See error appearing as a modal.

Update release notes:


jaclync (approved):
LGTM! :shipit: 

Code: ✅ 
Testing: ✅ (rotation will be fixed in https://github.com/woocommerce/woocommerce-ios/pull/5009)

non-blocking observation: maybe "Back to Sites" could be "Back to Store Picker" in case there is only one site in the picker?

Ecarrion:

PR #5008, opened 2021-09-16
[Mobile Payments] [WIP] Disable disconnect button while checking for reader software update

### LINKED ISSUE #4922: [Mobile Payments] Disconnect Reader button fails silently

allendav:
Fixes #4922 

Changes:
- Update status remains on screen the entire time, showing whether it is checking for updates, has an update, or if the reader is already running the latest software
- Update button is disabled unless an update is available, in which case it is enabled and primary
- Disconnect button is disabled while checking for whether an update is available. If the check completes and no update is available, it is made primary

@adamzelinski  - I've changed the info image to an activity spinner while we are checking for updates. I'll posted that movie here below. Please let me know if there other design changes you'd expect (see movies below)

@koke - I've based this off of issue/4990-display-version in order to pick up the changes there - once that is reviewed this can be rebased and merged to develop

Movie (with spinner added)

https://user-images.githubusercontent.com/1595739/133701488-a5db5db4-7a2a-4fff-91e9-7fc0ef27289f.MP4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
@@ -92,13 +92,15 @@ private extension CardReaderSettingsConnectedViewController {
 
         /// This section displays whether or not there is update for the reader software
         ///
-        sections.append(
-            Section(title: nil,
-                    rows: [
-                        .updatePrompt
-                    ]
-            )
-        )
+        let checkForReaderUpdateInProgress = viewModel?.checkForReaderUpdateInProgress ?? false
+        var rows = [Row]()
+        if checkForReaderUpdateInProgress {
+            rows = [.checkingForUpdate]
+        } else {
+            rows = [.updatePrompt]
+        }
+
+        sections.append(Section(title: nil, rows: rows))
 
         /// This section displays details about the connected reader
         ///
@@ -165,6 +167,8 @@ private extension CardReaderSettingsConnectedViewController {
     ///
     func configure(_ cell: UITableViewCell, for row: Row, at indexPath: IndexPath) {
         switch cell {
+        case let cell as ActivitySpinnerAndLabelTableViewCell where row == .checkingForUpdate:
+            configureCheckingForUpdate(cell: cell)
         case let cell as LeftImageTableViewCell where row == .updatePrompt:
             configureUpdatePrompt(cell: cell)
         case let cell as ConnectedReaderTableViewCell where row == .connectedReader:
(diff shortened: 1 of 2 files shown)


allendav:
Hey @joshheald - I was looking at the dark mode comment you posted [here](https://github.com/woocommerce/woocommerce-ios/pull/4992#issuecomment-921763603) and I cannot reproduce it with this branch - can you give it a spin? This is what I see:

ghost (approved):

joshheald:
> I cannot reproduce it with this branch - can you give it a spin? 

Yep, looks good to me too @allendav 

iPhone | iPad
------|------
Screenshot 2021-09-20 at 08 54 42 | Screenshot 2021-09-20 at 09 00 46

koke (approved):
I think this will work in most cases, but the state would be better kept on the service and not the view model. For instance, when you visit the settings page, it will start checking for an update. If you quickly navigate out of that view and go back in:
- The view triggers a second check for updates while the first one is still in progress
- The second check fails immediately because the SDK is busy with the first one
- It fails silently as if there was no update found, and the disconnect button becomes enabled
- If you tap to disconnect, the first check is still running, so the SDK is busy and it fails silently

I think the PR is good, but commenting more generally on the feature, I feel like we're giving too much weight to something that, while important, is quite infrequent. Every time you load that screen or connect to a reader, a new check is launched with prevents you from disconnecting until that's finished. The `checkForUpdate` operation is cancelable, so I think it might be better if we checked for updates as often as we wanted, but canceled that operation as soon as the user requested a different thing (disconnect, collect payment,...). I'd also hide the "Update Reader Software" button instead of having it disabled.

allendav:
> I think this will work in most cases, but the state would be better kept on the service and not the view model.

I think so too. 

> If you quickly navigate out of that view and go back in:

Good point! I'll open a issue for moving this responsibility into the service

> I feel like we're giving too much weight to something that, while important, is quite infrequent.

We are. The service could keep track of update checks and return a cached value if the last check was say within a day, with a pull to refresh allowing force refresh.

Another good idea for moving this to the service layer especially with SDK 2 changes

https://github.com/woocommerce/woocommerce-ios/issues/5024

PR #5009, opened 2021-09-17
StorePickerError: Landscape & Big Fonts support 

### LINKED ISSUE #4383: Errors: Modal message for errors on site picker screen

Ecarrion:
closes #4383 

# Why 

After #5006 is merged, the final bits to tackle are to properly support landscape and accessibility fonts.

This however turned to not be simple. Thanks `SwiftUI` :rage4: 

# How

## Landscape

The first issue is that apparently a `HostingViewController` presented with a `.custom` `modalPresentationStyle` does not inherit rotation support. 

After inspecting the view with the debugger, I noticed that the hosting controller view frame wasn't being updated properly.

To fix, this I'm creating a subclass of `FancyAlertPresentationController` where I manually update the `presentedView` in the `containerViewWillLayoutSubviews` method.

## Scroll & Big Fonts

Normally, in UIKIt, we would add a scroll view as a parent view of the stack view to make the content scroll if it overflows the scroll view size. This was not so trivial in `SwiftUI`. Adding a scroll view crates two problems: A. The scroll view takes all the available size, making the stack view content to be rendered at the top and not centered and B. There is no way to prevent the scroll view to scroll when the view is small to not need from a scroll.

To solve that, I've created a `ViewModifier` that wraps the content inside a scroll view if the view needs to scroll(this is achieved by using a couple of geometry readers.

Note: This view modifier only works for vertical scroll but it should be easy to adapt it to handle horizontal scroll.

# Demo

## Landscape

https://user-images.githubusercontent.com/562080/133705412-1b23c88f-03bf-4504-a4c2-54fc61bf0b84.mov

## Big Fonts

https://user-images.githubusercontent.com/562080/133705465-d6d7a231-cd1c-499a-bdf5-3eee04648370.mov

# Testing Steps

- Navigate to the store picker screen
- Turn off wifi in order to simulate a network error
- Select another store
- See error appearing as a modal.
- See that you can rotate it and scroll all the content.

Update release notes:


WooCommerce/Classes/View Modifiers/View+ScrollModifiers.swift
```diff
@@ -0,0 +1,54 @@
...
+    /// Updates the `contentSize` property by adding a clear background to the `content` that has a `GeometryReader` attached to it.
+    /// Updates the `shouldScroll` property when the content vertical geometry is greater than the parent vertical geometry.
+    ///
+    private func contentGeometryListener(content: Content, parentGeometry: GeometryProxy) -> some View {
+        content
+            .background(
+                GeometryReader { contentGeometry -> Color in
+                    DispatchQueue.main.async {
+                        contentSize = contentGeometry.size
+                        shouldScroll = contentGeometry.size.height > parentGeometry.size.height
```
jaclync, line 38: wow TIL'ed adding a clear background view with `GeometryReader` to achieve conditional scrolling 😮

WooCommerce/Classes/View Modifiers/View+ScrollModifiers.swift
```diff
@@ -0,0 +1,54 @@
...
+            )
+    }
+}
+
+// MARK: View Extensions
+
+extension View {
+    /// Allows the view to scroll vertically when the content height is greater than it's parent height.
+    ///
+    func scrollIfNeeded() -> some View {
```
jaclync: this is very cool!

nit: do you think if we might add horizontal scroll support to this extension in the future, maybe with optional params for scroll directions? For now, I wonder if `scrollVerticallyIfNeeded` would be more clear about the scroll direction 🙂

jaclync (approved):
Nice work figuring out these SwiftUI issues! 🥇 I tested on both iPhone and iPad simulators and rotation & dynamic type works perfectly!

For `UIHostingViewController` rotation issue with custom presentation, I tried wrapping it in a `UIViewController` and it seems to work ([testing branch](https://github.com/woocommerce/woocommerce-ios/compare/issue/4383-modal-presentation...issue/4383-modal-rotation-testing?expand=1)). Still a workaround though, I'm good with your approach since it's reusable - up to you!

koke:
Just dropping a note to point to a recent similar case I worked on with [ScrollableVStack](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI%20Components/ScrollableVStack.swift), which I think solves the first problem (stack view filling all the available space), but not the second (scrolling only when the content is bigger than the space available).

Ecarrion:
Thanks, @jaclync for the review!

> For UIHostingViewController rotation issue with custom presentation, I tried wrapping it in a UIViewController and it seems to work (testing branch). Still a workaround though, I'm good with your approach since it's reusable - up to you!

Nice, I didn't think about that, it would have saved me tons of time 🤦 

@koke 

> Just dropping a note to point to a recent similar case I worked on with ScrollableVStack, which I think solves the first problem (stack view filling all the available space), but not the second (scrolling only when the content is bigger than the space available).

Nice, I totally missed that component. I wonder how can we make those components more visible to the rest of the team 🤔 

Also, while I was creating this modifier I struggled if I should make it a regular `View` or a `ViewModifier`.

In the end, I went for the `ViewModifier` route because I believe it would be easy to update legacy views with this modifier. 

What is your reasoning for choosing one over the other?

koke:
> What is your reasoning for choosing one over the other?

None really, I didn't think of view modifiers. I would guess years of inertia subclassing things, I think the view modifier approach looks somewhat better in this case 😁

PR #5011, opened 2021-09-17
Add throttling of orders sync on every viewWillAppear

ealeksandrov:
## Description

This PR adds timeout check on orders sync action that is triggered on every `viewWillAppear`.
It doesn't impact manual pull-to-refresh and new order notification.

### Technical Details

Implementation is simple:
1. Save current datetime after each successful fetch of 1st page.
2. Check saved datetime if sync for 1st page is triggered from `viewWillAppear`.

## Test

1. Add some `print()` lines (OrderListViewController.swift:298) or launch a proxy.
2. Switch Stats-Orders-Stats-Orders tabs multiple times.
3. Observe that only one set of GET `/orders` is requested, on first open.
4. Wait 30s and switch tabs, observe non-throttled `/orders` requests.
5. Immediately trigger pull-to-refresh, observe non-throttled `/orders` requests.
6. Launch app on device, open orders tab and create new order from computer within refresh timeout.
7. Observe that orders list is refreshed immediately.

### Note

In the proxy you can still see requests on every tab open for `/reports/orders/totals` path. It's a known issue that is handled separately.

---
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/OrderListSyncActionUseCase.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrderListSyncActionUseCase.swift
@@ -75,9 +75,9 @@ struct OrderListSyncActionUseCase {
 
     /// The reasons passed to `SyncCoordinator` when synchronizing.
     ///
-    /// We're only currently tracking one reason.
     enum SyncReason: String {
         case pullToRefresh = "pull_to_refresh"
+        case viewWillAppear = "view_will_appear"
     }
 
     let siteID: Int64
--- a/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
@@ -71,6 +71,13 @@ final class OrderListViewController: UIViewController {
     ///
     private let syncingCoordinator = SyncingCoordinator()
 
+    /// Timestamp for last successful sync.
+    ///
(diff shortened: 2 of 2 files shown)


Ecarrion (approved):
This looks good to me!

----

Now that we are thinking more and more about Combine, I wonder if we could create a trigger subject that performs the network request and then add `debounce/throttle` on that trigger.

It will help us to keep the state contained in that publisher!

What do you think?

ealeksandrov:
Thanks for review @Ecarrion!

> It will help us to keep the state contained in that publisher!

This is a good idea, but we can't keep the logic contained on networking level, since we have different behavior on app level - no timeout for notifications and pull-to-refresh updates, for example. We can probably pass such arguments down to Yosemite-Networking (and I've tried it) but haven't found good clean and universal solution yet.

Ecarrion:
> This is a good idea, but we can't keep the logic contained on networking level, since we have different behavior on app level - no timeout for notifications and pull-to-refresh updates, for example. We can probably pass such arguments down to Yosemite-Networking (and I've tried it) but haven't found good clean and universal solution yet.

Hey @ealeksandrov, I tried to do this using the debounce / throttle operator but had troubles due to my lack of Rx knowledge, however, I did found a way to use combine and have all the needed states in the same scope.

[This is the branch](https://github.com/woocommerce/woocommerce-ios/compare/ec/throttled-orders?expand=1), FYI I'm not expecting you to change what is implemented. I only did it for learning purposes! 😅

PR #5012, opened 2021-09-17
[Mobile Payments] Remember at most one card reader

### LINKED ISSUE #4933: [Mobile Payments] Don't kick off automatic searches right after disconnecting from a reader
Right now, if you've successfully connected to a reader (and never explicitly disconnected from it), then, on entering Manage Card Reader, the reader search will automatically begin.

This does beg one question - if we keep this feature - if I have successfully connected to say ReaderA in the past, and I'm connected to ReaderB right now, if I disconnect from ReaderB, we probably shouldn't kick off an automatic search for ReaderA - right afterwards - right? That would probably seem unexpected / confusing to users.

Thoughts? @woocommerce/mobile-designers  @ctarda

### LINKED ISSUE #4953: [Mobile Payments] Only "remember" the most recently connected reader (i.e. a max of 1 known reader)

allendav:
Closes #4953 
Closes #4933 

Changes:
- Remembers at most one card reader in app settings
- "Forget" action no longer needs a card reader ID to forget (it just resets the array)

Note:
- For backwards compatibility, the storage in app settings is still an array
- If multiple reader IDs are found in app settings (i.e. persisted by a version prior to this change), just the last one added is returned

To test:
- Get two readers
- Initially reader A powered on, reader B off
- Connect to reader A (this will cause the app to "remember" just reader A)
- Exit the app
- Power off reader A, power on reader B
- Start the app and enter the connection flow. Search should begin automatically since there is a remembered reader
- Verify the app finds reader B and asks you if you'd like to connect to it. Connect to it.
- Exit the app
- Power on reader A, power off reader B
- Start the app and enter the connection flow. Search should begin automatically since there is a remembered reader
- Verify the app DOES NOT connect to reader A automatically any more (it should prompt you since it is no longer remembered - at this point only reader B is remembered)
- Do NOT connect to reader A
- Exit the app
- Power off reader A, power on reader B
- Start the app and enter the connection flow. Search should begin automatically since there is a remembered reader and it should connect automatically to reader B
- Disconnect from reader B
- Ensure search does not auto restart (since there are now no remembered readers)
- Exit the app
- Start the app and enter the connection flow. Search should NOT begin automatically since there is NO remembered reader

Update release notes:


allendav:
@ctarda or @koke - a review from either of you would be awesome - both of you are not, however, expected :)

koke (commented):
I like the idea of keeping the array for backwards compatibility, in case someone already has an array in the persisted settings. However, I think we can contain that to `GeneralAppSettings` and make sure that the methods from `AppSettingsStore` and upwards deal with a single optional value instead. What do you think about that?

allendav:
@koke wrote:

> I think we can contain that to GeneralAppSettings and make sure that the methods from AppSettingsStore and upwards deal with a single optional value instead

Agreed. Good idea. That changed the size of the PR a bit, but it is for the better. See https://github.com/woocommerce/woocommerce-ios/pull/5012/commits/013eddaf8c7628670edf19c852dafaa92212bf2a

koke (approved):
I still only have one reader so I can't test properly, but code looks good to me

PR #5017, opened 2021-09-20
Shipping Labels: Rename PackageItem to SinglePackage to avoid ambiguity

itsmeichigo:
Part of #4599 

# Description
As pointed out by @rachelmcr in #4959, the naming `ShippingLabelPackageItem` can be mistaken into an order item, so this PR aims to rename the view to remove the ambiguity.

# Changes
- Renames `ShippingLabelPackageItem` to `ShippingLabelSinglePackage`
- Renames `ShippingLabelPackageItemViewModel` to `ShippingLabelSinglePackageViewModel`.

# Testing
All unit tests should still pass. Configuring package details for shipping label should work properly as before:

4. Tap Continue on Package Details row. Edit the weight to an invalid value, notice that a validation error shows up and Done button is disabled.
5. Update the weight to a valid value and tap Done. Notice on the purchase form that the correct package name and weight is displayed. Tap on Package Details row again, notice that correct package details are displayed.

Update release notes:


pmusolino (approved):
LGTM :shipit: 

Code ✅ 
Testing ✅

PR #5019, opened 2021-09-20
Billing information accessibility improvements

### LINKED ISSUE #5007: Edit buttons on Order Details > Billing Information don't all have Accessibility Labels
**Describe the bug**
Found during #5004  
Buttons in the cell accessory view on the Billing Information screen, particularly the edit button for `Billing Address` has no accessibility label or hint, while the buttons by `Email` and `Phone Number` have an accessibility hint rather than a label.

Accessibility hints are to be used "very sparingly" (pbArwn-2d9-p2-comment-3174), as they are for providing additional context, not describing the UI. The practical implication of using a hint is that VoiceOver will read out last, after the label and control type. In the case of the "meatballs" icon next to the phone number, VoiceOver says "Button. {pause} Prompts with the option to call or message the billing customer."

There are some differences here; for the cells in the `Contact Details` section (phone and email), selecting the cell performs the indicated action, whereas for the billing address, tapping on the pencil icon itself is required. This may mean that `accessibilityHint` is the only option for the whole-cell actions.

**To Reproduce**
Steps to reproduce the behavior on simulator:
1. Enable the Accessibility inspector and build on simulator
2. Go to 'Orders' tab
3. Tap on an order
4. Tap `View Billing Information`
5. Inspect the icons (button) on right hand side of cells
6. Element is only identified as "Button", but may also have a hint afterwards

Steps to reproduce the behavior on device:
1. Go to 'Orders' tab
2. Tap on an order
3. Tap `View Billing Information`
4. Enable VoiceOver
5. Inspect the icons (button) on right hand side of cells
6. Element is only identified as "Button", but may also have a hint afterwards

**Screenshots**
CleanShot 2021-09-16 at 17 20 39@2x

**Expected behavior**
These buttons should have an Accessibility Label describing their purpose:
1. Update Address
2. Call or Message
3. Compose Email

These labels will be read aloud by VoiceOver when in use.

**Mobile Environment**
Please include:
- Device: iPhone 12 mini
- iOS version: iOS 14.5
- WooCommerce iOS version: 7.5

joshheald:
Fixes #5007

## Description
The accessibility of the Billing Information screen was incomplete/needing improvement. This change adds an accessibility label for the edit address button, and makes the actions to call, message, or email, the customer easier to notice and access.

## Testing
1. Navigate to `Orders > any order > View Billing Information`
2. Activate VoiceOver
3. Tap on, or flick to, the Pencil edit address icon
4. Observe the edit button is announced as "Update address, button"
5. Tap or flick to the Contact Details cells, and swipe up or down to access actions.
6. Observe that the contact details cells are announced with "Actions available", which when you swipe between them, expose "Call", "Message", "Email", "Copy Email Address"

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
@@ -321,6 +321,9 @@ private extension BillingInformationViewController {
         cell.onEditTapped = editingEnabled ? {
             print("Edit Billing Address Tapped")
         } : nil
+        cell.editButtonAccessibilityLabel = NSLocalizedString(
+            "Update Address",
+            comment: "Accessibility Label for the edit button to change the Customer Billing Address in Billing Information")
     }
 
     func setupBillingPhone(cell: WooBasicTableViewCell) {
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
@@ -195,7 +195,10 @@ private extension BillingInformationViewController {
     }
 
     private func copyEmailHandler() {
-        guard let email = order.billingAddress?.email else { return }
+        guard let email = order.billingAddress?.email else {
+            return
+        }
+
         ServiceLocator.analytics.track(.orderDetailCustomerEmailTapped)
         sendToPasteboard(email, includeTrailingNewline: false)
     }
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
```diff
@@ -188,6 +193,12 @@ private extension BillingInformationViewController {
 
         ServiceLocator.analytics.track(.orderDetailCustomerEmailMenuTapped)
     }
+
+    private func copyEmailHandler() {
+        guard let email = order.billingAddress?.email else { return }
```
allendav: We usually break the return out on a line by itself

allendav:
Tests fine. One suggestion. :shipit:

https://youtu.be/acxpmqEjbAk

allendav (approved):
Tests fine. One suggestion. :shipit:

https://youtu.be/acxpmqEjbAk

WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
joshheald: Thanks, good spot. I'd misremembered the exception for `guard self = self` as applying to all single line guard statements.

joshheald:
@allendav thanks for the video of your tests 😊
I think that kind of video is better because it shows the interactions you're making, but In case you didn't know, the built in recording function on iOS does also record the VoiceOver audio and selection boxes, so that works pretty well too. See the video on [this ticket](https://github.com/woocommerce/woocommerce-ios/issues/5022) for an example.

PR #5020, opened 2021-09-20
Order Editing: Make address form modal

ealeksandrov:
## Description
This PR changes presentation style of Address Edit Form from navigation to modal as previously discussed.

Ref: p1630933894035500-slack-CGPNUU63E

### Technical Details
SwiftUI navigation strikes again with broken nav links. To keep them working we need parent wrapper like `ZStack` instead of nesting in `NavigationView` directly.

## Screenshots
before|after
--|--

## Testing
1. Open existing order.
2. Tap "edit" icon on Shipping Address row.
3. Observe correct modal presentation style.
4. Tap on country and check that navigation inside form works.
5. Try dismissing with swipe gesture and "Close" button.

---
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
@@ -9,6 +9,14 @@ final class EditAddressHostingController: UIHostingController<EditAddressForm> {
 
     init(viewModel: EditAddressFormViewModel) {
         super.init(rootView: EditAddressForm(viewModel: viewModel))
+
+        // Needed because a `SwiftUI` cannot be dismissed when being presented by a UIHostingController
+        rootView.dismiss = { [weak self] in
+            self?.dismiss(animated: true, completion: nil)
+        }
+
+        // Set presentation delegate to track the user dismiss flow event
+        presentationController?.delegate = self
     }
 
     required dynamic init?(coder aDecoder: NSCoder) {
@@ -16,9 +24,21 @@ final class EditAddressHostingController: UIHostingController<EditAddressForm> {
     }
 }
 
+/// Intercepts to the dismiss drag gesture.
+///
+extension EditAddressHostingController: UIAdaptivePresentationControllerDelegate {
+    func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
+        // track dimiss gesture
+    }
+}
+
 /// Allows merchant to edit the customer provided address of an order.
 ///
 struct EditAddressForm: View {
+
+    /// Set this closure with UIKit dismiss code. Needed because we need access to the UIHostingController `dismiss` method.
+    ///
+    var dismiss: (() -> Void) = {}
 
     @ObservedObject private var viewModel: EditAddressFormViewModel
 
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
```diff
@@ -35,6 +55,17 @@ struct EditAddressForm: View {
     @State var showStateSelector = false
 
     var body: some View {
+        NavigationView {
+            // navigation liks require wrapping in a stack to work
```
Ecarrion:
```suggestion
            // navigation links require wrapping in a stack to work
```

Do you know why this happens? Is it due to a bug? or does it has a logical reason?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
```diff
@@ -35,6 +55,17 @@ struct EditAddressForm: View {
...
+            ZStack {
+                formContent
```
Ecarrion: Nit: Do you think it's worth it to take out the `NavigationLinks` to this level? It may be easier to read what is going on?

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -677,7 +677,7 @@ private extension OrderDetailsViewController {
     func editShippingAddressTapped() {
         let viewModel = EditAddressFormViewModel(siteID: viewModel.order.siteID, address: viewModel.order.shippingAddress)
         let editAddressViewController = EditAddressHostingController(viewModel: viewModel)
-        show(editAddressViewController, sender: self)
+        present(editAddressViewController, animated: true, completion: nil)
```
Ecarrion:
```suggestion
        let navigationController = WooNavigationController(rootViewController: editAddressViewController)
        present(navigationController, animated: true, completion: nil)
```

Now that I think about it, If we wrap the `EditAddressHostingController` we won't require the `ZStack` workaround & we will gain back the custom back button configuration.

What do you think? I think it's worth it.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
ealeksandrov: I don't have definite answer here, I remember using simple links directly inside NavigationView just fine.
Maybe it's because we use `EmptyView` in links.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
ealeksandrov: Agree, but after adding `WooNavigationController` we don't need this anymore.

ealeksandrov:
@Ecarrion thanks for review and nice idea about `WooNavigationController`! It simplifies the code and gives us correct back buttons 🎉
Ready for another look to confirm.

Ecarrion (approved):
Looks perfect!

PR #5023, opened 2021-09-20
Shipping Labels: add a new payment method

### LINKED ISSUE #4598: Add payment methods in Shipping Labels
We need to allow shop owners to add new payment methods.

Make sure to follow the release order: #4595

Feel free to split this issue into more subtasks.

Tasks:
- [x] Implement the button "Add another credit card". It should become "Add credit card" if there are no existing credit cards.
- [x] Only show Add Payment Button if the current user is the store owner.
- [x] We should open a web view that points to `https://wordpress.com/me/purchases/add-credit-card` since this functionality is not supported by API at the moment. Related PR on Android https://github.com/woocommerce/woocommerce-android/pull/4234
- [x] Tracks events
- [x] Display a notice `Payment method added` when a new payment method was added.

Before releasing this functionality, we should make sure those two PRs are merged: https://github.com/Automattic/wp-calypso/pull/54181 & https://github.com/Automattic/wp-calypso/pull/54189

Schermata 2021-07-13 alle 15 49 27

pmusolino:
Closes #4598

## Description
This PR adds support for adding new payment methods from the app in the Shipping Labels. The implementation relies on a WebView that opens the URL `https://wordpress.com/me/purchases/add-payment-method`, using a new SwiftUI component called `AuthenticatedWebView` that handles authenticating the user to their WordPress.com automatically.

## Testing
**Important testing notes:**
1) During the implementation of this functionality, I discovered that the 2-factor authentication is always shown. This happens only if you are under the Automattic Proxy. So, for testing this PR I suggest you disable the proxy on your Mac. More info here: pMz3w-bAI#comment-80199
2) For configuring your sandbox environment on the app, you should search for `Store Sandbox Cookie Secret` from the Secret Store, and put it under `AuthenticatedWebView`, substituting the string `[secret]`. Otherwise, you will not be able to test it with a fake credit card.
3) You can use one of the fake credit cards included in the post called `Testing the Store` under the Field Guide.

#### Case 1: Add another payment method
1. Make sure your WordPress.com account has some payment methods.
2. Open an order eligible for shipping label creation.
3. Click on Create shipping label.
4. Pass all steps until you reach the payment step.
5. Click on edit.
6. Click on Add another credit card.
7. Make sure the event `shipping_label_add_payment_method_tapped` was fired correctly in the console log.
8. Enter the card details and save.
9. Notice that the webview will be closed.
10. Notice that the "payment method" screen will get refreshed with the new payment method.
11. Make sure the event `shipping_label_payment_method_added` was fired correctly in the console log.
12. Select the new payment method.
13. Click on done.

#### Case 2: Add first payment method
1. Delete all credit cards from your WordPress.com account.
2. Open an order eligible for shipping label creation.
3. Click on Create shipping label.
4. Pass all steps until you reach the payment step.
5. Click on Continue.
6. Notice the placement of "Add a credit card" button.
7. Click on "Add a credit card".
8. Make sure the event `shipping_label_add_payment_method_tapped` was fired correctly in the console log.
9. Enter the card details and save.
10. Notice that the webview will be closed.
11. Notice that the "payment method" screen will get refreshed with the new payment method.
12. Make sure the event `shipping_label_payment_method_added` was fired correctly in the console log.
13. Click on done.

## Screenshots
| Add another credit card button            | WebView |
Simulator Screen Shot - iPhone 13 Pro - 2021-09-20 at 16 41 22 | Simulator Screen Shot - iPhone 13 Pro - 2021-09-20 at 16 43 13

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethods.swift
```diff
@@ -68,8 +69,35 @@ struct ShippingLabelPaymentMethods: View {
                         .padding(.horizontal, insets: geometry.safeAreaInsets)
                         .background(Color(.systemBackground))
                         .disabled(!viewModel.canEditNonpaymentSettings)
+
+                    Spacer()
+
+                    // Add credit card button
+                    if viewModel.canEditPaymentMethod && ServiceLocator.featureFlagService.isFeatureFlagEnabled(.shippingLabelsInternational) {
```
itsmeichigo: I think you'd want to check for feature flag `shippingLabelsAddPaymentMethods` instead?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethods.swift
```diff
@@ -4,6 +4,7 @@ import Yosemite
 struct ShippingLabelPaymentMethods: View {
     @ObservedObject private var viewModel: ShippingLabelPaymentMethodsViewModel
     @Environment(\.presentationMode) var presentation
+    @State private var showAddPaymentWebView: Bool = false
```
itsmeichigo: Nit: how about `showingAddPaymentWebView`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethods.swift
```diff
@@ -90,6 +118,27 @@ struct ShippingLabelPaymentMethods: View {
...
+                let notice = Notice(title: Localization.paymentMethodAddedNotice, feedbackType: .success)
+                ServiceLocator.noticePresenter.enqueue(notice: notice)
+            }
+            .navigationTitle(Localization.paymentMethodWebviewTitle)
+            .navigationBarTitleDisplayMode(.inline)
+            .navigationBarItems(trailing: Button(action: {
+                showAddPaymentWebView = false
+            }, label: {
+                Text(Localization.doneButtonAddPayment)
+            }))
```
itsmeichigo: I'm not sure if we should add the Done button here. I was confused when testing this because because I thought the Done button will behave like the Save button in the web view. Should we remove it?

The Figma design is showing only a full screen web view, in our case we're presenting the web view as a modal so I think it's fine that we don't have any navigation bar item here. User can dismiss the view by swiping down the model. WDYT?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethodsViewModel.swift
```diff
@@ -76,6 +80,23 @@ final class ShippingLabelPaymentMethodsViewModel: ObservableObject {
...
 
+    /// Syncs account settings specific to shipping labels, such as the last selected package and payment methods.
+    ///
+    func syncShippingLabelAccountSettings() {
+        let action = ShippingLabelAction.synchronizeShippingLabelAccountSettings(siteID: accountSettings.siteID) { [weak self] result in
+            guard let self = self else { return }
+
+            switch result {
+            case .success(let value):
+                self.accountSettings = value
```
itsmeichigo, line 91: There's an edge case that I found: Add a payment method, tap Done to get back to the purchase form, and then select the Payment Method row again, the new payment method does not show up.

I think we should also add a callback to save the account settings on the purchase form too - that way the new payment method will appear if we head out and back again to this screen.

itsmeichigo (changes requested):
I've built the branch but couldn't test some of the features, so I'm not sure if there's some part of the code missing. Here are what I found:

I could not reach the screen after removing all payment methods, and the implementation for this seems to be missing:
```Swift
    func configurePaymentMethod(cell: ShippingLabelFormStepTableViewCell, row: Row) {
        cell.configure(state: row.cellState,
                       icon: .creditCardImage,
                       title: Localization.paymentMethodCellTitle,
                       body: viewModel.getPaymentMethodBody(),
                       buttonTitle: Localization.continueButtonInCells) {
            // To be implemented as part of the "Add new payment method" flow
        }
    }
```

I've also found that delegate method `webView(_ webView:decidePolicyFor:decisionHandler:)` is not triggered after I tap the Save button on the web view. Should we observe the web view's URL instead? (I built the code on Xcode 12.5 to simulator iOS 14.5.)

I also have a few comments below 👇

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethods.swift
pmusolino: I had thought the same thing before introducing the button. In reality, this is needed because the swipe down doesn't work if you rotate the screen in landscape. This would mean blocking the user in the webview until they add a new payment method. I think it is always better to make the action clear and explicit, letting them understand that they can close the view at any time (among other things, it is a recurring pattern in our app).

pmusolino:
Thank you @itsmeichigo for your review! As I said in DM, I totally forgot to push some of the changes, that I lost after changing branches for a PR Review 🙊 my bad.

Can you take a second look at the PR, please?

> I've also found that delegate method webView(_ webView:decidePolicyFor:decisionHandler:) is not triggered after I tap the Save button on the web view. Should we observe the web view's URL instead? (I built the code on Xcode 12.5 to simulator iOS 14.5.)

This is strange, I tested it more than a hundred times and it always worked. Currently, I'm observing the view's URL and it works 🤔 I build the code on Xcode 13, but if this is the problem can be tricky (it means we should find a method that works on both versions of Xcode). Can you do a second check also on this? Maybe without the changes pushed there was something missing... not sure.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethods.swift
```diff
@@ -108,6 +170,19 @@ private extension ShippingLabelPaymentMethods {
...
+        static let addAnotherCreditCardButton = NSLocalizedString("Add another credit card",
+                                                                  comment: "Button title in the Shipping Label Payment Method" +
+                                                                    " screen if there is an existing payment method")
+        static let paymentMethodWebviewTitle = NSLocalizedString("Payment method",
+                                                            comment: "Title of the webview of adding a payment method in Shipping Labels")
+        static let doneButtonAddPayment = NSLocalizedString("Done",
+                                                            comment: "Done navigation button in Shipping Label add payment webview")
+        static let paymentMethodAddedNotice = NSLocalizedString("Payment method added",
+                                                                comment: "Notice that will be displayed after adding a new Shipping Label payment method")
+        static let pleaseAddPaymentMethodMessage = NSLocalizedString("Please, add a new payment method",
```
itsmeichigo, line 184: Nit: Maybe keep this message simple as "No payment method found"? Since the action button already says  `Add credit card` I think this sounds better.

itsmeichigo (approved):
Thanks for the update! The web view navigation delegate method works fine for me today, and I cannot find a proper explanation 😓. Good thing you haven't implemented the URL observing as I suggested! I hope this works consistently across different versions though.
✅  Code looks good
✅  Works as advertised

I think there's still room for improvements regarding UI of the screen:
- In portrait mode the empty view does not stay at the center of the screen. I think it's worth it to consider the optional scroll view introduced in #5009 or the ScrollableVStack mentioned in that discussion. I haven't experimented with those though, so I'm not sure if they're applicable in our case - just a suggestion.


- Should we show a loading view when syncing user account? That way we can avoid showing the empty view if it takes a while to load new data:
https://user-images.githubusercontent.com/5533851/134277722-1b87d210-fa22-4fe9-be72-64ee9c782a8c.mp4

- It looks like the web view does not look good in landscape mode when the keyboard is showing. I don't really have a good suggestion to get around this though:

Simulator Screen Shot - iPhone 12 Pro - 2021-09-22 at 09 46 49

I'm still approving this PR since all these issues are not too serious and can be handled later.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethods.swift
itsmeichigo: This makes sense to me! I'd prefer the button title to be "Cancel", but that may not be consistent with the rest of the app.

itsmeichigo:
A couple more things I forgot to mention regarding the Done button of the Payment Method screen: 

- Should we select the newly added payment method as the selected method automatically? Currently it seems that the old method is still selected, I think it'd be nice if the new one gets selected instead.
- Is it necessary to disable Done button when payment method is not updated? In the case when I have no payment method, after adding a card the card is selected automatically and the Done button is disabled. I was so confused and thought something was wrong 😅. My suggestion is that we only disable the Done button when there's no method to choose from.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Payment Methods/ShippingLabelPaymentMethods.swift
pmusolino: I'll do that in this PR before merging it 👍

pmusolino:
Thank you @itsmeichigo for your second review! 

Most of your suggestions to make an improvement make sense to me. Just to not block this PR and since they can be implemented later, I created a new issue here to track them: https://github.com/woocommerce/woocommerce-ios/issues/5037

> It looks like the web view does not look good in landscape mode when the keyboard is showing. I don't really have a good suggestion to get around this though.

It would be interesting, but I think that even if the view was native here, we could hardly find a way to improve the user experience. The height of the screen is occupied more by the keyboard, and there is no way to make it smaller. The same problem exists also in Safari, but also in our native views with forms in landscape.

PR #5026, opened 2021-09-21
Shipping Labels: Add Move button and action sheet for moving order items to different package

itsmeichigo:
Part of #4717

# Description
This PR aims to add the move button on each item row on Package Details screen. Tapping on the button should display an action sheet for moving the item elsewhere. The action sheet now has only one option: ship in original packaging. The action for the button will be implemented in the next PR.

# Changes
- Adds overlay to each `ItemToFulfillRow` for the extra Move button. This button drives the visibility of the action sheet.
- Adds new closure to `ShippingLabelSinglePackageViewModel` to let the packages form know when a request to move an item is sent.
- Updates the packages screen to configure and show the Move Item action sheet.

# Demo

https://user-images.githubusercontent.com/5533851/134162223-6609d749-497f-4f42-8bc2-33e0a207f8b4.mp4

# Testing
1. Make sure that your store has WCShip plugin installed.
2. Select an order eligible for creating shipping label and select Create Shipping Label.
3. Skip through Ship To and Ship From, and tap Continue on Package Details row.
4. Notice that there's a Move button on each of the order item row.
5. Tap on a Move button, notice that an action sheet is displayed with correct message, and 2 buttons: Ship in Original Packaging and Cancel.

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackagesFormViewModel.swift
```diff
@@ -13,6 +13,14 @@ final class ShippingLabelPackagesFormViewModel: ObservableObject {
         selectedPackages.count > 1
     }
 
+    /// Message displayed on the Move Item action sheet.
+    ///
+    @Published private(set) var moveItemActionSheetMessage: String = ""
```
pmusolino, line 18: Q: Why not using an optional string here? Also if it needs to be unwrapped, it's more clear in my opinion.

pmusolino (approved):
LGTM :shipit: concise and understandable code. Well done.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackagesFormViewModel.swift
itsmeichigo: Since the `Text` view for the title of the action sheet require a non-optional string, I think it's best that I use the view model as the single source of truth instead of using nil coalescing in the view. That makes sense right?

Issue #5032, opened 2021-09-22
Push notifications for all stores


jaclync:
Project thread p91TBi-5Oe-p2
iOS plan p91TBi-630-p2

## Goals

The main goal of this project is to notify merchants of sales and reviews for all of their stores on the go.

## Subtasks

- [x] Add a feature flag `pushNotificationsForAllStores` https://github.com/woocommerce/woocommerce-ios/pull/5033
- [x] Set `selected_blog_id` to an empty string when registering a device https://github.com/woocommerce/woocommerce-ios/pull/5033
- [x] Keep one `ReviewsCoordinator` per logged-in session for multiple stores, so that tapping on a new review notification from a different store opens the review details after switching store https://github.com/woocommerce/woocommerce-ios/pull/5044
- [ ] ~~(Pivoted to backend implementation) [Pending backend changes D67135] Create a `UNNotificationServiceExtension` target for modifying push notification content given payload from the server. Prerequisites & subtasks:~~
  - [x] Extract feature flag code to a separate Experiments framework https://github.com/woocommerce/woocommerce-ios/pull/5106
  - [ ] ~~Create a skeleton `UNNotificationServiceExtension` target https://github.com/woocommerce/woocommerce-ios/pull/5168~~
  - [ ] ~~Update notification content based on payload and feature flag in the notification service extension~~
- [x] [Pending backend changed D68195] Support new push notification payload format https://github.com/woocommerce/woocommerce-ios/pull/5299
- [x] Update badge number logic (not behind a feature flag): show 1 instead of the actual count for the selected store (ref p91TBi-66O-p2) https://github.com/woocommerce/woocommerce-ios/pull/5151
- [x] Change the copy of notice after switching store from "Switched to %1$@. You will only receive notifications from this store." to "Switched to %1$@." https://github.com/woocommerce/woocommerce-ios/pull/5137
- [x] Analytics: add event prop `is_from_selected_site` to `push_notification_received` and `push_notification_alert_pressed` events (ref p91TBi-6dd-p2) https://github.com/woocommerce/woocommerce-ios/pull/5222

Bug fixes:

- [x] Handle edge case: push notifications from a store that the app doesn’t know about yet (newly connected to Jetpack) https://github.com/woocommerce/woocommerce-ios/pull/5256

## Example screenshots

<img src="https://user-images.githubusercontent.com/1945542/134269139-878cf692-8876-4ac9-bdf4-74e90e4f4829.jpg" width="300" />

[...]

PR #5033, opened 2021-09-22
Enable push notifications for all stores behind a feature flag

jaclync:
Part of #5032 

## Why

The feature to support push notifications for all stores involves several steps as outlined in #5032, and I plan to make code changes behind a feature flag. This PR enables push notifications for all stores behind a new feature flag, the app shall start receiving remote push notifications from all stores when the feature is enabled.

## Changes

- Added a feature flag `pushNotificationsForAllStores` in https://github.com/woocommerce/woocommerce-ios/commit/dde8348f34ff76eb40da04ddcbe0cde2f8a6edf6
- Set `selected_blog_id` to an empty string when registering a device if the feature flag is enabled with unit tests

## Testing

Prerequisites: the logged-in user has more than one site connected to the app. If you are testing with your a8c WP.com account, you can resend previous new order/review notifications from the **MC Push Notifications Debug Console** (MC > Push Notifications > Debug Console > WP.com Notification, or with this path `/mobile/push-notifications/debug/?type=wpcom-notification`).

### Feature flag is enabled

- Launch the app logged in to a site, allow remote push notifications if you haven't already
- Place an order or resend a new order notification in the MC Debug Console **on a different site** --> a new order push notification should appear
- Leave a review or resend a new review notification in the MC Debug Console **on a different site** --> a new review push notification should appear
- Place an order or resend a new order notification in the MC Debug Console **on the selected site** --> a new order push notification should appear
- Leave a review or resend a new review notification in the MC Debug Console **on the selected site** --> a new review push notification should appear

### Feature flag is disabled

- In `DefaultFeatureFlagService`, return `false` for `pushNotificationsForAllStores` feature flag
- Launch the app logged in to a site, allow remote push notifications if you haven't already
- Place an order or resend a new order notification in the MC Debug Console **on a different site** --> a new order push notification should **not** appear
- Leave a review or resend a new review notification in the MC Debug Console **on a different site** --> a new review push notification should **not** appear
- Place an order or resend a new order notification in the MC Debug Console **on the selected site** --> a new order push notification should appear
- Leave a review or resend a new review notification in the MC Debug Console **on the selected site** --> a new review push notification should appear
- Tapping on the new order notification should show the order details as in the production
- Tapping on the new review notification should show the review details as in the production

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/DevicesRemote.swift
+++ b/Networking/Networking/Remote/DevicesRemote.swift
@@ -18,6 +18,7 @@ public class DevicesRemote: Remote {
                                applicationId: String,
                                applicationVersion: String,
                                defaultStoreID: Int64,
+                               pushNotificationsForAllStoresEnabled: Bool,
                                completion: @escaping (DotcomDevice?, Error?) -> Void) {
         var parameters = [
             ParameterKeys.applicationId: applicationId,
@@ -27,7 +28,7 @@ public class DevicesRemote: Remote {
             ParameterKeys.deviceModel: device.model,
             ParameterKeys.deviceName: device.name,
             ParameterKeys.deviceOSVersion: device.iOSVersion,
-            ParameterKeys.defaultStoreID: String(defaultStoreID)
+            ParameterKeys.defaultStoreID: pushNotificationsForAllStoresEnabled ? "": String(defaultStoreID)
         ]
 
         if let deviceUUID = device.identifierForVendor {
--- a/Networking/NetworkingTests/Extensions/MockNetwork+Path.swift
+++ b/Networking/NetworkingTests/Extensions/MockNetwork+Path.swift
@@ -13,6 +13,11 @@ extension MockNetwork {
         guard let urlComponents = URLComponents(string: url.absoluteString) else {
             return nil
         }
+
+        if let dotcomRequest = request as? DotcomRequest {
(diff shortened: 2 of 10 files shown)


pmusolino:
Thanks for opening this PR @jaclync! Before reviewing this PR, I need to request a new sandbox environment (the old one was closed), just give me the time to have it enabled 😃

pmusolino (approved):
Thank you @jaclync for this PR! I tested it in multiple cases with both feature flag enabled and disabled, and it works as expected.

Just one thing that I noticed, but it is probably expected, and we should support it in the future: I noticed for example that an order detail will be not opened if the app is completely closed. In that case, the switch to the new store will happen correctly, but the trigger will not fire. Instead, the order details are correctly opened, if the app is in the background (as in the test specifications).

jaclync:
Thanks for the review and testing a bunch of scenarios @pmusolino!

> Just one thing that I noticed, but it is probably expected, and we should support it in the future: I noticed for example that an order detail will be not opened if the app is completely closed. In that case, the switch to the new store will happen correctly, but the trigger will not fire. Instead, the order details are correctly opened, if the app is in the background (as in the test specifications).

I closed the app (so that the app is not in the app switcher) and was able to see the app switch to the new store and open order details 🤔 maybe there are some other conditions that were different in our testing, I'll keep an eye out for this case launching the app from closed state!

PR #5034, opened 2021-09-22
3722 Fix crash on media asset accessibilityLabel creation

### LINKED ISSUE #3722: NSInvalidArgumentException: -[WooCommerce.CancellableMedia filename]: unrecognized selector sent to instance 0x281898fc0
Sentry Issue: [WOOCOMMERCE-IOS-1C7X](https://sentry.io/organizations/a8c/issues/1854139684/?referrer=github_integration)

```
NSInvalidArgumentException: -[WooCommerce.CancellableMedia filename]: unrecognized selector sent to instance 0x281898fc0
  File "WPMediaPicker/Pod/Classes/WPMediaCollectionViewCell.m", line 213, in -[WPMediaCollectionViewCell configureAccessibility]
  File "WPMediaPicker/Pod/Classes/WPMediaCollectionViewCell.m", line 265, in -[WPMediaCollectionViewCell setAsset:]
  File "WPMediaPicker/Pod/Classes/WPMediaPickerViewController.m", line 880, in -[WPMediaPickerViewController configureCell:forIndexPath:]
  File "WPMediaPicker/Pod/Classes/WPMediaPickerViewController.m", line 871, in -[WPMediaPickerViewController collectionView:cellForItemAtIndexPath:]
  File "main.swift", line 7, in main
...
(23 additional frame(s) were not displayed)
```

Created by @shiki

---

It looks like we were supposed to implement `filename` for `CancellableMedia`. The `filename` property is expected by MediaPicker: https://github.com/wordpress-mobile/MediaPicker-iOS/blob/45726d0ba935f16846efca45d320f72ea4c74b2a/Pod/Classes/WPMediaCollectionViewCell.m#L212-L213

There does not seem to be a `protocol` we were supposed to implement though. ¯\\\_(ツ)\_/¯

---

As of Feb 22: 19 events, 5 users.

joshheald:
Fixes #3722 

## Description
A rare crash occurred when some users opened the media picker, due to unchecked access of the `filename` property on Media.

`filename` is an optional part of the protocol, so [this crash will also need to be fixed in MediaPicker-iOS](https://github.com/wordpress-mobile/MediaPicker-iOS/issues/371), as the asset class should be checked for conformance before the optional property is accessed. 

MediaPicker-iOS attempts to use the `filename` to create an accessibility label for a document, which will be of the form "Document: {filename}". This crash is very rare (4 incidents in the last month) likely because everywhere we use MediaPicker-iOS, we filter to show images only, which use the date in the accessibility label; for images, filename is not accessed.

This change adds the filename to the fields we request on the MediaRemote, and implements it on the WPMediaAsset protocol. Similarly, `fileExtension` is an optional conformance on the same protocol which we could trivially add, so that is included here too.

These changes will make this crash impossible within the WCiOS app, and slightly improve our users VoiceOver experience, but we'll also fix the library.

## Investigation
I've not been able to properly reproduce the crash in the app, since we [only request images](https://github.com/woocommerce/woocommerce-ios/blob/317ff6690660a6fe6ef163ba47052e725e5a3fbd/Networking/Networking/Remote/MediaRemote.swift#L25) for the media picker to display. We also set the [options of the media picker itself to only show images](https://github.com/woocommerce/woocommerce-ios/blob/317ff6690660a6fe6ef163ba47052e725e5a3fbd/WooCommerce/Classes/ViewRelated/Products/Media/WordPressMediaLibraryImagePickerViewController.swift#L15) for a belt-and-braces(suspenders!) approach, but that doesn't make any difference here.

However, by removing `ParameterKey.mimeType: "image"` from the media list request in `MediaRemote`, the crash can be trivially reproduced by opening the media picker on a site with any documents in its media library.

## Testing

Before (with images filter removed) | After
--------|--------
Crash | Fixed


koke (approved):
Now I'm puzzled as to why this can even happen with the image filter, but I guess that would get rid of the issue 🤷🏽‍♂️

joshheald:
> Now I'm puzzled as to why this can even happen with the image filter, but I guess that would get rid of the issue 🤷🏽‍♂️

Yeah, it's a bit weird. I suspect there's some strange mimetype which is falling into the API's `image` filter, but doesn't pass the on-device check [`UTType(mimetype:)`](https://github.com/woocommerce/woocommerce-ios/blob/317ff6690660a6fe6ef163ba47052e725e5a3fbd/Yosemite/Yosemite/Tools/Media/MediaType.swift#L21) which is what we use to provide MediaPicker the `assetType`. Then, when its [constructing the accesibilityLabel](https://github.com/wordpress-mobile/MediaPicker-iOS/blob/939737ff128587e001f23974ba5460e7758d83ce/Pod/Classes/WPMediaCollectionViewCell.m#L202), it's ending up in the `WPMediaTypeOther` case. But I've not been able to prove that.

PR #5042, opened 2021-09-23
Convert `MyStoreScreen` to be a `ScreenObject` subclass

mokagio:
My plan of action was to first create a UI tests foundation framework to share between UI tests and screenshot automation and then proceed with converting from `BaseScreen` to `ScreenObject`. But the other day @thehenrybyrd and I were pairing and decided to have some fun playing around with converting one screen and this is the result 😄 

To test it's enough to verify the UI test passed in CI. I'll be working on the tests more anyway soon, so any issue or suggestion will be addressed shortly.

For extra details on how `ScreenObject` subclasses work, see https://github.com/wordpress-mobile/WordPress-iOS/pull/17221

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. (N.A.)

### DIFF
--- a/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": null,
-          "revision": "df8ee4983e674b583b546bb0640ec83428be2465",
-          "version": "0.1.0"
+          "revision": "2d65beae47cdcaaf21703ce1b321f382fe0d0730",
+          "version": "0.2.0"
         }
       },
       {
--- a/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
+++ b/WooCommerce/WooCommerce.xcodeproj/project.pbxproj
@@ -523,6 +523,7 @@
 		31F92DE125E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31F92DE025E85F6A00DE04DF /* ConnectedReaderTableViewCell.swift */; };
 		31FE28C225E6D338003519F2 /* LearnMoreTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31FE28C125E6D338003519F2 /* LearnMoreTableViewCell.swift */; };
 		31FE28C825E6D384003519F2 /* LearnMoreTableViewCell.xib in Resources */ = {isa = PBXBuildFile; fileRef = 31FE28C725E6D384003519F2 /* LearnMoreTableViewCell.xib */; };
+		3F0CF2FE270420DD00EF3D71 /* ScreenObject+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3FF314D626FC55DE0012E68E /* ScreenObject+Extension.swift */; };
 		3F1CA81D26C3542600228BF2 /* XCUITestHelpers in Frameworks */ = {isa = PBXBuildFile; productRef = 3F1CA81C26C3542600228BF2 /* XCUITestHelpers */; };
 		3F1CA81F26C3543C00228BF2 /* XCUITestHelpers in Frameworks */ = {isa = PBXBuildFile; productRef = 3F1CA81E26C3543C00228BF2 /* XCUITestHelpers */; };
 		3FF2247326706AA3008FFA87 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3FF2247226706AA3008FFA87 /* ScreenObject */; };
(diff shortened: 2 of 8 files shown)


WooCommerce.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -24,8 +24,8 @@
         "repositoryURL": "https://github.com/Automattic/ScreenObject",
         "state": {
           "branch": null,
-          "revision": "df8ee4983e674b583b546bb0640ec83428be2465",
-          "version": "0.1.0"
+          "revision": "2d65beae47cdcaaf21703ce1b321f382fe0d0730",
+          "version": "0.2.0"
```
mokagio, line 28: This new ScreenObject version includes the init with support for multiple element getters ✨

WooCommerce/WooCommerce.xcodeproj/project.pbxproj
```diff
@@ -526,6 +526,8 @@
 		3F1CA81F26C3543C00228BF2 /* XCUITestHelpers in Frameworks */ = {isa = PBXBuildFile; productRef = 3F1CA81E26C3543C00228BF2 /* XCUITestHelpers */; };
 		3FF2247326706AA3008FFA87 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3FF2247226706AA3008FFA87 /* ScreenObject */; };
 		3FF2248A267073AE008FFA87 /* ScreenObject in Frameworks */ = {isa = PBXBuildFile; productRef = 3FF22489267073AE008FFA87 /* ScreenObject */; };
+		3FF314D726FC55DE0012E68E /* ScreenObject+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3FF314D626FC55DE0012E68E /* ScreenObject+Extension.swift */; };
+		3FF314D826FC577B0012E68E /* ScreenObject+Extension.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3FF314D626FC55DE0012E68E /* ScreenObject+Extension.swift */; };
```
mokagio: Noticed the duplicated entry. This is because this file needs to be available by both the UI tests and the screenshots automation targets. Once we'll have the UI tests foundation framework, this will no longer be necessary.

WooCommerce/WooCommerceUITests/Screens/MyStore/MyStoreScreen.swift
```diff
@@ -1,39 +1,39 @@
-import Foundation
```
mokagio: XCTest already imports Foundation, so we don't need to add import it explicitly

WooCommerce/WooCommerceUITests/Screens/MyStore/MyStoreScreen.swift
```diff
@@ -1,39 +1,39 @@
...
+import ScreenObject
 import XCTest
 
-final class MyStoreScreen: BaseScreen {
-
-    struct ElementStringIDs {
-        static let topBannerCloseButton = "top-banner-view-dismiss-button"
-        static let settingsButton = "dashboard-settings-button"
-    }
```
mokagio, line 9: `ScreenObject` allows us to centralize the logic that queries the `XCUIApplication` instance for the elements to interact with, making this type holding the accessibility identifiers unnecessary.

WooCommerce/WooCommerceUITests/Screens/MyStore/MyStoreScreen.swift
```diff
@@ -1,39 +1,39 @@
...
 
     @discardableResult
     func dismissTopBannerIfNeeded() -> MyStoreScreen {
+        guard topBannerCloseButton.waitForExistence(timeout: 3) else { return self }
 
-        if topBannerCloseButton.waitForExistence(timeout: 3) {
-            topBannerCloseButton.tap()
-        }
-
+        topBannerCloseButton.tap()
```
mokagio, line 33: This method is the same, I simply replaced the `if` with a `guard`.

Now that I'm writing this comment, I realize the change is a bit arbitrary. I guess it just felt more appropriate to use a `guard` in combination with a "wait for existence" method than an `if`. Happy to revert if this doesn't make sense to you.

WooCommerce/WooCommerceUITests/Screens/MyStore/MyStoreScreen.swift
```diff
@@ -1,39 +1,39 @@
...
+
+    private let topBannerCloseButtonGetter: (XCUIApplication) -> XCUIElement = {
+        $0.buttons["top-banner-view-dismiss-button"]
     }
 
-    init() {
-        settingsButton = XCUIApplication().buttons[ElementStringIDs.settingsButton]
-        topBannerCloseButton = XCUIApplication().buttons[ElementStringIDs.topBannerCloseButton]
+    private var settingsButton: XCUIElement { settingsButtonGetter(app) }
+    private var topBannerCloseButton: XCUIElement { topBannerCloseButtonGetter(app) }
```
mokagio: We use the getter instance properties to [_compute_ ](https://docs.swift.org/swift-book/LanguageGuide/Properties.html#ID259) the element at runtime so it's always up to date.

WooCommerce/WooCommerceUITests/Screens/MyStore/MyStoreScreen.swift
```diff
@@ -1,39 +1,39 @@
...
 
-        super.init(element: settingsButton)
+    static var isVisible: Bool {
+        guard let screen = try? MyStoreScreen() else { return false }
+        return screen.settingsButton.isHittable
+    }
```
mokagio, line 19: This is the same functionality as the implementation above. Initializing a `ScreenObject` already ensures it `.exists`, throwing an error otherwise.

With `try?` we can capture a thrown error and "pretend" that instead of receiving a value from the `init` method (`MyStoreScreen()`) we got `.none`. The `guard let screen...` will fail if the value is `.none` and run the `else` branch.

WooCommerce/WooCommerceUITests/Screens/MyStore/MyStoreScreen.swift
pachlava: To me, this makes sense, since this better depicts the logic, which is: leave as soon as we know we don't need to do anything in this method, or tap in the opposite case.

pachlava (approved):
Giving a :shipit: here. I executed the tests on CI (approval was needed) and also locally, all of them passed. Thank you for the extra details!

mokagio:
Thank you as always @pachlava 🙇

Issue #3823, opened 2021-03-12
Replaces "Mocks" For "Fakes"


Ecarrion:
Ref: #3813 

We have a lot of "Mock" classes for our networking models that are not ideal for tests.
We should replace them and use the `GeneratedFakeable` and `GeneratedCopiable` protocols instead.

Tasks:
- [ ] Unite both protocols into a single one for convenience
- [ ] Delete all Mocks and replace them for `Type.fake().copy(...)`

Mocks:
**In WooCommerce**
- [ ] MockReviews
- [ ] MockProductVariation
- [ ] MockOrders
- [ ] MockOrderItem
- [ ] MockRefunds
- [ ] MockShippingLabel
- [ ] MockShippingLabelAddress
- [ ] MockAggregateOrderItem

**In Yosemite**
- [ ] MockNote
- [ ] MockProductReview
- [ ] MockProductVariation
- [ ] MockShippingLabel
- [ ] MockShippingLabelAddress

jaclync:
Hey @Ecarrion, nice work on the Fakes! Do you think we can include classes/structs from Yosemite as well? I was hoping to replace a new mock class for `ProductReviewFromNoteParcel` in [this PR](https://github.com/woocommerce/woocommerce-ios/pull/5044#discussion_r715167994), but the fakes are only created for Networking classes so far. Maybe we were starting with Networking first?

Ecarrion:
@jaclync sure, [here is the PR](https://github.com/woocommerce/woocommerce-ios/pull/5056) to add `Yosemite` support.

Ecarrion:
Closing this as we are not actively following it

PR #5044, opened 2021-09-23
Show review details after tapping on a new review push notification from another store

jaclync:
Part of #5032 
Based on https://github.com/woocommerce/woocommerce-ios/pull/5033 (the code is independent but the base allows testing for multi-store push notifications)

## Why

`ReviewsCoordinator` nicely handles new review push notifications - it switches store if the notification is not from the currently selected store, navigates to Reviews tab, and then shows review details. After the [site ID DI work](https://github.com/woocommerce/woocommerce-ios/pull/2850/files#diff-447f41e2635ca816fc716d17130a9f977c281b1ea9d73948b0bf0d1369dc070d), `ReviewsCoordinator`'s lifetime becomes per site ID. This is fine when push notifications can only come from the default store, but the code to show review details (shown below) isn't executed anymore when we support multi-store push notifications. `self` becomes `nil` after switching store because a new `ReviewsCoordinator` is instantiated.

https://github.com/woocommerce/woocommerce-ios/blob/0e14c9a99daba7e460554a51b1028e74bcdea07f/WooCommerce/Classes/ViewRelated/Reviews/ReviewsCoordinator.swift#L80-L87

This PR updates the lifetime of `ReviewsCoordinator` from per site ID to per logged-in session where users might connect to multiple sites.

## Changes

- Removed `siteID` and Reviews tab navigation controller's setup from `ReviewsCoordinator`. The navigation controller setup is now in `MainTabBarController`.
  - I was hoping to configure `ReviewsCoordinator` when the user logs in, but couldn't find an ideal place to call it. Each tab is [updated for each site ID](https://github.com/woocommerce/woocommerce-ios/blob/0e14c9a99daba7e460554a51b1028e74bcdea07f/WooCommerce/Classes/ViewRelated/MainTabBarController.swift#L358-L367) when the site ID observable changes, and we need to make sure `ReviewsCoordinator` is configured before the first site ID is set. Since the site ID observable isn't obviously sending updates **after** the logged-in event, I decided to configure `ReviewsCoordinator` when it is `nil` (the first site ID for a logged-in session) when updating the tabs per site ID. Feel free to suggest other ways to do this!
- Unit tests:
  - Created `MockProductReviewFromNoteParcel` for mocking a new review notification parcel, and replaced existing usage in `ReviewsCoordinatorTests`
  - In `MainTabBarControllerTests`, added a new test case for handling new review push notifications from another store. I couldn't find an easy way to DI `PushNotesManager` and `SwitchStoreUseCaseProtocol` from `MainTabBarController` (instantiated from Storyboard) to `ReviewsCoordinator`, so I had to pass mocks for these dependencies in `ServiceLocator`. Because I had to mock `PushNotesManager` in `ServiceLocator`, I had to remove `ServiceLocatorTests.test_pushNotesManager_defaults_to_PushNotificationsManager` 😨 

## Testing

Prerequisites: the logged-in user has more than one site connected to the app.
If you are testing with your a8c WP.com account, you can resend previous new order/review notifications from the **MC Push Notifications Debug Console** (MC > Push Notifications > Debug Console > WP.com Notification, or with this path `/mobile/push-notifications/debug/?type=wpcom-notification`).

- Launch the app logged in to a site, allow remote push notifications if you haven't already
- Leave a review or resend a new review notification in the MC Debug Console **on a different site** --> a new review push notification should appear
- Tap on the new review push notification --> the app should switch to the target site, navigate to the reviews tab, then show the review details (before this PR, the app only switches to the target site because `ReviewsCoordinator` is deallocated after the store switch)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
+++ b/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
@@ -382,9 +382,15 @@ private extension MainTabBarController {
         let productsViewController = createProductsViewController(siteID: siteID)
         productsNavigationController.viewControllers = [productsViewController]
 
-        let reviewsTabCoordinator = createReviewsTabCoordinator(siteID: siteID)
-        self.reviewsTabCoordinator = reviewsTabCoordinator
-        reviewsTabCoordinator.start()
+        // Configure reivews tab coordinator once per logged in session potentially with multiple sites.
+        if reviewsTabCoordinator == nil {
+            let reviewsTabCoordinator = createReviewsTabCoordinator()
+            self.reviewsTabCoordinator = reviewsTabCoordinator
+            reviewsTabCoordinator.start()
+        }
+
+        let reviewsViewController = createReviewsViewController(siteID: siteID)
+        reviewsTabCoordinator?.navigationController.viewControllers = [reviewsViewController]
 
         // Set dashboard to be the default tab.
         selectedIndex = WooTab.myStore.visibleIndex()
@@ -402,9 +408,12 @@ private extension MainTabBarController {
         ProductsViewController(siteID: siteID)
     }
 
-    func createReviewsTabCoordinator(siteID: Int64) -> Coordinator {
-        ReviewsCoordinator(siteID: siteID,
-                           navigationController: reviewsNavigationController,
+    func createReviewsViewController(siteID: Int64) -> UIViewController {
+        ReviewsViewController(siteID: siteID)
+    }
+
+    func createReviewsTabCoordinator() -> Coordinator {
+        ReviewsCoordinator(navigationController: reviewsNavigationController,
                            willPresentReviewDetailsFromPushNotification: { [weak self] in
                             self?.navigateTo(.reviews)
         })
--- a/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
+++ b/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
@@ -87,7 +87,7 @@ final class MainTabBarController: UITabBarController {
     private let ordersNavigationController = WooTabNavigationController()
     private let productsNavigationController = WooTabNavigationController()
     private let reviewsNavigationController = WooTabNavigationController()
-    private var reviewsTabCoordinator: Coordinator?
+    private var reviewsTabCoordinator: ReviewsCoordinator?
 
     private var cancellableSiteID: ObservationToken?
 
@@ -389,8 +389,7 @@ private extension MainTabBarController {
             reviewsTabCoordinator.start()
         }
 
-        let reviewsViewController = createReviewsViewController(siteID: siteID)
-        reviewsTabCoordinator?.navigationController.viewControllers = [reviewsViewController]
+        reviewsTabCoordinator?.activate(siteID: siteID)
 
         // Set dashboard to be the default tab.
         selectedIndex = WooTab.myStore.visibleIndex()
@@ -408,11 +407,7 @@ private extension MainTabBarController {
         ProductsViewController(siteID: siteID)
     }
 
-    func createReviewsViewController(siteID: Int64) -> UIViewController {
-        ReviewsViewController(siteID: siteID)
-    }
-
-    func createReviewsTabCoordinator() -> Coordinator {
+    func createReviewsTabCoordinator() -> ReviewsCoordinator {
         ReviewsCoordinator(navigationController: reviewsNavigationController,
                            willPresentReviewDetailsFromPushNotification: { [weak self] in
                             self?.navigateTo(.reviews)
--- a/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
+++ b/WooCommerce/Classes/ViewRelated/MainTabBarController.swift
@@ -382,7 +382,7 @@ private extension MainTabBarController {
         let productsViewController = createProductsViewController(siteID: siteID)
         productsNavigationController.viewControllers = [productsViewController]
 
-        // Configure reivews tab coordinator once per logged in session potentially with multiple sites.
+        // Configure reviews tab coordinator once per logged in session potentially with multiple sites.
         if reviewsTabCoordinator == nil {
             let reviewsTabCoordinator = createReviewsTabCoordinator()
             self.reviewsTabCoordinator = reviewsTabCoordinator
--- a/WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
+++ b/WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
@@ -1,6 +1,5 @@
-import Foundation
 import XCTest
-@testable import Yosemite
+import Yosemite
 
 
 final class MockProductReviewFromNoteParcel {
@@ -34,7 +33,7 @@ final class MockProductReviewFromNoteParcel {
                         body: Data(),
                         meta: metaAsData)
         let product = Product.fake()
-        let review = MockReviews().anonymousReview()
+        let review = ProductReview.fake()
 
         return ProductReviewFromNoteParcel(note: note, review: review, product: product)
     }
--- a/WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
+++ b/WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
@@ -1,0 +1,41 @@
+import Foundation
+import XCTest
+@testable import Yosemite
+
+
+final class MockProductReviewFromNoteParcel {
+    func parcel(metaSiteID: Int64 = 0) -> ProductReviewFromNoteParcel {
+        let metaAsData: Data = {
+            var ids = [String: Int64]()
+            ids[MetaContainer.Keys.site.rawValue] = metaSiteID
+
+            var metaAsJSON = [String: [String: Int64]]()
+            metaAsJSON[MetaContainer.Containers.ids.rawValue] = ids
+            do {
+                return try JSONEncoder().encode(metaAsJSON)
+            } catch {
+                XCTFail("Expected to convert MetaContainer JSON to Data. \(error)")
+                return Data()
+            }
+        }()
+
+        let note = Note(noteID: 1,
+                        hash: 0,
+                        read: false,
+                        icon: nil,
+                        noticon: nil,
+                        timestamp: "",
+                        type: "",
+                        subtype: nil,
+                        url: nil,
+                        title: nil,
+                        subject: Data(),
+                        header: Data(),
+                        body: Data(),
+                        meta: metaAsData)
+        let product = Product.fake()
+        let review = MockReviews().anonymousReview()
+
+        return ProductReviewFromNoteParcel(note: note, review: review, product: product)
+    }
+}
--- a/WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
@@ -34,13 +34,13 @@ final class MainTabBarControllerTests: XCTestCase {
 
         // Assert
         XCTAssertEqual(tabBarController.viewControllers?.count, 4)
-        assertThat((tabBarController.viewControllers?[WooTab.myStore.visibleIndex()] as? UINavigationController)?.topViewController,
+        assertThat(tabBarController.tabNavigationController(tab: .myStore).topViewController,
                    isAnInstanceOf: DashboardViewController.self)
-        assertThat((tabBarController.viewControllers?[WooTab.orders.visibleIndex()] as? UINavigationController)?.topViewController,
+        assertThat(tabBarController.tabNavigationController(tab: .orders).topViewController,
                    isAnInstanceOf: OrdersRootViewController.self)
-        assertThat((tabBarController.viewControllers?[WooTab.products.visibleIndex()] as? UINavigationController)?.topViewController,
+        assertThat(tabBarController.tabNavigationController(tab: .products).topViewController,
                    isAnInstanceOf: ProductsViewController.self)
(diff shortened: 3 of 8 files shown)


WooCommerce/Classes/ViewRelated/MainTabBarController.swift
```diff
@@ -382,9 +382,15 @@ private extension MainTabBarController {
         let productsViewController = createProductsViewController(siteID: siteID)
         productsNavigationController.viewControllers = [productsViewController]
 
-        let reviewsTabCoordinator = createReviewsTabCoordinator(siteID: siteID)
-        self.reviewsTabCoordinator = reviewsTabCoordinator
-        reviewsTabCoordinator.start()
+        // Configure reviews tab coordinator once per logged in session potentially with multiple sites.
```
shiki, line 385: I pushed a commit to fix a typo. I'm experimenting with fixing small things like typos to help save some time for the author. But now I realized you might be surprised why you can't `git push` locally. Sorry! Please `git pull`.

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
```diff
@@ -382,9 +382,15 @@ private extension MainTabBarController {
...
+        if reviewsTabCoordinator == nil {
+            let reviewsTabCoordinator = createReviewsTabCoordinator()
+            self.reviewsTabCoordinator = reviewsTabCoordinator
+            reviewsTabCoordinator.start()
+        }
+
+        let reviewsViewController = createReviewsViewController(siteID: siteID)
+        reviewsTabCoordinator?.navigationController.viewControllers = [reviewsViewController]
```
shiki: > I was hoping to configure ReviewsCoordinator when the user logs in, but couldn't find an ideal place to call it. Each tab is updated for each site ID when the site ID observable changes, and we need to make sure ReviewsCoordinator is configured before the first site ID is set. Since the site ID observable isn't obviously sending updates after the logged-in event, I decided to configure ReviewsCoordinator when it is nil (the first site ID for a logged-in session) when updating the tabs per site ID. Feel free to suggest other ways to do this!

If I remember correctly, what we ultimately wanted a long time ago was for `MainTabBarController` itself to be recreated instead of the child view controllers. But that's an even bigger change. 🤪 And I think there were some UX problems. 

I think what you have here is fine. 

Another way to get around this is to move this switching to `ReviewsTabCoordinator` instead. The `reviewsTabCoordinator?.navigationController.viewControllers = [reviewsViewController]` line feels like the coordinator's responsibility. 

```swift
class ReviewsCoordinator {
    // Replaces start(). Can be called multiple times by MainTabBarController.
    func acrtivate(with siteID:) {
        // recreate
        navigationController.viewControllers = [ReviewsViewController(siteID: siteID)]

        if observationToken == nil {
            observationToken = pushNotificationsManager.inactiveNotifications.subscribe { [weak self] in
                // TODO: this probably needs to be changed to do nothing if there's no `siteID`. Maybe!
                // Or that can be done within `handleInactiveNotification`.
                self?.handleInactiveNotification($0)
            }
        }


    }
}
```

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
```diff
@@ -95,10 +95,72 @@ final class MainTabBarControllerTests: XCTestCase {
...
+            if case let .resetStoredOrders(completion) = action {
+                completion()
+            }
+        }
+        storesManager.whenReceivingAction(ofType: ProductReviewAction.self) { action in
+            if case let .resetStoredProductReviews(completion) = action {
+                completion()
+            }
+        }
+
```
shiki, line 129: I'd also suggest a precondition check. But maybe it's already implied. 🤷 

```suggestion

        let reviewsTabNavigationController = tabBarController.tabNavigationController(tab: .reviews)
        assertThat(reviewsTabNavigationController.topViewController, isAnInstanceOf: ReviewsViewController.self)
```

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
```diff
@@ -95,10 +95,72 @@ final class MainTabBarControllerTests: XCTestCase {
...
     var tabRootViewControllers: [UIViewController] {
         viewControllers?.compactMap { $0 as? UINavigationController }.compactMap { $0.viewControllers.first } ?? []
     }
+
+    func tabNavigationController(tab: WooTab) -> UINavigationController {
+        guard let navigationController = viewControllers?.compactMap({ $0 as? UINavigationController })[tab.visibleIndex()] else {
+            fatalError("Unexpected access to navigation controller at tab: \(tab)")
+        }
+        return navigationController
+    }
```
shiki: I think we should avoid `fatalError` in unit tests since they will cause crashes and halt the running tests. It does take a bit more effort to code around them though. 

```suggestion
    func tabNavigationController(tab: WooTab) -> UINavigationController? {
        viewControllers?.compactMap({ $0 as? UINavigationController })[tab.visibleIndex()]
    }
```

We'll have to unwrap all the time. 🙈 

```swift
        let reviewsTabNavigationController = try XCTUnwrap(tabBarController.tabNavigationController(tab: .reviews))
```

Maybe there's a better way somewhere. Anyway, up to you!

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import XCTest
+@testable import Yosemite
+
+
+final class MockProductReviewFromNoteParcel {
```
shiki: We probably need to rename some of these `Mock*` classes at some point. They're misleading. It took me a while to understand why we call them mocks. In reality, some are just builders of objects. 🤷

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
```diff
@@ -0,0 +1,41 @@
...
+                        type: "",
+                        subtype: nil,
+                        url: nil,
+                        title: nil,
+                        subject: Data(),
+                        header: Data(),
+                        body: Data(),
+                        meta: metaAsData)
+        let product = Product.fake()
+        let review = MockReviews().anonymousReview()
```
shiki: We can change this to use `fake` too:

```suggestion
        let review = ProductReview.fake()
```

The test still passed when I replaced it. 🤷

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import XCTest
+@testable import Yosemite
```
shiki: `@testable` is not needed and probably misleading. 🙂 

```suggestion
import XCTest
import Yosemite
```

shiki (approved):
@jaclync Thank you so much for the background info. I have forgotten some of the things that we changed a long time ago! 🤣 

I have not been able to test this yet since my certificate _conveniently_ expired yesterday. I'm waiting to get that fixed. (p1632427100338100-slack-CC7L49W13) But I can test this later if you like. 

I reviewed the code and I think that this will work just fine. Feel free to ignore all my comments and merge so you'll be unblocked. And I can test again when everything's ready. 

> In MainTabBarControllerTests, added a new test case for handling new review push notifications from another store. I couldn't find an easy way to DI PushNotesManager and SwitchStoreUseCaseProtocol from MainTabBarController (instantiated from Storyboard) to ReviewsCoordinator, so I had to pass mocks for these dependencies in ServiceLocator.

I think there's now a way to instantitate storyboard classes and inject dependencies. But I haven't used it before and no idea if it'll solve the problem. 🤷

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
jaclync: Thanks for the fix commit! Somehow my grammar check and other preferences were gone from deleting a few older Xcode versions 😬

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
jaclync: It looks like the future direction for these `Mock*` is to replace them with `Fakeable` https://github.com/woocommerce/woocommerce-ios/issues/3823 - however, only Networking models are supported right now and `ProductReviewFromNoteParcel` is in Yosemite. I asked about including Yosemite models in the issue, maybe we can add support for Yosemite in the future 😃 

Though for this case, we probably want to keep this class that creates `metaAsData` for the meta site ID. Maybe we can rename this class to like `ProductReviewFromNoteParcelFactory` to differentiate it from other `Mock*` classes?

jaclync: Btw, I tried generating Copiable for `Note` to replace the `Note` generation in this class. But because `Note` has a few properties that are not in the designated initializer like `metaAsData` and `timestampAsDate`, the generated `copy` function isn't calling the designated initializer. We can probably update the designated initializer to include all properties and mark the [current initializer](https://github.com/woocommerce/woocommerce-ios/blob/fb25492561c0d5bc16ab148b314969583880b886/Networking/Networking/Model/Note.swift#L100-L113) as a convenience initializer?

WooCommerce/Classes/ViewRelated/MainTabBarController.swift
jaclync: Having the reviews tab UI setup in `ReviewsCoordinator` makes more sense to me! Updated in https://github.com/woocommerce/woocommerce-ios/pull/5044/commits/6b2eb3e5e56579dcb2f42ec174dba99b63153d45

> If I remember correctly, what we ultimately wanted a long time ago was for MainTabBarController itself to be recreated instead of the child view controllers. But that's an even bigger change. 🤪 And I think there were some UX problems.

Yea, I remember having issues when the window doesn't start with `MainTabBarController` 😅

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
jaclync: Updated in: https://github.com/woocommerce/woocommerce-ios/pull/5044/commits/15d1131bbe5e1c8d293066161c905502a76d48e5

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
jaclync: Nice! Updated in: https://github.com/woocommerce/woocommerce-ios/pull/5044/commits/15d1131bbe5e1c8d293066161c905502a76d48e5

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
jaclync: Yea, it's not ideal for the tests to crash because of this scenario. I ended up just using the optional value, as in: https://github.com/woocommerce/woocommerce-ios/pull/5044/commits/8b9dc633fb1eddabfabb4f251fd3de9aa69849e5

WooCommerce/WooCommerceTests/ViewRelated/MainTabBarControllerTests.swift
jaclync: Makes sense! Added in: https://github.com/woocommerce/woocommerce-ios/pull/5044/commits/8b9dc633fb1eddabfabb4f251fd3de9aa69849e5

jaclync:
Thanks for your thoughtful review @shiki! 🙏 I just responded to your comments and updated the PR for another look. I also tested tapping on a review notification from another store after the code changes, feel free to skip the testing part or test it later since the feature is still at an early stage!

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
shiki: > Maybe we can rename this class to like ProductReviewFromNoteParcelFactory to differentiate it from other Mock* classes?

`*Factory` sounds fine by me! 

> We can probably update the designated initializer to include all properties and mark the current initializer as a convenience initializer?

Yeah, it's certainly worth a try. But up to you if you feel that's valuable to do right now. `Note` is probably not used a lot to warrant the effort. 🤷

shiki (approved):
Awesome! Thanks for double-checking. I pushed 2 small commits to fix the failing `ReviewCoordinatorTests` so it all passes now. 🙂 I still can't run on a device but I can definitely help test throughout the development of this project.

WooCommerce/WooCommerceTests/Mocks/MockProductReviewFromNoteParcel.swift
jaclync: > Yeah, it's certainly worth a try. But up to you if you feel that's valuable to do right now. Note is probably not used a lot to warrant the effort. 🤷

Going to do it in a separate issue/PR! https://github.com/woocommerce/woocommerce-ios/issues/5060

jaclync: > *Factory sounds fine by me!

Renamed in https://github.com/woocommerce/woocommerce-ios/pull/5044/commits/5626cd846c2647073348cb56cb65b5b05e13a93f

jaclync:
> I pushed 2 small commits to fix the failing ReviewCoordinatorTests so it all passes now. 🙂 I still can't run on a device but I can definitely help test throughout the development of this project.

Ohh thank you @shiki for fixing the tests! 🙏 I totally missed them since I was planning to make the `*Factory` change. And sounds good about the testing part! Merging soon 😄

PR #5046, opened 2021-09-23
4851 Wrap multiple lines on package details row

### LINKED ISSUE #4851: Shipping Labels: wrap multiple lines on Package Details screen
> When there are multiple lines, can we wrap up the text in the row as shown above?

image

joshheald:
Fixes #4851

## Description
Some package descriptions are relatively long, and wrap in the Package Selected row, but without any padding. This means they look a little squeezed.

This change allows the row to expand based on the height of the text, adding padding around the text as well.

## Testing
1. View an unfulfilled order on a site, and tap "Create Shipping Label"
2. Continue through the addresses
3. In Package Details, select a package with a long name, which you may need to add. Then tap Done.
4. Observe that the package selected row expands vertically to accommodate the name, along with 12px padding at the top and bottom of the cell.

After fix

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/TitleAndValueRow.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/TitleAndValueRow.swift
@@ -22,6 +22,7 @@ struct TitleAndValueRow: View {
                 Text(value)
                     .font(.body)
                     .foregroundColor(Color(.textSubtle))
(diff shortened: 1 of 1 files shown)


joshheald:
@designsimply I took a look at the truncation thing mentioned in the long label name, and I don't think we should do it, because it makes even normal labels difficult to use at large text sizes.
Truncated
..

pmusolino (approved):
LGTM :shipit:

PR #5047, opened 2021-09-23
Order Editing: Add State Selection Support

### LINKED ISSUE #4780: Order Editing: Fill Country & State UI

Ecarrion:
closes #4780

# Why

After the last bits to make the country selection work were implemented in #4981.
Now it is time to add support for state selection.

# How

- A lot of small changes are copied from `CountrySelectorCommand` & `CountrySelectorViewModel`

- The main change on the `EditAddressFormViewModel` is to make sure that a new state updates the state field and to make sure to `nil` the selected state if a different country is selected. PS I'm retaining the state name in fields to mimic core behavior.

# Demo

https://user-images.githubusercontent.com/562080/134587437-0b20baed-b3d3-420a-be00-4f53b941dcb1.mov

# Testing Steps

- Navigate to an order
- Tap on the "shipping address" edit icon
- See that the country & state field is property populated
- See that you can update the state.
- See that you change the country but the state value is persisted (to mimic core)

# Pending 

- Handle situations where there are no states. [Card 1](https://github.com/woocommerce/woocommerce-ios/projects/39#card-69408423) [Card 2](https://github.com/woocommerce/woocommerce-ios/projects/39#card-69336455)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
@@ -162,22 +162,24 @@ extension EditAddressFormViewModel {
             address2 = address.address2 ?? ""
             city = address.city
             postcode = address.postcode
-            state = address.state
         }
 
         mutating func update(with selectedCountry: Yosemite.Country?) {
             country = selectedCountry?.name ?? country
         }
 
+        mutating func update(with selectedState: Yosemite.StateOfACountry?) {
+            state = selectedState?.name ?? state
+        }
 
-        func toAddress(selectedCountry: Yosemite.Country?) -> Yosemite.Address {
+        func toAddress(selectedCountry: Yosemite.Country?, selectedState: Yosemite.StateOfACountry?) -> Yosemite.Address {
             Address(firstName: firstName,
                     lastName: lastName,
                     company: company.isEmpty ? nil : company,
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -145,57 +162,74 @@ extension EditAddressFormViewModel {
...
                 }
-                return .done(enabled: originalAddress != fields.toAddress(selectedCountry: selectedCountry))
+                return .done(enabled: originalAddress != fields.toAddress(country: selectedCountry, state: selectedState))
             }
             .assign(to: &$navigationTrailingItem)
     }
 
-    /// Update published fields when the selected country is updated.
+    /// Update published fields when the selected country and state is updated.
+    /// If the current selected state does not exists within the selected country, then the sate is nilled.
```
ealeksandrov:
```suggestion
    /// If the current selected state does not exists within the selected country, then the state is nilled.
```

ealeksandrov (approved):
LGTM :shipit:
Nice work handling country-state dependency! (But will still need textfield fallback, as you noted)

PR #5049, opened 2021-09-24
Polish GenerateCredentials aggregate settings

mokagio:
I've been experiencing the following issue on and off since starting to make changes to the UI tests targets:

```
error: input file '/Users/gio/Developer/a8c/wcios/WooCommerce/DerivedSources/ApiCredentials.swift' was modified during the build
error: input file '/Users/gio/Developer/a8c/wcios/WooCommerce/DerivedSources/ApiCredentials.swift' was modified during the build
Command CompileSwiftSources failed with a nonzero exit code
```

The funny things is that `ApiCredentials.swift` is an _output_ file, not an input one. :upside_smile:

That seems to happen on a non-clean build, and is happening [to other folks too](https://stackoverflow.com/questions/69256241/xcode-13-error-input-file-was-modified-during-the-build). It looks to be an Xcode 13.0 issue more than anything, but I thought I'd take some time to look into the setup around how that file is generated to see if there was something odd.

The changes I made are:

1. Remove the dependency on the `GenerateCredentials` aggregate target from the screenshots automation tests, because it's implicitly there already through the tests dependency on the main Woo app target – _I thought maybe the double dependency could have resulted in a race condition on the touching file_
2. Use an `.xcfilelist` for the outputs as well as the inputs of the Run Script build phase in `GenerateCredentials` – _I thought maybe Xcode 13.0 was getting confused using the list in the project_
3. Remove the `ApiCredentials.swift` membership from the screenshots target because it's not needed there anymore – _Again, I thought maybe because it was in multiple places there could have been some race condition..._

After making each of those changes, the problem persisted. In between the 2 and 3, I got [a different issue](https://stackoverflow.com/a/69308719/809944), after fixing that and coming back here, I can't reproduce this one anymore. So, I'm not sure whether one of my tweaks was effective and it was a caching issue on Xcode's end, or whether I just hit a lucky break and the issue will appear again later, which would again suggest it being related with some form of caching in the build system.

Still, I believe these changes are a net-positive improvement to our setup, so we might as well merge them.

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. – N.A.


AliSoftware:
> The funny things is that ApiCredentials.swift is an output file, not an input one. :upside_smile:

It's an output file of the script build phase, but an input file of the rest of the phases – especially the source compilation phases – afterwards. I remember a WWDC session explaining all the workings around this.

I'm going to take a deeper look at this and experiment to see if encounter the same issue as you did (just give me some time to sync back `bundle install`, `rake dependencies`, SPM, all over again then do an Xcode build from scratch, since it's been a while since I built WCiOS 😄 ).

But from what I know and learned from that WWDC session and the new build system, this issue happens because we generated/modified a file (`ApiCredentials.swift`) from a build phase A (`GenerateCredentials`) that was used as an input by a build phase B (Compile Sources) later in the phase dependency graph. And what that WWDC session said was that you need to ensure all your Input/Output files / filelists are properly configured in the build phases to tell Xcode which files might affect the phase dependency graph and its decision to re-run or not subsequent build phases during incremental builds.

AliSoftware:
Funny story: while Googling it, especially the terms "Xcode xcfilelist", the first result I finally found was a mention of it… [in one of my own GitHub repo](https://github.com/SwiftGen/SwiftGen/issues/441) – which finally led me to [WWDC'18 Session 408](https://developer.apple.com/videos/play/wwdc2018/408/).

Especially this is talked about around [this timestamp](https://developer.apple.com/videos/play/wwdc2018/408/?time=832) in the session). Basically Xcode starts by building the dependency graph _first_, using the info of those input/output files to know what phase it _should_ run or not, and in which order. And once the graph is constructed, then it will run the phases in order.

But if you don't specify the right output files in those phases, Xcode won't know that your `GenerateCredentials` build phase will generate that `ApiCredentials.swift` file when it builds the phases dependency graph, and will (wrongfully) think that other phases (like `Compile Sources` build phase) are ok to run in parallel of this one because they seem independent. And then, when it will _actually_ run the various phases and figure out that `GenerateCredentials` _did_ change a file that `Compile Sources` (which it was likely running in parallel at that time) is using, it will throw that error that you saw; because that meant the `Compile Sources` phases was not supposed to be able to run in parallel after all, but instead only after `GenerateCredentials`.

And the only way for Xcode to have known that in advance is for us to specify the output files explicitly and correctly, which explains why you made the error disappeared once you added `ApiCredentials.swift` to the output files of `GenerateCredentials`!

AliSoftware (approved):
- ✅ Soft clean build (Product > Clean Build Folder) of `WooCommerce` target
- ✅ Hard clean build (`rm -rf ~/Library/Developer/Xcode/DerivedData/`) (†) of `WooCommerce` target
- ✅ Incremental builds (1 right after hard clean build, then did 2 more after the previous incremental build) of `WooCommerce` target
- ✅ Incremental build of `WooCommerce` target, right after doing a `rm -rf WooCommerce/DerivedSources`
- ✅ Build of `WooCommerceScreenshots` target
- ✅ Build of `WooCommerceUITests` target

_(†) well, I did hit a SPM issue when building right after that nuking of DerivedData, but a {File > Swift Packages > Resolve Packages Versions} fixed it and made the build go back to green afterwards)_

mokagio:
Thanks for checking @AliSoftware, and for the link to the WWDC session.

> And the only way for Xcode to have known that in advance is for us to specify the output files explicitly and correctly, which explains why you made the error disappeared once you added `ApiCredentials.swift` to the output files of `GenerateCredentials`!

The only thing is... `ApiCredentials.swift` was already part of the `GenerateCredentials` output files 😳 The only change I made was [move it](https://github.com/woocommerce/woocommerce-ios/pull/5049/commits/b5e1f868e49aafe52a1d46bf6b4b275e69c63f43) from the list configured in the project to an `.xcfilelist` file 😞 

Maybe Xcode 13 is sensitive to inputs and outputs not being defined in the same way? 🤔

Regardless, let's merge this and see what happens. I have plenty of work left to do on this repo, which should give me heaps of opportunity to experience this issue again if none of those changes makes a difference.

mokagio:
For future reference, the R.swift issue mentioned [above](https://github.com/woocommerce/woocommerce-ios/pull/5049#ref-issue-995083529) contains lots of useful comments. The consensus seems to be that **this is a bug in Xcode 13**.

Also, [this StackOverflow answer](https://stackoverflow.com/a/69402038/809944) suggests going down to the Build Rules level, which is something we might want to look into if the issue reappears.

PR #5050, opened 2021-09-24
Shipping Labels: New model for package item to prepare for moving items between packages

itsmeichigo:
Part of #4717

# Description
While working on moving order items between packages, I realized that simply keeping product ID list in each package is not enough. Each package needs to have other information about the product it contains: quantity of items, weight, dimensions, attributes. 

# Changes
This PR aims to update `ShippingLabelPackageAttributes` by replacing product ID list with a list of information about each item contained in a package. This leads to changes in a lot of files:
- Purchase form view model: to transform the list of items to list of product IDs needed for selected / purchased packages.
- The legacy package details view model: to update the output of the selected package - a workaround for the old solution to work.
- Packages form and single package view models
- A lot of unit tests!

Also: extracted new reusable view `TitleAndSubtitleRow` from `ItemToFulfillRow`. I meant to reuse this in the new rows of original packages, but decided to move that logic for the next PR, so this change looks a bit irrelevant.

# Testing
Although this PR changes a lot of things under the hood, the expectation is that everything still works as before:
1. Make sure that your test store has WCShip plugin installed and configured packages in WPAdmin > WooCommerce > Settings > Shipping > WooCommerce Shipping.
2. Select an order eligible for creating shipping labels.
3. Configure Ship From and Ship To.
4. On Package Details screen, try selecting different packages and then tap Done. Notice that everything works as before with or without custom package weight.
5. Proceed to Carriers and Rates, notice that the rates list still loads correct items.
6. Configure payment method and proceed to purchase the label. Make sure that the purchase succeeds.
7. In `DefaultFeatureFlagService`, turn off feature flag for multiple packages. Repeat steps 2-6 and make sure that everything still works.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
@@ -41,7 +41,8 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
     @Published var selectedPackageID: String?
 
     /// List of selected package with basic info.
-    /// This is a workaround to work with multi-package solution.
+    /// This is a workaround to work with multi-package solution which requires a list of packages.
+    /// Since this legacy view model works only with one package, the returned list contains at most one item only.
     ///
     var selectedPackagesDetails: [ShippingLabelPackageAttributes] {
         guard let id = selectedPackageID, totalWeight.isNotEmpty else {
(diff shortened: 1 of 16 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
```diff
@@ -41,13 +41,16 @@ final class ShippingLabelPackageDetailsViewModel: ObservableObject {
     @Published var selectedPackageID: String?
 
     /// List of selected package with basic info.
-    /// TODO-4599: update this to properly support multi-package.
+    /// This is a workaround to work with multi-package solution.
```
pmusolino: nit: you didn't explain in the comment what is the workaround. It could be useful in my opinion for the future.

pmusolino (approved):
It works well for me, and also the size calculation seems correct like before 👍 

Code ✅ 
Testing ✅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/ShippingLabelPackageDetailsViewModel.swift
itsmeichigo: Thanks for spotting this. I've updated the comment in 8fd357897aec61505fa569205fa04ae5f1cb448d.

PR #5054, opened 2021-09-24
Order Editing: Add billing address handling

### LINKED ISSUE #4798: Order Editing: Fill address form content

ealeksandrov:
Part of https://github.com/woocommerce/woocommerce-ios/issues/4778, https://github.com/woocommerce/woocommerce-ios/issues/4784.
Closes https://github.com/woocommerce/woocommerce-ios/issues/4798.

## Description

Add ability to edit billing address and save it remotely.

### Technical Details

- Added `AddressType` enum as init arg to view model.
- Added `onEditTapped` callback closure to update `BillingInformationViewController` after save action.
- Updated data prefill logic and remote actions to handle shipping/billing based on `AddressType`.

### Note

I've noticed that **removing** (emptying) of phone or email fields ignored on remote action. Likely because of empty string/nil conversion.
Added card [here](https://github.com/woocommerce/woocommerce-ios/projects/39#card-69457552).

## Testing

1. Open existing order.
2. Tap "View Billing Information" in "Customer" section.
4. Tap "edit" icon on Billing Address row.
5. Update any of fields and tap "Done".
6. Wait for remote action to complete and modal to close automatically.
7. Observe correctly updated billing address.

## Screenshots

---
Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
```diff
@@ -93,6 +93,19 @@ private extension BillingInformationViewController {
             tableView.register(kind.loadNib(), forHeaderFooterViewReuseIdentifier: kind.reuseIdentifier)
         }
     }
+
+    /// Presents EditAddressForm modal view
+    ///
+    func editBillingAddress() {
+        let viewModel = EditAddressFormViewModel(order: order, type: .billing) { [weak self] updatedOrder in
+            self?.order = updatedOrder
```
Ecarrion, line 101: I wonder if we should refactor this screen to observe the order thru core data? 🤔 
Maybe add it as a "debt" card?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -93,12 +120,28 @@ final class EditAddressFormViewModel: ObservableObject {
...
+        switch type {
+        case .shipping:
+            modifiedOrder = order.copy(shippingAddress: updatedAddress)
+        case .billing:
+            modifiedOrder = order.copy(billingAddress: updatedAddress)
+        }
+
+        let action = OrderAction.updateOrder(siteID: order.siteID,
+                                             order: modifiedOrder,
+                                             fields: [type == .shipping ? .shippingAddress : .billingAddress]) { [weak self] result in
```
Ecarrion, line 151: This comparison worries me because it will break if we add another address type, but that is very unlikely to happen 😅

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -4,6 +4,19 @@ import Combine
 
 final class EditAddressFormViewModel: ObservableObject {
 
+    enum AddressType {
+        case shipping
+        case billing
+    }
```
Ecarrion, line 10: I was thinking if instead of having an enum, it would be a good idea to have a protocol. 

That way: we could reuse this screen to any other possible address form in the app. Not sure if it's worth it tho. Do you think we should add a "card" to tackle it leave an issue later?

Ecarrion (approved):
Looks good to me!
Just left a few thoughts.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
ealeksandrov: We can also add method like `update(order:)` and pass code data changes from order details VC. Or own core data logic would be better?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
ealeksandrov: That’s exactly why in 2 cases above I’ve used switch-case without default handling 😄
Tried here too, but field type is internal to Networking.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
ealeksandrov: Yes, moving all this stuff to protocol should work. I’ll make a quick draft next week to check this idea.
Let’s do it then on top of this implementation.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/Billing Information/BillingInformationViewController.swift
Ecarrion: While I like about the core data observations is that there is no way that one screen could get outdated because someone forgot to call a method or similar.

PR #5055, opened 2021-09-24
5022 Fix VoiceOver loop on Order Details screen

### LINKED ISSUE #4977: [Order Details] VoiceOver: Hidden Elements Should not be Available for Navigation
In an Order Details page, when navigating by element using VoiceOver (swipe-to-right), a hidden "Fees" element is accessible. And then right after swiping to right on that hidden element, the user is navigated back to the **topmost element** instead of the next element.

## Video

https://user-images.githubusercontent.com/198826/132887729-8445712b-e9f2-4854-8856-ad703fa49b46.mp4

## To Reproduce

1. Navigate to an order with no fees.
2. Turn on VoiceOver
3. Swipe-to-right until you reach the hidden Fees element. VoiceOver will say "Fees" but that is actually not visible. 

## Expectations

- [ ]  The "Fees" or any other hidden element should not be accessible. 
- [ ]  Swiping to right should continue on to the next element in the screen, not the topmost element. 

---

Internal ref: p1631199350210900-slack-C6H8C3G23

### LINKED ISSUE #5022: VoiceOver navigation loop on Order Details
## Description
When navigating the Order Details screen using VoiceOver and swiping between elements, the navigation loops back to the start after the Product Total field, which leaves much of the screen difficult to access (not impossible: directly tapping on later elements allows access.)

Additionally, the ℹ button and ⌄ related to showing information about Shipping Labels do not have accessibility labels, so are only announced as "button" by VoiceOver.

## To Reproduce

https://user-images.githubusercontent.com/2472348/134023123-148e0090-9e7a-418b-b0c8-366dcbb10584.mp4

1. Go to Orders
2. Tap on an order without fees
3. Enable VoiceOver
4. Swipe through controls until you get to Product Total
5. Continue swiping
6. Observe that you're returned to the top of the screen

## Expected behavior
It should be possible to swipe beyond Product Total through the entire screen

## Mobile Environment
Please include:
- Device: iPhone (7 plus, simulated 12 mini)
- iOS version: 14.6
- WooCommerce iOS version: 7.5.0.0

joshheald:
Fixes #5022 
Fixes #4977

## Description

The Order Details screen had hidden views, some of which were still announced by VoiceOver. These also caused a loop, making it impossible to swipe below the Payment (LedgerTableViewCell.swift) section of the screen.

The views in the ledger cell were separately exposed to VoiceOver, via the `accessibilityElement` array, however this included all the subviews, regardless of their visibility and their parent view's visibility.

Ordinarily, VoiceOver will remove hidden views from the announced set automatically, but by specifying subviews in `accessibilityElements`, we were breaking that or possibly falling foul of an a11y bug in Apple's frameworks. This change only adds visible rows to the `accessibilityElements`, which prevents the navigation loop.

The views being specified in `accessibilityElements` is a good thing for the VoiceOver experience; this cell is very information dense, and the ability to individually select individual rows and hear them separately is helpful. However, both labels in each row were individually announced, meaning the user had to swipe through two elements per row. This change groups the rows together, for easier and clearer navigation.

## Testing
1. Go to Orders
2. Tap on an order without fees
3. Enable VoiceOver
4. Swipe through controls until you get to Product Total
5. Continue swiping
6. Observe that you can swipe beyond Product Total through the entire screen

Additionally, the same test should be done on the Refund details screen:
1. Go to Orders
2. Tap on an order without fees, with a refund
3. Tap on the Refund row
4. Enable VoiceOver
5. Swipe through controls until you get to Product Total
6. Continue swiping
7. Observe that you can swipe beyond Product Total through the entire screen

https://user-images.githubusercontent.com/2472348/134694245-c4c0d418-20d5-4dd2-a0a9-9a6559d81748.mp4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/LedgerTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/LedgerTableViewCell.swift
@@ -36,66 +36,81 @@ final class LedgerTableViewCell: UITableViewCell {
     /// Configure an order payment summary "table"
     ///
     func configure(with viewModel: OrderPaymentDetailsViewModel) {
-        subtotalLabel.text = Titles.subtotalLabel
-        subtotalValue.text = viewModel.subtotalValue
-
-        discountLabel.text = viewModel.discountText
-        discountValue.text = viewModel.discountValue
-        discountView.isHidden = viewModel.shouldHideDiscount
-
-        feesLabel.text = Titles.feesLabel
-        feesValue.text = viewModel.feesValue
-        feesView.isHidden = viewModel.shouldHideFees
-
-        shippingLabel.text = Titles.shippingLabel
-        shippingValue.text = viewModel.shippingValue
-
-        taxesLabel.text = Titles.taxesLabel
-        taxesValue.text = viewModel.taxesValue
-        taxesView.isHidden = viewModel.shouldHideTaxes
-
-        totalLabel.text = Titles.totalLabel
-        totalValue.text = viewModel.totalValue
-
-        accessibilityElements = [subtotalLabel as Any,
-                                 subtotalValue as Any,
-                                 discountLabel as Any,
-                                 discountValue as Any,
-                                 feesLabel as Any,
-                                 feesValue as Any,
-                                 shippingLabel as Any,
-                                 shippingValue as Any,
-                                 taxesLabel as Any,
-                                 taxesValue as Any,
-                                 totalLabel as Any,
-                                 totalValue as Any
-                                ]
+        configureSubtotal(label: Titles.subtotalLabel, value: viewModel.subtotalValue)
+        configureDiscount(label: viewModel.discountText, value: viewModel.discountValue, hidden: viewModel.shouldHideDiscount)
+        configureFees(label: Titles.feesLabel, value: viewModel.feesValue, hidden: viewModel.shouldHideFees)
+        configureShipping(label: Titles.shippingLabel, value: viewModel.shippingValue, hidden: false)
+        configureTaxes(label: Titles.taxesLabel, value: viewModel.taxesValue, hidden: viewModel.shouldHideTaxes)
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/ReusableViews/LedgerTableViewCell.swift
```diff
@@ -36,66 +36,81 @@ final class LedgerTableViewCell: UITableViewCell {
...
+        configureAccessibility()
+    }
+
+    private func configureSubtotal(label: String, value: String) {
+        subtotalLabel.text = label
+        subtotalValue.text = value
+        subtotalView.accessibilityLabel = "\(label) \(value)"
+    }
+
+    private func configureDiscount(label: String?, value: String?, hidden: Bool) {
```
ctarda: Not a big deal, but maybe the body of these three methods could be extracted to a separate method, and pass the label as a parameter?

ctarda (approved):
Tests well, on device with voice over activated.

I left one question for your consideration. 

:shipit:

WooCommerce/Classes/ViewRelated/ReusableViews/LedgerTableViewCell.swift
joshheald: @ctarda See latest changes... I'd considered having a shared method before I PRed, but passing in the three views each time seemed like it would make the method call unwieldy and reduce readability.

You prompted me to revisit it, and I tried grouping the views to avoid this issue. I quite like the nested struct here, since it keeps things readable at the call site. What do you think? 

Thanks for prompting me to improve it.

ctarda:
Looks good @joshheald , :shipit:

WooCommerce/Classes/ViewRelated/ReusableViews/LedgerTableViewCell.swift
ctarda: Nice, merge whenever you are ready!

PR #5056, opened 2021-09-24
Add Yosemite support to Fakes.Framework

Ecarrion:
part of https://github.com/woocommerce/woocommerce-ios/issues/3823

# Why

As requested on https://github.com/woocommerce/woocommerce-ios/pull/5044#discussion_r715167994, this PR adds support to `Yosemite` entities within the `Codegen.package` and `Fakes.framework`

# How

- Update fakes sourcery template to import `Yosemite`
- Added a new `Yosemite` sourcery configuration file.
- Update sourcery templates to output generated files with different file names (one for `Yosemite` and one for `Networking`) 
- Update `rakefile` to run sourcery with the new `Yosemite` template.
- Make `ProductReviewFromNoteParcel` conform to `GeneratedFakeable`
- Add `Yosemite` as a dependency for `Fakes.Framework`
- Update docs

# Testing steps
- Make sure you can use `ProductReviewFromNoteParcel.fake()` in tests.

Update release notes:


Ecarrion:
Hi @jkmassel do you know what could be [happening here](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/14409/workflows/31543f02-b04d-4d16-b78f-93ab4d0d1477/jobs/26981)?

I've run it twice and the same error happens, locally works fine tho. 🤔

Fakes/Fakes/Networking.generated.swift
```diff
@@ -1,5 +1,7 @@
 // Generated using Sourcery 1.0.3 — https://github.com/krzysztofzablocki/Sourcery
 // DO NOT EDIT
+
+import Yosemite
```
jaclync, line 4: Networking probably doesn't have to know about Yosemite? It looks like it's from the codegen though, might have to hard code the framework dependency 🤔

jaclync (commented):
Wow thanks for the quick PR @Ecarrion! 🙇🏻‍♀️

I was able to reproduce the test failure locally as in CI:
`Underlying Error: The bundle “NetworkingTests” couldn’t be loaded. The bundle couldn’t be loaded. Try reinstalling the bundle.`

After some testing, I think the error is caused by the Fakes framework including Yosemite for NetworkingTests. If I remove the Yosemite framework from the Fakes framework and references to Yosemite from the Fakes Swift files, then the NetworkingTests run.

This might not be a simple task, but maybe we have to manually add the `*.generated.swift` files to individual Networking/Yosemite frameworks then point the codegen code to write to those file paths? I'm guessing that's how Copiable works for multiple frameworks (cc @shiki)?

shiki:
I am also unable to run the tests locally. 

```
2021-09-27 12:02:25.619651-0600 xctest[34221:12442083] The bundle “NetworkingTests” couldn’t be loaded because it is damaged or missing necessary resources. Try reinstalling the bundle.
2021-09-27 12:02:25.620027-0600 xctest[34221:12442083] (dlopen_preflight(/Users/shiki/Library/Developer/Xcode/DerivedData/WooCommerce-bpzhdeekaixnvkfxqawsuzqsipdk/Build/Products/Debug-iphonesimulator/NetworkingTests.xctest/NetworkingTests): Library not loaded: @rpath/FormatterKit.framework/FormatterKit
  Referenced from: /Users/shiki/Library/Developer/Xcode/DerivedData/WooCommerce-bpzhdeekaixnvkfxqawsuzqsipdk/Build/Products/Debug-iphonesimulator/Yosemite.framework/Yosemite
  Reason: image not found)
Program ended with exit code: 82
```

I think what Jaclyn said is correct. 

> This might not be a simple task, but maybe we have to manually add the *.generated.swift files to individual Networking/Yosemite frameworks then point the codegen code to write to those file paths? I'm guessing that's how Copiable works for multiple frameworks (cc @shiki)?

Maybe that'll work but Ernesto knows this Fakes framework better than me so he probably already has a solution in mind. 😄

Ecarrion:
Thanks for testing @jaclync and @shiki!

I'm now able to reproduce it! Stupid Xcode caches 😅 

>  but maybe we have to manually add the *.generated.swift files to individual Networking/Yosemite frameworks then point the codegen code to write to those file paths? I'm guessing that's how Copiable works for multiple frameworks 

I decided against that because as these types are only meant to be used for tests, we can't share code from a test target to another test target. 

A workaround would be to add those files into all test target's compile sources but it felt meh at that time!

----

As for the actual problem, I'm not sure why Xcode is complaining but as soon as I add any Yosemite code into the Fakes framework, the networking tests won't run.

I've been able to find a solution by using a static lib instead of a dynamic framework but I still need to validate a few extra cases.

As I'm a bit tight this sprint I'm going to turn this PR into a draft and will continue investigating the problem next sprint.

@jaclync If I'm blocking you, feel free to copy the `Yosemite.generate.swift` file from this PR and add it to the Yosemite test sources. I will take care of removing it later if/when I come up with a solution.

Sorry for any inconvenience

jaclync:
Thanks so much for investigating the issue and potential solutions @Ecarrion! 🙏  this is not blocking me, and I appreciate your effort to improve the test code. I agree that adding the files to test targets' compile sources doesn't feel ideal.

I wonder if we can try adding Yosemite to NetworkingTests target in the Podfile to includes Yosemite, will try it later this week or next sprint! (also a bit swamped this week from push notifications)

https://github.com/woocommerce/woocommerce-ios/blob/2925e5e4ca80058aa68e1557eaacb54f3eb65bc2/Podfile#L134

jaclync:
Hey @Ecarrion 👋 I got to try adding Yosemite dependencies to NetworkingTests target and it seems to work in [this testing branch](https://github.com/woocommerce/woocommerce-ios/compare/jaclync/yosemite-fakes-testing?expand=1)! My guess is that the Fakes framework includes Yosemite in the Xcode target settings, but there are some Yosemite dependencies defined in Cocoapods (like the `FormatterKit` reference in [Jayson's log](https://github.com/woocommerce/woocommerce-ios/pull/5056#issuecomment-928130207)). Would like to hear what you think about this solution!

Ecarrion:
Hi @jaclync, thanks a lot for the help, I'm happy with the workaround and will keep thinking if there is a way to handle those circular dependencies better!

I have applied the change and this is ready for review again!

jaclync (approved):
Yosemite fakes work nicely now! I tested using `ProductReviewFromNoteParcel.fake()` in `WooCommerceTests` and `YosemiteTests` targets and 💯 PR LGTM :shipit: 

Having a mix of dependency management tools also adds to the complexity now, we can see if it's easier to just consolidate into one in the future probably with SPM.

Issue #4943, opened 2021-09-06
Merge `updateOrder` and `updateOrderStatus` actions.


Ecarrion:
For the order editing project, we introduced a new `updateOrder` action that lets the consumer choose which order field to update remotely.

The order screen uses an action named `updateOrderStatus` to update the order status.

## Tasks

- [ ] Merge `updateOrderStatus` action into `updateOrder` action.
- [ ] Tests that updating the order status doesn't break anything or generates UI glitches.

talpert022:
I'll have a go at solving this issue!

Ecarrion:
We have noticed that we need the current `updateOrderStatus` to maintain the optimistic update UX when updating user status.

I'm going to close this for now.

PR #5059, opened 2021-09-26
Merged updateOrderStatus and updateOrder in order details vc

talpert022:
For issue #4943 

# Overview

This PR makes it so the order screen now uses the Order Action -- updateOrder instead of updateOrderStatus -- to change the status of an order. 

# Fix and technical details 

I noticed that the UI would lag a little bit after changing an order status using `OrderAction.updateOrder`. I think this is because `updateOrderStatus` would optimistically update the order status while `updateOrder` uses the function `upsertStoredOrdersInBackground` to update or insert the order entities in a background thread. This process I assume takes longer and only executes the completion handler in `OrderDetailsViewController`, which changes the UI of the order status, once this background process has completed. 

My fix is similar to changing the address in #5020. Starting the process of updating an order activates a spinner, which will run until the process of networking and storage is complete, and then dismiss the `OrderStatusListViewController` once the UI has been updated. This process is shown below. 

## Changes
- Added `status` field support to `updateOrder` action inside `OrdersRemote`
- Changed `updateOrderStatusAction` function in `OrderDetailsViewController.swift` to use `OrderAction.updateOrder` instead of `OrderAction.updateOrderStatus.`
- Added escaping closure to `statusList.didSelectApply` that dismisses the status list, which is called inside `updateOrderStatusAction`, after the status is changed in the UI. 

# Screen recording

https://user-images.githubusercontent.com/51763119/134824423-727c5b9b-634c-4b84-82d9-4eb70a205770.mp4

# Testing

1. Open existing order
2. Tap the edit button next to the order status
3. Change the status of the order and click apply


WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
```diff
@@ -10,6 +10,7 @@ final class OrderStatusListViewController: UIViewController {
     ///
     private var indexOfSelectedStatus: IndexPath? {
         didSet {
+            // Maybe don't change if the indexpath is the same as when the list view is presented?
```
Ecarrion:
```suggestion
            // Maybe don't change if the index-path is the same as when the list view is presented?
```

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
```diff
@@ -104,7 +105,12 @@ extension OrderStatusListViewController {
         navigationItem.setRightBarButton(rightBarButton, animated: false)
         deActivateApplyButton()
     }
-
+    func activateRightSpinner() {
+        let activityIndicator = UIActivityIndicatorView(frame: CGRect(x: 0, y: 0, width: 20, height: 20))
+        let rightBarButton = UIBarButtonItem(customView: activityIndicator)
+        navigationItem.setRightBarButton(rightBarButton, animated: true)
+        activityIndicator.startAnimating()
+    }
```
Ecarrion:
```suggestion

   func activateRightSpinner() {
        let activityIndicator = UIActivityIndicatorView(style: .medium)
        let rightBarButton = UIBarButtonItem(customView: activityIndicator)
        navigationItem.setRightBarButton(rightBarButton, animated: true)
        activityIndicator.startAnimating()
    }

```

Changes:
- We don't need to give the spinner a specific frame
- Please add new lines between methods

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -1004,7 +1009,9 @@ private extension OrderDetailsViewController {
         let title = String.localizedStringWithFormat(titleFormat, orderID)
         let actionTitle = NSLocalizedString("Retry", comment: "Retry Action")
         let notice = Notice(title: title, message: nil, feedbackType: .error, actionTitle: actionTitle) { [weak self] in
-            self?.setOrderStatus(to: status)
+            self?.setOrderStatus(to: status) {
+                print("Retry Action Complete")
```
Ecarrion: Could you remove this debug code?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
```diff
@@ -126,6 +132,7 @@ extension OrderStatusListViewController {
             didSelectCancel?()
             return
         }
+        activateRightSpinner()
```
Ecarrion: I'm not sure if this is a good idea. It looks like this view controller was designed to be decoupled from the actual action that is performed, so I think it will make sense for the consumer on `didSelectApply` side to decide if this VC should render a spinner or not.

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -929,8 +929,9 @@ private extension OrderDetailsViewController {
         }
 
         statusList.didSelectApply = { (selectedStatus) in
-            statusList.dismiss(animated: true) {
-                self.setOrderStatus(to: selectedStatus)
+            self.setOrderStatus(to: selectedStatus) {
+                statusList.configureRightButton()
```
Ecarrion: Could you explain why do you need to reconfigure the right button?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
Ecarrion: I think that would be a good idea, do you plan to make the change?

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -939,46 +940,50 @@ private extension OrderDetailsViewController {
...
 
         ServiceLocator.stores.dispatch(done)
 
         ServiceLocator.analytics.track(.orderStatusChange,
-                                       withProperties: ["id": orderID,
+                                       withProperties: ["id": order.orderID,
                                                         "from": undoStatus.rawValue,
                                                         "to": newStatus.rawValue])
-
+        // TODO: Move this to display once the list view dismisses
```
Ecarrion: What does this means?

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -939,46 +940,50 @@ private extension OrderDetailsViewController {
...
+        return OrderAction.updateOrder(siteID: siteID, order: newOrder, fields: [.status], onCompletion: { [weak self] result in
+            guard let self = self else { return }
+            switch result {
+            case .success:
                 NotificationCenter.default.post(name: .ordersBadgeReloadRequired, object: nil)
-                self?.syncNotes()
+                self.syncNotes()
                 ServiceLocator.analytics.track(.orderStatusChangeSuccess)
+                closeListView()
                 return
```
Ecarrion: I don't think we need this return anymore, right?

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -939,46 +940,50 @@ private extension OrderDetailsViewController {
...
                                                             "to": undoStatus.rawValue])
         }
     }
-
     /// Returns an Order Update Action that will result in the specified Order Status updated accordingly.
     ///
-    private func updateOrderStatusAction(siteID: Int64, orderID: Int64, status: OrderStatusEnum) -> Action {
-        return OrderAction.updateOrderStatus(siteID: siteID, orderID: orderID, status: status, onCompletion: { [weak self] error in
-            guard let error = error else {
+    private func updateOrderStatusAction(siteID: Int64, order: Order, status: OrderStatusEnum, closeListView: @escaping () -> Void) -> Action {
```
Ecarrion:
```suggestion
    private func updateOrderStatusAction(siteID: Int64, order: Order, status: OrderStatusEnum, onCompletion: @escaping () -> Void) -> Action {
```

We should probably keep the convention and call these closures, `onCompletion`

Ecarrion (changes requested):
Hi @talpert022! Thanks for contributing! 🎉 

I left you some suggestions to consider as code comments.

Also, as you are changing the UX of the flow. I think you should also update the scenario where the update fails, as we don't need to dismiss the VC or show the success notice.

----

@Garance91540 @adamzelinski 

This PR changes a bit the UX for changing the order status. Are you ok with it?

## Current

https://user-images.githubusercontent.com/562080/135109674-deaea600-ff90-48fe-9b54-fb10843645b0.mov


## Proposed in PR 

https://user-images.githubusercontent.com/562080/135109727-c5594719-7262-457c-a9b4-429ab1d3f62d.mov

talpert022:
Hi @Ecarrion, thank you for your comments. I realize that my update changes the UX a bit, for context here is the UI lag I was talking about without the spinner, while using `updateOrder` instead of `updateOrderStatus`. Let me know if you think it is a big deal or not. 

https://user-images.githubusercontent.com/51763119/135736316-24b37681-a319-42ed-bb1d-8011595b7923.mp4

I made the changes you suggested, including making sure the apply button only is available in the OrderStatusListView when the status has been changed. This specific change is in [e3475bf](https://github.com/woocommerce/woocommerce-ios/pull/5059/commits/e3475bf7f5d4b163032c5c8db680b6d7131c32b4). I also followed your suggestion for the UX to only show the success notice if the status change is successful (before the success notice would show followed by the error notice on a failed update). This is the resulting behavior from a failed status change. 

https://user-images.githubusercontent.com/51763119/135736513-cc27e41f-3ae7-4865-9903-f31e45d9066e.mp4

I was wondering if you had any suggestions for making the solution cleaner. I currently call two escaping closures inside `updateOrderStatusAction`, `closeList` and `showSuccess`, depending on a success or failure case. However, setOrderStatus doesn't always need to close a OrderStatusListView, for example, when pressing the undo button on the success notice: 
```
let notice = Notice(title: title, message: nil, feedbackType: .error, actionTitle: actionTitle) { [weak self] in
            self?.setOrderStatus(to: status) {}
}
```
This causes an awkward area where nothing is passed into the closeList closure of setOrderStatus. Is there a work around for this you would recommend. I could make the closure an optional and unwrap it inside `updateOrderStatusAction`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Order Summary Section/Edit Order Status/OrderStatusListViewController.swift
talpert022: Yes I made this change in [this commit](https://github.com/woocommerce/woocommerce-ios/pull/5059/commits/e3475bf7f5d4b163032c5c8db680b6d7131c32b4).

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
talpert022: You are right, I don't need to do this. I fixed it

Garance91540:
Hi @Ecarrion , this seems ok to me! thank you for checking ;)

Ecarrion:
Hi @talpert022, thanks for the effort you have put into this!
We are discussing internally what is the best way to move forward UX-wise and will let you know as soon as we have reached a conclusion.

Cheers!

Ecarrion (changes requested):
Hi @talpert022,

Upon discussions, we have decided that we don't want to drop the optimistic update experience, we think it is more intuitive and makes the app feel faster for slow hosts.

We appreciate your contribution. You've helped us realize how we want to approach our UX in the future.

I'm going to close this PR, but feel free to submit a new one if you have any other approaches!

Thanks from the Woo team.

talpert022:
Thanks @Ecarrion for the feedback. It's really fun getting to know this project, and glad I could help out. Would you be will to accept the changes I made in [this commit](https://github.com/woocommerce/woocommerce-ios/pull/5059/commits/e3475bf7f5d4b163032c5c8db680b6d7131c32b4), to make the apply button (on the StatusListView Controller) only highlight for a changed status, as a separate PR?

Ecarrion:
> Thanks @Ecarrion for the feedback. It's really fun getting to know this project, and glad I could help out. Would you be will to accept the changes I made in [this commit](https://github.com/woocommerce/woocommerce-ios/pull/5059/commits/e3475bf7f5d4b163032c5c8db680b6d7131c32b4), to make the apply button (on the StatusListView Controller) only highlight for a changed status, as a separate PR?

@talpert022, sure that looks like it could be useful!

PR #5061, opened 2021-09-27
Shipping Labels: Ship items in original packaging

itsmeichigo:
Part of #4717 

# Description
This PR adds implementation to move items to individual packages using their original packaging. It also includes validation for original packaging, which requires valid dimensions for individual items.

# Changes
- Packages view model:
   - Handles moving an item to an individual package using original packaging.
   - Updates validation for each package based on new validation variable.
- Single package view model:
   - Adds new parameter to initializer to determine if package is using original packing of its item.
   - Configures displayed dimensions for the original package
   - Updates validation to check for both total weight and dimensions if the package uses original packaging.
- Singe package view:
   - Adds new rows for original package
   - Adds validation error for original package 

# Demo
| Valid dimensions | Invalid dimensions |
| ----- | ----- |
|  |  |

# Testing steps
2. Select an order eligible for creating shipping label. The order item list should contain a product with valid dimensions configured for its packaging.
4. Proceed to configure Package Details. Select the product with valid dimensions and tap Move > select Ship in original packaging.
5. Notice that the item is displayed in another package, which shows a new row: Original packaging. Correct dimensions are displayed in another row: Item dimensions.
6. Repeat steps 2-5 for another order that contains a product with invalid dimensions. Notice that the invalid dimensions are highlight in red color, with a validation error underneath the Item dimensions row, explaining the error. Done button should be disabled.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
+++ b/WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
@@ -4,6 +4,9 @@ import Foundation
 ///
 struct ShippingLabelPackageAttributes: Hashable, Equatable {
 
+    /// Default box ID for boxes shipping in original packaging.
+    static let originalPackagingBoxID = "individual"
+
     /// ID of the selected package.
     let packageID: String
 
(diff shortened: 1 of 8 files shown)


pmusolino (approved):
This is just great, I didn't notice any issue 🎉 great work!

PS: Currently it is not possible to compact every package. Should be handled in a future PR?

itsmeichigo:
@pmusolino thanks for spotting the issue with collapsing packages. I'll update it in #5065!

PR #5065, opened 2021-09-27
Shipping Labels: Move item to new package and existing packages

### LINKED ISSUE #4717: Shipping Labels: allow moving items to new packages or shipping them in their original packaging.

itsmeichigo:
Closes #4717 
⚠️ This PR depends on #5061, so please make sure that PR is reviewed and merged first ⚠️

# Description
This PR completes moving items between packages by adding options to move item to a new package or existing packages. The general rules for each option to be available are:
- For each item in a package, there's an option to move to other existing packages that are not in original packaging.
- If an item is in original packaging, there's an option to move it back to a predefined / custom package.
- If an item is not in original packaging, and the item quantity is larger than 1 / or the package containing it has at least another item - then there's an option to move it to a new package.

# Changes
- Updates single package view model with a `packageName` variable to be accessed from outside.
- Updates logic for buttons on the Move Item action sheet.
- Handles logic to move item to a new package and to move it to other existing packages.

# Demo
https://user-images.githubusercontent.com/5533851/134892148-bc1829cc-6d27-409c-ad5d-1f5f2867b4ef.mp4

# Testing steps
2. Select an order eligible for creating shipping label. The order item list should contain more than one item / or there should be an order item whose quantity is larger one.
4. Proceed to configure Package Details. Tap Move on one of the items, select Add to new package. Notice that the item is now in a new package, the old package is also updated to reflect the change.
5. Continue moving each item to new package until no other package as more than one item. Notice that there's no more option to move to new package on any item now.
6. Choose an item and move it to an existing package. Notice that the old package is removed and the existing package is updated correctly.
7. Continue moving items to make sure that there's no issue at all.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
+++ b/WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
@@ -16,3 +16,9 @@ struct ShippingLabelPackageAttributes: Equatable {
     /// List of items in the package.
     let items: [ShippingLabelPackageItem]
 }
+
+extension ShippingLabelPackageAttributes {
+    var isOriginalPackaging: Bool {
+        packageID == Self.originalPackagingBoxID
+    }
+}
--- a/WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
+++ b/WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
@@ -21,4 +21,25 @@ extension ShippingLabelPackageAttributes {
     var isOriginalPackaging: Bool {
         packageID == Self.originalPackagingBoxID
     }
+
+    func partitionItems(using productOrVariationID: Int64) -> (matchingItem: ShippingLabelPackageItem?, rest: [ShippingLabelPackageItem]) {
+        var matchingItem: ShippingLabelPackageItem?
+        var updatedItems: [ShippingLabelPackageItem] = []
+        for item in items {
+            if item.productOrVariationID == productOrVariationID, matchingItem == nil {
+                // If found an item with matching product or variation ID,
+                // create a copy of the item with quantity = 1.
+                matchingItem = ShippingLabelPackageItem(copy: item, quantity: 1)
+
+                // If the item has quantity > 1, create a copy of the item with the reduced quantity and append to the list.
+                if item.quantity > 1 {
+                    let newItem = ShippingLabelPackageItem(copy: item, quantity: item.quantity - 1)
+                    updatedItems.append(newItem)
+                }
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
```diff
@@ -16,3 +16,30 @@ struct ShippingLabelPackageAttributes: Equatable {
     /// List of items in the package.
     let items: [ShippingLabelPackageItem]
 }
+
+extension ShippingLabelPackageAttributes {
+    var isOriginalPackaging: Bool {
+        packageID == Self.originalPackagingBoxID
+    }
+
+    func partitionItems(using productOrVariationID: Int64) -> (matchingItem: ShippingLabelPackageItem?, rest: [ShippingLabelPackageItem]) {
```
pmusolino, line 29: minor nit: I think a small description would be useful to understand what it does and where it will be used, without necessarily having to read the code.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackagesFormViewModel.swift
```diff
@@ -152,10 +151,39 @@ private extension ShippingLabelPackagesFormViewModel {
...
-            // if package is not original packaging, add option to ship in original package
-            if currentPackage.packageID != ShippingLabelPackageAttributes.originalPackagingBoxID {
+
+            // Add options to move to other packages.
+            for (index, package) in selectedPackages.enumerated() {
+                guard !package.isOriginalPackaging else {
+                    continue
+                }
+                if index != packageIndex {
+                    guard let name = itemViewModels[safe: index]?.packageName else {
```
pmusolino, line 161: 👍 for using `safe`

pmusolino (approved):
A lot of logic, but everything works as expected 💯 

Code ✅ 
Testing ✅

WooCommerce/Classes/Model/ShippingLabelPackageAttributes.swift
itsmeichigo: I have updated the comment in f29683e1e7e6031c34cb75075987e9b0328f084e - thanks for the suggestion!

PR #5067, opened 2021-09-27
Order Editing: Adds toggle to copy Shipping/Billing Address

### LINKED ISSUE #4808: Order Editing: Handle "copy address" button
Screen Shot 2021-08-11 at 9 54 50 AM

Tasks:
- [x] Add button to copy billing/shipping address
- [x] Make sure the order data model is properly populated before updating the order remotely.

Ecarrion:
closes #4808  #4778

# Why

Now that we can update shipping & billing addresses, it is super convenient to be able to update one and also update the other at the same time, this will be done by switching a toggle called "Use as Shipping|Billing address"

# How

- Update `AddressForm` to add the a new row that includes the toggle.
- Update `AddressFormViewModel` to hold the switch value.
- Update `AddressFormViewModel` to update the order correctly when the `useAs` toggle is on.

# Demo

https://user-images.githubusercontent.com/562080/134942368-73e63cfd-3231-4fa3-8779-33edc9625767.mov

# Testing Steps
- Navigate to edit an order shipping address
- Update any of the orders fields
- Enable the "Use as Billing Address" toggle.
- Tap the "Done" button
- Navigate to The Billing address screen
- See that it got updated with the shipping address values.

# Pending
- See if it's possible to save that "use as ... address" toggle state remotely. [Card](https://github.com/woocommerce/woocommerce-ios/projects/39#card-69557498)

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
@@ -194,6 +194,8 @@ extension EditAddressFormViewModel {
         var country: String = ""
         var state: String = ""
 
+        var useAsToggle: Bool = false
+
         mutating func update(with address: Address) {
             firstName = address.firstName
             lastName = address.lastName
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
@@ -137,18 +137,25 @@ final class EditAddressFormViewModel: ObservableObject {
     ///
     func updateRemoteAddress(onFinish: @escaping (Bool) -> Void) {
         let updatedAddress = fields.toAddress(country: selectedCountry, state: selectedState)
+        let orderFields: [OrderUpdateField]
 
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -137,18 +137,25 @@ final class EditAddressFormViewModel: ObservableObject {
     ///
     func updateRemoteAddress(onFinish: @escaping (Bool) -> Void) {
         let updatedAddress = fields.toAddress(country: selectedCountry, state: selectedState)
+        let orderFields: [OrderUpdateField]
 
         let modifiedOrder: Yosemite.Order
         switch type {
+        case .shipping where fields.useAsToggle:
```
ealeksandrov, line 144: Nice!

ealeksandrov (approved):
LGTM :shipit:

Couple of minor cases:
1. I can't just copy existing address without modifying it to enable "done" button.
2. Empty number/email doesn't get transferred and we already have a [card](https://github.com/woocommerce/woocommerce-ios/projects/39#card-69457552) for "nil" issues.

Ecarrion:
> I can't just copy existing address without modifying it to enable "done" button.

Yup, I've decided against it until there is a way to store that property remotely! 😅 
Will keep you updated! [Card](https://github.com/woocommerce/woocommerce-ios/projects/39#card-69557498)

PR #5068, opened 2021-09-27
[Mobile Payments] Payments Unavailable - Learn More - Replace AttributedText with Link

### LINKED ISSUE #5031: [Mobile Payments] Crash in attributed string initialization for InPersonPaymentsLearnMore
This is similar to the [fix made recently for UIKit](https://github.com/woocommerce/woocommerce-ios/pull/4996), but this is in the SwiftUI learn more component.

A crash occurs, frequently but not always, when initializing the learnMore string in the InPersonPaymentsLearnMore view

This should be fixed before/with #5030

allendav:
Closes #5031 

Changes
- Replaces HStack with AttributedText with a [Link](https://developer.apple.com/documentation/swiftui/link)
- Makes the entire learn more prompt tappable and avoids the crash
- Stand in until we move our minimum to iOS15

To test:
- Switch your store to Canada in wp-admin > WooCommerce > Settings > General > Country
- Launch the app and tap on the gear (Settings) then In-Person Payments

Screenshots and movies:

https://user-images.githubusercontent.com/1595739/134986731-3a832090-d55c-4cff-a726-58506687054b.MP4

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsLearnMore.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/InPersonPaymentsLearnMore.swift
@@ -4,25 +4,39 @@ struct InPersonPaymentsLearnMore: View {
     @Environment(\.customOpenURL) var customOpenURL
 
     var body: some View {
-        HStack(alignment: .center, spacing: 20) {
-            Image(uiImage: .infoOutlineImage)
-                .resizable()
-                .foregroundColor(Color(.textSubtle))
-                .frame(width: iconSize, height: iconSize)
-            AttributedText(Localization.learnMore)
-                .font(.subheadline)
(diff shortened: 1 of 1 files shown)


koke (approved):
This looks good, although these screens also show the `Contact us` link from `InPersonPaymentsSupportLink`, which would likely crash in the same way, so we probably want to change that one as well

allendav:
@koke - As I look closer at `InPersonPaymentsSupportLink` I don't see it kicking off HTML parsing ala `htmlToAttributedString` so I don't think it can crash like `InPersonPaymentsLearnMore`.

I'm going to proceed with this PR as is, I will open a separate issue for `InPersonPaymentsSupportLink` if you think we should change it anyways.

koke:
> @koke - As I look closer at `InPersonPaymentsSupportLink` I don't see it kicking off HTML parsing ala `htmlToAttributedString` so I don't think it can crash like `InPersonPaymentsLearnMore`.

Oh, that's right. I forgot I built the attributed string in chunks for that one

PR #5070, opened 2021-09-28
Shipping Labels: Fix validation for Package Details screen in multi-package cases

itsmeichigo:
Part of #4599 

# Description
Previously when working on validation for Package Details screen, I used selected package ID to as the key for the validation dictionary. The problem is we can have several packages with the same package ID, so this is not a good way to keep track of validation states of packages.

Another problem is when shipping items to original packaging and then moving them back together, the states of the old packages weren't cleared. This leads to the Done button not being enabled after moving an item without dimensions from original packaging to a predefined package.

This PR makes sure that the validation states are reset for every new set of `itemViewModels`.

# Changes
- Updates `packagesValidation` to keep track of package index instead of selected package ID.
- Resets `packagesValidation` at the beginning of `observeItemViewModels`.

# Demo
https://user-images.githubusercontent.com/5533851/135023901-33204c11-6e7f-401f-838c-a93ebb887cdb.mp4

# Testing steps
2. Select an order eligible for creating shipping label. The order item list should contain more than one item / or there should be an order item whose quantity is larger one. Try with products / variations without dimensions.
4. Proceed to configure Package Details. Move an item without dimensions to original packaging. Notice that the Done button is disabled.
5. Move the item back to a predefined / custom package (using Add to a new package), or move it to an existing package. Notice that the Done button should now be enabled.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackagesFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Package Details/Multi-package/ShippingLabelPackagesFormViewModel.swift
@@ -38,9 +38,9 @@ final class ShippingLabelPackagesFormViewModel: ObservableObject {
 
     private var cancellables: Set<AnyCancellable> = []
 
-    /// Validation states of all items.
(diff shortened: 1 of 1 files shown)


pmusolino (approved):
LGTM :shipit:

pmusolino:
I'm going to merge this PR for you @itsmeichigo since I need these changes 👯

PR #5072, opened 2021-09-28
Shipping Labels: Fix customs form issues for multiple packages

### LINKED ISSUE #4718: Shipping Labels: update the customs form to accept multiple packages
Update the customs form step in shipping labels form to handle multiple packages

Schermata 2021-08-03 alle 17 38 18

itsmeichigo:
Closes #4718 

# Description
Although I did prepare for multi-package support in customs forms, I found that there are still some issues with the current implementation:
- The customs forms are differentiated by package IDs, which can cause issues since we can have different packages of the same ID. Hence when we update one of the package, the change is reflected on another with the same ID and that's not what we want.
- Validation is not working properly as well since validation states are being tracked by package ID, which is again incorrect.
- The customs form list view model is currently fetching products and variations to update information customs form items. All these values however can now be fetched from the package details, so these trips to the database are redundant.

# Changes
- Updates customs form list by differentiating packages using their indices in the list.
- Updates customs form validation: keep track of each form using its index in the list.
- Shows error on the header of each package when validation for its fails.
- Updates `ShippingLabelPackageItem` with value of the item.
- Updates initializer of `ShippingLabelCustomsForm` to accept a list of items instead of just product IDs.
- Updates default customs forms and removes usage of database in customs form list.

# Demo
https://user-images.githubusercontent.com/5533851/135067055-46b0bf23-aa3c-48d8-9069-4db9af58d0d5.mp4

# Testing steps
2. Create an international order (i.e origin country different from destination country). The order item list should contain more than one item / or there should be an order item whose quantity is larger one.
3. In Orders tab, select the new order and select Create Shipping Label. Configure Ship From and Ship To.
4. Proceed to configure Package Details. Move items to different packages, keep at least 2 packages with the same custom / predefined package.
5. Proceed to configure Customs form. Notice that value for all items in the forms are correct.
6. Update content type / restriction type of one of the two packages that share the same package ID. Notice that only that one form is updated, the other stays intact.
7. Update values of any form so that validation fails; notice that an error icon is shown on the header of that form and Done button is disabled.
8. Turn off feature flag for multi-package support, notice that everything still works correctly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -606,8 +606,7 @@ private extension ShippingLabelFormViewModel {
         return nil
     }
 
-    /// Temporary solution for creating default customs forms.
-    /// When multi-package support is available, we should create separate form for each package ID.
+    /// Create customs forms based on `selectedPackageDetails` and default values for HS Tariff number and origin country.
     ///
     private func createDefaultCustomsFormsIfNeeded() -> [ShippingLabelCustomsForm] {
         guard customsFormRequired, selectedPackagesDetails.isNotEmpty else {
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -618,23 +618,21 @@ private extension ShippingLabelFormViewModel {
 
         return selectedPackagesDetails.map { package -> ShippingLabelCustomsForm in
             let packageName: String = {
-                guard let response = packagesResponse else {
-                    return ""
+                guard package.isOriginalPackaging else {
+                    return package.packageID
                 }
-
-                if let customPackage = response.customPackages.first(where: { $0.title == package.packageID }) {
-                    return customPackage.title
-                }
-
-                for option in response.predefinedOptions {
-                    if let package = option.predefinedPackages.first(where: { $0.id == package.packageID }) {
-                        return package.title
-                    }
-                }
-
-                return ""
+                return package.items.first?.name ?? ""
             }()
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -618,23 +618,21 @@ private extension ShippingLabelFormViewModel {
...
-
-                for option in response.predefinedOptions {
-                    if let package = option.predefinedPackages.first(where: { $0.id == package.packageID }) {
-                        return package.title
-                    }
-                }
-
-                return ""
+                return package.items.first?.name ?? ""
             }()
```
rachelmcr: If the package isn't in the original packaging, we still need to get the package title and return that instead of `package.packageID`. Otherwise, predefined packages (where the `title` and `id` aren't the same) will show the wrong name:

Before (package name)|After (package ID)
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-09-28 at 15 42 35|Simulator Screen Shot - iPhone 12 Pro - 2021-09-28 at 15 32 22

rachelmcr (commented):
I confirmed that the validation is working as expected now, even when packages have the same ID. It's great to see things simplified, too! 👍 

I just saw one issue with these changes: Since predefined package IDs aren't the same as the package title, if you use a predefined package the wrong package name appears on the customs form screen. See my comment below for more about that.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
itsmeichigo: @rachelmcr  Thanks for spotting the issue! I have updated the logic for package name in f69b6d8af04f3026dae9742ca4b29f444f7a4695, please take another look!

itsmeichigo: Update: I've switched to reuse private methods for searching custom and predefined packages in e6d9b5992005d970fc2f3b59edf5f8343aa33b1d.

rachelmcr (approved):
Looks great! :shipit:

Issue #5036, opened 2021-09-22
[Mobile Payments] [SDK 2] Handle error cases when installing updates


koke:
Right now, updates that fail to install will fail silently with the new 2.0 UI. We should address that and show an error.

Errors to address:

- [ ] Battery level is too low to do an update (see https://github.com/woocommerce/woocommerce-ios/issues/5036#issuecomment-933725244 below) - keep in mind that we have a specific design for when the update fails because of low battery
- [ ] Lost connection to reader during update
- [ ] ???

_Originally posted by @allendav in https://github.com/woocommerce/woocommerce-ios/pull/4974#discussion_r713303946_

allendav:
I end up here after connecting when an update is detected but my battery is only at 33%:

![IMG_0172](https://user-images.githubusercontent.com/1595739/135901433-11fe4114-3e8b-4cea-9773-1248fd196dee.PNG)

We should instead display a too-low-battery message.

I expect closing this issue will need to be done iteratively, starting with the most likely failure scenarios (like too-low-battery) first

koke:
Fixed in #5073 and #5109

PR #5073, opened 2021-09-28
[Mobile Payments] Show error when mandatory update fails because of low battery

koke:
Part of #5036 

This handles the case when a mandatory update fails to install because the reader has less than 50% battery.
Note that I've only been able to test this with the simulated readers, which only offers low battery failure simulation for mandatory updates. This doesn't address optional updates failing yet (they'd fail silently), or other update failures on mandatory updates (they would show up as a generic failure to connect).

When the charge % is known:

Light|Dark
-|-
Simulator Screen Shot - iPhone 12 mini - 2021-09-28 at 12 48 33|Simulator Screen Shot - iPhone 12 mini - 2021-09-28 at 12 48 36

When the charge % is unknown (@woocommerce/mobile-designers the Figma file didn't include this scenario, so I tweaked the copy a little bit):

Light|Dark
-|-
Simulator Screen Shot - iPhone 12 mini - 2021-09-28 at 12 48 05|Simulator Screen Shot - iPhone 12 mini - 2021-09-28 at 12 48 17

## To test

1. On Xcode, edit the WooCommerce scheme, and enable `-simulate-stripe-card-reader` in the Arguments tab
2. In `StripeCardReaderService.start(_:)`, after the SDK has been initialized (I add this right after `Terminal.shared.delegate = self`), add the following line: `Terminal.shared.simulatorConfiguration.availableReaderUpdate = .lowBattery`
3. Run the app
4. Go to Settings > In-Person Payments > Manage Card Reader and connect to a reader
5. After the reader connects, it should automatically attempt to update and show the new error

Note that sometimes the simulated reader reports 85% battery (even with the simulated low battery error 🙃), and sometimes it's `nil`. Follow the steps 2-3 times and you should see the UI for both cases

Update release notes:


ghost (approved):

Hardware/Hardware/CardReader/CardReaderEvent.swift
```diff
@@ -21,12 +21,6 @@ public enum CardReaderEvent {
     /// Low battery warning resolved.
     case lowBatteryResolved
 
-    /// A software update is available.
```
ctarda, line 24: Should this be part of the diff? As far as I understand, available updates are announced via the methods declared in BluetoothReaderDelegate, which seem to have been removed as well.

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -447,8 +457,6 @@ extension StripeCardReaderService: DiscoveryDelegate {
 // MARK: - ReaderDisplayDelegate.
 extension StripeCardReaderService: BluetoothReaderDelegate {
     public func reader(_ reader: Reader, didReportAvailableUpdate update: ReaderSoftwareUpdate) {
-        let softwareUpdate = CardReaderSoftwareUpdate(update: update)
```
ctarda, line 450: Are we sure this is safe to remove? How would we be notified that there is an update available?

ctarda (commented):
I left a couple of questions about how available updates are notified

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
koke: I think so. Before, reader updates were notified through the `readerEvents` publisher, but now they have a dedicated `softwareUpdateEvents` publisher. When I updated the code to handle optional updates, the previous method got unused, so I cleaned up all the related stuff.

I had doubts about removing the `CardReaderSoftwareUpdate`, but since we're not showing any details about the software update version in the UI, that's unused as well

ctarda: Ah, I see. I misread the diff, sorry. I thought we would be missing on available updates. 

> I had doubts about removing the CardReaderSoftwareUpdate, but since we're not showing any details about the software update version in the UI, that's unused as well

That's a good point. I would lean towards not removing it though, it seems like it's information that we might want to know later on, maybe attach to a Tracks event? Anyway, not a blocker. :shipit:

ctarda (approved):
Left a question about passing the software update metadata up to the app, but its not a blocker in my opinion.

:shipit:

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
koke: I thought about that, but we can always revert/restore that code if we need it

PR #5074, opened 2021-09-28
Shipping Labels: Update the print shipping label endpoint to accept multiple labels

rachelmcr:
Part of: #4714

## Description

As part of multi-package support for shipping labels, we need to support printing multiple shipping labels. This is the first step, to update the print endpoint to accept multiple labels, for example: `/v1/connect/label/print?paper_size=label&label_id_csv=1979,C1978`

There aren't any UI changes at this point, so the app should continue to work as before (printing a single shipping label).

## Changes

* Networking: Updates the print endpoint in `ShippingLabelRemote` to accept an array of shipping label IDs to print.
* Yosemite: Updates `ShippingLabelStore` and `ShippingLabelAction` to accept an array of shipping label IDs.
* Updates Networking and Yosemite unit tests and mocks to match those changes.

## Testing

2. Make sure you have at least one order with a purchased shipping label.
3. Open the order detail for the order that has a purchased shipping label.
4. Find the package section in the order detail and select "Print Shipping Label."
5. Select a paper size, if needed, and tap the Print Shipping Label button.
6. Confirm you get a print preview with the expected shipping label.

## Submitter Checklist

Update release notes:

### DIFF
--- a/Networking/Networking/Remote/ShippingLabelRemote.swift
+++ b/Networking/Networking/Remote/ShippingLabelRemote.swift
@@ -4,7 +4,7 @@ import Foundation
 public protocol ShippingLabelRemoteProtocol {
     func loadShippingLabels(siteID: Int64, orderID: Int64, completion: @escaping (Result<OrderShippingLabelListResponse, Error>) -> Void)
     func printShippingLabel(siteID: Int64,
-                            shippingLabelID: Int64,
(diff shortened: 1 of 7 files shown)


itsmeichigo (approved):
Code looks good and still works as before! 🎉

PR #5075, opened 2021-09-28
Add order number to receipt

### LINKED ISSUE #4264: [Mobile Payments] Add Order Number to Receipts
## Description
Add the Order Number to the receipt, as suggested in feedback on printed receipts

p91TBi-5bu#comment-4704

joshheald:
Resolves #4264 

## Description
Receipts did not include any way to link them back to an order, and this was identified as a useful improvement to make to them. This change adds them to the heading above the line details of the receipt.

## Testing

1. Go to the Orders tab
2. Open an unpaid "Pay on Delivery" order
3. Take card payment
4. Tap "Print Receipt"
5. Observe that the order details section now has a heading `SUMMARY: ORDER #85`

https://user-images.githubusercontent.com/2472348/135122505-c2ba123c-0163-4217-920d-fe3f5ea01292.mp4

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
@@ -10,7 +10,7 @@ extension PaymentIntent {
         self.created = intent.created
         self.amount = intent.amount
         self.currency = intent.currency
-        self.metadata = intent.metadata
+        self.metadata = intent.metadata as? [String: String]
         self.charges = intent.charges.map { .init(charge: $0) }
     }
 }
--- a/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
+++ b/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
@@ -86,7 +86,7 @@ public extension ReceiptRenderer {
                             <span class="card-icon \(parameters.cardDetails.brand.iconName)-icon"></span> - \(parameters.cardDetails.last4)
                         </p>
                     </header>
-                    <h3>\(Localization.summarySectionTitle.uppercased())</h3>
+                    <h3>\(summarySectionTitle.uppercased())</h3>
                     \(summaryTable())
                     <footer>
                         <p>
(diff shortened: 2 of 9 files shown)


Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -169,6 +169,13 @@ private extension ReceiptRenderer {
 
         return .localizedStringWithFormat(Localization.receiptFromFormat, storeName)
     }
+
+    private var summarySectionTitle: String {
+        guard let orderID = parameters.orderID else {
+            return Localization.summarySectionTitle
+        }
+        return .localizedStringWithFormat(Localization.summarySectionTitleWithOrderFormat, String(orderID))
```
koke: I'm not sure `localizedStringWithFormat` is useful here, as its main purpose is formatting numbers according to the user's locale, but we're passing the `orderID` as a `String`, and I'm not sure we wanted that formatted with a thousands separator anyway. I think `String(format:...)` would be better here.

Screen Shot 2021-09-30 at 07 53 32

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
```diff
@@ -10,7 +10,7 @@ extension PaymentIntent {
         self.created = intent.created
         self.amount = intent.amount
         self.currency = intent.currency
-        self.metadata = intent.metadata
+        self.metadata = intent.metadata as? [String: String]
```
koke, line 13: I didn't realize there was a `metadata` type coming from StripeTerminal already that's defined as `[AnyHashable: Any]`. I still think it's reasonable to expect it to only contain strings, but I'm a bit more worried than before, since I think this cast would fail for the whole dictionary if a single key or value isn't a string?

koke (approved):
Works great 👍🏽 I left a note about something that could potentially be an issue, but I think it should be fine as it is? 🤷🏽‍♂️

Hardware/Hardware/CardReader/StripeCardReader/PaymentIntent+Stripe.swift
joshheald: Yeah, agreed, and we're relying on the Stripe SDK maintaining that behaviour. Consequences of that seem fairly small at the moment, but having worked on another receipt field I might come back to this and rework it. For now I think it's fine, so I'm going to merge.

PR #5079, opened 2021-09-29
Shipping Labels: Update package details description for multi-package

itsmeichigo:
Part of #4599

# Description
This PR updates the details displayed on the Package Details row of the Shipping Label form for the multi-package case.

# Changes
- Adds new method for `WeightFormatter` to accept double as the weight type for formatting.
- Updates `ShippingLabelFormViewModel` to display appropriate description of package depending on the number of packages selected.

# Demo
https://user-images.githubusercontent.com/5533851/135201991-8cbc8db8-0dba-4b5d-9c9c-0a581c95e018.mp4

# Testing steps
2. Create an order with more than one item / or there should be an order item whose quantity is larger one.
4. Proceed to configure Package Details. Tap Done to proceed with every item in the same one package. Notice that the correct name and total weight of the package is then displayed on Package Details row.
5. Select Package Details again, this time move items to different packages. Tap Done, notice that the Package Details row now displays the correct number of items and packages as well as total weight.
6. Turn off multi-package feature flag and repeat step 4, make sure that everything still works properly.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -293,7 +293,9 @@ final class ShippingLabelFormViewModel {
 
         if selectedPackagesDetails.count == 1,
            let selectedPackage = selectedPackagesDetails.first {
-            packageDescription = searchCustomPackage(id: selectedPackage.packageID)?.title ?? searchPredefinedPackage(id: selectedPackage.packageID)?.title ?? ""
+            let customPackageTitle = searchCustomPackage(id: selectedPackage.packageID)?.title
+            let predefinedPackageTitle = searchPredefinedPackage(id: selectedPackage.packageID)?.title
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -283,18 +283,32 @@ final class ShippingLabelFormViewModel {
...
     ///
     func getPackageDetailsBody() -> String {
-        guard let packagesResponse = packagesResponse,
-              let selectedPackage = selectedPackagesDetails.first else {
+        guard selectedPackagesDetails.isNotEmpty else {
             return Localization.packageDetailsPlaceholder
         }
 
-        let packageTitle = searchCustomPackage(id: selectedPackage.packageID)?.title ?? searchPredefinedPackage(id: selectedPackage.packageID)?.title ?? ""
+        let packageTitle: String
```
rachelmcr: nit: Since this string is the package title when there's 1 package, or a count of the items and packages when there are multiple packages, maybe it could be named something more generic e.g. `packageDescription` instead of `packageTitle`.

rachelmcr (approved):
LGTM! I left a small nit/suggestion below about naming, but it isn't critical. 👍

PR #5080, opened 2021-09-29
Move all screens to UITestsFoundation

mokagio:
To make the review of the following work, in which I'll convert the `BaseScreen`s to be `ScreenObject` subclasses, I'm moving all of them in the UITestFoundation target in a decidated PR.

Notice that I've done the move in one "big bang" commit. In WordPress, I did it in multiple commits, moving isolated files first. The idea was to learn a bit about the structure of the test suite. Unfortunately, it turned out to bee a very tedious process that took way to long for what I got out of it. So in this instance I took a more pragmatic approach and did it all in one go. 😅 

As far as testing go, this PR doesn't change any code. You can see that CI runs the UI tests successfully, which is enough validation these changes don't break anything.

- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary. – **N.A.**

---

Also notice this is based on top of #5042, I'll wait for that one to merge first, but I'd still appreciate a review.

### DIFF
--- a/WooCommerce/UITestsFoundation/Screens/Login/GetStartedScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/Login/GetStartedScreen.swift
@@ -1,4 +1,3 @@
-import UITestsFoundation
 import XCTest
 
 private struct ElementStringIDs {
@@ -7,7 +6,7 @@ private struct ElementStringIDs {
     static let continueButton = "Get Started Email Continue Button"
 }
 
-final class GetStartedScreen: BaseScreen {
+public final class GetStartedScreen: BaseScreen {
     private let navBar: XCUIElement
     private let emailTextField: XCUIElement
     private let continueButton: XCUIElement
@@ -21,7 +20,7 @@ final class GetStartedScreen: BaseScreen {
         super.init(element: emailTextField)
     }
 
-    func proceedWith(email: String) -> PasswordScreen {
+    public func proceedWith(email: String) -> PasswordScreen {
         emailTextField.tap()
         emailTextField.typeText(email)
         continueButton.tap()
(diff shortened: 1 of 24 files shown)


WooCommerce/UITestsFoundation/Screens/Login/GetStartedScreen.swift
```diff
@@ -1,4 +1,3 @@
-import UITestsFoundation
```
mokagio, line 1: This import is no longer needed because the file is now part of the framework.

Conversely, some UI test target files will now have to `import UITestsFoundation` in order to access the screens they require.

WooCommerce/UITestsFoundation/Screens/Login/GetStartedScreen.swift
```diff
@@ -7,7 +6,7 @@ private struct ElementStringIDs {
     static let continueButton = "Get Started Email Continue Button"
 }
 
-final class GetStartedScreen: BaseScreen {
+public final class GetStartedScreen: BaseScreen {
```
mokagio, line 9: Because these types are now in a dedicated framework, we need to make them `public` to be used in the UI tests.

Notice that not all of the methods and screens had to become `public`, though. That's because only a portion of the API surface is used directly in the tests, the remainder is either unused (which I'll address as I convert them to `ScreenObject` subclasses, or only used internally.

pachlava (approved):
This was a fast review, because the changes follow the same pattern, and I only tested that UI tests also pass locally, besides passing on CI. Thank you for the extra details, and for this work in general, @mokagio !

mokagio:
Thank you @pachlava 🙇‍♂️ 

If you have the time, I'd appreciate a look at the PR on which this one is based, too: https://github.com/woocommerce/woocommerce-ios/pull/5042.

PR #5086, opened 2021-09-29
Order Edit: Better email handling

Ecarrion:
# Why

Address fields sent as `nil` is not a good way to represent empty values. Because `nil` won't be encoded in the network request. Making those fields remain unaltered.

Sending empty fields as `""` works well for all fields except for `email` because Core has a validation that requires a valid email address and an empty string does not pass that validation.

Also, we just realized that the shipping address form does not contain the email field.

For those reasons, this PR works around that limitation by:

- Hiding the email row in the shipping address form.
- Setting the email to `nil` before sending it to Core when the email is empty.

Note: This means that we can't clear an email address at the moment. Core appears to be using some kind of `admin-ajax` function to clear the email.

# How

- Add `GeneratedCopiable` to `Address`
- Update ViewModel to provide a `showEmailField` variable so the view can be rendered appropriately.
- Update ViewModel to not `nil` the rest of the fields when they are empty.

# Demo

https://user-images.githubusercontent.com/562080/135352037-8b24200e-3e50-4fa9-be92-870bde2a3a6d.mov

# Testing Steps

- Open the shipping address form and see that there is no email field.
- Remove content for phone, company, or last name.
- Tap on the "use as shipping address" row.
- Tap on the Done button.
- Corroborate that the correct changes were performed in the billing address form

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
@@ -259,15 +259,15 @@ extension EditAddressFormViewModel {
         func toAddress(country: Yosemite.Country?, state: Yosemite.StateOfACountry?) -> Yosemite.Address {
             Address(firstName: firstName,
                     lastName: lastName,
-                    company: company.isEmpty ? nil : company,
+                    company: company,
                     address1: address1,
-                    address2: address2.isEmpty ? nil : address2,
+                    address2: address2,
                     city: city,
                     state: state?.code ?? self.state,
                     postcode: postcode,
                     country: country?.code ?? self.country,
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressFormViewModel.swift
```diff
@@ -382,3 +393,15 @@ private extension EditAddressFormViewModel {
         .eraseToAnyPublisher()
     }
 }
+
+private extension Address {
+    /// Sets the email value to `nil` when it is empty..
```
ealeksandrov:
```suggestion
    /// Sets the email value to `nil` when it is empty.
```

ealeksandrov (approved):
LGTM :shipit:

Issue #5077, opened 2021-09-28
[Mobile Payments] [SDK 2.0] Bump minimum wcpay plugin version to 3.1.0 for Locations endpoint


allendav:
Why? Because connecting to readers with SDK 2.x requires Locations, and the [Locations endpoint was first available in woocommerce-payments plugin version 3.0.0](https://github.com/Automattic/woocommerce-payments/pull/2830)

Updated Oct 7: A nasty bug related to blank line2 addresses (which is a common address format) was fixed in 3.1, so 3.1 should be the new minimum. (3.1 was released 2 days ago)

<del>First, display an in-app notice (where?) for a release or two if a merchant using in-person payments is running a version < 3.0.0 - that notice should warn them that they need to update the plugin or they will lose the ability to collect payments when the app is updated to SDK 2.x</del>

<del>Then,</del> in the release that first includes SDK 2.x, update our on-boarding checks to check for woocommerce-payments **3.1** or newer (right now, `CardPresentPaymentsOnboardingUseCase` has it set to `2.5`)

cc @ctarda @koke and we should do something similar for Android

PR #5087, opened 2021-09-29
[Mobile Payments] [SDK 2] Add fetching actual location

### LINKED ISSUE #5078: [Mobile Payments] [SDK 2.0] Fetch Stripe Location ID for store base location for reader connection
Right now `StripeCardReaderService` `getBluetoothConfiguration` calls `CardPresentPaymentStore` `fetchDefaultLocationID` which calls `remote.loadDefaultReaderLocation` to returns a hard coded Location ID.

As of WCPay plugin version 3.0.0, an endpoint is available to return a Location ID for the store's current base location (in wp-admin > WooCommerce > Settings > General)

The app should be updated to request, map and have unit tests to hit that endpoint. I would not look into caching the location ID provided by the endpoint at this time, until we have a solid understanding of how we'd like to manage locations for highly mobile hardware like our Bluetooth card readers.

Important server side issue to be aware of: pc2DNy-vL-p2

Related but separate effort: https://github.com/woocommerce/woocommerce-ios/issues/5077

Related Epic: https://github.com/woocommerce/woocommerce-ios/issues/4497

fyi @koke - @ctarda asked me to tackle this one

allendav:
Closes #5078 

Changes:
- Adds WCPayReaderLocation for WCPayRemote to populate through WCPayReaderLocationMapper
- Adds WCPayReaderLocation+ReaderLocation.swift to translate WCPayReaderLocation into ReaderLocation
- Adds unit tests
- Removes hardcoded location value

Questions:
- Do we need to do something special for simulators?

Not included:
- Presenting specific failure message to user: https://github.com/woocommerce/woocommerce-ios/issues/5088

To test:
- Install WooCommerce Payments 3.0 or newer: https://github.com/woocommerce/woocommerce-ios/issues/5077
- Make sure your store is gated into the beta: (wcpay server 1199)
- Make sure you have both address line 1 AND address line 2 set up on your store : https://github.com/Automattic/woocommerce-payments/issues/3012
- Ensure you can connect to a reader
- For bonus points, set a breakpoint in`fetchDefaultLocationID` and make sure that the location ID provided is consistent for repeated connections to a reader. Then change the street address for the business in WooCommerce settings and connect again - the location ID should change.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -288,13 +288,11 @@ extension StripeCardReaderService: CardReaderService {
             self.readerLocationProvider?.fetchDefaultLocationID { (locationId, error) in
                 if let error = error {
                     let underlyingError = UnderlyingError(with: error)
-                    promise(.failure(CardReaderServiceError.connection(underlyingError: underlyingError)))
-                    return
+                    return promise(.failure(CardReaderServiceError.connection(underlyingError: underlyingError)))
                 }
 
                 if let locationId = locationId {
-                    promise(.success(BluetoothConnectionConfiguration(locationId: locationId)))
-                    return
+                    return promise(.success(BluetoothConnectionConfiguration(locationId: locationId)))
                 }
 
                 promise(.failure(CardReaderServiceError.connection()))
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -283,19 +283,21 @@ extension StripeCardReaderService: CardReaderService {
                 return
             }
 
-            if let locationId = reader.locationId {
-                promise(.success(BluetoothConnectionConfiguration(locationId: locationId)))
-            } else {
-                self.readerLocationProvider?.fetchDefaultLocationID { (locationId, error) in
-                    if let error = error {
-                        let underlyingError = UnderlyingError(with: error)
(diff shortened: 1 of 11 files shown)


koke (approved):
That seems to work, but I got confused by this:

> For bonus points, set a breakpoint in `fetchDefaultLocationID` and make sure that the location ID provided is consistent for repeated connections to a reader

I tried that, but I only saw `fetchDefaultLocationID` being called the first time. After that, reconnecting or relaunching the app seems to always use the location reported from the SDK.

allendav:
> After that, reconnecting or relaunching the app seems to always use the location reported from the SDK.

That's odd. I'll double check

allendav:
@koke wrote:

> I tried that, but I only saw fetchDefaultLocationID being called the first time. After that, reconnecting or relaunching the app seems to always use the location reported from the SDK.

I see that now. I missed that during my testing. Its actually a bug in the SDK 2 branch - I've written it up in #5116 

I'll remove using the locationId (if any) supplied by the SDK and double check this PR. If that proves true, I'll address #5116 here too - probably just ignoring the location Id for now, but we can keep #5116 open if desired to cache the value for some time for performance sake

allendav:
@koke - fixed cached location in https://github.com/woocommerce/woocommerce-ios/pull/5087/commits/ed27e02c5a5f508d2e4482608792e49936360e56 - we can optimize against https://github.com/woocommerce/woocommerce-ios/issues/5116

allendav:
@koke wrote:

> A little syntactic sugar that I like in cases like this: because this function returns Void, and promise returns Void as well, you can fold these two lines into return promise(...) and it looks a bit more compact. Obviously not a blocker, but I wanted to share that tip

Changed in https://github.com/woocommerce/woocommerce-ios/pull/5087/commits/072bd50edacfeaefd07a13a1952be353cb15c336

PR #5089, opened 2021-09-29
Order Editing: Edit Address form analytics

### LINKED ISSUE #4865: Order Editing: Tracks events

Ecarrion:
closes #4865

# Why

Before releasing the feature to the public we need to add the designed analytics to these screens.

# How

- Tracks `started` when the flows are being presented
- Tracks `completed` when the `updateOrder` action completes successfully.
- Tracks `failed` when the `updateOrder` action fails.
- Tracks `canceled` when either the user taps on the close button or when the user dismisses the flow via a drag gesture.

# Testing

- Launch the app, edit some order addresses, and see that the appropriate events are being printed on the console.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
@@ -146,8 +146,10 @@ struct EditAddressForm: View {
                                          text: $viewModel.fields.email,
                                          symbol: nil,
                                          keyboardType: .emailAddress)
+                        .renderedIf(viewModel.showEmailField)
                     Divider()
                         .padding(.leading, Constants.dividerPadding)
+                        .renderedIf(viewModel.showEmailField)
                     TitleAndTextFieldRow(title: Localization.phoneField,
                                          placeholder: "",
                                          text: $viewModel.fields.phone,
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
@@ -90,7 +90,12 @@ extension EditAddressHostingController: UIAdaptivePresentationControllerDelegate
     func presentationControllerDidAttemptToDismiss(_ presentationController: UIPresentationController) {
         UIAlertController.presentDiscardChangesActionSheet(viewController: self) { [weak self] in
             self?.dismiss(animated: true)
+            self?.rootView.viewModel.userDidCancelFlow()
         }
+    }
+
+    func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
+        rootView.viewModel.userDidCancelFlow()
     }
 }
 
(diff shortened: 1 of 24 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -676,6 +676,8 @@ private extension OrderDetailsViewController {
         let editAddressViewController = EditAddressHostingController(viewModel: viewModel)
         let navigationController = WooNavigationController(rootViewController: editAddressViewController)
         present(navigationController, animated: true, completion: nil)
+
+        ServiceLocator.analytics.track(event: WooAnalyticsEvent.OrderDetailsEdit.orderDetailEditFlowStarted(subject: .shippingAddress))
```
ealeksandrov: Can't we move start events to EditAddressFormViewModel `init` or `onLoadTrigger`?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
```diff
@@ -90,8 +90,13 @@ extension EditAddressHostingController: UIAdaptivePresentationControllerDelegate
     func presentationControllerDidAttemptToDismiss(_ presentationController: UIPresentationController) {
         UIAlertController.presentDiscardChangesActionSheet(viewController: self) { [weak self] in
             self?.dismiss(animated: true)
+            self?.rootView.viewModel.userDidCancelFlow()
         }
     }
+
+    func presentationControllerDidDismiss(_ presentationController: UIPresentationController) {
+        rootView.viewModel.userDidCancelFlow()
+    }
```
ealeksandrov, line 99: Q: since we have `DidAttemptToDismiss` handled, is there any case where `DidDismiss` called?

ealeksandrov (approved):
:shipit: LGTM, just a couple of minor questions.

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
Ecarrion: I normally don't like to add analytics code there because it is an unwanted side effect that people do not expect.

EG:
- I could create the VC and not present it yet, but if the analytics code is on `init` the event would be fired.


But now that I think about it, since this `onLoadTrigger()` does not follow `UIKit`, there is no chance for this to be double executed if the VC gets evicted from memory.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
Ecarrion: When `presentationControllerShouldDismiss` returns `true` then `presentationControllerDidDismiss` is called.

ealeksandrov: Right, I forgot about that case 😅

PR #5090, opened 2021-09-30
Show Variations screen on tapping "Add price to your variations..." warning row

### LINKED ISSUE #4849: Products: make the “Add price to your variations to make them visible…” banner clickable
> Can we make the “Add price to your variations to make them visible…” banner clickable? (same target as the “Variations” row, opening up the Variations List)

IMG_4955
<sup>Tested with WCiOS 7.3.0.0 TestFlight beta on iPhone 11 iOS 15.0 Beta 5 logged in as testuser01121 with hilarious-sea.jurassic.ninja selected and running WordPress 5.8, Jetpack 10.0, WooCommerce 5.6.0.</sup>

selanthiraiyan:
Closes #4849

## Description
“Add price to your variations to make them visible…” warning in `ProductFormViewController` was not tappable and an issue was reported to make it tappable and also launch the Variations screen on tap. 
This change includes the code to make the warning row actionable and also launch the `ProductVariationsViewController` on tap.

## Testing
You should add a Product with a variant, which doesn't have a price set. (Please refer to the attached recording.)

1. Go to the `Products` tab.
2. Tap on the product which has a variant and no price set.
3. In the tableview, you should see the "Add price to your variations to make them visible on your store" warning, just below the `Variations` row.
4. Tap on "Add price to your variations to make them visible on your store" warning and observe that `Variations` screen is shown.
5. Navigating back and tapping on `Variations` row should also end up showing the `Variations` screen.

https://user-images.githubusercontent.com/524475/135368677-cc101da8-4c29-43dc-9270-3ba702a9d2c2.mov

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -359,23 +359,22 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
...
                 }
-                let variationsViewModel = ProductVariationsViewModel(formType: viewModel.formType)
-                let variationsViewController = ProductVariationsViewController(initialViewController: self,
-                                                                               viewModel: variationsViewModel,
-                                                                               product: originalProduct.product)
-                variationsViewController.onProductUpdate = { [viewModel] updatedProduct in
-                    viewModel.updateProductVariations(from: updatedProduct)
+                showVariations()
+            case .noPriceWarning(let viewModel):
+                ServiceLocator.analytics.track(.productDetailViewVariationsTapped)
```
selanthiraiyan: Should we use a different event to track this?

joshbetz:
Thanks for the PR @selanthiraiyan! We're taking a look.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
Ecarrion: I think we can leave this, and treat it as it was a big cell.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -359,23 +359,22 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
...
+                guard viewModel.isActionable else {
+                    return
```
Ecarrion, line 368: I think this guard should go before the analytics, or we will be login taps on un-tappable rows, like on the product variation view.


https://user-images.githubusercontent.com/562080/135634944-e1d65664-8fba-472f-8288-b2bff10b4f2e.mov

Ecarrion (changes requested):
Hi @selanthiraiyan,

Thanks for giving this a go! 🎉 

I just left you a small suggestion and we should be good to go!

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
selanthiraiyan: Thanks for reviewing @Ecarrion ! ❤️ 
You are right. I left it this way, because we were previously logging the event before the guard. I will fix it.

Ecarrion (changes requested):
@selanthiraiyan I'm seeing that the analytics event is always fired, I guess that is because you are setting `WarningViewModel.actionable` to be always true.

We should modify that to be false for product variations.

selanthiraiyan:
> @selanthiraiyan I'm seeing that the analytics event is always fired, I guess that is because you are setting `WarningViewModel.actionable` to be always true.
> 
> We should modify that to be false for product variations.

I understand and thanks for your suggestion on changing `isActionable` to false for product variations. I have pushed those changes.

Ecarrion (approved):
This looks great now! @selanthiraiyan 
I'll merge it as soon as CI finishes! 🎉

PR #5091, opened 2021-09-30
Convert UI tests screens to be a `ScreenObject` subclasses

mokagio:
Converts all the `BaseScreen` subclass that are not part of the login and signup flow to be `ScreenObject` subclasses instead.

For more information on the structure of a `ScreenObject` subclass, see [this WordPress iOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/17221) description and comments.

As you can see, I run the UI tests in CI and [they pass](https://app.circleci.com/pipelines/github/woocommerce/woocommerce-ios/14673/workflows/b18119bb-7322-464f-a2d6-7bc525107d60). Feel free to run them locally, too, if you have time to spare.

The reason I didn't work on the login and signup screens is that the plan for the upcoming weeks is to extract what's in WordPress into a dedicated test helper framework, and use that instead. Working on them here would have been redundant.

---

### DIFF
--- a/WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
@@ -4,7 +4,8 @@ import XCTest
 public final class MyStoreScreen: ScreenObject {
 
     public let tabBar = TabNavComponent()
-    public let periodStatsTable = PeriodStatsTable()
+    // TODO: Remove force `try` once `ScreenObject` migration is completed
+    public let periodStatsTable = try! PeriodStatsTable()
 
     private let settingsButtonGetter: (XCUIApplication) -> XCUIElement = {
         $0.buttons["dashboard-settings-button"]
--- a/WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
@@ -3,7 +3,8 @@ import XCTest
 
 public final class MyStoreScreen: ScreenObject {
 
-    public let tabBar = TabNavComponent()
+    // TODO: Remove force `try` once `ScreenObject` migration is completed
+    public let tabBar = try! TabNavComponent()
     // TODO: Remove force `try` once `ScreenObject` migration is completed
     public let periodStatsTable = try! PeriodStatsTable()
 
--- a/WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
@@ -35,8 +35,8 @@ public final class MyStoreScreen: ScreenObject {
     }
 
     @discardableResult
-    public func openSettingsPane() -> SettingsScreen {
+    public func openSettingsPane() throws -> SettingsScreen {
         settingsButton.tap()
-        return SettingsScreen()
+        return try SettingsScreen()
     }
 }
--- a/WooCommerce/UITestsFoundation/Screens/MyStore/PeriodStatsTable.swift
+++ b/WooCommerce/UITestsFoundation/Screens/MyStore/PeriodStatsTable.swift
@@ -1,28 +1,29 @@
+import ScreenObject
 import XCTest
 
-public final class PeriodStatsTable: BaseScreen {
+public final class PeriodStatsTable: ScreenObject {
 
-    struct ElementStringIDs {
-        static let daysTab = "period-data-today-tab"
-        static let weeksTab = "period-data-thisWeek-tab"
-        static let monthsTab = "period-data-thisMonth-tab"
-        static let yearsTab = "period-data-thisYear-tab"
+    private let daysTabGetter: (XCUIApplication) -> XCUIElement = {
+        $0.cells["period-data-today-tab"]
     }
 
+    private let weeksTabGetter: (XCUIApplication) -> XCUIElement = {
+        $0.cells["period-data-thisWeek-tab"]
+    }
 
-    private let daysTab = XCUIApplication().cells[ElementStringIDs.daysTab]
-    private let weeksTab = XCUIApplication().cells[ElementStringIDs.weeksTab]
-    private let monthsTab = XCUIApplication().cells[ElementStringIDs.monthsTab]
-    private let yearsTab = XCUIApplication().cells[ElementStringIDs.yearsTab]
+    private let yearsTabGetter: (XCUIApplication) -> XCUIElement = {
+        $0.cells["period-data-thisYear-tab"]
+    }
 
-    init() {
+    private var daysTab: XCUIElement { daysTabGetter(app) }
+    private var weeksTab: XCUIElement { weeksTabGetter(app) }
+    private var yearsTab: XCUIElement { yearsTabGetter(app) }
 
-        super.init(element: daysTab)
-
-        XCTAssert(daysTab.waitForExistence(timeout: 3))
-        XCTAssert(weeksTab.waitForExistence(timeout: 3))
-        XCTAssert(monthsTab.waitForExistence(timeout: 3))
-        XCTAssert(yearsTab.waitForExistence(timeout: 3))
+    init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(
+            expectedElementGetters: [daysTabGetter, weeksTabGetter, yearsTabGetter],
+            app: app
+        )
     }
 
     @discardableResult
--- a/WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
@@ -20,8 +20,8 @@ public final class OrderSearchScreen: BaseScreen {
     }
 
     @discardableResult
-    public func cancel() -> OrdersScreen {
+    public func cancel() throws -> OrdersScreen {
         cancelButton.tap()
-        return OrdersScreen()
+        return try OrdersScreen()
     }
 }
--- a/WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
@@ -1,22 +1,28 @@
+import ScreenObject
 import XCTest
 
-public final class OrderSearchScreen: BaseScreen {
+public final class OrderSearchScreen: ScreenObject {
 
     struct ElementStringIDs {
         static let searchField = "order-search-screen-search-field"
         static let cancelButton = "order-search-screen-cancel-button"
     }
 
-    private let searchField: XCUIElement
-    private let cancelButton: XCUIElement
+    private let cancelButtonGetter: (XCUIApplication) -> XCUIElement = {
+        $0.buttons["order-search-screen-cancel-button"]
+    }
 
-    init() {
-        searchField = XCUIApplication().otherElements[ElementStringIDs.searchField]
-        cancelButton = XCUIApplication().buttons[ElementStringIDs.cancelButton]
-        super.init(element: cancelButton)
+    private var cancelButton: XCUIElement { cancelButtonGetter(app) }
 
-        XCTAssert(searchField.waitForExistence(timeout: 3))
-        XCTAssert(cancelButton.waitForExistence(timeout: 3))
+    init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(
+            expectedElementGetters: [
+                cancelButtonGetter,
+                // swiftlint:disable:next opening_braces
+                { $0.otherElements["order-search-screen-search-field"] }
+            ],
+            app: app
+        )
     }
 
     @discardableResult
--- a/WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
@@ -3,11 +3,6 @@ import XCTest
 
 public final class OrderSearchScreen: ScreenObject {
 
-    struct ElementStringIDs {
-        static let searchField = "order-search-screen-search-field"
-        static let cancelButton = "order-search-screen-cancel-button"
-    }
-
     private let cancelButtonGetter: (XCUIApplication) -> XCUIElement = {
         $0.buttons["order-search-screen-cancel-button"]
     }
--- a/WooCommerce/UITestsFoundation/Screens/Orders/OrdersScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/Orders/OrdersScreen.swift
@@ -3,7 +3,8 @@ import XCTest
 
 public final class OrdersScreen: ScreenObject {
 
-    public let tabBar = TabNavComponent()
+    // TODO: Remove force `try` once `ScreenObject` migration is completed
+    public let tabBar = try! TabNavComponent()
 
     private let searchButtonGetter: (XCUIApplication) -> XCUIElement = {
         $0.buttons["order-search-button"]
--- a/WooCommerce/UITestsFoundation/Screens/Orders/OrdersScreen.swift
+++ b/WooCommerce/UITestsFoundation/Screens/Orders/OrdersScreen.swift
@@ -1,31 +1,35 @@
+import ScreenObject
 import XCTest
 
-public final class OrdersScreen: BaseScreen {
-
-    struct ElementStringIDs {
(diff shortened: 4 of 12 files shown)


WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
```diff
@@ -3,8 +3,10 @@ import XCTest
 
 public final class MyStoreScreen: ScreenObject {
 
-    public let tabBar = TabNavComponent()
-    public let periodStatsTable = PeriodStatsTable()
+    // TODO: Remove force `try` once `ScreenObject` migration is completed
+    public let tabBar = try! TabNavComponent()
+    // TODO: Remove force `try` once `ScreenObject` migration is completed
+    public let periodStatsTable = try! PeriodStatsTable()
```
mokagio, line 9: I'm not sure how to address these `TODO`s yet.

Part of me wants to change the screens architecture. I understand the convenience of accessing the tab bar from a screen, but that is the wrong information flow. A screen embedded in a tab bar should not know about the tab bar, it should be the other way around only. As such, I would update the tests calling this by making them instantiate `TabNavComponent` directly.

What do you think?

WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
```diff
@@ -1,32 +1,34 @@
...
-        searchField = XCUIApplication().otherElements[ElementStringIDs.searchField]
-        cancelButton = XCUIApplication().buttons[ElementStringIDs.cancelButton]
-        super.init(element: cancelButton)
+    private var cancelButton: XCUIElement { cancelButtonGetter(app) }
+
+    init(app: XCUIApplication = XCUIApplication()) throws {
+        // Defined here to avoid a SwiftLint style violation if we did it inline in the array below
+        let searchFieldGetter: (XCUIApplication) -> XCUIElement = {
+            $0.otherElements["order-search-screen-search-field"]
+        }
```
mokagio: Note to self: replace with `swiftlint:disable:next` comment.

WooCommerce/UITestsFoundation/Screens/Orders/OrdersScreen.swift
```diff
@@ -1,37 +1,43 @@
...
-    private let searchButton: XCUIElement
-    private let filterButton: XCUIElement
+    private var searchButton: XCUIElement { searchButtonGetter(app) }
 
+    // TODO: There's only one usage of this and it can be replaced with a screen instantiation
     static var isVisible: Bool {
-        let searchButton = XCUIApplication().buttons[ElementStringIDs.searchButton]
-        return searchButton.exists && searchButton.isHittable
+        (try? OrdersScreen().isLoaded) ?? false
     }
```
mokagio, line 18: I will address this in a followup PR, to avoid going out of scope on this one, which is already quite dense.

WooCommerce/UITestsFoundation/Screens/Orders/OrdersScreen.swift
```diff
@@ -1,37 +1,43 @@
...
 
-    init() {
-        searchButton = XCUIApplication().buttons[ElementStringIDs.searchButton]
-        filterButton = XCUIApplication().buttons[ElementStringIDs.filterButton]
+    init(app: XCUIApplication = XCUIApplication()) throws {
+        // Defined here to avoid a SwiftLint style violation if we did it inline in the array below
+        let filterButtonGetter: (XCUIApplication) -> XCUIElement = {
+            $0.buttons["order-filter-button"]
+        }
```
mokagio: Note to self: replace with `swiftlint:disable:next` comment.

WooCommerce/UITestsFoundation/Screens/Settings/BetaFeaturesScreen.swift
```diff
@@ -1,21 +1,24 @@
+import ScreenObject
 import XCTest
 
-class BetaFeaturesScreen: BaseScreen {
+// This screen is currently unused. Given the purpose of UITestsFoundation is to provide developers
+// with an easy to use API to write tests, I'm going to keep the screen in the framework anyway, so
+// it'll be already available if someone needs to work with it or on it in the future.
```
mokagio, line 6: How do you feel about this?

Should I remove it? Or is it worth keeping just in case?

WooCommerce/UITestsFoundation/Screens/Settings/BetaFeaturesScreen.swift
```diff
@@ -1,21 +1,24 @@
...
+class BetaFeaturesScreen: ScreenObject {
 
     struct ElementStringIDs {
         static let enableProductsButton = "beta-features-products-cell"
     }
 
-    private let enableProductsButton = XCUIApplication().cells[ElementStringIDs.enableProductsButton]
+    // `expectedElement` is a `ScreenObject` utility to get the first element for the
+    // `expectedElementGetters` list.
```
mokagio: Do you think this comment is useful in the codebase? Or is it safe to assume a reader curious about `expectedElement` will be better served by the header doc (Alt-click):

image

_That reminds me, I should update the documentation to mention that, when multiple elements are given, it picks the first. Or, even better, take a page from the Swift API renames and call this `firstExpectedElement`.

WooCommerce/UITestsFoundation/Screens/TabNavComponent.swift
```diff
@@ -39,38 +46,40 @@ public final class TabNavComponent: BaseScreen {
...
-        return XCUIApplication().buttons[ElementStringIDs.myStoreTabBarItem].exists
+        (try? TabNavComponent().isLoaded) ?? false
     }
 
-    static func isVisible() -> Bool {
-        return XCUIApplication().buttons[ElementStringIDs.myStoreTabBarItem].isHittable
+    // TODO: This paradigm is used enough around the test suits that it would be worth extracting to `ScreenObject`.
+   static func isVisible() -> Bool {
+        guard let tabNavComponent = try? TabNavComponent() else { return false }
+        return tabNavComponent.isLoaded && tabNavComponent.expectedElement.isHittable
```
mokagio, line 83: I'll address this and the other `ScreenObject` refinements in dedicated PRs later on.

WooCommerce/UITestsFoundation/Screens/MyStore/PeriodStatsTable.swift
```diff
@@ -1,28 +1,29 @@
...
 
-public final class PeriodStatsTable: BaseScreen {
+public final class PeriodStatsTable: ScreenObject {
 
-    struct ElementStringIDs {
-        static let daysTab = "period-data-today-tab"
-        static let weeksTab = "period-data-thisWeek-tab"
-        static let monthsTab = "period-data-thisMonth-tab"
-        static let yearsTab = "period-data-thisYear-tab"
+    private let daysTabGetter: (XCUIApplication) -> XCUIElement = {
```
jkmassel, line 6: WDYT about something like:

```swift
typealias ElementGetter = (XCUIApplication) -> XCUIElement
```

to make these types a bit more concise?

WooCommerce/UITestsFoundation/Screens/MyStore/PeriodStatsTable.swift
```diff
@@ -1,28 +1,29 @@
...
-    private let monthsTab = XCUIApplication().cells[ElementStringIDs.monthsTab]
-    private let yearsTab = XCUIApplication().cells[ElementStringIDs.yearsTab]
-
-    init() {
+    private let yearsTabGetter: (XCUIApplication) -> XCUIElement = {
+        $0.cells["period-data-thisYear-tab"]
+    }
 
-        super.init(element: daysTab)
+    private var daysTab: XCUIElement { daysTabGetter(app) }
```
jkmassel, line 18: Not sure I'm 100% sold on the separation of getter vs element?

Could we use `self.app.cells["foo"]` to look this up instead of a separate getter block?

IIRC we discussed something about property wrappers too? Is this an option?

WooCommerce/UITestsFoundation/Screens/MyStore/PeriodStatsTable.swift
mokagio: I've been debating that. I'm glad you suggested it, it's a signal I should implement it. I'll do it a the `ScreenObject` level 👍

WooCommerce/UITestsFoundation/Screens/MyStore/PeriodStatsTable.swift
mokagio: You already raised concerns about boilerplate [here](https://github.com/woocommerce/woocommerce-ios/pull/4392#discussion_r686201816) and mentioned property wrappers [here](https://github.com/woocommerce/woocommerce-ios/pull/4392#pullrequestreview-726685856).

I had totally forgotten about property wrappers 😅 As for the boilerplate, I misinterpreted it as being related to a dirty diff more that to the approach itself, I should have followed up explicitly after removing the unused code I thought was the culprit.

> Could we use `self.app.cells["foo"]` to look this up instead of a separate getter block?

Sure can. In that case we would end up with duplicated calls to that query: one for the code we pass to the screen init to do the verification and one to get the element. E.g.:

```swift
var label: XCUIElement { app.staticTexts["foo"] }

init() throws {
  try super.init(elementsGetter: [ { $0.staticTexts["foo"] } ])
}
```

I prefer the current version, because the duplication we have here is one both the compiler and code completion can help us keep in sync.

I haven't tried, but I don't think we could use the computed variable as a parameter for the `super.init` (under the assumption we switched from `ElementGetter` to straight `XCUIElement`) because for the computation to run, `self` should already be instantiated.

mokagio: I didn't mention this in my earlier comment, but I do really like the property wrapper idea. I'll look into it and see if there's something that can be done, e.g. to define the getter and automatically give is a `get()` method that calls `self.app` internally.

WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
pachlava: I guess it depends on what we call a screen. From the user POV, it's all just a bunch of elements located on one screen. From the developer POV, I suppose these are different components in the app hierarchy.

I would say that having tests instantiate tab bar components separately will mean better depiction of app structure and having the framework more robust.

WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
```diff
@@ -1,27 +1,33 @@
...
+    private var cancelButton: XCUIElement { cancelButtonGetter(app) }
 
-        XCTAssert(searchField.waitForExistence(timeout: 3))
-        XCTAssert(cancelButton.waitForExistence(timeout: 3))
+    init(app: XCUIApplication = XCUIApplication()) throws {
+        try super.init(
+            expectedElementGetters: [
+                cancelButtonGetter,
+                // swiftlint:disable:next opening_braces
+                { $0.otherElements["order-search-screen-search-field"] }
```
pachlava, line 17: I'm curious why the getter is sometimes declared right inside the array, and in other cases as a constant of a function type. It looks like we lose the option of using the object if it's declared this way?

WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
```diff
@@ -1,27 +1,33 @@
+import ScreenObject
 import XCTest
 
-public final class OrderSearchScreen: BaseScreen {
+public final class OrderSearchScreen: ScreenObject {
 
     struct ElementStringIDs {
         static let searchField = "order-search-screen-search-field"
         static let cancelButton = "order-search-screen-cancel-button"
     }
```
pachlava: It looks like this is not needed anymore?

WooCommerce/UITestsFoundation/Screens/Settings/BetaFeaturesScreen.swift
```diff
@@ -1,16 +1,24 @@
...
 
-class BetaFeaturesScreen: BaseScreen {
+// This screen is currently unused. Given the purpose of UITestsFoundation is to provide developers
+// with an easy to use API to write tests, I'm going to keep the screen in the framework anyway, so
+// it'll be already available if someone needs to work with it or on it in the future.
+class BetaFeaturesScreen: ScreenObject {
 
     struct ElementStringIDs {
         static let enableProductsButton = "beta-features-products-cell"
     }
```
pachlava: This looks unused 🤔

WooCommerce/UITestsFoundation/Screens/Settings/BetaFeaturesScreen.swift
```diff
@@ -1,16 +1,24 @@
...
 
-    private let enableProductsButton = XCUIApplication().cells[ElementStringIDs.enableProductsButton]
+    // `expectedElement` is a `ScreenObject` utility to get the first element for the
```
pachlava: It looks like `from` is meant here?

```suggestion
    // `expectedElement` is a `ScreenObject` utility to get the first element from the
```

WooCommerce/UITestsFoundation/Screens/Settings/BetaFeaturesScreen.swift
pachlava: I'd say as long as this comment is the only place where return of the first element only is mentioned, it's useful. Also, for the people unaware of the `Alt+click` (I was one of them, until learning about it from your comment 😅), seeing this comment exposed is the fastest way to get this information.

WooCommerce/UITestsFoundation/Screens/Settings/BetaFeaturesScreen.swift
pachlava: I'm not sure if you meant keeping the comment or keeping the whole screen file 🤔. But I lean towards comment. In either case, I think keeping it will not hurt: it serves a purpose, it looks tidy, and it will hopefully save time for someone.

pachlava (approved):
I executed the tests locally, they were ✅, and also left some comments (non-blocking, mainly just to clarify some things). Thank you for the level of detail in your notes, and for such an amazing example of how things can be done!

And I'm sorry it took me so long.

WooCommerce/UITestsFoundation/Screens/MyStore/MyStoreScreen.swift
mokagio: > I would say that having tests instantiate tab bar components separately will mean better depiction of app structure and having the framework more robust.

I think that, too. But I might be biased...

WooCommerce/UITestsFoundation/Screens/Settings/BetaFeaturesScreen.swift
mokagio: I mean the file.

> In either case, I think keeping it will not hurt: it serves a purpose, it looks tidy, and it will hopefully save time for someone.

Cool, thanks 😄

WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
mokagio: >  It looks like we lose the option of using the object if it's declared this way?

Correct. I defined them a closures in the array only when they do not need to be used in other parts of the screen. That is, when the element is needed only to verify that the screen is loaded.

I did that to keep the code lean, only declaring a reusable getter when there's code that uses it.

One could argue that having the getter already declared would make it easier for folks to immediately use it in the future. I think we'll figure out what balance to strike in terms of minimal code vs spectrum of functionality offered out of the box once we start involving more people in writing tests. For the moment, I'm leaning towards keeping as little code as I can, _also because it makes my work easier_.

WooCommerce/UITestsFoundation/Screens/Orders/OrderSearchScreen.swift
mokagio: Good catch! Thank you 🙇

mokagio:
@jkmassel I'm going to merge this even though our conversation stalled. Just want you to know that I'm still thinking about your suggestions. For example, regarding this:

> Not sure I'm 100% sold on the separation of getter vs element?

I have been experimenting with a less verbose implementation where I call the getter directly instead of wrapping it in a computed `var` when it's only used one. E.g. [here](https://github.com/wordpress-mobile/WordPress-iOS/commit/82e7b15479bbea2231be3c527b7b7bdfa7c10ebe#diff-2b0de89d6dfec9cfeccb2b2991e308a880aea7e4ffdf6e7152ac1aec04ce57a8R5-R14). That's only a little step, though, I'm still keen to explore a property-wrapper-based solution 🤓

Issue #4716, opened 2021-08-03
Shipping Labels: update carriers step to accept multi packages


pmusolino:
Update the carriers step to handle multi-package and allow selecting rates for each of them

<img width="232" alt="Schermata 2021-08-03 alle 17 32 25" src="https://user-images.githubusercontent.com/495617/128043607-18f6c707-5d03-4b34-ba63-feb37560e38a.png">

PR #5094, opened 2021-09-30
Shipping Labels: update the parsing logic of Carriers and Rates for managing multi-packages

pmusolino:
Part of #4716

## Description
Since we are going to support the multi-packages in Shipping Labels, we need to update the logic for parsing Carriers and Rates from the endpoint `/rates`.
In this PR, I updated the `ShippingLabelCarriersAndRatesMapper`, which now will parse every single key which will have the `id` of the package (generated using UUID). The mapper now will return an array of carriers and rates for every single package.
Every `ShippingLabelCarriersAndRates` now will contain the package id, which will be used inside the Carrier and Rates screen in a later PR.
I also updated the logic in `ShippingLabelCarriersViewModel` for generating the rows of the view, because now we need to iterate around the array of carriers.

The new keys that we are going to parse to generate the `ShippingLabelCarriersAndRates` array are highlighted in the image below. Each of those keys represents the rates of each individual package passed to the endpoint.
Schermata 2021-09-30 alle 12 17 37

## Testing
1. Turn off the feature flag `shippingLabelsMultiPackage` under `DefaultFeatureFlagService`.
2. Try to create a new shipping label.
3. The Carriers and Rates screen should work like before.

Update release notes:


Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
```diff
@@ -5,8 +5,8 @@ import Codegen
 ///
 public struct ShippingLabelPackageSelected: Equatable, GeneratedFakeable {
 
-    /// The id will be always "default_box"
-    public let id: String = "default_box"
+    /// The id will be always the name of the package.
```
itsmeichigo, line 8: Nit: I don't think the ID is the same as the package's name?

Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
```diff
@@ -26,6 +34,7 @@ public struct ShippingLabelPackageSelected: Equatable, GeneratedFakeable {
     }
 
     public init(customPackage: ShippingLabelCustomPackage, totalWeight: Double, customsForm: ShippingLabelCustomsForm?) {
+        self.id = customPackage.title
```
itsmeichigo, line 37: From what I see when inspecting API requests on the web app, it seems that the custom packages always have ID `default_box`. Should we do the same on the app?

Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
```diff
@@ -36,6 +45,7 @@ public struct ShippingLabelPackageSelected: Equatable, GeneratedFakeable {
     }
 
     public init(predefinedPackage: ShippingLabelPredefinedPackage, totalWeight: Double, customsForm: ShippingLabelCustomsForm?) {
+        self.id = predefinedPackage.id
```
itsmeichigo, line 48: Should we use `default_box` as the ID here to be consistent with the web app too?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -71,7 +71,8 @@ final class ShippingLabelFormViewModel {
             if let customPackage = packagesResponse.customPackages.first(where: { $0.title == package.packageID }) {
                 let boxID = customPackage.title
                 let customsForm = customsForms.first(where: { $0.packageID == boxID })
-                return ShippingLabelPackageSelected(boxID: boxID,
+                return ShippingLabelPackageSelected(id: UUID().uuidString,
```
itsmeichigo, line 74: Should we use the initializer with custom package: `init(customPackage:totalWeight:customsForm:)` instead?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
```diff
@@ -84,7 +85,8 @@ final class ShippingLabelFormViewModel {
                 if let predefinedPackage = option.predefinedPackages.first(where: { $0.id == package.packageID }) {
                     let boxID = predefinedPackage.id
                     let customsForm = customsForms.first(where: { $0.packageID == boxID })
-                    return ShippingLabelPackageSelected(boxID: boxID,
+                    return ShippingLabelPackageSelected(id: UUID().uuidString,
```
itsmeichigo, line 88: Similar question as above - should we use initializer with predefined package here?

itsmeichigo (approved):
Great job transforming the mappers! I tested with the multi-package flag off and can still load the carrier list as before. I'm leaving some comments about the package ID, but these should not block the PR.

Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
pmusolino: Fixed in the other PR on which I'm working on 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: It doesn't change anything in this specific case, but for uniformity, I'll update the code to use the custom package initializer in the other PR that I'm preparing 👍

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: Same response as above, this will remain because probably I'll remove the other initializers, no more useful 👍

Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
pmusolino: I'll refactor this part of the code in the next PR, because with the multi-package support is no more useful. If the user will choose the same custom package for two different packages, it means that the API will return just one key with the name of the custom package, and it's wrong (because we have two different packages, with different specs, but with the same package).

Networking/Networking/Model/ShippingLabel/Packages/Carriers and Rates/ShippingLabelPackageSelected.swift
pmusolino: Same as above, I'll refactor this part in the next PR

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
pmusolino: Probably I'll remove this initializer because it's no more needed for the multi-package support

Issue #5051, opened 2021-09-24
Shipping Labels: impossible to digit more than two or three digits in phone number


pmusolino:
I discovered a serious bug (a regression) in Shipping Labels. This happens both on International and local shipping labels.
It seems that if the `Ship From` address has the phone missing, it's impossible to add a phone number with more than 3 digits. 

After typing 3 digits, the keyboard will be dismissed, the field is no more selectable and the done button will be enabled but by pressing it nothing will happen.

This issue happens also on `Ship To` address form, but the limit of digits will be 2 instead of 3. 

The only way to remedy this problem is to copy and paste a number with ten digits.

Video:

https://user-images.githubusercontent.com/495617/134660825-0a2a8623-2b60-4d7b-9f13-9084da78b261.mp4

itsmeichigo:
It seems that this bug only happens when building the app with Xcode 13.

itsmeichigo:
I can also reproduce the frozen screen by just tapping Done several times. I haven't yet found the cause for the issue though.

pmusolino:
@itsmeichigo just with Xcode 13? 😮 This is interesting...

rachelmcr:
Thanks @itsmeichigo! I hadn't tried with an Xcode 13 build and I do see the issue that way. I see the frozen screen when I change any of the address form fields (not just the phone number field).

In my investigation, it looks like the issue is linked to reloading the table data:

https://github.com/woocommerce/woocommerce-ios/blob/0cbf4c74fc0e2a7234cae7f7e148e064b0e9c6bc/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/Shipping%20Labels/Create%20Shipping%20Label%20Form/Shipping%20Address%20Validation/ShippingLabelAddressFormViewController.swift#L154

I haven't gotten farther than that, but removing that line avoids the freeze. (I might be mistaken about that, but if that is the cause of the freeze, we'll need to find a different way to update the form to display or hide validation errors.)

Since we aren't building releases with Xcode 13 yet, I think this doesn't have to be high priority/hotfixed right away, so I'll prioritize the feature work before investigating this one further.

itsmeichigo:
@rachelmcr thanks for finding where the issue lies! I think this might be caused by my fix for the instant validation in #4973. 

However I can't seem to reproduce the frozen screen when build with Xcode 12 so I can't confirm that. Can you clarify how you got the issue to happen?

rachelmcr:
[...]

PR #5095, opened 2021-09-30
Shipping Labels: Avoid reloading address table view when not neccessary

### LINKED ISSUE #5015: Shipping Labels: Address fields cannot be cleared by holding down delete key
**Describe the bug**
As reported by @astralbodies: On all of the Ship From fields, I can’t hold down the backspace key to keep deleting text. In my case, I was trying to delete the company name entirely and I had to keep tapping the backspace button.

**Expected behavior**
Holding down delete key should delete every character until the field is empty.

**Mobile Environment**
- WooCommerce iOS version: 7.5.0.2

itsmeichigo:
Partially fixes #5015

# Description
When fixing #4971, I added some changes to reload the table view when editing text fields on the Address Form to instantly show error rows when a field's validation fails. This leads to a lot of unnecessary reloading; and a side effect is that holding down the Delete key is impossible because the fields are constantly losing focus.

# Changes
To reload only when needed, I keep track of the list of local errors, and only reload the table view when local errors change.
I also keep track of the row index that's in focus to focus that row again after reloading the table view.

This helps avoiding unnecessary reloading - but only partially fixes the holding Delete key issue. I.e holding down the Delete key only works if there's no change to local error.

# Demo
https://user-images.githubusercontent.com/5533851/135444039-87501699-2fd7-4a7b-9f13-d05da5f99507.mov

# Testing
2. On Orders tab select an order that's eligible for creating shipping label.
3. Select Create Shipping Label and configure Ship From.
4. Type something on any field, then hold down Delete key. You should be able to delete the field as long as there's no change to validation errors of any field.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewModel.swift
@@ -1,3 +1,4 @@
+import Combine
 import UIKit
 import Yosemite
 
@@ -66,6 +67,10 @@ final class ShippingLabelAddressFormViewModel {
         return statesOfSelectedCountry.first { $0.code == address?.state }?.name
     }
 
+    private let localErrors = CurrentValueSubject<[ValidationError], Never>([])
+    private let currentRowIndex = CurrentValueSubject<Int?, Never>(nil)
+    private var validationSubscription: AnyCancellable?
+
     init(
         siteID: Int64,
         type: ShipType,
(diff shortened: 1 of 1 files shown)


pmusolino (approved):
Looks good to me. I just noticed that now the error reported here https://github.com/woocommerce/woocommerce-ios/issues/5051 is a little bit mitigated.
Now it seems that is possible to enter all the 10 digits in the phone field, but when an error occur (eg. you have exceeded 10 digits) the keyboard is dismissed and it is no longer possible to reselect the field.
I noticed that this behavior does not occur with the other fields: the keyboard is not dismissed, and it is always possible to select the text field.
So I wonder if perhaps there is a lack of control over that specific `phone` field, which is not handled like all the others.

https://user-images.githubusercontent.com/495617/135602206-5aac91f9-f216-40e4-ba3c-a738efc2d28e.mp4

itsmeichigo:
@pmusolino Actually I'm not sure if reloading the table view on change of text fields is a good way for instant validation at all. Since it seems to cause weird issues on iOS 15 and I don't think there's any workaround for those yet.

Do you think it's worth it to revert all logic for instant validation and accept that we'll only check for local validation when tapping the Done button?

**Edit**: Since #5051 can also be reproduced by tapping Done button 3 times, I think the issue will needs to be investigated further. I'll proceed to merge this for this release.

PR #5096, opened 2021-09-30
Shipping Labels: Update print label screen to support multiple purchased labels

rachelmcr:
Part of: #4714

## Description

As part of multi-package support for shipping labels, we need to support printing multiple shipping labels. This change allows the print screen to support multiple purchased labels:

* After purchasing multiple labels, all purchased labels are passed to the print screen and the UI reflects that there are multiple labels to print.
* After purchasing a single label, or when reprinting a single label, the UI is the same as before.

It isn't yet possible to purchase multiple labels, so that flow can't be directly tested but I've included screenshots below to show the changes in the UI.

## Changes

* `PrintShippingLabelCoordinator`, `PrintShippingLabelViewController`, and `PrintShippingLabelViewModel` are updated to handle an array shipping labels.
  * The print coordinator prints all of the labels and sends all of the labels' (non-nil, non-empty) customs forms to the `PrintCustomsFormsView` to be printed. (Full support for printing multiple customs forms will be done in another PR.) It also updates the printing progress view to show "Printing labels" when multiple labels are available to print.
  * The print view controller handles multiple labels and updates the UI to show plural strings ("Print Shipping Labels" and "Shipping labels purchased!") when there are multiple purchased labels.
* Updated unit tests to match the print screen changes.

Single label|Multiple labels
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-09-30 at 11 44 54|Simulator Screen Shot - iPhone 12 Pro - 2021-09-30 at 11 32 44
Simulator Screen Shot - iPhone 12 Pro - 2021-09-30 at 11 45 01|Simulator Screen Shot - iPhone 12 Pro - 2021-09-30 at 11 33 03

## Testing

2. Make sure you have at least one order with a purchased shipping label, and one order eligible to purchase a shipping label.
3. Turn off the feature flag `shippingLabelsMultiPackage` under `DefaultFeatureFlagService`.
4. Open the order detail for the order that has a purchased shipping label.
5. Find the package section in the order detail and select "Print Shipping Label."
6. Confirm the print screen reflects that there is a single label to print (no plural strings) and you can print the label.
7. Open the order detail for the order that is eligible to purchase a shipping label.
8. Select "Create Shipping Label" and fill out the form to complete the shipping label purchase.
9. After the purchase, confirm the print screen reflects that there is a single label to print (no plural strings) and you can print the label.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
@@ -135,8 +135,11 @@ private extension PrintShippingLabelCoordinator {
 private extension PrintShippingLabelCoordinator {
     /// Requests document data for printing a shipping label with the selected paper size.
     func requestDocumentForPrinting(paperSize: ShippingLabelPaperSize, completion: @escaping (Result<ShippingLabelPrintData, Error>) -> Void) {
+        guard let firstLabel = shippingLabels.first else {
+            return
+        }
         analytics.track(.shippingLabelReprintRequested)
-        let action = ShippingLabelAction.printShippingLabel(siteID: shippingLabels[0].siteID,
+        let action = ShippingLabelAction.printShippingLabel(siteID: firstLabel.siteID,
                                                             shippingLabelIDs: shippingLabels.map { $0.shippingLabelID },
                                                             paperSize: paperSize) { result in
             completion(result)
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
@@ -93,7 +93,8 @@ private extension PrintShippingLabelCoordinator {
     }
 
     func presentPrintInProgressUI() {
-        let viewProperties = InProgressViewProperties(title: Localization.inProgressTitle, message: Localization.inProgressMessage)
+        let viewProperties = InProgressViewProperties(title: Localization.inProgressTitle(labelCount: shippingLabels.count),
+                                                      message: Localization.inProgressMessage)
         let inProgressViewController = InProgressViewController(viewProperties: viewProperties)
         inProgressViewController.modalPresentationStyle = .overCurrentContext
         sourceNavigationController.present(inProgressViewController, animated: true, completion: nil)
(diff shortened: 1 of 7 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
```diff
@@ -135,8 +136,8 @@ private extension PrintShippingLabelCoordinator {
     /// Requests document data for printing a shipping label with the selected paper size.
     func requestDocumentForPrinting(paperSize: ShippingLabelPaperSize, completion: @escaping (Result<ShippingLabelPrintData, Error>) -> Void) {
         analytics.track(.shippingLabelReprintRequested)
-        let action = ShippingLabelAction.printShippingLabel(siteID: shippingLabel.siteID,
-                                                            shippingLabelIDs: [shippingLabel.shippingLabelID],
+        let action = ShippingLabelAction.printShippingLabel(siteID: shippingLabels[0].siteID,
```
itsmeichigo: Nit: Although most of the cases we have a non-empty shipping label list, but do you think it's worth it to `guard` that we have first item? I'd prefer avoiding direct subscripting unless we're 100% sure of all the cases.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
```diff
@@ -158,12 +159,14 @@ private extension PrintShippingLabelCoordinator {
     /// Show customs form printing if separate customs form is available
     ///
     func showCustomsFormPrintingIfNeeded() {
-        guard let url = shippingLabel.commercialInvoiceURL,
-              printType == .print else {
+        let urls = shippingLabels
+            .compactMap { $0.commercialInvoiceURL }
+            .filter { $0.isNotEmpty }
```
itsmeichigo, line 167: Thanks for checking this, I missed this before 👏

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
itsmeichigo: Another solution could be that we inject the `siteID` into the coordinator through the intializer, but I might be a bit over-engineering here 🤔 ⚙️

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelViewModel.swift
```diff
@@ -11,21 +11,21 @@ final class PrintShippingLabelViewModel {
...
 }
 
 // MARK: Public methods
 //
 extension PrintShippingLabelViewModel {
-    /// Sets the default selected paper size to the one from shipping label settings, if the user has not selected one in the print UI.
+    /// Sets the default selected paper size to the one from the first shipping label's settings, if the user has not selected one in the print UI.
     func loadShippingLabelSettingsForDefaultPaperSize() {
-        let action = ShippingLabelAction.loadShippingLabelSettings(shippingLabel: shippingLabel) { [weak self] settings in
+        let action = ShippingLabelAction.loadShippingLabelSettings(shippingLabel: shippingLabels[0]) { [weak self] settings in
```
itsmeichigo: Similar nit for direct subscripting as above 😅

itsmeichigo (approved):
This works great with the multi-package feature flag off! I have a couple of nits I'm leaving in the comments below, but I'm approving these changes anyway.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Print Shipping Label/PrintShippingLabelCoordinator.swift
rachelmcr: Good point! I thought about injecting the `siteID` through the initializer but it seemed like overkill since the labels also have the siteID as a property. I added `guard` statements for this and the other case in 18f92d9.

Issue #4440, opened 2021-06-16
[Mobile Payments] Show Coupons, if any, on receipts


allendav:
See also #4434

Issue #4439, opened 2021-06-16
[Mobile Payments] Show order fees, if any, on receipts


allendav:
See also #4434

Issue #4438, opened 2021-06-16
[Mobile Payments] Show shipping charges, if any, on receipts


allendav:
See also #4434

PR #5099, opened 2021-09-30
Add taxes to receipts

### LINKED ISSUE #4437: [Mobile Payments] Show tax collected, if any, on receipts
See also #4434

joshheald:
Resolves #4437 

## Description
Adds taxes to the receipt, where these are non-zero.
This implementation adds only the tax total, ignoring shipping tax and discount tax which are separately available. I initially included these in the receipt, but found that we don't split them out in the Cart, Checkout, or Confirmation on the web. Particularly since we don't have the items tax total, I think that adding just some of the components would add confusion, as the tax total wouldn't add up to the tax components that we print.

`ReceiptCartTotals` is a bit overkill at the moment, but will be used for more fields in #4438, #4439, and #4440.

## Testing

1. Go to the Orders tab
2. Open an unpaid "Pay on Delivery" order with non-zero taxes
3. Take card payment
4. Tap "Print Receipt"
5. Observe that the order lines include a "Total Tax" line.
6. Repeat with an order that has zero tax, observe that taxes are not shown.

Simulator Screen Shot - iPhone 12 mini - 2021-09-30 at 17 06 30

Update release notes:


Yosemite/YosemiteTests/Stores/ReceiptStoreTests.swift
```diff
@@ -82,6 +82,25 @@ final class ReceiptStoreTests: XCTestCase {
         XCTAssertEqual(UInt(mockOrder.total), parametersProvided?.parameters.amount)
         XCTAssertEqual(mockOrder.currency, parametersProvided?.parameters.currency)
     }
+
```
allendav, line 86: Would you add a test case for `0.00`?

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -127,21 +129,41 @@ private extension ReceiptRenderer {
...
+    }
+
+    private func summaryRow(title: String, amount: String) -> String {
+        """
+            <tr>
+                <td>
+                    \(title)
+                </td>
+                <td>
+                    \(amount) \(parameters.currency.uppercased())
```
allendav, line 139: One of these days we should switch to NSNumberFormatter. That doesn't have to be today. Would you open an issue?

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -242,5 +264,9 @@ private extension ReceiptRenderer {
             "Account Type",
             comment: "Reads as 'Account Type'. Part of the mandatory data in receipts"
         )
+
+        static let totalTaxLineDescription = NSLocalizedString(
+            "Total Tax",
```
allendav: In the Order Details Payment section we just say Taxes.

taxes

allendav (approved):
Tests well. I used taxable and non taxable products, discounts and taxable and non taxable shipping and everything presented correctly.

A few suggestions to consider then :shipit:

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -127,21 +129,41 @@ private extension ReceiptRenderer {
...
 
         return summaryContent
     }
 
+    private func taxesSummaryRow() -> String {
+        return taxSummaryRow(title: Localization.totalTaxLineDescription, amount: cartTotals.totalTax)
+    }
+
+    private func taxSummaryRow(title: String, amount: String) -> String {
+        guard amount != "0.00" else {
```
allendav: This will work for now, although comparisons to floats make me nervous and some currencies are non-decimal.

Could we at least try amount < 0.01 instead for now? Or are we safe because amount is a String (which is odd in its own right, but not our fault)

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
joshheald: For sure.

In this case, I wanted to use `CurrencyFormatter.formatAmount(amount: String, with currency: String)`, but couldn't easily do it because `CurrencyFormatter` is in the main target. Since we have user-displayable strings outside the main UI now, I think it would be useful to pull helper code like that out to a shared library. 

That seemed pretty disruptive for right now, and we could do with some team discussion about exactly what should be pulled out, and how we want to do that.

I've added #5108

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
joshheald: Hmm. I'd come to this conclusion because it's a String all the way from the API response. I didn't really want to do a conversion to a number in order to make this comparison, but I know what you mean about non-decimal currencies.

I'll have a noodle with turning it into an `NSDecimalNumber` to do a comparison, and see how it goes.

PR #5102, opened 2021-10-01
Order Detail: Allow state to be entered by text when needed

Ecarrion:
# Why

While comparing the app to Core I noticed that when a selected country does not have a list of states the merchant can manually enter a state as raw text.

This PR adjusts the address form to provide an input text field for those cases.

Additionally, it fixes a bug when the selected country changes so it properly clears the state fields.

# How

- Update `EditAddressForm` to show either an input text field or a selectable field when the ViewModel defines it.

- Update ViewModel to define `showStateFieldAsSelector` when the selected country does not have a state list.

- Update ViewModel to clear the selected state and the state field when a new country with states is selected.

# Demo

https://user-images.githubusercontent.com/562080/135550982-47d46528-9fef-47cd-87db-aaebf002967f.mov

# Testing Steps

- Navigate to the address form
- Select a country with states and select a state
- Select a country without states and see that you can enter another state manually.

Note: When switching to countries without a state list, the state is not cleared to mimic core behavior.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditAddressForm.swift
@@ -206,9 +206,7 @@ struct EditAddressForm: View {
                         }
                         Divider()
                             .padding(.leading, Constants.dividerPadding)
-                        TitleAndValueRow(title: Localization.stateField, value: viewModel.fields.state, selectable: true) {
-                            showStateSelector = true
-                        }
+                        stateRow()
                     }
                 }
                 .padding(.horizontal, insets: geometry.safeAreaInsets)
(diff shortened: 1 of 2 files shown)


ealeksandrov (commented):
Field logic is good ✅ 
But I think we need to clear `state` also when you select country with no states:

https://user-images.githubusercontent.com/3132438/135660507-11ef8278-d6d5-4984-b250-2cad41b12bcf.mp4

Ecarrion:
Hey @ealeksandrov, thanks for the review. I explicitly leave it like that because that is core behavior.

Uploading Screen Recording 2021-10-01 at 1.19.09 PM.mov…

Ecarrion:
Wow good catch @ealeksandrov, totally missed that.

I've fix that now.

https://user-images.githubusercontent.com/562080/135670922-3c4dd6cb-5f8c-44ae-9dfb-6325b56df84f.mov

Also in case you were wondering on why I'm not clearing the state field when a new country without state is selected, is to follow core behaviour.

https://user-images.githubusercontent.com/562080/135671036-ebb5ca58-9aa8-436c-9210-7411f90b0981.mov

Ready for another look!

ealeksandrov (approved):
:shipit:

PR #5103, opened 2021-10-01
Shipping Labels: Disable Done button on Package Details screen when no default package is set

itsmeichigo:
Part of #4599 

# Description
The new Package Details screen (supporting multi-package) currently only disables Done button when a package weight or dimension is invalid. If user has never purchased a label, no default package is set, the Done button should also be disabled to prevent proceeding without a selected package.

Extra: This PR also fixes the overlapping of the Move button on order item rows.

# Changes
- Sets default package ID an empty string if not found in settings.
- Updates Done button configuration to also consider a valid selected package ID.
- Updates Order Item rows to avoid overlapping Move button on contents.
- Adds validation error when selected package ID is empty to better communicate the error.

# Screenshot
| Before | After |
| ----- | ----- |
|  | 

# Testing
1. Make sure to test with a test store that's never created a shipping label before. Hint: use appstestadmin account with store named For The Birds (info can be found in the Shared Test Accounts in the Field Guide).
2. Select an order that's eligible for creating shipping label.
3. Select Create Shipping Label, and configure Ship From and Ship To.
4. On Package Details screen, notice that the Move button does not overlap with order item details if the details are long.
5. Notice that when no default package is selected, Done button is disabled and a validation error is displayed beneath the Selected Package row. Selecting a package should make this error go away and enable the Done button.

Update release notes:


pmusolino (approved):
LGTM :shipit:

PR #5104, opened 2021-10-01
Shipping Labels: Only show option for printing customs form if form URL is not empty

### LINKED ISSUE #5043: Shipping Labels: Customs form printing option is displayed when commercial invoice URL is an empty string
# Description
Currently we're checking to enable customs form printing when `commercial_invoice_url` field is present. However this field can sometimes be an empty string, which is invalid and should not be considered for printing form.

Two cases to consider:
- The Print Customs Form option in the Shipping Label More menu
- The Customs form printing screen displayed after printing the shipping label.

itsmeichigo:
Fixes #5043 

# Description
This PR makes sure that option to print customs form for shipping label is only present in Order Detail screen if `commercialInvoiceURL` is really present (not empty).

There's a similar check for `commercialInvoiceURL` in shipping label printing screen - however this has been fixed in #5096 so I'm not fixing it here to avoid conflict when merged to develop.

# Testing
1. Make sure you have installed and activated the WooCommerce WCShip plugin for your test store.
2. Select a non-international order and purchase shipping label for that order.
3. On Orders tab, select the order to see Order Details screen.
4. Select the overflow button in Shipping Label section, notice that the option to print customs form is not there.

Update release notes:


rachelmcr (approved):
Looks great! Confirmed in testing that the customs form print option shows up only when expected. Thanks for fixing this case! :shipit:

PR #5120, opened 2021-10-02
Observe keyboard and reposition `Notice` in `DefaultNoticePresenter`

### LINKED ISSUE #4942: NoticePresenter & keyboard
During the development of the Order Editing project, we noticed that notices presented by the `NoticePresenter` are hidden when the keyboard is present.

- Tasks

- [ ] Update `NoticePresenter` to move the notice up/down per keyboard presence
- [ ] Test changes don't break current notices presentation

selanthiraiyan:
Closes #4942

## Description
It has been observed and reported that Notices presented using the `NoticePresenter` don’t reposition themselves, when keyboard is presented or dismissed. i.e. Notice gets hidden by the keyboard.

This change includes the code to make the NoticePresenter update/animate the position of of the Notice, as per keyboard’s presence.

## Testing
The following is a scenario in which Notice can be tested.
1. Go to the `Products` tab.
2. Tap on the product and tap on the description.
3. Change the description and tap `Done`. 
4. In the Product form screen, tap on `Save` to save the edited description.
5. The description will be saved and observe that `Product saved` Notice is being presented from the bottom.
5. Quickly tap on anyone of the Title or Description textfields, and observe that the Notice is moved up as the keyboard appears.

By making local and temporary code changes, I have also tested the case in which **the keyboard is already presented and a Notice appears** and Notice is presented above the keyboard.
I also made sure that **Notice is moved to the bottom when keyboard is dismissed**.

https://user-images.githubusercontent.com/524475/135725340-e8b8ea5c-132d-4fd5-a276-e882d855def9.mov

Update release notes:


WooCommerce/Classes/Tools/Notices/DefaultNoticePresenter.swift
```diff
@@ -24,6 +24,9 @@ class DefaultNoticePresenter: NoticePresenter {
     ///
     weak var presentingViewController: UIViewController?
 
+    /// Observes keyboard and repositions Notice
+    ///
+    private var keyboardFrameObserver: KeyboardFrameObserver!
```
itsmeichigo: We want to refrain from forced wrapping whenever possible. Since this observer isn't created on init - maybe make it optional instead?

WooCommerce/Classes/Tools/Notices/DefaultNoticePresenter.swift
```diff
@@ -94,6 +97,25 @@ private extension DefaultNoticePresenter {
         noticeView.translatesAutoresizingMaskIntoConstraints = false
 
         let noticeContainerView = NoticeContainerView(noticeView: noticeView)
+
+        var onScreenBottomOffsetAdjustedForKeyboard: CGFloat = 0
+        keyboardFrameObserver = KeyboardFrameObserver { [weak self] keyboardFrame in
+            guard let self = self else { return }
+            // Subtract the tab bar height from keyboard height, if keyboard is visible
+            onScreenBottomOffsetAdjustedForKeyboard = -(keyboardFrame.height - (keyboardFrame.height > 0 ? self.offscreenBottomOffset : 0))
```
itsmeichigo: I'd suggest using a condition check to make it easier to read:
```Swift
if keyboardFrame.height > 0 {
    onScreenBottomOffsetAdjustedForKeyboard = self.offscreenBottomOffset - keyboardFrame.height
} else {
    onScreenBottomOffsetAdjustedForKeyboard = 0
}
```
WDYT?

itsmeichigo (commented):
Thanks for fixing this! I tested and it works as expected. I left a couple comments for code style below.

WooCommerce/Classes/Tools/Notices/DefaultNoticePresenter.swift
selanthiraiyan: Thanks for reviewing. 
You are right. I will make this an optional.

WooCommerce/Classes/Tools/Notices/DefaultNoticePresenter.swift
selanthiraiyan: I agree. I will introduce an if statement. 👍

itsmeichigo (approved):
The new changes look good to me - thanks for the contribution! 🚀

PR #5125, opened 2021-10-04
Shipping Labels: update carriers and rates step to accept multi packages

### LINKED ISSUE #4716: Shipping Labels: update carriers step to accept multi packages

pmusolino:
Closes #4716 

## Description
I updated the Shipping Label Carriers and Rates screen and view models to accept the multi packages. Now, the view will accept an array of packages and will return an array of rates, plus the sections of every package are compactable. 

## Changes
- Removed the unnecessary init methods in `ShippingLabelPackageSelected`.
- `ShippingLabelSelectedRate` now conforms to `Copiable`.
- Create a new SwiftUI view called `ShippingLabelCarrierSectionHeader` which represent the compactable header in the Carriers screen.
- Updated the `ShippingLabelCarriers` view for accepting an array of `ShippingLabelSelectedRate`, plus implemented a subview `ShippingLabelCarriersSection` which will represent a single compactable section.
- `ShippingLabelCarriersViewModel`: the rows have become sections, each section will contain the rows.
- Updated the unit tests.

#### Case 1: Single package
1. Select Create Shipping Label, proceed to fill all information. In Package Details, keep everything in one package.
2. Proceed to the Carriers and Rate selection.
3. Select the carrier and rate and press "Done".
4. Re-open the screen, the previous rate selected should be still selected.
5. Proceed, and buy the shipping label.

#### Case 2: Multiple packages
1. Select Create Shipping Label, proceed to fill all information. In Package Details, keep everything in multiple packages (two or more).
2. Proceed to the Carriers and Rate selection.
3. You should be able to select the carrier and rates for all the packages. The "Done" button will be enabled only when you will select a rate for all the packages.
4. Re-open the screen, the previous rates selected should be still selected.
5. Proceed, and buy the shipping labels.

## Screenshot

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriers.swift
```diff
@@ -42,9 +38,10 @@ struct ShippingLabelCarriers: View {
...
                                                               shippingCost: viewModel.shippingCost).renderedIf(viewModel.shouldDisplayTopBanner)
-                        ForEach(viewModel.rows) { carrierRowVM in
-                            ShippingLabelCarrierRow(carrierRowVM)
-                            Divider().padding(.leading, Constants.dividerPadding)
+                        ForEach(Array(viewModel.sections.enumerated()), id: \.offset) { index, sectionVM in
+                            ShippingLabelCarriersSection(section: sectionVM, safeAreaInsets: geometry.safeAreaInsets)
+                                .background(Color(.systemBackground))
+                            Spacer().frame(height: Constants.spaceBetweenSections)
                         }
                         .padding(.horizontal, insets: geometry.safeAreaInsets)
```
itsmeichigo: We may not need the padding here - instead how about adding padding for content of the section instead? Because it looks like the section is not extending edge-to-edge on landscape mode:

Simulator Screen Shot - iPhone 12 Pro - 2021-10-05 at 10 27 37

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewController.swift
```diff
@@ -331,24 +328,27 @@ private extension ShippingLabelFormViewController {
         } onButtonTouchUp: { [weak self] in
             guard let self = self else { return }
             ServiceLocator.analytics.track(.shippingLabelPurchaseFlow, withProperties: ["state": "purchase_initiated",
-                                                                                        "amount": self.viewModel.selectedRate?.rate ?? 0,
-                                                                                        "fulfill_order": self.shouldMarkOrderComplete])
+                 "amount": self.viewModel.selectedRates.map { $0.totalRate }
+                                                       .reduce(0, +),
```
itsmeichigo: Nit: I think it'd be better if we extract this into a variable so that we don't have to repeat the calculation in several places?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersSectionViewModel.swift
```diff
@@ -0,0 +1,8 @@
+import SwiftUI
+
+struct ShippingLabelCarriersSectionViewModel: Identifiable {
+    internal let id = UUID()
+
+    let packageNumber: Int
+    var rows: [ShippingLabelCarrierRowViewModel] = []
```
itsmeichigo: Nit: I think we should keep `rows` as a constant?

itsmeichigo (commented):
Thanks a lot for updating this view for multi-package support!

It seems like we haven't handled packages in original packaging when mapping from package attributes to `ShippingLabelPackageSelected`. So if I ship one of the items in its original package, I cannot find the package in the carriers screen:
https://user-images.githubusercontent.com/5533851/135964327-5ca9ad6c-6964-460a-8e86-53affc921318.mp4

I've found another minor issue regarding edge-to-edge layout in landscape mode (along with a couple of nits) that I've left in the comments down below. 

💡 A suggestion: should we also show an error icon on the header of each section if no rate is selected for that package yet? I think that way it'd be clearer for user why the Done button is not enabled.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriers.swift
itsmeichigo: Also, it seems like there's some extra horizontal padding on the header that we don't need.

pmusolino:
Thank you @itsmeichigo for testing it. I fixed all the nits, but I'm still investigating the items shipped in original packages. I didn't consider that we will not be able to find the package ID (which is custom and always `individual`). In that case, we are not considering it at all when we are generating the selected packages. I'm going to change this behavior to handle also the items shipped in original packaging.

pmusolino:
@itsmeichigo can you do a second check? Now everything should be ok with the original packaging.

itsmeichigo (changes requested):
Thank you for fixing the issue with the original packaging, I tested again and it works as expected! 🎉

However I noticed another issue just now 🤦‍♀️ . We're using `customsForms.first(where: { $0.packageID == boxID })` to find the matching customs form, which can be incorrect if we have several packages of the same ID.  

My suggestion is to have a unique ID for every package that ties customs form, selected package and selected rate all together. [Here](https://github.com/woocommerce/woocommerce-ios/compare/issue/4716-update-carriers-step-to-accept-multi-packages-main-logic...huong/shipping-label-attribute-update?expand=1) is how I imagine it could work. WDYT?

I also created a separated issue #5152 for further UI enhancements.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersViewModel.swift
```diff
@@ -61,30 +59,32 @@ final class ShippingLabelCarriersViewModel: ObservableObject {
...
     }
 
-    func generateRows(response: [ShippingLabelCarriersAndRates]) {
-        var tempRows: [ShippingLabelCarrierRowViewModel] = []
-        for resp in response {
-            tempRows += resp.defaultRates.map { rate in
+    func generateSections(response: [ShippingLabelCarriersAndRates]) {
+        var tempSections: [ShippingLabelCarriersSectionViewModel] = []
+        for (index, resp) in response.enumerated() {
+            let rows: [ShippingLabelCarrierRowViewModel] = resp.defaultRates.map { rate in
```
itsmeichigo, line 78: I also missed this earlier - in cases where no rates can be found for a package, this map will result in the missing row for that package. We should instead map rows by packages, and find the rates that match the package ID instead. How does that sound?

pmusolino:
> However I noticed another issue just now 🤦‍♀️ . We're using customsForms.first(where: { $0.packageID == boxID }) to find the matching customs form, which can be incorrect if we have several packages of the same ID.
My suggestion is to have a unique ID for every package that ties customs form, selected package and selected rate all together. Here is how I imagine it could work. WDYT?

You are right, this is not ideal if we have several packages of the same ID, and your solution seems to work ok, so feel free to open a PR targeting this PR 🙇

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriersViewModel.swift
pmusolino: If there is no rate for a package, I believe we should consider displaying the empty state, since it's mandatory for us to choose a rate for every package. In that case, we should compare the number of `packages` with the number of `sections`. They should match, and every section should contain at least 1 row. Otherwise, we should show an error, as it will not be possible to continue. WDYT?

itsmeichigo: Yes, I agree that we should show an error in that case, but I don't think displaying empty state screen is helpful for user since they may want to know which package needs updating. So how about we only display empty state when we cannot find rates for any package; otherwise we'll show a message under the empty sections only? Maybe a `ValidationErrorRow` is sufficient? We can bring this discussion to Slack to ask for design team's opinion and align with Android team.

pmusolino: Thank you @itsmeichigo for your suggestion. As discussed on slack, and as reported here https://github.com/woocommerce/woocommerce-ios/issues/5171 we decided to align the behavior to Android displaying the empty state, and to wait for a design proposal to improve the error state in a next iteration. 👍

pmusolino:
@itsmeichigo ready for a final 👀

itsmeichigo (approved):
Everything looks good now, thank you for your patience 😅 :shipit:

Issue #5053, opened 2021-09-24
[Mobile Payments] Modals dismiss without animation


koke:
I've noticed a few cases where the payments modals dismiss without animation.

It looks like ca23427f512b05040d1692efd7d3485ddf34c9cb intentionally sets `animated: false` when dismissing, so I'd like to understand the reasoning better before changing it back. cc @allendav

allendav:
It's because we don't animate between the modals in the series of card reader settings (discovery, found, connecting) modals when the view model changes, so I didn't want to introduce an animation just for the several-readers-found modal - which is basically another modal in the series.

The commit also avoids "presenting while presentation in progress", but I believe it will still do that if we change the several-readers-found presentation and dismissal to animated

koke:
Interesting, I thought/assumed we were animating them in, but the reality is that what's jarring is not the lack of animation but that the dimmed background disappears ~half a second before the modal does. When they are presented, both views appear at the same time:

https://user-images.githubusercontent.com/8739/135469431-7bcc69fa-abd0-49ff-b205-91989b87c999.MP4

ctarda:
Closing for now.

PR #5126, opened 2021-10-04
[Mobile Payments] [SDK 2.0] Prevent errors when canceling updates

### LINKED ISSUE #5124: [Mobile Payments] [SDK 2.0] Error when canceling update isn't cleared correctly
When an optional update is canceled, a modal error is shown for a split second, then it will show again whenever the reader disconnects/reconnects:

https://user-images.githubusercontent.com/8739/135838095-853de447-f5ac-435d-8a88-e2ab28b8528b.mp4

koke:
Fixes #5124 

There are two different issues being corrected in this PR:

1. When the user cancels an update, we don't treat that as an installation failure. Instead, we just leave the update as "available"
2. When an update fails and the user dismisses the error modal without retrying, we clear the error from the view model so the modal doesn't get triggered over and over.

## To test

1. Edit the WooCommerce scheme, and enable `-simulate-stripe-card-reader` in the Arguments tab.
2. In `StripeCardReaderService.start(_:)`, after the SDK has been initialized (I add this right after Terminal.shared.delegate = self), add the following line: `Terminal.shared.simulatorConfiguration.availableReaderUpdate = .available`
2. Go to Settings > In-Person Payments > Manage Card Reader and connect to a reader
3. After the reader connects, it should show an available update
4. Tap on "Update Reader Software"
5. Tap on Cancel before it completes
6. It should show no error. Disconnecting and reconnecting should show no error

Also tested with the same steps, but simulating an SDK error instead of user cancelation. To do that, replace the `installUpdate` method in `StripeCardReaderService` with this:

```swift
    public func installUpdate() -> Void {
        softwareUpdateSubject.send(.failed(error: CardReaderServiceError.bluetoothDenied))
        softwareUpdateSubject.send(.available)
    }
```

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -472,7 +472,10 @@ extension StripeCardReaderService: BluetoothReaderDelegate {
 
     public func reader(_ reader: Reader, didFinishInstallingUpdate update: ReaderSoftwareUpdate?, error: Error?) {
         if let error = error {
-            softwareUpdateSubject.send(.failed(error: error))
+            let underlyingError = UnderlyingError(with: error)
+            if underlyingError != .commandCancelled {
+                softwareUpdateSubject.send(.failed(error: error))
+            }
             softwareUpdateSubject.send(.available)
         } else {
             softwareUpdateSubject.send(.completed)
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
@@ -135,8 +135,8 @@ private extension CardReaderSettingsConnectedViewController {
                     tryAgain: {
                         viewModel.startCardReaderUpdate()
                     },
-                    close: { [settingsAlerts] in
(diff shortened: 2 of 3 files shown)


Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -472,7 +472,10 @@ extension StripeCardReaderService: BluetoothReaderDelegate {
 
     public func reader(_ reader: Reader, didFinishInstallingUpdate update: ReaderSoftwareUpdate?, error: Error?) {
         if let error = error {
-            softwareUpdateSubject.send(.failed(error: error))
+            let underlyingError = UnderlyingError(with: error)
+            if underlyingError != .commandCancelled {
```
joshheald, line 476: Nit: `UnderlyingError(with: error)` could probably be used in the `if` condition directly here, to make it a little more readable by removing repetition of underlyingError.

joshheald (approved):
Looks good and tests well. 

One very minor nit which I considered not even mentioning, so ignore it and merge if you prefer

Probably a separate issue: see this video, the shade gets dismissed before the alert by 0.5-1s, which looks a little odd. Very minor but while you're head's in it, any idea what might be causing it?

https://user-images.githubusercontent.com/2472348/135879264-61971275-9fae-4c4d-ab5e-592f5736ede9.mp4

koke:
> the shade gets dismissed before the alert by 0.5-1s, which looks a little odd

This is being tracked in #5053. It's been bugging me, but I haven't figured it out yet

PR #5127, opened 2021-10-04
[Mobile Payments] Add coupons and shipping to receipts

### LINKED ISSUE #4438: [Mobile Payments] Show shipping charges, if any, on receipts

### LINKED ISSUE #4440: [Mobile Payments] Show Coupons, if any, on receipts

joshheald:
Resolves #4440 
Resolves #4438 

## Description
Adds the "Discounts" and "Shipping" lines to In Person Payment receipts.

All lines populated

Shipping and Tax are shown whenever they are non-zero. I've changed behaviour here to check for 0 specifically, rather than use `<=`, in order to take account of potential for refund receipts in future having negative values here.

Discounts are shown when there are coupons applied to the order, or when the discount is non-zero. This is to account for coupons which don't have a discount value, e.g. free shipping coupons.

## Testing

### Shipping
1. Go to the Orders tab
2. Open an unpaid "Pay on Delivery" order with non-zero shipping
3. Take card payment
4. Tap "Print Receipt"
5. Observe that the order lines include a "Shipping" line.
6. Repeat with an order that has zero shipping, observe that the "Shipping" line is not shown.

Shipping line

### Coupons
1. Go to the Orders tab
2. Open an unpaid "Pay on Delivery" order with a coupon that provides a dollar or percentage discount
3. Take card payment
4. Tap "Print Receipt"
5. Observe that the order lines include a "Discount" line, including the coupon codes.
6. Repeat with an order that has a free shipping coupon, observe that the "Discount" line is still shown even though it's 0.00.
7. Repeat with an order that has no coupons, observe that the "Discount" line is not shown.

Free shipping coupon

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/ReceiptStore.swift
+++ b/Yosemite/Yosemite/Stores/ReceiptStore.swift
@@ -58,6 +58,24 @@ public class ReceiptStore: Store {
 
 
 private extension ReceiptStore {
+    func print(order: Order, parameters: CardPresentReceiptParameters, completion: @escaping (PrintingResult) -> Void) {
+        let content = generateReceiptContent(order: order, parameters: parameters)
+        receiptPrinterService.printReceipt(content: content, completion: completion)
+    }
+
+    func generateContent(order: Order, parameters: CardPresentReceiptParameters, onContent: @escaping (String) -> Void) {
+        let content = generateReceiptContent(order: order, parameters: parameters)
+        let renderer = ReceiptRenderer(content: content)
+        onContent(renderer.htmlContent())
+    }
+
+    func generateReceiptContent(order: Order, parameters: CardPresentReceiptParameters) -> ReceiptContent {
+        let lineItems = generateLineItems(order: order)
+        let cartTotals = generateCartTotals(order: order, parameters: parameters)
+
+        return ReceiptContent(parameters: parameters, lineItems: lineItems, cartTotals: cartTotals)
+    }
+
     func generateLineItems(order: Order) -> [ReceiptLineItem] {
         order.items.map { item in
             ReceiptLineItem(
@@ -76,24 +94,6 @@ private extension ReceiptStore {
         }
         totalLines.append(ReceiptTotalLine(description: ReceiptContent.Localization.amountPaidLineDescription, amount: parameters.formattedAmount))
         return totalLines
-    }
-
-    func generateReceiptContent(order: Order, parameters: CardPresentReceiptParameters) -> ReceiptContent {
-        let lineItems = generateLineItems(order: order)
-        let cartTotals = generateCartTotals(order: order, parameters: parameters)
-
-        return ReceiptContent(parameters: parameters, lineItems: lineItems, cartTotals: cartTotals)
-    }
-
-    func print(order: Order, parameters: CardPresentReceiptParameters, completion: @escaping (PrintingResult) -> Void) {
-        let content = generateReceiptContent(order: order, parameters: parameters)
-        receiptPrinterService.printReceipt(content: content, completion: completion)
-    }
-
-    func generateContent(order: Order, parameters: CardPresentReceiptParameters, onContent: @escaping (String) -> Void) {
-        let content = generateReceiptContent(order: order, parameters: parameters)
-        let renderer = ReceiptRenderer(content: content)
-        onContent(renderer.htmlContent())
     }
 
     func loadReceipt(order: Order, onCompletion: @escaping (Result<CardPresentReceiptParameters, Error>) -> Void) {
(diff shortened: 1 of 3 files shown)


koke (changes requested):
Overall it looks good, but I think discounts should show as a negative amount in the receipt. Otherwise it can get confusing.

Order|Receipt
-|-
Screen Shot 2021-10-04 at 17 03 39|Screen Shot 2021-10-04 at 17 04 10

joshheald:
> Overall it looks good, but I think discounts should show as a negative amount in the receipt. Otherwise it can get confusing.

@koke Yeah good call. That's done now. Thanks!
Negative discount receipt

Yosemite/Yosemite/Stores/ReceiptStore.swift
```diff
@@ -69,31 +87,52 @@ private extension ReceiptStore {
...
+
+        return subtotalLines + totalLine
     }
 
-    func generateReceiptContent(order: Order, parameters: CardPresentReceiptParameters) -> ReceiptContent {
-        let lineItems = generateLineItems(order: order)
-        let cartTotals = generateCartTotals(order: order, parameters: parameters)
+    func discountLine(order: Order) -> ReceiptTotalLine? {
+        let discountValue = NSDecimalNumber(apiAmount: order.discountTotal).decimalValue
+        if discountValue == 0 && order.coupons.isEmpty {
```
koke, line 104: I keep wondering if we want to check both these conditions. In the most common case, if there is a discount it will be because a coupon was applied. However, there are ways that these two things could go in separate ways:
- You can create a coupon with $0 value but that [allows free shipping](https://docs.woocommerce.com/document/free-shipping/#section-3)
- I see several extensions that let you set up discounts when you bundle products, or are quantity based, which would show up as a discount with no coupon.

I think in this second case we'd still want to show the Discount line without a coupon code, so the prices add up?

Yosemite/Yosemite/Stores/ReceiptStore.swift
```diff
@@ -69,31 +87,52 @@ private extension ReceiptStore {
...
+            couponCodes = "(\(couponCodes))"
+        }
+        return String.localizedStringWithFormat(ReceiptContent.Localization.discountLineDescription, couponCodes)
     }
 
-    func print(order: Order, parameters: CardPresentReceiptParameters, completion: @escaping (PrintingResult) -> Void) {
-        let content = generateReceiptContent(order: order, parameters: parameters)
-        receiptPrinterService.printReceipt(content: content, completion: completion)
+    func discountLineAmount(order: Order, value: Decimal) -> String {
+        if value > 0 {
```
koke, line 124: This seems a bit problematic. In a previous check, we don't show this if it's zero, and I don't think WooCommerce will let you create a coupon with a negative discount. That said, if that ever happens, both positive and negative `discountTotal` values would show a `-` sign. I think we should negate the numeric value before converting to string instead.

koke (approved):
Sorry, I just realized that the discount seems to be already applied to each item, both in the receipt and the order view. I don't think we need to hold back this PR for that, but it seems really confusing to me. Can you log that as a new issue, and get a second opinion from a designer?

Other than that, I noted a couple of edge cases that don't seem enough of a blocker, but it'd be good to address in this or a future PR.

Order|Receipt
-|-
Screen Shot 2021-10-05 at 11 50 08|Screen Shot 2021-10-05 at 11 50 13

Web|Email
-|-
Screen Shot 2021-10-05 at 11 51 18|Screen Shot 2021-10-05 at 11 52 29

Yosemite/Yosemite/Stores/ReceiptStore.swift
joshheald: I agree, and with the existing code both of these examples will show up... the only time this line won't show is for a Zero discount where no coupons were applied.

Yosemite/Yosemite/Stores/ReceiptStore.swift
joshheald: Yeah, I'm not super happy with the treatment of numbers here either and I've got a task to look in to how we can do it better. If we're going to convert the numbers from what the API gives us for display, I want to be really careful about how we do it. Do you know of any other examples in the app where we already do that?

It also probably ties in to getting the `CurrencyFormatter` available here, which is another thing on my list to look in to.

My reasoning for this code is:
1. The common case we want to cover is making a positive API value negative,
2. We don't want a zero value to be negative, because that's weird.
3. It seems like negative API values for discount aren't really a thing.

But that isn't super robust so I agree we should do something about point 3, so we don't confuse things. Most likely example I can think of for a positive discount is perhaps on a refund receipt. I'm going to add an issue to look at this after I have investigated numbers as mentioned above, but given we can't find a current use case where positive discounts exist, I think it's best to merge this for now.

PR #5132, opened 2021-10-04
Order Editing: Update Country/State selected style

### LINKED ISSUE #5113: Order Editing: Wrong cell style when country/state is selected
When a country or a state is selected its row should have a body style, instead of a text subtle style.

Expected | Current
--- | ---
expected | currect

Ecarrion:
closes #5113 

# Why

Selecting a country or a state had the wrong style on the Edit Address screen. This PR makes sure to update that.

# How

Updated `TitleAndValue` row to receive a new `Value` type instead of a plain string. This Value type has to options, `placeholder` or `content`. Depending on the provided option the view will render the underlying text with a different style.

# Screenshots

Country Selected | State Selected
--- | ---
first | second

# Testing Steps

- Go to edit an order shipping address.
- See that when no country or state(where there is a list of states) is selected, there is a placeholder text with a subtle text style.
- See that when a country or a state is selected, the country/state name is rendered with a body text style.

Update release notes:


ealeksandrov (approved):
Nice little improvement! :shipit:

PR #5133, opened 2021-10-05
[Mobile Payments] [SDK 1.4 Reader Updates] Disable update button if battery low

### LINKED ISSUE #5131: [Mobile Payments] [SDK 1.4] Something went wrong (reader software update fails when battery < 50%)
This isn't just an SDK 2 thing. It happens with Terminal SDK 1.4 too.

I have a Chipper CHB20xxx reader. It is apparently eligible for an update to configuration `SZZZ_Generic_v46` (which is NOT mentioned on https://stripe.com/docs/terminal/readers/bbpos-chipper2xbt#reader-software-releases, notably)

However, it also only has a 28% charge. As a result, the update immediately fails with "Something went wrong":

IMG_0174

The actual problem (low battery) is only visible in the debugger:

Screen Shot 2021-10-04 at 2 34 44 PM

We should display this error to the end user. See similar SDK 2.0 pull request here: https://github.com/woocommerce/woocommerce-ios/pull/5073

allendav:
Closes #5131 

@ctarda - I think we should try and squeeze this into 7.7 release since many merchants will have less-than-half-charged readers and will not understand that "something went wrong" means charge-your-reader. What do you think? I think the change is focused and thus low risk.

Applies to SDK 1.4 optional updates (SDK 1.4 doesn't have required updates)

SDK 1.4 doesn't have updates applied at connection time. Instead, we call Terminal.shared.checkForUpdate after connecting to see if there was an update. (Note: Per https://stripe.com/docs/terminal/sdk-migration-guide this has been removed in 2.0)

And further, per https://stripe.com/docs/terminal/sdk-migration-guide#bluetooth-reader-updates : “[With Terminal SDK 1.4] all Bluetooth reader software updates were optional—it was the responsibility of your integration to call the checkForUpdate: method and install an update if one was available. With the 2.0.0 Terminal iOS and Android SDKs, we’re categorizing some updates as mandatory, and others as optional.”

Changes:
- Disable the update button after connecting to a reader if the battery is less than 50% full

Screenshot:

IMG_0177

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -70,6 +70,9 @@ private extension CardReaderSettingsConnectedViewController {
         guard let viewModel = viewModel else {
             return
         }
+
+
+
```
koke: Was all this whitespace intentional?

```suggestion

```

koke (approved):
I'm not sure how to test this in the 1.4 SDK without a pending update, but the code looks good to me, and it's a nice change.

One thing that's not clear to me is whether the current code only reads the battery level on connection or it's updated more frequently. I'm wondering what happens if a user connects to the reader with 49% battery, see the message, then plug it in and wait for it to charge. Will the app react to that automatically, or only when reconnecting? I'm not sure the SDK provides a way to monitor battery, but in that case we might want to tweak the copy to make that clearer.

ctarda:
@allendav  if you want to get this into 7.7 you will need to diff against the release branch, not against develop, and then ping the release manager to cut a new beta

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: IIRC, SDK 1.4 only updates the battery level on connection. We've asked Stripe to improve this. Not sure if that is still the case with SDK 2 but I will check.

allendav: I'll fix the whitespace. Not intentional

allendav:
@ctarda wrote:

> If you want to get this into 7.7 you will need to diff against the release branch, not against develop, and then ping the release manager to cut a new beta

Can do

allendav:
@jkmassel - I've changed the base branch to release/7.7 just now. Is this ok to merge and ask for a new beta?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: Fixed whitespace in https://github.com/woocommerce/woocommerce-ios/pull/5133/commits/82bdc6fb2399c49f055a23934b0f1d945d83be95

PR #5136, opened 2021-10-05
Shipping Labels: Fix flaky customs form list test

itsmeichigo:
# Description
A missing quantity in the fake instance of `ShippingLabelCustomsForm` in the `test_done_button_is_disable_when_not_all_fields_are_valid` test of customs form list view model causes the test to fail. 

This PR fixes the test by adding a non-zero quantity to the fake model, to make sure that the total value of the test item is larger than zero.

# Testing
Run the test a few times to make sure that it always passes.

Update release notes:


rachelmcr (approved):
LGTM! :shipit:

PR #5137, opened 2021-10-05
Update notice title after switching stores based on feature flag

jaclync:
Part of #5032 

## Why

When we support multi-store push notifications, we no longer want to show the message `You will only receive notifications from this store` in the notice after switching stores. This PR shows a different notice message based on the feature flag.

## Changes

- In `SwitchStoreNoticePresenter`, updated store-switching notice title based on `pushNotificationsForAllStores` feature flag with a test case on the notice title. The plan is to delete the new test case after the feature is fully launched.

### Feature flag is on (default in debug builds)

- Launch the app from PR branch
- Switch to another store --> after switching stores, should see a notice that just says `Switched to \(siteName).`

### Feature flag is off

- In Xcode, disable the feature flag by returning `false` for `pushNotificationsForAllStores` in `DefaultFeatureFlagService`
- Launch the app from PR branch
- Switch to another store --> after switching stores, should see a notice that just says `Switched to \(siteName). You will only receive notifications from this store.`

## Example screenshots

feature flag off | feature flag on
-- | -- 
Simulator Screen Shot - iPhone 11 - 2021-10-05 at 15 02 47 | Simulator Screen Shot - iPhone 11 - 2021-10-05 at 14 30 52

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
+++ b/WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
@@ -7,11 +7,14 @@ final class SwitchStoreNoticePresenter {
 
     private let sessionManager: SessionManagerProtocol
     private let noticePresenter: NoticePresenter
+    private let featureFlagService: FeatureFlagService
 
     init(sessionManager: SessionManagerProtocol = ServiceLocator.stores.sessionManager,
-         noticePresenter: NoticePresenter = ServiceLocator.noticePresenter) {
+         noticePresenter: NoticePresenter = ServiceLocator.noticePresenter,
+         featureFlagService: FeatureFlagService = ServiceLocator.featureFlagService) {
         self.sessionManager = sessionManager
         self.noticePresenter = noticePresenter
+        self.featureFlagService = featureFlagService
     }
 
     /// Present the switch notice to the user, with the new configured store name.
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
```diff
@@ -24,13 +27,25 @@ final class SwitchStoreNoticePresenter {
...
+extension SwitchStoreNoticePresenter {
+    enum Localization {
+        static let titleFormatWithPushNotificationsForAllStoresDisabled =
+            NSLocalizedString("Switched to %1$@. You will only receive notifications from this store.",
+                              comment: "Message presented after users switch to a new store when multi-store push notifications are not supported. "
+                                + "Reads like: Switched to {store name}. You will only receive notifications from this store. "
+                                + "Parameters: %1$@ - store name")
+        static let titleFormat = NSLocalizedString("Switched to %1$@.",
+                                                   comment: "Message presented after users switch to a new store. "
+                                                    + "Reads like: Switched to {store name}. You will only receive notifications from this store. "
```
itsmeichigo: Nit: this comment should be `Reads like: Switched to {store name}.` only right? 😅

itsmeichigo (approved):
This works great! And thank you for the thorough unit tests. I left only a small nit in the comment.

WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
jaclync: Great catch! Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5137/commits/15981818fd0b1004434fda23708f380d619ac7ab

jaclync:
Thanks for catching the comment Huong! Just updated the PR, merging now 😄

PR #5138, opened 2021-10-05
Update to released WPMediaPicker

joshheald:
Updates WPMediaPicker to the non-beta 1.7.3 release. Should not make any user-facing difference, as we already used the beta version and it fixes a crash which we've already worked around.

## Testing
Run the app and use the WP Media Picker (e.g. from Products > Add Photo)

Update release notes:

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -50,7 +50,7 @@ target 'WooCommerce' do
 
   aztec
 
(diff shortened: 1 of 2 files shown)


koke:
Usually the release manager handles this when merging the release branch into develop, there's no need to do it right now. It's probably best to leave it until then, or at least double check with Platform 9 before merging

PR #5139, opened 2021-10-05
Fix navigation bar button color on iOS 15

### LINKED ISSUE #5083: Shipping Labels: colors of navigation bar buttons are inconsistent on iOS 15
**Describe the bug**
On iOS 15, navigation bar buttons on SwiftUI views are inconsistent - some are purple and some are pink:

 

**Expected behavior**
All Done button should be pink.

rachelmcr:
Fixes: #5083

## Description

On iOS 15, navigation bar buttons on SwiftUI views are inconsistent - some are purple and some are pink. This fixes them so navigation bar buttons on SwiftUI views are all pink as expected.

It also fixes the buttons so that confirmation buttons (e.g. "Done") are bold as expected.

## Changes

* Replaces the deprecated `.navigationBarItems` with `.toolbar` and `ToolBarItem`.
* Replaces `.navigationBarTrailing` with `.confirmationAction` and `.navigationBarLeading` with `.cancellationAction` where appropriate. Using `.confirmationAction` still uses the trailing placement but also makes the button bold.
* Adds a `.navigationBarStyle` modifier to set the accent color, making the navigation bar buttons consistently pink.
* Removes the unused `.minimalNavigationBarBackButton` modifier.

/|Before|After
-|-|-
Application log|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 19 31|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 09 56
Edit Order Address Form|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 21 47|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 09 35
Edit Customer Note|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 21 34|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 31 30
Shipping Labels - Package Details|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 26 00|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 02 54
Shipping Labels - Package Selected|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 23 52|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 02 56
Shipping Labels - Add New Package > Package Type|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 23 57|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 03 02
Shipping Labels - Customs Form|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 24 06|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 06 10
Shipping Labels - Carrier and Rates|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 26 26|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 03 15
Shipping Labels - Payment Method|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 24 32|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 03 19
Shipping Labels - Add New Payment Method|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 13 01 28|Simulator Screen Shot - iPhone 13 Pro - 2021-10-05 at 12 50 34

## Testing

1. Make sure the WCShip plugin is installed and activated on your store, and you have at least one order eligible for shipping labels.
2. Build the app in debug mode on a device or simulator running iOS 15.
3. Go to My Store > Settings icon > Help & Support > View Application Log > Current and confirm the share icon in the top right is pink and continues to work as expected.
4. Go to Orders and select an order eligible for shipping labels.
5. Scroll down to the "Customer" section and tap the edit icon next to the Shipping Details address.
6. Make a change to the address and confirm that the Done button is bold and pink.
7. Tap the "Create Shipping Label" button. Go through the entire shipping label flow and confirm the navigation bar buttons are all pink and bold when enabled.

## Submitter Checklist

Update release notes:


WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
```diff
@@ -86,6 +86,13 @@ struct ErrorStyle: ViewModifier {
     }
 }
 
+struct NavigationBarStyle: ViewModifier {
```
Ecarrion: What do you think about naming this something like `AccentBarStyle` or `AccentNavigationBarStyle`.

If I see `NavigationBarStyle` in the wild I would not know what style I am applying 😅

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
```diff
@@ -86,6 +86,13 @@ struct ErrorStyle: ViewModifier {
...
+    func body(content: Content) -> some View {
+        content
+            .accentColor(Color(.accent)) // The color of bar button items in the navigation bar
```
Ecarrion, line 92: I think someone mentioned that we have to use `tint` for iOS 15, right?

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
```diff
@@ -120,4 +127,8 @@ extension View {
     func errorStyle() -> some View {
         self.modifier(ErrorStyle())
     }
+
+    func navigationBarStyle() -> some View {
+        self.modifier(NavigationBarStyle())
+    }
```
Ecarrion: [Q] Do you think we should we add this extension only for `NavigationView`? Or is it useful for any view?

Also probably something like `wooNavigationBarStyle` or `accentNavigationBarStyle` could be more describing, what do you think?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressForm.swift
```diff
@@ -225,15 +225,18 @@ struct EditOrderAddressForm: View {
...
-            ToolbarItem(placement: .navigationBarLeading) {
+            ToolbarItem(placement: .cancellationAction) {
                 Button(Localization.close, action: {
                     dismiss()
                     viewModel.userDidCancelFlow()
                 })
             }
+            ToolbarItem(placement: .confirmationAction) {
+                navigationBarTrailingItem()
+            }
```
Ecarrion, line 239: TIL: I knew there must be an easy way to achieve that! 😅

Ecarrion (commented):
This looks good @rachelmcr, only some minor naming suggestions!

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
rachelmcr: I named it `NavigationBarStyle` with the idea that, if needed, other styling could be added in the future that also applies to all navigation bars / elements in the navigation bar. Along those lines, I like your suggestion below to use `wooNavigationBarStyle` so maybe this can be `WooNavigationBarStyle` — since this is used to add the default Woo styling to the navigation bar. (Updated in dcf2e04.)

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
rachelmcr: Eventually we'll want to add `tint`, but after trying it out I realized we'll have to wait on this. `tint` isn't available in Xcode 12.5, which we're still using for CI. I checked and `accentColor` does still work on iOS 15.0, so for now we can use it.

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
rachelmcr: > [Q] Do you think we should we add this extension only for NavigationView? Or is it useful for any view?

We have some SwiftUI views (e.g. `ApplicationLogDetailView`, `EditOrderAddressForm`) that need the modifier but don't have a `NavigationView`, so I think we need this extension for any view.

> Also probably something like wooNavigationBarStyle or accentNavigationBarStyle could be more describing, what do you think?

I think `wooNavigationBarStyle` makes a lot of sense here, thanks! I updated the naming in dcf2e04.

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
rachelmcr: Would it make sense to leave a comment with some sort of reminder about that upcoming change?

rachelmcr:
Thanks for the feedback, @Ecarrion! This is ready for another look.

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
Ecarrion: Yup, probably worth it to add a comment or create an issue so we can track this! 😅

Ecarrion (approved):
🎉

WooCommerce/Classes/View Modifiers/WooStyleModifiers.swift
rachelmcr: I created https://github.com/woocommerce/woocommerce-ios/issues/5147 to track this 👍

PR #5145, opened 2021-10-06
Shipping Labels: Make name optional for local address validation when company is specified

### LINKED ISSUE #4681: Shipping Labels: make name non mandatory if the company name is specified
To allow companies to create shipping labels without having to specify a person's name, we should update the address validation logic to make the name's optional if the company's name is specified.

And until 1812-gh-Automattic/woocommerce-connect-server is implemented, we should make sure that the address object in the API request doesn't include the name field completely if it's empty, as sending an empty string will throw an error.

itsmeichigo:
Fixes #4681 

# Description
This PR make name field in address validation form optional when company field is not empty. This also makes sure that name is only sent in requests to fetching carriers and rates & purchasing label when it's not empty.

# Changes
- Networking layer: 
   - Updates `ShippingLabelAddress` encoder to only encode name when it's not empty
   - Updates `ShippingLabelAddress` decoder to only decode name when the field is present.
- ShippingLabelAddressFormViewModel:
   - Adds new variable to check if name is optional
   - Updates local validation for name field
   - Fixes incorrect row index when validation error changes.
- ShippingLabelAddressFormViewController: updates correct placeholder text for Name field based on whether the field is optional.

# Demo
https://user-images.githubusercontent.com/5533851/136159595-c41a97d7-a1ee-40aa-915d-5eaf01c3591e.mp4

# Testing
2. On Orders tab select an order that's eligible for creating shipping labels.
3. Select Create shipping label and select Ship From.
4. Clear both Name and Company fields, notice that the placeholder on Name field is Required and a validation error shows up below the field.
5. Enter value for Company field, notice that placeholder on Name field is Optional and the validation error goes away. Tap Done and notice that the validation succeeds when other fields are valid.
6. Proceed to configure other info for the shipping label. Notice that the carriers and rows list can be loaded like before.
7. Proceed to purchase the label. Make sure that it still succeeds and information on the purchased label is correct.

Update release notes:

### DIFF
--- a/Networking/Networking/Model/ShippingLabel/ShippingLabelAddress.swift
+++ b/Networking/Networking/Model/ShippingLabel/ShippingLabelAddress.swift
@@ -54,11 +54,36 @@ public struct ShippingLabelAddress: GeneratedCopiable, Equatable, GeneratedFakea
 
 // MARK: Codable
 extension ShippingLabelAddress: Codable {
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let name = try container.decodeIfPresent(String.self, forKey: .name) ?? ""
+        let company = try container.decode(String.self, forKey: .company)
+        let phone = try container.decode(String.self, forKey: .phone)
+        let country = try container.decode(String.self, forKey: .country)
+        let state = try container.decode(String.self, forKey: .state)
+        let address1 = try container.decode(String.self, forKey: .address1)
+        let address2 = try container.decode(String.self, forKey: .address2)
+        let city = try container.decode(String.self, forKey: .city)
+        let postcode = try container.decode(String.self, forKey: .postcode)
+
+        self.init(company: company,
+                  name: name,
+                  phone: phone,
+                  country: country,
+                  state: state,
+                  address1: address1,
+                  address2: address2,
+                  city: city,
+                  postcode: postcode)
+    }
+
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
 
         try container.encode(company, forKey: .company)
-        try container.encode(name, forKey: .name)
+        if !name.isEmpty {
+            try container.encode(name, forKey: .name)
+        }
         try container.encode(phone, forKey: .phone)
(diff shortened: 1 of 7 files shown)


Networking/Networking/Model/ShippingLabel/ShippingLabelAddress.swift
```diff
@@ -54,11 +54,41 @@ public struct ShippingLabelAddress: GeneratedCopiable, Equatable, GeneratedFakea
...
+
     public func encode(to encoder: Encoder) throws {
         var container = encoder.container(keyedBy: CodingKeys.self)
 
         try container.encode(company, forKey: .company)
-        try container.encode(name, forKey: .name)
+        // Make sure to only send address name if it's not empty,
+        // otherwise requests to fetch rates and purchase label will fail.
+        // Reference: https://git.io/JVQzC
+        if !name.isEmpty {
```
pmusolino, line 89:
```suggestion
        if name.isNotEmpty {
```

pmusolino (changes requested):
Thank you @itsmeichigo for trying to address this issue.
I tested it, but I found a regression on my side. I'm testing it on Xcode 13 and iOS 15.
It seems that after deleting both fields (name and company), the company field is no more selectable.
It could be related only to iOS 13? Or you are not experiencing this issue on your side?
Here it is a demo:

https://user-images.githubusercontent.com/495617/136358355-871caa22-b046-4c7b-a836-155cbcba015f.mp4

itsmeichigo:
@pmusolino the issue with the frozen cell is the same as #5051 - i.e all the text fields on the table view cells' `canBecomeFirstResponder` return `false` after reloading the table view a few times. I still have yet to find out why.

Can you check again when building with Xcode 12?

Networking/Networking/Model/ShippingLabel/ShippingLabelAddress.swift
itsmeichigo: Oops, seems like the extension method `isNotEmpty` is only available in WooCommerce and not Networking 😅

pmusolino (approved):
> Can you check again when building with Xcode 12?

on Xcode 12 with iOS 14 everything seems ok 🚢

PR #5150, opened 2021-10-06
Quick Pay: Add Yosemite actions

Ecarrion:
# Why 

As the Quick Pay project is about to start, this PR makes sure that we have an action available in order to create a quick pay order using custom fees as discussed in p91TBi-6aO

# How

- Update `OrderFeeLine` to add `Encodable` support.

- Added into `OrdersRemote` a function to create orders, similar to the update orders functions, here we can choose what attribute from a given order will be used to create a new order. Currently, only `fee_lines` are supported.

- Added into `OrdersStore` an action to create a quick pay order. It will take just an arbitrary amount and will create an store an order as needed.

# Testing Steps

- Modify `DashboardViewController.viewDidLoad` in order to look like 
```swift
override func viewDidLoad() {
    super.viewDidLoad()
    configureNavigation()
    configureView()
    configureDashboardUIContainer()

    DispatchQueue.main.async {
        let action = OrderAction.createQuickPayOrder(siteID: self.siteID, amount: "10.99") { result in
            print(result)
        }
    }
}
```

- Launch the app and navigate to orders
- Go to the "All Orders" tab
- See that a new Guest order is there
- Confirm the value of the order is `10.99` and that you see the `Collect Payment` button if your test store is configured to receive payments.

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Actions/OrderAction.swift
+++ b/Yosemite/Yosemite/Actions/OrderAction.swift
@@ -58,4 +58,8 @@ public enum OrderAction: Action {
     /// Updates the specified fields from an order.
     ///
     case updateOrder(siteID: Int64, order: Order, fields: [OrderUpdateField], onCompletion: (Result<Order, Error>) -> Void)
+
+    /// Creates an order with the specified fields from a given order
+    ///
+    case createOrder(siteID: Int64, order: Order, fields: [CreateOrderField], onCompletion: (Result<Order, Error>) -> Void)
 }
--- a/Yosemite/Yosemite/Actions/OrderAction.swift
+++ b/Yosemite/Yosemite/Actions/OrderAction.swift
@@ -59,7 +59,7 @@ public enum OrderAction: Action {
     ///
     case updateOrder(siteID: Int64, order: Order, fields: [OrderUpdateField], onCompletion: (Result<Order, Error>) -> Void)
 
-    /// Creates an order with the specified fields from a given order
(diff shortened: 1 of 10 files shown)


Yosemite/Yosemite/Actions/OrderAction.swift
```diff
@@ -58,4 +58,8 @@ public enum OrderAction: Action {
     /// Updates the specified fields from an order.
     ///
     case updateOrder(siteID: Int64, order: Order, fields: [OrderUpdateField], onCompletion: (Result<Order, Error>) -> Void)
+
+    /// Creates a quick pay order with a specific amount value and no tax,
```
ealeksandrov: nit:
```suggestion
    /// Creates a quick pay order with a specific amount value and no tax.
```

Yosemite/Yosemite/Stores/OrderStore.swift
```diff
@@ -238,6 +241,22 @@ private extension OrderStore {
         }
     }
 
+    /// Creates a quick pay order with a specific amount value and no tax,
```
ealeksandrov: nit:
```suggestion
    /// Creates a quick pay order with a specific amount value and no tax.
```

ealeksandrov (approved):
Looks good and works as described :shipit:

Ecarrion:
Thanks @ealeksandrov for the review and for noticing those typos!

PR #5154, opened 2021-10-07
Shipping Labels: Fix incorrect item quantity in purchased label

### LINKED ISSUE #5153: Shipping Labels: Incorrect number of items in purchase request
**Describe the bug**
The number of items listed in purchased labels seems to be incorrect on Order Details screen.

**To Reproduce**
2. Create an order with 10 items of the same product.
3. On Orders tab, select the new order and select Create Shipping Label.
4. Configure Ship From and Ship To. 
5. On Package Details screen, move some items to another package.
6. Proceed to complete other fields and purchase the labels.
7. Navigate to Order Details, notice that the number items on both labels is 1.

**Screenshots**
https://user-images.githubusercontent.com/5533851/136353451-f027c242-96f7-49ad-8807-5901d7f9b5ae.mp4

**Expected behavior**
The correct number of each item should be displayed correctly.

itsmeichigo:
Fixes #5153 
⚠️ This PR depends on #5125, so please make sure to merge that PR first ⚠️

# Description
Currently when purchasing labels we're not taking into account the quantity of each item. Therefore the purchased labels are not containing the correct number of items.

This PR updates the product ID list sent to the purchase request to make sure the list reflects correctly the quantity of each product associating with the label.

# Demo
https://user-images.githubusercontent.com/5533851/136362173-e7cd1c07-0d53-463b-931d-7fc994eb9653.mp4

# Testing
2. Create an order with 10 items of the same product.
3. On Orders tab, select the new order and select Create Shipping Label.
4. Configure Ship From and Ship To. 
5. On Package Details screen, move some items to another package.
6. Proceed to complete other fields and purchase the labels.
7. Navigate to Order Details, notice that the number of items on both labels are correct.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -811,9 +811,17 @@ extension ShippingLabelFormViewModel {
                   let details = selectedPackagesDetails[safe: index] else {
                 return nil
             }
+            let productIDs: [Int64] = {
+                var ids: [Int64] = []
+                details.items.forEach { item in
+                    let quantity = item.quantity.intValue
+                    ids.append(contentsOf: Array(repeating: item.productOrVariationID, count: quantity))
+                }
+                return ids
+            }()
             return ShippingLabelPackagePurchase(package: package,
                                                 rate: selectedRate.rate,
-                                                productIDs: details.items.map { $0.productOrVariationID },
+                                                productIDs: productIDs,
                                                 customsForm: package.customsForm)
         }
 


rachelmcr (approved):
These changes look great and I confirmed the correct item quantity purchased for the shipping label. 👍 

I found a related issue on the Order Detail screen, however, when the items are product variations instead of simple products. In that case, the first time you navigate back to the Order Detail screen it shows "0 Items" for that shipping label:

Order Detail|Items detail for shipping label
-|-
|


But if you go back to the Orders list at that point and reopen the Order Detail, the correct number of items then appears for all the shipping labels. So it's only a problem the first time you go back to the Order Detail after purchasing the label.

I can't see why that's happening, and the shipping label purchase does have the correct item quantity, so we could follow up with this issue separately.

itsmeichigo:
@rachelmcr I think that could be the case when you first install the app / switch to a new store where the product list / variation list hasn't been fetched completely. Yeah we'll need a separate issue for that.

rachelmcr:
Thanks, I opened https://github.com/woocommerce/woocommerce-ios/issues/5170 to track that issue.

Issue #5037, opened 2021-09-22
Shipping Labels: improvements to add a new payment method


pmusolino:
As suggested by @itsmeichigo here https://github.com/woocommerce/woocommerce-ios/pull/5023 it would be great to have some enhancement (especially on the UI side) on the add a new payment method in the Shipping Label.

- [ ] In portrait mode the empty view does not stay at the center of the screen. I think it's worth it to consider the optional scroll view introduced in #5009 or the ScrollableVStack mentioned in that discussion (just a suggestion, should be tested).
<img src="https://user-images.githubusercontent.com/5533851/134277555-2a68921f-5913-4819-80d5-050f3c4d371c.png" width=350 />

- [x] Should we show a loading view when syncing user account? That way we can avoid showing the empty view if it takes a while to load new data:
https://user-images.githubusercontent.com/5533851/134277722-1b87d210-fa22-4fe9-be72-64ee9c782a8c.mp4
- [ ] It would be great to select the newly added payment method as the selected method automatically. Currently it seems that the old method is still selected, I think it'd be nice if the new one gets selected instead.
- [ ] Is it necessary to disable Done button when payment method is not updated? In the case when I have no payment method, after adding a card the card is selected automatically and the Done button is disabled. I was so confused and thought something was wrong 😅. My suggestion is that we only disable the Done button when there's no method to choose from.

JorgeMucientes:
Closing this because the mobile shipping labels flow was fully revamped in 2025, and this issue predates that revamp / targets legacy shipping labels behavior. If this still reproduces in the current shipping labels flow, please open a new issue with updated steps and screenshots.

PR #5155, opened 2021-10-07
Shipping Label: display loading state when syncing user account after adding a new payment method

pmusolino:
Part of https://github.com/woocommerce/woocommerce-ios/issues/5037

## Description
In this PR, the `Done` button will display the loading indicator after adding a new payment method. Before, nothing was shown, and there were no indicators that we were waiting for a response syncing the user accounts.
Plus, I fixed a localized string that display a wrong message after adding a payment method.

## Testing
**Important testing notes:**
1) For testing this PR I suggest you disable the proxy on your Mac. More info here: pMz3w-bAI#comment-80199
2) For configuring your sandbox environment on the app, you should search for `Store Sandbox Cookie Secret` from the Secret Store, and put it under `AuthenticatedWebView`, substituting the string `[secret]`. Otherwise, you will not be able to test it with a fake credit card.
3) You can use one of the fake credit cards included in the post called `Testing the Store` under the Field Guide.

1. Open an order eligible for shipping label creation.
2. Click on Create shipping label.
3. Pass all steps until you reach the payment step.
4. Click on edit.
5. Click on Add credit card.
6. Enter the card details and save.
9. Notice that the webview will be dismissed and that the top right button will become a loader until the new account data are fetched.

## Video
https://user-images.githubusercontent.com/495617/136389202-c16de7ce-f110-47a3-81bf-8d3a9482c592.mp4

Update release notes:


rachelmcr (approved):
These fixes look good to me! :shipit:

PR #5161, opened 2021-10-07
Ignores malformed sites when decoding the a merchant's site list

Ecarrion:
# Why

It has been reported that there are some sites without options that are causing the app to fail to display the site list.

This is very inconvenient because merchants can't use their other available sites.

This PR updates the site decoding logic to evict from the site list any site that fails to be decoded.

# How

- Added a custom `decoding` method on `SiteListMapper` where we catch the error when a site decode fails.

# Testing Step

- Test on a real device
- Connect a store without options. A8C store is a test subject.
- Logout from the app (if you are stuck in the site error screen you have to delete the app because there could be an unwanted `defaultSiteID` value stored)
- Login to your account
- See that all valid stores as listed in the site picker screen

Update release notes:


koke (approved):
I don't know how to get added back to the a8c store, so I can't test this on a real device. However, the test response matches what I was seeing, and the code looks good, so as long as tests are green, I'd say 👍🏽

PR #5162, opened 2021-10-07
Order Editing: Add empty customer note state

### LINKED ISSUE #5112: Order Editing: Empty Order Note cell design
It was reported that the empty space in this row when the note is empty feels a bit unexpected.

Screen Shot 2021-10-01 at 10 23 50 AM

@adamzelinski do you think that is ok, or do you want to adjust it in some way?

ealeksandrov:
Fixes https://github.com/woocommerce/woocommerce-ios/issues/5112.

## Description

Instead of showing empty field with edit icon we show "+ Add Customer Note" button.

### How

Cell layout updated to use couple of UIStackViews. Create button is hidden in vertical stack, edit button - in horizontal stack.

## Test

1. Navigate to order with empty customer note (or edit note to have empty content).
2. Verify that add button appears (instead of edit) and it opens modal view to edit note.
3. Update note with small and large content - double check the cell layout since it was significantly updated.

## Screenshots

before | after
--|--

---
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/CustomerNoteTableViewCell.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/CustomerNoteTableViewCell.swift
@@ -8,6 +8,8 @@ final class CustomerNoteTableViewCell: UITableViewCell {
 
     @IBOutlet private weak var editButton: UIButton!
 
+    @IBOutlet private weak var addButton: UIButton!
+
     /// Headline label text
     ///
     var headline: String? {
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Customer Section/CustomerNoteTableViewCell.swift
```diff
@@ -88,9 +99,23 @@ private extension CustomerNoteTableViewCell {
         editButton.addTarget(self, action: #selector(editButtonTapped), for: .touchUpInside)
     }
 
+    func configureAddButton() {
+        addButton.applyLinkButtonStyle()
+        addButton.setImage(.plusImage, for: .normal)
+        addButton.contentHorizontalAlignment = .leading
+        addButton.contentVerticalAlignment = .bottom
```
ealeksandrov, line 106: Shifting content down to add space between cell heading and button content.

shiki:
cc @ThomazFB. Though I suspect you already know about this change. 😄

Ecarrion (dismissed):
-deleted-

Ecarrion (commented):
Looks good in general but I just spotted a small regression where while the order is loading, the edit button, the body label, and the add buttons are visible.

### Video
https://user-images.githubusercontent.com/562080/136468655-7a2e0151-e2d7-4b14-a0a2-9e9176794fe6.mov

### Screenshot
Screen Shot 2021-10-07 at 5 06 06 PM


### Before this PR

https://user-images.githubusercontent.com/562080/136469189-0a6332eb-8bcb-427d-9002-749600feb6ae.mov

ealeksandrov:
@Ecarrion thanks for noticing this glitch! I've tested with longer orders and it was fine when scrolled.

Fixed by:
- setting `body = nil` in data source
- making edit (and create) button hidden by default in xib.

Ecarrion (approved):
Code and test looked great! 🚀

Issue #4773, opened 2021-08-11
Shipping Labels: Prefill phone number for shipping address


itsmeichigo:
# Description
Currently phone number is not prefilled on the Ship To form.
However, the phone number is available in 3 places (with priority as the order):
- shipping address (as it's available in Core since version 5.6.0-rc.1).
- order metadata field `_shipping_phone`
- billing address

We should be able to retrieve the number and prefill it on the shipping address form to make it more convenience for merchants.

Ref: p1628679660021300/1628674860.018500-slack-C026HH75ECA

PR #5163, opened 2021-10-07
Shipping Labels: Get shipping phone from order metadata if it isn't part of shipping address

rachelmcr:
Part of: #4773

## Description

Currently in Shipping Labels, the phone number is not pre-filled on the Ship To form unless it is sent as part of the shipping address (available in Core since version 5.6.0).

This updates the Order networking model to try to parse the shipping phone from the order metadata (`_shipping_phone`) if the phone number isn't part of the shipping address. (Another PR will include the billing phone number as a fallback if there is no shipping phone number.)

## Changes

* Adds an `OrderMetaData` model to decode the metadata within an order (currently only parses `String` values).
* In the `Order` model, if the shipping address has a `nil` phone number it tries getting the phone number from the metadata.

## Testing

1. Make sure WCShip is installed and activated on your store, and you have at least one order eligible for shipping labels that includes a shipping address with a phone number.
2. Open the Orders tab and select the order.
3. Select "Create Shipping Label."
4. Confirm the "Ship from" address.
5. Edit the "Ship to" address and confirm the phone number is included.

If you have a store that sends the phone number as metadata on the order, please test using that store. I wasn't able to reproduce that behavior on my store but I included a unit test to confirm the phone number is parsed correctly when sent as metadata.

## Submitter Checklist

Update release notes:

### DIFF
--- a/Networking/Networking/Model/OrderMetaData.swift
+++ b/Networking/Networking/Model/OrderMetaData.swift
@@ -30,7 +30,7 @@ struct OrderMetaData: Decodable {
 ///
 private extension OrderMetaData {
     enum CodingKeys: String, CodingKey {
-        case key = "key"
-        case value = "value"
+        case key
+        case value
     }
 }
--- a/Networking/Networking/Model/OrderMetaData.swift
+++ b/Networking/Networking/Model/OrderMetaData.swift
@@ -1,0 +1,36 @@
+import Foundation
+import Codegen
+
+/// Represents the metadata within an Order
+/// Currently only handles `String` metadata values
+///
+struct OrderMetaData: Decodable {
+    public let key: String
+    public let value: String
+
+    /// OrderMetaData struct initializer.
+    ///
+    public init(key: String, value: String) {
+        self.key = key
+        self.value = value
+    }
+
+    /// The public initializer for OrderMetaData.
+    ///
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let key = try container.decode(String.self, forKey: .key)
+        let value = container.failsafeDecodeIfPresent(String.self, forKey: .value) ?? ""
+
+        self.init(key: key, value: value)
+    }
+}
+
+/// Defines all of the OrderMetaData's CodingKeys.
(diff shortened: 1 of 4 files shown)


Networking/Networking/Model/OrderMetaData.swift
```diff
@@ -0,0 +1,36 @@
...
+        self.init(key: key, value: value)
+    }
+}
+
+/// Defines all of the OrderMetaData's CodingKeys.
+///
+private extension OrderMetaData {
+    enum CodingKeys: String, CodingKey {
+        case key = "key"
+        case value = "value"
```
itsmeichigo: Pro tip: if you have an `enum` with String as raw value, you can omit the explicit declaration of the raw value if you want the raw value to be the same as the case because Swift can determine that.

Networking/NetworkingTests/Mapper/OrderMapperTests.swift
```diff
@@ -63,9 +63,21 @@ final class OrderMapperTests: XCTestCase {
...
             XCTAssertEqual(address.country, "US")
+            XCTAssertEqual(address.phone, "333-333-3333")
         }
     }
 
+    /// Verifies that Order shipping phone is parsed correctly from metadata.
+    ///
+    func test_Order_shipping_phone_is_correctly_parsed_from_metadata() {
+        guard let order = mapLoadFullyRefundedOrderResponse(), let shippingAddress = order.shippingAddress else {
+            XCTFail()
```
itsmeichigo: Nit: Maybe also leave a message in the `XCTFail` so it's easier to understand the failure if reading the error in CI?

itsmeichigo (approved):
I tested and this works perfectly - thanks so much for the enhancement! I even tested on your site (the test site with long name) and it works too, I'm not sure why it didn't work for you?

I left some nits in the comments, but overall this is good to go!

Networking/NetworkingTests/Mapper/OrderMapperTests.swift
rachelmcr: Really good point, thanks for this feedback!

PR #5166, opened 2021-10-07
[Mobile Payments] [SDK 2] Bump minimum WCPay version, update merchant messaging

### LINKED ISSUE #5077: [Mobile Payments] [SDK 2.0] Bump minimum wcpay plugin version to 3.1.0 for Locations endpoint

allendav:
Closes #5077 

Changes
- Bumps minimum version to 3.1 to pick up locations API and line2 fix - see https://github.com/Automattic/woocommerce-payments/issues/3012#issuecomment-938215347
- Updates merchant facing text, which might have been confusing for existing (beta) in-person payments users

from

The WooCommerce Payments extension is installed on your store, but the installed version does not support In-Person Payments.

to

The WooCommerce Payments extension is installed on your store, but needs to be updated for In-Person Payments.

To test:

- Downgrade your store to WooCommerce Payments 3.1 and enter the mobile app settings, In-Person Payments. Ensure you see this:

IMG_0183

- Upgrade your store to WooCommerce Payments plugin 3.1 and ensure you are not stopped at that screen but advance or order/manage card reader instead.

cc @kidinov 

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
@@ -239,6 +239,6 @@ private extension PaymentGatewayAccount {
 
 private enum Constants {
     static let pluginName = "WooCommerce Payments"
-    static let supportedWCPayVersion = "2.5"
+    static let supportedWCPayVersion = "3.1"
     static let supportedCountryCodes = ["US"]
 }
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/Onboarding Errors/InPersonPaymentsPluginNotSupportedVersionView.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/Onboarding Errors/InPersonPaymentsPluginNotSupportedVersionView.swift
@@ -36,7 +36,7 @@ private enum Localization {
     )
 
     static let message = NSLocalizedString(
-        "The WooCommerce Payments extension is installed on your store, but the installed version does not support In-Person Payments. "
+        "The WooCommerce Payments extension is installed on your store, but needs to be updated for In-Person Payments. "
             + "Please update WooCommerce Payments to the most recent version.",
         comment: "Error message when WooCommerce Payments is installed but the version is not supported"
     )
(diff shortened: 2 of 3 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
```diff
@@ -239,6 +239,6 @@ private extension PaymentGatewayAccount {
 
 private enum Constants {
     static let pluginName = "WooCommerce Payments"
-    static let supportedWCPayVersion = "2.5"
+    static let supportedWCPayVersion = "3.1"
```
kidinov, line 242: I think there are still some bugs on 3.1 and especially [this one](https://github.com/Automattic/woocommerce-payments/issues/3070) is blocking, so we might need to bump it to 3.2

koke (approved):
Tested with 3.0, then 3.1, and it works perfectly 🚀

WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
allendav: Yeah, I expect us to bump to 3.2 soon too, but at least 3.1 mostly works correctly.

allendav:
@koke - I've just now merged this into stripe-sdk-2 - just to make sure you're aware

PR #5106, opened 2021-10-01
Extract feature flag code to a separate `Experiments` framework

jaclync:
Ref p91TBi-68e-p2

## Why

To support push notifications for all stores #5032, we are adding a new [notification service app extension](https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications). In order to use the feature flag inside an app extension, having the feature flag code in a separate Xcode framework seems like the ideal solution. This PR extracts the minimum amount of feature flag code from the main app to a separate `Experiments` framework, and allows any new features in an app extension or framework to use the feature flag code.

## Changes

- Created a new Xcode framework called Experiments (unfortunately, the new project file has ~600 diffs) with documentation (`EXPERIMENTS.md`)
- Moved 4 files (`FeatureFlag`, `BuildConfiguration`, `FeatureFlagService`, `DefaultFeatureFlagService`) to Experiments framework, and marked APIs as public when needed
- Moved 3 helpers originally in `BuildConfiguration` to a new file `ProcessConfiguration` in the main app since they aren't based on build configurations
- Imported Experiments framework in the main app when needed

## Testing

I will ensure all checks below pass when the PR is approved, feel free to just review the checklist!

- [x] All tests pass locally
- [x] All tests pass on CI
- [x] All optional UI tests pass on CI
- [x] Installable build works on CI, and the feature flags are working in the installable build as expected (e.g. barcode scanner CTA is visible on the products tab)
- [x] Screenshots generation works as before (depends on https://github.com/woocommerce/woocommerce-ios/pull/5123)

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #5168, opened 2021-10-08
Add notification service extension and framework (without any logic)

jaclync:
Part of #5032 
More details in p91TBi-68e-p2

## Why

To support multi-store push notifications, we want to modify notification content from the server payload in client side so that previous app versions are not affected. To achieve this, we are using [iOS notification service extension](https://developer.apple.com/documentation/usernotifications/modifying_content_in_newly_delivered_notifications). This PR added a notification service extension **WooCommerceNotificationServiceExtension**, and an Xcode framework **NotificationServiceExtension** that contains the main logic for unit testing because XCTest is not supported in app extensions. Since diffs from project files are huge already, this PR only includes the framework setup and skeleton files without any logic to modify notification content.

NotificationServiceExtension

## Changes

📝 Apology for the big PR! Lots of diffs are on the project files from framework changes. Feel free to review commit by commit to see the important changes in project files and Info.plists. You can also check out the documentation changes first.

- Added notification service extension called `WooCommerceNotificationServiceExtension` in https://github.com/woocommerce/woocommerce-ios/commit/7dcf8fe8b63c32662d01c3cee0b953f167ed5245, it's a target under the WooCommerce project
  - This extension depends on `NotificationServiceExtension` framework, and just has one class `NotificationService: UNNotificationServiceExtension`. Even though the framework below also has the same subclass where our logic will reside, we can't directly use the class from the framework in this app extension. I tried setting `NotificationServiceExtension.NotificationService` in the Info.plist's `NSExtensionPrincipalClass` field but it does not work (the code in the framework won't be called)
- Added an Xcode framework called `NotificationServiceExtension` in https://github.com/woocommerce/woocommerce-ios/commit/b80997c24d3b109e0ee1d7e961d2d5b27af9b37a, which `WooCommerceNotificationServiceExtension` depends on.
  - This framework depends on `Experiments` framework from https://github.com/woocommerce/woocommerce-ios/pull/5106, and it tests target depends on `TestKit`
- Added/updated documentation

## Testing

Prerequisites: the logged-in user has more than one site connected to the app.
If you are testing with your a8c WP.com account, you can resend previous new order/review notifications from the **MC Push Notifications Debug Console** (MC > Push Notifications > Debug Console > WP.com Notification, or with this path `/mobile/push-notifications/debug/?type=wpcom-notification`).

Let's make sure the push notifications work as before:

- Launch the app from PR branch
- Place an order on any connected site (you can also resend a notification in the MC Debug Console if you're using a8c WP.com account) --> a new order push notification should show up, and the content is the same as before (`You have a new order! 🎉`)
- Leave a review on any connected site --> a new review push notification should show up, and the content is the same as before (`{{customer name}} left a review on {{product name}}`)

Update release notes:


jaclync:
Closing this PR now that we implemented the notification content in the server side!

PR #5173, opened 2021-10-08
Shipping Labels: Use non-digit-only keyboard for international post code field

### LINKED ISSUE #5157: [Shipping Labels] Editing international Ship To address uses numeric keypad on Postcode field
**Describe the bug**
The edit address screen in the Shipping Label purchase flow shows the numeric keyboard when editing `Postcode`. The numeric keyboard does not allow entry of alphabetical characters, which are part of many international postcodes. This means that the postcode cannot be corrected without leaving the Shipping Label purchase flow

**To Reproduce**
1. Go to Orders
2. Open an international order, e.g. to the UK, with an incorrect address
3. Tap `Continue` through the `Ship from` screen, making any required corrections to progress
4. On the `Ship to` screen, tap in the postcode field
5. Try to enter or edit letters – this is not possible.

**Screenshots**

https://user-images.githubusercontent.com/2472348/136404649-3a333d91-bfa9-4e9a-8da8-df291c10a001.mp4

**Expected behavior**
An alpha-numeric keyboard should be shown, so that the full postcode can be edited.

**Mobile Environment**
Please include:
- Device: iPhone 7 Plus
- iOS version: iOS 14.8

itsmeichigo:
Fixes #5157 

# Description
We're currently using digit keyboard for post code field - but in some country post code can also contain letters.

This PR updates the keyboard to only force digit post code for US address; otherwise use a generic keyboard with both digits and letters.

# Testing
2. Open an international order, e.g. to the UK, without post code in the billing & shipping address.
3. Tap `Continue` through the `Ship from` screen, making any required corrections to progress.
4. On the `Ship to` screen, tap in the postcode field.
5. Notice that the keyboard showing up contains both digits and letters.
6. Repeat the same steps with an US order, notice that the post code keyboard is digit only.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
@@ -450,7 +450,7 @@ private extension ShippingLabelAddressFormViewController {
                                                                      text: viewModel.address?.postcode,
                                                                      placeholder: Localization.postcodeFieldPlaceholder,
                                                                      state: .normal,
-                                                                     keyboardType: .phonePad,
+                                                                     keyboardType: viewModel.isUSAddress ? .phonePad : .namePhonePad,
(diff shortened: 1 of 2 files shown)


pmusolino (approved):
:shipit: looks good to me.
It could be great in the future, to improve it, implementing a list of countries with just numeric phones (like Italy).

PR #5177, opened 2021-10-08
Shipping Labels: Use the order billing phone as a fallback if there is no shipping phone

### LINKED ISSUE #4773: Shipping Labels: Prefill phone number for shipping address

rachelmcr:
Fixes: #4773

## Description

When purchasing a shipping label, we want to use the phone number from the customer's billing address as a fallback in the "Ship to" address if there is no shipping address phone number. (This aligns with [the plugin's web behavior](https://github.com/Automattic/woocommerce-services/blob/2028479c4b2c75e09258843dd8563b0d60dc152a/woocommerce-services.php#L1558-L1565).)

## Changes

When initializing `ShippingLabelFormViewModel`, we were taking the destination address and using `fromAddressToShippingLabelAddress` to convert it from an `Address` to a `ShippingLabelAddress`. This PR adds a method `getDestinationAddress` that checks if there is a phone number and, if there isn't one, adds the phone number from `order.billingAddress` before converting the address to a `ShippingLabelAddress`.

Before|After
-|-
Simulator Screen Shot - iPhone 12 Pro - 2021-10-08 at 15 28 02|Simulator Screen Shot - iPhone 12 Pro - 2021-10-08 at 15 25 36

## Testing

1. Make sure you have WCShip installed and activated on your store.
2. Create an order with a phone number in the billing address, and **no** phone number in the shipping address.
3. Go to Orders and select that order.
4. Select "Create Shipping Label."
5. Confirm the "Ship from" address and then edit the "Ship to" address.
6. In the "Ship to" address form, confirm the billing address phone number appears.

## Submitter Checklist

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 - [*] Fix: Navigation bar buttons are now consistently pink on iOS 15. [https://github.com/woocommerce/woocommerce-ios/pull/5139]
 - [*] Fix incorrect info banner color and signature option spacing on Carrier and Rates screen. [https://github.com/woocommerce/woocommerce-ios/pull/5144]
 - [x] Fix an error where merchants were unable to connect to valid stores when they have other stores with corrupted information https://github.com/woocommerce/woocommerce-ios/pull/5161
+- [*] Shipping Labels: The shipping address now prefills the phone number from the billing address if a shipping phone number is not available. [https://github.com/woocommerce/woocommerce-ios/pull/5177]
 
 7.7
 -----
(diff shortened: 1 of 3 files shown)


itsmeichigo (approved):
✅ Code looks good!
✅ Tested with both cases: when shipping address phone is present and when not.
Thanks for fixing this! :shipit:

Issue #5108, opened 2021-10-01
Use formatters in the Yosemite target, e.g. on receipts


joshheald:
## Description
For user display of currency and other numbers, we use some shared formatters, e.g. [`CurrencyFormatter`](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/Tools/Currency/CurrencyFormatter.swift) which currently sit in the `WooCommerce` target, allowing them to be used in any UI code.

[`Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift`](https://github.com/woocommerce/woocommerce-ios/blob/develop/Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift) also produces user-displayable numbers and currency, but cannot access `CurrencyFormatter` at present, which means we are using a slightly jarring `10.05 USD` style format on receipts at present.

Another potential future use of these helpers outside of the main target would be any extensions we chose to add to the app, e.g. widgets, which would likely also need to use the same toolkit, but would not have access to code in the main target directly.

To resolve this, we could move the existing formatters to a shared library which could be used by the `WooCommerce`, `Hardware`, and any other targets we need it for in future.

joshheald:
Done in #7016

PR #5178, opened 2021-10-08
4439 Add total fees to In Person Payment receipts

### LINKED ISSUE #4439: [Mobile Payments] Show order fees, if any, on receipts

joshheald:
Resolves #4439 

## Description
Fees are used by some WC stores to cover various non-product services. They aren't really exposed in the core UI, and seem to generally be applied using plug ins, but they form part of the order total and when they're present it's important that we show them on the receipts so that the grand total makes sense.

## Context for review
The handling of numbers here and inability to use `CurrencyFormatter` for the receipts is a known issue that we're going to resolve. #5108 

Additionally, I'm looking in to improvements in how we might interpret numbers from the API (which often come to us as strings) in order to use them with more confidence for logic in the app. Any thoughts on this are very much welcome (saves me some time!) but I'll be looking to improve it in a future change rather than this PR.

## Testing

There are some plugins which can add fees, but the most direct route I found was to add the following to `functions.php` or in a snippet on your store:
```
add_action('woocommerce_cart_calculate_fees', function() {
	WC()->cart->add_fee('Extra-careful handling', 5);
});
```

This will add a $5 fee (for a US store) to every order.

1. Add the above snippet to a store
2. Checkout an In-Person Payment order
3. Go to the Orders tab
4. Open an unpaid "Pay on Delivery" order with fees
5. Take card payment
6. Tap "Print Receipt"
7. Observe that the order lines include a "Fees" line.
8. Repeat with an order that has no fees, observe that the "Fees" line is not shown.

Fees

Update release notes:

### DIFF
--- a/Yosemite/YosemiteTests/Stores/ReceiptStoreTests.swift
+++ b/Yosemite/YosemiteTests/Stores/ReceiptStoreTests.swift
@@ -270,6 +270,67 @@ final class ReceiptStoreTests: XCTestCase {
 
         XCTAssert(actualDiscountLine.description.contains("(1off, AVQW112)"))
     }
+
+    func test_print_callsPrint_passing_Fees() throws {
+        let mockParameters = try XCTUnwrap(MockPaymentIntent.mock().receiptParameters())
+        let fees = [makeFee(amount: "5.00"), makeFee(amount: "15.50")]
+        let mockOrder = makeOrder(fees: fees)
+
+        let receiptStore = ReceiptStore(dispatcher: dispatcher,
+                                        storageManager: storageManager,
+                                        network: network,
+                                        receiptPrinterService: receiptPrinterService,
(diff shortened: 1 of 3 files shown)


Yosemite/YosemiteTests/Stores/ReceiptStoreTests.swift
```diff
@@ -270,6 +270,67 @@ final class ReceiptStoreTests: XCTestCase {
 
         XCTAssert(actualDiscountLine.description.contains("(1off, AVQW112)"))
     }
+
+    func test_print_callsPrint_passing_Fees() throws {
```
ctarda, line 274: 10_000 extra points for adding tests

ctarda (approved):
Tests perfect, and the changes look 💯 

This is the result of capturing a payment in an order with two fees: 14.55 + 0.45



:shipit:

Issue #705, opened 2019-02-21
Hide shipping address for virtual products


nbradbury:
Right now we show the shipping address even for virtual products. We should hide the shipping address in this situation.

[Related Android issue](https://github.com/woocommerce/woocommerce-android/issues/826)

mindgraffiti:
@kyleaparker I am not sure what the expected behavior is here. On iOS, if the Shipping Address is blank, "No address specified." is the message displayed underneath the Shipping details heading. It looks like this:

<img src="https://user-images.githubusercontent.com/1062444/54386177-6238aa00-4666-11e9-822c-9c0482323f90.PNG" width="350" />

Android has a related task filed to show the same thing: https://github.com/woocommerce/woocommerce-android/issues/298.

If you mean "don't display the Customer Info section at all for digital / download-only orders", there isn't an explicit way to identify those. We don't have any properties that let us know that the order was for a virtual product. 

cc: @nbradbury, in case I missed something

This is what an order object response looks like for a digital download order:

```
{
    "id": 2198,
    "parent_id": 0,
    "number": "2198",
    "order_key": "wc_order_XDNP86jchb3oL",
    "created_via": "checkout",
    "version": "3.5.5",
    "status": "completed",
    "currency": "USD",
    "date_created": "2019-03-14T15:25:50",
    "date_created_gmt": "2019-03-14T19:25:50",
    "date_modified": "2019-03-14T15:25:53",
    "date_modified_gmt": "2019-03-14T19:25:53",
    "discount_total": "0.00",
    "discount_tax": "0.00",
    "shipping_total": "0.00",
    "shipping_tax": "0.00",
    "cart_tax": "1.78",
    "total": "21.78",
    "total_tax": "1.78",
    "prices_include_tax": false,
    "customer_id": 12,
    "customer_ip_address": "50.27.182.89",
    "customer_user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_2) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/72.0.3626.121 Safari/537.36",
    "customer_note": "",
    "billing": {
        "first_name": "ArtDesign",
        "last_name": "Love",
        "company": "Cooper Hewitt Smithsonian Design Museum",
        "address_1": "2 E 91st St",
        "address_2": "",
        "city": "New York",
        "state": "NY",
        "postcode": "10128",
        "country": "US",
        "email": "thuy.copeland@gmail.com",
[...]

Issue #4787, opened 2021-08-11
Order Editing: Review Accessibility


Ecarrion:
Let's make sure all of our screens are accessible.

- [x] Check Dark mode
- [x] Check big fonts
- [x] Check landscape (Screens to go edge to edge)
- [x] Check iPad

Ecarrion:
Current State for address picker in dark-mode + big font
<img width="382" alt="Screen Shot 2021-09-02 at 7 15 25 PM" src="https://user-images.githubusercontent.com/562080/132008233-72bb07c4-d42a-4d5d-abd9-91f7373ece88.png">

ealeksandrov:
"+" buttons issue is non-straightforward to resolve, created https://github.com/woocommerce/woocommerce-ios/issues/5335 for backlog.
Address Form fixed in https://github.com/woocommerce/woocommerce-ios/issues/5334.

PR #5180, opened 2021-10-08
Order Editing: Add empty shipping address state

### LINKED ISSUE #5114: Order Editing: Show empty row for shipping address
The iOS app does not show a row for empty shipping addresses, which removes the ability to the merchant to edit it.

@adamzelinski, should we add an empty row for it? How do you envision it?

### Current
Screen Shot 2021-10-01 at 10 53 04 AM

ealeksandrov:
Fixes https://github.com/woocommerce/woocommerce-ios/issues/5114.

## Description

Instead of showing "No address specified" with edit icon we show "+ Add Shipping Address" button.

### How

Create button is added in vertical stack, just like on `CustomerNoteTableViewCell`.

Initial issue (https://github.com/woocommerce/woocommerce-ios/issues/5114) suggests that app hides empty address and we need to display it first. I found that shipping address row is hidden when order contains only virtual products:
https://github.com/woocommerce/woocommerce-ios/blob/9c10f58bbfea741d530a751e002f9fb048838f95/WooCommerce/Classes/ViewModels/Order%20Details/OrderDetailsDataSource.swift#L962-L968

Related discussion:
- https://github.com/woocommerce/woocommerce-ios/issues/705

**Question:** should we change this logic for Quick Pay flow? To allow adding shipping address on order without products. We can differentiate "only virtual products" and "no products at all" cases.

BTW, I haven't found a way to have `nil` address. It's coming as address with all empty fields even for virtual products without shipping.

So here I've implemented different case - when shipping address is empty for a order with non-virtual products. You can do it by clearing all fields in wp-admin:
Screenshot

One more thing:
I've noticed strange glitch with `UIButton.distributeTitleAndImage` helper. Small improvement inspired by [this SO answer](https://stackoverflow.com/a/25559946/840588) helped. Check second set of screenshots for comparison.

## Test

1. Navigate to order with empty shipping address (or edit shipping address to have empty content from wp-admin).
2. Verify that add button appears (instead of edit) and it opens modal view to edit address.
3. Add some address fields, save it and verify that order details view changes with new content.

## Screenshots

before | after
--|--

before UIButton alignment fix | after UIButton alignment fix
--|--

---
Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UIButton+TitleAndImage.swift
+++ b/WooCommerce/Classes/Extensions/UIButton+TitleAndImage.swift
@@ -10,11 +10,13 @@ extension UIButton {
 
         switch layoutDirection {
         case .rightToLeft:
-            imageEdgeInsets = UIEdgeInsets(top: imageEdgeInsets.top, left: insetAmount, bottom: imageEdgeInsets.bottom, right: imageEdgeInsets.right)
-            titleEdgeInsets = UIEdgeInsets(top: titleEdgeInsets.top, left: titleEdgeInsets.left, bottom: titleEdgeInsets.bottom, right: insetAmount)
+            imageEdgeInsets = UIEdgeInsets(top: imageEdgeInsets.top, left: insetAmount, bottom: imageEdgeInsets.bottom, right: -insetAmount)
+            titleEdgeInsets = UIEdgeInsets(top: titleEdgeInsets.top, left: -insetAmount, bottom: titleEdgeInsets.bottom, right: insetAmount)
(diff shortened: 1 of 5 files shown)


Ecarrion (approved):
Looks good to me!

> Question: should we change this logic for Quick Pay flow? To allow adding shipping address on order without products. We can differentiate "only virtual products" and "no products at all" cases.

I don't see a reason for showing the shipping address row to an order that won't be shipped because it doesn't have products. will confirm.

----

I find a small glitch when using big fonts but feel free to tackle that later on #4787 

Screen Shot 2021-10-08 at 2 58 48 PM

ealeksandrov:
Thanks for review @Ecarrion! I'll keep in mind that dynamic type issue.

PR #5185, opened 2021-10-11
Shipping Labels: Only set limit to number of retries for failed requests of purchase status

### LINKED ISSUE #5169: Shipping Labels: Improve logic for purchasing labels
The current logic for purchasing labels is making polling requests with at most 3 retries. The purchase is assumed to fail if not getting response with successful status of the purchase after these 3 times.

We should update this naive logic to better determine the correct status of the purchase. We can use the approach introduced in the web client: [here](https://github.com/Automattic/woocommerce-services/blob/4f1b0e1bd8cba71608e4cfacf5716b69967b9183/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js#L993-L1018) and [here](https://github.com/Automattic/woocommerce-services/blob/4f1b0e1bd8cba71608e4cfacf5716b69967b9183/client/extensions/woocommerce/woocommerce-services/state/shipping-label/actions.js#L786-L805).

Explanation of the logic: p91TBi-4K3-p2
Internal discussion: p1633624095107900-slack-C6H8C3G23

itsmeichigo:
Closes #5169 

# Description
As discussed in p1633624095107900-slack-C6H8C3G23, we should only limit the number of polling requests when fail to fetch purchase status of shipping labels. In cases when the requests succeed and status of all labels are not all successful yet, we need to keep the polling requests going.

This PR makes sure that the limit of retries applies only to failed requests:
- Renames `maxRetries` parameter of `pollLabelStatus` method to `maxErrorRetries` to clarify the intention.
- Removes the check for number of retries when a purchase status request succeeds.

# Testing
2. On Orders tab, select an order that's eligible for creating shipping label.
3. Select Create Shipping Label and input information for all rows in the purchase form.
4. Proceed to purchase the label. Notice that the no error notice is displayed when the purchase succeeds (this may be hard to test if the backend responds with success status quickly; so I have also updated the unit test to ensure that we don't return error if the purchase status is in progress).

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/ShippingLabelStore.swift
+++ b/Yosemite/Yosemite/Stores/ShippingLabelStore.swift
@@ -257,9 +257,9 @@ private extension ShippingLabelStore {
                     guard let self = self else { return }
 
                     // Poll the status of the label purchases from the response above
-                    // with a delay of 1 second each time, with a maximum of 3 retries
+                    // with a delay of 1 second each time, with a maximum of 3 retries for error
                     self.pollLabelStatus(withDelayInSeconds: 1.0,
-                                         maxRetries: 3,
+                                         maxErrorRetries: 3,
                                          siteID: siteID,
                                          orderID: orderID,
                                          labelIDs: labelPurchaseIDs,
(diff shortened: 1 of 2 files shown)


pmusolino (approved):
:shipit:

PR #5186, opened 2021-10-11
Shipping Labels: Fix incorrect total weight for locales with comma as decimal point

### LINKED ISSUE #5184: Shipping Labels: Invalid total weight for package for locales with comma for decimal point
**Describe the bug**
When running the app in locales with comma for decimal points (e.g Italy), total weight displayed on the Package Details row of the purchase form cannot parse the weight so it displays the weight as 0.

**To Reproduce**
2. Set locale for your device to a region that uses comma for decimal point (e.g Italy).
3. Select an order eligible for creating shipping labels.
4. Select Create Shipping Label, configure Ship From and Ship To.
5. In Package Details, update the total weight of the package to a decimal value and tap Done. Notice that the total weight displayed on Package Details row is 0.

**Expected behavior**
The total weight should be correct regardless of locale.

itsmeichigo:
Fixes #5184 

# Description
Currently there are two places in `ShippingLabelsFormViewModel` where we're converting from string to double without taking locale into consideration:
- When calculating total weight of all packages to display on Package Details row.
- When getting the total weight of each package to set to `ShippingLabelPackageSelected` for fetching carriers and rates.

This causes issue with incorrect weights displayed and passed to carrier list requests when running the app with locales that use comma as decimal point. This PR fixes that by using `NumberFormatter` to convert string to double values instead of the naive way of converting to Double directly.

# Testing
2. Set locale for your device to a region that uses comma for decimal point (e.g Italy).
3. Select an order eligible for creating shipping labels.
4. Select Create Shipping Label, configure Ship From and Ship To.
5. In Package Details, update the total weight of the package to a decimal value and tap Done. Notice that the total weight displayed on Package Details row is correct.
6. Proceed to select Carriers and Rates. Notice that the list can be fetched properly with appropriate package size and weight.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/ShippingLabelFormViewModel.swift
@@ -57,7 +57,7 @@ final class ShippingLabelFormViewModel {
         }
 
         return selectedPackagesDetails.compactMap { package -> ShippingLabelPackageSelected? in
-            let weight = Double(package.totalWeight) ?? .zero
+            let weight = NumberFormatter.double(from: package.totalWeight) ?? .zero
             let customsForm = customsForms.first(where: { $0.packageID == package.id })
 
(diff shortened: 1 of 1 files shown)


pmusolino (approved):
:shipit:

PR #5188, opened 2021-10-11
Shipping Labels: display the discounted price rate in Carrier and Rates screen

### LINKED ISSUE #5156: Shipping Label: display the discounted price rate in Carrier and Rates screen
In Carriers and Rates, just on iOS, we display the `retail_rate` instead of the discounted price `rate`. This creates misalignment problems between iOS and Android. 
Showing the discounted rate makes the most sense because the merchant may be trying to make a decision on how much cost they're going to eat on shipping (if they collect a flat shipping fee of $5, they may be picking a rate close to that amount).

Original [discussion here](https://a8c.slack.com/archives/C026HH75ECA/p1633512595040600).

pmusolino:
Fixes #5156 

## Description
Previously, in Carriers and Rates, just on iOS, we were displaying the `retail_rate` instead of the discounted price rate. This creates misalignment problems between iOS and Android.

## Testing
1. Make sure that your test store has configured the WCShip plugin.
2. Select an order eligible for creating shipping labels.
3. Select Create Shipping Label, configure Ship From and Ship To.
5. Select a package.
6. Proceed to select Carriers and Rates. Notice that the price listed, it's the discounted price (if available) and not the retail rate. The prices between Android and iOS should be aligned.

Update release notes:


itsmeichigo (approved):
LGTM :shipit:

PR #5179, opened 2021-10-08
Quick Pay: Add Quick Pay View Prototype

Ecarrion:
# Why

This PR adds the view to enter an arbitrary amount for the quick pay prototype.

# How

- Added SwiftUI `QuickPayAmount` view.

# Screenshots

Normal | Dark | Big Font
--- | --- | ---
portrait | darkmode | portrait-big

Landscape | Landscape Big Font
--- | --- 
lanscape | landscape-big

**Note:** Accessibility does not work correctly in all orientations and sizes, but as the fix is not trivial and this is a prototype, I'll leave that for later.

# Testing Steps

- Modify `DashboardViewController.viewDidLoad` to look like
```swift
override func viewDidLoad() {
    super.viewDidLoad()
    configureNavigation()
    configureView()
    configureDashboardUIContainer()

    DispatchQueue.main.asyncAfter(deadline: .now()) {
        let vc = QuickPayAmountHostingController()
        let nav = WooNavigationController(rootViewController: vc)
        self.present(nav, animated: true)
    }
}
```
- Launch the app
- See the view presented.

# Next

- In the next PR I'll be working on adding the view model support & making sure the amount added is properly formatted.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #5189, opened 2021-10-11
Quick Pay: Launch QuickPay from orders list & format amount

Ecarrion:
# Why

Following the QuickPay prototype development, after adding the plain view in #5179, this PR makes sure the view can be launch from the orders list and adds some basic formatting to the amount provided by the merchant.

# How

- Launch `QuickPayAmountHostingController` from the orders list using a navigation bar item behind a feature flag.

- Adds a `QuickPayAmountViewModel` to prevent merchants to enter an invalid amount.

# Demo

https://user-images.githubusercontent.com/562080/136805755-830ef980-76dc-405f-a10a-8ad82c39aa21.mov

# Testing steps

- Launch the app and navigate to the orders list
- See that you can't enter an invalid amount or more than two decimals

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
@@ -193,10 +193,14 @@ extension OrdersTabbedViewController {
     /// Create a `UIBarButtonItem` to be used as a way to create a new quick pay order.
     ///
     func createAddQuickPayOrderItem() -> UIBarButtonItem {
-        UIBarButtonItem(image: .plusBarButtonItemImage,
-                        style: .plain,
-                        target: self,
-                        action: #selector(presentQuickPayAmountController))
+        let button = UIBarButtonItem(image: .plusBarButtonItemImage,
+                                     style: .plain,
+                                     target: self,
+                                     action: #selector(presentQuickPayAmountController))
+        button.accessibilityTraits = .button
+        button.accessibilityLabel = NSLocalizedString("Add quick pay order", comment: "Navigates to a screen to create a quick pay order")
+        button.accessibilityIdentifier = "quick-pay-add-button"
+        return button
     }
 
     /// Creates the view controllers to be shown in tabs.
--- a/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
@@ -85,6 +85,14 @@ final class OrdersTabbedViewController: ButtonBarPagerTabStripViewController {
         present(navigationController, animated: true, completion: nil)
     }
 
+    /// Presents `QuickPayAmountHostingController`.
+    ///
+    @objc private func presentQuickPayAmountController() {
+        let viewController = QuickPayAmountHostingController(viewModel: QuickPayAmountViewModel())
+        let navigationController = WooNavigationController(rootViewController: viewController)
+        present(navigationController, animated: true)
+    }
+
     // MARK: - ButtonBarPagerTabStripViewController Conformance
 
     /// Return the ViewControllers for "Processing" and "All Orders".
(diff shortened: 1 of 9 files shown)


WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
```diff
@@ -182,6 +190,15 @@ extension OrdersTabbedViewController {
...
     }
 
+    /// Create a `UIBarButtonItem` to be used as a way to create a new quick pay order.
+    ///
+    func createAddQuickPayOrderItem() -> UIBarButtonItem {
+        UIBarButtonItem(image: .plusBarButtonItemImage,
+                        style: .plain,
+                        target: self,
+                        action: #selector(presentQuickPayAmountController))
+    }
```
ealeksandrov, line 204: nit: will later need accessibility label, since it's an icon without text

WooCommerce/Classes/ViewRelated/Orders/QuickPay/QuickPayAmountViewModel.swift
```diff
@@ -0,0 +1,49 @@
...
+    func formatAmount(_ amount: String) -> String {
+        guard amount.isNotEmpty else { return amount }
+
+        // Removes any unwanted character
+        var formattedAmount = amount.filter { $0.isNumber || $0.isCurrencySymbol || $0 == "." }
+
+        // Prepend the `$` symbol if needed.
+        if formattedAmount.first != "$" {
+            formattedAmount.insert("$", at: formattedAmount.startIndex)
+        }
```
ealeksandrov, line 41: Will it work with other locales/currencies? Or we'll gate it to US for now?

ealeksandrov: Question also applies to `amountPlaceholder`. I think that `amountPlaceholder` shouldn't be localized - it's confusing for translators.
Should we create it with 0 passed to NumberFormatter setup with local (or better store) currency symbol?
Have you compared it to amount labels handling in IPP?

ealeksandrov (commented):
It works in my tests and looks good in general, but I have a question: have you considered using existing `CurrencyFormatter` or some custom `NumberFormatter`? There is even `PriceInputFormatter` (but I haven't looked much into it).

I'm afraid we may miss some edge cases around currency symbol handling.
Plus adding locales later will require a lot more complicated logic.

WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
Ecarrion: Right, good catch. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5189/commits/04286021c5d56c189303b3749508706e265e9c11

WooCommerce/Classes/ViewRelated/Orders/QuickPay/QuickPayAmountViewModel.swift
Ecarrion: Yes, this prototype is only meant to work with the initial release of IPP, so only US 
But you are right that this might confuse translators, so I've taken this string out from the translation string list and added a comment https://github.com/woocommerce/woocommerce-ios/pull/5189/commits/e077574edd1e7941c297a6250ba089a5093a3997

ealeksandrov:
BTW, even US user could have a locale where decimal separator is not `.`
https://developer.apple.com/documentation/foundation/nslocale/1643064-decimalseparator

It can be annoying when copy-pasting amounts. Even in [Alfred](https://www.alfredapp.com/) calculator for myself I use this setting:
image

Ecarrion:
Hi @ealeksandrov, thanks for the review 🙇 

> have you considered using existing CurrencyFormatter or some custom NumberFormatter?

Yes, that's the first thing I tried but it didn't play well with how `SwiftUI` manages `TextViews` input.

Basically the `CurrencyFormatter` formatted formats your input the way you want it, so if the user types `3` the formatter will output `$3.00` and then the `TextView` will be updated with that content. If the user then enters 2, they will end up with `$3.002` which will be trimmed automatically.

As this is a prototype I decided to continue with a simple ad-hoc formatting function and integrate later when/if we decide to release this for more countries.

> I'm afraid we may miss some edge cases around currency symbol handling.
Plus adding locales later will require a lot more complicated logic.

True, we deff need to spend more time considering all scenarios, once we know the prototype is useful. Right now I'm prioritizing speed over perfection.

> BTW, even US user could have a locale where decimal separator is not .

Wow, another good catch, this https://github.com/woocommerce/woocommerce-ios/pull/5189/commits/98fa871307dc0bce2df2c822e58b2d9bf4736caa uses the current locale separator

**Demo with coma separator**

https://user-images.githubusercontent.com/562080/136965147-e7690953-8acf-4947-9b7b-447780b5d624.mov

ealeksandrov (approved):
Thanks for clarification, I agree with the plan!

shiki:
Excellent progress! 👏

PR #5191, opened 2021-10-11
Show items at full price on receipts

### LINKED ISSUE #5141: [Mobile Payments] Show items at full price on receipts
Found during #5127

## Description
Discounts are pre-applied to each item, both in the receipt and the order view. The order view has a "Product total" line, after which the discount line is subtracted, and the taxes/shipping included below that too, which makes sense, though the products themselves won't add up to the product total.

Order|Receipt
-|-
Screen Shot 2021-10-05 at 11 50 08|Screen Shot 2021-10-05 at 11 50 13

Web|Email
-|-
Screen Shot 2021-10-05 at 11 51 18|Screen Shot 2021-10-05 at 11 52 29

The web and email approach seems better: display the products at full price, subtotal them, then subtract the discount and add fees/tax/shipping to give the grand total.

joshheald:
Resolves #5141 

## Prerequisites
#5178 is a separate PR still awaiting review, which this was based off. I've changed the base branch to that, to avoid doubling up that PR review... but it'll need to be merged before this is.

## Description
During review of #5127, we discovered that the discounted amounts were used for each line item on an IPP receipt. This meant that it was difficult to check the totals of the receipt because the totals lines did not sum up to the grand total. In particular, this was a problem on orders with a coupon discount.

This change matches the web and email receipts, by showing the item lines at full price, then a subtotal for all the items at full price, and after that subtracting the discount from the coupon.

## Testing

1. Go to the Orders tab
2. Open an unpaid "Pay on Delivery" order with multiple lines and a percentage-off coupon applied.
3. Take card payment
4. Tap "Print Receipt"
5. Observe that the item lines show the product full price, and there is a `Subtotal` line summing them up correctly

multi-line discounted

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/ReceiptStore.swift
+++ b/Yosemite/Yosemite/Stores/ReceiptStore.swift
@@ -88,7 +88,7 @@ private extension ReceiptStore {
 
     func generateCartTotals(order: Order, parameters: CardPresentReceiptParameters) -> [ReceiptTotalLine] {
         let subtotalLines = [
-            subtotalLine(order: order),
+            productTotalLine(order: order),
             discountLine(order: order),
             lineIfNonZero(description: ReceiptContent.Localization.feesLineDescription, amount: feesLineAmount(fees: order.fees)),
             lineIfNonZero(description: ReceiptContent.Localization.shippingLineDescription, amount: order.shippingTotal),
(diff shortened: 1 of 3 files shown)


Yosemite/Yosemite/Stores/ReceiptStore.swift
```diff
@@ -81,26 +81,33 @@ private extension ReceiptStore {
...
+            lineIfNonZero(description: ReceiptContent.Localization.shippingLineDescription, amount: order.shippingTotal),
+            lineIfNonZero(description: ReceiptContent.Localization.totalTaxLineDescription, amount: order.totalTax)
+        ].compactMap { $0 }
         let totalLine = [ReceiptTotalLine(description: ReceiptContent.Localization.amountPaidLineDescription,
                                          amount: parameters.formattedAmount)]
 
         return subtotalLines + totalLine
     }
 
+    func subtotalLine(order: Order) -> ReceiptTotalLine {
```
allendav: To disambiguate from things like fees, please consider naming this function something like `itemsSubtotalLine`

allendav (approved):
Tests well. Tested with and without coupons on an order. One suggested change. :shipit:

Issue #4430, opened 2021-06-16
[Mobile Payments] No prompt to remove chip card after payment collected


allendav:
I was not prompted to remove my card after completing a card-present payment with a chip card. I expected to have been.

cc @ctarda

designsimply:
In WCiOS 7.1 beta testing, the opposite problem was reported: 

> I started the payment flow for an open order, and at first it worked fine. But when the payment was collected I see the message “Please remove card,” which doesn’t make sense when the card was simply tapped on the reader rather than inserted into it.

<sup>(internal reference: p5T066-2uU-p2#comment-9160 /hat tip @nbradbury, tested with WCiOS 7.1.0.0 on an iPad running iPadOS 14.6)</sup>

I tested and confirmed:

<img height="420" alt="Screen Shot 2021-07-19 at Jul 19 2 29 32 PM" src="https://user-images.githubusercontent.com/1119271/126223496-e97b0e2e-fe1d-467e-ab6a-d462a9e086ed.png">
<sup>Tested with WCiOS 7.1.0.1 TestFlight beta on iPad Mini 4 (WiFi) (iPad5,1) iPadOS 14.6 logged in as testuser01121 with worthwhile-seahorse.jurassic.ninja selected and running WP 5.7.2 JP 9.9 WC 5.5.1 WCPay 2.7.0 WCS staging 1.0.</sup>

koke:
Testing the 2.0 SDK feature branch, it seems a bit different but still not working right:
- When tapping, I see "Please remove card" until it starts processing the payment
- When inserting, I see "Please remove card" for a split second, then it goes back to "Reader is ready" until it starts processing payment

https://user-images.githubusercontent.com/8739/136402126-0c174532-d55b-4b30-b341-5880ea0fc75a.mov

allendav:
Yup. I was looking at this in detail yesterday. For chip cards we need to listen for a card inserted reader event and then keep a "remove your card" dialog on the screen until the corresponding card removed reader event and THEN show the success (or failure) to collect payment.

Working on a fix now. Thankfully SDK2 makes this simpler with the single delegate

allendav:
Closed by #5192

PR #5192, opened 2021-10-11
[Mobile Payments] [SDK 2] Ensure "Remove Card" prompt displays when card is left inserted in reader

allendav:
@ctarda @joshheald - now ready to review

Addresses #4430 

Changes:

StripeCardReaderService
- listens to didReportReaderEvent (added to SDK 2) to listen for chip card insertion and removal events
- keeps track of whether a chip card was inserted
- adds another promise to the chain to wait for inserted card (if any) to be removed

PaymentCaptureOrchestrator
- It used to handle displayMessage and waitingForInput exactly the same - by calling onPresentMessage with the string (which was then ignored by the VM)
- Now it routes displayMessage to a new modal to display the message (the BBPOS is US only and so English strings from the reader can safely be passed to the VM and VC)
- waitingForInput still calls onPresentMessage as before, but without the ignored string (onPresentMessage is used to ask the user to “present” their payment method by tapping, swiping or inserting it)

OrderDetailsViewModel and OrderDetailsViewController
- onClearMessage (to have user remove card) was confusing to me so I replaced it with onDisplayMessage (which will display remove card message or any other message passed to it)

CardPresentModalDisplayMessage
- displays messages from headless card readers, like “Remove Card”

Feedback welcome:
- I'm not 100% happy with the spin lock approach to the promise. Suggestions welcome, but short of using PromiseKit, I'm not sure there are any other alternatives.

Background and Other approaches considered:

- SDK2 added the didReportReaderEvent that turns out to be essential for properly managing chip card state - it tells (for chip cards) when a card is inserted and when it is removed
- But, as currently written, the CardPresentPaymentStore only forwards reader events to its clients when in the midst of collectPaymentMethod. Those events stop when the payment completion fires. Unfortunately, that means that if a chip card is used and not removed before payment completes (as can happen - as the reader will allow payment to complete with a card in it), the clients will not be informed of the removal.
- I considered expanding the availability of those events outside the scope of collectPaymentMethod by adding new actions to CardPresentPaymentStore to allow subscribing to chip card reader events (inserted, removed) and having  PaymentCaptureOrchestrator use those actions to subscribe to those events and keep track of inserted card that needs to be removed
- But the more I worked on that approach, the more I didn't like that the orchestrator and view model were having to know details about how the card reader and chip cards worked.
- So, at the end of it, I went with this approach that keeps the details of headless chip card capture as close to the reader as possible and opens up the door for displaying other messages from headless readers if we want.

Screenshots/Movies:

https://www.youtube.com/watch?v=gk-fcToyFjQ

Note: The reader will send "Remove Card" as a message for display for tap, swipe or insert, but only sends the actual "card inserted" and "card removed" events for insert (chip) flows. That's why you see it for all flows, but it only remains on the screen until the card is removed for insert flows.

Note: When multiple contactless cards were near my reader and I attempted swipe payments, I would occassionally see a "Multiple Contactless Cards" error briefly. Not sure we need to worry about this as it seems unlikely to happen in real life, but I wrote it up https://github.com/woocommerce/woocommerce-ios/issues/5193 - this has probably been happening all along, but only now (with this PR) do we display messages from the reader.

To test:
- Collect a payment with tapping a card. No changes to the flow should be seen.
- Collect a payment with swiping a card. No changes to the flow should be seen.
- Collect a payment with inserting a card. Don't remove it right away. Make sure you are prompted to remove it before processing continues. 

Update release notes:


joshheald:
> Unfortunately, that means that if a chip card is used and not removed before payment completes (as can happen - as the reader will allow payment to complete with a card in it), the clients will not be informed of the removal.

My Chipper doesn't seem to let that happen, at least not with chip cards. It sits there and beeps at me until I take the card out, with the Remove Card message showing (with your changes). Is that expected and because of these changes?

https://youtu.be/F4qe7ccDnFE

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
```diff
@@ -538,9 +538,9 @@ extension OrderDetailsViewModel {
 
         paymentOrchestrator.collectPayment(for: self.order,
                                            paymentsAccount: self.cardPresentPaymentGatewayAccounts.first,
-                                           onPresentMessage: onPresentMessage,
-                                           onClearMessage: onClearMessage,
+                                           onPresentMessage: onPresentMessage, // AKA Present Payment
```
joshheald: 🏕️ I think it would be worth renaming this to `onPresentPaymentMessage`. When I first saw it in the event switch in `PaymentCaptureOrchestrator.swift`, I had to take a little while to figure out what it meant: initially I thought it meant "present this message" rather than "a message to prompt the user to present payment".

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -704,12 +704,12 @@ private extension OrderDetailsViewController {
                                     amount: value)
 
         ServiceLocator.analytics.track(.collectPaymentTapped)
-        viewModel.collectPayment { [weak self] readerEventMessage in
+        viewModel.collectPayment { [weak self] in // onPresentMessage - aka prompt to present a card for payment
```
joshheald: Suggested change added in new commit (because it's long even though minor!)

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -161,6 +161,7 @@ private extension CardPresentPaymentStore {
                         parameters: PaymentParameters,
                         onCardReaderMessage: @escaping (CardReaderEvent) -> Void,
                         onCompletion: @escaping (Result<PaymentIntent, Error>) -> Void) {
+
```
joshheald: Not a big deal, but I thought I'd point this out in case it was accidentally added? I'd seen in the [RW style guide](https://github.com/raywenderlich/swift-style-guide/#spacing) that we try to avoid blank lines after opening braces...

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -421,6 +425,30 @@ private extension StripeCardReaderService {
...
+            guard let self = self else {
+                return
+            }
+
+            // If there is no chip card inserted, it is ok to immediatedly return. The payment method may have been swipe or tap.
+            if !self.isChipCardInserted {
+                return promise(.success(intent))
+            }
+
+            self.chipCardTimer =  Timer.scheduledTimer(withTimeInterval: 1.0, repeats: true) { [weak self] _ in
```
joshheald: ❓ What do you think about [adding some tolerance](https://www.hackingwithswift.com/articles/117/the-ultimate-guide-to-timer) to the timer here? (Search for "adding some tolerance" on the page) It can reduce the energy cost of using a timer, and I don't think we mind too much about being called back exactly* every second...

*we couldn't get it _exactly_ even if we wanted to.

joshheald: 💭 Also... this might be an old thing, because I can't find much to back it up, but I was _sure_ that it was important to invalidate timers when screens go away. 

I thought this was because the run loop keeps a strong reference to them and so they will continue to fire for the lifetime of the app. From my experimenting, they fire in the background too. It's pretty low risk here because the only way to get rid of the modal seems to be removing the card, but it might be worth coding defensively with this. What do you think?

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -224,6 +228,8 @@ extension StripeCardReaderService: CardReaderService {
         return createPaymentIntent(parameters)
             .flatMap { intent in
                 self.collectPaymentMethod(intent: intent)
+            }.flatMap { intent in
+                self.waitForInsertedCardToBeRemoved(intent: intent)
             }.flatMap { intent in
                 self.processPayment(intent: intent)
             }.eraseToAnyPublisher()
```
joshheald, line 235: I couldn't find any pre-existing tests for the service... what's the thinking behind that? Do we have any good examples of testing Combine stuff, or is it just complicated too much in this case by the lack of hardware?

joshheald (approved):
Tests well.

A few questions and suggestions for minor improvements to this and the surrounding code. Also, [one suggestion I just pushed up as a change](https://github.com/woocommerce/woocommerce-ios/pull/5192/commits/2565346fa772adc41696a9b58f18413ce9f8f89a), because that seemed the easiest way to do it. Hope you don't mind!

I'm not yet super familiar with Combine, but my thinking went similar to these comments of yours: 

> I considered expanding the availability of those events outside the scope of collectPaymentMethod by adding new actions to CardPresentPaymentStore to allow subscribing to chip card reader events (inserted, removed) and having PaymentCaptureOrchestrator use those actions to subscribe to those events and keep track of inserted card that needs to be removed
> But the more I worked on that approach, the more I didn't like that the orchestrator and view model were having to know details about how the card reader and chip cards worked.
> So, at the end of it, I went with this approach that keeps the details of headless chip card capture as close to the reader as possible and opens up the door for displaying other messages from headless readers if we want.

I think having separate publishers, potentially including a `Zip` which will only emit when both the `collectPayment` and `waitForInsertedCardToBeRemoved` publishers both emit could do it, if those weren't `Future` publishers... because `Zip` will complete if either publisher completes. Having the card reader events available would be quite nice here as it might help us to present better messages e.g. acknowledging that the customer's inserted their card, before asking them to remove it, which we don't currently do and sometimes sits on screen for a little while.

You've tried it and found that you've got concerns though, so I'm not asking you to go back on that... but it would be good to chat on Zoom at some point about why it wasn't so nice, if you have the time 😊

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -704,39 +704,44 @@ private extension OrderDetailsViewController {
...
+                    ServiceLocator.analytics.track(.collectPaymentFailed, withError: error)
+                    self.paymentAlerts.error(error: error, tryAgain: {
+                        self.retryCollectPayment()
                     })
+                case .success(let receiptParameters):
+                    ServiceLocator.analytics.track(.collectPaymentSuccess)
+                    self.syncOrderAfterPaymentCollection {
+                        self.refreshCardPresentPaymentEligibility()
+                    }
+
```
joshheald: grr. Fixing

allendav:
Thank you @joshheald ! Sorry I didn't get to this feedback today. I will jump on it tomorrow.

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
allendav: Good catch. Let's fix that. Opened https://github.com/woocommerce/woocommerce-ios/issues/5232

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
allendav: Adding broad tolerance is a very good idea, and I will consider a timeout as well as additional defensive strategy.

I'll be changing this to a Timer Publisher and will make sure we are properly covering invalidating

Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5192/commits/fdd8c45b354202c3343d904894dcc58e24d01b2a

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
allendav: Thank you. That's makes it clearer.

allendav:
@joshheald wrote:

> My Chipper doesn't seem to let that happen, at least not with chip cards. It sits there and beeps at me until I take the card out, with the Remove Card message showing (with your changes). Is that expected and because of these changes?

Yes! This PR enables exactly that keep-displaying-remove-card-until-it-is-removed behavior - that the remove card message is NOT removed from the iPhone display until the card is actually removed from the reader.

The behavior **without this PR** causes the remove card to only briefly appear before the what-kind-of-receipt-do-you-want modal appears, meanwhile the Chipper keeps beeping (and pushing "Remove Card" messages to us) until you actually remove the card.

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsViewModel.swift
allendav: It confused me at first as well. I went with `onWaitingForInput` https://github.com/woocommerce/woocommerce-ios/pull/5192/commits/c1c2defc41b8f7f71219a0f0a2c4191431689ac7 to match [this](https://stripe.dev/stripe-terminal-ios/docs/Enums/SCPPaymentStatus.html#/c:@E@SCPPaymentStatus@SCPPaymentStatusWaitingForInput)

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
allendav: Added tolerance and switched to Timer Publisher in https://github.com/woocommerce/woocommerce-ios/pull/5192/commits/55807d8791e9a71f2f398e332a6d65d6e64c2e8f

I didn't add a timeout. I do want to think about ways the timer might keep firing and the only one that comes to mind is if the reader were to be powered off or moved out of BLE range from the iPhone with the card still in it, and we should consider those as part of a more comprehensive effort.

Net net: I think invalidating on card removal should be sufficient for now, and if feedback prompts us to add a timeout (e.g. the card has been in there for 15 seconds, proceed with the payment anyways) then lets do that then (and log and track that), but I think that is unlikely based on what I've seen with other card readers.

Sound good?

joshheald: Yep, sounds good 👍

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
ctarda: @allendav @joshheald The problem is not testing Combine itself, but providing a connection token to the service: p91TBi-4Es-p2

PR #5194, opened 2021-10-12
Adds a button style that shows a progress view while the view performs a loading operation

Ecarrion:
# Why

Following the Quick Pay prototype development, after the amount is correctly formatted in #5189, now it's time to properly submit that amount and create a quick pay order.

# How

- Call the `createQuickPayOrder` action from the view model when the user taps on the done button.

- Created a new `PrimaryLoadingButtonStyle` to allow a primary button to show a loading spinner while the network request is being performed

- Update ViewModel to disable the done button while no amount has been entered.

# Demo

https://user-images.githubusercontent.com/562080/136884313-262a9c1d-1dc7-40c9-bc48-75dab1b00588.mov

# Testing Steps
- Launch the app and navigate to the order list
- Tap on the plus icon
- See that the done button is disabled
- Enter an amount 
- See that the done button gets enabled
- Tap on the done button
- See that while the network call is being performed a spinner is shown within the done button
- After the operation is done, see that the view gets dismissed.
- Navigate to the All orders tab and see that the new order has been created with the correct amount.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrdersTabbedViewController.swift
@@ -88,7 +88,7 @@ final class OrdersTabbedViewController: ButtonBarPagerTabStripViewController {
     /// Presents `QuickPayAmountHostingController`.
     ///
     @objc private func presentQuickPayAmountController() {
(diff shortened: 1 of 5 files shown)


ealeksandrov (approved):
LGTM :shipit:

Issue #5202, opened 2021-10-12
Shipping Labels: ITN dialog is presented twice


shiki:
## Describe the Bug

After dismissing the dialog about ITN, a similar dialog is presented again but dismissed right away.

## To Reproduce

1. Create a shipping label for an order with an international address.
2. In the Customs step, tap Continue. 
3. Enter a random invalid number in the ITN field. 
4. Tap on Learn more below it.

## Screenshots

https://user-images.githubusercontent.com/198826/137017359-0fc0d668-e027-4532-ba13-14714213fd60.mp4

## Expected Behavior

No other dialog should be shown.

## Isolating the Problem 

- [x] I can reproduce this bug consistently using the steps above.

## Mobile Environment

- Device: iPhone 8 Plus
- iOS version: 14.3
- WooCommerce iOS version: 7.7

itsmeichigo:
Closing this because the issue was fixed in #5207.

PR #5207, opened 2021-10-13
Shipping Labels: Replace AttributedText with button to avoid issues with htmlToAttributedString

### LINKED ISSUE #5203: Shipping Labels: Crash after tapping chevron in Custom Line 1
## Describe the Bug

When tapping on the chevron to expand Custom Line 1, the app crashes.

## To Reproduce

1. Create a shipping label for an order with an international address.
2. In the Customs step, type something in the ITN field.
3. While the ITN field is selected, tap on the chevron in the Package Content → Custom Line 1. 

## Screenshots

https://user-images.githubusercontent.com/198826/137018701-c57947d7-7d61-4dd4-a69c-299dfe0caedf.mp4

## Expected Behavior

The app should not crash.

## Isolating the Problem 

## Mobile Environment

- Device: iPhone 8 Plus
- iOS version: 14.3

itsmeichigo:
Fixes #5203 and #5202

# Description
There is a [known issue](https://github.com/woocommerce/woocommerce-ios/issues/5031) that htmlToAttributedString can cause crashes when called to render SwiftUI view, so this PR replaces `AttributedText` with a button in the `LearnMoreRow` to avoid the need to use the problematic method.

When we're ready to build with Xcode 13, we'll need to update the row with option to use the shiny new ✨ markdown syntax ✨ for iOS 15.

# Testing
2. On Orders tab, select an international order that's eligible for creating shipping label.
3. Select Create Shipping Label, confirm Ship From, Ship To, Package Details.
4. On Customs form, input a valid ITN, e.g: AES X11111111111111. 
5. Tap on the chevron in the Package Content → Custom Line 1. Make sure that app doesn't crash.
6. Tap on the Learn more about Internal Transaction Number row - make sure that the web view shows up only once. Close the view, notice that no other view is presented.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -7,6 +7,7 @@
 - [*] Fix incorrect info banner color and signature option spacing on Carrier and Rates screen. [https://github.com/woocommerce/woocommerce-ios/pull/5144]
 - [x] Fix an error where merchants were unable to connect to valid stores when they have other stores with corrupted information https://github.com/woocommerce/woocommerce-ios/pull/5161
 - [*] Shipping Labels: Fix issue with decimal values on customs form when setting the device with locales that use comma as decimal point. [https://github.com/woocommerce/woocommerce-ios/pull/5195]
+- [*] Shipping Labels: Fix crash when tapping on Learn more rows of customs form. [https://github.com/woocommerce/woocommerce-ios/pull/5207]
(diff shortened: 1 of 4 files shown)


rachelmcr (approved):
Changes look great, and I confirmed that there is no crash and the web view only appears once. :shipit: 

Super excited about getting markdown soon, too 🤩

PR #5208, opened 2021-10-13
Shipping Labels: Fix issue validating string with thousand separator

### LINKED ISSUE #5197: Shipping Labels: Formatted weights with thousands separator aren't validated
**Describe the bug**

When purchasing a shipping label, we display formatted weights in both the Package Details and Customs Forms screens. This adds a thousands separator, which prevents the weight from being correctly validated. (This occurs no matter what locale the device has or what the thousands separator is.) You have to remove the thousands separator for the weight to be validated.

**To Reproduce**

2. Set up the locale for your device to a region that uses comma as decimal point (e.g Italy).
3. Select an international order that's eligible for creating shipping labels and has at least one product that is above 1,000 units of your store's weight unit. (For example, if your store is set to grams as the weight unit, make sure the order has at least one product that is 1,000 g.)
4. Select Create Shipping Label and input Ship From and Ship To.
5. In Package Details, notice that the Total package weight is invalid. Remove the thousands separator and notice it is now valid.
6. In Customs, notice the Weight for the item above 1,000 is invalid. Remove the thousands separator and notice it is now valid.

**Screenshots**

Package Details|Customs Form
-|-
|

itsmeichigo:
Fixes #5197 

# Description
This PR updates the formatter used for getting double value from string, to make sure that it uses the same styles as the formatter that converts double values to localized strings.

# Testing
2. Set up the locale for your device to a region that uses comma as decimal point (e.g Italy).
3. Select an international order that's eligible for creating shipping labels and has at least one product that is above 1,000 units of your store's weight unit. (For example, if your store is set to grams as the weight unit, make sure the order has at least one product that is 1,000 g.)
4. Select Create Shipping Label and input Ship From and Ship To.
5. In Package Details, notice that the Total package weight is valid with the thousand separator present.
6. In Customs, notice that the Weight for the item above 1,000 is valid with the thousand separator present.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/NumberFormatter+Localized.swift
+++ b/WooCommerce/Classes/Extensions/NumberFormatter+Localized.swift
@@ -6,6 +6,12 @@ extension NumberFormatter {
     static func double(from string: String, locale: Locale = .current) -> Double? {
         let formatter = NumberFormatter()
         formatter.locale = locale
+        formatter.usesGroupingSeparator = true
+        formatter.groupingSize = 3
+        formatter.formatterBehavior = .behavior10_4
+        formatter.numberStyle = .decimal
+        formatter.generatesDecimalNumbers = true
+        formatter.roundingMode = .halfUp
         let number = formatter.number(from: string)
         return number?.doubleValue
     }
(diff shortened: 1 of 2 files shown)


rachelmcr (approved):
LGTM! Thanks for the quick fix and for adding the test 👍

PR #5210, opened 2021-10-13
Shipping Labels: Fix corrupted navigation view for SelectionList

### LINKED ISSUE #5200: Shipping Labels: Package Type options disappear in landscape
## Describe the Bug

The Package Type options are not visible when in landscape mode.

## To Reproduce

I mainly followed the initial steps in https://github.com/woocommerce/woocommerce-ios/pull/4976. 

4. In the package details section, tap on “Package Selected.”
5. Tap Create new package.
6. Tap Package Type
7. Switch to landscape mode.

## Screenshots

https://user-images.githubusercontent.com/198826/137015447-b1ed5243-d3e8-47dc-9806-6932d1db5142.mp4

## Expected Behavior

The options should be visible.

## Isolating the Problem 

## Mobile Environment

- Device: iPhone 8 Plus
- iOS version: 14.8

---

cc @pmusolino @rachelmcr @itsmeichigo

itsmeichigo:
Fixes #5200 

# Description
On devices with regular width for horizontal size class (e.g iPad, iPhone Pro Max / Plus), navigation views that embed a list or scroll view will default to use column style. This style without proper setup can mess up the UI.

Since our `SelectionList` doesn't show any view on selection, I updated it to explicitly use stack style instead.

Extra: Add missing `wooNavigationBarStyle` for Package Details, Customs, Carriers and Rates, Payment method screens to make sure all Done buttons are pink on iOS 15.

# Screenshot
| Before | After |
| ----- | ----- |
|  | 

# Testing
2. Build the fix on an iPad / iPhone with Pro Max or Plus model.
3. Select an order that's eligible for creating shipping labels, select Create Shipping Label.
4. Configure Ship From and Ship To. On Package Details, select Package Selected > Add New Package > Custom Package > Package Type. Notice that the package type list is displayed properly on iPad and on landscape mode of iPhone with Pro Max or Plus models.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriers.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Carriers and Rates/ShippingLabelCarriers.swift
@@ -75,6 +75,7 @@ struct ShippingLabelCarriers: View {
                     .disabled(!viewModel.isDoneButtonEnabled())
                 }
             }
+            .wooNavigationBarStyle()
         }
     }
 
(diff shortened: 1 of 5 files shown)


pmusolino (approved):
:shipit: Looks great. Still, it seems like something you will easily forget. Maybe we should add a note somewhere about this?

shiki (approved):
I don't know what the code does but I can confirm that this is fixed for Package Type. 😅 👏 


https://user-images.githubusercontent.com/198826/137153073-6520af55-671c-4901-ad5f-6afc1d1d6fb5.mp4

itsmeichigo:
Thanks @pmusolino for the suggestion about noting the issue - but I'm not sure if it's big enough to go to a P2 post? For now I think it's best to try to test with devices of different sizes to better catch issues early on.

PR #5217, opened 2021-10-13
Quick Pay: Use the store decimal separator when creating an order

Ecarrion:
# Why

For the quick pay prototype, I was using the user's locale decimal separator in order to format the amount provided by the merchant. 

This has a problem when the user's locale has a decimal coma separator but the store setting uses a decimal dot separator because it will interpret that coma as a thousand separator.

This PR fixes that and additionally, it grabs the currency symbol from the store settings rather than forcing a "$". Keep in mind that this prototype is still only meant to be used in the US.

# How

- When entering formatting an amount, replace locale's decimal separator occurrences with the store decimal separator.

- Use the store currency symbol to format the amount instead of hardcoding "$"

- Build the amount placeholder based on the store currency settings.

# Demo
https://user-images.githubusercontent.com/562080/137209600-6cbff0b5-947d-4cbc-9cf2-3d04b5d1d298.mov

# Testing Step

- Change your device/simulator to a region like Argentina(Don't worry this doesn't change the language)
- Create a quick pay order (note that the keyboard shows a comma separator but the app formats the amount with a dot)
- See that the order is created with the correct amount.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/QuickPay/QuickPayAmount.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/QuickPay/QuickPayAmount.swift
@@ -76,7 +76,7 @@ struct QuickPayAmount: View {
                 .secondaryBodyStyle()
 
             // Amount Textfield
-            TextField(Localization.amountPlaceholder, text: $viewModel.amount)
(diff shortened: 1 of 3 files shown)


ealeksandrov (approved):
Works well :shipit:

PR #5219, opened 2021-10-13
[Mobile Payments] Log payment failures

### LINKED ISSUE #5215: [Mobile Payments] Log payment failed errors
@ctarda observed:

> while we do log all errors in Tracks, we don’t seem to do the same to the application log (DDLog). We could add a log in OrderDetailsViewController, in the failure case right after this ServiceLocator.analytics.track(.collectPaymentFailed, withError: error)

> there will be a card_present_collect_payment_failed tracked, with the actual error embedded

allendav:
Fixes #5215 

To test:
- Complete an order with cart total ending in .05
- A payment failure will result
- Ensure it appears in debugger, e.g.:

```
2021-10-13 15:14:00.583714-0700 WooCommerce[8439:6310483] Failed to collect payment: paymentCapture(underlyingError: Hardware.UnderlyingError.paymentDeclinedByPaymentProcessorAPI)
```

FWIW, it looks like this to the user:

IMG_0192

The bundler moved the screenshots line too without need to, but I included it anyways since it will likely happen for the next person to build anyways

Update release notes:


.bundle/config
```diff
@@ -2,5 +2,5 @@
 BUNDLE_RETRY: "3"
 BUNDLE_PATH: "vendor/bundle"
 BUNDLE_JOBS: "16"
-BUNDLE_WITHOUT: "screenshots"
```
ctarda: This should not be part of the diff

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -718,6 +718,7 @@ private extension OrderDetailsViewController {
             switch result {
             case .failure(let error):
                 ServiceLocator.analytics.track(.collectPaymentFailed, withError: error)
+                DDLogError("Failed to collect payment: \(error)")
```
ctarda: Can we please log the localized description as well? This is meant to be something that would help HEs support merchants.

ctarda (commented):
The PR should be against the 7.7 frozen branch, not develop. Given that we want to log these errors to help HEs can we please make sure that what we log contains human-readable information?

There is also a modification to Bundle's config file that should not part of the diff.

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5219/commits/e6e6686fd6ffda31349bc12559edebb2ebdc1587

The tracking and the logging now looks like

```
2021-10-14 06:16:19.165621-0700 WooCommerce[8619:6415552] 🔵 Tracked card_present_collect_payment_failed, properties: [AnyHashable("error_domain"): "Hardware.CardReaderServiceError", AnyHashable("error_description"): "Hardware.CardReaderServiceError.paymentCapture(underlyingError: Hardware.UnderlyingError.paymentDeclinedByPaymentProcessorAPI)", AnyHashable("error_code"): "5", AnyHashable("blog_id"): 197907611, AnyHashable("is_wpcom_store"): false]
2021-10-14 06:16:19.165979-0700 WooCommerce[8619:6415488] Failed to collect payment: The card was declined by the payment processor - please try another means of payment
```

.bundle/config
allendav: > The bundler moved the screenshots line too without need to, but I included it anyways since it will likely happen for the next person to build anyways

Nonetheless, fixed in https://github.com/woocommerce/woocommerce-ios/pull/5219/commits/71149b104af1257f40fc33371fedea21b5ac4ab6

ctarda:
@allendav I'll close this one as the work here was merged in #5221

PR #5222, opened 2021-10-15
Track if a notification is from the currently selected site

jaclync:
Part of #5032 

## Why

For analytics on the multi-store push notifications feature p91TBi-6dd-p2, we want to know whether a notification received in the app or tapped by the user is from the selected site. This way, we can compare the app usage between the users who receive/tap notifications from a different store and those who mainly get notifications from one store. We already have site ID in the notification events, but we don't know about the site ID in the push notifications.

## Changes

in `push_notification_received` and `push_notification_alert_pressed` events, set `is_from_selected_site property` based on the currently selected site ID and notification site ID

## Testing

Prerequisites: your account has more than one store, and you have a dev sandbox to receive push notifications in debug builds.

- Launch the app from Xcode
- In web, place an order or leave a review on the **currently selected site** --> a push notification should arrive, and the console should show log like `🔵 Tracked push_notification_received, properties: [AnyHashable("is_from_selected_site"): true, ...]`
- In web, place an order or leave a review on a **different site** --> a push notification should arrive, and the console should show log like `🔵 Tracked push_notification_received, properties: [AnyHashable("is_from_selected_site"): false, ...]`
- Tap on the push notification from the **currently selected site** (the first one) --> the console should show log like `🔵 Tracked push_notification_alert_pressed, properties: [AnyHashable("is_from_selected_site"): true, ...]`
- Tap on the push notification from **the different site** (the latest one) --> the console should show log like `🔵 Tracked push_notification_alert_pressed, properties: [AnyHashable("is_from_selected_site"): false, ...]`

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Notifications/PushNotificationsManager.swift
+++ b/WooCommerce/Classes/Notifications/PushNotificationsManager.swift
@@ -472,7 +472,7 @@ private extension PushNotificationsManager {
     /// Tracks the specified Notification's Payload.
     ///
     func trackNotification(with userInfo: [AnyHashable: Any]) {
-        var properties = [String: String]()
+        var properties = [String: Any]()
 
         if let noteID = userInfo.string(forKey: APNSKey.identifier) {
             properties[AnalyticKey.identifier] = noteID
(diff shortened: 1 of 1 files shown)


pmusolino (approved):
Tested all the cases 👍 works like a charm :shipit:

PR #5223, opened 2021-10-15
Shipping Labels: Move action sheet to individual package items to fix issue displaying popovers on iPad

### LINKED ISSUE #5211: Shipping Labels: Unable to display Move Item popover on iPad
**Describe the bug**
On iPad action sheet is displayed as popover - and the popover for the Move Item action sheet cannot be displayed properly from the packages form.

**To Reproduce**
4. In the package details section, tap on "Move". Notice that the popover is displayed partially on the top of the package list.

**Screenshots**
Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2021-10-13 at 18 22 49

itsmeichigo:
Fixes #5211 

# Description
Currently we're showing a single action sheet on the Package Details screen, which works fine on iPhone but breaks on iPad since action sheet is displayed as popover on bigger screens and popovers require correct anchor points to look good.

To make sure each popover points to the correct button, we need to display them on each item row, so this PR handles a lot of changes:
- Adds unique ID for `ShippingLabelPackageItem`, which is used for item row ID - to identify the correct row that's moved (instead of based on product ID, which is not unique in a package).
- Removes action sheet from package list.
- Moves "Move" button from single package to item row. Adds action sheet to item row.
- Updates single package view model:
  - Adds unique ID for the view model using ID of the associating package for easier identifying.
  - Adds package number needed to display on action sheet.
  - Adds method to update list of action sheet buttons for each package item.
  - Adds action sheet info when creating item rows.

Several workarounds to be aware of:
- SwiftUI doesn't allow displaying more than one popover on the same screen, so before displaying one, I have to trigger a callback to parent view to dismiss any presented one.
- The action buttons need other packages' names to display options to move to other packages, so action buttons cannot be set in initializer of a single package view model. I have to update these buttons as an extra step after creating the view models. 

# Demo
https://user-images.githubusercontent.com/5533851/137424774-06d322c1-914f-429d-afbd-63296f5631ed.mp4

# Testing
2. Create an order with multiple items.
3. Build the app to an iPad. Select the order and tap Create Shipping Label.
4. Confirm Ship From and Ship To address.
5. On Package Details screen, tap Move button on any item. Notice that the popover points to the correct button. Make sure that the popover contains all options available and each option works correctly.
6. With a popover present, try tapping Move button on any other item of the same or different package. Make sure that no glitch happens.
7. Build the app again on an iPhone, make sure that everything still works properly.

Update release notes:


WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ItemToFulfillRow.swift
```diff
@@ -7,9 +7,58 @@ struct ItemToFulfillRow: View, Identifiable {
...
+                Spacer()
+                Button(action: {
+                    guard !showingMoveItemDialog else {
+                        return
+                    }
+                    switch horizontalSizeClass {
+                    case .some(.regular): // popover is displayed instead of action sheet
+                        onMoveAction()
+                        // delay to make sure that any other popover has been dismissed.
+                        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
pmusolino, line 29: I noticed that when I click "Move", there is a small delay, which seems to match this. Is all this delay necessary? Maybe `0.1` is enough?

pmusolino (approved):
Looks good and the solution works perfectly, certainly better than the solution we discussed yesterday. I only noticed a small delay problem, and I think it is due to the 0.5 seconds set, which perhaps can be reduced.

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/ItemToFulfillRow.swift
itsmeichigo: From my testing it looks like the dismissing animation of popovers takes quite a while, I think the smallest delay has to be 0.4, but I keep 0.5 to be safe for slower devices (0.5 is not even safe in older devices I'm afraid). If we try to display a popover while another dismissing animation hasn't completed, we'll get a glitch that popover doesn't show up at all. I'd say we keep 0.5 for now and wait for feedback from users.

pmusolino: Ok, it seems reasonable. 👍

PR #5225, opened 2021-10-15
Add accessibility announcement when offline banner appears or disappears

### LINKED ISSUE #5164: [Accessibility] Draw attention of VoiceOver users to offline mode changes
The display of an offline banner is a significant event which a sighted user will pick up on easily, but with its position at the bottom of the screen it could be a while before a VoiceOver user notices it, and lead to some frustration.

We should consider posting [`UIAccessibilityLayoutChangedNotification`](https://developer.apple.com/documentation/uikit/uiaccessibilitylayoutchangednotification) which would allow us to draw the user's attention to a banner being displayed. 

This may also need a brief state transition of the banner to a "back online" state when connectivity is returned, so we have something to draw their attention to say everything's sorted again... or we could just use the `ScreenChanged` notification in that case.

itsmeichigo:
Fixes #5164 

# Description
Since offline banner is at the bottom of the screen, offline mode may be hard to noticed for users using voice over. This PR adds accessibility notification of type announcement to notify users of connectivity status.

# Testing
- Build the changes to a physical device.
- Turn on Voice Over on the device.
- Simulate network connectivity changes (e.g using airplane mode), notice that network state is correctly announced.

Update release notes:


joshheald (approved):
Looks good, tests well :shipit: 

TIL that accessibility notifications are a thing 👍

PR #5228, opened 2021-10-15
Shipping Labels: Update contact button in the warning banner of Ship To form with options to email or message customers

### LINKED ISSUE #4791: Shipping Labels: Update contact button in the warning banner of Ship To form with options to email or message customers
# Description
Currently when a destination address cannot be verified, we're showing a banner with option to contact customer via phone call. But this should not be the only option as @designsimply suggested that we should also let merchants send emails if customer's email address is available. 

@adamzelinski has put together a design for this:

ios-label-edit-invalid-address

Internal reference: p5T066-2wJ-p2#comment-9334

rachelmcr:
Fixes: #4791

## Description

When a destination address for shipping labels cannot be verified, we're showing a banner with the option to contact customer. This PR updates the "Contact Customer" action so the merchant can email, call, or message the customer (if email and phone are available) instead of just calling them.

## Changes

* The customer's email address comes from their billing address (there is no shipping email), so the billing email is passed from the Shipping Label Form to the Shipping Label Address Form, where it's used for the banner when editing the Ship To (destination) address.
* The banner now has an action sheet with options to Email, Call, or Message the customer.
   * As before, if there is no phone or email the option to Contact Customer will not appear in the banner.
   * If there is only a phone number, the action sheet will only show the Call or Message options.
   * If there is only an email address, the action sheet will only show the Email option.
* The banner factory only needs to know if there is a phone/email, so it was updated to accept `Bool` parameters `hasPhone` and `hasEmail` and uses those to determine what banner text and actions to display.
* Added unit tests for the `ShippingLabelAddressTopBannerFactory`, to check that the correct number of actions appear on the banner.

Phone & Email|Phone only|Email only
-|-|-
IMG_7850|IMG_7852|IMG_7851

## Testing

1. Make sure your test store has the WCShip plugin installed and activated.
2. Create an order that is eligible for shipping labels, with an invalid shipping address, a billing email address, and a billing or shipping phone number.
3. Build the app on a real device with the Mail app (so you can test the contact options).
4. Select the order and tap Create Shipping Label.
5. Confirm the Ship From address.
6. Select the Ship To section, and confirm it opens the address edit form with the warning banner at the top of the screen.
7. Tap the "Contact Customer" button and confirm an action sheet opens with the expected contact options.
8. Select each contact option and confirm each one works as expected (giving you the option to call the phone number, send a text message to the phone number, or email the email address).
9. Remove the phone number, tap Done, and confirm that the Contact Customer action sheet now only shows the Email option.
10. Remove the email address from the order (this must be done in wp-admin) or select an order with no billing email. Go back to the order detail, pull to refresh the order details, and repeat steps 4-8 above, confirming that no Email option appears.

## Submitter Checklist

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressTopBannerFactory.swift
```diff
@@ -5,25 +5,25 @@ import Yosemite
 ///
 final class ShippingLabelAddressTopBannerFactory {
     static func addressErrorTopBannerView(shipType: ShipType,
-                                          phoneNumber: String?,
+                                          hasPhone: Bool,
+                                          hasEmail: Bool,
```
itsmeichigo: Since the banner factory doesn't need to handle separate cases for missing phone or email, maybe we can use one parameter `hasContactInfo` instead of these two? I think that will clarify the intention better - WDYT?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
```diff
@@ -32,10 +36,34 @@ final class ShippingLabelAddressFormViewController: UIViewController {
...
+                    }
+                }
+                actionSheet.addDefaultActionWithTitle(Localization.contactActionMessage) { _ in
+                    ServiceLocator.messageComposerPresenter.presentIfPossible(from: self, recipient: phoneNumber)
+                }
+            }
+
+            let popoverController = actionSheet.popoverPresentationController
+            popoverController?.sourceView = self.view
+            popoverController?.sourceRect = self.view.bounds
```
itsmeichigo: Maybe we should set the top banner view as the source view, so that the popover points to it instead of the main view? We generally want the source view to be the button that cause the popover to show up, but it can be more complicated to handle.

itsmeichigo (commented):
I tested with an iPhone and this works great! However I can't seem to get the popover to show up on iPad, it's probably due to the source view of the popover being the main view of the view controller.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressTopBannerFactory.swift
rachelmcr: That makes a lot of sense, thanks for the suggestion! I made that change in ffb8588.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Shipping Address Validation/ShippingLabelAddressFormViewController.swift
rachelmcr: Agreed. I tried a couple approaches to set the source view to `topBannerView` or its buttons, but I found that to get the popover to appear where I expected I had to set it to the table header view (where the banner view is located). I made that change in bb2ea3b:

rachelmcr:
Ok, ready for another look @itsmeichigo! I think the change I made to the popover is an okay compromise for positioning it, but let me know if you know of a better way to do that.

itsmeichigo (commented):
Hi @rachelmcr - I think setting the button as source view for the popover can be achieved by adding a new parameter in the action handler closure of the `ActionButton` in `TopBannerViewModel` like how I did [here](https://github.com/woocommerce/woocommerce-ios/compare/issue/4791-shipping-contact-options...huong/4791-top-banner-action-button-source-view?expand=1). I'm not sure if this is too much of a hack, WDYT?

rachelmcr:
@itsmeichigo Thanks for that suggestion! I didn't think to set the source view that way but it does just what we need here. 👍  I think it's a reasonable addition to the top banner, so I've added your change and updated the unit tests. This is ready for another look.

itsmeichigo (approved):
LGTM 🚀

PR #5229, opened 2021-10-15
Shipping Labels: Update ITN placeholder text when it's required in customs form

### LINKED ISSUE #4972: Shipping Labels: ITN placeholder text should be updated when it's required in customs form.
# Description
When ITN field is required on customs form, the placeholder text should be updated to remove `(optional)`.

# Screenshot
image

rachelmcr:
Fixes: #4972

## Description

In the Shipping Label creation flow, when the ITN field is required on a customs form, the placeholder text is updated to remove `(optional)`.

## Changes

* In `ShippingLabelCustomsFormInput`, checks whether the ITN is missing for the destination and uses that to determine whether the placeholder text will be set to `Enter ITN` (when required) or `Enter ITN (Optional)` (when optional).

Required ITN|Optional ITN
-|-
Simulator Screen Shot - iPhone 13 Pro - 2021-10-15 at 14 09 29|Simulator Screen Shot - iPhone 13 Pro - 2021-10-15 at 14 10 08

## Testing

1. Make sure that your test store has the WCShip plugin installed and activated.
2. Create an order that requires international shipping (origin country different from destination country) where the destination country requires an Internal Transaction Number (ITN) on the customs form (for example, Iran).
3. Open the app, select the Orders tab, and select the created order.
5. Go through Ship From, Ship To, Package Details to configure Customs.
6. On the Customs screen, confirm you see the ITN row with the placeholder `Enter ITN` and the validation message below it noting that the ITN is required.
7. Change the destination country to a country that does NOT require an ITN on the customs form (for example, UK).
8. Go back to Customs, confirm you see the ITN row with the placeholder `Enter ITN (Optional)`.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormInput.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormInput.swift
@@ -133,7 +133,7 @@ struct ShippingLabelCustomsFormInput: View {
     private var itnRows: some View {
         VStack(spacing: 0) {
             TitleAndTextFieldRow(title: Localization.itnTitle,
-                                 placeholder: viewModel.missingITNForDestination ? Localization.itnRequiredPlaceholder : Localization.itnPlaceholder,
+                                 placeholder: viewModel.missingITNForDestination || viewModel.missingITNForClassesAbove2500usd ? Localization.itnRequiredPlaceholder : Localization.itnPlaceholder,
                                  text: $viewModel.itn)
             Divider()
                 .padding(.leading, Constants.horizontalPadding)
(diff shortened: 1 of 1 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormInput.swift
```diff
@@ -133,7 +133,7 @@ struct ShippingLabelCustomsFormInput: View {
     private var itnRows: some View {
         VStack(spacing: 0) {
             TitleAndTextFieldRow(title: Localization.itnTitle,
-                                 placeholder: Localization.itnPlaceholder,
+                                 placeholder: viewModel.missingITNForDestination ? Localization.itnRequiredPlaceholder : Localization.itnPlaceholder,
```
itsmeichigo: I think we should also check for `viewModel.missingITNForClassesAbove2500usd` here.

itsmeichigo (changes requested):
Thanks for fixing this issue! However I think there's a case missing - we need to also check for when items with the same valid tariff number has total value higher than $2500 the ITN field should be required too.

WooCommerce/Classes/ViewRelated/Orders/Order Details/Shipping Labels/Create Shipping Label Form/Customs/ShippingLabelCustomsFormInput.swift
rachelmcr: Thanks for catching that case! I fixed that in d5419ac:

itsmeichigo (approved):
It works perfectly now, thanks! :shipit:

Issue #5159, opened 2021-10-07
Product List: add support for product filtering by category


pmusolino:
At the moment, the product lists can be filtered by Stock Status, Status, and Product Type. The capability to filter by category has been requested a few times by merchants and will be good to add to the app.
Related task on Android: https://github.com/woocommerce/woocommerce-android/issues/4578

**TODO**:
- [x] Fetch a list of products, filtered by category. [Networking/Yosemite layer changes]
- [x] Get a list of products from core data filtered by category. [Storage layer]
- [x] Add a feature flag for not exposing this feature until is completed to our merchants.
- [x] When the product filter screen is opened, fetch a list of categories (if there's internet) or get from DB if they are stored, and add them to the list of filters. The categories view should manage also pagination. [UI]
- [x] When a category filter is selected and the "Show products" button is tapped, include the category as criteria for filtering and show the result. [UI]
- [x] On the products list screen, if a category filter is already selected and the user taps the "Filters" button again, make sure the selected category is persisted and already selected in the "Category" filter. [UI]
- [x] Send the new filter to analytics

toupper:
Added **TODO**:
- [x] Remove feature flag when the feature is done

toupper:
>  Get a list of products from core data filtered by category. [Storage layer]

As discussed in https://github.com/woocommerce/woocommerce-ios/pull/5310, this is not implemented because:

- Retrieving a list of products from core data does not include the products belonging to the subcategories of the selected categories, unlike the remote request.
- Retrieving a list of products from core data breaks the pagination, and it is not consistent with the remote request pagination. We could improve it in the future implementing that pagination synchronizing mechanism.

pmusolino:
Project completed with this PR: https://github.com/woocommerce/woocommerce-ios/pull/5388

PR #5230, opened 2021-10-15
Product List: Extract the product category list view and view model to their own classes.

toupper:
Sorry for the more than 500 lines of code changing, most of the changes are moving the logic to new classes.

## Description
In the context of https://github.com/woocommerce/woocommerce-ios/issues/5159, we want to extract the product category list view used for editing or adding a product into its own dedicated class so it can be reused when filtering products by category.
That way we will avoid code duplication for the product categories presentation and fetching logic.

In this. manner, we create a new `EditProductCategoryListViewController` and `EditProductCategoryListViewModel` to replicate the logic previously implemented by `ProductCategoryListViewController` and `ProductCategoryListViewModel`, composing the product list with the add button and the specific use case logic.

## Changes
- Extract product category list view to `ProductCategoryListViewController`
- Extract product category list presentation logic to `ProductCategoryListViewModel` Since it is not related to a specific product, we just need the `siteID` to operate
- Use view controller containment in `EditProductCategoryListViewController` to show the product category list together with the specific case UI (Add Category Button)
- Create `EditProductCategoryListViewModel`, that handles the business logic of this case holding a reference to the product category list view model
- Split unit tests files accordingly
- Reorganize the project group structure to accommodate this new implementation

## Testing
Since no new functionality is added here, we have to make sure that the product categories functionality works as before
Prerequisites: your account has to have one store, ideally with products and categories.

Use Case 1: Assign  a existing category to a product

- Open the app, go to the Products tab
- Select a product
- Select Categories in the menu
- Select any category and press Done, it should be added to the Product.

Use Case 2: Unassign a category from a product

- Open the app, go to the Products tab
- Select a product
- Make sure that it has categories assigned
- Select Categories in the menu
- Make sure that these categories are checked when opening the categories list
- Uncheck that category
- Make sure that that category is not assigned to the product anymore

Use Case 3: Create new category and add it to the product

- Open the app, go to the Products tab
- Select a product
- Select Categories in the menu
- Press add category
- Create a new category following the steps in the menu
- Make sure that that category is checked in the product category list
- When pressing Done, make sure that that category is assigned to the product

Use Case 3: Create new product and assign a category

- Open the app, go to the Products tab
- Press on the + button, at the top right
- Select a Product type
- Press on "+ Add more details"
- Select Categories
- Existing categories should be shown
- Check any category
- When pressing Done, make sure that that category is assigned to the product

Update release notes:


WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
```diff
@@ -6,24 +6,17 @@ import WordPressUI
 ///
 final class ProductCategoryListViewController: UIViewController {
 
-    @IBOutlet private weak var addButton: UIButton!
-    @IBOutlet private weak var addButtonSeparator: UIView!
-    @IBOutlet private var tableView: UITableView!
+    @IBOutlet var tableView: UITableView!
```
pmusolino: nit: this could be declared as `private`

WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
```diff
@@ -6,24 +6,17 @@ import WordPressUI
...
     private let ghostTableView = UITableView()
 
-    private let viewModel: ProductCategoryListViewModel
+    let viewModel: ProductCategoryListViewModel
```
pmusolino, line 14: nit: this one should be declared `private` or `private(set)`, since probably we don't want to change its behavior from outside.

WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
```diff
@@ -6,24 +6,17 @@ import WordPressUI
...
 
     private let siteID: Int64
 
-    // Completion callback
-    //
-    typealias Completion = (_ categories: [ProductCategory]) -> Void
-    private let onCompletion: Completion
+    init(siteID: Int64) {
+        self.viewModel = ProductCategoryListViewModel(siteID: siteID)
```
pmusolino: nit: there is no need to use `self` here 👍

WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
```diff
@@ -6,24 +6,17 @@ import WordPressUI
 ///
 final class ProductCategoryListViewController: UIViewController {
```
pmusolino, line 9: Since this view controller can’t be used as is, because it needs to be embedded since there are no configurations for navigation, it could be useful to add a comment on top of it just to explain how it should work.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewModel.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    private let product: Product
+    var selectedCategories: [ProductCategory] {
+        get {
+            baseProductCategoryListViewModel.selectedCategories
+        }
+        set {
+            baseProductCategoryListViewModel.selectedCategories = newValue
+        }
+    }
```
pmusolino: I'm not sure this is useful. It hides in a certain way the view model. Plus, we are writing the `selectedCategories` property of the view model two times: the first one during the `baseProductCategoryListViewModel` assignation, and the other one after assigning it to the property `selectedCategories`. So, in my opinion, we can remove this property simplifying it, and making `selectedCategories` under `ProductCategoryListViewModel` private, and allowing the initializer to take the selected Categories.

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+    private let siteID: Int64
+
+    // Completion callback
+    //
+    typealias Completion = (_ categories: [ProductCategory]) -> Void
+    private let onCompletion: Completion
+
+    init(product: Product, completion: @escaping Completion) {
+        self.product = product
+        self.productCategoryListViewController = ProductCategoryListViewController(siteID: product.siteID)
```
pmusolino: nit: `self` is not needed here

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+        self.viewModel = EditProductCategoryListViewModel(product: product, baseProductCategoryListViewModel: productCategoryListViewController.viewModel)
```
pmusolino: nit: `self` is not needed here

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+        self.siteID = product.siteID
```
pmusolino: nit: `self` is not needed here

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+        configureNavigationBar()
+        handleSwipeBackGesture()
+    }
+}
+
+// MARK: - View Configuration
+//
+private extension EditProductCategoryListViewController {
+    func configureAddButton() {
+        addButton.setTitle(NSLocalizedString("Add Category", comment: "Action to add category on the Product Categories screen"), for: .normal)
```
pmusolino: Since we are refactoring these view controllers, can we move all these `NSLocalizedString` under an enum `Localization` as we are doing in other classes? 🙇

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+        title = NSLocalizedString("Categories", comment: "Edit product categories screen - Screen title")
+    }
+
+    func configureProductCategoryListView() {
+        addChild(productCategoryListViewController)
+        attachSubview(productCategoryListViewController.view)
+        productCategoryListViewController.didMove(toParent: self)
+    }
+
+    private func attachSubview(_ subview: UIView) {
```
pmusolino: nit: `private` is not needed here since the method is already under a private extension

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    @objc private func addButtonTapped() {
+        ServiceLocator.analytics.track(.productCategorySettingsAddButtonTapped)
+        let addCategoryViewController = AddProductCategoryViewController(siteID: siteID) { [weak self] (newCategory) in
+            defer {
+                self?.dismiss(animated: true, completion: nil)
+            }
+            self?.viewModel.addAndSelectNewCategory(category: newCategory)
+            self?.productCategoryListViewController.tableView.reloadData()
```
pmusolino: Not sure if this is a viable way, but maybe, the view model can tell the category list to refresh?

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewModel.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    init(product: Product, baseProductCategoryListViewModel: ProductCategoryListViewModel) {
+        self.product = product
+        self.baseProductCategoryListViewModel = baseProductCategoryListViewModel
+
```
pmusolino: nit: I believe we can remove this extra space here.

pmusolino (changes requested):
Great idea @toupper to extract the code in this way, to make it reusable. 👏 It makes sense to me, I just added some suggestions and nits 👍 

I just noticed that there is a small issue in the portrait, in fact, it seems that the separator under the top button does not follow the same behavior of the right side (which is the correct one).
Simulator Screen Shot - iPhone 13 Pro - 2021-10-18 at 15 18 34

WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
toupper: Good hint, fixed in https://github.com/woocommerce/woocommerce-ios/commit/ae5dbe5e01af1311b86f14fc308eeb83b02259ba

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
toupper: That's true, fixed in ae5dbe5e01af1311b86f14fc308eeb83b02259ba

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
toupper: Indeed, changed in That's true, fixed in ae5dbe5e01af1311b86f14fc308eeb83b02259ba

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
toupper: Yeah, fixed in That's true, fixed in https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/ae5dbe5e01af1311b86f14fc308eeb83b02259ba

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
toupper: Good point, fixed in https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/52108d6f0d7695764b76136afe3cff3b17042fbc

WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
toupper: Great point, added in https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/74fadd4ae647eb7033a48c639339b5228b0d25e7

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
toupper: Indeed it looks much nicer like that, done in https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/07d2ac49eecd1c2ba155f6123a0f9d1188209ae3 and https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/099d50df10a83a55e81e53378593027820058d00

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewModel.swift
toupper: true, done in https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/76bc161dd672696590dee9850b009ab327a1666e

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewController.swift
toupper: Yeah, I added an observer to the `ProductCategoryListViewModel` that can be invoked when there is a need to reload. Please check the commit https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/58426b33c0b28e60ca7cd8412cadef3d25388a79 This also allows the `tableView` property in `ProductCategoryListViewController` to be private

WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
toupper: Yeah, there is no need for it to be public after https://github.com/woocommerce/woocommerce-ios/commit/58426b33c0b28e60ca7cd8412cadef3d25388a79 It is changed in that commit

WooCommerce/Classes/ViewRelated/Products/Edit Product/Edit Categories/EditProductCategoryListViewModel.swift
toupper: Good points, I removed the property in https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/682b5d1bd7d07d4cb15962b34a2c1b70cb1dec05 and https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/953ecd67d477751576621efcc5f68f0c251f5c71, and adapt the code to the changes:
- The completion closure is invoked in `EditProductCategoryListViewModel` and not in `EditProductCategoryListViewController` That way we do not need to expose the categories, and `baseProductCategoryListViewModel` can remain private.
- selectedCategories are passed in the initializer of `ProductCategoryListViewModel`, which is neater than setting it afterwards
- `selectedCategories` in `ProductCategoryListViewModel` has now `private(set)` (as it has in develop). Instead of writing it directly from `EditProductCategoryListViewModel` when adding a category, we do it through a public function, thus limiting the access to it.

We still need however the `selectedCategories` under `ProductCategoryListViewModel` to be public for reading as it was before, as it is used when completion to pass the categories along, or when checking whether there are unsaved changes

WooCommerce/Classes/ViewRelated/Products/Categories/ProductCategoryListViewController.swift
toupper: Actually, there is not need of `private(set)` for read-only properties (`let`), hence we get the error _`private(set)`' modifier cannot be applied to read-only properties_ if we try it

toupper:
@pmusolino Thanks for the feedback! it was very helpful. I fixed the UI issue with the separator:
Simulator Screen Shot - iPhone 12 - 2021-10-18 at 20 21 28

pmusolino:
> Thanks for the feedback! it was very helpful. I fixed the UI issue with the separator

I re-tested it, but perhaps I have explained myself badly, I'm sorry. Previously everything was correct, it was just the separator in the image that was wrong, the highlighted part. That section, it had to be the same as the one on the right (which was the correct one). Now, also the separator of each cell is wrong on the right side.

Simulator Screen Shot - iPhone 13 Pro - 2021-10-20 at 11 02 09

toupper:
> I re-tested it, but perhaps I have explained myself badly, I'm sorry. Previously everything was correct, it was just the separator in the image that was wrong, the highlighted part. That section, it had to be the same as the one on the right (which was the correct one). Now, also the separator of each cell is wrong on the right side.

Yeah, as we talked it was like that in develop/app store version, but I fixed it following your indications, adjusting both to the safe area. It looks like this now after my addition in https://github.com/woocommerce/woocommerce-ios/pull/5230/commits/5aba4d1f9f1156002717562b7e7538d215a24978:

Simulator Screen Shot - iPhone 12 - 2021-10-20 at 21 13 42

Let me know if that is fine :)

pmusolino (approved):
Looks good 👏 congrats @toupper on your first contribution on WCiOS! 

Just an extra thing on the UI side (nothing serious, it can be fixed later):
Simulator Screen Shot - iPhone 13 Pro - 2021-10-21 at 13 30 28

PR #5231, opened 2021-10-15
Refactor receipt printing

joshheald:
This is preparatory 🏕️ work for #5108, which I'll be able to resolve by moving ReceiptRenderer and the generation into the WooCommerce project.

## Testing
View IPP receipts in the following flows:

1. Take payment for an order using a card reader, then tap "Print Receipt" in the alert after payment is completed
2. Take payment for an order using a card reader, tap "Back to Order" in the alert, then tap "See Receipt" on the order.

Update release notes:


WooCommerce/Classes/ViewModels/CardPresentPayments/ReceiptPrintingCoordinator.swift
```diff
@@ -0,0 +1,21 @@
...
+        ServiceLocator.analytics.track(.receiptPrintTapped)
+
+        let action = ReceiptAction.print(order: order, parameters: params) { (result) in
+            switch result {
+            case .success:
+                ServiceLocator.analytics.track(.receiptPrintSuccess)
+            case .cancel:
+                ServiceLocator.analytics.track(.receiptPrintCanceled)
+            case .failure(let error):
+                ServiceLocator.analytics.track(.receiptPrintFailed, withError: error)
```
allendav: Let's log this error too please

WooCommerce/Classes/ViewModels/CardPresentPayments/ReceiptPrintingCoordinator.swift
```diff
@@ -0,0 +1,21 @@
+import Foundation
+import Yosemite
+
+struct ReceiptPrintingCoordinator {
```
allendav: Would this coordinator be an appropriate place for emailed receipt action creation and dispatch too? If so, you might want to choose a different name

allendav (approved):
Tests fine, including new unit tests. One request. Thanks for DRYing the tracking and print closures

ctarda:
@joshheald is this ready to merge?

joshheald:
> @joshheald is this ready to merge?

@ctarda no not yet I'm afraid, not forgotten though! I'm aiming to get to it today if I can wrap up the Tracks stuff first.

ctarda:
No worries! I'll tag it as "do not merge" then, so that it doesn't get merged by mistake

WooCommerce/Classes/ViewModels/CardPresentPayments/ReceiptPrintingCoordinator.swift
joshheald: Yeah, I think it might, except that there isn't an email receipt _action_, just a `generateContent` which we can use for the content of an email. It's one of those places where we might end up fighting with UIKit a bit because of the API of `MFMailComposeViewController`. 

That said, it might make sense to have the `ReceiptActionCoordinator` vend a `MailViewModel` struct which can be used to configure the `MFMailComposeViewController`, then pass it back to the `OrderDetailsViewController` for use in presenting the compose box. The configuration via the `MailViewModel` might be handled in an extension on the compose VC for convenient reuse of that struct.

What do you think? Seems out of scope for this change, especially since the email code isn't duplicated, but I've done the rename anyway.

PR #5233, opened 2021-10-16
Edit Order Address: Enable/Disable `Done` button and show/hide `Use as {Shipping/Billing} Address` toggle by comparing shipping and billing Addresses.

### LINKED ISSUE #5148: [Order Editing] "Use as {Billing|Shipping} address" toggle changes can't be saved alone
**Describe the bug**
When editing an address in Order Details, changes to the `Use as {Billing|Shipping} address` toggle do not enable the `Done` button, so can't be saved unless other changes are made as well.

**To Reproduce**
1. Go to Orders tab
2. Open a processing order with different Billing and Shipping addresses
3. Tap the pencil icon next to Shipping address
4. Set the toggle to on
5. Note that the `Done` button is unavailable
6. Cancel and check the other address anyway
7. Note that it has not been updated to match.

**Screenshots**

https://user-images.githubusercontent.com/2472348/136225911-125d91fc-af2b-4b15-8cca-7e278c781331.mp4

**Expected behavior**
Changing the toggle alone should enable the `Done` button and allow the address to be saved across to its alternate.

**Mobile Environment**
Please include:
- Device: iPhone 13 Pro
- iOS version: iOS 15
- WooCommerce iOS version: 7.7.0

selanthiraiyan:
Closes #5148 

# Description

Currently, the `Use as {Shipping/Billing} Address` toggle in "Edit Order Address screen" does not enable the Done button when turned on.

This has been identified and reported in #5148.

# Changes

1. Show the `Use as {Shipping/Billing} Address` toggle only if the currently entered address is different from its counterpart. 
For example, while editing “Shipping Address”, the address entered in the form will be compared with "Billing Address" and the toggle will be shown only if those addresses are not equal. 

1. Enable the `Done` button if the toggle is turned on. Please note that the toggle will be shown, only if point 1 is satisfied.

This way, we will avoid updating the address unnecessarily just because the toggle is turned on, even though the current address and its counterpart are equal.

# Demo

https://user-images.githubusercontent.com/524475/137592594-62558dfd-af33-4d05-84a8-6c5b6915d26f.mov

# Testing
1. Make sure that there is a processing order with different Billing and Shipping addresses.
1. Go to Orders tab and open the order.
1. Tap the pencil icon next to the shipping address to edit it.
1. Observe that `Use as Billing Address` toggle is shown.
1. Turn on the toggle and observe that `Done` button is enabled.
1. Tap `Done` and save the address. This will make Billing and Shipping addresses same.
1. Try editing the shipping address again and observe that `Use as Billing Address` toggle is NOT shown.
1. Follow the same process for editing billing address as well and observe `Use as Shipping Address` toggle and `Done` button behave in the same manner as described above.

# Points to consider
- This PR includes a UI level change. i.e. `Update {Shipping/Billing} Address` toggle is ONLY displayed when the addresses are different. This might make the toggle less noticeable or difficult to find, for the end-users.
- If the counterpart address is edited/changed using a different device/website, we will not know about the update until the order screen is refreshed, which means that we will not show the toggle in the “Edit address” screen until the order data is reloaded.

Update release notes:

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -11,7 +11,7 @@
 - [*] Shipping Labels: The shipping address now prefills the phone number from the billing address if a shipping phone number is not available. [https://github.com/woocommerce/woocommerce-ios/pull/5177]
 - [*] Shipping Labels: now in Carrier and Rates we always display the discounted rate instead of the retail rate if available. [https://github.com/woocommerce/woocommerce-ios/pull/5188]
 - [*] Accessibility: notify when offline mode banner appears or disappears. [https://github.com/woocommerce/woocommerce-ios/pull/5225]
-- [*] Use as {Shipping/Billing} Address toggle: Edit Order Address - Enable/Disable `Done` button and show/hide `Use as {Shipping/Billing} Address` toggle by comparing shipping and billing Addresses. [https://github.com/woocommerce/woocommerce-ios/pull/5233]
+- [*] Edit Order Address - Enable/Disable `Done` button and show/hide `Use as {Shipping/Billing} Address` toggle by comparing shipping and billing Addresses. [https://github.com/woocommerce/woocommerce-ios/pull/5233]
 7.7
 -----
 - [***] In-Person Payments: US merchants can now obtain a card reader and then collect payments directly from the app. [https://github.com/woocommerce/woocommerce-ios/pull/5030]
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -11,7 +11,7 @@
 - [*] Shipping Labels: The shipping address now prefills the phone number from the billing address if a shipping phone number is not available. [https://github.com/woocommerce/woocommerce-ios/pull/5177]
 - [*] Shipping Labels: now in Carrier and Rates we always display the discounted rate instead of the retail rate if available. [https://github.com/woocommerce/woocommerce-ios/pull/5188]
 - [*] Accessibility: notify when offline mode banner appears or disappears. [https://github.com/woocommerce/woocommerce-ios/pull/5225]
-
+- [*] Use as {Shipping/Billing} Address toggle: Edit Order Address - Enable/Disable `Done` button and show/hide `Use as {Shipping/Billing} Address` toggle by comparing shipping and billing Addresses. [https://github.com/woocommerce/woocommerce-ios/pull/5233]
 7.7
 -----
 - [***] In-Person Payments: US merchants can now obtain a card reader and then collect payments directly from the app. [https://github.com/woocommerce/woocommerce-ios/pull/5030]
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
@@ -135,7 +135,7 @@ final class EditOrderAddressFormViewModel: ObservableObject {
     }
 
     /// Defines whether "Use as {Shipping/Billing} Address" toggle should be shown.
-    /// Show the toggle, only if the entered address is different from it's counterpart.
+    /// Show the toggle, only if the entered address is different from its counterpart.
     /// i.e. Compare entered address with shipping address while editing billing address, and vice versa.
     ///
     var shouldShowUseAddressAsToggle: Bool {
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
@@ -134,6 +134,20 @@ final class EditOrderAddressFormViewModel: ObservableObject {
         }
     }
 
+    /// Defines whether "Use as {Shipping/Billing} Address" toggle should be shown.
+    /// Show the toggle, only if the entered address is different from it's counterpart.
+    /// i.e. Compare entered address with shipping address while editing billing address, and vice versa.
+    ///
+    var shouldShowUseAddressAsToggle: Bool {
+        let addressInForm = fields.toAddress(country: selectedCountry, state: selectedState)
+        switch type {
+        case .shipping:
+            return addressInForm != order.billingAddress
+        case .billing:
+            return addressInForm != order.shippingAddress
+        }
+    }
+
     /// Defines if the state field should be defined as a list selector.
     ///
     var showStateFieldAsSelector: Bool {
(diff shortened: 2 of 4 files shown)


RELEASE-NOTES.txt
```diff
@@ -11,7 +11,7 @@
 - [*] Shipping Labels: The shipping address now prefills the phone number from the billing address if a shipping phone number is not available. [https://github.com/woocommerce/woocommerce-ios/pull/5177]
 - [*] Shipping Labels: now in Carrier and Rates we always display the discounted rate instead of the retail rate if available. [https://github.com/woocommerce/woocommerce-ios/pull/5188]
 - [*] Accessibility: notify when offline mode banner appears or disappears. [https://github.com/woocommerce/woocommerce-ios/pull/5225]
-
+- [*] Edit Order Address - Enable/Disable `Done` button and show/hide `Use as {Shipping/Billing} Address` toggle by comparing shipping and billing Addresses. [https://github.com/woocommerce/woocommerce-ios/pull/5233]
```
jaclync, line 14: We might want to move this entry to release 7.9, since we code freeze on Oct 18 Monday 🙂 If 7.9 isn't available from the latest `develop` yet, you can just start a new 7.9 section at the top

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
```diff
@@ -134,6 +134,20 @@ final class EditOrderAddressFormViewModel: ObservableObject {
         }
     }
 
+    /// Defines whether "Use as {Shipping/Billing} Address" toggle should be shown.
+    /// Show the toggle, only if the entered address is different from its counterpart.
+    /// i.e. Compare entered address with shipping address while editing billing address, and vice versa.
+    ///
+    var shouldShowUseAddressAsToggle: Bool {
```
jaclync, line 141: nit:
```suggestion
    var showUseAddressAsToggle: Bool {
```
just to be more consistent with other similar variables in this class?

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/Addresses/EditOrderAddressFormViewModelTests.swift
```diff
@@ -588,15 +588,150 @@ final class EditOrderAddressFormViewModelTests: XCTestCase {
...
+        let address = sampleAddress()
+        return EditOrderAddressFormViewModel(order: order(shippingAddress: address, billingAddress: address),
+                                             type: addressType,
+                                             storageManager: testingStorage)
+    }
+
+    func viewModelWithDifferentShippingAndBillingAddresses(addressType: EditOrderAddressFormViewModel.AddressType) -> EditOrderAddressFormViewModel {
+        let johnny = sampleAddress()
+        let appleseed = johnny.copy(firstName: "Appleseed",
+                                    lastName: "Johnny")
```
jaclync, line 727: nit: I haven't seen naming like this in the codebase, maybe just `let address` and `let anotherAddress`? 🤔

WooCommerce/WooCommerceTests/ViewRelated/Orders/Order Details/Addresses/EditOrderAddressFormViewModelTests.swift
```diff
@@ -588,15 +588,150 @@ final class EditOrderAddressFormViewModelTests: XCTestCase {
...
+
+        // When
+        viewModel.onLoadTrigger.send()
+        XCTAssertEqual(viewModel.navigationTrailingItem, .done(enabled: false))
+
+        viewModel.fields.useAsToggle = true
+
+        // Then
+        XCTAssertEqual(viewModel.navigationTrailingItem, .done(enabled: true))
+    }
```
jaclync, line 712: nice test cases! how about adding one that is similar to `test_use_as_address_toggle_is_shown_for_order_with_same_shipping_and_billing_addresses_when_field_edited_and_updated_back_to_original_values` - edit the fields, undo the changes, then assert on `navigationTrailingItem` remains `.done(enabled: false)`? 😄

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
jaclync, line 141: I have a question about this on a conceptual level (since I'm still learning SwiftUI, tagging the authors of this file @Ecarrion @ealeksandrov here for confirmation):

Does it make sense to mark this variable as `@Published var` and then assign this variable to `Publishers.CombineLatest3($fields, $selectedCountry, $selectedState)` like the code below? So that when this variable changes (i.e. when the output of `fields`, `selectedCountry` and `selectedState` changes), the corresponding UI state changes. Right now, this use of computed variable also works probably because the view is always redrawn from any `fields`/`selectedCountry`/`selectedState` changes by the user. But I wonder if using `@Published var` here would make the reactive chain more clear? Feel free to share any pros & cons between these two approaches!

```swift
        Publishers.CombineLatest3($fields, $selectedCountry, $selectedState)
            .map { [weak self] fields, selectedCountry, selectedState -> Bool in
                guard let self = self else { return false }

                let addressInForm = fields.toAddress(country: selectedCountry, state: selectedState)
                switch self.type {
                case .shipping:
                    return addressInForm != self.order.billingAddress
                case .billing:
                    return addressInForm != self.order.shippingAddress
                }
            }
            .assign(to: &$shouldShowUseAddressAsToggle)
```

jaclync (commented):
Nice work on the fix with unit tests, @selanthiraiyan!

Code: ❓ I just had one major question about the use of computed variable vs. @Pusblished variable for the toggle visibility boolean
Testing: ✉️ I did have one issue after submitting the form with "Use as address" selected when the billing address has an email while the shipping address does not. Since [`Address`'s Equatable implementation](https://github.com/woocommerce/woocommerce-ios/blob/5d786627d3cde6aafc859ec4413b7d2b5e2c22f7/Networking/Networking/Model/Address.swift#L90-L109) also includes the email, it considers the two addresses as different when the emails don't match. However, email isn't in the address form and won't be copied when the user enables "Use as address" when submitting the form.

Steps to repro:
Prerequisite: the store has an order where the billing address has an email while the shipping address does not (there is no email field in wp-admin and iOS app), and the rest of address fields are exactly the same
- Go to the orders tab
- Tap on the order in the prerequisite
- Tap the pencil icon to edit the shipping address

Actual: "Use as Billing Address" toggle row is shown
Expected: "Use as Billing Address" is not shown

Since this exclusion of email is specific to the order editing use case, maybe creating another function to compare two addresses without the email makes more sense. What do you think?

Ecarrion (commented):
Hi @selanthiraiyan, thanks for contributing again! 

The UX change where you hide the "use as" button when the addresses are the same is a bit out of the scope of the original ticket.

I am personally not totally convinced because it deviates from the website experience(is always visible) and because it will limit the exposure of the feature just introduced.

Screen Shot 2021-10-19 at 3 30 57 PM


@adamzelinski do you have any particular opinion here?

WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
Ecarrion: I think you have a fair point @jaclync! 

And I'm probably the one to blame since I have added variables like that in this class relying on the fact that it works probably because the view is always redrawn from any fields/selectedCountry/selectedState changes by the user.

On the other hand, lets probably wait until the design team share their opinion on this UX change 😅

selanthiraiyan: Thanks for the detailed review @jaclync and I agree with your point as well. I have also listed my thoughts down below,

1. If this was UIKit and Combine, we might have to think about restricting unnecessary reloads to the view. But, SwiftUI takes care of that for us by comparing and identifying changes.
1. As long as we have one `combineLatest` stream responsible for `assign`ing one single value, without of modifying any other values in between, I think having one stream for each variable will be good and easy to understand.

As mentioned by @Ecarrion , I will work on including your suggested changes, once we have the design team's opinion. I will have to remove most of the code if the design team says NO to hiding the toggle. 😀

selanthiraiyan:
I understand @Ecarrion. It makes complete sense to avoid hiding the toggle if it is reducing the visibility of the feature.
I will wait for the final decision and work on changing this PR.
Thank you!

selanthiraiyan:
I have opened https://github.com/woocommerce/woocommerce-ios/pull/5254 to close this issue.
Thanks for your time @jaclync and @Ecarrion !

PR #5236, opened 2021-10-18
Order Details: Fix outdated product variation results controller when updating order

### LINKED ISSUE #5170: Shipping Labels: Purchased label shows 0 quantity for variation items immediately after purchase
**Describe the bug**

As reported on https://github.com/woocommerce/woocommerce-ios/pull/5154#pullrequestreview-773778126: 

When the items for a shipping label are product variations (instead of simple products), the first time you navigate back to the Order Detail screen after purchasing the label it shows "0 Items" for that shipping label:

Order Detail|Items detail for shipping label
-|-
|

But if you go back to the Orders list at that point and reopen the Order Detail, the correct number of items then appears for all the shipping labels. So it's only a problem the first time you go back to the Order Detail after purchasing the label.

**To Reproduce**

2. Create an order with simple products and variable products.
3. On Orders tab, select the new order and select Create Shipping Label.
4. Complete the shipping label form and purchase the label.
5. Navigate to Order Details, notice that the number of items only includes the simple products.
6. Tap on "# items" and notice that the variable product has a 0 quantity next to it.
7. Go back to the Orders list, and then reopen the Order Detail for that order. Notice the shipping label now shows the correct number of items (including the correct variable product quantity).

I repeated the same steps, with the same variable product, on another order and got the same result.

itsmeichigo:
Fixes #5170 

# Description
There is an edge case when an `OrderDetailsViewModel` is created with a storage model that has an empty list of order items. This list should be updated after syncing with remote is done, which then triggers the data source to update order with `update(order:)` method.

Currently the `resultsControllers` is created only once with the initial order, so when this edge case happens, the `resultsControllers` is tied with an order that has no items. Its `productVariationResultsController` needs the order item list to get the list of variation IDs for fetching, so if the order item list is not updated, the Order Details screen cannot display the correct item count for product variations in shipping labels.

This PR fixes the issue by forcing the data source to update the order in its `resultsControllers`.

# Testing
2. Create an order with simple products and variable products.
3. On Orders tab, select the new order and select Create Shipping Label.
4. Complete the shipping label form and purchase the label.
5. Navigate to Order Details, notice that the number of items only includes the simple products.
6. Tap on "# items" and notice that the variable product has correct quantity.

This issue can also be tested with orders with existing shipping labels.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -204,6 +204,8 @@ final class OrderDetailsDataSource: NSObject {
 
     func update(order: Order) {
         self.order = order
+        resultsControllers.update(order: order)
+        reloadSections()
     }
 
     func configureResultsControllers(onReload: @escaping () -> Void) {
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -205,7 +205,6 @@ final class OrderDetailsDataSource: NSObject {
     func update(order: Order) {
         self.order = order
         resultsControllers.update(order: order)
-        reloadSections()
     }
 
     func configureResultsControllers(onReload: @escaping () -> Void) {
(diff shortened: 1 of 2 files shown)


rachelmcr (approved):
Looks great and tests well! Thanks for tracking down the source of this issue. :shipit:

PR #5241, opened 2021-10-19
Keep sites in site picker up to date

### LINKED ISSUE #931: You need to log out of the app to view newly-added sites
**Describe the bug**
If I add a site to my account, I can't see it unless I log out.

**To Reproduce**
1. Create (or be invited to(?)) a new Woo site.
2. It doesn't show up in the list unless I log out and back in.

**Expected behavior**
I'd expect it to show up in my list of sites. When I view the site list, the button at the bottom has a spinner – I'd expect that spinner to indicate that it's checking whether there are any new sites.

**Mobile Environment**
Please include:
- Device: iPhone Xs
- iOS version: 12.2 
- WooCommerce iOS version: 1.6.0.1 

**WordPress Environment**
n/a

### LINKED ISSUE #2520: While Logged In, Connecting or Disconnecting a Site Does Not Refresh the Switch Store List
## Describe the bug

Connecting or disconnecting a site from Jetpack does not refresh the list in Settings → Switch Store. 

###  Disconnecting

1. Log in on the app. 
2. On a desktop browser, log in to WordPress.com.
3. Disconnect the self-hosted WooCommerce site.
4. Go back to the app. 
5. Navigate to Settings → Switch Store. 
6. Notice that the disconnected site is still there.

###  Connecting 

1. Log in on the app.
2. On a desktop browser, log in to a self-hosted WooCommerce site that you have not linked to Jetpack yet.
3. Link the site to Jetpack.
4. Go back to the app. 
5. Navigate to Settings → Switch Store.
6. Notice that the new linked site is not there and there is no way to refresh (e.g. pull-to-refresh).

Force-closing the app and reopening it will update the list.

## Expected behavior

The site list should be automatically refreshed.

jaclync:
Fixes #2520
Fixes https://github.com/woocommerce/woocommerce-ios/issues/931
Prerequisite for the edge case in #5032 

## Why

There are two issues with the site list in the site picker:

### A newly connected site does not show up in the list

The site list is synced by `AccountAction.synchronizeSites`, which is dispatched on app launch or sign-up epilogue. That's why the new site shows up after relaunching the app. This PR added sites syncing to the site picker `viewDidLoad`, while showing the existing sites in the storage before the results come back.

### A newly disconnected site still shows up in the list

Other than the same issue as the newly connected site above where the sites aren't synced in the site picker, we also never delete any sites during a logged-in session. In this PR, any local sites that aren't in the result from the server are now deleted.

## Changes

- In `StorePickerViewController`, dispatched `AccountAction.synchronizeSites` in `viewDidLoad` to refresh the results
  - I left the log event to record the `num_of_stores` before the sync completion block, since the event is meant for the time when the site picker is shown
- In `AccountStore`, deleted any sites in the storage that aren't in the site list from the remote service with a test case

### Disconnecting
- Log in on the app
- Navigate to Settings → Switch Store --> the site that is deleted in the next steps should be in the list
- On a desktop browser, log in to WordPress.com
- Disconnect the self-hosted WooCommerce site, or from wp-admin > Users with another account
- Go back to the app
- Navigate to Settings → Switch Store --> the disconnected site should not show up anymore

### Connecting
- Log in on the app
- Navigate to Settings → Switch Store --> the site that is added in the next steps should not be in the list
- On a desktop browser, log in to a self-hosted WooCommerce site that you have not linked to Jetpack yet
- Link the site to Jetpack
- Go back to the app
- Navigate to Settings → Switch Store --> the newly connected site should appear in the list

Update release notes:


Storage/Storage/Tools/StorageType+Extensions.swift
```diff
@@ -26,6 +26,12 @@ public extension StorageType {
         return firstObject(ofType: Site.self, matching: predicate)
     }
 
+    /// Retrieves all stored sites.
+    ///
+    func loadAllSites() -> [Site]? {
```
shiki: I don't think this has to be optional. Unless I'm missing something? 🤪 

```suggestion
    func loadAllSites() -> [Site] {
```

WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
```diff
@@ -47,7 +47,7 @@ enum StorePickerConfiguration {
 
 /// Allows the user to pick which WordPress.com (OR) Jetpack-Connected-Store we should set up as the Main Store.
 ///
-class StorePickerViewController: UIViewController {
+final class StorePickerViewController: UIViewController {
```
shiki, line 50: 😁

WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
```diff
@@ -262,6 +270,15 @@ private extension StorePickerViewController {
     }
 }
 
+// MARK: - Syncing
+//
+private extension StorePickerViewController {
+    func synchronizeSites(onCompletion: @escaping (Result<Void, Error>) -> Void) {
+        let action = AccountAction.synchronizeSites { _ in }
```
shiki: ⚠️ I wonder if you were planning to use the `onCompletion` here. It looks like if we don't, the list is not updated right away. Consider these steps:

1. Log in. 
2. Connect a site, Sad Panda, on the web.
3. Open the site picker. That Sad Panda site is not shown in the list. 
4. Close the site picker.
5. Open the site picker again. The Sad Panda is now shown. 

But if we do it this way, it's synchronized right away, and Sad Panda is immediately visible:

```suggestion
        let action = AccountAction.synchronizeSites(onCompletion: onCompletion)
```

Which means, Happy Panda? 🐼 😄 

Let me know if this is intentional! I can also share a video privately if it helps to understand the problem.

Yosemite/Yosemite/Stores/AccountStore.swift
```diff
@@ -195,6 +195,13 @@ extension AccountStore {
     func upsertStoredSitesInBackground(readOnlySites: [Networking.Site], onCompletion: @escaping () -> Void) {
         let derivedStorage = sharedDerivedStorage
         derivedStorage.perform {
+            // Deletes sites in storage that are not in `readOnlySites`.
+            let storageSites = derivedStorage.loadAllSites()
+            storageSites?.filter { readOnlySites.map { $0.siteID }.contains($0.siteID) == false }
+                .forEach { remotelyDeletedSite in
+                    derivedStorage.deleteObject(remotelyDeletedSite)
+                }
```
shiki: Do you think moving the map outside would be better for readability? I'm not sure. 🤔 

```suggestion
            let readOnlySiteIDs = readOnlySites.map(\.siteID)
            storageSites.filter { readOnlySiteIDs.contains($0.siteID) == false }
                .forEach { remotelyDeletedSite in
                    derivedStorage.deleteObject(remotelyDeletedSite)
                }
```

shiki (approved):
Hi, @jaclync! Thank you so much for looking into this. It's an annoying bug. 

I have one concern in the code marked with ⚠️ . I'm approving this to let you handle that, unblock you, and merge this whenever you're ready. If you would like to discuss further though, I'm totally fine doing that too. 🙂 

## Disconnecting a Currently Selected Site ⚠️ 

There is also another thing that I found that could be a concern. Though I think it's an edge case. Consider these steps:

1. Log in and pick a site. 
2. Disconnect that selected site from Jetpack
3. Open the site picker. 



Thankfully, the app will show an error and not let the user proceed until they've selected a site. I'm not quite sure what the priority of this is or if this should be fixed. Should we even allow deletion of the selected site? 🤷 I'll let you decide on that. 🙂 

## Testing

- [x] Disconnecting ✔️
- [x] Connecting ✔️
- [x] Disconnecting currently selected site ❓ 

Sometimes, when the picker is shown, the checkmark is not shown for the currently selected site's row. But this is a problem that's been there for a while now. Also, when Dismiss is pressed, the site picker still switches to that site. I think we have this logged somewhere. I just thought of mentioning these in case these are blockers for the project.

shiki:
I edited the description to say that this should also fix https://github.com/woocommerce/woocommerce-ios/issues/931.

WooCommerce/Classes/Authentication/Epilogue/StorePickerViewController.swift
jaclync: totally my bad to forget calling the completion block here 🤦🏻‍♀️ updated in https://github.com/woocommerce/woocommerce-ios/pull/5241/commits/09ef3f9fccdb51ccc4362f5dade97176acd6f535

Yosemite/Yosemite/Stores/AccountStore.swift
jaclync: yup! updated in https://github.com/woocommerce/woocommerce-ios/pull/5241/commits/2e841d6ad663298c0f31d41cdfd7e866225b855b

Storage/Storage/Tools/StorageType+Extensions.swift
jaclync: nice catch! I was following another similar function that returns an optional list, but it isn't necessary in this case. updated in https://github.com/woocommerce/woocommerce-ios/pull/5241/commits/3194c17ae5c23de8471d358e7e66b9b0c3e101e4

jaclync:
Thank you for the thorough review & testing @shiki, as always! 🙏 I updated the PR for another look, bunch of pre-existing issues with the site picker indeed 😅 

> Disconnecting a Currently Selected Site ⚠️

Great question! How about not deleting the selected site, even though it doesn't exist in the API response anymore? I made the changes in https://github.com/woocommerce/woocommerce-ios/pull/5241/commits/3194c17ae5c23de8471d358e7e66b9b0c3e101e4, so that the remotely deleted selected site is still visible in the site picker. An alert about the site not loading shows up when opening the site picker, and the user can select another site to connect or dismiss the alert. This way, the user can still see the selected site and the URL to fix the issue. Feel free to suggest other improvements too!

> Sometimes, when the picker is shown, the checkmark is not shown for the currently selected site's row. But this is a problem that's been there for a while now.

I also noticed this checkmark issue before, hopefully this commit that only compares the site ID fixes it: https://github.com/woocommerce/woocommerce-ios/pull/5241/commits/cb3dfdfa79803abfaab81f55f11e7e2a3bf9cd98

> Also, when Dismiss is pressed, the site picker still switches to that site. I think we have this logged somewhere. I just thought of mentioning these in case these are blockers for the project.

Good find! I didn't see a GH issue for this, created a new one https://github.com/woocommerce/woocommerce-ios/issues/5246 - I'll take a look this sprint

shiki (approved):
>  I made the changes in 3194c17, so that the remotely deleted selected site is still visible in the site picker.

I tested this and I love it! It's definitely better in terms of UX in my opinion. I also don't notice the checkmark disappearing anymore. 

Thank you for fixing all of these! 🙇

jaclync:
Thank you for testing again Jayson! Merging now 😃

PR #5242, opened 2021-10-19
Order Editing: clear billing email

### LINKED ISSUE #5115: Order Editing: Inability to clear email
The API has a limitation where we can't remove the email field while editing an address because the server expects a valid email address.

A PR on Core is waiting for approval https://github.com/woocommerce/woocommerce/pull/30850

Once that is merged we need to update the feature to allow email deletion in the address form screen

ealeksandrov:
Closes https://github.com/woocommerce/woocommerce-ios/issues/5115.

## Description

This PR adds support for clearing email in billing address, since support for it have been merged into core.
Before this change iOS were sending empty string as cleared email and this field was silently ignored by the backend. But all other changed fields saved fine in same request.

### Considerations

On WC <= 5.8.1 this change will result in error (code: `rest_invalid_param` message: `billing`) when email is cleared. This will be a blocker to save other fields as well.

Since email field is required for billing address (in normal checkout flow) and removing data is not a common action - so we'll keep the blocking error. To simplify debugging (for failed request with empty email) additional log line added with link to this PR.

WC 5.9.0 is scheduled to release on November 9 2021. WC 5.9.0 beta 1 is available now.

## Screenshots

WC 5.9.0 dev:

https://user-images.githubusercontent.com/3132438/137879226-9b699568-b19b-4028-a691-1798b99e7daa.mp4

WC 5.8.1:

---
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
@@ -199,6 +199,9 @@ final class EditOrderAddressFormViewModel: ObservableObject {
 
             case .failure(let error):
                 DDLogError("⛔️ Error updating order: \(error)")
+                if self.type == .billing, updatedAddress.hasEmailAddress == false {
+                    DDLogError("⛔️ Email is nil in address. It won't work in WC < 5.9.0 (https://git.io/J68Gl)")
+                }
                 self.presentNotice = .error(.unableToUpdateAddress)
                 self.analytics.track(event: WooAnalyticsEvent.OrderDetailsEdit.orderDetailEditFlowFailed(subject: self.analyticsFlowType()))
             }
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Orders/Order Details/Address Edit/EditOrderAddressFormViewModel.swift
```diff
@@ -197,6 +199,9 @@ final class EditOrderAddressFormViewModel: ObservableObject {
 
             case .failure(let error):
                 DDLogError("⛔️ Error updating order: \(error)")
+                if self.type == .billing, updatedAddress.hasEmailAddress == false {
+                    DDLogError("⛔️ Email is nil in address. It won't work in WC < 5.9.0 (https://git.io/J68Gl)")
+                }
                 self.presentNotice = .error(.unableToUpdateAddress)
```
Ecarrion, line 205: Nice, do you think we should also update the error presented to the user to something like:
> Unable to update address, please make sure you are running the latest version of WooCommerce try again.

Ecarrion (approved):
👍

ealeksandrov:
Thanks for review @Ecarrion! I've added suggestion for user. This required expanded 3-line notice, which was limited to 2 lines previously. The fix is straightforward (199ba6a), but do you think it's safe to do? I'm not missing something?

Simulator Screen Shot

Ecarrion (approved):
> but do you think it's safe to do? I'm not missing something?

It looks safe to me! Maybe just do a quick pass checking that most of the notices still look correctly and then 🚀

allendav:
@ctarda - we use billing email for receipts and probably customer creation and possibly payment intents - we should to check that we handle an empty billing email well for in-person payments (Android too)

Receipts on iOS is probably fine: https://github.com/woocommerce/woocommerce-ios/blob/5d786627d3cde6aafc859ec4413b7d2b5e2c22f7/WooCommerce/Classes/ViewRelated/Orders/Order%20Details/OrderDetailsViewController.swift#L785

PR #5244, opened 2021-10-19
[Mobile Payments] [SDK 2] Allow collect payment on orders where payment collection was interrupted by an error

### LINKED ISSUE #4655: [Mobile Payments] if a card present payment fails with "Networking.Dotcom error 0" the option to collect payment is removed
Steps to reproduce:

1. Create a situation where a payment will fail (i.e. a decimal stock quantity is used). You can install and activate a plugin that supports decimal quantities, such as [WPC Product Quantity for WooCommerce](https://wordpress.org/plugins/wpc-product-quantity/). To use this plugin as an example, go to WP Admin > WPClever > Product Quantity, change Decimal Quantities to Yes, add 0.5 for Minimum, Step, and Default.
1. On the web, submit an order containing a product with a decimal stock quantity.
1. In the app, open the order.
1. Tap the "Collect Payment" button and tap the test card to the reader.
1. Observe the error messages that pop up—do they make sense?
1. Pull-to-refresh on the order details screen.
1. Try to collect payment again.

**Result:** when a payment fails with the unexpected error "Networking.Dotcom error 0", the "Collect Payment" button turns back into an "Issue Refund" button and you become stuck awaiting payment but there is no longer an option to collect payment. ([5m29s](https://cloudup.com/caJgmGjc_Eh))

**Errors:**

> Payment failed
> The operation couldn't be completed. (Networking.Dotcom error 0.)
> Button: Try collecting payment again
> Button: No thanks

> Payment failed
> The system experienced an unexpected internal error - please try again
> Button: Dismiss

> A capture of $2.50 failed to complete with the following message: Error: Invalid integer: 0.5.

Error 1|Error 2|Error in notes
---|---|---
image|image|image

<sup>Tested with WCiOS 7.1.0.1 TestFlight beta on iPhone 6S iOS 14.7 logged in as testuser01121 with worthwhile-seahorse.jurassic.ninja selected and running WP 5.7.2 JP 9.9 WC 5.5.1 WCPay 2.7.0 WCS staging 1.0.</sup>

allendav:
Closes #4655

Changes:
- If collecting a payment fails and the payment method ID is changed to `woocommerce_payments` the logic prior to this PR would treat it as ineligible for in-person payments
- This PR adds `woocommerce_payments` as an acceptable payment method ID on an order. Coupled with the status of processing, on-hold, or none makes sense and continues to avoids the problem of collecting payments on orders which already have payment collected
- This PR also presents any error message provided from the backend during order payment capture directly to the user (e.g. "Payment capture failed to complete with the following message: Error: Invalid integer: 2.5") if preferred we could just show something simpler like "Payment capture failed to complete" - it would also avoid the risk of localization differences but would impart less information to the user. 

To test:
- Use the fractional quantity plugin described in @designsimply 's original report in #4655 to reproduce the error during payment collection
- Ensure that after pulling to refresh the order the Collect Payment button continues to be presented

Note: The backend does NOT emit different codes for different capture failures from Stripe, however. They all have the same code: `wcpay_capture_error`. See https://github.com/Automattic/woocommerce-payments/blob/6b2e9a5c9f372fdf9d585220846ac0154e83dea6/includes/admin/class-wc-rest-payments-orders-controller.php#L125-L139

Note: We could add coverage for other errors like wcpay_payment_uncapturable and wcpay_refunded_order_uncapturable and the less likely wcpay_missing_order  if desired. I can open separate issues for any of those. See that same part of the plugin linked above for those. See below for a screenshot of how other errors are presented (i.e. no change)

Screenshots:

Error when we attempt to process a fractional quantity:

IMG_0204

Error when the endpoint simply doesn't exist (edit includes/admin/class-wc-rest-payments-orders-controller.php and rename the path):

IMG_0205

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
+++ b/WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
@@ -1490,7 +1490,7 @@ private extension OrderDetailsDataSource {
 
     func isOrderPaymentMethodEligibleForCardPayment() -> Bool {
         let paymentMethod = OrderPaymentMethod(rawValue: order.paymentMethodID)
-        return paymentMethod == .cod || paymentMethod == .none
+        return paymentMethod == .cod || paymentMethod == .woocommercePayments || paymentMethod == .none
     }
 
     func hasCardPresentEligiblePaymentGatewayAccount() -> Bool {
--- a/Yosemite/Yosemite/Stores/PaymentGatewayAccountStore.swift
+++ b/Yosemite/Yosemite/Stores/PaymentGatewayAccountStore.swift
@@ -96,7 +96,7 @@ private extension PaymentGatewayAccountStore {
 
                 onCompletion(.success(()))
             case .failure(let error):
-                onCompletion(.failure(error))
+                onCompletion(.failure(PaymentGatewayAccountError(underlyingError: error)))
                 return
             }
         })
@@ -121,5 +121,53 @@ private extension PaymentGatewayAccountStore {
 
         storage.deleteObject(storageAccount)
         storage.saveIfNeeded()
+    }
+}
+
+/// Models errors thrown by the PaymentGatewayAccountStore. Not to be confused with
+/// errors originating from the card readers. Those are defined in CardReaderServiceError.
+///
+public enum PaymentGatewayAccountError: Error, LocalizedError {
+    case orderPaymentCaptureError(message: String?)
+    case otherError(error: AnyError)
+
+    init(underlyingError error: Error) {
+        guard case let DotcomError.unknown(code, message) = error else {
+            self = .otherError(error: error.toAnyError)
+            return
+        }
+
+        /// See if we recognize this DotcomError code
+        ///
+        self = ErrorCode(rawValue: code)?.error(message: message ?? Localizations.defaultMessage) ?? .otherError(error: error.toAnyError)
+    }
+
+    enum ErrorCode: String {
+        case wcpayCaptureError = "wcpay_capture_error"
+
+        func error(message: String) -> PaymentGatewayAccountError {
+            switch self {
+            case .wcpayCaptureError:
+                return .orderPaymentCaptureError(message: message)
+            }
+        }
+    }
+
+    public var errorDescription: String? {
+        switch self {
+        case .orderPaymentCaptureError(let message):
+            /// Return the message directly from the store, e.g. in the case of fractional quantities, which are not allowed
+            /// "Payment capture failed to complete with the following message: Error: Invalid integer: 2.5"
+            return message
+        case .otherError(let error):
+            return error.localizedDescription
+        }
+    }
+
+    enum Localizations {
+        static let defaultMessage = NSLocalizedString(
+            "Unknown error",
+            comment: "Message presented when no error message is available."
+        )
     }
 }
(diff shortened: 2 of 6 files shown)


WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
```diff
@@ -1490,7 +1490,7 @@ private extension OrderDetailsDataSource {
 
     func isOrderPaymentMethodEligibleForCardPayment() -> Bool {
         let paymentMethod = OrderPaymentMethod(rawValue: order.paymentMethodID)
-        return paymentMethod == .cod || paymentMethod == .none
+        return paymentMethod == .cod || paymentMethod == .woocommercePayments || paymentMethod == .none
```
ctarda, line 1493: Can you please make sure to update the FAQS (PdfdoF-V-p2) to reflect this as well?

allendav:
This PR should be targeting `develop.` Will fix.

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
ctarda: There are tests to cover the visibility of the "collect payment" button that touch this logic. Could we please update the tests to reflect that .woocommercePayments is a valid payment method as well?

Yosemite/Yosemite/Stores/PaymentGatewayAccountStore.swift
```diff
@@ -123,3 +123,51 @@ private extension PaymentGatewayAccountStore {
...
+            /// "Payment capture failed to complete with the following message: Error: Invalid integer: 2.5"
+            return message
+        case .otherError(let error):
+            return error.localizedDescription
+        }
+    }
+
+    enum Localizations {
+        static let defaultMessage = NSLocalizedString(
+            "Unknown error",
```
ctarda, line 169: This is an open question, maybe worth answering in a separate PR 😉 

Is there a way we can make this message a bit more helpful for users and Happiness Engineers? I understand it's a delicate balance, because we might not really know what went wrong at this point.

ctarda (approved):
Tests well. I would suggest adding a test to cover for the new payment method supported. That test could be deferred to a separate PR (your choice).

:shipit:

allendav:
This PR has been superseded by https://github.com/woocommerce/woocommerce-ios/pull/5245 which targets the same changes but on develop. Code review feedback will be addresed on https://github.com/woocommerce/woocommerce-ios/pull/5245

WooCommerce/Classes/ViewModels/Order Details/OrderDetailsDataSource.swift
allendav: Added test in https://github.com/woocommerce/woocommerce-ios/pull/5245/commits/95ccbea50f1021663ba94f46010da48cadaeb2bb

allendav: @ctarda I've updated the FAQ. Would you make sure we have a matching change for Android?

Yosemite/Yosemite/Stores/PaymentGatewayAccountStore.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5245/commits/a461942b5e11dddeeaf99c5df3299b34ad87b942

Issue #5243, opened 2021-10-19
Order List filters


pmusolino:
Enable filtering orders by status or date range. This task also included a new design for the screen were tabs will be removed and a single order list will be displayed.

More details p91TBi-5L9-p2

### 💻  Tasks

- [x] Add a feature flag that shows/hides the order filter bar https://github.com/woocommerce/woocommerce-ios/pull/5248
- [x] Display the order filter bar https://github.com/woocommerce/woocommerce-ios/pull/5248
- [x] Remove tabs and keep a single order list https://github.com/woocommerce/woocommerce-ios/pull/5400
- [x] Enable filtering by order status and date range https://github.com/woocommerce/woocommerce-ios/pull/5423
- [x] Persist selected filters between sessions https://github.com/woocommerce/woocommerce-ios/pull/5481
- [x] Add filter for given ranges: Today, Last 2 Days, This Week, This Month https://github.com/woocommerce/woocommerce-ios/pull/5374
- [x] Remove status filters from search view https://github.com/woocommerce/woocommerce-ios/pull/5442
- [x] Add filter for custom date range https://github.com/woocommerce/woocommerce-ios/pull/5374
- [x] Tracking events https://github.com/woocommerce/woocommerce-ios/pull/5454
- [x] Update empty state that now looks a bit different from the design. https://github.com/woocommerce/woocommerce-ios/pull/5444
- [x] Allow multiples selection of statuses https://github.com/woocommerce/woocommerce-ios/pull/5462

#### Filters: Order Status
![2](https://user-images.githubusercontent.com/495617/137908436-c2b83fb9-aabc-461b-89da-dc59c1296dab.jpg)

#### Filters: Date Range
![Schermata 2021-10-19 alle 14 24 22](https://user-images.githubusercontent.com/495617/137908511-ee7d89f3-9dab-439d-988a-55f4e7eda885.png)
![Schermata 2021-10-19 alle 14 25 34](https://user-images.githubusercontent.com/495617/137908672-316b4b2f-2936-4a3e-84a1-e2846b18ecd8.png)
Merchants will also be able to select a Date Range. Options include Today, Last 2 Days, This Week, This Month.

#### Custom Range
![Schermata 2021-10-19 alle 14 26 38](https://user-images.githubusercontent.com/495617/137908845-45205e4b-8549-4d60-bdef-9205c128438a.png)
[...]

PR #5248, opened 2021-10-20
Order list: feature flag and top bar that display order filters

pmusolino:
Part of #5243

## Description
As part of the Order List filters project, I added a new feature flag that will be useful to hide/show the new top bar in the Order List that will be useful for displaying and editing the filters on the order list.

## Testing
1. Open the app, and press the "Orders" tab.
2. You should see the new top bar that displays the label "All Orders" on the left, and a button "Filter" on the right.
3. Go into the code, and edit the feature flag `orderListFilters` under `DefaultFeatureFlagService`, changing the return to `false`.
4. Repeat from the beginning. This time, the top bar should be hidden.

## Screenshots
| Light mode             |  Dark mode |
Simulator Screen Shot - iPhone 13 Pro - 2021-10-20 at 15 57 59 | Simulator Screen Shot - iPhone 13 Pro - 2021-10-20 at 15 57 50

Update release notes:

### DIFF
--- a/Experiments/Experiments/DefaultFeatureFlagService.swift
+++ b/Experiments/Experiments/DefaultFeatureFlagService.swift
@@ -25,6 +25,8 @@ public struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .quickPayPrototype:
             return buildConfig == .localDeveloper || buildConfig == .alpha
+        case .orderListFilters:
+            return buildConfig == .localDeveloper || buildConfig == .alpha
         default:
             return true
         }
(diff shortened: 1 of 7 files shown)


itsmeichigo (approved):
✅  Tested and this works as expected for the feature flag. 

I notice on the task list that you planned for this feature, you intend to remove separate order tabs. I figure this filter header will then be moved to the order list instead? I'm approving this anyway because the focus of the PR is the feature flag.

pmusolino:
> I notice on the task list that you planned for this feature, you intend to remove separate order tabs. I figure this filter header will then be moved to the order list instead? I'm approving this anyway because the focus of the PR is the feature flag.

Since it will be one of the last tasks, the idea is to move (or mantain, we will evaluate later) it outside of the main container. For the moment, it's in the main container because we don't want to change the current behavior.

PR #5249, opened 2021-10-20
Added the new optional parameter `after` when fetching orders

pmusolino:
Part of #5243 

Added the parameter `after` inside the methods for fetching the orders (Networking + Yosemite layers).
From Woo API doc about the parameter `after`: `Limit response to resources published after a given ISO8601 compliant date.`
 This will be useful for completing #5243 where we will integrate the functionality for filtering orders in a range of dates. 

## Testing
- Just CI, still not used in our code.

Update release notes:


itsmeichigo (approved):
LGTM :shipit:

PR #5251, opened 2021-10-20
Quick Pay: Update tests to always use a concrete store settings

Ecarrion:
# Why

@pmusolino reported that some tests were failing locally, after a closer look, this was happening because the amount is formatted using the service locator currency store settings and the unit tests were assuming a US currency store settings.

# How

This is fixed by providing a concrete US currency store settings in tests.

# Testing steps

Run tests locally and make sure they pass.

Update release notes:


pmusolino (approved):
Tests now are ✅ thanks for fixing them 💯

PR #5252, opened 2021-10-20
Update GeneralAppSettings with a custom decoding

Ecarrion:
# Why

While working on adding the quick pay prototype behind the experimental features screen I noticed that at some point my "View Add-Ons" feature switch value got lost.

After inspecting, I noticed that adding(or updating/removing) a property to `GeneralAppSettings` makes the previously-stored `GeneralAppSettings` decoding fail because the data structure does not match the new type definition. Then a new `GeneralAppSettings` is created and stored with default values.

# How

To fix this, a custom decoding for `GeneralAppSettings` has been added, where if a property is not present or is different a default value is given but without aborting the whole decoding process.

# Testing Steps

- Launch the app, navigate to the experimental features screen and make sure the view add-ons feature is on.

- Modify `GeneralAppSettings` to include a new property.
```swift
public var someValue: String = "dsa"
```

- Launch the app and make sure the view add-ons feature is still on.

Update release notes:

### DIFF
--- a/Storage/Storage/Model/GeneralAppSettings.swift
+++ b/Storage/Storage/Model/GeneralAppSettings.swift
@@ -71,3 +71,18 @@ public struct GeneralAppSettings: Codable, Equatable, GeneratedCopiable {
         )
     }
 }
+
+// MARK: Custom Decoding
+extension GeneralAppSettings {
+    /// We need a custom decoding to make sure decoding doesn't fail when this type is updated (eg: when adding new properties)
+    /// Otherwise we risk to lose previously stored information
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.installationDate = try container.decodeIfPresent(Date.self, forKey: .installationDate)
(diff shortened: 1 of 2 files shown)


allendav:
> After inspecting, I noticed that adding(or updating/removing) a property

Does updating a property's value trigger the failure?

Asking for a friend: https://github.com/woocommerce/woocommerce-ios/pull/4499/files#diff-589afa6af8ad1530b89e871a8a3387339026dd4e6b57d837ae1613fad6084fd5R262

Ecarrion:
> Does updating a property's value trigger the failure?

@allendav Updating values is fine! it is just when updating the type definition.

Storage/Storage/Model/GeneralAppSettings.swift
```diff
@@ -71,3 +71,20 @@ public struct GeneralAppSettings: Codable, Equatable, GeneratedCopiable {
         )
     }
 }
+
+//// MARK: Custom Decoding
```
ealeksandrov, line 75: nit:
```suggestion
// MARK: Custom Decoding
```

ealeksandrov (approved):
LGTM! Nice work spotting and finding the issue! 🚀

PR #5253, opened 2021-10-20
Quick Pay: Hide quick pay behind the beta features screen

Ecarrion:
# Why

We will be releasing the Quick Pay prototype under an experimental flag. This PR enables that.
PS: It is still under the local feature flag, which will be removed later.

# How

- Update `GeneralAppSettings` to add a field to store the value of the feature toggle.

- Add Yosemite actions to save and load the feature toggle state.

- Update `BetaFeaturesViewController` to render and modify the state of the toggle(when the local feature flag is enabled)  

- Update `OrdersRootViewController` to check the toggle value on `ViewWillAppear` and render the appropriate navigation buttons.

# Demo

https://user-images.githubusercontent.com/562080/138187514-2aeec3dc-f501-47c8-b976-c35cca4b8769.mov

# Testing Steps

- Launch the app and see that the "+" button in the orders list is not present.
- Go to the beta features screen and enable the "Quick Pay" toggle. 
- Go back to the orders list and see that the "+" button in the orders list is present.

Update release notes:

### DIFF
--- a/Storage/Storage/Model/Copiable/Models+Copiable.generated.swift
+++ b/Storage/Storage/Model/Copiable/Models+Copiable.generated.swift
@@ -9,12 +9,14 @@ extension GeneralAppSettings {
         installationDate: NullableCopiableProp<Date> = .copy,
         feedbacks: CopiableProp<[FeedbackType: FeedbackSettings]> = .copy,
         isViewAddOnsSwitchEnabled: CopiableProp<Bool> = .copy,
+        isQuickPaySwitchEnabled: CopiableProp<Bool> = .copy,
         knownCardReaders: CopiableProp<[String]> = .copy,
         lastEligibilityErrorInfo: NullableCopiableProp<EligibilityErrorInfo> = .copy
     ) -> GeneralAppSettings {
         let installationDate = installationDate ?? self.installationDate
         let feedbacks = feedbacks ?? self.feedbacks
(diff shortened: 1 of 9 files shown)


rachelmcr (approved):
These changes all look good and the switch works well on a store with in-person payments (IPP) set up. 👍 

I just want to confirm that the behavior is what we want when the store isn't set up for IPP. When I test with a store that doesn't have IPP configured, here's what I see:

* I see the "Quick Order" toggle in experimental features and can enable it.
* I don't see the + on the Orders screen (since that also depends on IPP being configured).

Is that what we want/expect here? Since there's no other information about how to set up your store for Quick Pay, I wonder if we should use the same conditions for showing the experimental features toggle that we use for showing the + button on the Orders screen (so only show the toggle for stores that have IPP configured). Alternately, should we show the toggle for all stores but include additional information about what's needed to enable it?

Ecarrion:
@rachelmcr Nice catch, I totally forgot about that variable 🤦 :shame:

I will update the PR to only show the toggle to stores onboarded on in-person payments.

Ecarrion:
@rachelmcr Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5253/commits/c5944d54bbd1821aa032c572db650cadd47d3a26

https://user-images.githubusercontent.com/562080/138284780-28c23195-4029-4449-8a73-451f29a7e765.mov

PR #5256, opened 2021-10-21
Handle an edge case when switching to a newly connected site

jaclync:
Edge case for #5032 

### Background

When launching the app / logging in / opening the site picker, we currently sync all the sites that the app user is connected to via `AccountAction.synchronizeSites` through the [`me/sites` endpoint](P2gHKz-g-p2.1/get/me/sites/). We persist these sites in storage after each sync. On the other hand, when launching the app / logging in / switching stores, we update the global selected site ID (`sessionManager.defaultStoreID`) first and then the global selected site object (`sessionManager.defaultSite`) in storage from the latest sync.

Now that we are launching multi-store push notifications (hopefully in 7.9), there is a new trigger to switch stores. When the user taps on a push notification, the app switches to the store of the push notification and displays order/review details. There is no problem in production right now because push notifications can only come from the selected store, and the app does not have to switch stores. With multi-store push notifications, the app already handles store switching well except for an edge case when the app does not know about the store yet (not found in storage).

### Problem

When switching stores, there are a few things that happen in the app to update UI for the new store:

- `MainTabBarController` observes `StoresManager.siteID` and updates the tabs with the new site ID ——  ✅ the main tabs load data with the selected **site ID**, not site object
- When store switching completes, `SwitchStoreNoticePresenter` displays an in-app notice "Switched to {{site name}}" based on `sessionManager.defaultSite?.name` —— ⚠️ since the site is new to the app and we don't currently sync sites after switching stores, [`sessionManager.defaultSite` isn't updated](https://github.com/woocommerce/woocommerce-ios/blob/f12be96bbdb8d1c7970a8d0efb8275c2a18ec0f2/WooCommerce/Classes/Yosemite/DefaultStoresManager.swift#L404-L410) and the store name is from the previous store
- In the Dashboard tab, the store name is shown in the navigation bar based on `sessionManager.defaultSite?.name` —— ⚠️ similar to above, the store name is not updated for the new store
- There are other instances in the app that use `sessionManager.defaultSite` to display UI —— ⚠️ similar to above, the info that depends on `sessionManager.defaultSite` could be inaccurate

To summarize, the cause for the visual issues above with inaccurate store name is from `sessionManager.defaultSite` not being updated when the new store is not in storage.

### PR fix

- Sync the site from the server when switching stores if it is not found in storage, so that `sessionManager.defaultSite` is either nil or the site object of the selected site ID —— major changes in `DefaultStoresManager`, `AccountAction`, `AccountStore`
- Update the most prominent UI after switching stores that depends on `sessionManager.defaultSite` (in-app notice and Dashboard store name). I replace the `sessionManager.defaultSite` usage with an observable for the currently selected site `StoresManager.site`. This way, `SwitchStoreNoticePresenter` can subscribe to `StoresManager.site` and present the in-app notice only when the site with the new site ID becomes available. In `DashboardViewController`, it can subscribe to `StoresManager.site` and show the store name correctly —— major changes in `StoresManager`, `SessionManager` for the observable, and `DashboardViewController`, `SwitchStoreNoticePresenter` to subscribe to the observable

## Changes

The changes are a bit scattered, the major changes are in `DefaultStoresManager` to load or sync the target site whenever site ID changes. You can also review by commits as the following:

- Updated `AccountAction.loadSite` to `loadAndSynchronizeSiteIfNeeded` so that it syncs the site if the site does not exist locally https://github.com/woocommerce/woocommerce-ios/commit/d21bb5ca6e599d1bf0cfc03c6fb40f71d96e77b7
- Refactoring: update siteID observable from our own Observables implementation to Combine in `StoresManager` https://github.com/woocommerce/woocommerce-ios/commit/11cba2b540de61e7c8aad09c3b87c5dcfa4d9d66 https://github.com/woocommerce/woocommerce-ios/commit/7d893178898304b330080169665a3d36402f296b
- Added a site observable to `StoresManager`, and presented switch store notice with the site observable https://github.com/woocommerce/woocommerce-ios/commit/61e7cd16c373cad95aeb1662887e20e686f07264 (more comments added in https://github.com/woocommerce/woocommerce-ios/commit/4fe3c8ed6239cbb7bc35447c1dcba466106abd18)
- Updated the site name label in `DashboardViewController` with the site observable, and always showed the title as "My store" based on design

## Testing

Prerequisite: prepare a self-hosted WooCommerce site that the app user has not linked to Jetpack yet
- Launch the app, and make sure the app can receive push notifications
- Navigate to Settings → Switch Store --> make sure the site in the prerequisite does not appear in the site list
- Put the app in the background or close the app in order to receive push notifications
- In web, log in to the site in the prerequisite with the app account and connect to Jetpack
- Place an order or leave a product review on the newly connected site --> the app should receive a push notification
- Tap on the push notification --> the app should switch to the newly connected store and show the product/order details, and an in-app notice that says "Switched to {{site name}}." should appear
- Go to the Dashboard tab --> the site name should be shown below the navigation bar title

Update release notes:


WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
```diff
@@ -6,27 +7,46 @@ import Yosemite
...
-        }
+        observeSiteAndPresentWhenSiteNameIsAvailable()
+    }
+}
+
+private extension SwitchStoreNoticePresenter {
+    func observeSiteAndPresentWhenSiteNameIsAvailable() {
+        stores.site.compactMap { $0 }
+            .filter { $0.siteID == self.siteID }
+            .first()
```
Ecarrion, line 41: Is this `first` is because we don't expect the site name to change once it has been fetched?

WooCommerce/Classes/System/SessionManager.swift
```diff
@@ -115,6 +115,12 @@ final class SessionManager: SessionManagerProtocol {
 
     private let defaultStoreIDSubject: CurrentValueSubject<Int64?, Never>
 
+    var defaultSitePublisher: AnyPublisher<Site?, Never> {
+        defaultSiteSubject.eraseToAnyPublisher()
+    }
+
+    private let defaultSiteSubject: CurrentValueSubject<Site?, Never> = .init(nil)
+
```
Ecarrion: Lately, I've been wondering when we should use `@Published` property wrapper and when we should use a `CurrentValueSubject` with an erased publisher.

I think that the above code could be reduced to something like

```swift
@Published private(set) defaultSite: Site?
```

In the beginning, I thought that we should only use `@Published` when dealing with types that were consumed by `UIKit` but I've seen other parts of the app using `@Published` and I don't know anymore.

What do you think?

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -240,6 +232,16 @@ private extension DashboardViewController {
...
         spacerView.isHidden = true
     }
+
+    func updateUI(site: Site?) {
+        guard let siteName = site?.name, siteName.isEmpty == false else {
+            shouldShowStoreNameAsSubtitle = false
+            storeNameLabel.text = nil
+            return
+        }
+        shouldShowStoreNameAsSubtitle = true
```
Ecarrion, line 243: I think we are missing the feature flag condition here, right?
```swift
ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles)
```

Yosemite/Yosemite/Stores/AccountStore.swift
```diff
@@ -88,6 +88,23 @@ private extension AccountStore {
...
+    ///
+    func loadAndSynchronizeSiteIfNeeded(siteID: Int64, onCompletion: @escaping (Result<Site, Error>) -> Void) {
+        if let site = storageManager.viewStorage.loadSite(siteID: siteID)?.toReadOnly() {
+            onCompletion(.success(site))
+        } else {
+            synchronizeSites(selectedSiteID: siteID) { [weak self] result in
+                guard let self = self else { return }
+                guard let site = self.storageManager.viewStorage.loadSite(siteID: siteID)?.toReadOnly() else {
+                    onCompletion(.failure(SynchronizeSiteError.unknownSite))
+                    return
```
Ecarrion: Super nit: I think this can go in on line.
```swift
return onCompletion(.failure(SynchronizeSiteError.unknownSite))
```

Ecarrion (approved):
This is great @jaclync! Your description really made it easy for me to understand why all those changes were needed!

I only have concerns about not using the feature flag anymore
```swift
ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles)
```

But since bringing that(if needed) back should be straightforward I'm approving it anyway!

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
jaclync: my bad! I thought `.largeTitles` feature has been launched for a few releases and didn't add it back. But it's better to remove the feature flag separately, like as a subtask of https://github.com/woocommerce/woocommerce-ios/issues/3751. Added back in https://github.com/woocommerce/woocommerce-ios/pull/5256/commits/12f8caee21e1a037c82899d5a3e15b05bd40c9b3

WooCommerce/Classes/Authentication/Epilogue/SwitchStoreNoticePresenter.swift
jaclync: it's more like we only want to show the switch store notice once, as soon as the selected site object becomes available. Added a comment in https://github.com/woocommerce/woocommerce-ios/pull/5256/commits/08961ade5e63523ad3e904f28e8ec500507e4550

WooCommerce/Classes/System/SessionManager.swift
jaclync: Good idea! @Published var simplifies the code quite a bit and fits this use case. Updated in https://github.com/woocommerce/woocommerce-ios/pull/5256/commits/d24bd1de26262f016a61924e08e3caa7dca22b08

jaclync (commented):
Thank you @Ecarrion for reading the long description and feedback! 🙏  I updated the PR based on your suggestions, and tested it again. See if you'd like to take another look!

Ecarrion (approved):

PR #5257, opened 2021-10-21
[Mobile Payments] WisePOS E spike

ctarda:
This is a draft PR that explores what would be necessary to do to integrate the WisePOS E, with the only purpose of getting a ballpark estimation of the effort needed should we want to go ahead with the integration.

I have taken the shortest possible way, doing the bare minimum to be able to connect to a WisePOS E reader. This is definitely not ready or not intended to be formally reviewed.

But, yet again, it is a working prototype. 😄 

Update release notes:


ctarda:
The spike is done, we have a ballpark estimation of what would be needed to support the WisePOS E, so it's time to come this: pdfdoF-dj-p2

PR #5259, opened 2021-10-21
[Mobile Payments] Use system plugins, not site plugins response, to drive onboarding (for shop manager's sake)

### LINKED ISSUE #4960: [CPP Onboarding] Wrong error when trying to manage In-Person Payments as Shop Manager
I logged in to my test store with a different account than usual that has the Shop Manager role. When I go to Settings > In-Person Payments, I see an error saying the WCPay plugin isn't installed:

Simulator Screen Shot - iPhone 12 mini - 2021-09-09 at 10 53 54

allendav:
Closes #4960

Why are we doing this?
- To allow Shop Managers to have a proper on-boarding experience too. This fixes a problem where relying on WordPress's REST API endpoint for plugin enumeration fails for non Administrators by replacing it with a Shop Manager friendly endpoint in WooCommerce core.

What changed?
- Switch from SitePluginStore -> SystemStatusStore in CardPresentPaymentsOnboardingUseCase and tests
- Adds insertSampleSystemPlugin to MockStorageManager+Samples
- The "System Plugins" remote and mapper are actually a "System Status" remote and mapper, so I renamed them SystemPluginsRemote, SystemPluginsMapper -> SystemStatusRemote, SystemStatusMapper to make maintenance and understanding simpler down the road (and since we need more than just the active plugins nested object now
- Updated the JSON sample with a new one that includes active and inactive plugins
- Changed SystemStatusMapper to return both active and inactive plugins in a single array, using the networkActivated as a "hack" for now to indicate active or inactive state. See #5269 for our next step to undo this "hack"
- Updated CardPresentPaymentsOnboardingUseCase to look at networkActivated to determine if WooCommerce Payments is active or not
- Merged up from `release/7.8` to pick up fix for #5262 

What's next?
- Undo the networkActivated workaround we've introduced here by doing this with Core Data properly in #5269

To Test:

```
Install WooCommerce Payments 3.1.0 on the site but do not activate it
Log in to the app as a user with Shop Manager role
Navigate to Settings (gear) and then In-Person Payments
Tap on In-Person Payments
The “Connecting to your account” page will appear briefly
Ensure you are prompted to “Activate WooCommerce Payments”
Activate WooCommerce Payments in wp-admin
Tap on “Refresh After Activating”
The “Connecting to your account” page will appear briefly
Ensure you are now able to Order, Manage, or read the Manual for the card reader
Exit the App

Install WooCommerce Payments 2.4.0 on the site but do not activate it
Log in to the app as a user with Shop Manager role
Navigate to Settings (gear) and then In-Person Payments
Tap on In-Person Payments
The “Connecting to your account” page will appear briefly
Ensure you are prompted that you have installed an “Unsupported WooCommerce Payments version”
Activate WooCommerce Payments 2.4.0
Tap on “Refresh After Updating”
The “Connecting to your account” page will appear briefly
Ensure you are STILL prompted that you are running an “Unsupported WooCommerce Payments version”
In wp-admin, update to the latest WooCommerce Payments (e.g. 3.1.0)
Tap on “Refresh After Updating”
The “Connecting to your account” page will appear briefly
Ensure you are now able to Order, Manage, or read the Manual for the card reader

Activate all plugins 
Ensure settings still loads ( since `"inactive_plugins":[]`)

```

Props @fernandofgfer for the initial implementation of the system status active plugins remote and mapper

Update release notes:


Networking/Networking/Mapper/SystemStatusMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+            .siteID: siteID
+        ]
+
+        let systemStatus = try decoder.decode(SystemStatusEnvelope.self, from: response).systemStatus
+
+        /// For now, we're going to override the networkActivated Bool in each plugin to convey active or inactive -- in order to
+        /// avoid a core data change to add a Bool for activated
+        /// This will be undone in #5269
+        let activePlugins = systemStatus.activePlugins.map {
+            $0.overrideNetworkActivated(isNetworkActivated: true)
```
ctarda, line 26: I am not sure if this is already known, but we could also rely on the existing lenses for this. Something like this would also work:
```
        let activePlugins = systemStatus.activePlugins.map {
            $0.copy(networkActivated: true)
        }
```

ctarda (approved):
Tests well, code looks good. I left a comment just for your information.

:shipit:

Networking/Networking/Mapper/SystemStatusMapper.swift
allendav: I did not know this. I'll keep it in mind for the next time this comes up since I'll be removing this right away in #5269 anyways?

PR #5264, opened 2021-10-22
Issue/5159 add product category filter

toupper:

## Description
In the context of https://github.com/woocommerce/woocommerce-ios/issues/5159, we implement the UI for the  product filter list view, and the logic behind. We add a new row for Product Category filtering in the `FilterListViewController`, so the user can access the product categories list and select a category so the products list can be filter by category as well. However, please note that the `ProductCategoryFilterListViewController` that should hold the product categories list is not added yet, to keep the PR size in an acceptable level.
Furthermore, in this PR the product category filter logic is added to the relevant elements (filter view models, product list and storage) and the `ProductCategory` type is adapted to play as a filter. Existing unit tests are enhanced and new ones added, covering the new features added (feature flag, product category filter)
Lastly, a feature flag is added so the feature can be hidden in Release builds, as the feature is not yet finished.

## Changes
- Add feature flag for this case
- Make `ProductCategory` conform to `FilterType`
- Add a view model for Product Category in FilterListViewController so the row can be shown if the feature flag is enabled
- Handle clear filter logic also for Product Category
- Add a case for categories in `FilterListValueSelectorConfig`, so it can react to the selection (for now a no-op until the UI for the product categories is added)
- Add string to localizable strings
- Add `ProductCategory` parameter to `AppSettingStore`

## Testing
With feature flag enabled:
- Open the app
- Go to Products tab
- Open Filter
- Product Category row should be shown at the end, without any specific category selected (Any):

Please check that the existing filter flow works as expected, and the new addition did not break anything:
- Adding a filter option
- Clearing filter options
- Showing products with filters

With the feature flag disabled:
- Disable the filter by categories feature by doing it in `DefaultFeatureFlagService` on by running the app in a release version
- Open the app
- Go to Products tab
- Open Filter
- Product Category row should **not** be shown

As in the previous case, please check that the existing filter flow works as expected, and the new addition did not break anything:
- Adding a filter option
- Clearing filter options
- Showing products with filters

## Next Steps
- Show filter product category list view
- Handle selection: UI, storage
- Filter product list by category: networking

Update release notes:

### DIFF
--- a/Experiments/Experiments/FeatureFlag.swift
+++ b/Experiments/Experiments/FeatureFlag.swift
@@ -53,4 +53,8 @@ public enum FeatureFlag: Int {
     /// Display the bar for displaying the filters in the Order List
     ///
     case orderListFilters
+
+    /// Allows to filter products by a product category, persisting it so the filter can remain after restarting the app
+    ///
+    case filterProductsByCategory
 }
--- a/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
@@ -227,7 +227,7 @@ private extension FilterListViewController {
                 let staticListSelector = ListSelectorViewController(command: command, tableViewStyle: .plain) { _ in }
                 self.listSelector.navigationController?.pushViewController(staticListSelector, animated: true)
             case .categories:
-                // WIP https://github.com/woocommerce/woocommerce-ios/issues/5159: Show filter products by category view controller
+                // TODO-5159: Show filter products by category view controller
                 break
             }
         }
--- a/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
@@ -53,7 +53,7 @@ enum FilterListValueSelectorConfig {
     // Standard list selector with fixed options
     case staticOptions(options: [FilterType])
     // Filter list selector for categories, retrieved dynamically
-    case categories
+    case productCategories
 }
 
 /// Contains data for rendering a filter type row.
@@ -226,7 +226,7 @@ private extension FilterListViewController {
                 }
                 let staticListSelector = ListSelectorViewController(command: command, tableViewStyle: .plain) { _ in }
                 self.listSelector.navigationController?.pushViewController(staticListSelector, animated: true)
-            case .categories:
+            case .productCategories:
                 // TODO-5159: Show filter products by category view controller
                 break
             }
--- a/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
@@ -1,5 +1,6 @@
 import UIKit
 import Observables
+import Yosemite
 
 /// The view model protocol for filtering a list of models with generic filters.
 ///
@@ -51,8 +52,8 @@ final class FilterTypeViewModel {
 enum FilterListValueSelectorConfig {
     // Standard list selector with fixed options
     case staticOptions(options: [FilterType])
-    // Example: Categories
-    case custom
+    // Filter list selector for categories, retrieved dynamically
+    case categories
 }
 
 /// Contains data for rendering a filter type row.
@@ -225,7 +226,7 @@ private extension FilterListViewController {
                 }
                 let staticListSelector = ListSelectorViewController(command: command, tableViewStyle: .plain) { _ in }
                 self.listSelector.navigationController?.pushViewController(staticListSelector, animated: true)
-            case .custom:
+            case .categories:
                 break
             }
         }
--- a/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
@@ -227,6 +227,7 @@ private extension FilterListViewController {
                 let staticListSelector = ListSelectorViewController(command: command, tableViewStyle: .plain) { _ in }
                 self.listSelector.navigationController?.pushViewController(staticListSelector, animated: true)
             case .categories:
+                // WIP https://github.com/woocommerce/woocommerce-ios/issues/5159: Show filter products by category view controller
                 break
             }
         }
--- a/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductListViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductListViewModel.swift
@@ -1,5 +1,6 @@
 import UIKit
 import Yosemite
+import Experiments
 
 /// `FilterListViewModel` for filtering a list of products.
 final class FilterProductListViewModel: FilterListViewModel {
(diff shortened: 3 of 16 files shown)


WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
```diff
@@ -225,7 +226,8 @@ private extension FilterListViewController {
                 }
                 let staticListSelector = ListSelectorViewController(command: command, tableViewStyle: .plain) { _ in }
                 self.listSelector.navigationController?.pushViewController(staticListSelector, animated: true)
-            case .custom:
+            case .categories:
+                // WIP https://github.com/woocommerce/woocommerce-ios/issues/5159: Show filter products by category view controller
```
toupper: I thought about doing this more prominent so we can let the other devs know that it is still work in progress:
- By adding a log to the Xcode console, or
- By opening a view controller stating that this is still Work in Progress

Do you think that is necessary? is there any common guideline for this case?

Experiments/Experiments/FeatureFlag.swift
```diff
@@ -53,4 +53,8 @@ public enum FeatureFlag: Int {
     /// Display the bar for displaying the filters in the Order List
     ///
     case orderListFilters
+
+    /// Allows to filter products by a product category, persisting it so the filter can remain after restarting the app
```
pmusolino, line 57: Q: what do you mean with "persisting it"?

WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
pmusolino: Usually no, just because is a WIP feature under a feature flag. BTW, we usually use this format:

```
// TODO-5159: Show filter products by category view controller
```

WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
```diff
@@ -225,7 +226,8 @@ private extension FilterListViewController {
                 }
                 let staticListSelector = ListSelectorViewController(command: command, tableViewStyle: .plain) { _ in }
                 self.listSelector.navigationController?.pushViewController(staticListSelector, animated: true)
-            case .custom:
+            case .categories:
```
pmusolino: Since the plan is to reuse the `FilterListViewController` also for `Orders` and in the future for other cases, it would be great to be more explicit here, like `productCategories`.

WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductListViewModel.swift
```diff
@@ -41,27 +49,51 @@ final class FilterProductListViewModel: FilterListViewModel {
...
     private let productStatusFilterViewModel: FilterTypeViewModel
     private let productTypeFilterViewModel: FilterTypeViewModel
+    private let productCategoryFilterViewModel: FilterTypeViewModel
+
+    private let featureFlagService: FeatureFlagService
 
     /// - Parameters:
     ///   - filters: the filters to be applied initially.
-    init(filters: Filters) {
+    init(filters: Filters, featureFlagService: FeatureFlagService = ServiceLocator.featureFlagService) {
```
pmusolino, line 58: nit: it would be great to add a TODO in the original issue https://github.com/woocommerce/woocommerce-ios/issues/5159 to remember to remove the feature flag parameter. 👍

WooCommerce/Resources/en.lproj/Localizable.strings
```diff
@@ -1,4 +1,4 @@
-﻿/* In Order List, the pattern to show the order number. For example, “#123456”. The %@ placeholder is the order number. */
+/* In Order List, the pattern to show the order number. For example, “#123456”. The %@ placeholder is the order number. */
```
pmusolino: Localizable files should never be modified. Is there a reason why there is this change here? 🙇

WooCommerce/Resources/en.lproj/Localizable.strings
```diff
@@ -2466,6 +2466,9 @@
 /* Row title for filtering products by product type. */
 "Product Type" = "Product Type";
 
+/* Row title for filtering products by product category. */
+"Product Category" = "Product Category";
+
```
pmusolino: Same question as above 👍

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
```diff
@@ -28,14 +40,63 @@ final class FilterProductListViewModelTests: XCTestCase {
...
-        let expectedCriteria = FilterProductListViewModel.Filters(stockStatus: nil, productStatus: nil, productType: nil, numberOfActiveFilters: 0)
+        let expectedCriteria = FilterProductListViewModel.Filters(stockStatus: nil,
+                                                                  productStatus: nil,
+                                                                  productType: nil,
+                                                                  productCategory: nil,
+                                                                  numberOfActiveFilters: 0)
         XCTAssertEqual(viewModel.criteria, expectedCriteria)
     }
+
+    func testCriteriaWithNonNilFiltersAndFilterProductByCategoryIsDisabled() {
```
pmusolino: nit: since we are modifying this old test class, it would be great to follow the new naming convention for tests, maybe changing it everywhere in this class. https://github.com/woocommerce/woocommerce-ios/blob/develop/docs/naming-conventions.md#test-methods

pmusolino (commented):
Looks good to me 💯 I just left some questions and some nits to be solved before approving/merging it. 
A thing that it's not clear to me, it's why the `Localizable.string` file has some changes. Maybe a simple careless mistake?

WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
toupper: Yeah, done in https://github.com/woocommerce/woocommerce-ios/pull/5264/commits/74d2beaf1fd7ad27ee56fbe5f72f309f245177aa

WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
toupper: Great, done in 299c507fabcbf948a5f5932e50b77085e77d0ff0

WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductListViewModel.swift
toupper: Great, I added it as a comment as I cannot change the description :)

WooCommerce/Resources/en.lproj/Localizable.strings
toupper: Sorry, I misunderstood the way we work with localizable strings, I thought it was our task to just add the strings into the english .strings file. I revert this commit in https://github.com/woocommerce/woocommerce-ios/pull/5264/commits/2f3efac4b976c520863a39fe7511156cac138f80

WooCommerce/Resources/en.lproj/Localizable.strings
toupper: Same as above, reverted in https://github.com/woocommerce/woocommerce-ios/pull/5264/commits/2f3efac4b976c520863a39fe7511156cac138f80 :)

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductListViewModelTests.swift
toupper: Indeed, it looks more readable now :) done in https://github.com/woocommerce/woocommerce-ios/pull/5264/commits/6968feaca234a088e14312a63168eda3bb12d2aa

Experiments/Experiments/FeatureFlag.swift
toupper: I mean, when the user chooses a filter or more, we save them so the next time the app is restarted it is automatically filtered by that selection. I can remove that last part if it is redundant, to rephrase it to make it more meaningful :)

pmusolino (approved):
LGTM :shipit: thanks for fixing everything @toupper

PR #5265, opened 2021-10-22
Product screen: Fix `Sort by` and `Filter` toolbar's visibility on first time login or after switching stores. 

### LINKED ISSUE #5250: The filter options in Product List are not visible after login or after changing store
@JorgeMucientes while testing the v.7.8 of the app (dowloaded from Testflight), discovered an issue on product list. The first time you login or whenever you change store, the filter options in product list are not visible. You have to kill the app to see them.

## Video

https://user-images.githubusercontent.com/495617/138118063-16cb0ee3-78d1-456c-b66c-40840147c352.MP4

selanthiraiyan:
Closes #5250 

## Description
The toolbar which consists of `Sort by` and `Filter` options isn't visible, when we load products from a store for the first time. 

We load products for the first time in the following scenarios,
1. After logging in for the first time
1.  After switching stores. 

## Observations
I took a look and identified that `func showOrHideToolBar()` is not being called after the products are loaded by the `resultsController`.
```swift
    func showOrHideToolBar() {
        toolbar.isHidden = filters.numberOfActiveFilters == 0 ? isEmpty : false
    }
```

This is happening because `reloadTableAndView()` function is not being called as we directly reload the tableView using `tableView.reloadData()` from `func removePlaceholderProducts()`.

Even though we set back the `onDidChangeContent` and `onDidResetContent` closures by calling `setClosuresToResultController(resultsController, onReload: reloadTableAndView)` in `func removePlaceholderProducts()`, the `func reloadTableAndView()` is not called because the content is already loaded before we set those closures.

Also, this issue occurs only when we fetch products for the first time because, `showOrHideToolBar()` is called from `viewDidLoad()` as well. When we relaunch the app `resultsController.isEmpty` will be false because of the cached/stored products, which makes the `toolBar` visible.

## Fix
To fix this, I am calling `showOrHideToolBar()` for `.syncing` case in the following function.

This makes sure that toolBar's visibility is updated every time syncing is finished. i.e. When products are loaded.

## Demo
https://user-images.githubusercontent.com/524475/138455971-5c310576-95e9-4024-8fb5-2a7bf0378d2f.mov

## Testing
1. Do a fresh install and login into the app using a woocommerce account, preferably with more than one store.
1. Go to `Products` tab and wait for the products to load for the first time.
1. Observe that `Sort by` and `Filter` options are visible.
1. Go to settings and switch the store.
1. Go back to Products tab and observe that `Sort by` and `Filter` options are visible after switching stores.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
@@ -875,6 +875,7 @@ private extension ProductsViewController {
             ensureFooterSpinnerIsStopped()
             removePlaceholderProducts()
             showTopBannerViewIfNeeded()
+            showOrHideToolBar()
         case .results:
             break
         }


jaclync (approved):
Thanks a lot for contributing and fixing this issue @selanthiraiyan! 💟 The PR LGTM :shipit: 

Code: ✅ it makes sense to update the toolbar visibility when leaving the syncing state
Testing: ✅ I was able to reproduce before the PR by logging out and in, then switching stores. after this PR, I was able to see the toolbar with the same steps

I do wonder why `resultsController.onDidChangeContent` callback (which triggers the toolbar visibility update) isn't called when the product results are imported to storage 🤔 

Also noticed a separate memory issue where there is more than one `ProductsViewController` likely from a retain cycle: https://github.com/woocommerce/woocommerce-ios/issues/5272

selanthiraiyan:
@jaclync Thank you for reviewing. 🙇‍♂️ 

> I do wonder why resultsController.onDidChangeContent callback (which triggers the toolbar visibility update) isn't called when the product results are imported to storage 🤔

That's because we set those callbacks to `nil` by calling `resultsController.stopForwardingEvents()` while displaying ghost content in tableView. And, we are setting those callbacks back in `func removePlaceholderProducts()` only after leaving `.syncing` state, which I believe means that the product results are already imported into storage.

jaclync:
Makes sense @selanthiraiyan! Seems like `resultsController.onDidChangeContent` is triggered before the sync completion block where `onDidChangeContent` is set again.

Merging now, thanks again for contributing and sharing your findings with us! 🎉

PR #5266, opened 2021-10-22
Order list: implemented the Order Filters screen without actions

pmusolino:
Part of #5243 

## Description
In this PR, I implemented the screen for showing the first filters (Order Status and Date Range) that will be used for filtering the Order List. I implemented `FilterOrderListViewModel` which conforms to `FilterListViewModel`. The idea is to reuse `FilterListViewController`, the same components used in Products list for filtering the results.

## Testing
1. Launch the app it in debug (because of the feature flag).
2. Open the order list.
3. You should be able to tap the `Filter` button.
4. You should be able to view the `Filters` screen with two options: Order Status and Date Range.
5. Under Order Status, you should be able to select just one filter.
6. Under Date Range, you should be able to select just one filter (for the moment the custom range option is one of the selectable options. In a future PR, it will become an action).
7. After selecting at least an option, you should see in the Filters screen the "Clear all" button. Pressing it, should clear the filters.
8. Press Show Orders button. The view should be dismissed (no filters will be applied to order list for the moment. The functionality will be implemented in future PRs).

## Screenshots
| Filters screen            |  Order Status filters | Date Range filters |
Simulator Screen Shot - iPhone 13 Pro - 2021-10-22 at 14 51 17 | Simulator Screen Shot - iPhone 13 Pro - 2021-10-22 at 14 51 24 | Simulator Screen Shot - iPhone 13 Pro - 2021-10-22 at 14 51 28

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Order Filters/FilterOrderListViewModel.swift
```diff
@@ -0,0 +1,120 @@
...
+                       dateRange: dateRange,
+                       numberOfActiveFilters: numberOfActiveFilters)
+    }
+
+    func clearAll() {
+        let clearedOrderStatus: OrderStatusEnum? = nil
+        orderStatusFilterViewModel.selectedValue = clearedOrderStatus
+
+        let clearedDateRange: OrderDateRangeFilterEnum? = nil
+        dateRangeFilterViewModel.selectedValue = clearedDateRange
```
jaclync, line 60: nit: is it possible to set `selectedValue` to `nil` directly?

WooCommerce/Classes/ViewRelated/Orders/Order Filters/FilterOrderListViewModel.swift
```diff
@@ -0,0 +1,120 @@
...
+private extension FilterOrderListViewModel {
+    enum Localization {
+        static let filterActionTitle = NSLocalizedString("Show Orders", comment: "Button title for applying filters to a list of orders.")
+    }
+}
+
+private extension FilterOrderListViewModel.OrderListFilter {
+    enum Localization {
+        static let rowTitleOrderStatus = NSLocalizedString("Order Status", comment: "Row title for filtering orders by order status.")
+        static let rowTitleDateRange = NSLocalizedString("Date range", comment: "Row title for filtering orders by date range.")
```
jaclync: nit: in the [design](https://user-images.githubusercontent.com/495617/137908436-c2b83fb9-aabc-461b-89da-dc59c1296dab.jpg), "range" is capitalized 😄 
```suggestion
        static let rowTitleDateRange = NSLocalizedString("Date Range", comment: "Row title for filtering orders by date range.")
```

WooCommerce/WooCommerceTests/ViewRelated/Orders/Filters/FilterOrderListViewModelTests.swift
```diff
@@ -0,0 +1,41 @@
+import XCTest
+@testable import WooCommerce
+
+final class FilterOrderListViewModelTests: XCTestCase {
+    func test_criteria_With_default_filters() {
```
jaclync: super nit:

```suggestion
    func test_criteria_with_default_filters() {
```

Networking/Networking/Model/OrderStatusEnum.swift
```diff
@@ -56,6 +56,29 @@ extension OrderStatusEnum: RawRepresentable {
         case .custom(let payload):  return payload
         }
     }
+
+    /// Returns the localized text version of the Enum
+    ///
+    public var description: String {
```
jaclync: nit: maybe this variable can be moved to WooCommerce layer, and this doesn't have to be `public`?

WooCommerce/Classes/ViewRelated/Orders/Order Filters/FilterOrderListViewModel.swift
```diff
@@ -0,0 +1,120 @@
...
+            return Localization.rowTitleDateRange
+        }
+    }
+}
+
+extension FilterOrderListViewModel.OrderListFilter {
+    func createViewModel(filters: FilterOrderListViewModel.Filters) -> FilterTypeViewModel {
+        switch self {
+        case .orderStatus:
+            let options: [OrderStatusEnum?] = [nil, .pending, .processing, .onHold, .failed, .cancelled, .completed, .refunded]
```
jaclync, line 88: just wondering, do we plan to support custom order statuses that require fetching them like in order search placeholder?

jaclync (approved):
LGTM! :shipit: It's great that we are moving away from the top tabbed design which requires a workaround for the large titles 😄 Just some nits for your consideration!

WooCommerce/Classes/ViewRelated/Orders/Order Filters/FilterOrderListViewModel.swift
pmusolino: It was my first option, but the compiler says this.
Schermata 2021-10-26 alle 11 42 19

I believe that it needs to recognize a type conforming to `FilterType`.

WooCommerce/Classes/ViewRelated/Orders/Order Filters/FilterOrderListViewModel.swift
pmusolino: Nice catch, updated!

WooCommerce/Classes/ViewRelated/Orders/Order Filters/FilterOrderListViewModel.swift
pmusolino: For the moment it's not in the plan, we will allow filtering just by default statuses, but it's something that we could consider for a future improvement. 👍

pmusolino:
Thank you @jaclync for the review! Updated the code following your suggestions, and ready to merge this PR 🚀

PR #5267, opened 2021-10-22
Quick Pay: Adds protocol for CardPresentPaymentsOnboardingUseCase

Ecarrion:
# Why 

For the Quick Pay(Quick Order) experiment we are adding showing some banners conditionally when a store is onboarded for IPP.

To know if a store is onboarded we are using `CardPresentPaymentsOnboardingUseCase` but in order to unit test those conditions it is convenient to have a way to mock that use case instead of having to set the internal storage to the correct state.

# How 

- Creates a simple protocol to wrap the properties and functions that `CardPresentPaymentsOnboardingUseCase` exposes.

As `CardPresentPaymentsOnboardingUseCase` uses a `@Published` property wrapper, the protocol has to expose `state` and `statePublisher` due to swift unability to use `@Published` in protocols.

- Creates a Mock type for the new protocol to be used on tests.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
@@ -6,10 +6,36 @@ import Yosemite
 private typealias SitePlugin = Yosemite.SitePlugin
 private typealias PaymentGatewayAccount = Yosemite.PaymentGatewayAccount
 
-final class CardPresentPaymentsOnboardingUseCase: ObservableObject {
+/// Protocol for `CardPresentPaymentsOnboardingUseCase`.
+/// Right now, only used for testing.
+///
+protocol IPPOnboardingUseCaseProtocol {
+    /// Current store onboarding state.
+    ///
+    var state: CardPresentPaymentOnboardingState { get set }
+
+    /// Store onboarding state publisher.
+    ///
+    var statePublisher: Published<CardPresentPaymentOnboardingState>.Publisher { get }
+
+    /// Resynchronize the onboarding state if needed.
+    ///
+    func refresh()
+
+    /// Update the onboarding state with the latest synced values.
+    ///
+    func updateState()
+}
+
+final class CardPresentPaymentsOnboardingUseCase: IPPOnboardingUseCaseProtocol, ObservableObject {
     let storageManager: StorageManagerType
     let stores: StoresManager
+
     @Published var state: CardPresentPaymentOnboardingState = .loading
+
+    var statePublisher: Published<CardPresentPaymentOnboardingState>.Publisher {
+        $state
+    }
 
     init(
         storageManager: StorageManagerType = ServiceLocator.storageManager,
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
@@ -9,7 +9,7 @@ private typealias PaymentGatewayAccount = Yosemite.PaymentGatewayAccount
 /// Protocol for `CardPresentPaymentsOnboardingUseCase`.
 /// Right now, only used for testing.
 ///
-protocol IPPOnboardingUseCaseProtocol {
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
```diff
@@ -6,11 +6,37 @@ import Yosemite
 private typealias SitePlugin = Yosemite.SitePlugin
 private typealias PaymentGatewayAccount = Yosemite.PaymentGatewayAccount
 
-final class CardPresentPaymentsOnboardingUseCase: ObservableObject {
+/// Protocol for `CardPresentPaymentsOnboardingUseCase`.
+/// Right now, only used for testing.
+///
+protocol IPPOnboardingUseCaseProtocol {
```
allendav: Please replace IPP with CardPresent - it should help someone navigating the codebase tie it together in their head

Ecarrion: Sure! https://github.com/woocommerce/woocommerce-ios/pull/5267/commits/ed0d08e2f9165a446ee45627e55f78c2473244a5

WooCommerce/Classes/ViewRelated/Dashboard/Settings/In-Person Payments/CardPresentPaymentsOnboardingUseCase.swift
```diff
@@ -6,11 +6,37 @@ import Yosemite
...
+    /// Current store onboarding state.
+    ///
+    var state: CardPresentPaymentOnboardingState { get set }
+
+    /// Store onboarding state publisher.
+    ///
+    var statePublisher: Published<CardPresentPaymentOnboardingState>.Publisher { get }
```
allendav, line 19: Hmmm @koke advised against this in a PR of mine - see https://github.com/woocommerce/woocommerce-ios/commit/992361eaf8315e314396552dcbcc447a008679a3 for an alternative

Let me know what you think and I can continue the review

Ecarrion: @allendav That makes sense but I fail to see the risk you are referring to. The compiler won't let you use a `@Publihsed` property in a struct

Screen Shot 2021-10-22 at 9 53 34 AM

I see that `CardPresentPaymentsOnboardingUseCase` does conform to `ObservableObject` so I'm assuming what we do need that `ObservableObject` functionality?

In that case, I could update `CardPresentPaymentsOnboardingUseCaseProtocol` to
```swift
protocol CardPresentPaymentsOnboardingUseCaseProtocol: ObservableObject {
    // Code
}
```

Which will force consumers to be classes and will be more clear.

What do you think?

allendav: We were trying to avoid coding to what's in effect an unpublished API.

We do need the ObservableObject functionality for the views that depend on the onboarding state. I think your proposed update should work. You might not need to force clients to be classes if this generic approach holds water: https://stackoverflow.com/questions/59503399/how-to-define-a-protocol-as-a-type-for-a-observedobject-property

Ecarrion: > We were trying to avoid coding to what's in effect an unpublished API.

I don't think this is a [undocumented API](https://developer.apple.com/documentation/combine/published/publisher) but I do agree that it could contain more documentation 😅 

My rationale for exposing the `Publisher` type is that I don't want to change the current interface that `CardPresentPaymentsOnboardingUseCase` is exposing.

Consumers, like `InPersonPaymentsViewModel`, are accessing the `publisher` directly by using their `$state` property. And this protocol only aims to abstract that implementation but not change it. Does that makes sense?

If you are still feeling strong about it I can change it tho, as you are more familiar with this module. 😅

allendav: I experimented a bit with this yesterday and the more I think about it, the more I think this pattern is fine after all:

```Published<CardPresentPaymentOnboardingState>.Publisher { get }```

Go for it.

allendav (approved):

PR #5268, opened 2021-10-22
What's New: Prevent scrolling/scroll bounce when the feature list fits on screen

### LINKED ISSUE #4915: [What's New Component] Prevent scrolling when there are just a few features to show
### Description 📜
When there are just a few features to show on What's New component, it is possible to scroll the list, causing an odd behavior to users.

### Goal 🎯
Disable scrolling when there's just a couple of features to display

### Evidence 📱
https://user-images.githubusercontent.com/997521/131918839-1755293e-9815-4861-a344-38382bc91d59.mov

rachelmcr:
Fixes: #4915

## Description

On the What's New screen, when the feature list fits on screen it shouldn't scroll (or have any scroll bounce). This changes the layout to avoid that extra scroll/scroll bounce.

It also fixes a small issue with the spacing of list items on iPad (reduces the spacing to match designs).

## Changes

* Replaces the feature list `ScrollView` with a `VStack` using our modifier `scrollVerticallyIfNeeded()`. This allows the VStack to be scrolled when the list doesn't fit on the screen, but otherwise prevents scrolling.
* Replaces some of the `Spacer()` views with padding, to maintain the expected layout. (The `Spacer()` approach works with a `ScrollView` but not a `VStack`, since those views' heights are not the same.)
* Removes the vertical padding from `IconListItem` and replaces it with `Layout.listSpacing` on the list `VStack`. This allows the spacing between list items to be adjusted depending on the screen size, to better match the iPad design. This change also allows more list items to appear on screen in iPhone landscape mode.

Simulator Screen Recording - iPhone 12 Pro - 2021-10-22 at 15 47 07

Before|After
-|-
Before-iphone-portrait|After-iphone-portrait
Before-iphone-landscape|After-iphone-landscape
Before-ipad-portrait|After-ipad-portrait
Before-ipad-landscape|After-ipad-landscape

## Testing

⚠️ **Prerequisite** ⚠️ 

1. Build the app in debug mode (to ensure the feature flag is enabled) on an iPhone.
2. What's New component should be displayed upon app launch. (If not, you can go to Settings > What's New in WooCommerce to view it.)
3. With all the list items visible on screen (i.e. in iPhone portrait mode) confirm you _can't_ scroll the list.
4. With all the list items **not** visible on screen (i.e. in iPhone landscape mode) confirm you _can_ scroll the list.
5. Repeat steps 1-4 on an iPad and confirm the list doesn't scroll when all the items are visible on screen.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/IconListItem.swift
@@ -43,8 +43,6 @@ struct IconListItem: View {
             }
             Spacer()
         }
-        .padding(.horizontal, Layout.horizontalPadding)
-        .padding(.vertical, Layout.verticalPadding)
     }
 }
 
@@ -53,8 +51,6 @@ private extension IconListItem {
         static let iconSize = CGSize(width: 40, height: 40)
         static let contentSpacing: CGFloat = 16
         static let innerSpacing: CGFloat = 2
-        static let horizontalPadding: CGFloat = 40
-        static let verticalPadding: CGFloat = 16
     }
 }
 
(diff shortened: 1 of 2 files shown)


Ecarrion (approved):
Nice! 

Glad `scrollVerticallyIfNeeded()` was useful!

PR #5273, opened 2021-10-25
Fix multiple instances of `ProductsViewController` in memory after switching stores

### LINKED ISSUE #5272: ProductsViewController is not deallocated after switching stores
**Describe the bug**

After switching stores, each tab root view controller should be replaced including `ProductsViewController`. However, there is more than one `ProductsViewController` after switching stores. The count is the number of times switching stores + 1.

My guess is that there is a retain cycle from one of the callbacks in `ProductsViewController`.

**To Reproduce**

Prerequisite: the app account is connected to more than 1 store

- Launch the app in logged-in state
- Go to the products tab
- In Xcode, run memory debugger and filter by "ProductsViewController" --> there is one `ProductsViewController` instance
- Go to Settings > Switch Store, and switch to a different store
- In Xcode, run memory debugger and filter by "ProductsViewController" --> there are now two `ProductsViewController` instances

**Screenshots**

Screen Shot 2021-10-25 at 11 22 50 AM
Screen Shot 2021-10-25 at 11 22 37 AM

**Expected behavior**
There should always be one `ProductsViewController` instance in logged-in state.

**Mobile Environment**
Please include:
- Device: iPhone 11 simulator
- iOS version: iOS 14.5
- WooCommerce iOS version: 7.8.0.0

**WordPress Environment**
<details>
```
```
</details>

jaclync:
Fixes #5272 

## Why

When passing a closure that references `self`, if `self` isn't specified as weak it causes a retain cycle, and thus `ProductsViewController` is not deallocated after switching stores.

For example, after switching stores twice there are 3 instances of `ProductsViewController` and two of them are from the previous two stores:

Screen Shot 2021-10-25 at 11 22 37 AM

## Changes

Weak self when setting `ResultsController`'s callbacks.

## Testing

Prerequisite: the app account is connected to more than 1 store

- Launch the app in logged-in state
- Go to the products tab
- In Xcode, run memory debugger and filter by "ProductsViewController" --> there is one `ProductsViewController` instance
- Go to Settings > Switch Store, and switch to a different store
- In Xcode, run memory debugger and filter by "ProductsViewController" --> there should be only one `ProductsViewController` instance

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
@@ -65,7 +65,9 @@ final class ProductsViewController: UIViewController {
     ///
     private lazy var resultsController: ResultsController<StorageProduct> = {
         let resultsController = createResultsController(siteID: siteID)
-        configureResultsController(resultsController, onReload: reloadTableAndView)
+        configureResultsController(resultsController, onReload: { [weak self] in
+            self?.reloadTableAndView()
+        })
         return resultsController
     }()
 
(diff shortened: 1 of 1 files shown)


ealeksandrov (approved):
Nice find! After applying a branch I see only 1 instance of `ProductsViewController` :shipit:

PR #5275, opened 2021-10-25
Issue/5159 show product category filter list

toupper:

## Description

In the context of https://github.com/woocommerce/woocommerce-ios/issues/5159, we implement in this PR the `FilterProductCategoryListViewController` that shows the categories plus the "Any" filter row. Furthermore, we enhance the `productCategories` case of `FilterListValueSelectorConfig` with the `siteID` `Int64` necessary to fetch the relevant categories. Likewise, we pass that `siteID` to the `FilterProductListViewModel` from the `ProductsListViewController`, so it can be passed along. While the selection UI is implemented here, the selected product category is not passed to the filter yet, so products are not filtered by category yet. (will be implemented in the next PR)

## Challenges

In https://github.com/woocommerce/woocommerce-ios/pull/5230 we extracted the `ProductCategoryListViewController` from `EditProductCategoryListViewController`, so it can be reused in other cases, like in this one. There are however few features that are only specific to the filter product category list case:
- We should show an extra row on top with the title "Any"
- Unlike the case of `EditProductCategoryListViewController` where none or multiple categories can be selected, in this case **only and always one** should be selected (or Any)

To implement this business logic, we extend the functionality of `ProductCategoryListViewModel` with two protocols, 
`ProductCategoryListViewModelEnrichingDataSource`, that can modify the UI, and `ProductCategoryListViewModelDelegate`, that reacts to the selection event. The relevant logic for our case is encapsulated in `FilterProductCategoryListViewModel`, that conforms to these two protocols. It adds an extra row on top of the product category list (Any), and reacts to the selection event by reseting the categories so only one is selected, and updating the Any row selection state.  As you can see, nothing changes for the `EditProductCategoryListViewModel`, thus keeping its behaviour exactly as it was.

## Changes
- Pass that `siteID` to the `FilterProductListViewModel` from the `ProductsListViewController`
- Enhance the `productCategories` case of `FilterListValueSelectorConfig` with the `siteID` `Int64`
- Use it to retrieve product categories linked to that `siteID`
- Create `FilterProductCategoryListViewController` to show the product categories to be filtered
- Create `FilterProductCategoryListViewModel` to handle its logic
- Extend `ProductCategoryListViewModel` with `ProductCategoryListViewModelEnrichingDataSource` and `ProductCategoryListViewModelDelegate`
- Make `FilterProductCategoryListViewModel` conform to these protocols to implement its specific business logic
- Make `categoryID` in `ProductCategoryCellViewModel` optional so we can handle the "Any" cell case, that is not linked to any particular product category
- Refactor `ProductCategoryListViewController` initializer to accept the view model as parameter instead of multiple ones that are only used in the view model
- Move `MockProductCategoryStoresManager` to its own file, make it public
- Add Unit tests, TODOs

## Test
- Open the app
- Go to Products tab
- Open Filter
- Open Product Category:
  - Loading ghost effect should be visible
  - Product Categories are shown
  - Any row is shown on top
  - Only and always one product category (or the any row) can be selected (same as other filters)
  - UI looks good for landscape and portrait

## Next Steps
- Pass the selected product category along and filter the products according to it
- Remove the feature flag

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
@@ -1,4 +1,3 @@
-
 import Foundation
 import UIKit
 import Yosemite
--- a/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
@@ -1,0 +1,58 @@
+
+import Foundation
+import UIKit
+import Yosemite
+
+/// FilterProductCategoryListViewController: Displays the list of ProductCategories associated to the active Account,
+/// and allows the selection of one of them, or any.
+///
+final class FilterProductCategoryListViewController: UIViewController {
+
+    private let siteID: Int64
+    private let productCategoryListViewController: ProductCategoryListViewController
+    private let viewModel: FilterProductCategoryListViewModel
+
+    init(siteID: Int64) {
+        self.siteID = siteID
+        self.viewModel = FilterProductCategoryListViewModel()
+        let productCategoryListViewModel = ProductCategoryListViewModel(siteID: siteID,
+                                                                        selectedCategories: [],
+                                                                        dataSource: viewModel,
+                                                                        delegate: viewModel)
+
+        productCategoryListViewController = ProductCategoryListViewController(viewModel: productCategoryListViewModel)
+
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        configureProductCategoryListView()
+        configureNavigationBar()
+    }
+
+    func configureProductCategoryListView() {
+        addChild(productCategoryListViewController)
+        attachSubview(productCategoryListViewController.view)
+        productCategoryListViewController.didMove(toParent: self)
+    }
+
+    private func attachSubview(_ subview: UIView) {
+        subview.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(subview)
+        view.pinSubviewToAllEdges(subview)
+    }
+
+    private func configureNavigationBar() {
+        configureTitle()
+    }
+
+    private func configureTitle() {
+        title = NSLocalizedString("Categories", comment: "Filter product categories screen - Screen title")
+    }
+}
--- a/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
@@ -19,7 +19,7 @@ final class FilterProductCategoryListViewController: UIViewController {
         // TODO-5159: Initialize selected categories with the stored filter categories
         let productCategoryListViewModel = ProductCategoryListViewModel(siteID: siteID,
                                                                         selectedCategories: [],
-                                                                        dataSource: viewModel,
+                                                                        enrichingDataSource: viewModel,
                                                                         delegate: viewModel)
 
         productCategoryListViewController = ProductCategoryListViewController(viewModel: productCategoryListViewModel)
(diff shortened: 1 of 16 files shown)


WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
```diff
@@ -0,0 +1,62 @@
+
```
pmusolino: nit: this extra space can be removed

WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
```diff
@@ -0,0 +1,62 @@
...
+import Foundation
+import UIKit
+import Yosemite
+
+/// FilterProductCategoryListViewController: Displays the list of ProductCategories associated to the active Account,
```
pmusolino: nit: no needs to repeat the name of the view controller 👍

WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
```diff
@@ -0,0 +1,62 @@
...
+        view.pinSubviewToAllEdges(subview)
+    }
+
+    func configureNavigationBar() {
+        configureTitle()
+    }
+
+    func configureTitle() {
+        title = viewModel.title
+    }
```
pmusolino: nit: this seems redundant, maybe one method for setting the title is enough.

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductCategoryListViewModelTests.swift
```diff
@@ -0,0 +1,102 @@
...
+                                                                        name: NSLocalizedString("Any", comment: "Title when there is no filter set."),
+                                                                        isSelected: true,
+                                                                        indentationLevel: 0)
+        let enrichedViewModels = filterProductCategoryListViewModel.enrichCategoryViewModels([])
+        XCTAssertEqual(enrichedViewModels.first, expectedAnyCategoryViewModel)
+    }
+
+    func test_select_index_then_productCategoryListViewModel_reset_categories() {
+        // Given
+        let exp = expectation(description: #function)
```
pmusolino: nit: it would be great to use the new approach that we are using in tests, using `waitFor` instead of `expectation`

pmusolino (approved):
Looks great to me :shipit: I just left some minor nits and a suggestion on unit tests. I'll run the CI and merge it after the last changes 👍

WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
toupper: True, done in https://github.com/woocommerce/woocommerce-ios/pull/5275/commits/4faf255dad6726b66f12173409a67dc73b26a739

WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
toupper: Ooops :) removed in https://github.com/woocommerce/woocommerce-ios/pull/5275/commits/ba0826c2b34a541320034ac28cd798263457aa29

WooCommerce/Classes/ViewRelated/Products/Filters/FilterProductCategoryListViewController.swift
toupper: Yeah, I keep `configureTitle` and remove `configureNavigationBar` in https://github.com/woocommerce/woocommerce-ios/pull/5275/commits/13f44f82fefea4935360eb7193f9ab14168d168c

WooCommerce/WooCommerceTests/ViewRelated/Products/Filters/FilterProductCategoryListViewModelTests.swift
toupper: Great idea, done in https://github.com/woocommerce/woocommerce-ios/pull/5275/commits/ea870ac0c90acecdfbb9a29784f35ca9259a82f6

pmusolino (approved):
Thanks for taking care of all the suggestions 👍

PR #5276, opened 2021-10-25
Quick Pay: Adds announcement banners

Ecarrion:
# Why

In order to give the feature more discoverability and to allow merchants to give early feedback, this PR adds two quick pay banners on the order list screen.

The first banner will be shown when the store is IPP eligible but the merchant hasn't enabled the experimental feature

disabled

The second banner will be shown when the store is IPP eligible and the merchant has enabled the experimental feature.

enabled

**Note** Button actions will be done in the next PR. (This was getting quite big already)

# How

The order list screen already has a top banner set up when there is an error loading data, for this reason, this PR is refactoring that code in order to show a banner depending on the various states of the app.

The general rules are:

- If there is any error, show the error banner

- If there is no error, the store is IPP eligible and quick pay is disabled, show the quick pay disabled banner.

- If there is no error, the store is IPP eligible and quick pay is enabled, show the quick pay enabled banner.

To accomplish that, the main changes are:

- Created a `QuickPayTopBannerFactory` to create the new quick pay banners

- Updated `OrderListViewModel` to make the VM the one in charge of telling the VC what is the banner should be shown. I have turned some of the variables into published properties in order to combine those state streams.

- Updated `OrderListViewController` to not have a fixed top banner but have a dynamic one, that is driven by the view model content.

# Demo

https://user-images.githubusercontent.com/562080/138724892-0a9d9fe1-6e50-41ff-a9c3-6f04426437d4.mov

# Testing Steps

- Make sure you are in an IPP eligible store
- Go the exp features screen and make sure the quick pay toggle is off.
- Go to the order list and see a banner announcing the feature
- Toggle the quick pay exp feature
- Go back to the order list and see the banner with a "Give Feedback" button
- Turn of internet, and reload the order list
- See the error banner.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
@@ -301,7 +301,7 @@ extension OrderListViewController: SyncingCoordinatorDelegate {
         }
 
         transitionToSyncingState()
-        setErrorLoadingData(to: false)
+        viewModel.hasErrorLoadingData = false
 
         let action = viewModel.synchronizationAction(
             siteID: siteID,
@@ -314,7 +314,7 @@ extension OrderListViewController: SyncingCoordinatorDelegate {
 
                 if let error = error {
                     DDLogError("⛔️ Error synchronizing orders: \(error)")
-                    self.setErrorLoadingData(to: true)
+                    self.viewModel.hasErrorLoadingData = true
                 } else {
                     if pageNumber == self.syncingCoordinator.pageFirstIndex {
                         // save timestamp of last successful update
@@ -330,12 +330,6 @@ extension OrderListViewController: SyncingCoordinatorDelegate {
         }
 
         ServiceLocator.stores.dispatch(action)
-    }
-
-    /// Sets `hasErrorLoadingData` in the view model.
-    ///
-    private func setErrorLoadingData(to hasError: Bool) {
-        viewModel.hasErrorLoadingData = hasError
     }
 
     /// Sets the current top banner in the table view header
--- a/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
@@ -96,22 +96,9 @@ final class OrderListViewController: UIViewController {
 
     private let siteID: Int64
 
-    /// Top banner that shows an error if there is a problem loading orders data
+    /// Current top banner that is displayed.
     ///
-    private lazy var topBannerView: TopBannerView = {
-        ErrorTopBannerFactory.createTopBanner(isExpanded: false,
-                                              expandedStateChangeHandler: { [weak self] in
-                                                self?.tableView.updateHeaderHeight()
-                                              },
-                                              onTroubleshootButtonPressed: { [weak self] in
-                                                let safariViewController = SFSafariViewController(url: WooConstants.URLs.troubleshootErrorLoadingData.asURL())
-                                                self?.present(safariViewController, animated: true, completion: nil)
-                                              },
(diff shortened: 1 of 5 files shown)


WooCommerce/Classes/ViewRelated/Orders/QuickPay/QuickPayTopBannerFactory.swift
```diff
@@ -0,0 +1,54 @@
...
+        topBannerView.translatesAutoresizingMaskIntoConstraints = false
+        return topBannerView
+    }
+}
+
+private extension QuickPayTopBannerFactory {
+    enum Localization {
+        static let title = NSLocalizedString("Take payments from your device!", comment: "Title of the banner notice in the Quick Pay feature")
+        static let disabledInfo = NSLocalizedString("We are working on making it easier for you to take payments from your device! " +
+                                                    "Enable it from the Experimental Features section.",
```
rachelmcr: I don't have time for a full review right now, but I wanted to make a couple suggestions on this wording:

1. Instead of "Enable it" how about "Enable Quick Order"? (That way it's clear what feature needs to be enabled to be able to take payments.)
2. Do you think we should give specific instructions for where to find the Experimental Features section? Maybe "in Settings > Experimental Features."

Ecarrion: Thanks for the suggestions @rachelmcr! Would you be so kind to add your suggestions on p91TBi-6iQ#comment-6840 so they can be validated by product and design? 🙏

rachelmcr: Sure thing! I missed that discussion earlier, sorry about that. I added my suggestions at p91TBi-6iQ-p2#comment-6904 for confirmation.

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
```diff
@@ -324,20 +332,17 @@ extension OrderListViewController: SyncingCoordinatorDelegate {
         ServiceLocator.stores.dispatch(action)
     }
 
-    /// Sets `hasErrorLoadingData` in the view model and shows or hides the banner view accordingly
+    /// Sets `hasErrorLoadingData` in the view model.
     ///
     private func setErrorLoadingData(to hasError: Bool) {
         viewModel.hasErrorLoadingData = hasError
```
rachelmcr: Q: Since this method only sets `viewModel.hasErrorLoadingData` now, would it make sense to remove the method entirely and just set that value directly instead of calling a method for it?

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
```diff
@@ -205,6 +196,23 @@ private extension OrderListViewController {
...
+                    self.hideTopBannerView()
+                case .error:
+                    self.setErrorTopBanner()
+                case .quickPayEnabled:
+                    self.setQuickPayEnabledTopBanner()
+                case .quickPayDisabled:
+                    self.setQuickPayDisabledTopBanner()
+                }
+            }
+            .store(in: &cancellables)
```
rachelmcr, line 215: This is a really clear way to set the different banners! 🎉

rachelmcr (approved):
Looks great! I tested and confirmed the correct banners are shown (and hidden) as expected after switching between different states, and the error banner continues to work as expected on a site not enrolled in IPP. 👍 

Just left a small question (nitpicky, feel free to disregard it) below.

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
Ecarrion: Sure, why not!

Issue #5269, opened 2021-10-22
[Mobile Payments] [Core Data] Use SystemPlugin active attribute instead of networkActivated


allendav:
Replace the temporary overriding of `networkActivated` (as a proxy for plugin activated/not) introduced in #5259 with the 'active' attribute added by #5279 to the model for SystemPlugin in v56 of Core Data

Multiple places in the code have been commented with `5269` as an aid in executing this change

PR #5279, opened 2021-10-25
[Mobile Payments] Add `active` Bool to SystemPlugin and migrate Core Data to v56

allendav:
Partially addresses #5269 

Changes:
- This is the first of two planned PRs to properly reflect plugin activation status in core data. This PR is meant as the first step towards the more permanent fix to the temporary "hack" introduced in #5259 which stored the plugin activation status in the `networkActivated` attribute. We did this because the `networkActivated` Bool attribute was already in core data for SystemPlugin and not used by any client.
- This PR rolls our core data model version to 56, adds the `active` Bool as an attribute on the SystemPlugin model, and adds a migration test to ensure that when migrating SystemPlugins from v55 to v56, the `active` attribute is defaulted to YES/true
- #5259 targeted `release/7.8`. It will merge shortly to `develop`. After it does, we can undo the hack (tagged in the code contributed by #5259 with "5269")
- The name `active` (instead of `activated`) was chosen since the WooCommerce system status endpoint (from which SystemPlugins are obtained) divides the plugins in its response between `active_plugins` and `inactive_plugins`

To test:
- Run the migration test added in `test_migrating_from_55_to_56_adds_new_systemplugin_attribute_active_with_true_value` and ensure it completes successfully

Update release notes:


ctarda (approved):
I run develop first, then switched to this branch, and I didn't see any coredata warnings.

:shipit:

PR #5280, opened 2021-10-25
Quick Pay: Add banners button actions

Ecarrion:
# Why

#5276 introduced a couple of top banners for the quick pay prototype. This PR adds concrete actions to those banner buttons.

Actions added are:

- Dismiss: Closes the banner for this session
- Give Feedback: Launches a survey that allows merchants to give feedback about the prototype.

# How

- Added all the necessary cases to properly present the quick pay feedback survey.

- Update `OrderListViewController` to launch the survey when the user taps on the "Give Feedback" button.

- Update `OrderListViewModel` to expose a new published property `hideQuickPayBanners` to allow the view controller to hide the quick pay banners.

# Demo

https://user-images.githubusercontent.com/562080/138784401-9b1b0cbe-9187-4464-9c69-cdae2261397a.mov

# Testing Steps

- Make sure you are in an IPP eligible store.
- Go to the exp features screen and make sure the quick pay toggle is on.
- Go the the order list screen, expand the quick pay banner, and provide some feedback
- Tap on the dismiss button of the banner and see that it gets dismissed.

Note: Dismissing a banner in the "processing orders" tab won't dismiss it in the "all orders" tab.
As the solution is not that trivial, I'm still analyzing the best way to work around that.

Update release notes:

### DIFF
--- a/WooCommerce/WooCommerceTests/ViewRelated/Orders/OrderListViewModelTests.swift
+++ b/WooCommerce/WooCommerceTests/ViewRelated/Orders/OrderListViewModelTests.swift
@@ -356,7 +356,7 @@ final class OrderListViewModelTests: XCTestCase {
         }
     }
 
-    func test__does_not_show_banners() {
+    func test_dismissing_quick_order_banners_does_not_show_banners() {
         // Given
         let stores = MockStoresManager(sessionManager: .testingInstance)
         stores.whenReceivingAction(ofType: AppSettingsAction.self) { action in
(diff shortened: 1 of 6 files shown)


WooCommerce/WooCommerceTests/ViewRelated/Orders/OrderListViewModelTests.swift
```diff
@@ -355,6 +355,47 @@ final class OrderListViewModelTests: XCTestCase {
             viewModel.topBanner == .none
         }
     }
+
+    func test__does_not_show_banners() {
```
ealeksandrov: nit:
```suggestion
    func test_dismissing_quickPay_banner_does_not_show_banners() {
```

ealeksandrov (approved):
Code looks good, works as described. :shipit:

PR #5290, opened 2021-10-27
Product List: Use selected product category in filter

toupper:

## Description
In the context of Part of https://github.com/woocommerce/woocommerce-ios/issues/5159, the scope of this PR is to pass along the selected product category, so it can be stored and used to filter the products in the products view. Thus, that selected category is shown in the filter list menu, and the products network call includes now the product category filter information so the products can be filtered accordingly.

## Decisions
As mentioned above, we need to store the selected `ProductCategory` as the rest of the filters, so it can stay selected after the user closes and restarts the app. Because of that, I had to implement some additions when encoding a  ProductCategory, namely, `siteID` and `parentID` are now also encoded, so we can compare and select it in the list of categories.
Until now the `ProductCategory` `siteID` was not encoded: it was passed through the `userInfo` to the decoder after retrieving the categories remotely. In our case I preferred to encode it and decode it in the same way as the rest of properties, given that mimicking the former approach (using userInfo) would add more complexity to the local store process than needed. 
Precisely, we would have to add a `userInfo` parameter to the `FileStorage` protocol and the `PListFileStorage` class used in from `AppSettingsStore`. As this `userInfo` would be only used in our case, I find it a bit overkill to show that parameter and making it nil for all other cases. However, since with the new implementation in `ProductCategory` checks still wether the `siteID` is passed through the `Decoder` `userInfo`, decoding after fetching categories remotely works in the same way as before. Of course, any suggestions are welcome to improve this point.

## Changes
- Notify when a product category is selected in `ProductCategoryListViewModel`
- React to that event by updating the Filter List UI
- Show the stored filter `ProductCategory` as selected when opening `FilterProductCategoryListViewController`
- Add the product category filter parameter to the synchronize products network call
- Unit tests for the new cases. Adapt naming of related test functions to new convention

### Selecting a product category updates the filter list UI

- Open app
- Go to products
- Open filter 
- Select Product Category
- Select a product category
- Go back
- Selected product category is shown in the filter list UI:

Case when the app is closed and restarted should be checked as well to make sure that the filter product category has been properly stored

### Filter product category list starts with the previous category selected

- Open app
- Go to products
- Open filter 
- Select Product Category
- Select a product category
- Go back
- Open Product Category Filter List
- Selected product category is shown in the Product Category UI:

Case when the app is closed and restarted should be checked as well to make sure that the filter product category has been properly stored

### Clearing product category filter clears it in filter menu and product category filter list

Precondition: a product category filter should be selected

- Open app
- Go to products
- Open filter
- Press Clear All
- Filter should be cleared in filter list and product category filter list

### Products are filtered according to product category

- Open app
- Go to products
- Open filter 
- Select Product Category
- Select a product category
- Press "Show products" or
- Go back
- Press "Show products"
- Products list is filtered according to selected category. **Please note that the filter works for all the children categories. That is, if parent category is "Clothing" and is selected, and child category is "Shirts", "Shirts" products will be shown even if "Shirts" is not explicitly selected.**

## Next Steps
- Handle edge cases: errors, offline, categories removal
- Remove feature flag and update release notes

Update release notes:


Networking/Networking/Model/Product/ProductCategory.swift
```diff
@@ -56,13 +58,30 @@ public struct ProductCategory: Codable, Equatable, GeneratedFakeable {
...
         case name       = "name"
         case slug       = "slug"
         case parentID   = "parent"
     }
 }
 
+private extension ProductCategory {
+    /// Provides the siteID, that can be found as a encoded value or in the Decoder user info
+    ///
+    private static func siteID(from decoder: Decoder, container: KeyedDecodingContainer<ProductCategory.CodingKeys>) -> Int64? {
```
toupper, line 72: Here we retrieve the siteID as a encoded value (our case) or as part of the `Decoder.userInfo` (network fetching)

Yosemite/YosemiteTests/Stores/ProductStore+FilterProductsTests.swift
```diff
@@ -137,25 +144,58 @@ final class ProductStore_FilterProductsTests: XCTestCase {
...
 
         productStore.onAction(action)
         wait(for: [expectation], timeout: Constants.expectationTimeout)
     }
+
+    func test_synchronizeProducts_with_only_product_category_filter_then_it_sends_product_category_filter_param() {
+        let filterProductCategory = ProductCategory(categoryID: 213, siteID: 0, parentID: 0, name: "", slug: "")
+        let productStore = ProductStore(dispatcher: dispatcher, storageManager: storageManager, network: network)
+
+        let _: Bool = waitFor { [weak self] promise in
```
toupper, line 174: is there a common agreement to migrate the expectation tests to the new `waitFor` approach? shall I migrate all the other filter tests if I touch it? :)

Networking/Networking/Remote/ProductsRemote.swift
```diff
@@ -132,6 +136,16 @@ public final class ProductsRemote: Remote, ProductsRemoteProtocol {
         enqueue(request, mapper: mapper, completion: completion)
     }
 
+    /// Returns the category Id in string format, or empty string if the product category is nil
+    ///
+    private func filterProductCategoryParemeterValue(from productCategory: ProductCategory?) -> String {
```
pmusolino: I'd say to move this method under a private extension, maintaining only the public methods here.

WooCommerce/WooCommerceTests/ViewRelated/Products/Categories/ProductCategoryListViewModelTests.swift
```diff
@@ -43,7 +44,27 @@ final class ProductCategoryListViewModelTests: XCTestCase {
...
+        let productCategory = ProductCategory(categoryID: 443, siteID: 123, parentID: 0, name: name, slug: "")
+
+        let passedCategory: ProductCategory? = waitFor { [weak self] promise in
+            guard let self = self else {
+                return
+            }
+
+            let viewModel = ProductCategoryListViewModel(storesManager: self.storesManager, siteID: 0, onProductCategorySelection: { category in
+                promise(category)
+
```
pmusolino: nit: extra space that can be removed 👍

Yosemite/YosemiteTests/Stores/ProductStore+FilterProductsTests.swift
pmusolino: In general, if the work is not a big task we tend to update everything if possible when we are working on a specific class. So, up to you :) Not a big problem at the moment.

pmusolino (approved):
Very good job here César! 🚢 

Code ✅ 
Testing ✅ 

> Please note that the filter works for all the children categories. That is, if parent category is "Clothing" and is selected, and child category is "Shirts", "Shirts" products will be shown even if "Shirts" is not explicitly selected.

This makes sense. Not super intuitive for a new user, but makes sense. :nod:

Networking/Networking/Remote/ProductsRemote.swift
toupper: Makes total sense, done https://github.com/woocommerce/woocommerce-ios/pull/5290/commits/80f964380ce65d7b71eed6bd9b961ef11152e816

WooCommerce/WooCommerceTests/ViewRelated/Products/Categories/ProductCategoryListViewModelTests.swift
toupper: Indeed, removed in https://github.com/woocommerce/woocommerce-ios/pull/5290/commits/8b12ebebfb404376add1c28506b5206dda946be3

Yosemite/YosemiteTests/Stores/ProductStore+FilterProductsTests.swift
toupper: Then I migrated them all in https://github.com/woocommerce/woocommerce-ios/pull/5290/commits/fc43b2cd149e8127e86bed71c77eb605a7ad614b , it looks much better with `waitFor` :)

toupper:
@pmusolino thanks a lot! :blush:

PR #5292, opened 2021-10-27
What's New: Display base64 icons in feature announcements

### LINKED ISSUE #5291: What's New: Base64 icons are not included in announcement
**Describe the bug**

When an icon is uploaded to a feature announcement (instead of using an image URL) the image doesn't appear in the announcement.

It looks like the issue is that the `feature.iconBase64` string includes the prefix `data:image/png;base64,` before the image data, so the image data can't be decoded and used. Note how [WordPress removes this prefix](https://github.com/wordpress-mobile/WordPress-iOS/blob/42881bfca93bf616206c316131b447f0bbe58c68/WordPress/Classes/ViewRelated/What's%20New/Views/AnnouncementCell.swift#L56) before using the data.

**Screenshots**

rachelmcr:
Fixes: #5291

## Description

The "What's New" screen displays feature announcements that include an icon either from a URL or base 64 encoded. Icons that are base 64 encoded include the data URI scheme `data:image/png;base64,` in the string with the image data. This PR makes sure that data URI is removed before decoding the image data, so that those images can be displayed in the app.

## Changes

* Handles the `feature.iconBase64` string as a URL before decoding it, so the data URI scheme is handled as part of the URL.
* Adds unit tests to check that feature announcements contain the expected icon from either a URL or base 64 encoded data.

Before|After
-|-
Simulator Screen Shot - iPhone 13 Pro - 2021-10-27 at 12 33 08|Simulator Screen Shot - iPhone 13 Pro - 2021-10-27 at 12 48 38

## Testing

⚠️ **Prerequisite** ⚠️ 

1. Build the app in debug mode (to ensure the feature flag is enabled) on a device.
2. What's New component should be displayed upon app launch. (If not, you can go to Settings > What's New in WooCommerce to view it.)
3. Confirm you see icons appear next to each of the items in the list.

## Submitter Checklist

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewFactory.swift
+++ b/WooCommerce/Classes/ViewRelated/WhatsNew/WhatsNewFactory.swift
@@ -27,7 +27,7 @@ struct WhatsNewFactory {
     /// Get IconListItem.Icon from a Feature
     private static func icon(for feature: Feature) -> IconListItem.Icon? {
         var icon: IconListItem.Icon?
-        if let base64String = feature.iconBase64,
+        if let base64String = feature.iconBase64?.components(separatedBy: ";base64,")[safe: 1],
            let imageData = Data(base64Encoded: base64String),
            let image = UIImage(data: imageData) {
             icon = .base64(image)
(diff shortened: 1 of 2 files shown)


Ecarrion (approved):
Great!

PR #5293, opened 2021-10-27
What's New: Enable feature flag to release the feature

rachelmcr:
⚠️ Merge after #5292 ⚠️ 

## Description

This releases the "What's New" component, so we can display "What's New on WooCommerce" announcements. There are no scheduled announcements for upcoming releases yet, so this won't have any visible changes in the app yet.

## Changes

Enables the `whatsNewOnWooCommerce` feature flag to release the feature.

## Testing

1. If you have tested the What's New screen before, delete the app from your simulator/device to clear out any saved announcements from past testing.
2. Build and launch the app. Log in and confirm there are no visible changes on app launch or in Settings (since there are no scheduled announcements for upcoming releases yet).
3. Change the prerequisites noted below (to fake a different release) and confirm the test announcement displays on app launch.
4. Open Settings > What's New in WooCommerce and confirm the test announcement appears.

⚠️ **Prerequisites** ⚠️ 

## Submitter Checklist

Update release notes:

### DIFF
--- a/Experiments/Experiments/DefaultFeatureFlagService.swift
+++ b/Experiments/Experiments/DefaultFeatureFlagService.swift
@@ -19,8 +19,6 @@ public struct DefaultFeatureFlagService: FeatureFlagService {
             return true
         case .shippingLabelsMultiPackage:
             return true
-        case .whatsNewOnWooCommerce:
-            return true
         case .pushNotificationsForAllStores:
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .quickOrderPrototype:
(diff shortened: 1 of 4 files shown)


Ecarrion (approved):
What's the benefit of still having this under a feature flag that is always enabled rather than just removing the Feature flag? 😅

rachelmcr:
Ah! No particular reason; I can go ahead and remove the flag entirely. I was under the impression that it was normal to enable the feature flag for release and then remove it later, but there's really no need for that at least for this feature.

Ecarrion:
> Ah! No particular reason; I can go ahead and remove the flag entirely. I was under the impression that it was normal to enable the feature flag for release and then remove it later, but there's really no need for that at least for this feature.

I'll just go ahead and delete it, we can always revert the PR if something goes wrong later!

rachelmcr:
Feature flag removed entirely in 6e38b67. @Ecarrion Ready for another look!

Ecarrion (approved):

PR #5294, opened 2021-10-27
[Mobile Payments] Allow pass presentation after payment cancellation

### LINKED ISSUE #4649: [Mobile Payments] Restore Pass Presentation if Payment Collection is Cancelled
Related: #4218 

This was deferred from #4218 

Restore pass presentation in the event the merchant cancels payment collection

allendav:
Closes #4649 

@ctarda @joshheald  @jaclync - just one review needed from one of you, but all of you are welcome of course

Changes
- In CardPresentModalTapCard, instead of allowing the modal to create and enqueue the cancel payment action directly, it now accepts an onCancel closure to call when the user taps the button. This fixes a hole where the modal was bypassing PaymentCaptureOrchestrator.
- In OrderDetailsPaymentAlerts, allow that onCancel closure to be passed
- In OrderDetailsViewController, pass viewModel cancelPayment to tapOrInsertCard onCancel. Move the call to tracks (that used to be in the modal) here too
- Lastly, and most importantly, in PaymentCaptureOrchestrator, restore pass presentation in cancelPayment

To test:
- Use this branch to begin payment collection on a qualifying order with an iPhone and with the powered on reader directly in contact with the iPhone
- Ensure the iPhone’s wallet does NOT present
- Cancel the payment
- Remove the iPhone from contact with the reader
- Disconnect from the reader in Settings > In-Person Payments > Manage Card Reader
- Use this branch on another device (e.g. an iPad) to begin payment collection on that order
- Tap the iPhone to the reader
- Ensure the iPhone’s wallet DOES present
- Cancel the payment
- Begin payment collection again
- Tap a Point of Sale Test Card to the reader and ensure you can complete the payment

Update release notes:

### DIFF
--- a/Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
+++ b/Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
@@ -2,7 +2,6 @@ import Storage
 import Hardware
 import Networking
 import Combine
-import PassKit
 
 /// MARK: CardPresentPaymentStore
 ///
@@ -18,8 +17,6 @@ public final class CardPresentPaymentStore: Store {
 
     /// We need to be able to cancel the process of collecting a payment.
     private var paymentCancellable: AnyCancellable? = nil
-
-    private var walletSuppressionRequestToken: PKSuppressionRequestToken?
 
     public init(dispatcher: Dispatcher, storageManager: StorageManagerType, network: Network, cardReaderService: CardReaderService) {
         self.cardReaderService = cardReaderService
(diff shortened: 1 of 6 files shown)


Yosemite/Yosemite/Stores/CardPresentPaymentStore.swift
```diff
@@ -2,6 +2,7 @@ import Storage
 import Hardware
 import Networking
 import Combine
+import PassKit
```
ctarda: I am not sure we will want to import PassKit from Yosemite. 

I understand this (pass suppression) seems like business logic, but I think it is app-related business logic, and Yosemite should not need to be aware of it. If we ever wanted to share Yosemite across platforms or different apps this might end up being a problem.

I am curious about what others think though

allendav: Moved pass mgmt back into orchestrator in https://github.com/woocommerce/woocommerce-ios/pull/5294/commits/c554deff14db2d20f70ddf980c74c9ad8262c61d

Next commit will address cancellation

allendav:
@ctarda thanks for your initial review - would you mind re-viewing this? I've updated the description above with the new approach that no longer involves changes in the CardPresentPaymentStore

ctarda (approved):
Nice! Tests well, :shipit:

Issue #4428, opened 2021-06-15
[Mobile Payments] Connect to Reader hangs if live site is in test mode


allendav:
We shouldn't even allow the merchant to attempt to connect to a reader for live sites in test mode. Stripe doesn't support it.

I thought I had written this up before, but I can't find an issue for it. Feel free to close this issue if that's not the case.

cc @ctarda

ctarda:
How can we prevent the merchant from connecting to a reader if the site is in test mode?

allendav:
> How can we prevent the merchant from connecting to a reader if the site is in test mode?

We should be able to tell a site is live from the `is_live` boolean in the account response.

I wish we had `test_mode` in the account response, but it doesn't look like it. So... we have a few options:
1) Add test_mode to the account response
2) Extend the PaymentGateway model to parse `test_mode` (downside: most but not all gateways have the test mode concept and some may use a different key)
3) Fetch the connection token and then examine `test_mode` in its response

I am inclined to do the second option since it should be valuable down the road to have this data to work with for payment gateways besides than just WCPay

Edit: But for velocities sake we could do #3 faster and then do #2 later

Related: https://github.com/Automattic/woocommerce-payments/blob/80327e2c43aa838ca692b1c074d0e34348f198b2/includes/admin/class-wc-rest-payments-connection-tokens-controller.php#L45-L49

ctarda:
The previous ticket referred to might be #4515

allendav:
Looks like the backend may have the groundwork in place now. The wc/v3/payments/accounts endpoint returns is_live and test_mode with the expected values for my live and my test site as I toggle them in and out of test mode. Props @kalessil https://github.com/Automattic/woocommerce-payments/blame/6eaca4f05faa07acd0392a32ac1b1af051f93ca3/includes/admin/class-wc-rest-payments-accounts-controller.php#L73

In since 2.9.0 released on Aug 25

PR #5295, opened 2021-10-27
[Mobile Payments] Bump WCPay minimum from 2.5 to 2.9 for test_mode for WCPayAccount

allendav:
Partially addresses #4428 

@jaclync @joshheald - only one review needed - both of y'all are welcome of course cc @ctarda 

This is the first in a series of PRs to notify merchants that in-person payments are not supported on live accounts that are set to test mode.

Changes in this PR:

- This PR decodes the is_live and test_mode flags in the `wc/v3/payments/accounts` response ([test_mode](https://github.com/Automattic/woocommerce-payments/blame/6eaca4f05faa07acd0392a32ac1b1af051f93ca3/includes/admin/class-wc-rest-payments-accounts-controller.php#L73) was added in WCPay 2.9 and is_live is provided by our WPCOM backend - see compose_basic_account_info in the server repo)
- Updates unit tests and sample jsons to wit
- Per https://docs.swift.org/swift-book/LanguageGuide/Enumerations.html "When strings are used for raw values, the implicit value for each case is the text of that case’s name." - so I removed the unneeded strings for `country` and `supported` in the coding keys for `WCPayAccount`

Changes coming in later PRs:

The next PR will add these Bools to PaymentGatewayAccount. This will also require adding them as attributes to that Core Data entity.

A third PR will finish implementing isWCPayInTestModeWithLiveStripeAccount in CardPresentPaymentsOnboardingUseCase and a SwiftUI view to prompt the user that “In-Person Payments is Not Available When Test Mode is Enabled on Your Site” or somesuch, similar to the other “InPersonPayments*” views in Onboarding Errors

To test:
- Nothing user facing is changing with this PR. To test, ensure unit tests pass

Update release notes:


ctarda (approved):
This unblocks a lot of work, so I took the liberty of grabbing it and reviewing.

Code looks good. One question though. Will we have to bump the minimum version again for the support to the SDK 2.0? Even if we had to, I think it's okay to bump to 2.9 now and then bump again with the next release cycle of the mobile app, but I am curious of what you think.

Anyway, :shipit:

ctarda:
@allendav After merging this one, please make a note to update the documentation that references the minimum supported version

allendav:
@ctarda wrote

> Will we have to bump the minimum version again for the support to the SDK 2.0? Even if we had to, I think it's okay to bump to 2.9 now and then bump again with the next release cycle of the mobile app, but I am curious of what you think.

Yep! It's already enqueued: https://github.com/woocommerce/woocommerce-ios/blob/feature/stripe-sdk-2/WooCommerce/WooCommerceTests/ViewRelated/CardPresentPayments/CardPresentPaymentsOnboardingUseCaseTests.swift#L374-L375

allendav:
@ctarda wrote:

> After merging this one, please make a note to update the documentation that references the minimum supported version

https://github.com/woocommerce/woocommerce-ios/issues/5306
https://github.com/woocommerce/woocommerce-ios/issues/5307

PR #5298, opened 2021-10-28
Fix order request failing from spaces in `_fields` parameter value for certain sites

### LINKED ISSUE #5296: Products not loading in order details screen
**Describe the bug**

In release 7.7, some users mentioned the products section in the order details screen isn’t loading after updating the app. For certain stores, single order requests fail consistently with 500 errors.

**To Reproduce**

Please refer to p4a5px-2MI-p2

**Expected behavior**

The single order requests should not fail consistently.

**Mobile Environment**
Please include:
- Device: iPhone XR
- iOS version: iOS 14.5

jaclync:
Fixes #5296 

- [x] ⚠️ Update release notes when the target release is decided ⚠️ 

## Why

In release 7.7, we changed the `_fields` param values for order requests for performance improvements https://github.com/woocommerce/woocommerce-ios/pull/4997. The Swift multiline string literals have unexpected spaces for the single order `_fields` parameter value, which causes an exception in server-side for certain sites. It's possible that some servers cannot parse the `_fields` value when there are spaces.

### Why there are empty spaces

From the [official documentation](https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html) on multiline string literals, extra spaces are counted based on the **closing line `"""`**.

image
(Image from the [official doc](https://docs.swift.org/swift-book/LanguageGuide/StringsAndCharacters.html))

In release 7.7, the single order fields value lines have 4 extra spaces before the beginning of the closing line (the closing `"""`):

https://github.com/woocommerce/woocommerce-ios/blob/bcc01544ad85b69f14271345b145db46b3f091a1/Networking/Networking/Remote/OrdersRemote.swift#L253-L257

Compared to the order list fields value, the 3 lines with the fields are all aligned with the closing line so there are no spaces in the string:

https://github.com/woocommerce/woocommerce-ios/blob/bcc01544ad85b69f14271345b145db46b3f091a1/Networking/Networking/Remote/OrdersRemote.swift#L248-L252

cc @joshbetz 

## Changes

This PR fixes the empty spaces by using an array of field values in order to prevent such subtle incidents (I only learned about how spaces work in multiline string literals today). I also added a test case to make sure there are no spaces in the order fields param.

## Testing

Please refer to the "Steps to reproduce" section in p4a5px-2MI-p2#comment-12538

## Example screenshots

Update release notes:


pmusolino (approved):
Thanks for fixing it Jaclyn and for adding a unit test!

Code ✅ 
Testing ✅ 

> Update release notes when the target release is decided

Not sure about the impact on current users, but I believe that we can merge it in 7.9.

joshbetz:
> From the official documentation on multiline string literals, extra spaces are counted based on the closing line """.

Woah, nice. I did notice the closing `"""` looked wrong, but couldn't figure out why that would cause extra spaces.

jaclync:
Thanks for the quick review Paolo! After confirming with the team p1635429755081100/1635215591.051400-slack-C6H8C3G23, we are targeting release 7.9 and I updated the release notes in https://github.com/woocommerce/woocommerce-ios/pull/5298/commits/2bbea3cb8ffeab5209879570a89f9679edb50162. Merging now!

PR #5301, opened 2021-10-28
WC Tracker: Add telemetry action

ealeksandrov:
## Description

This PR adds telemetry endpoint support to report mobile app data (platform and version) to the Store.

## How

For telemetry endpoint usual Yosemite and Networking layers added. To reduce load on server:
- Action is dispatched only when wc-telemetry namespace available (we can check it in `RequirementsChecker`), WC 5.9.0 and up. This also prevents unwanted 404 errors in logs.
- Action is dispatched only once in 24h. Timeout is saved independently for every user’s store.

Server doesn't return any data in response in case of success. Jetpack wraps it in:
```
{
  "data": null
}
```
To ignore any kind of mapping here special `IgnoringResponseMapper` added.

Internal ref: p91TBi-6n5-p2

## Testing

iOS App:
1. Run it on WC <5.9.0. There shouldn't be any telemetry requests, including 404.
2. Install 5.9.0-rc.1 ([WooCommerce Beta Tester](https://wordpress.org/plugins/woocommerce-beta-tester/) plugin can help)
3. Switch on store with WC 5.9.0 and make sure only 1 telemetry request appeared in logs - for multiple app switches (background - foreground).
4. [Optional] Switch on different store with WC 5.9.0 and see same 1 telemetry request for that store.

[Optional] Verify on site:
1. Connect with ssh to the site after sending telemetry. 
2. Via WP-CLI run `wp eval 'var_dump(WC_Tracker::get_tracking_data());'`.
3. There should be a section like this:
```
["wc_mobile_usage"]=>
  array(1) {
    ["ios"]=>
    array(3) {
      ["platform"]=>
      string(3) "ios"
      ["version"]=>
      string(5) "7.8"
      ["last_used"]=>
      string(25) "2021-10-28T13:05:24+00:00"
    }
  }
```

---
Update release notes:


WooCommerce/Classes/Tools/RequirementsChecker.swift
```diff
@@ -92,4 +96,28 @@ private extension RequirementsChecker {
...
+    /// Dispatches a `TelemetryAction.sendTelemetry` action
+    ///
+    static func sendTelemetryIfNeeded(siteID: Int64) {
+        let minimalIntervalBetweenReports: TimeInterval = 60*60*24
+        if let telemetryLastReportedTime = UserDefaults.standard[.telemetryLastReportedTime] as? Date,
+           Date().timeIntervalSince(telemetryLastReportedTime) < minimalIntervalBetweenReports,
+           siteID == UserDefaults.standard[.telemetryLastReportedStoreID] {
+            // send telemetry for same store only once in 24h
+            return
+        }
```
Ecarrion: Wouldn't it be possible to fire a Yosemite action to perform this logic? Then we could use general app settings for storing the last reported time, and will be more similar to the rest of the app architecture. Also it will be easier to unit test I think!

Ecarrion (commented):
The code looks great! I just left a question/suggestion to move some logic out of `RequirementsChecker`!

One small thing that I found during testing is that launching the store picker and selecting a site, triggers the telemetry endpoint, it doesn't wait until the store picker is dismissed, maybe that is expected? Or an edge case?

Also, out of curiosity, what is this telemetry data going to be used for? 😅

ealeksandrov:
> Wouldn't it be possible to fire a Yosemite action to perform this logic? Then we could use general app settings for storing the last reported time, and will be more similar to the rest of the app architecture. Also it will be easier to unit test I think!

Thanks for idea! I've looked at GeneralAppSettings, but it's universal between all stores and I need store-specific info. I decided to add `GeneralStoreSettings` entity with very similar API, but for independent settings for stores. This allows us to have different timeouts for different stores and keep availability status saved away from RequirementsChecker.

One small drawback for testing - when you launch app updating from previous version it won't send telemetry, only fetch availability. If you relaunch app or switch store it'll be triggered. Fresh app install will work fine.

Let me know if you see more possible improvements for all this actions management 🙂

> One small thing that I found during testing is that launching the store picker and selecting a site, triggers the telemetry endpoint, it doesn't wait until the store picker is dismissed, maybe that is expected? Or an edge case?

This refactoring fixed it 🙂

> Also, out of curiosity, what is this telemetry data going to be used for? 😅

It's another perspective on stores analytics instead of mobile users behavior (where we have Tracks). This will help us with identifying sites that are using the mobile app and find interesting connections.

Ecarrion (approved):
🚀

PR #5303, opened 2021-10-28
[Mobile Payments] Cancel payment capture if order amount is below minimum (USD only)

### LINKED ISSUE #4911: Handle error code amount_too_small from createPaymentIntent
Context - pdfdoF-6N-p2#comment-233

ctarda:
Closes #4911 

> As Stripe’s processing fee combines a small fixed amount and a percentage, we enforce a minimum amount when creating a charge. This ensures you don’t lose money on a charge. The minimum amount you can charge depends on which bank account settlement currency the payment would be paid out to. [Source](https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts)

That minimum amount turns out to be 0.50 USD.

## Changes
* CardReaderService: Add a case in the error enumeration that the service exposes to model "api errors". 
* PaymentCaptureOrchestrator: check the order total amount **before** attempting to capture a payment, and throw a specific error.

This covers both ends of the problem:
* it prevents initiating a payment capture in the UI when we know that it is going to fail
* if an attempt to capture the payment got initiated somehow, we would still present a meaningful error.

However, there is a bit of duplication (the error is kind of defined in two different places). To me, this is one of those cases where a bit of duplication is better that introducing the wrong abstraction.

Also, this solution could be a bit better: we still attempt to connect to the card reader even though we know that the payment capture is going to fail. More about that in a comment in the actual diff.

## How to test
* Create three orders: Order A is a total 0.15 USD. Order B is a total 0.50 USD. Order C is over 0.50 USD
* Attempt to capture payment all those orders. Order B and C should work fine. For Order A, the app should present an error message before attempting to capture the payment.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/CardReaderServiceError.swift
+++ b/Hardware/Hardware/CardReader/CardReaderServiceError.swift
@@ -186,6 +186,9 @@ public enum UnderlyingError: Error {
     /// The SDK will attempt to auto-disconnect for you and you should instruct your user to reconnect it.
     case readerSessionExpired
 
+    /// The underlying request returned an API error.
+    case stripeAPIError
+
     /// Catch-all error case. Indicates there is something wrong with the
     /// internal state of the CardReaderService.
     case internalServiceError
@@ -305,6 +308,9 @@ updating the application or using a different reader
         case .readerSessionExpired:
             return NSLocalizedString("The card reader session has expired - please disconnect and reconnect the card reader and then try again",
                                      comment: "Error message when the card reader session has timed out.")
+        case .stripeAPIError:
+            return NSLocalizedString("Unable to process payment. Order Total amount is below the minimum amount you can charge.",
+                                     comment: "Error message when the order amount is below the minimum amount allowed.")
         case .internalServiceError:
             return NSLocalizedString("Sorry, this payment couldn’t be processed",
                                      comment: "Error message when the card reader service experiences an unexpected internal service error.")
--- a/Hardware/Hardware/CardReader/CardReaderServiceError.swift
+++ b/Hardware/Hardware/CardReader/CardReaderServiceError.swift
@@ -187,7 +187,7 @@ public enum UnderlyingError: Error {
     case readerSessionExpired
 
     /// The underlying request returned an API error.
-    case APIError
+    case processorAPIError
 
     /// Catch-all error case. Indicates there is something wrong with the
     /// internal state of the CardReaderService.
@@ -308,7 +308,7 @@ updating the application or using a different reader
         case .readerSessionExpired:
             return NSLocalizedString("The card reader session has expired - please disconnect and reconnect the card reader and then try again",
                                      comment: "Error message when the card reader session has timed out.")
-        case .APIError:
+        case .processorAPIError:
             return NSLocalizedString("Unable to process payment. Order Total amount is below the minimum amount you can charge.",
                                      comment: "Error message when the order amount is below the minimum amount allowed.")
         case .internalServiceError:
--- a/Hardware/Hardware/CardReader/CardReaderServiceError.swift
+++ b/Hardware/Hardware/CardReader/CardReaderServiceError.swift
@@ -309,8 +309,8 @@ updating the application or using a different reader
             return NSLocalizedString("The card reader session has expired - please disconnect and reconnect the card reader and then try again",
                                      comment: "Error message when the card reader session has timed out.")
         case .processorAPIError:
-            return NSLocalizedString("Unable to process payment. Order Total amount is below the minimum amount you can charge.",
-                                     comment: "Error message when the order amount is below the minimum amount allowed.")
+            return NSLocalizedString("The payment can not be processed by the payment processor.",
+                                     comment: "Error message when the payment can not be processed (i.e. order amount is below the minimum amount allowed.)")
         case .internalServiceError:
             return NSLocalizedString("Sorry, this payment couldn’t be processed",
                                      comment: "Error message when the card reader service experiences an unexpected internal service error.")
--- a/Hardware/Hardware/CardReader/CardReaderServiceError.swift
+++ b/Hardware/Hardware/CardReader/CardReaderServiceError.swift
@@ -187,7 +187,7 @@ public enum UnderlyingError: Error {
     case readerSessionExpired
 
     /// The underlying request returned an API error.
-    case stripeAPIError
+    case APIError
 
     /// Catch-all error case. Indicates there is something wrong with the
     /// internal state of the CardReaderService.
@@ -308,7 +308,7 @@ updating the application or using a different reader
         case .readerSessionExpired:
             return NSLocalizedString("The card reader session has expired - please disconnect and reconnect the card reader and then try again",
                                      comment: "Error message when the card reader session has timed out.")
-        case .stripeAPIError:
+        case .APIError:
             return NSLocalizedString("Unable to process payment. Order Total amount is below the minimum amount you can charge.",
                                      comment: "Error message when the order amount is below the minimum amount allowed.")
         case .internalServiceError:
--- a/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
+++ b/WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
@@ -20,9 +20,11 @@ final class PaymentCaptureOrchestrator {
                         onClearMessage: @escaping () -> Void,
                         onProcessingMessage: @escaping () -> Void,
                         onCompletion: @escaping (Result<CardPresentReceiptParameters, Error>) -> Void) {
-        guard let orderTotal = currencyFormatter.convertToDecimal(from: order.total), orderTotal.compare(Constants.minimumAmount) == .orderedDescending else {
-            DDLogError("💳 Error: failed to capture payment for order. Minim")
-            onCompletion(.failure(CardReaderServiceError.intentCreation(underlyingError: .processorAPIError)))
+        /// Bail out if the order amount is below the minimum allowed:
+        /// https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts
+        guard isTotalAmountValid(order: order) else {
+            DDLogError("💳 Error: failed to capture payment for order. Order amount is below minimum")
+            onCompletion(.failure(NotValidAmountError.belowMinimumAmount))
             return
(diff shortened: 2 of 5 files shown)


Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -186,6 +186,9 @@ public enum UnderlyingError: Error {
     /// The SDK will attempt to auto-disconnect for you and you should instruct your user to reconnect it.
     case readerSessionExpired
 
+    /// The underlying request returned an API error.
+    case processorAPIError
```
ctarda, line 190: A bit generic, non compromising error. References to Stripe removed.

Hardware/Hardware/CardReader/CardReaderServiceError.swift
```diff
@@ -305,6 +308,9 @@ updating the application or using a different reader
         case .readerSessionExpired:
             return NSLocalizedString("The card reader session has expired - please disconnect and reconnect the card reader and then try again",
                                      comment: "Error message when the card reader session has timed out.")
+        case .processorAPIError:
+            return NSLocalizedString("Unable to process payment. Order Total amount is below the minimum amount you can charge.",
```
ctarda: Completely made up error message, I assume it could be much better worded. At this point, all we know is that capturing the payment failed because of Stripe rejects it

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -20,6 +20,13 @@ final class PaymentCaptureOrchestrator {
                         onClearMessage: @escaping () -> Void,
                         onProcessingMessage: @escaping () -> Void,
                         onCompletion: @escaping (Result<CardPresentReceiptParameters, Error>) -> Void) {
+        /// Bail out if the order amount is below the minimum allowed:
+        /// https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts
+        guard isTotalAmountValid(order: order) else {
```
ctarda, line 25: Here we check that the order amount will not fail.

This happens **after** we have connected to the reader. The difficult solution here is that the connection to the reader is triggered in the view controller, not in the payment orchestrator.

I think we should consider pushing the knowledge of the reader down to the orchestrator, but that requires careful consideration and it seems to me like it would out of the scope of this PR

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -258,6 +265,20 @@ private extension PaymentCaptureOrchestrator {
...
+    enum Constants {
+        static let minimumAmount = NSDecimalNumber(string: "0.5")
+    }
+
+    func isTotalAmountValid(order: Order) -> Bool {
+        guard let orderTotal = currencyFormatter.convertToDecimal(from: order.total) else {
+            return false
+        }
+
+        return orderTotal.compare(Constants.minimumAmount) == .orderedDescending
```
joshheald:
```suggestion
        return orderTotal as Decimal >= Constants.minimumAmount as Decimal
```

🔧 I think this improves readability (I've never been able to quickly intuit the `orderedDescending` notation) and the Obj-C -> Swift bridging from `NSDecimalNumber` to `Decimal` takes care of everything. Using the [`decimalValue` property](https://developer.apple.com/documentation/foundation/nsnumber/1407409-decimalvalue) also works, so pick your favourite, I guess. I refreshed my memory with some research and couldn't find any difference between them.

I would also consider making the constant a `Decimal` and doing the cast for `orderTotal` in the `guard` here, to neaten the comparison further.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -268,3 +289,31 @@ private extension PaymentCaptureOrchestrator {
...
+                "Unable to process payment. Order Total amount is not valid.",
+                comment: "Error message when the order amount is not valid."
+            )
+
+            static let belowMinimumAmount = NSLocalizedString(
+                "Unable to process payment. Order Total amount is below the minimum amount you can charge.",
+                comment: "Error message when the order amount is below the minimum amount allowed."
+            )
+        }
+    }
```
joshheald, line 328: ⭐️ Nice error enum

WooCommerce/Classes/ViewRelated/Orders/Order Details/OrderDetailsViewController.swift
```diff
@@ -1087,6 +1087,7 @@ private extension OrderDetailsViewController {
         static let headerContainerInsets = UIEdgeInsets(top: 0, left: 0, bottom: 8, right: 0)
         static let rowHeight = CGFloat(38)
         static let sectionHeight = CGFloat(44)
+        static let minimumAmount = NSDecimalNumber(string: "0.5")
```
joshheald: 🔧 This isn't used, as far as I can see... (which is good because you can simply delete it and I don't really like it being in the view controller!)

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -258,6 +265,20 @@ private extension PaymentCaptureOrchestrator {
     }
 }
 
+private extension PaymentCaptureOrchestrator {
+    enum Constants {
+        static let minimumAmount = NSDecimalNumber(string: "0.5")
```
joshheald, line 272: 💭 🌱  My main sad-face about this is a US-specific value in code. This is fine in this PR with the context of getting stuff done for our existing customers, and I'm sure there are lots of other values like it... but I do worry about this kind of debt becoming difficult to repay when we want to ship in other countries. 

In particular, I think it'll be hard to find this stuff when we do want to fix it – there's nothing to particularly call out that this is a USD specific value, apart perhaps for a search for NSDecimalNumber, which is a bit of a clue.

Should we log this somewhere, or call it out with a comment to make it easier to identify in future?

Hardware/Hardware/CardReader/CardReaderServiceError.swift
joshheald: 🔧 "Order Total" both being capitalised looks strange here, I think the it should be "Order total", or just "Total amount" as it is in your screenshot, rather than this...
Simulator Screen Shot - iPhone 12 mini - 2021-10-28 at 18 13 30

Also, it seems strange to offer the "Try Collecting Again" button for this error, when we know it will definitely fail (based on `isTotalAmountValid`). I think we can live with it, since when the merchant does it, they're taken to this non-retriable error, so they don't get prompted to just keep trying...

Simulator Screen Shot - iPhone 12 mini - 2021-10-28 at 19 01 48

joshheald: ❓This seems like quite a specific interpretation of quite a generic error message... I couldn't find much about the error online and what else might cause it, is this worth asking Stripe about? 
It would be nice if there was anything we could use to have more confidence that minimum order amount was the reason for the error here.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
joshheald: 💭 What about passing the minimum amount to the `isTotalAmountValid` function here, and also include it (currency formatted) as a format param in the error message, which can be filled with an associated value on the enum? I'm thinking in terms of some idealised future when we get the minimum amount from the API, appropriate for the currency, or at least from some local store in the app with localization.

    public var errorDescription: String? {
        switch self {
        case .belowMinimumAmount(let minimum):
            return String(format: Localizations.belowMinimumAmount, minimum)
    }

    enum Localizations {
        static let belowMinimumAmount = NSLocalizedString(
            "Unable to process payment. Order total amount is below the minimum amount you can charge, which is %@.",
            comment: "Error message when the order amount is below the minimum amount allowed."
        )
    }

This would save our merchants some guesswork...

joshheald (approved):
A couple of small suggestions for change, and quite a few thoughts. I've gone a bit long on this review I'm afraid, but take what you like and leave/ticket what you don't, not much of this is a _need_ for now.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
ctarda: The minimum amounts are, according to Stripe: https://stripe.com/docs/currencies#minimum-and-maximum-charge-amounts

It does make sense to pass the actual amount as parameter though for whenever we expand to other currencies. I'll update!

Hardware/Hardware/CardReader/CardReaderServiceError.swift
ctarda: Fair point. I think we can go more generic here, specially considering we _should_ we preventing ourselves from hitting this error with the guarding that we are doing in the UI. I'll make it more generic.

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
ctarda: Yes, this is something we will need to tackle if and when we start expanding internationally. For now, let's go with a comment, and we will revisit later

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
ctarda: In this case, the APIs of currencyConverter expect NSDecimalNumber, so I'd go with the downcast here, even though, as you mentioned, it is not the neatest way to go.

ctarda:
Thanks for the feedback @joshheald I pushed an update to incorporate your suggestions (with a couple of minor adjustments). I'll let CI merge after the ✅

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
ctarda: Also, logged #5308 for enqueuing what we'd need to do to expand to Canada.

PR #5304, opened 2021-10-28
Merge 7.8 beta 3 into develop

jkmassel:
Merges beta 3 changes back into `develop`


allendav (approved):
Tests fine. I ensured that the unit tests in CardPresentPaymentsOnboardingUseCaseTests and AccountStoreTests continued to pass (they depend on insertSampleSystemPlugin and insertSampleSite which is where the merge conflict was) and that, as shop manager, I was still able to access the Manage Card Reader flow for my store.

PR #5310, opened 2021-10-29
Filter Products by Category: query Products locally

toupper:

## Description
In the context of https://github.com/woocommerce/woocommerce-ios/issues/5159, the scope of this PR is to apply the product category filter when querying the `CoreData` entities, in a similar way as it is done for the rest of product filters. With this addition the filter also works now in offline mode, by filtering the cached product entities.

## Decisions
As explained in https://github.com/woocommerce/woocommerce-ios/pull/5290, when retrieving filtered products by category remotely we obtained not only the products linked to the selected category, but also those linked to the descendant categories. This logic cannot be implemented however with a simple `NSPredicate`, as we do not have that information in the `ProductCategory` entity (only the `parentID`) nor it is delivered in the request JSON. Because of that, with the addition of the `ProductCategory` `NSPredicate` filter to the `ProductsViewController` `ResultsController` we retrieve only those products linked to the exact category set in the filter. I considered and discarded these alternatives as well:

- We could mimic the filter remote request logic by recursively filter the products comparing the filter `CategoryID` with `ParentID`, thus looping until a category has no children and gathering all the products along the way. This would add much complexity to our case, having probably to move away from the approach we use at the moment (`ResultsController` with a compound predicate).
- We could get all the products linked to a category whose parent is the selected category as well, thus showing products linked to a direct child category too. (e.g. selected category: Clothing, would show products with category pants, but not its children: jeans, shorts ...) I did not add it because it would add more confusion in my opinion: why we show the direct child category products, but not the rest of descendants? 
- Do not filter for category locally and rely on the remote filter results. This however would not work for offline mode.

## Changes
- Add NSPredicate for the ProductCategory filter in the `ProductsViewController` `ResultsController`
- Pass the ProductCategory to the ResultsController predicate related methods in ProductsViewController`, update and create
- Unit Tests. Make `ProductCategory` copiable, add the Product categories to the MockStorageManager so they can be stored and queried. Rename file tests according to new convention.

### Products are filtered according to product category (online)

- Open app
- Go to products
- Open filter 
- Select Product Category
- Select a product category
- Press "Show products" or
- Go back
- Press "Show products"
- Products list is filtered according to selected category. **Please note that now the filter works for the exact filter category, but not for its descendants. That is, if parent category is "Clothing" and is selected, and child category is "Shirts", "Shirts" products will not be shown.**

### Products are filtered according to product category (offline)

- Open app
- Go to products. Make them load, as the offline query will be applied to the cached products
- Go offline
- Open filter 
- Select Product Category
- Select a product category
- Press "Show products" or
- Go back
- Press "Show products"
- Products list is filtered according to selected category. Please note that the filter will be applied to the previously retrieved cached products, so the result can vary according to what was retrieved.

## Next Steps
- Synchronize stored `ProductCategory` with remote values to handle remote deletion or edition.
- Remove Feature Flag

Update release notes:


ealeksandrov:
Code looks good! ✅
Filter works as described for both remote and local actions ✅ (but offline mode is not really usable - too many edge cases)

I'm only concerned that wp-admin in product filter includes all subcategories:
Screenshot

But I don't see good solution to include subcategories in local search, as you described. Maybe rely only on remote search?
@pmusolino what do you think?

ealeksandrov (commented):
Code looks good! ✅
Filter works as described for both remote and local actions ✅ (but offline mode is not really usable - too many edge cases)

I'm only concerned that wp-admin in product filter includes all subcategories:
Screenshot

But I don't see good solution to include subcategories in local search, as you described. Maybe rely only on remote search?
@pmusolino what do you think?

pmusolino (commented):
As said by @ealeksandrov, I don't see a good solution to include subcategories in local search. 
Did you ask the Android team what was their solution on this side? Since they already implemented this functionality, it could be great to align the behavior with them. @hafizrahman worked on the Android functionality.

toupper:
@ealeksandrov @pmusolino  It seems like on Android they faced the same challenge we have here. For now they released this feature just filtering by the exact product category (without showing products belonging to children subcategories), but as @hafizrahman explains here https://github.com/woocommerce/woocommerce-android/issues/4578 they have an open issue https://github.com/woocommerce/woocommerce-android/issues/4714 to modify the filtering result to include products under selected category's sub-categories.

pmusolino:
What happens when there are a lot of elements in that category and we need to paginate it? I don't believe that the pagination will work as expected since the number of elements should be different from the one returned by APIs, or am I wrong?

toupper:
@pmusolino That's a good point. Since the results we show and those retrieved from the REST Api differ, the pagination behaviour would be inconsistent: we would show partial entries from a page result, being possible the case when we do not show any result at all (all products results belonging to subcategories).
It seems like querying the results locally does not add much value (as @ealeksandrov pointed out the offline logic is not really usable), rather the opposite. We can rely only on remote search as it is in develop at the moment thus closing this PR, and perhaps disabling/hiding the category filter in the future when going offline if necessary, as it won't work as expected.

pmusolino:
@toupper yes, I'd say to close this PR (unless you have another solution in mind), since it doesn't make sense to have a pagination that is not working properly. There is for sure a way to improve it (eg. storing the categories fetched, and doing a complex calc when we need to paginate with a different number of elements), but for the moment, this is not what we are looking for.

toupper:
@pmusolino thanks for your feedback, I close it then, and add a comment in the original issue to keep track of this behaviour 👍

PR #5311, opened 2021-10-29
Change Circle CI image to Xcode 13

pmusolino:
This PR changes the Circle CI image to use Xcode 13.

## Testing
Make sure Circle CI build is green.

## Notes
Before merging this PR in `develop`, we need to fix these [issues](https://github.com/woocommerce/woocommerce-ios/issues?q=is%3Aissue+is%3Aopen+xcode+13), in particular this one https://github.com/woocommerce/woocommerce-ios/issues/5018

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #5312, opened 2021-10-29
Fix: tab bar background color under iOS 15 and Xcode 13

### LINKED ISSUE #5018: iOS 15: tab bar black in light mode, and "My Store" menu no more visible during the first loading
I discovered this bug that is currently happening just on iOS 15.0 (I tested it in the simulator, on iPhone 13 Pro).

During the first loading of the dashboard, if the iPhone is in light mode, the tab bar will become black, and the "My Store" menu is invisible.

This happens just if you build the app with Xcode 13.

pmusolino:
Fixes #5018 

Compiling the project with Xcode 13 under iOS 15, we discovered a bug where the background of the tab bar has the wrong color. @jaclync proposed a solution to fix it, that works perfectly.
*Note*: This PR is targeting https://github.com/woocommerce/woocommerce-ios/pull/5311.

## Testing
1. Run the app under Xcode 13 (first in light mode, then in dark mode).
2. Navigate between tabs.
3. The background color of the tab should be correct.

## Screenshots
Simulator Screen Shot - iPhone 13 Pro - 2021-10-29 at 11 24 57 | Simulator Screen Shot - iPhone 13 Pro - 2021-10-29 at 12 02 10

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Extensions/UITabBar+Appearance.swift
+++ b/WooCommerce/Classes/Extensions/UITabBar+Appearance.swift
@@ -15,6 +15,13 @@ extension UITabBar {
         /// To fix it, we have to specifically set it in the `standardAppearance` object.
         ///
         appearance.standardAppearance = createWooTabBarAppearance()
+
+        /// This is needed because the tab bar background has the wrong color under iOS 15 (using Xcode 13).
+        /// More: issue-5018
+        ///
+        if #available(iOS 15.0, *) {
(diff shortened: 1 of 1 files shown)


jaclync (approved):
Thank you @pmusolino for fixing this and taking charge of Xcode 13 updates! :shipit: I'm especially looking forward to trying out async/await when Xcode 13.2 is released 😄 

Tested on iPhone 11 and iPad Pro simulators, Light and Dark mode

PR #5322, opened 2021-11-02
[Mobile Payments] [SDK 2] Switch to SDK 2

allendav:
This PR merges the long-lived feature/stripe-sdk-2 branch to develop. This will transition the app from SDK 1.x to SDK 2.x. This adds support for the M2 reader, changes the reader software updates to include mandatory un-cancellable updates, and other fixes.

To Test:

- Ensure you can
    - connect to a Chipper reader
    - connect to a M2 reader - see known issue #5326
    - choose from multiple discovered readers
    - connect to a reader after changing your store’s address in wp-admin (e.g. street address)
    - update a reader’s software - see known issues #5324 and #5325 
    - gracefully handle a reader with too low a battery to update - see known issue #5327
    - collect a tap payment
    - collect a chip (insert) payment
    - collect a swipe payment

Update release notes:


Hardware/Hardware/CardReader/CardReaderConfigProvider.swift
```diff
@@ -1,6 +1,16 @@
...
+
+public protocol ReaderTokenProvider {
     func fetchToken(completion: @escaping(String?, Error?) -> Void)
 }
+
+public protocol ReaderLocationProvider {
+    func fetchDefaultLocationID(completion: @escaping(String?, Error?) -> Void)
+}
+
+public protocol CardReaderConfigProvider: ReaderLocationProvider {
```
joshheald, line 13:
```suggestion
public protocol CardReaderConfigProvider: ReaderLocationProvider, ReaderTokenProvider {
```

🔧 It looks like this protocol also includes `ReaderTokenProvider`, but there's no explicit conformance to that one, I think we should add it.

Hardware/Hardware/CardReader/CardReaderConfigProvider.swift
```diff
@@ -1,6 +1,16 @@
...
+    func fetchToken(completion: @escaping(String?, Error?) -> Void)
+    func fetchDefaultLocationID(completion: @escaping(String?, Error?) -> Void)
```
joshheald, line 15: ❓♻️   What do you think about using `Result<String, Error>` in our adapter? I can see the completion blocks are modelled on `ConnectionTokenCompletionBlock` from `StripeTerminal`, but we could potentially adapt those as well? 

Perhaps something to ticket for later...

Hardware/Hardware/CardReader/CardReaderService.swift
```diff
@@ -52,13 +52,8 @@ public protocol CardReaderService {
...
-
-    /// Triggers a software update. This method requires that checkForUpdates
-    /// has been completed successfully
+    /// Triggers a software update.
     ///
-    /// The returned publisher will periodically publish the fraction of progress during the software update
-    /// and it will complete when it's finished, unless there is any error.
-    func installUpdate() -> AnyPublisher<Float, Error>
+    /// To check the progress of the update, observe the softwareUpdateEvents publisher.
+    func installUpdate() -> Void
```
joshheald, line 58: This is a nicer API 👍

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
```diff
@@ -207,7 +205,7 @@ private extension PaymentCaptureOrchestrator {
 
             switch result {
             case .success:
-                self?.celebrate()
+                self?.celebrate() // plays a sound, haptic
```
joshheald, line 208: 💭 Adding `let self = self` to the guard above would _probably_ be worthwhile here. I'm not sure we want to be doing the other things in this block but not saving the receipt, if the `PaymentCaptureOrchestrator` has gone away anyway... 

I traced that back as best I could see and while the `onCompletion` didn't seem like it would currently do anything too surprising in that case, this does seem like somewhere that strange bugs could creep in.

What do you think? Playing devil's advocate I can see the merits of trying to handle the success even if we can't save the receipt too... but generally I'd fall on the side of failing fast(er.)

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -66,20 +58,19 @@ final class CardReaderSettingsConnectedViewController: UIViewController, CardRea
...
         configureSections()
         configureTable()
         configureUpdateView()
     }
 
+    func isReaderUpdateAvailable() -> Bool {
+        guard let viewModel = viewModel else {
+            return false
+        }
+        return viewModel.readerUpdateAvailable == true
```
joshheald, line 71:
```suggestion
        return viewModel?.readerUpdateAvailable == true
```

This could also be inlined at the call sites of this function, if desired...

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -78,64 +114,32 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
...
     }
 
-    /// Allows the view controller to kick off a card reader update
-    ///
-    func startCardReaderUpdate() {
-        self.readerUpdateCompletedSuccessfully = false
+    func cancelCardReaderUpdate() {
+        softwareUpdateCancelable?.cancel(completion: { [weak self] result in
+            if case .failure(let error) = result {
+                print("=== error canceling software update: \(error)")
```
joshheald, line 127:
```suggestion
                DDLogError("=== error canceling software update: \(error)")
```

joshheald (approved):
Some minor suggestions added, all optional.

Testing progress
- [x] connect to a Chipper reader
- [ ] ~connect to a M2 reader~
- [ ] ~choose from multiple discovered readers~
- [x] connect to a reader after changing your store’s address in wp-admin (e.g. street address)
- [ ] update a reader’s software - done on sim, todo on physical reader after 👇 
- [ ] gracefully handle a reader with too low a battery to update - 🕐 trying to drain the battery to test this! 🔋 
- [x] collect a tap payment
- [x] collect a chip (insert) payment
- [ ] ~collect a swipe payment~

Strikethrough denotes tests I can't do for lack of equipment. @ctarda are there any of those you can do?

joshheald:
@allendav I found #5332 during testing. It's a bit of an edge case, so we could consider merging even with this issue. It seems like a completion block is being called when it shouldn't be, and also might be solved with a `guard self = self` in the right block, as I suspect the buttons don't work because of referencing something else on `self`. That's all guesswork at the moment I'm afraid though!

allendav:
Thank you for the review @joshheald ! In the interest of avoiding merge conflicts, would you be ok with me addressing the feedback in follow on issues? I'd like to merge this as quickly as possible.

joshheald:
> Thank you for the review @joshheald ! In the interest of avoiding merge conflicts, would you be ok with me addressing the feedback in follow on issues? I'd like to merge this as quickly as possible.

@allendav Yeah no problem at all... I should have called more attention to it but I meant that as one of your options when I wrote "Some minor suggestions added, *all optional*."

Hardware/Hardware/CardReader/CardReaderConfigProvider.swift
allendav: Opened https://github.com/woocommerce/woocommerce-ios/issues/5424

Hardware/Hardware/CardReader/CardReaderConfigProvider.swift
allendav: Opened https://github.com/woocommerce/woocommerce-ios/issues/5425

WooCommerce/Classes/ViewModels/CardPresentPayments/PaymentCaptureOrchestrator.swift
allendav: Opened https://github.com/woocommerce/woocommerce-ios/issues/5426

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
allendav: Opened https://github.com/woocommerce/woocommerce-ios/issues/5427

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
allendav: Opened https://github.com/woocommerce/woocommerce-ios/issues/5428

PR #5334, opened 2021-11-03
Order Editing: Add Address Form fields dynamic type support

ealeksandrov:
Part of https://github.com/woocommerce/woocommerce-ios/issues/4787.

## Description

This PR adds Dynamic Type support for address form. It splits title-value rows into 2 lines when accessibility text size is selected.

### How

AdaptiveStack wrapper view added. It switches between HStack and VStack depending on sizeCategory.

### Considerations

`TitleAndTextFieldRow` is also used in shipping labels screens. Quick test shows that they are broken before and after this change (screen doesn't fit in width in portrait phone mode) and require additional work. Example screenshots are below. Created issue https://github.com/woocommerce/woocommerce-ios/issues/5336.

## Test

1. Enable accessibility text size in iOS settings
2. Open existing order details
3. Tap edit icon on shipping address

## Screenshots

/ | before | after
--|--|--
regular size|Simulator Screen Shot|Simulator Screen Shot
accessibility size|Simulator Screen Shot|Simulator Screen Shot
shipping labels 1|Simulator Screen Shot|Simulator Screen Shot
shipping labels 2|Simulator Screen Shot|Simulator Screen Shot

---
Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AdaptiveStack.swift
+++ b/WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AdaptiveStack.swift
@@ -1,0 +1,30 @@
+import SwiftUI
+
+struct AdaptiveStack<Content: View>: View {
+    @Environment(\.sizeCategory) var sizeCategory
+
+    let horizontalAlignment: HorizontalAlignment
+    let verticalAlignment: VerticalAlignment
+    let spacing: CGFloat?
+    let content: () -> Content
+
+    init(horizontalAlignment: HorizontalAlignment = .center,
+         verticalAlignment: VerticalAlignment = .center,
+         spacing: CGFloat? = nil,
+         @ViewBuilder content: @escaping () -> Content) {
+        self.horizontalAlignment = horizontalAlignment
+        self.verticalAlignment = verticalAlignment
+        self.spacing = spacing
+        self.content = content
+    }
+
+    var body: some View {
+        Group {
+            if sizeCategory.isAccessibilityCategory {
+                VStack(alignment: horizontalAlignment, spacing: spacing, content: content)
+            } else {
+                HStack(alignment: verticalAlignment, spacing: spacing, content: content)
+            }
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AdaptiveStack.swift
```diff
@@ -0,0 +1,31 @@
...
+         @ViewBuilder content: @escaping () -> Content) {
+        self.horizontalAlignment = horizontalAlignment
+        self.verticalAlignment = verticalAlignment
+        self.spacing = spacing
+        self.content = content
+    }
+
+    var body: some View {
+        Group {
+            if sizeCategory.isAccessibilityCategory, horizontalSizeClass == .compact {
```
rachelmcr, line 24: I noticed in the [documentation for `isAccessibilityCategory`](https://developer.apple.com/documentation/swiftui/contentsizecategory/isaccessibilitycategory) that it's being deprecated. Do you think it's worth switching to something like this instead:

```suggestion
            if sizeCategory >= .accessibilityMedium, horizontalSizeClass == .compact {
```

rachelmcr (approved):
Looks good! I tested with dynamic type on both iPhone and iPad and this change works well on both. I left a small non-blocking suggestion below. Thanks for also testing the change in the Shipping Labels flow. 👍 

Do you think we should use `AdaptiveStack` for `TitleAndValueRow` as well? I noticed in your screenshot and my testing that the country row with dynamic type doesn't fit well on a single line (although it's not as problematic as `TitleAndTextFieldRow` was):

ealeksandrov:
Thanks for review @rachelmcr!

> Do you think we should use AdaptiveStack for TitleAndValueRow as well?

Yes, I overlooked it since it wasn't too broken compared to textfields 😄 But adding AdaptiveStack there is straightforward, so I've updated it:

before|after
--|--
preview1|preview2

And the final version of address form:

WooCommerce/Classes/ViewRelated/ReusableViews/SwiftUI Components/AdaptiveStack.swift
ealeksandrov: Good find! it's actually a deprecation for whole [ContentSizeCategory](https://developer.apple.com/documentation/swiftui/contentsizecategory) enum. It's replaced with [DynamicTypeSize](https://developer.apple.com/documentation/swiftui/dynamictypesize) on iOS 15.

I don't see obvious solution to conditionally add `@Environment` property only for iOS 15. Let's leave it for now and trust Xcode to remind us about migration in the future 🙂

PR #5337, opened 2021-11-03
Filter Products by Category: synchronize remotely the Filter Product Category

toupper:

## Description
In the context of https://github.com/woocommerce/woocommerce-ios/issues/5159, the user can select a product category to filter the products, what causes the category being stored locally so the filter is still enabled when the user restarts the application. With this PR I want to handle the case when that product category is deleted externally (e.g from wp-admin), an event that would trigger an inconsistent state when going back to the products list or opening the Filter view.
To address that, we compare the locally stored product category filter with is remote counterpart, removing the filter if the category was deleted.

## Decisions
In the the implementation suggested with this PR, we do not wait for the ProductCategory to be synchronized to apply the filters, but we show them as soon as they retrieved locally. After that we synchronize the `ProductCategory` asynchronously, and then update the view (filters + products) **only** if the `ProductCategory` changed or was removed. This way we can speed up the products load with the filters process.

## Notes
This logic is added to the `ProductsViewController` as I did not find any view model in charge of implementing the presentation logic of the view, thus assuming that MVC is the architecture here.

## Changes
- Add the relevant methods in `ProductCategoryStore` and `ProductCategoriesRemote` to retrieve the remote information of a single product category given its `categoryID`
- Create the  cases for `DotComError` and `ProductCategoryActionError` to encapsulate the event of a failure because the resource was not there
- Once the local product settings are retrieved when applying the filters in the `ProductsViewController`, the locally `ProductCategory` is synchronized against its remote entity. If the the category was deleted, we update the filters product information, and remove the filter locally.
- Unit tests are added. New naming convention and asynchronous testing approach are applied to all the tests in the modified files.

## Tests
- Open app
- Go to products
- Open filter 
- Select Product Category
- Select a Product Category
-  Press "Show products" so the Product Category filter is stored
- You can see that the filter number on top of the products list is updated to include the Product Category filter
- Close the app
- Go to your website wp-admin
- Delete that category
- Open the app again and go to products
- After the filter is synchronized, the Filters number does not include the Product Category filter
- Open filter. There is no Product Category selected, _Any_ is selected

Update release notes:


pmusolino:
Do this PR means that we can close this one https://github.com/woocommerce/woocommerce-ios/pull/5310?

toupper:
@pmusolino no, these two PRs are independent. https://github.com/woocommerce/woocommerce-ios/pull/5310 is about querying the products locally with the `ProductCategory`, and we can close it if we decide to rely on the remote results instead of the CoreData query ones because the latter approach doesn't add any value (more than offline filtering). 
This one is about synchronizing the stored ProductCategory remotely to avoid the case of having a `ProductCategory` filter that was deleted; in my opinion it is necessary even if we merge https://github.com/woocommerce/woocommerce-ios/pull/5310

pmusolino (commented):
Thanks @toupper for opening this PR. I tested it, but I encountered two main issues:
1) There are some glitches if a filter is applied and then deleted remotely. So, basically we refresh the data two times. I'm not sure if it makes sense. In my opinion, it was better to mantain the filter locally, also if it doesn't exist and it will return no results.
           - We need to do an extra API call.
           - The user may not understand what is happening under the hood: the remote filter may have been deleted a long time ago, or by someone else on the team. It seems basically glitchy. 
2) While I was testing it, I applied two filters (stock status and the category). Then, I deleted the category on `wp-admin` and re-launched the app. What happened is that the app loaded the data of some products, removed the category filter, reloaded the data showing only one filter at the top but no products in the list. A pull to refresh solved the problem. I have the impression that there is some problem with concurrent calls that overlap each other.

So, my main question is: why not left the stored category selected unless the user remote the filter or select a different category? @ealeksandrov do you believe that it makes sense for us?

toupper:
@pmusolino  thanks a lot for your feedback :) 
Yes, we can leave the stored category selected as you suggested, the only problem in that case  is that, once we open the `FIlterProductCategoryListViewController`, no category would be selected, as the stored one is not part of the retrieved categories:



To fix it we could easily compare the retrieved product categories with the initialized selected category parameter, and if it is not included in them as it happens in this case, add it to the list. What do you think?

pmusolino:
> To fix it we could easily compare the retrieved product categories with the initialized selected category parameter, and if it is not included in them as it happens in this case, add it to the list. What do you think?

In my opinion, it makes sense that it does not appear in categories, if it has been deleted, even if it is still selected in the previous view. Also, adding the category may not be as easy as we imagine I think. Here too there is pagination in the middle, and we should take into account the extra category. So, in my opinion it's ok to not add it between the categories, but I'd like to receive a feedback also from @ealeksandrov.

ealeksandrov:
I did 3 tests and got correct final data each time. But yes - it looks like a glitch to see states switching (loading -> empty -> loading -> data). Probably it's because of race condition:

```
GET data // 1
if (category selected in filter) {
   GET category {
    GET data // 2
  }
}
```

Requests log:
Proxy Screenshot

Can we update a sync flow (without overcomplicating already non-trivial module) like:

```
if (category selected in filter) {
   GET category {
    GET data // 2
  }
} else {
  GET data // 1
}
```

1. If we can prevent extra (first, empty result) GET /products - my vote is to add this edge case handling 👍
2. I haven't seen cases of wrong output. But it's only a matter of luck to have correct order of two /products responses.
3. I vote against keeping old category and inserting it into filter list - too complex, not much of profit.

toupper:
@ealeksandrov @pmusolino Thanks a lot for your feedback 🙂 In the latest commits I implemented the suggestion from @ealeksandrov, where we skip the extra call. With that addition, the flow now looks as follows:

open products screen > show cached products > load local settings > synchronize filter product setting > request products with filter settings > show products

With this flow, we would still have a "glitch" where we show first the cached products and then the updated values according to the filter. This is something that we had before, but it is more prominent in this case since the the deletion of the `ProductCategory` filter produces a different set of products. If we still think that this flow might be confusing for the user (e.g category was deleted long time ago) we can close this PR and go with the approach suggested by Paolo.

On that note, I also fixed here a bug existing before, but that is more prominent with this new addition:
 If the retrieved/syncrhonized local filter settings are the same as the default ones `FilterProductListViewModel.Filters()` where all the filters are nil and the sort order default (e.g when remove the category filter because it was deleted), the data is not remotely synchronized. 
 This happens for instance when the user adds a filter, show products,  removes the filter thus saving the same settings values and the default ones, and restarts the app. Since the settings are the same as the default ones, the sync call is not performed in the `filters didSet` when showing the products view:

```
private var filters: FilterProductListViewModel.Filters = FilterProductListViewModel.Filters() {
        didSet {
            if filters != oldValue {
                updateLocalProductSettings(sort: sortOrder,
                                           filters: filters)
                updateFilterButtonTitle(filters: filters)

                resultsController.updatePredicate(siteID: siteID,
                                                  stockStatus: filters.stockStatus,
                                                  productStatus: filters.productStatus,
                                                  productType: filters.productType)

                /// Reload because `updatePredicate` calls `performFetch` when creating a new predicate
                tableView.reloadData()

                syncingCoordinator.resynchronize {}
            }
        }
    }
```
This doesn't happen the first time the products screen is shown since there is no local filter settings stored, thus retrieving an error and resyncing:

```
syncLocalProductsSettings { [weak self] (result) in
            if result.isFailure {
                self?.syncingCoordinator.resynchronize()
            }
        }
```
Most likely this bug was unnoticed since, even if we didn't synced remotely, we always show data (the cached one). I fixed by checking if we synced remotely in the `filters didSet`:
```
private var filters: FilterProductListViewModel.Filters = FilterProductListViewModel.Filters() {
        didSet {
            let contentIsNotSyncedYet = syncingCoordinator.highestPageBeingSynced ?? 0 == 0
            if filters != oldValue ||
                contentIsNotSyncedYet {
                ...
```
Of course suggestions to optimize this fix are welcome.

pmusolino:
@toupper can you merge the last commit of `develop` in this branch? In this way we don't need to reconfigure the machine to use Xcode 12, an old version of ruby and all the other things while testing this PR 🙇

toupper:
@pmusolino sure, done 😎

Yosemite/YosemiteTests/Stores/ProductCategoryStoreTests.swift
```diff
@@ -58,65 +58,75 @@ final class ProductCategoryStoreTests: XCTestCase {
...
+        let errorResponse: ProductCategoryActionError? = waitFor { [weak self] promise in
+            guard let self = self else {
+                return
+            }
+
+            let action = ProductCategoryAction.synchronizeProductCategories(siteID: self.sampleSiteID, fromPageNumber: self.defaultPageNumber) { error in
+                promise(error)
+            }
+
+            self.store.onAction(action)
```
pmusolino, line 77: Thanks for refactoring our old tests 👏

pmusolino (approved):
I'd say that it looks good to me now. Also if there is a small glitch, it will happens just when there is a selected category in filter which was also deleted remotely. It will be a very rare case, but should it happen, it is not of great impact handled in this way. So thanks for updating the logic like this. 👏

ealeksandrov:
With couple of tests I confirm that category reloading looks good now (delay between local search and filter update is reasonable) and requests are correct. Nice work! 👍

PR #5343, opened 2021-11-03
Add-Ons: Update ProductAddOn decoding to support heading type

Ecarrion:
# Why 

I noticed that a while ago my test site started to fail when decoding global add-ons (add-on groups). I couldn't find any significant change in the add-ons extension(Also I don't understand PHP well) but the issue seems to be that the **Heading** add-on type is not returning all the properties that it should return, hence making the decoding to fail.

# How

To fix that I'm adding a custom decoding, where if a property is not present or fails to decode(`Ints`are sent as empty `strings`) I'm falling back to a default `zero` value.

# Testing Steps

- Install the add-ons extension, set up an add-ons group and a heading add-on (or login to my test store)
 Screen Shot 2021-11-03 at 3 04 59 PM
 
 - Launch the app and make sure add-on groups decode doesn't fail (no error printed in the console)

 
Update release notes:


ealeksandrov:
I can't get error on current `develop` version, with heading as you described. Also tried random additional options, all works.
Can you invite me to your test store, so I can compare responses?

```
(lldb) po groups
▿ 1 element
  ▿ 0 : AddOnGroup
    - siteID : 196010018
    - groupID : 89
    - name : "Add-ons Group #1"
    - priority : 10
    ▿ addOns : 1 element
      ▿ 0 : ProductAddOn
        - type : Networking.AddOnType.heading
        - display : Networking.AddOnDisplay.dropdown
        - name : "Test heading"
        - titleFormat : Networking.AddOnTitleFormat.label
        - descriptionEnabled : 0
        - description : ""
        - required : 0
        - position : 0
        - restrictions : 0
        - restrictionsType : Networking.AddOnRestrictionsType.any_text
        - adjustPrice : 0
        - priceType : Networking.AddOnPriceType.flatFee
        - price : ""
        - min : 0
        - max : 0
        ▿ options : 1 element
          ▿ 0 : ProductAddOnOption
            ▿ label : Optional<String>
              - some : ""
            ▿ price : Optional<String>
              - some : ""
            ▿ priceType : Optional<AddOnPriceType>
              - some : Networking.AddOnPriceType.flatFee
            - imageID : nil
```

Ecarrion:
Hey, @ealeksandrov thanks for testing!

That is weird, this is the response I'm getting on my test site.

Screen Shot 2021-11-04 at 7 28 49 AM

Anyway, I have added you to my test site using the A8C email, let me know if you want to be added with a different email!

ealeksandrov (approved):
After connecting to your store I can see error and it's fixed by the PR.
So the main issue of PR is fixed ✅

But! When I've tried to recreate similar structure in my store - I've got even worse issue (while heading is still correct).

```
⛔️ Failed to sync add-on groups for siteID: 196010018. Error: typeMismatch(Swift.Array<Any>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "data", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "fields", intValue: nil)], debugDescription: "Expected to decode Array<Any> but found a dictionary instead.", underlyingError: nil))
```

We have same version of Product Add-ons plugin - 4.3.0.
For so many differences - maybe we need additional server-side investigation?

Responses intercepted:
[api-ernesto.txt](https://github.com/woocommerce/woocommerce-ios/files/7475718/api-ernesto.txt)
[api-evgeny.txt](https://github.com/woocommerce/woocommerce-ios/files/7475719/api-evgeny.txt)

Sent invite to my store if you want to confirm.

Ecarrion:
I'm not going to merge this while we hear back from the Rubik team https://github.com/woocommerce/woocommerce-product-addons/issues/708

Ecarrion:
I'm closing this as there are no plans to improve this any time soon.

PR #5358, opened 2021-11-04
[Mobile Payments] Prompt user to take live account out of test mode to do in-person payments

### LINKED ISSUE #4428: [Mobile Payments] Connect to Reader hangs if live site is in test mode

allendav:
Closes #4428 

@adamzelinski - would you review the in-app screenshot at the bottom of this description, let me know if there are any changes needed, and add it to the Figma? Thanks!

A NOTE ON LIVE ACCOUNTS and TEST MODE:

Bear in mind while testing this that a "live account" and "test mode" are (mostly) independent concepts. You can have 1) a live account not in test mode, 2) a live account in test mode and 3) a developer account in test mode. Only the first and last support card reader operations. Live accounts in test mode DO NOT support card reader operations.

A merchant can set their live site for test mode here (the checkbox can be checked and unchecked by the merchant and is unchecked by default)

live

So... the message "WooCommerce Payments is in Test Mode" is meant for a **merchant.**.  Merchants get **live** accounts. I didn't mention live accounts versus developer accounts on this page because merchants are very unlikely to know about developer accounts.

If a developer sets up their site with a developer account using WCPay developer tools - 1) they automatically are placed in test mode - and 2) they will **NOT** see this screen because their site is using a developer account i.e. not a live account.

Developer mode looks like this in wp-admin (note: the checkbox CANNOT be unchecked):

devmode

@joshheald @jaclync - if either of you could take a look, that would be great - just one of y'all are needed of course, but both of y'all are welcome. If you don't have access to a live site to test with, you're welcome to ask me for access to mine, or hack `WCPayAccountMapper` to return `true` for `isLive`

Changes:
- This is the final PR in the sequence to hook up prompting users to deactivate test mode if active on WooCommerce Payments for stores with "live" accounts (i.e. accounts that can actually accept real money.)
- Adding isLive and isInTestMode to the PaymentGatewayAccount model required `rake generate` as expected

To test:
- Set your live site for test mode (see screenshot above)
- Launch the app and enter settings, in-person payments
- Ensure you are prompted to take your site out of test mode ("WooCommerce Payments is in Test Mode")
- Take your site out of test mode
- Refresh using the "Refresh After Updating" button provided on the "WooCommerce Payments is in Test Mode" screen
- Ensure you no longer see the "WooCommerce Payments is in Test Mode" screen under In-Person Payments

Screenshots:

IMG_0231

Update release notes:

### DIFF
--- a/Networking/Networking/Model/PaymentGatewayAccount.swift
+++ b/Networking/Networking/Model/PaymentGatewayAccount.swift
@@ -30,6 +30,14 @@ public struct PaymentGatewayAccount: GeneratedCopiable, GeneratedFakeable {
     /// A boolean flag indicating if this Account is eligible for card present payments
     public let isCardPresentEligible: Bool
 
+    /// Indicates if the account is live (i.e. can accept actual payments)
+    public let isLive: Bool
+
+    /// Indicates if the gateway is set for test mode. This is NOT the same as
+    /// whether the account is live or not. You can have a live account set for
+    /// test mode, although we cannot accept in-person payments in that situation.
+    public let isInTestMode: Bool
+
     /// Struct initializer
     ///
     public init(siteID: Int64,
@@ -42,7 +50,9 @@ public struct PaymentGatewayAccount: GeneratedCopiable, GeneratedFakeable {
                 defaultCurrency: String,
                 supportedCurrencies: [String],
                 country: String,
-                isCardPresentEligible: Bool
+                isCardPresentEligible: Bool,
+                isLive: Bool,
+                isInTestMode: Bool
         ) {
         self.siteID = siteID
         self.gatewayID = gatewayID
@@ -55,5 +65,7 @@ public struct PaymentGatewayAccount: GeneratedCopiable, GeneratedFakeable {
         self.supportedCurrencies = supportedCurrencies
         self.country = country
         self.isCardPresentEligible = isCardPresentEligible
+        self.isLive = isLive
+        self.isInTestMode = isInTestMode
     }
 }
--- a/Fakes/Fakes/Networking.generated.swift
+++ b/Fakes/Fakes/Networking.generated.swift
@@ -489,7 +489,9 @@ extension PaymentGatewayAccount {
             defaultCurrency: .fake(),
             supportedCurrencies: .fake(),
             country: .fake(),
-            isCardPresentEligible: .fake()
+            isCardPresentEligible: .fake(),
+            isLive: .fake(),
+            isInTestMode: .fake()
         )
     }
 }
--- a/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
+++ b/Networking/Networking/Model/Copiable/Models+Copiable.generated.swift
@@ -286,7 +286,9 @@ extension PaymentGatewayAccount {
         defaultCurrency: CopiableProp<String> = .copy,
         supportedCurrencies: CopiableProp<[String]> = .copy,
         country: CopiableProp<String> = .copy,
-        isCardPresentEligible: CopiableProp<Bool> = .copy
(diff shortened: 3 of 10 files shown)


allendav:
I forgot to add unit test coverage. Will do.

Networking/Networking/Model/PaymentGatewayAccount.swift
```diff
@@ -30,6 +30,14 @@ public struct PaymentGatewayAccount: GeneratedCopiable, GeneratedFakeable {
     /// A boolean flag indicating if this Account is eligible for card present payments
     public let isCardPresentEligible: Bool
 
+    /// Indicates if the account is live (i.e. can accept actual payments)
+    public let isLive: Bool
+
+    /// Indicates if the gateway is set for test mode. This is NOT the same as
+    /// whether the account is live or not. You can have a live account set for
+    /// test mode, although we cannot accept in-person payments in that situation.
+    public let isInTestMode: Bool
```
joshheald, line 39: ❓ I think I'm right in understanding that `testMode` is a property of the site, not the account... with that in mind, is there anywhere more semantically correct that this property might live? If the answer's no, that's fine!

joshheald (approved):
Tests well for the outlined scenario. One code query. 

When attempting to take a payment for a store in test mode, we get a generic error about being unable to connect to the reader. Is there anything else planned to display this same information in the collect payment flow? Or is that more of an edge case we don't need to cover?
test-mode-purchase-error

Networking/Networking/Model/PaymentGatewayAccount.swift
allendav: `testMode` is not a site property but a property of some payment gateways, like WooCommerce Payments. I'd daresay most payment gateways have a "test" or "debug" mode that the merchant can enable in settings for that gateway.

The major downside is there is no standard as to what that test mode / debug mode setting is called. So over the years developers that have added test modes to their gateways have used all sorts of things.

If there had been a standard, we could have more easily added the concept of test mode to `PaymentGateway` - which has a REST API at `wc/v3/payment_gateways` and specifically for WCPay at `wc/v3/payment_gateways/woocommerce_payments` 

But without that, we'd could probably entertain saving a whitelist of fields of various names (starting with the test_mode setting in WCPay) from `settings` portion for each gateway in the response to its `PaymentGateway` model, and then use that instead of getting test mode from the WCPay plugins' account response as we are doing here.

We'd still want this "associated account" model (PaymentGatewayAccount) for those concepts that are related to the backend account connected to a payment gateway (e.g. is a live account, has restrictions coming from the processor, etc) but we could at least stop relying on test_mode from here if we moved it to `PaymentGateway`

That's all assuming a shop manager can access `wc/v3/payment_gateways/woocommerce_payments` of course. I haven't checked that.

How about we start a refactor proposal post or issue for this? cc @ctarda

allendav:
> we get a generic error about being unable to connect to the reader. Is there anything else planned to display this same information in the collect payment flow?

We should fix that too - to show a more detailed error there. I'll start another PR for that.

Networking/Networking/Model/PaymentGatewayAccount.swift
joshheald: Thanks for the explanation, that makes a lot more sense... I'd not really twigged that this was an account for the gateway, rather than the. I should have, given the name, but I blame Friday brain!

That refactor would be nice but sounds like it might be pretty low priority

ghost:
The design looks good to me thanks @allendav

PR #5372, opened 2021-11-08
Issue/4385 empty filter view

### LINKED ISSUE #4385: Product List: the no results screen should specify when no products match the selected filter versus saying there are no products at all
Steps to reproduce:

1. Go to Products > Filter and filter until there you see the no results screen (i.e. draft status, out of stock).

**Result:** I expected the no results view when filtering products to let me know no results matched the filter and give me an easy way to clear all filters, and instead if I filter and there are no results I'm shown a "No products yet" screen and a call to action to add a product.

IMG_4218
<sup>Tested with WCiOS 6.8.0.0 TestFlight beta on iPhone 6S iOS 14.6.</sup>

/hat tip @malinajirka for raising this during 6.8 beta testing (internal reference: p5T066-2mh-p2#comment-8530).

> When I set a filter that doesn’t match any products, “No products yet – start selling today by adding …” empty screen is shown -> I’d expect an empty screen saying that no products match the selected filter

malinajirka:
Closes #4385 

Before this PR, the app would show identical empty view for both "Store has no products" and "No products match the selected filters" screens. This PR shows a more specific message "No matching products found" and action "Clear Filters" for the latter.

Questions:
1. I've created a new track event - do we need to do something else to start tracking this event (I've updated the internal xls)
2. I've removed "hint/details" message on the "No matching products found" screen. Do you think we should keep it?

To Test:
1. Open products screen
2. Tap on Filters
3. Select filter(s) that don't match any products (example for my store: Status - Pending Review, Product Type - Grouped)
4. Tap on Show Products
5. Notice an empty view saying "No matching products" and "Clear Filters" button is shown
6. Tap on "Clear filters" and notice the filters get cleared and the products re-loaded
7. Switch to a store which doesn't have any products
8. Open Product list screen
9. Notice the original "No products yet" and "Add Product" button is shown

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
@@ -680,7 +680,7 @@ private extension ProductsViewController {
 
     func clearFilter(sourceBarButtonItem: UIBarButtonItem? = nil, sourceView: UIView? = nil) {
         ServiceLocator.analytics.track(.productListClearFiltersTapped)
-        self.filters = FilterProductListViewModel.Filters()
+        filters = FilterProductListViewModel.Filters()
     }
 
     /// Presents products survey
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
@@ -253,15 +253,6 @@ private extension ProductsViewController {
         }
         coordinatingController.start()
     }
-
-    func clearFilter(sourceBarButtonItem: UIBarButtonItem? = nil, sourceView: UIView? = nil) {
-        ServiceLocator.analytics.track(.productListClearFiltersTapped)
-        self.filters = FilterProductListViewModel.Filters(stockStatus: nil,
-                                                          productStatus: nil,
-                                                          productType: nil,
-                                                          productCategory: nil,
-                                                          numberOfActiveFilters: 0)
-    }
 }
 
 // MARK: - View Configuration
@@ -685,6 +676,11 @@ private extension ProductsViewController {
             ServiceLocator.analytics.track(.productFilterListDismissButtonTapped)
         })
         present(filterProductListViewController, animated: true, completion: nil)
+    }
+
+    func clearFilter(sourceBarButtonItem: UIBarButtonItem? = nil, sourceView: UIView? = nil) {
+        ServiceLocator.analytics.track(.productListClearFiltersTapped)
+        self.filters = FilterProductListViewModel.Filters()
     }
 
     /// Presents products survey
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
@@ -255,6 +255,7 @@ private extension ProductsViewController {
     }
 
     func clearFilter(sourceBarButtonItem: UIBarButtonItem? = nil, sourceView: UIView? = nil) {
+        ServiceLocator.analytics.track(.productListClearFiltersTapped)
         self.filters = FilterProductListViewModel.Filters(stockStatus: nil,
                                                           productStatus: nil,
                                                           productType: nil,
--- a/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
+++ b/WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
@@ -733,9 +733,17 @@ private extension ProductsViewController {
     ///
     func displayNoResultsOverlay() {
         let emptyStateViewController = EmptyStateViewController(style: .list)
-        let config = filters.numberOfActiveFilters == 0 ? createNoProductsConfig() : createNoProductsMatchFilterConfig()
+        let config = createFilterConfig()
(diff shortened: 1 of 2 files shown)


WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -253,6 +253,15 @@ private extension ProductsViewController {
         }
         coordinatingController.start()
     }
+
+    func clearFilter(sourceBarButtonItem: UIBarButtonItem? = nil, sourceView: UIView? = nil) {
```
malinajirka: ❓ I added this next to `addProduct` action. However, it seems it's located in `Navigation Bar Actions` extension even though it's technically not a navigation bar action. Is this fine or should we move it somewhere else?

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -253,6 +253,15 @@ private extension ProductsViewController {
...
+        ServiceLocator.analytics.track(.productListClearFiltersTapped)
+        self.filters = FilterProductListViewModel.Filters(stockStatus: nil,
```
ctarda: The `self` should not be necessary, it is implicit and we tend to not make it explicit in Swift (it makes the code a bit more concise).

Also, given all the properties in Filters are passed as nil, we could also rely on the initializer without parameters that Filters provides. So this line could be:

`filters = FilterProductListViewModel.Filters()`

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -728,21 +737,43 @@ private extension ProductsViewController {
     ///
     func displayNoResultsOverlay() {
         let emptyStateViewController = EmptyStateViewController(style: .list)
+        let config = filters.numberOfActiveFilters == 0 ? createNoProductsConfig() : createNoProductsMatchFilterConfig()
```
ctarda: This is a bit of a personal preference, but I don't find these kind of assignments too easy to parse. I would suggest encapsulating if in a method, so that, here, the code could read like `let config = filterConfiguration()` or similar, hiding the ternary operator (which, I know from my own experience, is the opposite we do on the Android codebase)

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
```diff
@@ -728,21 +737,43 @@ private extension ProductsViewController {
...
+    ///
+    func createNoProductsMatchFilterConfig() ->  EmptyStateViewController.Config {
+        let message = NSLocalizedString("No matching products found",
+                                        comment: "The text on the placeholder overlay when no products match the filter on the Products tab")
+        let buttonTitle = NSLocalizedString("Clear Filters",
+                                            comment: "Action to add product on the placeholder overlay when no products match the filter on the Products tab")
+        return EmptyStateViewController.Config.withButton(
+            message: .init(string: message),
+            image: .emptyProductsTabImage,
+            details: "",
```
ctarda, line 777: I don't see a problem in removing the hint, but it might be a good idea to do the same Android does?

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
ctarda: There is a extension for Actions, that encapsulates, for example, `filterButtonTapped`. Moving this method to that particular extension might be slightly more cohesive, but to be honest, I think either would work.

malinajirka: Yeah that feels better, fixed in https://github.com/woocommerce/woocommerce-ios/pull/5372/commits/cbd9a6cdd0f0920421ad321ea37b141391e46d1c.

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
malinajirka: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5372/commits/cbd9a6cdd0f0920421ad321ea37b141391e46d1c and https://github.com/woocommerce/woocommerce-ios/pull/5372/commits/6a3f93674522be07a79e57ccd49f3747bc76cb98

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
malinajirka: Yeah, I see your point and I agree, fixed in https://github.com/woocommerce/woocommerce-ios/pull/5372/commits/a2e64927c12537356911ed797c80001717c8e8cc

WooCommerce/Classes/ViewRelated/Products/ProductsViewController.swift
malinajirka: Android doesn't use any hint :).

ctarda (approved):
Looks great @malinajirka ! 👏🏼 👏🏼 👏🏼 

This is how it looks when there are no products matching the filters:

(the Clear Filters button works as expected)

And this is a store without Products:


:shipit: 

As far as I am aware, if the Tracks event is in the xls, all that is left is to register it with Tracks (which might have been done on Android already?)

PR #5373, opened 2021-11-08
Simple Payments: Rename Quick Order to Simple Payments

Ecarrion:
# Why

As requested on p91TBi-6nk#comment-7153 we are renaming "Quick Order" to "Simple Payments".

Reasoning:
> Searches with the word ‘payments’ rank higher than searches for orders. While the end user may see it more of an order, for our merchants it’s much more about getting paid (and collecting payments). By using ‘payments’ in the term, we’re also creating a much stronger link between the umbrella of “in-person payments” and the sub-feature of the card reader products. As we continue to grow our in-person payments offerings, our focus should really be on the getting paid part, as much as possible.

# Event Registration
- [x] https://github.com/Automattic/tracks-events-registration/pull/633
- [x] https://github.com/Automattic/tracks-events-registration/pull/634
- [x] https://github.com/Automattic/tracks-events-registration/pull/635
- [x] https://github.com/Automattic/tracks-events-registration/pull/636
- [x] https://github.com/Automattic/tracks-events-registration/pull/637    

Update release notes:

### DIFF
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -124,9 +124,9 @@ extension WooConstants {
         case addPaymentMethodWCShip = "https://wordpress.com/me/purchases/add-payment-method"
 
 #if DEBUG
-        case quickOrderPrototypeFeedback = "https://automattic.survey.fm/woo-app-quick-order-testing"
+        case simplePaymentsPrototypeFeedback = "https://automattic.survey.fm/woo-app-simple-payments-testing"
 #else
-        case quickOrderPrototypeFeedback = "https://automattic.survey.fm/woo-app-quick-order-production"
+        case simplePaymentsPrototypeFeedback = "https://automattic.survey.fm/woo-app-simple-payments-production"
 #endif
 
         /// Returns the URL version of the receiver
--- a/WooCommerce/Classes/System/WooConstants.swift
+++ b/WooCommerce/Classes/System/WooConstants.swift
@@ -124,9 +124,9 @@ extension WooConstants {
         case addPaymentMethodWCShip = "https://wordpress.com/me/purchases/add-payment-method"
 
 #if DEBUG
-        case simplePaymentsPrototypeFeedback = "https://automattic.survey.fm/woo-app-simple-payments-testing"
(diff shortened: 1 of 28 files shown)


shiki:
Note that they're discussing using "Simple Payment" instead of the plural "Payments" right now. 😅

Ecarrion:
> Note that they're discussing using "Simple Payment" instead of the plural "Payments" right now. 😅

@shiki but I think that is for just for the copy on the quick menu where the flow is started, which we haven't built yet, so we should be fine?

shiki:
That's correct. My bad. 😅

WooCommerce/Classes/System/WooConstants.swift
```diff
@@ -124,9 +124,9 @@ extension WooConstants {
         case addPaymentMethodWCShip = "https://wordpress.com/me/purchases/add-payment-method"
 
 #if DEBUG
-        case quickOrderPrototypeFeedback = "https://automattic.survey.fm/woo-app-quick-order-testing"
+        case simplePaymentsPrototypeFeedback = "https://automattic.survey.fm/woo-app-simple-payments-testing"
 #else
-        case quickOrderPrototypeFeedback = "https://automattic.survey.fm/woo-app-quick-order-production"
+        case simplePaymentsPrototypeFeedback = "https://automattic.survey.fm/woo-app-simple-payments-production"
```
rachelmcr: It looks like these two surveys don't exist yet, so I'm getting a 404 error when I tap the "Give Feedback" button on the banner. Who has access to set these up?

rachelmcr (commented):
I confirmed that all instances of `Quick Order` (and `QuickOrder` or `quick_order`, in all cases) have been renamed.

There is an issue with the new survey URLs not working, though — we should make sure those surveys are set up before releasing this change.

WooCommerce/Classes/System/WooConstants.swift
Ecarrion: Wow, I totally missed that! Thanks for spotting it!

I have renamed the survey URLs to their previous value so we don't end up with two different surveys. (The one being released in 7.9 and this one 8.0+)

https://user-images.githubusercontent.com/562080/140949439-9461767f-6253-434f-a298-8d9d5b9fc30d.mov

rachelmcr (approved):
Good to go! :shipit:

PR #5374, opened 2021-11-08
Order list: implemented the screens for showing the date range filters

pmusolino:
Part of https://github.com/woocommerce/woocommerce-ios/issues/5243

## Description
In this PR, I implemented the login for showing the filters and for passing data between the flow. I also implemented the login for managing the custom date range filter. The PR is big because of the changes in the various `xib` files.

## Changes
- Added `ordersDateRange` between the options of `FilterListViewController`. This will allow us to present the view controller for filtering orders by dates.
- Implemented `OrderDatesFilterViewController` which allow us to select different options of dates (any, 2 days ago, etc…)
- Implemented `DateRangeFilterViewController`, which allow us to filter the orders by a range of dates, following the new design proposed by Adam.
- Implemented the new struct `OrderDateRangeFilter` (updated the old `OrderDateRangeFilterEnum`) which contains the various type (enum) of Date Range plus the start and end date in case the user will choose a custom range.
- Updated the constraints of `DatePickerTableViewCell` lateral constraints were incorrect.

## Testing
*Important*: Currently the filters are not filtering the orders. This will be part of another PR.
1. Launch the app it in debug (because of the feature flag).
2. Open the order list.
3. You should be able to tap the `Filter` button.
4. You should be able to view the `Filters` screen with two options: Order Status and Date Range.
5. Under Date Range, you should be able to select just one filter.
6. Tap the custom date range, you should be able to select just the start date, just the end date or both.
7. Go back, you should be able to see the dates selected in the Date Range screen.
8. Go back, you should be able to see the dates selected in the Filters screen.
9. You should be able also to change the date filters.

## Video
https://user-images.githubusercontent.com/495617/140803442-097bc519-7b19-44e6-bc9f-930b9d9174e8.mp4

Update release notes:


WooCommerce/Classes/ViewRelated/Filters/FilterListViewController.swift
```diff
@@ -234,6 +236,14 @@ private extension FilterListViewController {
                                                                                                       selectedCategory: selectedProductCategory,
                                                                                                       onProductCategorySelection: selectedValueAction)
                 self.listSelector.navigationController?.pushViewController(filterProductCategoryListViewController, animated: true)
+            case .ordersDateRange:
+                let selectedOrderFilter = selected.selectedValue as? OrderDateRangeFilter
+                let datesFilterVC = OrderDatesFilterViewController(selected: selectedOrderFilter) { dateRangeFilter in
+                    selected.selectedValue = dateRangeFilter
+                    self.updateUI(numberOfActiveFilters: self.viewModel.filterTypeViewModels.numberOfActiveFilters)
+                    self.listSelector.reloadData()
```
jaclync, line 244: nit: just wondering, can we call `selectedValueAction` here?

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    func configureStartDateTitle(cell: TitleAndValueTableViewCell) {
+        cell.updateUI(title: Localization.startDateTitle, value: startDate?.toString(dateStyle: .medium, timeStyle: .none))
+        cell.selectionStyle = .none
+    }
+
+    func configureStartDatePicker(cell: DatePickerTableViewCell) {
+        if let startDate = startDate {
+            cell.getPicker().setDate(startDate, animated: false)
```
jaclync, line 160: I'm guessing setting `animated` has some issues? 😅

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+            break
+        case let cell as TitleAndValueTableViewCell where row == .endDateTitle:
+            configureEndDateTitle(cell: cell)
+            break
+        case let cell as DatePickerTableViewCell where row == .endDatePicker:
+            configureEndDatePicker(cell: cell)
+            break
+        default:
+            fatalError()
+            break
```
jaclync: nit: `break` is probably not necessary in each case?

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+        ///
+        for row in Row.allCases {
+            tableView.registerNib(for: row.type)
+        }
+
+        tableView.backgroundColor = .listBackground
+        tableView.removeLastCellSeparator()
+    }
+
+    func configureRows() {
```
jaclync: nit: `configure*` sounds more like a one-time setup, maybe `update*` would be more clear?

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+
+// MARK: - UITableViewDelegate Conformance
+//
+extension DateRangeFilterViewController: UITableViewDelegate {
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+
+        switch rows[indexPath.row] {
+        case .startDateTitle:
+            startDateExpanded = !startDateExpanded
```
jaclync: nit: another way to toggle the value that might read more naturally 😄 

```suggestion
            startDateExpanded.toggle()
```

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
```diff
@@ -0,0 +1,231 @@
+import UIKit
+
+// MARK: - DateRangeFilterViewController
+// Allow to filter the orders by a range of dates
+//
+final class DateRangeFilterViewController: UIViewController {
```
jaclync, line 6: nice! I think this view controller can be used for any date range picker use cases

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
```diff
@@ -0,0 +1,231 @@
...
+
+    @IBOutlet private weak var tableView: UITableView!
+
+    // Completion callback
+    //
+    typealias Completion = (Date?, Date?) -> Void
```
jaclync: nit: might help clarify the parameters

```suggestion
    typealias Completion = (_ startDate: Date?, _ endDate: Date?) -> Void
```

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/OrderDatesFilterViewController.swift
```diff
@@ -0,0 +1,182 @@
...
+        for row in OrderDateRangeFilterEnum.allCases {
+            tableView.registerNib(for: row.cellType)
+        }
+
+        tableView.backgroundColor = .listBackground
+        tableView.removeLastCellSeparator()
+    }
+
+    func configureRows() {
+        rows = [.init(filter: .any),
```
jaclync, line 68: nit: might be nice to align the indentation for each filter :)

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/OrderDatesFilterViewController.swift
```diff
@@ -0,0 +1,182 @@
...
+        if let startDateFormatted = row.startDate?.toString(dateStyle: .medium, timeStyle: .none) {
+            formattedDates = startDateFormatted
+        }
+        if row.startDate != nil && row.endDate != nil {
+            formattedDates = formattedDates + " - "
+        }
+        if let endDateFormatted = row.endDate?.toString(dateStyle: .medium, timeStyle: .none) {
+            formattedDates = formattedDates + endDateFormatted
+        }
+
```
jaclync, line 169: How about using [`DateIntervalFormatter` for the date range string](https://developer.apple.com/documentation/foundation/dateintervalformatter), so that it can be localized automatically if both dates are non-nil?

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/OrderDatesFilterViewController.swift
```diff
@@ -0,0 +1,182 @@
...
+    ///
+    func configure(_ cell: UITableViewCell, for row: OrderDateRangeFilter, at indexPath: IndexPath) {
+        switch cell {
+        case let cell as BasicTableViewCell:
+            configureStandardFilter(cell: cell, row: row)
+        case let cell as TitleAndValueTableViewCell where row.filter == .custom:
+            configureCustomDateRangeFilter(cell: cell, row: row)
+        default:
+            fatalError()
+            break
```
jaclync: nit: similar to above, `break` might not be necessary?

jaclync (approved):
LGTM! Quite some work for the custom date range option 😅 

Code: ✅ just some non-blocking suggestions
Testing: ✅ is there a way to remove the start/end date like in the product sale dates?

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
pmusolino: It's okay actually, but as I expect the date to be already selected when the cell is rendered, I don't want the picker to move with an animation from the current date. If the chosen date and the current date were on different months, reopening the screen you would see the picker change month. It might seem strange.

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
pmusolino: TIL 💯

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/Custom Range Filter/DateRangeFilterViewController.swift
pmusolino: Absolutely. I have made the comment more generic.

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/OrderDatesFilterViewController.swift
pmusolino: I didn't used the `DateIntervalFormatter` because here we can select just the start date or just the end date, and the DateIntervalFormatter accept only non optionals value, maybe you had in mind another way to use `DateIntervalFormatter`?

pmusolino:
Thank you so much @jaclync for the review and all the suggestions! 

> Quite some work for the custom date range option

Yes, I tried initially to reuse the existing code (`ListSelectorViewController`) to implement the various options, but it was basically impossible without a massive refactoring to handle all the different actions needed for the date range filters.

I updated the code following your suggestions, and I left just two comments about the `animated` question and about the `DateIntervalFormatter`. 🙇 I'm going to merge this PR now, but free to reply if you have extra suggestions.

WooCommerce/Classes/ViewRelated/Orders/Order Filters/Date Range Filter/OrderDatesFilterViewController.swift
jaclync: I was thinking to use `DateIntervalFormatter` when both dates are non-nil, otherwise just the date string 😄

PR #5379, opened 2021-11-09
Jetpack CP: benefits detail UI

jaclync:
Part of #5362 
- [x] ⚠️ Make sure https://github.com/woocommerce/woocommerce-ios/pull/5378 is merged before merging ⚠️ 

## Why

We want to inform users with Jetpack Connection Package (JCP) sites (without Jetpack-the-plugin) about the benefits of installing Jetpack-the-plugin. In the My Store tab, we show a bottom banner for JCP sites with two actions:

- Tapping on the banner shows a screen with Jetpack benefits with a CTA to install Jetpack
- Tapping on the dismiss button dismisses the banner for the site

This PR just contains the UI for the detail screen for easier code review. The display under a feature flag will be in a separate PR.

## Changes

- Added 3 new images: `alarmBellRingImage`, `analyticsImage`, `multipleUsersImage` with PDF assets
- Created `JetpackBenefitsView` with a hosting controller, along with a reusable `JetpackBenefitItem` for each benefit item. The top container is scrollable when needed, for narrower screen height and larger font sizes

## Testing

This UI isn't used yet, feel free to play with SwiftUI previews with more configurations!

## Example screenshots

portrait | landscape
-- | --
Simulator Screen Shot - iPhone 11 - 2021-11-09 at 15 07 07 | Simulator Screen Shot - iPhone 11 - 2021-11-09 at 15 07 15
Simulator Screen Shot - iPhone 11 - 2021-11-09 at 15 07 28 | Simulator Screen Shot - iPhone 11 - 2021-11-09 at 15 07 30

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

Issue #5362, opened 2021-11-05
Jetpack CP: Jetpack benefits bottom banner


jaclync:
Refs:
- Android issue https://github.com/woocommerce/woocommerce-android/issues/4864
- Android PR - banner https://github.com/woocommerce/woocommerce-android/pull/5022
- Android PR - display banner https://github.com/woocommerce/woocommerce-android/pull/4995

Dependency:
- Backend to detect a JCP site https://github.com/woocommerce/woocommerce-ios/issues/5364

When the user visits the stats screen, show a bottom banner that links to a screen for them to know about Jetpack benefits.

Subtasks:
- [x] Create a Jetpack benefits banner https://github.com/woocommerce/woocommerce-ios/pull/5378
- [x] Create a screen about Jetpack benefits https://github.com/woocommerce/woocommerce-ios/pull/5379
- [x] Display the Jetpack benefits banner for JCP sites in the My Store tab, and navigate to the Jetpack benefits screen when tapping on the banner https://github.com/woocommerce/woocommerce-ios/pull/5380

Design:
<img width="363" alt="Screen Shot 2021-11-05 at 2 58 56 PM" src="https://user-images.githubusercontent.com/1945542/140470929-911706d5-ff50-4c91-ae65-56f853165002.png">

jaclync:
Moved the banner dismissal app settings to https://github.com/woocommerce/woocommerce-ios/issues/5368, closing this issue now

PR #5380, opened 2021-11-09
Jetpack CP: display benefits banner under a feature flag

jaclync:
Part of #5362 

- [x] ⚠️ Make sure https://github.com/woocommerce/woocommerce-ios/pull/5379 is merged before merging ⚠️ 

## Why

We want to inform users with Jetpack Connection Package (JCP) sites (without Jetpack-the-plugin) about the benefits of installing Jetpack-the-plugin. In the My Store tab, we show a bottom banner for JCP sites with two actions:
- Tapping on the banner shows a screen with Jetpack benefits with a CTA to install Jetpack
- Tapping on the dismiss button dismisses the banner for the site

This PR just contains the logic to display the banner at the bottom of the dashboard tab.

## Changes

Some SwiftUI challenges I ran into:
- The content height is higher than the actual banner height, on launch and after rotation
- After solving the issue above, the content height is 44px higher than the actual banner height when the offline banner is present

I spent quite some time figuring out how to resolve the two issues above. Even though the banner is set with `fixedSize(horizontal: false, vertical: true)` to fix its ideal height to content height, the height is still higher than expected. When the parent view frame changes in UIKit, it does not seem to notify SwiftUI to recalculate the height. In this PR, the solution I ended up with is to set a height constraint and calculate/update the height in `viewDidLayoutSubviews`, and pin the bottom edge with the safe area layout guide to accommodate the offline banner when it is present. Please feel free to suggest other ways to support self-sizing SwiftUI views in UIKit!

## Notes

- Because we use Auto Layout to pin the bottom of the content view, I created two separate constraints for the cases when the content bottom is pinned to the container (no banner) and to the top of the banner. It feels a bit ugly 😞 
- I plan to refactor the banner showing/hiding logic later to use an observable (combining dashboard content view publisher and a boolean publisher to show/hide the banner), when the criteria become async after we persist the banner dismiss state

## Testing

- Launch from PR branch in logged-in state --> there should be a banner shown at the bottom of the My Store tab
- Try rotating the simulator/device, or changing the font sizes --> the banner should stay at the bottom
- (Optional) Try turning the device offline (or rebuilding the app after setting [this line](https://github.com/woocommerce/woocommerce-ios/blob/ff70064acbd6f22da16bd11a8b6013aa0e0f0ded/WooCommerce/Classes/System/WooNavigationController.swift#L112)'s `status` parameter to `notReachable`) --> the banner should stay at the bottom above the offline banner
- Tap anywhere in the banner that is outside of the dismiss CTA --> a detail screen should be shown with all the Jetpack benefits
- Tap "Not Now" or drag down to dismiss the detail screen
- Tap the dismiss CTA ("X") on the banner --> the banner should disappear

## Example screenshots

portrait with offline banner | landscape
-- | --
Simulator Screen Shot - iPhone 11 - 2021-11-09 at 15 57 42 | Simulator Screen Shot - iPhone 11 - 2021-11-09 at 15 35 45

Update release notes:


Ecarrion (approved):
Hey @jaclync I faced a similar situation where the top banner was not being auto-resized in a `SwiftUI` View. I ended up using a similar strategy to yours but with the logic isolated in a wrapper view.

See if maybe [TopBannerWrapperView](https://github.com/woocommerce/woocommerce-ios/blob/develop/WooCommerce/Classes/ViewRelated/Top%20Banner/TopBannerWrapperView.swift) can help with with ideas.

jaclync:
> See if maybe `TopBannerWrapperView` can help with ideas.

Thanks @Ecarrion for the review and suggestion! I checked `TopBannerView` before because I remembered we were thinking about converting it to SwiftUI, but it looks like `TopBannerView` is still implemented with UIKit? I'm not sure if I missed something there 🤔 

Inspired by the wrapper view, I moved the height constraint logic from `DashboardViewController` to the hosting controller in https://github.com/woocommerce/woocommerce-ios/pull/5380/commits/29499aacbef0dddbf9f66b986d2c4586ac78228d - lemme know if you think it can be improved further!

Ecarrion:
> Inspired by the wrapper view, I moved the height constraint logic from DashboardViewController to the hosting controller in 29499aa - lemme know if you think it can be improved further!

I think that is cool! Mixing `SwiftUI` and `UIKit` has some many quirks that it is very hard(impossible?) to make it perfect and pretty! 😅

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -210,6 +215,22 @@ private extension DashboardViewController {
...
+            let benefitsController = JetpackBenefitsHostingController()
+            benefitsController.setActions {
+                // TODO: 5370 - Navigate to install Jetpack
+            } dismissAction: { [weak self] in
+                self?.dismiss(animated: true, completion: nil)
+            }
+            self.present(benefitsController, animated: true, completion: nil)
+        } dismissAction: {
+            // TODO: 5362 - Persist dismiss state per site
+            self.hideJetpackBenefitsBanner()
```
selanthiraiyan: Should we use `weak self` here?

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
```diff
@@ -303,6 +324,54 @@ private extension DashboardViewController {
...
+    func hideJetpackBenefitsBanner() {
+        contentBottomToJetpackBenefitsBannerConstraint?.isActive = false
+        contentBottomToContainerConstraint?.isActive = true
+        if isJetpackBenefitsBannerShown() {
+            bottomJetpackBenefitsBannerController.view?.removeFromSuperview()
+            remove(bottomJetpackBenefitsBannerController)
+        }
+    }
+
+    func isJetpackBenefitsBannerShown() -> Bool {
```
selanthiraiyan: nit: What do you think about changing this into a computed property?

jaclync: Makes sense 👍 updated in https://github.com/woocommerce/woocommerce-ios/pull/5380/commits/9d4660d0c4b8b7f9cefcca3eb0e3826998c765c5

WooCommerce/Classes/ViewRelated/Dashboard/DashboardViewController.swift
jaclync: Nice catch!! Updated in https://github.com/woocommerce/woocommerce-ios/pull/5380/commits/9d4660d0c4b8b7f9cefcca3eb0e3826998c765c5

jaclync:
Thank you both for the suggestions @Ecarrion @selanthiraiyan! Merging now 😃

PR #5382, opened 2021-11-09
[Mobile Payments] Improved minimum WCPay plugin version comparison

### LINKED ISSUE #5357: [Mobile Payments] Reconsider whether .numeric is appropriate for version comparison
Related to #5356 

We're using .numeric to compare versions in `isWCPayVersionSupported`. It seems to work, but Apple's documentation doesn't explicitly say it handles strings with multiple decimals (like a version string with a patch does, i.e. M.m.p) so we are relying on some undocumented behavior here.

We should reconsider using .numeric and at least add more unit tests to cover minimum supported versions without patch releases (e.g. 3.5) since plugins often drop the patch release for the first release of a minor (e.g. 3.5 instead of 3.5.0)

cc @joshheald

allendav:
Fixes #5357 

Changes:
- Adds a new String extension to do version comparison
- Changes onboarding to use that instead
- Expands onboarding unit tests to cover major.minor and major.minor.patch versions better

To test:
- Ensure unit tests pass
- Edit the WCPay plugin to change the version in the header to 3.1.8
- Launch the app, enter settings and in-person payments and confirm "unsupported ... version" is displayed
- Edit the WCPay plugin to change the version in the header to 3.2
- Refresh the "unsupported ... version" screen and ensure it still says "unsupported ... version"
- Edit the WCPay plugin to change the version in the header to 3.2.1 (the minimum currently required)
- Refresh the "unsupported ... version" screen and ensure it shows manage/buy/manual
- Edit the WCPay plugin to change the version in the header to 3.3
- Restart the app, enter settings and in-person payments and confirm it shows manage/buy/manual
- For bonus points, edit `minimumSupportedWCPayVersion` in `CardPresentPaymentsOnboardingUseCase` to 3.3, restart the app, enter settings and in-person payments and confirm it shows manage/buy/manual

Update release notes:


WooCommerce/WooCommerceTests/Extensions/StringHelpersTests.swift
```diff
@@ -0,0 +1,40 @@
+import XCTest
+@testable import WooCommerce
+
+/// String+Helpers: Unit Tests
+///
+class StringHelpersTests: XCTestCase {
```
jaclync: nit: we usually mark a class as `final` if we don't see it being subclassed in the near future (which is most likely the case for test classes)

```suggestion
final class StringHelpersTests: XCTestCase {
```

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,32 @@ extension String {
...
         return !isEmpty
     }
+
+    /// Compares two strings as versions.
+    ///
+    /// Returns `orderedAscending` if the argument represents a newer version
+    /// Returns `orderedSame` if the argument represents the same version
+    /// Returns `orderedDescending` if the argument represents an older version
+    ///
+    func compareAsVersion(to: String) -> ComparisonResult {
```
jaclync: How about we implement the [PHP `version_compare` function](https://www.php.net/manual/en/function.version-compare.php) since this is what [WordPress plugin versions](https://developer.wordpress.org/plugins/plugin-basics/header-requirements/) follow?

> The function first replaces _, - and + with a dot . in the version strings and also inserts dots . before and after any non number so that for example '4.3.2RC1' becomes '4.3.2.RC.1'. Then it compares the parts starting from left to right. If a part contains special version strings these are handled in the following order: any string not found in this list < dev < alpha = a < beta = b < RC = rc < # < pl = p. This way not only versions with different levels like '4.1' and '4.1.2' can be compared but also any PHP specific version containing development state.

jaclync (commented):
Left an initial suggestion to implement the same version check as PHP!

Also, do you know how to test this step with a WC store or in the code?

> Edit the WCPay plugin to change the version in the header to 3.1.8

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: Done in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/77a17af907d13048225fb0dea7203378168bc738

This (adding PHP style version compare) is a big commit. Would you mind re-reviewing it?

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+
+        /// Replace _ - and + with a .
+        ///
+        func replaceUnderscoreDashAndPlusWithDot(inString: String) -> String {
```
jaclync: Wow the comparison is quite complicated, apparently I shared the documentation without reading it thoroughly! 😆  Since the logic isn't trivial, I'd recommend moving all the code to a separate struct helper like `struct WordPressPluginVersionComparator`, that has a public-facing `func compare(_ lhs: String, _ rhs: String) -> ComparisonResult` and these helper functions can be private within the struct. What do you think?

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+            var outString = inString
+            let replaceTargets = ["_", "-", "+"]
+            for replaceTarget in replaceTargets {
+                outString = outString.replacingOccurrences(of: replaceTarget, with: ".")
+            }
+            return outString
+        }
```
jaclync: nit:

Perhaps these can be simplified by using regular expression? I haven't tested it yet, something like:

```suggestion
        func replaceUnderscoreDashAndPlusWithDot(_ string: String) -> String {
            string.replacingOccurrences(of: "[_-+]", with: ".", options: .regularExpression)
        }
```

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+
+                if characterAtIndex.isNumber && !characterBeforeIsNumberOrDot {
+                    outString += "."
+                }
+
+                outString += String(characterAtIndex)
+            }
+
+            return outString
+        }
```
jaclync: Is it possible to implement this with regex with capturing groups? I haven't tested it yet, maybe:

```suggestion
        func insertDotsBeforeAndAfterAnyNonNumber(_ string: String) -> String {
            string.replacingOccurrences(of: "([^0-9]+)", with: ".$1.", options: .regularExpression)
        }
```

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+
+        /// Ranked per https://www.php.net/manual/en/function.version-compare.php
+        ///
+        enum ComponentScore: Comparable {
```
jaclync: nit: this enum could be private within the struct if you're moving the logic to a separate comparator struct

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+            if componentCharacterSet.isSubset(of: .decimalDigits) {
+                return .number
+            }
+
+            if stringComponent == "pl" || stringComponent == "p" {
+                return .patch
+            }
+
+            return .other
+        }
```
jaclync: nit: these could be in the enum's initializer:

```suggestion
    extension ComponentScore {
        func init(stringComponent: String) {
            if stringComponent == "dev" {
                return self = .dev
            }
            if stringComponent == "alpha" || stringComponent == "a" {
                return self = .alpha
            }
            if stringComponent == "beta" || stringComponent == "b" {
                return self = .beta
            }
            if stringComponent == "RC" || stringComponent == "rc" {
                return self = .RC
            }
            let componentCharacterSet = CharacterSet(charactersIn: stringComponent)
            if componentCharacterSet.isSubset(of: .decimalDigits) {
                return self = .number
            }

            if stringComponent == "pl" || stringComponent == "p" {
                return self = .patch
            }

            self = .other
        }
```

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+
+        /// Score and compare two string components
+        ///
+        func compareStringComponent(stringComponent1: String, stringComponent2: String) -> ComparisonResult {
```
jaclync: nit: just a suggestion to follow the signature of some iOS comparators ([example](https://developer.apple.com/documentation/swift/string/standardcomparator/3802578-compare))

```suggestion
        func compareStringComponent(_ lhs: String, _ rhs: String) -> ComparisonResult {
```

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+                let component1Int = Int(stringComponent1) ?? 0
+                let component2Int = Int(stringComponent2) ?? 0
+
+                if component1Int < component2Int {
+                    return .orderedAscending
+                }
+
+                if component1Int > component2Int {
+                    return .orderedDescending
+                }
```
jaclync: nit: maybe can simplify this with `NSNumber(component1Int).compare(component2Int)`?

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+
+        var rightHandString = replaceUnderscoreDashAndPlusWithDot(inString: to)
+        rightHandString = insertDotsBeforeAndAfterAnyNonNumber(inString: rightHandString)
+
+        let leftComponents = leftHandString.components(separatedBy: ".")
+        let rightComponents = rightHandString.components(separatedBy: ".")
+
+        let maxComponents = max(leftComponents.count, rightComponents.count)
+
+        for index in 0...maxComponents - 1 {
```
jaclync: nit:
```suggestion
        for index in 0..<maxComponents {
```

WooCommerce/Classes/Extensions/String+Helpers.swift
```diff
@@ -68,4 +68,150 @@ extension String {
...
+        }
+
+        var leftHandString = replaceUnderscoreDashAndPlusWithDot(inString: self)
+        leftHandString = insertDotsBeforeAndAfterAnyNonNumber(inString: leftHandString)
+
+        var rightHandString = replaceUnderscoreDashAndPlusWithDot(inString: to)
+        rightHandString = insertDotsBeforeAndAfterAnyNonNumber(inString: rightHandString)
+
+        let leftComponents = leftHandString.components(separatedBy: ".")
+        let rightComponents = rightHandString.components(separatedBy: ".")
```
jaclync: nit: the two transformations on the string could probably be done in another helper function for reuse, like:

```swift
func versionComponents(of string: String) -> [String] {
    // Call `replaceUnderscoreDashAndPlusWithDot`
    // Call `insertDotsBeforeAndAfterAnyNonNumber`
    // Return components
}
```

Another idea is to update `replaceUnderscoreDashAndPlusWithDot` and `insertDotsBeforeAndAfterAnyNonNumber` to private `String` extension functions, so that you can just chain the calls like:

```swift
func versionComponents(of string: String) -> [String] {
    string.replacingUnderscoreDashAndPlusWithDot().insertingDotsBeforeAndAfterAnyNonNumber().components(separatedBy: ".")
}
```

jaclync (commented):
Great work implementing the plugin version check! 👏 The code looks good, I mostly had some suggestions on replacing some of the `replacingOccurrences` calls with regex and moving the logic to a separate struct. Would like to hear your thoughts!

If this is required for release 8.0, I will make sure to help with the merge tomorrow!

allendav:
> If this is required for release 8.0, I will make sure to help with the merge tomorrow!

It isn't "required" for 8.0 to my knowledge (cc @ctarda )

Edit: This issue is to ruggedize our existing version comparison, which has the gaps compared to PHP version compare that you noted and may not handle patchless version numbers correctly.

That said, I'll address all this wonderful feedback now and request your re-review when you are able. Thank you!

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: DRYed the pre-processing of the strings in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/2aabed7489bd190cf2bc7facd80eae989ff8e154

I'm going to pass on the chaining suggestion, but thank you! I prefer seeing the transformations on separate lines. OK?

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/a81193104a67fd74b4aa14d2db66db85f040b764

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/998b0d2a776b4427788215fce738ac4c5108c7cc

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: Love it! Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/0a631e38f0c3b5824438e11df9e2743d3af30363

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: Not quite. If I use that and say string = "3.4.13-beta-3" then after that is applied, I'll have "3...4...13.-beta-.3"

Lemme see if there is another way to apply regex here though. The idea of replacing lines and lines with a regex is enticing.

allendav: Looks like it might be as easy as changing the regex to `([^0-9.]+)`

allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/f52e39440d09f7484744862c500beb775f34b75c

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: Works with a little tweak to the regex. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/1e7b070116a4f919006e42166e5836d305331c55

WooCommerce/Classes/Extensions/String+Helpers.swift
allendav: I agree. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/face62882caf05243acd40f340c992022ead7dbf

WooCommerce/Classes/Tools/VersionHelpers.swift
```diff
@@ -0,0 +1,135 @@
+import Foundation
+
+/// Helpers for working with versions (e.g. comparing two version strings)
+///
+final class VersionHelpers {
```
jaclync, line 5: nit: since comparing WordPress plugin versions is quite different from other versions like app versions, it might be worth specifying that this is for WordPress plugin versions?

```suggestion
final class WordPressPluginVersionHelpers {
```

WooCommerce/Classes/Tools/VersionHelpers.swift
```diff
@@ -0,0 +1,135 @@
...
+    private init() {}
```
jaclync: nit: is this init needed especially since `compare` is a static function?

WooCommerce/Classes/Tools/VersionHelpers.swift
```diff
@@ -0,0 +1,135 @@
...
+                    return comparisonResult
+                }
+            }
+
+            return .orderedSame
+        }
+
+        /// Process the given string into version components
+        ///
+        func versionComponents(of string: String) -> [String] {
```
jaclync: nit: how about moving these local functions to a private extension, so that it's easier to see the main `compare` code below? totally up to you!

WooCommerce/Classes/Tools/VersionHelpers.swift
```diff
@@ -0,0 +1,135 @@
...
+            return stringToComponentize.components(separatedBy: ".")
+        }
+
+        let leftComponents = versionComponents(of: lhs)
+        let rightComponents = versionComponents(of: rhs)
+
+        let maxComponents = max(leftComponents.count, rightComponents.count)
+
+        for index in 0..<maxComponents {
+            /// Treat missing components (e.g. 1.2 being compared to 1.1.3 as "0", i.e. 1.2.0
```
jaclync: super nit:
```suggestion
            /// Treat missing components (e.g. 1.2 being compared to 1.1.3 as "0", i.e. 1.2.0)
```

WooCommerce/WooCommerceTests/Extensions/VersionHelpersTests.swift
```diff
@@ -0,0 +1,57 @@
...
+        for test in tests {
+            let meetsMinimum = VersionHelpers.compare(test.foundVersion, test.requiredMinimumVersion) != .orderedAscending
+            XCTAssertEqual(test.meetsMinimum, meetsMinimum)
+        }
+    }
+
+    struct VersionTestCase {
+        var foundVersion: String
+        var requiredMinimumVersion: String
+        var meetsMinimum: Bool
```
jaclync: nit: these properties can probably be `let` constants!
```suggestion
        let foundVersion: String
        let requiredMinimumVersion: String
        let meetsMinimum: Bool
```

WooCommerce/WooCommerceTests/Extensions/VersionHelpersTests.swift
```diff
@@ -0,0 +1,57 @@
+import XCTest
+@testable import WooCommerce
+
+/// String+Helpers: Unit Tests
```
jaclync: super nit: maybe update the class-level comment 😄

WooCommerce/Classes/Tools/VersionHelpers.swift
```diff
@@ -0,0 +1,135 @@
...
+/// Defines the score (rank) of a component string within a version string.
+/// e.g. the "3" in 3.0.0beta3 should be treated as `.number`
+/// and the "beta" should be scored (ranked) lower as `.beta`.
+///
+/// The scores of components of version strings are used when comparing complete version strings
+/// to decide if one version is older, equal or newer than another.
+///
+/// Ranked per https://www.php.net/manual/en/function.version-compare.php
+///
+fileprivate enum VersionComponentScore: Comparable {
```
jaclync: nit: I think this could be private?

```suggestion
private enum VersionComponentScore: Comparable {
```

WooCommerce/Classes/Tools/VersionHelpers.swift
```diff
@@ -0,0 +1,135 @@
...
+    case other
+    case dev
+    case alpha
+    case beta
+    case RC
+    case number
+    case patch
+}
+
+fileprivate extension VersionComponentScore {
```
jaclync: nit: same as above, I think this could be private:

```suggestion
private extension VersionComponentScore {
```

jaclync (approved):
Nice updates!! 💟 Code & testing LGTM 🚢  

(I wasn't sure how to edit the plugin version in the header, so I hard-coded the version in `SystemPlugin`)
- Ensure unit tests pass ✅ 
- WCPay plugin 3.1.8 --> unsupported ✅ 
- WCPay plugin 3.2 --> unsupported ✅ 
- WCPay plugin 3.2.1 --> shows manage/buy/manual ✅ 
- WCPay plugin 3.3 --> shows manage/buy/manual ✅ 
- Edit `minimumSupportedWCPayVersion` in `CardPresentPaymentsOnboardingUseCase` to 3.3, restart the app, enter settings and in-person payment --> shows manage/buy/manual ✅

WooCommerce/Classes/Tools/VersionHelpers.swift
allendav: > since comparing WordPress plugin versions is quite different from other versions like app versions

In what ways? The only difference I am aware of is that we tend to avoid double digit minors and roll majors to do so.

I'd like to keep this as generic as possible.

WooCommerce/Classes/Tools/VersionHelpers.swift
allendav: No. Good catch. This was leftover from when VersionHelpers was a struct. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/875194b9134217caabb47fe048ad203828b05e75

WooCommerce/Classes/Tools/VersionHelpers.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/2eb5cb0402c9bef48cd7b74fe3d81bbe031f3c44

WooCommerce/Classes/Tools/VersionHelpers.swift
allendav: Fixed as part of https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/2eb5cb0402c9bef48cd7b74fe3d81bbe031f3c44

WooCommerce/Classes/Tools/VersionHelpers.swift
allendav: Fixed as part of https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/2eb5cb0402c9bef48cd7b74fe3d81bbe031f3c44

WooCommerce/Classes/Tools/VersionHelpers.swift
allendav: Fixed as part of https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/2eb5cb0402c9bef48cd7b74fe3d81bbe031f3c44

WooCommerce/WooCommerceTests/Extensions/VersionHelpersTests.swift
allendav: Fixed as part of https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/ab06f43d2b2e7b1b350aed4f4957422a456b9ed0

WooCommerce/WooCommerceTests/Extensions/VersionHelpersTests.swift
allendav: Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5382/commits/ab06f43d2b2e7b1b350aed4f4957422a456b9ed0

PR #5388, opened 2021-11-10
Product List: Prepare "filter products by product category" issue for release

toupper:

## Description

In the scope of this PR we prepare the already implemented issue https://github.com/woocommerce/woocommerce-ios/issues/5159 for release. For that, we remove the feature flag so the feature is exposed in any build configuration, we include the product category filter in the tracked event with analytics, and update the Release notes file (`RELEASE-NOTES.txt`)

## Changes

- Remove the feature flag
- Add the `ProductCategory` `slug` to the Analytics event information when the products list is filtered. This is implemented in the same way as Android, where also the slug is added to the filters information. See `getFilterString` in https://github.com/woocommerce/woocommerce-android/blob/da7e069f12f46d18280327a1be72d39981d9d198/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/products/ProductFilterListViewModel.kt , that uses the `productFilterOptions` map property, being the values of that map _the slug associated with the property_.  This `getFilterString` function is called from` ProductFilterListFragment.kt` and `ProductFilterOptionListFragment.kt` to track the tap of the show products button.
- Update unit tests to remove the feature flag
- Update release notes

### Feature flag removal
The functionality should behave in the `appStore` `BuildConfiguration` in the same way as it was already tested in `localDeveloper` or `alpha`:
- Open app
- Go to Products
- Press the Filter button
- A menu filter entry with the title _Product Category_ should be shown

- If there was a Product Category filter persisted, it should be shown when opening there the menu:

- If the button with title "Clear all" is pressed, it should clear also the Product Category filter:

The rest of the issue implementation remains unaffected by the removal of the feature flag.

### Analytics tracking
In order to check that the event is tracked with the right information when showing products with filters:
- Open app
- Go to Products
- Press the Filter button
- Go to Product Category filter
- Select a category
- Press the Show Products button

Product category should be included in the event filters information. This can be checked in the Xcode console:

`WooCommerce[50427:8085398] 🔵 Tracked product_filter_list_show_products_button_tapped, properties: [AnyHashable("is_wpcom_store"): false, AnyHashable("filters"): "clothing", AnyHashable("blog_id"): 171745248]`

You can also check that it works fine when other filter is selected:

`WooCommerce[50427:8085398] 🔵 Tracked product_filter_list_show_products_button_tapped, properties: [AnyHashable("is_wpcom_store"): false, AnyHashable("filters"): "variable,clothing", AnyHashable("blog_id"): 171745248]`

### Release notes
Please check the `RELEASE-NOTES.txt` file changes 

Update release notes:

### DIFF
--- a/Experiments/Experiments/DefaultFeatureFlagService.swift
+++ b/Experiments/Experiments/DefaultFeatureFlagService.swift
@@ -25,8 +25,6 @@ public struct DefaultFeatureFlagService: FeatureFlagService {
             return buildConfig == .localDeveloper || buildConfig == .alpha
         case .orderListFilters:
             return buildConfig == .localDeveloper || buildConfig == .alpha
-        case .filterProductsByCategory:
-            return buildConfig == .localDeveloper || buildConfig == .alpha
         case .jetpackConnectionPackageSupport:
             return buildConfig == .localDeveloper || buildConfig == .alpha
         default:
--- a/Experiments/Experiments/FeatureFlag.swift
+++ b/Experiments/Experiments/FeatureFlag.swift
@@ -50,10 +50,6 @@ public enum FeatureFlag: Int {
     ///
     case orderListFilters
 
-    /// Allows to filter products by a product category, persisting it so the filter can remain after restarting the app
-    ///
-    case filterProductsByCategory
-
     /// Allows sites with plugins that include Jetpack Connection Package and without Jetpack-the-plugin to connect to the app
     ///
     case jetpackConnectionPackageSupport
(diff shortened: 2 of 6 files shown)


pmusolino (approved):
Looks great to me 🚀 @toupper this will be your first public contribution on WCiOS that will be visible to our merchants, congrats 🥳 

Just a nit for the future: it is not necessary to remove the code of the feature flag at this stage, it was enough to always return `true` as if a problem should occur it can be immediately disabled. However, the removed code is not very much, so it could easily be reintroduced if there were any problems in the release phase. 💯

pmusolino:
When the CI will be ✅ this PR will be automatically merged on `develop`.

Issue #5300, opened 2021-10-28
[Mobile Payments] Card reader update Tracks events


joshheald:
Android have implemented the following Tracks events for reporting on card reader updates. If our merchants are struggling with updates, it's important we know about that ASAP to deal with it if we can. Additionally, if people are leaving required updates to the last minute, that would be useful information to have.

- `card_reader_software_update_tapped`
- `card_reader_software_update_started`
- `card_reader_software_update_success`
- `card_reader_software_update_failed`
- `card_reader_software_update_cancel_tapped`
- `card_reader_software_update_canceled`

p1635419032116000-slack-C025A8VV728

Additionally, mirror [the properties Android use on these events](https://github.com/woocommerce/woocommerce-android/blob/48200d95ac882e8cacf507d548bda7b94a951839/WooCommerce/src/main/kotlin/com/woocommerce/android/ui/prefs/cardreader/update/CardReaderUpdateViewModel.kt#L162):

- `software_update_type`: `Required` or `Optional`
- `error_description`: error description

joshheald:
#5366 added the events, work still to do is to add the properties:

> Additionally, mirror the properties Android use on these events:
> software_update_type: Required or Optional
> error_description: error description

PR #5389, opened 2021-11-10
[Analytics] Add Track function combining errors and props

joshheald:
Supports #5300 

## Description
For Tracks events around card reader software update events, we need to log events with both errors and other properties. To do that, it's convenient to use a track function which accepts both as parameters. This PR does not add any direct use of the combined function, but it is well exercised.

### WooAnalytics changes
Logging events with both properties, and standard error details, required a new `track(_:properties:withError:)` function on WooAnalytics. It's worth calling out here that _all_ tracks events will now go through this function so _all the scrutiny_ is welcomed!

## Testing
Test that events are logged all the way back to Tracks.

Update release notes:

### DIFF
--- a/WooCommerce/Classes/Analytics/WooAnalytics.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalytics.swift
@@ -80,7 +80,7 @@ public extension WooAnalytics {
     ///   - properties: a collection of properties related to the event
     ///
     func track(_ stat: WooAnalyticsStat, withProperties properties: [AnyHashable: Any]?) {
-        track(stat, properties: properties, withError: nil)
+        track(stat, properties: properties, error: nil)
     }
 
     /// Track a specific event with an associated error (that is translated to properties)
@@ -90,7 +90,7 @@ public extension WooAnalytics {
     ///   - error: the error to track
     ///
     func track(_ stat: WooAnalyticsStat, withError error: Error) {
-        track(stat, properties: nil, withError: error)
+        track(stat, properties: nil, error: error)
     }
 
     /// Track a specific event with associated properties and an associated error (that is translated to properties)
@@ -100,7 +100,7 @@ public extension WooAnalytics {
     ///   - properties: a collection of properties related to the event
     ///   - error: the error to track
     ///
-    func track(_ stat: WooAnalyticsStat, properties passedProperties: [AnyHashable: Any]?, withError error: Error?) {
+    func track(_ stat: WooAnalyticsStat, properties passedProperties: [AnyHashable: Any]?, error: Error?) {
         guard userHasOptedIn == true else {
             return
         }
--- a/WooCommerce/Classes/Analytics/WooAnalytics.swift
+++ b/WooCommerce/Classes/Analytics/WooAnalytics.swift
@@ -80,15 +80,7 @@ public extension WooAnalytics {
     ///   - properties: a collection of properties related to the event
     ///
     func track(_ stat: WooAnalyticsStat, withProperties properties: [AnyHashable: Any]?) {
-        guard userHasOptedIn == true else {
-            return
-        }
-
-        if let updatedProperties = updatePropertiesIfNeeded(for: stat, properties: properties) {
-            analyticsProvider.track(stat.rawValue, withProperties: updatedProperties)
-        } else {
-            analyticsProvider.track(stat.rawValue)
-        }
+        track(stat, properties: properties, withError: nil)
     }
 
     /// Track a specific event with an associated error (that is translated to properties)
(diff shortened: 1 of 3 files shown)


WooCommerce/Classes/ServiceLocator/Analytics.swift
```diff
@@ -29,6 +29,15 @@ protocol Analytics {
...
     func track(_ stat: WooAnalyticsStat, withError error: Error)
 
+    /// Track a specific event with associated properties and an associated error (that is translated to properties)
+    ///
+    /// - Parameters:
+    ///   - stat: the event name
+    ///   - properties: a collection of properties related to the event
+    ///   - error: the error to track
+    ///
+    func track(_ stat: WooAnalyticsStat, properties: [AnyHashable: Any]?, withError error: Error?)
```
jaclync: nit: just a suggestion
```suggestion
    func track(_ stat: WooAnalyticsStat, properties: [AnyHashable: Any]?, error: Error?)
```

WooCommerce/Classes/Analytics/WooAnalytics.swift
```diff
@@ -98,16 +90,52 @@ public extension WooAnalytics {
...
+        let errorProperties = errorProperties(from: error)
+
+        if let passedProperties = passedProperties {
+            properties = passedProperties.merging(errorProperties ?? [:], uniquingKeysWith: { current, _ in
+                current
+            })
+        } else {
+            properties = errorProperties
+        }
+        return properties
```
jaclync, line 128: nit: non-blocking suggestion

```suggestion
        let passedProperties: [AnyHashable: Any] = passedProperties ?? [:]
        let errorProperties = errorProperties(from: error) ?? [:]
        return passedProperties.merging(errorProperties, uniquingKeysWith: { current, _ in
            current
        })
```

WooCommerce/WooCommerceTests/System/WooAnalyticsTests.swift
```diff
@@ -90,6 +90,36 @@ class WooAnalyticsTests: XCTestCase {
         }
     }
 
+    /// Verifies an event with an error and properties is received by the AnalyticsProvider
+    ///
+    func testEventsWithPropertiesAndErrorReceived() {
```
jaclync: nit: we use snake case for test case names now based on [this documentation](https://github.com/woocommerce/woocommerce-ios/blob/c5d83703d6c0565fc34bf410b6b27483f2318df2/docs/naming-conventions.md#test-methods) 😄 
```suggestion
    func test_events_with_properties_and_error_include_combined_properties() {
```

WooCommerce/WooCommerceTests/System/WooAnalyticsTests.swift
```diff
@@ -90,6 +90,36 @@ class WooAnalyticsTests: XCTestCase {
...
+        let testError = NSError(domain: Constants.testErrorDomain, code: Constants.testErrorCode, userInfo: Constants.testErrorUserInfo)
+        analytics.track(.applicationOpened, properties: Constants.testProperty1, withError: testError)
+        XCTAssertEqual(testingProvider?.receivedEvents.count, 1)
+        XCTAssertEqual(testingProvider?.receivedProperties.count, 1)
+        XCTAssertEqual(testingProvider?.receivedEvents.first, WooAnalyticsStat.applicationOpened.rawValue)
```
jaclync: nit: we generally follow Given/When/Then in test cases so that it's more clear to differentiate code for each step:

```suggestion
        // Given
        let testError = NSError(domain: Constants.testErrorDomain, code: Constants.testErrorCode, userInfo: Constants.testErrorUserInfo)
        
        // When
        analytics.track(.applicationOpened, properties: Constants.testProperty1, withError: testError)
        
        // Then
        XCTAssertEqual(testingProvider?.receivedEvents.count, 1)
        XCTAssertEqual(testingProvider?.receivedProperties.count, 1)
        XCTAssertEqual(testingProvider?.receivedEvents.first, WooAnalyticsStat.applicationOpened.rawValue)
```

WooCommerce/WooCommerceTests/System/WooAnalyticsTests.swift
```diff
@@ -90,6 +90,36 @@ class WooAnalyticsTests: XCTestCase {
...
+
+        /// Note: iOS 12 is shuffling several dictionaries (especially when it comes to serializing [:] > URL Parameters).
+        /// For that reason, we'll proceed with a bit of a more lengthy but robust check.
+        ///
+        for (key, value) in Constants.testErrorAndPropertyReceivedProperty {
+            XCTAssertEqual(value, receivedProperty1[key])
+        }
+
+        /// Second note: the error's userInfo, as a string, is getting swizzled. We'll ensure the expected payload is there,
+        /// but the exact position isn't guarranteed!
```
jaclync: nit:
```suggestion
        /// but the exact position isn't guaranteed!
```

jaclync (approved):
The code looks good to me! I didn't get the time to test it today, I can test it tomorrow if you'd like another confirmation 🙂

WooCommerce/Classes/Analytics/WooAnalytics.swift
joshheald: Unfortunately, nil-coalescing an empty dictionary for `passedProperties` will result in a change in behaviour, which causes [tests to fail](https://github.com/woocommerce/woocommerce-ios/blob/4fb64991809c42bdd99e5edad136062d0cb8ea8d/WooCommerce/WooCommerceTests/System/WooAnalyticsTests.swift#L28) by passing an empty dictionary where previously it would have returned nil. This [affects four events](https://github.com/woocommerce/woocommerce-ios/blob/4fb64991809c42bdd99e5edad136062d0cb8ea8d/WooCommerce/Classes/Analytics/WooAnalyticsStat.swift#L545).

I approached it the same as you have here initially, but wanted to preserve the existing behaviour of the two existing functions while I combined the code into a single function.

I think it would be good to change to this more readable approach, and since we're passing an empty dictionary around it shouldn't have any impact on what gets sent to Tracks, so I'll do that in a subsequent PR.

WooCommerce/Classes/ServiceLocator/Analytics.swift
joshheald: Yeah, agreed. I'd included `withError` to be consistent with the existing methods but just because the ones that are already there are too wordy, doesn't mean these need to be.

WooCommerce/WooCommerceTests/System/WooAnalyticsTests.swift
joshheald: Thanks. Again I was going for consistency, and again I think that was the wrong call.

joshheald (commented):
Thanks for the feedback and taking the time to add suggestions @jaclync

PR #5390, opened 2021-11-10
[Mobile Payments] Card reader software update Tracks properties

joshheald:
Closes #5300 

## Description
This covers the [remaining software update Tracks work](https://github.com/woocommerce/woocommerce-ios/issues/5300#issuecomment-964052643), i.e. setting the properties for `software_update_type` and those containing error details.

### Relevant events
- `card_reader_software_update_tapped`
- `card_reader_software_update_started`
- `card_reader_software_update_success`
- `card_reader_software_update_failed`
- `card_reader_software_update_cancel_tapped`
- `card_reader_software_update_canceled`

## Testing
- [x] Check that events are logged with `software_update_type: Optional` for optional updates
- [x] Check that events are logged with `software_update_type: Required` for required updates (`cancel` and `update_tapped` events not possible)
- [x] Check that `card_reader_software_update_failed` is logged with properties containing the error details, `error_description`, `error_code`, and `error_domain` in particular.

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #5391, opened 2021-11-10
[Mobile Payments] Maintain reader software update Required/Optional state correctly

joshheald:
Part of #5300 

## Description
In #5300, we want to Track a property of `software_update_type` with `Required` or `Optional` based on whether the a card reader software update is mandatory or not. To do this, we need to maintain this state accurately, through the flow of updating a reader.

### What's a required update?!
Required updates are communicated by Stripe as updates which start as soon as a reader is connected, without a notification from the Terminal SDK that there's an update available. We were tracking this using the property `CardReaderSettingsConnectedViewModel.readerUpdateAvailable`, however this wasn't clear from the original naming, so I've clarified that naming in 2e4e86c737b0f1d6444c98c2fac7916df612e11b.

Beyond that, I found a bug in our logic which meant that a re-run of a `Required` update which had failed (e.g. for low battery) was marked as `Optional`. This is fixed in f624724c15ead1d192fb878cd36c86611581684e, by checking the required date is in the future before reporting that an (optional) update is available in the failure handler.

## Testing
This is difficult to test in isolation, and might be best tested with #5390. However, it's worth running through some updates, failures, cancellations, and re-runs to check that the flows are unchanged by this work. The bug I found didn't actually manifest in the app in any way, so there are no repro steps to check.

See PdfdoF-fr-p2 for details on simulating different update scenarios.

Update release notes:

### DIFF
--- a/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
+++ b/Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
@@ -503,7 +503,12 @@ extension StripeCardReaderService: BluetoothReaderDelegate {
                 error: CardReaderServiceError.softwareUpdate(underlyingError: UnderlyingError(with: error),
                                                              batteryLevel: reader.batteryLevel?.doubleValue))
             )
-            softwareUpdateSubject.send(.available)
+            if let requiredDate = update?.requiredAt,
+               requiredDate > Date() {
+                softwareUpdateSubject.send(.available)
+            } else {
+                softwareUpdateSubject.send(.none)
+            }
         } else {
             softwareUpdateSubject.send(.completed)
             softwareUpdateSubject.send(.none)
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -11,7 +11,7 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     private var connectedReaders = [CardReader]()
     private let knownReaderProvider: CardReaderSettingsKnownReaderProvider?
 
-    private(set) var readerUpdateAvailable: Bool = false
+    private(set) var optionalReaderUpdateAvailable: Bool = false
     var readerUpdateInProgress: Bool {
         readerUpdateProgress != nil
     }
@@ -81,9 +81,9 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
                             self?.completeCardReaderUpdate(success: true)
                         }
                     case .available:
-                        self.readerUpdateAvailable = true
+                        self.optionalReaderUpdateAvailable = true
                     case .none:
-                        self.readerUpdateAvailable = false
+                        self.optionalReaderUpdateAvailable = false
                     }
                     self.didUpdate?()
                 }
@@ -148,7 +148,7 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     }
 
     private func completeCardReaderUpdate(success: Bool) {
-        readerUpdateAvailable = !success
+        optionalReaderUpdateAvailable = !success
         readerUpdateProgress = nil
         didUpdate?()
     }
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -27,6 +27,8 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     var connectedReaderBatteryLevel: String?
     var connectedReaderSoftwareVersion: String?
 
+    var delayToShowUpdateSuccessMessage: DispatchTimeInterval = .seconds(1)
+
     init(didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?, knownReaderProvider: CardReaderSettingsKnownReaderProvider? = nil) {
         self.didChangeShouldShow = didChangeShouldShow
         self.knownReaderProvider = knownReaderProvider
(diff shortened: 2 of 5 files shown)


Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
```diff
@@ -503,7 +503,12 @@ extension StripeCardReaderService: BluetoothReaderDelegate {
                 error: CardReaderServiceError.softwareUpdate(underlyingError: UnderlyingError(with: error),
                                                              batteryLevel: reader.batteryLevel?.doubleValue))
             )
-            softwareUpdateSubject.send(.available)
+            if let requiredDate = update?.requiredAt,
+               requiredDate > Date() {
+                softwareUpdateSubject.send(.available)
+            } else {
+                softwareUpdateSubject.send(.none)
```
jaclync, line 510: ❓ I am new to the Payments code, is this how we know a software update is required or optional - by comparing `ReaderSoftwareUpdate`'s `requiredAt` date to the current date? Why not including whether an update is required or optional in `softwareUpdateSubject`'s value like `.available(UpdateTypeEnum)` where the enum has two cases, required and optional?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -11,7 +11,7 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     private var connectedReaders = [CardReader]()
     private let knownReaderProvider: CardReaderSettingsKnownReaderProvider?
 
-    private(set) var readerUpdateAvailable: Bool = false
+    private(set) var optionalReaderUpdateAvailable: Bool = false
```
jaclync, line 14: Just to clarify, this variable indicates whether an optional update is available. Is there a way to know whether a required update is available?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
```diff
@@ -27,6 +27,8 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
     var connectedReaderBatteryLevel: String?
     var connectedReaderSoftwareVersion: String?
 
+    var delayToShowUpdateSuccessMessage: DispatchTimeInterval = .seconds(1)
```
jaclync: How about making this a private variable by injecting this as a parameter in the initializer like:

```swift
init(didChangeShouldShow: ((CardReaderSettingsTriState) -> Void)?, knownReaderProvider: CardReaderSettingsKnownReaderProvider? = nil, delayToShowUpdateSuccessMessage: DispatchTimeInterval = .seconds(1))
```

and then passing the value in unit tests? In the codebase, we generally do dependency injection with initializers for parameters that are different in unit tests

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/CardReaderSettings/CardReaderSettingsConnectedViewModelTests.swift
```diff
@@ -232,4 +233,66 @@ final class CardReaderSettingsConnectedViewModelTests: XCTestCase {
         wait(for: [expectation], timeout: Constants.expectationTimeout)
         XCTAssertFalse(analytics.receivedEvents.contains(WooAnalyticsStat.cardReaderSoftwareUpdateFailed.rawValue))
     }
+
+    func test_when_a_mandatory_update_succeeds_optional_updates_are_not_available() {
+        // Given
+        // .available is not sent
+        mockStoresManager.simulateUpdateStarted()
+        let expectation = self.expectation(description: #function)
```
jaclync, line 247: nit: we also have [a helper](https://github.com/woocommerce/woocommerce-ios/blob/dbea5342d2b1604904d109d6feb860b799b685d1/TestKit/Sources/TestKit/XCTestCase+Wait.swift#L72-L75) to wait for an expectation with a promise, you can find some examples in the codebase with `waitFor {` in the search. The helper makes it look a bit cleaner IMO 😄

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
ctarda: Does this need to be a var? It seems like it could be declared as a constant?

jaclync (commented):
Sorry @joshheald I didn't feel like I have the full knowledge to review this change! I left some suggestions and questions. If you'd like this one merged for release 8.0, is it possible for someone with more context to review it first?

It'd also help if there is a more detailed testing plan, like which steps to take since I'm new to the feature. This could also help with understanding the problem a bit more 🙂

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/CardReaderSettings/CardReaderSettingsConnectedViewModelTests.swift
```diff
@@ -232,4 +233,66 @@ final class CardReaderSettingsConnectedViewModelTests: XCTestCase {
         wait(for: [expectation], timeout: Constants.expectationTimeout)
         XCTAssertFalse(analytics.receivedEvents.contains(WooAnalyticsStat.cardReaderSoftwareUpdateFailed.rawValue))
     }
+
+    func test_when_a_mandatory_update_succeeds_optional_updates_are_not_available() {
```
ctarda, line 243: Nice!

ctarda (approved):
I didn't notice anything breaking. I run a couple of quick tests overriding what we get from the SDK (overriding ReaderSoftwareUpdate) and everything seems to test well.

:shipit:

Hardware/Hardware/CardReader/StripeCardReader/StripeCardReaderService.swift
joshheald: The canonical way is whether or not `.available` is sent by the Terminal SDK first, or if it just sends `started` and kicks the update off on connecting the reader. However, that's not useful after the required update has failed, so using the date is the next best thing, IMO

WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
joshheald: Not really, because they just start automatically.

WooCommerce/WooCommerceTests/ViewRelated/Dashboard/CardReaderSettings/CardReaderSettingsConnectedViewModelTests.swift
joshheald: Thanks for the heads up on this!

joshheald:
@jaclync @ctarda thanks for the reviews

PR #5393, opened 2021-11-10
Migrate `Settings` screen to MVVM

### LINKED ISSUE #4142: Migrate SettingsViewController to use MVVM for testability
## Problem
While working on the Settings screen for issue #3921, I had to add a feature flag for the Plugins section as this feature is not yet ready for public users. Then I wanted to add tests to make sure that my feature flag check works right for different build environments, but the `SettingsViewController` is now still using MVC and there's virtually no way to add tests to it.

## Suggested solution
To make the view controller testable, I'd propose migration to the MVVM pattern. The setup is simple:
- The view model takes `FeatureFlagService` as a dependency.
- The view model takes care of the table view contents (rows, sections) and exposes these so that the view controller only needs to display data.
- Unit tests can use a mock feature flag service and check if the displayed rows and sections are as expected.

Let's add tests for more code confidence!

selanthiraiyan:
Closes #4142 

#4142 proposes to migrate settings screen to MVVM pattern to make it testable. 

This pull request attempts to migrate it to MVVM and add basic tests.
SettingsScreen

## Changes
- Introduce `SettingsViewModel` and `SettingsFactory` and make settings screen follow MVVM architecture.
- Add basic unit tests using `SettingsViewModelTests`
- Fix typos in comments.

## Testing
- Launch the app and navigate to `My Store` screen.
- Navigate to Settings screen by tapping on the top right `gear` button.
- Observe that you can view all settings related to the account.
- Please navigate to child screens such as `Plugins`, `In-Person Payments` and make sure that the navigation works as expected.
- Switching the store should reload the sites in settings screen.

## Notes
- Being an outside contributor, I don't have a payment method set up in my store. I request you to test the child screens including `In-Person Payments` using your store. 

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/Settings/SettingsFactory.swift
```diff
@@ -0,0 +1,13 @@
...
+    /// Creates a Settings view controller
+    ///
+    static func settings() -> UIViewController {
+        let viewModel = SettingsViewModel(stores: ServiceLocator.stores,
+                                          storageManager: ServiceLocator.storageManager)
+        let settingsViewController = SettingsViewController(viewModel: viewModel)
+        viewModel.presenter = settingsViewController
+        return settingsViewController
+    }
+}
```
itsmeichigo: I'm just curious: why do we need a factory for Settings? Would it be simpler if we bring this setup to `DashboardViewController` instead?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Settings/SettingsViewModel.swift
```diff
@@ -0,0 +1,350 @@
+import UIKit
+import Combine
```
itsmeichigo: Nit: This import seems redundant, perhaps it can be removed?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Settings/SettingsViewModel.swift
```diff
@@ -0,0 +1,350 @@
...
+
+final class SettingsViewModel: SettingsViewModelOutput, SettingsViewModelActionsHandler, SettingsViewModelInput {
+
+    typealias Row = SettingsViewController.Row
+
+    weak var presenter: SettingsViewPresenter?
+
+    /// Table Sections to be rendered
+    ///
+    var sections: [Section] = []
```
itsmeichigo: Nit: I know this variable is defined as read-only on `SettingsViewModelOutput`, but how about adding `private(set)` for extra safety?

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Settings/SettingsViewModel.swift
```diff
@@ -0,0 +1,350 @@
...
+
+    /// Main Account's displayName
+    ///
+    var accountName: String {
+        stores.sessionManager.defaultAccount?.displayName ?? String()
+    }
+
+    /// Announcement for the current app version
+    ///
+    var announcement: Yosemite.Announcement?
```
itsmeichigo: Nit: Same comment as above about adding `private(set)`

itsmeichigo (commented):
Thank you @selanthiraiyan for the contribution!
Everything works great 🎉 I have a few comments and nits though, please check it out.

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Settings/SettingsViewModel.swift
selanthiraiyan: Nice catch. Fixed in https://github.com/woocommerce/woocommerce-ios/pull/5393/commits/e51af98e8c90b9196d1f38140f5e0b9c78fc4d5d

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Settings/SettingsViewModel.swift
selanthiraiyan: Great point. Changed in https://github.com/woocommerce/woocommerce-ios/pull/5393/commits/ea1f0951950a17c3bf8097032ce6ba13faf37342

WooCommerce/Classes/ViewRelated/Dashboard/Settings/Settings/SettingsFactory.swift
selanthiraiyan: Thanks a lot for reviewing @itsmeichigo 

My thought process was to hide the initialization (DI) and setup (setting presenter for `ViewModel`) details inside a factory. I am hoping that we will have to inject more types such as FeatureFlag and etc in future.

What do you think?

itsmeichigo: Thanks for the explanation. My understanding of the factory pattern is that we need it to create different kinds of instances that conform a shared protocol. I'm concerned that using this here may be misleading, since we're not having different kinds of Settings in the app. 

For DI we can set default values for `stores` and `storageManager` to simplify the SettingsViewModel's initilizer. We can also set up the presenter for the view model when initilizing the view controller. That way we can tidy up the setup - how does that sound?

selanthiraiyan: You are right @itsmeichigo. I made the necessary changes in https://github.com/woocommerce/woocommerce-ios/pull/5393/commits/7d971ce68b09caafe40853c3477e6e545f210635. 

Kindly take a look at `SettingsViewController` 's init method and share your thoughts.

I am not comfortable about having to change the `viewModel` property into `var` to be able to assign the `presenter` onto it from the view controller’s init method. (Even though `viewModel` is a private property)
```swift
private var viewModel: ViewModel
init(viewModel: ViewModel = SettingsViewModel()) {
    self.viewModel = viewModel
    super.init(nibName: nil, bundle: nil)
    self.viewModel.presenter = self
}
```
What do you think about this?

itsmeichigo: Good point! I think we can keep the view model a constant and still mutable by making `SettingsViewModelInput` conform to `AnyObject`. That'll work right?

selanthiraiyan: Great, that will work. Done in https://github.com/woocommerce/woocommerce-ios/pull/5393/commits/0cf6a3895b02baf302ba2d4a14b8ee03f9c3801b.
Thank you for your inputs! 🙇‍♂️

itsmeichigo (approved):
Everything looks great now 🎉

PR #5394, opened 2021-11-10
Products: Added the variation number to the variations list, details page and updated progress view text when saving a variation

### LINKED ISSUE #4846: Products: add the variation number to the variations list and details
The variation number should be shown in the variations list and details screens.

Here's what I'm seeing in 7.3.0.0 (Testflight beta):

IMG_4951  IMG_4952

Here's what the designs show:

screen-shot-2021-08-11-at-11 06 40-pm

### LINKED ISSUE #4847: Products: when saving a variation it should say "Saving your variation" not "Saving your product"
> Is it possible to update the text so it reads “Saving your variation” / “Please wait while we save your latest changes”. We know users get confused about watching their variations page vs their product page. So clarifying anywhere we can that they are looking at the Variations page when they are is a big help.

image

anas-p:
Closes #4846
Closes #4847

## Why
#4846 
 Variation number was not displayed in the Variation listing and details page.

#4847
Both Products and Variation details pages use the same controller `ProductFormViewController`,  So the message shown while saving variation changes was the same as that of product change. 

In this PR I added variation number in the variation listing and details page, also updated the progress view title and message when saving product variation.

## How
#4846
To add product variation number in variation listing page:
- Added `productVariation` of type `ProductVariation` in `ProductsTabProductViewModel` to get the variation ID (same as variation number)
- Above added `productVariation.productVariationID` is concatenated with cell name label to show variation number and name in variation listing
- 'nameLabel.numberOfLines' changed to 0 to avoid name clipping.
- Localization added for variation ID

To add product variation number in variation details page:
- Added `productionVariationID` in view model (`ProductFormViewModelProtocol`)
- Set navigation bar title in `updateNavigationBarTitle`
- If `viewModel.productionVariationID` is not `nil`, `updateNavigationBarTitle` function will set variation title.
- `viewModel.productionVariationID` becomes nil for production details page.

#4847
To update the texts in progress view page when saving a variation
- Added `saveVariation`, a new type in enum `SaveMessageType` to handle message type while showing progress view.
- Set messageType to `saveVariation` if the `viewModel` is `ProductVariationFormViewModel`.
- Based on the message type saving progress view will show a proper message for variation and product.

## Demo

#4648
https://user-images.githubusercontent.com/19687299/141671030-21d122a9-8d93-47a6-8eeb-cc2555e31e85.mov

#4847
https://user-images.githubusercontent.com/19687299/141671041-a8029f03-e14a-4aa5-852b-bc16b2f9d0fa.mov

## Testing Steps
#4846
- Open the Products tab
- Tap on any product with variations
- Tap on `Variations` option to show variation listing with variation number
- Tap on any variation item from variation listing to open the details page

#4847
- Open the Products tab
- Tap on any product with variations
- Tap on `Variations` option to open variation listing page
- Tap on any variation item from variation listing to open the details page
- Tap on `Description` or update any value
- Tap on `Save` button which appears on navigation bar

Update release notes:


WooCommerce.xcworkspace/contents.xcworkspacedata
```diff
@@ -23,13 +23,13 @@
...
    <FileRef
-      location = "group:Observables">
+      location = "group:CodeGeneration">
    </FileRef>
    <FileRef
       location = "group:TestKit">
    </FileRef>
    <FileRef
-      location = "group:CodeGeneration">
+      location = "group:Observables">
```
Ecarrion: Was this change intentional?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -410,6 +413,9 @@ private extension ProductFormViewController {
     func configureNavigationBar() {
         updateNavigationBar()
         updateBackButtonTitle()
+        if titleString.isNotEmpty {
```
Ecarrion: Is this condition really needed?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -58,10 +59,12 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
          eventLogger: ProductFormEventLoggerProtocol,
          productImageActionHandler: ProductImageActionHandler,
          currency: String = ServiceLocator.currencySettings.symbol(from: ServiceLocator.currencySettings.currencyCode),
+         titleString: String = String(),
```
Ecarrion:
```suggestion
         titleString: String = "",
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -39,6 +39,7 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
     private let productUIImageLoader: ProductUIImageLoader
 
     private let currency: String
+    private let titleString: String // Navigation title
```
Ecarrion: We tend to add documentation like:

```suggestion
    /// Navigation title
    ///
    private let titleString: String 
```

That way is picked up by Xcode and displayed on the help menu

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
```diff
@@ -423,10 +423,17 @@ extension ProductVariationsViewController: UITableViewDelegate {
             let updatedProduct = self.product.copy(variations: variationsUpdated)
             self.product = updatedProduct
         }
+
+        // Passing title to set as ProductFormViewController navigation title
+        let titleFormat = NSLocalizedString("Variation #%1$@",
+                                            comment: "Variation title. Parameters: %1$@ - Product variation ID")
+        let titleString = String.localizedStringWithFormat(titleFormat, "\(viewModel.productModel.productVariation.productVariationID)")
```
Ecarrion: We try to add localization code in enums at the bottom of the screen.
https://github.com/woocommerce/woocommerce-ios/blob/65be53923e8406339a5036b0b75644c94adc804d/WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift#L709-L730

Also, this could be a function so you can group those two formats. Something like

Eg:
```swift
static func title(variationID: String) {
    let titleFormat = NSLocalizedString("Variation #%1$@", comment: "Variation title. Parameters: %1$@ - Product variation ID")
    return let titleString = String.localizedStringWithFormat(titleFormat, variationID)
}
```

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce.xcscheme
```diff
@@ -102,6 +102,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      language = "id"
```
Ecarrion: I think you are changing the simulator device for everyone 😅 
Could you revert this please?

WooCommerce/Resources/en.lproj/Localizable.strings
```diff
@@ -1,4 +1,4 @@
-﻿/* In Order List, the pattern to show the order number. For example, “#123456”. The %@ placeholder is the order number. */
+/* In Order List, the pattern to show the order number. For example, “#123456”. The %@ placeholder is the order number. */
```
Ecarrion: I don't see where the difference is? Can we make sure to clear it?

Ecarrion (changes requested):
Congrats on submitting your first PR @anas-p! 
I left you some small suggestions, let me know what you think.

Also, I spotted a regression where variation names are be clipped, can we make sure this does not happen, please?

Expected | Current 
---- | ----
Screen Shot 2021-11-11 at 10 06 00 AM | Screen Shot 2021-11-11 at 10 06 00 AM

RELEASE-NOTES.txt
```diff
@@ -4,6 +4,8 @@
 -----
 - [***] Push notifications are now supported for all connected stores. [https://github.com/woocommerce/woocommerce-ios/pull/5299]
 - [*] Fix: in Settings > Switch Store, tapping "Dismiss" after selecting a different store does not switch stores anymore. [https://github.com/woocommerce/woocommerce-ios/pull/5359]
+- [*] Products: Variation number added in variation list and details page
+    [https://github.com/woocommerce/woocommerce-ios/issues/4846]
```
selanthiraiyan, line 8: Hello @anas-p 👋 
Please replace this URL with the PR's URL.

selanthiraiyan: Hello @anas-p 👋 
Please replace this URL with this pull-request's URL.

anas-p: Hi @selanthiraiyan I will change this and send a new PR with other changes.

WooCommerce.xcworkspace/contents.xcworkspacedata
anas-p: Not intentional.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
anas-p: This condition is needed here, but I have a better solution now with ViewModel. I will send this with the next PR.

I have changed it to this:
```
/// Configure navigation bar with the title
///
func configureNavigationBar(title: String = String()) {
        updateNavigationBar()
        updateBackButtonTitle()
        updateNavigationBarTitle(title: title)
 }
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
anas-p: I will change it.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
anas-p: I will remember this, thanks.

WooCommerce/Classes/ViewRelated/Products/Variations/ProductVariationsViewController.swift
anas-p: I will make the necessary changes.

WooCommerce/WooCommerce.xcodeproj/xcshareddata/xcschemes/WooCommerce.xcscheme
anas-p: Sorry, this was not intentional, I will revert this.

WooCommerce/Resources/en.lproj/Localizable.strings
anas-p: I tried to do add localization strings myself, but later I came to know that it will be added automatically.

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
```diff
@@ -64,7 +64,14 @@ extension ProductsTabProductTableViewCell: SearchResultCell {
...
     func update(viewModel: ProductsTabProductViewModel, imageService: ImageService) {
-        nameLabel.text = viewModel.name
+        if let variationID = viewModel.productVariation?.productVariationID {
+            // Add product variation ID with name
+            // See more: https://github.com/woocommerce/woocommerce-ios/issues/4846
+            nameLabel.text = "\(Localization.variationID(variationID: "\(variationID)"))\n\(viewModel.name)"
+        }
+        else {
+            nameLabel.text = viewModel.name
+        }
```
Ecarrion: What do you think if instead of having this logic in the view/cell, we move this to the view model, Ideally the view should not be doing calculations and should only be accessing `viewModel.name`. Also, that allows us to unit test that logic easily. Thoughts?

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
```diff
@@ -155,7 +162,7 @@ private extension ProductsTabProductTableViewCell {
 
     func configureNameLabel() {
         nameLabel.applyBodyStyle()
-        nameLabel.numberOfLines = 2
+        nameLabel.numberOfLines = 3
```
Ecarrion: Maybe this should be just `0`, what do you think?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -407,9 +416,12 @@ final class ProductFormViewController<ViewModel: ProductFormViewModelProtocol>:
 //
 private extension ProductFormViewController {
 
-    func configureNavigationBar() {
+    /// Configure navigation bar with the title
+    ///
+    func configureNavigationBar(title: String = String()) {
```
Ecarrion:
```suggestion
    func configureNavigationBar(title: String = "") {
```

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -605,7 +617,15 @@ private extension ProductFormViewController {
...
     func saveProduct(status: ProductStatus? = nil) {
-        let productStatus = status ?? product.status
+        var productStatus = status ?? product.status
+
+        // Set productStatus to custom type for saving a variation
+        // This will change Progress view title and message for variation save action.
+        // Fix for the issue: https://github.com/woocommerce/woocommerce-ios/issues/4847
+        if let _ = self.viewModel.productionVariationID {
+            productStatus = .custom("variation")
+        }
```
Ecarrion: ⚠️ I'm concerned about this, I don't think we should be changing the product status just to be able to detect if we are saving a variation or not.

You can see how in this class we detect if we are dealing with a variation with something like 
```swift
viewModel is ProductVariationFormViewModel
```

or by just checking the `productionVariationID ` property you just added.

Would you mind changing it, please?

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -801,6 +821,11 @@ private extension ProductFormViewController {
         navigationItem.backButtonTitle = viewModel.productModel.name.isNotEmpty ? viewModel.productModel.name : Localization.unnamedProduct
     }
 
+    func updateNavigationBarTitle(title: String) {
+        // Update navigation bar title with variation ID
+        self.title = title
+    }
```
Ecarrion: I think this would be the correct place to handle the conditional code about what title to set, rather than in `viewDidLoad`. You have access to `viewModel` so you can easily do something like
```swift
guard let variationID = viewModel.productionVariationID else {
    return
}
       
title = Localization.variationViewTitle(variationID: "\(variationID)"))
```

What do you think?

Ecarrion (changes requested):
Hey @anas-p, I left you some suggestions and I'm particularly concerned about changing the product status internally 😓.

Also, I think you may have deleted a folder(classes?) in our code project. could you double-check?

Screen Shot 2021-11-16 at 7 58 06 AM

anas-p:
> Hey @anas-p, I left you some suggestions and I'm particularly concerned about changing the product status internally 😓.
> 
> Also, I think you may have deleted a folder(classes?) in our code project. could you double-check?
> 
> Screen Shot 2021-11-16 at 7 58 06 AM

Sorry @Ecarrion I will never make this mistake again and the folder `Model` was accidentally dragged out of the `Class` folder.

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
anas-p: You are right, that would be better.

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
anas-p: If this is 3, then `nameLabel` will be clipped after 3 lines, Since it's a name we may not need more than 3 lines. If it's okay to display more than 3 lines we can keep it 0.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
anas-p: Sure, I will change it to `viewModel is ProductVariationFormViewModel`.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
anas-p: You are right. Shall I change it this way?

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
Ecarrion: Since a variation name depends on how many attributes it has, I think it would be fairly normal for the name to spawn over 3 lines. So I think 0 makes sense here.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
Ecarrion: I think so, yes!

WooCommerce/Classes/ViewRelated/Products/Cells/ProductsTabProductTableViewCell.swift
anas-p: Sure, I will change it to 0.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
```diff
@@ -606,7 +609,15 @@ private extension ProductFormViewController {
...
         let productStatus = status ?? product.status
-        let messageType = viewModel.saveMessageType(for: productStatus)
+        var messageType = viewModel.saveMessageType(for: productStatus)
+
+        // Set messageType to `saveVariation`
+        // This will change Progress view title and message for variation save action.
+        // Fix for the issue: https://github.com/woocommerce/woocommerce-ios/issues/4847
+        if viewModel is ProductVariationFormViewModel {
+            messageType = .saveVariation
+        }
```
Ecarrion: [Q] Why not include this logic in the `viewModel.saveMessageType(for: productStatus)` method?
Also it would allow you to test that logic in `ProductFormViewModelTests`

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModelProtocol.swift
```diff
@@ -148,12 +153,15 @@ extension ProductFormViewModelProtocol {
...
     /// Returns `.publish` when the product is going to be published from a different status (eg: from draft).
+    /// Returns `.saveVariation` when saving a variation
     /// Returns `.save` for any other case.
     ///
     func saveMessageType(for productStatus: ProductStatus) -> SaveMessageType {
         switch productStatus {
         case .publish where !productModel.existsRemotely || originalProductModel.status != .publish:
             return .publish
+        case .custom("variation"):
+            return .saveVariation
```
Ecarrion: I don't think this is needed now, is it?

With the above suggestion, I think you can modify this method to something like:

```swift
default:
    if self is ProductVariationFormViewModel {
        return .saveVariation
    } else {
        return .save
    }
```

WooCommerce/Classes/ViewRelated/Products/View Models/ProductsTabProductViewModel.swift
```diff
@@ -19,22 +19,42 @@ private extension ProductStatus {
...
+    /// Product will only show product name
+    /// See more: https://github.com/woocommerce/woocommerce-ios/issues/4846
+    /// 
+    func createNameLabel() -> String {
+        var name = name
+        if let variationID = productVariation?.productVariationID {
+            // Add product variation ID with name
+            name = "\(Localization.variationID(variationID: "\(variationID)"))\n\(name)"
+        }
+        return name
```
Ecarrion: Since that name is local,  I don't think you gain anything mutating it. something less confusing would be:
```swift
if let variationID = productVariation?.productVariationID {
    // Add product variation ID with name
    return "\(Localization.variationID(variationID: "\(variationID)"))\n\(name)"
}
return name
```

right?

Ecarrion (changes requested):
Hey @anas-p, I have a few more suggestions.
Let me know what you think.

WooCommerce/Classes/ViewRelated/Products/View Models/ProductsTabProductViewModel.swift
anas-p: Yes, You are right.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewModelProtocol.swift
anas-p: Obviously, this is a better solution. I will implement this way.

WooCommerce/Classes/ViewRelated/Products/Edit Product/ProductFormViewController.swift
anas-p: There is no specific reason to write this code in `saveProduct` and like you mentioned   `viewModel.saveMessageType(for: productStatus)` is the right place to add this logic. The below solution will help to reduce this code also. I will go ahead with the below solution.

anas-p:
@Ecarrion I have made the changes, please take a look.

Ecarrion (approved):
Thanks for helping with these issues @anas-p!
It now looks good! I wish you would have added some unit tests as [suggested here](https://github.com/woocommerce/woocommerce-ios/pull/5394#discussion_r751662101).

anas-p:
> Thanks for helping with these issues @anas-p!
> It now looks good! I wish you would have added some unit tests as [suggested here](https://github.com/woocommerce/woocommerce-ios/pull/5394#discussion_r751662101).

I'll try to add unit tests. Thanks for all your suggestions it helps a lot.

Issue #5332, opened 2021-11-03
[Mobile Payments] Cancelling card reader update and disconnecting reader presents undismissable error.


joshheald:
I found this during the SDK2 branch PR review #5322 

## Repro

1. Navigate to `⚙️ > In-Person Payments > Manage Card Reader`
2. Wait for discovery of the reader tap `Connect Card Reader`
3. Tap `Update Reader Software`
4. Cancel the update
5. Tap `Disconnect Reader`

## Result

A blocking modal is presented over the `Connect your card reader` screen, telling the user that "We couldn't update your reader's software". The `Try Again` and `Cancel` buttons on this alert don't have any effect.

## Expectation

This alert shouldn't be presented after disconnecting the reader.

https://user-images.githubusercontent.com/2472348/140061547-ac0ffc1f-e3ec-4109-8d62-ddb3a2f84462.mp4

joshheald:
This was inadvertently partially fixed in a785415b614df52328e19112e10901296219c2ec

However, this fix is only down to [stopping treating cancellation as an error](https://github.com/woocommerce/woocommerce-ios/blob/3ea5660d5da2216406cba54f841a4afc30e67dd8/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift#L74). 

This means that other errors may still cause this behaviour, if the user disconnects the reader after getting the error.

Unfortunately, that means that it is much harder to reproduce, as the only error-producing simulated Stripe response is `.lowBattery`, which uses the mandatory update flow, avoiding the reader ever being connected.

I can now only reproduce the error by using a real card reader that needs an optional update, and attempting to do that while it has low battery. That scenario isn't quite so bad, as the error produced has a cancel button which does work, unlike the previous behaviour. I'm still going to look at fixing this, because other errors may still have broken buttons, and @jaclync [noticed a memory leak which might well be to blame for the issue](https://github.com/woocommerce/woocommerce-ios/pull/5397#issuecomment-967213131).

joshheald:
> ... and @jaclync [noticed a memory leak which might well be to blame for the issue](https://github.com/woocommerce/woocommerce-ios/pull/5397#issuecomment-967213131).

... unfortunately, that wasn't it. I'll keep looking.

joshheald:
Reproduction of this is increasingly difficult, since my physical reader's update went to mandatory and I had to install it. 

[...]

PR #5397, opened 2021-11-11
Remove Cancel on card reader updates at 100%

### LINKED ISSUE #5324: [Mobile Payments] [SDK 2] Software Updated, 100% Complete, Should Probably Not Show Cancel Button
See also pdi1mn-2k-p2

Found during testing of https://github.com/woocommerce/woocommerce-ios/pull/5322

When a reader software update completes, it sits on this screen for 10-15 seconds. I think we should consider NOT showing the Cancel button once the update is 100% complete, to avoid the merchant being tempted to intervene when the update is so close to being "done done."

IMG_0223

joshheald:
Closes #5324 

## Description

Card reader updates hover at 100% when they are almost complete, displaying 100% in the progress view, for up to 15s. We've taken the decision to remove the Cancel button once 100% is displayed, to avoid a merchant ending an update right before it finishes.

## Testing
1. Simulate an optional update using the instructions here: PdfdoF-fr-p2
2. Go to `⚙️ > In-Person Payments > Manage Card Reader`
3. Connect to the card reader simulator
4. Start the update
5. Observe that the cancel button is removed as soon as the progress indicator gets to 100%
6. Redo the update, and confirm that before 100% the cancel button functions as expected.

Update release notes:


WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift
```diff
@@ -127,9 +127,10 @@ private extension CardReaderSettingsConnectedViewController {
             }
         } else if let readerUpdateProgress = viewModel.readerUpdateProgress {
             // If we are updating a reader, show the progress alert
-            settingsAlerts.updateProgress(from: self, requiredUpdate: false, progress: readerUpdateProgress, cancel: { [weak self] in
-                self?.viewModel?.cancelCardReaderUpdate()
-            })
+            settingsAlerts.updateProgress(from: self,
+                                          requiredUpdate: false,
+                                          progress: readerUpdateProgress,
+                                          cancel: viewModel.cancelCardReaderUpdate)
```
jaclync, line 133: nit: it took me some time to realize why this change hides the cancel CTA (by whether `cancel` closure is nil or not), would it be more clear if there is another boolean from the view model like `isCancelButtonShown`/`showsCancelButton` based on the progress? This way, the cancel closure can stay the same and we just change the cancel button visibility boolean.

jaclync (approved):
LGTM! Code and testing ✅ (just a suggestion for your consideration)

Some non-blocking side notes:
- During testing, I have a habit of checking memory debugger on code changes with a closure that references self. I noticed that `CardReaderSettingsConnectedViewController` was still in the memory after leaving the payments screen, and saw the retain cycle at a place unrelated to this PR:
https://github.com/woocommerce/woocommerce-ios/blob/dbea5342d2b1604904d109d6feb860b799b685d1/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift#L191-L193
Passing weak self there deallocates the view controller after leaving the screen.
- Minor visual suggestion: there seems to be quite some space at the bottom when the cancel CTA is not shown, though this is for like a second so it might not be worth the effort.

joshheald:
> * During testing, I have a habit of checking memory debugger on code changes with a closure that references self. I noticed that `CardReaderSettingsConnectedViewController` was still in the memory after leaving the payments screen, and saw the retain cycle at a place unrelated to this PR:
>   https://github.com/woocommerce/woocommerce-ios/blob/dbea5342d2b1604904d109d6feb860b799b685d1/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewController.swift#L191-L193
>   
>   Passing weak self there deallocates the view controller after leaving the screen.

My instant suspicion is that you've just found the cause of #5332, which is next on my list! 🙇‍♂️ 


> * Minor visual suggestion: there seems to be quite some space at the bottom when the cancel CTA is not shown, though this is for like a second so it might not be worth the effort.

Yeah, I know what you mean, but I think it's best to leave it as is in this case since it's so short a time.

PR #5396, opened 2021-11-11
Simple Payments: Add order note edit functionality

Ecarrion:
closes #5345 

# Why

The "Edit Order Note" component was introduced in the "Order Editing" project but it was designed to automatically submit changes to a remote source. This PR refactors the `EditCustomerNote` view in order to allow consumers to provide a different view model.

In this PR I'm creating a `SimplePaymentsNoteViewModel` to handle the in-memory editing of the order note to be created.

# How

- Add `EditCustomerNoteViewModelProtocol` to allow for different implementations.

- Updated `EditCustomerNote` and `EditCustomerNoteHostingController` to receive a **generic** `ViewModel`. This is needed because `EditCustomerNoteViewModelProtocol` conforms to `ObservableObject` hence can't be used as a generic constraint. This also required some changes like not being able to create enums or conform to `@objc` protocols in extensions due to swift limitations. 🙃 

- Update `EditCustomerNote` to conform to `EditCustomerNoteViewModelProtocol`.

- Added `SimplePaymentsNoteViewModel` to handle the note editing in memory and without hitting any remote source.

- Updated `SimplePaymentsSummary` to launch the `OrderEditNoteView` with the `SimplePaymentsNoteViewModel`.

# Demo
https://user-images.githubusercontent.com/562080/141228180-b8dea333-5fbd-4cb7-a61c-6788990f95da.mov

# Testing Steps
- Make sure the normal order editing feature keeps working as expected.
- Start the simple payment flow, enter an amount to collect, and tap next.
- Tap on the "Add note" button.
- See the edit note view, and enter some text, and tap done.
- Play with it as you please. 😛 

PS: Dismissing the note view via a drag gesture does not revert the note edit. This is being tracked [here](https://github.com/woocommerce/woocommerce-ios/projects/40#card-72655207)

Update release notes:
- [x] I have considered if this change warrants user-facing release notes and have added them to `RELEASE-NOTES.txt` if necessary.

PR #5398, opened 2021-11-11
Order Editing: Fix iOS 15 navigation bar style

Ecarrion:
# Why 

@ealeksandrov [reported](https://github.com/woocommerce/woocommerce-ios/pull/5396#pullrequestreview-803746084) that on iOS 15, the navigation bar on the order editing flow didn't show any navigation bar button and was using large titles.

This PR fixes that by moving the `.navigationViewStyle(StackNavigationViewStyle())` modifier inside the `NavigationView` content 🤷‍♂️ 

# Screenshots

Before iOS 15 | After iOS 15 | After iOS 14
---- | ---- | ----
ios-15-bad |  ios-15-good | ios-14

Update release notes:


ealeksandrov (approved):
You've fixed it faster than I downloaded the iOS14 sim 😅 
Strange bug, but I confirm that solution works on iOS15! :shipit:

PR #5400, opened 2021-11-11
Order Filters: new Order List without tabs on top of the view

pmusolino:
As part of #5243 I remove tabs and I keep a single order list, that in the next PR will allow merchants to filter by order status and date range.
Since this PR remove also a lot of code that will not be more useful, this will be used as a base branch for the next PRs for completing the project of Order Filters, so DO NOT MERGE it until all the other tasks are completed. The plan is to release it in milestone 8.1.

## Changes
- Deleted `OrdersTabbedViewController` and its view model, since this is no more in use. 
- Moved some of the logic/UI of `OrdersTabbedViewController` in `OrdersRootViewController`. This is the root view controller when tapping the “Orders” tab. This view controller contains a stackview with the filters bar and the `OrderListViewController` embedded.

## Testing
- You should be able to load the orders.
- You should be able to open an order detail.
- You should be able to paginate the order list.
- You should be able to do a pull to refresh.
- You should be able to press the search bar button.
- You should be able to present the simple payment amount controller if the payment in person is enabled.
- You should be able to open an order detail from a push notification.

## Screenshot

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/OrdersRootViewController.swift
```diff
@@ -2,35 +2,31 @@ import UIKit
...
+            image: .waitingForCustomersImage
+        )
+    )
 
     // Used to trick the navigation bar for large title (ref: issue 3 in p91TBi-45c-p2).
     private let hiddenScrollView = UIScrollView()
 
     private let siteID: Int64
 
+    private lazy var analytics = ServiceLocator.analytics
```
jaclync: nit: could this be a `let`? it doesn't seem to be modified anywhere and the object already exists as a global singleton

WooCommerce/Classes/ViewRelated/Orders/OrdersRootViewController.swift
```diff
@@ -170,8 +219,86 @@ private extension OrdersRootViewController {
...
+            self.dismiss(animated: true) {
+                self.navigateToOrderDetail(order)
+            }
+        }
+
+        let viewController = SimplePaymentsAmountHostingController(viewModel: viewModel)
+        let navigationController = WooNavigationController(rootViewController: viewController)
+        present(navigationController, animated: true)
+
+        ServiceLocator.analytics.track(event: WooAnalyticsEvent.SimplePayments.simplePaymentsFlowStarted())
```
jaclync: nit: could this call the property?

```suggestion
        analytics.track(event: WooAnalyticsEvent.SimplePayments.simplePaymentsFlowStarted())
```

WooCommerce/Classes/ViewRelated/Orders/OrdersRootViewController.swift
```diff
@@ -170,8 +219,86 @@ private extension OrdersRootViewController {
...
+        let button = UIBarButtonItem(image: .searchBarButtonItemImage,
+                                     style: .plain,
+                                     target: self,
+                                     action: #selector(displaySearchOrders))
+        button.accessibilityTraits = .button
+        button.accessibilityLabel = NSLocalizedString("Search orders", comment: "Search Orders")
+        button.accessibilityHint = NSLocalizedString(
+            "Retrieves a list of orders that contain a given keyword.",
+            comment: "VoiceOver accessibility hint, informing the user the button can be used to search orders."
+        )
```
jaclync: nit: maybe these two strings can be `Localization` constants?

WooCommerce/Classes/ViewRelated/Orders/OrdersRootViewController.swift
```diff
@@ -170,8 +219,86 @@ private extension OrdersRootViewController {
...
+
+    /// Create a `UIBarButtonItem` to be used as a way to create a new simple payments order.
+    ///
+    func createAddSimplePaymentsOrderItem() -> UIBarButtonItem {
+        let button = UIBarButtonItem(image: .plusBarButtonItemImage,
+                                     style: .plain,
+                                     target: self,
+                                     action: #selector(presentSimplePaymentsAmountController))
+        button.accessibilityTraits = .button
+        button.accessibilityLabel = NSLocalizedString("Add simple payments order", comment: "Navigates to a screen to create a simple payments order")
```
jaclync: nit: similar to above, maybe this can be a `Localization` constant?

WooCommerce/Classes/ViewRelated/Orders/OrdersRootViewController.swift
```diff
@@ -114,37 +159,41 @@ private extension OrdersRootViewController {
...
+
+    func configureChildViewController() {
+        // Configure large title using the `hiddenScrollView` trick.
         if ServiceLocator.featureFlagService.isFeatureFlagEnabled(.largeTitles) {
             hiddenScrollView.configureForLargeTitleWorkaround()
             // Adds the "hidden" scroll view to the root of the UIViewController for large title workaround.
             view.addSubview(hiddenScrollView)
+            view.sendSubviewToBack(hiddenScrollView)
             hiddenScrollView.translatesAutoresizingMaskIntoConstraints = false
             view.pinSubviewToAllEdges(hiddenScrollView, insets: .zero)
```
jaclync, line 191: Is this large title workaround still needed now that there is only one scroll view in the orders tab?

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
```diff
@@ -142,6 +142,8 @@ final class OrderListViewController: UIViewController {
         // and the toggle can be switch in the Settings section that resides in a different tab.
         viewModel.reloadSimplePaymentsExperimentalFeatureState()
 
+        viewModel.syncOrderStatuses()
+
         syncingCoordinator.resynchronize(reason: SyncReason.viewWillAppear.rawValue)
```
jaclync, line 154: nit: would be nice to group `viewModel.syncOrderStatuses` and `syncingCoordinator.resynchronize` in a separate function so that we know these two calls are needed for each sync

jaclync (approved):
LGTM, yay for less code :shipit: Mostly a question about keeping the large title workaround!

There are some comments on `OrdersTabbedViewController` in the codebase, maybe we can update them later.

You should be able to load the orders ✅ 
You should be able to open an order detail ✅ 
You should be able to paginate the order list ✅ 
You should be able to do a pull to refresh ✅ 
You should be able to press the search bar button ✅ 
You should be able to present the simple payment amount controller if the payment in person is enabled ✅ 
You should be able to open an order detail from a push notification ✅ 

Observations from testing that are unrelated to this PR:
- It looks like the top banner height doesn't resize after rotation



- The empty state looks a bit different from [the design](https://user-images.githubusercontent.com/495617/137908672-316b4b2f-2936-4a3e-84a1-e2846b18ecd8.png) in https://github.com/woocommerce/woocommerce-ios/issues/5243

pmusolino:
Thank you @jaclync for the review

> The empty state looks a bit different from the design in #5243

I added it in my todo list in #5243, so I'll update it in a future PR.

WooCommerce/Classes/ViewRelated/Orders/OrdersRootViewController.swift
pmusolino: It is still needed since the OrderListViewController is used by both `OrdersRootViewController` (the view controller presented by the tab) and by the order search. This means that the Filters Bar is put inside the `OrdersRootViewController` and it's indipendent from the Order List.

WooCommerce/Classes/ViewRelated/Orders/OrderListViewController.swift
pmusolino: It makes sense. Added to my todo list 👍

Issue #5327, opened 2021-11-02
[Mobile Payments] [SDK 2] Low battery blocks software update but user is not asked to charge battery


allendav:
Found during testing of #5322

If you attempt to update a card reader with battery < 50% it fails without saying specifically it was due to the low battery:

![IMG_0219](https://user-images.githubusercontent.com/1595739/139919935-fa65235b-742a-47c5-ae25-9e9b9a99dfd1.PNG)

We fixed this for SDK 1. Just need to bring that back to life for SDK 2. See https://github.com/woocommerce/woocommerce-ios/pull/5133

allendav:
I haven't seen this happen (and I have a low battery reader) outside of the case fixed in #5416 - I'm going to close this and we can open a fresh issue if it happens again.

PR #5416, opened 2021-11-12
[Mobile Payments] Clear connected view model update error state on successfully connecting to a reader

allendav:
Partially addresses #5327 

This was a tricky one to reproduce. There may be other ways to precipitate 5327, and I will keep looking.

Background of the changes:

As of SDK2 an update may be kicked off while connecting to a reader. Because of this an update error can happen when attempting to connect to a reader if the reader needs a software update but has a low battery. This is different than SDK 1 where updates (and update errors) could not happen until AFTER connection completed.

The card reader settings views have two underlying view models - one for aggregating and formatting state useful for the searching view and one for aggregating and formatting state useful for the connected view. Both view models are contemporary and observe software update events.

Unfortunately, with the change in SDK2, a software update failure (due to low battery) can now occur (and be handled) outside of the connected view controller's purview. When this happens, the connected view model will receive notification of the update failure, but not the clearing that occurs under the purview of the searching view model.

The most appropriate way to remedy this would be to have both the searching and connected views rely on a single source of update state since SDK2 has blurred the lines. I've opened issue #5417 for that. In the mean time, a quick fix will be to clear any update error in the connected view model when the list of connected readers updates.

To test
- You'll need two readers. One requiring an update and with a low battery and another not requiring an update
- Power on just the low battery reader initially
- Attempt to connect to it
- Observe the update fails due to low battery
- Power on the other reader
- Keep searching and connect to the other reader when you see it
- Disconnect from the other reader
- Ensure you do NOT see a update failed due to low battery alert at this point

Update release notes:

### DIFF
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -51,7 +51,6 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
             self.didGetConnectedReaders = true
             self.connectedReaders = readers
             self.updateProperties()
-
             self.reevaluateShouldShow()
         }
         ServiceLocator.stores.dispatch(action)
--- a/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
+++ b/WooCommerce/Classes/ViewRelated/Dashboard/Settings/CardReadersV2/CardReaderSettingsConnectedViewModel.swift
@@ -47,9 +47,11 @@ final class CardReaderSettingsConnectedViewModel: CardReaderSettingsPresentedVie
             guard let self = self else {
                 return
             }
+            self.readerUpdateError = nil
             self.didGetConnectedReaders = true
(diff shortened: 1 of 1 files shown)


joshheald (approved):
I was able to reproduce this with simulated readers by doing the following:

First, add this to `StripeCardReaderService.start(_:)`, to simulate the lowBattery scenario of the first connected reader.

        Terminal.shared.simulatorConfiguration.availableReaderUpdate = .lowBattery

Also add this to `CardReaderSettingsConnectedViewModel.completeCardReaderUpdate(success:)`, so that when the `.lowBattery` update finishes, the next reader gets an Optional update available (which does not trigger on start, `.none` would work here too.

        if Terminal.shared.simulatorConfiguration.availableReaderUpdate == .lowBattery {
            Terminal.shared.simulatorConfiguration.availableReaderUpdate = .available
        } else {
            Terminal.shared.simulatorConfiguration.availableReaderUpdate = .lowBattery
        }

Build and run, connecting to the Chipper simulator first, then wait for the `.lowBattery` failure. After that, tap `Cancel` to disconnect and then connect to the M2 simulator, and you would see the low battery error before this fix. With this fix, connection continues as expected.

It's difficult to test other errors, so I'm not sure what the implications could be for those. It could be argued that this is enough of an edge case, which is user-recoverable, that we don't need a workaround type fix at all, and should instead do the full fix when we have time... but fewer bugs is a good thing so as long as you have confidence about nilling the error out here, I'd go for merging.

Issue #5364, opened 2021-11-05
Jetpack CP: detect a JCP site in `me/sites` calls (workaround: extra API requests to fetch site metadata and for JCP sites to show up in site picker)


jaclync:
Refs:
- P2 that explains JCP workarounds p91TBi-6lK-p2
- Android issue https://github.com/woocommerce/woocommerce-android/issues/5137
- Android FluxC PR https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2169

(Adapted from p2 post above)

## Why

This workaround is for usage of `/me/sites` and `/sites/{site_id}` endpoints, so that JCP sites show up in the site picker and the app can fetch correct info about JCP sites.

The app currently uses the endpoint `/me/sites` during the login and site picker to fetch sites of the current user. It also keeps fetching sites on each app launch to keep the list of current sites up-to-date, and to confirm that the site is still connected.

And the endpoint `/sites/{site_id}` is used in two scenarios:

- During the login step in pretty specific circumstances: login with site URL, and when we reach the site picker, we find out that the site doesn’t have WooCommerce
- On each cold start of the app to keep an updated set of the site’s metadata.

Both those endpoints are currently broken for Jetpack Connection Package sites, `/me/sites` has data that was sent during the initial connection, so it might be outdated, and `/sites/{site_id}` doesn’t work at all.

**This issue focuses on the changes for `/me/sites` calls (on app launch and when showing the site picker).**

## High-level tasks

- When fetching the full list of sites, the app will hit the endpoint `/me/sites` first, then for each site that doesn’t have Jetpack-the-plugin, it will use two additional requests to the remote site: `/wc/v3/settings?_fields=""` to check if WooCommerce is installed, and `/wp/v2/settings` to get the updated site’s metadata, and then update the app’s local database accordingly
- When fetching a single site, if the site doesn’t have jetpack-the-plugin, it will use just the last two endpoints `/wc/v3/settings` and `/wp/v2/settings` to update the app’s local database

## Subtasks

- [x] Storage: schema update - add 2 new attributes `isJetpackConnected: Bool` and `isJetpackThePluginInstalled` to `Site` https://github.com/woocommerce/woocommerce-ios/pull/5418
- [x] Netwroking: schema update - add 2 new attributes `isJetpackConnected: Bool` and `isJetpackThePluginInstalled` to `Site` https://github.com/woocommerce/woocommerce-ios/pull/5420
[...]

PR #5418, opened 2021-11-15
Jetpack CP site detection: add `isJetpackConnected` and `isJetpackThePluginInstalled` attributes to `Storage.Site`

jaclync:
Part of #5364 

- [x] ⚠️ Only merge after release 8.0 has been cut ⚠️ 

Notes: the PR is large mostly from the new model version, which has 744 LOC!

## Why

In order to know if a site is connected to Jetpack via Jetpack Connection Package instead of Jetpack-the-plugin, we are fetching 2 values for each site in the `me/sites` endpoint:

- `isJetpackThePluginInstalled: Bool`: this corresponds to the `jetpack` field in the API response. It indicates whether Jetpack-the-plugin is installed on the site
- `isJetpackConnected: Bool`: this corresponds to the `jetpack_connection` field in the API response. It indicates whether the site is connected to Jetpack, either through Jetpack-the-plugin or other plugins that include Jetpack Connection Package

Then we can know whether a site is a JCP site by the derived value `isJetpackConnected && !isJetpackThePluginInstalled`. I thought about creating just one attribute for this calculated property, but it might be worth matching the API response in case we find one of these two Jetpack values useful in the future. Android also [persists both fields in FluxC](https://github.com/wordpress-mobile/WordPress-FluxC-Android/commit/138884276f2f2cbf48f9f347e1521e49f8c3e0e3#diff-ac504807ac3755802bc839947ec85b4b045a906264eaf35ab97885016704df8a).

This PR includes changes in the storage layer only for the 2 new attributes, to make it easier for code review.

## Changes

- Created a new model version https://github.com/woocommerce/woocommerce-ios/commit/b300b492aba34e4de4d724656c3e9e95f9b80039
- Added 2 attributes to `Site` https://github.com/woocommerce/woocommerce-ios/commit/93ebc12c90acb26a0bb09c7d1849f6f6d1508914
- Set current version to 58 https://github.com/woocommerce/woocommerce-ios/commit/e7972cdd4b16c59b76c0ae8bcd291311d393a0ee

## Testing

Feel free to launch the app that was last built from `develop` or any other branch and make sure the app does not crash from the new model version!

Update release notes:

### DIFF
--- a/Storage/Storage.xcodeproj/project.pbxproj
+++ b/Storage/Storage.xcodeproj/project.pbxproj
@@ -1752,7 +1752,7 @@
 				746A9D14214071F90013F6FF /* Model 2.xcdatamodel */,
 				B59E11D920A9D00C004121A4 /* Model.xcdatamodel */,
 			);
-			currentVersion = 313FC0B62732EE520067408D /* Model 57.xcdatamodel */;
+			currentVersion = 028A81022740E1B400F04D15 /* Model 58.xcdatamodel */;
 			path = WooCommerce.xcdatamodeld;
 			sourceTree = "<group>";
 			versionGroupType = wrapper.xcdatamodel;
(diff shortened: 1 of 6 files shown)


itsmeichigo (approved):
✅ Changes look good
✅ Builds and runs without error. 

:shipit:

PR #5419, opened 2021-11-15
[Mobile Payments] Remove html tags from receipt generation

### LINKED ISSUE #4137: [Mobile Payments] HTML entities in content of receipt
https://github.com/woocommerce/woocommerce-ios/pull/4128#discussion_r627730386

kidinov:
Closes #4137

### The PR removes HTML symbols from:
* Receipt item title
* Application preferred name
* Dedicated file name

## To test:
* Modify the code so one or all of the fields mentioned above have HTML tags (check `ReceiptRenderer`)
* Make a payment or open stored receipt
* Notice that the modification doesn't affect the generated receipt

@ctarda please try to point on anything you see that doesn't look good, even if it's nitpicks - should be good for educational purposes =)


Hardware/HardwareTests/AirPrintReceipt/ReceiptRendererTest.swift
```diff
@@ -0,0 +1,68 @@
+import XCTest
+@testable import Hardware
+import StripeTerminal
+import Foundation
+import CryptoKit
+
+final class ReceiptRendererTest: XCTestCase {
+    func test_TextWithoutHtmlSymbols() {
+        let expectedResultWithoutHtmlSymbolsMd5Description = "MD5 digest: 0a08fa3ac67c8a06fa460834ca876a7a"
```
kidinov: The approach with MD5 is questionable because we won't see what's broken in the final HTML if the test fails. On other hand, the HTML is so big that most probably you still won't see it either without using diff tools. Having  2 huge HTML stored in the file doesn't look neat as well, wdyt?

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -250,3 +251,19 @@ private extension ReceiptRenderer {
...
+    func stripHtml() -> String {
+        let data = Data(utf8)
+        do {
+            return try NSAttributedString(
+                data: data,
+                options: [.documentType: NSAttributedString.DocumentType.html],
+                documentAttributes: nil
+            ).string
+        } catch {
+            print(error)
```
ctarda: We tend to leave prints out of production code. DDLog might be the best approach here (if DDLog is available to Hardware). If not, I would just remove the print and move on.

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
```diff
@@ -250,3 +251,19 @@ private extension ReceiptRenderer {
         )
     }
 }
+
+private extension String {
+    func stripHtml() -> String {
```
ctarda: This function returns a modified copy of the original string, without modifying the original. In those cases, the convention in Swift is to name the method/property ending in -ed (participle tense?).

To me, `htmlStripped()` would sound "Swiftyer"

ctarda (approved):
Works perfect! I left a suggestion to rename a method, but it is not a blocker.

:shipit: 

👏🏼 👏🏼 👏🏼

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
kidinov: Got it 👍

Fixed: 65885a240a103e3917d187edb2345ef6acdf031d

Hardware/Hardware/Printer/AirPrintReceipt/ReceiptRenderer.swift
kidinov: Makes a lot of sense. Not sure why we don't call them like that in kotlin

Fixed: 7c2f860b17eddc315b5b25356605532c2b18e119

PR #5420, opened 2021-11-15
Jetpack CP site detection: add `isJetpackConnected` and `isJetpackThePluginInstalled` attributes to `Networking.Site`

jaclync:
Part of #5364 

## Why

In order to know if a site is connected to Jetpack via Jetpack Connection Package instead of Jetpack-the-plugin, we are fetching 2 values for each site in the `me/sites` endpoint:

- `isJetpackThePluginInstalled: Bool`: this corresponds to the `jetpack` field in the API response. It indicates whether Jetpack-the-plugin is installed on the site
- `isJetpackConnected: Bool`: this corresponds to the `jetpack_connection` field in the API response. It indicates whether the site is connected to Jetpack, either through Jetpack-the-plugin or other plugins that include Jetpack Connection Package

Then we can know whether a site is a JCP site by the derived value `isJetpackConnected && !isJetpackThePluginInstalled`. As a workaround, there are 2 more API requests to make for each JCP site to get the updated info about whether WooCommerce is active and also the site metadata.

This PR includes changes in the networking layer only for the 2 new attributes, to make it easier for code review. The workaround to make 2 extra API requests and integration with the storage layer will be in separate PRs.

## Changes

- Added `isJetpackThePluginInstalled` and `isJetpackConnected` booleans to `Networking.Site` that correspond to `jetpack` and `jetpack_connection` fields in the API
- Ran `rake generate` for Fakeable and Copiable support
- Updated unit tests: added `jetpack_connection` field to the existing `me/sites` API responses

## Testing

Please test for regression in the site picker (this is where we fetch the sites from `me/sites` endpoint):

- Log out from the app
- Log in to the app --> the site list in the site picker should be the same as before the PR branch
- Go to Settings > Switch Store --> the site list in the site picker should be the same as before

Update release notes:

### DIFF
--- a/Fakes/Fakes/Networking.generated.swift
+++ b/Fakes/Fakes/Networking.generated.swift
@@ -1314,6 +1314,8 @@ extension Site {
             description: .fake(),
             url: .fake(),
             plan: .fake(),
+            isJetpackThePluginInstalled: .fake(),
+            isJetpackConnected: .fake(),
             isWooCommerceActive: .fake(),
             isWordPressStore: .fake(),
             timezone: .fake(),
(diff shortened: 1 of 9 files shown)


itsmeichigo (approved):
LGTM :shipit:

PR #5421, opened 2021-11-15
Cancel button and swipe to cancel actions handled

### LINKED ISSUE #5302: Quick Order: Can tap "Cancel" to cancel quick order after tapping "Done," but order is still created
**Describe the bug**

After I tap "Done" on the Quick Order screen, the "Cancel" button (or drag to dismiss action) are still available. If I cancel after tapping "Done," though, the order is still created.

**To Reproduce**

1. On a store that's onboarded for IPP, enable Quick Order in Settings > Experimental Features.
2. Go to Orders and tap the + button.
3. Enter an amount and tap "Done."
4. Immediately tap the "Cancel" button or swipe to dismiss the quick order screen.
5. Notice that the quick order seems to be canceled, but the order is created and appears in the Orders list anyway.

**Screenshots**

https://user-images.githubusercontent.com/8658164/139257580-cb3fd04e-20e4-450b-bacb-a40668a28bf7.mp4

anas-p:
Closes #5302

## Why
Quick Order page can be dismissed using "Cancel" button or swipe down to dismiss action even after tapping "Done" button . Even If it is cancelled or dismissed after tapping "Done" the order is still created.

## How
- Created a `disableCancel` variable in `SimplePaymentsAmountViewModel` to get the status of create order operation (`createSimplePaymentsOrder`).
- Cancel button set to enable/disable state based on `disableCancel` status. (Disabled until the user get the result from `createSimplePaymentsOrder`).
- Disables swipe down to dismiss action when user tapped on "Done" button until user gets the payment order response.

## Demo
https://user-images.githubusercontent.com/19687299/141728049-090c56a8-6243-4999-bbe0-b1c50ce8621c.mov

## Testing Steps
1. Enable Quick Order in Settings > Experimental Features.
2. Go to Orders and tap the + button.
3. Enter an amount and tap "Done"
4. Try to dismiss the page using swipe to dismiss action
5. "Cancel" button set to disabled state until you get the result from `createSimplePaymentsOrder`

Update release notes:


WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmount.swift
```diff
@@ -48,6 +48,9 @@ final class SimplePaymentsAmountHostingController: UIHostingController<SimplePay
     override func viewDidLoad() {
         super.viewDidLoad()
 
+        // Disables interactive dismiss action.
+        isModalInPresentation = true
```
Ecarrion: Neat, I have never this property before! However, the idea is to only prevent the dismissal of the view while the network operation is happening, we still should be able to dismiss the view via drag gestures if the user hasn't tapped on the "done" button yet.

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmount.swift
```diff
@@ -126,10 +129,12 @@ struct SimplePaymentsAmount: View {
         .navigationTitle(Localization.title)
         .toolbar {
             ToolbarItem(placement: .cancellationAction) {
+                // Cancel button set to disabled state until get the result from `createSimplePaymentsOrder`
                 Button(Localization.cancelTitle, action: {
                     dismiss()
                     viewModel.userDidCancelFlow()
                 })
+                    .disabled(viewModel.loading)
```
Ecarrion: Ideally, we don't want the view to make assumptions about the view model state. Here we are assuming that we want the cancel button disabled when the view model is loading, and while that is the intended behavior a much more clear approach is to expose a `disableCancelButton` property on the view model that the view can consume without any doubt or assumption. Also, that allows us to unit test that new logic.

Does that make sense? What do you think?

Ecarrion (changes requested):

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmount.swift
anas-p: You are right, I will remember to write logics in `viewModel`s rather than in the view from now on. Thanks for pointing out.

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmount.swift
anas-p: I will check the possibility.

anas-p: @Ecarrion  I handled in this way: Now do swipe to dismiss if the user hasn't tapped on the "Done" button, otherwise can't.
```
 viewModel.$loading
             .sink { [weak self] loading in
                 // here handling interactive dismissal
                 self?.isModalInPresentation = loading
             }
             .store(in: &subscriptions)
```

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmount.swift
anas-p: Added `disableCancelButton` property in `ViewModel`, that is using in `SimplePaymentsAmount` view to enable/disable cancel button.

```
Button(Localization.cancelTitle, action: {
                    dismiss()
                    viewModel.userDidCancelFlow()
                })
                    .disabled(viewModel.disableCancelButton())
```

anas-p: Unit test added for `disableCancelButton`.

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmount.swift
```diff
@@ -43,6 +43,14 @@ final class SimplePaymentsAmountHostingController: UIHostingController<SimplePay
...
             }
             .store(in: &subscriptions)
+
+        // Observes the loading intent it will be true while performing the create order operation. False otherwise.
+        viewModel.$loading
+            .sink { [weak self] loading in
+                // Disables interactive dismissal while performing the create order operation, Enables otherwise.
+                self?.isModalInPresentation = loading
+            }
+            .store(in: &subscriptions)
```
Ecarrion: This looks good, another way to achieve this, since we already have a presentation controller delegate is to prevent the view dismissal like this.

```swift
func presentationControllerShouldDismiss(_ presentationController: UIPresentationController) -> Bool {
     !rootView.viewModel.loading
}
```

or even better
```swift
func presentationControllerShouldDismiss(_ presentationController: UIPresentationController) -> Bool {
    !rootView.viewModel.disableCancel
}
```

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmountViewModel.swift
```diff
@@ -107,6 +108,12 @@ final class SimplePaymentsAmountViewModel: ObservableObject {
         analytics.track(event: WooAnalyticsEvent.SimplePayments.simplePaymentsFlowCanceled())
     }
 
+    /// Disable cancel button while performing the create order operation
+    ///
+    func disableCancelButton() -> Bool {
+        return loading
+    }
```
Ecarrion: Question: Why did you decide to make this a function instead of a computed variable?

Also if you rename this as `var disableCancel: Bool { loading }` you can use it in the hosting view controller too.

Ecarrion (commented):
Hi @anas-p, this looks so much better, I left you some small suggestions to keep improving this!

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmount.swift
anas-p: Thank you for pointing out. I have change it to presentation controller delegate method.

WooCommerce/Classes/ViewRelated/Orders/Simple Payments/Amount/SimplePaymentsAmountViewModel.swift
anas-p: Yes, a func definition would be an overkill, it can be implemented as a variable declaration as you have suggested.
And I would rename "disableCancelButton" to "disableCancel".

Ecarrion (approved):
This looks good @anas-p! Good job!

